PR #2473, opened 2019-02-13
quicksetup: move check for existing profile

ltalirz:
the check whether a profile of the same name already exists was
happening too late (after finding the database of the same name etc.).
This should be the first thing to check.


ltalirz:
Small PR, ready to be merged

sphuber:
Thanks @ltalirz I am merging another PR first, then I will update this and merge it

PR #2478, opened 2019-02-14
Export file migration: v0.3 to v0.4 to v0.5

### LINKED ISSUE #2426: Implement migration of export archives from v0.3 to v0.4 to v0.5
Issue #2342 is responsible for making sure that the export/import functionality works properly after the changes of the provenance redesign and the export-import cycle is self-consistent. After that is guaranteed, migration functionality has to be implemented to update existing export archives with version `v0.3` to the new version `v0.4`. This will potentially require an export schema update and definitely a data migration.

### LINKED ISSUE #3010: Check new repo "aiida-export-migration-tests"
In connection with PR #2478 a new repo has been made to test migration of export files.

Together with the review of the PR, this new repo, found [here](https://github.com/aiidateam/aiida-export-migration-tests) should be "reviewed".

yakutovicha:
Fixes #2426.
Fixes #3010.


yakutovicha:
Note:
The DB migration 0009 does 'data.base.Bool.'  -> 'data.bool.Bool.' (without 'node.' prefix)
probably the node prefix was added in some other migration

In this commit I've implemented migration WITH 'node.' prefix

sphuber:
> The DB migration 0009 does 'data.base.Bool.' -> 'data.bool.Bool.' (without 'node.' prefix)
> probably the node prefix was added in some other migration

Yes, this was added later [in another migration](https://github.com/aiidateam/aiida_core/blob/provenance_redesign/aiida/backends/djsite/db/migrations/0025_move_data_within_node_module.py)

yakutovicha:
@CasperWA and @szoupanos I added both of you to my [aiida_core fork](https://github.com/yakutovicha/aiida_core/tree/export_migration), so you can also contribute.

I will sequentially go through all the migrations.

yakutovicha:
The migration 0010 does not need to be implemented, since the 'process_type' was never used before the migration 0020. In the migration 0020, however, the 'process_type' string will contain some information that was previously stored in 'type' (plugin entry point, I assume) for the calculation nodes.

Therefore, the type -> process_type conversion will be implemented together with the migration 0020.

yakutovicha:
The migration 0011 does not need to be implemented, as all the affected tables (kombu_message, kombu_queue, db_dbsetting) are not exported.

CasperWA:
> The migration 0010 does not need to be implemented, since the 'process_type' was never used before the migration 0020. In the migration 0020, however, the 'process_type' string will contain some information that was previously stored in 'type' (plugin entry point, I assume) for the calculation nodes.
> 
> Therefore, the type -> process_type conversion will be implemented together with the migration 0020.

For the conversion of the content in `type` to `process_type` see the [`infer_calculation_entry_point`-function](https://github.com/aiidateam/aiida_core/blob/b1b34f69da1bd43e1e6d97cee707b43c1e75c15c/aiida/manage/database/integrity/plugins.py#L77)

yakutovicha:
The migration 0012 deletes db_dblock table from the database that was not exported anyways. So skipping this migration as well.

yakutovicha:
The migration 0013 changes content type in two columns of `db_dbuser` table: `last_login` (to `DateTimeField`) and `email` (to `EmailField`).

- `last_login' column is not exported anyways, so will be skipped. 
- `email` column is exported as a string for before and after migration database. So no changes are required also.

yakutovicha:
The migration 0014 checks the uniqueness of the nodes' uuids. Based on the import file it is impossible to know what exactly should be done with those nodes, therefore the migration process will be aborted.

yakutovicha:
The migration 0015 invalidates node hashes stored in extras. Since extras are not exported in v0.3 - we skip this migration step.

yakutovicha:
The migration 0016 acts on the Code class used to be just a sub class of Node but was changed to act like a Data node.
Basically, the only change was to replace type 'code.Code.' with 'data.code.Code.'

yakutovicha:
The migration 0017 drops the DbCalcState table which is not exported 'as is' but as a 'state' item of a node_attributes dictionary.

I assume (@sphuber correct me if I am wrong) that the state became a node attribute, and therefore it is exported in exactly the same way as before the migration 0017.

So this migration does not need to be implemented

yakutovicha:
The migration 0018 imposes `uuid` uniqueness constraints to all tables that contain `uuid` column. 

@giovannipizzi, this is the list of such a tables in AiiDA 0.12.2:
```
=> select table_name from information_schema.columns where column_name = 'uuid';
  table_name   
---------------
 db_dbcomment
 db_dbcomputer
 db_dbgroup
 db_dbnode
 db_dbworkflow
(5 rows)
```
Among those columns only three (`db_dbcomputer`, `db_dbgroup`,  `db_dbnode`) were exported in AiiDA 0.12.2. So I added the uniqueness check only to those three dictionaries in the export file.

giovannipizzi:
Perfect

yakutovicha:
Migration 0020:

- remove namespace ‘simpleplugins’ from ‘templatereplacer’ and ‘addcalculation’ in ‘db_dbnode’ table’s ‘process_type’ and ‘type’ fields.

- remove 'simpleplugins' prefix from the 'input_plugin' attribute of 'arithmetic.add' and 'templatereplacer' codes

CasperWA:
Try, going forward, to rebase on provenance_redesign instead of merging

yakutovicha:
As the migration 0020 is quite large, I won't describe here what it is doing. One could just have a look inside the [corresponding file](https://github.com/aiidateam/aiida_core/blob/provenance_redesign/aiida/backends/djsite/db/migrations/0020_provenance_redesign.py) to get more information.

The thing that I did NOT implement is detection of "workflow nodes with isolated CREATE links" (third check [here](https://github.com/aiidateam/aiida_core/blob/provenance_redesign/aiida/backends/general/migrations/provenance_redesign.py)). All the rest should be working.

yakutovicha:
> Try, going forward, to rebase on provenance_redesign instead of merging

I would do debasing at the end, before requesting a review from Sebastiaan.

sphuber:
Yes, but for PRs that involve multiple authors, rebasing at the end becomes extremely painful if there have been intermediate merges, because in that case I cannot squash everything, because that would lose the attributions for the authors involved. So rule-of-thumb: for PRs with multiple authors that will need to keep individual attributions, only use rebasing during development to get branch up to date.

yakutovicha:
> Yes, but for PRs that involve multiple authors, rebasing at the end becomes extremely painful if there have been intermediate merges, because in that case I cannot squash everything, because that would lose the attributions for the authors involved. So rule-of-thumb: for PRs with multiple authors that will need to keep individual attributions, only use rebasing during development to get branch up to date.

ok

yakutovicha:
Migration 0021 renames the following columns of group table:
name -> label
type -> type_string

yakutovicha:
Migration 0022 modifies the group type_strings as follows:

'' -> 'user'
'data.upf.family' -> 'data.upf'
'aiida.import' -> 'auto.import'
'autogroup.run' -> 'auto.run'

yakutovicha:
@sphuber , could you please have a look at my last commit where I implemented the migration 0023? I am not sure I understood what exactly should be done here. 

@CasperWA , @szoupanos I think I stop here. Please continue further. The biggest migration among the remaining once is done by @szoupanos (0024) so I assume it would be much easier for him to work on it.

Let me know if I should already rebase the existing commits to simplify your life.

yakutovicha:
In the meanwhile I realized that the migration 0024 can be skipped entirely as it focuses on the migration of logs.

yakutovicha:
Migration 0025 changes the type string for `Data` nodes from `data.*` to `node.data.*

yakutovicha:
Migration 0026 moves symbols array of TrajectoryData object into attributes. Migration 0027 removes 'array|symbols' column from the attributes. Both are implemented in the last commit

yakutovicha:
Migration 0028 changes the type strings as folllows:
 * `node.data.` -> `data.`
 * `node.process.` -> `process.`

yakutovicha:
So I finished all the migrations and rebased/squashed the commits. 

Would @CasperWA or @szoupanos volunteer to have a look and/or add some tests? Please let me know.

szoupanos:
Thanks a lot Sasha for all this work! I will look at it (and I will try to contribute) but I will not have time today.

CasperWA:
Great work @yakutovicha and @szoupanos!

I have just rebased on the newest provenance_redesign and will slowly start to go through it.

yakutovicha:
@CasperWA I think you should have not squashed my commit (with migrations 0018 - 0028) to the one of @szoupanos. Because this transferred the ownership and responsibility for that part of work from me to Spyros.

sphuber:
@yakutovicha don't worry, I will correct the attribution

sphuber:
Attribution should be fixed now

yakutovicha:
> Attribution should be fixed now

thanks a lot @sphuber

CasperWA:
I am currently going through this work, trying to add the last migrations and make sure all tests succeed.

@sphuber, I have come across something strange. There are node attributes listed in the exported file `export_v0.1.aiida` for some USP files, however, there are no corresponding Nodes or Groups. Was this the norm for this export version, or has the file been manipulated/is it incomplete?

CasperWA:
Migration 0029 changes the type of `ParameterData` as follows:

- `data.parameter.ParameterData.` -> `data.dict.Dict.`

CasperWA:
Migration 0030 changes the name of the `Node` column `type` as follows:

- `type` -> `node_type`

The content/value is left unchanged.

CasperWA:
All changes that affect names of columns are also changed in `metadata.json` as well as in `data.json`, fx. Rev. 30.

CasperWA:
`Log` and `Comment` entities have been added to `metadata.json`'s `all_fields_info` and `unique_identifiers` dicts.

CasperWA:
`node_extras` and `node_extras_conversion` dicts have been added to `data.json`.

Since extras were not exportable previously, the dicts will be left empty.

@yakutovicha @sphuber @szoupanos since most `Node`s have the hash saved in extras, is it reasonable to add this always as an extra to `Node`s, irregardless of type? The value may be created afterwards or left empty?

CasperWA:
Added migrations for 0031, 0032, and 0033:

0031: Remove 'enabled' key from 'Computer' in 'export_data'.
0032: No effect on exported files, since legacy workflows were never exportable (according to @AntimoMarrazzo).
0033: Convert strings of dicts to actual dicts. (For 'metadata' and 'transport_params' keys in 'Computer' and 'Log')

CasperWA:
Restructured migrations functions, such that it is now available under `aiida.cmdline.utils.migration` and the new test file for export file migrations have been renamed and put under `aiida.backends.tests.cmdline.utils.migration`.

The restructuring has been done in a separate commit to more easily be able to see the original authors of the code before the restructuring.

CasperWA:
Missing work before this can be merged:

- [ ] Create export files that are representational of what could/can be exported for all export versions.
  - [ ] v0.1
  - [ ] v0.2
  - [ ] v0.3
  - [ ] v0.4
- [ ] Write tests for each migration function.
  - [ ] v01_to_v02
  - [ ] v02_to_v03
  - [ ] v03_to_v04
  - [ ] Make sure data is not/never lost when migrating, for as many use cases as possible.

CasperWA:
Export migration function tests along with their representative export files have been split away from aiida-core and can now be found in the repo [aiida-export_migration-testing](https://github.com/aiidateam/aiida-export_migration-testing).

This repo will be installed (v0.1.0), when doing `pip install aiida-core[testing]`, according to the current PR.

It should however be noted that the external repo has AiiDA v1.0.0 as a requirement.

CasperWA:
The only(TM) missing thing to do in this PR is to make sure the new JSONB changes do not introduce any incompatibilities/bugs with the export file migration.

ltalirz:
@CasperWA what are the remaining steps necessary for this to be merged (perhaps with links to the respective PRs?)

Would be great to have since we need to migrate export files for the tutorial.

yakutovicha:
I think we are waiting for https://github.com/aiidateam/aiida_core/projects/11 to be finalized

yakutovicha:
Another idea would be to merge it now, and add an issue about possible modifications after JSONB work is done.

sphuber:
> Another idea would be to merge it now, and add an issue about possible modifications after JSONB work is done.

The downside of this would be that *if* the JSONB work will require changes, we might be forced to change the export version again, with migration. Let's see what the timeline of the JSONB work is to see if this would be merged in time.

giovannipizzi:
I didn't check the code, but I believe JSONB shouldn't *require* a change to the export format.

However I just added #2815, where a format change would be beneficial for a number of reasons (smaller file size, faster import/export, simpler code, ...).

Still, my vote is to merge this now if it's ready, and add an issue in the [JSONB project]( https://github.com/aiidateam/aiida_core/projects/11) as this PR is needed now for the tutorial and already brings in a lot of benefits. We will then, if needed, create a 0.5 version.

CasperWA:
> I didn't check the code, but I believe JSONB shouldn't _require_ a change to the export format.

True, however, I prefer to update the entries in metadata, etc., to actual dicts instead of strings. Which is the way it is stored now with a Django backend. This has already been updated for other fields and results in a more backend-agnostic export file - even though it wouldn't affect the actual import.

> However I just added #2815, where a format change would be beneficial for a number of reasons (smaller file size, faster import/export, simpler code, ...).

This is very good. We need to make sure that it is only datetime objects that `node_attributes_conversion` is used for, but otherwise great. This will also remove the newly created `node_extras_conversion` dict from the export files.

> Still, my vote is to merge this now if it's ready, and add an issue in the [JSONB project](https://github.com/aiidateam/aiida_core/projects/11) as this PR is needed now for the tutorial and already brings in a lot of benefits. We will then, if needed, create a 0.5 version.

I agree. And it _is_ ready. The final reviewer should definitely go through the newly created repository [here](https://github.com/aiidateam/aiida-export-migration-tests) as well, since the code there is essentially part of this PR as well (it contains all the tests).

In the end, all this new work should make it "easier" to add a new export version, making it possible to merge this now and later update it.

**Note**: I have left a comment in the `migrate_v03_to_v04` function that alludes to the new JSONB changes, but no code exists yet. I will leave it up to the reviewer whether it should stay in or be included later, if necessary.

ltalirz:
> The final reviewer

So, we just need to figure out who this is.
I volunteer to have a quick look over the code changes in aiida-core, which seem quite manageable in the end.

I think there needs to be a review by second person, however, who is more familiar with how the export works.

aiida/backends/tests/__init__.py
```diff
@@ -101,6 +101,10 @@
         'engine.work_chain': ['aiida.backends.tests.engine.test_work_chain'],
         'engine.workfunctions': ['aiida.backends.tests.engine.test_workfunctions'],
         'export_and_import': ['aiida.backends.tests.test_export_and_import'],
+        'export_migration.migration': ['aiida-export-migration-tests.tests.test_migration'],
```
ltalirz: can you replace `export_migration.migration` with something that makes it more clear what is being tested here (vs the v01_to_v02, ... tests)?

aiida/backends/tests/cmdline/commands/test_calcjob.py
```diff
@@ -96,14 +91,13 @@ def setUpClass(cls, *args, **kwargs):
         calc.set_process_state(ProcessState.FINISHED)
         cls.calcs.append(calc)
 
-        # Uncomment when issue 2426 is addressed
-        # # Load the fixture containing a single ArithmeticAddCalculation node
-        # import_archive_fixture('calcjob/arithmetic.add.aiida')
+        # Load the fixture containing a single ArithmeticAddCalculation node
+        import_archive_fixture('calcjob/arithmetic.add.aiida')
```
ltalirz: just a question - in which sense is this a fixture? @yakutovicha was also confused by this.
In my understanding, the term `fixture` in python is mainly used by `pytest`, where it has a [very specific meaning](https://docs.pytest.org/en/latest/fixture.html).

Can we remove the `_fixture` and also move the `import_archive` simply to something like `aiida.backends.tests.utils.archive`?

aiida/backends/tests/cmdline/commands/test_export.py
```diff
@@ -155,18 +141,17 @@ def test_create_tar_gz(self):
...
             delete_temporary_file(filename)
 
-    @unittest.skip("Reenable when issue #2426 has been solved (migrate exported files from 0.3 to 0.4)")
     def test_migrate_versions_old(self):
         """Migrating archives with a version older than the current should work."""
         archives = [
-            'export_v0.1.aiida',
-            'export_v0.2.aiida',
-            'export_v0.3.aiida'
+            'export_v0.1_no_UPF.aiida',
```
ltalirz: just pointing out that the meaning of `no_UPF` might become unclear to people once atomic stuff has been migrated out of aiida-core.
up to you whether you want to keep this label.

aiida/backends/tests/__init__.py
CasperWA: It is testing the functions in `aiida.cmdline.utils.migration.__init__.py`, hence it gets the name of the "module" or folder: `migration`. This is the same reasoning for the other names.
However, if you have another name you prefer, which still makes sense in the context of the files and functions that are being tested, it can quickly be changed. I don't have a better name myself though, sorry.

aiida/backends/tests/cmdline/commands/test_calcjob.py
CasperWA: My understanding is that `fixture` means files that are present as "records", though I will have to ask @sphuber for the original meaning and reasoning behind this name.

aiida/backends/tests/cmdline/commands/test_export.py
CasperWA: Yeah, I don't want to keep this label.

They will be renamed to `simple` instead.

aiida/backends/tests/cmdline/commands/test_calcjob.py
sphuber: I used the term fixture as the archive is a resource required for the testing, which matches some description for example [found here](https://en.wikipedia.org/wiki/Test_fixture#Software). Note that test fixtures can come in various forms and especially in the software analogue there are various ways of 'delivering' or setting up the fixture. `pytest` specifically implements them as a magic way to inject them as arguments into the test function. Here the fixture is just a data archive that is imported to populate the database. I don't think it is any less valid, but I am fine with renaming it to just archive.

aiida/cmdline/utils/migration/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+"""Migration export files from old export versions to the newest, used by `verdi export migrate` command."""
+# pylint: disable=wildcard-import
+from __future__ import division
+from __future__ import print_function
+from __future__ import absolute_import
+
+from aiida.cmdline.utils import echo
+from aiida.common.exceptions import DanglingLinkError
+
+from .utils import *
```
ltalirz: this is potentially dangerous and can make tracking problems difficult.

is there not an easier way to achieve this, e.g. by importing the respective functions using `importlib` when needed?

aiida/cmdline/utils/migration/utils.py
```diff
@@ -0,0 +1,57 @@
...
+    If no version number is passed, it will just extract the version number
+    and return it.
+
+    :param metadata: the content of an export archive metadata.json file
+    :param version: string version number that the metadata is expected to have
+    """
+    try:
+        metadata_version = metadata['export_version']
+    except KeyError:
+        raise ValueError('could not find the export_version field in the metadata')
```
ltalirz:
```suggestion
        raise ValueError("metadata is missing the 'export_version' key")
```

aiida/cmdline/utils/migration/utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+    :param metadata: the content of an export archive metadata.json file
+    :param version: string version number that the updated metadata should get
+    """
+    from aiida import get_version
+
+    old_version = metadata['export_version']
+    conversion_info = metadata.get('conversion_info', [])
+
+    conversion_message = 'Converted from version {} to {} with external script'.format(old_version, version)
```
ltalirz: What does "with external script" mean? It's AiiDA who is converting here, right?

aiida/cmdline/utils/migration/__init__.py
CasperWA: I am copying the way it is done in other `__init__.py` files throughout AiiDA core (check e.g. [here](https://github.com/aiidateam/aiida_core/blob/develop/aiida/cmdline/__init__.py).

It would be good to know of an "AiiDA standard" for these things in internal sub-modules (which is what I would call this "utility").

CasperWA: Or do you mean that I should specify an `__all__` in `utils.py` as well perhaps?

aiida/cmdline/utils/migration/v01_to_v02.py
```diff
@@ -0,0 +1,72 @@
...
+                if key == 'requires' and value.startswith(old_start):
+                    new_data[key] = get_new_string(value)
+                else:
+                    new_data[key] = replace_requires(value)
+            return new_data
+
+        return data
+
+    for field in ['export_data']:
+        for key in list(data[field]):
```
ltalirz: what about replacing all this code with something like
```
data[field] = {  get_new_string(k) : v for k,v in data[field].items() }
```

this applies in other places, too

aiida/cmdline/utils/migration/utils.py
CasperWA: Done

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,452 @@
...
+
+        content['process_type'] = entry_point_string
+
+    if fallback_cases:
+        headers = ['UUID', 'type (old)', 'process_type (fallback)']
+        warning_message = 'found calculation nodes with a type string that could not be mapped onto a known entry point'
+        action_message = 'inferred `process_type` for all calculation nodes, using fallback for unknown entry points'
+        write_database_integrity_violation(fallback_cases, headers, warning_message, action_message)
+
+    # step2: detect and delete unexpected links
```
ltalirz: add # step1  comment

aiida/cmdline/utils/migration/utils.py
CasperWA: It was the original message. I agree it doesn't make sense and should be changed. Any suggestions?

E.g. "Converted from version x to y by AiiDA vVERSION"?
Or maybe simply: "Converted from version x to y"?

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,452 @@
...
+import numpy as np
+
+from aiida.cmdline.utils import echo
+from aiida.cmdline.utils.migration.utils import verify_metadata_version, update_metadata
+
+__all__ = ('migrate_v3_to_v4',)
+
+
+def migrate_provenance_redesign(data):
+    """Utility function for `migrate_v3_to_v4` to handle REV. 1.0.20 (provenance redesign)"""
```
ltalirz: is this a reference to a django migration?
might be helpful to add 1-2 words so that people know where to look for the specification of this.

aiida/cmdline/utils/migration/v01_to_v02.py
CasperWA: This is the original function written by @sphuber. We have not changed it, only moved it to this new file.

I can try and change it?

aiida/cmdline/utils/migration/v03_to_v04.py
ltalirz: found! ;-)

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,452 @@
...
+    # step1: rename the type column of process nodes
```
ltalirz: perhaps move this down - currently it seems like one can return after step1 before getting to step2

CasperWA: Done

aiida/cmdline/utils/migration/__init__.py
ltalirz: this would also allow to move some of the helper function definitions inside  `v03_to_v04` to the top level without needing to be worried about name conflicts.

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: Yeah. We have done this throughout the actual migration function.
It is easier to reference the django migrations than the sqla, since django's migrations are naturally numbered.

I will add some extra explanation of this in the file's documentation, i.e. in the top of the file, where it at the moment "only" states what the general contents of this particular file is and where it is mainly used.

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,452 @@
...
+    # 'data.base.Str.'   -> 'data.str.Str.'
+    # 'data.base.List.'  -> 'data.list.List.'
+    if 'Node' in data['export_data']:
+        for content in data['export_data']['Node'].values():
+            if 'type' in content and content['type'].startswith('data.base.'):
+                type_str = content['type'].replace('data.base.', '')
+                type_str = 'data.' + type_str.lower() + type_str
+                content['type'] = type_str
+
+    # Apply migrations: 0010 - REV. 1.0.10
```
ltalirz: unless they're super short, most of these migrations should probably be separate functions, would you agree?

aiida/backends/tests/__init__.py
ltalirz: ok, fine

aiida/cmdline/utils/migration/__init__.py
CasperWA: > is there not a better way to achieve this, e.g. by having a "migrate" function in all of those modules and importing the respective function using `importlib.import_module` when needed?

I think I would need a more detailed explanation of this idea, sorry. I am not completely following what you intend the change should be.

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: It would be the most ideal and "pythonic" way, I guess. I will gladly do it, if you think it will result in a more transparent code in the end?

aiida/backends/tests/cmdline/commands/test_calcjob.py
ltalirz: I guess what I meant is that in most of the aiida-core testbase we are not using `fixture` in this more general sense (although it would apply to many other functions as well).
Using it here for the archive seems to suggest this function is "special" and can lead to confusion.
I'd vote to remove it for now.

CasperWA: > Can we remove the `_fixture` and also move the `import_archive` simply to something like `aiida.backends.tests.utils.archive`?

I will do this.

aiida/cmdline/utils/migration/__init__.py
ltalirz: > I am copying the way it is done in other __init__.py files throughout AiiDA core (check e.g. here.

I see.

In the aiida-core example you gave, the purpose of `import *` is to make the corresponding names (which are many and are constantly growing) available to the user at a certain point of the module hierarchy. Except for this, the `__init__.py` does nothing.

In the PR here, your `__init__.py` actually does work and the functions it imports are needed to do this work.
Unless the purpose is to make the various migrate functions available to users, I think there is no good reason to use `import *` here.

Note: Instead of my suggestion with `importlib`, you can also simply hardcode the imports (which are unlikely to change over time).

aiida/cmdline/utils/migration/utils.py
ltalirz: > "Converted from version x to y by AiiDA vVERSION"

Sounds good!

aiida/cmdline/utils/migration/__init__.py
CasperWA: Okay! :)
Thank you for the clarification. I feel more confident in making the changes now.

I will take you up on your note and simply hardcode the functions that are needed.

aiida/cmdline/utils/migration/v01_to_v02.py
ltalirz: I see - unfortunately it's not straightforward for me to see what you have changed here and what has been moved.

It might still be worth streamlining some of this code where it is obvious and doesn't take much time.
Anyhow, up to you since it is not your code.

aiida/cmdline/utils/migration/v03_to_v04.py
ltalirz: > if you think it will result in a more transparent code in the end

Yes. Linters complain about too long/complex functions not just to make programmers miserable but because there are good reasons to keep them short.

One example is just variable scope - in a function that's hundreds of lines long it can be very difficult to figure out which parts touch which variables.
Isolating individual steps into functions (you've already done most of the job in the comments) makes this much clearer.

aiida/cmdline/utils/migration/utils.py
CasperWA: Done

aiida/cmdline/utils/migration/v01_to_v02.py
CasperWA: > I see - unfortunately it's not straightforward for me to see what you have changed here and what has been moved.

No. The only way would be to not look at all the commits at the same time, but go back to before the commit, where I moved everything around (which I have left as a single commit, and should probably be left like this in the final merge as well).

> It might still be worth streamlining some of this code where it is obvious and doesn't take much time.
> Anyhow, up to you since it is not your code.

I will take a look at it, but now have it at the top of my priority queue :)

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: Thank you. Will make this a priority.

aiida/backends/tests/cmdline/commands/test_calcjob.py
CasperWA: As an addition - should I rename `aiida.backends.tests.fixtures` to `aiida.backends.tests.archives` as well?

ltalirz: Aren't these tests for the actual plugin fixtures?

CasperWA: No. This path is the top-directory, wherein all export archives are sorted and stored to use with tests.

ltalirz: I see, they aren't. Yes, good to rename!

CasperWA: Okay. In this PR or a separate one, since this will probably touch a lot of non-migration-related tests?

ltalirz: can be done in a separate PR

CasperWA: > can be done in a separate PR

Great. Opened issue #2817 for this.

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: Done

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: Done

aiida/backends/tests/cmdline/commands/test_export.py
ltalirz: Why is this marked as resolved?

ltalirz: Ok, I guess I'm not yet supposed to look at the PR - just ping me when I should check it again.

aiida/cmdline/utils/migration/__init__.py
CasperWA: Done

aiida/cmdline/utils/migration/v01_to_v02.py
CasperWA: I tried to change this, also for `migrate_v2_to_v3`, but it seemed my tests became slower. Maybe it is the memory usage required by the single-line commands? In any case, I will not touch these functions at this point.

aiida/backends/tests/cmdline/commands/test_export.py
CasperWA: Reloved it because I changed it to `_simple`. But let's keep it unresolved if there should be an uproar for the label change.

ltalirz: should I check it again?

CasperWA: This particular thing, or the updated code in general?

If it's the first, no.
If it's the second, I guess, if you want to, yes? :)

aiida/cmdline/utils/migration/v01_to_v02.py
ltalirz: Ok, fine

giovannipizzi:
@ltalirz I don't have the time now so if you feel it's ok, I trust your review! However, discussing this morning with @sphuber we were wondering if it would be wiser to wait anyway for the JSONB work before merging - maybe it's easier to discuss in person and report back the decision

CasperWA:
Again, I would like to remind and stress that a reviewer should also go through the new repository [aiida-export-migrations-tests](https://github.com/aiidateam/aiida-export-migration-tests).

szoupanos:
> @ltalirz I don't have the time now so if you feel it's ok, I trust your review! However, discussing this morning with @sphuber we were wondering if it would be wiser to wait anyway for the JSONB work before merging - maybe it's easier to discuss in person and report back the decision

Theoretically speaking, the EAV to JSONB change for the Django backend is an AiiDA internal change that should be transparent at the AiiDA ORM level and above.

Practically speaking, since JSONB is almost finished (I am currently resolving the last bug that we/I are aware of - #2807), you do a check/test after this is fixed.

ltalirz:
@szoupanos Thanks! Please ping us here as soon as the merge has happened.

sphuber:
I will be taking over the coordination of JSONB from now, so I will keep you updated on progress and when we can start testing and hopefully integrate and merge this.

szoupanos:
@sphuber Thanks!

CasperWA:
Updated for the rework of `verdi import`. Should maybe not have done this and instead waited for later. Sorry. Will wait for reviewers to ping me to update it or similar.

CasperWA:
I am currently working on adding the new migrations that will be introduced, one has already been merged in `develop`: #2937.
The changes will be pushed _after_ they have all been merged into `develop`.

CasperWA:
Changes that will have to be included:

- [x] Issue #2925: Drop `nodeversion` and `public` from `Node`.
- [ ] Issue #2927: Drop `transport_params` from `Computer`.
- [ ] Issue #2926: Rename `name` from `Computer` to `label`.

Changes that will _not_ have to be included:

- Issue #2928: "Shave down" `User` model to the essentials.

CasperWA:
**Current state**:
Tests must fail, since the database migration have still not been written, nor has the export version been upped to v0.5. However, the export archive migrations should now work for v0.5, as is intended at this point in time.

**What is missing**:
Update testing-repository `aiida-export-migration-tests` with tests for v0.5, a new "advanced" archive, and an upped version.

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,528 @@
...
+
+    # Helper function
+    def _migration_calc_job_option_attribute_keys(attr_id, content):
+        """Apply migration 0023 - REV. 1.0.23 for both `node_attributes*` dicts in `data.json`"""
+        if data['export_data']['Node'][attr_id]['type'] == 'node.process.calculation.calcjob.CalcJobNode.':
+            for key in content:
+                # custom_environment_variables -> environment_variables
+                if key == 'custom_environment_variables':
+                    content['environment_variables'] = content.pop('custom_environment_variables')
+                elif key.startswith(r'custom\_environment\_variables'):
```
sphuber: These comparisons should not be "regex" strings but normal strings. This means you should also remove the back slashes as the underscores have no special meaning, i.e. `key.startswith('custom_environment_variables')` . Same for all below. For reference, these underscores were necessary in the SQL because there we are actually doing a like comparison and underscore is a special character. However, here in python we are doing simple string comparisons.

Actually, the above is even superfluous because this whole `elif` block is not necessary and can be remove. In the SQL it was necessary because due to the EAV schema in Django, the nested elements in this dictionary would have a key `custom_environmnet_variables.some_key` but in the archive it is just a dictionary `{'custom_environment_variables': {'some_key': 1}}`

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,528 @@
...
+    from aiida.plugins.entry_point import ENTRY_POINT_STRING_SEPARATOR
+
+    fallback_cases = []
+    nodes_to_migrate = {}
+
+    for key, value in data['export_data']['Node'].items():
+        if 'type' in value and value['type'].startswith('calculation.job.'):
+            nodes_to_migrate[key] = value
+
+    if not nodes_to_migrate:
```
sphuber: This return here is probably wrong. There could be an archive without calculation jobs, but that still has other entries that would match all the conditions and migrations below. However, these would not be applied. Probably just remove this check/return and adapt the code below to be able to deal with an empty `nodes_to_migrate` (maybe even renaming to `calc_jobs_to_migrate` for clarity).

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,528 @@
...
+                if key == 'custom_environment_variables':
+                    content['environment_variables'] = content.pop('custom_environment_variables')
+                elif key.startswith(r'custom\_environment\_variables'):
+                    new_key = key.replace(r'custom\_environment\_variables', 'environment_variables', 1)
+                    content[new_key] = content.pop(key)
+
+                # jobresource_params -> resources
+                elif key == 'jobresource_params':
+                    content['resources'] = content.pop('jobresource_params')
+                elif key.startswith(r'jobresource\_params.'):
```
sphuber: Same here, remove the `elif key.startswith(r'jobresource\_params.')`

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,528 @@
...
+                    new_key = key.replace(r'jobresource\_params', 'resources', 1)
+                    content[new_key] = content.pop(key)
+
+                # parser -> parser_name
+                elif key == 'parser':
+                    content['parser_name'] = content.pop('parser')
+
+        # _process_label -> process_label
+        elif data['export_data']['Node'][attr_id]['type'].startswith('node.process.'):
```
sphuber: this should be a normal `if` because it is now not hit for `CalcJobNodes`

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,528 @@
...
+    These were (according to Antimo Marrazzo) never exported.
+    """
+    old_version = '0.3'
+    new_version = '0.4'
+
+    verify_metadata_version(metadata, old_version)
+    update_metadata(metadata, new_version)
+
+    # Apply migrations on Nodes
+    if 'Node' in data['export_data']:
```
sphuber: This if else construction is a bit weird and difficult to understand. Please just call all functions and then make sure in the functions to properly check if the `Node` part is there.
Note you can easily often use `for value in data['export_data'].get('Node', {}).values()` to do the check inline so you don't have to have another `if 'Node' in data['export_data']` block

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,528 @@
...
+            "user": {
+                "related_name": "dbcomments",
+                "requires": "User"
+            }
+        }
+    }
+    metadata['all_fields_info'].update(new_entities)
+    metadata['unique_identifiers'].update({"Log": "uuid", "Comment": "uuid"})
+
+    # Update data.json with the new Extras
```
sphuber: Probably put this in a separate function as well, even though it does not correspond to an actual database migration

aiida/cmdline/utils/migration/v03_to_v04.py
```diff
@@ -0,0 +1,528 @@
...
+
+
+def migration_remove_dbcomputer_enabled(metadata, data):
+    """Apply migration: 0031 - REV. 1.0.31
+    Remove DbComputer.enabled
+    """
+    # For data.json
+    if 'Computer' in data['export_data']:
+        for content in data['export_data']['Computer'].values():
+            if 'enabled' in content:
```
sphuber: for reference: these two lines are typically written as `content.pop('enabled', None)` which means try to pop the key, but if it does not exist, no biggy

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: Didn't all calculations start with the type `'calculation.job.'` or was it only `JobCalculation`?

CasperWA: Never mind, I can see that further down we have `'calculation.inline'` as well.

CasperWA: done

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: done

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: done

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: done

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: done

aiida/cmdline/utils/migration/v03_to_v04.py
CasperWA: For the `Log` and `Comment` entries as well? (The lines above this comment)

CasperWA: Only moved the introduction of the extras dicts for now

sphuber:
Thanks a lot to @yakutovicha, @CasperWA and @szoupanos

CasperWA:
> Thanks a lot to @yakutovicha, @CasperWA and @szoupanos

woooot :sunglasses: so cool - thanks @sphuber !

PR #2214, opened 2018-11-17
Speed up creation of Nodes in the AiiDA ORM

ltalirz:
* move AiiDAManager from `aiida.work` to `aiida.manage`
 * introduce caching of .objects in `aiida.orm.entities.Collection`
 * cache default user in `aiida.orm.user.User.Collection` (fix #2216)

## Benchmark on MacBook Pro, python3
  * before: 58.5s to create 10k nodes (57.2 spent inside `Node.__init__`)
  * after: 2.8s to create 10k nodes (1.7s spent inside `Node.__init__`)

The timing of 58.5s is significantly worse than what we wrote down during the tests with @szoupanos  and @giovannipizzi  (which was ~10s for 10k nodes), i.e. probably someone broke the caching.

Anyhow, the target we set then was: 2s for 10k nodes, which we have reached now on python3 (it's the 1.7s that should scale linearly with the number of nodes).

Issue #2400, opened 2019-01-18
Add support for `Process` classes to `QueryBuilder.append`


sphuber:
If I have the `CalcJob` class `PwCalculation`, which has entry point `aiida.calculations:quantumespresso.pw`, calling `QueryBuilder.append(PwCalculation).all()` should tell the query builder to return all nodes that have a type string `node.process.calculation.calcjob.CalcJobNode.` and a `process_type` string `aiida.calculations:quantumespresso.pw`. This is essentially an `AND` statement of these two values. Note that since `append` supports adding tuples of classes, e.g. `QueryBuilder.append((PwCalculation, ArithmeticAddCalculation))`, this should be unwrapped into two sub filters that are `OR`'ed, i.e. in pseudo SQL
```
        WHERE (
            type = 'node.process.calculation.calcjob.CalcJobNode.' AND
            process_type = 'aiida.calculations:quantumespresso.pw'
        ) OR (
             type = 'node.process.calculation.calcjob.CalcJobNode.' AND
             process_type = 'aiida.calculations:arithmetic.add'
        )
```
technically since the `process_type` should also support querying for sub classes, the filter on process type should be an `OR` of `process_type = 'aiida.calculations:quantumespresso.pw'` and `process_type like 'aiida.calculations:quantumespresso.pw.'`

ltalirz:
Just to summarize the logic here:


 * parameter `cls`: need to provide a *class* (instances not allowed, as before)
   * if an instance is provided (instead of a class), raise an exception
   * if Process class is provided, get corresponding Node class and filter by `process_type`
   * if Node class is provided, try to infer `process_type`. If it cannot be found (e.g. not a ProcessNode), just use the `type` for filtering.
   * if a tuple of classes is provided, use combination logic mentioned in the issue description

 * parameter `type`: we would need to add an additional parameter `process_type`. At this point, the question arises whether we really need to have both `type` and `process_type` as parameters. Should they just be replaced by filters instead? @lekah 

 * parameter `subclassing`: if `True`, include subclasses both of `type` and `process_type`.

sphuber:
Fixed in PR #2421

PR #2480, opened 2019-02-14
Prevent tests from deleting the default user

### LINKED ISSUE #2407: running the tests deletes the default user
I encountered the following error while trying to run my first WorkChain on the new `provenance_redesign` branch, python 2.7.15:

```
  File "/Users/leopold/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/orm/implementation/django/node.py", line 127, in __init__
    user = orm.User.objects.get_default().backend_entity
AttributeError: 'NoneType' object has no attribute 'backend_entity'
```
This is in a profile that was migrated to the current `provenance_redesign` branch (not created from scratch).

To me, this raises a few questions:

 * how is it possible that my profile does not have a default user?
 * should a profile with no default user be allowed? I.e. should `orm.User.objects.get_default()` be allowed to return `None` or should it raise an exception?
 * if you don't have a default user, how do you add one? I tried adding a new user via `verdi user` (which worked) but, apparently, this is not enough. How do I make it the default one?

Perhaps @giovannipizzi could comment?

ltalirz:
Fix #2407 

PR #2214 (inadvertently?) removed safeguards that prevented the default user from
being deleted in between tests.
Since AiiDA expects a default user in many operations, leaving the DB
in a state with no default user is not ideal.

This PR reintroduces safeguards that result in the default user
remaining untouched.

 * reintroduce safeguards for django
 * introduce safeguards for sqlalchemy
 * add testimplbase to pre-commit

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,7 +87,6 @@
         aiida/backends/sqlalchemy/transition_06dj_to_07sqla.py|
         aiida/backends/sqlalchemy/utils.py|
         aiida/backends/testbase.py|
-        aiida/backends/testimplbase.py|
         aiida/backends/tests/backup_script.py|
         aiida/backends/tests/backup_setup_script.py|
         aiida/backends/tests/base_dataclasses.py|
--- a/aiida/backends/djsite/db/testbase.py
+++ b/aiida/backends/djsite/db/testbase.py
@@ -43,6 +43,9 @@ class DjangoTests(AiidaTestImplementation):
 
     def clean_db(self):
         from aiida.backends.djsite.db import models
+        from aiida.manage import get_config
+
+        email = get_config().current_profile.default_user_email
 
         # I first need to delete the links, because in principle I could not delete input nodes, only outputs.
         # For simplicity, since I am deleting everything, I delete the links first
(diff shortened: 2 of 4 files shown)


giovannipizzi:
I've approved, but actually a few tests fail.
Also, I realised that the Jenkins machine was in a weird state with problems with docker, I upgraded it and rebooted, let's see if also the Jenkins tests start to work again.

ltalirz:
This did not resolve the issue observed in #2400.
After discussion with @szoupanos , the behavior observed in #2407 may actually be acceptable.
Will close #2407 once the problem behind #2400 has been identified.

szoupanos:
Just a small comment on this:

I just wanted to say that since we decided to expect a clean database at the beginning of each test class and leave a clean database at the end of each test class, we should ensure that this happens in both backends.

Maybe what I say it is obvious and it is already done, but I wanted to mention it.

sphuber:
I would tend to agree that we should be able to delete all resources, including default users for tests databases, and if they are needed for certain operations it should be recreated. The problem you experienced during #2400 has another root cause I think, where as you said, certain classes in AiiDA can retain references to ORM instances that point to database model instances that have been cleaned, leading to problems mostly in SqlAlchemy. You already opened an issue in #2515 to address a (part) of this question, so I suggest we try to solve the greater problem instead.

PR #2481, opened 2019-02-14
Migrate `Node` to the new backend interface

### LINKED ISSUE #2233: Migrating Node to new backend interface
Here's the plan:

1. `orm.nodes.Node`

- [x] Create concrete class with required methods for all nodes
- [x] Create `Collection` with all the `.object` required methods that require customisation e.g. `delete_node` can go here
Do not expose this class yet in the `orm.__init__.py` so that the original is still being used

2. `orm.implementation.nodes.BackendNode`

- [x] Move `AbstractNode` here and rename to `BackendNode`
- [x] Create `NodeCollection` with appropriate methods
- [x] Rename to `DjangoNode` and `SqlaNode` and convert to use `DjangoEntity`/`SqlaEntity` generics (.e. get rid of `id()`, `pk()` etc)
- [x] Create Django/SQLA collection implementations

3. Putting it together

- [x] Join up `Node` to use `BackendNode`
- [x] Expose `Node` in `orm.__init__.py` and remove conditional imports in `implementation/__init__.py`

Things to check:

- [x] What to do with `delete_code`?
- [x] Does QueryBuilder work correctly?

Anything else I've forgotten?

sphuber:
Fixes #2233 

A new abstract interface has been defined for `BackendNode` in the
module `aiida.orm.implementation.nodes`. This has subsequently been
implemented for the two current database backends Django and SqlAlchemy.
The role of the `BackendNode` class is to provide a single interface,
independent on the database backend implementation, to store and
retrieve data from `Node` instances in the database. The front end
`Node` class will then only ever interact with the backend entity that
it encapsulates and shall never directly interact with the database
model that is encapsulated within the `BackendNode` class.

This commit deprecates quite a few methods from the old node class,
notably all methods of the variant `*_db_*` that retrieved or set data
directly from the database. These are now part of the `BackendNode`
interface, but since all its methods will operate directly on the
database, the `db` infix has been dropped.

All business logic that does not operate directly on database data, such
as the caching of links and attributes of unstored nodes, is implemented
in the front end ORM class `aiida.orm.node.Node`. Also the biggest part
of the business logic is implemented here, to keep the backend layer
thin and lean.


giovannipizzi:
That's great!!!!

Issue #2482, opened 2019-02-15
Migrate v1.0.0 backward incompatible changes page to Sphinx Docs for release


ConradJohnston:
Currently, the docs on backward-incompatible changes are being composed on this [Wiki page](https://github.com/aiidateam/aiida_core/wiki/Backward-incompatible-changes-in-1.0.0) .

Before v1.0.0b is released, the following should be done:

- [x] Finalise content of the wiki page
- [x] Convert the markdown format content to restructured text. 
      One can use `pandoc` for this:  `$ pandoc wikitext.md -o sphinxtext.rst`
- [x] Replace the placeholder in the docs (docs/source/install/updating_installation) with this finalised text

giovannipizzi:
I would do this for 1.0.0 and keep in the wiki for the beta lifetime

PR #2483, opened 2019-02-15
DOCS: Create placeholder for v1.0 breaking changes

### LINKED ISSUE #2320: Document backwards-incompatible things in 1.0.0
Merge things with this:
https://aiida-core.readthedocs.io/en/latest/concepts/workflows.html#upgrading-workchains-from-pre-v1-0-0
and move to Install->Updating installations.

**Please dump all changes in [this wiki page](https://github.com/aiidateam/aiida_core/wiki/Backward-incompatible-changes-in-1.0.0).**

ConradJohnston:
Resolves #2320

- Created a placeholder section in install/updating_installation
where the finalised prose describing all of the important behaviour changing and
backwards-incompatible change should be placed before the v1.0.0b release.
The content of this section is being drafted on the wiki at:
https://github.com/aiidateam/aiida_core/wiki/Backward-incompatible-changes-in-1.0.0
Before the beta release, the finalised content from the wiki page should be converted using
`pandoc` and then pasted in this section.

- Moved the section about converting workchains to the wiki. This section should be
included in the larger v1.0 changes text so that we don't have to maintain two seperate
versions of it.

Refer also to issue #2482.

### DIFF
--- a/docs/source/install/updating_installation.rst
+++ b/docs/source/install/updating_installation.rst
@@ -86,3 +86,13 @@ To find the update instructions for older versions of AiiDA follow the following
   SQLAlchemy backend. In case you were using SQLAlchemy before the introduction
   of Alembic, you may experience problems during your first migration. If it is
   the case, please see :ref:`this section <first_alembic_migration>`.
+
+
+
+.. _backward_incompatible_changes_beta_release:
+
+Backward incompatible changes in ``v1.0.0``
+===========================================
+ 
+The list of all backward-incompatible changes between the 0.x series and AiiDA 1.0 is 
+maintained `here <https://github.com/aiidateam/aiida_core/wiki/Backward-incompatible-changes-in-1.0.0>`_.


docs/source/install/updating_installation.rst
```diff
@@ -86,3 +86,15 @@ To find the update instructions for older versions of AiiDA follow the following
...
+
+
+.. _backward_incompatible_changes_beta_release:
+
+Backward incompatible changes in ``v1.0.0``
+===========================================
+
+.. note::
+ 
+  Insert the finalised content form the `Wiki <https://github.com/aiidateam/aiida_core/wiki/Backward-incompatible-changes-in-1.0.0>`_ here.
```
giovannipizzi: I would add a message more targeted for users, like "The list of all backward-incompatible changes between the 0.x series and AiiDA 1.0 is maintained [here](https://github.com/aiidateam/aiida_core/wiki/Backward-incompatible-changes-in-1.0.)."

ConradJohnston: Yes, no worries. Done.

PR #2488, opened 2019-02-15
Add more complicated blocking prompt for `verdi database migrate`

### LINKED ISSUE #2427: Insert automatic migration block for provenance redesign migrations
The migrations that have been introduced in the `provenance_redesign` (everything after `0015` in the django indexing) have big consequences and are irreversible. Before merging this branch into `develop` we should put a prompt that will block users from accidentally migrating their databases without understanding the consequences. The problem of the prompt will be that it will also block automatic migrations for the Travis environments. A solution should be found that for those situations the prompt is automatically closed.

ConradJohnston:
Fixes #2427 . 

- Adds a series of convoluted prompts to make the user aware that the migrations from `provenance redesign` are a serious business. Perhaps over the top, but I'll argue that they're effective.
- Disables the `--force' option, but retains it for the future. 

Automatic migrations are possible by not going through the click interface.

### DIFF
--- a/aiida/cmdline/commands/cmd_database.py
+++ b/aiida/cmdline/commands/cmd_database.py
@@ -35,14 +35,67 @@ def database_migrate(force):
     profile = manager.get_profile()
     backend = manager._load_backend(schema_check=False)  # pylint: disable=protected-access
 
-    if not force:
-        echo.echo_warning('Migrating your database might take a while.')
-        echo.echo_warning('Before continuing, make sure the daemon is stopped and you have a backup of your database.')
-        echo.echo_warning('', nl=False)
-        confirm_message = 'Are you really sure you want to migrate the database for profile "{}"?'.format(profile.name)
-        click.confirm(confirm_message, abort=True)
+    if force:
+        raise NotImplementedError('--force is disabled currently to prevent accidental migrations.')
 
-    backend.migrate()
+    echo.echo('\n' + '*' * 79 + '\n')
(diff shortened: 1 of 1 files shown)


aiida/cmdline/commands/cmd_database.py
```diff
@@ -35,14 +35,67 @@ def database_migrate(force):
...
+        echo.echo_warning('Have you stopped the daemon?')
+        echo.echo_warning('', nl=False)
+        confirm_message = 'Type "I HAVE STOPPED THE DAEMON"'
+        response = ''
+        while response != 'I HAVE STOPPED THE DAEMON':
+            response = click.prompt(confirm_message)
+
+        # Final prompt
+        echo.echo('')
+        message = 'Are you absolutely ready to migrate profile "{}"? This will be PERMENANT.'.format(profile.name)
```
giovannipizzi: ahah I like this :-D only typo PERMANANT -> PERMANENT, apart for this it can be approved

aiida/cmdline/commands/cmd_database.py
```diff
@@ -35,14 +35,67 @@ def database_migrate(force):
...
+        echo.echo_warning(message)
+        echo.echo_warning('', nl=False)
+        confirm_message = 'Type "MAKE IT SO"'
+        response = ''
+        while response != 'MAKE IT SO':
```
sphuber: Easy now, Jean-Luc

PR #2501, opened 2019-02-18
Clean up aiida.common

### LINKED ISSUE #2499: move aiida.common.serialize under aiida.orm
`aiida.common.serialize` imports `aiida.orm`, while "Modules in this sub package [common] have to run without a loaded database environment".

We should probably move it to aiida.orm.utils
@giovannipizzi @sphuber Do you agree?

### LINKED ISSUE #2500: move DBLogHandler from aiida.common.log to aiida.orm
The DBLogHandler clearly needs the ORM but currently lives in aiida.common.log

I guess it should be moved to aiida.orm (@giovannipizzi @sphuber just let me know where exactly).

ltalirz:
Move functions/modules that use the dbenv out of `aiida.common` and to `aiida.orm.utils`

 - [x] fix #2499 
 - [x] fix #2500


sphuber:
Do not forget to adapt the logging configuration in `aiida.common.log` that points to the `DbLogHandler`
```

        'dblogger': {
            'level': lambda: get_config_option('logging.db_loglevel'),
            'class': 'aiida.common.log.DBLogHandler',
        },
```
you need to change the path of the class

PR #2506, opened 2019-02-20
Make the node repository API backend agnostic

### LINKED ISSUE #2504: Prepare the node's repository API for upcoming repository redesign
Historically, the node repository only has a file system implementation and so its interface was geared towards interacting with a file system. However, given that in the near future we will be implementing other repository solutions, e.g. an object store, one can no longer assume that one can interact with the repository contents as if they were files on the local file system. The current node repository interface should be adapted with this future change in mind.

sphuber:
Fixes #2504 

Initially, the only backend supported for the file repository of nodes
was the local file system. In the near future, additional backends such
as an object store or a remote file system will be implemented. For this
to work the API of the node repository will have to be adapted because
now it is geared to work with a local file system. To prevent too many
changes after the release of the beta, we already define the API here
that should be able to support the different repository backends that
will be implemented in the future.

### DIFF
--- a/.ci/test_daemon.py
+++ b/.ci/test_daemon.py
@@ -161,14 +161,22 @@ def validate_cached(cached_calcs):
             valid = False
 
         if isinstance(calc, CalcJobNode):
-            if 'raw_input' not in calc.repository._get_folder_pathsubfolder.get_content_list():
-                print("Cached calculation <{}> does not have a 'raw_input' folder".format(calc.pk))
+            original_calc = load_node(calc.get_extra('_aiida_cached_from'))
+            files_original = original_calc.list_object_names()
+            files_cached = calc.list_object_names()
+
+            if not files_cached:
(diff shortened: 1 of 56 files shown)


sphuber:
>     * maybe now the 'disabling' of add_path doesn't work anymore?

actually, the `add_path` method no longer exists on the `Node` class, so there is no point in disabling it. I think we should just remove it. Currently, there exist no `Node` methods that operate directly on the repository, they are only exposed on the `Repository` class itself through the `repository` property. Probably we will need to expose the methods that should be available for the repository, directly on the node itself, which internally will just pipe through, because in the future changes to the repository will anyway need changes on the node instance in the database, so these have to be coordinated by the `Node` class and cannot be done directly on the repository. That will also make it possible again to disable certain actions for certain node subclasses.

Maybe we should already do this in this PR. Make it impossible to access the repository directly from the node class (i.e. remove the property) and only expose methods that should be callable directly on the node.

PR #2511, opened 2019-02-21
Added node label in the tree end point response

waychal:
Added missing node label in the response of rest api tree endpoint.


ltalirz:
Could you perhaps add one test?
Or is accessing the nodelabel already tested?

waychal:
@ltalirz there is test available for tree endpoint. We do not specifically check the information passed for every node or edge.

ltalirz:
> We do not specifically check the information passed for every node or edge.

Perhaps we could change this situation? ;-)

Does not have to be entirely exhaustive but checking at least once for each key that the expected value is returned would seem reasonable, no?

waychal:
I have updated the test where the important attributes displayed for node in front end are checked in response json data.

Issue #2502, opened 2019-02-19
Export/import of logs & comments - Recursive queries


szoupanos:
We should verify why the nodes don't reference the comments and the logs in the metadata.json and it is the other way around

```
        "Log": {
            "time": {
                "convert_type": "date"
            },
            "loggername": {},
            "levelname": {},
            "message": {},
            "metadata": {},
            "dbnode": {
                "requires": "Node",
                "related_name": "dblogs"
            },
            "uuid": {}
        },
        "Comment": {
            "uuid": {},
            "ctime": {
                "convert_type": "date"
            },
            "mtime": {
                "convert_type": "date"
            },
            "content": {},
            "dbnode": {
                "related_name": "dbcomments",
                "requires": "Node"
            },
            "user": {
                "related_name": "dbcomments",
                "requires": "User"
            }
        }
```

If the nodes would reference the logs and the comments then, the `fill_in_query` method of the export would construct a query that would automatically export the logs and the comments of the given nodes.

If there is no serious reason why we don't follow this approach, we should proceed to this change to be uniform with the way other entities are exported.

CasperWA:
When changing these it seems there is a reason why they are tied so tightly to the database schema:

```shell
Traceback (most recent call last):
  File "/home/candersen/virtualenv/test/aiida/aiida_core/aiida/orm/implementation/querybuilder.py", line 237, in get_column
    return getattr(alias, colname)
  File "/home/candersen/virtualenv/test/lib/python3.6/site-packages/sqlalchemy/orm/util.py", line 431, in __getattr__
    raise AttributeError(key)
AttributeError: dblog

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/candersen/virtualenv/test/aiida/aiida_core/aiida/backends/tests/test_export_and_import.py", line 505, in test_non_default_user_nodes
    export([sd2], outfile=filename1, silent=True)
  File "/home/candersen/virtualenv/test/aiida/aiida_core/aiida/orm/importexport.py", line 2948, in export
    export_tree(what, folder=folder, silent=silent, **kwargs)
  File "/home/candersen/virtualenv/test/aiida/aiida_core/aiida/orm/importexport.py", line 2365, in export_tree
    for temp_d in partial_query.iterdict():
  File "/home/candersen/virtualenv/test/aiida/aiida_core/aiida/orm/querybuilder.py", line 2050, in iterdict
    query = self.get_query()
[...]

PR #2512, opened 2019-02-21
Optimize export of Logs and Comments

### LINKED ISSUE #2503: The query for Logs and Comments should be outside of the export graph exploration loop
At the end of the graph exploration loop where we recursively try to find all the nodes that should be exported given an initial set of nodes, we search for the related comments and logs.

```
        # Universal "entities" attributed to all types of nodes
        if include_logs:
            # Get related log(s) - universal for all nodes
            qb = QueryBuilder()
            qb.append(Node, tag='node', filters={'id': curr_node_id})
            qb.append(Log, with_node='node', project=['id'])
            res = {_[0] for _ in qb.all()}
            given_log_entry_ids.update(res)

        if include_comments:
            # Get related comment(s) - universal for all nodes
            qb = QueryBuilder()
            qb.append(Node, tag='node', filters={'id': curr_node_id})
            qb.append(Comment, with_node='node', project=['id'])
            res = {_[0] for _ in qb.all()}
            given_comment_entry_ids.update(res)
```

Having these queries there is inefficient (there should be one query for the logs and one for the comments for all the found nodes and should not be launched per node) but we should first clarify that they are needed. Please look at issue #2502

CasperWA:
Fixes #2503.

Moving QueryBuilder that retrieves Logs (and another for Comments) outside while-loop exclusively for Nodes.

Furthermore, retrieving Logs and Comments without introducing another for-loop, instead fully utilizing QueryBuilder.

### DIFF
--- a/aiida/orm/importexport.py
+++ b/aiida/orm/importexport.py
@@ -2252,22 +2252,22 @@ def export_tree(what, folder, allowed_licenses=None, forbidden_licenses=None,
                 res = {_[0] for _ in qb.all()}
                 given_calculation_entry_ids.update(res - to_be_exported)
 
-        # Universal "entities" attributed to all types of nodes
-        if include_logs:
-            # Get related log(s) - universal for all nodes
-            qb = QueryBuilder()
-            qb.append(Node, tag='node', filters={'id': curr_node_id})
-            qb.append(Log, with_node='node', project=['id'])
-            res = {_[0] for _ in qb.all()}
-            given_log_entry_ids.update(res)
+    ## Universal "entities" attributed to all types of nodes
+    # Logs
+    if include_logs:
+        # Get related log(s) - universal for all nodes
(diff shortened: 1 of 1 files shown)


szoupanos:
This is OK. Thanks!
But I would try a bit more the approach of #2502

CasperWA:
> But I would try a bit more the approach of #2502

This will include a drastic change in the existing logic in my opinion. Or perhaps it may be achieved with yet another reference dict. However, I don't see the necessity of it at this point in time.

In any case, let's continue that discussion in #2502 :)

szoupanos:
This is ok for beta 1. However, we should avoid querying Nodes when exporting Logs & Comments in fill_in_query. Doing that will improve the speed (I don't know how much) but will make the code even more complex

I believe that the whole logic should be re-adapted & re-though based on the ideas described at #2516 (also other ideas are welcome)

Issue #2398, opened 2019-01-17
Rename `ParameterData` to `Dict`


sphuber:
Exactly what it says on the tin. The name `ParameterData` is often found confusing, especially when it is used as an output of a calculation. Given that we already have `List`, the name `Dict` describes exactly what it is and also directly mirrors the python built in data structure.

sphuber:
Fixed in PR #2530

PR #2517, opened 2019-02-22
Renaming `DbNode.type` to `DbNode.node_type`

### LINKED ISSUE #2510: Rename `type` attribute of `Node` to `node_type`
The `type` keyword is reserved in python and using it for a class attribute is causing quite a bit of trouble. Besides, with the addition of the `process_type` column, renaming it to `node_type` makes its role more clear.

sphuber:
Fixed #2510 

While we are at it....

### DIFF
--- a/aiida/backends/djsite/db/migrations/0029_dbnode_type_to_dbnode_node_type.py
+++ b/aiida/backends/djsite/db/migrations/0029_dbnode_type_to_dbnode_node_type.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# pylint: disable=invalid-name,too-few-public-methods
+"""Renaming `DbNode.type` to `DbNode.node_type`"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import absolute_import
+
+# Remove when https://github.com/PyCQA/pylint/issues/1931 is fixed
+# pylint: disable=no-name-in-module,import-error
+from django.db import migrations
+from aiida.backends.djsite.db.migrations import upgrade_schema_version
+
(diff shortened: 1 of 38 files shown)


sphuber:
@giovannipizzi faster than I anticipated. I could start working on #2398 . Originally we discussed that it would not be critical, but seeing as we still have a week and I could get started on it now, might be worth it to do it before the beta release. What do you think?

Also the migrations here were dead easy. Are we sure we are not missing anything (not including the migration of export archives)?

sphuber:
Uhm.... there is a reason it went so fast. I forgot to un-comment the migration tests :man_facepalming: 
but of course those are failing miserably.

giovannipizzi:
I think the migrations are ok (we're getting too used to complex ones :-D )
Ok to proceed to #2398 (a bit more complex, requires again a migration), but then I suggest to prioritise stating cleaning up the wiki page for backward-incompatible things (I just reopened #2320)

sphuber:
This might not actually be reasonably doable after all. Changing the `DbNode` model by changing the `type` column's name breaks all the migration tests. The actual database model can no longer be used anywhere, which therefore includes our backend and frontend classes. The only solution is to only use the `apps` to get the model at the time of the migration that is being tested. This is doable for quite a few tests and I have already migrated them to only use the django apps. But there are three `TestCalcAttributeKeysMigration`, `TestDuplicateNodeUuidMigration` and the `TestTrajectoryDataMigration` that use attributes, the repository and trajectory array setters and getters, that are not clear how to decouple from our ORM. At this point I am not sure whether the required effort would be worth it, even though I think the name change is important. Any ideas @giovannipizzi ?

PR #2520, opened 2019-02-24
Update python dependencies

ltalirz:
This PR updates all python dependencies of aiida-core to the latest version, except for:

 * pycifrw (version 4.4 still has an issue with python2, pinged maintainer about new release)
 * ipython (6 drops python2 compatibility)
 * pg8000 (<1.13.0 was hardcoded)
 * tornado (5 drops python2 compatibility)
 * pymatgen (2019 drops python2 compatibility)
 * prospector (we're currently using later versions of `pylint-django` than `prospector` demands because the versions demanded by `prospector` make pylint crash. While `prospector` 1.1.5 was fine with us increasing the `pylint-django` version, `prospector` 1.1.6.2 gets touchy and refuses to run, see [here](https://travis-ci.org/ltalirz/aiida_core/jobs/497709384#L1514))

I.e. we're pretty much up to date, except for dependencies that drop python2 compatibility.

Most notably, this PR upgrades to click 7

 - [x] @sphuber Please have a look at `_start_circus`.
    It seems click now does some character replacements (e.g. `_` to `-`) which forced me to rename this method. It is anyhow not clear to me why it is defined as a click command if people are not supposed to run it from the command line.


sphuber:
If I remember correctly and looking at the code, this was the easiest way of guaranteeing that the correct profile is used in the sub process that launches the circus controller. If we made it a normal function and called that somehow as a sub process, then we would still have to pass a specific profile to it somehow, because it is not necessarily the default, as the user can call `verdi daemon start` with the `-p` option, specifying another profile.

To still make sure the user does not call it directly, I "hid" it by prefixing. I also noticed a while ago that click 7 breaks this because it mangles names with underscores. In principle the change you made should be fine, I will just try and look now if we can define "hidden" commands in click 7

ltalirz:
Thanks, just let me know if you have a solution.

It still seems to me we should find a way of passing the profile on to a subprocess without having to go through a click command - that may be needed elsewhere as well.
However, this does not necessarily need to be fixed in this PR

sphuber:
If I am not missing anything, passing the correct profile to a sub process will almost *have* to be done through command line arguments, which will have to be parsed. It seems pointless to not use click in this case and try and reinvent the wheel to parse and validate the profile option.

sphuber:
[It was this commit](https://github.com/pallets/click/commit/5d1df0e042b2f8b0dee8f6dd9ce74edce331a950) that introduced the breaking change in `click==7.0.0`. Their reasoning is that command names are by default defined directly the python function name that is decorated. Since having dashes in function names is invalid python syntax, people are basically forced to use underscores. However, the standard in CLIs is to use dashes. This could be fixed by developers by explicitly setting the command name, but they just changed the default. Now by default underscored are replaced by dashes and the developer can specify a name with underscores explicitly if they want to do so.

sphuber:
I just realized that the strategy of using an underscore to hide it does not even work. It probably was still a remnant of our old home-grown CLI framework. Since that is the case, I am fine with accepting the change as is. If we really feel the command needs to not be autocompleted, we can tackle this problem in a separate issue

ltalirz:
> passing the correct profile to a sub process will almost _have_ to be done through command line arguments, which will have to be parsed.

hm, makes sense. I know too little about how the daemon works under the hood, I'll let you explain it to me some time ;-)

Feel free to merge when you're happy

PR #2530, opened 2019-02-25
Rename `ParameterData` to `Dict`

### LINKED ISSUE #2398: Rename `ParameterData` to `Dict`

sphuber:
Fixes #2398


sphuber:
@giovannipizzi this one is also ready for review. Don't worry about the test failing, that is just pre-commit yapf error after rebase.

sphuber:
We'll have a proper funeral once we release the beta, we'll celebrate the beautiful life it has had

PR #2540, opened 2019-02-27
Improve the efficiency of `DbLog` migration for SqlAlchemy

### LINKED ISSUE #2538: Make protected entity methods public if they should be accessible from the outside
In the original design, the approach was taken to make class methods protected by prefixing them with an underscore as a way of hiding them from the user. However, many of these methods should in fact be called from outside the class hierarchy, violating the fundamental purpose of making them protected and requiring us to put loads of linter exceptions `disable=protected-access`.

sphuber:
Fixes #2538 

The old implementation was using a `NOT IN` with the list of all node
ids to filter all log entries that had an `objpk` that did not
correspond to a node entry. This is extremely inefficient as it loads
the entire node pk column in memory and has to traverse it for each log
entry. In this commit, we change this to a construct that will simply
search for the objpk in the node table, which is indexed and simply
returns 1 if it exists, which is than checked in the filter of the logs.

Note: I did not touch the Django migration as I don't have a production database on which I could test it and it does the relevant queries through the Django ORM. I hope they will unwrap this into an efficient SQL query. What do you think?

### DIFF
--- a/aiida/backends/sqlalchemy/migrations/versions/041a79fc615f_dblog_cleaning.py
+++ b/aiida/backends/sqlalchemy/migrations/versions/041a79fc615f_dblog_cleaning.py
@@ -69,8 +69,8 @@ def get_logs_with_no_nodes_number(connection):
         text("""
         SELECT COUNT(*) FROM db_dblog
         WHERE
-            (db_dblog.objname LIKE 'node.%') AND
-            (db_dblog.objpk NOT IN (SELECT id FROM db_dbnode))
+            (db_dblog.objname LIKE 'node.%') AND NOT EXISTS
+            (SELECT 1 FROM db_dbnode WHERE db_dbnode.id = db_dblog.objpk LIMIT 1)
         """)).fetchall()[0][0]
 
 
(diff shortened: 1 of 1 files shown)


giovannipizzi:
For Django, it would be good to try even in a empty DB (e.g. during 'verdi setup') and print the query that is generated by Django - if you do please report and we try to see its efficiency (potentially you can just then run the SELECT query on the big DB you have - even if the DB is using the SQLA schema, the query should work the same as it does not touch attributes).

szoupanos:
Honestly speaking, I think that you exaggerate :-)

I really doubt that it is so inefficient as you describe and I would not immediately agree with you if I don't see a query plan of the execution of both queries on a big database which touches all the log records and all the node records.

I really think that it depends on the number of nodes that you have, on the number of logs per node, the available memory and how the database optimizer decider to execute each query.

szoupanos:
In order to conclude, I would focus more on correctness (and care more about correctness) and I would address speed when such a need appears.

sphuber:
Well I have a practical example where the original did not even finish within 24 hours, after which I killed it and the new version takes less than a minute

PR #2545, opened 2019-02-27
bug fix for `verdi node delete`

zhubonan:
Fix a bug such that if no argument is passed to `verdi node delete`. It will try to delete ALL notes in the database.
This is because `QueryBuilder().append(Node, filters={'or': []}, project='id').all()` match to every single node and projects its id. 

Perhaps this behaviour of `QueryBuilder` is not as expected? It makes more sense if an exception is raised when the `'or'` list contains no element at all. I am not sure if it is still the case in 1.0.0 versions.


ltalirz:
@zhubonan thanks for the bug & the fix!

> Perhaps this behaviour of `QueryBuilder` is not as expected? It makes more sense if an exception is raised when the `'or'` list contains no element at all.

As a general rule, I think it makes sense if an "empty" or condition is true.
This can be helpful in nested, autogenerated logic trees where you are passing a list of conditions into "or" that may or may not contain elements.

I can see why it may look a bit confusing, but I think it will also be rare that a user writes `{'or': []}` by hand.

The bug here is that not providing a node PK results in no filter being applied (rather than doing nothing because of the missing command line arguments).

aiida/cmdline/commands/node.py
```diff
@@ -351,8 +351,12 @@ def run(self, *args):
         for key, req_vals in (('id', parsed_args.pks), ('uuid', parsed_args.uuids)):
             if req_vals:
                 filters['or'].append({key:{'in':req_vals}})
-
-
+        
+        # If neither of the arguments is passed, exit the program
+        if not filters['or']:
+            print "Nothing to delete"
```
ltalirz: Please change this to
```suggestion
            print "No nodes specified."
```

zhubonan:
I see. Thanks.

ltalirz:
Just checked that `provenance_redesign` is not affected.

PR #2553, opened 2019-03-01
Update authors & bump version

ltalirz:


ltalirz:
@sphuber By the way, what strategy do you suggest for merging?

I believe the "canonical" way in [gitflow](https://gitversion.readthedocs.io/en/latest/git-branching-strategies/gitflow-examples/#support-branches) would be to 
  * merge `release_v0.12.3` into `master`
  * create a `support/1.x` branch from `master` 
  * create `release_v0.12.4` from `support/1.x` (and follow this model in the future)

By default, this would involve merging `develop` into `master`, when the 1.0 release comes.

If you really want to avoid this due to merge conflicts, we could, at that point, replace `master` with `develop`.
To me, however, it would seem worth attempting the merge - consistency of the branching system is certainly worth something as well.

sphuber:
Good question, so far we have always been merging the release branch into `master` and then after the merge, merging `master` into `develop` to get all the commits from the release branch in `develop`. We were recently discussing this with @giovannipizzi and we were not sure if this still makes sense in this case. The commits are mostly backports or fixes that do not affect `develop` anymore. I just discussed with @giovannipizzi again and we think it might be best to *not* merge into `master` for now, but just tagging directly on the release branch and doing the release and then creating a `release_v0.12.4` branch.

ltalirz:
> we think it might be best to _not_ merge into `master` for now, but just tagging directly on the release branch and doing the release and then creating a `release_v0.12.4` branch.

Ok, I can do the release without merging into master for the time being but I would suggest we should still discuss this - I would definitely suggest to merge into master. 
Machinery (documentation, ...) and people will assume that master contains the last released version, and rightfully so.

ltalirz:
P.S. By the way, if this is out of fear of merge conflicts, I'm happy to attempt a merge into develop over the weekend.

sphuber:
It is not really the fear of merge conflicts but rather if it makes sense. I agree that this other approach will mean that `master` will be in an unconventional state, but this would be "resolved" if we decide to merge the beta into master to release it. Anyway, I agree that a discussion would be good. Would it be possible to hold of the actual release to monday so we can discuss in person?

ltalirz:
> 
Would it be possible to hold of the actual release to monday so we can discuss in person?

I would prefer to do the release now, since there is lots of stuff that follows (conda, Quantum Mobile, ...) that I can do more easily on the weekend.
However, it's no problem to hold off merging into master and I can simply wait with announcing the release as well.

Issue #2342, opened 2018-12-11
Update export/import to provenance redesign


sphuber:
We should check that export/import does the right thing after the provenance redesign. Also, the schema version has been upped to `v0.4` and so a migration for `v0.3` archives should be implemented.

- [x] Check that export/import works correctly and respects provenance redesign rules
    * Find at which point the archive version was updated to `v0.3`
    * Determine what the database schema version was at that point
    * Look at all the database migrations that have been applied since, and make a list of what should affect the export archives
    * The cumulative changes should be defined in a migration that can be applied to an archive of `v0.3` to update it to `v0.4`.
- [x] Re-enable `verdi export create`
- [x] Implement migration of `v0.3 -> v0.4` archives will be done in a separate issue (see #2426)
- [x] Re-enable all skipped unittests (see #2426)
- [x] Migrate test fixture archives (see #2426)

CasperWA:
Import/export tests for updated db schema introduced by db migrations since v0.3 are available in branch [`fix_2342_update_export_import_to_prov_redesign`](https://github.com/aiidateam/aiida_core/tree/fix_2342_update_export_import_to_prov_redesign).

Furthermore, I have cleaned up the `export_and_import.py` test file a bit, moving tests into more-or-less relevant test classes.

**NOTE**: This does not fix #2342, since this also calls for writing a migration for an exported file from v0.3 to v0.4. As discussed with @sphuber and @szoupanos, this will not be implemented at this point.

Some tests in `TestLinks` have been removed, since they duplicate existing tests.

**[WIP]**: Other test classes will be rifled through and its tests weighed concerning duplication, with the attempt to streamline the test file further.

sphuber:
Closed in #2554

PR #2554, opened 2019-03-01
Update verdi export and import

CasperWA:
Skip specific cmd_import and cmd_export tests until issue #2426 has been solved.
Some were previously related to issue #2342.

Fixes issue #2342.

Open access to `verdi export create`.

Do recursive migration of export files when calling `verdi export migrate` in order to always create an export file that respects the latest export version.

Note! The migration from v0.3 to v0.4 is as of yet still not implemented. It is underway with PR #2478.

### DIFF
--- a/aiida/orm/importexport.py
+++ b/aiida/orm/importexport.py
@@ -717,6 +717,11 @@ def import_data_dj(in_path, user_group=None, ignore_unknown_nodes=False,
                     else:
                         new_entries[model_name] = data['export_data'][model_name].copy()
 
+            # Show Comment mode if not silent and Comments exist in existing_entries
+            if not silent:
+                if COMMENT_ENTITY_NAME in existing_entries:
+                    print("Comment mode: {}".format(comment_mode))
+
             # I import data from the given model
             for model_name in model_order:
                 cls_signature = entity_names_to_signatures[model_name]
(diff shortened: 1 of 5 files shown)


aiida/orm/importexport.py
```diff
@@ -2254,15 +2263,15 @@ def export_tree(what, folder, allowed_licenses=None, forbidden_licenses=None,
 
     ## Universal "entities" attributed to all types of nodes
     # Logs
-    if include_logs:
+    if include_logs and len(to_be_exported) is not 0:
```
sphuber: Don't check for empty list by comparing length to zero, if it is an empty list, you can just do `if include_logs and to_be_exported:`

CasperWA: Done - also removed this part completely from the other PR, when I realized it was duplicated.

sphuber:
:+1: thanks @CasperWA

aiida/orm/importexport.py
szoupanos: @sphuber @CasperWA 
This is true for sequences and it the advised way to check if they are empty
https://www.python.org/dev/peps/pep-0008/

`to_be_exported`  is a set, and a set is not a sequence (at least conceptually).
From a fast lookup, I see that the above check does by accident the correct thing
https://stackoverflow.com/questions/44813375/truth-value-of-empty-set
(from the discussion, I understand that it is not documented but since it is based on the implementation of the sequence, it has the 'behaviour' of a sequence - this is what I understand for the second comment).

If it is the case, I would do a comparison based on the size of the set.

PR #2564, opened 2019-03-04
Prevent launching or construction of `WorkChain` and `CalcJob` base class

### LINKED ISSUE #2474: What should be the process_type when creating a CalcJobNode by hand?
Currently, when storing a CalcJobNode by hand, the process_type is empty.
When you submit a CalcJob process (without subclassing), it stores the module path in the process_type (similar for WorkChain etc.).

Questions:
 1. Should the two operations produce the same result (e.g. storing by a user vs. submission) ?
 2. What should be the value of the string stored (empty, ...)?

Note: Current migrations will put an empty process_type for what used to be a JobCalculation (but will not touch the process_type column for existing WorkChains). Keep this in mind for migrations.

Once this is decided, make sure #2400 is not affected.

### LINKED ISSUE #2493: Prevent base `WorkChain` and `CalcJob` class from being run
Currently there is nothing stopping the users from passing `WorkChain` or `CalcJob` to `submit` or `run`. The engine will try to run them but it will fail. However, a node will have been created. It should be forbidden to launch these base classes.

sphuber:
Fixes #2493 and fixes #2474 

The `WorkChain` and `CalcJob` classes are base classes that are meant to
be sub classed as they miss the necessary functionality to successfully
run. However, it was still possible to launch these base classes through
the launchers or straight up constructing them. They would immediately
fail when launched, but they would create a node in the provenance graph
nonetheless. To prevent this from happening, a check is placed in the
constructors to make sure that it is actually a sub class that is being
constructed.

### DIFF
--- a/aiida/backends/tests/engine/test_work_chain.py
+++ b/aiida/backends/tests/engine/test_work_chain.py
@@ -21,9 +21,11 @@ from tornado import gen
 
 from aiida import orm
 from aiida.backends.testbase import AiidaTestCase
+from aiida.common import exceptions
 from aiida.common.links import LinkType
 from aiida.common.utils import Capturing
-from aiida.engine import ExitCode, Process, ToContext, WorkChain, if_, while_, return_, run, run_get_node
+from aiida.engine import ExitCode, Process, ToContext, WorkChain, if_, while_, return_, run, run_get_node, submit
+from aiida.engine import launch
 from aiida.engine.persistence import ObjectLoader
 from aiida.manage.manager import get_manager
 from aiida.orm import load_node, Bool, Float, Int, Str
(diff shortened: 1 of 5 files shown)


aiida/backends/tests/engine/test_work_chain.py
```diff
@@ -294,977 +296,994 @@ def tearDown(self):
...
+            launch.run_get_pid(WorkChain)
 
-            def s3(self):
-                test_case.assertEquals(self.ctx.r1.out.res, B)
-                test_case.assertEquals(self.ctx.r2.out.res, B)
+        with self.assertRaises(exceptions.InvalidOperation):
+            launch.submit(WorkChain)
 
-        run_and_check_success(Wf)
+#     def test_run(self):
```
giovannipizzi: Is this huge number of lines replaced with comments expected?

sphuber: No, I forgot to uncomment, I will fix it

sphuber: Should be fixed now

sphuber:
I canceled them since I will have to rebase this anyway after the others are merged, and so I am not blocking the build queue

PR #2567, opened 2019-03-04
Add efficient `count` method for `aiida.orm.groups.Group`

### LINKED ISSUE #1948: Group needs efficient implementation of number of nodes
Especially in for SqlAlchemy, the typical approach taken is to call `len(group.nodes)` which is horribly ineffective as it has to load all nodes into memory. Instead, it should have a backend specific implementation.

sphuber:
Fixes #1948 

Before the only way to get the number of entries in a `Group` was
through the iterator `nodes` and call `len()` on it. However, this is
almost always not the fastest way. Given that the count is used often in
for example `verdi group list -C` it is worthwhile to have a `count`
method that operates as direct on the database layer as possible.

### DIFF
--- a/aiida/orm/implementation/sqlalchemy/groups.py
+++ b/aiida/orm/implementation/sqlalchemy/groups.py
@@ -124,6 +124,15 @@ class SqlaGroup(entities.SqlaModelEntity[DbGroup], BackendGroup):  # pylint: dis
         self._dbmodel.save()
         return self
 
+    def count(self):
+        """Return the number of entities in this group.
+
+        :return: integer number of entities contained within the group
+        """
+        from aiida.backends.sqlalchemy import get_scoped_session
+        session = get_scoped_session()
+        return session.query(self.MODEL_CLASS).join(self.MODEL_CLASS.dbnodes).filter(DbGroup.id == self.pk).count()
+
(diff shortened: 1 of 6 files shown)


aiida/orm/implementation/sqlalchemy/groups.py
```diff
@@ -124,6 +124,15 @@ def store(self):
...
         return self
 
+    def count(self):
+        """Return the number of entities in this group.
+
+        :return: integer number of entities contained within the group
+        """
+        from aiida.backends.sqlalchemy import get_scoped_session
+        session = get_scoped_session()
+        return session.query(self.MODEL_CLASS).join(self.MODEL_CLASS.dbnodes).filter(DbGroup.id == self.pk).count()
```
giovannipizzi: Can't this be just piped to len(self.nodes)? This, below, should do the same thing? (see __len__ in the property returned by `nodes`?

sphuber: That's what I was originally using, but at least for the SqlA implementation it was a factor 8 slower on my production database

giovannipizzi: ok

PR #2576, opened 2019-03-06
Fix test for multiple log imports

### LINKED ISSUE #2575: Failing test for multiple imports of log entries
In a recent build on travis I just had one of the recently added tests for log import fail:
```
======================================================================
FAIL: test_multiple_imports_for_single_node (aiida.backends.tests.test_export_and_import.TestLogs)
Test multiple imports for single node with different logs are imported correctly
----------------------------------------------------------------------
  File "/home/travis/build/aiidateam/aiida_core/aiida/backends/tests/test_export_and_import.py", line 2253, in test_multiple_imports_for_single_node
    self.assertEqual(imported_log_uuid, log_uuids[0])
AssertionError: 'b8843d76-4307-47bc-9677-6dc486d62fc5' != 'cb8b97f3-7135-4301-b872-ec6d295e410a'
- b8843d76-4307-47bc-9677-6dc486d62fc5
+ cb8b97f3-7135-4301-b872-ec6d295e410a
```

I had to restart the build as it was an active PR. Let's monitor if we see this happen again or it was a one-off or a random error.

CasperWA:
Fixes #2575.

`Log` UUIDs were retrieved using `QueryBuilder`, which means that they may be retrieved in the wrong order, concerning later asserts.
The `Log` UUID retrieval is now split up.

Prettified `test_export_and_import.py` with `@with_temp_dir` decorator implemented by @ltalirz.

Prettified `test_export_and_import.py` by removing unneccessary imports in test functions for nodes, which can be used directly from `orm`.

### DIFF
--- a/aiida/backends/tests/test_export_and_import.py
+++ b/aiida/backends/tests/test_export_and_import.py
@@ -27,6 +27,7 @@ from aiida import orm
 from aiida.backends.testbase import AiidaTestCase
 from aiida.common.utils import get_new_uuid
 from aiida.orm.importexport import import_data, export
+from aiida.backends.tests.utils.configuration import with_temp_dir
 
 
 class TestSpecificImport(AiidaTestCase):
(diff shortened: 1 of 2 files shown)


ltalirz:
@CasperWA Looks good to me. Can you please paste the link to the line for comparing the versions wrt ordering? 
I didn't manage to find it (where I looked, the original code seemed already robust against reordering).
Once I've had a quick look, I'll approve

CasperWA:
> Can you please paste the link to the line for comparing the versions wrt ordering?
> I didn't manage to find it (where I looked, the original code seemed already robust against reordering).

Sure. The original line(s) can be found [here](https://github.com/aiidateam/aiida_core/pull/2576/files#diff-ab96b6b006ca6bca977ed1e29b7ddfffL2230).
The test does the following:
Create two export files, one after the other. In the second two more `Log`s (for a total of three) are added to a `CalculationNode`.
Since they are made using the `logger`, I cannot easily retrieve the `Log`'s UUIDs.
Originally, I chose to use the `QueryBuilder` to retrieve all the UUIDs at once, but since I import the files in steps, checking the database contents in between the steps, I cannot be sure that the first `Log` UUID the `QueryBuilder` retrieves is also the first `Log` message that was added to the node.
Therefore, the fix I did was to create a second `QueryBuilder`, exclusively for the first `Log` [here](https://github.com/aiidateam/aiida_core/pull/2576/files#diff-ab96b6b006ca6bca977ed1e29b7ddfffR2051).

sphuber:
Since you are after the logs in the order that they are created, you could always opt to use a single query builder but then sort on creation time.

CasperWA:
> Since you are after the logs in the order that they are created, you could always opt to use a single query builder but then sort on creation time.

Great Scott! Good idea. However, I think for this specific test I will not change things anymore. Unless of course you or anyone else insists.

PR #2581, opened 2019-03-06
Redefine the structure of the `CalcInfo.local_copy_list`

### LINKED ISSUE #2573: Redesign the interface of  the `local_copy_list` of the `CalcInfo` data structure
Currently, the developer of a `CalcJob` plugin is expected to define the `local_copy_list` and assign them to the `CalcInfo` data structure that is returned by the `CalcJob.prepare_for_submission` method. This `local_copy_list` should be a list of tuples, where each tuple specifies a local absolute path and a remote relative path. However, with the upcoming change of the repository, it is no longer guaranteed to have a local absolute path on the file system for node files and so the interface only provides methods to get a filelike object or the complete content.

sphuber:
Fixes #2573 

With the upcoming change of the node repository, files within it can no
longer be assumed to necessarily be stored on the local file system. It
is therefore impossible to address files within it through an absolute
file path, however, that is exactly what the `local_copy_list` expects
as the first item of its tuples.

Now, the `Node` class only exposes methods to get either the content of
a repository file or an filelike object. To comply with this change in
design the structure of the `local_copy_list` tuples is changed to
tuples of length three where each element represents:

 * UUID of the node
 * Relative key of the file within the node repository
 * Relative path within the working directory where to copy the file

The `upload_calculation` function of the execmanager has been updated to
this interface change, but because our `Transport` interface does not
yet provide a method to put an object from a filelike object, we have to
create a temporary file on the local file system whose absolute filepath
can be passed to the `Transport.put` call. Once the transport is updated
to provide a `put_object_from_filelike`, this inefficiency can be removed.

### DIFF
--- a/aiida/calculations/plugins/templatereplacer.py
+++ b/aiida/calculations/plugins/templatereplacer.py
@@ -137,7 +137,7 @@ class TemplatereplacerCalculation(CalcJob):
                 raise exceptions.InputValidationError("You are asking to copy a file link {}, "
                                                       "but there is no input link with such a name".format(link_name))
             if isinstance(fileobj, orm.SinglefileData):
-                local_copy_list.append((fileobj.get_file_abs_path(), dest_rel_path))
+                local_copy_list.append((fileobj.uuid, fileobj.filename, dest_rel_path))
             elif isinstance(fileobj, orm.RemoteData):  # can be a folder
                 remote_copy_list.append((fileobj.computer.uuid, fileobj.get_remote_path(), dest_rel_path))
             else:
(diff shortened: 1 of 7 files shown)


sphuber:
@ltalirz and @yakutovicha if this PR gets merged, we need to update the guide for the implementation of `CalcJob` classes. After this fix, for a given node, to copy one of the files in its repositories, one can do:
```
folder = self.inputs.folder # FolderData
singlefile = self.inputs.single # SinglefileData
local_copy_list = [
    (folder.uuid, 'relative/path.txt', 'relative/destination.txt')
    (single.uuid, single.filename, 'input_file.upf')
]
```

sphuber:
The failing test on Jenkins is a known bug that Casper is addressing in PR #2575

PR #2593, opened 2019-03-12
Release `v1.0.0b1`

sphuber:


setup.json
```diff
@@ -1,221 +1,220 @@
 {
-  "name": "aiida-core",
-  "version": "1.0.0a4",
-  "url": "http://www.aiida.net/",
-  "license": "MIT License",
-  "author": "The AiiDA team",
-  "author_email": "developers@aiida.net",
-  "include_package_data": true,
+  "name": "aiida-core",
```
ltalirz: hm... what is adding the space here? is this the `validate_consistency.py` script?

sphuber: Yes this was done automatically, I think it is the json dumper that is being used

ltalirz: we could use [this](https://stackoverflow.com/a/16311587)

sphuber: If it bothers you, I am fine with adding it

ltalirz: yes, please!

sphuber: Everything else ok? Like to do all necessary changes in one go, so we don't have to run too many builds

ltalirz: I looked through everything and did not see any other problems

sphuber: Done and done

PR #2596, opened 2019-03-13
Enhanced graphviz functionality

### LINKED ISSUE #2585: verdi graph generate should create a pdf by default
Of all people I've seen using `verdi graph generate`, 0% wanted to edit/work with the .dot file and 100% wanted the pdf.
To me, this means we should provide the PDF by default.

chrisjsewell:
fix #2585

So I don't expect this to be merged any time soon,
since all the tests would need to be rewritten (plus style fixes).
But the basic idea is I've:

1. Used the [python-graphviz](https://graphviz.readthedocs.io/) package which;
   - simplifies the code and make it more 'pythonic'. 
   - has a `_repr_svg_()` method for rendering in the notebook. 
2. Rewritten the function as a class, which
   - maintains a cache of the added nodes and graphs (as sets)
   - splits up the function and allows for multiple additions to the same graph
   - has separate `_style_` functions for different node classes

Here are two examples of using it in the notebook.

The first is a replication of the original function, with some extras.

Screen Shot 2019-03-13 at 20 47 04

The second is new functionality, that was the original reason I started writing this.
It visualizes a root node and filtered connecting nodes, without any intermediate connections.

Screen Shot 2019-03-13 at 21 27 36


chrisjsewell:
Oh and I've just noticed the actual name of the file would need to be change
(from graphviz.py), so there is not an import clash

chrisjsewell:
and if you were to get really fancy;
plugins could provide something like `aiida.graph.style` entry_points, to define how any of their node classes are styled in the graph (e.g. adding class specific information to the label)

ltalirz:
@chrisjsewell very cool!
I would propose to schedule a discussion on this during the [plugin migration week](http://www.aiida.net/aiida-plugins-migration-workshop-2019/) unless others want to move forward with this even before.
mentioning @giovannipizzi @sphuber

giovannipizzi:
Very nice! Indeed let's discuss in person - I'd also like to brainstorm if we can unify in a single place the routines to describe a node (e.g. we now have `get_desc()`that should probably move out of the Node subclasses, we should unify with the generation of text-only (for Graphviz), HTML string (for Materials Cloud), information on what to visualise (for Materials Cloud), ...

I.e., we should probably have a "visualiser" class that for every class know what an how to visualise it (and I see some possibilities to merge with @chrisjsewell idea of 'style' entry points).
We should also look at how jupyter provides multiple output formats when displaying classes (text, html, ...) to get inspiration.

chrisjsewell:
@ltalirz @sphuber  @giovannipizzi I have updated this module to be compatible with the current `develop` branch (i.e. v1.0.0b2). I have also streamlined (to some extent) the mapping of:

1. `LinkType`'s to graphviz edge styles
2. `Data.class_node_type` to graphviz node styles
3. Separate graphviz node styles for each process node type

Below, I attach a notebook and screenshot, replicating your `AddAndMultiplyWorkChain` example:

[Provenance_Graphviz.ipynb.txt](https://github.com/aiidateam/aiida_core/files/3091835/Provenance_Graphviz.ipynb.txt)

image

chrisjsewell:
All the tests are now passing, except Jenkins seems to randomly fail with:

```
django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.
```

sphuber:
That\s because a previous commit changed the module containing the django settings from a folder to a file, but due to the cache the folder is still there, albeit empty. Python will then wrongly import that instead of the file, causing the settings for django not to be set. We'd have to delete the build on Jenkins and restart it from scratch

giovannipizzi:
@chrisjsewell could you do this change and push?
```diff
diff --git a/.ci/Jenkinsfile b/.ci/Jenkinsfile
index b144e9506..5b2b43565 100644
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -160,6 +160,7 @@ pipeline {
         always {
             // Some debug stuff
             sh 'whoami ; pwd; echo $TEST_AIIDA_BACKEND'
+           cleanWs()
         }
         success {
             echo 'The run finished successfully!'
```
The first rebuild will still fail, but at the end it should clear the workspace. Then the second build should (hopefully) work. You can e.g. leave an additional trailing space in the first commit, then remove it, do a `commit --amend` and force push to trigger the second build.

Let us know if this works! (This also will help cleaning up the workspace, now it's almost at 100GB!)

ltalirz:
@chrisjsewell @sphuber @giovannipizzi I wonder, since we are making these changes now, could we converge on a consistent color scheme?

I.e. either the one used [in the docs](https://aiida-core.readthedocs.io/en/latest/concepts/provenance.html#graph-links) or some mix of it with the one [from the web site](http://www.aiida.net/)?

That would be really helpful going forward.

sphuber:
I fully agree with @ltalirz . In all the new sections for the docs that I have written, I have really paid a lot of attention to be very consistent, i.e. with respect to different node colors, node shapes, link line styles, link labels. I am by no means dead set on the colors that I have chosen. All the images are made using SVG the original of which are included in the `docs` tree, so can be adapted (potentially even scripted).

szoupanos:
Thats a good point @ltalirz. I also find a good idea to homogenise the colors & shapes

chrisjsewell:
> @chrisjsewell could you do this change and push?

@giovannipizzi yep that appears to have sorted it ta

chrisjsewell:
My thoughts...

# Link Styles

Similar to [the docs](https://aiida-core.readthedocs.io/en/latest/concepts/provenance.html#graph-links),
the colours are optional:

```python
  {
        LinkType.INPUT_CALC: {
            "style": "solid",
            "color": #000000"  # black
        },
        LinkType.INPUT_WORK: {
            "style": "dashed",
            "color": "#0000FF" # blue
        },
        LinkType.CALL_CALC: {
            "style": "dotted",
            "color":  "#000000"  # black
        },
        LinkType.CALL_WORK: {
            "style": "dotted",
            "color": "#0000FF" # blue
        },
        LinkType.CREATE: {
            "style": "solid",
            "color": "#006400" # green
        },
        LinkType.RETURN: {
            "style": "dashed",
            "color": "#006400" # green
  }
```

# Node Styles

## Shapes

Available shapes for graphviz can be found [here](https://www.graphviz.org/doc/info/shapes.html)

I thought about using the shapes in [the docs](https://aiida-core.readthedocs.io/en/latest/concepts/provenance.html#graph-links). However, triangles aren't great for fitting text in. Hence, I'd go for something like:

- Data: ellipse
- CalcJobNode: rectangle
- CalcFunctionNode: parallelogram
- WorkChainNode: octagon
- WorkFunctionNode: hexagon

## Colors

I was using fill colors to denote process success/failure, but you could switch this and use the outline color. Either way the *natural* colours for success/fail are green/red, so perhaps colors for nodes would be:

- Data: #FFFFF0 (ivory) or #D3D3D3 (light grey)
- CalcJobNode: #87CEFA (light blue)
- CalcFunctionNode: #87CEFA (light blue)
- WorkChainNode: #FFA500 (orange)
- WorkFunctionNode: #FFA500 (orange)

chrisjsewell:
> @giovannipizzi yep that appears to have sorted it ta

it broke again on the next commit, doh

giovannipizzi:
Ups... I think if there are concurrent builds it uses a different folder name internally, so that one probably wasn't cleaned... After Easter I'll try to check if it's safe to just clean up the whole workspace of Jenkins.

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,685 @@
...
+    elif class_node_type == "data.code.Code.":
+        sublabel = "{}@{}".format(os.path.basename(node.get_execname()), node.get_computer_name())
+    elif class_node_type == "data.singlefile.SinglefileData.":
+        sublabel = node.filename
+    elif class_node_type == "data.remote.RemoteData.":
+        sublabel = "@{}".format(node.get_computer_name())
+    elif class_node_type == "data.structure.StructureData.":
+        sublabel = node.get_formula()
+    elif class_node_type == "data.cif.CifData.":
+        formulae = [str(f).replace(" ", "") for f in node.get_attribute('formulae', []) if f]
```
ltalirz: This gives an error, when `formulae` is explicitly set to `None`, e.g.
```
In [1]: n = Dict()

In [2]: n.get_attribute('test', '1')
Out[2]: '1'

In [3]: n.get_attribute('test', [])
Out[3]: []

In [4]: n.set_attribute('test', None)

In [5]: n.get_attribute('test', [])  # returns None
```
This is the same behavior as with `dict` and `get`.

One could argue that `CifData` *should* not have `formulae` set to `None` but they exist in the wild and so we need to deal with them.

The solution would be simply
```
formulae = [str(f).replace(" ", "") for f in node.get_attribute('formulae') or [] if f]
```
(same for next line)

ltalirz:
@chrisjsewell sorry for the long wait - I'll be looking into this now. It would be great if we could get your nice contribution merged in time for it to be used during the May tutorial.

Before I get into the style discussion, there's one thing I'd like you to ask to look into:

Here is an export file (migrated to version 0.4): [export.tar.gz](https://github.com/aiidateam/aiida_core/files/3134128/export.tar.gz)

 * with the current `develop`, `verdi graph generate 31317b2e-d126-4fdc-a2a8-80f1a9aa773b` works fine
 * with your branch, `verdi graph generate 31317b2e-d126-4fdc-a2a8-80f1a9aa773b` hangs

I didn't leave it to run for a very long time but I have the impression it's getting into an infinite loop (might be wrong).

Once this is fixed,  I can use it to play around with the rendering and prepare my suggestions for link and node styles.

CasperWA:
Hey @chrisjsewell, how is the work coming along on this PR? Do you need help/should someone else take over to investigate the problem mentioned by @ltalirz? :)

chrisjsewell:
Hey @ltalirz and @CasperWA, sorry for not getting back sooner, bit busy at the moment!

FYI, I have currently implemented this module in
[`aiida_crystal17.graph.Graph`](https://aiida-crystal17.readthedocs.io/en/latest/apidoc/aiida_crystal17.html#aiida_crystal17.graph.Graph), for my own use, for example in my
[documentation tutorial](https://aiida-crystal17.readthedocs.io/en/latest/user_guide/calc_main.html#analysis-of-outputs). Its working fine for me, but obviously should really have a test module dedicated to it (not just the command-line part).

I have just merged in the latest `develop` and implemented @ltalirz's suggested fix, so perhaps you want to check if it still hangs?

> It would be great if the python API for manipulating graphs would take both uuids and PKs

This should also be fixed it the latest commit

aiida/tools/visualization/graph.py
chrisjsewell: Implemented in https://github.com/aiidateam/aiida_core/pull/2596/commits/2c0899046f6b20065777b9ef325e2b2de8b0e075

ltalirz:
@chrisjsewell thanks a lot for moving forward with this.
next week unfortunately it will be *us* who are busy with the tutorial.
will get back to you after!

ltalirz:
@chrisjsewell Just confirmed that `verdi graph generate 31317b2e-d126-4fdc-a2a8-80f1a9aa773b` is still hanging.
Should be very simple to reproduce for you:

```
verdi import https://github.com/aiidateam/aiida_core/files/3134128/export.tar.gz
verdi graph generate 31317b2e-d126-4fdc-a2a8-80f1a9aa773b
```

chrisjsewell:
@ltalirz in https://github.com/aiidateam/aiida_core/pull/2596/commits/175bfa970c2d530545164032fabf471ac58b29bf, I added functionality to print the graph traversal process:

```console
$ verdi graph generate 31317b2e-d126-4fdc-a2a8-80f1a9aa773b -p -d 2
Info: Initiating graphviz engine: dot
Info: Recursing ancestors, max depth=None
Info: - Depth: 1
Info:   32 -> [22, 25, 46]
Info: - Depth: 2
Info:   25 -> [14, 22, 46, 64, 82]
Info: - Depth: 3
Info: Recursing descendants, max depth=2
Info: - Depth: 1
Info:   32 -> [45]
Info: - Depth: 2
Info:   45 -> [19, 33, 38, 50, 53, 59, 85]
Success: Output file to 32.dot.pdf
```

From that, you could see what was happening was that, for multiple incoming links to a node, it was duplicating the same traversal of descendants, ~exponentially increasing the processing. 
This is fixed in https://github.com/aiidateam/aiida_core/pull/2596/commits/e58eac5f0e530cb9be3f819bb402e2de29a94ed3

chrisjsewell:
In https://github.com/aiidateam/aiida_core/pull/2596/commits/e8623ff768b1ba48a0a7bfc0b651b90a9c2cfd14, you can now create graphs with only certain link types, e.g.

`verdi graph generate 31317b2e-d126-4fdc-a2a8-80f1a9aa773b -p -l logic`

ltalirz:
Dear @chrisjsewell - first of all sincere apologies for the delay in processing this.
Anyhow, I would say let's now make progress with this to get this into 1.0 :-)

I have checked your new branch and things seem to work fine. A few minor things I noticed:

 * when choosing `-e circo`, the output file is still called `32.dot.pdf`
 * This is the output from `verdi graph generate -e circo 31317b2e-d126-4fdc-a2a8-80f1a9aa773b`
[32.dot.pdf](https://github.com/aiidateam/aiida_core/files/3328891/32.dot.pdf)
   Is this expected?

Before I start reviewing the code itself, I suggest we discuss the style of representation (will be my next comment)

ltalirz:
## Link styles

Using specific colors both for nodes and for links can quickly make graphs look messy. Also, colors can be difficult to see for a thin arrow.

I see the upside of distinguishing between the "data provenance" links and the workflow links and one could consider using two colors (black & gray) for this. But I think distinguishing this using line style (as done in the docs) would probably also already be sufficient.

I.e. I vote for one single link color: black
```
{
        LinkType.INPUT_CALC: {
            "style": "solid",
            "color": "#000000"  # black
        },
        LinkType.INPUT_WORK: {
            "style": "dashed",
            "color": "#000000"  # black
        },
        LinkType.CALL_CALC: {
            "style": "dotted",
            "color":  "#000000"  # black
        },
        LinkType.CALL_WORK: {
            "style": "dotted",
            "color":  "#000000"  # black
        },
        LinkType.CREATE: {
            "style": "solid",
            "color":  "#000000"  # black
        },
        LinkType.RETURN: {
            "style": "dashed",
            "color":  "#000000"  # black
  }
```
## Node styles

I would vote to keep the shapes for 
 - `Data`: ellipse
 - `CalcJobNode`: rectangle
 - `Code`: parallelogram

Given those, we can think whether/how to distinguish the rest

 - `WorkChainNode`: What about using a rectangle with a dashed border?
 - `CalcFunctionNode`/`WorkFunctionNode`: I think it's enough to distinguish those by color, i.e. : rectangle

The main things we want to distinguish are calculations, data and codes.


I'll now look into playing with the colors and perhaps come up with a suggestion.

chrisjsewell:
Hey @ltalirz, yeh no problem. As I mentioned, I currently have a copy of the module within [`aiida-crystal17.graph`](https://github.com/chrisjsewell/aiida-crystal17/blob/develop/aiida_crystal17/graph.py), which I've been using quite a lot, and haven't found any issues (apart from the fixes I've made).

Heres an example of a `PwRelaxWorkChain` run:

```python
wkchain_relax = orm.load_node(5268)
graph = Graph(graph_attr={'size': "20,20!", "rankdir": "LR"})
graph.add_node(wkchain_relax)
graph.recurse_ancestors(wkchain_relax, include_process_outputs=True, annotate_links="both", )
graph.recurse_descendants(wkchain_relax, include_process_inputs=True, annotate_links="both")
graph.graphviz.render("graphviz-pw-relax-example", format="png", cleanup=True)
```

graphviz-pw-relax-example


On style: 

- For the links, yes the colors aren't as necessary (could make them optional?). The most important thing, is that `INPUT`/`CREATE` are solid and  `CALL`/`RETURN` are dashed/dotted.
- For the nodes, I really prefer `Data` to be rectangle, `Calculations` ellipse and `WorkFlows` hexagon
- **TODO** excepted processes should be red, and perhaps processes still running should be orange (or different colors for waiting/queued/paused/...) 

> when choosing -e circo, the output file is still called 32.dot.pdf

Yeh at the moment they all have a `.dot.pdf` extension, although this could be changed? That is (I believe) how `circo` should look. To be honest, I haven't really found any use case for the other engines, apart from `dot`.

ltalirz:
I haven't checked the state of the viz code before your PR but I have to say that in your version playing around with the style settings was really a breeze - well done!

Playing around with the *shapes* I realized of course why you chose the ones you did.
Since the textual content of our nodes tends to be quite a bit wider than high, in my opinion the only really usable shapes are `rectangle` and `ellipse`. All other shapes tend to add significant space overhead (do you agree?).

If we choose a good color scheme, I think one can actually get away with just using rectangles and ellipses. See below the comparison between the original scheme and my draft

[original.pdf](https://github.com/aiidateam/aiida_core/files/3329315/orig.pdf)
[draft_tal_v1.pdf](https://github.com/aiidateam/aiida_core/files/3329317/tahoma.pdf)

The draft includes some modifications in the font and font size and was generated with

```dot -Tpdf -Nfontname=Tahoma -Efontname=Tahoma -Efontsize=10 32.dot.dot > tahoma.pdf```

I've pushed the modifications I made to my fork https://github.com/ltalirz/aiida_core/tree/graphviz-experiments

@giovannipizzi @sphuber your input on the style would be very welcome as well

**Comment by ltalirz** [comment]
P.S. I've taken the colors from Sebastiaan's [concepts section](https://aiida-core.readthedocs.io/en/latest/concepts/provenance.html) but I do find them a bit "dark".
Happy to make a version with a "lighter" color scheme if you agree.

**Comment by ltalirz** [comment]
@chrisjsewell While we wait for feedback from the others, two more questions from my side:

1. One alternative way of getting around the "width problem" would be an approach like the one taken by the provenance browser, where the text is simply displayed below the node (see e.g. [here](https://www.materialscloud.org/explore/2dstructures/details/2ac016d5-807e-454f-9bf7-865f7bce5080?nodeType=DATA). Is there a straightforward way of doing this in graphviz?

2. I think we'll probably want to control the font/fontsize in the output but it seems to me the graphviz wrapper currently does not have support for this (?) See e.g. [here](https://github.com/xflr6/graphviz/blob/master/graphviz/files.py#L168). We could consider making a PR - or we could go via the `dot` command line utility

**Comment by sphuber** [comment]
Regarding the style, I really like the look already of the `draft_tal_v1.pdf` that @ltalirz posted. I think it would be very useful if the style of these graphs and the documentation are as similar as possible. As far as I can see these are the remaining differences:

* Shapes: in the documentation we have circles for data and here ellipses. I think this is fine as the circle is just a special case of an ellipse
* Same for the `CalcJobNodes`: squares are a special kind of rectangle so is congruent.
* The `WorkChainNodes` are squares rotated by 90 degrees in the documentation. I can see how this is not practical in the generated graphs. If you guys think the coloring difference is sufficient, I would be fine with adapting the documentation SVG's.
* Difference between process functions. In documentation, I use triangles which again is clearly not usable for generated graphs that require more verbose labels. For the generated graphs I think I like the solution to use a transparent node and keeping the color and shape the same. Also here, I am in principle fine with applying these changes to the documentation images.
* Finally the special treatment of `Code` nodes. I can see why one would want a different style, but this might potentially be confusing as `Code` is simply another type of `Data` node. Why should it get a different style and others not. If you guys think this is not a problem and the distinction is more important then I am happy with the change. I would also use the blue color that it has now and keep the ellipse shape as it is a data node after all.

I would keep the current color scheme, but always open to consider alternatives.

**Comment by chrisjsewell** [comment]
> I think we'll probably want to control the font/fontsize in the output but it seems to me the graphviz wrapper currently does not have support for this (?) 

It certainly does! You use `global_node_style`.

```python
graph = Graph(graph_attr=
              {'size': "10,10!", "rankdir": "LR"}, 
              global_node_style={"fontname": "Tahoma", "fontcolor": "blue", "fontsize": "10"})
```

The attributes correspond to the table at https://www.graphviz.org/doc/info/attrs.html, `graph_attr` is any of the names with `Used By = G`, and `global_node_style` for `Used By = N`

chrisjsewell:
| playing around with the style settings was really a breeze

Thanks :) and I'd note that, from a user perspective, its also quite easy to change the default styles, e.g.

```python
def data_style_func(node):
    return {"shape": "triangle",
            "color": "black", "style": "solid"}

graph = Graph(graph_attr=
              {'size': "10,10!", "rankdir": "LR"},
             data_styles=data_style_func)
```

ltalirz:
Very cool. Is there also `global_edge_style`?

By the way, really like the `-l` option on the command line to limit to data/logical provenance only :)

chrisjsewell:
> Very cool. Is there also `global_edge_style`?

There is now

chrisjsewell:
**TODO** data nodes like `data.BaseType`, `CifData` and `UpfData` should be given `get_description` methods (outside of this PR). Then `graph.default_data_sublabels` could be shortened.

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,727 @@
...
+    types, and also color the nodes for successful/failed processes
+
+    The first line of the node text is always '<node.name> (<node.pk>)'.
+    Then, if ``include_sublabels=True``, subsequent lines are added,
+    which are node type dependant.
+
+    """
+    # pylint: disable=too-many-arguments
+    node_style = {}
+    if isinstance(node, BaseFactory("aiida.node", "data")):
```
ltalirz: Since this is using entry points, I'm not sure whether it's a good idea to create the factory for every node... will this be cached?
Anyhow, you could also just `from aiida.orm import Data`

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,727 @@
...
+
+        node_style = data_style_func(node)
+        label = ["{} ({})".format(node.__class__.__name__, node.pk)]
```
ltalirz: Another really needed feature would be to use (partial) uuids here instead of PKs, since verdi graphs that include PKs cannot be shared between people (see #2636, #2651).

Would you be willing to add an option to allow switching between uuid & pk?
No need to worry for now how exactly to get the right subset of the UUID, just have one function take care of this that, for now, takes the first block of a UUID (8 characters).

aiida/tools/visualization/graph.py
```diff
@@ -180,7 +180,7 @@ def default_process_styles(node):
     node_style = process_map.get(class_node_type, default)
 
     # style process node, based on success/failure of process
-    if node.is_failed:
+    if node.is_failed or node.is_excepted:
```
sphuber: What about `killed` then?

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,740 @@
...
+        :type global_edge_style: dict or None
+        :param include_sublabels: if True, the note text will include node dependant sub-labels (Default value = True)
+        :type include_sublabels: bool
+        :param link_styles: callable mapping LinkType to graphviz style dict;
+            link_styles(link_type) -> dict (Default value = None)
+        :param data_styles: callable mapping data node to a graphviz style dict;
+            data_styles(node) -> dict (Default value = None)
+        :param process_styles: callable mapping process node to a graphviz style dict;
+            process_styles(node) -> dict (Default value = None)
+        :param data_sublabels: callable mapping data node to a sublabel (e.g. specifying some attribute values)
```
ltalirz: Is there a reason to limit this to data nodes?

If somebody wants to overwrite the behavior, I can imagine they might also want to do it for other types of nodes (?)
If not, this parameter could be called `sublabel_fn`

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,740 @@
...
+            data_sublabels(node) -> str (Default value = None)
+
+        """
+        # pylint: disable=too-many-arguments
+        self._graph = Digraph(engine=engine, graph_attr=graph_attr)
+        self._nodes = set()
+        self._edges = set()
+        self._global_node_style = {}
+        if global_node_style:
```
ltalirz: what about
```suggestion
self._global_node_style = global_node_style or {}
```
(same below)

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,740 @@
...
+        else:
+            self._data_sublabels = default_data_sublabels
+
+    @property
+    def graphviz(self):
+        """return a copy of the graphviz.Digraph"""
+        return self._graph.copy()
+
+    @property
+    def nodes(self):
```
ltalirz: I guess this copying is also actually needed where `.nodes` is used?
I.e. there is no excessive copying going on?

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,740 @@
...
+        :type link_pair: None or aiida.orm.utils.links.LinkPair
+        :param style: graphviz style parameters (Default value = None)
+        :type style: dict or None
+        :param overwrite: whether to overrite existing edge (Default value = False)
+        :type overwrite: bool
+
+        """
+        in_node = self._load_node(in_node)
+        if in_node.pk not in self._nodes:
+            raise AssertionError("the in_node must have already been added to the graph")
```
ltalirz: Should this error message perhaps include the node PK? (same below)

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,740 @@
...
+        _add_graphviz_edge(self._graph, in_node, out_node, style)
+
+    def add_incoming(self, node, link_types=(), annotate_links=False, return_pks=True):
+        """add nodes and edges for incoming links to a node
+
+        :param node: node or node pk/uuid
+        :type node: aiida.orm.nodes.node.Node or int
+        :param link_types: filter by link types (Default value = ())
+        :type link_types: str or tuple[str] or aiida.common.links.LinkType or tuple[aiida.common.links.LinkType]
+        :param annotate_links: label edges with the link 'label', 'type' or 'both' (Default value = False)
```
ltalirz: would `None` perhaps be a better fit for "no label" here?

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,740 @@
...
+            elif annotate_links == "type":
+                style['label'] = link_triple.link_type.name
+            elif annotate_links == "both":
+                style['label'] = "{}\n{}".format(link_triple.link_type.name, link_triple.link_label)
+            self.add_edge(link_triple.node, node, link_pair, style=style)
+            nodes.append(link_triple.node.pk if return_pks else link_triple.node)
+
+        return nodes
+
+    def add_outgoing(self, node, link_types=(), annotate_links=False, return_pks=True):
```
ltalirz: this function shares quite a bit of code and interface with `add_incoming`.
Do we really need two functions here?

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,740 @@
...
+
+    def add_origin_to_target(self,
+                             origin,
+                             target_cls,
+                             target_filters=None,
+                             include_target_inputs=False,
+                             include_target_outputs=False,
+                             origin_style=(),
+                             annotate_links=False):
+        """add nodes and edges from an origin node to target nodes
```
ltalirz: If I understand correctly, it's actually rather
```suggestion
        """Add nodes and edges from an origin node to all nodes of a target node class.
```

ltalirz: In this case, the function should potentially be renamed `add_origin_to_targets` (also to be consistent with the next function)

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,740 @@
...
+    def add_origins_to_targets(self,
+                               origin_cls,
+                               target_cls,
+                               origin_filters=None,
+                               target_filters=None,
+                               include_target_inputs=False,
+                               include_target_outputs=False,
+                               origin_style=(),
+                               annotate_links=False):
+        """add nodes and edges from multiple origin node to target nodes
```
ltalirz:
```suggestion
        """Add nodes and edges from all nodes of an origin class to all node of a target node class.
```

ltalirz:
@chrisjsewell Please don't forget to ping me when you want me to review again. Cheers!

chrisjsewell:
> @chrisjsewell Please don't forget to ping me when you want me to review again. Cheers!
Will do, I’ll hopefully get round to it on Sunday

ltalirz:
thanks a lot @chrisjsewell !
for some reason I no longer see any "resolve" or "comment" buttons
Screenshot 2019-07-04 at 07 41 56
if anybody has an idea as to why, let me know.

I'll go through the changes later today.
By the way @sphuber and @giovannipizzi agreed to start with the default color scheme I added [here](https://github.com/ltalirz/aiida_core/commit/566beb912c27d41ca4f157e772f05311ca5a0a4e) - does this seem reasonable to you as well?

giovannipizzi:
only comment on the color scheme of @ltalirz, can we make the workfunction and the calcfunction a bit more different? e. g. pushing one group of colors more towards the red and one more towards the orange/yellow?

chrisjsewell:
> for some reason I no longer see any "resolve" or "comment" buttons

I was playing around with the new [VS Code GitHub PR](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github) functionality, so maybe I broke something!

But I think if you click on 'View Changes' here (and load diffs) you should be able to see all the comments and buttons.

image

chrisjsewell:
> does this seem reasonable to you as well

Yeh as I've mentioned, I added the functionality, for it to be fairly easy to change styles, at the user level. So I'm not necessarily to bothered what the defaults are, as I may choose not to use them. 
Personally, the edges are fine, but I'd prefer a shape difference between Work processes and Calc processes, and I think there's probably a case for two color "modes": 

- Process analysis: Data = white, Failed Processes = red, Running Processes = orange, Successful processes = green.
- Presentational: with distinct colors for different node types.

You could maybe mix them (using different background/font colors), but that gets a bit messy.

Perhaps I should add an easy way of switching?

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,727 @@
...
+def default_link_styles(link_type):
+    """map link_type to a graphviz edge style
+
+    :param link_type: a LinkType attribute
+    :type link_type: One of the members from
+        :class:`aiida.common.links.LinkType`
+    :rtype: dict
+
+    """
+    return {
```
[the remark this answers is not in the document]
ltalirz: > I thought it was a bad idea to have mutable objects as package constants?

Yes, one should probably make them a frozendict. Anyhow, it seems this isn't really used much in AiiDA at the moment... It's ok for now.

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,727 @@
...
+            "style": "filled",
+            "shape": "hexagon",
+            "pencolor": "black",
+            "fillcolor": "#FFA500"  # light blue
+        }
+    }
+
+    node_style = process_map.get(class_node_type, default)
+
+    # style process node, based on success/failure of process
```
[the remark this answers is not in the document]
ltalirz: > As per the discussion regarding default node styles, I think only shape is required to distinguish node types. 

It is certainly *possible* to use a different shape for each node but my personal impression (seems to be shared by sebastiaan and Giovanni) is that our need for very wide nodes makes the shapes other than ellipse and rectangle either unusable or very contorted.

> The default (white) is for processes that are not finised. Its also just more encouraging to see (hopefully) lots of green nodes :) 

While I share the sentiment, I think most of the time verdi graphs will be used to discuss finished calculations (e.g. when you want to show in a presentation how your workflow looks like). I.e. I think the "failed" state should be the one highlighted.

aiida/tools/visualization/graph.py
ltalirz: fantastic, thanks!

aiida/tools/visualization/graph.py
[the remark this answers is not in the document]
giovannipizzi: > It is certainly possible to use a different shape for each node but my personal impression (seems to be shared by sebastiaan and Giovanni) is that our need for very wide nodes makes the shapes other than ellipse and rectangle either unusable or very contorted.

Indeed. Maybe the only difference, if you really want to have one, is to use rectangles with sharp or rounded edges between e.g. Calc and Work - but other shapes become less good for writing text in them.

One could think to an option to highlight the state of a process e.g. with the color of the edge (green for OK, yellow if being processed, red if failed), I don't know though if this should be on by default or this can just go in the textual description?

chrisjsewell:
Ok, in b0c2be0, I have applied @ltalirz's style format. I have also merged `default_process_styles` and `default_data_styles` into `default_node_styles`, and added a `pstate_node_styles`.

From `aiida/backends/tests/tools/visualization/test_graph.py`, you can create (just uncomment the render line):

`test_graph_graphviz_source`:

```
graph = graph_mod.Graph()
graph.recurse_descendants(nodes.pd0)
```
image

`test_graph_graphviz_source`:

```
graph = graph_mod.Graph(node_style_fn=pstate_node_styles)
graph.recurse_descendants(nodes.pd0)
```
image

aiida/tools/visualization/graph.py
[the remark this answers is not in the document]
ltalirz: > One could think to an option to highlight the state of a process e.g. with the color of the edge

Yes, that's what I suggested above as well.
I suggested to highlight only "failed" ones but since it's just the edge I would be fine with including it for any status as well.

chrisjsewell:
Thanks @ltalirz 

> I'm now getting the following warnings

Oops, there was a typo in specifying styles for `WorkFunctionNode`s (the only process type I didn't include in the tests!), hopefully fixed in 0b5087f

ltalirz:
@chrisjsewell Thanks, for me it's good to merge
Let's give @sphuber and @giovannipizzi a few hours to reply - otherwise I'll go ahead.

I would simply "squash & merge" if that's OK with your.
If you like to keep any topical commits separated, I would ask you to rebase into, say, no more than 5.

chrisjsewell:
Yep "squash & merge" will be fine thanks @ltalirz

aiida/cmdline/commands/cmd_graph.py
```diff
@@ -37,33 +48,49 @@ def verdi_graph():
...
-        exit_status, output_file_name = draw_graph(
-            root_node,
-            ancestor_depth=ancestor_depth,
-            descendant_depth=descendant_depth,
-            image_format=output_format,
-            include_calculation_inputs=inputs,
-            include_calculation_outputs=outputs)
-    except OSError as exception:
-        echo.echo_critical(str(exception))
+    echo.echo_info("Initiating graphviz engine: {}".format(engine))
```
sphuber: Maybe also use the `print_func` here, or do we always want these lines to be printed regardless of `verbose` flag?

aiida/cmdline/commands/cmd_graph.py
```diff
@@ -37,33 +48,49 @@ def verdi_graph():
...
+    graph = Graph(engine=engine, node_id_label=id_label)
+    echo.echo_info("Recursing ancestors, max depth={}".format(ancestor_depth))
```
sphuber: Idem about `print_func` usage

aiida/cmdline/commands/cmd_graph.py
```diff
@@ -37,33 +48,49 @@ def verdi_graph():
...
+@click.option('-f', '--output-format', help="The output format used for rendering ('pdf', 'png', etc.).", default='pdf')
+@click.option('-s', '--show', is_flag=True, help="Open the rendered result with the default application.")
 @decorators.with_dbenv()
-def generate(root_node, ancestor_depth, descendant_depth, outputs, inputs, output_format):
-    """Generate a graph for a given ROOT_NODE."""
-    from aiida.tools.visualization.graphviz import draw_graph
+def generate(root_node, link_types, id_label, ancestor_depth, descendant_depth, process_out, process_in, engine,
+             verbose, output_format, show):
+    """
+    Generate a graph from a ROOT_NODE (specified by pk or uuid).
```
sphuber: I think technically the node label could also be used. Granted it won't be that common. I think it would maybe be better to use the term "identifier". This is used across the CLI to mean one of PK, UUID, or label.

aiida/cmdline/commands/cmd_graph.py
```diff
@@ -15,20 +15,31 @@
...
-    Create visual representations of part of the provenance graph.
-    Requires that `graphviz<https://graphviz.org/download>` be installed.
+    """Create visual representations of part of the provenance graph.
     """
 
 
 @verdi_graph.command('generate')
 @arguments.NODE('root_node')
+@click.option(
+    "-l",
```
sphuber: A nitpick, feel free to ignore. In `aiida-core` we are adhering to the convention to use single quotes for identifier strings, such as the keywords here and where possible in general. For "text"-like string, such as the help messages, which often contain single quotes, it is fine to use double quotes (whichever option leads to the least escaping characters).

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -38,13 +38,8 @@ def get_valid_test_paths():
...
     base_test_modules = [
-        'aiida.cmdline.utils',
-        'aiida.cmdline.params.types',
-        'aiida.cmdline.params.options',
-        'aiida.common',
-        'aiida.schedulers',
-        'aiida.transports',
-        'aiida.tools.dbimporters.plugins',
+        'aiida.cmdline.utils', 'aiida.cmdline.params.types', 'aiida.cmdline.params.options', 'aiida.common',
+        'aiida.schedulers', 'aiida.transports', 'aiida.tools.dbimporters.plugins', 'aiida.tools.visualization'
```
sphuber: I think all the tests are actually in the module `aiida.backends` correct? And I don't see any in `aiida.tools.visualization`, so it is not necessary to add it here

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,777 @@
...
+    """
+    class_node_type = node.class_node_type
+
+    default = {"shape": "rectangle", "pencolor": "black"}
+
+    process_map = {
+        "process.calculation.calcjob.CalcJobNode.": {
+            "shape": "ellipse",
+            "style": "filled",
+            # "pencolor": "black",
```
sphuber: If we are not using the pencolor, maybe get rid of the commented out code

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,777 @@
...
+
+    node_style = process_map.get(class_node_type, default)
+
+    if isinstance(node, ProcessNode):
+        # style process node, based on success/failure of process
+        if node.is_failed or node.is_excepted or node.is_killed:
+            node_style['fillcolor'] = '#de707fff'  # red
+        elif node.is_finished_ok:
+            node_style['fillcolor'] = '#8cd499ff'  # green
+        else:  # specifically look for waiting state?
```
sphuber: Note that this conditional will hit the states `CREATED` and `RUNNING` as well. Think that the same color as `WAITING` is fine, just thought I'd mention it. Maybe update the comment to include them

aiida/tools/visualization/graph.py
```diff
@@ -0,0 +1,777 @@
...
+    (e.g. specifying some attribute values)
+
+    :param node: the node to map
+    :type node: aiida.orm.nodes.node.Node
+    :rtype: str
+
+    """
+    # pylint: disable=too-many-branches
+
+    class_node_type = node.class_node_type
```
sphuber: I take it this code can be cleaned up when we address #3087 and we define a node method that will provide this type of descriptor?

aiida/cmdline/commands/cmd_graph.py
chrisjsewell: Well I guess they are a different level of verbosity. Your `VERBOSE` should probably employ the `count=True` functionality (see [click example](http://click.palletsprojects.com/en/5.x/options/#counting)), then `-v` would be these echos and `-vv` would also be `print_func`. I guess thats not a change for this PR, since it could affect other commands. Shall I switch to creating my own verbose option for now?

sphuber: No not necessary. If this was already there, it's indeed fine to leave it as it this

aiida/cmdline/commands/cmd_graph.py
chrisjsewell: Yeh, it makes. I'll fix a few, but might not get all. FYI, there's an open issue on YAPf to do this google/yapf#399

sphuber: That'd be great indeed to have yapf do this. And don't worry, as I said, just a nitpick, great if you take a few along

aiida/tools/visualization/graph.py
chrisjsewell: Yes indeed. Basically all the node classes should supply their own (default) description, and this function (which is injected into the `Graph` class) can just be used by the user, to override any descriptions they wish.

aiida/tools/visualization/graph.py
chrisjsewell: fixed in 6e599e9

aiida/tools/visualization/graph.py
chrisjsewell: fixed in 6e599e9

aiida/cmdline/commands/cmd_devel.py
chrisjsewell: fixed in 6e599e9

aiida/cmdline/commands/cmd_graph.py
chrisjsewell: fixed in 6e599e9

ltalirz:
@sphuber Looks to me like your points have been addressed
I leave the final merge up to you

sphuber:
Thanks a lot to both !

PR #2599, opened 2019-03-14
Make `Port.non_db` attribute inheritable from `PortNamespace`

### LINKED ISSUE #2598: Make `Ports` inherit `non_db` from the `PortNamespace` to which they are added
The default value for `non_db` is `False`, but often when defining a namespace that has `non_db=True` it is implied that all ports within it should have the same value. Currently one has to explicitly define it every time to override the default, but it would be nice if it could be inherited.

sphuber:
Fixes #2598 

If the `non_db` attribute is not explicitly defined, either upon
construction or through the setter, the value will inherited from the
namespace to which the port is added. This removes the requirement to
specify `non_db=True` for every port explicitly if the parent namespace
is already marked as non storable as well.

### DIFF
--- a/aiida/engine/processes/process.py
+++ b/aiida/engine/processes/process.py
@@ -54,7 +54,7 @@ class Process(plumpy.Process):
     # pylint: disable=too-many-public-methods
 
     _node_class = ProcessNode
-    _spec_type = ProcessSpec
+    _spec_class = ProcessSpec
 
     SINGLE_OUTPUT_LINKNAME = 'result'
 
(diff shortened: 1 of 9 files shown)


aiida/engine/processes/process.py
```diff
@@ -103,12 +103,11 @@ class SaveKeys(enum.Enum):
     @classmethod
     def define(cls, spec):
         super(Process, cls).define(spec)
-        spec.input_namespace(spec.metadata_key, required=False, non_db=True, default={})
+        spec.input_namespace(spec.metadata_key, required=False, non_db=True, valid_type=dict, default={})
```
giovannipizzi: As a note here, also `default={}` is not the right thing to do, because the *same* dictionary will be shared by all instances that use the default.

sphuber: You are right, but actually the `valid_type` I added is also wrong. I was intending to say that the value passed to this namespace should be a `dict`, but what it really is saying is that _each_ entry in the namespace should be of type `dict` which is wrong.

sphuber: The `default` was put there because without it, everything would break :+1:  but I have addressed the underlying issue in `plumpy` and made a new release `0.12.2` which is now included in this fix

PR #2604, opened 2019-03-15
Implement pre-commit hook to automatically format `verdi` docs

### LINKED ISSUE #143: verdi commands documentation
Originally reported by: **Nicolas Mounet (Bitbucket: [mounet](https://bitbucket.org/mounet), GitHub: [nmounet](https://github.com/nmounet))**

----------------------------------------

We should find a way to update dynamically (i.e. automatically from the code) the documentation for the 'verdi' commands.

----------------------------------------
- Bitbucket: https://bitbucket.org/aiida_team/aiida_core/issue/143

sphuber:
Fixes #143 

The hook `utils/validate_consistency verdi-autodocs` will automatically
two blocks of text in the documentation concerning the `verdi` CLI. The
first block is in `working_with_aiida/index.rst` which contains a short
list of the available sub commands, each of which has a link to a
section where more detailed information on its usage and available
parameters is displayed.

These two blocks are marked by restructured text literal markers, of
which the starting marker is a reference marker and the end marker is
just a comment. The script in the hook will find the block between these
markers and replace its content by the new content that is generated by
using the actual `click` commands of `verdi`. This ensures that if a
`verdi` command is updated in a commit, the documentation is
automatically updated.


sphuber:
@giovannnipizzi yes when generating the new blocks I sort the commands alphabetically so unless there are commands renamed the order should not change.

PR #2615, opened 2019-03-19
Try to get defaults from plug-in if not specified in node for `verdi calcjob [in/out]putcat` commands

### LINKED ISSUE #2611: No default input/output file defined by process.calculation.calcjob
Passing an `Arithmetic.Add` `CalcJob` UUID to `verdi calcjob inputcat` or `verdi calcjob outputcat` results in:

```shell
Critical: process.calculation.calcjob does not define a default [input or output] file. Please specify a path explicitly
```

The commands `verdi calcjob inputls` and `verdi calcjob outpuls` work fine, i.e. a path exists, and I see no reason why it shouldn't be able to locate and print the contents of a file in either folder.

CasperWA:
Fixes #2611 

Implemented logic for `verdi calcjob [in/out]putcat NODE path`:

1. Get path from `NODE.get_option('[in/out]put_filename')`
2. If path is `None`: Try: `path = 'aiida.[in/out]'`.
3. IF success: Issue warning that the "default" option was used after showing content of `aiida.[in/out]`
  IF fail: Throw existing critical message that `entry_point.name` does not define a default [in/out]put file.


aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -90,23 +90,31 @@ def calcjob_inputcat(calcjob, path):
     """
     from aiida.plugins.entry_point import get_entry_point_from_class
 
+    # Get path from the given CalcJobNode
     if path is None:
-
         path = calcjob.get_option('input_filename')
```
ltalirz: what about replacing this with
`path = calcjob.get_option('input_filename') or 'aiida.in'`
and getting rid of all this extra logic?

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -90,23 +90,31 @@ def calcjob_inputcat(calcjob, path):
...
 
-        if path is None:
-            cls = calcjob.__class__
-            _, entry_point = get_entry_point_from_class(cls.__module__, cls.__name__)
-            echo.echo_critical('{} does not define a default input file. Please specify a path explicitly'.format(
-                entry_point.name))
+    # Try "default" input_filename value
+    warn = False
+    if path is None:
+        path = "aiida.in"
```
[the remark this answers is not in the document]
CasperWA: No, but it is the default filename usually, if no specific is provided. Therefore also the comment in line 97.

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -90,23 +90,31 @@ def calcjob_inputcat(calcjob, path):
...
+        # No path available or is incorrect
+        cls = calcjob.__class__
+        _, entry_point = get_entry_point_from_class(cls.__module__, cls.__name__)
+        echo.echo_critical('{} does not define a default input file. Please specify a path explicitly.\n'
+                           'Exception: {}'.format(entry_point.name, exception))
     else:
         echo.echo(content)
 
+        if warn:
+            echo.echo_warning('Attribute/Option "input_filename" not found. Used default "aiida.in".\n'
```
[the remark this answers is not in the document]
CasperWA: Correct. The logic here being that if the 'input_filename' option is specified by the plug-in, it would never try to look for `aiida.in`, therefore there would be no need to warn.
But if a user thought he/she had already set this option, this may be helpful in finding out it wasn't set at all, and has instead defaulted to `aiida.in`.

I could see why the warning would be annoying with prolonged usage, and perhaps a note should be added to the critical message above saying something like "also tried the default 'aiida.in' without luck".

aiida/cmdline/commands/cmd_calcjob.py
CasperWA: Good idea.

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -90,23 +90,31 @@ def calcjob_inputcat(calcjob, path):
...
+        warn = True
 
     try:
         content = calcjob.get_object_content(path)
     except Exception as exception:  # pylint: disable=broad-except
-        echo.echo_critical('failed to get the content of file `{}`: {}'.format(path, exception))
+        # No path available or is incorrect
+        cls = calcjob.__class__
+        _, entry_point = get_entry_point_from_class(cls.__module__, cls.__name__)
+        echo.echo_critical('{} does not define a default input file. Please specify a path explicitly.\n'
```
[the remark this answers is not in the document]
CasperWA: I guess that's essentially what the first sentence also means implicitly. But there is no reason why it couldn't state it more transparently, sure.

sphuber:
I am not sure defaulting to `aiida.in` is a good idea, let alone putting that constant in the CLI function. The `CalcJob` does not specify a default `input_filename` as it is not required. Some codes might not need to write it. The `aiida.in` is just what was chosen years ago as the default for the QE plugins. But there is no rule in `aiida-core` that says that this file should be there or should have that value.

CasperWA:
So instead you would simply have it complain if `input_filename` was not defined and the input from the user cannot be found?

CasperWA:
> I am not sure defaulting to `aiida.in` is a good idea, let alone putting that constant in the CLI function. The `CalcJob` does not specify a default `input_filename` as it is not required. Some codes might not need to write it. The `aiida.in` is just what was chosen years ago as the default for the QE plugins. But there is no rule in `aiida-core` that says that this file should be there or should have that value.

Is the case the same for `aiida.out`?

CasperWA:
Have changed it so that it now looks through `node.options`, which accesses the process class directly, i.e. it checks the current version of the plug-in/process class for a default `input_filename` or `output_filename`, before giving up.

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -90,20 +90,25 @@ def calcjob_inputcat(calcjob, path):
...
         path = calcjob.get_option('input_filename')
 
-        if path is None:
-            cls = calcjob.__class__
-            _, entry_point = get_entry_point_from_class(cls.__module__, cls.__name__)
-            echo.echo_critical('{} does not define a default input file. Please specify a path explicitly'.format(
-                entry_point.name))
+    # Get path from current process class of CalcJobNode if still not defined
+    if path is None:
+        fname = calcjob.options.get('input_filename')
```
ltalirz: wow, so `calcjob.get_option` returns something different than `calcjob.options.get`?
@sphuber I think that is potentially very confusing. Should I open an issue for this?

ltalirz:
to me it looks fine - @sphuber what do you think?

aiida/cmdline/commands/cmd_calcjob.py
CasperWA: To me it makes sense if the goal is:
Show me all the options I can set: `calcjob.options`
Give me options that have been set for the specific node: `calcjob.get_option('x')`.

But yeah, the naming is a bit confusing, I agree, even if the above is the intended way it should work.

sphuber: The problem is exactly that. In principle, only `get_option` should exist, i.e. what was the value of this option that was set on this `CalcJobNode`. The `options` property is just a shortcut to loading the `Process` class used in generating the `CalcJobNode` and inspecting the `metadata.options` port from its `ProcessSpec`. Maybe we should remove this shortcut, such that *if* people use it, they realize damn well that it concerns the options of a spec, not option values that were actually set. Removing the shortcut would actually not make the alternative all that bad even. The current implementation of `CalcJobNode.options` is:
```
    @property
    def options(self):
        """Return the available process options for the process class that created this node."""
        try:
            return self.process_class.spec().inputs._ports['metadata']['options']  # pylint: disable=protected-access
        except ValueError:
            return {}
```
But this can actually be simplified a lot to `self.process_class.options` since I already added the `options` property to the `Process` class some time ago. So my proposal would simply be to remove `options` from the `CalcJobNode` class such that people have to do `self.process_class.options` which should hopefully make it clear that these are the abstract options

ltalirz: I think this would be a good solution.

CasperWA:
Updated with new `process_class.spec_options` method introduced in #2736.

CasperWA:
Updated critical message together with @ltalirz.

sphuber:
:+1: thanks Casper

PR #2608, opened 2019-03-16
enable conda test on travis

ltalirz:
@astamminger and me have fixed #2592 

enabling travis test for conda installation also on `develop`

PR #2617, opened 2019-03-20
Fix version requirements for `plumpy` dependency

### LINKED ISSUE #2616: topika/tornado making travis tests fail
I am unsure exactly what is happening, but from [Travis](https://travis-ci.org/aiidateam/aiida_core/jobs/508443058) it seems the cause for the failing tests are due to topika/tornado.

sphuber:
Fixes #2616 

Dependency `plumpy` just released `v0.13.0` which makes it compatible
with `pika==1.0.0b2`, which is expected to be the last beta with
breaking changes before the real release, and with `kiwipy==0.5.0` which
also was just released to be compatible with `pika==1.0.0b2`.
Since `kiwipy` now introduces proper limits on its dependencies `pika`
and `topika`, we no longer have to those explicitly in `aiida-core` but
can simply rely on fixing the version of `plumpy`. The versions of the
four libraries that should work with this version of `aiida-core` are:

 * `plumpy==0.13.0`
 * `kiwipy==0.5.0`
 * `topika==0.2.0`
 * `pika==1.0.0b2`


sphuber:
No, because from `aiida-core`s point of view `plumpy` is the only direct dependency. It then depends on pika and kiwipy. I guess we use some parts of kiwipy directly, albeit sporadically, so we could have added explicitly but it doesn't make a difference

ltalirz:
Aha - I guess there was a reason why #2608 was not merged ;-)

PR #2619, opened 2019-03-21
Fig bug for process functions allowing kwargs

### LINKED ISSUE #2618: process functions with kwargs can fail to link up inputs
It is allowed to use `**kwargs` in the definition of process functions, even in combination with normal named parameters, however, when called, *at least* as many positional parameters as named parameters should be passed. Currently this is not checked leading to problems in the following example:
```
from aiida import engine
from aiida import orm

@engine.calcfunction
def add(**kwargs):
    return Int(sum(kwargs.values))

add(orm.Int(1), orm.Int(2)
```
The function is called with positional arguments and the process will be run and added to the provenance graph, correctly linking the output, however, the inputs are not linked because they didn't have explicit names to be used for the link labels.

sphuber:
Fixes #2618 

A process function that supported `**kwargs` could be called with more
positional arguments then the function signature declares, causing them
to be interpreted as keyword arguments, except that now the engine could
not determine their name, since it is neither in the signature, nor in
the function call and so no links would be created. This bug allowed
a process function to be called that would have missing input links.

To fix this we add an explicit check on call time that if a function is
dynamic, i.e. its signature allows `**kwargs` then the number of
positional arguments may not surpass the number of named parameters
defined in the signature.

### DIFF
--- a/aiida/engine/processes/functions.py
+++ b/aiida/engine/processes/functions.py
@@ -229,8 +229,29 @@ class FunctionProcess(Process):
             })
 
     @classmethod
+    def validate_inputs(cls, *args, **kwargs):  # pylint: disable=unused-argument
+        """
+        Validate the positional and keyword arguments passed in the function call.
+
+        :raises TypeError: if more positional arguments are passed than the function defines
+        """
+        nargs = len(args)
+        nparameters = len(cls._func_args)
+
+        # If the spec is dynamic, i.e. the function signature includes `**kwargs` and the number of positional arguments
+        # passed is larger than the number of explicitly defined parameters in the signature, the inputs are invalid and
+        # we should raise. If we don't, some of the passed arguments, intended to be positional arguments, will be
+        # misinterpreted as keyword arguments, but they won't have an explicit name to use for the link label, causing
+        # the input link to be completely lost.
+        if cls.spec().inputs.dynamic and nargs > nparameters:
+            name = cls._func.__name__
+            raise TypeError('{}() takes {} positional arguments but {} were given'.format(name, nparameters, nargs))
+
+    @classmethod
     def create_inputs(cls, *args, **kwargs):
         """Create the input args for the FunctionProcess"""
+        cls.validate_inputs(*args, **kwargs)
+
         ins = {}
         if kwargs:
             ins.update(kwargs)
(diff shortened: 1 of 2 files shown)


aiida/engine/processes/functions.py
```diff
@@ -228,9 +228,29 @@ def _define(cls, spec):
...
+        nargs = len(args)
+        nparameters = len(cls._func_args)
+
+        # If the spec is dynamic, i.e. the function signature includes `**kwargs` and the number of positional arguments
+        # passed is larger than the number of explicitly defined parameters in the signature, the inputs are invalid and
+        # we should raise. If we don't, some of the passed arguments, intended to be positional arguments, will be
+        # misinterpreted as keyword arguments, but they won't have an explicit name to use for the link label, causing
+        # the input link to be completely lost.
+        if cls.spec().inputs.dynamic and nargs > nparameters:
+            raise ValueError('function has {} positional arguments, but received only {}'.format(nparameters, nargs))
```
giovannipizzi: I would try to return the same exception that Python returns:
```
In [1]: def f(a,b,**kwargs):
   ...:     pass
   ...: 

In [2]: f(1,2,3)
<ipython-input-2-b61741968103> in <module>()
----> 1 f(1,2,3)

TypeError: f() takes exactly 2 arguments (3 given)
```

sphuber: Fair enough, bit weird that they chose `TypeError` for this though. Is it because the "type" of `3` is positional whereas it should be keyword? Not sure what the reasoning was

giovannipizzi: I always had the same question in mind :-) Maybe just double check in python3 should they have changed something (I think I tested on py2 above)

sphuber: Same exception but they did improve the message:
```
In [1]: def test(a, **kwargs): 
   ...:     pass 
   ...:                                                                                                                                       

In [2]: test(1, 2)                                                                                                                            
<ipython-input-2-f4ef31738bb7> in <module>
----> 1 test(1, 2)

TypeError: test() takes 1 positional argument but 2 were given
```

sphuber: Maybe I should change my message to be the same, I like it, it is clear

sphuber:
Oops, accidentally clicked on `resolve` conflict. Anway, I changed the type to `TypeError` and also the message to match to that what py3 throws

PR #2632, opened 2019-03-22
Corrected the graph export set expansion rules that are described at …

szoupanos:
…issue 1102

https://github.com/aiidateam/aiida_core/issues/1102

There were problems when overriding the default behaviour of the rules. Moreover
they were moved to separate methods to resuce the complexity of the export method.

Various tests were added:
- Export tests that are based on a real workchain were added to ensure that a complex
  and real AiiDA graph based on workchains is exported correctly.
- Export tests based on a synthetic graph were added that explicitly check every
  flag that changes the default behaviour of the export set expansion rules.

### DIFF
--- a/aiida/backends/tests/export_and_import.py
+++ b/aiida/backends/tests/export_and_import.py
@@ -13,6 +13,8 @@ Tests for the export and import routines.
 
 from aiida.backends.testbase import AiidaTestCase
 from aiida.orm.importexport import import_data
+from aiida.orm.calculation.inline import make_inline
+
 
 class TestSpecificImport(AiidaTestCase):
 
@@ -168,6 +170,7 @@ class TestSpecificImport(AiidaTestCase):
             qb.append(RemoteData, project=['uuid'], output_of='parent', tag='remote')
             qb.append(Calculation, output_of='remote')
             self.assertGreater(len(qb.all()), 0)
+
 
 class TestSimple(AiidaTestCase):
 
@@ -746,10 +749,8 @@ class TestSimple(AiidaTestCase):
         from aiida.common.links import LinkType
         from aiida.orm.importexport import export
 
-        from aiida.common.exceptions import NotExistent
         # Creating a folder for the import/export files
         temp_folder = tempfile.mkdtemp()
-
 
         try:
             master = WorkCalculation().store()
@@ -777,7 +778,6 @@ class TestSimple(AiidaTestCase):
         finally:
             # Deleting the created temporary folder
             shutil.rmtree(temp_folder, ignore_errors=True)
-
 
     def test_reexport(self):
         """
@@ -896,6 +896,7 @@ class TestSimple(AiidaTestCase):
             # Deleting the created temporary folder
             shutil.rmtree(temp_folder, ignore_errors=True)
 
+
 class TestComplex(AiidaTestCase):
 
     def test_complex_graph_import_export(self):
@@ -980,6 +981,7 @@ class TestComplex(AiidaTestCase):
         finally:
             # Deleting the created temporary folder
             shutil.rmtree(temp_folder, ignore_errors=True)
+
 
 class TestComputer(AiidaTestCase):
 
@@ -1461,6 +1463,7 @@ class TestComputer(AiidaTestCase):
                              comp1_metadata,
                              "Not the expected metadata were found")
 
+
 class TestLinks(AiidaTestCase):
 
     def setUp(self):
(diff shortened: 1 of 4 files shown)


ltalirz:
@danieleongari has now done a first test

Here is what we noticed:

 * printing out the selected rules (very helpful!) made us realize that the current implementation of the `-C` option is very counterintuitive: while all other options turn the corresponding rules *on*,  `-C` turns the rule *off*
```
  -C, --create-reversed           Follow reverse CREATE links (recursively)
                                  when calculating the node set to export.
                                  [default: True]
```
I think the only proper way of resolving this is to have both `--create-reversed` and `--no-create-reversed` options (same for all other options).

 * When turning on all possible rules (`-RIX`), the export got stuck (we canceled after a few minutes). 
   I wonder what it was looking for - would it have tried to take the whole connected subset of the graph? (The DB contains 194125 nodes in total)

 * The option `-RX`now seems to produce what we were looking for originally :-)

 * Printing uuids instead of PKs for `verdi graph generate` is a good idea - however
   * it is important to use a shortened version of the uuid (see [graph.pdf](https://github.com/aiidateam/aiida_core/files/3000336/113.pdf); to be compared to the [original graph](https://github.com/aiidateam/aiida_core/files/2946116/103481.pdf) )
   * it should be possible to switch back to PKs if desired

ltalirz:
Concerning which subset of the uuid to show: git uses 7 characters for the commit hash; we could use 8 characters (the first part of the uuid).

Subsets of uuids are already supported in 0.12, e.g. you can do
`verdi  node show -u ea70cc12`

szoupanos:
Thanks a lot Leopold for the testing and the feedback!

Due to time constrains (and for keeping things simple), I would keep this PR to the absolutely necessary to make the export rules work correctly (the UUID at the graph generate were left there by accident because they were useful when debugging - but if we all like it, I can keep a shorter UUID version).

The discussion on overriding the arguments can be added at another ticket and the implementation at another PR.

ltalirz:
> Due to time constrains (and for keeping things simple), I would keep this PR to the absolutely necessary to make the export rules work correctly 

Sounds good to me. Let's do both the uuid stuff and the changes to options in a separate PR (please remove the uuid changes).

I will review the code changes this evening.

ltalirz:
However, please keep printing the rules that are selected in this PR (perhaps add a print "Rules on following links:" line before)

szoupanos:
The addition of the UUIDs at the graph will be addressed at issue #2636

szoupanos:
Further discussion on the export set expansion rules can be done at #2637

szoupanos:
For me it is OK.

- Switched back to pks for the AiiDA graph
- Created the necessary tickets for the AiiDA graph (with UUIDS) and the default behaviour of the export rules

aiida/backends/tests/export_and_import.py
```diff
@@ -1481,6 +1483,175 @@ def get_all_node_links(self):
                   edge_project=['label', 'type'], output_of='input')
         return qb.all()
 
+    class ComplexGraph:
+        """
+        A class responsible to create a"complex" graph with all available link types
+        (INPUT, CREATE, RETURN and CALL)
+        """
+
+        test_case = None
```
ltalirz: should these be class variables?
if they are just instance variables, you can remove these two lines

aiida/backends/tests/export_and_import.py
```diff
@@ -1481,6 +1483,175 @@ def get_all_node_links(self):
                   edge_project=['label', 'type'], output_of='input')
         return qb.all()
 
+    class ComplexGraph:
```
ltalirz: this file already has >1k lines
could you please move the ComplexGraph out into a separate file?

aiida/backends/tests/export_and_import.py
```diff
@@ -1481,6 +1483,175 @@ def get_all_node_links(self):
...
+            from aiida.common.links import LinkType
+
+            # Node creation
+            d1 = Int(1).store()
+            d2 = Int(2).store()
+            wc1 = WorkCalculation().store()
+            wc2 = WorkCalculation().store()
+
+            pw1 = JobCalculation()
+            pw1.set_computer(self.test_case.computer)
```
ltalirz: you use the test_case just to get the computer, correct?
you use the `test_case` just for the computer, correct?
can you just pass the `computer` to the constructor instead?

this is more explicit, while passing the test case creates a loop where the test case creates the ComplexGraph, which again has a reference to the test case, ...

ltalirz:
I have to stop the review now, will continue later

aiida/backends/tests/export_and_import.py
szoupanos: I prefer to keep them for clarity - I don't like the approach of using undefined variables even if Python allows it - I find it dirty

szoupanos: But I will document them though

aiida/backends/tests/export_and_import.py
szoupanos: Good point, I will do that.
I passed the test_case since I wanted to be more generic.
But it is cleaner what you propose.

aiida/backends/tests/export_and_import.py
szoupanos: I propose this to be done under another PR. And mainly to 1.0
There is already one created by Casper (#2702)

szoupanos:
I addressed some of the comments. I would propose to merge this and apply the same changes to 1.0 (since it is important to have correct export rules) and leave the further refactoring to be done under another ticket (since there is one - and Casper has already started working on it).

szoupanos:
@ltalirz @sphuber 
Can this be merged please?

ltalirz:
I'll finish the review tonight

szoupanos:
Thanks a lot Leo!
Yes, with these flags on, according to the following rules the full sub-graph that is connected to the node should be returned.
https://github.com/aiidateam/aiida_core/issues/1102#issuecomment-383090834

Regarding the speed of the graph exploration algorithm, I think that it is a simple graph traversal algorithm
https://en.wikipedia.org/wiki/Graph_traversal

If you find a case that you think that it is particularly slow, we can have a look at it (in case there is a bug etc).

PR #2638, opened 2019-03-25
Allow filelike in `Singelfile.set_file` and constructor

### LINKED ISSUE #2635: `SinglefileData` needs way to set file from filelike
Currently, the only way to set the file for a `SinglefileData` is through an absolute filepath. However, with the recent change in node repository interface, it is no longer possible to get an absolute filepath from for example a `FolderData` retrieved folder node at the end of a calculation. To create a `SinglefileData` from one of the files it contains, one would have to get the content and create a temporary file on disk only to pass it to `set_file` which will then read the content a gain. The `set_file` should allow both an absolute file path or filelike object.

sphuber:
Fixes #2635 

With the new node repository interface no longer providing access to
absolute filepaths, the requirement for a filepath in the
`SinglefileData` constructor and `set_file` method was too restrictive.
Here we add the possibility to add a filelike argument as well.


aiida/backends/tests/orm/data/test_data.py
```diff
@@ -1,72 +0,0 @@
-# -*- coding: utf-8 -*-
```
ltalirz: @sphuber Is deleting  `test_data.py` intended?
I just searched for  `validate_incoming` in this page and only found it here...

sphuber: Yes, those tests were fundamentally broken but never detected because were not run as they were not added to the list of tests to be run. In any case, they are not necessary, because what they were supposed to test is properly tests in `aiida.backends.tests.orm.node.test_node.py`

ltalirz: Ok, please include this in the commit message when you squash

aiida/backends/tests/orm/data/test_singlefile.py
```diff
@@ -0,0 +1,87 @@
...
+class TestSinglefileData(AiidaTestCase):
+    """Tests for the `SinglefileData` class."""
+
+    def test_reload_singlefile_data(self):
+        """Test writing and reloading a `SinglefileData` instance."""
+        content_original = 'some text ABCDE'
+
+        with tempfile.NamedTemporaryFile(mode='w+') as handle:
+            filepath = handle.name
+            basename = os.path.split(filepath)[1]
```
ltalirz: why not use os.path.basename as below?

sphuber: That test already existed and merely copied it from `test_basedataclasses.py` and simply didn't adapt it. I will change it

aiida/orm/nodes/data/singlefile.py
```diff
@@ -23,10 +24,15 @@
...
+    def __init__(self, file, **kwargs):
+        """Construct a new instance and set the contents to that of the file.
+
+        :param file: an absolute filepath or filelike object whose contents to copy
+        """
+        # pylint: disable=redefined-builtin
         super(SinglefileData, self).__init__(**kwargs)
-        if filepath is not None:
-            self.put_object_from_file(filepath)
+        if file is not None:
```
ltalirz: so, file is a required constructor argument but we want to allow passing `None`?
shouldn't we just let it fail if someone passes `None`?

aiida/orm/nodes/data/singlefile.py
```diff
@@ -56,20 +62,26 @@ def get_content(self):
...
-        """Add the file located at `path` on file system to repository, deleting any other existing objects."""
-        if not os.path.isabs(path):
-            raise ValueError('path `{}` is not absolute'.format(path))
+        :param file: an absolute filepath or filelike object whose contents to copy
+        """
+        # pylint: disable=redefined-builtin
 
-        if not os.path.isfile(path):
-            raise ValueError('path `{}` does not correspond to an existing file'.format(path))
+        # If `file` is filelike and not a string path, the following will raise an `AttributeError`
```
ltalirz: I guess this comment is now outdated

sphuber: correct, will remove

aiida/orm/nodes/data/singlefile.py
```diff
@@ -56,20 +62,26 @@ def get_content(self):
...
+        if isinstance(file, six.string_types):
+            is_filelike = False
 
-        if key is None:
-            key = os.path.split(path)[1]
+            key = os.path.split(file)[1]
```
ltalirz: another candidate for  `basename` ?

aiida/orm/nodes/data/singlefile.py
```diff
@@ -56,20 +62,26 @@ def get_content(self):
...
+            if not os.path.isabs(file):
+                raise ValueError('path `{}` is not absolute'.format(file))
+
+            if not os.path.isfile(file):
+                raise ValueError('path `{}` does not correspond to an existing file'.format(file))
+        else:
+            is_filelike = True
+            key = os.path.split(file.name)[1]
```
ltalirz: same here

aiida/orm/nodes/data/singlefile.py
sphuber: done

aiida/orm/nodes/data/singlefile.py
sphuber: done

aiida/orm/nodes/data/singlefile.py
sphuber: Agreed, does not make any sense, however, it might be nice to not 100% require a file be passed. For example sub classes may want different behavior. The `CifData` class currently also allows to create it from an `ase` or `values` object, in which case a `file` is not allowed. However, it still needs to call the parent, which will then fail. Leaving aside whether this `CifData` implementation makes sense, the idea that a sub class will set the file differently somehow, is a feasible use-case no?

ltalirz: yeah, also upfdata might make problems. let's leave it as it is for the moment.

ltalirz:
Looks good - will approve once you've pushed the changes

aiida/orm/nodes/data/singlefile.py
sphuber: I also just noticed that Github's code highlighting also seems to think that `file` is somehow special despite not being an official reserved keyword. Weird..

aiida/backends/tests/orm/data/test_data.py
sphuber: Will do

aiida/orm/nodes/data/singlefile.py
ltalirz: turns out it's a `builtin`
https://stackoverflow.com/a/24942363/1069467

sphuber: aha, good to know

PR #2639, opened 2019-03-25
Make sure exit code != 0 if incompatible archive version

### LINKED ISSUE #2633: verdi import returns exit code 0 also when import fails
this is on aiida-core 1.0.0b1

```
$ verdi import SSSP_accuracy_pseudos.aiida 
\/home/max/.virtualenvs/aiida/local/lib/python2.7/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
  """)
Info: importing archive SSSP_accuracy_pseudos.aiida
READING DATA AND METADATA...
EXTRACTING NODE DATA...
Warning: SSSP_accuracy_pseudos.aiida cannot be imported: Export file version is 0.3, can import only version 0.4
Use 'verdi export migrate' to update this export file.

$ echo $?
0
```

CasperWA:
Fixes #2633.

For `verdi import`:
Change the warning to a critical if exception "IncompatibleArchiveVersionError" is raised.

Added test `test_import_old_versions`.
It checks a non-zero value is returned, when an archives with old versions are imported.
Also, added asserts of exit code for other `verdi import` tests.

### DIFF
--- a/aiida/backends/tests/cmdline/commands/test_import.py
+++ b/aiida/backends/tests/cmdline/commands/test_import.py
@@ -15,6 +15,7 @@ import os
 
 import unittest
 from click.testing import CliRunner
+from click.exceptions import BadParameter
 
 from aiida.backends.testbase import AiidaTestCase
 from aiida.cmdline.commands import cmd_import
@@ -46,15 +47,17 @@ class TestVerdiImport(AiidaTestCase):
         options = []
         result = self.cli_runner.invoke(cmd_import.cmd_import, options)
 
-        self.assertIsNotNone(result.exception)
+        self.assertIsNotNone(result.exception, result.output)
         self.assertIn('Critical', result.output)
+        self.assertNotEqual(result.exit_code, 0, result.output)
 
     def test_import_non_existing_archives(self):
         """Test that passing a non-existing archive will lead to command failure."""
         options = ['non-existing-archive.aiida']
         result = self.cli_runner.invoke(cmd_import.cmd_import, options)
 
-        self.assertIsNotNone(result.exception)
+        self.assertIsNotNone(result.exception, result.output)
+        self.assertNotEqual(result.exit_code, 0, result.output)
 
     @unittest.skip("Reenable when issue #2426 has been solved (migrate exported files from 0.3 to 0.4)")
     def test_import_archive(self):
@@ -73,6 +76,7 @@ class TestVerdiImport(AiidaTestCase):
         result = self.cli_runner.invoke(cmd_import.cmd_import, options)
 
         self.assertIsNone(result.exception, result.output)
+        self.assertEqual(result.exit_code, 0, result.output)
 
     @unittest.skip("Reenable when issue #2426 has been solved (migrate exported files from 0.3 to 0.4)")
     def test_comment_mode(self):
@@ -85,8 +89,83 @@ class TestVerdiImport(AiidaTestCase):
         result = self.cli_runner.invoke(cmd_import.cmd_import, options)
         self.assertIsNone(result.exception, result.output)
         self.assertIn('Comment mode: newest', result.output)
+        self.assertEqual(result.exit_code, 0, result.output)
 
         options = ['--comment-mode', 'overwrite'] + archives
         result = self.cli_runner.invoke(cmd_import.cmd_import, options)
         self.assertIsNone(result.exception, result.output)
         self.assertIn('Comment mode: overwrite', result.output)
+        self.assertEqual(result.exit_code, 0, result.output)
+
+    def test_import_old_local_archives(self):
+        """ Test import of old local archives
+        Expected behavior: Return message in terminal and error code != 0
+        """
+        archives = [('export/migrate/export_v0.1.aiida', '0.1'), ('export/migrate/export_v0.2.aiida', '0.2'),
+                    ('export/migrate/export_v0.3.aiida', '0.3')]
+
+        for archive, version in archives:
+            options = [get_archive_file(archive)]
+            result = self.cli_runner.invoke(cmd_import.cmd_import, options)
+
+            self.assertIsNotNone(result.exception, result.output)
+            self.assertNotEqual(result.exit_code, 0, result.output)
+            self.assertIn(version, result.output, result.exception)
+
+    def test_import_old_url_archives(self):
+        """ Test import of old URL archives
+        Expected behavior: Return message in terminal and error code != 0
+        """
+        url = "https://raw.githubusercontent.com/aiidateam/aiida_core/f685844b290ed23df254873df62da7162bde0f1b/"
+        archive_path = "aiida/backends/tests/fixtures/export/migrate/"
+        archive = 'export_v0.1.aiida'
+        version = '0.1'
+
+        options = [url + archive_path + archive]
+        result = self.cli_runner.invoke(cmd_import.cmd_import, options)
+
+        self.assertIsNotNone(result.exception, result.output)
+        self.assertNotEqual(result.exit_code, 0, result.output)
+        self.assertIn(version, result.output, result.exception)
+
+    @unittest.skip("Reenable when issue #2426 has been solved (migrate exported files from 0.3 to 0.4)")
+    def test_import_url_and_local_archives(self):
+        """Test import of both a remote and local archive
+        TODO: Update 'url' to point at correct commit.
+        Now it is pointing to yakutovicha's commit, but when PR #2478 has been merged in aiidateam:develop,
+        url should be updated to point to the, essentially same, commit, but in aiidateam.
+        """
+        url = "https://raw.githubusercontent.com/yakutovicha/aiida_core/f5fff1846a62051b898f13db67f5eef18892d5f4/"
+        url_archive = "aiida/backends/tests/fixtures/export/migrate/export_v0.4_no_UPF.aiida"
+        local_archive = "export/migrate/export_v0.4_no_UPF.aiida"
+
+        options = [get_archive_file(local_archive), url + url_archive, get_archive_file(local_archive)]
+        result = self.cli_runner.invoke(cmd_import.cmd_import, options)
+
+        self.assertIsNone(result.exception, result.output)
+        self.assertEqual(result.exit_code, 0, result.output)
+
+    def test_import_url_timeout(self):
+        """Test a timeout to valid URL is correctly errored"""
+        from aiida.cmdline.params.types import ImportPath
+
+        timeout_url = "http://www.google.com:81"
+
+        test_timeout_path = ImportPath(exists=True, readable=True, timeout_seconds=0)
+        with self.assertRaises(BadParameter) as cmd_exc:
+            test_timeout_path(timeout_url)
+
+        error_message = 'Path "{}" could not be reached within 0 s.'.format(timeout_url)
+        self.assertIn(error_message, str(cmd_exc.exception), str(cmd_exc.exception))
+
+    def test_raise_malformed_url(self):
+        """Test the correct error is raised when supplying a malformed URL"""
+        malformed_url = "htp://www.aiida.net"
+
+        result = self.cli_runner.invoke(cmd_import.cmd_import, [malformed_url])
+
+        self.assertIsNotNone(result.exception, result.output)
+        self.assertNotEqual(result.exit_code, 0, result.output)
+
+        error_message = 'It may be neither a valid path nor a valid URL.'
+        self.assertIn(error_message, result.output, result.exception)
--- a/aiida/cmdline/commands/cmd_import.py
+++ b/aiida/cmdline/commands/cmd_import.py
@@ -8,6 +8,7 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """`verdi import` command."""
+# pylint: disable=too-many-locals
 from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
@@ -16,7 +17,7 @@ import click
 
 from aiida.cmdline.commands.cmd_verdi import verdi
 from aiida.cmdline.params.options import MultipleValueOption
-from aiida.cmdline.params.types import GroupParamType
+from aiida.cmdline.params.types import GroupParamType, ImportPath
 from aiida.cmdline.utils import decorators, echo
 from aiida.common import exceptions
 
@@ -36,7 +37,7 @@ class ExtrasImportCode(Enum):
 
 
 @verdi.command('import')
-@click.argument('archives', nargs=-1, type=click.Path(exists=True, readable=True))
+@click.argument('archives', nargs=-1, type=ImportPath(exists=True, readable=True))
 @click.option(
     '-w',
     '--webpages',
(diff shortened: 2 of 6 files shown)


aiida/cmdline/commands/cmd_import.py
```diff
@@ -127,7 +127,7 @@ def cmd_import(archives, webpages, group, extras_mode_existing, extras_mode_new,
                 extras_mode_new=extras_mode_new,
                 comment_mode=comment_mode)
         except exceptions.IncompatibleArchiveVersionError as exception:
-            echo.echo_warning('{} cannot be imported: {}'.format(archive, exception))
+            echo.echo_critical('{} cannot be imported: {}'.format(archive, exception))
             continue
```
sphuber: Since the `echo_critical` call will call `sys.exit` you can remove the `continue`

CasperWA: Done.

CasperWA:
I noticed I missed this correction for an archive URL, this was only done for a local archive.
I have corrected this initial oversight and removed `continue` in both places.
If you have a valid URL to an archive, please paste it and I will add another test for this if needed.

ltalirz:
@CasperWA well, I have lots of valid urls for archives (on the materials cloud archive) but those are all archives created by aiida-core 0.12.
I think it's important that we add a flag to `verdi import` that does the migration automatically (but not in this PR).

CasperWA:
After discussing it over with @ltalirz, I will import the same archives, but from GitHub, referencing a specific commit that contains the files.
There is still the problem of running the tests if one has no internet connection, however, specifically for the "make sure it fails"-test, this shouldn't be a problem, other than the fact we're not testing _exactly_ what we want to test...

CasperWA:
`verdi import` is broken with respect to URLs.

I will rewrite it in this PR, since it is essentially fixing the code to make the tests work and get the result that is sought after in issue #2633.

CasperWA:
> I will rewrite it in this PR, ...

Done :)

aiida/backends/tests/cmdline/commands/test_import.py
```diff
@@ -85,8 +88,67 @@ def test_comment_mode(self):
...
+            self.assertNotEqual(result.exit_code, 0, result.output)
+            self.assertIn(version, result.output, result.exception)
+
+    def test_import_old_url_archives(self):
+        """ Test import of old URL archives
+        Expected behavior: Return message in terminal and error code != 0
+        """
+        url = "https://raw.githubusercontent.com/aiidateam/aiida_core/f685844b290ed23df254873df62da7162bde0f1b/"
+        archive_path = "aiida/backends/tests/fixtures/export/migrate/"
+        archives = [('export_v0.1.aiida', '0.1'), ('export_v0.2.aiida', '0.2'), ('export_v0.3.aiida', '0.3')]
```
ltalirz: Fine to test the import via URL once; no need to do requests for all versions (it's already tested above)

aiida/backends/tests/cmdline/commands/test_import.py
```diff
@@ -85,8 +88,67 @@ def test_comment_mode(self):
...
+        local_archive = "export/migrate/export_v0.4_no_UPF.aiida"
+
+        options = [get_archive_file(local_archive), url + url_archive, get_archive_file(local_archive)]
+        result = self.cli_runner.invoke(cmd_import.cmd_import, options)
+
+        self.assertIsNone(result.exception, result.output)
+        self.assertEqual(result.exit_code, 0, result.output)
+
+    def test_import_url_timeout(self):
+        """Test a timeout to valid URL is correctly errored"""
```
ltalirz: hm... this I find a but problematic.
the rule of thumb for unit tests is that they should run "in the blink of an eye".
are you waiting for a timeout here? is there a way to reduce this test to ~100ms?

aiida/cmdline/commands/cmd_import.py
```diff
@@ -25,6 +26,52 @@
 COMMENT_MODE = ['newest', 'overwrite']
 
 
+class ImportPath(click.Path):
+    """AiiDA Verdi implementation of Click's Path-type.
+    Introduce tests to distinguish whether the path is to a local file
+    or a URL string to a file that needs to be downloaded.
+    """
```
ltalirz: Suggestion:

```
"""AiiDA extension of Click's Path-type to include URLs.

An ImportPath can either be a Click Path-type or a URL.
"""
```

aiida/cmdline/commands/cmd_import.py
```diff
@@ -25,6 +26,52 @@
...
+
+    def __init__(self, **kwargs):
+        super(ImportPath, self).__init__(**kwargs)
+
+    def __call__(self, value, param=None, ctx=None):
+        if value is not None:
```
ltalirz: is that check necessary here?
shouldn't we let covert handle that?

aiida/cmdline/commands/cmd_import.py
```diff
@@ -25,6 +26,52 @@
...
+            try:
+                res = super(ImportPath, self).convert(value, param, ctx)
```
ltalirz: since you are reusing the `convert` - perhaps it would be cleaner to simply overwrite `convert` directly?

i.e.
```
def convert(..):
   try:
       return super(ImportPath, self).convert(value, param, ctx)
    except click.exceptions.BadParameter:
       new code...
```

the way you do it currently, if somebody were to use your new `convert` method, the result would not be the expected one (since it only handles the URL case)

aiida/cmdline/commands/cmd_import.py
```diff
@@ -25,6 +26,52 @@
...
+        """Do checks for possible URL path
+        Parameters :param:`readable` and :param:`exists` are always set to `True`.
+        """
+        from six.moves import urllib
+        from django.core.validators import URLValidator
+        from django.core.exceptions import ValidationError as URLValidationError
+        from socket import timeout
+
+        url = value
+        timeout_seconds = 10
```
ltalirz: this should probably go as a constant in some configuration module

aiida/cmdline/commands/cmd_import.py
```diff
@@ -25,6 +26,52 @@
...
+
+        try:
+            URLValidator()(url)
+        except URLValidationError:
+            self.fail('%s "%s" is not a valid URL or local %s.' % (self.path_type, filename_to_ui(url), self.name),
```
ltalirz: why "or local"?

aiida/cmdline/commands/cmd_import.py
```diff
@@ -159,8 +205,7 @@ def cmd_import(archives, webpages, group, extras_mode_existing, extras_mode_new,
                     comment_mode=comment_mode)
             except exceptions.IncompatibleArchiveVersionError as exception:
                 echo.echo_warning('{} cannot be imported: {}'.format(archive, exception))
-                echo.echo_warning('download the archive file and run `verdi export migrate` to update it')
-                continue
+                echo.echo_critical('download the archive file and run `verdi export migrate` to update it')
```
ltalirz: since these are not two independent messages, probably  better to construct the string first and then pass it to `echo_critical` directly

ltalirz:
Thanks for looking into this!
Through with my review

aiida/backends/tests/cmdline/commands/test_import.py
CasperWA: Sure, will reduce.

aiida/backends/tests/cmdline/commands/test_import.py
CasperWA: Yes, I agree about this. But I want/need to test that we can get to this point in the code. If you have a better suggestion I'm happy to hear it. The timeout is otherwise "only" 10 seconds.

aiida/cmdline/commands/cmd_import.py
CasperWA: I'm simply reimplementing the `__call__` function as it is for `click.Path`'s parent class `ParamType`. I think it's fine to handle it here and let `convert` handle the more detailed checks.

aiida/cmdline/commands/cmd_import.py
CasperWA: True, This way I don't have to overwrite the `__call__` function as well.

aiida/cmdline/commands/cmd_import.py
CasperWA: Probably should. And I think the whole class should go somewhere else as well. I was kind of hoping @sphuber could help me out on this :)

aiida/cmdline/commands/cmd_import.py
CasperWA: Because if it gets to this point, we simply assume that the input is a URL, but it may just be a mistyped local path. So to not completely confuse the user, i put in "or local".

aiida/cmdline/commands/cmd_import.py
CasperWA: I agree, I thought to do this as well, but forgot it again as quickly, unfortunately. Good catch.

aiida/cmdline/commands/cmd_import.py
ltalirz: You're right. In this case, what about the following error message:
```
'"%s" is neither a valid filepath nor a valid URL.' % (filename_to_ui(url))
```

ltalirz: if you want to make things more "analogous" between checking whether it is a path or a URL, you can do a 
```
try:
   ...
except: 
    pass
```
for both and at the very end of the function fail with this error message.

CasperWA: That's a nice idea.

aiida/backends/tests/cmdline/commands/test_import.py
CasperWA: I've thought of something. I'll make `timeout` a property exactly for this purpose. Then we can test it with `timeout = 0`.

aiida/cmdline/commands/cmd_import.py
CasperWA: Done.

aiida/cmdline/commands/cmd_import.py
CasperWA: Done

aiida/cmdline/commands/cmd_import.py
CasperWA: I have moved `ImportPath` to `aiida.cmdline.params.path`.
The timeout seconds are moved into this file as well.

aiida/cmdline/params/types/path.py
```diff
@@ -31,3 +33,76 @@ def convert(self, value, param, ctx):
...
+        return url
+
+    @property
+    def timeout_seconds(self):
+        return self._timeout_seconds
+
+    # pylint: disable=attribute-defined-outside-init
+    @timeout_seconds.setter
+    def timeout_seconds(self, value):
+        if isinstance(value, str):
```
sphuber: This whole block can be cleaned up. First off, you can get either a `ValueError` or `TypeError` if the type of the timeout is incorrect. Furthermore, the combination of `if` conditional and then try-except is superfluous and not very pythonic. I would change this whole block simply in:
```
try:
     self._timeout_seconds = int(value)
except ValueError:
    raise TypeError('timeout should be an integer but got: {}'.format(type(value)))
```
The default should simply be set in the constructor of this object.

aiida/cmdline/params/types/path.py
```diff
@@ -31,3 +33,76 @@ def convert(self, value, param, ctx):
...
+            # Check if `click.Path`-type
+            return super(ImportPath, self).convert(value, param, ctx)
+        except click.exceptions.BadParameter:
+            # Check if URL
+            return self.checks_url(value, param, ctx)
+
+    def checks_url(self, value, param, ctx):
+        """Do checks for possible URL path"""
+        from six.moves import urllib
+        from django.core.validators import URLValidator
```
sphuber: Do we really need a django dependency here? I know we have it anyway, but if at some point it needs to be dropped. Isn't there a simply python analogue. Or better, do not validate it all, but literally just go to the next block and try to open it and catch the exception. Maybe the `urllib` library already throws  a specific exception for malformed URLs

aiida/backends/tests/cmdline/commands/test_import.py
```diff
@@ -85,8 +89,81 @@ def test_comment_mode(self):
...
+        result = self.cli_runner.invoke(cmd_import.cmd_import, options)
+
+        self.assertIsNotNone(result.exception, result.output)
+        self.assertNotEqual(result.exit_code, 0, result.output)
+        self.assertIn(version, result.output, result.exception)
+
+    @unittest.skip("Reenable when issue #2426 has been solved (migrate exported files from 0.3 to 0.4)")
+    def test_import_url_and_local_archives(self):
+        """Test import of both a remote and local archive
+        TODO: Update 'url' to point at correct commit.
```
sphuber: Is this TODO still relevant and if so what is the 'correct' URL?

CasperWA: I will deepen the explanation in the docstring to include the essence of the following:
It is to be changed when the "migration from v0.3 to v0.4" PR has gone through. Now it is referencing the new export files I created for that PR in Sasha's branch. However, as soon as it has been merged, and Sasha probably deletes his branch, then this URL should point to the, essentially same, commit, but in "aiidateam", not "yakutovicha".

aiida/cmdline/params/types/path.py
CasperWA: I have moved it to the next block and removed URL validation, other than through `urlopen()`.

aiida/cmdline/params/types/path.py
CasperWA: Done. I have reduced it to your more simple try-except block, keeping the following limit checks.
The default value is set in the constructor with a check `if timeout_seconds is None` set it to the default.
I think it is maybe better to keep a None value in the `__init__` parameters, since if the default is set already here and someone should pass a `None` value the default value will still be used. Unless we specifically want it to raise if a None value is passed?

aiida/cmdline/params/types/path.py
```diff
@@ -31,3 +33,65 @@ def convert(self, value, param, ctx):
...
+    An ImportPath can either be a `click.Path`-type or a URL.
+    """
+
+    # pylint: disable=protected-access
+
+    def __init__(self, timeout_seconds=None, **kwargs):
+        super(ImportPath, self).__init__(**kwargs)
+
+        # Set timeout_seconds
+        if timeout_seconds is None:
```
sphuber: What is the advantage of doing this instead of just putting it in the method signature? The default is immutable so that is fine?

CasperWA: Reiterating a previous comment:

The default value is set in the constructor with a check if timeout_seconds is None set it to the default.
I think it is maybe better to keep a None value in the __init__ parameters, since if the default is set already here and someone should pass a None value the default value will still be used. Unless we specifically want it to raise if a None value is passed?

CasperWA: According to our short talk, I'll simply define `timeout_seconds` in the `__init__` parameters.

CasperWA:
Writing some tests for the new class.

CasperWA:
Done. New tests for `ImportPath` to test specifically the `timeout_seconds` parameter.

PR #2640, opened 2019-03-25
pyproject: remove reentry from requirements

dev-zero:
The problem is that reentry starts to scan the entrypoints after
installation which picks up the current AiiDA directory as well, but
aiida-core is not yet installed.

I verified that pip is still able to build a wheel with `pip wheel .`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
 # Minimum requirements for the build system to execute.
-requires = ["pip>=18.1", "setuptools", "wheel", "reentry==1.2.2"]
+requires = ["pip>=18.1", "setuptools", "wheel"]


ltalirz:
@dev-zero does this mean, we can remove the fix to pip 18.1 from here?
https://github.com/aiidateam/aiida_core/blob/7a090a40608a457f42c058a0be03280016ee450b/.travis.yml#L53

sphuber:
The reason we added reentry to the build requirements is because for the automatic reentry scan hooks to be triggered each time a plugin is installed, reentry *had* to be installed before installing `aiida-core`. If we can verify that this is no longer required, for all installation paths, then we could remove the requirement I would say

sphuber:
I just did a test locally where I removed the `18.1` limitation from `pip` and `reentry` entirely, and build with `pip==19.0.3` from source in a completely clean environment with all reentry cache deleted and the install indeed worked and registered the entry points. I then installed from source a plugin and its entry points also got installed. I am not sure if there is something that I am missing in the ensuring a "clean" environment for testing, or if there are other installation paths that might not work. If you guys thinks this is not the case, then lets remove the version limitation on pip from the toml as well and merge this.

ltalirz:
I'll do a quick test as well

ltalirz:
There are two related open issues on the reentry tracker
https://github.com/DropD/reentry/issues/44
https://github.com/DropD/reentry/issues/43

I'll probably need to look into those first

dev-zero:
@sphuber that's the same test I ran and it just worked (tm)
Installing from a wheel or pip (DropD/reentry#43) might give a slightly different environment, so should be tested separately (locally generating wheels without the `reentry` and then installing from the local Python package cache should be sufficient for that).

ltalirz:
> Installing from a wheel or pip (DropD/reentry#43) might give a slightly different environment

I just tested this using a locally built wheel. For me (MacOS, conda environment, pip 19.0.3), installing from the wheel works fine (both with and without reentry in pyproject.toml) but in both cases the reentry map is empty afterwards.

For me it's ok to merge this for the moment in order to avoid problems with people installing develop.
The underlying issue of reentry will need to be fixed though.

sphuber:
Just as some additional information, I just setup Travis CI for `aiida-codtools` and there the hook was not triggered. When running the tests after installation, even the internal aiida entry points were not recognized. I had to add a manual `reentry scan` line in the installation section of the `travis.yml`

ltalirz:
I pinged @DropD concerning this, it would be great to have a bit of knowledge transfer before one of us goes after debugging this

xjf729:
When I use `pip install -e . --no-build-isolation` with pip version 19, "reentry error" was not appeared. So, I think we can use pip 19 to install aiida_core.

ltalirz:
closing in favor of #2741

PR #2646, opened 2019-03-26
return correct builder for CalcJob (fixes #2652)

### LINKED ISSUE #2652: Override get_builder() on CalcJob, to return CalcJobBuilder

chrisjsewell:
fixes #2652


chrisjsewell:
I also had to change `CalcJobBuilder.submit_test` to get it working.
You might want to check this change is correct, and possibly also add a test that calls it

chrisjsewell:
Update the `submit_test` method (on the node) has been removed.
[This method](https://github.com/aiidateam/aiida_core/blob/8bf2445ddb8a674598f633cadb04ba7316fffc36/aiida/orm/implementation/general/calculation/job/__init__.py#L1728) needs to be re-implemented

sphuber:
thanks @chrisjsewell

PR #2654, opened 2019-03-26
Fix #2653 repository.open as contextmanager

### LINKED ISSUE #2653: Repository.open() should be a context manager
[this is the method to change](https://github.com/aiidateam/aiida_core/blob/0ca5d991d039dca2261df3d6ad8da2f03d591caa/aiida/orm/utils/repository.py#L98)

chrisjsewell:
fixes #2653


chrisjsewell:
my mistake, this isn't actually required

PR #2656, opened 2019-03-26
Activate publish with travis

odarbelaeze:
Now we an publish to pypi by tagging a commit in the master branch,
the code is published if:

- The whole build matrix passes
- The build is a tagged commit
- The tag is eaither `v{major}.{minor}.{patch}` or `v{version}(a|b|rc){number}`.
- The repository matches ours (doesn't run in other contributors travis).

I made some checks in another repo to verify the behaviour on travis:

- Full deploy: https://travis-ci.org/odarbelaeze/aiida_core/builds/511607509
- A test fails so the deploy is not run: https://travis-ci.org/odarbelaeze/aiida_core/builds/511522325
- Deploy is skipped unless it's a tagged commit: https://travis-ci.org/odarbelaeze/aiida_core/jobs/511976867
- Deploy is skiped if someone pushes an arbitrary named tag https://travis-ci.org/odarbelaeze/aiida_core/jobs/512048835

After updates here are examples of the current behaviour:

* Any branch https://travis-ci.org/odarbelaeze/aiida_core/builds/515620496
* Non semver tag https://travis-ci.org/odarbelaeze/aiida_core/builds/515620534
* Semver tag in the right branch https://travis-ci.org/odarbelaeze/aiida_core/builds/515620556

We can check that the package was correctly published by installing the test package `aiida-core-bot==1.0.0b4`.


.travis.yml
```diff
@@ -72,9 +72,36 @@ before_script:
...
+            - echo "Deployed $TRAVIS_TAG"
+            - echo "We'll hook up some things in the future"
+          deploy: &pypi
+              provider: pypi
+              skip_existing: true
+              username: aiida-bot
+              password:
+                  secure: NhV11e1kVJaumNMb7YGENngZwa6qJjzoGWzmzqdU5BwFgpXaxbaFAk2fn+Cckrkz4MmNJaW1HaKEAhn07WsbrkjOXK7QVNK48/eagiquKbCQbyNZNjzF+C24EYQHI93y40JQcl2JaCahSqKXbYQCpLlX0Rbtob4psQH29uHRcbq4lm5t3Jmb8WckEzcDLMZX3+uPFwDJxMbbsDW+pONGF1z2/B0j+7m4beTgCepuvIEaXS97rTQj2egKYEJV+V3DbH2o2nr0+3z4lzH2FAdoAnZMlFwHfWoY3WIuYcHcwXDWyLGWQKvncaoh1sLU7gfjjup3dZ4Iq74Zp43x2eXUhbmZBp2cPN3CZpN9baE4NE6MNKeQ/erkg31qdq50OG/rjGgPKyfg5ShO3Kt1CyVbSJX5dqPxtDQblYE8TAGhHfO2O8M1kLhyWvrV5TaQuVB/IAsTkRC+t1NtdWTxaU6wdMz36O5so89oKbFljjk9744m4Ei8DCgc98WH6b96qn0VifjxGuBs6o1aqRoa8O7e29a0TwDVAwxkczgjxbXkP6okRY5IAxsr5Rsbm8urXEo874uQA8n1gGyN6+YKSbjvPbHD9RIRl9IddC6y2wKTv/1JBifNjHcLhadl7RVRPSU7vFSs2cknN+5Gkw3FgZ/jA8SoMxMLFsvHcFxchrurAkE=
+              on:
+                  repo: aiidateam/aiida_core
```
odarbelaeze: Hellop.

ltalirz:
@odarbelaeze thanks for your putting in your efforts, I like it!

I just have two minor questions:
 * the travis log of the deploy step always shows "skip" [even if it runs](https://travis-ci.org/aiidateam/aiida_core/builds/511976882?utm_source=github_status&utm_medium=notification). I guess this is because the `script` step is skipped. Is there a way to change this?
 * as you mentioned before, the fact that one has to manually skip all the "regular" steps of travis would seem to indicate that we are doing something unusual here. Did you manage to find out why what we are doing (wanting all parallel test builds to pass before deploying) is unusual?

@giovannipizzi I saw you talked about this with Oscar, are you OK with this?

I think this is a good step in the right direction and will already enable further niceties down the line (creating automatic PRs to dependent repositories after an AiiDA release etc.).

odarbelaeze:
@ltalirz it shows skip because of the "env: skip" directive. I could try with `env: ignore`.

odarbelaeze:
@ltalirz with the conditional deploy stage, the stage now doesn't even show in other branches, is that better?

.travis.yml
```diff
@@ -72,9 +72,37 @@ before_script:
...
   # upload coverage report to coveralls.io
   - if [ "$TEST_TYPE" == "tests" ]; then coveralls; fi
 
 git:
   depth: 3
+
+jobs:
+    include:
+        - stage: deploy
+          if: "branch = master AND tag =~ ^v[0-9]+\\.[0-9]+\\.[0-9]+(a|b|rc)?[0-9]*$"
```
ltalirz: Just checking - the single `=` is really the equality operator in this language?

Also, is there a reason you are using `tag` here and not `$TRAVIS_TAG` like below?

.travis.yml
```diff
@@ -72,9 +72,37 @@ before_script:
...
+              skip_existing: true
+              username: aiida-bot
+              password:
+                  secure: NhV11e1kVJaumNMb7YGENngZwa6qJjzoGWzmzqdU5BwFgpXaxbaFAk2fn+Cckrkz4MmNJaW1HaKEAhn07WsbrkjOXK7QVNK48/eagiquKbCQbyNZNjzF+C24EYQHI93y40JQcl2JaCahSqKXbYQCpLlX0Rbtob4psQH29uHRcbq4lm5t3Jmb8WckEzcDLMZX3+uPFwDJxMbbsDW+pONGF1z2/B0j+7m4beTgCepuvIEaXS97rTQj2egKYEJV+V3DbH2o2nr0+3z4lzH2FAdoAnZMlFwHfWoY3WIuYcHcwXDWyLGWQKvncaoh1sLU7gfjjup3dZ4Iq74Zp43x2eXUhbmZBp2cPN3CZpN9baE4NE6MNKeQ/erkg31qdq50OG/rjGgPKyfg5ShO3Kt1CyVbSJX5dqPxtDQblYE8TAGhHfO2O8M1kLhyWvrV5TaQuVB/IAsTkRC+t1NtdWTxaU6wdMz36O5so89oKbFljjk9744m4Ei8DCgc98WH6b96qn0VifjxGuBs6o1aqRoa8O7e29a0TwDVAwxkczgjxbXkP6okRY5IAxsr5Rsbm8urXEo874uQA8n1gGyN6+YKSbjvPbHD9RIRl9IddC6y2wKTv/1JBifNjHcLhadl7RVRPSU7vFSs2cknN+5Gkw3FgZ/jA8SoMxMLFsvHcFxchrurAkE=
+              on:
+                  repo: aiidateam/aiida_core
+                  branch: master
+                  tags: true
+                  # Make sure the tag is valid semver
+                  condition: "$TRAVIS_TAG =~ ^v[0-9]+\\.[0-9]+\\.[0-9]+(a|b|rc)?[0-9]*$"
```
ltalirz: Is this still necessary?
If not better to remove to reduce code duplication (imagine someone changes one regex but not the other; very difficult to spot)

.travis.yml
odarbelaeze: You know, safety for the win.
`tag` and `branch` are part of this grammar.

https://docs.travis-ci.com/user/conditions-v1

.travis.yml
odarbelaeze: :thinking: you're right.

However, I think I should test this in my own fork again. Just to make sure it'll behave correctly.

odarbelaeze:
@ltalirz @giovannipizzi @sphuber comments in this PR have been addressed, I ran some extra test runs in my own repo to verify taht all would work as expected once merged:

* Any branch https://travis-ci.org/odarbelaeze/aiida_core/builds/515620496
* Non semver tag https://travis-ci.org/odarbelaeze/aiida_core/builds/515620534
* Semver tag in the right branch https://travis-ci.org/odarbelaeze/aiida_core/builds/515620556

ltalirz:
@odarbelaeze Thanks for your work! for me this is good to go

sphuber:
Two questions:
1. Will this only take a "slot" on Travis when the conditions are met, or will this add a permanent 11th job that will fill one of our 5 consecutive slots?
2. I take it the people who created this have thought long and hard about the secure password that is included. But just for my information, do you know how this cannot be "hijacked" and be used by anyone else to publish anything using the `aiida-bot` account?

odarbelaeze:
@szoupanos 

1. Yes, in any branch or non-semver tag the job is going to be completely omitted (as you can see in the examples)
2. It's anti-symmetric encryption the process is described [here](https://docs.travis-ci.com/user/encryption-keys/#encryption-scheme). TL;DR; Travis creates a public private key pair for each repo, makes the public key publicly available and I can encrypt a message and put it in the file and only travis can read it with their private key, so... people can tell our travis their own secrets but they won't be able to read ours.

As you can see [in this commit](https://github.com/odarbelaeze/aiida_core/commit/693f07a507b420c82fb9e9e3a8d92a6707bad5d3) I had to encrypt our `aiida-core-bot` password with my repo's own public key to make those tests possible.

sphuber:
@odarbelaeze thanks for the explanation. All good for me

giovannipizzi:
Seems ok, let's try out if it works correctly for the next release!

PR #2657, opened 2019-03-26
silence psycopg2 warning

ltalirz:
I think we have been warned enough:

UserWarning: The psycopg2 wheel package will be renamed from release
2.8; in order to keep installing from binary please use "pip install
psycopg2-binary" instead. For details see:
<http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.

### DIFF
--- a/aiida/__init__.py
+++ b/aiida/__init__.py
@@ -57,6 +57,9 @@ else:
 if six.PY2:
     warnings.warn('python 2 will be deprecated in `aiida-core v2.0.0`', DeprecationWarning)  # pylint: disable=no-member
 
+# Disable psycopg2 warning. Remove this after upgrading psycopg2 to 2.8
+warnings.filterwarnings("ignore", category=UserWarning, module='psycopg2')
+
 
 def try_load_dbenv(*argc, **argv):
     """
--- a/aiida/__init__.py
+++ b/aiida/__init__.py
@@ -58,7 +58,7 @@ if six.PY2:
     warnings.warn('python 2 will be deprecated in `aiida-core v2.0.0`', DeprecationWarning)  # pylint: disable=no-member
 
 # Disable psycopg2 warning. Remove this after upgrading psycopg2 to 2.8
-warnings.filterwarnings("ignore", category=UserWarning, module='psycopg2')
+warnings.filterwarnings("ignore", category=UserWarning, module='psycopg2')  # pylint: disable=no-member
 
 
 def try_load_dbenv(*argc, **argv):


sphuber:
Couldn't we instead just already start installing from `psycopg-binary` if that is what we want anyway?

ltalirz:
It isn't what we want anyway - we should be installing `psycopg2`

https://pypi.org/project/psycopg2-binary/
> The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.

or do you disagree?

sphuber:
Ah ok, I think I understand now then. So does that mean that currently the package `psycopg` is the binary build and so far we have been running of the binary version? And if that is the case, if we don't change our installation target, once `psycopg==2.8` is released, we will automatically start installing  the version built from source. Given the description you quoted, that does seem to be what we want. Looking at the prerequisites, the only thing that might be missing is `libpq-dev`, but maybe that gets installed when you install postgres itself?

ltalirz:
> Ah ok, I think I understand now then. So does that mean that currently the package psycopg is the binary build and so far we have been running of the binary version?

That's what I understand as well.

Thinking again about your earlier comment, I'm no longer so convinced we actually want to follow their advice...
If we now start depending on people having installed `libpq`, python header files etc. (+ the potentially increased install time), it's maybe not such a good idea after all, also since we never heard of problems with the binary version.
And psycopg2 does offer pre-built wheels for a lot of operating systems
https://pypi.org/project/psycopg2-binary/#files

Feel free to close this PR and just switch back to psycopg2-binary

sphuber:
I still am a bit worried to go to the binaries, if they advice not to for production. I delved a bit further and it seems that [this is the reason for splitting off the binary package](https://www.postgresql.org/message-id/CA%2Bmi_8bd6kJHLTGkuyHSnqcgDrJ1uHgQWvXCKQFD3tPQBUa2Bw%40mail.gmail.com). In short the problem with the binary is that it comes with the libssl dependency, which may differ from that installed on the system, causing a potential segfault. Installing from source will ensure that the same library is used and there should be no problem. Further details [can be found here](http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/). Since `psycopg`'s build requisites are most likely shared with those of `postgres`, which we require to be installed anyway, maybe having those requirements is not bad. At least not as bad as having a database adaptor that can fail unpredictably.

@giovannipizzi your thoughts?

sphuber:
Anyway, it seems like the guy is about to release `2.8`. If we decide to stay with the build from source, the warning will be gone.

ltalirz:
Release just happened: https://pypi.org/project/psycopg2/

ltalirz:
(and promptly broke my azure builds...)

sphuber:
Oh the joy of dependencies...

ltalirz:
I do think going back to psycopg2-binary is the most user friendly thing to do, given that we never had problems with psycopg2

sphuber:
But we should be careful that this will work with downstream dependencies that require psycopg, e.g. [SqlAlchemy](https://github.com/sqlalchemy/sqlalchemy/blob/b90f9b5f367b16ddc1a2d5b06b12649c3fa9efdd/setup.py#L183) has the option to install both. We should make sure that if we select the binary everything downstream also does so. Otherwise we might end up with two versions of the package sharing the same namespace

sphuber:
Shall I in any case add `psycopg2==2.8` to the dependencies for now, before I release `1.0.0b2` so we get rid of the warning at least? Then we can think next week what to do about binary not-binary

giovannipizzi:
ok for me

ltalirz:
Let's see whether this works on travis
Still, I would predict complaints (e.g. on the azure macos, it was missing some of the prerequisites and the installation failed)

sphuber:
How did pip install whether to install from scratch or from the binary before though? If you install `psycopg2==2.7.6` through pip, will it always default to the binary?

ltalirz:
I believe it checks whether a wheel is available for your platform and uses it, if available

sphuber:
Travis seems to be fine with `psycopg2` except that the conda tests now fail. Who is responsible for making `2.8` available on conda?

ltalirz:
both anaconda and conda-forge provide it

https://anaconda.org/anaconda/psycopg2
https://github.com/conda-forge/psycopg2-feedstock

such a delay is quite unusual...

If you really want to merge this now, you can put the conda test with allow_failure for the moment
https://github.com/aiidateam/aiida-diff/blob/master/.travis.yml#L66
still, I think switching to 2.8 non-binary is risky

sphuber:
Thanks, I just created a PR on the [conda-forge feedstock](https://github.com/conda-forge/psycopg2-feedstock/pull/26). We could decide to update to `psycopg-binary==2.8` to at least keep the same behavior and then we can later discuss if we want to advise to use non-binary. Question then is whether that uses a separate feedstock as well?

ltalirz:
Unless specified explicitly (using `noarch`), conda will build binary packages by default.
That's also why we have this replacement  rule
https://github.com/aiidateam/aiida_core/blob/develop/utils/validate_consistency.py#L262

ltalirz:
For supported platforms, see here: https://anaconda.org/conda-forge/psycopg2

sphuber:
Closing in favor of PR #2712

PR #2661, opened 2019-03-27
Fix 2659 builder mutablemap

### LINKED ISSUE #2659: ProcessBuilderNamespace should be a MutableMapping
For convenience, it just needs a `__delitem__` and `__setitem__`,
then you could do:

```python
builder.update({
    'input1': Node(),
    'input2': Node()
})
```

rather than have to set each attribute separately.

chrisjsewell:
fixes #2659 

I guess the trade of with this is that you can no longer use `popitem`, `clear`, `update`, and `setdefault `
as attributes


sphuber:
Thanks @chrisjsewell but this has been taken care of #2694 with additional subtleties taken care of. Note that you can now update it with a dictionary but you have to use `builder._update`

PR #2671, opened 2019-03-29
Implement `__getitem__` on `Dict` node sub class

### LINKED ISSUE #2667: Implement `Dict.__getitem__` to allow to access keys directly
This would allow one to do:
```
node = Dict(dict={'a': 5})
node['a']
```
instead of having to call `get_dict` first.

sphuber:
Fixes #2667 

This shortcut will allow users to directly use `node[key]` instead of
having to go through either `get_dict` or `dict` property.

### DIFF
--- a/aiida/backends/tests/__init__.py
+++ b/aiida/backends/tests/__init__.py
@@ -108,6 +108,7 @@ db_test_list = {
         'orm.authinfos': ['aiida.backends.tests.orm.test_authinfos'],
         'orm.comments': ['aiida.backends.tests.orm.test_comments'],
         'orm.computers': ['aiida.backends.tests.orm.test_computers'],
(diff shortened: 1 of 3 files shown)


sphuber:
Small additional shortcut, maybe a bit superfluous when we already have `dict`, but don't think it will hurt. In any case it adds some missing unit tests

ltalirz:
I've been waiting for something like this for a long time, so both thumbs up!

If we do this, why not implement a `__setitem__` as well the pipes through to `set_attribute`?

Also, I'm wondering whether `__getitem__` instead of going through `self.dict` should use `get_attribute`?

sphuber:
Do we also want to support the mutable attributes and then just raise `ModificationNotAllowed` for stored nodes?

giovannipizzi:
Maybe yes? I would suggest to approve this as is, and then in an issue list the features that we would like to support

PR #2681, opened 2019-04-01
Warn user when copying group to non-empty destination group

### LINKED ISSUE #2679: Bug or unexpected behavior in `verdi group copy`
The command `verdi group copy` will create a clone of a certain group with a new name.
If the group with the new name already exists it will use that. If the user already created the new empty group, this may well be the expected behavior. But if the new name was misspelled by accident, now all the nodes of the source group will be copied to the existing group. There should be a warning when defining the target group that is not empty

CasperWA:
Fixes #2679

#### Logic:
If destination group was not created and is not empty, the user will be warned and asked to confirm the copy.

The copy will take place - or the user will have aborted.

#### Added test:
`test_copy_existing_group`: Creates source group with nodes.
First copies and creates destination group, checking all is well, then copies the same source group to the same destination group, checking an abort error is raised.

#### Minor things:
Corrected typo in doc-string for `__init__` of Group entity.

### DIFF
--- a/aiida/backends/tests/cmdline/commands/test_group.py
+++ b/aiida/backends/tests/cmdline/commands/test_group.py
@@ -210,3 +210,44 @@ class TestVerdiGroup(AiidaTestCase):
         result = self.cli_runner.invoke(group_remove_nodes, ['--force', '--clear', '--group=dummygroup1'])
         self.assertClickResultNoException(result)
         self.assertEqual(group.count(), 0)
+
+    def test_copy_existing_group(self):
+        """Test user is prompted to continue if destination group exists and is not empty"""
+        source_label = "source_copy_existing_group"
+        dest_label = "dest_copy_existing_group"
+
+        # Create source group with nodes
+        calc_s1 = orm.CalculationNode().store()
+        calc_s2 = orm.CalculationNode().store()
+        nodes_source_group = {str(node.uuid) for node in [calc_s1, calc_s2]}
+        source_group = orm.Group(label=source_label).store()
+        source_group.add_nodes([calc_s1, calc_s2])
+
+        # Copy using `verdi group copy` - making sure all is successful
+        options = [source_label, dest_label]
+        result = self.cli_runner.invoke(group_copy, options)
+        self.assertClickResultNoException(result)
+        self.assertIn(
+            "Success: Nodes copied from group<{}> to group<{}>".format(source_label, dest_label),
+            result.output,
+            result.exception)
+
+        # Check destination group exists with source group's nodes
+        dest_group = orm.load_group(label=dest_label)
+        self.assertEqual(dest_group.count(), 2)
+        nodes_dest_group = {str(node.uuid) for node in dest_group.nodes}
+        self.assertSetEqual(nodes_source_group, nodes_dest_group)
+
+        # Copy again, making sure an abort error is raised, since no user input can be made and default is abort
+        result = self.cli_runner.invoke(group_copy, options)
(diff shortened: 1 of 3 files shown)


aiida/cmdline/commands/cmd_group.py
```diff
@@ -332,6 +332,13 @@ def group_copy(source_group, destination_group):
     """Add all nodes that belong to source group to the destination group (which may or may not exist)."""
     from aiida import orm
 
-    dest_group = orm.Group.objects.get_or_create(label=destination_group, type_string=source_group.type_string)[0]
+    dest_group, created = orm.Group.objects.get_or_create(label=destination_group, type_string=source_group.type_string)
+
+    # Issue warning if destination group is not empty and get user confirmation to continue
+    if not created and not dest_group.is_empty:
+        echo.echo_warning('Destination group<{}> already exists and is not empty.'.format(destination_group))
```
sphuber: Let's use `group.label` in the echo'd messages. Same few lines below

aiida/backends/tests/cmdline/commands/test_group.py
```diff
@@ -210,3 +210,31 @@ def test_add_remove_nodes(self):
...
+        options = [source_label, dest_label]
+        result = self.cli_runner.invoke(group_copy, options)
+        self.assertClickResultNoException(result)
+        self.assertIn(
+            "Nodes copied from group<{}> to group<{}>".format(source_group, dest_label),
+            result.output,
+            result.exception)
+
+        # Copy again, making sure an abort error is raised, since no user input can be made and default is abort
+        result = self.cli_runner.invoke(group_copy, options)
```
sphuber: Will this ever finish? Won't it just be stuck at the prompt?

aiida/backends/tests/cmdline/commands/test_group.py
```diff
@@ -210,3 +210,31 @@ def test_add_remove_nodes(self):
...
+        calc_s1 = orm.CalculationNode().store()
+        calc_s2 = orm.CalculationNode().store()
+        source_group = orm.Group(label=source_label).store()
+        source_group.add_nodes([calc_s1, calc_s2])
+
+        # Copy using `verdi group copy` - making sure all is successful
+        options = [source_label, dest_label]
+        result = self.cli_runner.invoke(group_copy, options)
+        self.assertClickResultNoException(result)
+        self.assertIn(
```
sphuber: It would be good to check that the destination group was actually created and it contains the nodes you expect by loading it through normal ORM API. It is true that we are not testing the actually ORM functionality, but it could happen that due to a developer's mistake, the proper logic to copy over nodes to the new group is never called.

aiida/backends/tests/cmdline/commands/test_group.py
CasperWA: No, it uses the default answer instead (abort). The tests pass.

aiida/backends/tests/cmdline/commands/test_group.py
CasperWA: Done

aiida/cmdline/commands/cmd_group.py
CasperWA: I thought the idea was to use the general group representation, since it also displays information about the group, the user, etc. However, I agree it seems cleaner to have simply the label, and if a user wishes more information they can use `verdi group show`.

sphuber:
Thanks a lot @CasperWA

Issue #2568, opened 2019-03-05
Validation of link labels


giovannipizzi:
There are a number of places where, I think, we assume that link labels are valid python variable names, like:
- `.inputs.<label>` and `.outputs.<label>` methods of ProcessNodes
- when passing labels as kwargs to `run()` and `submit()` (actually I think cPython does not require this and works also with invalid variable names, but other python implementations might break)

Therefore, it might be a good idea to document rules on link labels and enforce/validate these rules in `add_incoming` (and possibly other places if applicable)

sphuber:
The validity of port names, and by extension the keys that need to be used in the launch functions, is taken care of in PR #2688 . This will also guarantee that the link labels that are generated automatically by the engine, by imploding the nested port namespaces, are valid. However, this validation is bypassed when calling `add_incoming` directly. We should add an explicit check on label validity in this method.

giovannipizzi:
Good point

Issue #1694, opened 2018-06-28
Use period as separator for namespaced input and output links


sphuber:
Currently, if a Process exposes inputs into a specific namespace, the link names will be a concatenation of the namespace and the input port, joined by an underscore. However, in specifying namespaces, we use `.` as the separator. Should these two made to be congruent?

sphuber:
Closing this in favor of a more general and well described #2688

PR #2687, opened 2019-04-01
Make the outputs of a live process nestable

### LINKED ISSUE #2666: Outputs registered for calculation job in parser are not validated

### LINKED ISSUE #2680: Process output namespaces are broken
Just like the `inputs` of a `Process`, the `outputs` are a `PortNamespace` and so in principle they should work exactly like the inputs of a process. However, the way outputs are currently attached to the a process and are validated when it terminates is broken. For example, the `self.out` method does not accept a dictionary that should be mapped onto the output port namespace of the spec. It expects a single `Data` instance as the value. One could try to access a nested output port by using the `.` character in the output key that is passed to `self.out` but also that will fail, e.g.:
```
self.out('integer.namespace.two', Int(2))
======================================================================
ERROR: test_output_validation_error (aiida.backends.tests.engine.test_process.TestProcess)
Test that a process is marked as failed if its output namespace validation fails.
----------------------------------------------------------------------
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/backends/tests/engine/test_process.py", line 237, in test_output_validation_error
    results, node = run_get_node(TestProcess, add_outputs=orm.Bool(True))
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/engine/launch.py", line 49, in run_get_node
    return runner.run_get_node(process, *args, **inputs)
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/engine/runners.py", line 218, in run_get_node
    result, node = self._run(process, *args, **inputs)
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/engine/runners.py", line 194, in _run
    process.execute()
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/processes.py", line 88, in func_wrapper
    return func(self, *args, **kwargs)
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/processes.py", line 1051, in execute
    return self.future().result()
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/futures.py", line 34, in result
    return super(Future, self).result(timeout)
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
  File "<string>", line 3, in raise_exc_info
TypeError: Error validating output 'uuid: 4d621fe2-3404-4c8f-8525-8fd15a20b40f (unstored) value: 2' for port 'n.a.m.e.s.p.a.c.e': Unexpected ports {'two': <Int: uuid: 4d621fe2-3404-4c8f-8525-8fd15a20b40f (unstored) value: 2>}, for a non dynamic namespace
```

sphuber:
Fixes #2680 and fixes #2666 

The outputs namespace of a process specification allows for arbitrary
nesting, just like the inputs. The inputs of a process, when passed,
will have to have a nesting structure that matches that of the inputs
process specification. However, the internal structure of outputs of a
process were not keeping the same nested structure as their process
specification counterpart. Of course when the processes finishes and the
outputs have to be linked up through links to the process node, the
support for nesting disappears and the mapping will have to be
flattened. However, while the process is "alive" the internal outputs
should keep the nesting structure of the process spec.

### DIFF
--- a/aiida/backends/tests/engine/test_process.py
+++ b/aiida/backends/tests/engine/test_process.py
@@ -15,21 +15,21 @@ import threading
 import plumpy
 from plumpy.utils import AttributesFrozendict
 
+from aiida import orm
 from aiida.backends.testbase import AiidaTestCase
 from aiida.backends.tests.utils import processes as test_processes
 from aiida.common.lang import override
-from aiida.engine import Process, run, run_get_pk
-from aiida.orm import load_node, Dict, Int, Str, WorkflowNode
+from aiida.engine import Process, run, run_get_pk, run_get_node
 
 
 class NameSpacedProcess(Process):
 
-    _node_class = WorkflowNode
+    _node_class = orm.WorkflowNode
 
     @classmethod
     def define(cls, spec):
         super(NameSpacedProcess, cls).define(spec)
-        spec.input('some.name.space.a', valid_type=Int)
+        spec.input('some.name.space.a', valid_type=orm.Int)
 
 
 class TestProcessNamespace(AiidaTestCase):
@@ -47,7 +47,7 @@ class TestProcessNamespace(AiidaTestCase):
         Test that inputs in nested namespaces are properly validated and the link labels
         are properly formatted by connecting the namespaces with underscores
         """
-        proc = NameSpacedProcess(inputs={'some': {'name': {'space': {'a': Int(5)}}}})
+        proc = NameSpacedProcess(inputs={'some': {'name': {'space': {'a': orm.Int(5)}}}})
 
         # Test that the namespaced inputs are AttributesFrozenDicts
         self.assertIsInstance(proc.inputs, AttributesFrozendict)
@@ -57,17 +57,16 @@ class TestProcessNamespace(AiidaTestCase):
 
         # Test that the input node is in the inputs of the process
         input_node = proc.inputs.some.name.space.a
-        self.assertTrue(isinstance(input_node, Int))
+        self.assertTrue(isinstance(input_node, orm.Int))
         self.assertEquals(input_node.value, 5)
 
         # Check that the link of the process node has the correct link name
         self.assertTrue('some_name_space_a' in proc.node.get_incoming().all_link_labels())
         self.assertEquals(proc.node.get_incoming().get_node_by_label('some_name_space_a'), 5)
 
-
 class ProcessStackTest(Process):
 
-    _node_class = WorkflowNode
+    _node_class = orm.WorkflowNode
 
     @override
     def run(self):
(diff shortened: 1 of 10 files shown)


giovannipizzi:
A few questions:
- when you output a node from a workchain, isn't this immediately attached? Or only at the very end?
- is there any check that if there is an output a_b and a namespace a with output a.b, these do not overwrite the same link? (Maybe this is another issue that we've been discussing about using double underscores and avoiding to start (or end) with an underscore? This is BTW something to discuss and then implement early as it changes link names and I don't think we want to make a migration for it)
- if I output a.b and there is no namespace `a`, is this considered an error? (Maybe there are already tests for this?)

sphuber:
A few answers:
 * When you call `self.out` in a process, that will straight away go through validation, but will then be stored in an internal cache in the process instance. This is all done in `plumpy`. At a state transition, for example when the workchain transitions to the next step in the outline, the AiiDA process will call the method `update_outputs`, which will determine the new outputs and create links between them and the process node. When the process terminates, the entire outputs port namespace is validated. This is the part that was missing and has been corrected in this PR.
 * This is indeed an ongoing problem that I on purpose did not address here. It is described more or less in #2568 and #1694 . The problem is not just to do with valid links but how we convert nested process port namespaces into a flat space of link labels. I think I will create a new issue that describes the exact situation and then I can close these two. I think this will be a blocking fix for 1.0.0
 * Yes, if there is no `a` namespace, this will fail, unless the top level namespace `outputs` is dynamic, then this will be allowed. This is very liberal and is why we removed the default dynamicity of input and output namespaces a few days ago.

giovannipizzi:
Ok - so nodes will be created at state transition, but not validated immediately? And validated only at process termination? Shouldn't they be validated before every node creation? Or there is a reason not to do it? (Maybe checking that all components of the namespace are present, that cannot be done at state transition since they could be created in a later step?)

sphuber:
> When you call self.out in a process, that will straight away go through validation`
The node itself will be validated straight away, but the whole output spec can only be validated at the end, because outputs can be registered at any time, so once you register the first, you might not yet have attached all required outputs.

You cannot validate the node "creation" so much, what we validate is that if someone "registers" an output with the process through `self.out`, it was validated against the spec, i.e. for type, value etc. but then is stored in an internal cache. When the process transitions state, the newly registered outputs are reflected in the graph, by storing the nodes (if they weren't already as in for example the case when calling `self.out` on one of the outputs produced by a calculation it called) and then adding the link. Then the process instance is dumped and saved as a checkpoint.

Issue #2673, opened 2019-03-29
Validate specified parser before launching `CalcJob`


sphuber:
The very least the validation can do is see if the specified parser through the `parser_name` option, can be loaded.

PR #2690, opened 2019-04-02
improve error message for missing parse method

ltalirz:


sphuber:
Would it perhaps be better to make `Parser` a proper abstract class and the `parse` method an abstract method. This would make the instantiation of a parser that does not implement `parse` raise a `TypeError`. I imagine this will also simplify the work for issue #2673

ltalirz:
That seems like the right thing to do.

Will the error message include the fact that the "parse" method is not implemented?
 (rather than just `NotImplemented` as currently)

Anyhow, unless you want to do it immediately, we can also merge this first - whatever you prefer.

sphuber:
This will be the error message
```
In [3]: p = Parser(node)
<ipython-input-3-d17a4041a4f2> in <module>()
----> 1 p = Parser(node)

TypeError: Can't instantiate abstract class Parser with abstract methods parse
```

ltalirz:
Ok. It's perhaps not the clearest message but `parse` is in there.

It's also good that it is raised already when the parser is instantiated rather then only when the (missing) method is called.

ltalirz:
closing in favor of #2693

PR #2691, opened 2019-04-02
Set default for `CalcJob.metadata.options.withmpi` to `False`

### LINKED ISSUE #2689: Change the default for `CalcJob.metadata.options.withmpi` to `False`
This is currently `True` in `1.0.0b1` as it is in [0.12.4](https://github.com/aiidateam/aiida_core/blob/8bf2445ddb8a674598f633cadb04ba7316fffc36/aiida/orm/implementation/general/calculation/job/__init__.py#L378). However, during the plugin migration there were some suggestions that a `False` default might be safer.

sphuber:
Fixes #2689 

Running codes with MPI by accident that don't support it is worse than
accidentally not using MPI for those that can. The safer default for
`aiida-core` therefore is to set it to `False`. Plugins can override this
on a per calculation basis when defining the process specification.

### DIFF
--- a/aiida/engine/processes/calcjobs/calcjob.py
+++ b/aiida/engine/processes/calcjobs/calcjob.py
@@ -78,7 +78,7 @@ class CalcJob(Process):
             help='Set the quality of service to use in for the queue on the remote computer')
         spec.input('metadata.options.computer', valid_type=orm.Computer, required=False,
             help='Set the computer to be used by the calculation')
-        spec.input('metadata.options.withmpi', valid_type=bool, default=True,
+        spec.input('metadata.options.withmpi', valid_type=bool, default=False,
             help='Set the calculation to use mpi',)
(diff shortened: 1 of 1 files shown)


ltalirz:
Added note on how to override to
https://github.com/aiidateam/aiida_core/wiki/AiiDA-1.0-plugin-migration-guide

giovannipizzi:
Shouldn't we also add a check around [here](https://github.com/aiidateam/aiida_core/blob/f9618189af9fb400fc923a289634d5944d986967/aiida/engine/processes/calcjobs/calcjob.py#L320) that `job_tmpl.job_resource.get_tot_num_mpiprocs()==1` if `withmpi==False`?

sphuber:
Thanks @ltalirz

ltalirz:
@giovannipizzi right, you mentioned this... sorry!

ltalirz:
We said we would provide some kind of a warning in this case

PR #2693, opened 2019-04-02
Make `Parser` an abstract class and `parse` abstract method

### LINKED ISSUE #2673: Validate specified parser before launching `CalcJob`

sphuber:
Fixes #2673 

This prevents one from instantiating a parser that has not implemented
the `parse` method, without which it will be useless. This also makes
the validation of the `CalcJob.metadata.options.parser_name` easier as
it will now be able to verify that the specified parser entry point can
both be resolved and imported, as well as constructed, or the input
validation will fail. This will prevent users from accidentally
launching a calculation job only to find after it has finished that an
incorrect parser was specified.

Note that we cannot use the `validator` key of the input port for this
task, as the ports are overridable by sub classes of the `CalcJob` and
if they forget to reattach the validator, the validation will not be
performed.

### DIFF
--- a/aiida/orm/nodes/process/calculation/calcjob.py
+++ b/aiida/orm/nodes/process/calculation/calcjob.py
@@ -162,11 +162,16 @@ class CalcJobNode(CalculationNode):
             raise exceptions.ValidationError('invalid calculation state `{}`'.format(self.get_state()))
 
         try:
-            self.get_parser_class()
-        except exceptions.EntryPointError:
-            raise exceptions.ValidationError("No valid class/implementation found for the parser '{}'. "
-                                             "Set the parser to None if you do not need an automatic "
-                                             "parser.".format(self.get_option('parser_name')))
+            parser_class = self.get_parser_class()
+        except exceptions.EntryPointError as exception:
+            raise exceptions.ValidationError('invalid parser specified: {}'.format(exception))
+
+        try:
+            # Since a parser is not required to be set, so `get_parser_class` will return `None` in that case
+            if parser_class is not None:
+                parser_class(self)
+        except TypeError as exception:
+            raise exceptions.ValidationError('invalid parser specified: {}'.format(exception))
 
         computer = self.computer
         scheduler = computer.get_scheduler()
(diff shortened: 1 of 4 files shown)


aiida/orm/nodes/process/calculation/calcjob.py
```diff
@@ -163,11 +163,16 @@ def _validate(self):
...
-                                             "Set the parser to None if you do not need an automatic "
-                                             "parser.".format(self.get_option('parser_name')))
+            parser_class = self.get_parser_class()
+        except (exceptions.MissingEntryPointError, exceptions.MultipleEntryPointError,
+                exceptions.LoadingEntryPointError) as exception:
+            raise exceptions.ValidationError('invalid parser specified: {}'.format(exception))
+
+        try:
+            if parser_class is not None:
+                parser_class(self)
```
ltalirz: `None(self)` also raises a `TypeError`.
If there is a reason why the `parser_class` can be `None`, could you put a comment here explaining what this means?

aiida/parsers/parser.py
```diff
@@ -15,16 +15,22 @@
...
 
 __all__ = ('Parser',)
 
 
-class Parser(object):  # pylint: disable=useless-object-inheritance
+@six.add_metaclass(ABCMeta)
+class Parser(object):
     """Base class for a Parser that can parse the outputs produced by a CalcJob process."""
 
+    # pylint: disable=useless-object-inheritance
```
ltalirz: interesting that this works, putting it inside the class.
is there a reason why you moved it there?
it would seem the original place is closer to the problem it complains about

ltalirz:
Just two minor comments.

aiida/orm/nodes/process/calculation/calcjob.py
sphuber: It can be `None` because we don't require a parser to be defined. I will put a comment

aiida/parsers/parser.py
sphuber: Only reason is that all of a sudden the linter started breaking my balls and complaining about this. Moving it seemed to have done the job. Don't understand why

sphuber:
@ltalirz this should be good to go

PR #2694, opened 2019-04-02
Formalize the mapping of port namespaces to link labels

### LINKED ISSUE #2688: Formalize the mapping between nested process port namespaces to flat link labels
The `Process` and `ProcessNode` duality forces us to define a mapping between the nested port namespaces of the process and the flat link label hierarchy of the corresponding node in the provenance graph. A process can define a nested port namespace through its spec:
```
class NestedProcess(Process):

    @classmethod
    def define(cls, spec):
        super(NestedProcess, cls).define(spec)
        spec.input('name.spaced.input')
        spec.output('name.spaced.output')
```

If we print the inputs port namespace, we get the following:

```
In [6]: print(NestedProcess.spec().inputs)
{
    "_attrs": {
        "default": [], 
        "dynamic": false, 
        "help": null, 
        "required": "True", 
        "valid_type": "None"
    },
    "name": {
        "_attrs": {
            "default": [], 
            "dynamic": false, 
            "help": null, 
            "required": "True", 
            "valid_type": "None"
        }, 
        "spaced": {
            "_attrs": {
                "default": [], 
                "dynamic": false, 
                "help": null, 
                "required": "True", 
                "valid_type": "None"
            }, 
            "input": {
                "name": "input", 
                "non_db": "False", 
                "required": "True"
            }
        }
    }
}
```
When constructing this process, or any process, the inputs are expected to map exactly onto this namespace.
So for example, `inputs = {'name': {'spaced': {'input': 1}}}` would be valid.
The same principles go for outputs and they should be registered in such a way that it maps one-to-one on the port namespace.
For the duration of the process, the inputs and outputs stay organized in this nested hierarchy.
However, finally the process needs to be represented in the provenance graph by a node, and the inputs and outputs should be linked up.
The link label, which is the only thing that can be used to represent the namespacing, is a string and therefore necessarily one-dimensional.
The nested namespace therefore needs to be mapped onto this flat space.

Currently, the namespaces are just concatenated with underscores.
So the input node in the example would get the link label `name_spaced_input`.
However, now it becomes indistinguishable from an input that was passed to a process with the name `name_spaced_input` in the top level namespace.

We should clearly define the rules for the mapping between nested namespaces and flat link labels.
This should also keep in mind that if there is any risk of overlapping labels when expanding namespaces.
For example, in the current situation, what would happen for the following input port namespace:

```
class NestedProcess(Process):

    @classmethod
    def define(cls, spec):
        super(NestedProcess, cls).define(spec)
        spec.input('name.spaced.input')
        spec.input('name_spaced_output')
```

This is currently legal as far as the process spec is concerned.
The first is a doubly nested namespace and the latter is a single port in the top level namespace.
However, the link labels that will be generated will overlap, making it impossible to link both inputs to the node.

The rules for links should also obey the rules of valid python variable names, as they are used in certain methods for provenance graph traversal, see #2568 

Mentioning @giovannipizzi for discussion

sphuber:
Fixes #2688 

The `Process` and `ProcessNode` duality forces us to define a mapping
between the nested port namespaces of the process and the flat link
label hierarchy of the corresponding node in the provenance graph.

Originally the nested namespace were imploded into a single string,
using a single underscore to concatenate the various port names. This
made it however impossible to distinguish a link label that came from a
nested namespace or simply from a single port that contained an
underscore. Given that underscores are allowed in valid python
identifiers we don't want to forbid the use of those. Instead, we
change the nesting character to be a double underscore and forbid a port
name if it satisfies any of the following rules:

 * Contains more than one consecutive underscore
 * Starts with an underscore
 * Ends with an underscore
 * Is otherwise not a valid python identifier

With these restrictions, mapping the flattened link labels back onto a
nested namespaced dictionary becomes an injective function.

### DIFF
--- a/aiida/common/lang.py
+++ b/aiida/common/lang.py
@@ -14,6 +14,50 @@ from __future__ import absolute_import
 
 import functools
 import inspect
+import keyword
+import six
+
+# Python 3 has a nice built-in solution, but while we support python 2 we need this ugly switch
+# Note that there is a difference between the PY2 and PY3 implementations as the latter allows unicode characters
+if six.PY2:
+
+    def isidentifier(identifier):
+        """Return whether the given string is a valid python identifier.
+
+        :return: boolean, True if identifier is valid, False otherwise
+        :raises TypeError: if identifier is not string type
+        """
+        import string
+
+        type_check(identifier, six.string_types)
+
+        if not identifier:
+            return False
+
+        if keyword.iskeyword(identifier):
+            return False
+
+        first = '_' + string.lowercase + string.uppercase  # pylint: disable=no-member
+        if identifier[0] not in first:
+            return False
+
+        other = first + string.digits
+        for character in identifier[1:]:
+            if character not in other:
+                return False
+
+        return True
+
+else:
+
+    def isidentifier(identifier):
+        """Return whether the given string is a valid python identifier.
+
+        :return: boolean, True if identifier is valid, False otherwise
+        :raises TypeError: if identifier is not string type
+        """
+        type_check(identifier, six.string_types)
+        return identifier.isidentifier() and not keyword.iskeyword(identifier)
 
 
 def type_check(what, of_type, msg=None, allow_none=False):
--- a/aiida/backends/tests/cmdline/utils/test_common.py
+++ b/aiida/backends/tests/cmdline/utils/test_common.py
@@ -26,13 +26,11 @@ class TestCommonUtilities(AiidaTestCase):
         from aiida.cmdline.utils.common import get_node_summary
 
         computer_label = self.computer.name  # pylint: disable=no-member
-        code_label = 'test_code'
 
         code = Code(
             input_plugin_name='arithmetic.add',
             remote_computer_exec=[self.computer, '/remote/abs/path'],
         )
-        code.label = code_label
         code.store()
 
         node = CalculationNode()
@@ -42,7 +40,6 @@ class TestCommonUtilities(AiidaTestCase):
 
         summary = get_node_summary(node)
         self.assertIn(node.uuid, summary)
-        self.assertIn(code_label, summary)
         self.assertIn(computer_label, summary)
 
     def test_get_process_function_report(self):
--- a/aiida/backends/tests/engine/test_ports.py
+++ b/aiida/backends/tests/engine/test_ports.py
@@ -64,3 +64,24 @@ class TestPortNamespace(AiidaTestCase):
         port = InputPort('not_storable')
         port_namespace['not_storable'] = port
         self.assertEqual(port.non_db, namespace_non_db)
+
+    def test_validate_port_name(self):
+        """This test will ensure that illegal port names will raise a `ValueError` when trying to add it."""
+        port = InputPort('port')
+        port_namespace = PortNamespace('namespace')
+
+        illegal_port_names = [
+            'two__underscores',
+            'three___underscores',
+            '_leading_underscore',
+            'trailing_underscore_',
+            'non_numeric_%',
+            'including.period',
+            'disallowed👻unicodecharacters',
+            'white space',
+            'das-hes',
+        ]
+
+        for port_name in illegal_port_names:
+            with self.assertRaises(ValueError):
+                port_namespace[port_name] = port
--- a/aiida/backends/tests/engine/test_process.py
+++ b/aiida/backends/tests/engine/test_process.py
@@ -61,8 +61,8 @@ class TestProcessNamespace(AiidaTestCase):
         self.assertEquals(input_node.value, 5)
 
         # Check that the link of the process node has the correct link name
-        self.assertTrue('some_name_space_a' in proc.node.get_incoming().all_link_labels())
-        self.assertEquals(proc.node.get_incoming().get_node_by_label('some_name_space_a'), 5)
+        self.assertTrue('some__name__space__a' in proc.node.get_incoming().all_link_labels())
+        self.assertEquals(proc.node.get_incoming().get_node_by_label('some__name__space__a'), 5)
 
 class ProcessStackTest(Process):
 
--- a/aiida/backends/tests/engine/test_process_builder.py
+++ b/aiida/backends/tests/engine/test_process_builder.py
@@ -11,9 +11,12 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
 
+from collections import MutableMapping
+
+from aiida import orm
 from aiida.backends.testbase import AiidaTestCase
+from aiida.common import LinkType
 from aiida.engine import WorkChain, Process
-from aiida.orm import Dict, Bool, Float, Int
 from aiida.plugins import CalculationFactory
 
 DEFAULT_INT = 256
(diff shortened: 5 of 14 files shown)


sphuber:
Marked it as `[WIP]` because I know the process builder restart is broken and the tests don't cover it. But in the meantime I wanted to get the discussion going. Where are other parts that this might affect the code that may not be tested? For sure the generated link labels. I haven't added explicit tests for each process type, but have only added the testing directly on the `PortNamespace`. Since any port creation should go through this class, that *should* cover everything. Might there be side-channels?

aiida/common/lang.py
```diff
@@ -14,6 +14,49 @@
...
+else:
+
+    def isidentifier(identifier):
+        """Return whether the given string is a valid python identifier.
+
+        :return: boolean, True if identifier is valid, False otherwise
+        :raises TypeError: if identifier is not string type
+        """
+        type_check(identifier, six.string_types)
+        return identifier.isidentifier() and not keyword.iskeyword(identifier)
```
giovannipizzi: Just as a note (maybe to comment, to avoid to forget) - there is a difference between the PY2 and PY3 implementations - the latter allows unicode characters. Tests pass because below in `validate_port_name` you also check the valid character set.

sphuber: But certain unicode characters *are* valid in identifiers, no?

giovannipizzi: I think: only ASCII are valid in py2, Unicode ([with rules](https://www.python.org/dev/peps/pep-3131/)) is allowed in py3, and in AiiDA we don't want Unicode at this stage, I think. The implementation of this PR is right, my only point here is that the py2 and py3 implementations here do different things on the same string (although they both do the "correct thing" of telling if it's an identifier in the current version of the language). A comment to clarify this would be enough I think.

sphuber: I see, will do

sphuber:
I have added a second commit on top, which implements the `get_builder_restart` for any `ProcessNode`. I found that by changing the `ProcessBuilderNamespace` to a `MutableMapping`, the implementation became very clean, and the same by extension for the more specific `CalcJobNode.get_builder_restart`. The downside is that we now block certain names for ports, but the potential solution of namespacing it would become very ugly. See commit message for example.

sphuber:
@giovannipizzi this is also ready for review

giovannipizzi:
Yes, I'm still a bit puzzled about the best approach about not overriding a bunch of names.
I also see that the approach of sub-namespacing is really ugly.
Do we need to inherit from MutableMapping? Can't we work on a 'clean' class? And maybe define, if needed, hidden methods for what we need, like _keys(), _values(), _items(), ...?
A way to implement, if we need a dictionary, is to have a `_data = MutableMapping()` as a member, and pipe the calls from self._keys() to self._data.keys() and so on?

sphuber:
I guess we only need `__setitem__`, `__delitem__` and `update` to be implemented. Maybe it is not so bad if we don't have other methods like `keys` and `values`. Wouldn't it be possible to "cast" to a regular `dict` and then you would have those? i.e. `dict(builder).keys()`

sphuber:
I tried making the `ProcessBuilderNamespace` a plain object and implementing only `update` as a public method, which mostly worked, except that we can no longer do `**builder`, which is what we used to extract the inputs as a dictionary. For `**` to work, the object **has** to be a mapping. Is there another easy way to extract the nested values as a dictionary from a builder if we don't have this? Will that require implementing our own recursive dictionary builder? And if so, is that worth the trade off of having some additional keywords available?

giovannipizzi:
Indeed, I also tried and it's not possible.
Probably there is no good solution, even if I quite dislike the namespace issue - e.g. I already see people doing `builder.update = Dict()` or `builder.values = Dict()` and then everything would break...

A trick (but possibly very dirty and so maybe worse) is to have the unary `~` operator return a Mapping, so you would do things like `**~builder` or `dict(~builder)`, but I think it risks to be a bit confusing.

If we really have to make this, would it make sense in the link name validator to forbid these keywords?

sphuber:
I still see the builder as a utility, mostly well suited for use in the shell, so I don't think that we should link names wholesale, just because they block a small sub part of the functionality of a utility. The most important part of the builder is that people can discover and set nested port values easily, and then pass to a launcher. If they really want to then use some mapping methods to investigate all contents, one could first cast to normal dict and then do those, i.e. `dict(builder).values()` would work right?

giovannipizzi:
I see your point. Still, I think now the __init__ of the ProcessBuilderNamespace would except if someone defines a `spec.input('values')`, because it will try to set the setter onto the property self.values that cannot be rewritten.

Would it be possible to:
1. add a test that tries to do this
2. modify the code so that it does not crash
3. still allow to access the value via `builder['values']`
3. document this "problem" and the strings that are problematic

Still, thinking better to it, I still a bit unconvinced. I agree I wouldn't block these strings (and I think `values` is a very good input port name, I believe so I guess someone sooner or later will use it). Indeed the builder is an utility, but I fear that if you do `builder.<TAB>` and you expect to see only the inputs, but you see also `values`, `keys`, `update`, ... it is not clear that these are other methods and not possible input links.

Is it so bad to ask to do `**builder.get_dict()` instead of simply **builder? And is the `update` method essential?

sphuber:
Having to do `**builder.get_dict()` wouldn't be bad at all, except that I don't think it will work, because the dictionaries nested within this dictionary would also be `ProcessBuilderNamespaces` and so I think the `**` will then choke if they are not proper `Mappings`. Unless we write our own recursive `get_dict()` unwrapper, maybe that could work.

Regarding the visibility of the mapping methods: they are currently not tab-completed, because we override `dir` to make sure that only the port names are completed.

I just checked as well for the `MutableMapping` case, if I specify a port with name `values`, it *is* tab-completed and I can set it. So it acts as the setter of the port value. Calling it like `builder.values()` will actually try to call the port value, i.e. you would get something like: `TypeError: 'NoneType' object is not callable`.

Playing a bit around, I actually think that the `MutableMapping` *is* the way to go. No port names are reserved and from the builders process port definition perspective everything works. The mapping methods are inaccessible however, but to make these available, one would only need to cast the builder to a `dict`, i.e. `dict(builder).items()` works just fine. Consider this example, where I added the `values` port to the `ArithmeticAddCalculation`.

```
builder = ArithmeticAddCalculation.get_builder()
builder.values = Int(2)
builder.x = Int(3)
In [11]: for key, value in dict(builder).items():
    ...:     print(key, value)
    ...:     
('x', <Int: uuid: 73f95be0-1c8c-4ab9-b9d1-f12f196c6969 (unstored) value: 3>)
('values', <Int: uuid: 12a1dc43-9857-40ab-acea-ccbc3a578442 (unstored) value: 2>)
('metadata', {'options': {}})
```
The tab-completion show:
builder.
                  builder.code        builder.values      
                  builder.metadata    builder.x           
                  builder.submit_test builder.y          
```
I think this gives us everything we want no?

giovannipizzi:
From what I'm saying, I think so!
But did you check that, now that .values() is inaccessible, thinks like `dict(builder)` and `f(**builder)` work?

If you add your example to a test (creating a temp calculation with a `values` input name and checking all the above) I'm happy to approve!! Thanks

sphuber:
OK, I have added a test specifically to test overlapping port names with mapping methods and I think everything works very nicely. In principle:

* All attributes act on ports not as mapping methods, when getting, setting and in auto-complete
* Simply casting `dict(builder)` will make all mapping methods available
* The `update` method has been implemented as `_update` in order to not block it as a port name

Issue #2384, opened 2019-01-11
REST API should be updated to reflect the changes in provenance redesign


sphuber:
An overview of name changes can be found in issue #2152

waychal:
- [ ] Move code endpoint under Data to match it with ORM hierarchy
- [ ] Create Node module as top hierarchy with Data and Process
- [ ] Implement Process hierarchy and update Calculation endpoint accordingly
- [ ] In translator classes, remove hard coded type string from query help and replace it with ORM class
- [ ] Find alternate way to get subclasses of translator class and remove hard coded node type string used there
- [ ] Use click interface to get rest api command line parameters
- [ ] Remove load_dbenv setup from run_api code

PR #2700, opened 2019-04-03
Remove the `enabled` column from the `DbComputer` table

### LINKED ISSUE #1154: Remove 'enabled' column from computer (leave it only in DbAuthInfo)
So that we reduce the number of columns that can change after storing in DbComputer. Anyway the logic is already in DbAuthInfo, one should just remove the column with a migration, making sure that if the computer was disabled, the information is moved to the DbAuthInfo table, and adapt the 'verdi node enable/disable' commands, so that if -u is not specified, instead of setting the disabled/enabled property in DbComputer, it still follows the logic used when -u is specified, but for the `get_automatic_user()` user.

sphuber:
Fixes #1154 

This information can already be specified on a per-user basis in the
`DbAuthInfo`. By removing this user specific configuration the computer
definitions become more broadly usable by multiple users.


sphuber:
Note that I explicitly did not do a data migration of any old `enabled` settings and transferring it to authentication infos. I don't think it is that crucial for quite a bit of extra work

aiida/orm/computers.py
```diff
@@ -108,7 +95,6 @@ def full_text_info(self):
         ret_lines.append(" * UUID:           {}".format(self.uuid))
         ret_lines.append(" * Description:    {}".format(self.description))
         ret_lines.append(" * Hostname:       {}".format(self.hostname))
-        ret_lines.append(" * Enabled:        {}".format("True" if self.is_enabled() else "False"))
```
giovannipizzi: Is there now an easy way to inquire whether the computer is disabled for the current user?

sphuber: In the API there is `get_user_enabled` and the `verdi computer list` still has the normal switches. Just the concept of globally enabled no longer applies

giovannipizzi: ok

sphuber:
@waychal and @asle85 in addition to @giovannipizzi comment regarding this specific change, I added issue #2384 to the list of blocking issues for `1.0.0` which should address updating the REST API in general to all the changes in `1.0.0`. It would be good if we can discuss this together sometime. I will gladly help in identifying the changes that may have affected it

PR #2703, opened 2019-04-03
Remove the legacy workflow tables

### LINKED ISSUE #2380: Remove legacy workflows models and add a migration
The logic of the legacy workflows has already been removed, but the database models were maintained, because once they are removed, a migration will have to be added as well that will drop the tables from existing databases. However, before they are dropped, users might want to dump the data for later reuse. This functionality will have to be implemented first.

sphuber:
Fixes #2380 

The migrations that will drop the tables will first check if there is
any data in the tables, in which the user will be prompted with a
warning message and a request to continue or abort. Should the user
decide to continue, the content of the tables will be serialized to JSON
format and written to a file.


sphuber:
I guess the most important question here is: how complete and correct are the JSON serialized dumps to file of the table content and how correct and complete do we want/require them to be.

I doubt that we can guarantee that the data can be fully recreated from the dumps, in fact it is impossible because the repository is not included. So I think it is justified to not include testing, because what would it be exactly that we would test? I have tested it locally on a test database for both backends and the dumping works. There is just no telling whether the data is consistent and complete.

aiida/backends/djsite/db/migrations/0031_remove_legacy_workflows.py
```diff
@@ -0,0 +1,115 @@
...
+    ]
+
+    operations = [
+        # Export existing data to a JSON file
+        migrations.RunPython(export_workflow_data, reverse_code=migrations.RunPython.noop),
+        migrations.RemoveField(
+            model_name='dbworkflow',
+            name='user',
+        ),
+        migrations.AlterUniqueTogether(
```
giovannipizzi: I guess this is automatically generated - interesting that it first removes each field

aiida/backends/djsite/db/migrations/0031_remove_legacy_workflows.py
```diff
@@ -0,0 +1,115 @@
...
+    count_workflow_data = DbWorkflowData.objects.count()
+    count_workflow_step = DbWorkflowStep.objects.count()
+
+    # Nothing to do if all tables are empty
+    if count_workflow == 0 and count_workflow_data == 0 and count_workflow_step == 0:
+        return
+
+    if not settings.TESTING_MODE:
+        echo.echo('\n')
+        echo.echo_warning('The legacy workflow tables contain data but will have to be dropped to continue.')
```
giovannipizzi: Maybe you can add another line that also says something like "if you really care, make also sure that you save a DB dump of the current version"

aiida/backends/djsite/db/migrations/0031_remove_legacy_workflows.py
sphuber: It was indeed auto generated by `makemigrations`

aiida/backends/djsite/db/migrations/0031_remove_legacy_workflows.py
sphuber: You can pre-approve it. I will make sure to add it before merging

PR #2705, opened 2019-04-04
Change the heuristics of the `AIIDA_PATH` environment variable

### LINKED ISSUE #2337: Automatic creation of the aiida directory fails if AIIDA_PATH is set
When setting `AIIDA_PATH` in the environment, the `.aiida` folder is not created automatically and it defaults to the home directory `.aiida` (which is does create automatically). Creating it manually works. Could we consider to change this behavior?

sphuber:
Fixes #2337 

The location of the configuration folder will be determined and
optionally created following these heuristics:

 * If the `AIIDA_PATH` variable is set, all the paths will be checked
   to see if they contain a configuration folder. The first one to be
   encountered will be set as `AIIDA_CONFIG_FOLDER`. If none of them
   contain one, a configuration folder will be created in the last
   path considered.

 * If the `AIIDA_PATH` variable is not set the `DEFAULT_AIIDA_PATH`
   value will be used as base path and if it does not yet contain a
   configuration folder, one will be created.

This new approach will guarantee that a configuration folder will
always be set and created if necessary. The one downside is that if
someone sets the environment variable to the wrong value, a config
folder will be created without a warning.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -112,8 +112,6 @@ pipeline {
                         environment name: 'RUN_ALSO_DJANGO', value: 'true'
                     }
                     steps {
-                        // An empty .aiida folder must be present in the AIIDA_PATH
-                        sh 'mkdir -p "$AIIDA_PATH/.aiida"'
                         sh '.ci/setup_profiles.sh'
                         sh '.ci/before_script.sh'
                         sh '.ci/test_script.sh'
@@ -131,8 +129,6 @@ pipeline {
                         environment name: 'RUN_ALSO_SQLALCHEMY', value: 'true'
                     }
                     steps {
-                        // An empty .aiida folder must be present in the AIIDA_PATH
-                        sh 'mkdir -p "$AIIDA_PATH/.aiida"'
                         sh '.ci/setup_profiles.sh'
                         sh '.ci/before_script.sh'
                         sh '.ci/test_script.sh'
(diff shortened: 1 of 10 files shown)


sphuber:
Pinging @espenfl and @greschd if they are interested to comment

greschd:
Looks good to me.

Just to make sure I understand it right, if AIIDA_PATH is set it will not fall back to DEFAULT_AIIDA_PATH, correct? And the other difference is that there's no more going up the directory tree?

It could probably be debated whether it should be the _first_ one in AIIDA_PATH which gets the newly created config dir (since that's the first one to be checked), but not sure there's a clear right or wrong way to do that.

So yeah, all good for me.

Am 05.04.2019 00:02 schrieb Coveralls <notifications@github.com>:

[Coverage Status]<https://coveralls.io/builds/22598945>

Coverage decreased (-5.9%) to 64.892% when pulling 3968589<https://github.com/aiidateam/aiida_core/commit/3968589149d701a0a6bc3e366a2d731492e25f56> on sphuber:fix_2337_aiida_path into a297600<https://github.com/aiidateam/aiida_core/commit/a2976007d3c3e5403597b389ad6cbe72aa522e3e> on aiidateam:develop.

sphuber:
> Looks good to me. Just to make sure I understand it right, if AIIDA_PATH is set it will not fall back to DEFAULT_AIIDA_PATH, correct?
Yes, if `AIIDA_PATH` is set, it will either find an existing folder in there and use it, or create one.

> And the other difference is that there's no more going up the directory tree?
Yes, in no cases will a directory be traversed upwards in search of a config directory

> It could probably be debated whether it should be the _first_ one in AIIDA_PATH which gets the newly created config dir
This is currently just the way I implemented it because the original implementation allowed for multiple paths to be set in `AIIDA_PATH`. To keep this, I had to decide what to do if none contained a config directory, since we decided that whatever the case, a directory would be created. My reasoning is that if you have multiple paths specified, but none contain a directory, then it does not really matter which one we take to create one in. Happy to reconsider if there are arguments against this

sphuber:
Closing until I figure out the cause of the single test failing

PR #2706, opened 2019-04-04
Replace `TextField` stored for storing JSON with proper `JSONField`

### LINKED ISSUE #2699: Replace `TextField` columns in Django models for JSONB
Now that Django supports JSONB, maybe it is better to use that instead of `TextField` which was used as a temporary replacement in the original schema.
Fields that are affected:

    * `DbComputer.metadata`
    * `DbComputer.transport_params`
    * `DbAuthInfo.auth_params`
    * `DbAuthInfo.metadata`
    * `DbLog.metadata`

Ideally, this is done before the release of `v1.0.0` as this requires database migrations

sphuber:
Fixes #2699 

The following model fields were using a text field to store JSON:

 * `DbComputer.metadata`
 * `DbComputer.transport_params`
 * `DbAuthInfo.auth_params`
 * `DbAuthInfo.metadata`
 * `DbLog.metadata`

Now that Django supports a JSON field for the Postgres JSONB field it is
better to replace the old implementation such that these columns become
properly queryable and we are no longer responsible for proper
serialization and deserialization

### DIFF
--- a/aiida/backends/djsite/db/migrations/0033_replace_text_field_with_json_field.py
+++ b/aiida/backends/djsite/db/migrations/0033_replace_text_field_with_json_field.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+# pylint: disable=invalid-name,too-few-public-methods
+"""Replace use of text fields to store JSON data with builtin JSONField."""
+from __future__ import unicode_literals
+from __future__ import absolute_import
+
+# Remove when https://github.com/PyCQA/pylint/issues/1931 is fixed
+# pylint: disable=no-name-in-module,import-error,no-member
+import django.contrib.postgres.fields.jsonb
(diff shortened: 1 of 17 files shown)


szoupanos:
Thanks @sphuber !
@giovannipizzi please let me do the final merge since I would like to compare it with my work on JSONB (overlaps, conflicts etc)

sphuber:
When you do, please use rebase and merge (so not squash or normal merge) because I think it is important to keep these two commits separate

sphuber:
@giovannipizzi yes they were all already JSONB, and you are right, I will add some tests

sphuber:
I added tests for all 5 columns on the first commit

szoupanos:
Thanks Seb. Very minor conflicts on my side.
Please proceed with the merge

PR #2707, opened 2019-04-04
Added `delete_all` and `delete_many` methods for comments and logs

### LINKED ISSUE #2683: Implement `delete_many` method for CommentCollection's for both Django and SQLA
The `delete_many` method should be implemented for `DjangoCommentCollection` and `SqlaCommentCollection`.

It should take `filters` as a dict parameter similar to the known `filters` parameter for the QueryBuilder.

### LINKED ISSUE #2684: Implement `delete_many` method for LogCollection's for both Django and SQLA
The `delete_many` method should be implemented for `DjangoLogCollection` and `SqlaLogCollection`.

It should take `filters` as a dict parameter similar to the known `filters` parameter for the QueryBuilder.

CasperWA:
Fixes #2683.
Fixes #2684.

The methods are implemented specifically for each backend in their `Collection`s.

#### Added methods
- `delete_all`: Deletes all entities in the Collection.
- `delete_many`: Delete one or more entities based on a QueryBuilder filter.

Methods are added for Logs and Comments and for both backends (Django and SQLAlchemy).
A `utils.py` file is also added in `aiida.orm.implementations` for utility functions for `delete_many`.

#### The intended logic for `delete_many` is:
- When an empty dict is provided, delete all entities.
- If an entity cannot be found by using unique identifier filter keys, such as `id` or `uuid`, do not delete anything and raise `NotExistent` with helpful message.
- If an entity cannot be found by using any other filter keys, do not raise.

The reasoning here is that if a user chooses to delete comments for a series of nodes, the result shouldn't raise if no comments are found for one or more of the nodes, since this is exactly the desired outcome from the user's point-of-view.
However, if a user chooses to filter directly on a specific `id` of a comment, then it might be helpful to find that the comment didn't exist in the first place.

#### Tests
Several tests are added to test `delete_many` and one for `delete` and `delete_all`.
Furthermore, a few tests for the new utility functions `ignore_qb_operators` and `get_filter_values` have been added.

Test files for Logs and Utils are new and the former is based on the similar one for Comments.
They are included in `__init__.py` for tests.

#### Minor corrections:
Updated exiting test `test_creation` for Comments.


sphuber:
Question: why do we need `delete_all` if `delete_many()` without filters does the same thing?

CasperWA:
> Question: why do we need `delete_all` if `delete_many()` without filters does the same thing?

Yes.


No, but well, I presume to have a general function that you don't need to supply any inputs, but simply wipes a table and the entries' corresponding repository files?
Also, I think it may be because it was there originally?

sphuber:
In the current develop I cannot find any other `delete_all` in the ORM. In any case, I think it is fine to keep the `delete_all` for internal use, but I would definitely remove the behavior of `delete_many` deleting everything on empty filters. Imagine someone passing a variable `filters=filters` where it is generated dynamically and due to a mistake the filters are empty. Poof!  Let's just make `delete_many` raise when no filters are specified. Then there is also a right to existence for `delete_all` :+1:

CasperWA:
You got it! :sunglasses:

CasperWA:
> In the current develop I cannot find any other `delete_all` in the ORM.

I don't think it makes sense for any other ORM entities to have it either. It makes sense for Comments and Logs though - more so Comments than Logs, but still. Right? :)

aiida/backends/tests/orm/test_comments.py
```diff
@@ -73,6 +71,49 @@ def test_comment_collection_delete(self):
...
+        # Delete all Comments
+        Comment.objects.delete_all()
+
+        with self.assertRaises(exceptions.NotExistent):
+            Comment.objects.delete(comment_pk)
+
+        with self.assertRaises(exceptions.NotExistent):
+            Comment.objects.get(id=comment_pk)
+
+    def test_comment_collection_delete_many(self):
```
sphuber: Is there a test to check that passing empty filters raises?

aiida/orm/comments.py
```diff
@@ -34,6 +34,23 @@ def delete(self, comment_id):
...
+            self._backend.comments.delete_all()
+
+        def delete_many(self, filters):
+            """
+            Delete Comments based on `filters`.
+            If `filters` is an empty dict, raise ValidationError.
+            If at least one entity is found, that entity is deleted, nothing is raised.
+
+            :param filters: similar to QueryBuilder filters
+            :type filters: dict
```
sphuber: add the note of exception being raised as `:raises ValidationError: if filters is an empty dictionary`

aiida/orm/implementation/comments.py
```diff
@@ -79,9 +79,27 @@ def create(self, node, user, content=None, **kwargs):
...
+    def delete_all(self):
+        """
+        Delete all Comments from the collection
+        """
+
+    @abc.abstractmethod
+    def delete_many(self, filters):
+        """
+        Delete Comments based on `filters`.
+        If `filters` is an empty dict, raise ValidationError.
```
sphuber: Same add as `:raises ...:`

aiida/orm/implementation/django/comments.py
```diff
@@ -117,16 +120,81 @@ def create(self, node, user, content=None, **kwargs):
...
-        from django.core.exceptions import ObjectDoesNotExist
-        assert comment is not None
+        # Check
+        if not isinstance(comment_id, int):
+            raise TypeError("comment_id must be an int")
+
+        # Delete
         try:
-            models.DbComment.objects.get(id=comment).delete()
+            with transaction.atomic():
```
sphuber: Why does the deletion of a single object need to go in an atomic transaction?

aiida/orm/implementation/django/comments.py
```diff
@@ -117,16 +120,81 @@ def create(self, node, user, content=None, **kwargs):
...
+        if not filters:
+            raise exceptions.ValidationError("filters must not be empty")
+
+        # Initialization
+        not_found_entities = {}
+        column_mapping = {'node_id': 'dbnode_id'}
+        unique_keys = ['id', 'uuid']
+
+        # Special case for unique key values returning single entities
+        for key in unique_keys:
```
sphuber: I am a bit befuddled by this logic here. What is it you are trying to do here? Why are you manipulating the filters passed by the user in the arguments?

CasperWA: Since this comment reiterates the general comment of this review, I'll answer here.
This logic is quite convoluted, yes. Even though it has gone through some iterations. Its point is to let the user use QueryBuilder (qb) filters, along with the operators available there. So I am essentially trying to parse the filter. However, I see now that I am, indeed, not doing it properly, since the `ignore_qb_operators` utility function basically does what it says on the box; ignore the operators, i.e., the operators have no effect.
Do we have a utility function to parse qb filters meaningfully, not creating SQL queries ....

Wait, I have just remembered (it's been a long time since I wrote this PR). This logic essentially shouldn't be here, since the filter should be passed directly to the qb to retrieve the PKs. However, there must have been some problem with this, since I did this logic.
Let me just do some further testing and I'll revise this accordingly...

CasperWA: Ah! Now I remember. The problem is that if a user specifically requests to delete two Comments (or Logs) according to a unique identifier, then if one of the two entities do not exist, the QueryBuilder will not raise, but simply return the one entity that does exist. Hence, the user will never know that the non-existing entity could not be found (which may be annoying if it was a simple typo).

This (incorrect) logic was an attempt at providing the user with this information. However, since it doesn't handle the `or` and `and` operators correctly, it should either be removed or updated.

aiida/backends/tests/orm/implementation/test_comments.py
```diff
@@ -98,3 +116,299 @@ def test_creation_with_time(self):
...
+        for comment_uuid in comment_uuids:
+            self.assertIn(comment_uuid, found_comments_uuid)
+
+        # Delete last two comments (comment2, comment3)
+        filters = {'or': [{'id': comment2.id}, {'uuid': str(comment3.uuid)}]}
+        self.backend.comments.delete_many(filters=filters)
+
+        # Check they were deleted
+        builder = orm.QueryBuilder().append(orm.Comment, project='uuid')
+        self.assertEqual(builder.count(), no_of_comments - 2)
```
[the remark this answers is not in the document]
CasperWA: Not exactly, since I am not deleting all Comments between tests (to avoid using `delete_all()` outside of a class that tests this exact method). However, I have rewritten it using filters for the QueryBuilder, making it possible to do this.

aiida/backends/tests/orm/test_comments.py
CasperWA: Not here, but under `.implementation.test_comments` there is. Do you want one here as well?

aiida/orm/implementation/django/logs.py
```diff
@@ -95,24 +97,59 @@ def delete(self, log_id):
...
+        """
+        Delete all Log entries.
+
+        :raises `~aiida.common.exceptions.IntegrityError`: if all Logs could not be deleted
+        """
+        from django.db import transaction
+        try:
+            with transaction.atomic():
+                models.DbLog.objects.all().delete()
+        except Exception as exc:
```
sphuber: surprised pylint doesn't throw a fit here about the broad-exception. This file is not excluded from pre-commit no?

CasperWA: Nope - pylint just knows about my awesomeness, and cannot bear to complain :smile:

PR #2708, opened 2019-04-04
Change the heuristics of the `AIIDA_PATH` environment variable

### LINKED ISSUE #2337: Automatic creation of the aiida directory fails if AIIDA_PATH is set

sphuber:
Fixes #2337 

The location of the configuration folder will be determined and
optionally created following these heuristics:

 * If the `AIIDA_PATH` variable is set, all the paths will be checked
   to see if they contain a configuration folder. The first one to be
   encountered will be set as `AIIDA_CONFIG_FOLDER`. If none of them
   contain one, a configuration folder will be created in the last
   path considered.

 * If the `AIIDA_PATH` variable is not set the `DEFAULT_AIIDA_PATH`
   value will be used as base path and if it does not yet contain a
   configuration folder, one will be created.

This new approach will guarantee that a configuration folder will
always be set and created if necessary. The one downside is that if
someone sets the environment variable to the wrong value, a config
folder will be created without a warning.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -112,8 +112,6 @@ pipeline {
                         environment name: 'RUN_ALSO_DJANGO', value: 'true'
                     }
                     steps {
-                        // An empty .aiida folder must be present in the AIIDA_PATH
-                        sh 'mkdir -p "$AIIDA_PATH/.aiida"'
                         sh '.ci/setup_profiles.sh'
                         sh '.ci/before_script.sh'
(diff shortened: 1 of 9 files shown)


sphuber:
Github wouldn't allow me to open #2705 so I had to create a new PR. Ah well

PR #2712, opened 2019-04-05
Update requirement for dependency `psycopg==2.8`

### LINKED ISSUE #2710: Update requirement for dependency `psycopg==2.8`

sphuber:
Fixes #2710 

This should get rid of the warning about the package moving the binary
builds to a the separate package `psycopg2-binary`. Whether we want to
use the binary build remains to be decided.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -32,7 +32,7 @@ dependencies:
 - tabulate==0.8.3
 - ete3==3.1.1
 - uritools==2.2.0
-- psycopg2==2.7.6
+- psycopg2==2.8
 - paramiko==2.4.2
 - ecdsa==0.13
 - ipython>=4.0,<6.0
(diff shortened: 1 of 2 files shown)


ltalirz:
Just saying that building the psycopg2 wheel takes 9s on my machine while downloading the wheel takes just 1.4s.

sphuber:
@ltalirz I updated the commit message to explain the situation and decision. I noticed that pre-commit checker for `environment.yml` has an explicit exception for `pyscopg2-binary` and will always map it to `psycopg2`. Is this still correct?

ltalirz:
> I noticed that pre-commit checker for environment.yml has an explicit exception for pyscopg2-binary and will always map it to psycopg2. Is this still correct?

Yes, it is still correct. conda always provides binary packages and the correct package is still named `psycopg2`

PR #2715, opened 2019-04-05
Set upper limit for `pip` in `setup.py` to `<19.0.0`

### LINKED ISSUE #2714: Update `setup.py` to also check for `pip<19.0.0`
For now `pip>=19.0.0` still breaks our build process. I have had to help a few too many people because of this. We should update the check in `setup.py`

sphuber:
Fixes #2714 

The build process is broken for versions bigger than this limit.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -10,7 +10,7 @@ node {
 }
 
 pipeline {
-    /* The tutorial was setting here agent none, and setting the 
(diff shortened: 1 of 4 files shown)


ltalirz:
by the way, I organized a brief skype call with Rico yesterday. he gave me some pointers on where to look for the reentry issues.
I'll see whether I can make some progress over the weekend.

PR #2718, opened 2019-04-05
Remove duplicate `verdi` command tests

### LINKED ISSUE #2717: Tests for `verdi node` and `verdi data` are virtually identical
Having many unit tests is great but simply duplicating tests in different files seems ineffective.
The two files `aiida/backends/tests/cmdline/commands/test_node.py` and `aiida/backends/tests/cmdline/commands/test_data.py` are almost identical with the latter adding tests for `verdi data upf`. I propose to remove the former

sphuber:
Fixes #2717 

The tests in `aiida/backends/tests/cmdline/commands/test_node.py`
and `aiida/backends/tests/cmdline/commands/test_data.py` were
essentially identical, with the latter adding some tests, so the former
is deleted.


giovannipizzi:
less is more

PR #2719, opened 2019-04-05
Implement generic list print function for command line

### LINKED ISSUE #2408: verdi user list "color option"
verdi user list has a "--color" option

 * on my terminal, the behavior is the reverse of the expected one: the one user I currently have is colored red in output, if I do *not* specify the --color option (and it *isn't* colored when I specify it)
 * do we need this option? how are the users colored?

sphuber:
Fixes #2408 

Multiple `verdi` commands will print a simple list of entries in a
collection where optionally some entries are highlighted. This concept
is abstracted in `aiida.cmdline.utils.echo.echo_formatted_list` to make
the output across the commands more homogeneous. The command that
implement it for now:

 * `verdi computer list`
 * `verdi profile list`
 * `verdi user list`

The function provides optional lambdas that can help sort the
collection, determine highlighting and which entries should be hidden.

### DIFF
--- a/aiida/cmdline/commands/cmd_computer.py
+++ b/aiida/cmdline/commands/cmd_computer.py
@@ -388,50 +388,22 @@ def computer_disable(computer, user):
 @with_dbenv()
 def computer_list(all_entries, raw):
     """List available computers."""
-    from aiida import orm
-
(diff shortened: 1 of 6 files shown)


ltalirz:
very nice!

PR #2723, opened 2019-04-08
Add config file option

ltalirz:
This PR adds a `--config` option to

 * `verdi quicksetup`
 * `verdi setup`
 * `verdi code setup`
 * `verdi computer setup`

that allows to simply read parameters from a config file in yaml format.

E.g. create a file `config.yaml`:
```yaml
---
first_name: Leopold
last_name: Talirz
institution: EPFL
profile_name: testing
backend: django
email: 123@234.de
```

And run
`verdi quicksetup --config config.yaml`

To do:
 - [x] use `functools.partial` (or similar) to use the yaml config file provider by default so that one can use it simply as `options.CONFIG_FILE`. I played a bit around with this but somehow didn't manage to get it to work - would appreciate help on this (should not be too hard)
 - [x] to decide: it adds a dependency on `click_config_file`, which is a very minimal (1 file). we could also decide to include it inside AiiDA but I'd be fine with adding the dependency.
 - [x] once decisions have been made, extend this to other commands
    - [x] computer setup
    - [x] code setup

### DIFF
--- a/aiida/backends/tests/cmdline/commands/test_code.py
+++ b/aiida/backends/tests/cmdline/commands/test_code.py
@@ -96,6 +96,27 @@ class TestVerdiCodeSetup(AiidaTestCase):
         self.assertClickResultNoException(result)
         self.assertIsInstance(Code.get_from_string('{}'.format(label)), Code)
 
+    def test_from_config(self):
+        """Test setting up a code from a config file"""
+        from aiida.orm import Code
+        import tempfile
+        import os
+
+        label = 'noninteractive_config'
+
+        with tempfile.NamedTemporaryFile('w') as handle:
+            handle.write("""---
+label: {l}
+input_plugin: arithmetic.add
+computer: {c}
+remote_abs_path: /remote/abs/path
+""".format(l=label, c=self.comp.name))
+            handle.flush()
+            result = self.cli_runner.invoke(setup_code, ['--non-interactive', '--config', os.path.realpath(handle.name)])
+
+        self.assertClickResultNoException(result)
+        self.assertIsInstance(Code.get_from_string('{}'.format(label)), Code)
+
     def test_mixed(self):
         from aiida.orm import Code
         label = 'mixed_remote'
--- a/aiida/cmdline/params/options/config.py
+++ b/aiida/cmdline/params/options/config.py
@@ -24,41 +24,42 @@ from .overridable import OverridableOption
 
 def yaml_config_file_provider(file_path, cmd_name):  # pylint: disable=unused-argument
     """Read yaml config file."""
-    try:
-        with open(file_path, 'r') as handle:
-            return yaml.load(handle)
-    except IOError:
-        return {}
+    with open(file_path, 'r') as handle:
+        return yaml.load(handle)
 
 
 class ConfigFileOption(OverridableOption):
     """
-    Wrapper around click_config_file.configuration_option that increases reusability
+    Wrapper around click_config_file.configuration_option that increases reusability.
 
     Example::
 
         CONFIG_FILE = ConfigFileOption('--config', help='A configuration file')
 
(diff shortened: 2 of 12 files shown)


sphuber:
You had merged in my unmerged PR so now there were a bunch of conflicts. I rebased and fixed the issues and force pushed. I kept a backup of your original branch on my fork just in case. I will no go through it

aiida/cmdline/params/options/config.py
```diff
@@ -0,0 +1,74 @@
...
+import click_config_file
+
+from .overridable import OverridableOption
+
+
+def yaml_config_file_provider(file_path, cmd_name):  # pylint: disable=unused-argument
+    """Read yaml config file."""
+    try:
+        with open(file_path, 'r') as handle:
+            return yaml.load(handle)
```
sphuber: What will `yaml.load` throw if the content is invalid `yaml`? Should we also catch this and just return empty dict? Maybe add a unit test for this option

sphuber:
Aside from the one comment, looks really good and I like it a lot. You have my blessing to add it to the other commands you proposed as well :+1:  Regarding the option checkboxes, I think the first you have already fixed, correct? For the second, I would prefer to just add the dependency instead of copy pasting the file. Last but not least, yes please add to code and computer setup

aiida/cmdline/params/options/config.py
ltalirz: Actually, what I would prefer is for the exception to simply propagate to the top - I put this try/except in as a hack because the error message was unusable.
I've opened a PR on `click_config_file` to fix this https://github.com/phha/click_config_file/pull/2

ltalirz: Fixing the error message surfaced another problem: by default, the option is *always* looking for a config file (by default in a standard directory with a standard name). I.e. if I don't put the try/except here it will always fail.

This is being addressed here (this PR is still to be discussed): https://github.com/phha/click_config_file/pull/4

Since the error messages now are processed correctly, a yaml parser error would just bubble up to the top, so I think it's fine for us for the moment.

aiida/backends/tests/cmdline/commands/test_code.py
```diff
@@ -96,6 +96,27 @@ def test_noninteractive_upload(self):
...
 
+    def test_from_config(self):
+        """Test setting up a code from a config file"""
+        from aiida.orm import Code
+        import tempfile
+        import os
+
+        label = 'noninteractive_config'
+
+        with tempfile.NamedTemporaryFile('w', delete=False) as handle:
```
sphuber: I think you are not deleting the test config files here, correct?

aiida/cmdline/params/options/config.py
```diff
@@ -0,0 +1,76 @@
...
+        computer_setup --config config.yml
+
+    with config.yml::
+
+        ---
+        computer_name: computer1
+
+    """
+
+    def __init__(self, *args, **kwargs):  # pylint: disable=super-init-not-called
```
sphuber: I think it is better here to just create the correct `kwargs` and then call the super, instead of completely overriding.

ltalirz:
>  I did the rebase locally and can push it if you agree.

Sure, please do.

sphuber:
> Sure, please do.

Done

aiida/backends/tests/cmdline/commands/test_code.py
ltalirz: thanks for pointing this out, fixed (also in other places)

aiida/cmdline/params/options/config.py
ltalirz: done

ltalirz:
Given passing tests (always a nice assumption...) this should be good to merge.

PR #2736, opened 2019-04-10
Remove the `CalcJobNode.options` shortcut property

### LINKED ISSUE #2735: Remove the `CalcJobNode.options` shortcut
This can lead to confusion with the `get_option` method, which retrieve actual option values set on the node, whereas `options` will return the options of the process spec which created the node. By removing the shortcut, one will have to do `self.process_class.inputs` instead, which is not so bad but makes it very clear that you are accessing a property of the process.

sphuber:
Fixes #2735 

This can lead to confusion with the `get_option` method, which retrieves
actual option values set on the node, whereas `options` will return the
options of the process spec which created the node. By removing the
shortcut, one will have to do `self.process_class.inputs` instead, which
is not so bad but makes it very clear that you are accessing a property
of the process.

### DIFF
--- a/aiida/engine/processes/process.py
+++ b/aiida/engine/processes/process.py
@@ -111,15 +111,31 @@ class Process(plumpy.Process):
             self.set_logger(self.node.logger)
 
     @classproperty
-    def exit_codes(self):
-        """
-        Return the namespace of exit codes defined for this WorkChain through its ProcessSpec.
+    def exit_codes(cls):  # pylint: disable=no-self-argument
+        """Return the namespace of exit codes defined for this WorkChain through its ProcessSpec.
+
         The namespace supports getitem and getattr operations with an ExitCode label to retrieve a specific code.
         Additionally, the namespace can also be called with either the exit code integer status to retrieve it.
 
         :returns: ExitCodesNamespace of ExitCode named tuples
         """
-        return self.spec().exit_codes
+        return cls.spec().exit_codes
+
+    @classproperty
+    def spec_metadata(cls):  # pylint: disable=no-self-argument
+        """Return the metadata port namespace of the process specification of this process.
+
+        :return: metadata dictionary
+        """
+        return cls.spec().inputs['metadata']
+
+    @classproperty
+    def spec_options(cls):  # pylint: disable=no-self-argument
+        """Return the metadata options port namespace of the process specification of this process.
+
+        :return: options dictionary
+        """
+        return cls.spec_metadata['options']  # pylint: disable=unsubscriptable-object
 
     @property
     def node(self):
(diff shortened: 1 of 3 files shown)


aiida/engine/processes/process.py
```diff
@@ -121,6 +121,22 @@ def exit_codes(self):
...
+    @classproperty
+    def spec_metadata(self):
+        """Return the metadata port namespace of the process specification of this process.
+
+        :return: metadata dictionary
+        """
+        return self.spec().inputs['metadata']
+
+    @classproperty
+    def spec_options(self):
```
ltalirz: perhaps replace with cls for clarity?

sphuber: I did, but then pylint complains because it does not understand our `classproperty` decorator. I think across the code base we now use `self` instead of `cls`. Maybe it is indeed better to just add the `pylint` disable statement

aiida/orm/nodes/process/calculation/calcjob.py
```diff
@@ -215,7 +207,7 @@ def get_options(self):
         :return: dictionary of the options and their values
         """
         options = {}
-        for name in self.options.keys():
+        for name in self.process_class.spec_options.keys():
```
CasperWA: Will this introduce an actual change in the behavior of the method?
Since `get_option()` will still act on the node, the extra keys that will be found through this will always be None from the node options.
Furthermore, if one should change the process class spec, removing a key, I guess this method, `get_options`, won't actually find all options set for this node. Unless your intention is to ignore "legacy" options for a node, even though they exist?

sphuber: It does *not* change the behavior of the method before this change. However, you are right, if the process spec of the process class is changed and for example removes an option, then this will no longer find it. The problem is however, that "options" are simply stored as attributes and just from their key in the attributes it is impossible to determine whether it is actually an option or a random other attribute. The only solution would be to put a requirement on the key of option attributes and say for example they start with `option_` prefix. However, this would require a data migration that is far from easy. For example, how do you determine for the currently existing attributes which are options and which are not?

CasperWA: > ... The only solution would be to put a requirement on the key of option attributes and say for example they start with `option_` prefix. ...

Or they could be separated into a single `options` attribute dict.

> ... However, this would require a data migration that is far from easy. For example, how do you determine for the currently existing attributes which are options and which are not?

Yes. Yes, this is not something that seems like an urgent thing at all to do, nor even something that is necessary I think?
If I should shoot in the dark I would suggest going through the attributes based on the process class spec, but this will not necessarily get all option items in the end. Yeah --- let's not :)

aiida/engine/processes/process.py
sphuber: I have changed `self` to `cls`

aiida/orm/nodes/process/calculation/calcjob.py
sphuber: So we'll leave it as it is for now

PR #2737, opened 2019-04-10
Refactor the Orbital code and fix bugs

### LINKED ISSUE #2419: OrbitalFactory('realhydrogen') broken
version: 1.0.0a4

Since there is no 'realhydrogen' entry-point in the ``setup.py``, the ``OrbitalFactory('realhydrogen')`` no longer works. Is this intentional or not?

giovannipizzi:
The code was not tested (and now we add tests)
Moreover, the code required quite some cleanup and
refactoring.
Now the code should be cleaner (even if probably still
not perfect). The intended aim of this commit is to avoid
breaking backward-compatibility.

This fixes #2419


giovannipizzi:
Marked as [WIP] because I'd like @DanielMarchand to test if now this works (you can check out the branch in my fork) and if I didn't change the API.

Also, @sphuber: I apparently get some weird errors from prospector in my tests (at least [In my previous runs](https://travis-ci.org/giovannipizzi/aiida_core/jobs/518364028), like cyclic imports etc.) - is this related to the caching of pip or to something else?

aiida/backends/tests/tools/data/orbital/test_orbitals.py
```diff
@@ -0,0 +1,172 @@
...
+                'magnetic_number': 0,
+                'radial_nodes': 2,
+                'spin': 1,
+                'diffusivity': 3.1
+            })
+        self.assertEqual(orbital.get_orbital_dict()['spin'], 1)
+        self.assertEqual(orbital.get_orbital_dict()['diffusivity'], 3.1)
+
+        with self.assertRaises(ValidationError) as exc:
+            _ = RealhydrogenOrbital(
```
sphuber: No need to assign to placeholder if you are just catching exception from constructor

aiida/backends/tests/tools/data/orbital/test_orbitals.py
```diff
@@ -0,0 +1,172 @@
...
+                    'angular_momentum': 1,
+                    'magnetic_number': 0,
+                    'radial_nodes': 2,
+                    'spin': 1,
+                    'diffusivity': "a"
+                })
+        self.assertIn("diffusivity", str(exc.exception))
+
+        with self.assertRaises(ValidationError) as exc:
+            _ = RealhydrogenOrbital(
```
sphuber: same

aiida/tools/data/orbital/realhydrogen.py
```diff
@@ -153,86 +321,53 @@ def _validate_keys(self, input_dict):
...
         angular_momentum=1 and magnetic_number=1 will return "Px"
         """
-        # importing a copy of the conversion_dict
-        convert_ref = copy.deepcopy(conversion_dict)
-        orbital_name = [x for x in convert_ref if
-                any([convert_ref[x][y]['angular_momentum'] ==
-                angular_momentum for y in convert_ref[x]])]
-        if len(orbital_name) == 0:
+        # importing a copy of the CONVERSION_DICT
+        convert_ref = copy.deepcopy(CONVERSION_DICT)
```
sphuber: You don't seem to be mutating the `convert_ref` so do you really need a deep copy here?

DanielMarchand:
> Marked as [WIP] because I'd like @DanielMarchand to test if now this works (you can check out the branch in my fork) and if I didn't change the API.

I, have just tested and the import works now! Thanks!!

aiida/tools/data/orbital/realhydrogen.py
giovannipizzi: indeed - I just left the old logic but just renamed to have pylint happy - now I fixed this

giovannipizzi:
@sphuber I fixed your comments and I also added a couple more tests - I think this is ready to be merged (if tests pass)

sphuber:
@giovannipizzi the linter is known to be a bit buggy when it comes to cyclic imports. It is a bit weird that touching other files seems to trigger it. You could just try to add the disable statement in one of the files that it highlights as problematic. The only thing I can see quickly is that `aiida.engine.processes.ports` import `validate_link_label` from `aiida.orm.utils.links` and in that same module, `LinkManager.nested` calls `from aiida.engine.processes.ports import PORT_NAMESPACE_SEPARATOR`. I thought that if these things are within the scope of methods they should be lazily called and so fine.

giovannipizzi:
The thing that is actually needed is `PORT_NAMESPACE_SEPARATOR`. Can I move it one level up, in `aiida/engine/processes/__init__.py`? Anyway it's used also in `aiida/engine/processes/process.py` so it seems to me that it is a more appropriate place, and this would avoid the cyclic import?
Should I move also `PORT_NAME_MAX_CONSECUTIVE_UNDERSCORES`?

sphuber:
> The thing that is actually needed is `PORT_NAMESPACE_SEPARATOR`. Can I move it one level up, in `aiida/engine/processes/__init__.py`? Anyway it's used also in `aiida/engine/processes/process.py` so it seems to me that it is a more appropriate place, and this would avoid the cyclic import?
> Should I move also `PORT_NAME_MAX_CONSECUTIVE_UNDERSCORES`?

:+1: Seems fair enough

giovannipizzi:
I added a commit that should solve the cyclic import problem - to avoid other analogous problems, the right approach seems to be to define the constants in a submodule `constants.py` and then expose the relevant values one level up.

giovannipizzi:
py3.6 linter is pickier... and the problem got worse!
https://travis-ci.org/giovannipizzi/aiida_core/jobs/518832224#L1541
Any further suggestions?

sphuber:
@giovannipizzi  I ended up fixing the problem in a different way, no more cyclic imports

sphuber:
Thanks a lot @DanielMarchand

PR #2738, opened 2019-04-10
add Framework :: AiiDA trove classifier

ltalirz:


ltalirz:
@sphuber if there are still issues with the deployment of the b2, perhaps we can even get this in before?

sphuber:
Unfortunately I already [released it manually this morning](https://pypi.org/project/aiida-core/1.0.0b2/). Haven't had the time yet to advertise where needed. Do we need to telegraph this release?

ltalirz:
what about writing to the plugin-migration channel from the tutorial, if people are interested?
I think we don't need more than this at this point.

PR #2741, opened 2019-04-11
Upgrade reentry dependency

ltalirz:
I've just released a new alpha version of reentry, which should work with pip 19 (and before).

 * upgrade reentry dependency to 1.3.0a1
 * remove `reentry_scan` from setup.py. The hook has been deprecated, see [reentry changelog](https://github.com/DropD/reentry/blob/develop/CHANGELOG.md) for details
 * remove pip dependency from build system (should not be needed)
 * remove check for pip version in setup.py
 * remove check for pip from setup.py
 * remove pinning of pip version from travis

### DIFF
--- a/docs/requirements_for_rtd.txt
+++ b/docs/requirements_for_rtd.txt
@@ -57,7 +57,7 @@ python-memcached==1.59
 python-mimeparse==1.6.0
 pytz==2018.9
 qe-tools==1.1.2
-reentry==1.3.0a1
+reentry>=1.3.0
 seekpath==1.8.4
 simplejson==3.16.0
 singledispatch>=3.4.0.3; python_version<'3.5'
--- a/docs/requirements_for_rtd.txt
+++ b/docs/requirements_for_rtd.txt
@@ -57,7 +57,7 @@ python-memcached==1.59
 python-mimeparse==1.6.0
 pytz==2018.9
 qe-tools==1.1.2
-reentry==1.2.2
+reentry==1.3.0a1
 seekpath==1.8.4
 simplejson==3.16.0
 singledispatch>=3.4.0.3; python_version<'3.5'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -234,7 +234,8 @@
     language: system
     files: >-
       (?x)^(
-        setup_requirements.py|
+        setup.json|
+        setup.py|
         utils/validate_consistency.py|
         environment.yml|
       )$
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,9 +50,12 @@ before_install:
 
 install:
     # Upgrade pip setuptools and wheel to be able to run the next command
-    - pip install -U pip==18.1 wheel setuptools coveralls
+    - pip install --upgrade pip wheel setuptools coveralls
     # Install AiiDA with some optional dependencies
     - if [ "$TEST_TYPE" == "docs" ]; then pip install . && pip install -r docs/requirements_for_rtd.txt; else pip install --no-cache-dir .[all]; fi
+    - reentry map
+    - reentry scan
+    - reentry map
 
 env:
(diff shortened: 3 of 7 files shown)


ltalirz:
On the guards for the pip version in aiida-core's `setup.py`:
As far as I understand, the only reason for these guards is that pip < 10.0 will disregard the `pyproject.toml` and therefore the `reentry_register` hook won't be triggered.

This is solved by doing a `reentry scan` right after installing aiida, which one could consider the less intrusive fix.

Still, I'm fine with putting the guards for checking pip>=10.0 back into the setup.py if this is what people prefer.

ltalirz:
Let me know what you guys prefer - I'll fix this PR and tomorrow also release a reentry 1.3.0 - both on pypi and on conda.

sphuber:
Two questions:
1. With the current state then, entry points are automatically discovered upon install as long as `pip>=10.0.0` is used? So only users with a smaller version will have to run `reentry scan` explicitly. To me that seems reasonable. We can keep this in a section of the troubleshooting page on the documentation.
2. What is the difference between the `build-system` of the `pyproject.toml` and the `install_requires` in the `setup.json`? Why do we need `django` for example in the latter? Could one of the two be dropped?

ltalirz:
Re 1. Yes. I also tried installing aiida with pip 9 just to make sure that it works.

Re 2. The build system is what needs to be available *before* you execute the `setup.py`.
The setup.py contains the `reentry_register` hook, which will only trigger reentry if it is already installed when `setup.py` is executed.

Note: This means that also aiida *plugins* should, strictly speaking, include the same pyproject.toml (I will add it to the cookiecutter).
However, this will only matter if reentry is not yet present when the plugin is being installed.

It might be relevant in some cases, though - e.g. for plugins that are installed *together* with aiida-core (as part of the aiida metapackage).

sphuber:
> Note: This means that also aiida _plugins_ should, strictly speaking, include the same `pyproject.toml` (I will add it to the cookiecutter).
> However, this will only matter if reentry is not yet present when the plugin is being installed.

Or when one installs a plugin directly, without installing `aiida-core` first, which would be very common at some point I think. Since `aiida-core` should be a direct dependency for the plugins, people would for example only install `aiida-quantumespresso` which would then install `aiida-core` as one of its dependency. Are you saying that if `aiida-quantumespresso` does not have the toml, the entry points will not be automatically registered? Because that would be problematic

ltalirz:
> Are you saying that if aiida-quantumespresso does not have the toml, the entry points will not be automatically registered? 

I believe that is correct - without having the `pyproject.toml` in `aiida-quantumespresso` you would need to do `reentry scan` after the install.

ltalirz:
Anyhow, it has been my experience in the past that `reentry scan` was also needed in a number of corner cases even before the introduction of build system isolation.
I think it's simply good practice to keep it in the documentation (fine to remove it from travis etc. where we are testing how things *should* be).

ltalirz:
The PR should be ok for review - I've just opened an [issue](https://github.com/DropD/reentry/issues/49) on what to recommend to plugin developers.

In principle, we could also be using the `setup_requires` key, which would make life easier (no extra file, no need to add to MANIFEST.in) but comes with some slight downsides + is potentially less future proof.

docs/requirements_for_rtd.txt
```diff
@@ -57,7 +57,7 @@ python-memcached==1.59
 python-mimeparse==1.6.0
 pytz==2018.9
 qe-tools==1.1.2
-reentry==1.2.2
+reentry>=1.3.0
```
sphuber: I am approving even though this might gives us problems at some point, given that we are breaking compatibility quite a bit in `reentry` minor versions. If we do, we will have to quickly release a patch of `aiida-core`

sphuber:
Thanks @ltalirz

PR #2743, opened 2019-04-12
Reinstate the test submit functionality for `CalcJob`

sphuber:
The old `JobCalculation` had a method `submit_test` which would perform
a test submission, by mocking the remote working directory that would be
created in a real submission in a local sandbox folder. This method was
removed because the replacement class `CalcJobNode` no longer provided
methods to actually run anything. Instead this is now the responsability
of the `CalcJob` process class.

To provide similar functionality a new metadata input `dry_run` is
introduced for the `CalcJob` class. When passed in the inputs when the
process is ran, the same procedure as a normal submission is started but
the `upload_calculation` step will the local transport instead of the
targeted remote computer and will write the files to a local sandbox
folder instead of a remote working directory. After the upload task is
completed, the process is stopped instead of going to the submission
step.

This new approach has the advantage over the old implementation that the
pathway that is followed in the test submission is alsmost identical to
the code that would be executed in an actual submission. Additionally,
by using a metadata option, instead of a special launcher function or a
specific method on the `ProcessBuilder` ensures that the user does not
have to import and call a separate function, nor does the method on the
builder have to block a name in the input namespace.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -34,7 +34,7 @@ REMOTE_WORK_DIRECTORY_LOST_FOUND = 'lost+found'
 execlogger = AIIDA_LOGGER.getChild('execmanager')
 
 
-def upload_calculation(node, transport, calc_info, script_filename):
+def upload_calculation(node, transport, calc_info, script_filename, dry_run=False):
     """Upload a `CalcJob` instance
 
     :param node: the `CalcJobNode`.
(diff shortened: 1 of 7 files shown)


sphuber:
@giovannipizzi I think this solution is really elegant as it does not require a separate builder with a method spoiling the namespace, nor a separate launcher. The user only has to specify `metadata.dry_run = False`. It can even run while storing the provenance, but this can also be disabled. I don't think we should set provenance to False by default when `dry_run == True` but we should document it. Which I will do in other PR for one of the many open documentation issues

aiida/engine/daemon/execmanager.py
```diff
@@ -55,77 +55,78 @@ def upload_calculation(node, transport, calc_info, script_filename):
     transport.set_logger_extra(logger_extra)
     logger = LoggerAdapter(logger=execlogger, extra=logger_extra)
 
-    if node.has_cached_links():
-        raise ValueError("Cannot submit calculation {} because it has cached input links! If you just want to test the "
-                         "submission, use the test_submit() method, otherwise store all links first".format(node.pk))
+    if not dry_run and node.has_cached_links():
+        raise ValueError('Cannot submit calculation {} because it has cached input links! If you just want to test the '
+                         'submission, set `metadata.dry_drun` to True in the inputs.'.format(node.pk))
```
giovannipizzi: typo here (`dry_drun`)

PR #2744, opened 2019-04-12
Capture `KeyboardInterrupt` in `Runner.run` and kill processes

### LINKED ISSUE #2711: Try and kill a `Process` when running in local interpreter gets interrupted
Currently, if a user runs a `Process` in a local interpreter and then interrupts the interpreter the process will be lost. However, the node will reflect the last active state. This can be confusing to new users, who will still for example see a "process" in the "Waiting" state. We should try and catch the interrupt signal and try to kill the process properly before shutting down. At the very least we should try to set an `Excepted` state on the node.

sphuber:
Addresses #2711 partially but does not fix it 

Currently, if a user runs a `Process` in a local interpreter and then
interrupts the interpreter the process will be lost. However, the node
will reflect the last active state. This can be confusing to new users,
who will still for example see a "process" in the `Waiting` state.

To prevent this situation, the `Runner._run` method is modified and
attaches listeners to the `SIGINT` and `SIGTERM` signals that when
emitted will trigger a clean function that will log the interrupt and
call `process.kill` on the process that is currently in the scope.
This should ensure that if a user runs a process in a local interpreter
and interrupts it, the process will be properly killed, which will cause
the node to properly reflect what has happened.

The current solution works for a single process being run, but has
issues for nested processes. Sometimes only the current process is
killed and the others in the chain are unaffected before the interpreter
shuts down.

### DIFF
--- a/aiida/common/escaping.py
+++ b/aiida/common/escaping.py
@@ -38,6 +38,9 @@ def escape_for_bash(str_to_escape):
     within triple quotes to make it work, getting finally: the complicated
     string found below.
     """
+    if str_to_escape is None:
+        return ''
+
     escaped_quotes = str_to_escape.replace("'", """'"'"'""")
     return "'{}'".format(escaped_quotes)
 
--- a/aiida/engine/processes/functions.py
+++ b/aiida/engine/processes/functions.py
@@ -15,6 +15,8 @@ from __future__ import absolute_import
 import collections
 import functools
 import inspect
+import logging
+import signal
 
 from six.moves import zip  # pylint: disable=unused-import
 
@@ -24,6 +26,8 @@ from aiida.manage.manager import get_manager
 from .process import Process
 
 __all__ = ('calcfunction', 'workfunction', 'FunctionProcess')
+
+LOGGER = logging.getLogger(__name__)
 
 
 def calcfunction(function):
@@ -122,6 +126,15 @@ def process_function(node_class):
                 raise ValueError('{} does not support these kwargs: {}'.format(function.__name__, kwargs.keys()))
 
             process = process_class(inputs=inputs, runner=runner)
+
+            def kill_process(_num, _frame):
+                """Send the kill signal to the process in the current scope."""
+                LOGGER.critical('runner received interrupt, killing process %s', process.pid)
+                process.kill(msg='Process was killed because the runner received an interrupt')
+
+            signal.signal(signal.SIGINT, kill_process)
+            signal.signal(signal.SIGTERM, kill_process)
+
             result = process.execute()
 
             # Close the runner properly
--- a/aiida/engine/processes/process.py
+++ b/aiida/engine/processes/process.py
@@ -22,6 +22,7 @@ import traceback
 import six
 from six.moves import filter, range
 from pika.exceptions import ConnectionClosed
+from kiwipy.communications import UnroutableError
 
 import plumpy
 from plumpy import ProcessState
@@ -239,6 +240,8 @@ class Process(plumpy.Process):
                         killing.append(result)
                 except ConnectionClosed:
                     self.logger.info('no connection available to kill child<%s>', child.pk)
+                except UnroutableError:
+                    self.logger.info('kill signal was unable to reach child<%s>', child.pk)
 
             if isinstance(result, plumpy.Future):
                 # We ourselves are waiting to be killed so add it to the list
--- a/aiida/engine/runners.py
+++ b/aiida/engine/runners.py
@@ -13,8 +13,9 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
 
-from collections import namedtuple
+import collections
 import logging
+import signal
 import tornado.ioloop
 
 import plumpy
@@ -31,8 +32,8 @@ __all__ = ('Runner',)
 
 LOGGER = logging.getLogger(__name__)
 
-ResultAndNode = namedtuple('ResultAndNode', ['result', 'node'])
-ResultAndPk = namedtuple('ResultAndPk', ['result', 'pk'])
+ResultAndNode = collections.namedtuple('ResultAndNode', ['result', 'node'])
+ResultAndPk = collections.namedtuple('ResultAndPk', ['result', 'pk'])
 
 
 class Runner(object):  # pylint: disable=useless-object-inheritance
@@ -198,6 +199,15 @@ class Runner(object):  # pylint: disable=useless-object-inheritance
 
         with utils.loop_scope(self.loop):
             process = instantiate_process(self, process, *args, **inputs)
+
+            def kill_process(_num, _frame):
+                """Send the kill signal to the process in the current scope."""
+                LOGGER.critical('runner received interrupt, killing process %s', process.pid)
+                process.kill(msg='Process was killed because the runner received an interrupt')
+
+            signal.signal(signal.SIGINT, kill_process)
+            signal.signal(signal.SIGTERM, kill_process)
+
             process.execute()
             return process.outputs, process.node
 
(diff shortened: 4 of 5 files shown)


sphuber:
@muhrin : would welcome your input if `plumpy._process_stack` is the advised way of getting the running process and if the location of the signal handlers is fine. I tested this locally for calc jobs and it works pretty well. Don't really know how to setup up decent unit tests for this. Problem is that with workchains it does not seem to work. The workchain might get killed, but I am getting all sort off other exceptions elsewhere, I think when the callback of the child gets excepted. Its different everytime and can even go down to `postgres` :sweat: 
```
04/12/2019 11:48:50 AM <4723> tornado.application: [ERROR] Exception in callback <functools.partial object at 0x7fe4aa169940>
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/tornado/ioloop.py", line 605, in _run_callback
    ret = callback()
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/tornado/stack_context.py", line 277, in null_wrapper
    return fn(*args, **kwargs)
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/tornado/ioloop.py", line 626, in _discard_future_result
    future.result()
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/tornado/gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/processes.py", line 1109, in step_until_terminated
    yield self.step()
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/tornado/gen.py", line 1069, in run
    yielded = self.gen.send(value)
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/processes.py", line 1100, in step
    self.transition_to(next_state)
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/base/state_machine.py", line 326, in transition_to
    self.transition_failed(initial_state_label, label, *sys.exc_info()[1:])
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/base/state_machine.py", line 339, in transition_failed
    six.reraise(type(exception), exception, trace)
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/base/state_machine.py", line 310, in transition_to
    self._enter_next_state(new_state)
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/base/state_machine.py", line 374, in _enter_next_state
    self._fire_state_event(StateEventHook.ENTERED_STATE, last_state)
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/base/state_machine.py", line 288, in _fire_state_event
    callback(self, hook, state)
  File "/home/sphuber/code/aiida/env/dev/plumpy/plumpy/processes.py", line 308, in <lambda>
    lambda _s, _h, from_state: self.on_entered(from_state))
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/engine/processes/process.py", line 285, in on_entered
    self.update_node_state(self._state)
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/engine/processes/process.py", line 495, in update_node_state
    self.node.set_process_state(state.LABEL)
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/nodes/process/process.py", line 181, in set_process_state
    return self.set_attribute(self.PROCESS_STATE_KEY, state)
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/utils/mixins.py", line 207, in set_attribute
    super(Sealable, self).set_attribute(key, value, stored_check=False, **kwargs)
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/nodes/node.py", line 395, in set_attribute
    self.backend_entity.set_attribute(key, value)
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/implementation/django/nodes.py", line 178, in set_attribute
    self.ATTRIBUTE_CLASS.set_value_for_node(self.dbmodel, key, value)
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/backends/djsite/db/models.py", line 1167, in set_value_for_node
    stop_if_existing=stop_if_existing)
  File "/home/sphuber/code/aiida/env/dev/aiida-core/aiida/backends/djsite/db/models.py", line 663, in set_value
    transaction.savepoint_rollback(sid)
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/django/db/transaction.py", line 66, in savepoint_rollback
    get_connection(using).savepoint_rollback(sid)
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/django/db/backends/base/base.py", line 348, in savepoint_rollback
    self._savepoint_rollback(sid)
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/django/db/backends/base/base.py", line 308, in _savepoint_rollback
    cursor.execute(self.ops.savepoint_rollback_sql(sid))
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/django/db/backends/base/operations.py", line 361, in savepoint_rollback_sql
    return "ROLLBACK TO SAVEPOINT %s" % self.quote_name(sid)
  File "/home/sphuber/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/django/db/backends/postgresql/operations.py", line 111, in quote_name
    if name.startswith('"') and name.endswith('"'):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

giovannipizzi:
@sphuber there are three `except:` in the code, two are in the part related to rolling back (even if in a different place than the traceback). But what could happen for instance is that you send the signal exactly in that part of the code, then `except` is triggered, you roll back, and then you see the problem elsewhere? Just guessing, though. If you replace those, temporarily, with `except Exception:` instead do you still see the problem? (BTW the try/except in TrajectoryData can be removed, and for the other two maybe we should rethink better how to deal with those rollbacks ([link to the Django docs for reference](https://docs.djangoproject.com/en/1.11/topics/db/transactions/#savepoint-rollback)).

giovannipizzi:
to test it, you could try something like `os.kill(os.getpid(), signal.SIGINT)` even if you should run a different process with `subprocess` and `verdi run`, otherwise you kill the test interpreter?

sphuber:
Regarding the test suggestion: the difficult thing is not the killing but to launch the sub process. It should be a sub process that runs an aiida environment in which we run a process in a local runner. This can only be done by calling `run` on a process with some inputs. So the only way I can envisage this happening is to write a dummy `runaiida` script that runs a long running dummy calcjob and then send the signal. But as you can see it is not very clean

sphuber:
@giovannipizzi I have a solution that works partially. It is not tested through unit tests and doesn't work in all cases, but I think it will be important to have this partial non-ideal solution in `v1.0.0b3` of tomorrow. I am sure that people during the tutorial will CTRL+C after running a process and will end up with a zombie process. It would be good that it will show as killed. When launching nested processes not everything will get properly killed, but at least it is something.

sphuber:
@giovannipizzi could you please reapprove, I have rebased

PR #2747, opened 2019-04-12
Raise when returning an unstored `Data` from a `WorkflowNode`

### LINKED ISSUE #2675: Warn users when they are creating new `Data` in a `WorkChain`
Although this is (or will be) well documented, I predict that many novice users will try creating new `Data` inside Workflows - it's simply something very natural to do: you run some calculations,  you combine some of their outputs and return them in a new `Dict`.

I'm not saying we should allow this (although at some point one could perhaps think of a mechanism to automatically wrap this in a `@calcfunction`), but I think it is very important that we **warn** users that they are doing something wrong.
Otherwise, we end up with a situation like here in the lsmo, where users have run their `WorkChains` and, at the end when they want to export them, they look at the graph and notice that links are missing etc.

@sphuber Is there a way for us to detect that `Data` is created from within a Workflow?
If we can detect this, I would actually even be in favor of (by default) throwing an exception rather than printing just a warning, since it really is something we don't want people to do.

sphuber:
Fixes #2675 

This will happen if one tries to return a `Data` node from within the
body of a `WorkChain` or `workfunction`, which means that they probably
created a new node based on its inputs or data returned by one of the
processes it calls. However, this is strictly forbidden as the
provenance of the new node will be lost. Given that beginning users are
likely to make this mistake, instead of issuing a warning we explicitly
forbid this behavior and raise in the link validation.

### DIFF
--- a/aiida/orm/nodes/process/workflow/workflow.py
+++ b/aiida/orm/nodes/process/workflow/workflow.py
@@ -23,8 +23,6 @@ __all__ = ('WorkflowNode',)
 class WorkflowNode(ProcessNode):
     """Base class for all nodes representing the execution of a workflow process."""
 
-    # pylint: disable=too-few-public-methods
-
     # Workflow nodes are storable
     _storable = True
     _unstorable_message = 'storing for this node has been disabled'
@@ -52,3 +50,23 @@ class WorkflowNode(ProcessNode):
         :return: `NodeLinksManager`
         """
         return NodeLinksManager(node=self, link_type=LinkType.RETURN, incoming=False)
+
+    def validate_outgoing(self, target, link_type, link_label):
+        """Validate adding a link of the given type from ourself to a given node.
+
+        A workflow cannot 'create' Data, so if we receive an outgoing link to an unstored Data node, that means
+        the user created a Data node within our function body and tries to attach it as an output. This is strictly
+        forbidden and can cause provenance to be lost.
+
+        :param target: the node to which the link is going
+        :param link_type: the link type
+        :param link_label: the link label
+        :raise TypeError: if `target` is not a Node instance or `link_type` is not a `LinkType` enum
+        :raise ValueError: if the proposed link is invalid
+        """
+        super(WorkflowNode, self).validate_outgoing(target, link_type, link_label)
+        if link_type is LinkType.RETURN and not target.is_stored:
+            raise ValueError(
+                'Workflow<{}> tried returning an unstored `Data` node. This likely means new `Data` is being created '
+                'inside the workflow. In order to preserve data provenance, use a `calcfunction` to create this node '
+                'and return its output from the workflow'.format(self.process_label))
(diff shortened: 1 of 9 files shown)


aiida/orm/nodes/process/workflow/workflow.py
```diff
@@ -52,3 +50,23 @@ def outputs(self):
...
+        :param target: the node to which the link is going
+        :param link_type: the link type
+        :param link_label: the link label
+        :raise TypeError: if `target` is not a Node instance or `link_type` is not a `LinkType` enum
+        :raise ValueError: if the proposed link is invalid
+        """
+        super(WorkflowNode, self).validate_outgoing(target, link_type, link_label)
+        if link_type is LinkType.RETURN and not target.is_stored:
+            raise ValueError(
+                'trying to return an unstored Data node from a workflow is not allowed because it will cause the '
```
ltalirz: this is already good. What about something like:
"Workflow '{}' tried returning an unstored Data node. This likely means new Data is being created inside the workflow. In order to preserve data provenance, use a `calcfunction` to create this data, and return its output from the workflow.".format(...)

sphuber:
My thirst through was to override the `out` method for the workchain, but then that would not catch workfunctions. There I would have to do it for the `FunctionProcess` but since that is shared with `calcfunction` I would have to make a switch on the node type, which is ugly. Since it is then ultimately the node type that decides anyway, might as well put there in one place. This has the added benefit that this also prevent sneaky pete's from calling `my_illegally_created_node.add_incoming(self, link_type=LinkType.RETURN)` in the work chain :wink: which would not be caught by the check in the overridden `out`

aiida/orm/nodes/process/workflow/workflow.py
sphuber: done

aiida/orm/nodes/process/workflow/workflow.py
```diff
@@ -52,3 +50,23 @@ def outputs(self):
...
+        :param target: the node to which the link is going
+        :param link_type: the link type
+        :param link_label: the link label
+        :raise TypeError: if `target` is not a Node instance or `link_type` is not a `LinkType` enum
+        :raise ValueError: if the proposed link is invalid
+        """
+        super(WorkflowNode, self).validate_outgoing(target, link_type, link_label)
+        if link_type is LinkType.RETURN and not target.is_stored:
+            raise ValueError(
+                'Workflow<{}> tried returning an unstored `Data` node. This likely means new `Data` is being created '
```
ltalirz: I was thinking more of the name of the workflow (which should tell people where in the code they need to change things).
Happy to re-approve if you still want to change this.

sphuber: I adapted it to print the process label and also added two more tests on workchain and workfunction level, to make sure it also works through there

PR #2757, opened 2019-04-14
Print formatted spec for `verdi plugin list <entry_point>`

### LINKED ISSUE #2645: Bug in `verdi plugin list`
Calling the command with a specific entry point for `aiida.calculations` raises the following:
```
(aiida_dev) sph@bastion:~/code/aiida/env/dev/aiida-core$ verdi plugin list aiida.calculations arithmetic.add
Info: arithmetic.add
  File "/home/sph/.virtualenvs/aiida_dev/bin/verdi", line 10, in <module>
    sys.exit(verdi())
  File "/home/sph/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/click/core.py", line 764, in __call__
  File "/home/sph/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/sph/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
  File "/home/sph/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
  File "/home/sph/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/click/core.py", line 956, in invoke
  File "/home/sph/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/click/core.py", line 555, in invoke
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/cmdline/utils/decorators.py", line 75, in decorated_function
    return function(*args, **kwargs)
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/cmdline/commands/cmd_plugin.py", line 52, in plugin_list
    echo.echo(plugin.get_description())
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/cmdline/utils/echo.py", line 42, in echo
    click.secho(message, bold=bold, nl=nl, err=err)
  File "/home/sph/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/click/termui.py", line 477, in secho
    message = style(message, **styles)
  File "/home/sph/.virtualenvs/aiida_dev/local/lib/python2.7/site-packages/click/termui.py", line 449, in style
    return ''.join(bits)
TypeError: sequence item 1: expected string, dict found
```

sphuber:
Fixes #2645 

When a valid entry point is passed an the corresponding class is a sub
class of `Process`, the `get_description` message will return the
process spec which is then passed to a function that will print it in a
human-readable formatted way to screen. It will give an overview of the
inputs, outputs and exit codes defined on the spec.

### DIFF
--- a/aiida/cmdline/commands/cmd_plugin.py
+++ b/aiida/cmdline/commands/cmd_plugin.py
@@ -11,6 +11,7 @@
 from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
+
 import click
 
 from aiida.cmdline.commands.cmd_verdi import verdi
@@ -30,6 +31,8 @@ def verdi_plugin():
 def plugin_list(entry_point_group, entry_point):
     """Display a list of all available plugins."""
     from aiida.common import EntryPointError
+    from aiida.cmdline.utils.common import print_process_spec
+    from aiida.engine import Process
     from aiida.plugins.entry_point import get_entry_point_names, load_entry_point
 
     if entry_point_group is None:
@@ -45,13 +48,15 @@ def plugin_list(entry_point_group, entry_point):
         try:
             plugin = load_entry_point(entry_point_group, entry_point)
         except EntryPointError as exception:
-            echo.echo_critical(exception)
+            echo.echo_critical(str(exception))
         else:
-            echo.echo_info(entry_point)
             try:
-                echo.echo(plugin.get_description())
-            except AttributeError:
-                echo.echo('No description available')
+                if issubclass(plugin, Process):
+                    print_process_spec(plugin.spec())
+                else:
+                    echo.echo(str(plugin.get_description()))
+            except (AttributeError, TypeError):
+                echo.echo_error('No description available for {}'.format(entry_point))
     else:
         entry_points = get_entry_point_names(entry_point_group)
         if entry_points:
--- a/aiida/cmdline/utils/common.py
+++ b/aiida/cmdline/utils/common.py
@@ -16,6 +16,7 @@ from __future__ import absolute_import
 import os
 import sys
 
+import click
 from tabulate import tabulate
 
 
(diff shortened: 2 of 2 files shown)


sphuber:
Example output for `verdi plugin list aiida.calculations arithmetic.add`:
```
Inputs
           code:  required  Code        The Code to use for this job.
              x:  required  Int, Float  The left operand.
              y:  required  Int, Float  The right operand.
       metadata:  optional              
Outputs
  remote_folder:  required  RemoteData  Input files necessary to run the process will be stored in this folder node ...
      retrieved:  required  FolderData  Files that are retrieved by the daemon will be stored in this node. By defa ...
            sum:  required  Int, Float  The sum of the left and right operand.
Exit codes
             11:  The process did not register a required output
            100:  The retrieved folder data node could not be accessed.
            110:  The output file could not be read from the retrieved folder.
            120:  The output file contains invalid output.
```

giovannipizzi:
I like the output, but I'm not sure this should be in `get_description` - at the moment we are using it for two purposes (short string for the Materials Cloud, and short description for the `verdi graph generate`, and I think this should stay "short" (that fits in 1 or few lines, inside a node of Graphviz or in a web tooltip).

Maybe we should document this better, and decide if we want to have two different methods - one for a one-liner, and one for an extensive text description like the one above

sphuber:
Note that we are talking about `get_description` of the `Process` class, not a `Node` class, which I think is what you were thinking off. In addition, the example output is still not that returned by `get_description`, which just return the `process spec (as that is in a sense the best and most complete description of a process class), but is the output generated by `aiida.cmdline.utils.common.print_process_spec`.

giovannipizzi:
I see - then I'm approving this, even if probably we should make sure we and others don't mix the two.

ltalirz:
I think there might be some value in having `get_description` work consistently across different AiiDA classes (meaning: returning a 1-line description).

>  the example output is still not that returned by get_description, which just return the process spec

Do you think this is the right thing to do?
I would expect a description to be directly printable.

ltalirz:
by the way, is there a potential for code sharing with the aiida sphinx extension for documenting workchains?
https://github.com/aiidateam/aiida_core/blob/develop/aiida/sphinxext/workchain.py

sphuber:
> Do you think this is the right thing to do?
> I would expect a description to be directly printable.

It *is* directly printable, because the `ProcessSpec` implements the `__str__` method that will recursively call the `__str__` methods of its relevant attributes resulting in a print of a nested dictionary, however, this is not very readable. For it to be really nice and readable you need some fancy formatting, which is what I implemented here.

sphuber:
@ltalirz I have removed `get_description` from the `Process` class for now and just call the spec directly. This way we can always add it in the future to have similar behavior to the other `get_description` methods.

sphuber:
And regarding the code of the sphinx extension: in principle they have the same goal, formatting the spec, but I am not sure if it is easy to write it in such a way that it can be reused easily in both places. Probably interesting cleanup for the future though

PR #2758, opened 2019-04-14
Add test to make sure `Process.out` accepts dictionaries

### LINKED ISSUE #1571: Allow workchain namespace outputs using dictionary
A suggestion from Matthieu Mottet that it would be nice to be able to write:

    d = {'a': Int(5), 'b': Int(5)}
    self.out('namespace', d)
instead of having do it it one by one:

    for key, val in d.items():
        self.out('namespace.key', node)`

sphuber:
Fixes #1571 

The recent commit `e5abcd9b114efe` already ensured that output port
namespaces are properly nested and so it took care of the issue that
`Process.out` could not accept a dictionary to be automatically mapped
onto a nested namespace. Here we add an explicit tests for this.

### DIFF
--- a/aiida/backends/tests/engine/test_process.py
+++ b/aiida/backends/tests/engine/test_process.py
@@ -217,6 +217,28 @@ class TestProcess(AiidaTestCase):
         recovered_process = process.node.process_class
         self.assertEqual(recovered_process, process.__class__)
 
+    def test_output_dictionary(self):
+        """Verify that a dictionary can be passed as an output for a namespace."""
+
+        class TestProcess(Process):
+
+            _node_class = orm.WorkflowNode
+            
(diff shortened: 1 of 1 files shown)


giovannipizzi:
Note that the commit ID in the PR message is probably wrong?

sphuber:
@giovannipizzi no it is a partial commit, but it is correct

giovannipizzi:
ok! but why then github does not link to it? maybe something you rebased?

sphuber:
No it is there: https://github.com/aiidateam/aiida_core/commit/e5abcd9b114efe
It doesn't format it because it is maybe a partial commit hash or because I put it in between backticks and it doesn't interpret it as a commit

Issue #2763, opened 2019-04-16
Clean the logic in `verdi quicksetup` implementation


sphuber:
The logic in `aiida.cmdline.commands.setup.quicksetup` tries through a bunch of prompts to fill in the blanks before passing all information to `verdi setup`. The code, however, is still a bit messy and does not use our `InteractiveOption` for the prompting for example, leading to different styles of prompts.

ltalirz:
Just to list a few other changes that were made in #2740 : 
 * `verdi setup` profile name argument became an option
 * `verdi quicksetup` profile name argument became an option

We should briefly discuss whether we want these changes and what the desired behavior of `verdi setup` and `verdi quicksetup` is. I list my suggestions below:

### `verdi quicksetup`

1.1 Running `verdi quicksetup` interactively:
 * general idea: prompt as little as possible
 * I believe it nevertheless makes sense to always prompt for the profile name. Otherwise many aiida users never get introduced to the concept of profiles. This would speak for @sphuber 's change where the profile name becomes an option (note: I believe this used to be the case in earlier aiida versions as well)
 * We need to prompt for the user info. Pre-filling it helps a lot but we can't entirely hide it in this case (otherwise, you risk creating all your data with inappropriate user information).
 * if you provide the name of an *existing* profile, I suggest quicksetup should fail immediately.
   Don't use quicksetup to re-configure an existing profile.

1.2. Running `verdi quicksetup` non-interactively:
 * it would make sense to be able to specify the options that are available in verdi setup, since quicksetup also creates the database for you.
 * if you provide the name of an *existing* profile, it should fail.

### `verdi setup`

2.1. Running `verdi setup` interactively:
 * general idea: prompt for everything but provide sensible defaults
 * we could introduce a `--force` option that allows verdi setup to reconfigure an *existing* profile, (a use case mentioned by @sphuber ). If the `--force` option is not provided, `verdi setup` should fail immediately.

2.2. Running `verdi setup` non-interactively:
 * Requiring the `--force` option makes it safe enough to allow re-configuring non-interactively as well. We would need to check whether the `contextual_default` does the right thing in this case, though

mentioning @giovannipizzi for comment

ltalirz:
Summarizing the discussion with @sphuber and @giovannipizzi: 

 * We change the profile name to being an *option*, not an argument. 
   Note: This was already the case for `quicksetup` in 0.12.x (but not for `setup`).
[...]

Issue #2759, opened 2019-04-15
Allow the switching of profile and corresponding backend within interpreter


sphuber:
The mechanism of loading a profile was changed significantly in PR #2740 which make it possible in principle to dynamically load another profile and its corresponding database backend while staying in the same interpreter. For the time being this behavior is blocked and changing the backend in a non-test profile will raise. After tests are introduced to make sure this can be done safely, this limitation should be lifted.

sphuber:
Adding this additional comments from #2813 that really just describes a sub problem of the issue described here.

Copied content below
-----

If we want profiles to be switchable in a single interpreter, the database environment needs to be switchable as well. However, currently for Django, once the `aiida.backends.djsite.settings` file is imported, the variables for example the database connection, which are based on the currently loaded profile, will be locked in. So even if the profile would be swapped for another django profile, with a different database, the original database will be used anyway.


By @giovannipizzi 

A comment to be taken into account: the purpose of `recreate_after_fork()` is to be called as a callback when the process is forked (e.g. by a demoniser - this was the case with celery), here:
```
https://github.com/aiidateam/aiida-core/blob/0644e7d9ff492fdb2c94d608c575c697daee6d32/aiida/backends/sqlalchemy/__init__.py#L78
```
By the way, I realise now that probably:
1. We should dispose of the passed engine, and not the global one
2. When switching backend, probably we should deregister the callback if it was registered

chrisjsewell:
Fixed in #5330

PR #2766, opened 2019-04-16
Rework creation of configuration and profile and loading database backends

### LINKED ISSUE #899: verdi setup detects existing user but asks to reenter information regardless
When I run `verdi setup <PROFILE>` that was previously setup, for the general parameters the command correctly detects the preexisting settings and asks if you want to change or keep them. However, directly followed by this is the prompt to enter the information for the user, like the name and institution, even though the user already exists. Should this not similarly prompt with the question if one wants to keep the existing data or change it?

### LINKED ISSUE #1416: clever quicksetup
Asking a user *once* to enter his email, name, last name and institution is fine.
Asking a user ten times is... suboptimal.

Let `verdi quicksetup` autofill these fields with those of the `default` profile, if one exists (still editable, of course).

### LINKED ISSUE #1483: Non-interactive and interactive versions of verdi setup do not use same code
Especially the code that validates the user input is not shared by these two pathways. The validation should be abstracted and used by both versions of the command

sphuber:
Fixes #1416 fixes #1483 fixes #899 

This is a rather complex PR that addresses a few entangled problems:
- Centralize the loading of configuration and profile to `aiida.manage.configuration`
- Make the loading of the database backend environment go through the `Manager`
- Allow reloading a different profile with a different database backend in the same interpreter as long as a database environment has not yet been loaded (see #2759 that will release this constraint)
- Clean up all the various "settings" files and move still relevant parts to `aiida.manage.configuration.settings`
- Refactor the `Config` and `Profile` class interfaces. They no longer store their literal `config.json` representation but have a different internal storage and return the config format when requested. This allows for a nicer API which will be more stable with respect to changes in how their content is actual written to the `config.json` file.
- Rework the `verdi setup` and `verdi quicksetup` commands to go completely through `click`. With the also added concept of a default "configuration user", which can be set through `verdi config`, in principle, user information will now only be have to typed in *once per AiiDA instance*.

Things that remain to be done, but potentially in other PRs
- [x] Unit tests were added for `verdi setup` and `verdi quicksetup` but are currently skipped because they will cause the tests to hang locally if the Postgres setup fails. This should be migrated to use the fixtures
**Decision:** See issue #2759 
- [x] Extensive testing for dynamic switching of profiles within single interpreter. Since we do not actively encourage this, I think it is fine to have the possibility for now.
**Decision:** it is allowed to load a different profile, but for the time being we do not allow switching the database backend. This means that a profile can be changed until the very first time its backend is loaded and at that time the profile becomes frozen. See issue #2759 
- [x] Further cleaning of the `verdi quicksetup` logic that tries to automatically determine or prompt for information that will be passed to `verdi setup`. The Postgres detection is still fragile, it fails on my Ubuntu 18.04 machine for example, and the prompt use plain `click` instead of our own, which looks a bit ugly.
**Decision:** See issue #2763 
- [x] I moved "configuration options", i.e. those that can be set per profile or globally, into the `options` dictionary. They used to live in the top level namespace for both the config and the profiles. This runs the risk of options at some point overlapping with "normal" keys, plus it makes the API more complicated. By moving them to a different location in the `config.json` and since the `Config` and `Profile` classes are no longer just a wrapper around the data structure of that file, the constructor currently of those classes simply ignores and removes them, emitting a warning if it does so. Since some older options had already been deprecated anyway and these used to be `verdi devel` commands, I think it is fine to not invest the amount of work it would require to write a proper migration for this, but I thought to mention it here.
**Decision:** commit has been added to in this case create a backup of the configuration file and warn the user.


szoupanos:
@sphuber @giovannipizzi 

Very likely this PR is the result of the following exception
```
(aiidapy2) aiida@ubuntu_vm1_new:~$ verdi -p dj1 shell
  File "/home/aiida/aiidapy2/bin/verdi", line 10, in <module>
    sys.exit(verdi())
  File "/home/aiida/aiidapy2/local/lib/python2.7/site-packages/click/core.py", line 764, in __call__
  File "/home/aiida/aiidapy2/local/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/aiida/aiidapy2/local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
  File "/home/aiida/aiidapy2/local/lib/python2.7/site-packages/click/core.py", line 956, in invoke
  File "/home/aiida/aiidapy2/local/lib/python2.7/site-packages/click/core.py", line 555, in invoke
  File "/home/aiida/aiida-code/aiida_core/aiida/cmdline/utils/decorators.py", line 68, in decorated_function
    load_backend_if_not_loaded()
  File "/home/aiida/aiida-code/aiida_core/aiida/cmdline/utils/decorators.py", line 44, in load_backend_if_not_loaded
    get_manager().get_backend()
  File "/home/aiida/aiida-code/aiida_core/aiida/manage/manager.py", line 104, in get_backend
    self._load_backend()
  File "/home/aiida/aiida-code/aiida_core/aiida/manage/manager.py", line 83, in _load_backend
    load_backend(profile)
  File "/home/aiida/aiida-code/aiida_core/aiida/backends/djsite/utils.py", line 30, in load_dbenv
    _load_dbenv_noschemacheck(profile)
  File "/home/aiida/aiida-code/aiida_core/aiida/backends/djsite/utils.py", line 45, in _load_dbenv_noschemacheck
    django.setup()
  File "/home/aiida/aiidapy2/local/lib/python2.7/site-packages/django/__init__.py", line 22, in setup
    configure_logging(settings.LOGGING_CONFIG, settings.LOGGING)
  File "/home/aiida/aiidapy2/local/lib/python2.7/site-packages/django/conf/__init__.py", line 56, in __getattr__
    self._setup(name)
  File "/home/aiida/aiidapy2/local/lib/python2.7/site-packages/django/conf/__init__.py", line 41, in _setup
    self._wrapped = Settings(settings_module)
  File "/home/aiida/aiidapy2/local/lib/python2.7/site-packages/django/conf/__init__.py", line 129, in __init__
    raise ImproperlyConfigured("The SECRET_KEY setting must not be empty.")
django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty. 
```

This appeared after checking out the latest develop state (8b6169c9f)
If it is the case (and whatever the case is), there should have been a warning or a migration of the config files (I suppose that the problem is there since all the tests pass on travis).

szoupanos:
Sorry for the (unnecessary) noise.
The above exception resulted from not deleted .pyc files.

Something more on this PR or related commits: After getting the latest updates, I don't see a way to recreate a database that corresponds to a specific profile after deleting it and creating it.

This was possible until recently with verdi setup <the_profile_that_corresponds_to_the_db>

sphuber:
@szoupanos there is `verdi database migrate`

szoupanos:
Great! Thanks!

szoupanos:
But there is no default user - OK we discuss it in person and if needed we create an issue etc
Sorry for the spamming

sphuber:
For that you can use `verdi user` commands

Issue #2764, opened 2019-04-16
REST API should not be based on Hybrid properties


szoupanos:
Currently REST API is based on hybrid properties like the user_email defined in the DbNode table in SQLA.

This allows us to create queries where we may project node properties and also the user_email and implicitly to do the join of the DbNode with the DbUser to provide the user email.

This is a nice feature but it is a feature that it is defined at the SQLA model (not in the Django model) and we should not make assumptions that such feature exist at the REST API which is a few levels above the backends.

In general the REST API queries should change from 

```
qh = {'project': {
    'calculations': ['id', 'label', 'node_type', 'ctime', 'mtime', 'uuid', 'user_id', 'user_email',
                     'attributes', 'extras']},
      'path': [{'tag': 'calculations', 'entity_type': 'process.calculation.calcjob.CalcJobNode.'}],
      'order_by': {'calculations': {}}, 'filters': {}}
```

to

```
qh = {
    'project':
        {
            'calc': ['id', 'label', 'node_type', 'ctime', 'mtime', 'uuid', 'user_id', 'attributes', 'extras'],
            'user_1': ['email']
        },
    'path':
        [
            {'entity_type': 'process.calculation.calcjob.CalcJobNode.', 'tag': 'calc'},
            {'edge_tag': 'calc--user_1', 'joining_keyword': 'with_node', 'entity_type': 'user',
             'joining_value': 'calc', 'tag': 'user_1'}
    ],
    'order_by': {'calc': {}}
```

This issue became visible after the removal of the dummy_model (for the querying of Django backrend). Work which is done to support JSONB in Djnago

PR #2767, opened 2019-04-16
Remove hybrid properties from SqlaAlchemy and dummy models

### LINKED ISSUE #2765: SQLA model hybrid properties should be removed
We defined hynrid properties in the SQLA models for convenience.
E.g. We can project user_email when projecting DbNode properties, and this will implicitly do the join of DbNode with DbUser. For  more info, please look at #2764.

These are backend depended properties, introduced for convenience, which for various reasons started to be used at layers above the backend like REST API (where such choices should have been transparent).

In this issue, we remove the hybrid properties and we change the code in aiida-core code to create the proper queries.

Such assumptions for layers above the aiida-core/aiida-orm like the REST-API will be handled at #2764

szoupanos:
Fixes #2765 

Soon the current hard-coded dummy model, to map the Django models on
SqlAlchemy models, which drives the `QueryBuilder` implementation, will
be replaced with dynamically generated Aldjemy models. However, these do
not support hybrid properties so they have to be removed.

### DIFF
--- a/aiida/backends/sqlalchemy/migrations/env.py
+++ b/aiida/backends/sqlalchemy/migrations/env.py
@@ -20,7 +20,8 @@ from aiida.backends.sqlalchemy.models.comment import DbComment
 from aiida.backends.sqlalchemy.models.computer import DbComputer
 from aiida.backends.sqlalchemy.models.group import DbGroup
 from aiida.backends.sqlalchemy.models.log import DbLog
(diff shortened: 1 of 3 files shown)


szoupanos:
After the #2764 is solved and this PR is merged to develop, then the aiida_core/django_jsonb branch should be rebased and the rest-api tests should be reactivated

PR #2768, opened 2019-04-16
Programatically access CalcJob dry_run folder

chrisjsewell:
Hey @sphuber, just going through your recent changes, looks good :)

#2743 makes sense, however, previously with the `submit_test` approach,
I could use it within my test suite to check the correct input files were being created
(and ensure the folder was removed after testing), e.g.

```python
with SandboxFolder() as folder:
    subfolder, script_filename = builder.submit_test(folder=folder)
    with subfolder.open(process.metadata.options.input_file_name) as f:
            input_content = f.read()
    assert input_content == "..."
```

I didn't see an obvious way to achieve this with the new approach?

If not, the simple fix I propose here, is to store the path as an attribute on the `CalcJobNode`.


chrisjsewell:
Ok, so I realised my first approach only worked if `store_provenance=False`.
I think maybe a better is to drop the `SubmitTestFolder` approach and simply copy the contents of the sandbox folder into a folder in the `CalcJobNode` repository, at the end of `upload_calculation()` (see latest commit).
In this manner, you are not creating folders in spurious locations, and its simple to interrogate with the normal AiiDA mechanics.

sphuber:
@chrisjsewell I discussed your solution with @giovannipizzi and the reason that we use a folder in the current working directory is for a reason. A lot of people will use this for manual testing purposes and having the possibility to just inspect an actual folder in either a editor or the shell is a lot quicker than through the AiiDA API. That is why we are a bit reluctant to move the generated files to the repository. That being said, we see the usefulness of your use case as well. What about if we attach the created `Folder` in the dry run as an attribute to the node. This way you would be able to do:
```
results, node = launch.run.get_node(ProcessClass, a=Int(1), metadata={'dry_run': True})
node.dry_run_folder
>>  `aiida.common.folders.Folder object`
```
You have then full access to the folder API for your unittests.
The attribute name `dry_run_folder` remains to be decided.
Would that work?

chrisjsewell:
@sphuber yep that seems like a good compromise, I've moved your suggestion to #2884 . So you can close this PR if you want.

PS I see you have also added a global variable for the top level folder `aiida.common.folders import CALC_JOB_DRY_RUN_BASE_PATH`

sphuber:
Closing this for now. Discussion continues in #2884

PR #2769, opened 2019-04-16
Clean the implementation of `verdi quicksetup`

### LINKED ISSUE #2595: Bug in creation of `dbinfo` of `Postgres` class in `verdi quicksetup`
The `Postgres` class, used to connect to postgres, has an internal attribute `_dbinfo` which stores a dictionary of parameters that should be used to connect to the database. Defaults for the `port` and `hostname` are only set if the `dbinfo` is `None` during construction. However, in `verdi quicksetup` an actual `dbinfo` argument *is* passed, however, the `port` and `hostname` are set to `None` if they are not explicitly passed by the user, because the default values for these options are set to `None`. This leads to problems further on, because without a valid hostname and port, the connections will fail. Moreover, at the end of `quicksetup` the `port` and `hostname` are retrieved from the same `dbinfo` obtained from the `Postgres` instance, with the `get` construction, passing the defaults in the case they are not set. However, they *are* set, they are just set to `None`. As a result, `None` is set for these keys in the newly created profile and all subsequent connections will fail as a result.

### LINKED ISSUE #2763: Clean the logic in `verdi quicksetup` implementation

sphuber:
Fixes #2763 and fixes #2595 

The logic for automatically determining the required options if not
specified, as well as reprompting if incorrect values are selected are
now implemented directly in `click` options instead of the function
body. The various database connection options are also restored for
`verdi quicksetup` that were removed in a recent commit. This allows to
override the defaults while still profiting of the automatic resource
creation.

To avoid code duplication, the reusable options are defined for use in
`verdi quicksetup`, which should not prompt for the database connection
options, and these are then extended to become fully interactive options
for `verdi setup` which will make them prompt even if there is a
default.


sphuber:
Note @ltalirz I uncommented the re-added `test_quicksetup` test because it kept failing and I don't know why and I cannot test it locally. Would be good if this branch works for you with `verdi quicksetup`.

.ci/test_script.sh
```diff
@@ -25,6 +25,7 @@ case "$TEST_TYPE" in
         coverage erase
 
         # Run preliminary tests
+        # coverage run -a "${CI_DIR}/test_quicksetup.py"
```
ltalirz: maybe run the test as well ;-)

sphuber: ah no that is on purpose, my comment *meant* to say "I *commented*" the test. I did so because I wanted to see if at least all the rest would work

ltalirz: oh sorry, I went directly to the changes and missed your comment. I will look into it

ltalirz:
@sphuber I did some improvements in https://github.com/ltalirz/aiida_core/tree/fix_2763_clean_verdi_quicksetup_implementation

The tests will still fail, however, since quicksetup actually needs the backend (and I now switched to the plugin fixtures for the quicksetup tests).

Why not use the quicksetup tests as the first testbed for testing loading/unloading of profiles?
By keeping them in a separate python file, we don't run the risk to impact other tests.

aiida/backends/testbase.py
```diff
@@ -185,6 +185,22 @@ def assertClickResultNoException(self, cli_result):
         self.assertIsNone(cli_result.exception, ''.join(traceback.format_exception(*cli_result.exc_info)))
 
 
+class AiiDAPostgresTestCase(AiidaTestCase):
```
ltalirz:
```suggestion
class AiidaPostgresTestCase(AiidaTestCase):
```

aiida/backends/testbase.py
```diff
@@ -185,6 +185,22 @@ def assertClickResultNoException(self, cli_result):
...
+        """Setup the PGTest postgres test cluster."""
+        super(AiiDAPostgresTestCase, cls).setUpClass(*args, **kwargs)
+        from pgtest.pgtest import PGTest
+        cls.pg_test = PGTest()
+
+    @classmethod
+    def tearDownClass(cls, *args, **kwargs):
+        """Close the PGTest postgres test cluster."""
+        super(AiiDAPostgresTestCase, cls).tearDownClass(*args, **kwargs)
+        cls.pg_test.close()
```
ltalirz: this order is a bit inconsistent with the order in the setupclass.
I guess it doesn't matter since the AiidaTestCase does not interact with pgtest directly.

Still, we should probably consider the AiidaTestCase the "inner" thing, i.e. create the pgcluster first and delete it last

aiida/backends/tests/cmdline/commands/test_profile.py
```diff
@@ -128,3 +128,38 @@ def test_delete(self):
...
         self.assertNotIn(self.profile_list[1], result.output)
+
+    @with_temporary_config_instance
+    def test_delete(self):
+        """Test for verdi profile delete command."""
+        from aiida.cmdline.commands.cmd_profile import profile_delete, profile_list
+
+        configuration.reset_profile()
+
+        # Create mock profiles
```
ltalirz: either delete the `mock_profiles` method or use it here (I did delete it in `.ci`, that's why the copied-over code is here)

aiida/cmdline/params/options/commands/setup.py
```diff
@@ -61,12 +66,93 @@ def get_repository_path_default(ctx):
     return os.path.join(AIIDA_CONFIG_FOLDER, 'repository', ctx.params['profile'].name)
 
 
+def get_quicksetup_repository_uri(ctx, param, value):  # pylint: disable=unused-argument
+    """Determine the repository URI to be used as default in `verdi quicksetup`
+
+    :param ctx: click context which should contain the contextual parameters
+    :return: the password
```
ltalirz:
```suggestion
    :return: the repository URI
```

aiida/cmdline/params/options/commands/setup.py
```diff
@@ -61,12 +66,93 @@ def get_repository_path_default(ctx):
...
+    if value is not None:
+        return value
+
+    username = ctx.params['db_username']
+    config = get_config()
+
+    for available_profile in config.profiles:
+        if available_profile.database_username == username:
+            value = available_profile.database_password
+            echo.echo_info('using found password for existing user `{}`'.format(username))
```
ltalirz: this now also prints appears in auto completion.
perhaps we simply get rid of it?

sphuber:
Addressed the changes and also restored authorship to you for last commit, which I noticed went lost accidentally

Issue #2770, opened 2019-04-17
Move process_state from JSONB attribute to a dedicated column


ConradJohnston:
Experience with `verdi process list` shows that queries using the new `process_state` variable are expensive for large databases. This is because  there is no index on the JSONB attributes, and the query must do a sequential scan over these.

Proposed solutions:
1. Move `process_state` to a dedicated column with an index.
2. Add a GIN index to the JSONB (benchmarking needed).

Benchmarking from @sphuber below:

sphuber:
Useful and detailed information on this subject by @giovannipizzi in another thread:

https://github.com/aiidateam/aiida-core/issues/4841#issuecomment-830866937
https://github.com/aiidateam/aiida-core/issues/4841#issuecomment-830867591

Pasted below

-------

There's quite some discussion and debugging on GIN indexes on #4664 done by @CasperWA and me.
The main lessons we learnt are:
- making a GIN index on *everything* is not useful and actually bad (you create a huge index, that will fill your disk and in the end not really help speed up your queries)
- it's possible to make a GIN index only a specific field. This is in general not something we can do in AiiDA core as we don't know which extras or attributes a user will need to search (but an experienced user could add them).
  On the other hand, for the `process_state`, this is something known to AiiDA - so a quick solution, even before trying to move this field to a proper column, would be to add an index *just* for this.

From #4664, the syntax to create a GIN index only on a specific field is this:
```sql
CREATE INDEX idx_gin_optimade_elements ON db_dbnode USING GIN ((extras -> 'optimade' -> 'elements') jsonb_path_ops);
```
and, what's even more interesting, you can create a *partial index* with a WHERE clause (e.g. to only add the index to processes, for instance, to reduce even more the size - note that the index will be used, however, only if the corresponding queries actually have the very same filter - so, for testing, maybe it's better to start with an index without WHERE clause).

The actual command to run should be:
```sql
CREATE INDEX idx_process_state ON db_dbnode USING GIN ((attributes -> 'process_state') jsonb_ops);
```
(one should check which operators are actually used, and if we can replace `jsonb_ops` with the faster but more limited `json_path_ops`, see [bottom of this docs page](https://www.postgresql.org/docs/9.5/gin-builtin-opclasses.html).)

(Note: when timing, note that there a few last operations that are also slow: checking if there are too many slots used and issuing a warning, and checking when the status was last changed - in timing, you should remove that part; I don't remember if the --raw option does that, or you need to change the source code for timing).

[...]

PR #2774, opened 2019-04-17
Add slots warning to verdi process list

### LINKED ISSUE #1397: Deadlock when creating many processes
When many processes are created, the workers will clog up and keep checking the status of a few processes, without actually running the processes that it depends on.

This is due to the prefetch limit.

Possible ways to solve this:
- Set the prefetch limit to 0 (no limit). Maybe this would create some problems with the heartbeat?
- Add a scheduler which tells the workers to "drop" waiting processes and pick up new ones.

ConradJohnston:
Fixes #1397 

- Adds a warning to be printed by `verdi process list` when the number of active processes exceeds 90% of the available worker slots, calculated by _RMQ_TASK_PREFETCH_COUNT * the number of active workers.
- Adds a test to make sure it fires the warning in the correct conditions.
- Adds a get_numprocess() utility method to the AiiDA circus client class to simplify getting the number of active workers

Currently, this solution requires running a second query in `verdi profile list` which can be slow for large DBs. This will be fixed by issue #2770 .


giovannipizzi:
Some tests are failing

ConradJohnston:
Yes, these are a bit wierd. The behaviour of these tests on Travis/Jenkins
is different from my local machine.

ConradJohnston:
Tests should pass now. 

I did a slightly dangerous thing in the test setup which we can revisit. I need the daemon PID file to be written in order for a check to pass in the calculation of the number of available worker slots. The existing test setup for `verdi process pause/play/stop`  does create a daemon runner, but when setup in this way, the daemon PID file is not created and so `is_daemon_running` always returns `False`.
We use the CLI runner and start the daemon using `verdi` in the StartUp fixture and then stop in the TearDown fixture. This is non-ideal as if `verdi daemon` is broken or someone hits control+c, the daemon may be left in a weird state for the test profile.

ConradJohnston:
Marking this as WIP while I try to improve the test setup.

ConradJohnston:
Now good to go, @sphuber .
Refactored the call to circus to get the number of active workers into a separate function. This allows us to bluff the number of active workers in the tests by temporarily redefining this function with a lamda function, removing the need for a running daemon, messing with PID files and suprocesses, etc.

aiida/backends/tests/cmdline/commands/test_process.py
```diff
@@ -219,6 +220,58 @@ def test_list(self):
...
+    def test_list_worker_slot_warning(self):
+        """
+        Test that the if the number of used worker process slots exceeds a threshold,
+        that the warning message is displayed to the user when running `verdi process list`
+        """
+        from aiida.engine import ProcessState
+
+        # Create 6 WorkFunctionNodes (one for each ProcessState)
+        # Do this 28 times such that we have an additional 168 nodes, of which 84 are 'active' states
+        # Total now is 90 active nodes and 90 inactive nodes.
```
sphuber: You create 84 active but the total now is 90??? Also, since we only care about active states, why not just create however many you need with `Running`. No need for the 'complex' logic in the creation of those nodes.

Edit: I see now why there are 90 total, the other 6 come from the `setupClass`. Maybe add this in the comment

aiida/backends/tests/cmdline/commands/test_process.py
```diff
@@ -219,6 +220,58 @@ def test_list(self):
...
+                if state == ProcessState.FINISHED:
+                    calc.set_exit_status(0)
+
+                calc.store()
+
+        # Override the call to the circus client to retrieve the number of workers
+        # As we don't have a running circus client, this will normally fail, so here we simulate the
+        # response by redefining the function to get the final value we want.
+        import aiida.cmdline.utils.common
+        aiida.cmdline.utils.common.get_num_workers = lambda: 1
```
sphuber: We should probably reset the original function after the test. I can imagine some other test being used in the future that calls it indirectly, and this will send somebody potentially on a wild goose chase! :)

aiida/backends/tests/cmdline/commands/test_process.py
```diff
@@ -219,6 +220,58 @@ def test_list(self):
...
+
+        # Default cmd should not throw the warning as 90/100 !> 90
+        result = self.cli_runner.invoke(cmd_process.process_list)
+        self.assertIsNone(result.exception, result.output)
+        warning_phrase = "of the available daemon worker slots have been used!"
+        self.assertTrue(all([warning_phrase not in line for line in get_result_lines(result)]))
+
+        # Add 6 more nodes
```
sphuber: Here I would not add 6 of all states. Just make sure you have `limit - 1` at the beginning of the test and then here you create one more with `ProcessState.RUNNING`.

aiida/cmdline/utils/common.py
```diff
@@ -410,3 +410,56 @@ def build_entries(ports):
...
+    :param active_slots: the number of currently active worker slots
+    """
+    from aiida.cmdline.utils import echo
+    from aiida.manage.external.rmq import _RMQ_TASK_PREFETCH_COUNT
+
+    warning_threshold = 0.9  # 90%
+
+    slots_per_worker = _RMQ_TASK_PREFETCH_COUNT
+
+    active_workers = get_num_workers()
```
sphuber: This can potentially throw the `CallError`. It would be better to catch it here and print a `echo_critical` instead of letting the exception bubble up. That said, I would use an aiida internal exception to throw from `get_num_workers` instead of the circus one. Normally you would import those to be able to catch them, but if you are throwing them yourself, might as well use an internal exception

aiida/backends/tests/cmdline/commands/test_process.py
```diff
@@ -219,6 +220,51 @@ def test_list(self):
...
+        calc.store()
+
+        # Now the warning should fire
+        result = self.cli_runner.invoke(cmd_process.process_list)
+        self.assertIsNone(result.exception, result.output)
+        warning_phrase = "% of the available daemon worker slots have been used!"
+        self.assertTrue(any([warning_phrase in line for line in get_result_lines(result)]))
+
+        # Reset the redifined function
+        aiida.cmdline.utils.common.get_num_workers = real_get_num_workers
```
sphuber: I am afraid that this might still be problematic, if the tests excepts before it gets here. I think the safest think would be to separate this test in its own class, so you can properly change the `get_num_workers` function in the `setUp` and `tearDown`. Sorry to be asking yet another change, but the branch anyway had to be brought up to date, with develop. I promise this is the last change I ask for :)

ConradJohnston: Ha, no problem. 
I've not refactored this test in to it's own class, and we do the method overloading in the setup and teardown.

sphuber: *not* should be *now* ? :wink:

sphuber:
Absolute blinder mate! Thanks a lot

aiida/backends/tests/cmdline/commands/test_process.py
ConradJohnston: Brutal

PR #2778, opened 2019-04-17
Remove `aiida_tests` entry point group and `parser_tests`

### LINKED ISSUE #1091: Automated parser test discovery for installed plugins
I implemented an addition to the `PwParser` of the `aiida-quantumespresso` and wanted to write some tests for this as well. I found a folder `parser_tests` in the `tests` folder of the plugin repository and subsequently found in the developers documentation that these used to live in `aiida-core` in the path `aiida.backends.tests.parser_tests` and the test class in `aiida.backends.tests.parsers` would know how to read the content of those directories and automatically run a standard test. Since the tests were moved to the `aiida-quantumespresso` package, the `verdi devel tests db.parsers` command now runs zero tests. I guess the functionality to discover parser tests through the plugin system has not yet been implemented? Is this the case? I remember that we released `v0.11.1` this month because it contained new functionality regarding the testing of plugins. Was this not included?

sphuber:
Fixes #1091 

The `aiida_tests` entry point was added as a temporary replacement until
the plugin test fixtures came available. These have been available for a
while now and since no plugins rely on the entry point group we remove
it. Additionally, the `test_parsers` was a first implementation to test
parser plugins, but this can be done more elegantly with `pytest` and
the plugin fixtures, therefore it is removed.


sphuber:
Note, I have gone through all the plugins on the registry and looked at their `setup.json` and none of them use the `aiida.tests` entry point group except for `aiida-nwchem` and `aiida-castep`. The former we control and we can easily migrate to use pytest with the fixtures. @zhubonan are you still relying on the `aiida.tests` entry point, which contains one test for the `CastepCalculation` or have you also migrated to using `pytests`?

zhubonan:
I still use it for 0.12.x but it is no longer used in the 1.0.0 branch as all tests has been migrated. I have not updated the setup.json yet probably. Please go ahead and remove the entrypoint.

On 17 April 2019 18:21:25 Sebastiaan Huber <notifications@github.com> wrote:

Note, I have gone through all the plugins on the registry and looked at their setup.json and none of them use the aiida.tests entry point group except for aiida-nwchem and aiida-castep. The former we control and we can easily migrate to use pytest with the fixtures. @zhubonan<https://github.com/zhubonan> are you still relying on the aiida.tests entry point, which contains one test for the CastepCalculation or have you also migrated to using pytests?

PR #2779, opened 2019-04-17
Ensure `KpointsData.reciprocal_cell()` works again

### LINKED ISSUE #2749: .reciprocal_cell not found when KPointsData has already been stored
When I store a k point data, the .reciprocal_cell disappears.
Simple example:
alat = 5.430  # angstrom
cell = [[0.5 * alat, 0.5 * alat, 0.,], [0., 0.5 * alat, 0.5 * alat,],[0.5 * alat, 0., 0.5 * alat,],]
structure = StructureData(cell=cell)
structure.append_atom(position=(0.000 * alat, 0.000 * alat, 0.000 * alat), symbols=['Si'])
structure.append_atom(position=(0.250 * alat, 0.250 * alat, 0.250 * alat),  symbols=['Si'])
k=KpointsData()
k.set_cell(structure)
k.store_all()

Now compare k.reciprocal_cell with load_node(pk)reciprocal_cell

Is it meant to be like that or is it a bug??
This is breaking all my siesta bands parser. Please let me know

giovannipizzi:
This fixes #2749
The `reciprocal_cell` is now a property instead of being
an attribute that is attached by calling a method.

Also, a couple of tests have been added, and the module
kpoints.py fixed to make prospector happy.


aiida/orm/nodes/data/array/kpoints.py
```diff
@@ -228,16 +219,17 @@ def set_cell(self, cell, pbc=None):
...
-    def _set_reciprocal_cell(self):
+    @property
+    def reciprocal_cell(self):
         """
-        Sets the reciprocal cell in units of 1/Angstrom from the internally set cell
+        Return the reciprocal cell in units of 1/Angstrom from the internally set cell
         """
         the_cell = numpy.array(self.cell)
-        self.reciprocal_cell = 2. * numpy.pi * numpy.linalg.inv(the_cell).transpose()
+        reciprocal_cell = 2. * numpy.pi * numpy.linalg.inv(the_cell).transpose()
```
sphuber: Would it maybe make sense to construct this lazily and cache it? Or is it too cheap a computation to even bother

aiida/orm/nodes/data/array/kpoints.py
```diff
@@ -268,8 +262,7 @@ def set_kpoints_mesh(self, mesh, offset=[0., 0., 0.]):
         # I cannot have both of them at the same time
         try:
             _ = self.get_array('kpoints')
-            raise ModificationNotAllowed("KpointsData has already a kpoint-"
-                                         "list stored")
+            raise ModificationNotAllowed("KpointsData has already a kpoint-" "list stored")
```
sphuber: There is quite a few occurrences of `" "` due to `yapf`

giovannipizzi: Good catch, will fix it

aiida/orm/nodes/data/array/kpoints.py
giovannipizzi: I thought about it, I decided it's very cheap, and also "dangerous" to cache because in principle one could "re-set" the cell and then one should invalidate the cache. Possible, but probably not worth it unless we have a real usecase of the benefits.

PR #2782, opened 2019-04-18
Change the way hash is calculated for float

### LINKED ISSUE #2631: Storing a node changes its hash, due to float rounding
```python
p = ParameterData(dict={'test': [3.5615599999999996, 4.508439999999999, 0.8715600000000004]})
print(p.get_attrs())
print(p.get_hash())
p.store()
print(p.get_attrs())
print(p.get_hash())
```
```
{'test': [3.5615599999999996, 4.508439999999999, 0.8715600000000004]}
5632c77efbfe1d0690e71321d0674715f76f5fc14d01b28f4f19bb23
{u'test': [3.56156, 4.50844, 0.87156]}
6f9926c50e8447dff7b7f89751d466ece9f392449db37ebdcf1e9838
```

Only tested on v0.12.3, but I assume this will be the same in v1.

Bit of a nuisance when, for example, you're trying to compare/reuse input Structure nodes, 
created *via* ASE/Pymatgen functions.

zhubonan:
Fixes #2631 

This pull request will change the hash is calculated for float. 
Now the hash is computed by converting a float to string with only 14 significant
digits (e.g by `{:.14g}.format(val)`. This is consistent with the way rounding error may when saving a float to the database.
This should fix the issue where the hash of a node containing float attributes may change when it is stored #2631 .

PS: The existing `truncate_float64` function is kept.


zhubonan:
@dev-zero @giovannipizzi Would you be able to review this PR? I understand it is probably of low priority and if merged require all hashes to be recomputed.

dev-zero:
hmm, I see the way you want to fix it, but I don't think that conversion to string is the right approach. If at all, can't we decrease the precision by increasing the mask?

Furthermore this still leaves the issue with floats occurring inside numpy arrays since there we go directly from an array to a byte-representation (and then to a hash of it).

zhubonan:
@dev-zero The problem is that increasing the mask does not work as expected when rounding occurs during communication with the database backend, which I suspect is based on strings. For example, 1.99999999999 would round to 2.000000000. If we just drop the last significant digits (not quite the same as applying the mask to the float but the similar idea), it will be 1.99999999990 and no matter how the mask has increased the rounding behaviour cannot be reproduced.
Hence, the mask works for some cases but fail for the rest.

sphuber:
@zhubonan please don't keep updating the branch, until a PR review has been completed. The reason is that each time a build is triggered, which takes a full 30 minutes and blocks any other build from running. This goes for *all* projects on `aiidateam` and is significantly blocking other repos. Until the PR has been accepted there is no need to update. The person responsible for merging it will make sure it is up to date. Thanks

zhubonan:
@sphuber  I see, sorry about that. I thought it had to pass the building before review.....

giovannipizzi:
Sorry about the delay, I also wasn't convinced of the correctness of this approach and this got me delayed.

I also think that this is not a perfect fix (see below) - still, if we don't have a better solution and this already improves the situation in most cases with respect to the current serialiser we might think to replace the function before 1.0.

Anyway, to show that this is not enough, try this:
```python
import math
import json

def float64_to_text(value, sig=14):
    """
    Convert float to text string for computing hash.
    Preseve up to N significant number given by sig.
    """

    fmt = '{:.' + str(sig) + 'g}'
    return fmt.format(value)

def my_hash(val):
    return float64_to_text(val).encode('utf-8')

def json_serialize(val):
    return json.dumps(val)

def json_deserialize(val):
    return json.loads(val)

def custom_serialize(val):
    return float64_to_text(val, sig=13)

def custom_deserialize(val):
    return float(val)

if __name__ == "__main__":
    for serializer_label, (serialize, deserialize) in [
            ('json', (json_serialize, json_deserialize)),
            ('custom', (custom_serialize, custom_deserialize)),
        ]:
        for val in [1., 0.3, 2., math.pi, math.e, math.sqrt(2.)]:
            before = my_hash(val)
            roundtrip = deserialize(serialize(val))
            after = my_hash(roundtrip)
            assert before == after, "Error with serializer={}, val={}, before={}, roundtrip={}, after={}".format(
                serializer_label, val, before, roundtrip, after
            )
```
that (at least to me) returns
```
  File "example.py", line 38, in <module>
    serializer_label, val, before, roundtrip, after
AssertionError: Error with serializer=custom, val=3.14159265359, before=3.1415926535898, roundtrip=3.14159265359, after=3.14159265359
```

I don't know if, in the case the `sig` of the hashing function is less precise than the one of the DB, if this would be enough to guarantee consistency after a roundtrip, I'll need to think about this. If this is the case, we need to decrease the default `sig` probably, and test this for both backends [that have different precision when serialising](https://github.com/aiidateam/aiida_core/issues/2124#issuecomment-434706777)

zhubonan:
@giovannipizzi Thanks for testing. Yes, the number of significant values need to be reduced for json serialization.  I only tested when storing to DB with the django backend in the first place.
I agree that this approach is not perfect - we will still have collisions beyond the defined precision (e.g. false positive).  The main thing I was trying to fix is to make sure the hash does not change after calling `store()`.

giovannipizzi:
Superseded by #3108

PR #2793, opened 2019-04-24
Added separate user projection in query help for rest api

### LINKED ISSUE #2764: REST API should not be based on Hybrid properties

waychal:
Resolves #2764

### DIFF
--- a/aiida/restapi/translator/node.py
+++ b/aiida/restapi/translator/node.py
@@ -66,8 +66,9 @@ class NodeTranslator(BaseTranslator):
         super(NodeTranslator, self).__init__(Class=Class, **kwargs)
 
         self._default_projections = [
-            "id", "label", "node_type", "ctime", "mtime", "uuid", "user_id", "user_email", "attributes", "extras"
+            "id", "label", "node_type", "ctime", "mtime", "uuid", "user_id", "attributes", "extras"
         ]
+        self._default_user_projections = ["email"]
 
         ## node schema
         # All the values from column_order must present in additional info dict
@@ -258,8 +259,7 @@ class NodeTranslator(BaseTranslator):
         if not self._is_qb_initialized:
             raise InvalidOperation("query builder object has not been initialized.")
 
-        ## Count the total number of rows returned by the query (if not
-        # already done)
+        # Count the total number of rows returned by the query (if not already done)
         if self._total_count is None:
             self.count()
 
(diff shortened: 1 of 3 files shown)


aiida/restapi/translator/node.py
```diff
@@ -268,7 +269,7 @@ def _get_content(self):
             return {}
 
         # otherwise ...
-        node = self.qbobj.first()[0]
+        node = self.qbobj.all()[0][1]
```
szoupanos: We should see with Leonid why this is not possible anymore.
If we don't find a solution, a ticket should be opened.

szoupanos:
This should be merged after it is tested with the django_jsonb branch where the hybrid properties are removed.

I will need help from @sphuber to rebase the django_jsonb branch since after his last PRs I have conflicts that are not obvious to resolve.

Then we can test this PR with the rebased django_jsonb and merge it to develop.

aiida/restapi/translator/node.py
sphuber: The issue describing this #2808 I have fixed in PR #2824 so this change can now be reverted

sphuber:
I have tested this with the `django_jsonb` branch and confirmed that it fixes the tests. I will merge this in `develop` and then rebase `django_jsonb` which will allow the REST API tests to be re-enabled.

PR #2795, opened 2019-04-24
Deal with stale daemon PID files during stop attempts

### LINKED ISSUE #1865: Daemon fails to start after being killed
When killing the daemon and removing the sockets directory given in ``circus-*.sockets``, the damon can no longer start:

```
$ verdi daemon start
Starting the daemon... FAILED
Try to run 'verdi daemon start --foreground' to potentially see the exception
```

When called with ``--foreground`` it hangs at 
```
$ verdi daemon start --foreground
Starting the daemon... 2018-08-10 02:13:53 circus[17615] [INFO] Starting master on pid 17615
2018-08-10 02:13:53 circus[17615] [INFO] Arbiter now waiting for commands
2018-08-10 02:13:53 circus[17615] [INFO] aiida-default started
2018-08-10 02:13:53 circus[17615] [INFO] circusd-stats started
2018-08-10 02:13:53 circus[17628] [INFO] Starting the stats streamer
```

I would guess the reason is that the daemon tries to connect on the inexisting sockets, which obviously doesn't work anymore.

It's not a big issue, since obviously it shouldn't happen too often that the daemon is manually ``SIGKILL``ed. However, maybe there's an easy solution to prevent this issue? When starting the daemon, can we make a sanity check whether the socket directory exists?

### LINKED ISSUE #2775: Stale PID files are not always removed
Currently, there is a possibility that the PID file in the .aiida folder containing the PID of the running `verdi` process is not cleared when the process is killed. 

When one then tries to start the daemon, `verdi` complains that the daemon is already running because it finds the PID file. This file must then be removed by hand.

ConradJohnston:
Fixes #2775 and fixes #1865 

In some circumstances, the circus PID file which stores the PID of the running circus process can be corrupted, causing calls to the `get_daemon_pid()` method to return the wrong PID, or causing the `is_daemon_running` attribute to return `True` when in reality the daemon is not running.

This function checks if the PID contained in the circus pid file (`circus-{PROFILE_NAME}.pid`) matches a valid running `verdi` process. If it does not, the PID file is stale and will be removed.

This situation can arise if a system is shut down suddenly and so the process is killed but the PID file is not deleted in time. Alternatively, another process or the user may have meddled with the PID file in some way, corrupting it.

Example of broken behaviour:
`verdi daemon start`
`pid=$(cat ~/.aiida/daemon/circus-{PROFILE_NAME}.pid)`
`echo '9999999999' >  ~/.aiida/daemon/circus-{PROFILE_NAME}.pid`
`kill $pid`
`verdi daemon stop`
In this scenario, the `daemon stop` command will timeout, as will the `start` command, due to the stale PID file which is lying about there being a running daemon with the claimed PID.

### DIFF
--- a/aiida/cmdline/commands/cmd_daemon.py
+++ b/aiida/cmdline/commands/cmd_daemon.py
@@ -22,7 +22,8 @@ from click_spinner import spinner
 from aiida.cmdline.commands.cmd_verdi import verdi
 from aiida.cmdline.utils import decorators, echo
 from aiida.cmdline.utils.common import get_env_with_venv_bin
-from aiida.cmdline.utils.daemon import get_daemon_status, print_client_response_status
+from aiida.cmdline.utils.daemon import get_daemon_status, \
+    print_client_response_status, delete_stale_pid_file, _START_CIRCUS_COMMAND
 from aiida.manage.configuration import get_config
 
 
@@ -46,9 +47,9 @@ def start(foreground):
     echo.echo('Starting the daemon... ', nl=False)
 
     if foreground:
-        command = ['verdi', '-p', client.profile.name, 'daemon', 'start-circus', '--foreground']
+        command = ['verdi', '-p', client.profile.name, 'daemon', _START_CIRCUS_COMMAND, '--foreground']
     else:
-        command = ['verdi', '-p', client.profile.name, 'daemon', 'start-circus']
+        command = ['verdi', '-p', client.profile.name, 'daemon', _START_CIRCUS_COMMAND]
 
     try:
         currenv = get_env_with_venv_bin()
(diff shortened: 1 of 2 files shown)


ConradJohnston:
@giovannipizzi , requested your review. Even though you co-authored the commit and the idea for the implementation was yours, you didn't touch any of the real code in the end so maybe it is okay for you to review.

ConradJohnston:
What do we want to do get this over the line? Remove the dependence on  psutil? I think this is the only place it's used in AiiDA now.

sphuber:
For now, I think we can leave the `psutil` dependency, because this fix is useful enough. Thanks a lot @ConradJohnston and sorry for the delay with the review

PR #2798, opened 2019-04-26
Fix bugs in migrations of calculation job node attribute keys

### LINKED ISSUE #2797: Migration 0022_dbgroup_type_string_change_content (Django) takes too long
This migration very likely is wrong.
I started to look at it since it takes **more than 2 days** on a 7M node database

According to the description we would like to migrate CalcJobNode attributes for metadata options whose key changed.

    Renamed attribute keys:

      * `custom_environment_variables` -> `environment_variables`
      * `jobresource_params` -> `resources`
      * `_process_label` -> `process_label`
      * `parser` -> `parser_name`

    Deleted attributes:
      * `linkname_retrieved` (We do not actually delete it just in case some relies on it

If I have a look at the first update sql query, I don't see any join between the attributes and the nodes:
```
UPDATE db_dbattribute AS attribute
SET key =  regexp_replace(attribute.key, '^custom_environment_variables', 'environment_variables')
FROM db_dbnode AS node
WHERE
	(
		attribute.key = 'custom_environment_variables' OR
		attribute.key LIKE 'custom\_environment\_variables.%'
	) AND
	node.type = 'node.process.calculation.calcjob.CalcJobNode.';
```

I analyze the execution plan of the query (to ensure that no join will be added - e.g. implicitly) and I see the following:

```
                                                                          QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------
 Update on db_dbattribute attribute  (cost=1856.29..14419903.13 rows=810381824 width=569)
   ->  Nested Loop  (cost=1856.29..14419903.13 rows=810381824 width=569)
         ->  Bitmap Heap Scan on db_dbattribute attribute  (cost=1510.70..205773.67 rows=77504 width=63)
               Recheck Cond: (((key)::text = 'custom_environment_variables'::text) OR ((key)::text ~~ 'custom\_environment\_variables.%'::text))
               Filter: (((key)::text = 'custom_environment_variables'::text) OR ((key)::text ~~ 'custom\_environment\_variables.%'::text))
               ->  BitmapOr  (cost=1510.70..1510.70 rows=57439 width=0)
                     ->  Bitmap Index Scan on db_dbattribute_key_6936ff5c4f96a1be_like  (cost=0.00..1467.37 rows=57439 width=0)
                           Index Cond: ((key)::text = 'custom_environment_variables'::text)
                     ->  Bitmap Index Scan on db_dbattribute_key_6936ff5c4f96a1be_like  (cost=0.00..4.58 rows=1 width=0)
                           Index Cond: (((key)::text ~>=~ 'custom_environment_variables.'::text) AND ((key)::text ~<~ 'custom_environment_variables/'::text))
         ->  Materialize  (cost=345.59..32473.68 rows=10456 width=6)
               ->  Bitmap Heap Scan on db_dbnode node  (cost=345.59..32421.40 rows=10456 width=6)
                     Recheck Cond: ((type)::text = 'node.process.calculation.calcjob.CalcJobNode.'::text)
                     ->  Bitmap Index Scan on db_dbnode_type_4cda33f938ccd765_like  (cost=0.00..342.98 rows=10456 width=0)
                           Index Cond: ((type)::text = 'node.process.calculation.calcjob.CalcJobNode.'::text)
```

Which mentions at the top level, that it results to 810.381.824 rows (!!!) when the attribute table has "only" 282.505.715. So it seems to apply the attribute update to every attribute that satisfies the condition (irrespective to which type of node it corresponds to) and this is repeated for all the nodes that satisfy the condition.

I didn't check if it just runs the update db_dbnode times (for the calc nodes) or if it adds db_dbnode times (for the calc nodes) updated attributes to the table. But this is less important.

After adding the join condition to the update
```
UPDATE db_dbattribute AS attribute
SET key =  regexp_replace(attribute.key, '^custom_environment_variables', 'environment_variables')
FROM db_dbnode AS node
WHERE
	(
		attribute.key = 'custom_environment_variables' OR
		attribute.key LIKE 'custom\_environment\_variables.%'
	) 
	AND
	node.type = 'node.process.calculation.calcjob.CalcJobNode.'
	AND
	node.id = attribute.dbnode_id;
```

I get the following execution plan which is more normal (in the number of expected rows)
```
                                                                          QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------
 Update on db_dbattribute attribute  (cost=34062.80..238618.06 rows=110 width=569)
   ->  Hash Join  (cost=34062.80..238618.06 rows=110 width=569)
         Hash Cond: (attribute.dbnode_id = node.id)
         ->  Bitmap Heap Scan on db_dbattribute attribute  (cost=1510.70..205773.67 rows=77504 width=63)
               Recheck Cond: (((key)::text = 'custom_environment_variables'::text) OR ((key)::text ~~ 'custom\_environment\_variables.%'::text))
               Filter: (((key)::text = 'custom_environment_variables'::text) OR ((key)::text ~~ 'custom\_environment\_variables.%'::text))
               ->  BitmapOr  (cost=1510.70..1510.70 rows=57439 width=0)
                     ->  Bitmap Index Scan on db_dbattribute_key_6936ff5c4f96a1be_like  (cost=0.00..1467.37 rows=57439 width=0)
                           Index Cond: ((key)::text = 'custom_environment_variables'::text)
                     ->  Bitmap Index Scan on db_dbattribute_key_6936ff5c4f96a1be_like  (cost=0.00..4.58 rows=1 width=0)
                           Index Cond: (((key)::text ~>=~ 'custom_environment_variables.'::text) AND ((key)::text ~<~ 'custom_environment_variables/'::text))
         ->  Hash  (cost=32421.40..32421.40 rows=10456 width=10)
               ->  Bitmap Heap Scan on db_dbnode node  (cost=345.59..32421.40 rows=10456 width=10)
                     Recheck Cond: ((type)::text = 'node.process.calculation.calcjob.CalcJobNode.'::text)
                     ->  Bitmap Index Scan on db_dbnode_type_4cda33f938ccd765_like  (cost=0.00..342.98 rows=10456 width=0)
                           Index Cond: ((type)::text = 'node.process.calculation.calcjob.CalcJobNode.'::text)
```

I will update the migration queries and I report again how it goes.
We should add also 1 test that demonstrates that attributes of other nodes are nodes are not updated

### LINKED ISSUE #2831: SqlAlchemy migration of calculation job keys incorrectly converts to text
The migration `7ca08c391c49_calc_job_option_attribute_keys.py` for SqlAlchemy that renames the keys of certain attributes of `CalcJobNodes` incorrectly converts the dictionaries of `resources` and `environment_variables` from a dictionary to text.

szoupanos:
Fixes #2797 and fixes #2831 

This PR fixes two distinct bugs in the migrations for key name changes in the attributes of calculation job nodes.

The faulty migration has been added in commit [2c961cbbb89aa8f2a2ab257dcb34322f7cc98bb9](https://github.com/aiidateam/aiida_core/commit/2c961cbbb89aa8f2a2ab257dcb34322f7cc98bb9#diff-f712699834f2257f35791ed0de4c788e) and so was present between version `v1.0.0b1 - v1.0.0b3`.

The Django migration `0023_calc_job_option_attribute_keys` failed to properly join the attributes on the filtered nodes, potentially causing attribute keys of non `CalcJobNode` to erroneously be updated. The chances of this having to occurred while the bug was present are negligible.

The bug in the SqlAlchemy migration `7ca08c391c49_calc_job_option_attribute_keys` is more serious. The used JSONB operator was wrong and erroneously converted the dictionaries for the attributes `resources` and `environment_variables` to text. All SqlAlchemy databases that have activated the migration between `v1.0.0b1 - v1.0.0b3` will now have values with type text instead of dictionaries for the attributes `resources` and `environment_variables`.

### DIFF
--- a/aiida/backends/djsite/db/migrations/0023_calc_job_option_attribute_keys.py
+++ b/aiida/backends/djsite/db/migrations/0023_calc_job_option_attribute_keys.py
@@ -8,7 +8,7 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 # pylint: disable=invalid-name,too-few-public-methods
-"""Migration of CalcJobNode attributes for metadata options whose key changed."""
+"""Migration of ProcessNode attributes for metadata options whose key changed."""
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
@@ -25,14 +25,14 @@ DOWN_REVISION = '1.0.22'
 
 
 class Migration(migrations.Migration):
-    """Migration of CalcJobNode attributes for metadata options whose key changed.
+    """Migration of ProcessNode attributes for metadata options whose key changed.
 
     Renamed attribute keys:
 
-      * `custom_environment_variables` -> `environment_variables`
-      * `jobresource_params` -> `resources`
-      * `_process_label` -> `process_label`
-      * `parser` -> `parser_name`
+      * `custom_environment_variables` -> `environment_variables` (CalcJobNode)
+      * `jobresource_params` -> `resources` (CalcJobNode)
+      * `_process_label` -> `process_label` (ProcessNode)
+      * `parser` -> `parser_name` (CalcJobNode)
 
     Deleted attributes:
       * `linkname_retrieved` (We do not actually delete it just in case some relies on it)
@@ -54,7 +54,8 @@ class Migration(migrations.Migration):
                     attribute.key = 'custom_environment_variables' OR
                     attribute.key LIKE 'custom\_environment\_variables.%'
                 ) AND
-                node.type = 'node.process.calculation.calcjob.CalcJobNode.';
+                node.type = 'node.process.calculation.calcjob.CalcJobNode.' AND
+                node.id = attribute.dbnode_id;
             -- custom_environment_variables -> environment_variables
 
             UPDATE db_dbattribute AS attribute
@@ -65,7 +66,8 @@ class Migration(migrations.Migration):
                     attribute.key = 'jobresource_params' OR
                     attribute.key LIKE 'jobresource\_params.%'
                 ) AND
-                node.type = 'node.process.calculation.calcjob.CalcJobNode.';
+                node.type = 'node.process.calculation.calcjob.CalcJobNode.' AND
+                node.id = attribute.dbnode_id;
             -- jobresource_params -> resources
 
             UPDATE db_dbattribute AS attribute
@@ -73,7 +75,8 @@ class Migration(migrations.Migration):
             FROM db_dbnode AS node
             WHERE
                 attribute.key = '_process_label' AND
-                node.type LIKE 'node.process.%';
+                node.type LIKE 'node.process.%' AND
+                node.id = attribute.dbnode_id;
             -- _process_label -> process_label
 
             UPDATE db_dbattribute AS attribute
@@ -81,7 +84,8 @@ class Migration(migrations.Migration):
             FROM db_dbnode AS node
             WHERE
                 attribute.key = 'parser' AND
-                node.type = 'node.process.calculation.calcjob.CalcJobNode.';
+                node.type = 'node.process.calculation.calcjob.CalcJobNode.' AND
+                node.id = attribute.dbnode_id;
             -- parser -> parser_name
             """,
             reverse_sql=r"""
@@ -93,7 +97,8 @@ class Migration(migrations.Migration):
                     attribute.key = 'environment_variables' OR
                     attribute.key LIKE 'environment\_variables.%'
                 ) AND
-                node.type = 'node.process.calculation.calcjob.CalcJobNode.';
+                node.type = 'node.process.calculation.calcjob.CalcJobNode.' AND
+                node.id = attribute.dbnode_id;
             -- environment_variables -> custom_environment_variables
 
             UPDATE db_dbattribute AS attribute
@@ -104,7 +109,8 @@ class Migration(migrations.Migration):
                     attribute.key = 'resources' OR
                     attribute.key LIKE 'resources.%'
                 ) AND
-                node.type = 'node.process.calculation.calcjob.CalcJobNode.';
+                node.type = 'node.process.calculation.calcjob.CalcJobNode.' AND
+                node.id = attribute.dbnode_id;
             -- resources -> jobresource_params
 
             UPDATE db_dbattribute AS attribute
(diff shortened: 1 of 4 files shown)


sphuber:
You changed all `CalcJobNode` checks to `ProcessNode` but that is incorrect. All of the attributes, with the exception of `_process_label`, only apply to `CalcJobNode` and not `ProcessNodes` in general

szoupanos:
> You changed all `CalcJobNode` checks to `ProcessNode` but that is incorrect. All of the attributes, with the exception of `_process_label`, only apply to `CalcJobNode` and not `ProcessNodes` in general

Why do you say so?

I followed the following rules:
* `custom_environment_variables` -> `environment_variables` (CalcJobNode)
* `jobresource_params` -> `resources` (CalcJobNode)
* `_process_label` -> `process_label` (ProcessNode)
* `parser` -> `parser_name` (CalcJobNode)

With node.type `'node.process.%'`  to identify the ProcessNodes and `'node.process.calculation.calcjob.CalcJobNode.'` to identify the CalcJobNode.

The above applies for both migrations. SQLA and Django

sphuber:
You are right, for some reason I misread and thought the node type condition was changed, but is was just in the comments. My bad. In principle this looks good, thanks, I was just looking at the other thing you noticed though where I added the `ast.literal_eval` for the dictionary values in the SQLA tests to convert back to a dictionary. I created a test calculation for an SQLA backend on the latest develop and there if I go through the API i.e. `type(node.attributes['resources']` I get a dictionary, and there we don't do any explicit converting as far as I know. Also directly accessing the model column will give a dictionary: `type(node.backend_entity.dbmodel.attributes['resources'])`. So maybe this is just a side effect of the `DbNode` class we use in the tests? It is not the actual database model class but something returned by alembic correct? Maybe that one does not do a full JSONB deserialization. In any case, the normal API seems to do the correct thing.

szoupanos:
No worries. And thanks for the tests!

Just for the history, the old migration apart from being wrong, it was taking too long (I stopped it after at 3-4 days). The new one (the one mentioned above) takes 10 hours on a 7M database with 282M rows in the attribute table.

Regarding the node.attributes.resources, the above test is on-top of develop. So it has nothing to do with the new alembic schema (in any case, alembic is for Django).

What is even more weird is that when the migration is applied, for the records that it changes, are a string. E.g. node_calc.attributes.get(self.KEY_RESOURCES_NEW)

The records that are not migrated, the result of  node_calc.attributes.get(self.KEY_RESOURCES_OLD) is a dictianary. If you look at the extension that I did to aiida.backends.sqlalchemy.tests.test_migrations.TestCalcAttributeKeysMigration#test_attribute_key_changes, I am not forced to use ast.literal for the check of node_other.attributes.get(self.KEY_ENVIRONMENT_VARIABLES_OLD)

sphuber:
@giovannipizzi I have reviewed the first commit by @szoupanos (although it would be good if you can also give it a second look) which addressed the bug in the Django migration. Then I have added a second commit that fixes a separate bug on the SqlAlchemy version of the same migration. This one is more problematic, although given that this bug was only present for the beta versions, which should not have been used with production databases, it might be ok.

PR #2799, opened 2019-04-26
remove useless chainmap dependency

ltalirz:
it seems the dependency was introduced together with the caching
mechanism but ended up not being used at all.
Will remove the dependency in 1.0.0 as well.

Note: this is breaking the conda build - the dependency is currently not installed (because it is not available on conda-forge) and so you get an import error when doing `verdi quicksetup`.


ltalirz:
Just want to point out that the same travis test passes on my fork:
https://travis-ci.org/ltalirz/aiida_core/builds/525093445?utm_medium=notification&utm_source=email

ltalirz:
@waychal after merging this  & updating your REST API, the docs build should, hopefully, pass

Issue #2813, opened 2019-04-30
Database environment needs to be properly "unloadable"


sphuber:
If we want profiles to be switchable in a single interpreter, the database environment needs to be switchable as well. However, currently for Django, once the `aiida.backends.djsite.settings` file is imported, the variables for example the database connection, which are based on the currently loaded profile, will be locked in. So even if the profile would be swapped for another django profile, with a different database, the original database will be used anyway.

giovannipizzi:
A comment to be taken into account: the purpose of `recreate_after_fork()` is to be called as a callback when the process is forked (e.g. by a demoniser - this was the case with celery), here:
```
https://github.com/aiidateam/aiida-core/blob/0644e7d9ff492fdb2c94d608c575c697daee6d32/aiida/backends/sqlalchemy/__init__.py#L78
```
By the way, I realise now that probably:
1. We should dispose of the passed engine, and not the global one
2. When switching backend, probably we should deregister the callback if it was registered

sphuber:
Closing as it is a sub problem required for fixing #2759

PR #2814, opened 2019-04-30
Properly unload database environment when swapping profiles

### LINKED ISSUE #2810: Unittests for `verdi quicksetup/setup` use default and not test backend
The `--db-backend` is not specifically specified, so the default `django` is used, even when running the tests for `sqlalchemy`

### LINKED ISSUE #2811: Database connections and session need to be properly "swapped"
The `temporary_config_instance` testing context manager will temporarily create a new AiiDA instance, complete with configuration folder file etc. When it is opened the current configuration and profile globals are stored as they will be overwritten, such that they can be restored at the end of the scope of the context manager. However, it is just swapping global variables and certain resources are not correctly reset/recreated. An example is the session for SqlAlchemy. The test session will be replaced by a session for the duration of the context manager, however the original session is never recreated.

sphuber:
Fixes #2810 and fixes #2811 

This tackles two problems that were brought to light by the changes made to bring JSONB to Django and use Aldjemy to dynamically generate SqlAlchemy models for the QueryBuilder. Essentially we discovered that when a profile is switched, the database environment is not properly reset. This experimental functionality was added recently and is only being used to make running unittests for `verdi quicksetup/setup` within the normal test framework possible. To do so, a new AiiDA instance has to be generated as well as swapping the profile from the generic test session one, to one that is created on the fly by the setup commands. We realized that the incorrect backend was used for the commands, which is handled by the first commit. But fixing this revealed the more serious problem. When swapping profiles, existing database connections are not properly reset, causing the supposedly isolated test to use the wrong (i.e. global test session connection). This is now fixed for SqlAlchemy, but the required functionality of `unload_dbenv` is not yet implemented for Django. This will be taken care of in issue #2813 .

### DIFF
--- a/aiida/backends/djsite/utils.py
+++ b/aiida/backends/djsite/utils.py
@@ -7,10 +7,8 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-"""
-This modules contains a number of utility functions specific to the
-Django backend.
-"""
+# pylint: disable=no-name-in-module, no-member, import-error
+"""Utility functions specific to the Django backend."""
 from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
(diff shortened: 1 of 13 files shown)


szoupanos:
Thanks a lot Seb!

PR #2818, opened 2019-05-01
Fix calling show_mpl() on a BandsData object

borellim:
We were getting an error from matplotlib while calling show_mpl() on a BandsData object. The line:
```
from __future__ import print_statement
```
in the `matplotlib_header_template` string seems incorrect, so we changed it to `print_function` and now it works (although to be honest I don't fully understand the context).


borellim:
Tagging @d-tomerini

giovannipizzi:
See also #2838 that is doing something similar (but there the line is being removed)

@yakutovicha - I think that unless there is a reason we should keep it, but with this fix?

PR #2820, opened 2019-05-01
Automate export file migration for `verdi import`

### LINKED ISSUE #2739: verdi import should automatically migrate export files
currently, when trying to import an export file that was generated with an older format, I get the error message
```
Critical: /.../SSSP_efficiency_pseudos.aiida cannot be imported: Export file version is 0.3, can import only version 0.4
Use 'verdi export migrate' to update this export file.
```

From a user perspective, most of the time you simply want the data imported, so I propose that, by default, this happens automatically (with migration in memory).
We can add an option to disable this automatic migration for people who don't like it for some reason (large export file perhaps, memory issues).

CasperWA:
Fixes #2739.

Invoke `verdi export migrate` in `verdi import`, if it turns out the export file archive cannot be imported due to an `IncompatibleArchiveVersionError`.

The default is to automatically migrate it (via a temporary file).

Introduce the `--non-interactive` flag for `verdi import`.
This will avoid printing confirm messages.

Add hidden `--manual-migration` flag to turn off automatic migration.
This is hidden, because it should only be used for tests, in order to switch the default to _not_ do automatic migration. This should also make sure that an error code != 0 is still present when automatic migration is not performed on old export files.

### DIFF
--- a/aiida/backends/tests/cmdline/commands/test_import.py
+++ b/aiida/backends/tests/cmdline/commands/test_import.py
@@ -97,9 +97,10 @@ class TestVerdiImport(AiidaTestCase):
         self.assertIn('Comment mode: overwrite', result.output)
         self.assertEqual(result.exit_code, 0, result.output)
 
+    @unittest.skip("Reenable when issue #2426 has been solved (migrate exported files from 0.3 to 0.4)")
     def test_import_old_local_archives(self):
         """ Test import of old local archives
-        Expected behavior: Return message in terminal and error code != 0
+        Expected behavior: Automatically migrate to newest version and import correctly.
         """
         archives = [('export/migrate/export_v0.1.aiida', '0.1'), ('export/migrate/export_v0.2.aiida', '0.2'),
                     ('export/migrate/export_v0.3.aiida', '0.3')]
@@ -108,25 +109,32 @@ class TestVerdiImport(AiidaTestCase):
             options = [get_archive_file(archive)]
             result = self.cli_runner.invoke(cmd_import.cmd_import, options)
 
-            self.assertIsNotNone(result.exception, result.output)
-            self.assertNotEqual(result.exit_code, 0, result.output)
-            self.assertIn(version, result.output, result.exception)
+            self.assertIsNone(result.exception, msg=result.output)
+            self.assertEqual(result.exit_code, 0, msg=result.output)
+            self.assertIn(version, result.output, msg=result.exception)
+            self.assertIn("Success: imported archive {}".format(options[0]), result.output, msg=result.exception)
 
+    @unittest.skip("Reenable when issue #2426 has been solved (migrate exported files from 0.3 to 0.4)")
     def test_import_old_url_archives(self):
         """ Test import of old URL archives
-        Expected behavior: Return message in terminal and error code != 0
+        Expected behavior: Automatically migrate to newest version and import correctly.
+        TODO: Update 'url' to point at correct commit and file.
+        Now it is pointing to yakutovicha's commit, but when PR #2478 has been merged in aiidateam:develop,
+        url should be updated to point to the, essentially same, commit, but in aiidateam.
+        Furthermore, the filename should be changed from '_no_UPF.aiida' to '_simple.aiida'.
         """
-        url = "https://raw.githubusercontent.com/aiidateam/aiida_core/f685844b290ed23df254873df62da7162bde0f1b/"
+        url = "https://raw.githubusercontent.com/yakutovicha/aiida_core/f5fff1846a62051b898f13db67f5eef18892d5f4/"
         archive_path = "aiida/backends/tests/fixtures/export/migrate/"
-        archive = 'export_v0.1.aiida'
-        version = '0.1'
+        archive = 'export_v0.3_no_UPF.aiida'
+        version = '0.3'
 
         options = [url + archive_path + archive]
         result = self.cli_runner.invoke(cmd_import.cmd_import, options)
 
-        self.assertIsNotNone(result.exception, result.output)
-        self.assertNotEqual(result.exit_code, 0, result.output)
-        self.assertIn(version, result.output, result.exception)
+        self.assertIsNone(result.exception, msg=result.output)
+        self.assertEqual(result.exit_code, 0, msg=result.output)
+        self.assertIn(version, result.output, msg=result.exception)
+        self.assertIn("Success: imported archive {}".format(options[0]), result.output, msg=result.exception)
 
     @unittest.skip("Reenable when issue #2426 has been solved (migrate exported files from 0.3 to 0.4)")
     def test_import_url_and_local_archives(self):
@@ -169,3 +177,59 @@ class TestVerdiImport(AiidaTestCase):
 
         error_message = 'It may be neither a valid path nor a valid URL.'
         self.assertIn(error_message, result.output, result.exception)
+
+    @unittest.skip("Reenable when issue #2426 has been solved (migrate exported files from 0.3 to 0.4)")
+    def test_non_interactive_and_migration(self):
+        """Test options `--non-interactive` and `--migration`/`--no-migration`
+        `migration` = True (default), `non_interactive` = False (default), Expected: Query user, migrate
+        `migration` = True (default), `non_interactive` = True, Expected: No query, migrate
+        `migration` = False, `non_interactive` = False (default), Expected: No query, no migrate
+        `migration` = False, `non_interactive` = True, Expected: No query, no migrate
+        """
+        archive = get_archive_file('export/migrate/export_v0.3.aiida')
+        confirm_message = "Do you want to try and migrate {} to the newest export file version?".format(archive)
+        success_message = "Success: imported archive {}".format(archive)
+
+        # Import "normally", but explicitly specifying `--migration`, make sure confirm message is present
+        # `migration` = True (default), `non_interactive` = False (default), Expected: Query user, migrate
+        options = ['--migration', archive]
+        result = self.cli_runner.invoke(cmd_import.cmd_import, options)
+
+        self.assertIsNone(result.exception, msg=result.output)
+        self.assertEqual(result.exit_code, 0, msg=result.output)
+
+        self.assertIn(confirm_message, result.output, msg=result.exception)
+        self.assertIn(success_message, result.output, msg=result.exception)
+
+        # Import using non-interactive, make sure confirm message has gone
+        # `migration` = True (default), `non_interactive` = True, Expected: No query, migrate
+        options = ['--non-interactive', archive]
+        result = self.cli_runner.invoke(cmd_import.cmd_import, options)
+
+        self.assertIsNone(result.exception, msg=result.output)
+        self.assertEqual(result.exit_code, 0, msg=result.output)
+
+        self.assertNotIn(confirm_message, result.output, msg=result.exception)
+        self.assertIn(success_message, result.output, msg=result.exception)
+
+        # Import using `--no-migration`, make sure confirm message has gone
+        # `migration` = False, `non_interactive` = False (default), Expected: No query, no migrate
+        options = ['--no-migration', archive]
+        result = self.cli_runner.invoke(cmd_import.cmd_import, options)
+
+        self.assertIsNotNone(result.exception, msg=result.output)
+        self.assertNotEqual(result.exit_code, 0, msg=result.output)
+
+        self.assertNotIn(confirm_message, result.output, msg=result.exception)
+        self.assertNotIn(success_message, result.output, msg=result.exception)
+
+        # Import using `--no-migration` and `--non-interactive`, make sure confirm message has gone
+        # `migration` = False, `non_interactive` = True, Expected: No query, no migrate
+        options = ['--no-migration', '--non-interactive', archive]
+        result = self.cli_runner.invoke(cmd_import.cmd_import, options)
+
+        self.assertIsNotNone(result.exception, msg=result.output)
+        self.assertNotEqual(result.exit_code, 0, msg=result.output)
+
+        self.assertNotIn(confirm_message, result.output, msg=result.exception)
+        self.assertNotIn(success_message, result.output, msg=result.exception)
--- a/aiida/cmdline/commands/cmd_import.py
+++ b/aiida/cmdline/commands/cmd_import.py
@@ -8,15 +8,16 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """`verdi import` command."""
-# pylint: disable=too-many-locals
+# pylint: disable=broad-except,too-many-arguments,too-many-locals,too-many-branches
 from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
 from enum import Enum
+import traceback
 import click
 
 from aiida.cmdline.commands.cmd_verdi import verdi
-from aiida.cmdline.params.options import MultipleValueOption
+from aiida.cmdline.params import options
 from aiida.cmdline.params.types import GroupParamType, ImportPath
 from aiida.cmdline.utils import decorators, echo
 from aiida.common import exceptions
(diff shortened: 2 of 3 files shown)


aiida/backends/tests/cmdline/commands/test_import.py
```diff
@@ -108,25 +109,64 @@ def test_import_old_local_archives(self):
...
 
-        self.assertIsNotNone(result.exception, result.output)
-        self.assertNotEqual(result.exit_code, 0, result.output)
-        self.assertIn(version, result.output, result.exception)
+        self.assertIsNone(result.exception, msg=result.output)
+        self.assertEqual(result.exit_code, 0, msg=result.output)
+        self.assertIn(version, result.output, msg=result.exception)
+        self.assertIn("{} after migration".format(options[0]), result.output, msg=result.exception)
+
+    def test_import_old_url_archive_no_migrate(self):
```
ltalirz: since you've tested the migrate/no-migrate already before, this test is not strictly necessary

aiida/cmdline/commands/cmd_import.py
```diff
@@ -78,8 +78,13 @@ class ExtrasImportCode(Enum):
     help="Specify the way to import Comments with identical UUIDs: "
     "newest: Only the newest Comments (based on mtime) (default)."
     "overwrite: Replace existing Comments with those from the import file.")
+@click.option(
+    '--manual-migration', is_flag=True, hidden=True, default=True, help="Do not by default do automatic migration.")
```
ltalirz: I'd suggest `--no-migration`.

Also: `default=True`? Why does this even work?
I would also be fine with unhiding this option, since it provides a functionality that is otherwise not available.

aiida/cmdline/commands/cmd_import.py
```diff
@@ -108,7 +116,8 @@ def cmd_import(archives, webpages, group, extras_mode_existing, extras_mode_new,
             except Exception:
                 echo.echo_error('an exception occurred while trying to discover archives at URL {}'.format(webpage))
                 echo.echo(traceback.format_exc())
-                click.confirm('do you want to continue?', abort=True)
+                if not non_interactive:
```
ltalirz: well spotted!

aiida/cmdline/commands/cmd_import.py
```diff
@@ -127,15 +136,64 @@ def cmd_import(archives, webpages, group, extras_mode_existing, extras_mode_new,
                 extras_mode_existing=ExtrasImportCode[extras_mode_existing].value,
                 extras_mode_new=extras_mode_new,
                 comment_mode=comment_mode)
-        except exceptions.IncompatibleArchiveVersionError as exception:
-            echo.echo_critical('{} cannot be imported: {}'.format(archive, exception))
+        except exceptions.IncompatibleArchiveVersionError:
+            echo.echo_warning('{} cannot be imported.'.format(archive))
```
ltalirz: Let's change this now to a more explanative message about the version conflict.

aiida/cmdline/commands/cmd_import.py
```diff
@@ -127,15 +136,64 @@ def cmd_import(archives, webpages, group, extras_mode_existing, extras_mode_new,
...
+            if non_interactive:
+                migrate_archive = manual_migration
```
ltalirz: I would have `no_migration=False` by default and make it the top-level logic.
If it is `True`, no need to even ask.
(you can print one line as the reason why no migration was performed)

aiida/cmdline/commands/cmd_import.py
```diff
@@ -157,14 +215,64 @@ def cmd_import(archives, webpages, group, extras_mode_existing, extras_mode_new,
...
                 echo.echo(traceback.format_exc())
-                click.confirm('do you want to continue?', abort=True)
+                if not non_interactive:
+                    click.confirm('do you want to continue?', abort=True)
             else:
                 echo.echo_success('imported archive {}'.format(archive))
+
+            # Migrate archive if needed and desired
+            if migrate_archive:
+                migrate_archive = False
```
ltalirz: if beer then: no beer ;-)

I guess this line is anyhow not needed...

aiida/cmdline/commands/cmd_import.py
```diff
@@ -157,14 +215,64 @@ def cmd_import(archives, webpages, group, extras_mode_existing, extras_mode_new,
...
+                        'Download the archive file and run `verdi export migrate` to update it.\n{}'.format(
+                        archive, exception)
+                    echo.echo_critical(crit_message)
+                except Exception:
+                    echo.echo_error('an exception occurred while importing the archive {}'.format(archive))
+                    echo.echo(traceback.format_exc())
+                    if not non_interactive:
+                        click.confirm('do you want to continue?', abort=True)
+                else:
+                    echo.echo_success('imported archive {} after migration'.format(archive))
```
ltalirz: will it also go here, if no migration was needed?

aiida/cmdline/commands/cmd_import.py
CasperWA: > I'd suggest `--no-migration` instead.

Yes, but this makes it a bit cryptic to use in the code - but maybe it will be okay?

> Also: `default=True`? Why does this not break the tests?

Yeah, I wondered this myself.

> I would also be fine with unhiding this option, since it provides a functionality that is otherwise not available.

Personally, I think it should be hidden, since under normal circumstances, invoking `verdi import` from the terminal will not do anything - it will wait for the user to say "yes" or "no" to doing the migration. This is only for those cases, where you want to invoke `verdi import` from a Python function, and don't want it to automatically migrate your export files - for one reason or the other.

aiida/cmdline/commands/cmd_import.py
CasperWA: It can be either a version conflict of the export file or of AiiDA.

Maybe "{} cannot be imported due to a version conflict" - and then we do not mention what the version belongs to? :)

aiida/cmdline/commands/cmd_import.py
CasperWA: True. If one specifically specified `--manual-migration` / `--no-migration`, then there shouldn't be a prompt.

aiida/cmdline/commands/cmd_import.py
CasperWA: Yeah, it looks weird. But if my analysis of the for-loop and everything is correct, I believe it is needed.

We can have multiple archives, where one may have to be migrated (setting `migrate_archive = True`). Then a later archive, in the same for-loop, will not have to be migrated, `migrate_archive` is still `True`, so it will try to migrate it and fail.

aiida/cmdline/commands/cmd_import.py
CasperWA: It shouldn't, since all this is under the `if migrate_archive:`

aiida/cmdline/commands/cmd_import.py
ltalirz: What I meant is: it *can* be imported, right? E.g. replace the message with something like:

`'{}' has export version '{}' while AiiDA v'{}' uses export version '{}'`

aiida/cmdline/commands/cmd_import.py
ltalirz: Well, the question about hiding or not just depends on whether you can imagine a user who might want to use this.
E.g. somebody setting up ansible scripts who wants to make sure things are not migrated without notice.
Anyhow, I see no harm in showing that this option exists.

aiida/cmdline/commands/cmd_import.py
ltalirz: Ok, then rather put `migrate_archive=False` at the beginning of the loop

aiida/cmdline/commands/cmd_import.py
CasperWA: Okay. I surrender :) And will also be more strict in forcing this option through in the function as suggested by you in another comment.

aiida/cmdline/commands/cmd_import.py
CasperWA: At that point in time it cannot. But I understand what you mean.

I'd rather not have to retrieve the version from the file, since this requires a bit too much code simply for an error message. But it could be something like: `{} has an old export version, while AiiDA v{} uses export version {}'. That will be easy. However, it may still not be a sufficient error message.

What I mean above is that if an `IncompatibleArchiveVersionError` is raised it can mean two different things:

1. the export version of the archive is older than the current export version of the installed version of AiiDA.
2. the export version of the archive is newer than the currect export version of the installed version of AiiDA.

For the 1st point the archive should be migrated, for the 2nd one needs to install a newer version of AiiDA to import it.

aiida/cmdline/commands/cmd_import.py
CasperWA: True! Thanks.

aiida/backends/tests/cmdline/commands/test_import.py
CasperWA: True. It shall be tossed aside.

aiida/cmdline/commands/cmd_import.py
sphuber: I wrote an `Archive` class that can be used as a context manager to very simply get access to some of the metadata. I used it in `verdi export inspect` and you can take inspiration from there. To get the version all you need something like this:
```
from aiida.common.archive import Archive

with Archive(archive) as archive_object:
        echo.echo(archive_object.version_format)
```

aiida/cmdline/commands/cmd_import.py
sphuber: I take it that you guys want the automatic migration to be the default behavior. Shouldn't a flag `--no-auto-migrate` then suffice? No need to make it hidden either. In scripts you can then specify `--non-interactive --no-auto-migrate`. I am trying to see if we really need the double flag and *if* we really need it, is it not better to keep the `--something/--no-something` dichotomy, i.e. `--auto-migrate/--no-auto-migrate`?

CasperWA: > I take it that you guys want the automatic migration to be the default behavior.

Yes, see reasons and discussion conclusions [here](https://github.com/aiidateam/aiida_core/issues/2739#issuecomment-484109917).

>  ... is it not better to keep the `--something/--no-something` dichotomy, i.e. `--auto-migrate/--no-auto-migrate`?

I thought of this to begin with as well. But `--non-interactive` has other uses as well. There are several "Do you want to continue?" queries when an `Exception` is caught - for _any_ reason.
So I feel it is fine to have both, but they are definitely overlapping a bit, when it comes to the "Do you want to migrate?" query.

aiida/cmdline/commands/cmd_import.py
CasperWA: Thanks @sphuber. Will take a look at it.
If I am using this anyway, I might as well provide the user with "intelligent" error messages.

aiida/backends/tests/cmdline/commands/test_import.py
CasperWA: Done.

aiida/cmdline/commands/cmd_import.py
sphuber: What I meant with the double flag was not combination of non-interactive and the auto-migrate, but rather why `--no-auto-migrate/--auto-migrate` if we can simply have `--no-auto-migrate` only and by default do auto migration?

ltalirz: I see no harm in having both the on- and off-switch.
I don't think it is really needed at the moment but it's one more line of code (or not even) and it could be used to explicitly force one behavior, even if the aiida developers at some point decide to change the default behavior.

I prefer `--no-migrate` over `--no-auto-migrate` since it's shorter and simpler (i.e. should I migrate or not?) but I'm fine with both.

aiida/cmdline/commands/cmd_import.py
CasperWA: Merci, Monsieur :)

aiida/cmdline/commands/cmd_import.py
CasperWA: I will vote for `--no-migrate` as well, for the simplicity. And meaning is not lost, I think. 

But okay, I will make an on/off-switch: `--migration` / `--no-migration`

CasperWA: Finally, do we agree on the following behaviour?:

| `migration` | `non_interactive` | Result CLI | Result script |
| ------------- | ---------------------- | ------------- | ---------------- |
| `True`* | `False`* | Query user, migrate | Migrate |
| `True`* | `True` | No query, migrate | Migrate |
| `False` | `False`* | No query, no migrate | No migrate |
| `False` | `True` | No query, no migrate | No migrate |

*Default

I do not know if there is a way to check whether a flag (`--migrate`) was specifically requested or not - since I could then implement that no query was performed, if it can be detected that it _was_ requested.

aiida/cmdline/commands/cmd_import.py
CasperWA: Done.

aiida/cmdline/commands/cmd_import.py
CasperWA: I have realized I can simply use the exception message thrown by `importexport.py`. However, to make it "intelligent" I would have to edit the exception message.
What is better: Edit "external" exception message or implement the same logic used "externally" to create the, essentially same, exception message "locally"?

aiida/cmdline/commands/cmd_import.py
CasperWA: This discussion is continued in another review comment. Will resolve this one to simplify.

aiida/cmdline/commands/cmd_import.py
ltalirz: I'm not quite sure what the difference between "result cli" and "result script" is here...
anyhow, the "result cli" is correct.

>  since I could then implement that no query was performed, if it can be detected that it was requested.

I think it's fine displaying a "no migration was performed" message whenever that is correct (whether a migration was requested or not). No need for more complex logic.

ltalirz: i.e. if you literally mean "putting the same bash line into a script vs typing it on the terminal" - the behavior will be the same

CasperWA: The difference was between calling `verdi import` in a terminal and calling it from, e.g. ansible.

aiida/cmdline/commands/cmd_import.py
CasperWA: I have gone the way of the less involved: edit "external" exception message. Basically, I am just keeping the first line that mentions the export versions.

aiida/cmdline/commands/cmd_import.py
CasperWA: Great, done.

aiida/cmdline/commands/cmd_import.py
ltalirz: Ok, I see now that there is a lot of code duplication for url vs local archives, which was perhaps partially already there before but is now growing even more because of the migration.

Can we essentially get rid of that? There isn't really much difference in the two code paths, right?

CasperWA: I agree with this sentiment. I can try. The main difference is when to use the SandboxFolder.
This will however have to wait until tomorrow at the earliest, probably not until beginning of next week.

CasperWA: Done, but not yet completely tested.

CasperWA: Done.

aiida/cmdline/commands/cmd_import.py
```diff
@@ -108,36 +212,54 @@ def cmd_import(archives, webpages, group, extras_mode_existing, extras_mode_new,
...
+        import_opts['archive'] = archive
+        import_opts['file_to_import'] = import_opts['archive']
+
+        # First attempt to import archive
+        migrate_archive = _try_import(first=True, **import_opts)
+
+        # Migrate archive if needed and desired
+        if migrate_archive:
+
+            echo.echo_info("migrating archive {}".format(archive))
```
ltalirz: let's move that inside `_migrate_archive`

aiida/cmdline/commands/cmd_import.py
```diff
@@ -36,13 +37,108 @@ class ExtrasImportCode(Enum):
...
+        if not kwargs['non_interactive']:
+            click.confirm('do you want to continue?', abort=True)
+    else:
+        echo.echo_success('imported archive {}'.format(kwargs['archive']))
+
+    return migrate_archive
+
+
+def _migrate_archive(ctx, temp_folder, **kwargs):
+    """Utility functiuon for `verdi import` to migrate archive"""
```
ltalirz: a little more info on the behavior of this function would be great

aiida/cmdline/commands/cmd_import.py
```diff
@@ -36,13 +37,108 @@ class ExtrasImportCode(Enum):
...
+    from aiida.cmdline.commands.cmd_export import migrate
+
+    # Checks
+    expected_keys = ['file_to_import', 'archive', 'non_interactive']
```
ltalirz: why not simply make those keys explicit arguments?
that saves you from this check. you can still have "kwargs" to capture the rest

aiida/cmdline/commands/cmd_import.py
```diff
@@ -36,13 +37,108 @@ class ExtrasImportCode(Enum):
...
+        echo.echo(traceback.format_exc())
+        if not kwargs['non_interactive']:
+            click.confirm('do you want to continue?', abort=True)
+    else:
+        echo.echo_success("archive migrated, proceeding with import")
+
+    kwargs['file_to_import'] = temp_folder.get_abs_path(temp_out_file)
+
+    # Try again to import archive
+    _try_import(first=False, **kwargs)
```
ltalirz: to me, this does not belong in `_migrate_archive` but should be called explicitly.

aiida/cmdline/commands/cmd_import.py
```diff
@@ -36,13 +37,108 @@ class ExtrasImportCode(Enum):
     ask = 'kca'
 
 
+def _try_import(first, **kwargs):
+    """
+    Utility function for `verdi import` to try-except to import archive
+
+    :param first: Boolean to determine what exception message to throw for IncompatibleArchiveVerisonError
```
ltalirz: it is not obvious to me why this is called `first`.
Could it be given a more meaningful name (or a description that makes this name clear?).

aiida/cmdline/commands/cmd_import.py
CasperWA: Done

aiida/cmdline/commands/cmd_import.py
CasperWA: Done

CasperWA: Also added more info for `_try_import`-function.

aiida/cmdline/commands/cmd_import.py
CasperWA: For `_migrate_archive` it makes sense to make these few keys explicit. In `_try_import` there are so many, I thought it easier simply to do `**kwargs` for all of them. But sure, some of them, as in the keyword inputs to `import_data()` may simply be `**kwargs`. However, this is only three, making five explicit keyword inputs to `_try_import`, but maybe that is fine?

aiida/cmdline/commands/cmd_import.py
CasperWA: Okay, I agree now you bring it up. I will remove it out and put it in the click command.

aiida/cmdline/commands/cmd_import.py
CasperWA: `first_try` or `first_run` or `migrated_import` or `already_migrated`? :)

ltalirz: I'd vote for `migration_performed`

CasperWA: Done

aiida/cmdline/commands/cmd_import.py
CasperWA: Done

PR #2829, opened 2019-05-02
Remove usage of `aiida.cmdline.utils.echo` from non CLI code

### LINKED ISSUE #2819: Remove use of `echo_critical` inside the main code
`echo_critical` calls sys.exit, so it should only be called from within command-line code and should not be present in the main ORM code (or elsewhere in AiiDA code).

I spotted one occurrence:
https://github.com/aiidateam/aiida_core/blob/16f105ecd820eb04358ac67c4c954cd0db93ad36/aiida/orm/groups.py#L70

but we should probably check if it's also used elsewhere (and see, when replacing, in which parts of the cmdline this was used - probably the cmdline is now assuming that `echo_critical` is being called, while instead it should catch an exception and call `echo_critical` in the cmdline code instead

sphuber:
Fixes #2819 

The `echo` functions should only be used for code that is directly
called in command line interface utilities. Internal code should simply
communicate by throwing exceptions.

### DIFF
--- a/aiida/tools/visualization/graphviz.py
+++ b/aiida/tools/visualization/graphviz.py
@@ -202,8 +202,9 @@ def draw_graph(origin_node,
     try:
         exit_code = subprocess.call(['dot', '-T', image_format, fname, '-o', output_file_name])
     except OSError:
-        from aiida.cmdline.utils import echo
-        echo.echo_critical('Operating system error - perhaps Graphviz is not installed?')
+        raise OSError('call to `dot` failed: perhaps graphviz is not installed?')
+
     # cleaning up by removing the temporary file
     os.remove(fname)
+
     return exit_code, output_file_name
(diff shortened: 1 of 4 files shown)


aiida/tools/visualization/graphviz.py
```diff
@@ -202,8 +202,7 @@ def draw_link_settings(inp_id, out_id, link_label, link_type):
     try:
         exit_code = subprocess.call(['dot', '-T', image_format, fname, '-o', output_file_name])
     except OSError:
-        from aiida.cmdline.utils import echo
-        echo.echo_critical('Operating system error - perhaps Graphviz is not installed?')
+        raise OSError('Operating system error - perhaps graphviz is not installed?')
```
giovannipizzi: Maybe remove "Operating system error - " that seems a blue screen of death problem? :-)

sphuber: done

PR #2840, opened 2019-05-06
Changes required for the REST API to work with the Materials Cloud Ex…

asle85:
…plore section

Fix REST API to work correctly with the develop branch of Materials Cloud Explore section


aiida/restapi/translator/calculation/__init__.py
```diff
@@ -91,10 +92,11 @@ def __init__(self, **kwargs):
         }
 
     @staticmethod
-    def get_files_list(dirobj, files=None, prefix=None):
+    def get_files_list(nobj, dirobj=None, files=None, prefix=None):
```
ltalirz:
```suggestion
    def get_files_list(node_obj, dir_obj=None, files=None, prefix=None):
```
the `n` prefix usually stands for a number, `node_obj` makes the content more clear

aiida/restapi/translator/calculation/__init__.py
```diff
@@ -91,10 +92,11 @@ def __init__(self, **kwargs):
...
         """
         Get list of all files present in given directory.
```
ltalirz: Can you explain a bit here the purpose of the function, i.e. should one provide a node object *or* a directory object?

aiida/restapi/translator/node.py
```diff
@@ -587,9 +587,10 @@ def get_node_shape(ntype):
             nodetype = main_node.node_type
             nodelabel = main_node.label
             display_type = nodetype.split('.')[-2]
-            description = main_node.get_description()
-            if description == '':
-                description = main_node.node_type.split('.')[-2]
+            try:
+                description = main_node.get_description()
+            except ValueError:
+                description = display_type
```
ltalirz: there seems to be a few places where this is duplicated.
can this be encapsulated in one function that is used in these three places?

aiida/backends/tests/test_restapi.py
```diff
@@ -69,14 +73,37 @@ def setUpClass(cls, *args, **kwargs):
...
         calc.set_option('resources', resources)
         calc.set_attribute("attr1", "OK")
         calc.set_attribute("attr2", "OK")
 
         calc.add_incoming(structure, link_type=LinkType.INPUT_CALC, link_label='link_structure')
         calc.add_incoming(parameter1, link_type=LinkType.INPUT_CALC, link_label='link_parameter')
+
+        aiida_in = 'The input file\nof the CalcJob node'
+        # Add the file to the CalcJobNode folder calcjob_inputs
+        tmpdir_in = tempfile.mkdtemp()
```
sphuber: If you are only writing a single file, it is perhaps nicer to use the `put_object_from_filelike` that was designed for such use cases. You can then do something like:
```
with tempfile.TemporaryFile() as handle:
     handle.write()
     handle.flush()
     handle.seek(0)
     calc.put_object_from_filelike(handle, key='calcjob_inputs', force=True)
```
This way you do not need to create a whole directory.

aiida/restapi/translator/calculation/__init__.py
```diff
@@ -104,12 +106,19 @@ def get_files_list(dirobj, files=None, prefix=None):
...
-        for fname in dirobj.get_content_list():
-            if dirobj.isfile(fname):
+
+        if dirobj:
+            flist = nobj.list_objects(dirobj)
+        else:
+            flist = nobj.list_objects()
+
+        for fname, ftype in flist:
+            print(fname, ftype)
```
sphuber: left-over print statement

aiida/restapi/translator/calculation/__init__.py
```diff
@@ -28,12 +29,12 @@ class CalculationTranslator(NodeTranslator):
     # A label associated to the present class (coincides with the resource name)
     __label__ = "calculations"
     # The AiiDA class one-to-one associated to the present class
-    from aiida.orm import CalcJobNode
-    _aiida_class = CalcJobNode
+    from aiida.orm import CalculationNode
+    _aiida_class = CalculationNode
     # The string name of the AiiDA class
-    _aiida_type = "process.calculation.calcjob.CalcJobNode"
+    _aiida_type = "process.calculation.calculation.CalculationNode"
```
sphuber: these hard coded strings are all code smells and most likely should be replaced with something else. I remember the other PR that is still open that removes the `_qb_type` but the same goes for `_aiida_type`. It is ok to not address this in this PR, but for future reference that this should most likely be refactored

aiida/restapi/translator/calculation/__init__.py
asle85: I agree, it's more clear using `node_obj` and `dir_obj`. I changed it, thanks!

aiida/restapi/translator/node.py
asle85: done!

aiida/restapi/translator/calculation/__init__.py
asle85: I modified the docstring to:

        Return the list of all files contained in the node object repository
        If a directory object `dir_obj` of the repository is passed, get the list of all files
        recursively in the specified directory

aiida/backends/tests/test_restapi.py
asle85: done, thanks!

aiida/restapi/translator/calculation/__init__.py
asle85: thanks!

aiida/restapi/translator/calculation/__init__.py
asle85: Ok, I'm mentioning @waychal so she also reads this comment.

aiida/restapi/translator/node.py
asle85: I was checking how to move also the rest of the code inside a function and I noted that in 
```
            nodes.append({
                "id": node_count,
                "nodeid": pk,
                "nodeuuid": uuid,
                "nodetype": nodetype,
                "nodelabel": nodelabel,
                "displaytype": display_type,
                "group": "main_node",
                "description": description,
                "shape": get_node_shape(nodetype)
            })
```

not only  `group` and `node_count` values differ, but also in the 'duplicated' chunks there are two additional fields `linklabel` and `linktype`. Do you think that it still makes sense to put everything into a function?

ltalirz: it's fine like this

aiida/restapi/translator/calculation/__init__.py
ltalirz: thanks!

PR #2841, opened 2019-05-06
Update dependency requirement `plumpy==0.14.0`

### LINKED ISSUE #2839: Add dependency to plumpy 0.14 when released
See this issue for why: https://github.com/aiidateam/plumpy/issues/102

sphuber:
Fixes #2839 

This version fixes two bugs in process port validation.

### DIFF
--- a/aiida/backends/tests/engine/test_process_spec.py
+++ b/aiida/backends/tests/engine/test_process_spec.py
@@ -35,19 +35,19 @@ class TestProcessSpec(AiidaTestCase):
         """Test a process spec with dynamic input enabled."""
         node = Node()
         data = Data()
-        self.assertIsNotNone(self.spec.validate_inputs({'key': 'foo'}))
-        self.assertIsNotNone(self.spec.validate_inputs({'key': 5}))
-        self.assertIsNotNone(self.spec.validate_inputs({'key': node}))
-        self.assertIsNone(self.spec.validate_inputs({'key': data}))
+        self.assertIsNotNone(self.spec.inputs.validate({'key': 'foo'}))
+        self.assertIsNotNone(self.spec.inputs.validate({'key': 5}))
+        self.assertIsNotNone(self.spec.inputs.validate({'key': node}))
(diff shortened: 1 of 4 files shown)


borellim:
Thanks Sebastiaan!

Does the change in the plumpy API reflect into a change in the aiida-core API?

sphuber:
The change in `plumpy` API was not directly part of `aiida-core`'s API so users should remain unaffected

sphuber:
> Thanks, seems to work.

Did you install `plumpy==0.14.0` and tried to run the `PwBandStructureWorkChain` without the work around?

borellim:
> > Thanks, seems to work.
> 
> Did you install `plumpy==0.14.0` and tried to run the `PwBandStructureWorkChain` without the work around?

Yes.

However, I think it makes sense to wait for the next version of aiida-core before removing the workaround (https://github.com/aiidateam/aiida-quantumespresso/issues/276), so that I can make aiida-qe depend on aiida-core 1.0.0b3 and avoid any problems. Alternatively I could add a dependency on plumpy 0.14, but it's less clean. Do you agree?

borellim:
To be clear: I did remove the workaround when I tested locally.

borellim:
@sphuber, I just noticed that the validator function is still called twice, although with the correct argument both times now. Is this expected?

sphuber:
Excellent. I think we can just remove it since we will have to update the version of `aiida-core` anyway for the tutorial VM. At least you can remove it and open the PR. We can then see when we merge it

borellim:
@sphuber, I just noticed that the validator function is still called twice, although with the correct argument both times now. Is this expected?

sphuber:
> @sphuber, I just noticed that the validator function is still called twice, although with the correct argument both times now. Is this expected?

Yes this is expected. The first time is when the `Port` itself is created and the specified default value is actually validated, calling the validator. The second time is when the process is instantiated and the actual inputs are validated against the process spec.

PR #2846, opened 2019-05-07
Fix verdi quicksetup on ubuntu 

### LINKED ISSUE #2836: quicksetup fails on ubuntu 18.04 with postgresql 10.7
The reason for this is the following:

On ubuntu 18.04 with the default postgresql 10.7, when logged in as the `postgres` user this command asks for a password:
```psql -d template1 -h localhost -p 5432 -tc '\q'```
while this command doesn't
```psql -d template1 -p 5432 -tc '\q'```

The docs say:
```
  -h, --host=HOSTNAME      database server host or socket directory (default: "/var/run/postgresql")
```

It turns out that on ubuntu 18.04, both the local IPv4 and IPv6 connections are configured to require a password, while the connection over sockets doesn't (on MacOS 10.14 with postgresql 9.6 both work without password).

The naive fix would be not to pass the `-h localhost` flag to psql if the host is detected to be `localhost` (in which case it will just work out of the box).
However, this might break cases where the postgresql server is running inside docker and you *want* to pass `localhost` (which is not an invented use case, see e.g. #2588).

I.e. we'll probably need to change the default from `localhost` to `None` (which doesn't seem to work for psycopg2, though, so here we need again some logic...). I'll look into this.

Edit: It's even a bit more complicated: 
In Ubuntu, the connection via sockets has authentication method 'peer' which means database users also have to match the current unix user.
For authentication via IPv4/v6, the method is password-based.

So: For Ubuntu we *need* `host='localhost'` for psycopg2, while we *cannot* specify a host for psql.
(to check whether in MacOS both works).

### PR DESCRIPTION
**Pull request by ltalirz**
fix #2836 
prepare #2837  

On the default postgres configuration for ubuntu 18.04, both the local IPv4 and IPv6 connections are configured to require a password, while the connection over sockets doesn't (on MacOS 10.14 with postgresql 9.6 both work without password).
We were passing the `-h localhost` flag, which switched to IPv4/IPv6 and thus required the password of the `postgres` user (in addition to the sudo password). We are now no longer doing this when going via the `psql` shell.
Note: When going via `psycopg2`,  you *do* need `host='localhost'` on Ubuntu. `pgsu` is now handling this correctly.

This issue demonstrates how dependent the `verdi quicksetup` functionality is on the details of the postgres setup of the OS. This PR therefore also extracts the core of the `aiida.manage.external.postgres.Postgres` class into into a separate class in the `pgsu.py` module (PGSU for postgres superuser), which only has the purpose to connect to a given DB as the postgres superuser and execute SQL commands. Once the API has converged, this should enable extracting pgsu from aiida-core into a separate package and test this package for a number of OS/postgres-setup combinations.
Work in this has started in https://github.com/ltalirz/pgsu

Further changes:
 * add three new command line options to `verdi quicksetup` in order to allow to configure the `superuser`: 
   - `--su-db-username`
   - `--su-db-name`
   - `--su-db-password`

To decide
 - [ ] whether to change the `DEFAULT_DBINFO` to some explicitly constant construct
 - [ ] tests for the new superuser options?

### DIFF
--- a/aiida/cmdline/commands/cmd_status.py
+++ b/aiida/cmdline/commands/cmd_status.py
@@ -82,10 +82,9 @@ def verdi_status():
         with Capturing(capture_stderr=True):
 
             postgres = Postgres.from_profile(profile)
-            pg_connected = postgres.try_connect()
 
-        dbinfo = postgres.get_dbinfo()
-        if pg_connected:
+        dbinfo = postgres.dbinfo.copy()
+        if postgres.is_connected:
             print_status(ServiceStatus.UP, 'postgres', "Connected to {}@{}:{}".format(
                 dbinfo['user'], dbinfo['host'], dbinfo['port']))
         else:
--- a/aiida/manage/external/pgsu.py
+++ b/aiida/manage/external/pgsu.py
@@ -1,0 +1,348 @@
+# -*- coding: utf-8 -*-
+"""Connect to an existing PostgreSQL cluster as the `postgres` superuser and execute SQL commands.
+
+Note: Once the API of this functionality has converged, this module should be moved out of aiida-core and into a
+  separate package that can then be tested on multiple OS / postgres setups. Therefore, **please keep this
+  module entirely AiiDA-agnostic**.
+"""
+from __future__ import division
+from __future__ import print_function
+from __future__ import absolute_import
+
+try:
+    import subprocess32 as subprocess
+except ImportError:
+    import subprocess
+
+from enum import IntEnum
+import click
+
+DEFAULT_DBINFO = {
+    'host': 'localhost',
+    'port': 5432,
+    'user': 'postgres',
+    'password': None,
+    'database': 'template1',
+}
+
+
+class PostgresConnectionMode(IntEnum):
+    """Describe mode of connecting to postgres."""
+
+    DISCONNECTED = 0
+    PSYCOPG = 1
+    PSQL = 2
+
+
+class PGSU(object):  # pylint: disable=useless-object-inheritance
+    """
+    Connect to an existing PostgreSQL cluster as the `postgres` superuser and execute SQL commands.
+
+    Tries to use psycopg2 with a fallback to psql subcommands (using ``sudo su`` to run as postgres user).
+
+    Simple Example::
+
+        postgres = PGSU()
+        postgres.execute("CREATE USER testuser PASSWORD 'testpw'")
+
+    Complex Example::
+
+        postgres = PGSU(interactive=True, dbinfo={'port': 5433})
+        postgres.execute("CREATE USER testuser PASSWORD 'testpw'")
+
+    Note: In postgresql
+     * you cannot drop databases you are currently connected to
+     * 'template0' is the unmodifiable template database (which you cannot connect to)
+     * 'template1' is the modifiable template database (which you can connect to)
+    """
+
+    def __init__(self, interactive=False, quiet=True, dbinfo=None, determine_setup=True):
+        """Store postgres connection info.
+
+        :param interactive: use True for verdi commands
+        :param quiet: use False to show warnings/exceptions
+        :param dbinfo: psycopg dictionary containing keys like 'host', 'user', 'port', 'database'
+        :param determine_setup: Whether to determine setup upon instantiation.
+            You may set this to False and use the 'determine_setup()' method instead.
+        """
+        self.interactive = interactive
+        self.quiet = quiet
+        self.connection_mode = PostgresConnectionMode.DISCONNECTED
+
+        self.setup_fail_callback = prompt_db_info if interactive else None
+        self.setup_fail_counter = 0
+        self.setup_max_tries = 1
+
+        self.dbinfo = DEFAULT_DBINFO.copy()
+        if dbinfo is not None:
+            self.dbinfo.update(dbinfo)
+
+        if determine_setup:
+            self.determine_setup()
+
+    def execute(self, command, **kwargs):
+        """Execute postgres command using determined connection mode.
+
+        :param command: A psql command line as a str
+        :param kwargs: will be forwarded to _execute_... function
+        """
+        # Use self.dbinfo as default kwargs, update with provided kwargs
+        kw_copy = self.dbinfo.copy()
+        kw_copy.update(kwargs)
+
+        if self.connection_mode == PostgresConnectionMode.PSYCOPG:  # pylint: disable=no-else-return
+            return _execute_psyco(command, **kw_copy)
+        elif self.connection_mode == PostgresConnectionMode.PSQL:
+            return _execute_psql(command, **kw_copy)
+
+        raise ValueError('Could not connect to postgres.')
+
+    def set_setup_fail_callback(self, callback):
+        """
+        Set a callback to be called when setup cannot be determined automatically
+
+        :param callback: a callable with signature ``callback(interactive, dbinfo)``
+          that returns a ``dbinfo`` dictionary.
+        """
+        self.setup_fail_callback = callback
+
+    def determine_setup(self):
+        """Determine how to connect as the postgres superuser.
+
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

**Comment by ltalirz** [comment]
@sphuber I had one strange pre-commit error on travis: https://travis-ci.org/ltalirz/aiida_core/jobs/529221856#L1388

Do you happen to know where this is coming from?

**Comment by ltalirz** [comment]
I think this PR is the last thing I need to prepare the next iteration of the tutorial VMs.

**Comment by sphuber** [comment]
> Do you happen to know where this is coming from?

Tracking down cyclic import errors is notoriously difficult with pylint because it does not always correctly point to the source. I have never seen it this wrong however. It seems that you have not touched these files at all. I wouldn't know where this would come from. Maybe rebase your branch and pray that they will disappear :laughing: Does the same warning show up locally?

**Comment by ltalirz** [comment]
@sphuber I've rebased - unfortunately this did not fix the issue.

**Comment by ltalirz** [comment]
@sphuber  just got a notification for your passing build: https://travis-ci.org/sphuber/aiida_core/builds/529298825?utm_medium=notification&utm_source=email

what did you do?

**Comment by sphuber** [comment]
> @sphuber just got a notification for your passing build: https://travis-ci.org/sphuber/aiida_core/builds/529298825?utm_medium=notification&utm_source=email
> 
> what did you do?

I tracked down the cyclic import to `aiida.engine.processes.process.Process.on_entered` and simplified the import and added the pylint exclusion. I can force push to your branch if you want

**Comment by ltalirz** [comment]
You'll have to self-review ;-)

File: aiida/cmdline/commands/cmd_status.py [hunk 1]
```diff
@@ -82,10 +82,9 @@ def verdi_status():
         with Capturing(capture_stderr=True):
 
             postgres = Postgres.from_profile(profile)
-            pg_connected = postgres.try_connect()
 
-        dbinfo = postgres.get_dbinfo()
-        if pg_connected:
+        dbinfo = postgres.dbinfo.copy()
```
sphuber: Why do we need a copy here? Will the `postgres.is_connected` call modify it?

aiida/manage/external/pgsu.py
```diff
@@ -0,0 +1,348 @@
...
+        """Print a warning message and calls the failed setup callback
+
+        :returns: False, if no successful try.
+        """
+        message = '\n'.join([
+            'Warning: Unable to autodetect postgres setup - do you know how to access it?',
+        ])
+
+        if not self.quiet:
+            click.echo(message)
```
sphuber: Maybe use `echo_warning` instead here

aiida/manage/external/pgsu.py
```diff
@@ -0,0 +1,348 @@
...
+    if database:
+        option_str += '-d {}'.format(database)
+    # to do: Forward password to psql; ignore host only when the password is None.  # pylint: disable=fixme
+    kwargs.pop('password', None)
+
+    host = kwargs.pop('host', 'localhost')
+    if host and host != 'localhost':
+        option_str += ' -h {}'.format(host)
+    elif not quiet:
+        click.echo("Warning: Found host 'localhost' but dropping '-h localhost' option for psql " +
```
sphuber: consider replacing with `echo_warning`

aiida/cmdline/commands/cmd_status.py
ltalirz: It was just the safe replacement. It's not needed here.

aiida/manage/external/pgsu.py
ltalirz: As mentioned at the top of the file, the idea is to move this out of aiida-core eventually so it can be easily tested.

ltalirz:
Thanks for testing this. 
```
[sudo] password for sphuber: 
Success: created new profile `quicksetup`.
Info: migrating the database.
Critical: database migration failed, probably because connection details are incorrect:
FATAL:  password authentication failed for user "aiida_qs_sphuber_eb52c82774454f8a69b8c2a1d9b0d059"
FATAL:  password authentication failed for user "aiida_qs_sphuber_eb52c82774454f8a69b8c2a1d9b0d059"
```
That sucks - fixing it for one Ubuntu breaks it for another... this is exactly why we need CI on different OSs for this.
Anyhow, could you please let me know:

 * what it wrote in the profile configuration (host, port, db and user)?
 * whether the DB & user specified there have actually been created?

Thanks!

sphuber:
The profile it created:
```
aiidadb_backend         django
aiidadb_engine          postgresql_psycopg2
aiidadb_host            localhost
aiidadb_name            quicksetup2_sphuber_eb52c82774454f8a69b8c2a1d9b0d059
aiidadb_pass            8t4QSYoEk8fe0yyh
aiidadb_port            5432
aiidadb_repository_uri  file:///home/sphuber/.virtualenvs/aiida_dev/.aiida/repository/quicksetup2
aiidadb_user            aiida_qs_sphuber_eb52c82774454f8a69b8c2a1d9b0d059
profile_uuid            a18eb39da04a4767810c685b554eb063
```
The database and the database user *were* created.
It seems there is something wrong with the credentials because also manually logging in fails:
```
psql -U aiida_qs_sphuber_eb52c82774454f8a69b8c2a1d9b0d059 -h localhost -p 5432 -d quicksetup2_sphuber_eb52c82774454f8a69b8c2a1d9b0d059 -W
Password for user aiida_qs_sphuber_eb52c82774454f8a69b8c2a1d9b0d059: 
psql: FATAL:  password authentication failed for user "aiida_qs_sphuber_eb52c82774454f8a69b8c2a1d9b0d059"
FATAL:  password authentication failed for user "aiida_qs_sphuber_eb52c82774454f8a69b8c2a1d9b0d059"
```
Because if I change the password in psql `ALTER ROLE aiida_qs_sphuber_eb52c82774454f8a69b8c2a1d9b0d059 WITH PASSWORD 'test';`
Then I can login just fine. So I think the setting of the password during the creation of the role does not happen correctly

aiida/manage/external/pgsu.py
sphuber: Ah of course, missed that, thanks

sphuber:
And as an addendum: we are not sure if these changes "broke" it on Ubuntu 16.04 because quicksetup never worked for me. It always just hung after having passed my sudo password.

ltalirz:
Thanks for reporting - so the user is created correctly but the password is not set correctly.

The name of the DB user is based on the current system user and the path to the config directory. I.e. it should actually be *the same* for all your profiles that were created this way (by the way, I wonder whether we want to keep this behavior or whether we prefer to createwant one user per profile).

Now, if such a user already exists, then quicksetup will not touch it and instead look in the config.json to figure out the user's password.
If what it finds there is incorrect (it should just take the first occurence of the username), this could explain the behavior you saw.

@sphuber Could you please check whether you have the same problems with a vanilla config folder?

Things to think about:
  * If this turns out to be the source of the problem, we should probably open an issue for this case - we could at least to do a connection check inside quicksetup and provide a helpful error message if this happens
  * By default, do we want to keep one user per config file path or do we rather want one DB user per profile?
    Having one user per profile would make it easier down the line to share access to only one AiiDA profile. This would be a very simple change in `get_quicksetup_username`

sphuber:
I think you were right and the problem is that it tried to use an existing role. I think the role already existed but had been created with another password algorithm, so the one that was deduced now was wrong. After I deleted all quicksetup roles and databases and rerun `verdi quicksetup` now works.

Regarding whether to use a separate role for each profile: I think it is not necessary. A separate database makes sense, but I don't see why we would need a separate role for each profile. Unless that makes the implementation a lot easier, I would just keep a single quicksetup role per AiiDA instance

sphuber:
Given that it seems to work, I will update the branch and merge it ok? We can then address the other problems in your comment in separate issues

ltalirz:
Good, let's merge.

Indeed the suggestion above can be implemented separately.

PR #2865, opened 2019-05-13
swap order of columns in verdi process list

ltalirz:
to have "State" and "Process status" side by side


ltalirz:
@sphuber I didn't see where it gets the headers of `verdi process list` from... does this come somehow from deep down in the database?

sphuber:
@ltalirz you can find the mapping of projection key to header label here: https://github.com/aiidateam/aiida_core/blob/6d2e2dbdc95b1c4b759e6f486b10af829e84d330/aiida/cmdline/utils/query/mapping.py#L100

ltalirz:
@sphuber thanks, done!

PR #2867, opened 2019-05-13
Clean up `submit_test` folder in `CalcJob` dry-run tests

### LINKED ISSUE #2866: Folder 'submit_test' is created in work dir when running tests
Folder should be deleted after tests finish, but it is not.

sphuber:
Fixes #2866 

The launcher tests running with `dry_run=True` will create the
`submit_test` folder in the current working directory that was not being
cleaned up. The test is adapted to delete this folder in the `tearDown`
of the unit test. Note that his is not the most elegant solution as it
depends on the `submit_test` name not changing, which is currently not
directly configurable and it not being another folder in the current
working directory. However, since it is not configurable and the cleanup
*has* to be done in a `tearDown`, because otherwise if the test fails
the cleanup might be skipped if placed in the test itself, there is
currently no better solution.


giovannipizzi:
Is there an easy way to make the test(s) fail if the name of the folder (`submit_test`) changes? Or can we make the folder name at least be a constant and reuse the value here?

sphuber:
I made the `submit_test` value a module constant and use that. That should already make it a lot more robust even though it is still not the cleanest

sphuber:
> I guess there is a reason why the `submit_test` folder cannot be used as any other SandboxFolder is used in other tests, where it is automatically removed after usage?

Yes, because in normal usage, the whole point is that even after the code is executed, the folder remains for the user inspect manually

CasperWA:
@giovannipizzi if you are satisfied with the recent changes by @sphuber, I will merge this?

giovannipizzi:
@CasperWA please go ahead!

PR #2873, opened 2019-05-13
Make `verdi import` -G (group) option function as intended.

### LINKED ISSUE #2863: Exception in `verdi import` when specifying a group
The problem is that the command passes the `Group` instance to `import_data` but in [this line in `import_data_sqla`](https://github.com/aiidateam/aiida_core/blob/6d2e2dbdc95b1c4b759e6f486b10af829e84d330/aiida/orm/importexport.py#L1891) the variable is treated as a list.

As a side-note: the `verdi import` command should use the pre-defined `options.GROUP` overridable option instead of defining its own

CasperWA:
Fixes #2863.

Set `verdi import`'s `-G` (group) option to the general `options.GROUP()`, but allowing to create new `Group` if needed.
Only a `Group` label can be passed.
The handling of 'user_group' in `importexport.py` has been optimized.
Tests have been added for `verdi import` and for `importexport.py`.

**Note**: If one has created a `Group` to be used for the purpose of placing the imported `Node`s in, and has worked with it, but _not_ stored it and subsequently passes _its label_ to `import_data()`, the work will be overwritten. ~~Since the method `Group.objects.get_or_create(label="whatever_label")` is used.~~ However, if the `Group` object itself is passed to `import_data()`, the work will _not_ be overwritten. Instead, the `Group` will be stored and used.

**Edit**: The method `Group.objects.get_or_create(label="whatever_label")` is not used anymore, since there is need to specify a specific `type_string` if a new `Group` is created. However, the "quirk" described above is still valid.

### DIFF
--- a/aiida/backends/tests/test_export_and_import.py
+++ b/aiida/backends/tests/test_export_and_import.py
@@ -665,6 +665,47 @@ class TestGroups(AiidaTestCase):
         builder.append(orm.Group, filters={'label': {'like': grouplabel + '%'}})
         self.assertEqual(builder.count(), 2)
 
+    @with_temp_dir
+    def test_import_to_group(self, temp_dir):
+        """Test `group` parameter
+        Make sure an unstored Group is stored by the import function, forwarding the Group object.
+        Make sure the Group is correctly handled and used for imported nodes.
+        """
+        from aiida.orm import load_group
+
+        # Create Nodes to export
+        data1 = orm.Data().store()
+        data2 = orm.Data().store()
+        node_uuids = [node.uuid for node in [data1, data2]]
+
+        # Export Nodes
+        filename = os.path.join(temp_dir, "export.aiida")
+        export([data1, data2], outfile=filename, silent=True)
+        self.reset_database()
+
+        # Create Group, do not store
+        group_label = "import_madness"
+        group = orm.Group(label=group_label)
+        group_uuid = group.uuid
+
+        # Try to import to this Group, providing only label - this should fail
+        with self.assertRaises(TypeError, msg="Labels should no longer be passable to `import_data`") as exc:
+            import_data(filename, group=group_label, silent=True)
+        exc = exc.exception
+        self.assertEqual(str(exc), "group must be a Group entity",
+            msg="The error message should be the same for both backends.")
+
+        # Import properly now, providing the Group object
+        import_data(filename, group=group, silent=True)
+
+        # Check Group for content
+        builder = orm.QueryBuilder().append(orm.Group, filters={'label': group_label}, project='uuid')
+        self.assertEqual(builder.count(), 1, msg="There should be exactly one Group with label {}. "
+                                                 "Instead {} was found.".format(group_label, builder.count()))
+        imported_group = load_group(builder.all()[0][0])
+        self.assertEqual(imported_group.uuid, group_uuid)
+        for node in imported_group.nodes:
+            self.assertIn(node.uuid, node_uuids)
 
 class TestCalculations(AiidaTestCase):
     """Test ex-/import cases related to Calculations"""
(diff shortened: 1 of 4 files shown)


aiida/orm/importexport.py
```diff
@@ -1162,12 +1168,27 @@ def import_data_dj(in_path, user_group=None, ignore_unknown_nodes=False,
             if pks_for_group:
                 # If user specified a group, import all things in it
                 if user_group:
-                    group = user_group[0]
+                    if isinstance(user_group, str):
+                        # Will try to retrieve Group with provided label
+                        res_group = Group.objects.find(label=user_group)
```
sphuber: This pattern should be replaced with `Group.objects.get_or_create`

aiida/orm/importexport.py
```diff
@@ -583,6 +588,7 @@ def import_data_dj(in_path, user_group=None, ignore_unknown_nodes=False,
     detect the compression format (zip, tar.gz, tar.bz2, ...) and calls the
     correct function.
     :param in_path: the path to a file or folder that can be imported in AiiDA
+    :param user_group: Group wherein all imported Nodes will be placed.
```
sphuber: Here it is called `user_group` but in the generic `import_data` it is just `group`. I think the latter is better so let's keep that

aiida/orm/importexport.py
```diff
@@ -1162,12 +1168,27 @@ def import_data_dj(in_path, user_group=None, ignore_unknown_nodes=False,
             if pks_for_group:
                 # If user specified a group, import all things in it
                 if user_group:
-                    group = user_group[0]
+                    if isinstance(user_group, str):
```
sphuber: I feel all this logic is way too complicated for what it should be doing. Wouldn't it be good enough to just accept only a `Group` instance? If one really wants to import everything in a group, creating or loading it first is not that difficult. It would allow to cleanup the following logic mess a lot

aiida/orm/importexport.py
```diff
@@ -1888,12 +1909,28 @@ def import_data_sqla(in_path, user_group=None, ignore_unknown_nodes=False,
...
+                        else:
+                            # Create new import Group with provided label
+                            group = Group(label=user_group, type_string=IMPORTGROUP_TYPE)
+                            session.add(group.backend_entity._dbmodel)
+                    elif isinstance(user_group, Group):
+                        # Use provided Group
+                        group = user_group
+
+                        # Make sure provided Group is stored
+                        if not group.is_stored:
```
sphuber: Shouldn't it be stored here then? (Note I think that all of this logic should be removed anyway and just accept an existing group in the arguments)

aiida/orm/importexport.py
CasperWA: done

aiida/orm/importexport.py
CasperWA: done

aiida/orm/importexport.py
CasperWA: Have fixed it so that labels can no longer be passed - only Groups

CasperWA: Have fixed it so that labels can no longer be passed - only Groups

aiida/orm/importexport.py
CasperWA: It should - Will remove this as well

aiida/backends/tests/test_export_and_import.py
```diff
@@ -665,6 +665,67 @@ def test_group_import_existing(self, temp_dir):
         builder.append(orm.Group, filters={'label': {'like': grouplabel + '%'}})
         self.assertEqual(builder.count(), 2)
 
+    @with_temp_dir
+    def import_to_group_using_label(self, temp_dir):
```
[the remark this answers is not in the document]
CasperWA: Agree, on both accounts.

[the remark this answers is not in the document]
CasperWA: However, it is weird that this test doesn't fail. I will make one that should fail when a label is given.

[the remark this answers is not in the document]
CasperWA: Never mind, I found the mistake (hint: `test_`)

sphuber:
Thanks a lot @CasperWA

PR #2876, opened 2019-05-15
Respect option `-p/--profile` in `verdi profile show`

### LINKED ISSUE #2875: Explicit profile in `verdi -p <name> profile show` not respected
It will always show the currently default profile

sphuber:
Fixes #2875 

Even when a non-default option was specified with `verdi profile show`
the default specified in the configuration was taken. The order of
precedence for profile in verdi commands should be:

 * Configuration default
 * Profile option `-p/--profile`
 * Command specific profile parameter

By setting an attribute `profile_option_used` in the `ctx.obj` of the
`verdi` command, other commands can see whether the user specified an
explicit profile that should potentially be respected.

### DIFF
--- a/aiida/cmdline/utils/defaults.py
+++ b/aiida/cmdline/utils/defaults.py
@@ -22,13 +22,20 @@ def get_default_profile(ctx, param, value):  # pylint: disable=unused-argument
 
     This should be used if the default profile should be returned lazily, at a point for example when the config
     is not created at import time. Otherwise, the preference should go to calling `get_config` to load the actual
-    config and using `config.default_profile_name` to get the default profile name
+    config and using `config.default_profile_name` to get the default profile name.
+
+    This all of course unless the `-p/--profile` option was used by the user which trumps everything, in which case the
+    `profile_option_used` attribute on the `ctx.obj` object will have been set and we simply return the profile that
+    was already loaded by `verdi` itself and set in `ctx.obj.profile`.
 
     :raises click.UsageError: if the config could not be loaded or no default profile exists
     :return: the default profile
     """
     if value:
         return value
+
+    if ctx.obj.profile_option_used:
+        return ctx.obj.profile
 
     try:
         config = get_config()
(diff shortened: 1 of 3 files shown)


sphuber:
@ltalirz note this fix is important for the `verdi quicksetup` section I am writing for the tutorial

docs/source/verdi/verdi_user_guide.rst
```diff
@@ -367,6 +367,7 @@ Below is a list with all available subcommands.
       structure   Manipulation of StructureData objects.
       trajectory  View and manipulate TrajectoryData instances.
       upf         Manipulation of the upf families.
+      diff        Command line interface for aiida-diff
```
ltalirz: this is the danger of the automatic docs creation.
is there any way of avoiding this? conrad and me also already ran into this

aiida/cmdline/utils/defaults.py
```diff
@@ -22,14 +22,21 @@ def get_default_profile(ctx, param, value):  # pylint: disable=unused-argument
...
+    `profile_option_used` attribute on the `ctx.obj` object will have been set and we simply return the profile that
+    was already loaded by `verdi` itself and set in `ctx.obj.profile`.
 
     :raises click.UsageError: if the config could not be loaded or no default profile exists
     :return: the default profile
     """
     if value:
         return value
 
+    if ctx.obj.profile_option_used and ctx.obj.profile:
```
ltalirz: so what if `not ctx.obj.profile`? should this be allowed?
if not, do we even need to test this here?

docs/source/verdi/verdi_user_guide.rst
sphuber: Good catch. It sure is annoying, but still the amount of things we would have missed without auto doc generation would have been far greater. I will see if I can exclude external plugins from this. Should be possible

ltalirz: > the amount of things we would have missed without auto doc generation would have been far greater

fully agree

sphuber: Quite an ad-hoc solution, but still acceptable I would say in PR #2878

PR #2878, opened 2019-05-15
Implement an isolated mode for the `Pluginable` sub of `click.Group`

### LINKED ISSUE #2877: Auto-docs generator `verdi` includes plugin entry points
The `verdi data` command supports plugins to add entry points, which when installed, get erroneously picked up by the `utils.validate_consistency` pre-commit script that will auto-generate the documentation for `verdi`. There needs to be a way to disable external entry points in `verdi` for the purpose of this documentation generation.

sphuber:
Fixes #2877 

The `Pluginable` sub class of `click.Group` allows sub commands to be
added to it dynamically from external plugins through entry points.
However, this gave problems for the auto-docs generator in the
pre-commit script `utils/validate_consistency.py` which would include
these external commands, even though it shouldn't. To solve this, we
implement the isolation mode for the `Pluginable` class which the
pre-commit script will enable before generating the documentation.

### DIFF
--- a/aiida/cmdline/utils/pluginable.py
+++ b/aiida/cmdline/utils/pluginable.py
@@ -22,20 +22,38 @@ class Pluginable(click.Group):
 
     def __init__(self, *args, **kwargs):
         """Initialize with entry point group."""
+        self._exclude_external_plugins = False  # Default behavior is of course to include external plugins
         self._entry_point_group = kwargs.pop('entry_point_group')
         super(Pluginable, self).__init__(*args, **kwargs)
 
     def list_commands(self, ctx):
         """Add entry point names of available plugins to the command list."""
         subcommands = super(Pluginable, self).list_commands(ctx)
-        subcommands.extend(get_entry_point_names(self._entry_point_group))
+
+        if not self._exclude_external_plugins:
+            subcommands.extend(get_entry_point_names(self._entry_point_group))
+
         return subcommands
 
     def get_command(self, ctx, name):  # pylint: disable=arguments-differ
         """Try to load a subcommand from entry points, else defer to super."""
         command = None
-        try:
-            command = load_entry_point(self._entry_point_group, name)
-        except exceptions.EntryPointError:
+        if not self._exclude_external_plugins:
+            try:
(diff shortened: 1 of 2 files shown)


aiida/cmdline/utils/pluginable.py
```diff
@@ -22,20 +22,36 @@ class Pluginable(click.Group):
 
     def __init__(self, *args, **kwargs):
         """Initialize with entry point group."""
+        self._isolated_mode = False
```
ltalirz: Perhaps add a comment like
`# see set_isolated_mode() for what this variable does`

aiida/cmdline/utils/pluginable.py
```diff
@@ -22,20 +22,36 @@ class Pluginable(click.Group):
...
+        if not self._isolated_mode:
+            try:
+                command = load_entry_point(self._entry_point_group, name)
+            except exceptions.EntryPointError:
+                command = super(Pluginable, self).get_command(ctx, name)
+        else:
             command = super(Pluginable, self).get_command(ctx, name)
         return command
+
+    def set_isolated_mode(self, isolated_mode):
```
ltalirz: what about calling the variable something like `include_external_plugins` instead of `isolated_mode`?

aiida/cmdline/utils/pluginable.py
```diff
@@ -22,20 +22,36 @@ class Pluginable(click.Group):
...
+        """Set the isolated mode.
+
+        If the isolated mode is enable, i.e. set to `True`, the plugins that belong to the `entry_point_group` defined
```
ltalirz:
```suggestion
        Set to `True` to enable isolated mode, preventing the plugins that belong to the `entry_point_group` defined for this  `click.Group` from being discovered. This is useful to limit the available commands to those provided by aiida-core (excluding those provided by plugins).
```

aiida/cmdline/utils/pluginable.py
sphuber: `exclude_external_plugins` would make more sense, no? Since the normal behavior should be to include it. The excluding of them is the exceptional behavior.

PR #2882, opened 2019-05-16
Include only input links in creating a process restart builder

### LINKED ISSUE #2881: `Process.get_builder_restart` should specify to only retrieve incoming links
If not, for work chain nodes, any call links will also be retrieved and will cause an exception

sphuber:
Fixes #2881  

The `Process.get_builder_restart` was populating the builder by getting
all the nodes with incoming links, including call links which of course
are not defined in the spec and so would cause an exception. Simply by
adding a link filter to only get input data nodes, this problem is fixed.


sphuber:
Bug reported by @vdikan should be addressed by this commit

vdikan:
> Bug reported by @vdikan should be addressed by this commit

Confirm, this bug is resolved.

PR #2886, opened 2019-05-18
Documentation: add section on working with calculation jobs

### LINKED ISSUE #2628: Documentation: add section working with calculation job

sphuber:
Fixes #2628 

A section is added explaining how to implement a calculation job and an
accompanying parser.


sphuber:
Note: I have added stubs for all the calculation job options, but I have not yet documented them. However, I think this already inserts the bulk of the documentation necessary to work with calculation jobs and it would be good to have in before the tutorial. We can always add more detail later

PR #2887, opened 2019-05-18
improve installation & 'getting started' docs

ltalirz:
* quick install:
   * fixes suggested by @giovannipizzi
   * add starting daemon
 * install: add starting daemon
 * configuration: move adding verdi to PATH to troubleshooting
 * getting started:
   * move daemon section to install (& shorten)
   * move plugin installation to 1st place (& shorten)
 * developer docs:
   * reorder, putting plugins first (which is what most developers will be interested in)

computer/code setup still untouched (but should definitely be
shortened).


docs/source/get_started/index.rst
```diff
@@ -5,11 +5,55 @@
...
+.. note::
+    What does ``pip install aiida-diff`` do?
+
+    * resolves and installs the dependencies on other python packages
+    * creates a folder ``aiida_diff.egg-info/`` with metadata about the package
+    * if the ``-e`` option is given, creates a symbolic link from the python package
+      search path to the ``aiida-diff`` directory and puts the ``.egg-info``
+      folder there.
+      Changes to the **source code** will be picked up by python without reinstalling, 
+      but changes to the **metadata** will not.
```
sphuber: Maybe add another point saying that for changes will also not be picked up by running interpreters, i.e. daemon needs to be restarted and open shells need to be reopened

docs/source/developer_guide/design/changes.rst
```diff
@@ -1,7 +1,7 @@
-Design change justifications
-============================
+Design evolution
+================
 
-This document records big changes in the design and architecture of AiiDA and especially details the justification for the change
+This document records the evolution of the design and architecture of AiiDA, including the underyling reasoning.
```
sphuber: `underyling` -> `underlying`

docs/source/get_started/index.rst
```diff
@@ -5,11 +5,55 @@
...
+
+.. _plugins:
+
+***************
+Install Plugins
+***************
+
+While the ``aiida-core`` package provides the workflow engine and database model, it relies on *plugins* for connecting to specific simulation codes.
+
+Find your AiiDA plugins on the `AiiDA plugin registry <https://aiidateam.github.io/aiida-registry>`_.
```
sphuber: What if the plugin they need doesn't exist? Should we put a link to the documentation for creating a plugin and the `aiida-plugin-cutter`?

docs/source/install/installation.rst
```diff
@@ -294,144 +294,21 @@ If you open a new terminal for example, be sure to activate it first with::
...
-Alternatively, if you have a ``aiida_core`` repository checked out locally,
-you can just copy the file ``<aiida_core>/aiida/tools/ipython/aiida_magic_register.py`` to the same folder.
-The current ipython profile folder can be located using::
-
-  ipython locate profile
-
-After this, if you open a Jupyter notebook as explained above and type in a cell::
-
-    %aiida
+    While operational, the daemon logs its activity to a file in ``in ~/.aiida/daemon/log/``.
```
sphuber: Should take 'in' out of the inline-literal. Note also that the path here will of course only be correct for default `AIIDA_PATH`

docs/source/get_started/index.rst
```diff
@@ -5,11 +5,55 @@
...
-2) A computer has to be configured
-3) A code has to configured
-4) A plugin has to be installed
+1) :ref:`plugins`
+2) :ref:`setup_computer`
+3) :ref:`setup_code`
+
+.. _plugins:
+
+***************
```
sphuber: why put the `plugins` section in the same file, but use the index to include the ones for setting up a computer and code?

docs/source/install/installation.rst
```diff
@@ -294,144 +294,21 @@ If you open a new terminal for example, be sure to activate it first with::
 At this point, you can choose to read on for additional installation details and configuration options, or you can choose to start using
 AiiDA and go straight to the section :ref:`get started<get_started>`.
```
sphuber: This should be changed, because now the other sections assume that the daemon is started and running. But following this line people might skip it. Either put this line after the start the daemon section or return the start daemon section to one of the points in `get_started`

docs/source/install/installation.rst
```diff
@@ -294,144 +294,21 @@ If you open a new terminal for example, be sure to activate it first with::
...
+.. _start_daemon:
 
-.. _configure_aiida:
+Start the daemon
+================
 
-Configure AiiDA
-===============
+The AiiDA daemon process runs in the background and takes care of
+submitting new calculations, checking their status, retrieving their results once they are finished and storing them in the AiiDA database.
```
sphuber: This is not strictly true, for two reasons:
* it will also run work chains that are submitted
* it is not really only the daemon that can deal with calculation jobs. You can also run a calculation job, in which case a local runner will do all the things listed here.

docs/source/install/installation.rst
```diff
@@ -294,144 +294,21 @@ If you open a new terminal for example, be sure to activate it first with::
 At this point, you can choose to read on for additional installation details and configuration options, or you can choose to start using
 AiiDA and go straight to the section :ref:`get started<get_started>`.
 
+.. _start_daemon:
```
sphuber: Before going here, should we let them check `verdi status` just as in the quick install section to make sure that everything is working. Or better, after starting the daemon, because otherwise that will show a cross

docs/source/get_started/index.rst
ltalirz: thanks, will add some info concerning this in brackets.
alternatively, we could delete the entire note - it's all python /pip basics and I don't really think we should be explaining this in the aiida docs (let me know)

docs/source/get_started/index.rst
ltalirz: good point - it's exactly the right point for a link to the plugin development section

docs/source/install/installation.rst
ltalirz: > Should take 'in' out of the inline-literal.

thanks!

> Note also that the path here will of course only be correct for default AIIDA_PATH

will add info on this.

docs/source/get_started/index.rst
ltalirz: I know it's not consistent with the other sections at this point but now at least one sees some relevant information when clicking on "first things first".
Eventually, when code and computer setup have been reasonably shortened, they could all go in one page.

docs/source/install/installation.rst
ltalirz: > Either put this line after the start the daemon section 

thanks, will do!

docs/source/get_started/index.rst
sphuber: I agree, maybe it would be good to add the extra detail I suggested and then move this whole section to troubleshooting

docs/source/get_started/index.rst
sphuber: fair enough

docs/source/install/installation.rst
ltalirz: thanks - my rewrite is hopefully better

docs/source/install/installation.rst
ltalirz: thanks, added!

docs/source/get_started/index.rst
ltalirz: there currently isn't a troubleshooting section for "get started".
I've left it in for the moment, let's deal with this later.

ltalirz:
@sphuber thanks for the thorough read, suggestions are incorporated!

docs/source/get_started/index.rst
sphuber: No I meant the general troubleshooting section in the installation section, because installing plugins can be viewed as the installation part. But it is okay for now

sphuber:
thanks a lot @ltalirz

PR #2889, opened 2019-05-18
make aiida.cmdline.data entry point visible

ltalirz:
make aiida.cmdline.data entry point visible in

```
verdi plugin list
```


ltalirz:
@sphuber
My impression is that the `entry_point_group_to_module_path_map` is misused at the moment, in the sense that it is both used for mapping entry points to module paths (where this mapping exists) and to display the available entry points in `verdi plugin list`.

I guess these are actually two different sets of entry points - not all entry points have contents that correspond to the internal module hierarchy.

I.e. the string I'm putting for `aiida.cmdline.data` is kind of arbitrary here.
Let me know how to proceed.

Anyhow, the data cli commands now show up properly, together with the ones provided by plugins:

```
$ verdi plugin list aiida.cmdline.data
Registered entry points for aiida.cmdline.data:
* array
* bands
* cif
* dict
* diff
* structure
* trajectory
* upf
```

sphuber:
You are right that the `entry_point_group_to_module_path_map` is being abused. This is an old remnant of the old plugin system, where certain groups had a base path within `aiida-core`, think the old `aiida.orm.calculation.job.plugins`. This should not be necessary anymore. Although I also think that adding the internal data commands to the entry points is necessary. They are already being loaded simply by importing them in `aiida.cmdline.commands.cmd_data.__init__.py`.

ltalirz:
>  Although I also think that adding the internal data commands to the entry points is necessary.

I guess you wanted to write "*not* necessary"?

Perhaps not absolutely necessary. However, we are already adding lots of AiiDA-core functionality as entry points (data types etc.). And it somehow feels wrong to see
```
$ verdi plugin list aiida.cmdline.data
Registered entry points for aiida.cmdline.data:
* diff
```
but then have many other commands available for `verdi data`

PR #2921, opened 2019-05-24
Installation docs: fix inconsistent virtualenv path and a broken ref

### LINKED ISSUE #2920: installation instructions use inconsistent virtualenv paths

borellim:
Fixes #2920


borellim:
I concur; I also prefer to use `~/venvs/` rather than `/.virtualenvs/` as a folder to store virtual environments. But I didn't want to change too much, it was more urgent to make the docs consistent.

Issue #2923, opened 2019-05-24
postgres problem psycopg2.InterfaceError: connection already closed


marco-foscato:
On MacOS
aiida_core 1.0.0.b3
postgres 11.2 installed via Homebrew
python 3.7.3

Error from pytest on aiida-diff after the number of connection (see https://github.com/jamesnunn/pgtest/issues/14):
<pre>self = <django.db.backends.postgresql.base.DatabaseWrapper object at 0x1141e6b38>
autocommit = False

    def _set_autocommit(self, autocommit):
        with self.wrap_database_errors:
>           self.connection.autocommit = autocommit
E           psycopg2.InterfaceError: connection already closed

../aiida_virtualenv/lib/python3.7/site-packages/django/db/backends/postgresql/base.py:236: InterfaceError
</pre>

ltalirz:
@marco-foscato sorry for the long wait - could you check whether adding `determine_setup=False` in the following line fixes the issue?
https://github.com/aiidateam/aiida_core/blob/develop/aiida/manage/fixtures.py#L160

You need to keep the fix we added for pgtest with the number of connections (which I'll now update).

marco-foscato:
Tired, but it raises the <code>ValueError: Could not connect to postgres.</code> from <code>aiida_core/aiida/manage/external/pgsu.py:111</code>. With <code>determine_setup=False</code> the connection_mode of Postgress is never changed from its default DISCONNECTED value (see https://github.com/aiidateam/aiida_core/blob/develop/aiida/manage/external/pgsu.py#L78).

You probably know this already, but without the <code>determine_setup=False</code> the database cluster is created and the aiida-diff test runs at least to some extent. Then, while the <code>aiida_diff/tests/test_calculations.py</code> is hanging, I can still get the psql promt and see the <code>aiida_db</code> database with some aiida nodes:
<pre>aiida_db=# SELECT * FROM db_dbnode;
 id |                 uuid                 |                node_type                 | label | description |             ctime             |             mtime             | nodeversion | public | dbcomputer_id | user_id |      process_type
----+--------------------------------------+------------------------------------------+-------+-------------+-------------------------------+-------------------------------+-------------+--------+---------------+---------+-------------------------
[...]

PR #2924, opened 2019-05-24
Fix false negative of postgres connection in `verdi status`

### LINKED ISSUE #2911: verdi status: reports false negative database connection
`verdi status` reports that the database cannot be connected to, however, this is not the case. All other commands run as expected. The problem is most likely because `Postgres.from_profile(profile)` is broken and does not actually correctly read the DB information from the profile.
Printing the `dbinfo` we get the following:
```
{'host': 'localhost',
 'port': 5432,
 'user': 'postgres',
 'password': None,
 'database': 'template1'}
```
This the default and not the actual profile.

sphuber:
Fixes #2911 

The output of `verdi status` was reporting a problem in connecting to
the postgres database even though all other commands seemed to work. The
problem was with the `Postgres.from_profile` used to perform the check
that did not correctly use the database connection details from the
profile but just the defaults.

### DIFF
--- a/aiida/backends/tests/cmdline/commands/test_status.py
+++ b/aiida/backends/tests/cmdline/commands/test_status.py
@@ -19,7 +19,7 @@ from aiida.backends.tests.utils.configuration import with_temporary_config_insta
 
 
 class TestVerdiStatus(AiidaTestCase):
-    """Tests for `verdi rehash`."""
+    """Tests for `verdi status`."""
 
     def setUp(self):
         self.cli_runner = CliRunner()
@@ -29,11 +29,13 @@ class TestVerdiStatus(AiidaTestCase):
         """Test running verdi status.
 
         Note: The exit status may differ depending on the environment in which the tests are run.
+            Also cannot check for the exit status to see if connecting to all services worked, because
+            the profile might not be properly setup in this temporary config instance unittest.
         """
         options = []
         result = self.cli_runner.invoke(cmd_status.verdi_status, options)
         self.assertIsInstance(result.exception, SystemExit)
-        self.assertIn("profile", result.output)
-        self.assertIn("postgres", result.output)
-        self.assertIn("rabbitmq", result.output)
-        self.assertIn("daemon", result.output)
+        self.assertIn('profile', result.output)
+        self.assertIn('postgres', result.output)
+        self.assertIn('rabbitmq', result.output)
+        self.assertIn('daemon', result.output)
--- a/aiida/cmdline/commands/cmd_status.py
+++ b/aiida/cmdline/commands/cmd_status.py
@@ -13,6 +13,7 @@ from __future__ import print_function
 from __future__ import absolute_import
 
 import sys
+
 from enum import IntEnum
 import click
 
@@ -21,8 +22,7 @@ from ..utils.echo import ExitCode
 
 
 class ServiceStatus(IntEnum):
-    """Describe status of services for 'verdi status' command.
-    """
+    """Describe status of services for 'verdi status' command."""
     UP = 0  # pylint: disable=invalid-name
     ERROR = 1
     DOWN = 2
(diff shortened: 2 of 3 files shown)


ltalirz:
@sphuber thanks a lot for fixing my bugs!

actually, this could possibly also fix #2923 where we noticed that the fixture manager was not able to connect to the database, despite the fact that using the connection parameters produced by `pgtest` by hand worked fine.

sphuber:
Actually, it is now causing the profile delete test to fail. This might have just been silently passing because of the bug that is now fixed, though:
```
In python3, it is not possible to serialize a `numpy.int64` with JSON. This
was causing exceptions when storing nodes that contained these values. To fix
this, we add a specific check in the `clean_value` method, to convert those
types to plain python integers.
```

ltalirz:
I think the issue is that if you want to delete the database, you actually cannot connect with the credentials of the profile but, like for quicksetup, you need to connect with the credentials of the database superuser.
For some reason, those seem to be incorrect here.

ltalirz:
I think  `def create_mock_profile` is missing suitable values for the superuser

aiida/manage/external/postgres.py
```diff
@@ -69,13 +69,15 @@ def from_profile(cls, profile, **kwargs):
         :returns: Postgres instance pre-populated with data from AiiDA profile
         """
         dbinfo = DEFAULT_DBINFO.copy()
-
-        get = profile.dictionary.get
         dbinfo.update(
             dict(
-                host=get('AIIDADB_HOST', DEFAULT_DBINFO['host']),
```
ltalirz: actually, the docstring of the function explains why it is using only host and port

ltalirz: I think the solution to the problem is to switch the way `verdi status` determines whether postgres is working.
The `Postgres` class is meant to connect to a postgres cluster as the superuser.
`verdi status` could simply go via `psycopg2` like `load_dbenv` does

sphuber: Aha ok, gotchya

sphuber:
@ltalirz I have restored the original changes to the `Postgres.from_profile` method and replaced the method to check for database connection in the status command  by trying to get a cursor from the backend.

aiida/cmdline/commands/cmd_status.py
```diff
@@ -74,43 +74,31 @@ def verdi_status():
...
-        print_status(
-            ServiceStatus.ERROR,
-            'postgres',
-            "Error connecting to {}:{}".format(profile.database_hostname, profile.database_port),
-            exception=exc)
+        backend = manager.get_backend()
+        backend.cursor()
+    except Exception:
+        database_data = [profile.database_hostname, profile.database_port]
+        print_status(ServiceStatus.DOWN, 'postgres', "Unable to connect to {}:{}".format(*database_data))
```
ltalirz: since this is now (correctly) using all the info from the profile, the error message can also include the username (like the success message below).
this means you can move `database_data` outside and reuse it in both places.

ltalirz:
```suggestion
        print_status(ServiceStatus.DOWN, 'postgres', "Unable to connect as {}@{}:{}".format(*database_data))
```

aiida/backends/tests/cmdline/commands/test_status.py
```diff
@@ -29,11 +29,13 @@ def test_status_1(self):
...
         """
         options = []
         result = self.cli_runner.invoke(cmd_status.verdi_status, options)
         self.assertIsInstance(result.exception, SystemExit)
-        self.assertIn("profile", result.output)
-        self.assertIn("postgres", result.output)
-        self.assertIn("rabbitmq", result.output)
-        self.assertIn("daemon", result.output)
+        self.assertIn('profile', result.output)
+        self.assertIn('postgres', result.output)
```
ltalirz: in order to prevent regression of this bug, perhaps we should search for `Connected to` as well?

ltalirz: hm, but would not help on travis

ltalirz: scratch that

PR #2933, opened 2019-05-27
Make backend type for `verdi quicksetup --config` test dynamic

### LINKED ISSUE #2931: Unit test for `verdi quicksetup --config` uses hardcoded backend type
It hardcodes `django` as backend type, which will cause an exception when running the tests for an SqlAlchemy test profile

sphuber:
Fixes #2931 

It was hardcoded to `django` causing the SqlAlchemy tests to fail.

### DIFF
--- a/aiida/backends/tests/cmdline/commands/test_setup.py
+++ b/aiida/backends/tests/cmdline/commands/test_setup.py
@@ -83,8 +83,8 @@ profile: testing
 first_name: Leopold
 last_name: Talirz
 institution: EPFL
-backend: django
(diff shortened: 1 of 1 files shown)


sphuber:
They are skipped for django only because of #2813 . They were now failing only for SqlAlchemy and python 3. I have no clue why to be honest. I have an inkling it has to do with the change we made in the way we connect to postgres in the setup commands. [Here is example failure](https://travis-ci.org/sphuber/aiida_core/jobs/537824818) in build of branch that changed something completely unrelated

PR #2936, opened 2019-05-28
Acknowledge process task if corresponding node cannot be loaded

### LINKED ISSUE #2930: Acknowledge process task if corresponding node cannot be loaded
Currently, the `ProcessLauncher._continue` method will raise a `plumpy.TaskRejected` exception if the node of a process task cannot be loaded, either because it does not exist or it cannot be uniquely resolved. However, this will cause RabbitMQ to requeue the task, causing the task to be resent. This will result in the task ping-ponging between RabbitMQ and the daemon workers without end. This situation typically occurs when a user deletes a process node while it has not yet properly terminated and therefore the tasks has not yet been acknowledged. This situation is unrecoverable so the task should simply be acknowledged.

sphuber:
Fixes #2930 

Before this commit the `ProcessLauncher._continue` method was raising a
`plumpy.TaskRejected` exception if the node of a process task cannot be
loaded, either because it does not exist or it cannot be uniquely
resolved. However, this will cause RabbitMQ to requeue the task, causing
the task to be resent. This will result in the task ping-ponging between
RabbitMQ and the daemon workers without end. This situation typically
occurs when a user deletes a process node while it has not yet properly
terminated and therefore the tasks has not yet been acknowledged. This
situation is unrecoverable so the task should simply be acknowledged.

### DIFF
--- a/aiida/manage/external/rmq.py
+++ b/aiida/manage/external/rmq.py
@@ -136,7 +136,6 @@ class ProcessLauncher(plumpy.ProcessLauncher):
         :param nowait: if True don't wait for the process to finish, just return the pid, otherwise wait and
             return the results
         :param tag: the tag of the checkpoint to continue from
-        :raises plumpy.TaskRejected: if the node corresponding to the task cannot be loaded
         """
         from aiida.common import exceptions
         from aiida.engine.exceptions import PastException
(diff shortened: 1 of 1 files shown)


aiida/manage/external/rmq.py
```diff
@@ -145,9 +144,9 @@ def _continue(self, communicator, pid, nowait, tag=None):
 
         try:
             node = load_node(pk=pid)
-        except (exceptions.MultipleObjectsError, exceptions.NotExistent) as exception:
+        except (exceptions.MultipleObjectsError, exceptions.NotExistent):
             LOGGER.exception('Cannot continue process<%d>', pid)
-            raise plumpy.TaskRejected('Cannot continue process: {}'.format(exception))
```
giovannipizzi: Coud you put here, as a comment, a summary of the PR message?

sphuber: Done

PR #2942, opened 2019-05-28
Force tags to zero when using ASE to convert a CIF file.

### LINKED ISSUE #2893: Changed behavior of the kind name when creating `StructureData` from CIF file
When reading a cif file and storing it as an AiiDA node it pads tag from ASE if it is not zero to the kind name.

`structure = get_data_node('structure', ase=structure.get_ase())`

Where structure which `get_ase()` is executed on is of `CifData` type.

Happens for ASE 3.17, but not for 3.12 for instance. It is related to the treatment of the tag in ASE. Seems ASE have changed the behavior of the setting of `tag` when converting cif files.
[test.txt](https://github.com/aiidateam/aiida_core/files/3201850/test.txt)

espenfl:
Fixes #2893 

Somewhere from 3.12 to 3.17 the tag concept was bundled with each Atom object. When reding a CIF file, this is incremented and signifies the atomic species, even though file do not have specific tags embedded. On reading CIF files we thus force the ASE tag to zero for all Atom elements.

Fixing this on the ASE side seems not so easy, ref. issue:
https://gitlab.com/ase/ase/issues/357

I suggest we fix this on our end. If we use ASE to convert from CIF to an `Atoms` object (which we again convert to `StructureData`), we force the `tag` attribute to zero for all elements.


espenfl:
@giovannipizzi I updated according to recent `develop`. Can you please review again. Thanks.

giovannipizzi:
Thanks @espenfl !

Indeed, this now reproduced the old behaviour of ASE for version 3.12.

Just as a comment, for future reference, I'm not sure this is the 'best' way of doing it: in principle, CIF contains information on species.
So if we have As with multiplicity 2, and In with multiplicity 2, the new versions give `[0, 0, 1, 1]` as tags, while the correct thing to do (for the way AiiDA deals with tags) is to restart from 0 for every chemical element, so setting `[0, 0, 0, 0]`.

However, in principle, if we have As with multiplicity 2, and In1 and In2 each with multiplicity 1, the old ASE would return `[0, 0, 0, 0]`, the new `[0, 0, 1, 2]`, and the "correct" thing to do from the AiiDA point of view would be probably `[0, 0, 0, 1]` (or actually keep `[0, 0, 1, 2]` in this case, so they get named In1 and In2, but if the names were In3 and In4, then we should set `[0, 0, 3, 4]`).

However, I'm going to approve this on the following grounds:
- it reproduces the old behavior
- anyway a round trip in ASE (CIF -> ASE -> CIF -> ASE) is lossy: when storing, ASE always stores as P1 symmetry, so it stores `As1, As2, In1, In2` so reloading in ASE in the new version would give all atoms a different tag. This is probably the correct thing to do when looking just at the CIF file (the CIF file is saying that the atoms are not symmetrically equivalent) but it's annoying that storing a proper CIF file with symmetry info and reloading it would create multiple species.

In the future, the correct thing is to use a better CIF writer, and make sure that when reading (ideally with a different parser?) from CIF, the names from CIF are used directly as kind names in AiiDA, rather than going through a convoluted logic involving integer tags ASE.

espenfl:
@giovannipizzi Thanks. Yes, I agree. Somehow we need to determine how to proceed with CIF in the future. It seems most people agree it would be nice if someone could do and maintain an up to date Python CIF parser that followed standard (well, as close to whatever is considered a standard that is), in close collaboration with core CIF people and database containers. Maybe a sensible start would be to join the efforts in pymatgen and ASE and see if the CIF parsers in there can be merged and maintained outside as a separate more lightweight package.

PR #2946, opened 2019-05-28
Drop column `transport_params` from the `Computer` database model 

### LINKED ISSUE #2927: Drop the `transport_params` of the `Computer` model
Metadata that can affect the execution of a calculation job on that computer needs to go in the `metadata` column, which will be included in the hash, once implemented for the computer. All other data that relate to how the connection to the computer should behave should be included in the `AuthInfo` instead (this data should *not* affect the calculation execution, e.g. connection cooldown intervals etc.)

sphuber:
Fixes #2927 

The column was intended to contain customized parameters to be used to
configure the transport of a computer. However, these fit better in the
`AuthInfo`, as these settings can change per user. The design is to keep
the `Computer` as generic as possible as to not be user dependent, such
that its configuration can be easily shared and reused. Generic computer
configuration can go in the `metadata` column. User specific settings
need to go in the `AuthInfo.metadata` instead.


sphuber:
@giovannipizzi note that I built this on top of commits that are waiting review in PR #2944 
The migrations here simply drop the column. Are we sure we don't need a data migration? Do we know if people had been putting data in `transport_params`. And if so, is there any way to deterministically migrate it? Where should it go? `Computer.metadata` or `Authinfo.metadata`, and in the case of the latter, which ones?

giovannipizzi:
From the current code, do I understand correctly that there is no direct way for the user to set these from the CLI? If this is the case, I wouldn't be worried (are the prepend/append_text going into the metadata, right?) If instead these are settable from the command line then we must be more careful.

In the first case, if we really want to be sure, we can check if they are empty, and print them in case on file with warnings similar to when we dropped DbWorkflow etc.

sphuber:
@giovannipizzi you are right, it was not exposed in the CLI and all relevant data was stored in the `metadata`.

PR #2949, opened 2019-05-29
Improve the efficiency of `DbLog` migration for Django 

### LINKED ISSUE #2539: Implementation of DbLog migration is extremely slow
In PR #2507 the migration of the `DbLog` table was adapted to also account for log entries whose corresponding node was deleted. The query that was used to select these is as follows:
```
SELECT COUNT(*) FROM db_dblog
WHERE
       (db_dblog.objname LIKE 'node.%') AND
       (db_dblog.objpk NOT IN (SELECT id FROM db_dbnode))
```
The usage of   `NOT IN` with the entire set of node ids makes this query terribly slow. On theossrv5 on a database with roughly `800,000` nodes and `400,000` entries, the migration was still not finished after 24 hours.

sphuber:
Fixes #2539 

The old implementation was using Django's ORM to get the count and rows
of certain rows in the `DbLog` table. However, this was generating
extremely inefficient queries that made the migration run for days on
big databases. A similar problem was encountered with the original
SqlAlchemy implementation. That was solved in an earlier commit
`80a0aa9117f399a5385a30ba0ce3e2afc7016a4e`, which created a more
efficient query. We now replace the Django ORM solution with the same
custom SQL query as used for SqlAlchemy. This should significantly
improve the efficiency of this migration.

### DIFF
--- a/aiida/backends/djsite/db/migrations/0024_dblog_update.py
+++ b/aiida/backends/djsite/db/migrations/0024_dblog_update.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 
 from __future__ import print_function
 import sys
+from six.moves import zip
 import click
 
 # Remove when https://github.com/PyCQA/pylint/issues/1931 is fixed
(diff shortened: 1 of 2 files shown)


sphuber:
I tested this on the mat2D database and the `DbLog` migrations now runs in roughly 10 minutes. Good enough I would say.

giovannipizzi:
@sphuber just double checking that you also verified that the DbLogs are still there and correct in the mat2D, before merging

sphuber:
I did, but it is hard to say. The unit tests should cover this. I forgot to note the number of rows in the original database for the `DbLog` table (did do this for a bunch of other tables). After migration there are still 220.000 logs remaining and some have been written to the two JSON dumps. But cannot say for certain that everything is there. Open to suggestions

giovannipizzi:
I think it's ok then (maybe just run a few verdi commands to get logs and check the they work). I agree that tests should ideally already cover this

szoupanos:
Thanks Seb! I also believe that if the tests pass and you did a basic check that everything is as it should, it should be OK

PR #2420, opened 2019-01-24
add instructions when verdi import fails

ltalirz:
* when verdi import fails because of an old/new export file version, add a hint on what to do
 * fix issue with pip 19.x
 * fix issue with prospector failing to install
 * fix nasty travis issue with build of pymatgen (built using incompatible numpy version provided by travis)

PR #2952, opened 2019-05-31
docs: add note on increasing work_mem

### LINKED ISSUE #2859: profiling the daemon
In aiida-core 0.12 we noticed that running a few 100 calculations of a particular workflow led to very high CPU usage, almost exclusively spent inside Postgres.

Could you please suggest a way to profile what is going on, i.e. to figure out which queries are causing this and where in the python code they are originating from?

Mentioning @sphuber and @giovannipizzi  for comment
@danieleongari and @kjappelbaum for notice

ltalirz:
fix #2859 

 * add note on increasing work_mem
 * restructure troubleshooting section a bit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ before_install:
 install:
     # Upgrade pip setuptools and wheel to be able to run the next command
     - pip install -U pip==18.1 wheel setuptools coveralls
+    - pip install numpy==1.15.4
     # Install AiiDA with some optional dependencies
     - if [ "$TEST_TYPE" == "docs" ]; then pip install --no-cache-dir . && pip install --no-cache-dir -r docs/requirements_for_rtd.txt; else pip install .[all]; fi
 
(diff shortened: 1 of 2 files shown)


ltalirz:
thanks for the suggestions - I've tried to incorporate them
in order not to overcomplicate things I don't include instructions for tuning the value depending on the size of the table but I think it contains everything needed to get started debugging

ltalirz:
@sphuber do you have any idea why the tests failed? Seems to be exactly the same error as in #2420

It was thought that this could be prevented by adding the
`--no-cache-dir` argument to pip here: #2799
but perhaps that was fortuitous?

I also tried resetting the travis cache...

sphuber:
It does seem to be the same error, and indeed the `--no-cache-dir` did consistently fix the problem last time. Maybe now there is an incompatibility even when installing everything from scratch, i.e. the version of `pymatgen` that we tell to install, will automatically install a version of `numpy` and `scipy` that are incompatible.

ltalirz:
I'm going to merge this now... will hopefully fix this travis problem at some point as well.

PR #2958, opened 2019-06-04
Update dependency `kiwipy==0.5.1` with critical bug patch

### LINKED ISSUE #2957: Verify that tasks in the process queue of RabbitMQ are properly persisted
There has been a report here and there that processes did not continue after rebooting the system. The only thing I can think of is that the tasks that were with RabbitMQ were not properly persisted. This would be an enormous problem, but also hard to believe. Needs to first be reproduced

sphuber:
Fixes #2957 

The persistence of a task message, which represents a process that is to
be executed, is guaranteed by RabbitMQ, but only if the exchange, queue
and the task message itself are configured correctly. Up till now the
exchange and task queue were properly set to be durable, however, the
individual messages were sent with the default delivery mode, which
means non-persistent. This caused tasks to be lost on broker restart.
The patch release 0.5.1 of `kiwipy` fixes this by now always marking
task messages as persistent.

### DIFF
--- a/docs/requirements_for_rtd.txt
+++ b/docs/requirements_for_rtd.txt
@@ -32,7 +32,7 @@ flask-marshmallow==0.9.0
 futures; python_version=='2.7'
 ipython>=4.0,<6.0
 itsdangerous==1.1.0
-kiwipy[rmq]==0.5.0
+kiwipy[rmq]==0.5.1
 marshmallow-sqlalchemy==0.16.0
 mock==2.0.0
 numpy==1.16.1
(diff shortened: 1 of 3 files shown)


sphuber:
Yeah, I saw but didn't merge it, because in the PR message was written:
> It is very likely that the current package version for this feedstock is out of date.
And so I thought something was wrong, despite everything seeming to have passed.
Is that message normal?

ltalirz:
Yes - they mean: the bot noticed it is out of date, so it created a PR to bring it up to date.

ltalirz:
By the way, so if I understand correctly, this big bug would have hit when you restart the rabbitmq service, eg when you restart the computer (but not when the computer is put to sleep, correct?
This information may be helpful to users wondering whether they were affected

sphuber:
In principle it would occur if the RabbitMQ service reboots. If that is not the case when the machine goes to sleep (which I don't think it is) then that should have been fine.

PR #2960, opened 2019-06-04
Always attach all registered outputs in `Parser.parse_from_node`

### LINKED ISSUE #2959: Let `Parser.parse_from_node` attach outputs in case of returned exit code
When a `Parser` is called normally, an exit code can be returned to indicate failure, but outputs can be attached before returning. The `Parser.parse_from_node` utility to easily re-perform the parsing manually, does not have the same behavior on the `CalcFunctionNode` that represents the parsing. If an exit code is returned by the `Parser.parse` call, that is also what is returned from the wrapped `calcfunction` and that is what will be returned to the caller as the result. However, any outputs that may have been attached during `Parser.parse` will not be attached as outputs to the `CalcFunctionNode`.

sphuber:
Fixes #2959 

This was not done when an exit code was returned by `Parser.parse` but
the actually parser could have attached outputs before returning this.
By failing to attach these outputs on the `CalcFunctionNode` that
represents the re-parsing, the caller would receive an empty dictionary
as a result.

### DIFF
--- a/aiida/parsers/parser.py
+++ b/aiida/parsers/parser.py
@@ -119,9 +119,13 @@ class Parser(object):
         It's storing method will also be disabled, making it impossible to store, because storing it afterwards would
         not have the expected effect, as the outputs it produced will not be stored with it.
 
+        This method is useful to test parsing in unit tests where a `CalcJobNode` can be mocked without actually having
+        to run a `CalcJob`. It can also be useful to actually re-perform the parsing of a completed `CalcJob` with a
+        different parser.
+
         :param node: a `CalcJobNode` instance
         :param store_provenance: bool, if True will store the parsing as a `CalcFunctionNode` in the provenance
-        :return: a tuple of the parsed results and the `CalcFunctionNode`
+        :return: a tuple of the parsed results and the `CalcFunctionNode` representing the process of parsing
         """
         parser = cls(node=node)
 
@@ -129,12 +133,27 @@ class Parser(object):
         def parse_calcfunction(**kwargs):
             """A wrapper function that will turn calling the `Parser.parse` method into a `CalcFunctionNode`.
 
+            .. warning:: This implementation of a `calcfunction` uses the `Process.current` to circumvent the limitation
+                of not being able to return both output nodes as well as an exit code. However, since this calculation
+                function is supposed to emulate the parsing of a `CalcJob`, which *does* have that capability, I have
+                to use this method. This method should however not be used in process functions, in other words:
+
+                    Do not try this at home!
+
             :param kwargs: keyword arguments that are passed to `Parser.parse` after it has been constructed
             """
+            from aiida.engine import Process
+
             exit_code = parser.parse(**kwargs)
             outputs = parser.outputs
 
             if exit_code and exit_code.status:
+                # In the case that an exit code was returned, still attach all the registered outputs on the current
+                # process as well, which should represent this `CalcFunctionNode`. Otherwise the caller of the
+                # `parse_from_node` method will get an empty dictionary as a result, despite the `Parser.parse` method
+                # having registered outputs.
+                process = Process.current()
+                process.out_many(outputs)
                 return exit_code
 
             return dict(outputs)
--- a/aiida/backends/tests/parsers/test_parser.py
+++ b/aiida/backends/tests/parsers/test_parser.py
@@ -12,6 +12,8 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
 
+import io
+
 from aiida import orm
 from aiida.backends.testbase import AiidaTestCase
 from aiida.common import LinkType
@@ -85,3 +87,31 @@ class TestParser(AiidaTestCase):
         self.assertEqual(outputs_for_parsing['retrieved'].uuid, retrieved.uuid)
         self.assertIn('output', outputs_for_parsing)
         self.assertEqual(outputs_for_parsing['output'].uuid, output.uuid)
+
+    def test_parse_from_node(self):
+        """Test that the `parse_from_node` returns a tuple of the parsed output nodes and a calculation node.
+
+        The calculation node represents the parsing process
+        """
+        summed = 3
+        output_filename = 'aiida.out'
+
+        # Mock the `CalcJobNode` which should have the `retrieved` folder containing the sum in the outputfile file
+        # This is the value that should be parsed into the `sum` output node
+        node = orm.CalcJobNode(computer=self.computer, process_type=ArithmeticAddCalculation.build_process_type())
+        node.set_option('resources', {'num_machines': 1, 'num_mpiprocs_per_machine': 1})
+        node.set_option('max_wallclock_seconds', 1800)
+        node.set_option('output_filename', output_filename)
+        node.store()
+
+        retrieved = orm.FolderData()
+        retrieved.put_object_from_filelike(io.StringIO(u'{}'.format(summed)), output_filename)
+        retrieved.store()
+        retrieved.add_incoming(node, link_type=LinkType.CREATE, link_label='retrieved')
+
+        result, node = ArithmeticAddParser.parse_from_node(node)
+
+        self.assertIsInstance(result['sum'], orm.Int)
+        self.assertEqual(result['sum'].value, summed)
+        self.assertIsInstance(node, orm.CalcFunctionNode)
+        self.assertEqual(node.exit_status, 0)


aiida/parsers/parser.py
```diff
@@ -131,10 +131,17 @@ def parse_calcfunction(**kwargs):
...
+
             exit_code = parser.parse(**kwargs)
             outputs = parser.outputs
 
             if exit_code and exit_code.status:
+                # In the case that an exit code was returned, still attach all the registered outputs on the current
+                # process as well, which should represent this `CalcFunctionNode`. Otherwise the caller of the
+                # `parse_from_node` method will get an empty dictionary as a result.
+                process = Process.current()
+                process.out_many(outputs)
```
ltalirz: So, why do you need to explicitly attach the outputs here when there's an exit code, but not when there is no exit code?
Wouldn't this suggest that this attachment should be done somewhere else?

sphuber: This is because of the difference in operation of `calcfunctions` and `CalcJobs`. For the latter, during parsing, you can call `self.out` to attach any outputs (just like in a `WorkChain`) to the process node, before returning optionally an exit code. The engine will set the relevant attributes on the node based upon the exit code, but the value returned to whoever called the process (if called with `run` or its variants) is the dictionary of the outputs. For a process function there is no `self`, so you cannot use `self.out` to attach an output. Instead there you return a dictionary of nodes from the function, which the engine will then attach for you. Or you return an exit code, but you cannot return both at the moment, so it's either-or.

This `Parser.parse_from_node` then is a nifty little trick to re-parse an already completed calculation and it does so by dynamically creating a `calcfunction` that represents that parsing. But for this calculation function I need to reproduce the capability of the `Parser` to attach outputs *even* when you also return an exit code. Before this change, in the case of an exit code, that is what was being returned, and so the outputs that were actually "attached" by the parser, were not actually also attached on the calculation function. The caller of `Parser.parse_from_node` would then get an empty dictionary as result, even if the parsing did actually attach some outputs.

ltalirz: Wow, this is complicated!

I'm still not sure I follow everything. Let me try to disentangle this a bit for my own understanding:

CalcJobs: 
 * return an exit code
 * engine takes care of attaching output nodes to CalcJobNode
 * `run` will return dict of output nodes

ProcessFunctions: 
 * return dict of nodes *or* exit code
 * engine takes care of attaching what is returned to ProcessFunctionNode (exit code or dictionary of output nodes)
 * `run` will return whatever the ProcessFunction returned

Parser.parse_from_node:
 * is a function never used (nor tested!) in the entire aiida-core codebase(?)
 * takes an existing CalcJobNode and re-parses it by creating a CalcFunction to do the job
 * Unfortunately, a CalcFunction is not a good match here, since it cannot return an exit code *and* have the engine attach/return the output nodes (by the way, when calling `parser.parse(**kwargs)` inside the calcfunction, what does `self.out` do?)
 * So, we somehow hack this attachment in by hand

Am I getting this more or less right?

If yes, then I would have some suggestions:
 * write 1-2 tests
 * expand the documentation of Parser.parse_from_node a bit based on the information above, including an example of when this function may be useful
 * how about having some version of `parse` that can take the process class as an argument so that you don't have to attach the outputs by yourself but can let `parse` do its job? You can then still use it inside the CalcFunction and pass it the `Process.current()`. This version of `parse` could e.g. be a free function that is reused in `Parser.parse`

ltalirz: I also find this mixing of different meanings for the return value of ProcessFunctions problematic... but maybe we run into this issue here only because we want to mimic a CalcJob with a CalcFunction.

I guess there should really be no distinction whether a ProcessFunction returns an exit code or something else.

sphuber: I think the best thing is to take a step back and go back to basics. The confusion arises mostly due to the nature and limitations of functions in python (and by extension AiiDA's process functions) and this particular nifty trick to allow easy re-parsing.

1. Any process can have one or multiple outputs and optionally an exit status and message can be set to indicate "failure".
2. When you "run" a process, i.e. with the launcher `run`, the return value will be the dictionary of output nodes registered to the process node. If no outputs were registered, this will be an empty dictionary.
3. The typical API to "register" outputs for a process is to call `self.out`. Note that this only naively works for `CalcJobs` and `WorkChains` as those are classes that have an instance `self` to call `out` on.
4. Process functions also need a way to attach outputs, but they don't have a `self`, so the `self.out` method, used for the other processes are not possible. Instead, the rule here is that the outputs are simply returned from the function. This can be either a single output node, or a dictionary of nodes.
5. How does one set the exit status and exit message on a process? One simply returns from a process step (be it a work chain outline step or the `parse` method) an instance of `ExitCode`. This return value will be detected by the engine who then sets the corresponding status and message as attributes on the node. In the case of the `WorkChain` this will also immediately instruct the engine to halt the execution. In this sense, returning an exit code from an outline step acts as an abort. Note that this of course makes no sense in the `CalcJob` context.
6. Since process functions are also functions and also can "fail" we also want to have the possibility to set their exit status and message through an exit code. The only way to "communicate" with the engine from a function, is by returning a value. This then means however that we cannot also any outputs. Note that this is a design choice. We could have chosen to allow returning a tuple, e.g. `(exit_code, outputs)` but I chose not to. The reasoning being that if there is a reason to want to mark the function as "failed" by wanting to return an exit code, you most likely don't want to return some partial outputs. Since process functions are limited in multiple ways, if you find yourself writing one where you do need this kind of complexity, you should probably be writing a `CalcJob` or `WorkChain` anyway.

This describes in a nutshell the design philosophy and consequent interface and usage of these process types. Now we get to the `Parser.parse_from_node`.

1. The reason this is never used is because I added it as a useful utility to re-trigger the parsing of a completed `CalcJob`. In a normal usage, this will be triggered by the engine when the calculation is successfully retrieved *and* a parser was defined in the input options.
2. For unit testing of parsers, I don't want to run the actual `Process`, but I want to "mock" the `CalcJobNode` with its retrieved folder and then simply call the parsing. I developed this solution for this purpose. I realized that this however may also be useful in real life where a calculation has already been completed and parsed, but you want to parse it again with another parser. In this case, you want the process to be as similar as possible and so want it to be represented it in the database by a process node, to which you can then attach the new parsed nodes as outputs  (or course you couldn't and wouldn't attach those to the original process node).
3. The best and easiest way to accomplish this was to represent the parsing as a calculation function as implemented by the `Parser.parse_from_node`.
4. The behavior of this calculation function should be "exactly" that of the original `CalcJob`, i.e. when you `run` it, it will return the outputs registered during parsing and if the parser returned an exit code, that needs to also be mirrored on the `CalcFunctionNode`.
5. This is where this fix comes in. When the parser attached outputs but *also* returned an exit status, I *had* to chose between one of them to return from the function, because the interface only allows for one to be returned. But this would then violate the requirement that the calculation function mirrors the normal calculation job exactly. This commit fixes this problem.

N.B. when you do `Parser.out` it does not directly link up the output node to the process node. Instead, it keeps it in an internal cache and when the parsing is finished, the engine will take care of linking them actually up to the node. This is another reason why with the current implementation, it is not possible to re-parse by reconstructing a parser instance manually and calling `parse`.

I hope that clears things up. Not sure where to put all this or how much of it to include in the docstring of `parse_from_node`. A lot of it is just basic rules of the current engine, nothing specific to this function. At some point this would be good to have in developers documentation of `aiida-core` but I am not sure if or when I will ever get to that.

ltalirz: Ok, thanks for the clarification. So, I wasn't too far off then.

I still think that 
 1. there needs to be at least one test for `Parser.parse_from_node` and I suggest this should be included in this PR.
 2. ProcessFunctions are of course free to return whatever they like (including exit code instances) but writing logic in aiida-core for the special case when they return exit codes looks like a problematic design to me. This is not for this PR to fix but we should probably revisit this design decision. By the way, there are languages like Go, where returning a tuple of value & exit code (or rather an "error" instance in that case) is the default way of error handling. 
 3. This hack of attaching outputs via a ProcessFunction using some kind of global variable (`Process.current()`) is dirty and invites problems down the line (e.g. say someone starts depending on the fact that what a ProcessFunction returns determines its output nodes). I understand that, because of the design decisions described above, the `Parser.parse` function, fundamentally, cannot be represented as a `CalcFunction` in a clean way. I think this fact should be food for thought

Happy to open an issue for 2./3.

sphuber: > there needs to be at least one test for Parser.parse_from_node and I suggest this should be included in this PR.
Agreed. I didn't do it, because testing the case for which I added logic in this PR is very difficult. I would need to write a dummy `Parser` class and mock `CalcJobNode` that attaches outputs *and* returns an exit code. I will add at least a test case for the normal case without an exit code.

> ProcessFunctions are of course free to return whatever they like (including exit code instances) but writing logic in aiida-core for the special case when they return exit codes looks like a problematic design to me.
I think there may still be some misunderstanding

I am not writing special logic here for process functions in general. I am writing a specific implementation of a `calcfunction`. It's required behavior needs it to be able to return both an exit code, while still attaching outputs. Because the standard API doesn't allow this, I employ the solution of `Process.current()`. I think this is perfectly fine in this case, given that `Parser.parse_from_node` is not part of code that is run by AiiDA itself (as you noted since it is never called in the code base) but is simply there as a useful utility.

> e.g. say someone starts depending on the fact that what a ProcessFunction
This change does not affect the `FunctionProcess` class behavior in general. It is just *a single* implementation of a `calcfunction`. That is all. Nothing of the base behavior of process functions is altered here.

sphuber:
@ltalirz I added a test and I updated the docstring heavily, so this should be good to go

Issue #2965, opened 2019-06-05
Should process functions be made submittable


sphuber:
Currently, it is not possible to `submit` process functions. This means that they will always "run" blockingly in the current interpreter, be it a local interpreter or one of a daemon worker. However, this also means that when a daemon worker is shutdown while it is working on a process function, that will be interrupted. Any calling processes will most likely than also fail. The fact that process functions always run blockingly means that shutting down the daemon becomes dangerous, because one risks killing active process functions. I am not sure yet if it would even be possible to make process functions submittable like other processes and make them restartable when the daemon is shutdown

Pinging @muhrin for input

muhrin:
I just spoke with @unkcpz about this as he may be interested in such a feature for his work.  Looking at it, it *should* be quite doable (and I definitely agree that it's useful).  Maybe using the `ProcessFunction.build()` to create the process class inside the `submit()` call (where we now assert that we're not being passed a process function) could be enough.  The rest of the `submit()` call is just expecting a `Process` class so it could be good to go.

I've left it up to @unkcpz if he has enough time to attempt this but I'm happy to look over any WIP PRs that open up about this.

unkcpz:
I guess it is safe to close this. @sphuber

sphuber:
Fixed in PR #4539

PR #2966, opened 2019-06-05
Fix bug in process function interrupt handlers causing `verdi daemon stop` to timeout

sphuber:
Fixes #2693 

The issue is addressed in two separate commits, because the source of the issue was two-fold. Although this solves this particular issue and makes the handling of interrupts in local runners more robust, a more general problem was laid bare as documented in issue #2965 .
Unfortunately, I haven't found a way to add unit tests for the issue that this PR solves.

### DIFF
--- a/aiida/engine/processes/functions.py
+++ b/aiida/engine/processes/functions.py
@@ -127,8 +127,8 @@ def process_function(node_class):
             :return: tuple of the outputs of the process and the process node pk
             :rtype: (dict, int)
             """
-            current_runner = get_manager().get_runner()
-            runner = get_manager().create_runner(with_persistence=False)
+            manager = get_manager()
+            runner = manager.create_runner(with_persistence=False)
             inputs = process_class.create_inputs(*args, **kwargs)
 
             # Remove all the known inputs from the kwargs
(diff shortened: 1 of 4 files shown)


aiida/engine/processes/functions.py
```diff
@@ -119,14 +119,15 @@ def run_get_node(*args, **kwargs):
...
-            because otherwise if this workfunction were to call another one from within its scope, that would use
+            because otherwise if this process function were to call another one from within its scope, that would use
             the same runner and it would be blocking the event loop from continuing.
 
             :param args: input arguments to construct the FunctionProcess
             :param kwargs: input keyword arguments to construct the FunctionProcess
             :return: tuple of the outputs of the process and the process node pk
             :rtype: (dict, int)
             """
+            current_runner = get_manager().get_runner()
```
ltalirz: maybe move this line directly before the `if not ...`

aiida/engine/processes/functions.py
```diff
@@ -140,18 +141,30 @@ def run_get_node(*args, **kwargs):
 
             process = process_class(inputs=inputs, runner=runner)
 
-            def kill_process(_num, _frame):
-                """Send the kill signal to the process in the current scope."""
-                LOGGER.critical('runner received interrupt, killing process %s', process.pid)
-                process.kill(msg='Process was killed because the runner received an interrupt')
+            # Only add handlers for interrupt signal to kill the current process if we are not in a daemon runner
```
ltalirz: maybe add what the alternative is (i.e. if you are not in a daemon runner, what type of runner are you in then?).

aiida/engine/processes/functions.py
```diff
@@ -140,18 +141,30 @@ def run_get_node(*args, **kwargs):
...
+            # Without this check, running process functions in a daemon worker would be killed if the daemon is shutdown
+            original_handler = None
+            original_signal = signal.SIGINT
```
ltalirz: original_signal somehow suggests there will be a new signal as well, but there isn't right?
I guess you could call it `kill_signal`...

ltalirz: same for the handler

sphuber: Good point for the signal, I will change it. However, I am not sure if the change for the handler would be good. This stems from the somewhat weird way of unbinding a signal in python. You cannot actually unbind it as such, you simply need to call `signal.signal(signal, handler)` again and here then `handler` needs to be what it was before. That is why I store that handle in the `original_handler` before overriding it, and then later you restore it

aiida/engine/processes/functions.py
sphuber: done

sphuber:
Thanks for this and all the other thoughtful reviews Leo

Issue #2880, opened 2019-05-16
connection cooldown time for local computers?


ltalirz:
AiiDa currently configures a connection cooldown time for local computers:
```
max@ip-172-31-1-140:~$ verdi computer configure local localhost2
Info: enter "?" for help
Connection cooldown time (sec) [2.0]:
Info: Configuring computer localhost2 for user aiida@localhost.
Success: localhost2 successfully configured for aiida@localhost
```

Not sure this really makes sense... one could leave it as an option for testing but I don't think we should prompt for it.

Also, the help shows the time as an integer, while the default is shown as a float (which makes more sense I find):
```
max@ip-172-31-1-140:~$ verdi computer configure local -h
Usage: verdi computer configure local [OPTIONS] COMPUTER

  Configure COMPUTER for local transport.

Options:
  -u, --user USER          Email address of the user.
  --safe-interval INTEGER  Minimum time between connections in sec
  -n, --non-interactive    Non-interactive mode: never prompt for input.
  -h, --help               Show this message and exit.
```

sphuber:
The `safe_interval` is a property of the `Transport` base class and therefore also needs to be set for the `LocalTransport`. If we want to keep it configurable, we need to keep the option in the configuration command. Since a default is provided (currently incorrectly set to `2.0`) all one has to do is press enter. With the `-n` flag even this is not necessary. I think that is perfectly acceptable for now.

ltalirz:
>  Since a default is provided (currently incorrectly set to 2.0) 

Ok, so I guess you'll set the default to 0 for local connections? That would already make more sense.

I still think it is preferable that `verdi computer configure local` does not prompt for it (and silently uses the default). I think in 99% of cases this does not make sense for a local connection, so we should not show this to users.

sphuber:
Yes I am fixing this now, as well as adding additional validation and cleaning things up. Currently you can set it to a negative number for example. It will be difficult to not have this prompt only for the local transport. These configure commands are created dynamically based on the registered transport entry points and the options are created based on the auth options defined in that class. Since this is a property of the base `Transport` class it will be difficult to hide this just for the local configure command. If this is really necessary, this would require a big rewrite.

ltalirz:
>  If this is really necessary, this would require a big rewrite.

Ok, then let's not invest time in this right now.

ltalirz:
>  Since a default is provided (currently incorrectly set to 2.0) all one has to do is press enter. With the -n flag even this is not necessary

[...]

PR #2967, opened 2019-06-05
Improve `safe_interval` validation `verdi computer configure`

sphuber:
Addresses #2880 partially

The computer configuration command for the transport specified that
an integer was required. However, a float value makes more sense. In
addition to this change, an additional validation callback was added to
make sure that the interval is positive.

The addition of the validation through the default option callback
mechanism required a small adaptation to the `InteractiveOption`.
The behavior of error handling was different if the option was
explicitly defined on the command line but without an explicit
`--non-interactive` flag. If a string value was passed, the type
validation would fail and prompt loop was never entered. However, if a
negative number was passed, the type validation would pass, but then the
normal callback would fail and in this case the prompt would start.
However, this was with the context improperly configured and so
exceptions would occur in the contextual defaults.

### DIFF
--- a/aiida/transports/transport.py
+++ b/aiida/transports/transport.py
@@ -28,6 +28,19 @@ DEFAULT_TRANSPORT_INTERVAL = 30.
 __all__ = ('Transport',)
 
 
+def validate_positive_number(ctx, param, value):  # pylint: disable=unused-argument
+    """Validate that the number passed to this parameter is a positive number.
+
+    :param ctx: the `click.Context`
+    :param param: the parameter
+    :param value: the value passed for the parameter
+    :raises `click.BadParameter`: if the value is not a positive number
+    """
+    if value is not None and value < 0:
+        from click import BadParameter
+        raise BadParameter('value needs to be a positive number')
+
+
 @six.add_metaclass(ABCMeta)
 class Transport(object):
     """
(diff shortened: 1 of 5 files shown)


sphuber:
Not entirely sure about this solution. To me it seems weird that doing
```
verdi computer configure local --safe-interval string
```
will go through the basic click error message path, whereas
```
verdi computer configure local --safe-interval -1
```
instead would start the prompt cycle. Here I chose to make it consistent by having the second follow the path of the first. But maybe in both cases, the prompting should actually be engaged unless the `--non-interactive` flag is *explicitly* set. If this is the desired behavior, then that is a problem in the `InteractiveOption` independent of this change and would require a separate issue probably.

aiida/transports/transport.py
```diff
@@ -28,6 +28,19 @@
...
 
+def validate_positive_number(ctx, param, value):  # pylint: disable=unused-argument
+    """Validate that the number passed to this parameter is a positive number.
+
+    :param ctx: the `click.Context`
+    :param param: the parameter
+    :param value: the value passed for the parameter
+    :raises `click.BadParameter`: if the value is not a positive number
+    """
+    if isinstance(value, (int, float)) and value < 0:
```
ltalirz: so, why do you need to check here that it is an int/float when the option requires it to be a float?
does the built-in validation of click happen later?
if yes, then this is indeed backwards - we would like the click validation to go first here.

ltalirz:
> To me it seems weird that doing...

agreed!

> Here I chose to make it consistent by having the second follow the path of the first. But maybe in both cases, the prompting should actually be engaged unless the --non-interactive flag is explicitly set.

I like the solution you implemented better. If I provide an option with a wrong value on the command line, I would expect the command to fail, not to prompt.

aiida/transports/transport.py
sphuber: No `click` validation seems to go first, but in some code pathways this gets called with a `None` value. I am thinking now that it is probably called *before* the contextual default is populated. Not sure though

ltalirz: Ok, in this case how about transforming this to
`if value is not None and value < 0` ?
or at least you can drop the `int` since you should be getting a `float` here

sphuber: Yeah good point, I like the `is not None` best

sphuber:
@ltalirz is there still something you wanted me to address here?

PR #2972, opened 2019-06-06
fix typo in the serializer for `dict`

zhubonan:
The `to_aiida_type` serializer for `dict` was not working because of a typo. This is a quick patch to make it work.
PS: It seems `to_aiida_type` does not have any tests written as the bug is not caught.


sphuber:
Would it be possible to write some unit tests for these? As you noted, that is the reason of this bug

zhubonan:
@sphuber No problem, I will come up with some unit tests for this and other types.

zhubonan:
Cosed in favour of  #2973

PR #2986, opened 2019-06-08
Ensure that stacktrace is included in transport task exceptions

### LINKED ISSUE #2898: Add stack trace of failed transport tasks to log messages
Currently, only the exception message of the exponential backoff mechanism is shown in the report, but this is not very helpful to understand *why* it failed. The daemon log *will* contain it, but this is not useful if one needs to know what happened to a specific calculation job

### LINKED ISSUE #2984: `ProcessLauncher._continue` used removed `create_future`
This method on the communicator no longer exists and will except

sphuber:
Fixes #2898 and fixes #2984 

When a transport task of a calculation job excepts, the exponential
back off retry wrapper calls the logger of the process node with the
exception. This will trigger the `DbLogHandler` such that the log
message is entered into the database with a link to the process node.
However, only the message was included and not the stack trace which
made it difficult to determine the cause from `verdi process report`.
To fix this, in the `Log.objects.create_from_record` we attempt to
format a traceback from the `exc_info` if it is included in the record
which will be the case for an exception log.

This commit also fixes a bug in the `ProcessLauncher._continue` method
that was calling the `create_future` method of the communicator, which
has long been removed. Instead a normal future should just be created.

### DIFF
--- a/aiida/engine/utils.py
+++ b/aiida/engine/utils.py
@@ -161,6 +161,7 @@ def exponential_backoff_retry(fct, initial_interval=10.0, max_attempts=5, logger
     :param fct: the function to call, which will be turned into a coroutine first if it is not already
     :param initial_interval: the time to wait after the first caught exception before calling the coroutine again
     :param max_attempts: the maximum number of times to call the coroutine before re-raising the exception
+    :param ignore_exceptions: list or tuple of exceptions to ignore, i.e. when caught do nothing and simply re-raise
     :raises: ``tornado.gen.Result`` if the ``coro`` call completes within ``max_attempts`` retries without raising
     """
     if logger is None:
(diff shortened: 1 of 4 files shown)


aiida/engine/utils.py
```diff
@@ -180,13 +180,15 @@ def exponential_backoff_retry(fct, initial_interval=10.0, max_attempts=5, logger
             if ignore_exceptions is not None and isinstance(exception, ignore_exceptions):
                 raise
 
+            count = iteration + 1
+            core_name = coro.__name__
```
ltalirz: should this be `coro_name`?

aiida/engine/utils.py
```diff
@@ -180,13 +180,15 @@ def exponential_backoff_retry(fct, initial_interval=10.0, max_attempts=5, logger
             if ignore_exceptions is not None and isinstance(exception, ignore_exceptions):
```
ltalirz: ignore_exceptions seems to be missing a docstring. Can you add one?
It's not self-evident that "ignoring exceptions" means re-rasing them

sphuber: Done

aiida/engine/utils.py
sphuber: absolutely, done

sphuber:
thanks for the review @ltalirz

PR #2989, opened 2019-06-09
Fix problem in validation of `InteractiveOption` CLI option type

### LINKED ISSUE #2987: `verdi computer configure --non-interactive` does not properly validate `safe_interval`
Executing `verdi computer configure local localhost --non-interactive` for the first ime, will pass validation just fine, despite no default value being set for `safe_interval`. The auth info record in the database will look like:
```
 id |       auth_params       | metadata | enabled | aiidauser_id | dbcomputer_id 
----+-------------------------+----------+---------+--------------+---------------
  3 | {"safe_interval": null} | {}       | t       |            1 |             1
```

This in turn causes #2983 

Also, the following behavior is problematic and really weird:
```
(aiida_dev) sph@bastion:~/code/aiida/env/dev/aiida-core$ verdi computer configure local localhost
Info: enter "?" for help
Connection cooldown time (s) [-1.0]: -2
Error: -2.0 is not a valid positive number
Info: Minimum time interval in seconds between consecutive connection openings
Connection cooldown time (s) [-1.0]: -2
Info: Configuring computer localhost for user sebastiaan.huber@epfl.ch.
Success: localhost successfully configured for sebastiaan.huber@epfl.ch
```
So the first time the validation properly catches the wrong input. However, if one just enters the same faulty value again, it is accepted

sphuber:
Fixes #2987 

The option allows to specify a callback function to perform validation
in addition to the type. However, in the prompt loop, only the normal
type validation was called. This caused the validation in the callback
to be skipped if called more than once, allowing illegal values to be
accepted in interactive mode if simply passed more than once.

Additionally, the `non_interactive_default` setting for the
`safe_interval` option of `verdi computer configure` in combination with
the implementation of the callback function, which ignored `None`
values, allowed `None` values to be set in non-interactive mode. The
callback now properly checks for the correct type and raises an error if
`None` is passed.

### DIFF
--- a/aiida/cmdline/params/options/interactive.py
+++ b/aiida/cmdline/params/options/interactive.py
@@ -22,25 +22,6 @@ from aiida.cmdline.utils import echo
 from .conditional import ConditionalOption
 
 
-def noninteractive(ctx):
-    """check context for non_interactive flag"""
-    return ctx.params.get('non_interactive')
-
-
-def string_wrapper(msg, max_length=100):
-    """Wraps a string into multiple lines according to ``max_length``"""
-    words = msg.split(' ')
-    lines = ['']
-    for word in words:
-        if not lines[-1]:
-            lines[-1] = word
-        elif len(lines[-1]) + len(word) + 1 <= max_length:
-            lines[-1] += ' ' + word
-        else:
-            lines.append(word)
-    return lines
-
-
 class InteractiveOption(ConditionalOption):
     """
     Intercepts certain keyword arguments to circumvent :mod:`click`'s prompting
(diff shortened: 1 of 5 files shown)


sphuber:
Damn, writing a CLI without bugs is hard :sweat_smile:

ltalirz:
`Error: missing option --safe-interval` https://travis-ci.org/aiidateam/aiida_core/jobs/543435227#L1398

Also, could you have a look at `def prompt_callback`?
It seems to me that `after_callback` can now be called too many times.

> However, in the prompt loop, only the normal
type validation was called. 

Would it be possible to call the same (single) function in both cases?
I.e. can we eliminate the possibility that the loop does something else than the non-loop?

aiida/cmdline/params/options/interactive.py
```diff
@@ -191,10 +191,12 @@ def safely_convert(self, value, param, ctx):
         successful = False
         try:
             value = self.type.convert(value, param, ctx)
+            value = self.after_callback(ctx, param, value)
```
ltalirz: as mentioned in my previous comment, looking at the definition of `def prompt_callback`  this seems to result in `after_callback` being called twice in the last run.
Is this correct?

sphuber: I think you were right. I added it because if an invalid value was passed twice in a row in interactive mode, the second time, the callback was not called and the value was accepted. However, the correct approach is to make sure the normal `callback` is called, which will ensure that the prompt loop is reentered properly. Now I think the `after_callback` is only called once.

PR #2991, opened 2019-06-11
Correct decorator implementation using the `wrapt` library

### LINKED ISSUE #2446: Current implementation of decorators loses function signature
The current implementation of decorators in `aiida-core` cause the functions that are wrapped to lose their function signature. This will cause `inspect` to fail miserably. The same problem occurs with it losing its `__name__` and `__doc__`, which is sometimes fixed by using `functools.wraps`, however, that does not properly restore the function signature. Example:
```
import inspect
from aiida.cmdline.utils import decorators

def unwrapped(a, b):
    pass

inspect.getargspec(wrapped)
ArgSpec(args=['a', 'b'], varargs=None, keywords=None, defaults=None)
```
Now when we wrap it, for example with `with_dbenv` decorator, we get:
```
@decorators.with_dbenv()
def wrapped(a, b):
    pass

inspect.getargspec(wrapped)
ArgSpec(args=[], varargs='args', keywords='kwargs', defaults=None)
```

Now this is currently not causing any problems, that we know of, but it might in the future. In addition, I use the `inspect` functionality in launch scripts that I have to wrap with `with_dbenv` and those now stopped working. There are [various solutions](https://hynek.me/articles/decorators/), of which the [wrapt](https://wrapt.readthedocs.io/en/latest/) library is the best as in most correct, but also slowest. Let's discuss if we would be willing to put this in our dependencies and fix our decorators.

sphuber:
Fixes #2446 

The naive implementation of a decorator will break some of the
functionality of the functions they wrap. For example, the original
docstring will be lost. The module `functools` provides the `wraps`
decorator to fix this problem, but this is not the only issue. In
addition to the docstring, other properties, such as the function
signature are also lost. This means that using the `inspect` module on
decorated functions is hopelessly broken. The `wrapt` library provides
the `decorator` decorator that simplifies writing decorators, while
ensuring that all the important properties of wrapped function are
maintained.

### DIFF
--- a/aiida/backends/tests/__init__.py
+++ b/aiida/backends/tests/__init__.py
@@ -74,6 +74,7 @@ DB_TEST_LIST = {
         'common.extendeddicts': ['aiida.backends.tests.common.test_extendeddicts'],
         'common.folders': ['aiida.backends.tests.common.test_folders'],
         'common.hashing': ['aiida.backends.tests.common.test_hashing'],
+        'common.lang': ['aiida.backends.tests.common.test_lang'],
         'common.links': ['aiida.backends.tests.common.test_links'],
         'common.logging': ['aiida.backends.tests.common.test_logging'],
         'common.serialize': ['aiida.backends.tests.common.test_serialize'],
--- a/aiida/backends/tests/common/test_lang.py
+++ b/aiida/backends/tests/common/test_lang.py
@@ -1,0 +1,107 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida_core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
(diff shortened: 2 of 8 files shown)


sphuber:
A few notes:
1. I have not adapted the implementation of the `protected` nor `override` decorators yet because they are a bit more complicated.
2. The same goes for the `calcfunction` and `workfunction` decorator
3. The `with_dbenv` currently does not take an argument, so could be simplified by removing a wrapping layer. However, this will require it to be used as `@with_dbenv` instead of `@with_dbenv()`. Since this is a breaking change, I have left it. Moreover, if in the future we want to allow to specify which profile to load through the decorator, this can be done without breaking the interface, I think

Before I try to see if I can tackle 1 and 2, I wanted to see if you guys feel this is worth the added dependency. I certainly do in any case.

ltalirz:
wrapt seems like a nice library, and it's available on conda as well.
the only criticism I could have here is that it seems to have a [bus factor of 1](https://github.com/GrahamDumpleton/wrapt/graphs/contributors) and that I don't consider the features we add here terribly important - I agree that they are nice to have, though, and would be ok with including wrapt as a dependency.

giovannipizzi:
Same comment as @ltalirz - anyway it seems that the library is quite stable and only doing bugfixes in the past two years.

PR #2993, opened 2019-06-12
Add `errors_handled` updatable attribute to `WorkChainNode`

### LINKED ISSUE #2992: Implement updatable attribute `errors_handled` for `WorkChainNode`
This is a feature that only really makes sense in the framework of the `BaseRestartWorkChain` and the `register_error_handlers` that is currently implemented in `aiida-quantumespresso` but since that should probably be included in `aiida-core` soon, I think this prerequisite is worthy to be included now.

The idea of the `BaseRestartWorkChain` is to provide the basic infrastructure for a robust work chain to wrap around a `CalcJob`. Through the `register_error_handler` one can dynamically update the error handlers available for a specific implementation. This decorator can already fire a log message if triggered, but this is not queryable. I have realized that in practice it would be very valuable to be able to query for work chains that triggered certain error handlers. To this end the decorator can simply append the error handlers name when triggered to a list that is an attribute on the work chain node. But for this to work, the attribute needs to be marked as an `updatable_attribute`.

sphuber:
Fixes #2992 

This new updatable attribute for the `WorkChainNode` class will be
useful once the `BaseRestartWorkChain` and `register_error_handler`
concepts will be added to `aiida-core`. These utilities provide a basic
framework to quickly write powerful work chains around `CalcJob`
implementations. Through the `register_error_handler` methods can be
attached to the `BaseRestartWorkChain` implementation to extend the
error handling capability in the case of a failed calculation. To make
it possible to query for work chains that had certain error handlers
triggered during their execution, their method names need to be added to
some list in the attributes. This is the intended purpose for the new
`errors_handled` attribute.

### DIFF
--- a/aiida/backends/tests/__init__.py
+++ b/aiida/backends/tests/__init__.py
@@ -126,6 +126,7 @@ DB_TEST_LIST = {
         'orm.logs': ['aiida.backends.tests.orm.test_logs'],
         'orm.mixins': ['aiida.backends.tests.orm.test_mixins'],
         'orm.node.calcjob': ['aiida.backends.tests.orm.node.test_calcjob'],
+        'orm.node.workchain': ['aiida.backends.tests.orm.node.test_workchain'],
         'orm.node.node': ['aiida.backends.tests.orm.node.test_node'],
         'orm.querybuilder': ['aiida.backends.tests.orm.test_querybuilder'],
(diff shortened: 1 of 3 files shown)


sphuber:
After some more thought, this concept relies a lot on the specific implementation of the `BaseRestartWorkChain` and `register_error_handler` tools and so it makes more sense to store this data in extras as opposed to attributes. If at some point this gets included in `aiida-core` the switch can potentially be made

PR #2995, opened 2019-06-13
 Improve SchedulerError message in PbsBaseClass._parse_joblist_output 

chrisjsewell:
A minor improvement to the `SchedulerError` message, so it is easier to debug using `verdi process report`


sphuber:
Thanks @chrisjsewell

PR #3000, opened 2019-06-13
Set the `process_status` if `WorkCalculation` goes to `Wait` for subprocess

### LINKED ISSUE #1941: Update process status for processes waiting on child process
With the addition of the `process_status` attribute of `Processes`, we can know add a useful message to this field whenever a process enters the waiting state because it is waiting on a child process. This can then be displayed on `verdi process list`

sphuber:
Fixes #1941

As soon as an `awaitable` is added to a `WorkChain` instance, its
process status is updated. This process status will contain a human
readable string with the pks of the sub processes for which it is
waiting. If an `awaitable` is removed and no awaitables remain, the
process status of the work chain node is cleared.

### DIFF
--- a/aiida/backends/tests/engine/test_work_chain.py
+++ b/aiida/backends/tests/engine/test_work_chain.py
@@ -24,11 +24,11 @@ from aiida.backends.testbase import AiidaTestCase
 from aiida.common import exceptions
 from aiida.common.links import LinkType
 from aiida.common.utils import Capturing
-from aiida.engine import ExitCode, Process, ToContext, WorkChain, if_, while_, return_, run, run_get_node, submit
-from aiida.engine import launch, calcfunction
+from aiida.engine import ExitCode, Process, ToContext, WorkChain, if_, while_, return_, launch, calcfunction
 from aiida.engine.persistence import ObjectLoader
 from aiida.manage.manager import get_manager
 from aiida.orm import load_node, Bool, Float, Int, Str
+from six.moves import range
 
 
 def run_until_paused(proc):
@@ -79,7 +79,7 @@ def run_and_check_success(process_class, **kwargs):
     :returns: instance of process
     """
     process = process_class(inputs=kwargs)
-    run(process)
+    launch.run(process)
     assert process.node.is_finished_ok is True
 
     return process
@@ -196,7 +196,7 @@ class TestExitStatus(AiidaTestCase):
     """
 
     def test_failing_workchain_through_integer(self):
-        result, node = run_get_node(PotentialFailureWorkChain, success=Bool(False))
+        result, node = launch.run.get_node(PotentialFailureWorkChain, success=Bool(False))
         self.assertEquals(node.exit_status, PotentialFailureWorkChain.EXIT_STATUS)
         self.assertEquals(node.exit_message, None)
         self.assertEquals(node.is_finished, True)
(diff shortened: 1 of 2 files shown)


ltalirz:
@giovannipizzi Let me know in case you don't want to review this. In this case I will take care of it (just not very familiar with "awaitables")

aiida/backends/tests/engine/test_work_chain.py
```diff
@@ -636,6 +636,39 @@ def do_run(self):
...
+
+            def do_run(self):
+                pks = []
+                for i in range(2):
+                    node = self.submit(SubWorkChain)
+                    pks.append(node.pk)
+                    self.to_context(subwc=node)
+
+                assert 'Waiting' in self.node.process_status
+                assert str(pks[0]) in self.node.process_status
```
giovannipizzi: Just to double check, this should always work, right? In the sense: SubWorkChain returns immediately, but if tests are run with 1 worker only, the process will have to terminate first to process this function before running the subworkflow? Or since even a single worker can process more messages from RabbitMQ, it is technically possible that the status is not 'Waiting' anymore by the time we get here?

aiida/engine/processes/workchains/workchain.py
```diff
@@ -123,30 +122,32 @@ def on_run(self):
...
-        Precondition: must be a awaitable that was previously inserted
+        Precondition: must be an awaitable that was previously inserted.
 
-        :param awaitable: The awaitable to remove
+        :param awaitable: the awaitable to remove
         """
         self._awaitables.remove(awaitable)
 
+        if not self._awaitables:
+            self.node.set_process_status(None)
```
giovannipizzi: Do I understand correctly that:
- every time a new await able is added, the status is added with the list of PKs of th awaitables currently interted
- when one is removed, the status is not changed (unless all have gone)
- Also, if an awaitable is added after a few have been already removed, the string is updated with the currently inserted ones.
Is this the intended behaviour? It seems a bit counterintuitive.

Also, is there the possibility that the last two finish approximatively at the same time and both have a non-empty `self._awaitables` or this is not possible and the whole logic is "thread-safe"?`

aiida/backends/tests/engine/test_work_chain.py
sphuber: No this will always work, but to do so, I did have to put the checks in the same step of the main work chain as where the sub workchains are submitted. The reason is exactly as you expected, since the sub work chains will execute and return straight away, if I try to do it outside, I can only really do it after the `do_run` step, at which point the awaitables have been cleared. Since the message is set as soon as `to_context` is called, I can check straight after.

aiida/engine/processes/workchains/workchain.py
sphuber: yes the behavior you describe is correct. The only problematic part there, I think, is where the status is not updated until **all** awaitables are completed. I can change this to update the status each time an awaitable is removed. Your third bullet point scenario is currently not really possible since sub processes won't be found to be finished by the one who launched them, until he tries to go to the next step. But while waiting to go to the next step, he cannot insert. So there is that guarantee that once he added all awaitables and started waiting, no new ones will be added.

sphuber:
@giovannipizzi  I have updated the changes to now properly update the status upon every awaitable insertion **and** removal

PR #3001, opened 2019-06-13
Allow to specify number of workers in `verdi daemon start`

### LINKED ISSUE #2998: Allow to specify the number of workers in `verdi daemon start`
Now one has to do `verdi daemon start && verdi daemon incr 7` whereas I want to be able to do `verdi daemon start 8`.

sphuber:
Fixes #2998 

Now you can start more than just one worker when starting the daemon.

### DIFF
--- a/aiida/cmdline/commands/cmd_daemon.py
+++ b/aiida/cmdline/commands/cmd_daemon.py
@@ -27,6 +27,17 @@ from aiida.cmdline.utils.daemon import get_daemon_status, \
 from aiida.manage.configuration import get_config
 
 
+def validate_positive_non_zero_integer(ctx, param, value):  # pylint: disable=unused-argument,invalid-name
+    """Validate that `value` is a positive integer."""
+    if not isinstance(value, int):
+        raise click.BadParameter('{} is not an integer'.format(value))
+
+    if value <= 0:
+        raise click.BadParameter('{} is not a positive non-zero integer'.format(value))
+
+    return value
+
+
 @verdi.group('daemon')
 def verdi_daemon():
     """Inspect and manage the daemon."""
@@ -34,12 +45,11 @@ def verdi_daemon():
 
 @verdi_daemon.command()
 @click.option('--foreground', is_flag=True, help='Run in foreground.')
+@click.argument('number', default=1, type=int, callback=validate_positive_non_zero_integer)
 @decorators.with_dbenv()
 @decorators.check_circus_zmq_version
-def start(foreground):
-    """
-    Start the daemon
-    """
+def start(foreground, number):
+    """Start the daemon with NUMBER workers [default=1]."""
     from aiida.engine.daemon.client import get_daemon_client
 
     client = get_daemon_client()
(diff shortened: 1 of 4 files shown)


aiida/cmdline/commands/cmd_daemon.py
```diff
@@ -64,13 +74,17 @@ def start(foreground):
 
     print_client_response_status(response)
 
+    if number > 1:
```
ltalirz: is there a reason to do it this way?
i.e. why not make the number of workers a parameter of `get_daemon_client` and just start with the right number of workers?

sphuber: I couldn't do this, because we do not actually start the daemon through the client. That just serves to query it once it is running and to stop it. The daemon is started through the hidden internal command `verdi daemon start-circus`.

However, I agree that the method I chose is sub-optimal so I changed it. I now pass `number` through to the sub command that we call, and there in the `arbiter` configuration, I launch the number of processes we want straight away.

I added an exception being raised if more than one processes are requested while running in the foreground, since that doesn't make any sense and added some more tests for this

aiida/cmdline/commands/cmd_daemon.py
```diff
@@ -27,35 +27,46 @@
...
+
+
 @verdi.group('daemon')
 def verdi_daemon():
     """Inspect and manage the daemon."""
 
 
 @verdi_daemon.command()
 @click.option('--foreground', is_flag=True, help='Run in foreground.')
+@click.argument('number', default=1, type=int, callback=validate_positive_non_zero_integer)
```
ltalirz: Could this be an option instead?
The command currently is `verdi start-circus 12` and I think `verdi start-circus --nworkers 12` (or similar) would be more self-explanatory.

sphuber: One never calls `verdi daemon start-circus` it is a hidden command. But it would be `verdi daemon start [NUMBER]`. I chose the argument because that is inline with the already existing commands `verdi daemon incr/decr [NUMBER]`.

ltalirz:
Thanks, that looks already much better!

Sorry to keep nagging here - just two more questions:

 * Is there a configuration setting for the default number of workers? Or is this something one has to put every time one starts the daemon? If there is no configuration setting for this, perhaps we should add an issue to add one?
 * One thing I never quite understood: Do we *need* to go via click here? Can't we just spawn a process directly from a function like [so](https://stackoverflow.com/a/2046630/1069467)?
  Anyhow, this is just out of curiosity.

sphuber:
> Is there a configuration setting for the default number of workers? Or is this something one has to put every time one starts the daemon?

There isn't a configuration setting yet, as this functionality is new, so yes, for now one would always have to specify an explicit number if one wants more than the default one. I don't think it is a problem, but adding a new issue with a feature request to make this configurable would be a good idea.

> Do we need to go via a hidden click command here?

It sure is not necessary, it would certainly be possible to implement it differently. I think Rico implemented the very first command this way when migrating to `circus`. I just kept it and especially now that it is properly hidden, I don't see any harm

ltalirz:
Ok, then I would kindly request to 

 * open an issue as discussed above
 * mention somewhere in an appropriate docstring (or in a separate issue) the alternative of going through multiprocessing

Will approve then - am offline for 2h now

sphuber:
> open an issue as discussed above

Done, see #3007 

> mention somewhere in an appropriate docstring (or in a separate issue) the alternative of going through multiprocessing

I just realized why keeping the `click` route is preferable. Since we are launching a sub process, the correct profile needs to be loaded. This is currently taken care of automatically by the `verdi` click infrastructure that correctly parses and loads the desired profile. If we were to go through a normal function that we would call as a sub process, that function would have to provide some way to take a profile different from the default. You will basically end up implementing another command line parser.

PR #3002, opened 2019-06-13
add order options to `verdi process list`

chrisjsewell:
As it says on the tin! Particularly helpful if you want to list the last *x* processes, i.e. `verdi process list -D desc -l 10`

### DIFF
--- a/aiida/cmdline/params/options/__init__.py
+++ b/aiida/cmdline/params/options/__init__.py
@@ -280,6 +280,11 @@ ORDER_BY = OverridableOption(
     type=click.Choice(['id', 'ctime']), default='ctime', show_default=True,
     help='Order the entries by this attribute.')
 
+ORDER_DIR = OverridableOption(
+    '-D', '--order-dir', 'order_dir',
+    type=click.Choice(['asc', 'desc']), default='asc', show_default=True,
+    help='List the entries in ascending or descending order')
+
 PAST_DAYS = OverridableOption(
     '-p', '--past-days', 'past_days',
     type=click.INT, metavar='PAST_DAYS',
--- a/aiida/cmdline/commands/cmd_process.py
+++ b/aiida/cmdline/commands/cmd_process.py
@@ -33,6 +33,8 @@ def verdi_process():
 @verdi_process.command('list')
 @options.PROJECT(
     type=click.Choice(CalculationQueryBuilder.valid_projections), default=CalculationQueryBuilder.default_projections)
+@options.ORDER_BY()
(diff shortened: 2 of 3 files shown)


aiida/cmdline/params/options/__init__.py
```diff
@@ -280,6 +280,11 @@ def active_process_states():
     type=click.Choice(['id', 'ctime']), default='ctime', show_default=True,
     help='Order the entries by this attribute.')
 
+ORDER_DIR = OverridableOption(
+    '-D', '--order-dir', 'order_dir',
```
sphuber: Let's not use an abbreviation here, i.e. let's use `--order-direction`, especially since the flags are tab completed

aiida/cmdline/params/options/__init__.py
```diff
@@ -280,6 +280,11 @@ def active_process_states():
     type=click.Choice(['id', 'ctime']), default='ctime', show_default=True,
     help='Order the entries by this attribute.')
 
+ORDER_DIR = OverridableOption(
```
sphuber: Also name this `ORDER_DIRECTION`

sphuber:
Also a question: why do your PR branches always have a commit with message `CJS fix`, then a shit-ton of `Merge remote-tracking branch 'upstream/develop' into develop` and then the actual commit with the fix?

chrisjsewell:
> Also a question: why do your PR branches always have a commit with message `CJS fix`, then a shit-ton of `Merge remote-tracking branch 'upstream/develop' into develop` and then the actual commit with the fix?

Yeh its because of that that stupid `CJS fix` commit I made to my `develop` fork branch. So every time I've updated it, by merging in `upstream/devlop`, it was adding these extra commits.
I've just run `git reset --hard upstream/develop`, which looks to have fixed it. However, I've royally screwed up this branch now, so I'm going to close this PR and open a new one with a fresh branch.

sphuber:
If your fork's `develop` is messed up, you should do the following:
```
git checkout develop
git fetch upstream
git reset --hard upstream/develop
git push -f fork develop
```
The key is to push to your fork again, otherwise you will only just fix it locally.
Also, the best way for PRs in terms of a workflow is to always start from the `upstream/develop`.
You never really need your fork's `develop.
So always something like:
```
git fetch upstream
git checkout -b fix_something upstream/develop
git commit # after changes of course
git push fork fix_something
```
And then make a PR off of that branch.
Then if you need to update, just use `rebase` instead of merge, i.e.
```
git fetch upstream
git rebase upstream/develop
```

PR #3003, opened 2019-06-13
Allow to define link label for `CALL` links

### LINKED ISSUE #2559: Allow to specify the `CALL` link label when submitting processes
This could be passed in the `metadata` that now properly namespaces such additional inputs to a process that are not "actual" inputs such as nodes

sphuber:
Fixes #2559 

The `Process` class defines a new metadata input `call_link_label` that
when set for a process that is called by another, will be used as the
link label for the `CALL` link, instead of the default.


aiida/engine/processes/process.py
```diff
@@ -65,12 +65,18 @@ class SaveKeys(enum.Enum):
...
-        spec.input('{}.description'.format(spec.metadata_key), valid_type=six.string_types[0], required=False)
-        spec.input('{}.label'.format(spec.metadata_key), valid_type=six.string_types[0], required=False)
+        spec.input('{}.store_provenance'.format(spec.metadata_key), valid_type=bool, default=True,
+            help='If set to `False` provenance will not be stored in the database.')
+        spec.input('{}.description'.format(spec.metadata_key), valid_type=six.string_types[0], required=False,
+            help='Description to set on the process node.')
+        spec.input('{}.label'.format(spec.metadata_key), valid_type=six.string_types[0], required=False,
+            help='Label to set on the process node.')
+        spec.input('{}.call_link_label'.format(spec.metadata_key), valid_type=six.string_types[0], default='CALL',
+            help='The label to use for the `CALL` link if the process is called by another process.')
```
ltalirz: So, the process being called sets the label of the link - not the one calling?
From a naive perspective this seems a bit strange since the calling process is the one "acting"... but maybe this comes from a specific use case?

sphuber: This is simply due to the user-facing API. A user never explicitly adds a `CALL` link, but this is done automatically. Currently there is two ways:
* Through `self.submit` in a `WorkChain`
* Or when calling another process from within a `workfunction`
The only way for the user to pass the desired link label is through the `metadata`.
Now note also that it is never the "parent" process literally calling the child. It just so happens that the child will be in "the scope" of the parent. Since this scope always corresponds to a unique parent, we can use this fact to set the correct `CALL` link in the construction of the child process. So as you can see, the creation of the `CALL` link was already happening this way, I am merely adding the possibility for the user to change the link label through the `metadata` input.

I will add a test that also demonstrates and tests the example of something being called from within a work function.

sphuber:
@ltalirz I have added two more tests to verify that calling a process function from a work chain and another process function also works when passing the label in the metadata.

aiida/engine/processes/process.py
ltalirz: Hm, just from a logical point of view, I still think that both of the following statements are statements a user may want to make:

 1. When this process calls other processes, the link label should be X
 2. When this process is called by other processes, the link label should be Y

We may choose to only support one of them in order not to run into conflicts, but I still find 1. more "natural".
Anyhow, perhaps the example will show which one is more practically useful.

ltalirz: Ok, I see the examples are still "abstract".
Can you give (just in a comment) a real-life example to understand what these labels are used for?

sphuber: Imagine in a `WorkChain` I launch multiple `CalcJobs`. Currently there is no way to set the `CALL` link label, so they will all have `CALL` as default. For example calling `verdi process show` on such a work chain will show:
```
Called         PK  Type
---------  ------  ----------------
CALL_CALC  138252  CalcJobNode
CALL_CALC  133813  CalcJobNode
CALL_CALC  127091  CalcJobNode
CALL_CALC  120598  CalcJobNode
CALL_CALC  120594  CalcFunctionNode
```
This is not necessarily bad, but with this new feature you can have it such that it will be:
```
Called         PK  Type
---------  ------  ----------------
create_kpoints  120594  CalcFunctionNode
iteration_01  120598  CalcJobNode
iteration_02  127091  CalcJobNode
iteration_03  133813  CalcJobNode
iteration_04  138252  CalcJobNode
```

ltalirz: Thanks, this example makes total sense, and I understand now that it *is* actually the calling process that defines the label (because it will pass the label when creating the processes it calls).
I believe it was a bit too late in the day when I did the review ;-)

ltalirz:
One last comment (not a problem): This means a Process class can define a *default* CALL label that is different from the standard one, which might come a bit unexpected to its users.

Anyhow, I guess this is unlikely, and even then it will be done in the Process spec.

Issue #3035, opened 2019-06-17
Move away the section for developers on "aiida-plugins"


giovannipizzi:
Either:
- move it to the aiida-diff where appropriate (_and link to this doc from the AiiDA-core docs_), or 
- integrate it with the sections on how to write for instance a calculation plugin (depending on the content). 

It is possible the content will need to be split between the two locations.

The section is now called "Development/AiiDA plugins" and will soon be renamed to "aiida-plugins [Section to be moved]"

Also:
- make sure it's updated with AiiDA 1.0 and does not contain wrong/old content.
- remove the old section on parser warnings policy that is old and not used anymore

Issue #3022, opened 2019-06-17
Check documentation for "working with AiiDA/Data types"


giovannipizzi:
Check if it the section is updated to AiiDA 1.0, if it is correct, and if remove possible redundancies

https://aiida.readthedocs.io/projects/aiida-core/en/latest/working_with_aiida/index.html#data-types

giovannipizzi:
Note that probably it makes sense to fix at the same time #3019 by bringing those examples into this section

sphuber:
Fixed in PR #3053

Issue #3019, opened 2019-06-17
Restructure Data examples in the documentation (used to be in a "Tutorials" category)


giovannipizzi:
There are currently some examples on how to use "StructureData" and "Pseudopotentials" in a category called "Tutorials", see e.g. [https://aiida.readthedocs.io/projects/aiida-core/en/latest/tutorial/index.html](this page).

The section "tutorials" is not really a tutorials section and should be removed. The content should be 
- double checked and verified
- moved inside the page on [working with data types](https://aiida.readthedocs.io/projects/aiida-core/en/latest/working_with_aiida/index.html#data-types)

giovannipizzi:
Note that this should be probably done together with #3022

PR #3036, opened 2019-06-17
First restructuring of devel docs

giovannipizzi:
This is preliminary to #3035 to move some docs out
of the main documentation or at least revised.

Moreover, also the main devel section has been
renamed from "Development" to "For AiiDA developers"
and the only page under "AiiDA design" has been moved 
inside the same section "for AiiDA developers".

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -65,7 +65,6 @@ The software is available at http://www.aiida.net.
     working/calculations
     working/workflows
 
-
 .. toctree::
     :maxdepth: 1
     :caption: Working with AiiDA
@@ -77,11 +76,10 @@ The software is available at http://www.aiida.net.
 
 .. toctree::
     :maxdepth: 1
-    :caption: Development
+    :caption: For AiiDA developers
     :hidden:
 
     developer_guide/index
-
 
 .. toctree::
     :maxdepth: 1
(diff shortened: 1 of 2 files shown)


docs/source/index.rst
```diff
@@ -77,12 +76,11 @@ The software is available at http://www.aiida.net.
...
-    :caption: Development
+    :caption: For AiiDA developers
     :hidden:
 
     developer_guide/index
 
-
 .. toctree::
     :maxdepth: 1
     :caption: Tutorials
```
sphuber: Should we also include the sections of `Tutorials` in this restructuring? Moving their content to `StructureData` and `UpfData` respectively? And the `plugin development` sectino should be moved to the temporary `aiida-plugins`, as I think it should be moved out of the `aiida-core` docs

giovannipizzi: I suggest that I mark this in the respective issues, so at least this restructuring is done and we can work on this in the next days during the coding days.
In particular moving tutorials into the section requires probably some thought.

giovannipizzi: For the `plugin development` section, is there something that I have missed, apart for the section that I moved below under the title `aiida-plugins [Section to be moved]`?

giovannipizzi: Added comments for #3019 and #3022 (for the part on `Tutorials`)

PR #3042, opened 2019-06-19
Docs: add virtualenv to quick-install; add python3-dev as dependency.

### LINKED ISSUE #3013: Discuss virtualenv in quick installation page
Also in the quick installation it is important (even if it makes the quick installation less quick) to discuss how to create a new venv and enter it.

borellim:
Fix #3013

Note that this is tested with Ubuntu 16.04, not other OSs.


borellim:
@d-tomerini Would you mind reviewing this again? I added some more things (see commit messages)... Thanks!

sphuber:
Thanks to both

Issue #2918, opened 2019-05-23
decide when to point default documentation back to `latest`


ltalirz:
the new documentation is getting to a useful state and, in several regards, starts to be better than the old one.

I propose something very simple:
  * put in bold on the landing page the version of AiiDA that a documentation refers to (for develop take the `__version__`), both for the current "stable" documentation (release_v0.12.4) and "latest"
  * start pointing readthedocs.org to "latest" by default.

@sphuber @giovannipizzi do you agree?

giovannipizzi:
I agree!

sphuber:
Me too :+1:

borellim:
Would something like this be OK?
![Schermata del 2019-06-19 17-42-56](https://user-images.githubusercontent.com/6683587/59780121-cdbf9280-92b9-11e9-82c4-8679ef6262b2.png)

This is achieved by adding this line:
```rst
**aiida-core version:** |release|
```

I wasn't able to make it all bold, nor bold and italics.

ltalirz:
That's already much better, although still not very visible.
For inspiration see
https://docs.sqlalchemy.org/en/12/
https://bokeh.pydata.org/en/1.0.4/

ltalirz:
Already pointed back rtd to latest. After merging @borellim 's PR this can be closed

PR #3057, opened 2019-06-20
Add aiida-core version to docs home page (1.0 branch).

borellim:
Addresses issue #2918 for the 1.0 branch.


ltalirz:
@borellim have you figured out the cause?
If not, you might want to open an issue here: https://github.com/lemurheavy/coveralls-public/issues

PR #3063, opened 2019-06-20
fix a bug for printing scheduler outputs

zhubonan:
Replace `format` with string concatenation to avoid interpreting unicode under python2.


zhubonan:
The problem is that the scheduer error is unicode but format tries to decode it into ascii string. This is a problem for Python 2 only. When the scheduler error is just ASCII it is fine (most cases) but error will be thrown if it does contain Unicode.

Using concatenation will automatically convert the result to unicode.

ltalirz:
I see - in this case, would keeping `.format` and simply prefixing the string with `u` work as well?
This would make your fix a bit more explicit - otherwise, people reading this won't know why the split is necessary and might change it back.

zhubonan:
Indeed that should work as well. I will change it back that way.

ltalirz:
Hm... I just tried and am not sure...

ltalirz:
```
In [40]: s="123"

In [41]: s2=u"34ü'3ée"

In [42]: "abc{}".format(s)
Out[42]: 'abc123'

In [43]: "abc{}".format(s2)
--------------------------------------------------------
UnicodeEncodeError     Traceback (most recent call last)
<ipython-input-43-550dc6cef559> in <module>()
----> 1 "abc{}".format(s2)

UnicodeEncodeError: 'ascii' codec can't encode character u'\xfc' in position 2: ordinal not in range(128)

In [44]: u"abc{}".format(s2)
Out[44]: u"abc34\xfc'3\xe9e"
```

ltalirz:
Ah, ok, that's an issue of the terminal though
```
In [48]: s2
Out[48]: u"34\xfc'3\xe9e"
```

zhubonan:
Yes when you 'print', it will be displayed correctly

ltalirz:
You're right - just for [future reference](http://www.macfreek.nl/memory/Unicode_Strings_in_Python)

> The repr() function also encodes from Unicode strings to byte strings in Python 2, but is generally safe, since the output of repr() should be valid Python, which is generally ASCII-encoded in Python 2.

That's what is happening in
```
In [48]: s2
Out[48]: u"34\xfc'3\xe9e"
```
and why
```
In [49]: print(s2)
Out[49]: 34ü'3ée
```
works fine (i.e. it had nothing to do with the encoding of my terminal).

ltalirz:
@zhubonan by the way, you haven't pushed yet ;-)

zhubonan:
Just changed it to `u'{}'.format()`. 
BTW Should I update the branch?

sphuber:
Thanks guys!

PR #3072, opened 2019-06-23
Move import in `aiida.cmdline.commands` to speed up `verdi`

sphuber:
The `aiida.cmdline.utils.migration` is importing quite some modules that
were slowing the usual loading time of 200 ms to near 350 ms.

### DIFF
--- a/aiida/cmdline/commands/cmd_export.py
+++ b/aiida/cmdline/commands/cmd_export.py
@@ -24,7 +24,6 @@ from aiida.cmdline.params import arguments
 from aiida.cmdline.params import options
 from aiida.cmdline.utils import decorators
 from aiida.cmdline.utils import echo
-from aiida.cmdline.utils import migration
 
 
 @verdi.group('export')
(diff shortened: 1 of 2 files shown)


sphuber:
We have indeed been talking about this with Gio and we think the best option would be to have some tests on Jenkins. This would be part of a larger set of performance tests. Now we have no clue when we introduce changes that might affect the performance in big ways that are not trivial.

For example, on the Django JSONB branch we refactored in addition the backend node interface with respect to attributes and extras. To make sure we did not affect performance too much, I was running my cif clean setup with this. Turns out it became slower by an order of magnitude :sweat:  This is something we would never have noticed without these tests, so I think they are becoming more and more critical.

However, this is still no guarantee that we would find it. I just spend the whole weekend profiling and hunting this down and the source is due to somewhere in the stack me calling the pymatgen `CifParser` with a stored `Float` node instead of a normal float. The reason this now appears is due to a small change in behavior of the backend node, that we introduced _to make things more efficient_. Long story short, without performance tests we would never have caught this, but even with tests, it is still not guaranteed

CasperWA:
This is interesting. I thought long about the placement of imports for the refactoring of the import/export module, and it seems there could be some optimization there as well, by moving imports into functions.

I will take add a comment to the PR.

PR #3077, opened 2019-06-24
Bump version to `v1.0.0b4`

sphuber:
That's what you get for making releases late at night on a Sunday :sweat:


sphuber:
Is there anything else you can see I might have missed in the previous PR? I did the following:
* Update change log
* Update contributors
* Run helpers scripts to fix headers
* Update version number

That should be it, right?

ltalirz:
seems fine!

PR #3080, opened 2019-06-24
Fix `UpfData.get_upf_groups`

chrisjsewell:
Change filter attribute from `type` -> `type_string`, and remove duplicates.

### DIFF
--- a/aiida/backends/tests/orm/data/test_upf.py
+++ b/aiida/backends/tests/orm/data/test_upf.py
@@ -19,6 +19,7 @@ import tempfile
 import shutil
 import os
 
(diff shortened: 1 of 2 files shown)


sphuber:
Thanks Chris, looks good, but it would be good if you could add a small test

sphuber:
@chrisjsewell I took your branch and added the tests on top. Hope you don't mind

chrisjsewell:
> @chrisjsewell I took your branch and added the tests on top. Hope you don't mind

Yeh no problem ta, I probably was going to get round to it any time soon!

PR #3081, opened 2019-06-25
Only print `retrieve_singlefile_list` deprecation when necessary

sphuber:
While the functionality still exists, the engine will keep calling the
`get_retrieve_singlefile_list` method and try to set them from the
calculation info object, so the warnings were printed even when the user
was not actually using it. Make sure that the relevant methods are only
called when the user explicitly sets the `retrieve_singlefile_list`.

### DIFF
--- a/aiida/orm/nodes/process/calculation/calcjob.py
+++ b/aiida/orm/nodes/process/calculation/calcjob.py
@@ -373,7 +373,6 @@ class CalcJobNode(CalculationNode):
             Will be removed in `v2.0.0`, use
             :meth:`aiida.orm.nodes.process.calculation.calcjob.CalcJobNode.get_retrieve_temporary_list` instead.
         """
-        warnings.warn('method is deprecated, use `get_retrieve_temporary_list` instead', AiidaDeprecationWarning)  # pylint: disable=no-member
         return self.get_attribute(self.RETRIEVE_SINGLE_FILE_LIST_KEY, None)
 
     def set_job_id(self, job_id):
--- a/aiida/engine/processes/calcjobs/calcjob.py
+++ b/aiida/engine/processes/calcjobs/calcjob.py
@@ -294,7 +294,8 @@ class CalcJob(Process):
                 raise PluginInternalError(
                     "[presubmission of calc {}] retrieve_singlefile_list subclass problem: {} is "
                     "not subclass of SinglefileData".format(self.node.pk, file_sub_class.__name__))
-        self.node.set_retrieve_singlefile_list(retrieve_singlefile_list)
+        if retrieve_singlefile_list:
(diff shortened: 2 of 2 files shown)


aiida/orm/nodes/process/calculation/calcjob.py
```diff
@@ -373,7 +373,6 @@ def get_retrieve_singlefile_list(self):
             Will be removed in `v2.0.0`, use
             :meth:`aiida.orm.nodes.process.calculation.calcjob.CalcJobNode.get_retrieve_temporary_list` instead.
         """
-        warnings.warn('method is deprecated, use `get_retrieve_temporary_list` instead', AiidaDeprecationWarning)  # pylint: disable=no-member
```
CasperWA: Could we move this warning into the other file then, possibly? - In the new (and very much needed) if block.

sphuber: The thing is, this function is most likely never called by the user and if they do, they will have called the `set_retrieve_singlefile_list`. But even that function they never really call, because that is done automatically by the engine. The problem is that the `set_retrieve_singlefile_list` call I can make conditional in the engine only when the user specifies it. But the `get_retrieve_singlefile_list` always has to be called by the engine, so this warning would always show up

CasperWA: So let's hit what we can, by the philosophy that if a user manually sets the `retrieve_singlefile_list` a warning occurs, elsewhere, not?

CasperWA: At least then, if a user forces the use of this, they are notified. Or did I misunderstand and it will be flashed every time the engine calls it as well?

sphuber: that's the thing, the warning does not make a distinction between who calls it. But in any case, the engine is the only one who is supposed to call it anyway. The user just specified it in `CalcInfo.retrieve_singlefile_list` which they return from the `CalcJob.prepare_for_submission`.

CasperWA: Hmm. Okay. Then either we create a proxy-function that includes the warning, which the engine can go around, or we remove the message as you have done, I guess? And I don't like the first option at all, honestly.

CasperWA:
@sphuber, unless you have a novel idea on how to keep the warning in some way, I will merge this :)

sphuber:
No there is no other way, but I think it is fine. Even if users were using this mechanism, they never were calling `get_retrieve_singlefile_list` themselves anyway. And if they use the concept, they *will* see the warning coming from the `set_retrieve_singlefile_list` that the engine will call for them. So merge away!

PR #3084, opened 2019-06-25
Fixed a few deprecation warnings

espenfl:
Tried to fix a few deprecation warnings related to `aiida_core` that is a bit annoying when running `pytest`. Disable with `pylint` does not fix this and we would like not to turn of warnings.


espenfl:
See also:
https://github.com/aiidateam/plumpy/pull/118

and:
https://github.com/materialsproject/pymatgen/issues/1511

PR #3090, opened 2019-06-26
Migrating node attributes and extras to JSONB for the Django backend

### LINKED ISSUE #864: sqlalchemy: very expensive recursive check for dates
The following function contains a rather expensive recursive procedure, whose only purpose (if I understand correctly) is to try and replace strings by datetime objects:

https://github.com/aiidateam/aiida_core/blob/17484b5920ecc4f6906c73867cfaf6b16658e219/aiida/backends/sqlalchemy/utils.py#L172-L198

Any ideas for a better way of achieving this?

In my use case, which involves importing a large number of .cif files into the database, this rather benign check is the single most expensive operation (it causes millions of `is_instance` calls).
In the total run time of the import process it's just 20%, so it's not absolutely critical for the moment, but still...

### LINKED ISSUE #1559: Add sorting functionality for attributes.state in QB in django
QB supports sorting by attributes.state in SQLA but not in django backend.
Check if it could also be supported in django?

### LINKED ISSUE #2786: [Testing] Test the migration speed of Django from the EAV schema to JSONB
Given the implementation found at aiida_core/django_jsonb branch, we should take and report the time that is needs for a complete migration of a database of a reasonable size (5m to 10M of nodes).

This is related to issue #2762 and can be inspired by that issue.

Based on the performance we can see if the migration needs improvement.

### LINKED ISSUE #2788: Remove support for datetime objects in attributes and extras
Currently when the attributes and extras are stored, there is a datetime_to_isoformat conversion and the opposite action (isoformat_to_datetime) is performed when a JSON field is read.

If we remove datetime serialization (issue #864) the implementation of Django JSONB should be changed and adapted as the SQLA JSONB.

We should see in both issues what we do with DbSetting that also contains dates and is stored in JSONB.

### LINKED ISSUE #2789: [Testing] Test that the migrated database works as expected
Since a lot of information are migrated (attributes, extras and settings) it is good to get a project that has started in Django EAV and migrate it to Django JSONB and see if the calculations continue correctly and if the engine works as expected with the migration of the settings.

### LINKED ISSUE #2815: Simplify export file format after JSONB migration is performed
Currently there is a convoluted way of storing datetimes in the export file, to mark them as date times and not as strings (the attributes dictionary is "duplicated", having `None` values in every place that maps a standard data type, except for value related a date time, where a special string is used). This is the `node_attributes_conversion` field, see [docs](https://aiida-core.readthedocs.io/en/stable/import_export/main.html#data-json)

If in #2788 we remove support for datetimes, we can safely remove this dictionary from the data.json - this will reduce the export file size, simplify the code and most probably also make import/export faster.

### LINKED ISSUE #2821: Fix the `extras_keys`/`extras_items` methods of Node
This line
https://github.com/aiidateam/aiida_core/blob/16f105ecd820eb04358ac67c4c954cd0db93ad36/aiida/orm/nodes/node.py#L623
and this line
https://github.com/aiidateam/aiida_core/blob/16f105ecd820eb04358ac67c4c954cd0db93ad36/aiida/orm/nodes/node.py#L615
are wrong. 
First, `extras` is a property and should not be called.
In any case, we shouldn't go via `extras` but via the methods to get only the keys, for instance. See, as a reference, the implementation for attributes:
https://github.com/aiidateam/aiida_core/blob/16f105ecd820eb04358ac67c4c954cd0db93ad36/aiida/orm/nodes/node.py#L496

This also means this is not tested, so tests should be added for these methods (and maybe we might want to check on coverage.io/coveralls which other methods are currently untested).

### LINKED ISSUE #2854: Add data migration for old calculation jobs
`JobCalculation` nodes in old databases will currently automatically be migrated to `CalcJobNode` instances. However, these nodes still have some old attributes that are not migrated and miss many attributes that they *would* have had, had they been run with `aiida-core==1.0.0`, e.g. the process state and process label attributes.
We should consider adding a data migration that tries as best we can to transform these old calculations into something compatible.

1. The old `DbCalcJobState` was dropped, but the proxy attribute `state` remains there. The `CalcJobNode.get_state` method will try to cast this into the new `CalcJobState` enum, but the old calculations can contain values that are not supported by this enum, causing a `ValueError` to be thrown. In the new system, this `state` attribute is a temporary sub-state, in addition to the process state, that only applies to `CalcJob` processes, while they are active and are deleted once the process terminates. Therefore, I propose we simply delete this state for all `CalcJobNode`s that are in a terminal state. We should be careful that we only apply this to terminated process nodes, because there might be active ones during the migration that we do not want to touch.
2. All `CalcJobNodes` should have a `process_state`, `process_label` and optionally an `exit_status`. The `process_label` is currently set as the name of the process class or function name (in the case of a process function). I am not sure if an equivalent was stored on the old node, but otherwise we could maybe use the value in the `process_type`, which should have been added during existing migrations. The problem here will be how to set the `process_state`. The old calculations only had the `state` attribute, the old job calc state, but those values do not map one-to-one on the new process states. For example, any of the old "failed" states, such as `SUBMISSIONFAILED`, or `FAILED`: should they become `Excepted` or `Finished` with a non-zero exit status. But in the case of the latter, what exit status should be used?

This can be summed up in the following table which should tell how to translate the `state` attribute to the correct `process_state` and `exit_status`:

Old state            | Process state  | Exit status | Process status
---------------------|----------------|---------------|-----------------------
`NEW`                | `Killed`      |      `None`  | `Legacy JobCalculation with state 'NEW' `
`TOSUBMIT`           | `Killed`     |      `None`  |  `Legacy JobCalculation with state 'TOSUBMIT' `
`SUBMITTING`         | `Killed`     |      `None`  |  `Legacy JobCalculation with state 'SUBMITTING' `
`WITHSCHEDULER`      | `Killed`     |      `None`  |  `Legacy JobCalculation with state 'WITHSCHEDULER' `
`COMPUTED`           | `Killed`     |      `None`  |  `Legacy JobCalculation with state 'COMPUTED' `
`RETRIEVING`         | `Killed`     |      `None`  |  `Legacy JobCalculation with state 'RETRIEVING' `
`PARSING`            | `Killed`     |      `None`  |  `Legacy JobCalculation with state 'PARSING' `
`FINISHED`           | `Finished`     |      0    |
`SUBMISSIONFAILED`   | `Excepted`        |      `None`  |  `Legacy JobCalculation with state 'SUBMISSIONFAILED' `
`RETRIEVALFAILED`    | `Excepted`        |      `None`  |  `Legacy JobCalculation with state 'RETRIEVALFAILED' `
`PARSINGFAILED`      | `Excepted`        |      `None`  |  `Legacy JobCalculation with state 'PARSINGFAILED' `
`FAILED`             | `Finished`        |      1    |
`IMPORTED`           |       -        |      -    |

Notes;
* Note the `IMPORTED` state was never actually stored in the `state` attribute, so we do not have to consider it in the migration.
* All the actives states will be set to `Killed`. Before the migration is performed, the user will be warned (note this still needs to be added) to first finish all running calculations and workflows. Then stop the daemon and perform the migration. In a sense then, if we find these "active" calculation states during the migration, it means the user either has ignored the warning and is effectively telling us to kill them or the calculation got stuck in an active state and was lost, in which case the `Killed` state is also appropriate.
* For the `process_label`, inferring the original class name is difficult, although the `process_type` could potentially be used. However, since this complicated to do in SQL we simply put `Legacy JobCalculation`.

### LINKED ISSUE #3067: Test the efficiency of the new Django JSONB implementation
Note that especially we need to guarantee that appending to attributes while the node is unstored is efficient and scales linearly. The old implementation required [a fix](https://github.com/aiidateam/aiida_core/pull/733/files#diff-c251db38fcdbaac0d8bebe068face5cfR839) once to enable this.

sphuber:
Fixes #2821 
Fixes #2788 
Fixes #2789 
Fixes #2786 
Fixes #864 
Fixes #3067 
Fixes #2815 
Fixes #2854 
Fixes #1559 

The main purpose of this PR is to migrate the old EAV schema for node attributes and extras for the Django backend. The change of the models and the database migrations have been implemented in the first commit by @szoupanos . But this change allowed to homogenize and streamline a lot of the rest of the code, since now both backends use JSONB for node attributes and extras. For example this now means that the attribute cache mechanism, that allowed one to store attributes, before the node itself was stored, can be removed. This cache is now in a sense delegated to the database model, where the attributes and extras column is represented as a dictionary. In addition, the restriction of only being able to store extras after a node was stored could also be lifted. To simplify and robustify the access of attributes and extras, the `BackendNode` interface has been significantly improved and tested.

Then an unrelated commit was added to migrate old `state` attributes from legacy `JobCalculations`. We decided to do this on this branch, since the migration was made a lot easier after both backends used JSONB, meaning the SQL used in the migration was identical for both backends.

The final step was to add migrations for existing export archives to account for the two database migrations added in this PR, while upping the version number to `v0.6`. The migrations will find `state` attributes of `CalcJobNodes` and migrate the data. Serialized datetime strings in attributes are also updated to reattach the timezone information, which was originally never included since everything was first converted to UTC before converting.


ltalirz:
Fantastic! Killing those issues like flies ;-)

PR #3091, opened 2019-06-27
Add the `verdi process call-root` command

### LINKED ISSUE #1150: Add 'verdi node getcallroot' command
Add a command which follows the CALL links to the top-level caller. This can be used to easily identify which workflow a particular node comes from.

sphuber:
Fixes #1150 

This command will go up the call stack for a given process node and
return the root caller, if there is one.

### DIFF
--- a/aiida/backends/tests/cmdline/commands/test_process.py
+++ b/aiida/backends/tests/cmdline/commands/test_process.py
@@ -411,6 +411,57 @@ class TestVerdiProcessListWarning(AiidaTestCase):
         self.assertTrue(any([warning_phrase in line for line in get_result_lines(result)]))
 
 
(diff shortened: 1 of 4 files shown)


giovannipizzi:
Wouldn't it be even more useful to implement instead a `verdi callstack` with the whole stack from root to current? (and maybe with an optional flag to see also called processes)? This would implicitly give the required info, but also much more

sphuber:
The callstack is already implemented in `verdi process status`. This command serves simply to find the call-root. Its output is not graphical in anyway, it just prints a pk. The idea is that the result of this command can be piped into another one, e.g.:
```
verdi process status $(verdi process call-root 901)
```

PR #3093, opened 2019-06-27
Enable "deep" reconstruction of yaml deserialization mapping constructor

### LINKED ISSUE #3092: YAML serialization round trip of mixed nested `AttributeDicts` loses information
The custom YAML serializer and deserializer in `aiida.orm.utils.serializer` is badly broken for nested `AttributeDicts` that also contain normal dictionaries. If all nested dictionaries are instances of `AttributeDict` the round trip does not lose information. But any normal dictionary in the `AttributeDict` will lose its content in the round trip. Demonstration:
```
#!/usr/bin/env runaiida
from aiida.common.extendeddicts import AttributeDict
from aiida.orm.utils import serialize

dictionary = {'nested': {'dict': 'string', 'value': 1}}
adict = AttributeDict(dictionary)
adict['nested']['normal'] = {'test': 'ting'}

serialized = serialize.serialize(adict)
deserialized = serialize.deserialize(serialized)

print('ORIGINAL', adict)
print('SERIALIZED', serialized)
print('DESERIALIZED', deserialized)
```
This will yield:
```
('ORIGINAL', AttributeDict({'nested': AttributeDict({'dict': 'string', 'value': 1, 'normal': {'test': 'ting'}})}))
('SERIALIZED', '!aiida_attributedict\nnested: !aiida_attributedict\n  dict: string\n  normal: {test: ting}\n  value: 1\n')
('DESERIALIZED', AttributeDict({'nested': AttributeDict({'dict': 'string', 'value': 1, 'normal': AttributeDict({})})}))
```

Note that the content of the `normal` nested dictionary is still present in the serialized yaml dump, but it is lost in the deserialized dictionary.

Since `AttributeDicts` are used heavily through the code and the serializing is used to save `Process` checkpoints, this bug risks losing information of the `Process`, such as its `context` contents, in the checkpoint. As soon as the process is reloaded from the checkpoint, for example after a daemon restart, parts of the context can be lost completely.

sphuber:
Fixes #3092 

The yaml mapping constructor in `aiida.orm.utils.serialize` for
`AttributeDicts` was not using the `deep=True` argument to recursively
reconstruct nested dictionaries. Combined with the recursive constructor
of the `AttributeDict` class, this would result in the loss of content
of normal dictionaries nested in an attribute dict in a serialization
round-trip.

Since this serialization is used to store `Process` instances as
checkpoints in the database, from which the daemon reloads the instance
into memory after a restart, this bug resulted in the loss of
information in those checkpoints. For example, if one stored a normal
dictionary in the context of a `WorkChain` instance, which is an
`AttributeDict`, the contents of that nested normal dictionary would not
be restored when the `WorkChain` was reloaded from the checkpoint
causing all sorts of exceptions in the business logic of the work chain.

### DIFF
--- a/aiida/backends/tests/common/test_serialize.py
+++ b/aiida/backends/tests/common/test_serialize.py
@@ -16,10 +16,9 @@ from aiida import orm
 from aiida.orm.utils import serialize
 from aiida.backends.testbase import AiidaTestCase
 
-# pylint: disable=missing-docstring
-
 
 class TestSerialize(AiidaTestCase):
+    """Tests for the YAML serializer and deserializer."""
 
     def test_serialize_round_trip(self):
         """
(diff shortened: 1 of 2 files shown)


sphuber:
This was a really nasty one. The only real change is setting `deep=True` in the mapping constructor of the `aiida.orm.utils.serialize` module. All the other changes is simple docstring fixes and code reorganzation. This tiny omission that caused basically all work chains to except after reloading from checkpoint due to a daemon restart.

PR #3096, opened 2019-06-28
Respect the minimum polling interval for scheduler updates

### LINKED ISSUE #2535: Minimum scheduler polling interval not respected
The `JobManager` which is supposed to limit the number of scheduler queries by batching the requests and not exceeding the minimum poll interval, does in fact not respect this interval.

sphuber:
Fixes #2535 

The `aiida.engine.processes.calcjobs.manager.JobsList` container was
introduced for two purposes, related to managing running calculation
jobs in high-throughput mode:

 * It bundles the scheduler update calls for all active calculation jobs
   for given authentication info (the comination of a computer and user)
 * It ensures that consecutive scheduler update calls are separated at
   least by a time that is given by the `get_minimum_job_poll_interval`
   as defined by the computer of the authentication info.

However, the final requirement was not being respected. The problem was
twofold. The internal attribute `_last_updated` that records the
timestamp of the last update was not being updated after the scheduler
was queried for a status update.

The use of the `RefObjectStore` to create and delete these `JobsLists`
instances, however, was an even bigger problem. The reference store
would delete the `JobsLists` instance as soon as no-one held a reference
to it any more, since they are created through the
`request_job_info_request` context manager of the `JobManager` that each
runner has. As soon as no requests were active, the object was deleted
and with it the record of the last time the scheduler was queried. The
next time a request comes in a new `JobsList` would be created that
straight away call the scheduler as it has no recollection the last time
it was called for the given authinfo, if at all. This would result in
the minimum poll interval essentially never being respected.

Fundamentally, the problem lies in the fact that the data that needs to
be persistent, the "last updated" timestamp, was being stored in the
container `JobsList` that by implementation was made non-persistent.
The solution is to simply make the `JobsList` instances live for as long
as the python interpreter is alive. Each daemon runner creates a single
`JobManager` instance on start up and this will now only create a new
`JobsList` once for each authinfo and will keep returning the same
instance for the rest of its lifetime.

### DIFF
--- a/aiida/engine/processes/calcjobs/manager.py
+++ b/aiida/engine/processes/calcjobs/manager.py
@@ -7,61 +7,82 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-"""
-Module containing utilities and classes relating to job calculations running
-on systems that require transport.
-"""
+"""Module containing utilities and classes relating to job calculations running on systems that require transport."""
 from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
 
 import contextlib
-from functools import partial
 import time
 
 from six import iteritems, itervalues
 from tornado import concurrent, gen
 
 from aiida import schedulers
-from aiida.common import exceptions
-from ...utils import RefObjectStore
+from aiida.common import exceptions, lang
+from aiida.common.log import AIIDA_LOGGER
 
 __all__ = ('JobsList', 'JobManager')
 
 
 class JobsList(object):  # pylint: disable=useless-object-inheritance
-    """
-    A list of submitted jobs on a machine connected to by transport based on the
-    authorisation information.
+    """Manager of calculation jobs submitted with a specific ``AuthInfo``, i.e. computer configured for a specific user.
+
+    This container of active calculation jobs is used to update their status periodically in batches, ensuring that
+    even when a lot of jobs are running, the scheduler update command is not triggered for each job individually.
+
+    In addition, the :py:class:`~aiida.orm.computers.Computer` for which the :py:class:`~aiida.orm.authinfos.AuthInfo`
+    is configured, can define a minimum polling interval. This class will guarantee that the time between update calls
+    to the scheduler is larger or equal to that minimum interval.
+
+    Note that since each instance operates on a specific authinfo, the guarantees of batching scheduler update calls
+    and the limiting of number of calls per unit time, through the minimum polling interval, is only applicable for jobs
+    launched with that particular authinfo. If multiple authinfo instances with the same computer, have active jobs
+    these limitations are not respected between them, since there is no communication between ``JobsList`` instances.
+    See the :py:class:`~aiida.engine.processes.calcjobs.manager.JobManager` for example usage.
     """
 
-    def __init__(self, authinfo, transport_queue):
-        """
+    def __init__(self, authinfo, transport_queue, last_updated=None):
+        """Construct an instance for the given authinfo and transport queue.
+
         :param authinfo: The authinfo used to check the jobs list
         :type authinfo: :class:`aiida.orm.AuthInfo`
         :param transport_queue: A transport queue
         :type: :class:`aiida.engine.transports.TransportQueue`
+        :param last_updated: initialize the last updated timestamp
+        :type: float
         """
+        lang.type_check(last_updated, float, allow_none=True)
+
         self._authinfo = authinfo
         self._transport_queue = transport_queue
         self._loop = transport_queue.loop()
+        self._logger = AIIDA_LOGGER.getChild('calcjobs')
 
         self._jobs_cache = {}
-        self._last_updated = None  # type: float
         self._job_update_requests = {}  # Mapping: {job_id: Future}
+        self._last_updated = last_updated
         self._update_handle = None
 
-    def get_minimum_update_interval(self):
+    @property
+    def logger(self):
+        """Return the logger configured for this instance.
+
+        :return: the logger
         """
-        Get the minimum interval that can be expected between updates of the list
-        :return: The minimum interval
+        return self._logger
+
+    def get_minimum_update_interval(self):
+        """Get the minimum interval that should be respected between updates of the list.
+
+        :return: the minimum interval
         :rtype: float
         """
         return self._authinfo.computer.get_minimum_job_poll_interval()
 
-    def get_last_updated(self):
-        """
-        Get the timestamp of when the list was last updated as produced by `time.time()`
+    @property
+    def last_updated(self):
+        """Get the timestamp of when the list was last updated as produced by `time.time()`
 
         :return: The last update point
         :rtype: float
@@ -70,10 +91,9 @@ class JobsList(object):  # pylint: disable=useless-object-inheritance
 
     @gen.coroutine
     def _get_jobs_from_scheduler(self):
-        """
-        Get the current jobs list from the scheduler
+        """Get the current jobs list from the scheduler.
 
-        :return: A dictionary of {job_id: job info}
+        :return: a mapping of job ids to :py:class:`~aiida.schedulers.datastructures.JobInfo` instances
         :rtype: dict
         """
         with self._transport_queue.request_transport(self._authinfo) as request:
@@ -89,7 +109,11 @@ class JobsList(object):  # pylint: disable=useless-object-inheritance
                 kwargs['jobs'] = self._get_jobs_with_scheduler()
 
             scheduler_response = scheduler.get_jobs(**kwargs)
+
+            # Update the last update time and clear the jobs cache
+            self._last_updated = time.time()
             jobs_cache = {}
+            self.logger.info('AuthInfo<{}>: successfully retrieved status of active jobs'.format(self._authinfo.pk))
 
             for job_id, job_info in iteritems(scheduler_response):
                 # If the job is done then get detailed job information
@@ -107,12 +131,10 @@ class JobsList(object):  # pylint: disable=useless-object-inheritance
 
     @gen.coroutine
     def _update_job_info(self):
-        """
-        Update all of the job information objects for a given authinfo, that is to say for
-        all the jobs on a particular machine for a particular user.
+        """Update all of the job information objects.
 
-        This will set the futures for all pending update requests where the corresponding job
-        has a new status compared to the last update.
+        This will set the futures for all pending update requests where the corresponding job has a new status compared
+        to the last update.
         """
         try:
             if not self._update_requests_outstanding():
@@ -135,16 +157,16 @@ class JobsList(object):  # pylint: disable=useless-object-inheritance
 
     @contextlib.contextmanager
     def request_job_info_update(self, job_id):
-        """
-        Request job info about a job when it next changes it's job state.  If the job is not
-        found in the jobs list at the update the future will resolve to None.
+        """Request job info about a job when the job next changes state.
 
-        :param job_id: The job identifier
-        :return: A future that will resolve to a JobInfo object when the job changes state
+        If the job is not found in the jobs list at the update, the future will resolve to `None`.
+
+        :param job_id: job identifier
+        :return: future that will resolve to a `JobInfo` object when the job changes state
         """
         # Get or create the future
         request = self._job_update_requests.setdefault(job_id, concurrent.Future())
-        assert not request.done(), "The future should be no be in the done state"
+        assert not request.done(), 'Expected pending job info future, found in done state.'
 
         try:
             self._ensure_updating()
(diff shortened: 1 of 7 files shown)


aiida/engine/processes/calcjobs/manager.py
```diff
@@ -7,61 +7,81 @@
...
+    """A list of submitted jobs on a machine connected to by transport based on the authorisation information.
+
+    This container of active calculation jobs is used to update their status periodically in batches. This ensure that
+    even when a lot of jobs are running the scheduler update command is not triggered for each job individually.
+
+    In addition, the `Computer` for which the `AuthInfo` is configured, can define a minimum polling interval. This
+    class will guarantee that the time between update calls to the scheduler is larger or equal to that minimum
+    interval.
+
+    Note that since each instance operates on a specific `AuthInfo`, the guarantees of batching scheduler update calls
```
ltalirz: Note that there will be one `JobsList` instance *per daemon worker* and *per `AuthInfo`*.
I.e. when, hypothetically, configuring the same computer for multiple users, and running jobs for different users simultaneously, the minimum polling interval would be respected only *per user*.

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -70,10 +90,9 @@ def get_last_updated(self):
 
     @gen.coroutine
     def _get_jobs_from_scheduler(self):
-        """
-        Get the current jobs list from the scheduler
+        """Get the current jobs list from the scheduler.
 
-        :return: A dictionary of {job_id: job info}
+        :return: a dictionary of {job_id: job info}
```
ltalirz: is the "job info" of a particular format?
either put a string or some info about the format

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -89,7 +108,11 @@ def _get_jobs_from_scheduler(self):
                 kwargs['jobs'] = self._get_jobs_with_scheduler()
 
             scheduler_response = scheduler.get_jobs(**kwargs)
+
+            # Update the last update time and clear the jobs cache
+            self._last_updated = time.time()
```
ltalirz: just wondering: are there different standard ways of getting the time?
If not, it's fine - otherwise this might be something to standardize in aiida.common

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -135,16 +156,16 @@ def _update_job_info(self):
 
     @contextlib.contextmanager
     def request_job_info_update(self, job_id):
-        """
-        Request job info about a job when it next changes it's job state.  If the job is not
-        found in the jobs list at the update the future will resolve to None.
+        """Request job info about a job when it next changes it's job state.
```
ltalirz:
```suggestion
        """Request job info about a job when the job next changes state.
```

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -135,16 +156,16 @@ def _update_job_info(self):
...
+
+        If the job is not found in the jobs list at the update the future will resolve to `None`.
```
ltalirz:
```suggestion
        If the job is not found in the jobs list at the update, the future will resolve to `None`.
```

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -135,16 +156,16 @@ def _update_job_info(self):
...
 
-        :param job_id: The job identifier
-        :return: A future that will resolve to a JobInfo object when the job changes state
+        :param job_id: job identifier
+        :return: future that will resolve to a `JobInfo` object when the job changes state
         """
         # Get or create the future
         request = self._job_update_requests.setdefault(job_id, concurrent.Future())
-        assert not request.done(), "The future should be no be in the done state"
+        assert not request.done(), 'The future should be no be in the done state'
```
ltalirz: could this be more explicit? e.g. something like
```suggestion
        assert not request.done(), 'Expected pending job info future, found in done state.'
```

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -153,14 +174,14 @@ def request_job_info_update(self, job_id):
             pass
 
     def _ensure_updating(self):
-        """
-        Ensure that we are updating the job list from the remote resource.
+        """Ensure that we are updating the job list from the remote resource.
```
ltalirz: this is not only ensuring things but also performing the update as well, right?
should this be called something like `_update_loop` ?

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -189,63 +211,72 @@ def _has_job_state_changed(old, new):
         return old.job_state != new.job_state or old.job_substate != new.job_substate
 
     def _get_next_update_delay(self):
-        """
-        Calculate when we are next allowed to call the scheduler get jobs command
-        based on when we last called it, how long has elapsed and the minimum given
-        update interval.
+        """Calculate when we are next allowed to call the scheduler get jobs command.
```
ltalirz:
```suggestion
        """Calculate when we are next allowed to poll the scheduler.
```

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -189,63 +211,72 @@ def _has_job_state_changed(old, new):
...
 
-        :return: The delay (in seconds) for when it's safe to call the get jobs command
+        This delay will be calculated as the minimum safe interval that needs to be guaranteed, as defined by the
```
ltalirz:
```suggestion
        This delay will be calculated as the minimum polling interval defined by the
```

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -189,63 +211,72 @@ def _has_job_state_changed(old, new):
...
+        authentication info for this instance, minus the last time we updated the list.
+
+        :return: delay (in seconds) after which the minimum update interval has passed since last update
```
ltalirz:
```suggestion
        :return: delay (in seconds) after which the scheduler may be polled again
```

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -189,63 +211,72 @@ def _has_job_state_changed(old, new):
...
         :rtype: float
         """
-        if self._last_updated is None:
+        if self.last_updated is None:
+            self.logger.info('AuthInfo<{}>: next scheduler update will be in {} seconds'.format(self._authinfo.pk, 0))
```
ltalirz:
```suggestion
            self.logger.info('AuthInfo<{}>: next scheduler update allowed in {} seconds'.format(self._authinfo.pk, 0))
```

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -189,63 +211,72 @@ def _has_job_state_changed(old, new):
...
             # Never updated, so do it straight away
             return 0.
 
-        # Make sure to actually 'get' it here, so that if the user changed it
-        # between times we use the current value
+        # Make sure to actually 'get' it here, so that if the user changed it between times we use the current value
```
ltalirz:
```suggestion
        # Make sure to actually 'get' the minimum interval here, in case the user changed since last time
```

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -189,63 +211,72 @@ def _has_job_state_changed(old, new):
...
         minimum_interval = self._authinfo.computer.get_minimum_job_poll_interval()
```
ltalirz: just wondering: is this a database query? I guess this might be called a lot?

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -189,63 +211,72 @@ def _has_job_state_changed(old, new):
...
-        elapsed = time.time() - self._last_updated
+        elapsed = time.time() - self.last_updated
+
+        delay = max(minimum_interval - elapsed, 0.)
+        self.logger.info('AuthInfo<{}>: next scheduler update will be in {} seconds'.format(self._authinfo.pk, delay))
```
ltalirz:
```suggestion
        self.logger.info('AuthInfo<{}>: next scheduler update allowed in {} seconds'.format(self._authinfo.pk, delay))
```

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -189,63 +211,72 @@ def _has_job_state_changed(old, new):
...
         """
         return [str(job_id) for job_id, _ in self._job_update_requests.items()]
 
 
-class JobManager(object):  # pylint: disable=useless-object-inheritance
-    """
-    A manager for jobs on a (usually) remote resource such as a supercomputer
-    """
+class JobManager(object):
+    """A manager for jobs on a (usually) remote resource such as a supercomputer."""
```
ltalirz:
```suggestion
    """A manager for jobs on a ``Computer``."""
```

aiida/engine/processes/calcjobs/manager.py
ltalirz: Also, I'm wondering: how often can this be called per calculation?
Is there a risk of overflowing the log with this (even though it's the "INFO" level)?

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -7,61 +7,81 @@
...
+from aiida.common.log import AIIDA_LOGGER
 
 __all__ = ('JobsList', 'JobManager')
 
 
 class JobsList(object):  # pylint: disable=useless-object-inheritance
-    """
-    A list of submitted jobs on a machine connected to by transport based on the
-    authorisation information.
+    """A list of submitted jobs on a machine connected to by transport based on the authorisation information.
```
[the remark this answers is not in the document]
sphuber: I think this line needs to include the `authinfo` to make sure that it does not cover necessarily *all* the jobs on that computer, but only those submitted with the `authinfo` used in the constructor

ltalirz:
Regarding backticks in sphinx, I just checked and it simply prints the contents in italic
https://aiida.readthedocs.io/projects/aiida-core/en/latest/apidoc/aiida.tools.data.html#module-aiida.tools.data.cif

I.e. where that is what you want, that's fine. Often times, it's probably better to use something like ```:py:class`~aiida.orm.Computer` ``` (but may not be necessary everywhere)

aiida/engine/processes/calcjobs/manager.py
sphuber: the `time.time` call just returns a float measured in seconds. Don't think it is worth to standardize this

aiida/engine/processes/calcjobs/manager.py
sphuber: I don't think so, the actual updating is happening in `_update_job_info`. This `_ensure_updating` is really just making sure that, each time it is called, a callback has already been registered with the appropriate delay, to perform the updating. If not, it schedules it, if is already scheduling, it is a no-op

aiida/engine/processes/calcjobs/manager.py
sphuber: Yes, it will fetch it from the computer attribute, which since it is stored, will cause a refresh. This was done on purpose however, to allow someone to change the value dynamically, without having to restart the daemon

aiida/engine/processes/calcjobs/manager.py
sphuber: Not really, because if a new request comes in when a request is already active (from another calculation) it will be simply added and the delay is not requested. See the `_ensure_updating` method which will be a no-op when the `_update_handle` is already defined

aiida/engine/processes/calcjobs/manager.py
sphuber: Even though your change is correct here, it made me realize that it is better to put a single info message at the point when the actual update is performed.

aiida/engine/processes/calcjobs/manager.py
sphuber: I explicitly did not reference the "daemon runner" here, because this applies to any runner. Also when you "run" a job, it creates a runner, which has its own job manager and own job lists for each authentication info. Since all these extra conditions have nothing to do with the `JobsList` I did not include them here. This behavior should rather be documented in a more general place. This is already done in the documentation of the `get_minimum_job_poll_interval`.

ltalirz: In this case, it still makes sense to mention that it applies to any runner individually.
This is not obvious.

aiida/engine/processes/calcjobs/manager.py
ltalirz: Ok, so maximum frequency of this being called is 1/min_polling_interval, correct?
I guess that's fine.

aiida/engine/processes/calcjobs/manager.py
sphuber: But I don't think that is a limitation of this class and therefore should not be in this docstring. This class is simply used by the `JobManager` which then makes sure there is one per `AuthInfo`. However, then still we do not have a guarantee that this is unique per authinfo. A runner could just this instantiate two `JobManager` instances and then the condition that is written in the dosctring of the `JobsList` would be broken. I think it is just the wrong place for this information as it is not enforced here at all.

ltalirz: I get your point - I'm just saying that, from a user perspective, the use case of running jobs using multiple authinfos on the same computer is quite esoteric compared to the highly relevant case of having multiple runners. 
So, as a user reading this docstring, that is what I would like it to tell me.

Perhaps one way to resolve this is to mention that the JobList is per JobManager and put a link to the JobManager's  documentation (which then should somehow mention its relation to the runners and what this means for any per-computer limits).

sphuber: I have put a reference to the `JobManager` in the `JobsList` docstring and explained the bigger picture with the runners in the `JobManager` docstring

PR #3102, opened 2019-07-01
Allow to ignore default values in `InteractiveOption`

### LINKED ISSUE #2905: Interactive verdi commands need way to unset default values
Options that already have a value, either a default or from an existing object that is being reconfigured, will have that value preset in interactive mode. It is now impossible to "unset" that value, because simply pressing enter will assume the user wants to use that default value. There needs to be a way to say I want *no* value for this option, even if there is a default. Example is the option `proxy` in `verdi computer configure`. If you set it by accident and then reexecute the command, you cannot "unset" it.

sphuber:
Fixes #2905 

The `InteractiveOption` in interactive mode, will populate the value
with a default, if defined. If no value is specified by the user, this
is interpreted as the user wanting to use the default. However, since
the defaults are often contextual and are determined by existing data,
the situation can occur where one wants to set no specific value, even
though a default has been defined through pre-existing data. To fix this
situation, a special character `!` is introduced that when passed in
interactive mode is interpreted as passing `None`. Note that this only
works in interactive mode. When `!` is used non-interactively, the value
is interpreted as is.

### DIFF
--- a/aiida/cmdline/params/options/interactive.py
+++ b/aiida/cmdline/params/options/interactive.py
@@ -48,7 +48,10 @@ class InteractiveOption(ConditionalOption):
         def foo(label):
             click.echo('Labeling with label: {}'.format(label))
     """
+
     PROMPT_COLOR = 'yellow'
+    CHARACTER_PROMPT_HELP = '?'
+    CHARACTER_IGNORE_DEFAULT = '!'
 
     def __init__(self, param_decls=None, switch=None, prompt_fn=None, contextual_default=None, **kwargs):
         """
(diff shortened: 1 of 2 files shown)


aiida/cmdline/params/options/interactive.py
```diff
@@ -209,7 +219,7 @@ def simple_prompt_loop(self, ctx, param, value):
                     return value
 
     def after_callback(self, ctx, param, value):
-        """If a callback was registered on init, call it and return it's value"""
+        """If a callback was registered on init, call it and return it's value."""
```
giovannipizzi: It should have been `its` instead of `it's`... anyway we can fix this in another PR

Issue #2964, opened 2019-06-04
REST API - docs still mention v2


ltalirz:
See e.g. https://aiida.readthedocs.io/projects/aiida-core/en/latest/restapi/index.html#

I guess this should be v3 now?

@asle85 @waychal Can one of you please fix this?

ltalirz:
As mentioned before, it would also be very useful if `verdi restapi` would show the actual URL that people need to use (including the /api/v<x>)  suffix.

ltalirz:
Finally, this is also **the** place to advertise the provenance browser.

In addition to the `.../api/v2` link, include a link to https://www.materialscloud.org/explore/connect

BONUS: If you implement https://github.com/materialscloud-org/frontend-explore/issues/142 first, then you can display a link that you can click and it will work directly

asle85:
The message with the base url printed when running ``verdi restapi`` comes from ``werkzeug.serving.run_simple`` and cannot be customized in a simple way.

What we can do is to print a message with the url prefix **before** the default messages, e.g.

    verdi restapi
    * Add the suffix api/v3 to form the base url
    * Serving Flask app "aiida.restapi.run_api" (lazy loading)
    * Environment: production
      WARNING: Do not use the development server in a production environment.
      Use a production WSGI server instead.
    * Debug mode: off
    * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)

ltalirz:
That's fine as well - why don't you construct the full URL there already?
E.g.
`REST API running on http://127.0.0.1:5000/api/v3`

CasperWA:
This issue was fixed through PR #3106

PR #3106, opened 2019-07-01
Update documentation for version v3 of REST API

### LINKED ISSUE #3048: Check documentation from REST API
Check if REST API section is updated for AiiDA v1.0.0

[https://aiida.readthedocs.io/projects/aiida-core/en/develop/restapi/index.html](https://aiida.readthedocs.io/projects/aiida-core/en/develop/restapi/index.html)

asle85:
Fix ~~#3048~~ #2964.

Changed v2 to v3 in the base url, and updated documentation and examples to reflect the changes in the ORM. 
Added full base url (including /api/v* suffix) to the message printed by ``verdi restapi`` command, so the user can click on the link and directly access the API. 
As soon as the functionality of passing the REST endpoint as a variable in the URL of Materials Cloud Explore is implemented in production, this should print also the proper link to [https://www.materialscloud.org/explore/connect](https://www.materialscloud.org/explore/connect), so it can directly connect to the explore interface of a database.

Fixed help text in rest api schema for Groups.

### DIFF
--- a/aiida/orm/groups.py
+++ b/aiida/orm/groups.py
@@ -330,7 +330,7 @@ class Group(entities.Entity):
         return {
             "description": {
                 "display_name": "Description",
-                "help_text": "short description of the Computer",
+                "help_text": "Short description of the group",
                 "is_foreign_key": False,
                 "type": "str"
             },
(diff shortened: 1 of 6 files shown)


sphuber:
Isn't this supposed to rather fix issue #2964 ? Issues #3048 (and the duplicate it seems #2384 ) are the generic ones to make sure the REST API documentation is fully up to date. Is that the case with this PR? I remember @waychal saying that the only two end-points remaining would be `node` and `process`. However, I see many more, among which `calculation` which seems a remnant from `0.12`.

PR #3108, opened 2019-07-01
Enforcing specific precision in clean_value for floats

### LINKED ISSUE #2631: Storing a node changes its hash, due to float rounding

giovannipizzi:
Fixes #2631

This will guarantee a larger uniformity in what gets stored across backends.
Note that this commit also fixes the problem of the bigint vs. float
part of the discussion that is mentioned in the thread, i.e.
that numbers like 1.2e20, floats in python, become long ints
in the database. Here we forcefully convert numbers >=1.e14 to
integers to alleviate the problem and ensure that a roundtrip
via the database does not change the hash.

IMPORTANT: in this commit, we enforce a maximum precision of 14
digits in AiiDA - this is a change with respect to what used
to be before. Note also that we are *not* going to truncate old
AiiDA values, so behavior before and after this commit will change
even if the stored value is the same.

Note that this commit changes the hashing functionality, so we
will need to ask users to rehash everything.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,6 @@
         aiida/backends/djsite/db/migrations/0012_drop_dblock.py|
         aiida/backends/djsite/db/migrations/0013_django_1_8.py|
         aiida/backends/djsite/db/migrations/0014_add_node_uuid_unique_constraint.py|
-        aiida/backends/djsite/db/migrations/0015_invalidating_node_hash.py|
         aiida/backends/djsite/db/migrations/0016_code_sub_class_of_data.py|
         aiida/backends/djsite/db/migrations/0017_drop_dbcalcstate.py|
         aiida/backends/djsite/db/migrations/__init__.py|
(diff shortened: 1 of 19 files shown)


giovannipizzi:
Tests added in a second commit; also removed annoying output during tests of migrations

dev-zero:
for the record: converting floats to string and back again to control decimal rounding/truncation seems to be a very efficient way to do it, see discussion here https://news.ycombinator.com/item?id=20816827

giovannipizzi:
Thanks!

Issue #2549, opened 2019-02-28
improvements to caching documentation / ease of use


ltalirz:
- [ ] aiida needs to add a `aiida_caching.yml` file by default (with commented example of configuration)
 - [ ] the first section of the documentation should focus on the useful hashing of calculations (the example should be about that). An example on data is optional. Mentioning things that are *not* hashable should go to the end
 - [ ] The "configuration" section should first provide the simplest option:
```
profile-name:
  default: True
```
- [ ] it also needs to mention `verdi rehash` which people will need to do in order to enable to take advantage of hashing in an existing database

Finally, I actually would argue that caching should be enabled by default since it is such a useful feature but that is up for discussion.

zhubonan:
By the way, I realized that the hash of a `CalcJobNode` returned by `get_hash` is not the same as the actual hash used for matching (stored in `extras`) when it get stored. This is because the updatable attributes are added and not excluded from calculating the hash, such as *last_jobinfo*, *exit_message* etc. This is not affect matching though. These attributes are not there when the node gets stored.

chrisjsewell:
yeh +1 for this, there should definitely be `verdi` methods for interfacing with the `cache_config.yml`, to show/modify current settings

greschd:
>
> By the way, I realized that the hash of a `CalcJobNode` returned by `get_hash` is not the same as the actual hash used for matching (stored in `extras`) when it get stored. This is because the updatable attributes are added and not excluded from calculating the hash, such as _last_jobinfo_, _exit_message_ etc. This is not affect matching though. These attributes are not there when the node gets stored.

For future reference, this was fixed in #3130

greschd:
> (to decide) provide helper function to help figuring out why hashes of two nodes differ

FYI, at least a simple version of this would be relatively straightforward: Go through the ``get_objects_to_hash`` of each object, and see where they differ. If you want to get fancy, one could recursively go through the differing objects and figure out where they are different, too. That's a bit more complicated though, because it would be _inside_ the `make_hash` function.

greschd:
A comment that came up in #3878 is that `_get_objects_to_hash` is hard to interpret: While some of its contents are self-explanatory, others come "out of nowhere", and it's not clear what they correspond to without going into the code.

This doesn't _necessarily_ mean `_get_objects_to_hash` needs to be changed (as this would invalidate current hashes), but needs to be addressed in creating a helper for finding hash differences.

[...]

Issue #2990, opened 2019-06-11
per process_type caching configuration is not avalaible for CalcJobNode


zhubonan:
At the moment the default behaviour of caching for each node classes can be tuned in `cache_config.yml`. 

In 1.0.0 various types of `CalcJob` processes all store the same type of nodes `CalcJobNodes` in the database. The default caching configuration for `CalcJobNode` can be turned on or off but not for individual types of processes. 

I think this can be fixed by including the `process_type` of `CalcJobNode` when retrieving default the configurations.

sphuber:
Instead of making the switch in the configuration file on the `process_type` the `CalcJobNode` will have, we should just update the configuration parser to accept `CalcJob` sub classes. The fact that it still switches on the node class is because the old calculation jobs used to be implemented as a sub class of a calculation instead of a process.

zhubonan:
Yes, this would be ideal as it keeps the old behaviour. 
In fact, I initially tried to turn on the cache for my calculations (sub-class of `CalcJob`) this way but it did not work - it turned out to be the `CalcJobNode` class that needs to be enabled for caching of calculations to work.

sphuber:
Exactly, but that is only because at the time that the caching was implemented calculation jobs were implemented as a sub class of a `Node`. When we address this issue however, we should simply adapt the logic of the caching mechanism. The user should simply define the calculation job class to cache and the implementation in `aiida-core` should then simply make sure that this is mapped to the correct nodes. The user should not have to think about this

PR #3111, opened 2019-07-01
update caching docs

### LINKED ISSUE #3026: Check documentation for "working with AiiDA/caching"

https://aiida.readthedocs.io/projects/aiida-core/en/latest/working_with_aiida/index.html#caching

ltalirz:
fix #3026 
start fixing #2549 (some more to go)

 * reorder caching documentation, starting with how to enable it
 * add sketch of how caching works for calculations
 * move some technical details to developer documentation
 * add note on how to investigate why a calculation isn't being reused
 * general restructuring along similar topics


ltalirz:
@sphuber ready to review now

ltalirz:
@CasperWA would you be available for review?

sphuber:
@ltalirz just FYI, I started working on the review this morning. Fine if @CasperWA also wants to review of course

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
 
-    In [7]: print('UUID of n1:', n1.uuid)
-    UUID of n1: 956109e1-4382-4240-a711-2a4f3b522122
+In order to ensure that the provenance graph with and without caching is the same,
+AiiDA creates a *copy* of the orginial calculation node, linked to the original inputs, as well as a to a copy of the original outputs, as shown in :numref:`fig_caching`.
 
-    In [8]: print('n2 is cached from:', n2.get_extra('_aiida_cached_from'))
-    n2 is cached from: 956109e1-4382-4240-a711-2a4f3b522122
+.. note:: Caching is **not** implemented at the WorkChain/workfunction level (see :ref:`caching_limitations` for details).
+
```
sphuber: I think we need to add some clarification here that for the caching of a `CalculationNode` to happen, one does not need to add the identical input nodes. Rather, since the hash of the input nodes is included in the hash of the `CalculationNode` all inputs merely need to have the same content (and therefore the same hash). I fear giving only this example may suggest that a calculation will only be cached if the exact same input nodes are used. The following example shows that is not the case:
```
#!/usr/bin/env runaiida
from aiida import orm
from aiida.engine import calcfunction
from aiida.manage.caching import enable_caching


@calcfunction
def add(x, y):
    return x + y


x_01 = orm.Int(1)
y_01 = orm.Int(2)

x_02 = orm.Int(1)
y_02 = orm.Int(2)

results_01, node_01 = add.run_get_node(x_01, y_01)
print('Running calcfunction')
print('Inputs pks:', [node.pk for node in node_01.get_incoming().all_nodes()])

results_02, node_02 = add.run_get_node(x_01, y_01)
print('Running calcfunction without caching')
print('Inputs pks:', [node.pk for node in node_02.get_incoming().all_nodes()])

with enable_caching():
    results_03, node_03 = add.run_get_node(x_01, y_01)
print('Running calcfunction with caching')
print('Inputs pks:', [node.pk for node in node_03.get_incoming().all_nodes()])
print('Cache source:', node_03.get_cache_source())

with enable_caching():
    results_04, node_04 = add.run_get_node(x_02, y_02)
print('Running calcfunction with caching and different inputs but with identical value')
print('Inputs pks:', [node.pk for node in node_04.get_incoming().all_nodes()])
print('Cache source:', node_04.get_cache_source())
```
will print:
```
Running calcfunction
('Inputs pks:', [2977, 2976])
Running calcfunction without caching
('Inputs pks:', [2977, 2976])
Running calcfunction with caching
('Inputs pks:', [2977, 2976])
('Cache source:', u'420d73d1-8e18-429f-b9d4-a67bde3f72f3')
Running calcfunction with caching and different inputs but with identical value
('Inputs pks:', [2985, 2984])
('Cache source:', u'420d73d1-8e18-429f-b9d4-a67bde3f72f3')
```
As you can see, despite different input nodes being used in the last case, because the content is identical, the calculation is still cached

docs/source/developer_guide/core/caching.rst
```diff
@@ -3,10 +3,30 @@ Caching: implementation details
 
 This section covers some details of the caching mechanism which are not discussed in the :ref:`user guide <caching>`. If you are developing a plugin and want to modify the caching behavior of your classes, we recommend you read :ref:`this section <caching_matches>` first.
 
+Controlling caching
+-------------------
+
+When developing calculation and data classes, there are some methods you can use to determine how the hash is created:
+
+* To ignore specific attributes, a ``Node`` subclass can have a ``_hash_ignored_attributes`` attribute. This is a list of attribute names which are ignored when creating the hash.
+* For calculations, the ``_hash_ignored_inputs`` attribute lists inputs that should be ignored when creating the hash.
```
CasperWA: This may not even be a comment that can be resolved here - but isn't the plural of input, input?

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
+* the ``__version__`` of the package which defined the node class
+* the content of the repository folder of the node
+* the UUID of the computer, if the node is associated with one
 
-When running a ``CalcJob`` through the ``Process`` interface, you cannot directly set the ``use_cache`` flag when the calculation node is stored internally. Instead, you can pass the ``_use_cache`` flag to the ``run`` or ``submit`` method.
+Once a node is stored in the database, its hash is stored in the ``_aiida_hash`` extra, and this extra is used to find matching nodes.
+If a node of the same class with the same hash already exists in the database, this is considered a cache match. 
 
-Caching is **not** implemented for workchains and workfunctions. Unlike calculations, they can not only create new data nodes, but also return exsting ones. When copying a cached workchain, it's not clear which node should be returned without actually running the workchain. This is explained in more detail in the section :ref:`caching_provenance`.
+Use the :meth:`.get_hash() <aiida.orm.nodes.Node.get_hash>` method to check the hash of any node.
```
sphuber: the links :meth:`.get_hash() <aiida.orm.nodes.Node.get_hash>` can be written as :meth:`~aiida.orm.nodes.Node.get_hash` i.e. simply a tilde in front to get only the method (or class) name

ltalirz:
ah sorry - thanks!

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
+
+.. _caching_matches:
 
-As you can see, passing ``use_cache=True`` to the ``store`` method enables using the cache. The fact that ``n2`` was created from ``n1`` is stored in the ``_aiida_cached_from`` extra of ``n2``.
+How are nodes hashed?
+---------------------
+
+By default, the hash of a node is computed from:
+
+* all attributes of the node, except the ``_updatable_attributes``
```
sphuber: This also includes the `_hash_ignored_attributes` (currently only defined for `CalcJob`)

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
+* the ``__version__`` of the package which defined the node class
+* the content of the repository folder of the node
+* the UUID of the computer, if the node is associated with one
```
sphuber: For `ProcessNodes` this also includes all the hashes of nodes with an `INPUT_CALC` or `INPUT_WORK` link, i.e. input `Data` nodes.

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+    In [6]: n2.store(use_cache=True)
+    Out[6]: u'test string'
 
-What to do when caching is used when it shouldn't
--------------------------------------------------
+    In [7]: print('UUID of n1:', n1.uuid)
+    UUID of n1: 956109e1-4382-4240-a711-2a4f3b522122
 
-In general, the caching mechanism should trigger only when the output of a calculation will be exactly the same as if it is run again. However, there might be some edge cases where this is not true.
+    In [8]: print('n2 is cached from:', n2.get_extra('_aiida_cached_from'))
```
sphuber: Better to use `get_cache_source` instead of `get_extra('_aiida_cached_from')` as this may move in the future

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+    n2 is cached from: 956109e1-4382-4240-a711-2a4f3b522122
 
-For example, if the parser is in a different python module than the calculation, the version number used in the hash will not change when the parser is updated. While the "correct" solution to this problem is to increase the version number of a calculation when the behavior of its parser changes, there might still be cases (e.g. during development) when you manually want to stop a particular node from being cached.
+When running a ``CalcJob`` through the ``Process`` interface, you can achive the same effect by passing ``_use_cache=True`` to the ``run`` or ``submit`` method.
```
sphuber: achive -> achieve

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
 
-In such cases, you can follow these steps to disable caching:
+If you suspect a node is being reused in error (e.g. during development),
+it is also possible to manually *prevent* a specific node from being reused:
 
-1. If you suspect that a node has been cached in error, check that it has a ``_aiida_cached_from`` extra. If that's not the case, it is not a problem of caching.
-2. Get all nodes which match your node, and clear their hash:
+1. Check that the node in question has a ``_aiida_cached_from`` extra.
```
sphuber: use `get_cache_source` instead, which return `None` if not cached, otherwise UUID of cache source

docs/source/developer_guide/core/caching.rst
```diff
@@ -3,10 +3,30 @@ Caching: implementation details
...
+* To add things which should be considered in the hash, you can override the :meth:`_get_objects_to_hash <aiida.orm.nodes.Node._get_objects_to_hash>` method. Note that doing so overrides the behavior described above, so you should make sure to use the ``super()`` method.
```
CasperWA: I think you can write these as :meth:`~aiida.orm.nodes.Node._get_objects_to_hash` (using the now hidden code markers), and the tilde will make sure only the last part is used as the link. But I guess either way is fine?

docs/source/developer_guide/core/caching.rst
```diff
@@ -3,10 +3,30 @@ Caching: implementation details
...
+* Pass a keyword argument to :meth:`.get_hash <aiida.orm.nodes.Node.get_hash>`. These are passed on to ``aiida.common.hashing.make_hash``. For example, the ``ignored_folder_content`` keyword is used by the :class:`JobCalculation <aiida.orm.nodes.process.calculation.calcjob.CalcJobNode>` to ignore the ``raw_input`` subfolder of its repository folder.
```
CasperWA: encapsulate "aiida.common.hashin.make_hash" in a :func: as well, or similar.

CasperWA: Class "JobCalculation" doesn't exist anymore. Just use the tilde method here, I think.

CasperWA: Do we still have a subfolder called "raw_input"?

docs/source/developer_guide/core/caching.rst
```diff
@@ -3,10 +3,30 @@ Caching: implementation details
...
+
+Additionally, there are two methods you can use to disable caching for particular nodes:
+
+* The :meth:`~aiida.orm.nodes.Node.is_valid_cache` property determines whether a particular node can be used as a cache. This is used for example to disable caching from failed calculations.
+* Node classes have a ``_cachable`` attribute, which can be set to ``False`` to completely switch off caching for nodes of that class. This avoids performing queries for the hash altogether.
+
+Finally an important remark: There are two ways in which the hash match can go wrong: False negatives, where two nodes should have the same hash but do not, or false positives, where two different nodes have the same hash.
```
CasperWA: Perhaps a comma after "Finally"?

CasperWA: Comma before "but do not"

docs/source/developer_guide/core/caching.rst
```diff
@@ -3,10 +3,30 @@ Caching: implementation details
...
+Be mindful of this when modifying the caching behaviour of your calculation and data classes.
+
+
 Disabling caching for ``WorkflowNode``
 --------------------------------------
 
-As discussed in the :ref:`user guide <caching_provenance>`, nodes which can have ``RETURN`` links cannot be cached. This is enforced on two levels:
+As discussed in the :ref:`user guide <caching_limitations>`, nodes which can have ``RETURN`` links cannot be cached. This is enforced on two levels:
 
 * The ``_cachable`` property is set to ``False`` in the :class:`aiida.orm.nodes.process.ProcessNode`, and only re-enabled in :class:`aiida.orm.nodes.process.calculation.calcjob.CalcJobNode` and :class:`CalcFunctionNode <aiida.orm.nodes.process.calculation.calcfunction.CalcFunctionNode>`. This means that a ``WorkflowNode`` will not be cached.
```
CasperWA: There should be tildes in these class references.

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
+Enabling caching
+----------------
 
-When a calculation is cached, a copy of the original calculation is created. This copy will keep the input links of the new calculation. The outputs of the original calculation are also copied, and linked to the new calculation. This allows for the new calculation to be a separate Node in the provenance graph and, critically, preserves the acyclicity of the graph.
+There are numerous reasons why you may need to re-run calculations you’ve already done before.
+Since AiiDA stores the full provenance of each calculation, it can *detect* whether a calculation has been run before
+and reuse its outputs without wasting computational resources.
 
-Caching is also implemented for Data nodes. This is not very useful in practice (yet), but is an easy way to show how the caching mechanism works:
+This feature is not (yet) enabled by default. In order to enable hashing:
```
sphuber: this is not enabling "hashing" but it is enabling "caching". We need to define and explain these terms better. We are using the terms "hashing" and "caching" too loosely and interchangeably. The "caching" only makes sense for `ProcessNodes`, where, for reasons explained later on, this is disabled for `WorkflowNodes` and therefore only possible for `CalculationNodes`. The "hashing" is merely a method to detect equivalent nodes and therefore allow to implement the "caching".

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
 Caching
 *******
 
-When working with AiiDA, you might sometimes re-run calculations which were already successfully executed. Because this can waste a lot of computational resources, you can enable AiiDA to **cache** calculations, which means that it will re-use existing calculations if a calculation with the same inputs is submitted again.
+Enabling caching
+----------------
 
-When a calculation is cached, a copy of the original calculation is created. This copy will keep the input links of the new calculation. The outputs of the original calculation are also copied, and linked to the new calculation. This allows for the new calculation to be a separate Node in the provenance graph and, critically, preserves the acyclicity of the graph.
+There are numerous reasons why you may need to re-run calculations you’ve already done before.
+Since AiiDA stores the full provenance of each calculation, it can *detect* whether a calculation has been run before
```
CasperWA: Keep complete sentences in a single line (see previous comment). I just wanted to write a comment again, since this is the "opposite" example :smiley:

CasperWA: Perhaps also comma before "and reuse its..."?

docs/source/working_with_aiida/caching.rst
CasperWA: Remove the (yet) and perhaps make a note after the fact instead? I.e., make a note block after describing what it is, saying circa when we intend to make it default.

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
-    In [2]: from aiida.orm import Str
+        aiida2:
+          default: True
 
-    In [3]: n1 = Str('test string')
+ 2. Run ``verdi -p aiida2 rehash`` in order to compute hashes of all nodes in your existing database
 
-    In [4]: n1.store()
-    Out[4]: u'test string'
+From this point onwards, when you submit a new calculation, AiiDA will compute hashes of the type of calculation and its inputs (see :ref:`caching_matches`). If the hash of the calculation is found to be already present in the database, AiiDA will reuse its results without repeating the actual calculation.
```
sphuber: Maybe replace "submit" with "launch". In AiiDA lingo, "submit" specifically means submitting to a daemon, however, the caching also works when "running" locally (the "submit" here could also misconstrue that it only applies to submitting `CalcJobs`). The term "launch" is used as the umbrella term for running or submitting a process.

CasperWA: For the reference here, you may want to have it say the section title?

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
 
-    In [5]: n2 = Str('test string')
+.. _fig_caching:
+.. figure:: include/images/caching.png
+  :align: center
+  :height: 350px
 
-    In [6]: n2.store(use_cache=True)
-    Out[6]: u'test string'
+  When resusing a cached calculation, AiiDA will link up to the input nodes as usual, and make a copy of both the calculation node **C** and its outputs **D3**.
```
CasperWA: reusing

docs/source/working_with_aiida/caching.rst
sphuber: I would here just say that it computes the hash of the calculation. Don't reference the inputs. Later on you explain what is included when computing a hash. I would explain there that for `ProcessNodes` the hashes of the input data nodes are included

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
 
-    In [7]: print('UUID of n1:', n1.uuid)
-    UUID of n1: 956109e1-4382-4240-a711-2a4f3b522122
+In order to ensure that the provenance graph with and without caching is the same,
+AiiDA creates a *copy* of the orginial calculation node, linked to the original inputs, as well as a to a copy of the original outputs, as shown in :numref:`fig_caching`.
 
-    In [8]: print('n2 is cached from:', n2.get_extra('_aiida_cached_from'))
-    n2 is cached from: 956109e1-4382-4240-a711-2a4f3b522122
+.. note:: Caching is **not** implemented at the WorkChain/workfunction level (see :ref:`caching_limitations` for details).
```
CasperWA: Same concerning this reference as my comment above.

docs/source/working_with_aiida/caching.rst
CasperWA: This is not `_updateable_attributes`?

docs/source/working_with_aiida/caching.rst
CasperWA: tilde?

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
+* the UUID of the computer, if the node is associated with one
 
-When running a ``CalcJob`` through the ``Process`` interface, you cannot directly set the ``use_cache`` flag when the calculation node is stored internally. Instead, you can pass the ``_use_cache`` flag to the ``run`` or ``submit`` method.
+Once a node is stored in the database, its hash is stored in the ``_aiida_hash`` extra, and this extra is used to find matching nodes.
+If a node of the same class with the same hash already exists in the database, this is considered a cache match. 
 
-Caching is **not** implemented for workchains and workfunctions. Unlike calculations, they can not only create new data nodes, but also return exsting ones. When copying a cached workchain, it's not clear which node should be returned without actually running the workchain. This is explained in more detail in the section :ref:`caching_provenance`.
+Use the :meth:`.get_hash() <aiida.orm.nodes.Node.get_hash>` method to check the hash of any node.
+
+In order to figure out why a calculation is *not* being reused, the :meth:`._get_objects_to_hash() <aiida.orm.nodes.Node._get_objects_to_hash>` may come in useful:
```
CasperWA: "be" instead of "come in"

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
+ 
 
 Configuration
 -------------
 
-Of course, using caching would be quite tedious if you had to set ``use_cache`` manually everywhere. To fix this, the default for ``use_cache`` can be set in the ``.aiida/cache_config.yml`` file. You can specify a global default, or enable / disable caching for specific calculation or data classes. An example configuration file might look like this:
+Class level
+...........
+
+Besides a on/off switch per profile, the ``.aiida/cache_config.yml`` provides control over caching at the level of specific calculation or data classes:
```
CasperWA: an (second word)

docs/source/working_with_aiida/caching.rst
CasperWA: achieve

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
-    @workfunction
-    def select(a, b):
-        return b
+Limitations
+-----------
 
-    d = Int(1)
-    r1 = select(d, d)
-    r2 = select(Int(1), Int(1))
+1. The current implementation of caching for data nodes clones not only the graph representation of the reused node but also the underlying data in the database and file repository.
```
sphuber: I find this still more confusing than that it helps. Even when we implement the automated deduplication in the future, what will then be the definition of a "cached" `Data` node. It will still simply be a clone that happens to have the same hash as some other node. There is no link between them whatsoever. I really think that this stems from the fact that we are using the terms "hashing" and "caching" too loosely and interchangeably as explained in one of the first comments in this review. I would simply remove this entire point about `Data` nodes

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+When running a ``CalcJob`` through the ``Process`` interface, you can achive the same effect by passing ``_use_cache=True`` to the ``run`` or ``submit`` method.
 
-In such cases, you can follow these steps to disable caching:
+If you suspect a node is being reused in error (e.g. during development),
+it is also possible to manually *prevent* a specific node from being reused:
 
-1. If you suspect that a node has been cached in error, check that it has a ``_aiida_cached_from`` extra. If that's not the case, it is not a problem of caching.
-2. Get all nodes which match your node, and clear their hash:
+1. Check that the node in question has a ``_aiida_cached_from`` extra. 
+   If that's not the case, it anyhow is not reused.
```
CasperWA: "that is" instead of "that's". Contractions are chat language or flavour for novel writers, if you like :smile:

CasperWA: Also the second bi-sentence should be reworded: "it is anyhow not reused".

docs/source/working_with_aiida/caching.rst
CasperWA: Some commas are needed here for clarification

CasperWA: This goes for the next lines as well.

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+3. The constraint that the shape of the provenance graph should be independent of whether caching is enabled or not, imposes limitations on the possible caching operations.
+   In order to reuse a cached node, we need to know not only its contents but also how the new node should be linked to its parents and children.
 
-Overall, this limitation is acceptable: The runtime of AiiDA workchains is usually dominated by time spent inside expensive calculations. Since these can be avoided with the caching mechanism, it still improves the runtime and required computer resources a lot.
+   * **Data nodes:** Making a copy of a data node should not change its links, so AiiDA just needs to link the new node to the direct ancestors of the old node.
+   * **Calculations nodes:** The node can have inputs and creates new data nodes as outputs. Again, the new node needs to replicate these links. In order to make it look as if the calculation had produced its own outputs, the output nodes are copied and linked as well.
```
CasperWA: No "s" at the end of Calculations

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+   * **Workflows nodes:** Workflows differ from calculations in that they can *return* an input node or an output node created by a calculation.
```
CasperWA: No "s" at the end of Workflows

CasperWA: Why is return italic here, but create is not in the previous point?

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
 
-Overall, this limitation is acceptable: The runtime of AiiDA workchains is usually dominated by time spent inside expensive calculations. Since these can be avoided with the caching mechanism, it still improves the runtime and required computer resources a lot.
+   * **Data nodes:** Making a copy of a data node should not change its links, so AiiDA just needs to link the new node to the direct ancestors of the old node.
+   * **Calculations nodes:** The node can have inputs and creates new data nodes as outputs. Again, the new node needs to replicate these links. In order to make it look as if the calculation had produced its own outputs, the output nodes are copied and linked as well.
+   * **Workflows nodes:** Workflows differ from calculations in that they can *return* an input node or an output node created by a calculation. 
+     This is not compatible with the cloning model chosen for calculations (leave input nodes untouched, clone output nodes). For this reason, workflows are not cached.
+     
+  Overall, this limitation is acceptable since the runtime of AiiDA workchains is usually dominated by expensive calculations, which are covered by the caching mechanism.
```
CasperWA: comma before "since the runtime..."

CasperWA: "are" instead of "is" after "AiiDA workchains". Perhaps also write workchains as WorkChains?

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
 
-While this example might seem contrived, the conclusion is valid more generally: Because workchains can return nodes from their history, they cannot be cached. Since even two equivalent workchains (with the same inputs) can have a different history, there is no way to deduce which links should be created on the new workchain without actually running it.
+3. The constraint that the shape of the provenance graph should be independent of whether caching is enabled or not, imposes limitations on the possible caching operations.
+   In order to reuse a cached node, we need to know not only its contents but also how the new node should be linked to its parents and children.
 
-Overall, this limitation is acceptable: The runtime of AiiDA workchains is usually dominated by time spent inside expensive calculations. Since these can be avoided with the caching mechanism, it still improves the runtime and required computer resources a lot.
+   * **Data nodes:** Making a copy of a data node should not change its links, so AiiDA just needs to link the new node to the direct ancestors of the old node.
```
sphuber: This is not what is currently happening and also not what should be happening I think.
```
#!/usr/bin/env runaiida
from aiida import orm
from aiida.engine import calcfunction


@calcfunction
def add(x, y):
    return x + y


x = orm.Int(1)
y = orm.Int(2)

result, node = add.run_get_node(x, y)

cached = orm.Int(result.value)
cached.store(use_cache=True)
print('Original', result.pk, result.get_incoming().all_nodes())
print('Cached', cached.pk, cached.get_incoming().all_nodes())
```
will print
```
('Original', 2996, [<CalcFunctionNode: uuid: 12f52293-3e3a-4604-9bef-4671f3a6bb82 (pk: 2995) (__main__.add)>])
('Cached', 2997, [])
```
Even though the `cloned` node has the same value as the output of the calcfunction, I don't think it should get the links of the parent. What does that even mean? Imagine we have many calculations that all produce an output integer with value 3, through different inputs and computations. When we now "cache" an `Int(3)`, which one of these output nodes do we take? To me this shows again why "caching" of `Data` nodes makes no sense and we should not talk about it. The only thing that is important is the "hashing" of data nodes

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
 
-    d = Int(1)
-    r1 = select(d, d)
-    r2 = select(Int(1), Int(1))
+1. The current implementation of caching for data nodes clones not only the graph representation of the reused node but also the underlying data in the database and file repository. 
+   This could be improved in order to reduce data duplication.
 
-The two ``select`` workfunctions have the same inputs as far as their hashes go. However, the first call uses the same input node twice, while the second one has two different inputs. If the second call should be cached from the first one, it is not clear which of the two input nodes should be returned.
+2. The caching mechanism for calculations *should* trigger only when the inputs and the calculation to be performed are exactly the same.  
+   Edge cases where this assumption might be violated include cases where the calculation parser is in a different python module than the calculation and the developer made changes without updating the version number of the plugin.
```
sphuber: Made changes to the calculation or the parser? It is even worse because as for as I know, for `CalcJob` only the `parser_name` is stored in the attributes and so that's the only thing that is included in computing the hash. Even if the parser version *was* updated properly, this will not be detected by the hashing I think

docs/source/working_with_aiida/caching.rst
sphuber: Calculation nodes (minus -s)

docs/source/working_with_aiida/caching.rst
sphuber: Workflow nodes (minus -s)

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +113,69 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+   In order to reuse a cached node, we need to know not only its contents but also how the new node should be linked to its parents and children.
 
-Overall, this limitation is acceptable: The runtime of AiiDA workchains is usually dominated by time spent inside expensive calculations. Since these can be avoided with the caching mechanism, it still improves the runtime and required computer resources a lot.
+   * **Data nodes:** Making a copy of a data node should not change its links, so AiiDA just needs to link the new node to the direct ancestors of the old node.
+   * **Calculations nodes:** The node can have inputs and creates new data nodes as outputs. Again, the new node needs to replicate these links. In order to make it look as if the calculation had produced its own outputs, the output nodes are copied and linked as well.
+   * **Workflows nodes:** Workflows differ from calculations in that they can *return* an input node or an output node created by a calculation. 
+     This is not compatible with the cloning model chosen for calculations (leave input nodes untouched, clone output nodes). For this reason, workflows are not cached.
```
sphuber: I understand that you wanted to simplify the reasoning why `WorkflowNodes` are not cached, but I think this makes it worse. Sure the sentence here is less complicated, but it does not explain the *why* at all. I would really keep the example that demonstrates the problem, it is really important

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
+.. note:: Caching is **not** implemented at the WorkChain/workfunction level (see :ref:`caching_limitations` for details).
+
+
+.. _caching_matches:
 
-As you can see, passing ``use_cache=True`` to the ``store`` method enables using the cache. The fact that ``n2`` was created from ``n1`` is stored in the ``_aiida_cached_from`` extra of ``n2``.
+How are nodes hashed?
+---------------------
+
+By default, the hash of a node is computed from:
```
sphuber: "By default"? There is no way to configure this behavior

docs/source/working_with_aiida/caching.rst
sphuber: This is not correct, the `_use_cache` keyword does not exist. Instead the `from aiida.manage.caching import enable_caching` context manager can be used

CasperWA:
> Thanks a lot @ltalirz , already it seems to be a great improvement, but there are still quite some problems I think that we need to address. For now I mostly just reviewed the semantics of the changes and haven't looked at the structuring or spelling that kind of stuff.

That's perfect @sphuber! I did the opposite :sweat_smile:

docs/source/developer_guide/core/caching.rst
ltalirz: not in this case https://english.stackexchange.com/a/180366

CasperWA: Great! Thanks :smiley: That saves me and us a lot of changes :sweat_smile:

docs/source/developer_guide/core/caching.rst
ltalirz: I think so as well, was not introduced by me but I will fix it.

docs/source/developer_guide/core/caching.rst
zhubonan: I think the "raw_inputs" is removed and now input files are stored at the top level of a `CalcJobNode` #3125

docs/source/developer_guide/core/caching.rst
ltalirz: Thanks, will add comma after "Finally".

> Comma before "but do not"

Not in this case https://www.dailywritingtips.com/comma-before-but/

docs/source/working_with_aiida/caching.rst
ltalirz: I don't think there's an official consensus on this yet... so I'm not sure I can say much more than what I wrote there...

docs/source/working_with_aiida/caching.rst
ltalirz: Not quite sure what you mean... using the reference like this *will* display the section title.

docs/source/working_with_aiida/caching.rst
CasperWA: Then just remove the (yet) ?

docs/source/working_with_aiida/caching.rst
CasperWA: Ah! - Then never mind :)

docs/source/working_with_aiida/caching.rst
ltalirz: Good question - no it's updatable and [that is prevalent use today](https://english.stackexchange.com/a/56433)

CasperWA: .. loving all the proper referencing :+1: :smile:

docs/source/working_with_aiida/caching.rst
ltalirz: I don't think any commas would be needed here from the language point of view but I agree that the sentence is a bit long and will split it in half.

docs/source/working_with_aiida/caching.rst
ltalirz: > Why is return italic here, but create is not in the previous point?

It's just highlighting the most important word in this sentence.

docs/source/working_with_aiida/caching.rst
ltalirz: > "are" instead of "is" after "AiiDA workchains".

Actually "is" is correct, since "runtime" is the subject of this sentence.

docs/source/working_with_aiida/caching.rst
CasperWA: Sure - but it is also underlining the link type. Create being a link type as well, why not uniformize? Unless the point here is that _because_ it is return (and not create) then it will not be included?

docs/source/working_with_aiida/caching.rst
CasperWA: You're right! Sorry.

docs/source/developer_guide/core/caching.rst
ltalirz: thanks for bringing this up. will fix this once #3125 is resolved.

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,90 @@
...
+
+    In [5]: calc=load_node(1234)
+    
+    In [6]: calc.get_hash()
+    Out[6]: '62eca804967c9428bdbc11c692b7b27a59bde258d9971668e19ccf13a5685eb8'
+    
+    In [7]: calc._get_objects_to_hash()
+    Out[7]:
+    ['1.0.0b4',
+     {'_sealed': True,
```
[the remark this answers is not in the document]
ltalirz: Ok, will do!

docs/source/working_with_aiida/caching.rst
ltalirz: There *are* ways of controlling this. Will add a :ref: at this point.

docs/source/working_with_aiida/caching.rst
ltalirz: Thanks, this is new. Will incorporate!

docs/source/working_with_aiida/caching.rst
ltalirz: I changed this, since I felt the example given did not actually illustrate the problem.
Let's rediscuss this in person, I will then adapt the text accordingly.

docs/source/working_with_aiida/caching.rst
ltalirz: thanks - mentioned by Casper as well.
will fix the occurences (copied from dominik's text)

docs/source/working_with_aiida/caching.rst
ltalirz: Yes, this was the point.
I'll keep it like this for the moment but it might change anyhow since sebastiaan liked the example that was provided originally.

docs/source/working_with_aiida/caching.rst
ltalirz: I've added this now.

docs/source/working_with_aiida/caching.rst
ltalirz: Added as well.

docs/source/developer_guide/core/caching.rst
sphuber: The `raw_input` sub folder is still there, but its existence is hidden from the user and the `Repository` interface automatically deals with it

docs/source/working_with_aiida/caching.rst
sphuber: How exactly? I can't think of one without changing the node classes, but I wouldn't call that configurable

docs/source/developer_guide/core/caching.rst
ltalirz: @sphuber So, do I understand correctly that the sentence "For example, ..." is still accurate (after fixing the class reference)?

I see this is still in the code
```python
    def get_hash(self, ignore_errors=True, ignored_folder_content=('raw_input',), **kwargs):  # pylint: disable=arguments-differ
```

docs/source/working_with_aiida/caching.rst
ltalirz: Ok, I do it for you ;-)

ltalirz: you are right and I will adapt.

docs/source/working_with_aiida/caching.rst
ltalirz: I get your point but I think it's helpful to mention the basic info of what goes into the hash already here since the question comes up immediately.
I'll put this info into the brackets, together with the reference

docs/source/developer_guide/core/caching.rst
sphuber: Ah, we must have missed this in PR #3130 . I guess this is how the problem of mutable `CalcJobNode` hash was dealt with. I suspect it was actually working but was broken by PR #3108 . That dealt with a similar problem, where the hash computed before storing, changed after storing, because the point of "cleaning" attributes was changed. To fix this, @giovannipizzi made the `get_hash` function not callable on unstored nodes, to prevent people from getting a hash, that would change once stored. The `store` method then calls the `_get_hash` method. The overridden `CalcJobNode.get_hash` however now never gets called by the `store` method and so the exclusion of the `raw_input` folder from the hash was circumvented. I think, since we solved the problem differently now, we can simply remove the `CalcJobNode.get_hash` override

ltalirz: Opened #3312 for this

docs/source/working_with_aiida/caching.rst
ltalirz: Thanks - I've added a sentence to the figure caption
>  While this sketch shows **C'** reuse the input nodes of **C'**, it is only the *content* of the input nodes (their hash) that matters.

docs/source/working_with_aiida/caching.rst
ltalirz: done

docs/source/working_with_aiida/caching.rst
ltalirz: After we had our chat, I now understand what you mean.
I think it is still important to mention here but I've adapted the terminology.

docs/source/working_with_aiida/caching.rst
ltalirz: I've removed the mention of data nodes in this list as I agree it is not helpful.

docs/source/working_with_aiida/caching.rst
ltalirz: I've updated this section, and it does now give a hint as to why caching workflows is not straightforward.

docs/source/working_with_aiida/caching.rst
ltalirz: > "By default"? There is no way to configure this behavior
> How exactly? I can't think of one without changing the node classes, but I wouldn't call that configurable

I'm not calling it `configurable`, I'm just writing down how the has his computed by default.
See section "controlling caching" for ways to modify this behavior (some of which includes modifying node classes).
I think this wording is reasonable.

docs/source/developer_guide/core/caching.rst
```diff
@@ -4,13 +4,38 @@ Caching: implementation details
...
+* Pass a keyword argument to :meth:`.get_hash <aiida.orm.nodes.Node.get_hash>`.
+  These are passed on to :meth:`aiida.common.hashing.make_hash`.
+  For example, the ``ignored_folder_content`` keyword is used by the :class:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode` to ignore the ``raw_input`` subfolder of its repository folder.
+
+Additionally, there are two methods you can use to disable caching for particular nodes:
+
+* The :meth:`~aiida.orm.nodes.Node.is_valid_cache` property determines whether a particular node can be used as a cache. This is used for example to disable caching from failed calculations.
+* Node classes have a ``_cachable`` attribute, which can be set to ``False`` to completely switch off caching for nodes of that class. This avoids performing queries for the hash altogether.
+
+Finally, an important remark: A cache match can go wrong in two ways: False negatives, where two nodes *should* have the same hash but do not, and false positives, where two different nodes have the same hash.
```
sphuber: Two consecutive colons is a bit weird to read. For the second one, since the following clause is a list, it should not use a capital. Maybe just make the second colon and following list an itemized list.

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,92 @@
...
 
-    In [1]: from __future__ import print_function
+    .. code:: yaml
 
-    In [2]: from aiida.orm import Str
+        aiida2:
+          default: True
 
-    In [3]: n1 = Str('test string')
+ 2. Run ``verdi -p aiida2 rehash`` in order to compute hashes of all nodes in your existing database.
```
sphuber: I don't think step 2 is really necessary. Not even for older databases from before caching existed since there are migrations to compute the hash. If you want to keep it, maybe add a note that this is optional and may take a long time for big databases.

docs/source/working_with_aiida/caching.rst
```diff
@@ -4,45 +4,92 @@
...
+* the UUID of the computer, if the node is associated with one
+
+The hash of a :class:`~aiida.orm.ProcessNode` will, on top of this, include the hashes of any of its input ``Data`` nodes.
+
+Once a node is stored in the database, its hash is stored in the ``_aiida_hash`` extra, and this extra is used to find matching nodes.
+If a node of the same class with the same hash already exists in the database, this is considered a cache match.
+
+Use the :meth:`~aiida.orm.nodes.Node.get_hash` method to check the hash of any node.
+
+In order to figure out why a calculation is *not* being reused, the :meth:`._get_objects_to_hash() <aiida.orm.nodes.Node._get_objects_to_hash>` may be useful:
```
sphuber: For consistency maybe change to `:meth:`~aiida.orm.nodes.Node._get_objects_to_hash``

docs/source/working_with_aiida/caching.rst
```diff
@@ -54,7 +101,10 @@ Of course, using caching would be quite tedious if you had to set ``use_cache``
       disabled:
         - aiida.orm.nodes.data.float.Float
 
-This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str``, and disabled for all other classes. In this example, manually disabling ``aiida.orm.nodes.data.float.Float`` is actually not needed, since the ``default: False`` configuration means that caching is disabled for all classes unless it is manually enabled. Note also that the fully qualified class import name (e.g., ``aiida.orm.nodes.data.str.Str``) must be given, not just the class name (``Str``). This is to avoid accidentally matching classes with the same name. You can get this name by combining the module name and class name, or (usually) from the string representation of the class:
+In this example, caching is enabled for ``TemplatereplacerCalculation`` and ``Str``, and disabled for all other classes.
```
sphuber: I realize now that this may and probably is broken since we separated the node and process. For example adding `aiida_quantumespresso.calculations.pw.PwCalculation` is now a process class and probably won't work as intended. Instead we will probably have to support to use entry point string in the config file, e.g. `aiida.calculations:quantumespresso.pw`

There is already an open issue for this: https://github.com/aiidateam/aiida-core/issues/2990

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +115,76 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+    with enable_caching(node_class=CalcJobNode):
+       run(...)
 
-In such cases, you can follow these steps to disable caching:
+If you suspect a node is being reused in error (e.g. during development),
+it is also possible to manually *prevent* a specific node from being reused:
 
-1. If you suspect that a node has been cached in error, check that it has a ``_aiida_cached_from`` extra. If that's not the case, it is not a problem of caching.
-2. Get all nodes which match your node, and clear their hash:
+1. For one of the cloned nodes, check that :meth:`~aiida.orm.nodes.Node.get_cache_source` returns a UUID.
```
sphuber: For one of the _accidentally_ cloned nodes?

Or _unwantedly_ maybe? In any case, first time reading it was not clear that the it referenced a calculation that was cached from another that was not supposed to be used.

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +115,76 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+   If it returns `None`, the node was not cloned.
+2. Clear the hashes of all nodes that are considered identical to this node:
```
sphuber: It seems a good extra step to clear the hash for all other nodes that share the same hash, but with this snippet people will not clear the hash of the original faulty node. Maybe add an explicit step also clearing the hash of `n` itself.

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +115,76 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+   * **Calculation nodes:** Calculation nodes can have data inputs and create new data nodes as outputs.
+     In order to make it look as if a cloned calculation produced its own outputs, the output nodes are copied and linked as well.
+   * **Workflow nodes:** Workflows differ from calculations in that they can *return* an input node or an output node created by a calculation.
+     Since caching does not care about the *identity* of input nodes but only their *content*, it is not straightforward to figure out which node to return in a cached workflow.
 
-While this example might seem contrived, the conclusion is valid more generally: Because workchains can return nodes from their history, they cannot be cached. Since even two equivalent workchains (with the same inputs) can have a different history, there is no way to deduce which links should be created on the new workchain without actually running it.
+  For the moment, this limitation is acceptable since the runtime of AiiDA WorkChains is usually dominated by expensive calculations, which are covered by the current caching mechanism.
```
sphuber: Remove the empty line of the usage of `#` will not work properly and the final item will start with `1.` again

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +115,76 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
 
-    from aiida.engine import workfunction
-    from aiida.orm import Int
+Limitations
+-----------
 
-    @workfunction
-    def select(a, b):
-        return b
+#. While caching saves unnecessary computations, the current implementation does not yet save disk space:
```
sphuber: This is not so much a limitation (as is the uncachability of workflow processes) as an implementation inefficiency.
I would maybe rather add this as a note or warning in an earlier section where we introduce the caching and its benefits. In any case I would not add it as the first item here, but rather start with the uncachability of workflows

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +115,76 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+   The output nodes of the cached calculation are full copies of the original outputs.
+   The plan is to add data deduplication as a global feature at the repository and database level (independent of caching).
 
-    d = Int(1)
-    r1 = select(d, d)
-    r2 = select(Int(1), Int(1))
+#. Workflow nodes are not cached. As explained below, in the current design this follows from the requirement that the provenance graph be independent of whether caching is enabled or not.
```
sphuber: "As explained below" this is probably remnant from before reorganization. Does not make sense now.

docs/source/working_with_aiida/caching.rst
```diff
@@ -65,88 +115,76 @@ This means that caching is enabled for ``TemplatereplacerCalculation`` and ``Str
...
+     Since caching does not care about the *identity* of input nodes but only their *content*, it is not straightforward to figure out which node to return in a cached workflow.
 
-While this example might seem contrived, the conclusion is valid more generally: Because workchains can return nodes from their history, they cannot be cached. Since even two equivalent workchains (with the same inputs) can have a different history, there is no way to deduce which links should be created on the new workchain without actually running it.
+  For the moment, this limitation is acceptable since the runtime of AiiDA WorkChains is usually dominated by expensive calculations, which are covered by the current caching mechanism.
 
-Overall, this limitation is acceptable: The runtime of AiiDA workchains is usually dominated by time spent inside expensive calculations. Since these can be avoided with the caching mechanism, it still improves the runtime and required computer resources a lot.
+#. The caching mechanism for calculations *should* trigger only when the inputs and the calculation to be performed are exactly the same.
```
sphuber: I think we discussed this before somewhere but cannot find it. I think the problem is worse as in that only the parser name is considered through the `parser_name` attribute. That is to say that as long as that stays the same, the parser can be changed completely and the hash wouldn't change.

docs/source/working_with_aiida/caching.rst
sphuber: Ah I see, this is for developers when implementing sub classes of `Node`.

docs/source/developer_guide/core/caching.rst
sphuber: This has been resolved

ltalirz: Coming back to this: The sentence
> For example, the ``ignored_folder_content`` keyword is used by the :class:`JobCalculation <aiida.orm.nodes.process.calculation.calcjob.CalcJobNode>` to ignore the ``raw_input`` subfolder of its repository folder.

now no longer applies, and I will simply delete it.

docs/source/working_with_aiida/caching.rst
ltalirz: @sphuber Thanks, I will rephrase this.
One question: is the node C' actually an exact copy of the node C (except for the UUID) or not (e.g. does it have different modification time etc.)?
I.e. should one even speak of the node C' being a copy of C or should one rather speak of C' as a new calculation that simply reuses the outputs of C?

docs/source/developer_guide/core/caching.rst
```diff
@@ -4,13 +4,38 @@ Caching: implementation details
...
+
+Below are some methods you can use to control caching when developing calculation and data classes:
+
+* To ignore specific attributes, a `~aiida.orm.nodes.Node` subclass can have a ``_hash_ignored_attributes`` attribute.
+  This is a list of attribute names, which are ignored when creating the hash.
+* For calculations, the ``_hash_ignored_inputs`` attribute lists inputs that should be ignored when creating the hash.
+* To add things which should be considered in the hash, you can override the :meth:`~aiida.orm.nodes.Node._get_objects_to_hash` method. Note that doing so overrides the behavior described above, so you should make sure to use the ``super()`` method.
+* Pass a keyword argument to :meth:`.get_hash <aiida.orm.nodes.Node.get_hash>`.
+  These are passed on to :meth:`aiida.common.hashing.make_hash`.
+  For example, the ``ignored_folder_content`` keyword is used by the :class:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode` to ignore the ``raw_input`` subfolder of its repository folder.
```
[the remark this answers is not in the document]
ltalirz: Ok.

docs/source/developer_guide/core/caching.rst
ltalirz: thanks, I've rephrased this.

docs/source/working_with_aiida/caching.rst
ltalirz: I don't quite follow.
So, say caching is disabled, and then I enable it by putting the `.yml` file. 
Will all my nodes be automatically hashed the next time I do anything in AiiDA?
That would be strange, no?

docs/source/working_with_aiida/caching.rst
ltalirz: Hm... so do we remove this configuration example entirely?
I realize that the terminology is probably also unclear, since the example includes Data classes but speaks of "caching".
What does it mean to disable "caching" for a data class?
Does it mean its hash is not computed? Will it affect calculations that have this Data type as an input?

docs/source/working_with_aiida/caching.rst
ltalirz: Thanks, I've rephrased this.

docs/source/working_with_aiida/caching.rst
ltalirz: >  but with this snippet people will not clear the hash of the original faulty node

It will. First of all, `node` is one of the cloned nodes, but `get_all_same_nodes()` anyhow returns all same nodes (even the node on which it is called, I checked).

docs/source/working_with_aiida/caching.rst
ltalirz: I've now moved it to the end of the list.
I still consider it a limitation, since it is something that users might naturally expect to happen when they read "caching" (I would).

docs/source/working_with_aiida/caching.rst
ltalirz: it was meant to refer to the itemized list below, but it is not really necessary.
I've removed it.

docs/source/working_with_aiida/caching.rst
ltalirz: Yes, I've seen your comment on this - I'll resolve this in order not to duplicate things

docs/source/working_with_aiida/caching.rst
ltalirz: fixed. the line can actually stay - perhaps it was the wrong indentation.

docs/source/working_with_aiida/caching.rst
sphuber: Hashing and caching are two different concepts. The hash of a node is _always_ computed and stored as an extra, regardless whether caching is enabled for that node type. That means that even when you do not have the caching config file yet, all nodes already should have their hash computed, so there is no need to call `verdi rehash`

ltalirz: Ah, thanks for the clarification. 
I thought hashing was turned on using the same switch.

docs/source/working_with_aiida/caching.rst
sphuber: I am not sure I agree. This relates to how we define "caching" in AiiDA and boils down to the same discussion as in the other thread about the configuration file enabling the caching of `Data` nodes. Since caching in AiiDA does explicitly _not_ mean the deduplication of content storage, but just and only just the skipping of executing calculations that have already been executed before, the absence of content deduplication is **not** a limitation. It is just an efficiency of the database implementation. Note that this problem exists completely independent of AiiDA. If you store a `Data` node with the same content a million times, its content will be stored on disk a million times. This is why I am a strong proponent to making more clear what we understand with "caching" in AiiDA.

docs/source/working_with_aiida/caching.rst
sphuber: I agree that caching of a `Data` node, in the way that caching is defined in AiiDA, makes no sense whatsoever. I would be for removing it and maybe adding more explanation to the introduction to stress what caching means and that it _only_ applies to the prevention of running two identical calculations twice.

The practical problem we then face is that the only classes that make sense to add to the configuration do not actually work. This is not acceptable, so I will now start to work on this issue and try to fix it a.s.a.p. You can already update this PR to remove mentioning of Data nodes being cachable and use entry point strings to target specific calc job classes.
Notation should be `aiida.calculations:quantumespresso.pw` for example.

ltalirz: Thanks, will do!

docs/source/working_with_aiida/caching.rst
sphuber: It will not be an exact copy. If you look at the implementation, it will simply overload all relevant properties (label, description, attributes and repository) from the source node to the "clone" and store it, i.e. the `ctime` and `mtime` will be different. In terms of phrasing then I am not sure if there is any difference between "C' being a copy of C" or "C' as a new calculation that simply reuses the outputs of C". In terms of implementation it is a mix of the two, but I don't think users care about implementation. Ultimately it is a copy with the exact same content that define it. The only difficulty then is defining what properties belong to the "content", e.g. the repository does, but the `ctime` and `mtime` do not.

ltalirz: Rephrased now.

docs/source/working_with_aiida/caching.rst
ltalirz: Done!

docs/source/working_with_aiida/caching.rst
ltalirz: Sure, we both agree that the lack of content deduplication is not a limitation of caching as we define it here.
However, I think most people will not have such a clear separation of these two concepts in their mind, and they might expect deduplication to happen as well.

It was already the last item in the list - I've now made it a comment after the list (no longer a list item). I think it should be kept.

docs/source/working_with_aiida/caching.rst
ltalirz: what about this:
https://github.com/aiidateam/aiida-core/blob/56e0100c8098796f8a9f8ed120be06d8f5f29b97/aiida/orm/nodes/process/calculation/calcjob.py#L106-L117

docs/source/working_with_aiida/caching.rst
sphuber: I agree that it is important and should be kept, I was just suggesting to move it higher up in the introduction. But like this is fine as well, thanks

docs/source/working_with_aiida/caching.rst
sphuber: That only includes the version of the module that contains the calculation job class though. If the parser lives in a completely different module, or it lives in the same, but its implementation was changed without the version being adapted, it will not be caught.

ltalirz:
Hopefully all fixed now ;-)

sphuber:
Thanks a lot @ltalirz !

Issue #2664, opened 2019-03-29
Store version of core and plugin in `ProcessNode` attributes


chrisjsewell:
@sphuber as discussed, I expect this will be wrapped up in a more general `Process` versioning issue soon. But I wanted to note that, because the `Parser` is decoupled from the `CalcJob`, and potentially could come from a different plugin, it would probably be good to store the version of both

sphuber:
The current idea would be the following.
Whenever a node get stores, the engine will try to retrieve a version number from the package in which the class is defined. For the data classes this will go through the entry points and for the process nodes, it should go through the entry points of the process class that created it. The interpreter should build an internal map in a cache that maps the class to a version number of the plugin that it belongs to. This version number is then stored as an attribute. For `CalcJobs`, as @chrisjsewell rightfully mentioned, the `Parser` version should also be explicitly mentioned as it is decoupled and does not have to be in the same plugin.

For the moment, we will not allow to override the version number on a per-class basis. This approach is simple enough with respect to the required implementation in `aiida-core` as well as what is required from plugins. The developers of the latter will just have to be aware of this mechanism and update the version number and release it whenever a change in any of the relevant classes merits a different version number on the nodes. If this is the case, then it would be more than reasonable to also expect a bump in the plugin version.

A downside of this approach is that this would only work if the plugin is not used from a local copy that has unstored changes or unreleased commits. A potential solution would require the core code to start dynamically inspecting the packages to see if they are from repositories and then need to use git libraries to fetch current commits and/or even local changes. This will open a whole can of worms. Besides, this should only happen in development environments, in which case it is not wholly unreasonable that the provenance cannot be perfectly kept.

sphuber:
See also #126

sphuber:
The minimal required functionality for `v1.0.0` is the following:
* All `ProcessNodes` get a `version_core` and `version_plugin` attribute
* The value of these two attributes is based on the version of the package, i.e. `aiida.__version__` and e.g. `aiida_quantumespresso.__version__`
* If the plugin does not specify this property, no attribute is set at all
* The properties on the node for retrieving this value will simply return `None` if it does not exist.
* The runner responsible for creating the `ProcessNodes` will lazily build a mapping of objects (`CalcJob` and `WorkChain` classes, and `calcfunction` implementations) to the corresponding module version.

PR #3116, opened 2019-07-02
Properly overload function module and name to `FunctionProcess`

### LINKED ISSUE #3115: Module class and name of process functions not properly defined
The `process_type` of a process function is meant to be set to the corresponding entry point string, based on its module and name if it exists, or otherwise the concatenation of module and function name. When the function is ran, the process node created will get the value for `process_type` by calling `build_process_type` which tries to do exactly this. However, the `FunctionProcess` class that is dynamically built for the decorated function does not correctly inherit the `__module__` and `__name__` attribute. Therefore, for example for the function:
```
from aiida_quantumespresso.workflows.functions.create_kpoints_from_distance import create_kpoints_from_distance
```
which should be getting the `process_type`:
```
aiida_quantumespresso.workflows.functions.create_kpoints_from_distance.create_kpoints_from_distance
```
instead gets the generic
```
abc.create_kpoints_from_distance
```

sphuber:
Fixes #3115 

The `FunctionProcess` class that is dynamically built based on the
function that is decorated with the `calcfunction` or `workfunction`
decorator was not overloading the `__module__` and `__name__` attributes
of the wrapped function. Since the `process_type` attribute of the
process node is defined by those attributes, an incorrect basic value
was being stored of the form `abc.name_of_wrapped_function`.

### DIFF
--- a/aiida/backends/tests/engine/test_process_function.py
+++ b/aiida/backends/tests/engine/test_process_function.py
@@ -131,6 +131,12 @@ class TestProcessFunction(AiidaTestCase):
         self.assertEqual(node.is_finished_ok, True)
         self.assertEqual(node.is_failed, False)
 
+    def test_process_type(self):
+        """Test that the process type correctly contains the module and name of original decorated function."""
+        _, node = self.function_defaults.run_get_node()
+        process_type = '{}.{}'.format(self.function_defaults.__module__, self.function_defaults.__name__)
+        self.assertEqual(node.process_type, process_type)
+
     def test_exit_status(self):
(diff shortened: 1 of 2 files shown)


sphuber:
I just discovered this while starting work on #2664 . This PR fixes the problem, but data generated with this bug will have lost information. It is not clear yet when this bug was introduced, but anyway I don't think it is possible to restore the data in a deterministic manner since the bug is lossy in nature.
For example, what should have been:
```
aiida_codtools.workflows.functions.primitive_structure_from_cif.primitive_structure_from_cif
```
has become due to the bug:
```
abc.primitive_structure_from_cif
```

PR #3118, opened 2019-07-03
remove stale daemon pid in status check

unkcpz:
The problem (https://github.com/aiidateam/aiida_core/pull/2795) that stale daemon PID not found after reboot when daemon is not stop correctly also occurred in command `verdi daemon status` and `verdi status`.

Also calling function `delete_stale_pid_file` in verdi status check.

Or maybe it is better to print tips on how to solve the problem in `get_daemon_status` rather than remove files automatically?


sphuber:
> Or maybe it is better to print tips on how to solve the problem in `get_daemon_status` rather than remove files automatically?

The problem with this approach is that it is difficult to determine the source of the problem when a call to daemon times out. Note that the `delete_stale_pid_file` should only remove the pid file if the corresponding process can no longer be found. This should be fine in principle. Did you find a situation where the pid file was removed unjustly? If not, why do you think we should remove these lines?

unkcpz:
I didn't find situation where the pid file was removed unjustly.

My PR suggests also adding `delete_stale_pid_file` in `verdi status` and `verdi daemon status`.

sphuber:
Ah you are right, I glossed fast over the diff and interpreted it exactly the other way around.

unkcpz:
I think this is not an urgent PR, but it can be confusing for users, especially beginners.

Issue #1762, opened 2018-07-18
Immutability and export


szoupanos:
The majority of the entity instances that are exported and imported are immutable. However, when an instance can change its values after storing it, problems can arise. The biggest problem is how we discover such changes and how we treat them.

For example, let a code C that exists with the same UUID at the databases of users U1 and U2.
U1 decides to update the code (prepend text) and use it for some calculations.
U2 uses the code without any change.
When exchanging data between U1 and U2, the import code will not update the code node because the UUID exists and it is already in the user's database. However, the code is not the same.

How should we address such issues (but also similar issues)?
Should we check the UUID of the immutable entities and the hashes of the rest?
How sealable entities can help us on the above problem?

This ticket is initially intended for discussion and to take decisions on this subject.

Idea below sealing:
- Before store you can change everything
- After store:
  - for Data, you cannot change anything anymore (this is enforced in aiida 1.0.0aX, not in 0.12.X)
  - the same should happen for Code, but not sure it if it is the case
  - for Calculations, you change only the updatable attrs
- After sealing (the concept exists only for Calculations), you can not change anything
A note: the logic of Sealing was fixed/improved in 1.0.0aX

giovannipizzi:
Things to do (these can be moved to smaller issues once it's all clear):
- [ ] document the logic of sealing. Also, note that there are notable exceptions to what can be changed. E.g., label and description. The logic should be already in the code for caching/hashing. We should document this as well in the same section, and define the behaviour (e.g. overwrite the label, ignore the imported label, ...)
- [ ] make sure Code sealing works as for codes (connected to the issue of having Code be a subclass of Data #278)
- [ ] ensure we don't provide command line options to modify entities (e.g. `verdi computer update`, `verdi code update`; check also if there are API calls that would allow this). Note: if they exist, it's because this is useful/easy for the user. We shouldn't just disable those features, but replace with convenient commands to achieve the same goal but in the correct way (e.g. `verdi computer duplicate` instead of `verdi computer update`, etc.). Important to check also non-node entities where the Sealing concept does not exist (i.e. they should behave like data).
- [ ] have a flag to hash the nodes that are being imported and check if they changed (and in this case warn the user/give the option e.g. to keep the existing data, or replacing with imported one.
   - note that this should not happen unless the user messes up with its db (but might happen e.g. if there are data migrations and one is importing between different AiiDA versions...)
[...]

PR #3134, opened 2019-07-06
Add data migration for legacy process attributes

### LINKED ISSUE #3120: Missing data migration for change in sealed key from `_sealed` to `sealed`
This means that for older processes that still have the `_sealed` attribute, the `ProcessNode.is_sealed` will return `False`, which is problematic

### LINKED ISSUE #3157: Export version 0.7
Things to do/include for export version 0.7:

"Easy":
- [x] Remove `Attributes` and `Links` from `metadata.json` of export archives in migration
- [x] Remove last remnants of `Attributes` and `Links` entities in import functions
- [x] Make sure only sealed process Nodes are exported
- [x] Migration 0040 - Data migration legacy Process attributes

"Hard":
- ~Make import functions backend-agnostic~ (will be postponed - may not even change the version!)

sphuber:
Fixes #3120 
Fixes #3157 

Attribute keys that are renamed:

  * `_sealed` -> `sealed`

Attribute keys that are removed entirely:

  * `_finished`
  * `_failed`
  * `_aborted`
  * `_do_abort`

The `sealed` key is still used, the key has just changed, so existing
attributes are migrated. The last four attributes are left over control
attributes of one of the first process based engines, but are obsolete
and are therefore deleted to clean up some space.


giovannipizzi:
Should we seal very old JobCalculations that were run by the old exec manager and didn't know about the sealed concept? (any calc that happens to be in a final state but is not sealed - I think these can be safely sealed?)

sphuber:
> Should we seal very old JobCalculations that were run by the old exec manager and didn't know about the sealed concept? (any calc that happens to be in a final state but is not sealed - I think these can be safely sealed?)

What would you filter on though? All nodes starting with `process.%` that do not have the `sealed` attribute? How do we exclude calculations that are currently still active (even though we explicitly tell people to finish running everything before migrating, but you know..) ? I guess we could exclude any of the modern process states, i.e. any node that has `process_state` attribute with value `created`, `running` or `waiting`.

Something like this:
```
UPDATE db_dbnode
SET attributes = jsonb_set(attributes, '{"sealed"}', to_jsonb(True))
WHERE
    node_type LIKE 'process.%' AND
    NOT (attributes ? 'sealed') AND
    attributes->>'process_state' = ANY(array['created', 'running', 'waiting', 'finished']);
```

Am I missing something?

giovannipizzi:
Yes for the SQL, but we will need also an export file migration, anyway :-(

sphuber:
@CasperWA would you be willing to add a commit on top that implements an export archive migration for the data migration I added here? Pretty please? :)

CasperWA:
> @CasperWA would you be willing to add a commit on top that implements an export archive migration for the data migration I added here? Pretty please? :)

Derp to the rescue!

image

CasperWA:
@sphuber, one question: If a 'process.*' Node is found in an export archive. Is there ever a case where it should _not_ be sealed?

sphuber:
> @sphuber, one question: If a 'process.*' Node is found in an export archive, is there then ever a case where it should _not_ be sealed?

If all export archives are 100% consistent and correct, then no. Because the rules state that only sealed nodes are exportable. So by that definition, any unsealed nodes in archives are "illegal" and should be sealed anyway. The exception in the database migration is for new and actually "active" processes, that we do not want to seal. But those _should_ anyway not be present in the archive. So I think we can safely just add the `sealed` key to all process nodes in existing archives (and remove `_sealed` and the other deprecated keys).

giovannipizzi:
Question: is this check on exporting only sealed processes already implemented?

CasperWA:
> Question: is this check on exporting only sealed processes already implemented?

A quick search for "sealed", "seal", and "store" reveals no instances of it in the `export_tree` and `export` functions, nor in any of the export utility functions.
From this, I deduce that `.is_sealed` or similar is never checked.

I presume there are more intricate ways of checking if nodes are sealed, or simply some implicit ways. It would take a lot more time for me to dig around and check if these are present.

sphuber:
I am pretty sure that just means it is actually never checked. This goes back why #1762 really is so important and blocking for 1.0.0.  This means we cannot assume that exported nodes were always going to be sealed.

Maybe the best thing would be to actually include the check that nodes are sealed before they are exported in this PR and then in the export migration always add the `sealed` attribute if it doesn't yet exist and potentially emit a warning. Anyway, even if people exported a recent active process node, I think it is fine to say that it won't be able to be continued to run when imported. Thoughts @giovannipizzi ?

giovannipizzi:
I agree - I would have been indeed surprised if the logic was already there.
My suggestion is to put the same check we do in the DB also on the export file, at this stage.
If we add a check on what is exportable in this PR is maybe ok, but anyway not sufficient (we need to be able to migrate also old export files) and maybe it's better to implement the logic while fixing #1762 with a proper fix after we make the final decisions.

CasperWA:
I have updated/rewritten my original commit to now up the EXPORT_VERSION to 0.7. This is accompanied by the recently upped and released `aiida-export-migration-tests` module/repo that now includes archives up to v0.7.

#### The original migration ...
... has been made a migration that is part of the step from 0.6 to 0.7.
It will add a `"sealed": true` attribute to all ProcessNodes.
The only time it will _not_, is when the `"sealed"` or `"_sealed"` attribute already existed in the archive and was _not_ `true`. At this point the migration will raise `IntegrityError` stating that the export archive is corrupt, hence, cannot be migrated to the newest export version, being excluded from being imported.
Yes, one can always manually change this in the archive, but this is something users can _always_ do - as well as changing the DB directly, if they so choose.

#### As an added bonus ...
... I have removed the `Attribute` and `Link` entities from `metadata.json` for v0.7 archives, resulting in the fact that I could remove a couple of lines in total for the import functions.

giovannipizzi:
A comment: I think it never happens that that sealed attribute is there and False. By default IIRC unsealed nodes just don't have it, and the attribute is added upon sealing. Old style calcjobs never got the attribute. 
@sphuber am I right?

@CasperWA do you prefer to have a different logic than in the SQL statements for a specific reason? I prefer to have the same logic if possible, so that you get the same result if you migrate your db or import an export file from an old version into a new version.

giovannipizzi:
sorry, pressed the wrong button

sphuber:
> A comment: I think it never happens that that sealed attribute is there and False. By default IIRC unsealed nodes just don't have it, and the attribute is added upon sealing. Old style calcjobs never got the attribute.
> @sphuber am I right?

Yes you are right, I was just discussing this with @CasperWA suggesting that he keeps the logic of the migration. That is to say, if `_sealed` exists, just rename to `sealed`. Then for all process nodes that still don't have `sealed`, simply add it with value `True`. This should be the logic as `False` would have never been set.

CasperWA:
Okay. So now, I'll simply remove all "old" attributes, and create `"sealed" = true`, if `"sealed"` does not already exist.
The only raise is, if by some chance a Node was listed in the attributes dict, but not in the export_data dict.

aiida/backends/tests/tools/importexport/migration/test_v06_to_v07.py
```diff
@@ -0,0 +1,144 @@
...
+        metadata, data = get_json_files('export_v0.6_manual.aiida', **self.external_archive)
+        verify_metadata_version(metadata, version='0.6')
+
+        # Migrate to v0.7
+        migrate_v6_to_v7(metadata, data)
+        verify_metadata_version(metadata, version='0.7')
+
+        self.maxDiff = None  # pylint: disable=invalid-name
+        # Check attributes of process.* nodes
+        illegal_attrs = {"_sealed", "_finished", "_failed", "_aborted", "_do_abort"}
```
sphuber: please consistently use single quotes for string identifiers

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -658,16 +659,22 @@ def export_tree(what,
                 else:
                     groups_uuid[str(res[0])] = [str(res[1])]
 
+    #######################################
+    # Final check for unsealed ProcessNodes
+    #######################################
+    process_nodes = set()
+    for node_pk, content in export_data.get(NODE_ENTITY_NAME, {}).items():
+        if content['node_type'].startswith('process.'):
+            process_nodes.add(node_pk)
```
sphuber: Do we not have the node instances that are going to be exported at some point? If so, it would be better to keep that as a set and pass them to `check_process_nodes_sealed` such that there they won't have to be reloaded again. Never mind: I just saw the implementation of the function that you filter on the pks. I guess this is fine then

aiida/backends/tests/cmdline/commands/test_export.py
```diff
@@ -68,6 +70,9 @@ def setUpClass(cls, *args, **kwargs):
 
         # Utility helper
         cls.fixture_archive = 'export/migrate'
+        cls.newest_archive = "export_v{}_simple.aiida".format(EXPORT_VERSION)
+        cls.penultimate_archive = "export_v{}_simple.aiida".format(
+            round(float(EXPORT_VERSION) - 0.1, len(EXPORT_VERSION) - len("0.")))
```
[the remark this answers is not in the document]
CasperWA: Yeah, I know.
If you have any suggestions here, they are more than welcome!
I tried making it general and nice, but the floating point numbers made it quite difficult to make it precise (e.g. hitting 0.799999999 instead of 0.8).
I could just go back to a more hard-coded way of doing it?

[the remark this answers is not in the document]
CasperWA: Have now hardcoded it (it is not the most important thing to remember to update, luckily).
Furthermore, I have removed the same logic that was present in the test `test_wrong_versions` in `aiida.backends.tests.tools.importexport.migration.test_migration.TestExportFileMIgration`. Here it was supposed to add 0.1 to the current export version. Instead, I simply added 0.99.

aiida/backends/tests/tools/importexport/migration/test_v05_to_v06.py
```diff
@@ -116,3 +138,19 @@ def test_migrate_v5_to_v6_datetime(self):
         else:
             raise RuntimeError('the archive `export_v0.5_simple.aiida` did not contain a node with the attribute '
                                '`scheduler_lastchecktime` which is required for this test.')
+
+    def test_migrate_v5_to_v6_complete(self):
+        """Test migration 0040 - Data migration for some legacy process attributes"""
```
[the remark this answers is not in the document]
CasperWA: Good catch. This was also the case for the similar test in v06 to v07.

sphuber:
This is good to go as far as I am concerned. Have at it @giovannipizzi if you would

aiida/tools/importexport/migration/v06_to_v07.py
```diff
@@ -0,0 +1,99 @@
...
+    for node_pk, content in data['node_attributes'].items():
+        try:
+            if data['export_data']['Node'][node_pk]['node_type'].startswith('process.'):
+                # Remove attributes
+                for attr in attrs_to_remove:
+                    content.pop(attr, None)
+
+                # Set `'sealed' = True`, if 'sealed' does not already exist
+                if 'sealed' not in content:
+                    content['sealed'] = True
```
giovannipizzi: I know this has been already discussed, but is there a real reason not to put the *same* logic checking the state (as in the migrations) also here?

aiida/backends/djsite/db/migrations/0040_data_migration_legacy_process_attributes.py
```diff
@@ -0,0 +1,88 @@
...
+                UPDATE db_dbnode SET attributes = attributes - '_aborted'
+                WHERE attributes ? '_aborted' AND node_type LIKE 'process.%';
+                -- Delete `_aborted`
+
+                UPDATE db_dbnode SET attributes = attributes - '_do_abort'
+                WHERE attributes ? '_do_abort' AND node_type LIKE 'process.%';
+                -- Delete `_do_abort`
+
+                UPDATE db_dbnode
+                SET attributes = jsonb_set(attributes, '{"sealed"}', to_jsonb(True))
```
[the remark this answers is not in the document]
CasperWA: @sphuber

aiida/tools/importexport/migration/v06_to_v07.py
CasperWA: One would never have been able to export a running Node, I presume? So no reason to check state.

giovannipizzi: Well, I'm pretty sure that even in the current develop version it *is* possible, that's why I'd prefer to do the same check. The goal of the data migration is to mirror the SQL migrations, so I'd rather go for explicitly implementing the same logic (that, BTW, isn't so complex in this case) rather than using assumptions that might be not true in some cases

aiida/backends/djsite/db/migrations/0040_data_migration_legacy_process_attributes.py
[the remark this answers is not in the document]
sphuber: done

aiida/backends/djsite/db/migrations/0040_data_migration_legacy_process_attributes.py
```diff
@@ -0,0 +1,88 @@
...
+                WHERE
+                    node_type LIKE 'process.%' AND
+                    NOT (attributes ? 'sealed') AND
+                    attributes->>'process_state' != ANY(array['created', 'running', 'waiting']);
```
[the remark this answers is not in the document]
sphuber: Added the test

sphuber:
@giovannipizzi @CasperWA I added the tests for both backends and indeed the migration failed. Good catch Gio :+1: The migrations have now been fixed. So after Casper updates the conditional in the export migration we are good to go

aiida/tools/importexport/migration/v06_to_v07.py
CasperWA: This is fine, I will implement this.
My only question is now - what should the migration do for the case where a Node is in an active state? Since we should not allow these ProcessNodes to be imported, I don't want the migration to create a valid, importable archive.
For this case, shouldn't we acknowledge the difference between the archive and the database and simply raise, stating that the archive is corrupt - print out the affected ProcessNodes, and have them contact us? Or perhaps even (dangerous) give them the option to override the process_state to finished and "seal" the Node...?

CasperWA: > The goal of the data migration is to mirror the SQL migrations, (...)

Yes, I agree, but only to the extent of what is "allowed" to export/import.
If we agree _only_ sealed ProcessNodes can be exported from here on out (something that this PR is implementing), then we should consider what to do when migrating exported ProcessNodes that could clearly never have been sealed when originally exported (i.e., they have an active `process_state`).
Personally, I think the archive should be marked as corrupted, and they should contact us to sort it out (as I wrote above). But perhaps we cannot come up with any case, where it is _not_ okay to just change the `process_state` to `"finished"`, set the `sealed` attribute and be done with it?

In any case - I am sorry for writing two almost similar comments here, I just felt like the first one didn't properly explain my position, but perhaps I was wrong :)

CasperWA:
The latest push, was to address the review of @giovannipizzi concerning the application of migration 0040 on the export archives. However, due to the "hole" mentioned above, the tests will fail when applying this to very old export archives that have other process nodes than JobCalculation.

CasperWA:
@sphuber, I am done with this at this point I believe. Will not be pushing anything more now until you give an all clear in any case.
I have a weird case though, where my django test profile suddenly stops working during the tests and reverts to database version 1.0.35...? Let's see what happens on Travis. (It worked fine for the sqla test profile).

aiida/tools/importexport/migration/v06_to_v07.py
```diff
@@ -0,0 +1,126 @@
...
+                else:
+                    # Either the ProcessNode is in a non-active state or its 'process_state' hasn't been set.
+                    # In both cases we claim the ProcessNode 'sealed' and make it importable.
+                    content['sealed'] = True
+
+                # Remove attributes
+                for attr in attrs_to_remove:
+                    content.pop(attr, None)
+        except KeyError as exc:
+            raise IntegrityError("Your export archive is corrupt! Org. exception: {}".format(exc))
```
giovannipizzi: I don't like too much `Org.` for `Original`...

aiida/tools/importexport/migration/v06_to_v07.py
```diff
@@ -0,0 +1,126 @@
...
+        try:
+            if data['export_data']['Node'][node_pk]['node_type'].startswith('process.'):
+                # Check if the ProcessNode has a 'process_state' attribute, and if it's non-active.
+                # Raise if the ProcessNode is in an active state, otherwise set `'sealed' = True`
+                process_state = content.get('process_state', '')
+                if process_state in active_states:
+                    # The ProcessNode is in an active state, and should therefore never have been allowed
+                    # to be exported. The Node will be added to a log that is saved in the working directory,
+                    # then an IntregityError will be raised, since the archive needs to be migrated manually.
+                    uuid_pk = data['export_data']['Node'][node_pk].get('uuid', node_pk)
```
giovannipizzi: Do we really have cases where there is no `uuid` and be safe we need to use the `node_pk`?

CasperWA: Hopefully not, but for what is supposed to be a helpful log for us/the user, I'd rather try and catch the ball as many times as possible, so to say. However, if the identifier was used in the actual code, it should definitely be more strict, i.e. only use either UUID or PK.

PR #3135, opened 2019-07-07
Update documentation of querying/querybuilder/append for 1.0.0b4

atztogo:
Fixed some of the first examples in `querying/querybuilder/append.html` that doesn't work at the latest aiida-core. I roughly checked working for those examples above **Attributes and extras** section.

Probably users will come to this page once number of data becomes huge and `verdi process list ...` can't be of much help. In this case, **Limiting the number of results** section is useful because it may be the first easiest one coming to their minds than advanced query. So I think putting  **Limiting the number of results** section more above could help users.


docs/source/querying/querybuilder/append.rst
```diff
@@ -127,7 +127,7 @@ This will be the next example::
     qb.append(
         CalcJobNode,
         filters={
-            'state':{'in':['FINISHED', 'RETRIEVING']}
+            'attributes.process_state':{'in':['finished', 'retrieving']}
```
sphuber: `retrieving` is not a valid `process_state`. Please replace it by `excepted` and or `killed`, which are valid `ProcessState` values.

docs/source/querying/querybuilder/append.rst
```diff
@@ -140,7 +140,7 @@ So, to ask for all calculations that are not in 'FINISHED' or 'RETRIEVING'::
     qb.append(
         CalcJobNode,
         filters={
-            'state':{'!in':['FINISHED', 'RETRIEVING']}
+            'attributes.process_state':{'!in':['finished', 'retrieving']}
```
sphuber: Same here

docs/source/querying/querybuilder/append.rst
atztogo: Thanks. I pushed this fix to my clone's develop. I don't know the correct manner of PR. Was this OK?

sphuber: This was fine. For your own convenience, typically it is nicer to make a specific branch for a fix. You can do that by doing `git checkout -b fix_some_isse` and make your commits there. Then just push that branch to your fork and you can create the PR from there. Then there is one final thing: as you can see now, your branch is "out-of-data", meaning our develop has commits that your branch does not have yet. Typically when you push your PR branch, make sure it is up to date by checking out your branch and doing `git fetch origin && git rebase origin/develop`. This will update your branch to include the latest changed of our develop

atztogo: OK. I do it the next. Thanks for your explanation!

PR #3146, opened 2019-07-08
Make the `execmanager.upload_calculation` idempotent'ish

sphuber:
Fixes #3142

The `upload_calculation` would cause an exception if called multiple
times for the same calculation, which can happen if the first time that
the runner was working on it got interrupted, for example due to a
daemon shutdown. The reason is that the second time around the adding of
the `remote_folder` data node will raise a uniqueness exception,
because there can only be one output with the same label.

Note that full idem-potency is impossible, but this change should make
the problem a lot less likely to occur. The idea is to delay the actual
attaching of the remote folder data node to the last moment possible.
This way, if the method is called again and the folder is already there,
we can be reasonably sure that the files were already retrieved
successfully and we simply return, leaving the call a no-op. This is
done in the beginning of the function to check if the output node already
exists using the `LinkManager.first()` call. If the node exists, the
upload function has apparently already been called before and reached
the end of the function where it adds the remote folder. This means all
the files were already successfully uploaded so we can safely skip it.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -42,10 +42,19 @@ def upload_calculation(node, transport, calc_info, script_filename, dry_run=Fals
     :param transport: an already opened transport to use to submit the calculation.
     :param calc_info: the calculation info datastructure returned by `CalcJobNode.presubmit`
     :param script_filename: the job launch script returned by `CalcJobNode.presubmit`
+    :return: tuple of ``calc_info`` and ``script_filename``
     """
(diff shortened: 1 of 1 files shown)


sphuber:
@giovannipizzi this is now unblocked and ready for review

giovannipizzi:
Could you put a comment also here when calling `add_incoming` that this should be the last thing to do in the function? (And also in this case put a note in the issue about transactions)

sphuber:
Done. I also noticed I made a mistake in the return value of the initial check. Now it also returns the correct tuple

PR #3153, opened 2019-07-09
Add commands `verdi node attributes` and `verdi node extras`

chrisjsewell:
The commands follow the same logic as `verdi calcjob res`. As such the options (for filtering and formatting dictionaries) have been standardised.
Also an additional 'yaml' format option has been added to `echo_dictionary`.


aiida/cmdline/commands/cmd_node.py
```diff
@@ -162,6 +162,36 @@ def show(nodes, print_groups):
...
+        if keys is not None:
+            attr_dict = {k: v for k, v in attr_dict.items() if k in keys}
+        echo.echo_dictionary(attr_dict, fmt=fmt)
+
+
+@verdi_node.command('extras')
+@arguments.NODES()
+@options.DICT_KEYS()
+@options.DICT_FORMAT()
+@with_dbenv()
```
giovannipizzi: Thanks a lot!
Could you add a `--raw` option (we should have it already) that instead of printing `Pk: XXX` before every line just creates a dictionary `{"PK1": load_node(PK1).attributes, "PK2": load_node(PK2).attributes, ...)` and prints only that, so you can pipe it into another code able to read JSON or YAML? (Now one would have to grep away `Pk:`)

Same for `attributes`

aiida/cmdline/params/options/__init__.py
```diff
@@ -391,3 +391,17 @@ def active_process_states():
 
 CONFIG_FILE = ConfigFileOption(
     '--config', help='Load option values from configuration file in yaml format.')
+
+DICT_FORMAT = OverridableOption(
+    '-f', '--format', 'fmt',
+    type=click.Choice(["json+date", "yaml"]),
+    default="json+date",
+    help="format for the output"
```
sphuber: Please use a proper sentence for the help string just like all the other options, i.e. start with a capital and end with a full stop

aiida/backends/tests/cmdline/commands/test_node.py
```diff
@@ -0,0 +1,93 @@
...
+from aiida.backends.testbase import AiidaTestCase
+from aiida.cmdline.commands import cmd_node as command
+
+
+def get_result_lines(result):
+    return [e for e in result.output.split('\n') if e]
+
+
+class TestVerdiCalculation(AiidaTestCase):
+    """Tests for `verdi calcjob`."""
```
sphuber: wrong docstring

aiida/backends/tests/cmdline/commands/test_node.py
```diff
@@ -0,0 +1,93 @@
...
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida_core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+# pylint: disable=protected-access,too-many-locals,invalid-name,too-many-public-methods
+"""Tests for `verdi calcjob`."""
```
sphuber: wrong docstring

aiida/backends/tests/cmdline/commands/test_node.py
```diff
@@ -0,0 +1,93 @@
...
+from aiida import orm
+from aiida.backends.testbase import AiidaTestCase
+from aiida.cmdline.commands import cmd_node as command
+
+
+def get_result_lines(result):
+    return [e for e in result.output.split('\n') if e]
+
+
+class TestVerdiCalculation(AiidaTestCase):
```
sphuber: wrong class name

aiida/cmdline/utils/echo.py
```diff
@@ -169,7 +169,8 @@ def echo_dictionary(dictionary, fmt='json+date'):
     :param dictionary: the dictionary
     :param fmt: the format to use for printing, valid options: ['json+data']
     """
-    valid_formats_table = {'json+date': _format_dictionary_json_date}
+    from aiida.orm.utils.serialize import serialize
+    valid_formats_table = {'json+date': _format_dictionary_json_date, 'yaml': serialize}
```
giovannipizzi: Could you move this as a module variable (capitalised, e.g. called `VALID_DICT_FORMATS_MAPPING`) and then reuse it here, but also when you do `type=click.Choice(["json+date", "yaml"])` you replace this with `VALID_DICT_FORMATS_MAPPING.keys()`?

Also, could you add the `format` option to the other functions calling the function (by grepping `json+date` we found at least `verdi data array` and `verdi data dict`

aiida/backends/tests/cmdline/commands/test_node.py
```diff
@@ -0,0 +1,93 @@
...
+        """Test verdi node extras"""
+        options = [str(self.node.uuid)]
+        result = self.cli_runner.invoke(command.extras, options)
+        self.assertIsNone(result.exception, result.output)
+        self.assertIn(self.EXTRA_KEY_ONE, result.output)
+        self.assertIn(self.EXTRA_VAL_ONE, result.output)
+        self.assertIn(self.EXTRA_KEY_TWO, result.output)
+        self.assertIn(self.EXTRA_VAL_TWO, result.output)
+
+        for flag in ['-k', '--keys']:
```
sphuber: Add a test for the format flag

aiida/cmdline/params/options/__init__.py
```diff
@@ -393,3 +393,23 @@ def active_process_states():
 
 CONFIG_FILE = ConfigFileOption(
     '--config', help='Load option values from configuration file in yaml format.')
+
+IDENTIFIER = OverridableOption(
+    '-id', '--identifier', 'identifier',
```
sphuber: We should stick to the rule of using only a single letter for short form options i.e. `-id` -> `-i` or something like that

aiida/cmdline/utils/echo.py
```diff
@@ -196,6 +181,25 @@ def default_jsondump(data):
...
 
+VALID_DICT_FORMATS_MAPPING = OrderedDict((('json+date', _format_dictionary_json_date), ('yaml', serialize)))
+
+
+def echo_dictionary(dictionary, fmt='json+date'):
+    """
+    Print the given dictionary to stdout in the given format
+
+    :param dictionary: the dictionary
+    :param fmt: the format to use for printing, valid options: ['json+data']
```
sphuber: typo in `'json+data'` but also incorrect, right? `yaml` is also a valid option? Did you mean that is the default perhaps?

aiida/cmdline/commands/cmd_node.py
sphuber: Do I understand correct that for `yaml` the format of the actual extras or attributes content is still printed in `json`? If so, that would be a bit unexpected, no?

aiida/backends/tests/cmdline/commands/test_node.py
```diff
@@ -19,8 +19,36 @@
 from aiida.cmdline.commands import cmd_node
 
 
-class TestVerdiNode(AiidaTestCase):
-    """Tests for `verdi node`"""
+def get_result_lines(result):
+    return [e for e in result.output.split('\n') if e]
+
+
+class TestVerdiCalculation(AiidaTestCase):
```
sphuber: Nitpick, `TestVerdiCalculation` -> `TestVerdiNode`

aiida/cmdline/commands/cmd_node.py
sphuber: Ah ok, good to know, thanks

sphuber:
Thanks @chrisjsewell for me this is good to go. @giovannipizzi you want to have a final looksy?

sphuber:
Thanks a lot @chrisjsewell

chrisjsewell:
Ta

PR #3155, opened 2019-07-10
Fix bug with never resolving calculation job update futures

### LINKED ISSUE #2388: Calculation stuck on waiting
Some calculations get stuck in `Waiting` state with the process status: `Waiting for transport task: update`.
All the calculations are however complete on Daint. There's no particular error in the daemon log and restarting the daemon sometimes retrieves one or two calculations but not always.

Edit:
version 1.0.0a4

### LINKED ISSUE #3136: Daemon runner can appear to be inactive and not retrieve finished jobs
This is observed for daemon runners with jobs that spend a long time in the queue. My suspicion is that a problem with the transport, which triggers the exponential backoff mechanism, renders a component of the runner disabled. The activity of the runner reported by circus seems low, so it is doing something. Maybe just the backup polling mechanism, but it is not retrieving jobs that are completed on the remote for sure. Simply restarting the runner, will cause it to reload the processes and it will start retrieving the jobs.

sphuber:
Fixes #3136 
Fixes #2388 

The `JobsList` class is a container and manager of requests for updates
of the status of calculation jobs. For each request, it keeps a future
that it will update in bulk after retrieving the status of all the jobs
that it manages. This is necessary to ensure that multiple jobs use the
same update call to scheduler and not all individually, in order to not
overload the scheduler.

If the scheduler update call excepts, all the futures were properly
excepted as well and the internal cache of outstanding requests would be
cleared, before re-raising the exception. The excepted futures would
each be caught by the exponential backoff mechanism and a new request to
update the status would be scheduled. This would be registered with the
same `JobsList` instance, however, the next scheduler update would never
be called and so the futures would never be resolved, causing the
calculation jobs to "hang" indefinitely.

The problem lies in the `_ensure_updating` method. When a new request
comes in, this is the method that is called first to see if a scheduler
update call has already been scheduled. If the case, a new one does not
have to be scheduled. This is based on whether the `_update_handle`
attribute is defined. This attribute is properly reset after a scheduler
update goes through successfully, in the `updating` coroutine after the
`_update_job_info` yields. However, in the case of an exception in that
last method, the code path that resets the `_update_handle` is never
reached because the exception is re-raised after the job requests cache
is reset. The solution is simply to manually set `_update_handle` to
`None` in the except clause.

### DIFF
--- a/aiida/engine/processes/calcjobs/manager.py
+++ b/aiida/engine/processes/calcjobs/manager.py
@@ -13,6 +13,7 @@ from __future__ import print_function
 from __future__ import absolute_import
 
 import contextlib
+import logging
 import time
 
 from six import iteritems, itervalues
@@ -20,7 +21,6 @@ from tornado import concurrent, gen
 
 from aiida import schedulers
 from aiida.common import exceptions, lang
-from aiida.common.log import AIIDA_LOGGER
 
 __all__ = ('JobsList', 'JobManager')
 
@@ -57,7 +57,7 @@ class JobsList(object):  # pylint: disable=useless-object-inheritance
         self._authinfo = authinfo
         self._transport_queue = transport_queue
         self._loop = transport_queue.loop()
-        self._logger = AIIDA_LOGGER.getChild('calcjobs')
+        self._logger = logging.getLogger(__name__)
 
         self._jobs_cache = {}
         self._job_update_requests = {}  # Mapping: {job_id: Future}
@@ -97,6 +97,7 @@ class JobsList(object):  # pylint: disable=useless-object-inheritance
         :rtype: dict
         """
         with self._transport_queue.request_transport(self._authinfo) as request:
+            self.logger.info('waiting for transport')
             transport = yield request
 
             scheduler = self._authinfo.computer.get_scheduler()
@@ -147,6 +148,13 @@ class JobsList(object):  # pylint: disable=useless-object-inheritance
             for future in itervalues(self._job_update_requests):
                 if not future.done():
                     future.set_exception(exception)
+
+            # Reset the `_update_handle` manually. Normally this is done in the `updating` coroutine, but since we
+            # reraise this exception, that code path is never hit. If the next time a request comes in, the method
+            # `_ensure_updating` will falsely conclude we are still updating, since the handle is not `None` and so it
+            # will not schedule the next update, causing the job update futures to never be resolved.
+            self._update_handle = None
+
             raise
         else:
             for job_id, future in iteritems(self._job_update_requests):
--- a/.pylintrc
+++ b/.pylintrc
@@ -50,7 +50,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=bad-continuation,locally-disabled,useless-suppression,django-not-available,bad-option-value
+disable=bad-continuation,locally-disabled,useless-suppression,django-not-available,bad-option-value,logging-format-interpolation
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/aiida/engine/processes/calcjobs/tasks.py
+++ b/aiida/engine/processes/calcjobs/tasks.py
@@ -61,7 +61,7 @@ def task_upload_job(node, transport_queue, calc_info, script_filename, cancellab
     :raises: TransportTaskException if after the maximum number of retries the transport task still excepted
     """
     if node.get_state() == CalcJobState.SUBMITTING:
-        logger.warning('calculation<{}> already marked as SUBMITTING, skipping task_update_job'.format(node.pk))
+        logger.warning('CalcJob<{}> already marked as SUBMITTING, skipping task_update_job'.format(node.pk))
         raise Return(True)
 
     initial_interval = TRANSPORT_TASK_RETRY_INITIAL_INTERVAL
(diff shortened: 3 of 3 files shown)


sphuber:
@giovannipizzi this is a pretty important bug, but it is almost impossible to unittest. I will see if I can deploy it for production environment on theossrv5 and see if the problem is resolved. The problem here is that were trying to prove a problem _doesn't_ show up, which is always difficult

giovannipizzi:
Great catch!
One (probably naive) question: is it enough to set it to None and wait for the next call to re-update, or should we do something like it's done outside, `self._update_handle = self._loop.call_later(self._get_next_update_delay(), updating)`?

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -113,7 +114,7 @@ def _get_jobs_from_scheduler(self):
             # Update the last update time and clear the jobs cache
             self._last_updated = time.time()
             jobs_cache = {}
-            self.logger.info('AuthInfo<{}>: successfully retrieved status of active jobs'.format(self._authinfo.pk))
+            self.logger.info('AuthInfo<%s>: successfully retrieved status of active jobs', self._authinfo.pk)
```
giovannipizzi: Is there a reason to switch to the `%s` string interpolation instead of `{}` and `.format`?

sphuber: This makes the evaluation of the arguments to the `info` call lazily evaluated. This is useful when those operations are potentially costly and there is a chance that the log will be discarded anyway. Using the lazy evaluation is the [preferred method](https://docs.python.org/3/howto/logging.html#optimization) for logging for this reason, even though in this case the effect will be minimal

sphuber:
> Great catch!
> One (probably naive) question: is it enough to set it to None and wait for the next call to re-update, or should we do something like it's done outside, `self._update_handle = self._loop.call_later(self._get_next_update_delay(), updating)`?

I think not, but I am not a 100% sure. The reason I set it to `None` is because at that point, there are no active requests, since we just cancelled all the futures. It then doesn't make sense to schedule the next update, when as soon as the next request comes in, this is done anyway in `_ensure_updating`. So I think this is the correct thing to do. 

Still I wouldn't merge this just yet. Since we have no tests, we rely on testing manually in production which is what I am doing now. I also applied the path for Thibault who was experiencing the same problems and he hasn't had the problem since. Daemon ran all night without issues. Same for me on theossrv5, except that this morning I noticed that all daemon runners were running at 100% even though there were 0 active processes and 0 tasks. So I am not sure what was going on, but it worries me a bit. I will keep running for a few more days to watch the behavior

giovannipizzi:
Ok thanks for the explanation! I agree to wait a bit more, and that it's hard to test properly - one should emulate a failure when multiple things are in the queue etc. - it would be ideal if in the mid term we find ways to test this kind of things, but I have no good idea right now

aiida/engine/processes/calcjobs/manager.py
giovannipizzi: I'm not sure that's the point. Anyway python needs to evaluate the argument before passing it to the `.info()` method, and the `.info()` method cannot do anything to defer it. The only way to do it would be if the method were to accept a callable and then decides if/when to call it, but this cannot be done with properties like `pk` (unless you return a lambda returning the value, but anyway I'm not sure the logger accepts lambdas or callables). 
I think the point of the link is not on whether to use `%s` vs. `format`, but rather to wrap the logging call into an `if` block `checking `logger.isEnabledFor(CURRENT_LEVEL)`.
For this reason (since anyway here we don't need to speed this up, if `pk` is slow we've got a problem anyway :-) ) I would suggest consistency in the code and using `format` as we do (almost?) everywhere else.

giovannipizzi: Example proving the lambda concept:
```python
import time

def slow():
    time.sleep(2)
    return "slow return"

def myprint(value, shouldprint=True):
    print('inside myprint, before printing')

    # Potential checks if I need to print here
    if not shouldprint:
        return

    # Lazy evaluation
    if callable(value):
        value = value()

    print("Value: {}".format(value))
    print('inside myprint, after printing')


print("*"*72)

print("*** Fast computations")
# Everything fast
myprint("A")

print("*"*72)

print("*** Slow evaluation *before* calling myprint")
# Slow evaluation, then three messages printed quickly
myprint(slow())

print("*"*72)

print("*** Slow evaluation deferred inside myprint")
# First print occurs quickly, then value is slowly evaluated inside the method,
# then final two messages are printed
myprint(lambda: slow())

print("*"*72)

print("*** Should be slow")
t = time.time()
# Does not print but it's slow
myprint(slow(), shouldprint=False)
print("Time spent: {} s".format(time.time() - t))

print("*"*72)

print("*** Should be fast")
t = time.time()
# Does not print and returns quickly
myprint(lambda: slow(), shouldprint=False)
print("Time spent: {} s".format(time.time() - t))

print("*"*72)
```

sphuber: Agreed, I misread the link I gave. I'm happy with reverting the change, but then we probably want to add a global ignore rule for the warning that pylint will start to throw. They advise to pass string formatting arguments as arguments to the logging call to delay interpolation to last moment possible. Do you want me to revert?

giovannipizzi: I think this exception makes sense only if you are doing this inside a loop that is very fast (I would say microseconds) where interpolating the string is slow. For us it does not matter so much I believe. There are also various discussions online, also on how from py3.6 there are even faster ways to do this with f-strings, see e.g. [here](https://github.com/PyCQA/pylint/issues/2395).

My suggestion would be to revert and add an ignore for this (does it even make sense to put a global ignore in the pylintrc for this error, in AiiDA?)

sphuber: I would say that if we are not going to use it, we should just put the ignore in the `.pylintrc` instead of littering the code with inline ignores

PR #3164, opened 2019-07-15
Improve the efficiency of link validation

### LINKED ISSUE #3088: Improve the efficiency of link validation
The link validation in `Node.add_incoming` was written to be correct when the `Node` interface was rewritten and efficiency was not given priority. However, in certain use cases the validation becomes too slow and forms a bottleneck. The problem arises when there are so-called "god" nodes in the database. These are typically input nodes that are used over-and-over again for certain calculation. This is either by design, e.g. `UpfData` nodes, or in an attempt by the user to overcome another inefficiency of AiiDA that duplicate node content is not automatically deduplicated. Meaning that if a node is stored with the exact same content of a node that is already in the database, the content is duplicated.

The problem of "god"-nodes is that if a new link is added from or to one, the link validation will need to consider all other links that it already has. This operation becomes progressively more heavy as more links are added.

If possible, the implementation of `validate_incoming` and `validate_outgoing` on the ORM level should be made more efficient. Potentially, other solutions that do not rely on the ORM, but for example perform the checks on the database level (which are potentially a lot faster) may have to be considered. See issue #2238 for example.

zhubonan:
Fixes #3088 

This PR is aimed to improve the performance for input link validation #3088. The original approach in my comment in #3088 does not work since the `validate_link` the function has to deal with cached links.

The root of the problem is that `get_stored_link_triples` method returns a list of `LinkTriple` which has the node instance as its `node` attribute. When doing validations loading individual `Node` is not necessary. When link triples are requested for a node with many links, getting every connected nodes can be very slow and sometimes become a performance bottleneck. This PR adds a `only_uuid` switch to the function to include only uuids in the `LinkTriple` constructed instead of the full `Node`. When it is set to True, the `node` attribute of `LinkTriple` returned becomes a string of the uuid. The keyword is also added for `get_incoming` and `get_outgoing` methods. The `validate_link` is updated accordingly to utilize this for faster checks.

The test showed that the efficiencies are improved significantly when storing nodes with inputs links from *god-like* node which has many output links. The time used for adding a single node is plotted with the total number of nodes added below. The sqla backend is used in the test.

performance

The time taken to store one node increases proportionally with the number of stored nodes before and after the commit, which means that the total time to store N nodes will scale with N^2. However, the slope is much smaller (by ~ 1/40) with the fix.

The code for generating the timings: 

```
from random import choice
from tqdm import tqdm
from time import time
comp = Computer.objects.all()[0]
node_data = Data()
node_data.store()



def create_calc_node():
    node = CalcJobNode()
    for i in range(10):
        k = choice('abcds')
        node.set_attribute(k, k)
    node.computer = comp
    node.set_option('resources', {'num_machines': 1, 'tot_num_mpiprocs': 12})
    return node

start = time()
timings = []
for i in tqdm(range(3000), total=3000):
    node = create_calc_node()
    node.add_incoming(node_data, link_type=LinkType.INPUT_CALC, link_label='INPUT')
    node.store()
    timings.append(time() - start)


with open('save_timings', 'w') as fh:
    for t in timings:
        fh.write('{:.3f}\n'.format(t))

```

### DIFF
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -829,7 +829,12 @@ class Node(Entity):
 
         self._incoming_cache.append(link_triple)
 
-    def get_stored_link_triples(self, node_class=None, link_type=(), link_label_filter=None, link_direction='incoming'):
+    def get_stored_link_triples(self,
+                                node_class=None,
+                                link_type=(),
+                                link_label_filter=None,
+                                link_direction='incoming',
+                                only_uuid=False):
         """Return the list of stored link triples directly incoming to or outgoing of this node.
 
         Note this will only return link triples that are stored in the database. Anything in the cache is ignored.
@@ -839,6 +844,7 @@ class Node(Entity):
         :param link_label_filter: filters the incoming nodes by its link label. This should be a regex statement as
             one would pass directly to a QuerBuilder filter statement with the 'like' operation.
         :param link_direction: `incoming` or `outgoing` to get the incoming or outgoing links, respectively.
+        :param only_uuid: Store the uuid of nodes in the NodeTriple instead of the actual node
         """
         if not isinstance(link_type, tuple):
             link_type = (link_type,)
@@ -859,24 +865,25 @@ class Node(Entity):
         builder = QueryBuilder()
         builder.append(Node, filters=node_filters, tag='main')
 
+        node_project = ['uuid'] if only_uuid else ['*']
         if link_direction == 'outgoing':
             builder.append(
                 node_class,
                 with_incoming='main',
-                project=['*'],
+                project=node_project,
                 edge_project=['type', 'label'],
                 edge_filters=edge_filters)
         else:
             builder.append(
                 node_class,
                 with_outgoing='main',
-                project=['*'],
+                project=node_project,
                 edge_project=['type', 'label'],
                 edge_filters=edge_filters)
 
         return [LinkTriple(entry[0], LinkType(entry[1]), entry[2]) for entry in builder.all()]
 
-    def get_incoming(self, node_class=None, link_type=(), link_label_filter=None):
+    def get_incoming(self, node_class=None, link_type=(), link_label_filter=None, only_uuid=False):
         """Return a list of link triples that are (directly) incoming into this node.
 
         :param node_class: If specified, should be a class or tuple of classes, and it filters only
@@ -885,18 +892,22 @@ class Node(Entity):
             link type, if None then returns all inputs of all link types.
         :param link_label_filter: filters the incoming nodes by its link label.
             Here wildcards (% and _) can be passed in link label filter as we are using "like" in QB.
+        :param only_uuid: Use the uuid of the nodes instead of the `Node` objects
         """
         if not isinstance(link_type, tuple):
             link_type = (link_type,)
 
         if self.is_stored:
-            link_triples = self.get_stored_link_triples(node_class, link_type, link_label_filter, 'incoming')
+            link_triples = self.get_stored_link_triples(
+                node_class, link_type, link_label_filter, 'incoming', only_uuid=only_uuid)
         else:
             link_triples = []
 
         # Get all cached link triples
         for link_triple in self._incoming_cache:
 
+            if only_uuid:
+                link_triple = LinkTriple(link_triple.node.uuid, link_triple.link_type, link_triple.link_label)
             if link_triple in link_triples:
                 raise exceptions.InternalError('Node<{}> has both a stored and cached link triple {}'.format(
                     self.pk, link_triple))
@@ -910,7 +921,7 @@ class Node(Entity):
 
         return LinkManager(link_triples)
 
-    def get_outgoing(self, node_class=None, link_type=(), link_label_filter=None):
+    def get_outgoing(self, node_class=None, link_type=(), link_label_filter=None, only_uuid=False):
         """Return a list of link triples that are (directly) outgoing of this node.
 
         :param node_class: If specified, should be a class or tuple of classes, and it filters only
@@ -920,7 +931,7 @@ class Node(Entity):
         :param link_label_filter: filters the outgoing nodes by its link label.
             Here wildcards (% and _) can be passed in link label filter as we are using "like" in QB.
         """
-        link_triples = self.get_stored_link_triples(node_class, link_type, link_label_filter, 'outgoing')
+        link_triples = self.get_stored_link_triples(node_class, link_type, link_label_filter, 'outgoing', only_uuid)
         return LinkManager(link_triples)
 
     def has_cached_links(self):
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -844,7 +844,7 @@ class Node(Entity):
         :param link_label_filter: filters the incoming nodes by its link label. This should be a regex statement as
             one would pass directly to a QuerBuilder filter statement with the 'like' operation.
         :param link_direction: `incoming` or `outgoing` to get the incoming or outgoing links, respectively.
-        :param only_uuid: Store the uuid of nodes in the NodeTriple instead of the actual node
+        :param only_uuid: project only the node UUID instead of the instance onto the `NodeTriple.node` entries
         """
         if not isinstance(link_type, tuple):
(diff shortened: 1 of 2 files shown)


zhubonan:
As an interesting note, there seem to be periodic slowdowns:

performance_zoom

giovannipizzi:
Thanks a lot @zhubonan !
For the periodic jumps, I think it's ok, it might just be that the DB is writing things in a memory cache and when you go above some threshold is has to dump to disk.

zhubonan:
Thanks for the comment. I think there is still room of improvements. So please don't merge yet. The scaling is linear since all links triples has to be returned for checking uniqueness. The latter can be coded in the querybuilder directly. And a separate check can be done with the cache. I will update the code soon.

zhubonan:
I have improved the code so the filtering of unique source-link-target is done at the query level whenever possible, instead of the at the python level. Only the cached links are compared at the python level. The *god-like* nodes now do not have any noticeable performance impact on storing speed anymore. 

performance

sphuber:
I have checked out your changes and performed my CIF cleaning high-throughput benchmark to see how it performs. This test I already used before to test the new JSONB implementation for node attributes and extras for Django. It is based on a real life production project I run that really suffered from the link validation so I always had to run it with it completely disabled. For details see the issue on the Django JSONB efficiency tests #3067 

Essentially, I submit 500 `CifCleanWorkChains` each of which launches two `CalcJobs` and a `calcfunction`. I run this with 20 daemon runners on our production server with all calculation jobs over local transport and direct scheduler.


| Setup                      | Total (submission) [s]    |
|--------------------------|-------------------------------|
| Current link validation  |  23:10 (23:00)  |
| Without link validation                 |  02:50 (01:52)  |
| Changes by @zhubonan           |  04:00 (03:41)  |

As you can see, the current link validation slows down the submission by an order of magnitude compared to when it is completely turned off. With your fix, the submission is only a factor of two slower than the base line. I think this is already an amazing improvement! I will now go over the code changes to see if they can be merged or maybe there is even some more to be gained.

A huge thank you @zhubonan , excellent work! :+1:

giovannipizzi:
thanks, amazing work!

zhubonan:
@sphuber Thanks for testing. I just pushed some minor changes. Please reiview the code and merge if you think it is fine.

zhubonan:
@sphuber Thanks for simplifying the code!

aiida/orm/nodes/node.py
```diff
@@ -901,6 +910,9 @@ def get_incoming(self, node_class=None, link_type=(), link_label_filter=None):
                 raise exceptions.InternalError('Node<{}> has both a stored and cached link triple {}'.format(
                     self.pk, link_triple))
 
+            if only_uuid:
```
giovannipizzi: Shouldn't this go above? the `link_triples` from the DB already have only UUIDs

aiida/orm/utils/links.py
```diff
@@ -122,37 +122,59 @@ def validate_link(source, target, link_type, link_label):
     if not isinstance(source, type_source) or not isinstance(target, type_target):
         raise ValueError('cannot add a {} link from {} to {}'.format(link_type, type(source), type(target)))
 
-    # Validate the outdegree of the source node
-    outgoing = source.get_outgoing(link_type=link_type)
+    def link_triple_exists(source, target, link_type, link_label):
```
giovannipizzi: This will redefine this function at every execution of the enclosing function - wouldn't it be better to move the definition outside the function?

aiida/orm/utils/links.py
```diff
@@ -122,37 +122,59 @@ def validate_link(source, target, link_type, link_label):
...
+        :param link_label: the link label
+        :return: boolean, True if the link triple exists, False otherwise
+        """
+        from aiida.orm import QueryBuilder
+
+        if target._incoming_cache and LinkTriple(source, link_type, link_label) in target._incoming_cache:  # pylint: disable=protected-access
+            return True
+
+        builder = QueryBuilder()
+        builder.append(Node, filters={'uuid': source.uuid}, project=['id'])
```
giovannipizzi: We can also use `id`s here, right? Queries on the ID are typically much faster (same for the target node one line below)

aiida/orm/utils/links.py
```diff
@@ -122,37 +122,59 @@ def validate_link(source, target, link_type, link_label):
...
+        builder.append(Node, filters={'uuid': target.uuid}, edge_filters={'type': link_type.value, 'label': link_label})
+        builder.limit(1)
+
+        return builder.count() != 0
+
+    if outdegree == 'unique_triple' or indegree == 'unique_triple':
```
giovannipizzi: Let's set a default value (e.g. to False) to avoid `NameError`s? Or better to leave it like this so we notice if we change the code and forget to define the variable?

aiida/orm/nodes/node.py
sphuber: correct, will change it

aiida/orm/utils/links.py
sphuber: Thought it didn't, but turns out the function object is recreated at every invocation even though the code block is compiled once. But still, given that this is called often I will move it out

aiida/orm/utils/links.py
sphuber: We can. If you say it is faster I will change it

aiida/orm/utils/links.py
sphuber: I'd rather raise indeed then the conditional below accidentally passing

sphuber:
@zhubonan I have rebased the PR but I cannot force push to your repository. Would you be willing to give me collaboration rights? Or otherwise I can open a new PR

sphuber:
@giovannipizzi I addressed your comments and rebased/cleaned the commits. Should be good to go

aiida/orm/utils/links.py
zhubonan: In addition, the query part can be skipped if any of the nodes do not have a pk (if it is not stored).

aiida/orm/nodes/node.py
```diff
@@ -885,18 +892,22 @@ def get_incoming(self, node_class=None, link_type=(), link_label_filter=None):
...
-            link_triples = self.get_stored_link_triples(node_class, link_type, link_label_filter, 'incoming')
+            link_triples = self.get_stored_link_triples(
+                node_class, link_type, link_label_filter, 'incoming', only_uuid=only_uuid)
         else:
             link_triples = []
 
         # Get all cached link triples
         for link_triple in self._incoming_cache:
 
+            if only_uuid:
```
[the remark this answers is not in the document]
zhubonan: Don't quite understand. Has this been resolved?

aiida/orm/utils/links.py
sphuber: actually this is a good point. I think this might now except if one of the nodes is not stored since it won't have a pk. I guess we can revert to using UUIDs or just add `if source.pk is None or target.pk is None: return False`

zhubonan: I would second the latter as it skips the need of querying the database. One of the major usage of this function is to validate the links of an unstored node just before having it stored.

aiida/orm/utils/links.py
```diff
@@ -122,37 +122,80 @@ def validate_link(source, target, link_type, link_label):
...
+        query = QueryBuilder()
+        query.append(Node, filters={'uuid': source.uuid})
+        query.append(
+            Node,
+            filters={'uuid': target.uuid},
+            edge_filters={
+                'type': link_type.value,
+                'label': link_label
+            },
+            project=['uuid'],
```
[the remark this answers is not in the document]
sphuber: After some testing does not seem to have much of an influence

aiida/orm/nodes/node.py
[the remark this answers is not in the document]
sphuber: Yes it has been resolved. My original comment is wrong, because in the case of `only_uuid` the link triples from the stored nodes will already only have the uuid, so the conversion of the cached triples needs to go before the check. Which is what you already had.

aiida/orm/utils/links.py
sphuber: I have updated the function to include this check to potentially skip the query if either of the two nodes is not stored.

PR #3187, opened 2019-07-17
Fix updating existing Node Extras and clean code for import of Node extras

### LINKED ISSUE #3186: Updating existing Node Extras fails
The import functions has a general problem reusing `unique_id` from a previous for-loop during import of Node attributes and extras.
Except for two places, it is only used for error messages, where it will not give any problems (only for tracing back the problem).
However, when updating existing Node extras it is used, as if it was the iterator for the Node UUIDs. This will fail, since where the iterator was used, was in a previous for-loop.

This whole segment should be rewritten/updated to reflect its proper place in the code. Specifically, the use of invalid iterators should be remedied.

CasperWA:
Fixes #3186

Clean up a bit the code for importing Node Extras.
There were several iterators used, which were not "legal".

At its worst, Extras were unable to be updated for existing Nodes.

Created a new exception `ArchiveIntegrityError`, which may be raised when a key is missing for an entity PK or similar in the `data.json` file.
I have yet to implement it completely in the import/export code.

The timing should be revisited for the import in this section - when to update the Attributes/Extras, when to create new, etc.

### DIFF
--- a/aiida/common/exceptions.py
+++ b/aiida/common/exceptions.py
@@ -243,3 +243,11 @@ class CircusCallError(AiidaException):
     """
     Raised when an attempt to contact Circus returns an error in the response
     """
+
+
+class ArchiveIntegrityError(IntegrityError):
+    """
+    Raised when there is an integrity error in an export archive.
+    This may be if a Node's repository folder is missing or a Node is missing a representation in a dictionary of
+    `data.json`.
+    """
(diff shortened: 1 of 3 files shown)


sphuber:
I looked through the changes and was unable to really spot changes in logic, just renaming things. Is that correct? If stuff was failing first and this PR is supposed to fix it, could we add a test to probe this?

CasperWA:
Yeah, I tried to change the logic, but was unable to (quickly). Instead, the logic change is at this point a wish. It ended up being a simple fix instead.

I don't know that I can replicate the exact case for this error - it should already be tested in the Extras tests, however, it may be that only a single Node is exported, which would result in no problems for the wrong code. Indeed, even if more Nodes are exported for a test, the test will still not work 1/N times (where N is the number of exported Nodes).

sphuber:
Alright, I have gotten @CasperWA that this fixes the problem, so I will approve and merge it. I hereby denounce all responsibility :smile:

PR #3188, opened 2019-07-18
Make the `execmanager.submit_calculation` idempotent'ish

### LINKED ISSUE #3183: Make the `execmanager.submit_calculation` idempotent as best as possible
Currently it is possible to have the same job submitted twice if the daemon is interrupted and restart during the `submit_calculation` call.

sphuber:
Fixes #3183 

The `submit_calculation` is responsible for submitting a job, that was
already uploaded to the remote machine, to the scheduler. There was a
risk of the daemon submitting the same job twice. If the first time
around it successfully submitted the job and set the job id as an
attribute on the node, but then was interrupted (for example due to a
shutdown) before it could transition the `CalcJob` to the `UPDATE`
transport task, once reloaded the command would be executed again. This
would not complain and simply submit the job again to the scheduler.
This is both a waste of resources and can cause various complications.

It is impossible to make the function fully idempotent, but by checking
if the job id attribute is already set for the node at the beginning and
returning it should minimize the risk of double submission.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -236,8 +236,7 @@ def upload_calculation(node, transport, calc_info, script_filename, dry_run=Fals
 
 
 def submit_calculation(calculation, transport, calc_info, script_filename):
-    """
-    Submit a calculation
+    """Submit a previously uploaded `CalcJob` to the scheduler.
 
     :param calculation: the instance of CalcJobNode to submit.
     :param transport: an already opened transport to use to submit the calculation.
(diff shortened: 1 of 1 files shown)


giovannipizzi:
see e.g. slurm commands here https://ubccr.freshdesk.com/support/solutions/articles/5000686909-how-to-retrieve-job-history-and-accounting

sphuber:
Should be easy enough. To get a unique list of all job ids for given user since given data:
```
sacct -S 2019-06-18 -u sphuber --format=JobId -n | cut -d . -f1 | sort | uniq -u
```
this can then be compared with the job ids stored in the attributes.

PR #3190, opened 2019-07-18
Fix bug in tree printer of `verdi node tree`

### LINKED ISSUE #223: verdi node tree should show link types as edge labels
Originally reported by: **Martin Uhrin (Bitbucket: [muhrin](https://bitbucket.org/muhrin), GitHub: [muhrin](https://github.com/muhrin))**

----------------------------------------

In addition links between the same parent -- child should be combined e.g. return+create

We probably need to use short hands e.g. r = return, c = create, ca = call etc.

----------------------------------------
- Bitbucket: https://bitbucket.org/aiida_team/aiida_core/issue/223

sphuber:
Fixes #223 

The method illegally passed `None` for `link_type` to the `get_outgoing`
method. Now it will pass all existing link types. The sorting lambda
also did not correctly dereference the `ctime` from the node instance.


sphuber:
Note that technically this PR does not solve #223 but I think we should close it regardless. This PR makes sure that at least in the current state the command works.

PR #3192, opened 2019-07-19
allow wildcard to be used while filtering process label in cmdline

zhubonan:
This PR changes the filter for matching `process_label` in `verdi process list` from the default
'==' (implicit exact match) to 'like', which allows wildcards to be used. For example, processes of  `PwBaseWorkChain`, `PwRelaxWorkChain` and be filtered easily with `'-L Pw%Chain'`.

### DIFF
--- a/aiida/cmdline/params/options/__init__.py
+++ b/aiida/cmdline/params/options/__init__.py
@@ -260,7 +260,7 @@ PROCESS_STATE = OverridableOption(
 PROCESS_LABEL = OverridableOption(
     '-L', '--process-label', 'process_label',
     type=click.STRING, required=False,
-    help='Only include entries with this process label.')
+    help='Only include entries match this process label.')
 
 EXIT_STATUS = OverridableOption(
     '-E', '--exit-status', 'exit_status',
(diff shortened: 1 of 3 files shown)


aiida/cmdline/params/options/__init__.py
```diff
@@ -260,7 +260,7 @@ def active_process_states():
 PROCESS_LABEL = OverridableOption(
     '-L', '--process-label', 'process_label',
     type=click.STRING, required=False,
-    help='Only include entries with this process label.')
+    help='Only include entries match this process label.')
```
sphuber: Change to 'Only include entries whose process label matches this filter.'

aiida/cmdline/utils/query/calculation.py
```diff
@@ -80,7 +80,7 @@ def get_filters(self,
             filters[process_state_attribute] = {'in': process_state}
 
         if process_label is not None:
-            filters[process_label_attribute] = process_label
+            filters[process_label_attribute] = {'like': process_label}
```
sphuber: I think the `like` operation is significantly slower than equals. Let's check if the `process_label` contains the wildcard operator in which case we use `like` otherwise use equality operator

zhubonan: I see your point. For ~1000 entries in a database of 15K nodes, I did not see much difference. I can add a check to only use `like` when wildcard characters are present in the passed option.

zhubonan:
@sphuber I have made the changes and also updated the test for `verdi process list`.

PR #3200, opened 2019-07-23
Docs: update the section on working with schedulers

### LINKED ISSUE #3024: Check documentation for "working with AiiDA/schedulers"

https://aiida.readthedocs.io/projects/aiida-core/en/latest/working_with_aiida/index.html#schedulers

sphuber:
Fixes #3024 

 * Fix the hyperlinks to supported schedulers
 * Remove usage of outdated `CalcJobNode.set_option` with metadata options
 * `JobResource` now validates upon construction of the class instance
   so it is no longer possible to set as properties after construction

### DIFF
--- a/docs/source/scheduler/index.rst
+++ b/docs/source/scheduler/index.rst
@@ -24,7 +24,7 @@ All the main features are supported with this scheduler.
 
 The :ref:`JobResource <job_resources>` class to be used when setting the job resources is the :ref:`NodeNumberJobResource`.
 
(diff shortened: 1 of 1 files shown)


sphuber:
I didn't change the structure much, even though I doubt that users would ever need access to the `JobResource` classes and its sub classes, since all of this is set automatically by the engine based on the `metadata.options` passed in the process inputs. Should we leave it like this?

sphuber:
I will merge this for now. Once we do the reorganization of everything we can see if we move part of this.

PR #3202, opened 2019-07-24
Add tests for `dry_run=True` and `store_provenance=False`

### LINKED ISSUE #3194: Respect return signature of launchers for `CalcJob` in dry-run
The metadata of a `CalcJob` allow to launch a dry-run, however, in this case, the return value is different compared to a normal launch. For example, `run_get_node` returns a tuple of the result and process node, but this is not the case for a dry-run

sphuber:
Fixes #3194 

These tests ensure that the various launchers respect the return value
signature irrespective of the values of the `dry_run` and
`store_provenance` metadata inputs.

### DIFF
--- a/aiida/backends/tests/engine/test_launch.py
+++ b/aiida/backends/tests/engine/test_launch.py
@@ -136,7 +136,7 @@ class TestLaunchersDryRun(AiidaTestCase):
 
         code = orm.Code(
             input_plugin_name='arithmetic.add',
-            remote_computer_exec=[self.computer, '/bin/true'])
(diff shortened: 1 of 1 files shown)


sphuber:
@giovannipizzi I opened #3194 after you told me that you encountered this problem. Either I have misunderstood it or you misidentified the problem because the described problem does not exist. I already had added tests to validate the return value of the launchers under these conditions. I added another one for the case of `store_provenance=False` for good measure. Can you maybe post the exact problem you were experiencing?

Issue #3201, opened 2019-07-23
`load_profile` is not enough to load the profile


giovannipizzi:
At least for Django backends. I think it's a bug.

Snippet to reproduce (make sure to have a default Django backend, or specify a Django profile with as a parameter to `load_profile`):
```python
from aiida import load_profile
load_profile()
from aiida.manage.backup.backup_django import Backup
```
This fails with 
```
ImproperlyConfigured: Requested setting DEFAULT_INDEX_TABLESPACE, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
```
while it works from the `verdi shell`.

I think the old `load_dbenv`, when called, was enough to make sure everything was loaded. Is this not anymore the case with `load_profile`? What does `verdi` does in addition to actually load fully the profile?

By the way, this seems to work instead:
```python
from aiida import load_profile
load_profile()
from aiida.manage.manager import get_manager
manager = get_manager()
manager.get_backend()
from aiida.manage.backup.backup_django import Backup
``` 
but I'm not sure if this the right thing to do (I have the feeling that the `get_backend()` should be put inside `load_profile`).

sphuber:
The problem is that the concept of the "profile" and "backend" are now properly separated. That is to say, when the profile is loaded, does not mean that the backend itself is also loaded. This is necessary to make the backend load lazily and to be able to "unload" it and switch to another one. The backend is loaded automatically whenever any component that requires it is constructed. The problem here is that the `Backup` class is not properly going through this layer. Like other ORM components it should not be imported from a backend specific path, but the manager should be asked for an instance.

PR #3205, opened 2019-07-24
Update "working with AiiDA/backups"

### LINKED ISSUE #3028: Check documentation for "working with AiiDA/backups"

https://aiida.readthedocs.io/projects/aiida-core/en/latest/working_with_aiida/index.html#backups

CasperWA:
Fixes #3028 

Ran through most of the documentation on setting up backups for the database and the repository and updated the documentation accordingly.

*Setup*:
Ubuntu 18.04.2 LTS
AiiDA v1.0.0b5
PostgreSQL 9.6

### DIFF
--- a/aiida/manage/backup/backup_setup.py
+++ b/aiida/manage/backup/backup_setup.py
@@ -239,7 +239,7 @@ load_profile(profile='{}')
 {}
 
 # Create the backup instance
-backup_inst = Backup(backup_info_filepath="{}", additional_back_time_mins = 2)
+backup_inst = Backup(backup_info_filepath="{}", additional_back_time_mins=2)
 
 # Define the backup logging level
 backup_inst._logger.setLevel(logging.INFO)
--- a/aiida/manage/backup/backup_setup.py
+++ b/aiida/manage/backup/backup_setup.py
@@ -239,7 +239,7 @@ load_profile(profile='{}')
 {}
 
 # Create the backup instance
-backup_inst = Backup(backup_info_filepath="{}", additional_back_time_mins=2)
+backup_inst = Backup(backup_info_filepath="{}", additional_back_time_mins = 2)
 
 # Define the backup logging level
 backup_inst._logger.setLevel(logging.INFO)
--- a/docs/source/backup/backup-aiidadb-USERNAME
+++ b/docs/source/backup/backup-aiidadb-USERNAME
@@ -1,3 +1,2 @@
 #!/bin/bash
 su USERNAME -c "/home/USERNAME/.aiida/backup_postgresql.sh"
-
--- a/docs/source/backup/index.rst
+++ b/docs/source/backup/index.rst
@@ -1,7 +1,6 @@
 .. _backup:
 
-In this page you will find useful information on how to backup your database,
-how to move it to a different location and how to backup your repository.
+On this page you will find useful information on how to backup your database, how to move it to a different location, and how to backup your repository.
 
 
 .. _backup_postgresql:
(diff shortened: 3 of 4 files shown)


ltalirz:
I'll take a look at this now

giovannipizzi:
Did you check also if you reproduce #3201 and if this can be fixed?

CasperWA:
> Did you check also if you reproduce #3201 and if this can be fixed?

I can still reproduce this.
I can try and fix it, i.e. attempt to use @sphuber 's insight from his comment in the issue.

CasperWA:
All right, so I have done a "minimum effort" fix - in the sense that I have created a backend-agnostic `backup_general.py` file that uses the ORM Node class and QueryBuilder to query the data.
While this may be a bit slower than using the backend-specific query API, this is much cleaner.

There are still a lot of optimizations that can be done in this area, e.g. merging the Backup and AbstractBackup classes, since there is now no reason to split them up (as I see it?).
However, as it is, it works fine - the test times are similar (if just a bit more steady around ~0.35s for the backend-agnostic implementation) when going through the two backup test-files, if one is using the backend-specific or -agnostic Backup classes.

Finally, I have squashed down to three commits:

1. Concerns the backup docs.
1. Is an addition of a global "pylint: disable=useless-object-inheritance", since we are keeping with Py2 and Py3 support for the time being, there is no reason to keep adding this disable comment in the code every time a base class is created. As a consequence I have removed it throughout the code and kept it in `.pylintrc`.
1. The re-imagining of Backups - now backend-agnostic.

ltalirz:
Since togo added a fix as well (not sure if more is needed), do you agree it's nicer to merge this first?
https://github.com/aiidateam/aiida-core/pull/3210

sphuber:
I don't think merging #3210 makes much sense, since it does not solve the problem. The import will still except because the backend is not properly loaded.

giovannipizzi:
@sphuber note that the change in #3210 is on a different file than the one discussed in #3201 so it might make sense to merge it first

sphuber:
I know, but @CasperWA also fixes the file in this PR so I don't see why we shouldn't just include it here. Moreover, @CasperWA said it worked in his testing and he did not make the same change. So I don't fully understand what the problem was. Is it still not working and/or tested?

CasperWA:
> I don't think merging #3210 makes much sense, since it does not solve the problem. The import will still except because the backend is not properly loaded.

Yeah, I don't think #3210 will fix #3201. But it's difficult to know if it is that same error that is being fixed with #3210. Indeed, I don't think I even see why it would fail with the relative import? Other than it's hitting the same error as for #3201.

sphuber:
Exactly, the relative import should just work and it does for me
```
from aiida.manage.backup import backup_setup
```
without #3210 works just fine, so I am not sure what issue @atztogo originally ran in, but it makes sense to investigate it a bit further before merging

giovannipizzi:
He mentioned he had problems with py 3.7, I don't know if this is the problem

CasperWA:
> He mentioned he had problems with py 3.7, I don't know if this is the problem

I'm sure we'll find out in good time - when the Earth spins a bit further :)

sphuber:
@CasperWA the fix by @atztogo changing the relative import to an absolute one *was* important, so I am now wondering what you tested exactly? If you ran the command that is suggested in the documentation, i.e. `verdi run aiida/.../backup_script.py` then it *should* have failed.

CasperWA:
> @CasperWA the fix by @atztogo changing the relative import to an absolute one _was_ important, so I am now wondering what you tested exactly? If you ran the command that is suggested in the documentation, i.e. `verdi run aiida/.../backup_script.py` then it _should_ have failed.

Ah. Well I tested everything explicitly up to "Restore databases backup". This would explain it, because then I would have also not run that command (which is under the "Setup repository backup" section). Sorry 😞 
So the fix by @atztogo was not for #3201, but for this other thing. Great to catch it! And definitely important.

CasperWA:
> @CasperWA I've read through the docs and I would be happy to approve the changes (I have a couple of suggestions for changing the structure of the backup section but I will do this in a separate PR).

Great! On both accounts.

> With the changes in the code, however, we'll need someone with a big aiida 1.0 database (which I don't have) to test the speed of the new script. After all, the whole point of this script is to speed up backups.
> ...
> Otherwise I'd like to ask someone who has a large 1.0 database to benchmark this.

Again, just to be clear. The script is the same as the previous backend-specific scripts. With the only difference that all queries are now going through the QueryBuilder. So to test that this can be done through the QueryBuilder and not with backend-specific queries, I think you are right in having it properly benchmarked.

> ...
> Do you think it makes sense to split this into 2 PRs, one for the text and one for the code?

Yes, I do think that may be the correct approach. Also a more elegant one, since this should really only have involved updating and checking the docs.

Even more so, maybe I should split it in three PRs (one also for the pylint stuff)?

CasperWA:
The three original commits in this PR have now been split into separate PRs (#3240, #3241, and this one).

This PR is now _only_ concerned with the Docs update of the Backup section.

aiida/manage/backup/backup_setup.py
```diff
@@ -239,7 +239,7 @@ def run(self):
 {}
 
 # Create the backup instance
-backup_inst = Backup(backup_info_filepath="{}", additional_back_time_mins = 2)
+backup_inst = Backup(backup_info_filepath="{}", additional_back_time_mins=2)
```
CasperWA: This should probably be moved to #3240

PR #3206, opened 2019-07-24
Docs: updating "Modifying the schema" for `aiida-core==1.0.0`

### LINKED ISSUE #3176: Check documentation for "for AiiDA developers/Modifying the schema"
Check if the section is updated to AiiDA 1.0.

https://aiida.readthedocs.io/projects/aiida-core/en/latest/developer_guide/core/modifying_the_schema.html

sphuber:
Fixes #3176

### DIFF
--- a/docs/source/developer_guide/core/modifying_the_schema.rst
+++ b/docs/source/developer_guide/core/modifying_the_schema.rst
@@ -3,220 +3,193 @@ Modifying the schema
 
 Django
 ------
-The Django database schema can be found in :py:mod:`aiida.backends.djsite.db.models`.
 
+The Django database schema can be found in :py:mod:`aiida.backends.djsite.db.models`.
 If you need to change the database schema follow these steps:
 
-1. Make all the necessary changes to :py:mod:`aiida.backends.djsite.db.models`
-2. Create a new migration file.  From ``aiida/backends/djsite``, run::
+ 1. Make all the necessary changes to :py:mod:`aiida.backends.djsite.db.models`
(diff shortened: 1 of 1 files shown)


docs/source/developer_guide/core/modifying_the_schema.rst
```diff
@@ -3,220 +3,192 @@ Modifying the schema
...
+            statement = text('\n'.join(forward_sql))
+            conn.execute(statement)
+        def downgrade():
+            conn = op.get_bind()
+            statement = text('\n'.join(reverse_sql))
+            conn.execute(statement)
+
+    If you want to learn more about the migration operations, you can have a look at the Alembic documentation.
+
+ 4. Your database will be automatically migrated to the latest revision as soon as you run your first verdi command.
```
giovannipizzi: Is this still true?

sphuber: Nope, you are right. It now follows the same paradigm as Django and simply prompts a command to be executed

sphuber: I also realize now that the Django section still instructs to use the Django manage tool directly, i.e. `python aiida/backends/djsite/manage.py migrate`. I think people, even developers, should use `verdi database migrate` instead. Should I homogenize the two sections to reflect this?

giovannipizzi: Sure, if you can it would be great!

PR #3220, opened 2019-07-26
Update requirements for for `dev_precommit` dependencies

### LINKED ISSUE #2546: Revert pinning version of `astroid` and `pylint`
Versions of `astroid` and `pylint` were pinned in PR #2543 because of a bug in `pylint-plugin-utils` in this [issue](https://github.com/PyCQA/pylint-plugin-utils/issues/12#event-2169384481). This has been resolved with the release of `pylint-plugin-utils==0.5` and `pylint-django==2.0.6`

sphuber:
Fixes #2546 

An update in `prospector` which finally fixes incompatibility issues
with `pylint` and `astroid` allows us to unpin their version
requirements. At the same time, the versions of `yapf` and `precommit`
are upped. The former results in a lot of new formatting rules which are
also applied and committed.

### DIFF
--- a/.ci/polish/cli.py
+++ b/.ci/polish/cli.py
@@ -25,48 +25,55 @@ from aiida.cmdline.utils import decorators
 @options.CODE(
     type=types.CodeParamType(entry_point='arithmetic.add'),
     required=False,
-    help='Code to perform the add operations with. Required if -C flag is specified')
+    help='Code to perform the add operations with. Required if -C flag is specified'
+)
 @click.option(
     '-C',
     '--use-calculations',
     is_flag=True,
     default=False,
     show_default=True,
-    help='Use job calculations to perform all additions')
+    help='Use job calculations to perform all additions'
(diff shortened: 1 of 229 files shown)


sphuber:
> I always wonder if it's a good thing that at every version change most of the code needs to change, though...

I was wondering the same thing and I think it depends. As in, we should not blindly accept all changes, but sometimes it really is an improvement. In this case also for example I added the `raise-no-else` to the list of exceptions because I don't think it is a good change. However, the biggest change which I do find an improvement is the change in hanging indent.
It used to be:
```
some_extremely_long_name.some_method(followed,
                                                                       by,
                                                                       some,
                                                                       arguments)
```
which has now become:
```
some_extremely_long_name.some_method(
    follwed, by, some, arguments
)
````
Which I find a lot more readable, especially when you have of these following one another because at least the indent is consistent and does not depend on the length of the callable.

PR #3223, opened 2019-07-27
Docs: add conventions on use of exit status integers

### LINKED ISSUE #2997: Formalization of meaning of exit code ranges
This issue will be a design document to agree on a guideline for the meaning of certain exit code ranges to be used by plugins and potentially even reserve some for internal use in `aiida-core`.
The proposal in this message will be updated as the discussion progresses:

* `0 - 99` Reserved for internal use by the engine of `aiida-core`
* `100 - 199` Reserved for exit codes by scheduler parsers (This only applies to `CalcJobs` and the scheduler parsing is not yet implemented)
* `200 - 299` Suggested for compound (not automated by engine based on process spec) input validation errors
* `300 - 399` Suggested for unrecoverable errors
* `400 - ...` Suggested for recoverable errors

sphuber:
Fixes #2997 

There is currently no enforcement of the proposed integer ranges and it
is not even clear if this would be possible. Therefore, for now, we just
document the suggested usage in the hope to create as much homogeneity
between the various plugins.

### DIFF
--- a/docs/source/working/processes.rst
+++ b/docs/source/working/processes.rst
@@ -245,6 +245,24 @@ This is an infinitely more robust way of communcating specific errors to a non-h
 Additionally, The exit codes make it also very easy to query for failed processes with specific error codes.
 
 
+.. _working_processes_exit_code_conventions:
+
+Exit code conventions
+.....................
+In principle, the only restriction on the exit status of an exit code is that it should be a positive integer or zero.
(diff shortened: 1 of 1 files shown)


giovannipizzi:
maybe add that in the future we might decide to enforce them, so better to follow the guidelines?

sphuber:
Updated the text

PR #3222, opened 2019-07-27
Make sure `verdi` commands respect PEP 257 docstring conventions

sphuber:
Fixes #3196 

The PEP states that a docstring should start with a single short
description ending with a full stop. The `click` library will take this
as the short help message when displaying the help message of the parent
group. When the message is also sufficienty short, no ellipsis will be
displayed.

P.S.: the "easter-egg" is now a proper hidden easter egg.

PR #3224, opened 2019-07-27
Make sure all CLI help strings are short and consistent

### LINKED ISSUE #3196: verdi help has calculation descriptions that are too long
When running `verdi`, one gets a list of available commands with a short description.

Now, some of them have a too-long description that is truncated:
```
$ verdi 
Usage: verdi [OPTIONS] COMMAND [ARGS]...

  The command line interface of AiiDA.

Options:
  -p, --profile PROFILE  Execute the command for this profile instead of the
                         default profile.
  -v, --version          Show the version and exit.
  -h, --help             Show this message and exit.

Commands:
  calcjob            Inspect and manage calcjobs.
  code               Setup and manage codes.
  comment            Inspect, create and manage node comments.
  completioncommand  Return the bash code to activate completion.
  computer           Setup and manage computers.
  config             Set, unset and get profile specific or global...
  daemon             Inspect and manage the daemon.
  data               Inspect, create and manage data nodes.
  database           Inspect and manage the database.
  devel              Commands for developers.
  export             Create and manage export archives.
  graph              Create visual representations of part of the
                     provenance...
  group              Create, inspect and manage groups.
  import             Import one or multiple exported AiiDA archives The...
  node               Inspect, create and manage nodes.
  plugin             Inspect installed plugins for various entry point...
  process            Inspect and manage processes.
  profile            Inspect and manage the configured profiles.
  quicksetup         Setup a new profile where the database is
                     automatically...
  rehash             Recompute the hash for nodes in the database The set
                     of...
  restapi            Run the AiiDA REST API server Example Usage: verdi
                     -p...
  run                Execute an AiiDA script.
  setup              Setup a new profile.
  shell              Start a python shell with preloaded AiiDA environment.
  status             Print status of AiiDA services.
  user               Inspect and manage users.
```
I think we should fix this to avoid `...`

giovannipizzi:
Fixes #3196
Supersedes #3222 and therefore closes #3222

This makes all commands conform to PEP 257
(in particular dot at the end of the first line)
so that they are properly displayed by click.

I have also used the occasion to put a confirmation
prompt in `verdi rehash` as I realised it would
just rehash without asking. I've added also a force
option to avoid questions asked (e.g. for tests).

I have noticed also that `verdi comment add` and
`verdi comment delete` did not have some arguments
marked as required, but then the command would fail.
Since not passing the argument does not make much sense,
I have made those required.

Finally, I have moved the `verdi rehash`, `verdi graph`
and `verdi comment` inside `verdi node`, but left them
there with a deprecation warning to be backwards-compatible.

Tests have been added for the actual `verdi node xxx` commands,
and the old ones left in there to make sure also the deprecated
commands still work. These tests will be removed when the
deprecated commands will be removed.

### DIFF
--- a/aiida/cmdline/commands/__init__.py
+++ b/aiida/cmdline/commands/__init__.py
@@ -23,8 +23,3 @@ from aiida.cmdline.commands import (
     cmd_daemon, cmd_devel, cmd_export, cmd_graph, cmd_group, cmd_import, cmd_node, cmd_plugin, cmd_process, cmd_profile,
     cmd_rehash, cmd_restapi, cmd_run, cmd_setup, cmd_shell, cmd_status, cmd_user
 )
-
-# Import to populate the `verdi data` sub commands
-from aiida.cmdline.commands.cmd_data import (
-    cmd_array, cmd_bands, cmd_cif, cmd_dict, cmd_remote, cmd_structure, cmd_trajectory, cmd_upf
-)
--- a/aiida/cmdline/commands/cmd_data/cmd_cif.py
+++ b/aiida/cmdline/commands/cmd_data/cmd_cif.py
@@ -29,7 +29,7 @@ VISUALIZATION_FORMATS = ['jmol', 'vesta']
 
 @verdi_data.group('cif')
 def cif():
-    """Manipulation of CIF data objects."""
+    """Manipulate CifData objects (crystal structures in .cif format)."""
 
 
 @cif.command('list')
@@ -92,7 +92,7 @@ def cif_show(data, fmt):
 @arguments.DATA(type=types.DataParamType(sub_classes=('aiida.data:cif',)))
 @decorators.with_dbenv()
 def cif_content(data):
-    """Show the content of the file behind CifData objects."""
+    """Show the content of the CIF file."""
     for node in data:
         try:
             echo.echo(node.get_content())
@@ -121,7 +121,7 @@ def cif_export(**kwargs):
 @click.argument('filename', type=click.Path(exists=True, dir_okay=False, resolve_path=True))
 @decorators.with_dbenv()
 def cif_import(filename):
-    """Import structure into CifData object."""
+    """Import .cif file into CifData object."""
     from aiida.orm import CifData
 
     try:
--- a/aiida/cmdline/commands/cmd_data/cmd_remote.py
+++ b/aiida/cmdline/commands/cmd_data/cmd_remote.py
@@ -23,7 +23,13 @@ from aiida.common.files import get_mode_string
 
 @verdi_data.group('remote')
 def remote():
-    """Managing RemoteData objects."""
+    """Manipulate RemoteData objects (reference to remote folders).
+
+    A RemoteData can be thought as a "symbolic link" to a folder on one of the
+    Computers set up in AiiDA (e.g. where a CalcJob will run).
+    This folder is called "remote" in the sense that it is on a Computer and
+    not in the AiiDA repository. Note, however, that the "remote" computer
+    could also be "localhost"."""
 
 
 @remote.command('ls')
@@ -31,7 +37,7 @@ def remote():
 @click.option('-l', '--long', 'ls_long', is_flag=True, default=False, help='Display also file metadata.')
 @click.option('-p', '--path', type=click.STRING, default='.', help='The folder to list.')
 def remote_ls(ls_long, path, datum):
-    """List directory content on remote RemoteData objects."""
+    """List content of a (sub)directory in a RemoteData object."""
     import datetime
     try:
         content = datum.listdir_withattributes(path=path)
@@ -58,7 +64,7 @@ def remote_ls(ls_long, path, datum):
 @arguments.DATUM(type=types.DataParamType(sub_classes=('aiida.data:remote',)))
 @click.argument('path', type=click.STRING)
 def remote_cat(datum, path):
-    """Show the content of remote files in RemoteData objects."""
+    """Show content of a file in a RemoteData object."""
     import os
     import sys
     import tempfile
@@ -81,7 +87,7 @@ def remote_cat(datum, path):
 @remote.command('show')
 @arguments.DATUM(type=types.DataParamType(sub_classes=('aiida.data:remote',)))
 def remote_show(datum):
-    """Show information on a RemoteData object."""
+    """Show information for a RemoteData object."""
     click.echo('- Remote computer name:')
     click.echo('  {}'.format(datum.get_computer_name()))
     click.echo('- Remote folder full path:')
--- a/aiida/cmdline/commands/cmd_data/cmd_structure.py
+++ b/aiida/cmdline/commands/cmd_data/cmd_structure.py
@@ -54,7 +54,7 @@ def _store_structure(new_structure, dry_run):
 
 @verdi_data.group('structure')
 def structure():
-    """Manipulation of StructureData objects."""
+    """Manipulate StructureData objects (crystal structures)."""
 
 
 # pylint: disable=too-many-locals,too-many-branches
@@ -64,7 +64,7 @@ def structure():
 @list_options
 @decorators.with_dbenv()
 def structure_list(elements, raw, formula_mode, past_days, groups, all_users):
-    """List stored StructureData objects."""
+    """List StructureData objects."""
     from aiida.orm.nodes.data.structure import StructureData, get_formula, get_symbols_string
     from tabulate import tabulate
 
@@ -150,7 +150,7 @@ def structure_show(data, fmt):
 @export_options
 @decorators.with_dbenv()
 def structure_export(**kwargs):
-    """Export StructureData object."""
+    """Export StructureData object to file."""
     node = kwargs.pop('datum')
     output = kwargs.pop('output')
     fmt = kwargs.pop('fmt')
@@ -163,7 +163,7 @@ def structure_export(**kwargs):
 
 @structure.group('import')
 def structure_import():
-    """Import crystal structures from a file."""
+    """Import a crystal structure from file into a StructureData object."""
 
 
 @structure_import.command('aiida-xyz')
--- a/aiida/cmdline/commands/cmd_devel.py
+++ b/aiida/cmdline/commands/cmd_devel.py
@@ -160,9 +160,9 @@ def devel_tests(paths, verbose):  # pylint: disable=too-many-locals,too-many-sta
         sys.exit(len(test_failures) + len(test_errors))
 
 
-@verdi_devel.command('play')
+@verdi_devel.command('play', hidden=True)
 def devel_play():
-    """Open a browser and play the Aida triumphal march by Giuseppe Verdi."""
+    """Play the Aida triumphal march by Giuseppe Verdi."""
     import webbrowser
 
     webbrowser.open_new('http://upload.wikimedia.org/wikipedia/commons/3/32/Triumphal_March_from_Aida.ogg')
--- a/aiida/cmdline/commands/cmd_group.py
+++ b/aiida/cmdline/commands/cmd_group.py
@@ -23,7 +23,7 @@ from aiida.cmdline.utils.decorators import with_dbenv
 
 @verdi.group('group')
 def verdi_group():
-    """Create, inspect and manage groups."""
+    """Create, inspect and manage groups of nodes."""
 
 
 @verdi_group.command('add-nodes')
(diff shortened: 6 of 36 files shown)


giovannipizzi:
Three comments:
- for a quick check of the new strings, one can check the autogenerated docs. I'm open to suggestions on my edits.
- I realised that `verdi rehash` should probably be instead `verdi node rehash`. Do you agree?
- Maybe also `verdi graph` and `verdi comment` should go under `verdi node`?

@sphuber if you agree, we can either change it here or maybe better in a different PR.
The only doubt is about this being a backward-incompatible change. But being on the command line, it's not so bad (and if we agree that the change is the right thing to do, we can anyway move it, and leave the old command as well with some deprecation warning).

aiida/cmdline/commands/__init__.py
```diff
@@ -25,6 +25,8 @@
 )
 
 # Import to populate the `verdi data` sub commands
-from aiida.cmdline.commands.cmd_data import (
-    cmd_array, cmd_bands, cmd_cif, cmd_dict, cmd_remote, cmd_structure, cmd_trajectory, cmd_upf
-)
+## Note: this does not seem to be necessary. With this, running
```
sphuber: If it is not necessary let's remove it and not keep dead code

aiida/cmdline/commands/cmd_data/cmd_cif.py
```diff
@@ -29,7 +29,7 @@
 
 @verdi_data.group('cif')
 def cif():
-    """Manipulation of CIF data objects."""
+    """Manipule CifData objects (crystal structures in .cif format)."""
```
sphuber: Manipulate

aiida/cmdline/commands/cmd_data/cmd_cif.py
```diff
@@ -92,7 +92,7 @@ def cif_show(data, fmt):
 @arguments.DATA(type=types.DataParamType(sub_classes=('aiida.data:cif',)))
 @decorators.with_dbenv()
 def cif_content(data):
-    """Show the content of the file behind CifData objects."""
+    """Show the content of the .cif file."""
```
sphuber: -> CIF file?

aiida/cmdline/commands/cmd_data/cmd_remote.py
```diff
@@ -23,15 +23,15 @@
 
 @verdi_data.group('remote')
 def remote():
-    """Managing RemoteData objects."""
+    """Manipulate RemoteData objects (reference to remorte folders)."""
```
sphuber: remorte -> remote

aiida/cmdline/commands/cmd_data/cmd_remote.py
```diff
@@ -23,15 +23,15 @@
...
 
 
 @remote.command('ls')
 @arguments.DATUM(type=types.DataParamType(sub_classes=('aiida.data:remote',)))
 @click.option('-l', '--long', 'ls_long', is_flag=True, default=False, help='Display also file metadata.')
 @click.option('-p', '--path', type=click.STRING, default='.', help='The folder to list.')
 def remote_ls(ls_long, path, datum):
-    """List directory content on remote RemoteData objects."""
+    """List content of directories in a linked RemoteData object."""
```
sphuber: why "linked"? What does that mean in this context?

aiida/cmdline/commands/cmd_data/cmd_remote.py
```diff
@@ -58,7 +58,7 @@ def remote_ls(ls_long, path, datum):
 @arguments.DATUM(type=types.DataParamType(sub_classes=('aiida.data:remote',)))
 @click.argument('path', type=click.STRING)
 def remote_cat(datum, path):
-    """Show the content of remote files in RemoteData objects."""
+    """Show content of a file in a linked RemoteData object."""
```
sphuber: same as above

aiida/cmdline/commands/cmd_data/cmd_structure.py
```diff
@@ -54,7 +54,7 @@ def _store_structure(new_structure, dry_run):
 
 @verdi_data.group('structure')
 def structure():
-    """Manipulation of StructureData objects."""
+    """Manipule StructureData objects (crystal structures)."""
```
sphuber: Manipule -> manipulate

aiida/cmdline/commands/cmd_data/cmd_structure.py
```diff
@@ -64,7 +64,7 @@ def structure():
 @list_options
 @decorators.with_dbenv()
 def structure_list(elements, raw, formula_mode, past_days, groups, all_users):
-    """List stored StructureData objects."""
+    """List StructureData objects stored in the database."""
```
sphuber: stored in the database is a bit superfluous, no? I mean what else is it going to show? You don't say this phrase anywhere else either even though it also applies there

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -162,7 +162,7 @@ def devel_tests(paths, verbose):  # pylint: disable=too-many-locals,too-many-sta
 
 @verdi_devel.command('play')
```
sphuber: Make this `hidden=True` to make it a truly hidden easter egg ;)

aiida/cmdline/commands/cmd_group.py
```diff
@@ -32,7 +32,7 @@ def verdi_group():
 @arguments.NODES()
 @with_dbenv()
 def group_add_nodes(group, force, nodes):
-    """Add NODES to the given GROUP."""
+    """Add modes to the a group."""
```
sphuber: modes -> nodes

aiida/cmdline/commands/cmd_group.py
```diff
@@ -136,7 +136,9 @@ def group_description(group, description):
 @arguments.GROUP()
 @with_dbenv()
 def group_show(group, raw, uuid):
-    """Show information on a given group. Pass the GROUP as a parameter."""
+    """Show information on a given group.
```
sphuber: show information *for* a given

aiida/cmdline/commands/cmd_group.py
```diff
@@ -136,7 +136,9 @@ def group_description(group, description):
...
+
+    Pass the group as a parameter."""
```
sphuber: "Pass the group .." is superfluous. It just states that it takes an argument, but that should be clear

aiida/cmdline/commands/cmd_group.py
```diff
@@ -323,7 +325,7 @@ def group_list(
 @click.argument('group_label', nargs=1, type=click.STRING)
 @with_dbenv()
 def group_create(group_label):
-    """Create a new empty group with the name GROUP_NAME."""
+    """Create a new empty group with a given name."""
```
sphuber: remove "new"

aiida/cmdline/commands/cmd_node.py
```diff
@@ -27,15 +27,15 @@ def verdi_node():
 
 @verdi_node.group('repo')
 def verdi_node_repo():
-    pass
+    """Manage and inspect the content of a node repository folder."""
```
sphuber: There are no "managing" commands, only to inspect, correct?

aiida/cmdline/commands/cmd_plugin.py
```diff
@@ -21,7 +21,7 @@
 
 @verdi.group('plugin')
 def verdi_plugin():
-    """Inspect installed plugins for various entry point categories."""
+    """Inspect and manage AiiDA plugins."""
```
sphuber: For now there is only inspection of plugins, no managing

aiida/cmdline/commands/cmd_rehash.py
```diff
@@ -29,14 +29,35 @@
...
+    """Recompute the hash for nodes in the database.
 
     The set of nodes that will be rehashed can be filtered by their identifier and/or based on their class.
     """
     from aiida.orm import Data, ProcessNode, QueryBuilder
 
+    if not force:
+        echo.echo_warning('This command will recompute and overwrite the hashes of all nodes.')
+        echo.echo_warning('Note that this can take a lot of time for big databases.')
+        echo.echo_warning('Also, do it only if you are sure of what you are doing.')
```
sphuber: Is the third warning necessary? Apart from it taking long, it cannot do much harm can it? Often we even tell them to do it, such as after a migration.

aiida/cmdline/commands/cmd_rehash.py
```diff
@@ -29,14 +29,35 @@
...
+        echo.echo_warning('')
+        echo.echo_warning('', nl=False)
+
+        confirm_message = 'Do you want to continue? [y/n]'
+
+        try:
+            while True:
```
sphuber: why the while loop? We can do just a single if statement:
```
if not click.prompt():
    echo.echo_critical()
```

aiida/cmdline/commands/cmd_setup.py
```diff
@@ -123,7 +123,7 @@ def quicksetup(
     ctx, non_interactive, profile, email, first_name, last_name, institution, db_engine, db_backend, db_host, db_port,
     db_name, db_username, db_password, su_db_name, su_db_username, su_db_password, repository
 ):
-    """Setup a new profile where the database is automatically created and configured."""
+    """Setup a new profile if a fully automated fashion."""
```
sphuber: if -> in

sphuber: And it is not fully automated, because still asks for user details, but ok

sphuber (changes requested):

aiida/cmdline/commands/__init__.py
giovannipizzi: OK, I was just checking this. Indeed, the reason is that they are defined as entry points.
The latest test fails because `remote` was not in the entry points. I've fixed this now.

aiida/cmdline/commands/cmd_data/cmd_remote.py
giovannipizzi: I meant to say that RemoteData is a "link" to a folder (I didn't want to say "remote folder" because it might be on localhost). If it's not clear, let me know and I remove "linked"

aiida/cmdline/commands/cmd_group.py
giovannipizzi: Agreed. It was there and I didn't want to to remove it, but I agree it's better to remove.

aiida/cmdline/commands/cmd_node.py
giovannipizzi: Correct.

aiida/cmdline/commands/cmd_rehash.py
giovannipizzi: True. I'm still not 100% sure that the hash will be always properly calculated for Floats, but for now I'll remove this warning.

aiida/cmdline/commands/cmd_rehash.py
giovannipizzi: It's in case the user replies with a different letter. Do you want to stop also in that case?

giovannipizzi:
Thanks and sorry for the multiple typos (Saturday evening... ;-)) I fixed all but two, let me know what you want me to do there.

aiida/cmdline/commands/cmd_rehash.py
giovannipizzi: Ok, I changed to use click.prompt's list of valid choices

aiida/cmdline/commands/cmd_data/cmd_remote.py
giovannipizzi: Ok, I fixed by removing "linked" and explaining better what a RemoteData is

giovannipizzi:
Ok, I decided how to address the two remaining comments and pushed.
Let's see if I didn't do any more mistake and if the tests pass ;-)

Also, let me know about the three commands `rehash`, `graph`, `comment` and moving them under `node`.

sphuber:
Just one more change. If you could also rebase, so I can merge straight after and dont have to rerun the tests. Regarding moving the commands, I'm fine with it in principle. But you are right that we should leave them in and deprecate them. There is a decorator for this in `aiida.cmdline.utils.decorators`

giovannipizzi:
I did all the git magic and moved the commands to be moved (keeping the old ones as deprecated) so this is ready to be merged for me

sphuber (approved):

sphuber:
Thanks a lot @giovannipizzi

PR #3233, opened 2019-07-29
Work on queries documentation

### LINKED ISSUE #3025: Check documentation for "working with AiiDA/Querying data"

https://aiida.readthedocs.io/projects/aiida-core/en/latest/working_with_aiida/index.html#querying-data

yakutovicha:
fixes #3025
adapt queries documentation to aiida-1.0.0b5 release


yakutovicha:
The appender part seems to be alright, while the queryhelper is a bit more tricky. I have come across three issues that I am going to describe in further comments

yakutovicha:
*problem 1*: direction

if I do a similar query as described in the docs: 
```
qh1 = {
    'path':[
        {
            'cls':Cp2kCalculation
        },
        {
            'cls':Dict
        },
        {
            'cls':Dict,
            'direction':-2
        }
    ]
}
```

I get the following error:
```
/home/epfl/work/aiida_core/aiida/orm/querybuilder.py in _get_tag_from_specification(self, specification)
   1214                                            'The classes that I can do this for are:{}\n'
   1215                                            'The tags I have are: {}'.format(specification, self._cls_to_tag_map.keys(),
-> 1216                                                                             self.tag_to_alias_map.keys()))
   1217         return tag
   1218 

InputValidationError: You specified as a class for which I have to find a tag
The classes that I can do this for are:-2
The tags I have are: dict_keys([<class 'aiida_cp2k.calculations.Cp2kCalculation'>, <class 'aiida.orm.nodes.data.dict.Dict'>])
```
not sure how to fix this though.

yakutovicha:
*problem 2* is a bit confusing. The documentation says:

Notice that the tag for the link, by default, is the tag of the two connecting
nodes delimited by two dashes '--'. The order does not matter, the following queryhelp would results in the same query::
```
queryhelp = {
    'path':[
        {'cls':Cp2kCalculation, 'tag':'relax'},         # Relaxation with structure as output
        {'cls':StructureData, 'tag':'structure'}
    ],
    'filters':{
        'structure':{
            'id':{'>': 50}
        },
        'relax--structure':{
            'label':{'like':'output_%'},
        }
    },
    'project':{
        'relax--structure':['label'],
        'structure':['label'],
        'relax':['label', 'state'],
    }
}
```
And provides *exactly the same* query as before.

I've tried to replace `relax--structure` with `structure--relax` but got an error:
```
InputValidationError: tag structure--relax is not among my known tags
My tags are: dict_keys(['relax', 'structure', 'relax--structure'])
```

For the time being I commented this part out.

docs/source/querying/querybuilder/append.rst
```diff
@@ -707,8 +707,7 @@ Let's take the above example, but put a filter on the label of the link and proj
     qb = QueryBuilder()
```
CasperWA: In the text section above, it says something about the keyword being preceeded by '*link*', but this should be `edge`, right?

docs/source/querying/querybuilder/append.rst
```diff
@@ -749,7 +748,6 @@ You can also limit the number of rows returned with the method *limit*::
...
     qb.append(
         CalcJobNode,
-        filters={'ctime':{'>': now - timedelta(days=3)}},
```
CasperWA: Has this filter been removed for all examples? If not, I guess it should be?

docs/source/querying/querybuilder/intro.rst
```diff
@@ -16,7 +16,7 @@ Before starting to write a query, it helps to:
     Nodes of an AiiDA graph (vertices) are connected with links (edges).
     A *Node* can be either input or output of another *Node*, but also an
     ancestor or a descendant.
```
CasperWA: Do we still use 'ancestor' and 'descendant'? Aren't we using 'successor' and 'predecessor' now, or something less family-like? Just a question, since I cannot remember - I can only remember there was a discussion about this.

docs/source/querying/querybuilder/queryhelp.rst
```diff
@@ -176,14 +162,17 @@ What do you have to specify:
     but only the ones that were added
     after a certain time `t` and have an id higher than 50::
 
+        from aiida.common import timezone
```
CasperWA: When you update this, you should update the text above as well, which mentions the (now deleted) time `t`.

docs/source/querying/querybuilder/queryhelp.rst
```diff
@@ -317,73 +306,71 @@ Let's take an example that we had and add a few filters on the link::
...
             'relax--structure':{
-                'time':{'>': t},
                 'label':{'like':'output_%'},
             }
         },
         'project':{
             'relax--structure':['label'],
             'structure':['label'],
-            'relax':['label', 'state'],
+            'relax':['label', 'uuid'],
```
CasperWA: I guess you could use 'process_state' instead? Also in an example above, where you changed 'state'.

docs/source/querying/querybuilder/append.rst
yakutovicha: Sorry, I did not find anything that is related to links in the text above. The only sentence that says about links below is:
> Another feature that had to be added are projections, filters and labels on the edges of the graphs, that is to say links or paths between nodes. It works the same way, just that the keyword is preceeded by ‘link’. Let’s take the above example, but put a filter on the label of the link and project the link label:

docs/source/querying/querybuilder/append.rst
yakutovicha: Well, not everywhere. If the description explicitly says
> How, can we have multiple filters? Suppose you are interested in all calculations in your database that are in state ‘FINISHED’ and were created **in the last n days**:

Then it should be kept.

docs/source/querying/querybuilder/intro.rst
yakutovicha: Valid keywords for queries are: 'with_incoming', 'with_outgoing', 'with_descendants', 'with_ancestors'

docs/source/querying/querybuilder/queryhelp.rst
yakutovicha: right, thanks!

docs/source/querying/querybuilder/queryhelp.rst
yakutovicha: I just tried and this is what I got:
```
InputValidationError: process_state is not a column of aliased(DbNode)
Valid columns are:
id
uuid
node_type
process_type
label
description
ctime
mtime
attributes
extras
user_id
dbcomputer_id
```

docs/source/querying/querybuilder/append.rst
yakutovicha: sorry, you were right.

docs/source/querying/querybuilder/append.rst
```diff
@@ -700,15 +700,14 @@ Working with edges
 ++++++++++++++++++
 
 Another feature that had to be added are projections, filters and labels on
-the edges of the graphs, that is to say links or paths between nodes.
-It works the same way, just that the keyword is preceeded by '*link*'.
+the edges of the graphs, that is to say, links or paths between nodes.
+It works the same way, just that the keyword is preceded by '*link*'.
```
CasperWA:
```suggestion
It works the same way, just that the keyword is preceded by '*edge*'.
```

yakutovicha: I just didn't push it. Sorry :(

sphuber:
Thanks guys!

PR #3240, opened 2019-08-05
Create backend-agnostic Backup class

### LINKED ISSUE #3201: `load_profile` is not enough to load the profile

CasperWA:
Fixes #3201

The new file `backup_general.py` hosts a backend-agnostic Backup class
that will query through `aiida.orm.QueryBuilder`, instead of using the
backend specific query language.
This makes the backup-module completely backend-agnostic, drastically
simplifying it, however, it may also become slightly slower.
A mid-way solution should be considered, where perhaps one may employ
backend-specific Backup classes, performing backend-specific queries,
increasing speed - but only if the backup is otherwise exceedingly slow.
It is recommended to be as agnostic as possible.

It is recommended to do benchmark tests of the speed of this change before merging.

This PR is derived from PR #3205


aiida/orm/entities.py
```diff
@@ -102,7 +102,6 @@ def query(self):
         :return: a new query builder instance
         :rtype: :class:`aiida.orm.QueryBuilder`
         """
-        # pylint: disable=no-self-use
```
CasperWA: This should probably be moved to PR #3241

CasperWA:
It should be noted that anyone interested is welcome to perform benchmark tests of this :)

CasperWA:
I have done some testing.

Creating two new databases using the different backends, they have been filled with the same Nodes.
Each Node is an `Int(1)` and has a 1 kB `extra` binary string.
I have filled up the databases with 100 MB worth of 1 kB Nodes, amounting to 102400 Nodes (When 1 MB = 1024 kB).

For each backup run, the `backup_info.json` has its key `oldest_object_backedup` reset to the value `'2019-01-01 00:00:00+00:00'`, so that a backup run can be run in a for loop for N times, where N can be varied, and all Nodes are backed up for each run.

The following table shows the number of backup runs performed, as well as the average time for the runs. Furthermore, 'before' corresponds to the develop branch as is (with the split Backup classes for each backend) and 'after' then corresponds to the changes introduced in this PR.

|   | Django | SQLAlchemy |
| --- | --- | --- |
| before | 100 runs, **0.66s** | 100 runs, **14s** |
| after | 110 runs, **1.6s** | 105 runs, **18s** |

The reason for the different number of runs is due to initial testing of the scripts used.

Below is a graph show all timings for the runs:
backup_times

CasperWA:
What is missing is the timings for different sized databases, however, I am not sure it is needed?

In the end, what it shows is a decrease in efficiency of **142%** for the Django backend and of **29%** for the SQLAlchemy backend.

This sounds bad, but in absolute time the increase for a database of 100k+ Nodes is not severe, I think.
The gain is a more neat Backup class hierarchy and one less place in the code, where the use of direct backend querying is avoided.

giovannipizzi:
Thanks for the accurate testing! Do you know why SQLA is so much slower?

sphuber:
I looked through the code and for me this is good to merge. I'm ok with the decrease in performance for reduced code complexity/fragility. If performance becomes a problem, we can address it in the future

CasperWA:
> Thanks for the accurate testing! Do you know why SQLA is so much slower?

No. Actually, it quite surprised me, since I believe we have a bit more steps to go through when querying the Django backend, versus the SQLAlchemy backend through the QueryBuilder?

For completion, (and this has nothing to do with the question above) it is worth mentioning that this testing is done only for the difference introduced in this PR to how the database is queried. Therefore, while the Backup classes indeed also copies the repository, it does not the target of the current investigation. Therefore, all Nodes have empty repository folders and the "meat" of the Nodes have been placed in such a way as to fill the database, as well as create a multitude of Nodes - making querying more taxing.

PR #3242, opened 2019-08-05
Rework export archive extraction

### LINKED ISSUE #3193: Move `verdi export migrate` logic to export/import module or do like `verdi import`
All logic and checks performed in `verdi export migrate` should be moved either to the export/import module or in a separate helper function, as is done for `verdi import` with `_try_import` and more.

The verdi command should not have to unpack the export archive and load the json files.

### LINKED ISSUE #3199: Pre-extracted archives can not be imported
The function `extract_tree` does not work as intended.

At the moment only compressed file archives can be imported into AiiDA. If one has already extracted the archive and direct the import function(s) to the folder, it is not possible to import the archive.

CasperWA:
Fixes #3199 
Fixes #3193 (since `Archive` is now utilized)

The function `extract_tree` now returns a `Folder` object, pointing to
the archive folder, which should not automatically erase the folder
after import.

The difference between the functions `extract_zip`, `extract_tar` and `extract_tree` now, is that `extract_zip` and `extract_tar` unpacks the compressed files to a `SandboxFolder` under `.aiida/repository/<AiiDA user>/sandbox` (or similar), and then uses the files from there to whatever purpose and finally erases the extracted files (and folders) again.

`extract_tree` will instead return a `Folder` object (parent class of `SandboxFolder`), which points to the "original" folder/tree - the source.
This can potentially create some problems, since one can use `extract_tree` to get a `Folder` object and then use the `erase` method to remove the only known instance of the data.
This may be remedied by copying the contents of the specified path for `extract_tree` to a temporary `SandboxFolder` (as is essentially done when unpacking the zip or tar files) and then exclusively operate on the `SandboxFolder`. The problem, however, with this is the extra space this will take up for large archives/folders/trees. Therefore, I have chosen the solution of pointing to the "original" folder/tree, making sure `erase` is not called upon `__exit__`. The caveat being that I could have missed a specific call to `erase` somewhere, however, this does not seem to be the case - according to the tests :)

**Update**
`extract_tree` is now `get_tree` and all extraction functions have had their uses cut down to a single place; the `Archive` class. This class will henceforth be used in import/export to deal with export archives. I believe this was also @sphuber's original intent for this class(?)

*To do*
~- [ ] Thoroughly introduce `Archive` to the export function~ No need.
- [x] Update migrations to rely on `Archive`
- [x] Move some code from `verdi export migrate` to the `aiida.tools.importexport.migration` module
- [x] Update all tests to use either `Archive` or the extraction functions directly, where it may be needed
- [x] Consider updating test utility functions to use `Archive`

### DIFF
--- a/aiida/tools/importexport/common/archive.py
+++ b/aiida/tools/importexport/common/archive.py
@@ -18,12 +18,15 @@ import zipfile
 from wrapt import decorator
 
 from aiida.common import json
-from aiida.common.exceptions import ContentNotExistent, InvalidOperation
+from aiida.common.exceptions import InvalidOperation
 from aiida.common.folders import SandboxFolder
 from aiida.tools.importexport.common.config import NODES_EXPORT_SUBFOLDER
 from aiida.tools.importexport.common.exceptions import CorruptArchive
 
 __all__ = ('Archive', 'extract_zip', 'extract_tar', 'extract_tree')
+
+FILENAME_DATA = 'data.json'
+FILENAME_METADATA = 'metadata.json'
 
 
 class Archive:
@@ -38,11 +41,10 @@ class Archive:
 
     """
 
-    FILENAME_DATA = 'data.json'
-    FILENAME_METADATA = 'metadata.json'
-
-    def __init__(self, filepath):
+    def __init__(self, filepath, silent=True, sandbox_in_repo=True):
         self._filepath = filepath
+        self._silent = silent
+        self._sandbox_in_repo = sandbox_in_repo
         self._folder = None
         self._unpacked = False
         self._data = None
@@ -50,7 +52,7 @@ class Archive:
 
     def __enter__(self):
         """Instantiate a SandboxFolder into which the archive can be lazily unpacked."""
-        self._folder = SandboxFolder()
+        self._folder = SandboxFolder(sandbox_in_repo=self._sandbox_in_repo)
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
@@ -77,17 +79,24 @@ class Archive:
     @ensure_within_context
     def unpack(self):
         """Unpack the archive and store the contents in a sandbox."""
-        if os.path.isdir(self.filepath):
-            self._folder = extract_tree(self.filepath, self.folder)
-        elif tarfile.is_tarfile(self.filepath):
-            extract_tar(self.filepath, self.folder, silent=True, nodes_export_subfolder=NODES_EXPORT_SUBFOLDER)
-        elif zipfile.is_zipfile(self.filepath):
-            extract_zip(self.filepath, self.folder, silent=True, nodes_export_subfolder=NODES_EXPORT_SUBFOLDER)
-        else:
-            raise CorruptArchive('unrecognized archive format')
+        try:
+            if os.path.isdir(self.filepath):
+                self._folder = extract_tree(self.filepath, self.folder)
+            elif tarfile.is_tarfile(self.filepath):
+                extract_tar(
+                    self.filepath, self.folder, silent=self.silent, nodes_export_subfolder=NODES_EXPORT_SUBFOLDER
+                )
+            elif zipfile.is_zipfile(self.filepath):
+                extract_zip(
+                    self.filepath, self.folder, silent=self.silent, nodes_export_subfolder=NODES_EXPORT_SUBFOLDER
+                )
+            else:
+                raise CorruptArchive('unrecognized archive format')
+        except Exception as why:
+            raise CorruptArchive('Error during unpacking of {}: {}'.format(self.filepath, why))
 
         if not self.folder.get_content_list():
-            raise ContentNotExistent('the provided archive {} is empty'.format(self.filepath))
+            raise CorruptArchive('the provided archive {} is empty'.format(self.filepath))
 
         self._unpacked = True
 
@@ -98,6 +107,11 @@ class Archive:
         :return: the archive filepath
         """
         return self._filepath
+
+    @property
+    def silent(self):
+        """Return whether unpacking should be done silently."""
+        return self._silent
 
     @property
     def folder(self):
@@ -115,7 +129,7 @@ class Archive:
         :return: dictionary with contents of data file
         """
         if self._data is None:
-            self._data = self._read_json_file(self.FILENAME_DATA)
+            self._data = self._read_json_file(FILENAME_DATA)
 
         return self._data
 
@@ -127,7 +141,7 @@ class Archive:
         :return: dictionary with contents of meta data file
         """
         if self._meta_data is None:
-            self._meta_data = self._read_json_file(self.FILENAME_METADATA)
+            self._meta_data = self._read_json_file(FILENAME_METADATA)
 
         return self._meta_data
 
@@ -215,8 +229,11 @@ class Archive:
         :param filename: the filename relative to the sandbox folder
         :return: a dictionary with the loaded JSON content
         """
-        with io.open(self.folder.get_abs_path(filename), 'r', encoding='utf8') as fhandle:
-            return json.load(fhandle)
+        try:
+            with io.open(self.folder.get_abs_path(filename), 'r', encoding='utf8') as fhandle:
+                return json.load(fhandle)
+        except OSError as error:
+            raise CorruptArchive('Error reading {}: {}'.format(filename, error))
 
 
 def extract_zip(infile, folder, nodes_export_subfolder=None, silent=False):
@@ -240,6 +257,8 @@ def extract_zip(infile, folder, nodes_export_subfolder=None, silent=False):
         incorrect formats
     """
     # pylint: disable=fixme
+    json_files = [FILENAME_DATA, FILENAME_METADATA]
+
     if not silent:
         print('READING DATA AND METADATA...')
 
@@ -252,18 +271,14 @@ def extract_zip(infile, folder, nodes_export_subfolder=None, silent=False):
     try:
         with zipfile.ZipFile(infile, 'r', allowZip64=True) as handle:
 
-            if not handle.namelist():
+            if not handle.infolist():
                 raise CorruptArchive('no files detected')
 
-            try:
-                handle.extract(path=folder.abspath, member='metadata.json')
-            except KeyError:
-                raise CorruptArchive('required file `metadata.json` is not included')
-
-            try:
-                handle.extract(path=folder.abspath, member='data.json')
-            except KeyError:
-                raise CorruptArchive('required file `data.json` is not included')
+            for json_file in json_files:
+                try:
+                    handle.extract(path=folder.abspath, member=json_file)
+                except KeyError:
+                    raise CorruptArchive('required file `{}` is not included'.format(json_file))
 
             if not silent:
                 print('EXTRACTING NODE DATA...')
@@ -300,6 +315,8 @@ def extract_tar(infile, folder, nodes_export_subfolder=None, silent=False):
         incorrect formats
     """
     # pylint: disable=fixme
+    json_files = [FILENAME_DATA, FILENAME_METADATA]
+
     if not silent:
         print('READING DATA AND METADATA...')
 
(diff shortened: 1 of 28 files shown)


CasperWA:
Found and error with this.
When the import functions copy in the new repository subfolder for each Node, it creates a `SandboxFolder` object for each subfolder, meaning the "source" subfolder will be deleted upon a successful transfer.
This should either be fixed in the import functions, when creating the subfolder objects, or in the extract_tree function, where the files should first be copied and then used.
Personally, I still prefer not to copy the tree, and will instead implement a "switch"/if-else in the import functions, when creating the subfolder objects.

CasperWA:
I am currently rewriting this, in order to properly utilize the "new" `Archive` class made by @sphuber. The idea is that the extraction functions will be used solely by `Archive`, and the import and migration functions will use an `Archive` instance/object to load the files needed. This should reduce lines of code, but have no effect on performance (i.e., speed and stability should be roughly the same as before).

CasperWA:
The `Archive` class has bloated a bit - I have tried to keep it reasonable, but wanting to make it "repackable", I had to introduce several minor functions and properties.
On the other side, it _is_ now repackable, i.e., one can use the `repack()` method (when in a context) to write the `meta_data` and `data` properties/dictionaries back to their respective JSON files in the `Archive.folder.abspath` and subsequently automatically write it all to a specified compressed tar or zip file. The format cannot be chosen manually, but will automatically be chosen based on the original archive format (if it is a folder/tree, the format will be zip).

This new `repack()` method makes it easy to use the `Archive` class for migrations. Hence, a new migration workflow has been introduced that utilizes this method and minimizes the amount of "backend" code in the click function for `verdi export migrate`.

aiida/tools/importexport/common/archive.py
```diff
@@ -40,59 +60,135 @@ class Archive(object):
...
+        self._output_filepath = output_filepath
+        self._overwrite = overwrite
+        self._silent = silent
+        self._sandbox_in_repo = sandbox_in_repo
         self._folder = None
         self._unpacked = False
+        self._keep = False
         self._data = None
         self._meta_data = None
+        self._archive_format = None
```
sphuber: You are not using this consistently. For starters, I cannot find where you _set_ this. But then you _do_ use the value to switch on in `unpack` and `repack`. You are also sometimes using the attribute and sometimes the property

aiida/tools/importexport/common/archive.py
```diff
@@ -102,14 +198,56 @@ def filepath(self):
...
+    @property
+    def output_filepath(self):
+        """Return the filepath of where the archive should be repacked
+
+        :return: the archive output filepath
+        """
+        return self._output_filepath
+
+    @property
+    def overwrite(self):
```
sphuber: Since overwrite should no longer be a class attribute but rather just a write method argument, it no longer needs a setter.

aiida/tools/importexport/common/archive.py
```diff
@@ -102,14 +198,56 @@ def filepath(self):
...
+        """Return whether repacking should be overwrite a possible existing file."""
+        return self._overwrite
+
+    @property
+    def silent(self):
```
sphuber: I don't think `silent` needs to be set or retrieved from the outside. One should just open the `Archive` class with the correct `silent` value and use it like that. Why would we need the functionality of changing this value during the context manager. These properties are not really necessary and are what contributes a lot to the bloating

aiida/tools/importexport/common/archive.py
```diff
@@ -102,14 +198,56 @@ def filepath(self):
         """
         return self._filepath
 
+    @property
```
sphuber: See comments below. If this attribute is and only should be used internally, there is no need for a property.

aiida/tools/importexport/common/archive.py
```diff
@@ -130,7 +268,7 @@ def meta_data(self):
         :return: dictionary with contents of meta data file
         """
         if self._meta_data is None:
-            self._meta_data = self._read_json_file(self.FILENAME_METADATA)
+            self._meta_data = self._read_json_file(FILENAME_METADATA)
```
sphuber: Why did you make `FILENAME_METADATA` and `FILENAME_DATA` global variables. Do they need to be accessed outside of this class? Wouldn't it make more sense to make these class attributes

aiida/tools/importexport/common/archive.py
```diff
@@ -217,9 +352,103 @@ def _read_json_file(self, filename):
...
+        :param key: Key matching the value to be returned.
+        :type key: str
+
+        :return: Value of key in JSON file
+        :rtype: dict, str, list
+
+        :raises ~aiida.tools.importexport.common.exceptions.CorruptArchive: if the key cannot be found in JSON file
+        """
+        if json_filename == FILENAME_DATA:
+            json_file = self.data
```
sphuber: `json_file` can be a bit misleading. It is not actually the file, but already the loaded content, so maybe `json_content` is a clearer name?

aiida/tools/importexport/common/archive.py
```diff
@@ -217,9 +352,103 @@ def _read_json_file(self, filename):
...
+            mandatory_keys = self.MANDATORY_DATA_KEYS
+        elif json_filename == FILENAME_METADATA:
+            json_file = self.meta_data
+            mandatory_keys = self.MANDATORY_METADATA_KEYS
+        else:
+            raise CorruptArchive('{} is not a valid archive JSON file'.format(json_file))
+
+        if key in json_file:
```
sphuber: I think a try/except `KeyError` would be more elegant. You can keep the second if-conditional in the except block

aiida/tools/importexport/common/archive.py
```diff
@@ -217,9 +352,103 @@ def _read_json_file(self, filename):
...
+            return json_file[key]
+        if key not in mandatory_keys:
+            return {}
+        # else
```
sphuber: I would remove this comment. We use the early-return pattern throughout the code base.

aiida/tools/importexport/common/archive.py
```diff
@@ -217,9 +352,103 @@ def _read_json_file(self, filename):
...
+        raise CorruptArchive('{} is missing the mandatory key "{}"'.format(json_filename, key))
+
+    @ensure_within_context
+    def read_data(self, key):
```
sphuber: I would rename this. The `read_` prefix suggests that the content is being read from file, but this is actually done lazily by the `data` and `metadata` property. Maybe `get_data_member` and `get_meta_data_member`

aiida/tools/importexport/common/archive.py
```diff
@@ -217,9 +352,103 @@ def _read_json_file(self, filename):
...
+
+        :raises ~aiida.tools.importexport.common.exceptions.CorruptArchive: if there was an error writing the JSON file
+        """
+        try:
+            with io.open(self.folder.get_abs_path(filename), 'wb') as fhandle:
+                json.dump(data, fhandle)
+
+            self._keep = True
+        except OSError as error:
+            raise CorruptArchive('Error writing {}: {}'.format(filename, error))
```
sphuber: The `CorruptArchive` exception seems a bit weird to throw here. It is not that the original archive is corrupt it is just that writing the altered one fails. Should be something like an operational error

aiida/tools/importexport/common/archive.py
```diff
@@ -217,9 +352,103 @@ def _read_json_file(self, filename):
...
+        :param data: the data to be written to a JSON file
+        :type data: dict
+
+        :raises ~aiida.tools.importexport.common.exceptions.CorruptArchive: if there was an error writing the JSON file
+        """
+        try:
+            with io.open(self.folder.get_abs_path(filename), 'wb') as fhandle:
+                json.dump(data, fhandle)
+
+            self._keep = True
```
sphuber: Why do we have to change `_keep` here? Is the idea that as soon as anything is written, it should always be kept?

aiida/tools/importexport/common/archive.py
```diff
@@ -40,59 +60,135 @@ class Archive(object):
...
-    FILENAME_DATA = 'data.json'
-    FILENAME_METADATA = 'metadata.json'
-
-    def __init__(self, filepath):
+    def __init__(self, filepath, output_filepath=None, overwrite=False, silent=True, sandbox_in_repo=True):
         self._filepath = filepath
+        self._output_filepath = output_filepath
+        self._overwrite = overwrite
+        self._silent = silent
+        self._sandbox_in_repo = sandbox_in_repo
```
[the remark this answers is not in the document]
CasperWA: This was introduced in an attempt to make use of the `Archive` class during migration, i.e., specifically when calling `verdi export migrate`. Here, for some reason, this was set to `False`.

I agree with your sentiment, but have not yet come around to addressing this, mostly due to the lack of understanding why it was set to `False` for migration in the first place. My guess would be for testing purposes during development.

I will remove it.

aiida/tools/importexport/common/archive.py
CasperWA: I set it in `unpack`, where it is set to `zip`, unless the archive was a `tar` file, in which case it is set to `tar`.

I will go through its uses. Also, I think this is an attribute that should be completely reworked from a design point of view, in order to make it possible to choose the archive you are outputting (if repacking).

aiida/tools/importexport/common/archive.py
```diff
@@ -40,59 +60,135 @@ class Archive(object):
...
+        """Clean the sandbox folder if it was instantiated."""
+        if not self.keep and self.folder:
             self.folder.erase()
 
-    @decorator
-    def ensure_within_context(wrapped, instance, args, kwargs):  # pylint: disable=no-self-argument
-        """Decorator to ensure that the instance is called within a context manager."""
-        if instance and not instance.folder:
-            raise InvalidOperation('the Archive class should be used within a context')
+    def _ensure_within_context(self):
```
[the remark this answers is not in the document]
CasperWA: True - it comes from the desire to have a class used only for tests that subclasses `Archive`, but otherwise does not need everything to be in a context. For this I needed the decorator to call an "overridable" class method.

aiida/tools/importexport/common/archive.py
CasperWA: These are used in the `extract_*` functions - outside the class.

aiida/tools/importexport/common/archive.py
CasperWA: Exactly.

aiida/tools/importexport/common/archive.py
```diff
@@ -40,59 +60,135 @@ class Archive(object):
...
-    def __init__(self, filepath):
+    def __init__(self, filepath, output_filepath=None, overwrite=False, silent=True, sandbox_in_repo=True):
         self._filepath = filepath
+        self._output_filepath = output_filepath
+        self._overwrite = overwrite
+        self._silent = silent
+        self._sandbox_in_repo = sandbox_in_repo
         self._folder = None
         self._unpacked = False
+        self._keep = False
```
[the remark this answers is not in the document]
CasperWA: Very sensible idea!

However, this is somewhat coupled with the fact that I am still trying to get around having to copy all the files (if they are an extracted archive - a folder on the hdd) before repacking. But maybe there is no getting around this, if we wish to use this class for migrations and the archived node repository files need to be altered. If it was only the JSON files, it may have been possible. Unless the class could handle a selective copying and altering during migrations that the migration functions do not need to know about?

aiida/tools/importexport/common/archive.py
CasperWA: Completely agree. This is why more eyes are needed on such things :) Thanks.

sphuber:
Since this has been open for very long and we first need to find some time to discuss the redesign, is it ok to close this for now @CasperWA ?

CasperWA:
> Since this has been open for very long and we first need to find some time to discuss the redesign, is it ok to close this for now @CasperWA ?

Since this needs redesign. Yes, it can be closed (if you absolutely do not wish it to be open here).

PR #3249, opened 2019-08-12
Remove `qe-tools` dependency

### LINKED ISSUE #3248: Remove `qe-tools` dependency
Even though this dependency is optional through the `atomic_tools` extra, it only adds functionality to deal with Quantum ESPRESSO specific input files, which should therefore be added in the `aiida-quantumespresso` plugin and not in `aiida-core`.

sphuber:
Fixes #3248 

Even though this dependency is optional through the `atomic_tools` extra,
it only adds functionality to deal with Quantum ESPRESSO specific input
files, which should therefore be added in the `aiida-quantumespresso`
plugin and not in `aiida-core`.


CasperWA:
All the removed functionality here will be added to the `aiida-quantumespresso` plugin, correct?

sphuber:
Yes, I am working on this now

sphuber:
Functionality recreated in `aiida-quantumespresso` with [this PR](https://github.com/aiidateam/aiida-quantumespresso/pull/374/files)

CasperWA:
Ah wait - should we possibly add a deprecation warning and not just remove the `verdi` command?

sphuber:
No, not necessary, this was recently added anyway. I will take responsibility ;)

giovannipizzi:
I'm not sure I 100% agree with this removal, even if it's been already done.

My reasoning is that, as long as the `StructureData` class is in QE, importing a file in QE format is something generic and shouldn't necessarily need to have the `aiida-quantuemespresso` plugin installed (use case: a VASP user gets an input file with a crystal structure in QE input format, and wants to run from that crystal structure).

On the other hand probably once AiiDA is installed, installing the plugin is straightforward so it's ok as it is now? I'd like to hear @sphuber and @CasperWA opinions on this

sphuber:
I came to this problem due to a dependency problem with `qe-tools` once i added it to `aiida-quantumespresso`. Since `aiida-core` also defined it, but another version, the documentation stopped building on RTD. I personally think that `qe-tools` should not be a dependency of `aiida-core` even if its optional. I can see your point now that `StructureData` is still part of `aiida-core` this could be argued, but just raising this point shows to me that it does not really belong there. In addition, just having the `qe-tools` as dependency was not enough. Specific tools had to be implemented and added to `aiida-core` that again only serve to parse and read a QE input file. I don't think it is a good idea to start adding this kind of code for all types of external formats specific to a code. In any case you would have to install another library, why not an official AiiDA plugin. With pip it is just as easy to install it as a normal library.

giovannipizzi:
ok, I see your point. Fine as it is, then

PR #3254, opened 2019-08-13
Always show Comment mode

CasperWA:
It is only relevant to show what the Comment mode is for imports, when
importing a Comment that already exists in the database (i.e. the UUIDs
are the same).

In order to properly test the comment mode flags for `verdi import`, an
archive with a Comment is needed.


CasperWA:
After discussing this offline with @ltalirz, we have decided to simply leave the print on, no matter if it is relevant or not. This is as to not fill up the repo with files (the archive) used only for a print statement.

ltalirz:
Should the PR be closed then?

CasperWA:
> Should the PR be closed then?

It could, but a tiny part of the code can be removed still (checking for the Comment key) - so I simply updated the PR to remove this. However, there is no particular need, i.e. it could be done in another PR as a "side note", if you want?

PR #3258, opened 2019-08-15
Add support for `CalcJobNode` in `Process.exposed_outputs`

### LINKED ISSUE #3162: Allow `Process.exposed_outputs` to work with `CalcJobNode`s
At present the link type filter is set only for `RETURN`:

https://github.com/aiidateam/aiida-core/blob/d79287c3cb9801ff854e9a24c5fcfe55ab433224/aiida/engine/processes/process.py#L843

Should just be a matter of adding `linktype=(LinkType.RETURN, LinkType.CREATE)`?

sphuber:
Fixes #3162 

The `get_outgoing` call was filtering only for `RETURN` links, thereby
limiting outputs returned by `WorkflowNodes` making the `exposed_ouputs`
call broken for use with `CalcJobNodes`. Simply adding `CREATE` links as
well fixes the problem.

### DIFF
--- a/aiida/engine/processes/process.py
+++ b/aiida/engine/processes/process.py
@@ -828,14 +828,11 @@ class Process(plumpy.Process):
 
         return AttributeDict(exposed_inputs)
 
-    def exposed_outputs(self, process_instance, process_class, namespace=None, agglomerate=True):
-        """
-        Gather the outputs which were exposed from the ``process_class`` and emitted by the specific
-        ``process_instance`` in a dictionary.
+    def exposed_outputs(self, node, process_class, namespace=None, agglomerate=True):
+        """Return the outputs which were exposed from the ``process_class`` and emitted by the specific ``node``
 
-        :param process_instance: Instance of Process class whose outputs to try and retrieve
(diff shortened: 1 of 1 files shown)


aiida/engine/processes/process.py
```diff
@@ -828,14 +828,11 @@ def exposed_inputs(self, process_class, namespace=None, agglomerate=True):
...
-        Gather the outputs which were exposed from the ``process_class`` and emitted by the specific
-        ``process_instance`` in a dictionary.
-
-        :param process_instance: Instance of Process class whose outputs to try and retrieve
-        :type process_instance: :class:`aiida.engine.Process`
+    def exposed_outputs(self, node, process_class, namespace=None, agglomerate=True):
+        """Return the outputs which were exposed from the ``process_class`` and emitted by the specific ``node``
 
+        :param node: process node whose outputs to try and retrieve
+        :type node: :class:`aiida.orm.nodes.process.ProcessNode`
```
CasperWA: I think you need to add another `.process` to this? Also perhaps a tilde to begin with?

CasperWA:
Thanks @sphuber ! :)

PR #3259, opened 2019-08-16
Add `--all` option to `verdi process play/pause`

### LINKED ISSUE #2904: Implement `verdi process play/pause --all`
This is useful if lots of processes have been pauses due to temporary connection issues.

sphuber:
Fixes #2904

For the `play` command, passing this option will play all paused
processes unless specific processes are passed as arguments. Likewise,
for the `pause` command all "active" processes will be paused, meaning
any process with state `CREATED`, `RUNNING` or `WAITING`.

### DIFF
--- a/aiida/cmdline/commands/cmd_process.py
+++ b/aiida/cmdline/commands/cmd_process.py
@@ -170,11 +170,7 @@ def process_status(processes):
 @verdi_process.command('kill')
 @arguments.PROCESSES()
 @options.TIMEOUT()
-@click.option(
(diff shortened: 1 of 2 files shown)


sphuber:
You are correct that the explicit process ids will be ignored if the `--all` flag is specified. I am happy to change the behavior to raise an error if you think that is safer.

ltalirz:
In this case I find it safer to throw an exception, since the input is ambiguous.

sphuber:
Done

Issue #2431, opened 2019-01-31
get job status from scheduler


ltalirz:
Once a job has been dropped from the queue, it would be great if AiiDA would query the scheduler for the job status rather than assuming that the job completed correctly (e.g. to detect whether the walltime was exceeded, the job was cancelled via a scheduler command, ...).

When AiiDA thinks a job is finished, it could simply run something like (add --format=State to just show the state)

```
$ sacct  --parsable --jobs=11566621
JobID|JobName|Partition|Account|AllocCPUS|State|ExitCode|
11566621|aiida-140|normal|s888|192|TIMEOUT|0:0|
11566621.batch|batch||s888|24|CANCELLED|0:15|
11566621.extern|extern||s888|192|COMPLETED|0:0|
11566621.0|pw.x||s888|96|FAILED|1:0|
```

The slurm plugin already has commands that use sacct, e.g. here:
https://github.com/aiidateam/aiida_core/blob/6e9711046753332933f982971db1d7ac7e7ade58/aiida/scheduler/plugins/slurm.py#L225

@sphuber suggest to implement this in the form of a builtin generic scheduler output parser that is called before the actual calculation parser.

> Ideally we would reserved some exit codes for scheduler specific errors, that will be available to the calculation parser, which can then decide what to do, if they want to simply return the scheduler error code or do some parsing before

ltalirz:
See [the mailing list](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/aiidausers/b64jjy1CujA/NQaA8szuDgAJ) for the initial discussion.

@sphuber writes:

> This would require however that this functionality would have to be supported by all the scheduler plugins.

That would of course be ideal but I would say this is not necessary in order to get started.
If a scheduler plugin does not support getting the "job exit status", then we can simply default to the current behavior (which assumes everything went fine).

zhubonan:
If we just want to know if a job has reached the end of execution, a simple solution would be to append a command each as `touch .JOB_DONE` to the job submission script. This way it won't be cluster/scheduler dependent.

ltalirz:
@zhubonan thanks for the suggestion.
To me, the purpose of this is more to profit from the information that schedulers already provide (instead of re-implementing this functionality ourselves).

adegomme:
I would also like this feature, as sometimes jobs end up with timeouts or oom_kill when memory is exhausted, and aiida states that everything was fine, which is not really logical.

sphuber:
[...]

PR #3261, opened 2019-08-16
WIP: Restructured the detailed job info from SLURM

espenfl:
This is work in progress and the first attempt to realize:

- a standardized way to return detailed job info, ideally the keys should be available to all schedulers
- a way to pick the exit code from the detailed job info
- a way to, if the exit code is larger than zero, parse the files of the scheduler in order to give a more sensible exit code to the calculation/workchain

The last point is crucial as a lot of calculations for instance hit memory issues and need to be restarted with different memory and node/cpu configurations. Corrections need to be performed on the calculation plugin as the way to correct might be different based on input and job configuration.

What still remains:

- a parser to analyze the schedulers stdout and stderr (should reside in the `scheduler.plugins.somescheduler`)
- a nice way to return an exit code or information about why the job is listed as `Done`, but still not okey

Some questions:
- how do I get access to the schedulers stdout and stderr in `scheduler.plugins.somescheduler`?
- should we introduce a `Failed` state in addition to the `Done`?


espenfl:
Okey, it seems there is no existing parser for the stdout and stderr of SLURM. Problem is that the exit code is inherited from the bash script return code. Some packages do not return sensible exit codes rendering the inspection of the exit code itself rather useless. However, at least for memory overflow, a message exists and is dumped in stdout/stderr, but there is no record of that in  `sacct` for instance. So it seems we would have to write a simple parser.

Happy to do this, but right now I am a bit unsure how to access the schedulers stdout and stderr from `scheduler.plugins.somescheduler`.

espenfl:
Investigating this a bit further it seems the sensible approach would be to do the following:

- make sure we have an entry under `JobInfo()` that contains the detailed job information (exit codes)
- implement parsers of stdout/stderr in `schedulers.plugins.somescheduler`
- inspect the exit code, if not zero, trigger parser and raise a defined exit code on the AiiDA side

In order to do the last step we need to decide where to put the exit codes. I am a bit tempted to suggest that we implement the same stuff that is now sitting on the calculation and workchain for the scheduler, e.g. putting it on `JobInfo()` or something similar. Then, on the plugin side we can modify those returns by issuing something like `get_scheduler_exit_status` (which would have to be implemented of course). It would just be a shortcut to fetch the exit status in the `JobInfo()` dict.

Is this the way to go? Other suggestions? Would be nice to have something rolling in not too distant of future as this stuff is really hindering production runs.

ltalirz:
Just mentioning this is related to #2431 and @FelixBoelle is interested in this topic as well

sphuber:
Superseded by PR #3906 and #3931

PR #3262, opened 2019-08-20
Properly export using Links flags

### LINKED ISSUE #3295: Default rules for following Links (following edges of the provenance graph) when exporting
Following an offline discussion of issue #1762 these are the rules we have decided when exporting. Note, the Node types in bold symbolize the Node that is currently being exported, i.e., it is this Node onto which the Link in question has been found, from which it should be decided whether or not to export the linked Node.

**LinkType** | **From** | **To** | Follow (default) | Togglable (flag) |
------------ | -------- | ------ | ---------------- | ----------------
INPUT_CALC | **Data** | CalculationNode | False | True (`input_forward`)
INPUT_CALC | Data | **CalculationNode** | True | False
CREATE | **CalculationNode** | Data | True | False
CREATE | CalculationNode | **Data** | True | True (`create_reversed`)
RETURN | **WorkflowNode** | Data | True | False
RETURN | WorkflowNode | **Data** | False | True (`return_reversed`)
INPUT_WORK | **Data** | WorkflowNode | False | True (`input_forward`)
INPUT_WORK | Data | **WorkflowNode** | True | False
CALL_CALC | **WorkflowNode** | CalculationNode | True | False
CALL_CALC | WorkflowNode | **CalculationNode** | False | True (`call_reversed`)
CALL_WORK | **WorkflowNode** | WorkflowNode | True | False
CALL_WORK | WorkflowNode | **WorkflowNode** | False | True (`call_reversed`)

For now, there will be no way to distinguish between toggling to follow certain links (e.g., CALL_CALC and CALL_WORK reversed - both will be toggled on or off by the flag `call_reversed`). If needed, this may be further divided at a later point.

Now, these rules should be implemented in the export logic.

CasperWA:
Closes #3295 

It seemed to me the four Links flags for exporting did not properly work.
Two were not tested, but are now tested.

The code has been restructured slightly and moved to a utility function in an attempt to make the flags' effects more transparent.

Furthermore, from the various discussions (mostly in #3325), the flags are now as follows:

- `input_calc_forward`: Follow INPUT_CALC links forward (from `Data` nodes to `CalculationNode` nodes).
  Default: `False`
- `input_work_forward`: Follow INPUT_WORK links forward (from `Data` nodes to `WorkflowNode` nodes).
  Default: `False`
- `create_backward`: Follow CREATE links backwards (from output `Data` nodes to `CalculationNode` nodes).
  Default: `True`
- `return_backward`: Follow RETURN links backwards (from output `Data` nodes to `WorkflowNode` nodes).
  Default: `False`
- `call_calc_backward`: Follow CALL_CALC links backwards (from called `CalculationNode` nodes to `WorkflowNode` nodes).
  Default: `False`
- `call_work_backward`: Follow CALL_WORK links backwards (from called `WorkflowNode` nodes to `WorkflowNode` nodes).
  Default: `False`

Since Links were collected twice during export, this PR comprises it into a single collection providing both purposes of the original collections (retrieve linked Nodes and produce list of Links for `data.json`).

Also, I am adding a bit of helpful information to the top-level `__init__.py` file in the `importexport` module, guiding people to the commit in `aiida-core` just before the creation of the `importexport` module. This should make it easier to find the git blame/origins of the code related to import/export; both the functionality and the tests.

Lastly, the export of Node Attributes and Extras have been collected within a single QueryBuilder query, instead of having two consecutive queries.

### DIFF
--- a/aiida/backends/tests/tools/importexport/orm/test_calculations.py
+++ b/aiida/backends/tests/tools/importexport/orm/test_calculations.py
@@ -58,7 +58,7 @@ class TestCalculations(AiidaTestCase):
         not_wanted_uuids = [v.uuid for v in (b, c, d)]
         # At this point we export the generated data
         filename1 = os.path.join(temp_dir, 'export1.tar.gz')
-        export([res], outfile=filename1, silent=True, return_reversed=True)
+        export([res], outfile=filename1, silent=True, return_backward=True)
         self.clean_db()
         self.insert_data()
         import_data(filename1, silent=True)
--- a/aiida/backends/tests/tools/importexport/orm/test_links.py
+++ b/aiida/backends/tests/tools/importexport/orm/test_links.py
@@ -122,8 +122,8 @@ class TestLinks(AiidaTestCase):
         Graph::
 
             data1 ---------------INPUT_WORK----------------+
+              |                                            |
               |     data2 -INPUT_WORK-+                    |
-              |                       |                    |
               |                       V                    V
               +-------INPUT_WORK--> work1 --CALL_WORK--> work2 ----+
               |                                            |       |
@@ -374,9 +374,14 @@ class TestLinks(AiidaTestCase):
     def prepare_link_flags_export(nodes_to_export, test_data):
         """Helper function"""
         from aiida.tools.importexport.common.config import LINK_FLAGS
-        rules = LINK_FLAGS.copy()
-        for rule in rules:
-            rules[rule] = False
+        rules = {}
+        togglable_link_rules = {
+            'input_calc_forward', 'create_backward', 'return_backward', 'input_work_forward', 'call_calc_backward',
+            'call_work_backward'
+        }
+        for rule in LINK_FLAGS:
+            if rule in togglable_link_rules:
+                rules[rule] = LINK_FLAGS[rule]
 
         for export_file, rule_changes, expected_nodes in test_data.values():
             rules.update(rule_changes)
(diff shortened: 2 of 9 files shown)


CasperWA:
Note: Due to the offline discussion of issue #1762, this PR needs to be re-checked for the rules and their implementation throughout the import/export.

CasperWA:
This is ready to be reviewed.

The latest commit updates the functionality of how Links can be followed in the provenance graph, when collecting the Nodes to export.
All twelve flags from the discussion of this subject (see, e.g., #3325 or #3295) have been implemented in the export functionality and can essentially be used. However, if the flag is not set to be togglable it will raise, stating so, unless one is on a test profile.
The flag defaults, as well as which are togglable and which aren't have all been condensed in a single variable in `aiida.tools.importexport.common.config.LINK_FLAGS`.

It should be noted that in my attempt to create meaningful tests, I have removed some existing tests for the old flags and created a new "beast". Everything is in `aiida.backends.tests.tools.importexport.orm.test_links`.
The main idea of the test(s) is:
1. have a single "complex" graph (here the existing method `construct_complex_graph()` has been used)
2. set all flags to `False` and then match related flags in the same direction together, depending on if the focus is on the data or logic provenance (e.g., `INPUT_CALC_forward` and `INPUT_WORK_forward`)
3. mix and match them in their four different configurations (i.e., (`False`, `False`), (`True`, False`), (`False`, `True`), (`True`, `True`))
4. generate export archives using these rules, choosing reasonable Nodes to export and what the expected exported Nodes will be,
5. repeat steps 2.-4.,
6. and then finally make sure it's all correct/as expected (this is where the actual tests/assertions are performed).

The tests thereby make sure all flags work properly, however, it does _not_ test the recursivity. For this, we rely on the existing tests, e.g., the ones using the same `construct_complex_graph()` method with the default rules for how to follow Links.

CasperWA:
Updated according to discussion in #3325 (from [this](https://github.com/aiidateam/aiida-core/pull/3325#issuecomment-534949094) point onwards, circa).

The togglability of the rules has been hardcoded into the utility function that retrieves Nodes to be exported, but most importantly, the tests no longer check scenarios that are unreachable for users.

CasperWA:
@sphuber As a note, I initially creaed this PR with two separate commits in mind, although it is not strictly needed, they were just different enough for it to make sense.
The first commit is still the first commit here. The next would now consist of the remaining commits really.
If you think it is better with two commits I can rebase it, otherwise I will leave it for you to squash and merge eventually.

sphuber:
> As a note, I initially creaed this PR with two separate commits in mind,

Apologies, I only read this message now after having squash merged. Hope that is ok. Will read more carefully in the future

Issue #3283, opened 2019-08-28
verdi data bands show broken


ltalirz:
This is on the latest develop branch
```
$ verdi data bands show 247
Traceback (most recent call last):
  File "/home/max/.virtualenvs/aiida/bin/verdi", line 10, in <module>
    sys.exit(verdi())
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/max/codes/aiida_core/aiida/cmdline/utils/decorators.py", line 68, in wrapper
    return wrapped(*args, **kwargs)
  File "/home/max/codes/aiida_core/aiida/cmdline/commands/cmd_data/cmd_bands.py", line 99, in bands_show
    show_function(fmt, data)
  File "/home/max/codes/aiida_core/aiida/cmdline/commands/cmd_data/cmd_show.py", line 228, in _show_xmgrace
    tempf.write(text)
  File "/home/max/.virtualenvs/aiida/lib/python3.6/tempfile.py", line 624, in func_wrapper
    return func(*args, **kwargs)
TypeError: write() argument must be str, not bytes
```

One could fix this issue by switching from `w+` to `w+b` here
[...]

PR #3284, opened 2019-08-28
Release `v1.0.0b6`

sphuber:


ltalirz:
Thanks a lot @sphuber !!
Perhaps wait until we know how to proceed with the BandsData issue?
https://github.com/aiidateam/aiida-core/issues/3283

I could fix this if you like...

sphuber:
I am fine with holding on, but given that this is not the only problem, but is a more prolific problem across the data export/show implementation, how long do you want to wait? Do you need the bands export specifically for the tutorial?

ltalirz:
> Do you need the bands export specifically for the tutorial?

I would have used the export, yes, but not necessarily `show`.

If the solution is indeed going to be larger, perhaps you are right and we should not hold off. 
Feel free to just go ahead.

sphuber:
Merged the PRs that we wanted to merge and added the corresponding lines to `README.md`. This should be good to go

CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
+## v1.0.0b6
+
+### Deprecations
+- To be removed in `v2.0.0`: the `mode` and `encoding` arguments of the `put_object_from_file` of the `Node` class and `Repository` utility mixin [[#3289]](https://github.com/aiidateam/aiida-core/pull/3289)
+
+### Minor bug fixes
+- Allow to construct `SinglefileData` nodes from binary files [[#3285]](https://github.com/aiidateam/aiida-core/pull/3285)
+- Fix eagerness of `verdi setup --help` being overruled [[#3257]](https://github.com/aiidateam/aiida-core/pull/3257)
```
ltalirz:
```suggestion
- Fix eagerness of `verdi setup/quicksetup --help` being overruled [[#3257]](https://github.com/aiidateam/aiida-core/pull/3257)
```

PR #3285, opened 2019-08-29
Allow to construct `SinglefileData` nodes from binary files

### LINKED ISSUE #3237: Impossible to store binary file as `SinglefileData`
As of aiida v1.0.0beta, it became impossible to store binary files as `SinglefileData`: it seems that AiiDA tries to convert it to Unicode and fails (it worked nicely in pre-beta versions). Also, `SinglefileData.put_object_from_file` does not allow changing file open mode to `wb` as needed.

Steps to reproduce:
1. Create binary file:
`head /dev/random > ~/file.dat `
2. Try to make a `SinglefileData` node from it:
```
from os.path import expanduser
from aiida.orm import SinglefileData
node = SinglefileData(expanduser('~/file.dat'))
```
The error:
```
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-2-1a6822a9910c> in <module>()
----> 1 SinglefileData('/home/aiida/code/file.txt')

/home/aiida/code/aiida_core/aiida/orm/nodes/data/singlefile.pyc in __init__(self, file, **kwargs)
     35         super(SinglefileData, self).__init__(**kwargs)
     36         if file is not None:
---> 37             self.set_file(file)
     38 
     39     @property

/home/aiida/code/aiida_core/aiida/orm/nodes/data/singlefile.pyc in set_file(self, file)
    100             self.put_object_from_filelike(file, key, mode='wb')
    101         else:
--> 102             self.put_object_from_file(file, key)
    103 
    104         # Delete any other existing objects (minus the current `key` which was already removed from the list)

/home/aiida/code/aiida_core/aiida/orm/nodes/node.pyc in put_object_from_file(self, path, key, mode, encoding, force)
    678         :raises aiida.common.ModificationNotAllowed: if repository is immutable and `force=False`
    679         """
--> 680         self._repository.put_object_from_file(path, key, mode, encoding, force)
    681 
    682     def put_object_from_filelike(self, handle, key, mode='w', encoding='utf8', force=False):

/home/aiida/code/aiida_core/aiida/orm/utils/repository.pyc in put_object_from_file(self, path, key, mode, encoding, force)
    186 
    187         with io.open(path) as handle:
--> 188             self.put_object_from_filelike(handle, key, mode=mode, encoding=encoding)
    189 
    190     def put_object_from_filelike(self, handle, key, mode='w', encoding='utf8', force=False):

/home/aiida/code/aiida_core/aiida/orm/utils/repository.pyc in put_object_from_filelike(self, handle, key, mode, encoding, force)
    212             folder = folder.get_subfolder(basepath, create=True)
    213 
--> 214         folder.create_file_from_filelike(handle, key, mode=mode, encoding=encoding)
    215 
    216     def delete_object(self, key, force=False):

/home/aiida/code/aiida_core/aiida/common/folders.pyc in create_file_from_filelike(self, filelike, filename, mode, encoding)
    248                 except (UnicodeDecodeError, UnicodeEncodeError):
    249                     filelike.seek(0)
--> 250                     shutil.copyfileobj(filelike, handle)
    251             else:
    252                 shutil.copyfileobj(filelike, handle)

/usr/lib/python2.7/shutil.pyc in copyfileobj(fsrc, fdst, length)
     61     """copy data from file-like object fsrc to file-like object fdst"""
     62     while 1:
---> 63         buf = fsrc.read(length)
     64         if not buf:
     65             break

/usr/lib/python2.7/codecs.pyc in decode(self, input, final)
    312         # decode input (taking the buffer into account)
    313         data = self.buffer + input
--> 314         (result, consumed) = self._buffer_decode(data, self.errors, final)
    315         # keep undecoded input until the next call
    316         self.buffer = data[consumed:]

UnicodeDecodeError: 'utf8' codec can't decode byte 0xc1 in position 0: invalid start byte
```

sphuber:
Fixes #3237 

The `SinglefileData.set_file` method was opening the target file in
non-binary mode with default utf-8 encoding with no way for the user to
override it through the constructor. This would fail when a binary file
was passed. Since the purpose of this operation is just to copy the file
wholesale to the repository this should simply be done in binary mode to
copy the bytes over.

### DIFF
--- a/aiida/backends/tests/orm/data/test_singlefile.py
+++ b/aiida/backends/tests/orm/data/test_singlefile.py
@@ -107,3 +107,29 @@ class TestSinglefileData(AiidaTestCase):
 
         self.assertEqual(content_stored, content_original)
         self.assertEqual(node.list_object_names(), [SinglefileData.DEFAULT_FILENAME])
+
+    def test_binary_file(self):
+        """Test that the constructor accepts binary files."""
+        byte_array = [120, 3, 255, 0, 100]
+        content_binary = bytearray(byte_array)
+
+        with tempfile.NamedTemporaryFile(mode='wb+') as handle:
+            basename = os.path.basename(handle.name)
+            handle.write(bytearray(content_binary))
+            handle.flush()
+            handle.seek(0)
+            node = SinglefileData(handle.name)
+
+        with node.open(mode='rb') as handle:
+            content_stored = handle.read()
+
+        self.assertEqual(content_stored, content_binary)
+        self.assertEqual(node.list_object_names(), [basename])
+
+        node.store()
+
+        with node.open(mode='rb') as handle:
+            content_stored = handle.read()
+
+        self.assertEqual(content_stored, content_binary)
+        self.assertEqual(node.list_object_names(), [basename])
--- a/aiida/orm/nodes/data/singlefile.py
+++ b/aiida/orm/nodes/data/singlefile.py
@@ -99,7 +99,7 @@ class SinglefileData(Data):
         if is_filelike:
             self.put_object_from_filelike(file, key, mode='wb')
         else:
-            self.put_object_from_file(file, key)
+            self.put_object_from_file(file, key, mode='wb')
 
         # Delete any other existing objects (minus the current `key` which was already removed from the list)
         for existing_key in existing_object_names:
--- a/aiida/orm/utils/repository.py
+++ b/aiida/orm/utils/repository.py
@@ -179,13 +179,14 @@ class Repository(object):
         :param force: boolean, if True, will skip the mutability check
         :raises aiida.common.ModificationNotAllowed: if repository is immutable and `force=False`
         """
+        # pylint: disable=unused-argument
         if not force:
             self.validate_mutability()
 
         self.validate_object_key(key)
 
-        with io.open(path) as handle:
-            self.put_object_from_filelike(handle, key, mode=mode, encoding=encoding)
+        with io.open(path, mode='rb') as handle:
+            self.put_object_from_filelike(handle, key, mode='wb')
 
     def put_object_from_filelike(self, handle, key, mode='w', encoding='utf8', force=False):
         """Store a new object under `key` with contents of filelike object `handle`.


sphuber:
This PR fixes the referenced issue but this really has shown a problem in the design of the repository interface. I am starting to think strongly that the `put_object_*` methods should not accept a `mode` and `encoding` and should simply always open a binary handle and transfer the bytes. The `put_object_from_file` simply calls through to `put_object_from_filelike`. The latter will simply fail if the source handle given as argument is not a binary one. We can optionally catch the exception to be explicit that only binary handles are supported. This way the responsibility is passed to the caller and the callee should not have to try and guess about the nature of the filelike object. The added responsibility for the user might be problematic in terms of user-experience.

I would propose then to remove the `mode` and `encoding` arguments from the `put_object_*` methods and those that call it going up the stack. They were only added for `aiida-core==1.0.0b1` but this means still that we have to go through the deprecation route, I am afraid. I am not sure if we can be sure that any plugins used these arguments and/or depended on the default. I have already tested what changes would have to be applied when the interface is updated in this way and can be found in [this branch](https://github.com/sphuber/aiida_core/commits/fix_3237_repository_mode_encoding). Note that here I deleted instead of deprecated, but it is just to see the affect on the rest of the code.

@giovannipizzi @ltalirz comments/ideas?

ltalirz:
so, I definitely agree that when passing a file path to `set_file`, we should be using binary mode.

As for deprecating the encoding arguments for the `put_...` methods: again for `put_object_from_file` I would tend to agree that they are not needed (?).

For `put_object_from_filelike`, however, I'm not sure. What if what you have is simply a non-binary file handle?

sphuber:
> For `put_object_from_filelike`, however, I'm not sure. What if what you have is simply a non-binary file handle?

In principle it should always be possible to turn that into a normal handle. But it might be inconvenient yes. Would maybe turning the defaults into `mode='wb'` and `encoding=None` be a better solution perhaps? Of course this then still puts the burden on a user to understand how all this works and that if they pass a normal handle they _have_ to at least change the `mode` of the command will fail. My worry is just that users will think this is a bug in AiiDA, but there is nothing we can or should do in that case, I guess.

ltalirz:
> Would maybe turning the defaults into mode='wb' and encoding=None be a better solution perhaps?

For `put_object_from_filelike` this looks like the better solution to me.

> My worry is just that users will think this is a bug in AiiDA, but there is nothing we can or should do in that case, I guess.

If that is your worry, you can try to be more clever - you could simply detect the mode of the handle:
```
In [1]: handle = open('a', 'rb')

In [2]: handle.mode
Out[2]: 'rb'
```
and provide a helpful error message.

ltalirz:
It's not unusual to have a default mode (`open` has this) but I guess it is unusual to have binary as the default mode.
So, perhaps it is good to help users along a bit in this case.

giovannipizzi:
@ltalirz as @sphuber pointed out to me yesterday, the mode only works for opened files, not for e.g. StringIO or BytesIO (that BTW have quite different behaviour in python 2 and 3).

ltalirz:
I see, I wasn't aware of that.

So, in the end the method you use is `shutil.copyfileobj`, which will do the basic test:
```
In [8]: h=StringIO("abc")

In [20]: f=open("test","wb")

In [22]: shutil.copyfileobj(h, f)
<ipython-input-22-d0fff4447d1e> in <module>()
----> 1 shutil.copyfileobj(h, f)

/Users/leopold/Applications/miniconda3/lib/python3.6/shutil.py in copyfileobj(fsrc, fdst, length)
     80         if not buf:
     81             break
---> 82         fdst.write(buf)
     83
     84 def _samefile(src, dst):

TypeError: a bytes-like object is required, not 'str'

In [24]: f2=open("test","w")

In [25]: shutil.copyfileobj(h, f2)
```
I.e. you can either do the same test they are doing earlier (and warn about it) or you can just leave it at the error message that it will throw.

sphuber:
Ok, I think the best thing to do then is to remove (deprecate) `mode` and `encoding` from `put_object_from_file` and for `put_object_from_filelike` keep the current defaults. Agreed?

If so, I will do this in a separate PR with separate issue for reference and we can merge this PR. If you both agree I will do those things first before making the release.

ltalirz:
> Ok, I think the best thing to do then is to remove (deprecate) mode and encoding from put_object_from_file and for put_object_from_filelike keep the current defaults. Agreed?

Agreed (deprecate).

> If so, I will do this in a separate PR with separate issue for reference and we can merge this PR. If you both agree I will do those things first before making the release.

Sounds good!

PR #3289, opened 2019-08-29
Deprecate `mode` and `encoding` of `Repository.put_object_from_file`

### LINKED ISSUE #3286: Deprecate `mode` and `encoding` in `put_object_from_file`
When copying over a file to the repository it should be done in binary mode without decoding/encoding the content in between.

sphuber:
Fixes #3286 

When copying a file to the repository from a filepath, it does not make
sense to be able to specify the `mode` and `encoding` in which the
target file handle should be opened. It should simply copy over the
bytes and therefore always open in binary mode.

To be able to detect if the arguments are being used, we have to change
the defaults to `None` in order to be able to emit the deprecation
warning. This is justified because despite changing the defaults the
intended side-effect of the method, copying over a file, will be
unchanged in its behavior.

### DIFF
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -665,7 +665,7 @@ class Node(Entity):
         """
         self._repository.put_object_from_tree(path, key, contents_only, force)
 
-    def put_object_from_file(self, path, key, mode='w', encoding='utf8', force=False):
+    def put_object_from_file(self, path, key, mode=None, encoding=None, force=False):
         """Store a new object under `key` with contents of the file located at `path` on this file system.
 
         .. warning:: If the repository belongs to a stored node, a `ModificationNotAllowed` exception will be raised.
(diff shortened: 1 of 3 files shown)


aiida/orm/nodes/node.py
```diff
@@ -674,10 +674,28 @@ def put_object_from_file(self, path, key, mode='w', encoding='utf8', force=False
         :param path: absolute path of file whose contents to copy to the repository
         :param key: fully qualified identifier for the object within the repository
         :param mode: the file mode with which the object will be written
+            Deprecated: will be removed in `v1.1.0`
```
ltalirz: hehe, how was this thing with increasing the **major** version number for api changes? ;-)

aiida/orm/utils/repository.py
```diff
@@ -175,18 +175,35 @@ def put_object_from_file(self, path, key, mode='w', encoding='utf8', force=False
...
+            )
+
         if not force:
             self.validate_mutability()
 
         self.validate_object_key(key)
 
         with io.open(path, mode='rb') as handle:
-            self.put_object_from_filelike(handle, key, mode='wb')
+            self.put_object_from_filelike(handle, key, mode='wb', encoding=None)
```
ltalirz: Just out of curiosity: why add this encoding argument here?

aiida/orm/nodes/node.py
sphuber: Honestly, it wasn't me :) I put `v2.0.0` but @giovannipizzi strongly suggested to put `v1.1.0`. He will be my witness

aiida/orm/utils/repository.py
sphuber: because the default encoding is `utf-8` which does not make sense and if the lower lying code does the wrong thing, it will crash. It so happens that there is a line in `Folder.create_from_filelike` that sets `encoding` to `None` if `b` is in the mode, but again, it is better to do this here.

aiida/orm/nodes/node.py
ltalirz: So @giovannipizzi , what is the reasoning here?
We say 1.1 means "only minor little API changes"?
What is minor and what is major?

giovannipizzi: The reasoning is that if, in 1.0.0, something is already deprecated, people shouldn't use it (and if they have been using it in a beta, well, they know it's a beta and things can still change - we still guarantee that when we release 1.0 things work, and they have time to adapt before 1.1).

If we deprecated something in 1.0.1 or later instead, I think it has to be changed in 2.0.

ltalirz: I remember that we said the API would be frozen from v1.0.0b1 onwards unless we discover real problems.
This was communicated to plugin developers as a reson why it makes sense to invest the time for the migration now, and is mentioned in the roadmap https://github.com/aiidateam/aiida-core/wiki/AiiDA-release-roadmap

I don't think this change here qualifies as a big problem.
What is the issue with leaving the parameters (deprecated) until 2.0.0?

I know this is a rather small issue here, but I think this is a matter of principle.
In my view, if a plugin breaks in aiida 1.1 that worked with aiida 1.0.0b1 it is our fault, not the fault of the developer.

giovannipizzi: I don't agree, but since it's not critical it's ok to put 2.0 here

PR #3299, opened 2019-09-05
Enabled extracting Sphinx documentation from CalcJob

espenfl:
Due to the similarities between `CalcJob` and `WorkChain` I think such an extension makes sense. Tried to refactor as much as possible, but some things was not so easy to do. This could possibly be improved, but since we are probably not going to generate many more of these, maybe it is okey as is.

There is now a warning, `WARNING: while setting up extension aiida.sphinxext: directive 'autoworkchain' is already registered, it will be overridden` when you compile with Sphinx.

Which happens due to the two calls to `setup_aiida_*`. I do not yet know how to remove that, but it does not seem to void functionality.

### DIFF
--- a/aiida/sphinxext/__init__.py
+++ b/aiida/sphinxext/__init__.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 
 import aiida
 from . import workchain
+from . import calcjob
 
 
 def setup(app):
@@ -24,5 +25,6 @@ def setup(app):
     Setup function to add the extension classes / nodes to Sphinx.
     """
     workchain.setup_aiida_workchain(app)
+    calcjob.setup_aiida_calcjob(app)
 
     return {'version': aiida.__version__, 'parallel_read_safe': True}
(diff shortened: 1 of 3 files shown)


ltalirz:
@espenfl I started working on something similar in July but somehow didn't find the time to push it. Will look into this

espenfl:
Thanks. What is the current status of this?

ltalirz:
working on it (amongst other things). I was looking into addressing some of the points you mentioned in #3300 but no luck yet.
If you like, we can merge it as-is without the "readability fixes" so that at least the basic functionality is there.

espenfl:
Thanks. Great. Yes, that would be nice. Maybe we also could incorporate the solution to point 1 in #3300? That is at least better than nothing (variables displayed in red). For the remaining points, we should maybe merge these issues into one.

ltalirz:
Closing this as this was addressed in #3314 .
Apologies again @espenfl for the delay.

PR #3305, opened 2019-09-10
remove alterations of sys.path

### LINKED ISSUE #3293: verdi shell imports deprecated module
The following happens in the tutorial VM:

```
$ verdi shell
/home/max/codes/aiida-core/aiida/backends/profile.py:23: AiidaDeprecationWarning: this module is deprecated
  warnings.warn('this module is deprecated', AiidaDeprecationWarning)  # pylint: disable=no-member
Python 3.6.8 (default, Jan 14 2019, 11:02:34)
Type "copyright", "credits" or "license" for more information.

IPython 5.8.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.
```

I really wonder why... I've searched throughout the codebase and don't really understand what triggers the import of `aiida.backends.profile`.

Note: This does not happen with other `verdi` commands that use the database, e.g. `verdi process list`...

ltalirz:
fix #3293 

python allows to modify the search path for modules programmatically via the `sys.path` variable (`sys.path` is initialized from the `PYTHONPATH` environment variable).
Doing so, however, can easily have unintended consequences (one of them surfaced in #3293) - in particular, when new paths are inserted at the *beginning* of `sys.path`.

AiiDA was manipulating `sys.path` in a number of places:

1. https://github.com/aiidateam/aiida-core/blob/a500d2b2801efa393ce6f567dc9893a62696b8f1/aiida/cmdline/commands/cmd_run.py#L113

1. https://github.com/aiidateam/aiida-core/blob/d750d513f7adf5126d43a230c9fb378de02f0772/aiida/sphinxext/tests/workchain_source/conf.py#L32

1. https://github.com/aiidateam/aiida-core/blob/cc94a1e0127cd9cab0a4049e69a9fa31cb82941c/utils/validate_consistency.py#L190

1. https://github.com/aiidateam/aiida-core/blob/cc94a1e0127cd9cab0a4049e69a9fa31cb82941c/aiida/common/files.py#L194

1. https://github.com/aiidateam/aiida-core/blob/91b205bcdeebc3d6f9a019c77c38ffdf6fc95321/aiida/backends/djsite/settings.py#L31

This PR removes/remedies those manipulations (except for point 4, which is part of `shutil.which` and is relevant only for `sys.platform == 'win32'`).


aiida/cmdline/commands/cmd_run.py
```diff
@@ -109,8 +108,6 @@ def run(scriptname, varargs, group, group_name, exclude, excludesubclasses, incl
             # Must add also argv[0]
             new_argv = [scriptname] + list(varargs)
             with update_environment(new_argv=new_argv):
-                # Add local folder to sys.path
-                sys.path.insert(0, os.path.abspath(os.curdir))
```
ltalirz: Note: This adds the current working directory to the path - however this is done by python automatically (I checked sys.path before and after this line to check)

aiida/backends/djsite/settings.py
```diff
@@ -13,23 +13,13 @@
...
-import os
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.dialects.postgresql import UUID
 
 from aiida.common import exceptions
 from aiida.common.timezone import get_current_timezone
 from aiida.manage.configuration import get_profile, settings
 
-# Assumes that parent directory of aiida is root for
-# things like templates/, SQL/ etc.  If not, change what follows...
```
ltalirz: Not sure what exactly this comment means / why this should be needed.
Anyhow, it seems the code works also without this

aiida/cmdline/commands/cmd_run.py
CasperWA: For both version of Python?

ltalirz: I didn't check for both versions but I'm pretty sure this is the case almost since the beginning of python.

CasperWA: Fair enough, I am simply spitballing, trying to come up with a reason why it may have been added to begin with - and it seemed a reasonable explanation.

aiida/sphinxext/tests/conftest.py
```diff
@@ -7,47 +7,58 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
+"""Pytest fixtures for AiiDA sphinx extension tests."""
 from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
 import os
+from os import path
```
CasperWA: Why import both `os` and `path` from `os` separately?

aiida/sphinxext/tests/conftest.py
```diff
@@ -7,47 +7,58 @@
...
 @pytest.fixture
 def build_dir():
+    """Create directory to build documentation."""
     # Python 2 doesn't have tempfile.TemporaryDirectory
     dirname = tempfile.mkdtemp()
     try:
         yield dirname
-    except Exception as e:
-        raise e
+    except Exception as exc:
```
CasperWA: Is this a standard way to do this, simply raising the exception if/when it is already raised? Why not skip the `try-except` and simply have the `yield` and then the `shutil` line?

aiida/sphinxext/tests/test_workchain.py
```diff
@@ -15,12 +15,16 @@
 from os.path import join, dirname
 
 import six
-import pytest
+import pytest  # pylint: disable=unused-import
```
CasperWA: If it is unused, I guess simply remove it?

utils/validate_consistency.py
```diff
@@ -187,9 +187,11 @@ def validate_version():
     they match.
     """
     # Get version from python package
-    sys.path.insert(0, ROOT_DIR)
+    sys.path.insert(1, ROOT_DIR)
     import aiida  # pylint: disable=wrong-import-position
     version = aiida.__version__
+    # Clean up sys.path again
```
CasperWA: This seems weird, but I guess this is the main reason why this PR is relevant, I presume? To make sure `sys.path` is properly cleaned up from AiiDA side?

aiida/sphinxext/tests/conftest.py
ltalirz: to make things shorter for `path` commands?
I just fixed a linter complaint here...

aiida/sphinxext/tests/conftest.py
ltalirz: yeah, the try/except is kind of pointless here.

aiida/sphinxext/tests/test_workchain.py
ltalirz: hm, I wasn't sure whether it is unused in the sense that there's all this magic needed to inject the fixtures.
I'll try removing it.

utils/validate_consistency.py
ltalirz: What is weird about this? It's just adding a path to import a module (which we need just for the aiida version) and then cleaning up the path again so that things aren't messed up further down the line.

aiida/sphinxext/tests/conftest.py
ltalirz: other functions from os are used.
I think it's fine.

utils/validate_consistency.py
ltalirz: Feel free to suggest a better approach

ltalirz:
@giovannipizzi Casper is done with his review - could you please give a quick look as well?

PR #3314, opened 2019-09-12
generalize workchain sphinx extension to processes

### LINKED ISSUE #3300: Documenting WorkChain and CalcJob plugins
I notice that upon including https://aiida.readthedocs.io/projects/aiida-core/en/latest/developer_guide/aiida_sphinxext.html for workchains and similar for calculations (after #3299) that it might be convenient to try to change the formatting slightly. Here is my suggestions:

1. Since we use a lot of inline literals for variables etc. it would be nice if we could get the `spec.inputs` and `spec.outputs` names to be rendered as inline literals. Now they are rendered with `lineral_strong` function from the `addnode` from Sphinx. I have tried to see it it is possible to render reStructuredText in any of the `docutils.nodes` or `sphinx.addnodes` but gave up due to lack of examples that I considered to be relevant. If anyone could chime in on this, that would be great.

2. Similarly, I would like to have the data type following the name to be projected as `:py:class:'aiida.orm.nodes.data.Dict` or similar. Then they are clickable etc. and fits with the documentation otherwise. Again, tried to search for solutions, but did not find any similar requests.

3. It seems when using `expose_inputs` etc. we pull along everything from the bottom, meaning that for instance `metadata` is shown for all `WorkChain`. I think ideally we would like to offer some granularity to this as we would like to restrict what is projected to the users.

4. Also, when the docstring is taken from the class definition, reStructuredText is not supported, making it a bit difficult to support a good API documentation and at the same time leveraging the functionality to get the workchain and calculation documentation into a more general documentation on the workchains and calculation functionality.

Comments, suggestions, limitations etc?

ltalirz:
Fixes #3300 

This PR 

 * generalizes the "workchain" sphinx autodoc extension
 * uses it to replace documentation of the `CalcJob` interface

I.e. you can now do:

```rst
.. aiida-process:: CalcJob
    :module: aiida.engine.processes.calcjobs

.. aiida-calcjob:: CalcJob
    :module: aiida.engine.processes.calcjobs

.. aiida-workchain:: WorkChain
    :module: aiida.engine.processes.workchains
```

Currently, the implementation of the `aiida-workchain` and `aiida-calcjob` directives are identical to `aiida-process` (simply inherit from the class) but they can easily be customized in the future.

This PR includes some modifications introduced by #3299


aiida/sphinxext/tests/workchain_source/index.rst
```diff
@@ -11,13 +11,13 @@ This is a demo documentation to show off the features of the ``sphinx-aiida`` ex
     :module: demo_workchain
 
 
-You can use the ``:hide-unstored-inputs:`` option to not show the inputs which are not stored in the DB:
+If you don't want to show only the inputs that are stored in the database, use the ``:hide-unstored-inputs:`` option.
```
sphuber:
```suggestion
If you want to hide the inputs that are not stored as nodes in the database, use the ``:hide-unstored-inputs:`` option.
```

ltalirz:
@sphuber ready for review now :-)

sphuber:
Thanks a lot @ltalirz

ltalirz:
Hm... while [the docs build on travis seems to work fine](https://travis-ci.org/aiidateam/aiida-core/jobs/591654958), it seems this [broke the docs build on readthedocs](https://readthedocs.org/projects/aiida-core/builds/9736015/).

Espen mentioned something like this as well but I may have misunderstood him before.
We'll need to look into this.

d-tomerini:
Hi @italirz,

as I mentioned, just a typo at line 468 of your new version (dimnensional) in the page.

For the ``mpirun_extra_params`` I'd specify that the *mpi run command*  is defined in the :ref:`computer setup<computer_setup>` section, for anyone interested, for the ``resources`` section I'd refer to the scheduler manual page to access the options. For the rest, the help does its job.

Also, I'd clearly states always when the value to be supplied is not a string (Boolean or AiiDA object)

sphuber:
> Also, I'd clearly states always when the value to be supplied is not a string (Boolean or AiiDA object)

For that we should use the `valid_type` keyword in the spec. If this is currently not used for some `CalcJob` options, we should add them. This keyword is automatically parsed by the Sphinx extenstion

Issue #868, opened 2017-10-27
verdi code setup does not check the remote absolute path


sphuber:
When you setup a code through `verdi code setup` and (accidentally) provide an incorrect path to the remote binary, the command will not complain and successfully create the code. If you then try to use the code for a calculation the calculation will fail. You can now also no longer delete the code or correct it, as it now has an associated calculation and so you are forced to create a new code node, choosing a different label, because the original is now already occupied.

Since when specifying the remote absolute path, the computer is already selected, we should be able to verify whether the given path even exists and put out a warning otherwise. I don't see any impossibilities or blocking things to this approach but maybe I am missing something

ltalirz:
I guess it should still be possible to set up codes if you don't have an internet connection, i.e. some logic is needed as well.

ltalirz:
@sphuber I think you mentioned in #3319 that you had second thoughts about this feature request - perhaps it should be limited to setting up local codes?

Also, it would create a bit of a discrepancy in the sense that you *can* have local codes in your database (imported from somewhere) that don't correspond to an actual binary, while you can't set such a code up.

I guess the right way forward would be to do the check by default (even for codes on remote computers) but offer a way to disable it (e.g. via an interactive prompt that defaults to "proceed").

sphuber:
I definitely have second thoughts about doing this for remote codes, because adding the complexity with the transport is not worth what we will gain I think. If that transport is `local` though, so "remote code on localhost" it might still be useful. My original request came from the situation where I configured a code on localhost and accidentally mistyped the path. The code was then unusable and since we basically also do not provide a way to edit a code, I basically had to delete and recreate it. At the time there was not even a simple way of deleting it I think. So my thinking then was that it should be nice to prevent this annoyance by checking the path.

So maybe we should just check for local codes and remote code on localhosts?

ltalirz:
> So maybe we should just check for local codes and remote code on localhosts?

Yes, I think that is a good next step (and, as mentioned above, this check should happen during creation of the Code only - it's ok to have (imported) codes in the DB that point to non-existing paths).

bosonie:
I'm randomly getting the chance to read this....and it's been a while that I think that would be great to have a command `verdi code configure` (similar to the computer one). The scope would not only be to test the code executable, but ideally could be extended by plugin developers in order to parse the version of the code for instance. To have an `extra` attached to the code and reporting the version would make the `parser` development much easier.
Just a dream though, I guess it's very difficult to implement.

[...]

PR #3319, opened 2019-09-12
reorganize plugin testing code

### LINKED ISSUE #2648: faster plugin tests
It is now convenient to run just `pytest` in the test folder using the fixture defined in `aiida.manage.fixtures.fixture_manager`. This creates a complete separate aiida profile and database. However, this process is relatively slow (~20 seconds) compared to the old `verdi -p <test_profile> devel tests` (no delay) method. This is becuase the postgres cluster needs to be created every time be migrated afterwards. 

It would be better to allow this `fixture_manager` to reuse the existing test_profile to speed things up. During plugin development sometimes there can be just a small change made and it is desirable to get the test results faster.

### LINKED ISSUE #2660: Make pytest fixtures publically available
Similar to [sphinx.testing.fixtures](https://github.com/sphinx-doc/sphinx/blob/2.0/sphinx/testing/fixtures.py). Plugin developers could then add:  `pytest_plugins = "aiida.testing.fixtures"` to their `conftest.py` and automatically inherit the fixtures.
They could then be maintained centrally, and potential improvements such as #2648 would be automatically propagated to plugins.

@zhubonan

### LINKED ISSUE #3462: resetting default User cache
Since the default user of an AiiDA profile is requested a lot, we've introduced a cache for this property:

https://github.com/aiidateam/aiida-core/blob/cc94a1e0127cd9cab0a4049e69a9fa31cb82941c/aiida/orm/users.py#L51-L70

However, short of resetting the manager, there currently doesn't seem to be a way to reset this cache, as one might want to do, for example, after setting a new default user programmatically, deleting the default user, etc.

Resetting this cache could be done at two levels:
 1. A specific command for the `User.objects` collection that resets its `_default_user` property
 2. A generic `reset()` method for `objects` - `objects` are loaded lazily and so you can add a generic method 
    ```python
        @classmethod
        def reset(cls):
            cls._COLLECTIONS = datastructures.LazyStore()
    ```
     to `aiida.orm.entities.Collection`
     This would allow to do `User.objects.reset()` (same for `XXX.objects.reset()`).

I'm leaning towards the second option - what do you guys think?

@sphuber 
@giovannipizzi

ltalirz:
fix #2648 
fix #2660 
fix #3462 

The "helper" code for plugin testing is reorganized from a single file `fixtures.py` (that actually did not contain a single fixture) over multiple files.

  * `manage/tests/__init__.py` for the manager, which is now called `TestManager`
    (it has nothing to do with fixtures)
  * `unittest_classes.py` for test classes & runners for use with unittest
  * `pytest_fixtures.py` for the fixtures for use with pytest (=the main new content in this PR)

In order to avoid breaking the API, the old way of importing the main classes is still supported (with a corresponding deprecation warning).
 
Further additions:

 * `TestManager` can now use either a `TemporaryProfileManager` (which will set up the postgres cluster etc.) or a `ProfileManager` (which will use an existing test Profile). The `ProfileManager` is used automatically when the `TEST_AIIDA_PROFILE` environment variable is set.
 *  Fixes an issue where the temporary postgres cluster could remain running when a test failed

Up for discussion:

 1. The file names `unittest_classes.py` and `pytest_fixtures.py` are a bit long.
    I previously tried `unittest.py` and `pytest.py` but this caused confusion with the external packages of the same name. Open to alternative suggestions
 1. One of the new fixtures (`aiida_localhost_code`) is of a different type than the others, since it needs extra parameters. Depending on the needs of a particular code/plugin even the two provided may not be enough.
    I still think it is useful to have - if people feel this should not be in aiida-core, I can also remove it again.


yakutovicha:
thanks, @ltalirz, I am going to adapt the cp2k plugin to use the fixtures introduced here. Once they work fine I am going to approve your PR.

yakutovicha:
We should add a possibility to not remove the folder where the calculation was run, useful for the tests.

yakutovicha:
thanks, @ltalirz, for the ebd3f68a67e6d5898b775f5b30969953041452cd. I think this feature is absolutely critical to make tests useful for the developers.

yakutovicha:
after taking the last changes I get the following error while running pytest for the aiida-cp2k plugin:

```
self = <aiida.manage.configuration.profile.Profile object at 0x11ea78a10>

    @property
    def default_user(self):
>       return self._attributes[self.KEY_DEFAULT_USER]
E       KeyError: 'default_user_email'

../aiida_core/aiida/manage/configuration/profile.py:124: KeyError
```

ltalirz:
Yes, adding the possibility to avoid the profile creation required a lot of
reorganization and I still need to fix a few things. Will let you know.

sphuber:
To give an idea of how testing is done in `aiida-quantumespresso`.
* Fixtures for `pytest` are written in this [`conftest.py`](https://github.com/aiidateam/aiida-quantumespresso/blob/develop/tests/conftest.py) file. Some that are added in this PR directly overlap so they could be removed and use directly from `aiida-core`.
* There is a fixture to test [`CalcJob`](https://github.com/aiidateam/aiida-quantumespresso/blob/89172686a824d0838e0da2c017bf793184aa92ce/tests/conftest.py#L59) classes. Example usage can be found here for [`PwCalculation`](https://github.com/aiidateam/aiida-quantumespresso/blob/develop/tests/calculations/test_pw.py).
* Similar fixture exists to test [`Parser`](https://github.com/aiidateam/aiida-quantumespresso/blob/89172686a824d0838e0da2c017bf793184aa92ce/tests/conftest.py#L86) implementations. It essentially mocks a completed `CalcJobNode` with the `retrieved` folder data node that contains the retrieved files. Example usage for [`PwParser`](https://github.com/aiidateam/aiida-quantumespresso/blob/89172686a824d0838e0da2c017bf793184aa92ce/tests/parsers/test_pw.py#L77) shows.

The automatic comparison files generation and checking is done through the [`pytest-regressions`](https://github.com/aiidateam/aiida-quantumespresso/blob/89172686a824d0838e0da2c017bf793184aa92ce/setup.json#L65) plugin which is added as a dependency. With this pytest plugin and the two fixtures, it is really easy to write and run tests for calculation jobs and parsers without actually having to run the calculations. Note of course, as mentioned before, that some tests that actually run the code will still be useful and desirable.

aiida/manage/tests/pytest_fixtures.py
```diff
@@ -0,0 +1,147 @@
...
+            scheduler_type='direct'
+        )
+        computer.store()
+        computer.configure()
+
+    return computer
+
+
+@pytest.fixture(scope='function')
+def aiida_code_factory(aiida_localhost):  # pylint: disable=redefined-outer-name
```
greschd: I'm ambivalent on having only local computers: On the one hand, testing with remote computers probably isn't the greatest idea. But on the other hand, there might be some codes which can not be run locally.

Maybe to leave this option open, we should rename this to ``aiida_local_code_factory``?

ltalirz:
Just to have everything in one place - below the pointers from @greschd concerning aiida-pytest:

* example for the config file: https://github.com/greschd/aiida-tbextraction/blob/master/tests/config_example.yml
  On the actual machine I would rename this to ‘config.yml’ (which is in .gitignore), and make the necessary changes.
 * Getting the config: https://github.com/greschd/aiida_pytest/blob/migrate_v1/aiida_pytest/_configure.py
 * Setup of computers: https://github.com/greschd/aiida_pytest/blob/migrate_v1/aiida_pytest/_computer.py
 * Setup of codes: https://github.com/greschd/aiida_pytest/blob/migrate_v1/aiida_pytest/_code.py

One thing that I’ve found to be helpful is having an option to wait for user input before wiping the test DB. That allows for inspecting the state better before it’s gone. See here: https://github.com/greschd/aiida_pytest/blob/migrate_v1/aiida_pytest/_configure.py#L152

sphuber:
@ltalirz was this something that needed to go in before the release?

ltalirz:
I'll continue to work on this today; also to add at least the possibility to use the pytest fixtures together with an existing AiiDA profile.

However, this is not absolutely necessary for the 1.0 release.

ltalirz:
closing this while doing some more work.

ltalirz:
@sphuber Do you remember whether configuring logging is needed here:
https://github.com/aiidateam/aiida-quantumespresso/blob/22d54c868c3c2322a83116476ee369c07bede800/tests/conftest.py#L31-L32

I.e. do you think this is something the fixture manager should do automatically?

aiida/manage/tests/pytest_fixtures.py
ltalirz: Will do!

ltalirz:
@sphuber 
Regarding the `generate_calc_job` and `generate_calc_job_node` fixtures: The `generate_calc_job` fixture seems compact and agnostic and could be added as-is - the `generate_calc_job_node` fixture less so. 
If you would like to include the latter here as well, would you mind making a version that is entirely aiida-qe agnostic?

Finally, as far as I can tell, the fixtures don't make direct use of `pytest-regressions` (but it's rather in how they are used). 
What would be the best way to make it easy for people to use it? Should we just include something in the documentation?

sphuber:
> @sphuber Do you remember whether configuring logging is needed here:
> https://github.com/aiidateam/aiida-quantumespresso/blob/22d54c868c3c2322a83116476ee369c07bede800/tests/conftest.py#L31-L32
> 
> I.e. do you think this is something the fixture manager should do automatically?

I actually happen to have added that just today. It was necessary to get the `DbLogHandler` loaded in time. Otherwise the logging in `Parser` classes, that I trigger through the `parser_from_node` function would not end up triggering the handler, therefore attaching the log messages to the node. The weird thing was that the first test the handlers were not configured, but at the second time it was. So it seems the normal logging configuration pathway was being triggered to late. I didn't have time to figure out why, so put this in for the time being.

sphuber:
> If you would like to include the latter here as well, would you mind making a version that is entirely aiida-qe agnostic?

I think the only thing that is specific is how I determine the path of the directory that contains the output files for the retrieved folder. It is expected to be in `parsers/fixtures/<calc_name>/<test_name>` where `test_name` is passed to the fixture but the `<calc_name>` is deduced from the entry point name. Since these are always something like `quantumespresso.pw` and I don't want to include `quantumespresso.` in the folder name, I strip that, which is hardcoded. I could make it simply so the exact path of the retrieved folder needs to be passed in, but that will add a slightly bigger burden on the caller.

> Finally, as far as I can tell, the fixtures don't make direct use of `pytest-regressions` (but it's rather in how they are used).
> What would be the best way to make it easy for people to use it? Should we just include something in the documentation?

the `pytest-regressions` I just use to easily create, update and compare the expected results files. Since their content really just depends on what you want to test, which changes per test, it does not make sense to put in a fixture. It is just used in the tests themselves. I could write some examples to include in the documentation on how use all the fixtures

> P.S. You have rights to push to my fork, so feel free to proceed however is easiest for you.

Did you put in all the changes you wanted?

P.S.: regarding naming, I was wondering whether it makes sense to prefix all fixtures with `fixture_`. Having `aiida_profile` does not really convey what it does.

ltalirz:
Re `generate_calc_job_node`: There's more hardcoded stuff like the file names of input and output files. On the one hand it looks a bit verbose; on the other hand I see how it could be useful, so it could be included. Perhaps we should wait with that one until someone else has used it? 

Re `pytest-regressions`: Yes, I think writing something in the docs would be nice.
I'll then use that to adapt the `aiida-diff` demo plugin (i.e. create a "mock" `diff` executable).

> Did you put in all the changes you wanted?

Not yet, but free to go ahead (I can always pull ;-) ).

> regarding naming, I was wondering whether it makes sense to prefix all fixtures with fixture_. Having aiida_profile does not really convey what it does.

I think we're using pytest fixtures as recommended (also in terms of naming, see e.g. [here](https://docs.pytest.org/en/latest/fixture.html)). I've never seen fixture names prefixed with `fixture_` - that would need to be added to every fixture...
I would vote to keep it as it is.

ltalirz:
@sphuber This would basically be ready for review.

I haven't yet incorporated Dominik's additions - perhaps I'll find time but if not I think this could also be done in a separate PR.

aiida/manage/tests/pytest_fixtures.py
```diff
@@ -0,0 +1,145 @@
...
+
+@pytest.fixture(scope='session', autouse=True)
+def aiida_profile():
+    """Set up AiiDA test profile for the duration of the tests.
+
+    Note: scope='session' limits this fixture to run once per session. Thanks to ``autouse=True``, you don't actually
+     need to depend on it explicitly - it will activate as soon as you import it in your ``conftest.py``.
+    """
+    # create new TestManager instance
+    with test_manager(backend=get_test_backend()) as test_mgr:
```
sphuber: We either need to configure the logger here, or find the correct place of doing it in the manager.
In any case, in this setup I don't think the logging will be configured for the first test.
```
with test_manager(backend=get_test_backend()) as test_mgr:
    from aiida.common.log import configure_logging
    configure_logging(with_orm=True)
    yield test_mgr
```

aiida/manage/tests/pytest_fixtures.py
```diff
@@ -0,0 +1,145 @@
...
+        yield test_mgr
+    # here, the TestManager instance has already been destroyed
+
+
+@pytest.fixture(scope='function')
+def clear_database(aiida_profile):  # pylint: disable=redefined-outer-name
+    """Clear the database after each test.
+
+    Note: Thanks to ``autouse=True``, this happens as soon as you import this fixure in your ``conftest.py``.
```
sphuber: `autouse=True` is no longer the case, correct?

aiida/manage/tests/pytest_fixtures.py
```diff
@@ -0,0 +1,145 @@
...
+def temp_dir():
+    """Get a temporary directory.
+
+    E.g. to use as the working directory of an AiiDA computer.
+
+    :return: The path to the directory
+    :rtype: str
+
+    """
+    dirpath = tempfile.mkdtemp()
```
sphuber: This should probably be in a `try/finally` to make sure it gets cleaned even when exceptions are hit, or does the `pytest` fixture somehow magically deal with this?

aiida/manage/tests/pytest_fixtures.py
```diff
@@ -0,0 +1,145 @@
...
+        computer = Computer(
+            name=name,
+            description='localhost computer set up by test manager',
+            hostname=name,
+            workdir=temp_dir,
+            transport_type='local',
+            scheduler_type='direct'
+        )
+        computer.store()
+        computer.configure()
```
sphuber: Should we really `configure` the computer as well already? There might be use cases where one does not want a configured computer yet and if one does, doing so is one line plus gives the opportunity of determining exactly how to configure

aiida/manage/tests/pytest_fixtures.py
```diff
@@ -0,0 +1,145 @@
...
+
+    return computer
+
+
+@pytest.fixture(scope='function')
+def aiida_local_code_factory(aiida_localhost):  # pylint: disable=redefined-outer-name
```
sphuber: Why `_factory`? The previous fixture did not include this suffix

aiida/manage/tests/pytest_fixtures.py
```diff
@@ -0,0 +1,145 @@
...
+        :rtype: :py:class:`aiida.orm.Code`
+        """
+        from aiida.orm import Code
+        from aiida.common.files import which
+
+        codes = Code.objects.find(filters={'label': executable})  # pylint: disable=no-member
+        if codes:
+            return codes[0]
+
+        executable_path = which(executable)
```
sphuber: What happens if `which` cannot find it? I also use a similar fixture to create a `Code` instance on localhost, but I actually never intend to run it. I just need the `Code` object. This fixture would make that impossible. For reference this is what I use:
```
@pytest.fixture
def generate_code_localhost():
    """Return a `Code` instance configured to run calculations of given entry point on localhost `Computer`."""

    def _generate_code_localhost(entry_point_name, computer):
        from aiida.orm import Code
        plugin_name = entry_point_name
        remote_computer_exec = [computer, '/bin/true']
        return Code(input_plugin_name=plugin_name, remote_computer_exec=remote_computer_exec)

    return _generate_code_localhost
```
Can we make it optional of the `executable` really having to correspond to a physical executable on the machine?

aiida/manage/tests/unittest_classes.py
```diff
@@ -0,0 +1,85 @@
...
+from __future__ import print_function
+import unittest
+
+from aiida.manage.manager import get_manager
+from . import _GLOBAL_TEST_MANAGER, test_manager, get_test_backend
+
+__all__ = ('PluginTestCase', 'TestRunner')
+
+
+class PluginTestCase(unittest.TestCase):
```
sphuber: Do we want to keep supporting `unittest`? I thought we discussed deprecating this to be removed in `2.0.0` and directing people to use `pytest` instead

docs/source/developer_guide/plugins/plugin_tests.rst
```diff
@@ -1,165 +1,146 @@
...
-The recommended way is to use the `pytest`_ framework, while the `unittest`_ package is also supported.
-Internally, test environments are created and managed by the :py:func:`aiida.manage.fixtures.fixture_manager` defined in :py:mod:`aiida.manage.fixtures`.
+Testing AiiDA plugins
+=====================
 
-.. _pytest: https://pytest.org
-.. _unittest: https://docs.python.org/library/unittest.html
-.. _fixture: https://docs.pytest.org/en/latest/fixture.html
+When developing a plugin it is important to write tests.
+We recommend using the `pytest`_ framework, while he `unittest`_ framework is also supported.
```
sphuber:
```suggestion
We recommend using the `pytest`_ framework, while the `unittest`_ framework is also supported.
```

docs/source/developer_guide/plugins/plugin_tests.rst
```diff
@@ -1,165 +1,146 @@
...
 
-Using the pytest framework
---------------------------
+One concern when running tests for AiiDA plugins is to separate the test environment from your production environment.
+Ideally, each individual test should be run against a fresh AiiDA database.
```
sphuber:
```suggestion
Depending on the kind of test, each should even be run against a fresh AiiDA database.
```

docs/source/developer_guide/plugins/plugin_tests.rst
```diff
@@ -1,165 +1,146 @@
...
+thus letting you focus on testing the functionality of your plugin without having to worry about this separation.
 
+.. note::
+   The overhead for setting up the temporary environment is of the order of a few seconds and occurs only once per test session.
 
-To utilize the ``fixture_manager``, we first need to define the actual fixtures:
+If you prefer to run tests on an existing profile, say ``test_profile``, simply set the following environment variable before running your tests::
 
-.. literalinclude:: conftest.py
+  export AIIDA_TEST_PROFILE=test_profile
```
sphuber: is `AIIDA_TEST_PROFILE` correct? I think I saw you use `TEST_AIIDA_PROFILE` in our own tests

aiida/orm/entities.py
```diff
@@ -163,6 +163,16 @@ def all(self):
...
 
+    @classmethod
+    def reset(cls):
+        """
+        Reset collection to initial state.
+
+        Resets collection instance back to its initial state (to be constructed from scratch when next requested).
+        Useful to clear any internal caches of the collection.
+        """
+        cls._COLLECTIONS = datastructures.LazyStore()
```
sphuber: I don't think this does what you intend it to do. This will reset the collection for _all_ entities and backend. So when you call `User.object.reset` it will reset all collections not just that of `User`. I would really go for the alternative you proposed for now and implement a specific default user reset on the `User` collection

yakutovicha:
Just a comment from my side: I used those modifications to test `aiida-cp2k` plugin and it worked fine on my local machine: https://github.com/aiidateam/aiida-cp2k/pull/49

Beware that on travis the tests still fail because they are based on 1.0.0b6 release of `aiida_core`.

aiida/manage/tests/pytest_fixtures.py
ltalirz: Thanks for the hint - I'll do this in the `test_manager` contextmanager directly.

aiida/manage/tests/pytest_fixtures.py
ltalirz: Well spotted!

aiida/manage/tests/pytest_fixtures.py
ltalirz: great point - no, I don't think pytest deals with this.

aiida/manage/tests/pytest_fixtures.py
ltalirz: Well, if you don't configure the computer, you can't use it, right?
Do you want to add another fixture that configures the unconfigured computer?

aiida/manage/tests/pytest_fixtures.py
ltalirz: As you can see in the code of the fixture, it does not return a code but rather a function (factory?) to create a code.
In principle, that function could be a standalone function - were it not for the `aiida_localhost`  default value of the computer.

Happy to change if you have better suggestions for the interface.

aiida/manage/tests/pytest_fixtures.py
sphuber: In all my tests I never actually _use_ the computer, just like I don't use the `Code` instances. I just need those nodes as placeholders. I think that is a perfectly valid way of testing. In this case for me it does not matter if the computer _is_ configured, but I was just thinking to potential use cases where that would be a problem

aiida/manage/tests/pytest_fixtures.py
sphuber: I see, this one allows to pass inputs to change the behavior. Fair enough, factory does make sense in that case

aiida/manage/tests/pytest_fixtures.py
ltalirz: I think if an executable is specified and not found, the fixture should fail.

One could think of setting the default value of the `executable` parameter to `None` and then doing something like
```python
        if executable is None:
            executable_path = '/fake/path' 
        else:        
            executable_path = which(executable)
```

I would be ok with this - still, it feels a bit like cheating since we are taking advantage of the fact that the `Code` constructor does not check whether the path that is provided actually exists.
There is actually an issue https://github.com/aiidateam/aiida-core/issues/868 (opened by you ;-) ) to fix this - after that, this solution would no longer work.

Let me know what you think.

aiida/manage/tests/unittest_classes.py
ltalirz: Good point - the docs already state that `pytest`  is the preferred way.

However, we can add a deprecation warning here as well.

docs/source/developer_guide/plugins/plugin_tests.rst
ltalirz: Thanks for spotting this - this actually highlights a bit of an issue:

We've been using `TEST_AIIDA_BACKEND` so far in the travis tests, so `TEST_AIIDA_PROFILE` would follow this pattern.
However, the other environment variables, such as `AIIDA_PATH` start with `AIIDA` and if we would like to keep this pattern, now would be a good point to decide.

Perhaps `AIIDA_TEST_PROFILE` and `AIIDA_TEST_BACKEND` would be better?

aiida/orm/entities.py
ltalirz: Thanks, for the moment I've implemented the modification described in #3462 

I can still change it if this is thought to be the worse solution.

ltalirz:
@sphuber Thanks a lot for the thorough review and the great suggestions.
I've gone through all your comments now & pushed some fixes.

aiida/manage/tests/pytest_fixtures.py
sphuber: I see, although in reference to #868 I now don't think that in the case of a remote machine, code creation should validate the path, given that this would have to open transport adding a lot more complexity and potential for errors. For local codes and codes on localhost computers there is still something to be said for this check and this would amount to your `which` check in the tests. 

That is why I was using `/bin/true` though, which would at least work on typical Linux and MacOS systems. I still think that there should be a way of creating a `Code` instance for testing purposes that does not necessarily have an actual existing path. All the testing in `aiida-quantumespresso` is done this way, so I think this is really valuable and it is not necessary to actually run the calculations to provide useful testing. Would your current implementation work if I pass `/bin/true`, I think so right?

ltalirz: > Would your current implementation work if I pass /bin/true, I think so right?

Yes - by the way, `/bin/true` does not exist on my system:

```python
In [1]: from aiida.common.files import which

In [2]: which('true')
Out[2]: '/opt/local/libexec/gnubin/true'

In [3]: which('/bin/true')

In [4]: which('/opt/local/libexec/gnubin/true')
Out[4]: '/opt/local/libexec/gnubin/true'
```

docs/source/developer_guide/plugins/plugin_tests.rst
ltalirz: To summarize: we discussed to switch all variables to `AIIDA_...`, and to introduce a warning when running `verdi devel tests` while the `AIIDA_TEST_PROFILE` is set, so that as long as the two testing systems coexist, people don't get confused by how to control each.

ltalirz:
@sphuber I've made the changes we discussed.
I still left three comments open - feel free to resolve or to follow up further.

aiida/manage/tests/unittest_classes.py
sphuber: If we are not going to support it, which I don't think we should, we should add the deprecation warning

ltalirz: Well, the `PluginTestCase` has been in AiiDA since some time and people could rely on it.
This goes back to our promise not to break the API after 1.0.0b2.

sphuber: I am not saying that we remove it now or even in `1.*` but we can deprecate it and remove it in `2.0.0` no? Unless you think it makes sense to keep it and keep supporting it of course

ltalirz: Oops, sorry, I misread your comment.
I did add the deprecation warning in the TestRunner, not at the module level (because I wanted to avoid deprecation warnings popping up accidentally).

Let me know whether you think there is a better place to put it.
I'm even ok with moving it at the module level if you think this is better.

aiida/orm/entities.py
```diff
@@ -163,6 +163,15 @@ def all(self):
...
         return [_[0] for _ in self.query().all()]
 
+    def reset(self):
+        """
+        Remove lazily constructed collection instance.
+
+        Removes lazily constructed collection instance from LazyStore (constructed again upon next request).
+        Useful to clear any internal caches of the collection.
+        """
+        self._COLLECTIONS.pop((self.entity_type, self.backend))
```
sphuber: Won't this raise if the `(self.entity_type, self.backend)` tuple does not exist?

ltalirz: It should definitely have tests ;-) 
If you agree to add this functionality, I'm happy to add some - just let me know.

sphuber: It feels like it is quite important functionality and with potentially big effects to just include in this PR. Would it be possible to just use the specific solution instead for now on the `User` collection. Just a `User.objects.set_default` or `User.objects.reset_default` even. I would feel a bit more comfortable with a more limited scope as part of this bigger PR that has really a completely different aim

aiida/manage/tests/unittest_classes.py
sphuber: Ah, I missed that one, then it is fine :+1:

aiida/orm/entities.py
ltalirz: Ok, will do!

ltalirz: I'm offline for a while now (will fix this PR over the weekend)

ltalirz:
@sphuber https://github.com/aiidateam/aiida-core/pull/3319/commits/61d696e9a2263f1a8d8dd1e032a900ee52821cdb includes the modification we discussed (moving the cache reset to the `User` collection only).

If there's anything else left, let me know!

sphuber:
Thanks a lot @ltalirz

greschd:
@ltalirz I'm just migrating ``aiida-pytest`` to match these changes -- is there a nicer way to get at the root directory than through the ``aiida_profile`` fixture's ``aiida_profile._manager.root_dir``?

ltalirz:
@greschd probably not - if you have any small suggestions for improvements, now would be a good time to make a PR.

By the way, sorry for not yet having included more from aiida-pytest's features, it was just lack of time on my side. I'll open an issue so that these things don't get lost and we can keep improving the fixtures.

greschd:
I haven't had the time to look at it in detail, but from the surface it looks pretty good - better than that hot mess that is `aiida-pytest` for sure 😅

PR #3322, opened 2019-09-13
Also export empty repository folders (for zip)

### LINKED ISSUE #3321: Empty repository folders not exported
If a `Node` has an empty repository folder, it will not be exported, which will result in it not being importable.

This is due to the way the repository folders are found and exported in the export function.
A comment in [this](https://github.com/aiidateam/aiida-core/blob/develop/aiida/tools/importexport/dbexport/__init__.py#L789) line also confirms it, essentially.

Found by @d-tomerini.

CasperWA:
Fixes #3321 

Blocked by PR #3242, since this fixes the issue of the `extract_tree` function, which the new test makes use of.

Edit: The actual fix is not blocked, only the test. I have skipped the new test until PR #3242 has been merged/issue #3199 has been fixed.
It should now be good-to-go.

### DIFF
--- a/aiida/backends/tests/tools/importexport/test_specific_import.py
+++ b/aiida/backends/tests/tools/importexport/test_specific_import.py
@@ -17,6 +17,8 @@ import os
 import shutil
 import tempfile
 
+import unittest
+
 import numpy as np
 
 from aiida import orm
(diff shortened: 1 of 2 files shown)


aiida/backends/tests/tools/importexport/test_specific_import.py
```diff
@@ -268,3 +270,44 @@ def test_missing_node_repo_folder_import(self, temp_dir):
...
+
+    @unittest.skip('Reenable when issue #3199 is solve (PR #3242): Fix `extract_tree`')
+    @with_temp_dir
+    def test_empty_repo_folder_export(self, temp_dir):
+        """Check a Node's empty repository folder is exported properly"""
+        from aiida.common.folders import Folder
+        from aiida.tools.importexport.dbexport import export_zip, export_tree
+
+        node = orm.Dict().store()
+        node_uuid = node.uuid
```
d-tomerini: This seems to be used exactly once in the test? Although it makes more sense in the others

CasperWA: In the other tests? Sure. I believe the other tests are very old and can easily be updated to reflect the change in the code in general, i.e., checking Nodes based on their UUID, like this, and more. It should be noted that this is how it is done in all new export/import tests (as can be seen in the other test-files).

Or do you mean that I should check the Node's UUID after each import in _this_ test? I can do this :)

d-tomerini: Hi, it was a much less complex comment :) . 
I meant that you use the variable ``node_uuid`` exactly once in the test, so you could use directly ``node.uuid`` a few lines later. But I do see it is consistent with the rest of the tests in the file...

CasperWA: Ah right :)
Well, I think I will just use it more and check that we actually correctly import the Node each time.

The reason it is saved like this, is due to the `self.reset_database()` that makes it impossible to retrieve the Node's UUID after it has been deleted. However, in this case, you're right, it is used before it is deleted. It should be used afterwards to make sure we are getting the Node back into the database.

CasperWA:
@d-tomerini I have now updated the test, made sure it passes (when rebased on the branch for PR #3242), and all should be good-to-go :)

Issue #3296, opened 2019-09-02
Rules for deleting Nodes


CasperWA:
Following an offline discussion of issue #1762 these are the rules we have decided on concerning Node deletion.
Note, the Node types in bold symbolize the Node that is currently being deleted, i.e., the table answers the question "should I follow the found provenance graph edge/link and delete the Linked Node?"

**LinkType** | **From** | **To** | Deletion rule
------------ | -------- | ------ | -------------
INPUT_CALC | **Data** | CalculationNode | Remove CalculationNode
INPUT_CALC | Data | **CalculationNode** | Keep Data, remove Link
CREATE | **CalculationNode** | Data | *OPTIONAL*, default: Delete Data
CREATE | CalculationNode | **Data** | Remove CalculationNode
RETURN | **WorkflowNode** | Data | Keep Data, remove Link
RETURN | WorkflowNode | **Data** | Remove WorkflowNode
INPUT_WORK | **Data** | WorkflowNode | Remove WorkflowNode
INPUT_WORK | Data | **WorkflowNode** | Keep Data, remove Link
CALL_CALC | **WorkflowNode** | CalculationNode | *OPTIONAL*, default: Keep CalculationNode, remove Link
CALL_CALC | WorkflowNode | **CalculationNode** | Remove WorkflowNode
CALL_WORK | **WorkflowNode** | WorkflowNode | *OPTIONAL*, default: Keep WorkflowNode, remove Link
CALL_WORK | WorkflowNode | **WorkflowNode** | Remove WorkflowNode

These rules should be implemented in the new deletion logic.

PR #3325, opened 2019-09-18
Delete/Export - Documentation and Final Details

### LINKED ISSUE #3309: Documenting delete_node and export_node functionalities
Following the discussion of issue #1762 (related to issues #3262 and #3296 ) the documentation for both functionalities should probably be in the same place and the proposed location for now would be as a sub-section of the `concepts` section.

ramirezfranciscof:
Fixes #3309 

Unifies the discussions in #3295 and #3296 in order to provide a consistent interface for both features. The content still needs re-arrangement in order to be more clear (more graphics, less text) but I'm uploading it now since I would need an agreement on the interface for the final version. Soon @sphuber will be making his suggestions here.


sphuber:
Discussed this in person with @ramirezfranciscof and decided to write my comments down here so it can be discussed with @giovannipizzi .

I think it is a great start but we need to see if we can make it even clearer, because while reading I was struggling a bit to understand. Now to be fair, this is just because it is an innately complex problem with lots of rules and details. Francisco opted to start with a basic summary without going into details, followed by a practical example to show how these rules manifest in practice. Normally I agree this is the way to go, but here I am afraid it won't be feasible. Trying to explain the "algorithm" for graph traversal at the hand of a single big example will never cover all possible cases and therefore will always leave open questions. Instead then, I think we should clearly introduce the problem and then explain the algorithm that even the implementation uses. I propose a structure like the following:

1. **Simple problem sketch**: we want to delete or export a sub graph of our provenance graph. Since the goal of AiiDA is to maintain a _consistent_ graph, the resulting graph after deletion/export should still be consistent. This means that we cannot export just any sub graph. If we chose one or more nodes to be deleted/exported, we need to potentially include more nodes to have the sub graph be internally consistent. Here we can show a trivial example of calculation with single input and output and that if you chose the calculation to be deleted/exported you need the input and output as well.
2. **Explain the algorithm**: how, given a starting set of nodes, is the full consistent sub graph determined?
   - Select a starting set consisting of one or more nodes
   - For each node apply the "traversal rules" (which we explain later)
   - Add each new node reached by traversal rule to the current set.
   - Recursively apply the traversal rules until no new nodes are discovered.
3. **Explain the concept of a traversal rule**: a traversal rule is a rule that states that given a certain node in your sub set, whether you should follow a certain link type (`LinkType`) in a certain direction (forward or backward). Now we can show the full 6 by 2 table (link type in rows and direction as column) what the rule is: `True` or `False`, i.e. it should be followed or should _not_ be followed, respectively. Note that there are two options for a rule:
    - Required: the value of the rule is _always_ enforced and cannot be changed
    - Default: the rule has a default but can be overridden in the API with a flag
The table may also be done with 12 rows instead of 6,  to directly include the direction. The link type + direction in this table should probably be visualized instead of written. For example, like we did during our own discussion, draw two nodes with a single link and circle or highlight the one that is currently in the set, making the other node the candidate considered for addition to the set.
4. **Demonstrate the algorithm by visual example**: with the information above any question about any hypothetical provenance graph will be answerable by a user. It perfectly covers all situations and the exact behavior of the implementation. Granted this might still be a bit technical, so at this point it would be a good time to demonstrate the concept step-by-step with a graphical example.
Start with a reasonably complex graph and propose to delete/export a single node. Using the algorithm and table explained before, let the set grow by applying the traversal rules one by one.


I told Francisco to hold on with implementing this so we get to chance to discuss this first to see if others agree. This to prevent a lot of changes being done for nothing.

sphuber:
I think the two tables from #3295 and #3296 can be combined and condensed in the following traversal rule matrix that I described in the previous comment. For all the 12 possible rules, one for each link type both forwards and backwards, it gives whether the link should be followed `True` or not `False`. This is done separately for export delete as the rules differ. If and only if the rule is not required and the default is allowed to be overridden, a keyword is added in parentheses. The names are temporary.

**Rule**               | **Export**                    | **Delete**
---------------------- | ----------------------------- | -----------------------------
`INPUT_CALC_FORWARD`   |  `False` (`input_forward`)    | `True`
`INPUT_CALC_BACKWARD`  |  `True`                       | `False`
`CREATE_FORWARD`       |  `True`                       | `True`  (`create_forward`)
`CREATE_BACKWARD`      |  `True`  (`create_reversed`)  | `True`
`RETURN_FORWARD`       |  `True`                       | `False`
`RETURN_BACKWARD`      |  `False` (`return_reversed`)  | `True`
`INPUT_WORK_FORWARD`   |  `False` (`input_forward`)    | `True`
`INPUT_WORK_BACKWARD`  |  `True`                       | `False`
`CALL_CALC_FORWARD`    |  `True`                       | `False`  (`call_forward`)
`CALL_CALC_BACKWARD`   |  `False` (`call_reversed`)    | `True`
`CALL_WORK_FORWARD`    |  `True`                       | `False`  (`call_forward`)
`CALL_WORK_BACKWARD`   |  `False` (`call_reversed`)    | `True`

ramirezfranciscof:
Just as a comment to the last table and keywords used: although we do not need (in principle) to make the distinction between `call_reversed` for export in a `call_work` or `call_calc`, we do need it in `call_forward` so as to have the option of deleting the logical provenance without it also deleting the data provenance.

giovannipizzi:
> Now to be fair, this is just because it is an innately complex problem with lots of rules and details. 

Agreed.

> Francisco opted to start with a basic summary without going into details, followed by a practical example to show how these rules manifest in practice. 

Indeed, I think it was me who suggested this originally.

> I propose a structure like the following:
> 
> 1. **Simple problem sketch**: we want to delete or export a sub graph of our provenance graph. Since the goal of AiiDA is to maintain a _consistent_ graph, the resulting graph after deletion/export should still be consistent. This means that we cannot export just any sub graph. If we chose one or more nodes to be deleted/exported, we need to potentially include more nodes to have the sub graph be internally consistent. Here we can show a trivial example of calculation with single input and output and that if you chose the calculation to be deleted/exported you need the input and output as well.

Ok for me. I think Francisco is already discussing why a calculation cannot "lose" its inputs/outputs, maybe just some rephrasing is needed.

> 2. **Explain the algorithm**: how, given a starting set of nodes, is the full consistent sub graph determined?
>    
>    * Select a starting set consisting of one or more nodes
>    * For each node apply the "traversal rules" (which we explain later)
>    * Add each new node reached by traversal rule to the current set.
>    * Recursively apply the traversal rules until no new nodes are discovered.

Ok. I think this can be easily explained (even as an itemised list as above).

> 3. **Explain the concept of a traversal rule**: a traversal rule is a rule that states that given a certain node in your sub set, whether you should follow a certain link type (`LinkType`) in a certain direction (forward or backward). So for a given node, consider all its neighbors and based on the link type and direction, the traversal rules will tell you whether the neighbor should be included in the set.

Ok. However, I don't know if we should explain here the details of how the algorithm actually work. I would keep this section to a minimum, what is really critical for users to understand what is going to be deleted.

> 4. **Complete table of traversal rules**: now we can show the full 6 by 2 table (link type in rows and direction as column) what the rule is: `True` or `False`, i.e. it should be followed or should _not_ be followed, respectively. Note that there are two options for a rule:
>    
>    * Required: the value of the rule is _always_ enforced and cannot be changed
>    * Default: the rule has a default but can be overridden in the API with a flag

Ok to put a table here (I think it's the better thing to make things well defined).

>      The link type + direction in this table should probably be visualized instead of written. 

I suggest writing it, plus adding on the side a small graphical representation. Having it written might be easier to interpret for quick reference for experienced users, but for others the graphics might be easier instead.

> 5. **Demonstrate the algorithm by visual example**: with the information above any question about any hypothetical provenance graph will be answerable by a user. It perfectly covers all situations and the exact behavior of the implementation. Granted this might still be a bit technical, so at this point it would be a good time to demonstrate the concept step-by-step with a graphical example.

Ok. What I think we should also keep here (maybe shorter than now) is a description of some common use cases, which combination of flags to use and an example on that graph of what will happen.

> I would be hesitant to start to introduce "compound" flags in the API because they run the risk to make the situation even more complex. 

I agree.

> I do agree with Casper that having to expose all twelve flags at the highest user level might also be undesirable. We will have to see how this works out.

Yes. However, if we want to use compound flags, let's do it in a way that allows us to split them in the future in a backward-compatible way, without having to wait for 2.0. It is typically doable.

> One final thing that I think is important and will make all of this more consistent is that we should stop to think and talk of export and delete as different things.

I agree.

> I think it would be wise to move this to `aiida-core` and have both the export and delete functionality leverage it for the sub graph determination.

If possible it would be great.

giovannipizzi:
About the table: 

> I think the two tables from #3295 and #3296 can be combined and condensed in the following traversal rule matrix that I described in the previous comment. [...]

Maybe the table can be further simplified? I think that any forward rule for delete maps to the corresponding backward rule for export. I don't know if making the table more compact will reduce its clarity (in which case, ok to keep the table as @sphuber suggested) but I would in some way clarify this pairing (at least with a sentence somewhere close to where we put the table).

Regarding this, I was wondering again on why we miss flags for following INPUT_CALC_BACKWARD, RETURN_FORWARD and INPUT_WORK_BACKWARD for delete (while we have the corresponding flags for export [with forward/backward inverted]). IIRC it was to be safer when deleting, and avoid too much cascading being automatically applied, risking to delete much more than expected, right? Maybe we if we discuss the relationship I'm mentioning above, we should also mention this as well.

Also,
> Just as a comment to the last table and keywords used: although we do not need (in principle) to make the distinction between call_reversed for export in a call_work or call_calc, we do need it in call_forward so as to have the option of deleting the logical provenance without it also deleting the data provenance.

Correct.

giovannipizzi:
a note on the `cawork`/`cacall` shortening in one of the graphs: while `inwork` is clear, `cawork`/`cacall` is not so clear I believe. Also, I think the two are not consistent: if `cawork` stands for `call_work`, then it shouldn't be `cacall` but `cacalc`. 

Maybe it's clearer though (if a shortening is really needed) to call them `call_w`, `call_c`, `input_c`, `input_w`? (Other opinions welcome)

ramirezfranciscof:
> Regarding this, I was wondering again on why we miss flags for following INPUT_CALC_BACKWARD, RETURN_FORWARD and INPUT_WORK_BACKWARD for delete (while we have the corresponding flags for export [with forward/backward inverted]). IIRC it was to be safer when deleting, and avoid too much cascading being automatically applied, risking to delete much more than expected, right? Maybe we if we discuss the relationship I'm mentioning above, we should also mention this as well.

For INPUT_CALC_BACKWARD in delete my guess is that we just thought "if the user wanted to delete the input, he should have targeted it instead of the calculation". One could make the case that you might want to delete ALL inputs associated with a calculation without picking them one by one in which case it would be useful to have this option. The counter argument, however, is that since we necessarily delete calculations when removing any of its outputs, activating INPUT_CALC_BACKWARDS effectively deletes the whole upstream data provenance associated with that branch, making it a bit...dangerous? And also maybe limiting the usefulnes (at least until we can implement a limit of the kind "apply INPUT_CALC_BACKWARDS but just for the starting nodes").

A similar argument can be made for INPUT_WORK_BACKWARD.

For RETURN_FORWARD the idea was that if you want to delete the return, you do it by first deleting the calculations that created them (CALL_CALC_FORWARD + CREATE_FORWARD), since anyways if you delete the returned output first it will also delete that calculation afterwards (because of CREATE_BACKWARD always being true). A case could be made if you wanted to target a workflow and just delete its returned value (and consequently the last calculation as well) but wanted to keep the intermediate calculations and data nodes of the workflow intact (which would otherwise be deleted with CALL_CALC_FORWARD, but not necessarily with a RETURN_FORWARD). I'm not sure if there would be many cases in which people would want to "decouple" a workflow like this, besides the fact that would only be useful to delete the very last process and not just a general way of 'cutting' a workflow.

ramirezfranciscof:
I would need to know which convention should I use for the AiiDA schematic graphs, specially for the processes. Because in the Concepts/Provenance section the first graphic uses rhomboid for workflows and square for calculation but the next example uses pointing down triangle for workflows and pointing up triangle for calculations (unless there is any reason for this that I'm missing).

sphuber:
The reason for this is historical. I initially created these graphs by hand in SVG specifically for the documentation. Chris Sewell then improved `verdi node graph generate` to try and match this style as much as possible because the documentation by then already almost exclusively that consistent style. The only problem is that the triangle shapes that I originally chose for the documentation is not ideal for auto generated graphs, because it does not fit a lot of content, so often text would run over the borders. We then decided to use the rhomboids instead in the auto-generated. I would say that if you are making new graphs for the documentation and the labels do not have to be big, you can keep the same convention. 

If, however, we think having two conventions will be detrimental in the long run, we should maybe decide to update the current hand-generated docs graphs to use the same style as the auto-generated one.

ramirezfranciscof:
Thanks for the response. So whatever we end up doing with the rest of the documentation (multiple possible shapes or always autogen convention), I'll use rhomboinds and squares which should comply with our final decision.

CasperWA:
> > Seb: I do agree with Casper that having to expose all twelve flags at the highest user level might also be undesirable. We will have to see how this works out.
> 
> Gio: Yes. However, if we want to use compound flags, let's do it in a way that allows us to split them in the future in a backward-compatible way, without having to wait for 2.0. It is typically doable.

Referring to the (at this moment in time) most recent offline AiiDA team meeting, is the consensus to expose all twelve flags at the `verdi` level, both for delete and export?

In any case, I will implement the use of all twelve flags in the export, however, I could still keep exposing only the flags we have chosen to be togglable from previous meeting?

ramirezfranciscof:
My understanding was that yes, when we were talking about exposing the 'base flags', we meant only the ones that can be actually change (at least in my mind this was the case). I think that what we were meant to exclude was flags that dealt with more than one rule (i.e. currently export uses `call_reversed` for both `call_calc_reverse` and `call_work_reverse`).

giovannipizzi:
I agree with @ramirezfranciscof - we shouldn't expose all flags - these are merely for our reference, but many of them must not be togglable by the user.
Also for the others, if we know their behaviour is not very common, we can also for now not expose them (see e.g. [this comment](https://github.com/aiidateam/aiida-core/pull/3325#issuecomment-533548169) by @ramirezfranciscof). And yes, I would avoid for the time being to expose compose flags - let's just expose the few we need to expose, and in later versions depending on real user patterns we can decide to create compose flags.

sphuber:
Just to be sure, when we are talking about exposing flags, are we just talking on the `verdi export` and `verdi node delete` levels? Or also on the highest API layer in the `export*` functions?

CasperWA:
> Just to be sure, when we are talking about exposing flags, are we just talking on the `verdi export` and `verdi node delete` levels? Or also on the highest API layer in the `export*` functions?

In my view, it is on the `verdi` level only. On the highest API layer, all flags should essentially exist - so that they can be introduced easily in `verdi` if needed, however, I think some logic should be in place in the `export` and `delete`functions to determine whether the (valid) flag is allowed to be toggled or not. This it how I have now implemented all flags for the export in #3262.

giovannipizzi:
I'm not sure I fully agree. I think that if we don't allow something, we should simply not show it.
In the python API we can leave all flags that users can modify, and in the `verdi` CLI only those we think are relevant maybe? But I wouldn't expose flags that the user is not supposed to activate and/or deactivate

CasperWA:
> In the python API we can leave all flags that users can modify, and in the `verdi` CLI only those we think are relevant maybe? But I wouldn't expose flags that the user is not supposed to activate and/or deactivate

Fair enough. In the end, I am not exposing anything, since all is enveloped in kwargs. However, there will be a different Exception when using a valid, but not togglable, keyword/flag, and when using an invalid flag.
Essentially, I have done what you specify here for the export (see #3262), while still implementing _all_ flags for the use of the future graph traversal interface and engine.
What is exposed, is what one can see from `verdi`.

CasperWA:
A note from talking offline with @giovannipizzi just now (please correct any misunderstandings):
For the Python API, he thinks the best implementation would be to have the togglable flags as explicit parameters (no matter the amount) and then set the complete list of rules within the function using the defaults. The complete set may then be used later with the coming agnostic graph traversal interface.
Exceptions for test profiles should not be in the code in general.
The flags and the functionality of the functions should reflect, what the user is supposed to be able to do, not introduce flags that cannot or should not be used. In this sense, the function interface will match the `verdi` command in terms of explicit flags.
In line with this, the tests should also not test functionality that the user should not have, i.e., non-togglable flags should _not_ be tested.

For me it comes down to a design and implementation question. Considering the work I have done already in PR #3262, I would like to have consensus on this before reverting some changes in said PR. So what do you think @sphuber, @ltalirz?

sphuber:
I agree with only exposing those flags on the top level python API that are allowed to be changed anyway. Whether we reduce these even further towards the CLI is debatable. I think that if they make sense in the python API they also make sense for `verdi`. However, in terms of testing, if our backend implementation does in principle have knowledge of all the flags (which I think it should) then it should also be tested there.

giovannipizzi:
> I agree with only exposing those flags on the top level python API that are allowed to be changed anyway. 

I agree.

> Whether we reduce these even further towards the CLI is debatable. I think that if they make sense in the python API they also make sense for verdi. 

Probably. Anyway if there are a couple that we think are not critical we can skip them in the cli for now (but ok to keep the same).

> However, in terms of testing, if our backend implementation does in principle have knowledge of all the flags (which I think it should) then it should also be tested there.

My point was: the testing of the traversal should not be done when testing the export, but when testing the AGE or whichever tool we implement.
When testing the export, we should only test the functionality that the user will have the possibility to do (or test that the user cannot do it and gets an exception).
The main point is that I don't think we should have in the main code things like:
```python
if not in_testing_mode:
    raise ...
```
to allow to do more testing than needed - having this flags clutters the code and risks to introduce bugs testing different code paths than the ones the user will actually encounter.

sphuber:
> My point was: the testing of the traversal should not be done when testing the export, but when testing the AGE or whichever tool we implement.
> When testing the export, we should only test the functionality that the user will have the possibility to do (or test that the user cannot do it and gets an exception).
> The main point is that I don't think we should have in the main code things like:

100% agreed on both counts

ramirezfranciscof:
The adaptations requested were implemented in the last commits, pre-aproved by @giovannipizzi (the very last one was actually some modifications made by him because there were some errors when making the documentation). Let me know if there is something else you want me to correct or if you deem this ready to merge. Apologies for the double push, btw.

sphuber:
Thanks a lot @ramirezfranciscof

PR #3326, opened 2019-09-18
Fixing the drop in test coverage reported

### LINKED ISSUE #3324: Coveralls dropped suddenly
It seems that since PR #3305 coveralls has dropped to 20-30 % (from the original ~75 %).

Looking at [this build](https://coveralls.io/builds/25730940) versus the one [immediately prior to it](https://coveralls.io/builds/25730282) it seems coveralls now also takes into account the absolute file paths on the Travis builds (see under TREE under SOURCE FILES ...).
It seems to be when testing `sphinxext`, since this subfolder has a high coverage under this tree - and this is also what was touched by the PR.

giovannipizzi:
Fixes #3324

This should set properly the coverage configuration
file also in the tests run via pytests, avoiding that
some files are reported with a relative path and some with
an absolute path (and, as a consequence, with a low coverage
both for the relative and for the absolute path, resulting in
a low overall coverage).


giovannipizzi:
Marked as [WIP] to first double check that this is solving the issue, and solving it in full.
If it gets back to normal (no absolute path reported) feel free to remove [WIP] and merge, otherwise suggestions are welcome.

sphuber:
It still seems to include the `/home` directory in the end report by coverage.

giovannipizzi:
Do you see also the report? It's not reported here, I think it's [this build](https://coveralls.io/builds/25788115) but I get a 500 error from Coveralls...

CasperWA:
> Do you see also the report? It's not reported here, I think it's [this build](https://coveralls.io/builds/25788115) but I get a 500 error from Coveralls...

Same here. Other builds are the same, so it is not your change, I think.

giovannipizzi:
In my previous commit (and in my comment in the issue) I was wrong: adding `--cov-config` is useless since `.coveragerc` is the default. Instead, I checked and replacing `pytest --cov` (using the `pytest-cov` module) with `coverage run -m pytest` seems to fix the issue (at least locally on my laptop) of creating a report with also absolute files.

I think this sentence from [the pytest-cov README file](https://github.com/pytest-dev/pytest-cov/blob/master/README.rst) explains the reason (and why this is connected with removing `'.'` from the `sys.path`):

> Compared to just using coverage run this plugin does some extras: [...] Consistent pytest behavior. If you run coverage run -m pytest you will have slightly different sys.path (CWD will be in it, unlike when running pytest).

Note that above "this plugin" refers to `pytest-cov`.

Since we actually want to have `.` in the path, we can use directly `coverage run -m pytest`, and so I am also removing the dependency on pytest-cov. Let's see if the coverage improves.

sphuber:
Thanks a lot @giovannipizzi

giovannipizzi:
> sorry for not acting on this sooner, I'm currently struggling just to keep up with replying to emails...

No problem!

PR #3327, opened 2019-09-18
Fix the identifier format of the caching configuration file

### LINKED ISSUE #2990: per process_type caching configuration is not avalaible for CalcJobNode

sphuber:
Fixes #2990 

The old caching configuration file implementation expected the module
path of node sub classes that should be either enabled or disabled for
caching, for example:

    `aiida.orm.node.calculation.job.TemplatereplacerCalculation`

The old caching implementation in the `Node.store` method would
then compare these directives to the node type of the current node being
stored to determine whether it should be considered for caching. This
method broke after the separation of the "process" and "node" concepts.
Since calculation job classes, for example, are now no longer a sub
class of `Node`, but of `Process`, the corresponding node that will be
created will all be `CalcJobNode` instances, regardless of the exact
process sub class. That particular bit of information is stored as its
entry point string in the `process_type` column of the node. For example:

    `aiida.calculations:templatereplacer`

This entry point string is now also the identifier that determines the
caching behavior and should be used in the configuration file.

### DIFF
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -15,6 +15,7 @@ from __future__ import absolute_import
 
 import copy
 import importlib
+import warnings
 import six
 
 from aiida.common import exceptions
@@ -22,6 +23,7 @@ from aiida.common.escaping import sql_string_match
 from aiida.common.hashing import make_hash, _HASH_EXTRA_KEY
 from aiida.common.lang import classproperty, type_check
 from aiida.common.links import LinkType
+from aiida.common.warnings import AiidaDeprecationWarning
 from aiida.manage.manager import get_manager
 from aiida.orm.utils.links import LinkManager, LinkTriple
 from aiida.orm.utils.repository import Repository
@@ -680,19 +682,15 @@ class Node(Entity):
         :param force: boolean, if True, will skip the mutability check
         :raises aiida.common.ModificationNotAllowed: if repository is immutable and `force=False`
         """
-        # pylint: disable=no-member
-        import warnings
-        from aiida.common.warnings import AiidaDeprecationWarning
-
         # Note that the defaults of `mode` and `encoding` had to be change to `None` from `w` and `utf-8` resptively, in
         # order to detect when they were being passed such that the deprecation warning can be emitted. The defaults did
         # not make sense and so ignoring them is justified, since the side-effect of this function, a file being copied,
         # will continue working the same.
         if mode is not None:
-            warnings.warn('the `mode` argument is deprecated and will be removed in `v2.0.0`', AiidaDeprecationWarning)
+            warnings.warn('the `mode` argument is deprecated and will be removed in `v2.0.0`', AiidaDeprecationWarning)  # pylint: disable=no-member
 
         if encoding is not None:
-            warnings.warn(
+            warnings.warn(  # pylint: disable=no-member
                 'the `encoding` argument is deprecated and will be removed in `v2.0.0`', AiidaDeprecationWarning
             )
 
@@ -970,6 +968,11 @@ class Node(Entity):
 
         :parameter with_transaction: if False, do not use a transaction because the caller will already have opened one.
         """
+        if use_cache is not None:
+            warnings.warn(  # pylint: disable=no-member
+                'the `use_cache` argument is deprecated and will be removed in `v2.0.0`', AiidaDeprecationWarning
+            )
+
         if self.is_stored:
             raise exceptions.ModificationNotAllowed('Node<{}> is already stored'.format(self.id))
 
@@ -979,9 +982,9 @@ class Node(Entity):
 
         for link_triple in self._incoming_cache:
             if not link_triple.node.is_stored:
-                link_triple.node.store(with_transaction=with_transaction, use_cache=use_cache)
+                link_triple.node.store(with_transaction=with_transaction)
 
-        return self.store(with_transaction, use_cache=use_cache)
+        return self.store(with_transaction)
 
     def store(self, with_transaction=True, use_cache=None):
         """Store the node in the database while saving its attributes and repository directory.
@@ -997,6 +1000,11 @@ class Node(Entity):
         # pylint: disable=arguments-differ
         from aiida.manage.caching import get_use_cache
 
+        if use_cache is not None:
+            warnings.warn(  # pylint: disable=no-member
+                'the `use_cache` argument is deprecated and will be removed in `v2.0.0`', AiidaDeprecationWarning
+            )
+
         if not self._storable:
             raise exceptions.StoringNotAllowed(self._unstorable_message)
 
@@ -1007,9 +1015,8 @@ class Node(Entity):
             # Verify that parents are already stored. Raises if this is not the case.
             self.verify_are_parents_stored()
 
-            # Get default for use_cache if it's not set explicitly.
-            if use_cache is None:
-                use_cache = get_use_cache(type(self))
+            # Determine whether the cache should be used for the process type of this node.
+            use_cache = get_use_cache(identifier=self.process_type)
 
             # Clean the values on the backend node *before* computing the hash in `_get_same_node`. This will allow
             # us to set `clean=False` if we are storing normally, since the values will already have been cleaned
--- a/.ci/test_daemon.py
+++ b/.ci/test_daemon.py
@@ -359,7 +359,7 @@ def main():
     else:
         # Launch the same calculations but with caching enabled -- these should be FINISHED immediately
         cached_calcs = []
-        with enable_caching(node_class=CalcJobNode):
+        with enable_caching(identifier='aiida.calculations:templatereplacer'):
             for counter in range(1, number_calculations + 1):
                 inputval = counter
                 calc, expected_result = run_calculation(code=code, counter=counter, inputval=inputval)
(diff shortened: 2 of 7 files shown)


aiida/orm/nodes/node.py
```diff
@@ -970,6 +968,11 @@ def store_all(self, with_transaction=True, use_cache=None):
 
         :parameter with_transaction: if False, do not use a transaction because the caller will already have opened one.
         """
+        if use_cache is not None:
```
greschd: Since the ``use_cache`` argument is now just ignored, would it make sense to drop it before 1.0? Or is the interface for 1.0 frozen already?

sphuber: We froze the interface with `v1.0.0b3` and decided not to change it anymore. I doubt anyone was using it, but just to be consistent as much as possible with what we promised towards users

sphuber:
> One thing I'm unsure about is whether it would be useful to allow "short-hand" identifiers: Since (I think) the type will always be `aiida.calculations`, we could think about allowing e.g. `templatereplacer` instead of `aiida.calculations:templatereplacer`.

We thought about this but decided to be explicit. The reason being that, even though now only calculations (`CalcJob`'s or calculation functions) are cacheable, in the future we might solve the limitation of workflows (not sure if this is at all possible or it is impossible by definition, we actually were discussing recently what the original reasoning was for proving that it is impossible) and re-enable it. This is also why I propose the caching business logic to the `ProcessNode` class and not just the `CalculationNode` sub class. If then in the future we would allow workflows to be cached, we need the `aiida.calculations` and `aiida.workflows` prefixes to make them uniquely resolvable

greschd:
Yeah, on balance I agree with that decision.

The problem with caching in workflows is that they can return something which belongs to their "ancestry". When called again with the same inputs, the graph relationships between ancestors are not necessarily the same -- so it's unclear which ancestor the return link should point to. For calculations this problem doesn't exist because outputs are always new nodes.

For workflows which don't have return links to ancestors however it should be possible to just re-create the graph relationship of their children, and enable caching in that way.

ltalirz:
> When called again with the same inputs, the graph relationships between ancestors are not necessarily the same -- so it's unclear which ancestor the return link should point to.

While this is a limitation of the system, a very simple practical solution can be to figure out which inputs to return based on the hash of the returned node.
If a workflow contains multiple input nodes with the same hash, simply pick one of them.

Not saying that this addresses 100% of the cases but I guess it will address 99%.

ltalirz:
P.S. I'm still not convinced that the benefits that come with having the option to directly return data really outweigh the cost imposed by the added complexity, both conceptual and in terms of code...
But that's not a discussion for today or anytime soon ;-)

greschd:
In principle a "nasty" workflow could return anything from its ancestry, not just direct inputs. That's probably a very rare occurrence though, and I agree we can probably get caching working for the majority of realistic workflows.

Heh, if (once?) we have data-deduplication I think the argument for return links would be much weaker ;)

sphuber:
> The problem with caching in workflows is that they can return something which belongs to their "ancestry". When called again with the same inputs, the graph relationships between ancestors are not necessarily the same -- so it's unclear which ancestor the return link should point to. For calculations this problem doesn't exist because outputs are always new nodes.

I remember you provided the following example to illustrate this point, a workflow that takes two inputs and returns one of them:
```
       ____   ____
      |    | |    |
      | D1 | | D2 |
      |____| |____|
           |  |  ^
     input |  |  |
      	   v__v  | return
      	  |    |_|
      	  | W1 |
      	  |____|
```
Assuming `D1` and `D2` have different hashes, for simplicity let's say `D1 = Int(1)` and `D2 = Int(2)`, wouldn't it be possible to deterministically decide how to link up the cached node. For example:
```
d3 = Int(1)
d4 = Int(2)
test_work_function(d3, d4)
```
The hash of the to be created workflow node `W2` would be identical to that of `W1` and so it can be cached and so the resulting graph would look like:
```
       ____   ____
      |    | |    |
      | D3 | | D4 |
      |____| |____|
           |  |  ^
     input |  |  |
      	   v__v  | return
      	  |    |_|
      	  | W2 |
      	  |____| 
```
Now, the know that the return link _has_ to point to `D4` and not `D3` based on the hashes of the inputs nodes. Of course this breaks down if the hash of `D1` and `D2` happen to be equal, but wouldn't it be arbitrary which one would get the link. Either choice would be semantically identical, no?

aiida/orm/nodes/node.py
greschd: One could argue that simply ignoring the argument is still a change in interface, since the semantics of the code changed.

sphuber:
> Heh, if (once?) we have data-deduplication I think the argument for return links would be much weaker ;)

This was never part of the argument of adding return links. The main argument was that when having heavily nested provenance graphs, trying to access an output node of a calculation deep in the hierarchy, when starting from the top-level workflow node is not obvious. Take the `PwBandStructureWorkChain` for example. Imagine you have its pk and you want to get the final `BandsData` or any other of the results that it currently returns, but now without the existence of return links. It will not be easy to find them. Also here, it being a "turn-key" solution, the typical user won't care about the nested provenance, they just want to see the collapsed top level, i.e. the workflow node with its inputs and outputs (that are returned). I don't see how you can achieve this without returning. Of course you could have some concept where the workflow node simply "clones" the outputs of the calculations it calls to add a shortcut, but this seems like the wrong solution to me.

aiida/orm/nodes/node.py
sphuber: You are absolutely right, except in this case I think it is acceptable because the resulting behavior is identical to that before the change (with the exception that `Data` nodes stored with `use_cache=True` in the past would have gotten the `_aiida_cached_from` extra, which is no longer the case, but this extra is meaningless anyway in this scenario).

greschd:
On the workflow caching: the problem is that a workflow can not only return its direct input, but really _anything_.

A (somewhat contrived) example would be a workflow which takes several CalcJobNodes as input, somehow selects the best, and then returns all outputs of _that_ calculation. Now you somehow need to do graph matching between the ancestors to figure out what the new workflow should return.

So in the most general case it's probably impossible to replicate what the workflow would have done if it wasn't cached. That just means that we need to define for which "restricted" workflows we want to make it work.

On return links, I see your point there. Also, it's really not something we should probably touch in the near to mid-term future.

sphuber:
> A (somewhat contrived) example would be a workflow which takes several CalcJobNodes as input,

Does this counter-example rely on the inputs being `CalcJobNodes`? Because that is illegal anyway, only `Data` nodes can be inputs.

greschd:
I don't think it does, you just have to get a bit more creative in how you get a reference to an ancestor that isn't also an input. I can have that be outputs of different calculations, and then the optimal input is returned.

Is that good workflow design? Probably not. But since it's possible, we should make sure the caching system doesn't crash and burn when someone does that.

For example, we could restrict caching to workflows where the output is a "closed" sub-tree, i.e. doesn't have any links higher than the workflow itself. Re-creating that would still be more complex than for simple calculations (since you might need to follow multiple layers of create links), but should in principle be possible.

ltalirz:
> This was never part of the argument of adding return links. The main argument was that when having heavily nested provenance graphs, trying to access an output node of a calculation deep in the hierarchy, when starting from the top-level workflow node is not obvious.

I fully agree that the "explicit lineage" that the return link provides can be valuable.
To me, the important question is how much one is willing to pay for this in terms of complicating the provenance model.
I somehow vaguely remember Martin mentioning during the discussions as one argument in favor of this approach that most programming languages also have the `return` - but in fact, in the programming languages that I know the `return` is more closely related to our CREATE link in the sense that there is no explicit connection made from some input of a function to its return value  (but perhaps I misunderstood this argument in the first place).

> Of course you could have some concept where the workflow node simply "clones" the outputs of the calculations it calls to add a shortcut, but this seems like the wrong solution to me.

I wouldn't call it wrong. It's different and it contains less explicit information.
One could still partially recover information by having some graphical way of highlighting/connecting nodes within a workflow that have the same hash, but of course you cannot recover the information provided by RETURN links.

What I'm mainly uncomfortable with is the somewhat arbitrary distinction between "logical provenance" (=workflows) and "data provenance" (=calculations), which I find a bit difficult to explain to new users (and to myself).
In my view, there is no fundamental difference between a workflow that decides to return one out of many inputs and a calculation that computes a new result (in principle, you could replace any calculation by a workflow that takes all possible results as input and returns the right one).

However, I agree that the distinction between workflows and calculations *is* sensible in practice for our typical use cases, and that the current model is practical.

>  Also, it's really not something we should probably touch in the near to mid-term future.

Indeed. Sorry, I did not intend this thread to explode ;-)

ltalirz:
I guess this PR is good to merge, right?

aiida/orm/nodes/node.py
greschd: Yeah, true -- since ``_aiida_cached_from`` is an internal extra, there's no public-facing change.

greschd:
Yes :)

sphuber:
Thanks for the input guys. @greschd for future reference, for single commit PRs we have been using squash-and-merge to prevent the pointless merge commits.

greschd:
Ok, I'll keep that in mind :+1:

PR #3330, opened 2019-09-19
Allow empty `enabled` and `disabled` keys in caching configuration

### LINKED ISSUE #3329: Node store method excepts for invalid caching configuration files
For a configuration file as follows:
```
django:
    default: False
    enabled:
        # - aiida.calculations:quantumespresso.pw
    disabled:
        # - aiida.calculations:templatereplacer
```
the `Node.store()` method will except because the caching configuration parsing in `aiida.manage.caching` will except. It does not deal with the fact that the `enabled` and `disabled` keys can be empty.

sphuber:
Fixes #3329 

If the `enabled` or `disabled` key is specified in the caching
configuration but contains no values, or all values are commented, the
parsed dictionary will have `None` as the value for the corresponding
key. This was causing an exception in`aiida.manage.caching._get_config`
because it was always iterating over it without checking that it exactly
contained an iterable.


sphuber:
Don't merge yet. I just might have found another issue

sphuber:
Thanks @giovannipizzi for the approval but just found another issue. The last commit adds a test for it and solves it.

PR #3335, opened 2019-09-24
reentry scan missing from installation docs

### LINKED ISSUE #3337: Describe/explain `reentry scan` in the Docs
`reentry scan` is a useful feature to be aware of when installing and updating AiiDA. Furthermore, when installing, updating or developing plugins it is essential knowledge.

It should have some sentences in the documentation.

ltalirz:
Closes #3337


ltalirz:
> I would prefer a bit of an explanation of what it is, if it doesn't already exist.

Done.

> could you add it to the quick installation documentation as well?

Done

PR #3344, opened 2019-09-24
Fix unspecific flush in `DjangoNode` causing data override

### LINKED ISSUE #3338: setting extra on finished calculation seems to put them into `CREATED` state
setting extra on finished CalcJobNode seems to put them into `CREATED` state

Encountered during SINTEF tutorial
@sphuber will edit & provide more details

sphuber:
Fixes #3338 

The methods of `DjangoNode` related to the mutation of attributes and
extras were indirectly causing a flush of the entire node state instead
of flushing only the property they were mutating. For example, changing
the extras, would also cause the attributes to be flushed. This could
lead to problems, if the attributes had been changed in another instance
and flushed to the database, because the original instance will not
refresh the state of the attributes when flushing its state, these
changes will be overridden.

The `SqlaNode` properly avoided this problem by "flagging" the changed
property before calling the flush. This is done by a SqlAlchemy utility
function which is not present for Django. Instead we pipe this through
to the `DbNode.save` method which supports the `updated_fields` argument
to target only specific model fields to be flushed.

### DIFF
--- a/aiida/backends/tests/orm/implementation/test_nodes.py
+++ b/aiida/backends/tests/orm/implementation/test_nodes.py
@@ -504,6 +504,28 @@ class TestBackendNode(AiidaTestCase):
         node.set_attribute_many(attributes)
         self.assertEqual(set(attributes), set(node.attributes_keys()))
 
+    def test_attribute_flush_specifically(self):
+        """Test that changing `attributes` only flushes that property and does not affect others like extras.
+
+        Regression fix for #3338
+        """
+        node = self.create_node().store()
+        extras = {'extra_one': 1, 'extra_two': 2}
+        node.set_extra_many(extras)
+        node.store()
+
+        # Load the stored node in memory in another instance and add another extra
+        reloaded = self.backend.nodes.get(node.pk)
+        reloaded.set_extra('extra_three', 3)
+
+        # The original memory instance should now not include `extra_three`. Changing an attribute should only flush
+        # the attributes and not the whole node state, which would override for example also the extras.
+        node.set_attribute('attribute_one', 1)
+
+        # Reload the node yet again and verify that the `extra_three` extra is still there
+        rereloaded = self.backend.nodes.get(node.pk)
+        self.assertIn('extra_three', rereloaded.extras.keys())
+
     def test_extras(self):
         """Test the `BackendNode.extras` property."""
         node = self.create_node()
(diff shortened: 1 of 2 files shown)


sphuber:
This does in fact fix the problem, but I think there might still be an important weakness in the `BackendNode` and `ModelWrapper` design. I wrote a test that exposed the bug but it only directly tests the `set_extra` method to trigger it. In principle it could be done by any of the other `attribute/extra` mutation methods. They now all currently flush only the property they are changing but in the future we might add another method and use the `ModelWrapper.save` method instead of the `_flush` method. This is identical to forgetting to call the `_flag_modified` line in the `SqlaNode` implementation. I am wondering then if it wouldn't be better to _only_ expose a single method (whether we call it `save` or `flush`) that _requires_ to specify the fields to be flushed, instead of making it an option. On the other hand, there might be use cases where one changes various properties without flushing and then wants to flush in one go, for efficiency purposes, in which case having a generic `save` that flushes the whole node state, would be very useful if not necessary.

sphuber:
> That is, that if both instances change a (different) attribute, or a different extra, the changes from the other instance are equally lost. I.e., changes to Json fields act as a whole and not at the level of single sub keys.

This is exactly not the case, as I think the following test (which passes) demonstrates:
```


    def test_set_attribute_override(self):
        """Test the `BackendNode.set_attribute` method."""
        node = self.create_node()

        node.set_attribute('attribute_one', 1)
        node.store()

        reloaded = self.backend.nodes.get(node.pk)
        reloaded.set_attribute('attribute_two', 2)

        self.assertNotEqual(id(node.dbmodel), id(reloaded.dbmodel))

        node.set_attribute('attribute_three', 3)
        self.assertIn('attribute_two', node.attributes)
```
The reason is that in the `set_attribute` method uses `self._dbmodel.attributes[key] = value` to set the new attribute. The `self._dbmodel.attributes` will be evaluated first which goes through the model wrapper, which triggers the refreshing of the state from the database. This ensures that the changes of another instance are correctly "merged" with the additional change. This is difficult to see from the code and should maybe be commented. But then this should be commented in many places. Again, this might be another potential weakness of the current design that this kind of critical logic is relatively obscure.

giovannipizzi:
I see. Good that at least from the same interpreter this is not going to happen, if we don't use multithreading.

However note that in general the pattern "refesh data + modify + push everything back" is problematic if we use multithreading (but luckily we don't) and in multiple processes (e.g. if extras are written at the same time by two CLIs, or by the CLI and the daemon).

I know this might seem low probability (and it's harder to test, one needs probably to run a `subprocess.check_output` to run in a different unix process during the tests) but actually it might happen (e.g. if you have a bash loop to update a lot of extras, and at the same time you are working in a shell)...

PR #3348, opened 2019-09-25
Catch exceptions raised by `aiida.engine.daemon.runner.start_daemon`

### LINKED ISSUE #3333: Daemon runners launched by Circus could except
This was encountered when preparing the AWS machine for the SINTEF tutorial starting from the Xiamen image. The daemon, started automatically upon boot through systemd, worked just fine. However, when restarting it, the new daemon appeared to be unresponsive, even though Circus seemed to be working fine and it reported a running worker through `verdi daemon status`. The problem was that the worker was dying as soon as it was launched and Circus was just reviving it, giving the impression one was alive at all times.

To show the problem one should run the Circus daemon directly and in the foreground, through the command `verdi daemon start-circus --foreground`. Note that this is a hidden command and so won't be auto-completed. Once doing this, the problem becomes obvious as the daemon runner excepts:
```
SyntaxError: invalid syntax
  File "/home/max/.virtualenvs/aiida/bin/verdi", line 6, in <module>
  File "/home/max/codes/aiida-core/aiida/cmdline/commands/__init__.py", line 21, in <module>
    from aiida.cmdline.commands import (
  File "/home/max/codes/aiida-core/aiida/cmdline/commands/cmd_process.py", line 18, in <module>
    from kiwipy import communications
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/kiwipy/__init__.py", line 5, in <module>
    from .communications import *
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/kiwipy/communications.py", line 3, in <module>
    import concurrent.futures
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/concurrent/futures/__init__.py", line 8, in <module>
    from concurrent.futures._base import (FIRST_COMPLETED,
  File "/home/max/.virtualenvs/aiida/lib/python3.6/site-packages/concurrent/futures/_base.py", line 381
    raise exception_type, self._exception, self._traceback
                        ^
SyntaxError: invalid syntax
```

This was with Ubuntu 18.04 and `aiida-core==1.0.0b6` and it is not yet clear why this exception does not occur when the daemon is launched through systemd. Besides finding the origin to this problem and fixing it, we should see if we can change the configuration of Circus such that these exceptions in the daemon runners also make it to the Circus log, which currently does not happen.

sphuber:
Fixes #3333 

Exceptions in this call, performed by the Circus daemon when launching a
new daemon worker, were not being caught. This problem would then go
unnoticed as Circus would silently keep reviving the daemon process as
soon as it died due to the uncaught exception. The exception would not
make its way into the daemon log either, so unless one would notice that
the daemon runner pid was changing constantly, it would seem like the
daemon had a worker running even when it was not really. By catching the
exception and echoing it to stdout, it will now show up in the daemon
log file.

### DIFF
--- a/aiida/cmdline/commands/cmd_devel.py
+++ b/aiida/cmdline/commands/cmd_devel.py
@@ -11,6 +11,7 @@
 from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
+
 import click
 
 from aiida.cmdline.commands.cmd_verdi import verdi
--- a/aiida/cmdline/commands/cmd_daemon.py
+++ b/aiida/cmdline/commands/cmd_daemon.py
@@ -63,7 +63,7 @@ def start(foreground, number):
 
     try:
         currenv = get_env_with_venv_bin()
-        subprocess.check_output(command, env=currenv, stderr=subprocess.STDOUT)
+        subprocess.check_output(command, env=currenv, stderr=subprocess.STDOUT)  # pylint: disable=unexpected-keyword-arg
     except subprocess.CalledProcessError as exception:
         click.secho('FAILED', fg='red', bold=True)
         echo.echo_critical(exception.output)
@@ -262,6 +262,10 @@ def start_circus(foreground, number):
             'virtualenv': client.virtualenv,
             'copy_env': True,
             'stdout_stream': {
+                'class': 'FileStream',
+                'filename': client.daemon_log_file,
+            },
+            'stderr_stream': {
                 'class': 'FileStream',
                 'filename': client.daemon_log_file,
             },
--- a/aiida/engine/daemon/runner.py
+++ b/aiida/engine/daemon/runner.py
@@ -23,9 +23,7 @@ LOGGER = logging.getLogger(__name__)
 
 
 def start_daemon():
-    """
-    Start a daemon runner for the currently configured profile
-    """
+    """Start a daemon runner for the currently configured profile."""
     daemon_client = get_daemon_client()
     configure_logging(daemon=True, daemon_log_file=daemon_client.daemon_log_file)
 


sphuber:
I am not sure how the Circus arbiter can detect this exception and stop or, the other way around, how the `verdi devel run_daemon` call can stop the parent that is calling it. In the `verdi daemon start-circus` command where we start the `Arbiter` which controls the `Watcher` which in turn launches the daemon runners takes a `max_retry` which I tried setting but it does not seem to respect this argument.

sphuber:
I improved the PR a bit adding a timestamp to the log and setting a cooldown between worker spawns. Without it, it was running one each second. I now delayed that to 10 seconds. Funnily enough, this key `warmup_delay` _is_ respected. I have been trying to figure out why the `max_retry` is ignored. What's more, the default setting for that is `5`, so even the default should have not launched indefinitely. To do that one needs to set `-1`. But in reality it was actually launching indefinitely. Very weird.

sphuber:
The `warmup_delay` was clashing with a bunch of other code, for example tests timing out that rely on a daemon starting and the daemon commands themselves. There is a timeout timer there as well, that is not yet configurable but hard coded. So for the time being I removed it from this PR

sphuber:
I tried to understand this, but honestly I don't know. Note that the logging has potentially not yet been configured as that happens in `start_daemon`. If the exception happens on import of that file, we cannot rely on configuration to the daemon file, through the file handler, to be setup. I am not sure what is going on exactly, but when printing to stderr, it won't end up in the daemon log file. I think because it relies on the basic console handler that only redirects stdout, but I am not sure

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -61,8 +64,16 @@ def get_valid_test_paths():
 @decorators.with_dbenv()
 def devel_run_daemon():
     """Run a daemon instance in the current interpreter."""
-    from aiida.engine.daemon.runner import start_daemon
-    start_daemon()
+    try:
+        # Keep the import also within the try-except such that if the import itself excepts, it is also caught
+        from aiida.engine.daemon.runner import start_daemon
```
ltalirz: sorry for the nit-picking but what happens if the import here excepts?
wouldn't that be handled properly anyhow, i.e. wouldn't the exception appear in the regular way?

sphuber: No, it would have the same problem. This verdi command is what is launched by circus in a different sub process. If that sub process excepts before the daemon logging handling is configured (which is what happens in the `aiida.engine.daemon.runner.start_daemon` call, the exception that will be raised will never end up in the log file and so the exception will never be seen by the user.

sphuber:
Changed the solution. The actual problem was just that the Circus watchers were only forwarding stdout of the daemon runner processes to the daemon log file. Simply forwarding the `stderr` as well solves the issue

PR #3350, opened 2019-09-26
`ProcessLauncher`: except processes whose class cannot be imported

### LINKED ISSUE #3347: Except submitted processes whose class cannot be imported
Currently, the process will simply except and acknowledge the task, but the node will remain in the `CREATED` state.

sphuber:
Fixes #3347 

The `ProcessLauncher` would hit an exception if the class of the process
it is continuing could not be imported. The `plumpy.DefaultObjectLoader`
would throw a `ValueError` in this case which was not caught and would
simply except the future. The process task would be acknowledged and so
would not pose a problem, but the corresponding process node would not
be properly wrapped up and would keep the `CREATED` process state for
ever. The exception is now properly caught and the node is cleaned up
before being sealed.

Note that normally this is done by the process engine, but since the
exception happens before the process instance can be recreated this
pathway is never hit and we have to clean up the node manually.

### DIFF
--- a/aiida/manage/external/rmq.py
+++ b/aiida/manage/external/rmq.py
@@ -138,6 +138,7 @@ class ProcessLauncher(plumpy.ProcessLauncher):
         :param tag: the tag of the checkpoint to continue from
         """
         from aiida.common import exceptions
+        from aiida.engine import ProcessState
         from aiida.engine.exceptions import PastException
         from aiida.orm import load_node, Data
         from aiida.orm.utils import serialize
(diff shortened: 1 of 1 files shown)


sphuber:
Note I tried since yesterday to write a test for this but I couldn't figure it out. I tried it as a unit test, like we do in `aiida.backends.cmdline.commands.cmd_process` for example where we also spawn a daemon in the background, but for a workchain defined in the same file, it does get submitted but not picked up. I also tried it as a stand alone script and adding it to `test_daemon.py` but there it is not clear how to make a work chain class that _is_ importable in the launch script and submittable, but _not_ importable by the daemon

ltalirz:
> I also tried it as a stand alone script and adding it to test_daemon.py but there it is not clear how to make a work chain class that is importable in the launch script and submittable, but not importable by the daemon

How about simply modifying `sys.path` in your run script?
I.e. you put your workchain inside `hidden/workchain.py` and in your script you do

```python
sys.path.insert(0,'/path/to/hidden')
from workchain import MyWorkChain
run(MyWorkChain, ...)
sys.path.pop(0)
```

For an even cleaner method, see here:
https://github.com/aiidateam/aiida-core/blob/6299cf6c961f27c64c0a146130f52adafcba9d96/utils/validate_consistency.py#L189-L196

aiida/manage/external/rmq.py
```diff
@@ -168,7 +169,14 @@ def _continue(self, communicator, pid, nowait, tag=None):
 
             raise gen.Return(future.result())
 
-        result = yield super(ProcessLauncher, self)._continue(communicator, pid, nowait, tag)
+        try:
+            result = yield super(ProcessLauncher, self)._continue(communicator, pid, nowait, tag)
+        except ValueError as exception:
```
ltalirz: One more question: You are only catching `ValueError` here.
What if other exceptions happen in the call above?
Shouldn't the process node always be taken care of?

sphuber:
> How about simply modifying `sys.path` in your run script?
> I.e. you put your workchain inside `hidden/workchain.py` and in your script you do
>

This would indeed make it possible to launch a workchain that is submittable but then won't be loadable by the daemon. The next problem then is how to actually perform the test. It would have to call the script to submit it, somehow get the pk of the node that was submitted and wait for it to go in a terminal state (or timeout) and then check that it is excepted. Not sure how to do this in a way that is not super fragile

aiida/manage/external/rmq.py
sphuber: You are right, I had to update the logic and extended the comment with explanation

PR #3354, opened 2019-09-26
Multiple RETURN links import now possible

### LINKED ISSUE #3353: Import multiple return links with non-unique labels
@giovannipizzi and @AntimoMarrazzo found a bug when importing first the data provenance and then the whole provenance of the same data, where, if the link label for a `create` and `return` link (to the same `Data` node) has the same label, an error is raised - which is wrong, since multiple `return` links are okay.

CasperWA:
Fixes #3353

### DIFF
--- a/aiida/tools/importexport/dbimport/backends/django/__init__.py
+++ b/aiida/tools/importexport/dbimport/backends/django/__init__.py
@@ -503,28 +503,28 @@ def import_data_dj(
 
             if not silent:
                 print('STORING NODE LINKS...')
-            ## TODO: check that we are not creating input links of an already
-            ##       existing node...
             import_links = data['links_uuid']
             links_to_store = []
 
             # Needed, since QueryBuilder does not yet work for recently saved Nodes
             existing_links_raw = models.DbLink.objects.all().values_list('input', 'output', 'label', 'type')
-            existing_links = {(l[0], l[1], l[2]): l[3] for l in existing_links_raw}
-            existing_output_links = {(l[0], l[3]): (l[1], l[2]) for l in existing_links_raw}
-            existing_input_links = {(l[1], l[3]): (l[0], l[2]) for l in existing_links_raw}
+            existing_links = {(l[0], l[1], l[2], l[3]) for l in existing_links_raw}
+            existing_outgoing_unique = {(l[0], l[3]) for l in existing_links_raw}
+            existing_outgoing_unique_pair = {(l[0], l[2], l[3]) for l in existing_links_raw}
+            existing_incoming_unique = {(l[1], l[3]) for l in existing_links_raw}
+            existing_incoming_unique_pair = {(l[1], l[2], l[3]) for l in existing_links_raw}
 
-            CalculationNode = {'CalculationNode', 'CalcJobNode', 'CalcFunctionNode'}  # pylint: disable=invalid-name
-            WorkflowNode = {'WorkflowNode', 'WorkChainNode', 'WorkFunctionNode'}  # pylint: disable=invalid-name
-            Data = {'data'}  # pylint: disable=invalid-name
+            calculation_node_types = 'process.calculation.'
+            workflow_node_types = 'process.workflow.'
+            data_node_types = 'data.'
 
             link_mapping = {
-                LinkType.CALL_CALC: (WorkflowNode, CalculationNode, 'unique_triple', 'unique'),
-                LinkType.CALL_WORK: (WorkflowNode, WorkflowNode, 'unique_triple', 'unique'),
-                LinkType.CREATE: (CalculationNode, Data, 'unique_pair', 'unique'),
-                LinkType.INPUT_CALC: (Data, CalculationNode, 'unique_triple', 'unique_pair'),
-                LinkType.INPUT_WORK: (Data, WorkflowNode, 'unique_triple', 'unique_pair'),
-                LinkType.RETURN: (WorkflowNode, Data, 'unique_pair', 'unique_triple'),
+                LinkType.CALL_CALC: (workflow_node_types, calculation_node_types, 'unique_triple', 'unique'),
+                LinkType.CALL_WORK: (workflow_node_types, workflow_node_types, 'unique_triple', 'unique'),
+                LinkType.CREATE: (calculation_node_types, data_node_types, 'unique_pair', 'unique'),
+                LinkType.INPUT_CALC: (data_node_types, calculation_node_types, 'unique_triple', 'unique_pair'),
+                LinkType.INPUT_WORK: (data_node_types, workflow_node_types, 'unique_triple', 'unique_pair'),
+                LinkType.RETURN: (workflow_node_types, data_node_types, 'unique_pair', 'unique_triple'),
             }
 
             for link in import_links:
@@ -545,8 +545,7 @@ def import_data_dj(
 
                 # Check if link already exists, skip if it does
                 # This is equivalent to an existing triple link (i.e. unique_triple from below)
-                existing_type = existing_links.get((in_id, out_id, link['label']), None)
-                if existing_type and existing_type == link['type']:
+                if (in_id, out_id, link['label'], link['type']) in existing_links:
                     continue
 
                 # Since backend specific Links (DbLink) are not validated upon creation, we will now validate them.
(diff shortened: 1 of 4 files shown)


aiida/tools/importexport/dbimport/backends/django/__init__.py
```diff
@@ -559,24 +555,20 @@ def import_data_dj(
                         # is the case of the RETURN links of workflows/
                         # workchains. If it is not this case, then it is
                         # an error.
+                        from aiida.common.links import LinkType
                         existing_input = existing_input_links[out_id, link['label']]
 
-                        if link['type'] != LinkType.RETURN:
+                        if link['type'] != LinkType.RETURN.value:
```
giovannipizzi: I'm not sure this is the correct logic (in the sense that there are other cases to take into account - e.g. CALL links).
I think we should use the logic that @sphuber implemented to check the validity and uniqueness constraints of link triples - @sphuber what do you think?

sphuber: I have not looked at the issue nor the solution in this PR, but yes, in general, the import code should not try and do the link validation itself. This is bound to become out of sync with the actual ORM implementation. I hope the import is done in a transaction, so I suggest to just try to import the nodes through the ORM and if a link is invalid it will raise the validation error in `Node.add_incoming` coming from `aiida.orm.utils.link.validate_link`. This can be caught and a more useful message can be formatted and printed. The only problem I can see is that going through the front-end ORM may be too slow for bulk imports.

CasperWA: Makes sense. However, the biggest problem with export/import is how to make it general and keep up the speed (as hinted by @sphuber).
I suggest making another issue for this, since testing is necessary for the rewrite, which will take some time. The current quick-fix can then be reviewed as is, it should (according to the new test) fix the issue at hand, and the proper workflow can be implemented in a later PR?

giovannipizzi: Probably this fix is enough for now for 1.0.
Maybe indeed this fix is critical for 1.0, and we don't have time to benchmark a better solution before 1.0.
So unless we see other problems with this PRI, I would be in favour of this temporary fix, adding a comment also in the code of why this has to be changed, and opening another issue

sphuber: Does this fix also now allow the scenario where a single `Data` node has two incoming `RETURN` links with the same label but from two different `WorkflowNodes`?

CasperWA: > Does this fix also now allow the scenario where a single `Data` node has two incoming `RETURN` links with the same label but from two different `WorkflowNodes`?

I believe this is caught in the last `except KeyError`, where a new link is created .... Never mind, it isn't. It should be added as an `else` to the `if`-sentence that makes sure the link type is not a RETURN link. That should do it, I believe?

sphuber: Looking at the code it also seems that it is not even considering `CALL` links whatsoever. There may very well be quite some bugs with those as well.

sphuber:
So if I understand correctly, the only real change was that the if-condition was comparing to `LinkType.RETURN` instead of `LinkType.RETURN.value` and therefore was always false, causing an unnecessary exception being raised.

Unfortunately, looking at the current code, there are a lot of problems that really should be addressed sooner rather than later. The whole point of #1762 was to get to a state where we know what we export and import, but seeing this code I am afraid we do not actually have much of an idea. But I agree that we cannot tackle that right now. So for me also fine to merge this now

CasperWA:
> So if I understand correctly, the only real change was that the if-condition was comparing to `LinkType.RETURN` instead of `LinkType.RETURN.value` and therefore was always false, causing an unnecessary exception being raised. Is that correct @CasperWA ?

This is correct.

> Unfortunately, looking at the current code, there are a lot of problems that really should be addressed sooner rather than later. The whole point of #1762 was to get to a state where we know what we export and import, but seeing this code I am afraid we do not actually have much of an idea. But I agree that we cannot tackle that right now. So for me also fine to merge this now

Yeah, I have been trying slowly to take out each step of the export and import functions, where possible, trying to create backend-agnostic utility functions. This should hopefully make each step of the process easier to deal with and understand, however, it is taking forever to do.

aiida/tools/importexport/dbimport/backends/django/__init__.py
CasperWA: Again, to make it "function" somewhat, I can quick-fix it, however, the more I do that, the more I feel that we should simply bite the bullet and go through the ORM properly - making sure that this works properly and then prioritize optimization after 1.0?

sphuber:
I understand and already your work has made the organization of the code and the code itself a lot better, but eventually small steps won't solve this particular problem. At some point we will have to face the music and properly design an interface for export/import that matches ORM implementation but that allows for efficient bulk inserts. This is a big challenge, but is the only way to once and for all fix it.

CasperWA:
I completely agree.

aiida/tools/importexport/dbimport/backends/django/__init__.py
CasperWA: Concerning CALL links, this should be a check for existing output links, right? For RETURN links, the check is for existing input links.

sphuber: If we make certain assumptions about what can and can't be exported, then yes, but I think we should not make those assumptions. The only thing we should be checking is if adding the new link does not violate the rules. For the time being though, I guess one can only have the situation of additional incoming `CALL` links for an already existing node. The reverse, where an existing node, gets a new outgoing `CALL` link should not be possible.

CasperWA: After discussing offline, we have found that the best thing to do now, will be to remove the existing logic and implement a "dumb" logic, i.e., go through each LinkType in turn and deal with them separately.
In this way, we make sure that everything works as intended for _this_ version.

CasperWA:
@sphuber @giovannipizzi this is ready for review. I have re-implemented the `validate_link` function from `aiida.orm.utils.links`. Note that for "unique_triple" Links, this is essentially similar to stating "this Link already exists", so if the util function `link_triple_exists` returns `True`, the import simply skips the Link in question instead of raising.

It is not the fastest, nor is it the prettiest fix, but it should be more rigorous in testing the created Links from import.

aiida/tools/importexport/dbimport/backends/django/__init__.py
```diff
@@ -519,68 +519,110 @@ def import_data_dj(
...
+
+                if source.uuid == target.uuid:
+                    raise exceptions.ImportValidationError('Cannot add a link to oneself')
+
+                link_type = LinkType(link['type'])
+                types_source, types_target, outdegree, indegree = link_mapping[link_type]
+
+                # Check if source Node is a valid type
+                for type_source in types_source:
+                    if type_source in source.node_type:
```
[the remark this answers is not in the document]
CasperWA: This doesn't really work for `Data` Nodes, so I will use `process.calculation.`, `process.workflow.`, `data.` and `startswith` instead.

aiida/tools/importexport/dbimport/backends/django/__init__.py
```diff
@@ -519,68 +519,110 @@ def import_data_dj(
...
+                # If the indegree is `unique` there cannot already be any other incoming links of that type
+                if indegree == 'unique' and (out_id, link['type']) in existing_input_links:
+                    raise exceptions.ImportValidationError(
+                        'Node<{}> already has an incoming {} link'.format(target.uuid, link_type)
+                    )
+
+                # If the indegree is `unique_pair`,
+                # then the link labels for incoming links of this type should be unique
+                elif indegree == 'unique_pair' and (out_id, link['type']) in existing_input_links:
+                    (_, label) = existing_input_links[in_id, link['type']]
```
CasperWA: should've been `out_id`, not `in_id` here

CasperWA:
I got insecure whether the ad hoc rules would actually work, feeling there is something missed between the usage of `.get_outgoing` and `.get_incoming` in the actual `validate_links()` function.
It seemed to me that if the Link type and label was the same for an existing incoming Link for a Node and we are only concerned with the existing outgoing Links being similar (possible only for `CALL_WORK`, I think), then the ad hoc rules here would not catch it.

However, when writing the test, I came to the realization that there would never be a situation where this is a possible case, due to a combination of the fixed Link rule to follow `CALL_WORK` forward and the fact that _only_ sealed Nodes may be exported.
I.e., a graph:
work1 -> work2 -> work3
cannot be exported, following an export of 
work1 -> work2 -> work3
                   |
                   +-> work4
with the same Node UUIDS, since they must be sealed - which means the Link between work2 and work4 cannot be created after export of the first graph, and the sub-graph equal to the first graph cannot be exported from the second graph exists due to the Link follow rules (non-changeable `call_work_forward=True`, changeable `call_work_backward=False`).

aiida/tools/importexport/dbimport/backends/django/__init__.py
```diff
@@ -562,48 +561,57 @@ def import_data_dj(
...
+                elif indegree == 'unique_pair' and \
+                (out_id, link['label'], link['type']) in existing_incoming_unique_pair:
+                    raise exceptions.ImportValidationError(
+                        'Node<{}> already has an incoming {} link with label "{}"'.format(
+                            target.uuid, link_type, link['label']
                         )
+                    )
 
                 # New link
                 links_to_store.append(
```
sphuber: The link checking itself now seems correct, very nice. There is just one thing we realized. In this logic we should not rely on export logic, i.e. what export _can_ legally export. So for example, we should not assume that the contents of an archive are necessarily consistent. This means we have to check that the links _within_ an archive are also consistent. To do this, we simply have to update the `existing_` sets at this point. Then in the next iterations, the added link will be included in the existence check and if the archive itself contains duplicate links, it will be noticed.

CasperWA: Done. I have done the same for the SQLAlchemy backend, however, the only way to do this was to not "bulk create" the Links after the for loop, but instead add the new Link in every run using `session.add(Link)`.

CasperWA: Btw, also for the SQLAlchemy import, I did a check that the new Links are correctly continuously added and can be found using the QueryBuilder in the subsequent run of the for loop.

sphuber:
Thanks a million @CasperWA great stuff ! :+1:

PR #3359, opened 2019-09-27
Allow passing directory path in `FolderData` constructor

### LINKED ISSUE #3357: FolderData should take folder path as constructor argument
We discovered during the SINTEF tutorial that it is quite difficult to figure out how to create a `FolderData` from an existing folder (I figured it out by searching the aiida-core source code, which is not something a regular user can do).

 * When creating `fd = FolderData` in verdi shell and doing `fd.` + <tab> you see lots of methods, but the relevant one ( `.put_object_from_tree`)  was not even visible (you would have to scroll to the end)
 * When you just want to create a FolderData to wrap a folder on your file system, "put_object_from_tree" is maybe not what you are thinking...

In my opinion, the `FolderData` should have a constructor argument `folder` to which you can pass an absolute path to a folder that will make the `FolderData` wrap this folder.
I think this would cover the majority of cases with a simple user interface.

Mentioning @sphuber and @giovannipizzi for comment

sphuber:
Fixes #3357 

This will initialize the node and copy over the tree pointed to by the
`tree` keyword argument into the node's repository.

### DIFF
--- a/aiida/orm/nodes/data/folder.py
+++ b/aiida/orm/nodes/data/folder.py
@@ -19,3 +19,22 @@ __all__ = ('FolderData',)
 
 class FolderData(Data):
     """`Data` sub class to represent a folder on a file system."""
+
+    def __init__(self, **kwargs):
+        """Construct a new `FolderData` to which any files and folders can be added.
+
+        To add files to a new node use the various repository methods:
+
+            folder = FolderData()
+            folder.put_object_from_tree('/absolute/path/to/directory')
+            folder.put_object_from_filepath('/absolute/path/to/file.txt')
+            folder.put_object_from_filelike(filelike_object)
+
+        Alternatively, in order to simply wrap a directory, the `path` keyword can be used in the constructor:
+
+            folder = FolderData(tree='/absolute/path/to/directory')
+        """
+        tree = kwargs.pop('tree', None)
+        super(FolderData, self).__init__(**kwargs)
+        if tree:
+            self.put_object_from_tree(tree)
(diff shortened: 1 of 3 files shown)


aiida/orm/nodes/data/folder.py
```diff
@@ -19,3 +19,22 @@
...
+        To add files to a new node use the various repository methods:
+
+            folder = FolderData()
+            folder.put_object_from_tree('/absolute/path/to/directory')
+            folder.put_object_from_filepath('/absolute/path/to/file.txt')
+            folder.put_object_from_filelike(filelike_object)
+
+        Alternatively, in order to simply wrap a directory, the `path` keyword can be used in the constructor:
+
+            folder = FolderData(tree='/absolute/path/to/directory')
```
ltalirz: Would it make sense to add an explicit documentation of the `tree` argument as well?
```
    :param tree: Absolute path to a folder to wrap.
    :type tree: str
```

aiida/orm/nodes/data/folder.py
```diff
@@ -19,3 +19,22 @@
 
 class FolderData(Data):
     """`Data` sub class to represent a folder on a file system."""
+
+    def __init__(self, **kwargs):
+        """Construct a new `FolderData` to which any files and folders can be added.
```
ltalirz:
```suggestion
        """Construct a new `FolderData` to which files and folders can be added.
```

aiida/orm/nodes/data/folder.py
```diff
@@ -19,3 +19,22 @@
...
+
+        To add files to a new node use the various repository methods:
+
+            folder = FolderData()
+            folder.put_object_from_tree('/absolute/path/to/directory')
+            folder.put_object_from_filepath('/absolute/path/to/file.txt')
+            folder.put_object_from_filelike(filelike_object)
+
+        Alternatively, in order to simply wrap a directory, the `path` keyword can be used in the constructor:
```
ltalirz: I would put this first, and then mention the other methods.
```suggestion
        Use the `tree` keyword in order to simply wrap a directory:
```

PR #3362, opened 2019-09-28
Enable auto-completion for the `ComputerParamType`

### LINKED ISSUE #3361: Enable auto-completion for `ComputerParamType`

sphuber:
Fixes #3361 

This will allow CLI commands that expose an option or argument that
takes a `Computer` to auto-complete partial identifiers. Note that since
this will require loading the database environment, the completion
itself may take a second.

### DIFF
--- a/aiida/orm/utils/loaders.py
+++ b/aiida/orm/utils/loaders.py
@@ -82,7 +82,7 @@ class OrmEntityLoader(object):
         raise NotImplementedError
 
     @abstractclassmethod
-    def _get_query_builder_label_identifier(cls, identifier, classes):
+    def _get_query_builder_label_identifier(cls, identifier, classes, operator='==', project=None):
         """
         Return the query builder instance that attempts to map the identifier onto an entity of the orm class,
         defined for this loader class, interpreting the identifier as a LABEL like identifier
@@ -96,7 +96,7 @@ class OrmEntityLoader(object):
         raise NotImplementedError
 
     @classmethod
-    def _get_query_builder_id_identifier(cls, identifier, classes):
+    def _get_query_builder_id_identifier(cls, identifier, classes, operator='==', project=None):
         """
         Return the query builder instance that attempts to map the identifier onto an entity of the orm class,
         defined for this loader class, interpreting the identifier as an ID like identifier
@@ -105,14 +105,20 @@ class OrmEntityLoader(object):
         :param classes: a tuple of orm classes to which the identifier should be mapped
         :returns: the query builder instance
         """
+        if project is None:
+            project = ['*']
+
+        if operator == 'like':
+            identifier = str(identifier) + '%'
+
         builder = QueryBuilder()
-        builder.append(cls=classes, tag='entity', project=['*'])
-        builder.add_filter('entity', {'id': identifier})
+        builder.append(cls=classes, tag='entity', project=project)
+        builder.add_filter('entity', {'id': {operator: identifier}})
 
         return builder
 
     @classmethod
-    def _get_query_builder_uuid_identifier(cls, identifier, classes, query_with_dashes):
+    def _get_query_builder_uuid_identifier(cls, identifier, classes, query_with_dashes, project=None):
         """
         Return the query builder instance that attempts to map the identifier onto an entity of the orm class,
         defined for this loader class, interpreting the identifier as a UUID like identifier
@@ -123,6 +129,9 @@ class OrmEntityLoader(object):
         """
         from uuid import UUID
 
+        if project is None:
+            project = ['*']
+
         uuid = identifier.replace('-', '')
 
         if query_with_dashes:
@@ -131,7 +140,7 @@ class OrmEntityLoader(object):
                     uuid = '{}-{}'.format(uuid[:dash_pos], uuid[dash_pos:])
 
         builder = QueryBuilder()
-        builder.append(cls=classes, tag='entity', project=['*'])
+        builder.append(cls=classes, tag='entity', project=project)
 
         # If a UUID can be constructed from the identifier, it is a full UUID and the query can use an equality operator
         try:
@@ -144,7 +153,9 @@ class OrmEntityLoader(object):
         return builder
 
     @classmethod
-    def get_query_builder(cls, identifier, identifier_type=None, sub_classes=None, query_with_dashes=True):
+    def get_query_builder(
+        cls, identifier, identifier_type=None, sub_classes=None, query_with_dashes=True, operator='==', project=None
+    ):
         """
         Return the query builder instance that attempts to map the identifier onto an entity of the orm class,
         defined for this loader class, inferring the identifier type if it is not defined.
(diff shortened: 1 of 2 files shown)


aiida/orm/utils/loaders.py
```diff
@@ -105,14 +105,20 @@ def _get_query_builder_id_identifier(cls, identifier, classes):
...
-        builder.append(cls=classes, tag='entity', project=['*'])
-        builder.add_filter('entity', {'id': identifier})
+        builder.append(cls=classes, tag='entity', project=project)
+        builder.add_filter('entity', {'id': {operator: identifier}})
 
         return builder
 
     @classmethod
-    def _get_query_builder_uuid_identifier(cls, identifier, classes, query_with_dashes):
+    def _get_query_builder_uuid_identifier(cls, identifier, classes, query_with_dashes, project=None):
```
ltalirz: would you mind adding a docstring for `project` as well?

sphuber:
This is a first implementation to provide auto-completion for CLI parameters that target database entities. In this case the `ComputerParamType` has auto-completion. If one has a computer `localhost` for example, typing `verdi computer show local` followed by double-tab will autocomplete the computer name. This of course requires loading the database environment and with the query itself it takes a small second. However, this is only triggered when the auto-complete is triggered so it won't slow down the other parts of the CLI. One questionable point of this implementation is that PK and UUID identifiers are also autocompleted, which might be very useful, but potentially confusing, but I am not sure yet. If they _are_ autocompleted the value is transformed into the corresponding label. Also here I am not sure if this is the expected/desired behavior.

aiida/orm/utils/loaders.py
sphuber: Not at all, but since I realize there are some design decisions to be made and maybe in practice this is not necessarily very useful (see my post below), I first wanted to pass it by in this form

ltalirz:
> This is a first implementation to provide auto-completion for CLI parameters that target database entities.

I see - I wasn't aware this is the first.

> If they are autocompleted the value is transformed into the corresponding label. Also here I am not sure if this is the expected/desired behavior.

This could be nasty, if there are multiple computers with the same label, where the PK/uuid is unique but the label isn't.
So, it would autocomplete to something unusable.

Another problem is that currently you get an error when you press tabtab before writing the first letter/symbol to identify the computer:
```
$ verdi computer test Traceback (most recent call last):
  File "/Users/leopold/Applications/miniconda3/envs/aiida_rmq_py3/bin/verdi", line 10, in <module>
    sys.exit(verdi())
  File "/Users/leopold/Applications/miniconda3/envs/aiida_rmq_py3/lib/python3.7/site-packages/click/core.py", line 764, in __call__
  File "/Users/leopold/Applications/miniconda3/envs/aiida_rmq_py3/lib/python3.7/site-packages/click/core.py", line 712, in main
    _bashcomplete(self, prog_name, complete_var)
  File "/Users/leopold/Applications/miniconda3/envs/aiida_rmq_py3/lib/python3.7/site-packages/click_completion/patch.py", line 111, in _shellcomplete
    do_bash_complete(cli, prog_name)
  File "/Users/leopold/Applications/miniconda3/envs/aiida_rmq_py3/lib/python3.7/site-packages/click_completion/core.py", line 168, in do_bash_complete
    echo('\t'.join(re.sub(r"""([\s\\"'()])""", r'\\\1', opt) for opt, _ in choices), nl=False)
  File "/Users/leopold/Applications/miniconda3/envs/aiida_rmq_py3/lib/python3.7/site-packages/click_completion/core.py", line 168, in <genexpr>
    echo('\t'.join(re.sub(r"""([\s\\"'()])""", r'\\\1', opt) for opt, _ in choices), nl=False)
  File "/Users/leopold/Applications/miniconda3/envs/aiida_rmq_py3/lib/python3.7/site-packages/click_completion/core.py", line 113, in get_choices
    choices += [c if isinstance(c, tuple) else (c, None) for c in optctx.type.complete(ctx, incomplete)]
  File "/Users/leopold/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/cmdline/utils/decorators.py", line 68, in wrapper
  File "/Users/leopold/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/cmdline/params/types/computer.py", line 47, in complete
    return [(option, '') for option, in self.orm_class_loader.get_options(incomplete, project=['name'])]
  File "/Users/leopold/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/orm/utils/loaders.py", line 197, in get_options
    builder, _ = cls.get_query_builder(incomplete, operator='like', project=project)
  File "/Users/leopold/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/orm/utils/loaders.py", line 173, in get_query_builder
    identifier, identifier_type = cls.infer_identifier_type(identifier)
  File "/Users/leopold/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/orm/utils/loaders.py", line 285, in infer_identifier_type
    raise ValueError('the value for the identifier cannot be empty')
ValueError: the value for the identifier cannot be empty
```
As a user, I would find this behavior unexpected since this is different from the other stages of the verdi command.

Perhaps you're right and a bit more iteration could be useful here.
I would suggest to autocomplete only labels, and - if possible - to do that also when you haven' t started typing the label (just to see what all the labels out there are)

sphuber:
Fully agree that an empty value should not except but just show the available values. I will see if I can nicely change the behavior to just autocomplete the actual matched value and not always convert to label. If it becomes too complicated I can restrict it to label only for the time being.

giovannipizzi:
I didn't try, but indeed I would tend to autocomplete only the label (and maybe only if unique?). I think autocompleting the PK/UUID is not the behavior of other parts of the code (am I right?)

sphuber:
> I didn't try, but indeed I would tend to autocomplete only the label (and maybe only if unique?)

Are you referring if the incomplete identifier can be resolved uniquely or if the final computer label is unique? For computers the uniqueness constraint already exists for the label on the database level.
If you have two computers called `localhost-slurm` and `localhost-direct`, it would make a lot of sense to complete to `localhost-` if a user types `local` and hits tab, no?

> I think autocompleting the PK/UUID is not the behavior of other parts of the code (am I right?)

Correct, this functionality did not exist at all for database identifiers. I have now implemented it in a generic way in the entity loaders so enabling it for other param types, even as basic as `Node` would be very easy. Of course if we decide only to auto-complete labels, there it won't make as much sense.

giovannipizzi:
> Are you referring if the incomplete identifier can be resolved uniquely or if the final computer label is unique? For computers the uniqueness constraint already exists for the label on the database level.
If you have two computers called localhost-slurm and localhost-direct, it would make a lot of sense to complete to localhost- if a user types local and hits tab, no?

Sure, I was thinking to the final label, but indeed as you say for computers it's unique. And anyway I'm not sure that even e.g. for codes it would make sense not to complete labels that are used by multiple Codes, so maybe the simplest solution (just provide back the list of all computer labels and let bash take care of the completion) is the best.

sphuber:
Superseded by #3375 with auto-completion of only `LABEL` identifiers

PR #3367, opened 2019-10-01
Official Delete Feature

### LINKED ISSUE #3296: Rules for deleting Nodes

ramirezfranciscof:
Closes #3296

Delete feature implemented with a temporary internal search engine that looks for the nodes required to keep the provenance consistent. All tests have to be changed because they used different flags (for example, had an option to activate "return_forward" but no option for "create_forward") and the process worked differently (it would go recursively downstream, but only do checks upstream). Added a new set of tests for the feature and one test for the command line that checks there is no exceptions with any optional argument.


aiida/backends/tests/test_nodes.py
```diff
@@ -1702,100 +1621,385 @@ def test_deletion_non_existing_pk(self):
...
-        uuids_check_deleted = [n.uuid for n in (wf, slave1, slave2, outp2)]
 
+        # By default deleting input data should eliminate everything that it
+        # generated downwards.
+        di, do, c0, w1, w2 = self._create_minimal_graph()
+        uuids_check_existence = [n.uuid for n in ()]
+        uuids_check_deleted = [n.uuid for n in (di, do, c0, w1, w2)]
         with Capturing():
-            delete_nodes([wf.pk], verbosity=2, force=True, follow_calls=True, follow_returns=False)
+            delete_nodes([di.pk], verbosity=2, force=True)
```
giovannipizzi: Remove `verbosity=2` from here and below when capturing output.

Otherwise, put `with Capturing() as output`, then pass `output` as an additional parameter to `_check_existence()` (you'll need to move `_check_existence()` inside the `with` block) and then, if there is an exception, put the output in the exception message.

aiida/backends/tests/test_nodes.py
```diff
@@ -1702,100 +1621,385 @@ def test_deletion_non_existing_pk(self):
...
-        wf.add_incoming(in2, link_type=LinkType.INPUT_WORK, link_label='link2')
-        wf.store()
-        in2.add_incoming(wf, link_type=LinkType.RETURN, link_label='link3')
+        # Again, deleting with call_calc/work_forward also deletes other parts of
+        # the graph.
+        di1, di2, di3, di4, do1, pws, pcs, pwm = self._create_looped_graph()
+        uuids_check_existence = [n.uuid for n in (di1, di2, di3, di4)]
+        uuids_check_deleted = [n.uuid for n in (do1, pws, pcs, pwm)]
+        with Capturing():
+            delete_nodes([pcs.pk], verbosity=2, force=True, call_calc_forward=True, call_work_forward=True)
```
giovannipizzi: I think what you wanted to check is the following
```suggestion
            delete_nodes([pws.pk], verbosity=2, force=True, call_calc_forward=True, call_work_forward=True)
```
You can still keep both tests if you want, but the modified one is more important to test the flags, I would say.

aiida/backends/tests/test_nodes.py
```diff
@@ -1702,100 +1621,385 @@ def test_deletion_non_existing_pk(self):
...
+        self._check_existence(uuids_check_existence, uuids_check_deleted)
+
+    def _create_indep2w_graph(self):
+        """
+        Creates a simple graph with one workflow handling two independent workflows
+        (with one simple calculation each). It was designed and used mainly to point
+        out how the deletion behaviour works when setting call_calc_forward to true:
```
giovannipizzi:
```suggestion
        out how the deletion behaviour works when setting call_work_forward to true:
```
Also, add that if you set also `call_calc_forward` on top of this, actually also unrelated data will be deleted (DOB)

aiida/backends/tests/test_nodes.py
```diff
@@ -1702,100 +1621,385 @@ def test_deletion_non_existing_pk(self):
...
+        in this case if PWA is deleted, the somewhat independent PWB will be deleted
+        as well.
```
giovannipizzi: Maybe mention that this is the same graph as in the docs.

aiida/backends/tests/test_nodes.py
```diff
@@ -1702,100 +1621,385 @@ def test_deletion_non_existing_pk(self):
...
-            delete_nodes([called.pk], verbosity=2, force=True, follow_returns=True)
+        for ii in range(20):
+            new_calc = orm.CalculationNode()
+            new_data = orm.Data().store()
+            new_calc.add_incoming(old_data, link_type=LinkType.INPUT_CALC, link_label='inp' + str(ii))
+            new_calc.store()
+            new_data.add_incoming(new_calc, link_type=LinkType.CREATE, link_label='out' + str(ii))
+            node_list_id.append(new_calc.id)
+            node_list_id.append(new_data.id)
+            old_data = orm.load_node(new_data.id)
```
giovannipizzi:
```suggestion
            old_data = new_data
```

aiida/backends/tests/test_nodes.py
```diff
@@ -1702,100 +1621,385 @@ def test_deletion_non_existing_pk(self):
...
+
+        node_list = []
+        for node_id in node_list_id:
```
giovannipizzi: also in this case, just create (above) `node_list` directly instead of creating a list of ids

ramirezfranciscof:
(Disregard latest commit, not all comments have been addressed yet)

aiida/manage/database/delete/nodes.py
```diff
@@ -58,43 +80,70 @@ def delete_nodes(
...
-    # Operational set always includes the recently (in the last iteration added) nodes.
-    operational_set = set().union(set(starting_pks))  # Union to copy the set!
-    pks_set_to_delete = set().union(set(starting_pks))
+    follow_backwards = []
+    follow_backwards.append(LinkType.CREATE.value)
+    follow_backwards.append(LinkType.RETURN.value)
+    follow_backwards.append(LinkType.CALL_CALC.value)
+    follow_backwards.append(LinkType.CALL_WORK.value)
+
+    follow_fowards = []
```
giovannipizzi: Sorry to bother, but it should be `forwards` and not `fowards`...

ramirezfranciscof: Uf, aplogies, my bad!

sphuber:
Great work @ramirezfranciscof thanks a lot

PR #3374, opened 2019-10-02
Fix loading of database environment in ReadTheDocs environment

### LINKED ISSUE #3371: readthedocs build broken
see https://github.com/aiidateam/aiida-core/pull/3314#issuecomment-537269013

sphuber:
Fixes #3371 

The environment of ReadTheDocs does not actually have a database and so
the typical configuration and profile loading had to be hotwired. A
special global flag `IN_RTD_DOC_MODE` in `aiida.manage.configuration`
would circumvent loading an actual configuration for a dummy one. Since
the actual database environment never had to be loaded, the fact that
this contained non-sensical database connection parameters was
irrelevant. However, recently the documentation was changed to use the
Sphinx extension to document the `CalcJob` class. This now requires
loading the `aiida.orm` module, which in turn requires the database
environment being loaded.

The fix is to actually load the database environment through the backend
manager, but without performing the schema check. This will work despite
the unusable database configuration because no actual connection is
established in this way.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -257,10 +257,21 @@ if not on_rtd:  # only import and set the theme if we're building docs locally
     load_profile(config.default_profile_name)
     get_manager().get_backend()
 else:
-    # Back-end settings for readthedocs online documentation.
     from aiida.manage import configuration
+    from aiida.manage.configuration import load_profile, reset_config
+    from aiida.manage.manager import get_manager
+
+    # Set the global variable to trigger shortcut behavior in `aiida.manager.configuration.load_config`
     configuration.IN_RT_DOC_MODE = True
-    configuration.BACKEND = 'django'
+
+    # First need to reset the config, because an empty one will have been loaded when `aiida` module got imported
+    reset_config()
+
+    # Load the profile: this will first load the config, which will be the dummy one for RTD purposes
+    load_profile()
+
+    # Finally load the database backend but without checking the schema because there is no actual database
+    get_manager()._load_backend(schema_check=False)
 
 
 def run_apidoc(_):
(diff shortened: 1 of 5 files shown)


docs/update_req_for_rtd.py
```diff
@@ -34,7 +34,7 @@ def update_req_for_rtd(pre_commit):
     extras = setup_json['extras_require']
     reqs = set(extras['testing'] + extras['docs'] + extras['rest'] + extras['atomic_tools'] +
                # To avoid that it requires also the postgres libraries
```
giovannipizzi: We should probably remove this comment

giovannipizzi:
BTW, we should probably set some configuration in coveralls so that if the coverage decreases less than a threshold, it's still considered ok

docs/update_req_for_rtd.py
sphuber: Done

docs/source/conf.py
```diff
@@ -257,10 +257,21 @@
...
-    configuration.BACKEND = 'django'
+
+    # First need to reset the config, because an empty one will have been loaded when `aiida` module got imported
+    reset_config()
+
+    # Load the profile: this will first load the config, which will be the dummy one for RTD purposes
+    load_profile()
+
+    # Finally load the database backend but without checking the schema because there is no actual database
+    get_manager()._load_backend(schema_check=False)
```
ltalirz: @sphuber One question: I could imagine that loading a dummy profile could be useful not just on read the docs but in other contexts as well (e.g. for a quick test that doesn't need the DB... perhaps others).

So I'm wondering: instead of this slightly hackish way of getting there, how about we either
 * add a function `load_dummy_profile()` or
 * add an optional argument `dummy_profile` (or similar) to`load_profile` 

that does the job?
Then people who need this (even if just to enable building their docs on RTD) don't need to know about AiiDA internals.

sphuber: I guess that makes sense. I didn't do it now because as soon as we make this an "official" public function we should now what potential problems and dangers could be. But I am very willing to discuss this and see if we can implement a useful solution

PR #3375, opened 2019-10-03
Enable auto-completion for `IdentifierParamType`

### LINKED ISSUE #3361: Enable auto-completion for `ComputerParamType`

sphuber:
Fixes #3361 

The `OrmEntityLoader` class is extended to allow specifying a query
operator in the `get_query_builder` method. This allows the introduction
of the `get_options` method, which unlike `load_entity` will not try to
load one specific entity, but provide a list of possibilities for a
given identifier. This in turn allows the CLI `IdentifierParamType` to
enable auto-completion. Here we only enable it for `LABEL` identifiers
and do not try to infer the identifier type. Finally, the new
functionality is hooked up for the `ComputerParamType`.

### DIFF
--- a/aiida/orm/utils/loaders.py
+++ b/aiida/orm/utils/loaders.py
@@ -82,13 +82,15 @@ class OrmEntityLoader(object):
         raise NotImplementedError
 
     @abstractclassmethod
-    def _get_query_builder_label_identifier(cls, identifier, classes):
+    def _get_query_builder_label_identifier(cls, identifier, classes, operator='==', project='*'):
         """
         Return the query builder instance that attempts to map the identifier onto an entity of the orm class,
         defined for this loader class, interpreting the identifier as a LABEL like identifier
 
         :param identifier: the LABEL identifier
         :param classes: a tuple of orm classes to which the identifier should be mapped
+        :param operator: the operator to use in the query
+        :param project: the property or properties to project for entities matching the query
         :returns: the query builder instance
         :raises ValueError: if the identifier is invalid
         :raises aiida.common.NotExistent: if the orm base class does not support a LABEL like identifier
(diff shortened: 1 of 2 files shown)


sphuber:
This supersedes #3362 as this PR only implements auto-completion for `LABEL` identifiers

aiida/orm/utils/loaders.py
```diff
@@ -375,19 +406,24 @@ def orm_base_class(self):
...
         :param identifier: the LABEL identifier
         :param classes: a tuple of orm classes to which the identifier should be mapped
+        :param operator: the operator to use in the query
+        :param project: the property or properties to project for entities matching the query
         :returns: the query builder instance that should retrieve the entity corresponding to the identifier
         :raises ValueError: if the identifier is invalid
         :raises aiida.common.NotExistent: if the orm base class does not support a LABEL like identifier
         """
+        if operator == 'like':
+            identifier = str(identifier) + '%'
```
giovannipizzi: use `escape_for_sql_like` here (and everywhere below)

aiida/orm/utils/loaders.py
```diff
@@ -342,19 +368,24 @@ def orm_base_class(self):
...
```
giovannipizzi: use `escape_for_sql_like` here

aiida/orm/utils/loaders.py
```diff
@@ -309,19 +330,24 @@ def orm_base_class(self):
...
```
giovannipizzi: use `escape_for_sql_like` here

aiida/orm/utils/loaders.py
```diff
@@ -427,7 +465,7 @@ def _get_query_builder_label_identifier(cls, identifier, classes):
             raise ValueError('the identifier needs to be a string')
 
         builder = QueryBuilder()
-        builder.append(cls=classes, tag='code', project=['*'], filters={'label': {'==': label}})
+        builder.append(cls=classes, tag='code', project=project, filters={'label': {'==': label}})
 
         if machinename:
             builder.append(Computer, filters={'name': {'==': machinename}}, with_node='code')
```
giovannipizzi: As a note for the future, you might always want to append the Computer, just put the filter if you have a `machine name`, project the computer name, and then when you return (in general, for all _get_query_... functions) return also a string to say how to construct the output string from the projected values, e.g. doing a `.format(*projected_values)` (e.g. in general this will be `"{0}"` and here it would be `"{0}@{1}"`).

PR #3384, opened 2019-10-04
Check existence before adding to ZipFolder

### LINKED ISSUE #3383: 'UserWarning: DuplicateName' when exporting zip files
A `UserWarning: DuplicateName: 'some/internal/path/'` warning message shows up when exporting Nodes with empty repository folders using `verdi export create`.

This can be mediated by checking whether the internal path already exists before attempting to write it (again).

CasperWA:
Fixes #3383


aiida/tools/importexport/dbexport/zip.py
```diff
@@ -127,23 +136,18 @@ def insert_path(self, src, dest_name=None, overwrite=True):
...
 
-        if not overwrite:
-            try:
-                self._zipfile.getinfo(base_filename)
-                exists = True
-            except KeyError:
-                exists = False
-            if exists:
-                raise IOError('destination already exists: {}'.format(base_filename))
+        if not overwrite and self.exists(base_filename):
```
sphuber: Should this check then also not be done on line 155? Anyway, if this is a bug to be solved. Shouldn't there be a test for this. And why are we writing our own zip folder implementation? Surely there must be something out of the box in python.

aiida/tools/importexport/dbexport/zip.py
```diff
@@ -127,23 +136,18 @@ def insert_path(self, src, dest_name=None, overwrite=True):
...
 
         # print src, filename
         if os.path.isdir(src):
             for dirpath, dirnames, filenames in os.walk(src):
                 relpath = os.path.relpath(dirpath, src)
                 if not dirnames and not filenames:
                     real_src = dirpath
-                    real_dest = os.path.join(base_filename, relpath)
-                    self._zipfile.write(real_src, real_dest)
+                    real_dest = os.path.join(base_filename, relpath) + '/'
```
sphuber: Why the trailing slash? If this is really necessary, one should use the `os.path.sep`

CasperWA: Cheers. Well, it seems the slash was really the issue all along here, since _with_ the slash, the string now represents a possible existing path in the zip-file. This is due to zip adding an ending slash to signify a folder type.

aiida/tools/importexport/dbexport/zip.py
CasperWA: There is no test, simply because I didn't have time to add it. But can do that at a later time if needed.
And yes, perhaps this check (similar to the one added in line 149) should be added in line 155 as well. Seems like a sound idea :)

Finally, there actually isn't (to my knowledge). But mostly, because the purpose of this class is to "mimic" the `Folder` class' methods, so that the classes can be used interchangeably in the `export_tree` function.
In the end, this is a quick-fix to the current situation that will/should all be obsolete with the use of the `Archive` classes.

CasperWA: I don't think I am going to write a test for this, as it is not, as such, a critical bug. The zipfile will still be written correctly, it will just simply spit out "harmless" UserWarnings.

However, I can always make a general test for when one exports a simple graph to a ZipFile, if a UserWarning is raised or not?

PR #3385, opened 2019-10-04
Docs: remove `--pre` from pip install

### LINKED ISSUE #3339: Docs: Remove `--pre` for PyPI installation documentation (when releasing v1.0.0)
When releasing the final AiiDA v1.0.0, the `--pre` flag should be removed from [these](https://github.com/aiidateam/aiida-core/blame/ab4571e9edf06b591518324fe78a2bfeddddf81e/docs/source/install/installation.rst#L59) kind of lines in the documentation, i.e.:

```shell
pip install --pre aiida
```

CasperWA:
Closes #3339 

Also update installation information to be against the meta-package `aiida` instead of `aiida-core`.
Note, I have not checked that the installation of the meta-package `aiida` would introduce any issues instead of referencing to install `aiida-core`.


sphuber:
What I meant to say in the issue is that we actually should install change the docs to say to install `aiida-core` and _not_ the meta-package. But before you update this PR, let's discuss with @giovannipizzi and @ltalirz to see if they agree.

giovannipizzi:
I agree - for now let's reference aiida-core only, and then let's try to see if we can make `aiida` just an alias to `aiida-core`.

sphuber:
Leo also agrees with just `aiida-core` and added that we should remove the `--pre` _only_ when release is imminent. I agree here, so we should hold of with this PR until we are about to release, because otherwise the `latest` RTD branch will show incorrect information

CasperWA:
> What I meant to say in the issue is that we actually should install change the docs to say to install `aiida-core` and _not_ the meta-package.

Ah. This was a bit unclear. Sorry.

> Leo (...) added that we should remove the `--pre` _only_ when release is imminent. I agree here, so we should hold of with this PR until we are about to release, because otherwise the `latest` RTD branch will show incorrect information

This is a given - I simply prepared the PR so that it was ready for you (@sphuber) to merge in when ready :smiley:

PR #3389, opened 2019-10-04
Prune input mapping of `ProcessBuilder` before creating `Process`

### LINKED ISSUE #3377: `ProcessBuilder` will always return mapping of namespace even if completely empty
**Steps to reproduce:**
Create a ``PwBandsWorkChain`` using the process builder, without specifying any ``relax`` inputs.

**Expected behavior:**
The workchain skips the relaxation step.

**Actual behavior:**
The workchain fails to validate its inputs, claiming a missing ``code`` input for the ``relax`` part.

**Notes:**
The issue seems to come from the fact that the builder contains
```
'relax': {'metadata': {}, 'base': {'metadata': {}, 'pw': {'metadata': {'options': {}}, 'pseudos': {}}}}
```
Deleting the ``relax`` attribute entirely isn't possible, and replacing it with ``None`` or ``{}`` doesn't fix the issue.

Creating the same input as a dictionary works as expected.

I'm not sure if this is a problem with ``aiida-core``, the way the spec is built in ``aiida-quantumespresso``, or kind of inevitable. @sphuber is the expert on the topic in any case :smile:

sphuber:
Fixes #3377 

The validation of a `ProcessSpec` will trigger for a port as soon as a
value has been passed. For a port namespace this is also the case, even
if it contains no actual values. This was a problem when launching a
process from a `ProcessBuilder`. If it contained an optional namespace,
it would always pass at least an empty dictionary for said port, even if
the user did not specify any specific input within it. This would then
always trigger the validation which would fail. This limitation made the
concept of optional port namespaces fundamentally incompatible with the
`ProcessBuilder`. Here we add a method to "prune" the input mapping of a
builder before it is being passed to the process constructor. This will
remove any portnamespace that is completely empty, meaning it contains
no other values other than other empty dictionaries (namespaces).

### DIFF
--- a/aiida/backends/tests/engine/test_process_builder.py
+++ b/aiida/backends/tests/engine/test_process_builder.py
@@ -11,7 +11,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
 
-from collections import MutableMapping
+from collections import Mapping, MutableMapping
 
 from aiida import orm
 from aiida.backends.testbase import AiidaTestCase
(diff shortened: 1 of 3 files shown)


sphuber:
@greschd with this fix I am now able to use a builder to run the `PwBandsWorkChain` without the relax option. If we don't specify any of its inputs, the call to `_inputs(prune=True)` when constructing the process will now make sure that the empty nested mapping is pruned and so does not trigger default population and subsequent validation which fails. Please give that a try and see if you think the solution can be improved. I still have the feeling that there is something dirty, but can't quite see an alternative yet.

greschd:
Is there a case where the empty mapping that should be pruned is not of type ``ProcessBuilderNamespace``? I'm not sure how to feel about pruning all empty ``collections.Mapping`` instances. 

The AiiDA ``Dict`` is something that could in principle be extended with ``Mapping`` capability, and I think the more cautious approach would be not to prune an empty ``Dict``.  Maybe we should just exclude any AiiDA data types?

Other than that, I think this is a good solution to a somewhat ugly problem. I like that it's implemented completely in the builder (doesn't touch the main validation), and pretty easy to get around should it cause trouble.

sphuber:
I see your point about not wanting to prune all empty `collections.Mapping` instances. However, I think the problem is that in order to get the actual input mapping from the builder instance, we have to either call `**self` or `dict(self)`, no? In which case the resulting value, which is passed to `_prune` is just a plain `dict` and no longer nested `ProcessBuilderNamespace` instances.

In addition, if there is an empty mapping, it has to be either a namespace, or value. If it is a namespace it can safely be pruned. If it is intended as an actual value, however, then it should be a sub class of `Node` unless the corresponding port is `non_db=True`. I think that may be the only case where the value will be pruned that might not be desirable. An empty mapping for a `non_db` port. I am not sure how to check for this.

greschd:
Well, since it's only the top-level that will be a ``dict`` you could just change it to 

```python
return {key: self._prune(value) for key, value in self.items()}
```

in the ``_inputs`` function.

Not sure if this is more elegant, though. However, I would probably explicitly exclude AiiDA ``Node`` from being pruned -- after all, we don't know if some plug-in data type might match the ``Mapping`` test.

sphuber:
I have put an additional clause to prevent `Node` instance that are empty mappings from being pruned. Also added a test for that

PR #3394, opened 2019-10-05
Centralize definition of graph traversal rule toggleability and default

### LINKED ISSUE #3393: Centralize definition of graph traversal rule toggleability and default
The graph traversal rules that are used during export and node deletion need to be defined in a single place since they apply to both scenarios. Just the toggleability and default value is different.

sphuber:
Fixes #3393 and resolves `verdi` loading speed issue

The available graph traversal rules are identical for node export and
deletion. Only the toggleability and defaults differs for each of the
two scenarios. To prevent mistakes, we centralize their definition in
one location. Two dictionaries in `aiida.common.links` provide a mapping
of each traversal rule onto a named tuple that records the default value
and toggleability.

This notation allows for dynamic code, for example, the command line
options that toggle the rules are now generated dynamically for the
`verdi export create` and `verdi node delete` commands.

In addition, the old `LINK_FLAG` data structure was defined in the
`aiida.tools.importexport` module which indirectly causeed `verdi` to
slow down. The dictionary was needed to generate the command options
and by importing it, indirectly the `aiida.orm` module was imported.

### DIFF
--- a/aiida/manage/database/delete/nodes.py
+++ b/aiida/manage/database/delete/nodes.py
@@ -15,13 +15,11 @@ from __future__ import print_function
 import click
 
 from aiida.cmdline.utils import echo
+from aiida.common import InternalError
 
 
-def delete_nodes(
-    pks, verbosity=0, dry_run=False, force=False, create_forward=True, call_calc_forward=False, call_work_forward=False
-):
-    """
-    Delete nodes by a list of pks.
+def delete_nodes(pks, verbosity=0, dry_run=False, force=False, **kwargs):
+    """Delete nodes by a list of pks.
 
     This command will delete not only the specified nodes, but also the ones that are
     linked to these and should be also deleted in order to keep a consistent provenance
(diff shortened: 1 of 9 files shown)


sphuber:
> Shouldn't one use the new data structures also in the function `from aiida.manage.database.delete.nodes import delete_nodes`?

I looked at that, but it already correctly exposed only the toggleable flags. I could make it **kwargs like in export and apply the same technique for validating them. You think that is better?

giovannipizzi:
I think so, because in such a way we centralise the logic - now it's hardcoding the defaults and what is toggleable into the function `delete_nodes` in the following "implicit" way:
https://github.com/aiidateam/aiida-core/blob/85238b6aa54e04ff5d34950b5cc70d35f6dd2a10/aiida/manage/database/delete/nodes.py#L89-L106

sphuber:
Absolutely right, I'll make the change.

aiida/manage/database/delete/nodes.py
```diff
@@ -86,24 +74,23 @@ def delete_nodes(
...
 
-    if call_calc_forward:
-        follow_forwards.append(LinkType.CALL_CALC.value)
+        if follow and rule.direction == 'forward':
+            follow_forwards.append(rule.link_type.value)
 
-    if call_work_forward:
-        follow_forwards.append(LinkType.CALL_WORK.value)
+        elif follow and rule.direction == 'backward':
+            follow_backwards.append(rule.link_type.value)
```
giovannipizzi: I'm going to approve, but do we want to put here (or somewhere else?) something to check the `else` clause to verify we didn't mistype some of the values, being neither `forward` nor `backward`? E.g. raising an InternalError

sphuber: Done

Issue #3395, opened 2019-10-06
Exception raised "no computer specified" when submitting a `CalcJob`


mrlixuec:
I want to try run a calcjob,however it always note me that i has not specified a computer.But the verdi computer test and verdi status is good.Below is my code:
```
from aiida.engine import run,submit
from aiida.orm import load_code,Int,load_computer
from aiida.plugins import CalculationFactory
from aiida import load_profile

load_profile()

ArithmeticAddCalculation = CalculationFactory('arithmetic.add')

inputs = {
        'code':load_code(label='add1'),
        'x':Int(1),
        'y':Int(2),
        'metadata':
             {'options':
                 {'resources':{'num_machines':1,'num_mpiprocs_per_machine':1},
                  'computer':load_computer(label='localhost'),
                     }
              },
         }

submit(ArithmeticAddCalculation,**inputs)
```

sphuber:
In principle your `code` object already has a connected computer.
You can make sure by doing the following in a `verdi shell`:
```
from aiida.orm import load_code
code = load_code('add1')
code.computer
```
This should print the computer that is configured for that code.
This always means you don't have to specify the computer in the inputs of the calculation.
Just specifying the `code` is enough. So try removing `computer` from the `metadata.options` and submit again

sphuber:
Also, could you please report what version of `aiida-core` you are using? Thanks

mrlixuec:
> Also, could you please report what version of `aiida-core` you are using? Thanks

Thank you very much!!! I will try your solution,and the version of my aiida-core is 1.0.0b6

mrlixuec:
> Also, could you please report what version of `aiida-core` you are using? Thanks
I'm so sorry!! I find it's i hadn't connect my code to a conputer,so the mistake was made .
The code could running normally withou `code.computer`.
Again,I'm so sorry make a problem for you .And thank for your effort on aiida!Thank you very much .

sphuber:
No problem. So I take it you are able to run the example now?

I will keep this issue open because I am a bit surprised that it is possible to store a `Code` without a `Computer`. I will investigate how this is possible and whether we should consider it as a bug.

PR #3400, opened 2019-10-07
Move `CalcJob` input `computer` from `options` to `metadata`

### LINKED ISSUE #3396: Setting `computer` directly in `metadata.options` will except `CalcJob` submission
The problem is because since `computer` is part of the options and options are set as attributes on the `CalcJobNode`, the attributes will contain a `Computer` instance. When the node is then stored and the attributes are serialized, an exception is raised, because `Computer` instances are not serializable. The `computer` option should not be set as an attribute like all others, but rather be set through the `computer` setter property. This will then create a proper foreign key relationship.

@giovannipizzi I feel this problem might suffer from some historical design. Since the `code` is a required input for all `CalcJobs` and as far as I know, a `code` _always_ has to have an associated computer: what is then the use-case for being able to specify an explicit computer in the options? Do we really need this? Or can I just remove it from the spec?

sphuber:
Fixes #3396 

All options are stored as attributes on the `CalcJobNode` but the
`computer` input will have a `Computer` instance as type and needs to be
stored as a property. It therefore makes more sense to be passed as a
metadata like the `label` and `description`.

### DIFF
--- a/aiida/backends/tests/engine/test_calc_job.py
+++ b/aiida/backends/tests/engine/test_calc_job.py
@@ -93,6 +93,25 @@ class TestCalcJob(AiidaTestCase):
         with self.assertRaises(TypeError):
             launch.run(SimpleCalcJob, code=self.code, metadata={'options': orm.Dict(dict={'a': 1})})
 
+    def test_computer(self):
+        """Test passing an explicit `computer` in the `metadata`."""
+        inputs = {
+            'code': self.code,
+            'x': orm.Int(1),
+            'y': orm.Int(2),
+            'metadata': {
+                'computer': self.code.computer,
+                'options': {
+                    'resources': {
+                        'num_machines': 1,
+                        'num_mpiprocs_per_machine': 1
+                    },
+                }
+            }
+        }
+
+        ArithmeticAddCalculation(inputs=inputs)
+
     def test_invalid_parser_name(self):
         """Passing an invalid parser name should already stop during input validation."""
         inputs = {
--- a/aiida/engine/processes/calcjobs/calcjob.py
+++ b/aiida/engine/processes/calcjobs/calcjob.py
@@ -50,9 +50,11 @@ class CalcJob(Process):
     def define(cls, spec):
         # yapf: disable
         super(CalcJob, cls).define(spec)
-        spec.input('code', valid_type=orm.Code, help='The Code to use for this job.')
+        spec.input('code', valid_type=orm.Code, help='The `Code` to use for this job.')
         spec.input('metadata.dry_run', valid_type=bool, default=False,
-            help='When set to True will prepare the calculation job for submission but not actually launch it.')
+            help='When set to `True` will prepare the calculation job for submission but not actually launch it.')
+        spec.input('metadata.computer', valid_type=orm.Computer, required=False,
+            help='When using a "local" code, set the computer on which the calculation should be run.')
         spec.input_namespace('{}.{}'.format(spec.metadata_key, spec.options_key), required=False)
         spec.input('metadata.options.input_filename', valid_type=six.string_types, required=False,
             help='Filename to which the input for the code that is to be run will be written.')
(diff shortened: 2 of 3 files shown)


sphuber:
Technically this breaks backwards compatibility as one can no longer specify `metadata.options.computer`. However, that was excepting anyway, so I don't think it is a problem in this case.

ltalirz:
Ah, I wasn't even aware that one could pass a computer separately :-D
Has there ever been a use case for this? If not (and if the current way to pass it was excepting), perhaps even better to remove it completely?

sphuber:
It is not necessary for "remote" codes, because there the `code` will have a computer attached, which will be automatically set by the engine. However, for "local" codes one has to manually specify the `computer`, which since it is not a `Node` cannot be an input but has to be set in metadata. The reason it ended up in `metadata.options` is for historical reasons and the reason it was never encountered suggests there are not many users running local codes with the `aiida-core==1.0.0`

ltalirz:
> since it is not a Node cannot be an input but has to be set in metadata.

Hm... one more reason to make `Computer` a Node, I guess.
I let you guys decide this.

ltalirz:
P.S. We should just make sure that we don't confuse people by making the Computer input more visible than it was before (e.g. it might show up in the `verdi plugin` output now, and people might start wondering whether they need to provide a computer in addition to a code).

sphuber:
> P.S. We should just make sure that we don't confuse people by making the Computer input more visible than it was before (e.g. it might show up in the `verdi plugin` output now, and people might start wondering whether they need to provide a computer in addition to a code).

As chance would have it, since the output of `verdi plugin list` does not yet support showing nested namespaces, it will just show the `metadata` input, so for the time being it is not a problem. But that might change of course

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -53,6 +53,8 @@ def define(cls, spec):
         spec.input('code', valid_type=orm.Code, help='The Code to use for this job.')
         spec.input('metadata.dry_run', valid_type=bool, default=False,
             help='When set to True will prepare the calculation job for submission but not actually launch it.')
+        spec.input('metadata.computer', valid_type=orm.Computer, required=False,
+            help='Set the computer to be used by the calculation')
```
ltalirz: Perhaps mention here that this is for local codes only, something like
```suggestion
            help='Set the computer to be used by the calculation (for "local" codes only)')
```

aiida/backends/tests/engine/test_calc_job.py
```diff
@@ -93,6 +93,25 @@ def prepare_for_submission(self, folder):
         with self.assertRaises(TypeError):
             launch.run(SimpleCalcJob, code=self.code, metadata={'options': orm.Dict(dict={'a': 1})})
 
+    def test_computer(self):
+        """Test passing an explicit `computer` in the `metadata`."""
+        inputs = {
+            'code': self.code,
```
ltalirz: Looking here https://github.com/aiidateam/aiida-core/pull/3400/files#diff-4faa11e49a5f253a316f6d5f1e156cccR31
this seems to be a code with a remote executable.

So, I'm wondering: Shouldn't this test actually fail?
And shouldn't we have another test with a "local" code?
So, what is this test actually testing?

sphuber: It is just testing that the process spec validation, which is triggered by constructing the process instance, does not except, which used to be the case when specifying the `computer` in `metadata.options`, even though it had a corresponding port. None of the tests in the module actually "run" the calculation. Once we move to `pytest` we could take the fixtures I wrote for `aiida-quantumespresso` to easily check more, such as the content of the input files written by `prepare_for_submission`.

Also regarding local and remote codes: there are no real restrictions on this. The only check is that when the `CalcJobNode` is stored, `_validate` will check that _a_ computer has been set. If a remote code is passed in the inputs, its corresponding computer is automatically set. For a local code, one has to do this manually.

ltalirz: > Also regarding local and remote codes: there are no real restrictions on this.

I see. Would it make sense to add one, though?
I think we should warn when people are setting a computer for a remote code since this likely means they're using aiida incorrectly/

sphuber: I agree, but I was taking care of this in a separate PR. A user reported an issue in #3395 that was most likely him accidentally using a local code. I told him how to fix it but kept the issue open to improve the validation. I think instead of the validation on the node level, it should be done on the process spec.

ltalirz: Ok, that's fine of course.
We should just keep in mind that this test here should fail once this extra check is implemented

Issue #2914, opened 2019-05-23
verdi status ; rabbitmq error loop?


rkovacik:
```
kovacik@iff1610:~/CODES/aiida_core 0> verdi status 
 ✓ profile:     On profile local
 ✓ repository:  /Users/kovacik/.aiida/repository-local
 ✗ postgres:    Unable to connect to iff504:5432
05/23/2019 06:17:14 PM <6424> kiwipy.rmq.communicator: [ERROR] Error starting the communicator:
Traceback (most recent call last):
  File "/Users/kovacik/CODES/venv_aiida1.0/lib/python2.7/site-packages/kiwipy/rmq/communicator.py", line 535, in do_connect
    yield self._communicator.connect()
  File "/Users/kovacik/CODES/venv_aiida1.0/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/Users/kovacik/CODES/venv_aiida1.0/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "/Users/kovacik/CODES/venv_aiida1.0/lib/python2.7/site-packages/tornado/gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "/Users/kovacik/CODES/venv_aiida1.0/lib/python2.7/site-packages/kiwipy/rmq/communicator.py", line 363, in connect
    yield self._message_subscriber.connect()
  File "/Users/kovacik/CODES/venv_aiida1.0/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/Users/kovacik/CODES/venv_aiida1.0/lib/python2.7/site-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "/Users/kovacik/CODES/venv_aiida1.0/lib/python2.7/site-packages/tornado/gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "/Users/kovacik/CODES/venv_aiida1.0/lib/python2.7/site-packages/kiwipy/rmq/communicator.py", line 179, in connect
    self._channel = yield self._connection.channel()
  File "/Users/kovacik/CODES/venv_aiida1.0/lib/python2.7/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
[...]

PR #3408, opened 2019-10-09
Suppress kiwipy log message for verdi status check (#2914).

csadorf:
The kiwipy package logs the full traceback in case that the connection
to the RabbitMQ server cannot be established. This patch suppresses this
log message to reduce the noise presented to the user for the status
check.

### DIFF
--- a/aiida/cmdline/commands/cmd_status.py
+++ b/aiida/cmdline/commands/cmd_status.py
@@ -14,11 +14,24 @@ from __future__ import absolute_import
 
 import sys
 
+import logging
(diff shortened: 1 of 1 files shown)


csadorf:
I'll fix my development environment and install the pre-commit hooks.

csadorf:
Updated patch to be compliant with pre-commit hooks.

I am requesting merge into develop instead of the release branch, because to my understanding that makes more sense at the current stage, but please let me know if you would like me to backport this.

PR #3409, opened 2019-10-09
Suppress kiwipy log message for verdi status check

csadorf:
The kiwipy package logs the full traceback in case that the connection
to the RabbitMQ server cannot be established. This patch suppresses this
log message to reduce the noise presented to the user for the status
check.

Addresses issue #2914.

### DIFF
--- a/aiida/cmdline/commands/cmd_status.py
+++ b/aiida/cmdline/commands/cmd_status.py
@@ -18,6 +18,7 @@ from enum import IntEnum
 import click
 
 from aiida.cmdline.commands.cmd_verdi import verdi
+from aiida.common.log import override_log_level
 from ..utils.echo import ExitCode
 
 
@@ -79,8 +80,9 @@ def verdi_status():
     # Getting the postgres status by trying to get a database cursor
     database_data = [profile.database_username, profile.database_hostname, profile.database_port]
     try:
-        backend = manager.get_backend()
-        backend.cursor()
+        with override_log_level():  # temporarily suppress noisy logging
+            backend = manager.get_backend()
+            backend.cursor()
     except Exception:
         print_status(ServiceStatus.DOWN, 'postgres', 'Unable to connect as {}@{}:{}'.format(*database_data))
         exit_code = ExitCode.CRITICAL
@@ -90,8 +92,9 @@ def verdi_status():
     # getting the rmq status
     try:
         with Capturing(capture_stderr=True):
-            comm = manager.create_communicator(with_orm=False)
-            comm.stop()
+            with override_log_level():  # temporarily suppress noisy logging
+                comm = manager.create_communicator(with_orm=False)
+                comm.stop()
     except Exception as exc:
         print_status(ServiceStatus.ERROR, 'rabbitmq', 'Unable to connect to rabbitmq', exception=exc)
         exit_code = ExitCode.CRITICAL
--- a/aiida/common/log.py
+++ b/aiida/common/log.py
@@ -15,6 +15,7 @@ from __future__ import absolute_import
 import copy
 import logging
 import types
+from contextlib import contextmanager
 import six
 
 if six.PY2:
@@ -24,7 +25,7 @@ else:
 
 from aiida.manage.configuration import get_config_option  # pylint: disable=wrong-import-position
 
-__all__ = ('AIIDA_LOGGER',)
+__all__ = ('AIIDA_LOGGER', 'override_log_level')
 
 # Custom logging level, intended specifically for informative log messages reported during WorkChains.
 # We want the level between INFO(20) and WARNING(30) such that it will be logged for the default loglevel, however
@@ -201,3 +202,13 @@ def configure_logging(with_orm=False, daemon=False, daemon_log_file=None):
         config['loggers']['aiida']['handlers'].append(handler_dblogger)
 
     dictConfig(config)
+
+
+@contextmanager
+def override_log_level(level=logging.CRITICAL):
+    """Temporarily adjust the log-level of logger."""
+    logging.disable(level=level)
+    try:
+        yield
+    finally:
+        logging.disable(level=logging.NOTSET)


csadorf:
Sorry about potential confusion due to opening a second PR. I did not realize I would not be able to re-open the previous one.

csadorf:
It is not clear to me why this patch would lead to such a significant reduction in coverage.

sphuber:
> It is not clear to me why this patch would lead to such a significant reduction in coverage.

Don't worry about this. If one of Travis jobs fails, the total coverage will always be less. If you rerun the Travis job and it succeeds the coverage will be more accurate. However, even in that case sometimes it will report a decrease of `0.0*%` i.e. something negligible, but since it is a decrease it will still show it as red. We should try to find if we can put some threshold on the coverage configuration

sphuber:
And a PR into `develop` is all we need. We don't need a backport onto `release_0.12*` because RabbitMQ only exists for `aiida-core==1.0.0`

aiida/cmdline/commands/cmd_status.py
```diff
@@ -14,13 +14,26 @@
...
+from contextlib import contextmanager
 import click
 
 from aiida.cmdline.commands.cmd_verdi import verdi
 from ..utils.echo import ExitCode
 
 
+@contextmanager
+def set_log_level(logger, level):
+    'Temporarily adjust the log-level of logger.'
```
sphuber: For docstrings we try to follow the conventions of [PEP257](https://www.python.org/dev/peps/pep-0257/). So in this case the single quotes should be replaced with triple double quotes, i.e. `"""Temporarily adjust the log-level of logger."""`

aiida/cmdline/commands/cmd_status.py
```diff
@@ -14,13 +14,26 @@
...
 from enum import IntEnum
+from contextlib import contextmanager
 import click
 
 from aiida.cmdline.commands.cmd_verdi import verdi
 from ..utils.echo import ExitCode
 
 
+@contextmanager
+def set_log_level(logger, level):
```
sphuber: I think this utility function might be nice to put in `aiida.common.log` so it can be reused. Maybe good to check the code if something similar is already done and replace with this utility

aiida/cmdline/commands/cmd_status.py
```diff
@@ -90,8 +103,9 @@ def verdi_status():
     # getting the rmq status
     try:
         with Capturing(capture_stderr=True):
-            comm = manager.create_communicator(with_orm=False)
-            comm.stop()
+            with set_log_level(logging.getLogger('kiwipy'), logging.CRITICAL):
```
sphuber: Looking at the original issue #2914 I wonder if this will be enough.
For one it seems to throw log messages in two checks:
 * postgres
 * rabbitmq

Apparently, for the code path of the first, it also hits the `self._communicator.connect()` which tries to establish a connection to the RabbitMQ and so that fails just as the RabbitMQ test itself fails. It would be good if we can figure out why the database test hits code that tries to connect to RabbitMQ and see if we can isolate it to have a more pure database-only test.

Secondly, the log messages do not just come from `kiwipy`, but also from `tornado` and `pika`, two other libraries that `kiwipy` depends on. Are we sure that all messages will now be surpressed? I know this is very difficult to put in a unit test, but did you test this locally?

Should we maybe update the `set_log_level` context manager to simply suppress all loggers for the time being? Or capture it and redirect it optionally if the `Exception` clause is triggered?

aiida/cmdline/commands/cmd_status.py
csadorf: Got it. This was automatically changed from double to single quotes by the pre-commit hook.

aiida/cmdline/commands/cmd_status.py
csadorf: I checked that module and didn't see anything similar, but I'll check again and make an assessment whether this is worth generalizing.

csadorf: I searched the code and found that many tests use the following pattern:
```
logging.disable(level=logging.CRITICAL)
# noisy code
logging.disable(level=logging.NOTSET)
```
I think that's fine as long as it's properly placed in `try: ... finally:` blocks and has the added benefit that it suppresses logging from **all** modules.

I'll adjust this patch accordingly.

aiida/cmdline/commands/cmd_status.py
csadorf: > Looking at the original issue #2914 I wonder if this will be enough.
> For one it seems to throw log messages in two checks:
> 
> * postgres
> * rabbitmq
> 
> Apparently, for the code path of the first, it also hits the `self._communicator.connect()` which tries to establish a connection to the RabbitMQ and so that fails just as the RabbitMQ test itself fails. It would be good if we can figure out why the database test hits code that tries to connect to RabbitMQ and see if we can isolate it to have a more pure database-only test.

I'd argue that this is out-of-scope for this PR.

> Secondly, the log messages do not just come from `kiwipy`, but also from `tornado` and `pika`, two other libraries that `kiwipy` depends on. Are we sure that all messages will now be surpressed? I know this is very difficult to put in a unit test, but did you test this 

I've tested this locally. How are the command line entrypoints tested at the moment?

> Should we maybe update the `set_log_level` context manager to simply suppress all loggers for the time being? Or capture it and redirect it optionally if the `Exception` clause is triggered?

Agreed, which is why I've adjusted the patch as described in my comment above.

csadorf:
I've kept the original per-logger adjustment approach as a separate commit in case that we want to go that route after all, but I'd recommend to squash this branch prior to merging.

aiida/cmdline/commands/cmd_status.py
sphuber: I agree that removing the `self._communicator.connect()` path from the postgres connect pathway is out of the scope, however, it is resulting in error logs printed to screen in `verdi status` _does_ seem to be in scope. It is almost exactly the title of the issue. All I would suggest is to place the same temporary logging disabling around the attempt to connect to postgres as well. It is the try-catch block just above the one you are already targeting.

The `verdi` command tests are in `aiida.backends.tests.cmdline.commands`. They are run through `verdi -p <test_profile> devel tests db.cmdline.commands`.

aiida/cmdline/commands/cmd_status.py
sphuber: I liked your context manager approach a lot better though. What I suggest is that you place that in `aiida.common.log` and have it just disable logging completely for its duration, just as you have done now locally in the status command. This utility can then replace the same pattern you found in the other tests.

aiida/common/log.py
```diff
@@ -201,3 +202,15 @@ def configure_logging(with_orm=False, daemon=False, daemon_log_file=None):
         config['loggers']['aiida']['handlers'].append(handler_dblogger)
 
     dictConfig(config)
+
+
+@contextmanager
+def overide_log_level(level=None):
```
sphuber: overide -> override

aiida/common/log.py
```diff
@@ -201,3 +202,15 @@ def configure_logging(with_orm=False, daemon=False, daemon_log_file=None):
...
+    """Temporarily adjust the log-level of logger."""
+    if level is None:
+        level = logging.CRITICAL
```
sphuber: since `logging.CRITICAL` is an `int` and thus immutable, you can directly set it as the default in the function signature, i.e. `def override_log_level(level=logging.CRITICAL)`

aiida/common/log.py
csadorf: Ah damn.. I was so close to double checking that and then just went ahead. :D

aiida/common/log.py
csadorf: True, I usually avoid that, because it makes it harder to _explicitly_ use the default argument, but in this case it probably makes sense.

aiida/cmdline/commands/cmd_status.py
csadorf: I've made an attempt at implementing a unit test for this issue and have so far not succeeded. Unless you know an elegant way to trigger the faulty behavior, I'd like to merge this as-is.

sphuber: That is fine, this would be tricky to test properly. If you just address the last two comments then we can merge

csadorf:
Should be ready to merge.

PR #3410, opened 2019-10-09
Move and improve `CalcJob` input validation to process spec

### LINKED ISSUE #3395: Exception raised "no computer specified" when submitting a `CalcJob`

sphuber:
Fixes #3395 

Compound input validation for the `CalcJob` was mostly implemented on
the node level in `CalcJobNode._validate`. This has been moved to a
validator on the input portnamespace of the `CalcJob` in order to catch
inputs problems as soon as possible. It implements existing validation:

 * A `Computer` has been defined
 * The parser class, if specified, can be successfully loaded
 * The specified resources are valid and sufficient for the scheduler

In addition, the validation on the `Computer` has been improved. It now
checks that if it is specified both directly in the `metadata` as well
as indirectly through the `Code` input (in the case of a remote code)
that it concerns the same computer.

### DIFF
--- a/aiida/backends/tests/engine/test_calc_job.py
+++ b/aiida/backends/tests/engine/test_calc_job.py
@@ -12,6 +12,9 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
 
+import copy
+import unittest
+
 from aiida import orm
 from aiida.backends.testbase import AiidaTestCase
 from aiida.common import exceptions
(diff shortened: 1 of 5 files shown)


sphuber:
Question: a big part of the validation now done in the `CalcJob` process spec validator is still present in the `CalcJobNode._validate`. In principle users will never instantiate a `CalcJobNode` manually, i.e. without going through a `CalcJob` process. Is it safe to remove the validation from the node class, or is it better to keep the duplication?

giovannipizzi:
Re the validation, if I understand correctly the validation code is still in its place in this PR? Is it possible to call a common function to leave it there without code duplication? I was thinking to prevent storing a CalcJobNode by hand, but I already foresee problems (e.g. when importing etc.) so maybe not a good idea; but on the other hand if it's created "by hand", it will not be able to do much as it will not be picked up by the engine, correct? In this case, maybe we can indeed delete the validation code from the Node class.

Issue #3331, opened 2019-09-19
Graph consistent/recursive exploration tool


ramirezfranciscof:
Both the delete (#3296) and export (#3295) functionalities need to take a list of nodes and get a second list where all connections of a given list of types have been exhausted.
The idea would be to define an interface that both features could use and create a simple engine for it that could later be improved upon.
So in principle this function would take not the optional keywords that delete and export use (since we could later add others that turn on or off multiple of them at the same time), but the ones for the exact kind of links that are optional ("translation" would be in charge of delete/export). That is:

* INPUT_CALC_FORWARD
* CREATE_FORWARD
* CREATE_REVERSED
* RETURN_REVERSED
* INPUT_WORK_FORWARD
* CALL_CALC_FORWARD
* CALL_CALC_REVERSED
* CALL_WORK_FORWARD
* CALL_WORK_REVERSED

I was thinking as implementing it literally as a single function that takes a list of nodes and those logicals and returns the requested list that exhaust all possible paths. I would call it something like `exhaustive_explore` maybe? Comments welcome.

CasperWA:
Sounds like a good idea :+1:

sphuber:
Things that have be improved/merged:
- #3262 `aiida.tools.importexport.dbexport._retrieve_linked_nodes_query` should be moved to generic AGE
- #3262 `aiida.tools.importexport.dbexport.retrieve_linked_nodes` data structures of `links_uuid_dict` are probably unnecessarily complex and should be simplified
- #3262 `aiida.tools.importexport.dbexport.retrieve_linked_nodes` is building queries with redundant information. In principle just knowing the pk of the starting node and the link type to follow is enough. Adding filters on the source and target node type is already encoded perfectly in the link type. The future AGE can leverage this to make this query efficient. It will only have to query the `DbLink` table as all information needed is contained in there.
- #3262 definition of `togglable_link_rules` is done in multiple places and should be put close to or even merged with the `aiida.tools.importexport.common.config.LINK_FLAGS` dictionary.

CasperWA:
>     * #3262 definition of `togglable_link_rules` is done in multiple places and should be put close to or even merged with the `aiida.tools.importexport.common.config.LINK_FLAGS` dictionary.

This I actually already did originally, but it was thought to be too complex.

PR #3418, opened 2019-10-11
[WIP] Generic Graph Traverser

ramirezfranciscof:
Closes issue #3331

Created a general, "exhaustive" (it will iterate the application of the rules until all possible connections are exhausted) and "consistent" (it will apply the same set of rules to all nodes) graph traversal function. The words are between inverted commas because I'm not sure this is the correct way to describe these, but I it's the best I could think of.

The function uses the engine that was previously used in the delete function, and the delete function now calls on this one to get the set of nodes. I didn't yet replaced the engine in the export module because:

1) I'm not sure how this engine performs compared to the one in the export function. And although I agree that optimization follows implementation, in this case we already have an implementation and I wouldn't want to risk making it slower in the process of unifying the engine. I think it would be best to at least compare the performance of these two before replacing the export node selection engine.

2) The interface selected to interact with the delete function is different than the one that would be required for the export function (for example, the export function 'retrieve_linked_nodes' receives and returns not only the nodes, opposed to node pks, but also link information). Although translation is not difficult (just get the pks before the call and load the nodes after the return, etc) it might be inefficient (thus adding to the previous point). More than that, we should probably first talk and discuss what would be the best interface for this utility function (for example, since all verdi commands automatically load nodes, might be better to use that as the input instead of pks).

### DIFF
--- a/aiida/backends/tests/__init__.py
+++ b/aiida/backends/tests/__init__.py
@@ -147,6 +147,7 @@ DB_TEST_LIST = {
         'restapi.identifiers': ['aiida.backends.tests.restapi.test_identifiers'],
         'restapi': ['aiida.backends.tests.test_restapi'],
         'tools.data.orbital': ['aiida.backends.tests.tools.data.orbital.test_orbitals'],
+        'tools.graph.traversers': ['aiida.backends.tests.tools.graph.test_graph_traversers'],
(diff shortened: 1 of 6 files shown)


sphuber:
Superseded by PR #3686

Issue #2794, opened 2019-04-24
hard times in the circus


ltalirz:
I am getting the impression that the development of circus is slowing down to the point where it starts to become a problem for us:

 * still no release with python3.7 support (master branch seems updated, though)
 * latest release forces pyzmq<17.0, which conflicts with the whole jupyter notebook stack, even on python2 (all versions of `notebook` after 5.4.1 (March 2018) require pyzmq>=0.17)
  * Out of the top 10 contributors, 9 did not make a single commit last year https://github.com/circus-tent/circus/graphs/contributors

Is there an alternative in sight?

ltalirz:
mentioning @giovannipizzi @sphuber for comment

ltalirz:
E.g. just `pip install aiida-core[notebook]==1.0.0b2` already has dependency conflicts because of circus

giovannipizzi:
Mmm, that's bad... 
Could you try to ping them to see if they can support pyzmq > 17?
We already spent quite a few weeks to transition to circus, and if possible I'd like to avoid investing a few more weeks of time in switching to (yet another) manager of the daemon processes, if this is possible.

ltalirz:
> Could you try to ping them to see if they can support pyzmq > 17?

An issue on this was opened beginning of february but nothing much is happening
https://github.com/circus-tent/circus/issues/1099#issuecomment-467986991

giovannipizzi:
I see. It is also true that we are using circus just for a small functionality (namely demonising a process), I think.
We could look, at some point, into other libraries to see if they are easy to replace - just to mention one, https://pypi.org/project/python-daemon/ but I didn't investigate/did a thorough analysis of options yet

sphuber:
It is annoying, and we already had a problem quite some time ago, when `tornado==5.0` was released which also broke circus, and that also has still not been addressed. However, I would be hesitant to invest a lot of time in changing the daemonizer now. Even though it is true that what we use it for seems trivial and not a lot, I spent a lot of time, even after the first implementation to get things to work across the various platforms. I fear we might encounter similar problems with a potential replacement that might seem adequate on the surface.

ltalirz:
Ok. Of course I also don't want to cause unnecessary work.
So, we stay with circus for the moment and accept the dependency conflict.

[...]

PR #3424, opened 2019-10-14
Update version requirements of dependencies

### LINKED ISSUE #3423: Incompatible version requirement for seekpath (spglib)
Slightly related to #3184 (in that this also addresses incompatible version requirements).

The following error is now present when installing `aiida-core`:

```shell
ERROR: seekpath 1.9.3 has requirement spglib>=1.14.1, but you'll have spglib 1.12.2.post0 which is incompatible.
```

sphuber:
Fixes #3423 

A new CLI script is added `utils/update_dependencies` that helps with
the procedure that is necessary to update the dependency version
requirements. It is still a manual process but the two CLI commands
should already help quite a bit in the time it takes.


sphuber:
@ltalirz the tests run but the conda builds fail because `django==1.11.25` is not available. We were using `1.11.20` before, so apparently that _was_ available. Do you know how to get `1.11.25` on there? We cannot upgrade to `django>=2.0.0` (or even [v3](https://pypi.org/project/Django/3.0b1/) :smile: ) because that is py3 only.

utils/update_dependencies.py
```diff
@@ -0,0 +1,146 @@
...
+
+    The command will update the `setup.json` to remove all explicit limits, except for those packages specified by the
+    `--exclude` option. After this step, install `aiida-core` through pip with the `[all]` flag to install all optional
+    extra requirements as well. Since there are no explicit version requirements anymore, pip should install the latest
+    available version for each dependency.
+
+    Now run the unit test suite and reinstate dependency version requirements on packages that break the tests. Once all
+    the tests complete successfully, run the following command:
+
+        pip freeze > requirements.txt
```
CasperWA: I guess all from here on down could be condensed to be a part of `python update_dependencies.py update`, and the `requirements` argument can be left out - unless you plan on having several different `requirements.txt` files that should be applied here?

docs/requirements_for_rtd.txt
```diff
@@ -1,73 +1,73 @@
...
 enum34==1.1.6; python_version<'3.5'
 ete3==3.1.1
-flask-marshmallow==0.9.0
+flask-cache==0.13.1
+flask-cors==3.0.8
+flask-httpauth==3.3.0
+flask-marshmallow==0.10.1
+flask-restful==0.3.7
+flask-sqlalchemy==2.4.1
+flask==1.1.1
```
CasperWA: Shouldn't these explicitly be `Flask` and not `flask`?

sphuber:
> I.e., let's say that all these updated version now work for Python3.6+, but how about Python2.7?
> Travis should catch it, sure - but I guess that kind of misses the point with this CLI utility?

You are right, the problem of different python versions is not solved, but that is not the point of this utility though. It would be very complicated to have everything done automatically by the CLI because it would have to create a new virtual env with the various python versions. As the docstrings report, the functionality is limited and there are still some manual operations to perform. However, the amount of manual operations is reduced. Without this CLI, one would have to manually update the setup.json to remove all version requirements before reinstalling. I haven't found a more clever way of telling pip to update all current dependencies to the latest version available. Also the second step of taking the freeze contents and applying them to the setup.json is annoying to do manually. It is these two steps that the CLI automates, which is still better than nothing.

CasperWA:
> @ltalirz the tests run but the conda builds fail because `django==1.11.25` is not available. We were using `1.11.20` before, so apparently that _was_ available. Do you know how to get `1.11.25` on there? We cannot upgrade to `django>=2.0.0` (or even [v3](https://pypi.org/project/Django/3.0b1/) smile ) because that is py3 only.

For this, btw, I believe @astamminger is our go-to-django-feedstock supplier? :) (see feedstock maintainer at the [conda-forge GitHub page](https://github.com/conda-forge/django-feedstock))

docs/requirements_for_rtd.txt
sphuber: I decided for simplicity to convert everything to lowercase. I think pip is capital insensitive.

utils/update_dependencies.py
sphuber: Yes, we could do all in a single command and perform a sys-call to `pip freeze` to get the versions, but since the whole thing is not fully automated anyway, I opted to keep it in two explicit steps. I feel this makes it clearer what is "happening".

CasperWA:
> You are right, the problem of different python versions is not solved, but that is not the point of this utility though. (...) Without this CLI, one would have to manually update the setup.json to remove all version requirements before reinstalling. I haven't found a more clever way of telling pip to update all current dependencies to the latest version available. Also the second step of taking the freeze contents and applying them to the setup.json is annoying to do manually. It is these two steps that the CLI automates, which is still better than nothing.

Agree. This will simplify a lot of manual labour.

I suggest you add a sentence in the `cli()` doc string about how to handle different Python versions.
Perhaps it may be helpful to simply copy over the newly generated `setup.json` in another "clean" virtualenv with the "other" Python version and see that all tests run smoothly there as well. This may result in adapting versions again, and then one would have to copy it back to the starting virtualenv, going back and forth...
This seems like a lot of trouble as well, but is maybe the "easiest", if one wants to avoid Travis? Or we could simply state that this would be the approach if the first Travis run fails for "the other" Python version?

astamminger:
> For this, btw, I believe @astamminger is our go-to-django-feedstock supplier? :)

@CasperWA that's damn right! :D
@sphuber I'll add ``django==1.11.25`` to conda ASAP!

sphuber:
> I suggest you add a sentence in the `cli()` doc string about how to handle different Python versions.
> Perhaps it may be helpful to simply copy over the newly generated `setup.json` in another "clean" virtualenv with the "other" Python version and see that all tests run smoothly there as well. This may result in adapting versions again, and then one would have to copy it back to the starting virtualenv, going back and forth...
> This seems like a lot of trouble as well, but is maybe the "easiest", if one wants to avoid Travis? Or we could simply state that this would be the approach if the first Travis run fails for "the other" Python version?

In principle the workflow would have to consist of creating two new virtual envs for each python version, but in practice one can get away with doing one and running the tests locally. Since the CLI only deals with requirements that have no additional qualifiers, all packages with `; python >= 3` are not touched. Chances that the others will have conflicts in python versions are then relatively small. Of course these will have to be updated manually as it is very difficult to automate this in any way shape or form

sphuber:
Thanks a lot @astamminger ! And at @CasperWA I added a bit of text in the docstring about different python versions.

ltalirz:
sorry for not participating in the discussion - I'm busy in moscow at the moment

I just wanted to briefly mention two tools that may be helpful/relevant in this context:

 * the tool might ideally want use here is pipenv (https://github.com/pypa/pipenv) - you write a Pipfile with only necessary requirements, potentially with some limits, and from that can generate the fully concretized dependencies. However, we already have conflicting dependencies, and currently it's difficult in pipenv to resolve this (see also https://github.com/aiidateam/aiida-core/issues/2794)
 * pipdeptree: an easy way to find out why package x was installed on your system

happy to look at this PR when I'm back (thursday)

docs/requirements_for_rtd.txt
CasperWA: > I think pip is capital insensitive.

Yeah, I couldn't find anything on PyPI that says it is - there was only something about case sensitivity for entry points in pip.

CasperWA:
> In principle the workflow would have to consist of creating two new virtual envs for each python version, but in practice one can get away with doing one and running the tests locally. Since the CLI only deals with requirements that have no additional qualifiers, all packages with `; python >= 3` are not touched. Chances that the others will have conflicts in python versions are then relatively small. Of course these will have to be updated manually as it is very difficult to automate this in any way shape or form

This is what I was implying as well, with the separate virtualenvs.
Btw, could we do something smart with the `; python >= 3` as well, i.e., figure out the python version in the cli functions and update the requirements with this knowledge as well? Or maybe this is an addition for a later revision/addition of this new tool?
At least for now, it's fine with me if we keep those few packages out of this update-workflow - but they should definitely be considered later.

PR #3428, opened 2019-10-16
Include breadcrumbs in exceptions thrown by `PortNamespace.serialize`

### LINKED ISSUE #3427: Include namespace breadcumbs if `PortNamespace.serialize` throws `TypeError`
Currently it looks like:
```
TypeError: Got object of type '<class 'aiida.orm.nodes.data.dict.Dict'>', expecting '<class 'collections.abc.Mapping'>'
```
and figuring out which namespace was the violator is difficult for deeply nested namespaces.

sphuber:
Fixes #3427 

Having the full namespace path in the exception message becomes
invaluable when serializing deeply nested namespaces.

### DIFF
--- a/aiida/engine/processes/ports.py
+++ b/aiida/engine/processes/ports.py
@@ -17,8 +17,6 @@ import re
 
 from plumpy import ports
 
-from aiida.common.lang import type_check
-
 __all__ = (
     'PortNamespace', 'InputPort', 'OutputPort', 'CalcJobOutputPort', 'WithNonDb', 'WithSerialize',
     'PORT_NAMESPACE_SEPARATOR'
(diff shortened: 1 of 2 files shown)


aiida/engine/processes/ports.py
```diff
@@ -185,25 +182,35 @@ def validate_port_name(port_name):
...
+        It will recursively call this function on any nested `PortNamespace` or the serialize function on any `Ports`.
 
         :param mapping: a mapping of values to be serialized
+        :param breadcrumbs: a tuple with the namespaces of parent namespaces
         :returns: the serialized mapping
         """
+        from plumpy.ports import breadcrumbs_to_port
+
+        breadcrumbs += (self.name,)
+
```
espenfl: Maybe put after return on `None`?

sphuber: Done

Issue #3411, opened 2019-10-10
Error when retrieving an irregular multidimensional array


d-tomerini:
AiiDA does not allow to retrieve irregularly shaped arrays (list of lists with different lengths) when saved to disk. Numpy saves them as pickles and does not allow retrieval without an `` allow_pickle=True`` flag.

This results in being able to save them to disk, but not to load them afterwards.

Test code by @sphuber 

```
import numpy
from aiida.orm import ArrayData
data = [[1, 1, 1], [2, 2], [3, 3, 3, 3]]
array = ArrayData()
ndarray = numpy.array([numpy.array(element) for element in data])

array.set_array('scf_convergence', ndarray)
array.store()
array.get_array('scf_convergence') 
```

greschd:
This seems to apply for all ``numpy`` arrays with ``dtype=object``. For example, ``np.array([1, 'a', dict()])`` has the same behavior.

But, I'd be quite sceptical of allowing ``pickle`` as a long-term storage format. Unpickling breaks as soon as one of the classes used in the data structure changes names, for example. It should probably be detected already when calling ``set_array``, though.

In this case, wouldn't it be possible to just use a ``List`` instead?

sphuber:
The thing I don't understand is why it would have a `dtype=object` in the example I gave. I explicitly make all lists numpy arrays. So every element should be a numpy type. Even the integers are `numpy.int64`. So why is it than still storing as a pickle? I even tried to call `numpy.save` manually with `allow_pickle=False` and that also fails with the same exception.

greschd:
Yes, numpy arrays are supposed to be homogeneous -- that's necessary for them to be contiguous in memory with O(1) random access.

The example given doesn't produce a "single" numpy array, but rather an "array of arrays" that contains the lower arrays as objects.

greschd:
Fun side note: Strings in numpy arrays are automatically truncated (to the size set when the array is created).

```python
In [32]: x = np.array(['a', 'a', 'a'])                                                                                  
In [33]: x[0] = 'test'                                                                                                  
In [34]: x                                                                                                              
Out[34]:                                                                                                                
array(['t', 'a', 'a'],
      dtype='<U1')
```

sphuber:
[...]

PR #3434, opened 2019-10-17
Disallow pickle when storing numpy array in ArrayData

greschd:
Following the discussion in #3411.

### DIFF
--- a/aiida/orm/nodes/data/array/array.py
+++ b/aiida/orm/nodes/data/array/array.py
@@ -165,7 +165,7 @@ class ArrayData(Data):
 
         # Write the array to a temporary file, and then add it to the repository of the node
         with tempfile.NamedTemporaryFile() as handle:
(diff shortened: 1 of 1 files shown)


greschd:
We could also add ``allow_pickle=False`` on the load method: In principle it's not necessary because the ``numpy`` version required by ``aiida-core`` does that already, but it's possible some people have an older version installed for some reason.

sphuber:
If we decide that this is always the behavior we want, I would vote for also putting it explicitly on the load method, just in case `numpy` ever decide to change the default (not very likely, but still).

greschd:
Ok, I updated it accordingly.

sphuber:
@giovannipizzi I would be OK with this change with the idea "better safe than sorry".

sphuber:
Thanks a lot @greschd

Issue #3436, opened 2019-10-17
use `verdi node graph generate` to understand effect of node deletion / export


ltalirz:
It is often difficult for new users to predict the effect of `verdi node export` or `verdi node delete` on a given node. It would be extremely useful, if one could use `verdi node graph generate` with a corresponding `--export/--delete` option (or similar) to understand what would be happening, i.e.:

 - [ ] Given a node identifier, draw the graph of the export file created by `verdi export create -N <identifier>` (in order to check whether it will contain everything necessary)
 - [ ] Given a node identifier, draw the graph of the nodes that will be deleted by `verdi node delete <identifier>`

Currently, this is not really possible since the logic of graph generation and node deletion/export are slightly different.

Furthermore, it would seem wise to harmonize the cli options of node deletion & export (reproduced below), since these operations are kind of "dual" to each other.
Graph generation is a different operation, but reusing some options here might make sense as well (less important).

Finally, I suggest to make the help strings of the options more self-explanatory, e.g. replacing
```
Toggle the call_calc_backward graph traversal rule 
  used in computing the complete node set.[default: False]
```
with 
```
Whether to follow CALL links to calculations backwards 
when constructing the node set. [default: False]
```

Mentioning @chrisjsewell @giovannipizzi @CasperWA for comment and @danieleongari for info.

### Node deletion
```
$ verdi node delete -h
  --call-calc-forward / --no-call-calc-forward
                                  Toggle the call_calc_forward graph traversal
                                  rule used in computing the complete node
                                  set.  [default: False]
  --call-work-forward / --no-call-work-forward
                                  Toggle the call_work_forward graph traversal
                                  rule used in computing the complete node
                                  set.  [default: False]
  --create-forward / --no-create-forward
                                  Toggle the create_forward graph traversal
                                  rule used in computing the complete node
                                  set.  [default: True]
```

### Export
```
$ verdi export create -h
...
  --call-calc-backward / --no-call-calc-backward
                                  Toggle the call_calc_backward graph
                                  traversal rule used in computing the
                                  complete node set.  [default: False]
  --call-work-backward / --no-call-work-backward
                                  Toggle the call_work_backward graph
                                  traversal rule used in computing the
                                  complete node set.  [default: False]
  --create-backward / --no-create-backward
                                  Toggle the create_backward graph traversal
                                  rule used in computing the complete node
                                  set.  [default: True]
  --input-calc-forward / --no-input-calc-forward
[...]

PR #3438, opened 2019-10-18
update help strings for graph traversal

ltalirz:
as mentioned in #3436 

* make graph traversal rules a bit more self explanatory
* update help texts of `verdi node delete` and `verdi export create`


ltalirz:
@sphuber So far that's a quick fix, with all the options that were necessary to make the `verdi node delete`  and `verdi export create` commands work.

I see now that other rules exist as well - happy to add help strings for those as well.

aiida/cmdline/commands/cmd_export.py
```diff
@@ -89,10 +89,13 @@ def create(
     create_backward, return_backward, call_calc_backward, call_work_backward, include_comments, include_logs
 ):
     """
-    Export parts of the AiiDA database to file for sharing.
+    Export subsets of the provenance graph to file for sharing.
 
-    Various entities can be exported, such as Codes, Computers, Groups, Nodes,
-    Comments, Logs, ...
+    Besides Nodes of the provenance graph, you can export Groups, Codes, Computers, Comments, Logs and more.
```
sphuber: "and more" like what? I think we have listed everything there.
Also technically "Codes" are nodes, but I can see how it makes sense to list them invidivually given that there is a separate flag, so fine to keep.

aiida/cmdline/params/options/__init__.py
```diff
@@ -33,6 +33,18 @@
     'WITH_ELEMENTS_EXCLUSIVE'
 )
 
+TRAVERSAL_RULE_HELP_STRING = {
+    'call_calc_backward': 'CALL links to calculations backwards',
```
sphuber: There are 3 rules missing. I would just add them so that if at some point they are made toggleable this won't crash

aiida/cmdline/params/options/__init__.py
```diff
@@ -65,7 +77,9 @@ def decorator(command):
             if traversal_rule.toggleable:
                 option_name = name.replace('_', '-')
                 option_label = '--{option_name}/--no-{option_name}'.format(option_name=option_name)
-                help_string = 'Toggle the {} graph traversal rule used in computing the complete node set.'.format(name)
+                help_string = 'Whether to follow {} when constructing the node set.'.format(
```
sphuber:
```suggestion
                help_string = 'Whether to follow {} when constructing the full node set.'.format(
```

aiida/cmdline/commands/cmd_export.py
ltalirz: Ok, will delete (this was just a translation of the ...)

aiida/cmdline/params/options/__init__.py
ltalirz: will add!

aiida/cmdline/params/options/__init__.py
ltalirz: I deleted `full` on purpose since the meaning wasn't so clear to me.
I do get what you mean now ... how about this:
`Whether to follow {} when selecting the nodes to export.`

I think this is quite self-explanatory.

ltalirz: Hm... that would be specific to export. Now going with 
` Whether to expand the node set by following {}.`

ltalirz:
> I was really confused there for a moment because I started working on the review and requested to change 'AiiDA graph' to 'provenance graph' and when I came back it was already changed. Scratched my head for a second thinking how you could have read my comment before submitting the review :)

Hehe, yes - I guess that just sprung directly in the eye when opening the "files" tab ;-)

sphuber:
Thanks @ltalirz

PR #3440, opened 2019-10-18
Introduce `count()` for entity Collections

CasperWA:
This is useful for a number of reason, one being a quick way of getting a count of how many entities there are of a specific AiiDA entity, but further because this should introduce some optimizations for the Collection helper methods.


aiida/orm/entities.py
```diff
@@ -144,25 +160,32 @@ def find(self, filters=None, order_by=None, limit=None):
...
         """
         Get all entities in this collection
 
-        :return: A collection of users matching the criteria
+        :return: A list of all entities
         :rtype: list
         """
         return [_[0] for _ in self.query().all()]
 
+    def count(self, filters=None, limit=None, offset=None):
```
sphuber: do `limit` and `offset` make sense in the context of `count`? Couldn't you just do `max(Entity.objects.count() - offset, limit)` if you really wanted to bound the answer? Or is this for efficiency reasons?

aiida/backends/tests/orm/test_entities.py
```diff
@@ -28,3 +28,15 @@ def test_collections_cache(self):
         # Now check passing an explicit backend
         backend = user_collection.backend
         self.assertIs(user_collection, user_collection(backend))
+
+    def test_collections_count(self):
+        """Make sure we count() works for collections"""
```
sphuber:
```suggestion
        """Make sure count() works for collections"""
```

aiida/orm/entities.py
```diff
@@ -95,18 +95,34 @@ def entity_type(self):
...
+        :param limit: the maximum number of results to return
+        :type limit: int
+
+        :param offset: number of initial results to be skipped
+        :type offset: int
+
         :return: a new query builder instance
         :rtype: :class:`aiida.orm.QueryBuilder`
         """
         # pylint: disable=no-self-use
```
sphuber: This pylint can now be removed

aiida/orm/entities.py
CasperWA: I guess one could, but at the same time, since `limit` and `offset` are already implemented in the `QueryBuilder` and they are not mandatory for `.objects.count()`, it doesn't seem to be a big issue.
The only problem I can see is a definition question of why the `QueryBuilder` `count()` method does not have similar parameter definition as `.objects.count()` - however, I see this as being fine, since I view the two methods to be philosophically different (although both are `QueryBuilder`s).
In the end, I prefer the elegance of this implementation (`Entity.objects.count(offset=offset, limit=limit)`), instead of `max(Entity.objects.count() - offset, limit)` :sweat_smile:

sphuber: > The only problem I can see is a definition question of why the `QueryBuilder` `count()` method does not have a similar parameter definition as `.objects.count()`

`QueryBuilder.count` does not have those arguments because they don't really make sense for that function.

> In the end, I prefer the elegance of this implementation (`Entity.objects.count(offset=offset, limit=limit)`), instead of `max(Entity.objects.count() - offset, limit)`

What is the point of `Entity.objects.count(offset=offset, limit=limit)` though. Why would one ever want to do that?

CasperWA: > What is the point of `Entity.objects.count(offset=offset, limit=limit)` though. Why would one ever want to do that?

Well --- now that you put it that way (and I actually am forced to think it through) you do make a lot of sense. Hmm. Limit definitely doesn't make sense. Offset, sort of, although it is as simple to take the difference with the total count, for sure. Hmm. Yes, I see. Silly of me.
There shouldn't be any large optimization benefits of doing it like this either, I think? ... I will remove them as parameters.

aiida/orm/entities.py
```diff
@@ -118,15 +134,15 @@ def get(self, **filters):
 
         :return: the entry
         """
-        res = self.find(filters=filters)
-        if not res:
+        res = self.query(filters=filters, limit=1)
+        if res.count() == 0:
```
sphuber: Getting the count twice, here and two lines below, will most likely not be very efficient, unless we cache the result under the hood, which I doubt. So unless you can verify that the database is called once, maybe it is best to call `count` once, if you want to use that new interface

CasperWA: Sure.

Looking at this method (`get`), I guess it's point has been superseded by `find`? It's point now is to return a single entity object through a `QueryBuilder` filter. However, the rigidity of this is not properly implemented I guess - and does the exception message for the `res.count() > 1` even return something meaningful? `id` is not a property of the `Collection` I guess, but rather of `Entity`...
It would make sense to simply remove `get` from `Collection` and have `Entity.get()` create a specific filter to be used with `find`, right? Or?

sphuber: > I guess it's point has been superseded by `find`?

No. The intention of `get` is to get a single specific entity of the collection. It should be uniquely resolvable, i.e. if none or more than one are found and exception should be raised. This should therefore also be a method on the collection and not the entity class itself.

aiida/orm/entities.py
```diff
@@ -118,15 +133,16 @@ def get(self, **filters):
 
         :return: the entry
         """
-        res = self.find(filters=filters)
-        if not res:
+        res = self.query(filters=filters, limit=1)
```
sphuber: This is incorrect the `limit=1` will now hide if there are multiple objects, i.e. your exception in line 141 will never be hit

sphuber: By the way, check `QueryBuilder.first`, probably that already includes all the necessary logic to do what is done here to check for a uniquely resolvable identifier.

CasperWA: > By the way, check `QueryBuilder.first`, probably that already includes all the necessary logic to do what is done here to check for a uniquely resolvable identifier.

Good idea.

Hmm, concerning my longer comment above. Shouldn't a filter then be made in the `Entity` class' `get` method that calls `Collection.get()`?

CasperWA: Simply using the `.one()` method from `QueryBuilder`, which indeed has the logic already in place.

Concerning the `Entity.get()` filter, I guess it doesn't make sense, but maybe the method name should be changed to show that _only_ one entity is expected to be found and returned. E.g., `get_one()` or something?

sphuber: Of course I meant `one`, my bad. The name of `get` is actually modeled after [Django](https://docs.djangoproject.com/en/2.2/topics/db/queries/#retrieving-a-single-object-with-get) and is the common name for this type of query

CasperWA: > Of course I meant `one`, my bad. The name of `get` is actually modeled after [Django](https://docs.djangoproject.com/en/2.2/topics/db/queries/#retrieving-a-single-object-with-get) and is the common name for this type of query

Doesn't make it clearer - only makes it traditional and conservative ... :unamused: :sweat_smile: Nah, I see your point. Let's keep it like this, but just use `one()` (as is now committed) and perhaps consider updating the method documentation for `Entity.get()` and perhaps others where/if needed?

PR #3441, opened 2019-10-18
Collapsible namespace in sphinxext

greschd:
Use the HTML ``<details>`` tag to create collapsible namespaces in the sphinx extension for process documentation.
 
All collapsed:
image

All open:
image

We also discussed (#3300) directly making the heading clickable (without the extra "Namespace Ports" line), but I couldn't get this to work -- at least not without messing with the CSS. Although I'm not exactly an HTML wizard, so a more capable person might be more successful. Changing CSS is a bit of a risk IMO, it might break when the theme is updated, or people could decide to use a different theme for their plugin docs.

Note also: Edge doesn't support ``<display>``, it will show like this:
image
But that should probably be resolved soon, with the Chrome-based Edge.

I wouldn't consider #3300 resolved, this is just an incremental improvement.


greschd:
If you're wondering about the spacing after the namespace port list, that seems to be just how that particular theme shows nested bullet lists - I did fix am inconsistency between regular ports and port namespaces, though.

aiida/sphinxext/process.py
```diff
@@ -125,45 +125,57 @@ def build_portnamespace_doctree(self, port_namespace):
...
-        Build the paragraph that describes a single port.
+        Build the content that describes a single port.
         """
-        paragraph = nodes.paragraph()
-        paragraph += addnodes.literal_strong(text=name)
-        paragraph += nodes.Text(', ')
-        paragraph += nodes.emphasis(text=self.format_valid_types(port.valid_type))
-        paragraph += nodes.Text(', ')
-        paragraph += nodes.Text('required' if port.required else 'optional')
+        res = []
```
ltalirz: I guess this switch from `paragraph` was necessary to make the spacing consistent?
Perhaps a comment to explain this might be useful.

greschd: Yeah, that's exactly why. I did document that in the commit message, but since the paragraph is no longer there in the code I'm not sure a code comment would really be understandable.

setup.json
```diff
@@ -83,8 +83,11 @@
       "markupsafe==1.1.1",
       "pygments==2.4.2",
       "sphinx-rtd-theme==0.4.3",
-      "sphinx==1.8.5",
-      "sphinxcontrib-contentui==0.2.2"
+      "sphinx==1.8.5; python_version<'3'",
+      "sphinx==2.2.0; python_version>='3.0'",
+      "sphinxcontrib-contentui==0.2.2; python_version<'3'",
+      "sphinxcontrib-contentui==0.2.4; python_version>='3.0'",
```
greschd: @ltalirz this is the main change compared to before: For the Python3 version, Sphinx needed to be updated (to at least 2.0) to be compatible with ``sphinxcontrib-details-directive``, and that in turn required an update to ``sphinxcontrib-contentui``.

sphuber:
Very nice @greschd , thanks for this.

> We also discussed (#3300) directly making the heading clickable (without the extra "Namespace Ports" line), but I couldn't get this to work -- at least not without messing with the CSS. Although I'm not exactly an HTML wizard, so a more capable person might be more successful. Changing CSS is a bit of a risk IMO, it might break when the theme is updated, or people could decide to use a different theme for their plugin docs.

This would indeed be nice but if it comes at the risk of not working at all in some cases, I agree it is better to leave it for now.

> Note also: Edge doesn't support `<display>`, it will show like this:

I was wondering why you would use Edge, but then I remembered ... :wink: 

Final question: does this work with nested namespaces? Technically your example does contain a doubly nested one, since `inputs` itself is a namespace, but maybe that is treated differently. Did you try putting another namespace in `y` for example?

greschd:
> I was wondering why you would use Edge, but then I remembered ... 

I wouldn't have noticed if it wasn't mentioned in the docs for the ``<detail>`` tag :smile:

> Final question: does this work with nested namespaces? Technically your example does contain a doubly nested one, since `inputs` itself is a namespace, but maybe that is treated differently. Did you try putting another namespace in `y` for example?

Yeah good point -- it works, but I added a nested namespace to the test to make sure.

Screenshot from 2019-10-19 15-56-24

PR #3442, opened 2019-10-19
Define custom object loader that raise `ImportError`

### LINKED ISSUE #3191: Except submitted processes whose class cannot be loaded
Currently the task is just acknowledged if the node cannot be loaded. This is necessary because often the reason for the node failure is that it no longer exists. Since the node cannot be loaded, its state can also not be changed. However, in the case where it is the loading of the class that fails, the node _can_ be loaded. In this case it is better to set the process state to excepted instead of leaving it in created. This will depend on [this issue in plumpy](https://github.com/aiidateam/plumpy/issues/123) to be fixed such that the specific import error can be caught.

sphuber:
Fixes #3191 

The default object loader from `plumpy` was raising `ValueError`
exceptions if the process class could not be loaded. However, catching
the `ValueError` in the `ProcessLauncher._continue` is a bit broad if we
only want to target problems with loading the process class. Therefore
we create our own `ObjectLoader` that raises a `ImportError` instead.

### DIFF
--- a/aiida/engine/persistence.py
+++ b/aiida/engine/persistence.py
@@ -12,6 +12,8 @@
 from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
+
+import importlib
 import logging
 import traceback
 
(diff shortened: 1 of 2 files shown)


muhrin:
Sweet, this tickles my design bone a lot more pleasantly than the `plumpy` PR.

If you're willing to go a little further on the journey with me I have one or two more questions but if you're pressed for time no prob to merge this in.

`<rabbit hole>`
I was looking at the stack trace you posted (thanks! super useful):
```
aiida.manage.external.rmq.ProcessLauncher._continue
plumpy.ProcessLauncher._continue
plumpy.Bundle.unbundle
plumpy.Savable.load
plumpy.DefaultObjectLoader.load_object
```
Look at this, it seems like the whole `ValueError` thing is in general not the right way to go because clearly `Saveable.load` can't react appropriately and so has to propagate up, however whoever is 'up' will just get the super generic `ValueError` it seems to me that even at the `plumpy.ProcessLauncher` point you would want to know that it was a `LoadObjectError` with an appropriate message.

So, my question is this: If the interface were changed so that the object loaders emitted their own error and the `ProcessLauncher` (or above) could use this to distinguish between problems loading objects and any other problems of unbundling would this be enough to avoid having to implement a custom `ProcessLaunche` in AiiDA?  I ask as this would be my ultimate best case goal for this stack of interfaces.
`</rabbit hole>`

sphuber:
> Look at this, it seems like the whole `ValueError` thing is in general not the right way to go because clearly `Saveable.load` can't react appropriately and so has to propagate up, however whoever is 'up' will just get the super generic `ValueError` it seems to me that even at the `plumpy.ProcessLauncher` point you would want to know that it was a `LoadObjectError` with an appropriate message.

This was exactly my point and why I started the PR on `plumpy`. I don't want to just be catching `ValueErrors` in the `ProcessLauncher` and assume it is because the object could be loaded.

> So, my question is this: If the interface were changed so that the object loaders emitted their own error and the ProcessLauncher (or above) could use this to distinguish between problems loading objects and any other problems of unbundling would this be enough to avoid having to implement a custom ProcessLaunche in AiiDA?

Having to implement a custom `ProcessLauncher` in `aiida-core` is not the problem, as we have to do this anyway, regardless of how we deal with the not being able to load the object. If you look at the current implementation it contains a lot more logic specific to `aiida-core`, such as loading the node and cleaning the node up if the `_continue` call excepts. So for me the solution in this PR works fine. Not sure what you would like to change in `plumpy` but if it works there, then I am fine with that as well.

PR #3444, opened 2019-10-19
Include more detail in process state of node summary

### LINKED ISSUE #3443: Include more information on process state in `verdi process show`
If the process is excepted it should report the exception message and if it is failed the exit message.

sphuber:
Fixes #3443 

The node summary is used by `verdi node show` and `verdi process show`
to display basic information about the node. The process state was
already included, but is often not enough. If the process is excepted we
add the exception message and in case of a failed process the exit
message associated with the exit code.

### DIFF
--- a/aiida/cmdline/utils/common.py
+++ b/aiida/cmdline/utils/common.py
@@ -80,8 +80,7 @@ def print_last_process_state_change(process_type=None):
 
 
 def get_node_summary(node):
-    """
-    Return a multi line string with a pretty formatted summary of a Node
+    """Return a multi line string with a pretty formatted summary of a Node.
 
     :param node: a Node instance
     :return: a string summary of the node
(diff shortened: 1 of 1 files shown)


ltalirz:
Ok, after trying it out I see that there are already differences beyond the one that is introduced here (e.g. `verdi node show` showing the Inputs and Outputs for processes).

I'm wondering a bit what our overall strategy is here. Are we trying to have a `verdi node show` command that adapts to the type as much as possible or do we want people to use specific commands for specific nodes?
I guess adaptation is anyhow only possible to a degree, and we need people to provide data cli plugins for their data types.

With this in mind, perhaps it would make sense to limit the output of `verdi node show` to what is general to all nodes?

Otherwise the purpose of having a `verdi process show` command is a bit unclear, since the output currently seems to be the same as the one of `verdi node show`

sphuber:
In the current form it doesn't really make sense that `verdi node show` has identical output to `verdi process show`, i agree. However, I am not sure if we should encourage parsing of their output as that seems really fragile, but if we want this and so we cannot change layout based on type, we _have_ to have separate commands I would say. Having the show for processes without the process state, computer, report hint and/or all the links is useless. However, for a generic node, let's say `Int` this would give a lot of empty lines. So maybe `verdi node show` could be showing only the things common to all node types. However, then we would need a generic `verdi data show` because sometimes it might be nice to see the links coming in and going out of a `Data` node as well.

PR #3446, opened 2019-10-20
Allow passing `retrieved_temporary_folder` to `Parser.parse_from_node`

### LINKED ISSUE #3061: Allow parsing of `retrieved_temporary_folder` to `Parser.parse_from_node`
should be pretty simple :)

sphuber:
Fixes #3061 

A `CalcJob` class can specify a `retrieve_temporary_list`, the files
of which will be retrieved in a separate temporary folder that is then
passed by the daemon to the `Parser.parse` function in the keyword
arguments as `retrieved_temporary_folder`. To allow one to pass the
same temporary folder through the `Parser.parse_from_node` wrapper, the
`retrieved_temporary_folder` keyword is added to the signature.


sphuber:
@chrisjsewell could you please verify that this works for your use-case?

PR #3459, opened 2019-10-23
Modification of graph traversal rules for node export and deletion

### LINKED ISSUE #2637: Changing the default behaviour of the export set expansion rules
In this issue we can discuss the default behaviour of the export rules.

These rules, based on a given set of nodes, they expand it to include nodes that are "expected" to be exported too.

The rules are described at issue #1102 as well as the default behaviour of these rules, and which of the they can be overridden.

Some ideas on how they can be changed are proposed at #2632 by @ltalirz 
https://github.com/aiidateam/aiida_core/pull/2632#issuecomment-475951114

ramirezfranciscof:
Export will now follow `call_calc` and `call_work` links in the backward direction by default.
Changed the defaults, and adapted the documentation and the tests.
A lot of the tests for the export feature now ended up being very similar (since they were designed for a different behavior); I only "fixed" them (so that they pass), but we might want to consider in the future removing some or maybe adding others.

closes  #2637 (correct?)


ltalirz:
@giovannipizzi  I discussed with @ramirezfranciscof and @sphuber concerning the corresponding `deletion` rules.

Currently, deleting a workflow node does not touch the underlying data provenance, and I believe this will be counterintuitive to new users. 
Say, I run a workflow and realize I made some mistake in the input. I want to delete the workflow, so I pass the PK of the workflow node and expect that all the nodes it generated would be gone as well.
With the current default settings, I will create a mess because I've now lost the information of how the data provenance was generated (which might not even be properly linked everywhere).

The reason for the current defaults setting is that deleting a sub-workflow always "bubbles up" to the top workflow, which users may not always realize. 
Currently, this just gets rid of the logical provenance of the top-workflow, but if we change the default, it would also get rid of the underlying data provenance.

I would still vote to change the default to delete the data provenance as well, and rather focus on making it easier for users to understand what they are deleting (e.g. by making it possible to graph those nodes).
@sphuber agrees that, if we don't change the default, he would basically always turn on the corresponding flags.
Finally, this will make the defaults of node deletion more symmetric with export (except for `return`).

What do you say?

giovannipizzi:
I like symmetry and I see the reasoning, but I am a bit worried by the automatic deletion rules... I fear you get easily in a lot of cascading that ends up deleting a lot of things (like calculations called by top workflows, and their outputs, that is maybe a bit unexpected?). But I see why this default would make sense.

I feel that one reason of the problem is that we apply all rules at each iteration of the recursion. Probably there should be some rules that are applied recursively, and others only at the end, even if this probably requires some thinking and a better graph traversal tool that @ramirezfranciscof is working on but won't be ready by 1.0.

Anyway, in the end I think it's ok to change the default if this makes things more consistent, and ideally if you find a good way to alert people that the cascading rules are deleting a lot of things, it would be great (but maybe what is written is already ok, I let you judge).

ltalirz:
Thanks for the feedback!

@ramirezfranciscof I think it makes sense to do these changes in the same PR.
I'll review it in one go then.

docs/source/concepts/provenance.rst
```diff
@@ -281,29 +282,27 @@ The results of a parent workflow depend critically on the subworkflows or calcul
...
+The possibility to follow ``call`` links in the other direction is enabled by default, which causes that targeting a given process for either export or deletion will also include in the process not only all of his child processes, but also all the childs of any of its parent processes as well.
+As a result of all ``call`` links being traversed in both directions, targeting any of the process nodes in a workflow will mean the inclusion of the other processes of that workflow as well.
+The option of disabling the traversal of the ``call`` links in one of the directions (forward if deleting, backward if exporting) provides the users with the ability to select parts within their workflows to affect, but this usually requires a much deeper understanding of the conectivity of the nodes and the complex behavior that arises from the combination of the different traversal rules (see below).
```
ltalirz:
```suggestion
Users can disable the traversal of ``call`` links in one of the directions (forward for deletion, backward for export) for fine-grained control (see examples below).
```

docs/source/concepts/provenance.rst
```diff
@@ -320,64 +319,71 @@ Let us first focus on the data provenance only (i.e., only ``input_calc`` and ``
...
    :scale: 80%
 
-Notice that we arrived at this result through the following traversal rules (illustrated by the red arrows in the figure):
-
-* |D_3| will be deleted because |C_1| is being deleted (``create_forward=True``).
-* |W_1| will be deleted because |C_1| is being deleted (``call_calc_backward=True``).
-* |W_0| will be deleted because |W_1| is being deleted (``call_work_backward=True``)
+The nodes |W_1| and |W_2| would be included because |W_0| is being targeted (``call_work_forward=True``), then the nodes |C_1| and |C_2| would also be included (``call_calc_forward=True``), and finally the nodes |D_3| and |D_4| would end up being included as well (``create_forward=True``).
+In the end, only the inputs |D_1| and |D_2| remain (since ``input_work_backward=False`` always and ``input_calc_backward=False`` by default).
+The same result would occurr if the user were to target the output nodes instead (intending to delete everything associated with the obtention of those results):
```
ltalirz:
```suggestion
The same result would occur if the user were to target the output nodes instead (intending to delete everything associated with the obtention of those results):
```

docs/source/concepts/provenance.rst
```diff
@@ -320,64 +319,71 @@ Let us first focus on the data provenance only (i.e., only ``input_calc`` and ``
...
-In this particular case, just targeting |C_1| with the default behavior gives the following result, that is probably the desired final state of the database (in the following figures, the dash-circled node is the targeted one, and nodes highlighted in red are those that are eventually deleted):
+As you can see, |W_1| and |W_2| describe two similar but independent procedures that were launched by a single parent workflow |W_0|.
+A typical user would have obtained this by directly running this workflow |W_0| to obtain the results |D_3| and |D_4| from the inputs |D_1| and |D_2|, and may even be unaware of the internal division of |W_0| into two sub-Workflows |W_1| and |W_2|.
+Hence, if the user considers the workflow (meaning, the whole set of nodes produced by it) no longer necessary, the intuitive thing to do in order to remove it from its database would be by targeting the workflow node |W_0| for deletion.
+Indeed, this would produce the desired result:
 
-.. _delexp_example02a:
-.. image:: include/images/delexp_example02-a00.png
+.. _delexp_example02-a01:
+.. image:: include/images/delexp_example02-a01.png
```
ltalirz: Is there a need for this figure, where both D3 and D4 are targeted?

To me it would seem to suffice (and make things shorter) to keep only the next figure, which is identical except that only D3 is targeted.

docs/source/concepts/provenance.rst
```diff
@@ -320,64 +319,71 @@ Let us first focus on the data provenance only (i.e., only ``input_calc`` and ``
...
-.. image:: include/images/delexp_example02-b00.png
+.. _delexp_example02-a03:
+.. image:: include/images/delexp_example02-a03.png
    :scale: 80%
 
-As you see, only |W_0| and |W_1| have been deleted, but |C_1| is still in the database.
-In this case in particular, to achieve the same result as above, it would suffice to enable ``call_calc_forward=True`` to traverse the ``call_calc`` link from |W_1| to |C_1| and then recover our desired result, starting from a different target (|W_1| here, instead of |C_1| above).
-The second workflow |W_2| would still be unaffected because there was no need to forward traverse any ``call_work`` link so far.
+This results from a combination of the rules mentioned in the previous two paragraphs: so the first section (|D_3| to |C_1| to |W_1| to |W_0|) is the same as the last case viewed, and then the rest (|W_0| to |W_2| to |C_2| to |D_4|) is the same as the case when the target is |W_0|.
+This dependency between nodes becomes particularly relevant when, for example, a user with more knowledge of the internal procedures of the parent workflow |W_0| wants to only delete the calculations and results associated to workflow |W_1|.
```
ltalirz: I suggest you introduce a new sub-section before this sentence - e.g.

```
Deleting sub-workflows
------------------------

Say, a user with knowledge of the internal workings of the parent workflow |W_0| wants to delete *only* the calculations and results associated to workflow |W_1|.
With default traversal rules, targeting |W_1| does not produce the desired outcome:
```

docs/source/concepts/provenance.rst
```diff
@@ -320,64 +319,71 @@ Let us first focus on the data provenance only (i.e., only ``input_calc`` and ``
...
-.. _delexp_example02c:
-.. image:: include/images/delexp_example02-c00.png
+.. _delexp_example02-b01:
+.. image:: include/images/delexp_example02-b01.png
    :scale: 80%
 
-But what if some of the child processes of |W_1| are workflows instead of calculations?
-The naive answer would be to enable ``call_work_forward=True`` as well. However, this will delete much more that you might want! In fact, since we are also deleting |W_0|, this last rule would also imply going through the ``call_work`` link between |W_0| and |W_2|, thus producing the following final undesired result, where most nodes have been deleted:
+Indeed |C_1| and |D_4| will be deleted (through ``call_calc_forward`` from |W_1| to |C_1| and ``create_forward`` from |C_1| to |D_3|), but so will |W_0| (through ``call_work_reverse`` from |W_1|), |W_2| (``call_work_forward`` from |W_0|), |C_2| (``call_calc_forward`` from |W_2|) and |D_4| (``create_forward`` from |C_2|).
+The way to achieve the desired outcome is not at all obvious, although in this particular case one could propose some simple 'hacks', like targeting |W_1| setting the toggable flag ``call_work_forward=False`` (preventing the traversal from |W_0| to |W_2|):
```
ltalirz:
```suggestion
In this particular case, deletion of |W_2| could be avoided by setting the flag ``call_work_forward=False`` (preventing the traversal from |W_0| to |W_2|):
```

docs/source/concepts/provenance.rst
```diff
@@ -320,64 +319,71 @@ Let us first focus on the data provenance only (i.e., only ``input_calc`` and ``
...
+.. image:: include/images/delexp_example02-b02.png
    :scale: 80%
 
-So what can you do in the general case where you want to delete all processes (calculations and workflows) contained under |W_1| without affecting |W_2|?
-First you would need to get rid of the connection between these two nodes by targetting the node |W_0| (with the default keywords, and in particular ``call_work_forward=False``).
-This will delete only |W_0| and no other node.
-
-Only then you can target |W_1| by activating the keywords to include all call links to its subprocesses (``call_work_forward=True`` and ``call_calc_forward=True``).
-After this two-step procedure, you will get the desired result for this more general case:
+However, these kind of solutions are not very generalizable (for example, this would no longer work if |W_1| also had other sub-workflows that needed to be deleted as well).
```
ltalirz:
```suggestion
However, this approach is not generally applicable, and wouldn't work if |W_1| had sub-workflows that needed to be deleted as well.
```

docs/source/concepts/provenance.rst
```diff
@@ -320,64 +319,71 @@ Let us first focus on the data provenance only (i.e., only ``input_calc`` and ``
...
+A better approach would be to first delete, using the most conservative setup (disabling all toggable rules), the node that is linking the workflows that the user wishes to now treat separatedly (in this case |W_0|, which is linking |W_1| and |W_2|).
```
ltalirz:
```suggestion
A more general approach is to first sever the connection to |W_2| by deleting node |W_0| with all switchable traversal rules turned off.
```

docs/source/concepts/provenance.rst
```diff
@@ -320,64 +319,71 @@ Let us first focus on the data provenance only (i.e., only ``input_calc`` and ``
...
+Then, once the independence of these two workflows (|W_1| and |W_2|) is explicitly reflected in the graph, the node |W_1| can be targeted for deletion with the default settings and nothing will happen to the procedures in |W_2|.
```
ltalirz:
```suggestion
Then, once the independence of |W_1| and |W_2| is explicitly reflected in the graph, delete node |W_1| with the default settings.
```

docs/source/concepts/provenance.rst
```diff
@@ -320,64 +319,71 @@ Let us first focus on the data provenance only (i.e., only ``input_calc`` and ``
...
-.. _delexp_example02e:
-.. image:: include/images/delexp_example02-e00.png
+.. _delexp_example02-b03:
+.. image:: include/images/delexp_example02-b03.png
    :scale: 80%
 
+It is worth noticing that if the workflow |W_0| was itself a sub-process of a more complicated workflow, all that high-level logic will be deleted and lost due to the non-toggable rule ``call_work_reverse=True``.
+This is an inevitable outcome of deleting part of a workflow, since due to the loss of that information it has become incomplete and it makes no sense to keep it.
 
 The number of possible scenarios and desired outcomes is impossible to cover entirely, but hopefully this example helped to show how you need to analyze the outcome of applying the delete or export procedures in your own cases of interest, especially when not using the default rules.
```
ltalirz:
```suggestion

```

docs/source/concepts/provenance.rst
```diff
@@ -320,64 +319,71 @@ Let us first focus on the data provenance only (i.e., only ``input_calc`` and ``
...
+A better approach would be to first delete, using the most conservative setup (disabling all toggable rules), the node that is linking the workflows that the user wishes to now treat separatedly (in this case |W_0|, which is linking |W_1| and |W_2|).
+Then, once the independence of these two workflows (|W_1| and |W_2|) is explicitly reflected in the graph, the node |W_1| can be targeted for deletion with the default settings and nothing will happen to the procedures in |W_2|.
 
-.. _delexp_example02e:
-.. image:: include/images/delexp_example02-e00.png
+.. _delexp_example02-b03:
+.. image:: include/images/delexp_example02-b03.png
    :scale: 80%
 
+It is worth noticing that if the workflow |W_0| was itself a sub-process of a more complicated workflow, all that high-level logic will be deleted and lost due to the non-toggable rule ``call_work_reverse=True``.
```
ltalirz:
```suggestion
It is worth noting that if the workflow |W_0| was itself part of a higher-level workflow, all that higher-level logic would be deleted due to the non-switchable rule ``call_work_reverse=True``.
```

ramirezfranciscof:
@ltalirz I tried to incorporate all the suggestions, please check if the documentation now looks ok and if so I'll merge it.

@sphuber just to check how you usually do things, now this branch is out-of-date but it also has a review by @ltalirz, so to update it should I merge develop into it or is it also ok to rebase in this situation (I'm not sure what would happen to the whole display here of the PR and the review)?

sphuber:
@ramirezfranciscof in this case it is fine to do it like this. Typically @ltalirz can then perform a review of the changes. If he is happy, he can approve and update the branch here, Github will then merge develop for us, after which it can be merged.

docs/source/concepts/provenance.rst
```diff
@@ -320,66 +320,72 @@ Let us first focus on the data provenance only (i.e., only ``input_calc`` and ``
...
+
+.. _delexp_example02-b01:
+.. image:: include/images/delexp_example02-b01.png
    :scale: 80%
 
-But what if some of the child processes of |W_1| are workflows instead of calculations?
-The naive answer would be to enable ``call_work_forward=True`` as well. However, this will delete much more that you might want! In fact, since we are also deleting |W_0|, this last rule would also imply going through the ``call_work`` link between |W_0| and |W_2|, thus producing the following final undesired result, where most nodes have been deleted:
+Indeed |C_1| and |D_4| will be deleted (through ``call_calc_forward`` from |W_1| to |C_1| and ``create_forward`` from |C_1| to |D_3|), but so will |W_0| (through ``call_work_reverse`` from |W_1|), |W_2| (``call_work_forward`` from |W_0|), |C_2| (``call_calc_forward`` from |W_2|) and |D_4| (``create_forward`` from |C_2|).
+The way to achieve the desired outcome is not at all obvious, although in this particular case one could propose some simple 'hacks', like targeting |W_1| setting the switchable flag ``call_work_forward=False`` (preventing the traversal from |W_0| to |W_2|):
```
ltalirz: I did have a suggestion for this sentence as well that removed the "not at all obvious" and "hacks", which I would probably still prefer...

ramirezfranciscof: I have now changed the "hack" and "not obvious" phrasings. I used a different version of what you proposed, not out of stuborness, but because I keep missing comments/corrections when github hides/colapses them (only now I have noticed it; this is also the reason why I didn't adress it in the previous commit). Luckily I think in this case I still adressed what you wanted change, correct? I will try to pay more attention to that in the future, apologies.

ltalirz: thanks & no worries, happens to me as well!

sphuber:
Thanks a lot to both

Issue #3447, opened 2019-10-21
Docs: restructure "Working with AiiDA" tab


sphuber:
Currently, there is very little to no structure or organization in this tab and it makes it difficult to navigate the index bar on the left

csadorf:
Resolved with merge of #3464 .

PR #3464, opened 2019-10-25
Restructure top-level of documentation

csadorf:
Draft to reduce the number of top-level sections in the core-documentation.

Addresses issue #3447.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -91,31 +91,11 @@ For developers
 
 .. toctree::
    :maxdepth: 1
-   :caption: AiiDA plugin development
+   :caption: Development
    :hidden:
 
-   developer_guide/plugins/basics
-   developer_guide/plugins/quickstart
-   developer_guide/plugins/entry_points
-   developer_guide/plugins/documenting
-   developer_guide/plugins/plugin_tests
-   developer_guide/plugins/publish
-   developer_guide/plugins/update_plugin
-
-.. toctree::
-    :maxdepth: 1
-    :caption: AiiDA core development
-    :hidden:
-
-    developer_guide/core/internals
-    developer_guide/core/transport
-    developer_guide/core/modifying_the_schema
-    developer_guide/core/extend_restapi
-    developer_guide/core/caching
-    developer_guide/core/plugin_system
-    developer_guide/tools/sphinx_cheatsheet
-    developer_guide/tools/visual_aids
-    developer_guide/design/changes
+   developer_guide/plugins
+   developer_guide/core
 
 .. toctree::
     :maxdepth: 1
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -64,23 +64,12 @@ For developers
     install/updating_installation
     install/troubleshooting
 
-
-.. toctree::
-    :maxdepth: 1
-    :caption: Getting started
-    :hidden:
-
-    get_started/index
-    get_started/computers
-    get_started/codes
-
 .. toctree::
     :maxdepth: 1
     :caption: Working with AiiDA
     :hidden:
 
     working_with_aiida/index
-
 
 .. toctree::
     :maxdepth: 1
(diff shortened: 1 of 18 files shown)


csadorf:
@sphuber I'd appreciate some preliminary feedback before I complete this PR.

csadorf:
@sphuber So, I'm done applying the changes we discussed offline with the exception of two:

1. I did not find any sub-sections within the Install/troubleshooting section that could be reasonably be moved into the general "cookbook" section (renamed).
2. You asked me to restore [this paragraph as introduction to processes](https://aiida-core.readthedocs.io/en/latest/working/processes.html#processes) , however I think that this concept is actually already sufficiently explained in the first two paragraphs [here](https://aiida-core.readthedocs.io/en/latest/concepts/processes.html), which is now the intro to the combined section. I may have misunderstood what you meant earlier, so please pay extra attention to that.

docs/source/index.rst
```diff
@@ -64,80 +64,50 @@ For developers
...
-.. toctree::
-    :maxdepth: 1
-    :caption: Terms and Concepts
-    :hidden:
-
-    concepts/provenance
-    concepts/processes
-    concepts/calculations
-    concepts/workflows
+    topics/data
```
ltalirz: I feel the order should be something like:

topics/plugins
topics/computers_codes
topics/cli
topics/scripting
topics/data

Caching could be moved into tips&tricks.
Schedulers also doesn't need to be top-level and could be moved there.
Schedulers and Caching could be moved into tips&tricks

ltalirz:
I like the new organisation!

Except for the order in "Working with AiiDA" I have no further comments.

Two minor suggestions:
 * `REST API for AiiDA` => `AiiDA REST API`
 * `Calculation and work functions`: Would be nice to find something shorter, e.g. `Function decorators`?

sphuber:
>     * `Calculation and work functions`: Would be nice to find something shorter, e.g. `Function decorators`?

That is too generic, could just refer to python function decorators. "Process functions" would be shorter and correct and would have my vote. The only potential problem is that since people really only ever use calculation function and work function that they might overlook it when the section is called process functions.

docs/source/index.rst
csadorf: I have mixed feelings about moving the "Caching" topic to "Tips & Tricks", however, I strongly disagree that "Schedulers" should be part of "Tips & Tricks". That section should be strictly reserved for topics that are not essential to working with AiiDA, however, that does not apply to the topic of "Schedulers" in my opinion.

Finally, I don't think that plugins should be at the very top, but I believe that this would probably need to be hashed out in a separate and larger discussion. For now, I would need clear guidance from senior developers on how to move forward.

@sphuber @giovannipizzi Could you weigh in, please?

ltalirz: > Finally, I don't think that plugins should be at the very top

You could move "installing plugins" to the installation section.
it's kind of important, though, to have a plugin since there isn't much you can do with AiiDA without it.

> I strongly disagree that "Schedulers" should be part of "Tips & Tricks". That section should be strictly reserved for topics that are not essential to working with AiiDA

Ideally, users should not have to care about this. The basic configuration of the scheduler is done at the computer setup, i.e. from there one can link to wherever the more detailed explanation is placed.
I agree that "Tips & tricks" is not how I would call this "collection" - but I would also argue that the page on schedulers does not belong at the same level as a section explaining how to use the cli / shell / manage data etc.

csadorf:
@sphuber Per your request, I've updated this PR to be rebased on develop and to be less invasive overall, however it is in turn also slightly less comprehensive. 

Changes are now limited to modifications of indexes wherever possible and I have only applied minimal changes to the individual changes where this could not be avoided (unless you know a better way).

Compared to the last revision, I have left the verdi-related API sections more or less *as-is*, since they would be more invasive. Please let me know if you would like me to apply those changes after all.

I have also left the "Get started" section intact in order to better accomodate @ltalirz 's notion of how the documentation should be structured.

The branch has no conflicts with develop. Please let me know if you want me to reintroduce any of the other changes from the [previous revision](https://github.com/csadorf/aiida-core/tree/fix/issue-3447-major-revision/docs/source), otherwise this is ready for review.

Issue #2515, opened 2019-02-22
Should Process classes be allowed to contain references to ORM entities?


ltalirz:
At the moment, `Process` classes, such as work chains can contain references to ORM entities, e.g. as default values in the spec:
https://github.com/aiidateam/aiida_core/blob/b22f55155e7a602ac42b00edf62e57e4a458f55b/aiida/backends/tests/work/test_work_chain.py#L94

One problem with this is that it prevents creating a "clean slate" for tests:
 1. Test A imports and runs workchain X (running may not even be necessary).
 2. Database is reset as usual.
 3. Test B imports workchain X and tries to run it. It fails because the default values point to database entries that no longer exist.

As noticed in #2421 this can be hard to debug, since test A and B can be otherwise completely unrelated.

One suggestion by @sphuber  is to switch to lambda functions for defaults of workchains

Issue #3143, opened 2019-07-08
Issue with `aiida.manage.fixtures.FixtureManager.reset_db()` and testing WorkChains 


chrisjsewell:
In [this file](https://github.com/chrisjsewell/aiida-crystal17/blob/de068ac762d969725e8dd3a1d306dd1845f56a95/aiida_crystal17/tests/test_workflows/test_crystal_main.py) I have two tests that call `run_get_node` on a `WorkChain` (a replica of your `PwBaseWorkChain` thanks @shuber!). If I run either test individually, they work fine. But, if I run them in the same session, the 2nd one fails. Also, if I comment out `reset_db()` and run the test file, both tests pass.

```python

aiida_crystal17/tests/test_workflows/test_crystal_main.py:120: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../aiida_core_develop/aiida/engine/launch.py:62: in run_get_node
    return runner.run_get_node(process, *args, **inputs)
../aiida_core_develop/aiida/engine/runners.py:244: in run_get_node
    result, node = self._run(process, *args, **inputs)
../aiida_core_develop/aiida/engine/runners.py:210: in _run
    process = instantiate_process(self, process, *args, **inputs)
../aiida_core_develop/aiida/engine/utils.py:61: in instantiate_process
    process = process_class(runner=runner, inputs=inputs)
//anaconda/envs/aiida-workshop/lib/python3.6/site-packages/plumpy/base/state_machine.py:187: in __call__
    inst.transition_to(inst.create_initial_state())
//anaconda/envs/aiida-workshop/lib/python3.6/site-packages/plumpy/base/state_machine.py:326: in transition_to
    self.transition_failed(initial_state_label, label, *sys.exc_info()[1:])
//anaconda/envs/aiida-workshop/lib/python3.6/site-packages/plumpy/base/state_machine.py:339: in transition_failed
    six.reraise(type(exception), exception, trace)
//anaconda/envs/aiida-workshop/lib/python3.6/site-packages/six.py:693: in reraise
    raise value
//anaconda/envs/aiida-workshop/lib/python3.6/site-packages/plumpy/base/state_machine.py:310: in transition_to
    self._enter_next_state(new_state)
[...]

PR #3465, opened 2019-10-25
Allow the usage of lambdas for `InputPort` default values

### LINKED ISSUE #1604: Allow function as default values for WorkChain input.
Would it be possible to allow the runtime generation of the default node of a Worchain input.

This would allow to have, for example, random seeds stored in inputs node allowing better reproduciblity:
 
```python
cls.input('seed', valid_type=Int, default=lambda: Int(np.random.randint(2**32 - 1)))
```

sphuber:
Fixes #1604 

This is made possible by upgrading `plumpy` to version `0.14.3`. This
functionality is useful to define defaults that should change at each
invocation, for example the generation of a seed. However, it also
provides a way to avoid the issues that arise when defining a default to
be an instance of a `Node`. Since a node instance is always regarded as
a mutable type by python, regardless whether it is stored or not, it can
cause unexpected behavior. For example when the node appointed to the
default value of a port is deleted in the database, the memory instance
remains and points to a no longer existing database row. The next time
it is accessed an exception will be thrown. Using a lambda avoids this
problem.

Edit: this will also address #2515 and #3143 indirectly. It will still require users to replace mutable defaults with lambdas. Therefore I prefer to close them in a separate PR where the `InputPort` constructor implementation will warn when mutable values are specified for the port's default

### DIFF
--- a/aiida/backends/tests/engine/test_ports.py
+++ b/aiida/backends/tests/engine/test_ports.py
@@ -14,7 +14,7 @@ from __future__ import absolute_import
 
 from aiida.backends.testbase import AiidaTestCase
 from aiida.engine.processes.ports import InputPort, PortNamespace
-from aiida.orm import Dict
+from aiida.orm import Dict, Int
 
 
 class TestInputPort(AiidaTestCase):
@@ -98,3 +98,31 @@ class TestPortNamespace(AiidaTestCase):
         # The `assertRaisesRegexp` method is deprecated in python 3 but assertRaisesRegex` does not exist in python 2
         with self.assertRaisesRegexp(TypeError, '.*{}.*{}.*'.format(base_namespace, nested_namespace)):
             port_namespace.serialize({'some': {'nested': {'namespace': {Dict()}}}})
+
+    def test_lambda_default(self):
+        """Test that an input port can specify a lambda as a default."""
+        port_namespace = PortNamespace('base')
+
+        # Defining lambda for default that returns incorrect type should not except at construction
+        port_namespace['port'] = InputPort('port', valid_type=Int, default=lambda: 'string')
+
+        # However, pre processing the namespace, which shall evaluate the default followed by validation will fail
+        inputs = port_namespace.pre_process({})
+        self.assertIsNotNone(port_namespace.validate(inputs))
+
+        # Passing an explicit value for the port will forego the default and validation on returned inputs should pass
+        inputs = port_namespace.pre_process({'port': Int(5)})
+        self.assertIsNone(port_namespace.validate(inputs))
+
+        # Redefining the port, this time with a correct default
+        port_namespace['port'] = InputPort('port', valid_type=Int, default=lambda: Int(5))
+
+        # Pre processing the namespace shall evaluate the default and return the int node
+        inputs = port_namespace.pre_process({})
+        self.assertIsInstance(inputs['port'], Int)
+        self.assertEqual(inputs['port'].value, 5)
+
+        # Passing an explicit value for the port will forego the default
+        inputs = port_namespace.pre_process({'port': Int(3)})
+        self.assertIsInstance(inputs['port'], Int)
+        self.assertEqual(inputs['port'].value, 3)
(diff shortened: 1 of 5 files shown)


aiida/backends/tests/engine/test_ports.py
```diff
@@ -98,3 +98,29 @@ def test_serialize_type_check(self):
...
             port_namespace.serialize({'some': {'nested': {'namespace': {Dict()}}}})
+
+    def test_lambda_default(self):
+        """Test that an input port can specify a lambda as a default."""
+        port_namespace = PortNamespace('base')
+
+        # Defining lambda for default that returns incorrect type should not except at construction
+        port_namespace['port'] = InputPort('port', valid_type=Int, default=lambda: 'string')
+
+        # However, pre processing the namespace, which shall evaluate the default and validate it will fail
```
ltalirz: Would you mind checking this explicitly (that validation raises?)

docs/source/working/processes.rst
```diff
@@ -123,9 +123,19 @@ An example input port that explicitly sets all these attributes is the following
 
 .. code:: python
 
-    spec.input('positive_number', required=False, default=Int(1), valid_type=(Int, Float), validator=is_number_positive)
+    spec.input('positive_number', required=False, default=lambda: Int(1), valid_type=(Int, Float), validator=is_number_positive)
+
+Here we define an input named ``positive_number`` that is not required, if a value is not explicitly passed, the default will be used and if a value *is* passed, it should be of type ``Int`` or ``Float`` and it should be valid according to the ``is_number_positive`` validator.
```
ltalirz:
```suggestion
Here we define an input named ``positive_number`` that should be of type ``Int`` or ``Float`` and should pass the test of the ``is_number_positive`` validator.
If no value is passed, the default will be used.
```

docs/source/working/processes.rst
```diff
@@ -123,9 +123,19 @@ An example input port that explicitly sets all these attributes is the following
...
+
+.. warning::
+
+    In python, it is important to `never use mutable defaults for function arguments <https://docs.python.org/3/reference/compound_stmts.html#function-definitions>`_, since they are instantiated at function definition and reused for each invocation.
```
ltalirz:
```suggestion
    In python, it is good practise to avoid mutable defaults for function arguments, `since they are instantiated at function definition and reused for each invocation. <https://docs.python.org/3/reference/compound_stmts.html#function-definitions>`_
```

docs/source/working/processes.rst
```diff
@@ -123,9 +123,19 @@ An example input port that explicitly sets all these attributes is the following
...
+    This can lead to unexpected results where the default value can change between function calls.
```
ltalirz:
```suggestion
    This can lead to unexpected results when the default value is changed between function calls.
```

docs/source/working/processes.rst
```diff
@@ -123,9 +123,19 @@ An example input port that explicitly sets all these attributes is the following
...
+    The same concept holds for process ports in AiiDA and so defaults should *only use immutable types*.
```
ltalirz:
```suggestion
    In the context of AiiDA, nodes (both stored and unstored) are considered *mutable* and should therefore *not* be used as default values for process ports.
```

docs/source/working/processes.rst
```diff
@@ -123,9 +123,19 @@ An example input port that explicitly sets all these attributes is the following
...
+    A node, whether unstored or stored, is considered mutable and so should not be used as a default value.
```
ltalirz:
```suggestion
```

chrisjsewell:
FYI this should also fix #2515 and #3143 :)

ltalirz:
Ah, I see @chrisjsewell has me beat by 1 minute ;-)

sphuber:
> FYI this should also fix #2515 and #3143 :)

Indirectly it would yes, but I was thinking of maybe adding a check in the `InputPort` constructor to warn the user when a mutable value is specified for a default. With that PR I would close those issues

aiida/backends/tests/engine/test_ports.py
sphuber: Done

PR #3479, opened 2019-10-29
move yapf/prospector to virtualenv

ltalirz:
This PR moves the yapf/prospector hooks back to a virtual env.
Hopefully, this will reduce issues with installing the hooks & with incompatibilities for different python versions.


ltalirz:
Closing for further changes.

PR #3480, opened 2019-10-29
Replace site.species_and_occu with site.species property for compatible pymatgen versions 

### LINKED ISSUE #3297: Pymatgen interface: species_and_occu deprecated
When creating pymatgen structures, I get the following deprecation warning:

```
  /home/a-dogres/programming/aiida/aiida-core/aiida/orm/nodes/data/structure.py:908: DeprecationWarning: species_and_occu is deprecated
  Use site.species instead. This will be deprecated with effect from pymatgen 2020.
    'weights': [x for x in site.species_and_occu.values()],
```

I would guess this should be a pretty simple replacement. What we should be careful about is when the ``site.species`` interface was introduced, _if_ we want to maintain compatibility with older pymatgen versions.

astamminger:
Fixes #3297 

Added check for the used pymatgen version and dynamically switch to ``Site.species`` property if applicable (i.e. ``pymatgen>=2019.3.13``). Prior to ``2019.3.13`` the old ``Site.species_and_occu`` property is used.


aiida/orm/nodes/data/structure.py
```diff
@@ -896,16 +896,23 @@ def build_kind_name(species_and_occu):
         self.pbc = [True, True, True]
         self.clear_kinds()
 
+        pymatgen_version = tuple(map(int, get_pymatgen_version().split('.')))
         for site in struct.sites:
 
+            # site.species property first introduced in pymatgen version 2019.3.13
+            if pymatgen_version < (2019, 3, 13):
```
greschd: I think here it'd be better to just do plain string comparison - something like ``"2019.3.13b2"`` would also be a legal version string (although pymatgen doesn't seem to use it). That would break the ``int`` conversion above.

Sorry for the nit-pick :grimacing:

greschd:
I think you can ignore the coverall failure, but I'll leave that to @sphuber to confirm.

aiida/orm/nodes/data/structure.py
sphuber: Better still I would say is to use something like:
```
from distutils.version import StrictVersion
current_version = StrictVersion(get_pymatgen_version())
required_version = StrictVersion('2019.3.13')
if current_version < required_version:
    species_and_occu = site.species_and_occu
```

greschd: Nice :+1:

Btw, do we want to bump the ``pymatgen`` version in ``setup.json``? It's currently at ``"pymatgen<=2018.12.12"``. Not sure what the differences are though, tbh.

This issue came up because some dependency changed the ``pymatgen`` version.

sphuber: We can't unfortunately because they dropped python 2 support after that.

greschd: Soon... :sunglasses:

aiida/orm/nodes/data/structure.py
```diff
@@ -18,6 +18,7 @@
 import itertools
 import copy
 from functools import reduce
+from packaging import version
```
sphuber: Thanks for the update @astamminger . One question: is there a reason you prefer the dependency `packaging` over the standard library `distutils`?

astamminger: Was about to write a comment on that but you were too fast :D Mainly the reason was that AFAIK ``StrictVersion`` provided by ``distutils`` does not comply to PEP440 so I tought``packaging`` might be a better alternative (But I'm happy to change that if you prefer to use ``distutils``)

sphuber: Thanks, I didn't know that. It seems indeed that `packaging` seems the more reliable option. I was just worried about it being third-party and therefore adding an implicit dependency, but since it gets installed by `setuptools` that is ok

greschd: From looking at the list of maintainers on PyPI, ``packaging`` seems pretty solid. We should add it as an explicit dependency in ``setup.json`` if we use it explicitly, though.

sphuber: Could you add the dependency to the `atomic_tools` extra @astamminger ? Since it is only used in combination with `pymatgen` I think it is ok to just put it in that extra

greschd: Shouldn't we move the import to the function itself then? Might be easier to just put it in the general dependencies.

astamminger: We could also use ``pkg_resources.parse_version()`` which is distributed with ``setuptools`` so we can avoid adding an additional dependency plus being PEP440 conform. What do you think?

greschd: Go for it 👍 Let's try not to [bike-shed](https://en.wikipedia.org/wiki/Law_of_triviality) too hard here..

I have a feeling the coveralls can be a bit flaky sometimes.

sphuber:
@astamminger if you can move the `pkg_resources` import within the pymatgen function, we do not have to add it to the explicit dependencies. If you can do that final thing I will approve and merge this, thanks!

Issue #3473, opened 2019-10-28
fixtures - 


ltalirz:
aiida-core 1.0 now ships with a number of fixtures for running tests, but there is clearly room for improvement.

We should decide which of the following features should also be included:

fixtures by @sphuber in aiida-qe:

* There is a fixture to test [`CalcJob`](https://github.com/aiidateam/aiida-quantumespresso/blob/89172686a824d0838e0da2c017bf793184aa92ce/tests/conftest.py#L59) classes. Example usage can be found here for [`PwCalculation`](https://github.com/aiidateam/aiida-quantumespresso/blob/develop/tests/calculations/test_pw.py).
* Similar fixture exists to test [`Parser`](https://github.com/aiidateam/aiida-quantumespresso/blob/89172686a824d0838e0da2c017bf793184aa92ce/tests/conftest.py#L86) implementations. It essentially mocks a completed `CalcJobNode` with the `retrieved` folder data node that contains the retrieved files. Example usage for [`PwParser`](https://github.com/aiidateam/aiida-quantumespresso/blob/89172686a824d0838e0da2c017bf793184aa92ce/tests/parsers/test_pw.py#L77) shows.

The automatic comparison files generation and checking is done through the [`pytest-regressions`](https://github.com/aiidateam/aiida-quantumespresso/blob/89172686a824d0838e0da2c017bf793184aa92ce/setup.json#L65) plugin which is added as a dependency. With this pytest plugin and the two fixtures, it is really easy to write and run tests for calculation jobs and parsers without actually having to run the calculations. Note of course, as mentioned before, that some tests that actually run the code will still be useful and desirable.

pointers from @greschd concerning aiida-pytest:

* example for the config file: https://github.com/greschd/aiida-tbextraction/blob/master/tests/config_example.yml
  On the actual machine I would rename this to ‘config.yml’ (which is in .gitignore), and make the necessary changes.
 * Getting the config: https://github.com/greschd/aiida_pytest/blob/migrate_v1/aiida_pytest/_configure.py
 * Setup of computers: https://github.com/greschd/aiida_pytest/blob/migrate_v1/aiida_pytest/_computer.py
 * Setup of codes: https://github.com/greschd/aiida_pytest/blob/migrate_v1/aiida_pytest/_code.py

[...]

PR #3481, opened 2019-10-30
Accept arguments to aiida_profile

espenfl:
An initial solution to how to pass arguments to the recently introduced `aiida_profile`. The problem with this approach is that we on the plugin side relies on fixtures, which typically use the `aiida_profile` fixture as well. Since there is no way, using this approach to pass arguments from one fixture to another (say if we want to set `pg_ctl` in the `pgtest` argument to `aiida_profile`).

We would thus severely limit the usability of fixtures on the plugin side.

### DIFF
--- a/aiida/manage/tests/pytest_fixtures.py
+++ b/aiida/manage/tests/pytest_fixtures.py
@@ -61,7 +61,7 @@ def clear_database(aiida_profile):  # pylint: disable=redefined-outer-name
     """
     yield
     # after the test function has completed, reset the database
-    aiida_profile.reset_db()
+    aiida_profile().reset_db()
 
 
 @pytest.fixture(scope='function')
--- a/aiida/manage/tests/pytest_fixtures.py
+++ b/aiida/manage/tests/pytest_fixtures.py
@@ -23,18 +23,27 @@ from __future__ import print_function
 import tempfile
 import shutil
 import pytest
+from aiida.common.extendeddicts import AttributeDict
 from aiida.manage.tests import test_manager, get_test_backend_name, get_test_profile_name
 
 
 @pytest.fixture(scope='session', autouse=True)
-def aiida_profile():
+def aiida_profile(request):
     """Set up AiiDA test profile for the duration of the tests.
 
     Note: scope='session' limits this fixture to run once per session. Thanks to ``autouse=True``, you don't actually
      need to depend on it explicitly - it will activate as soon as you import it in your ``conftest.py``.
     """
     # create new TestManager instance
-    with test_manager(backend=get_test_backend_name(), profile_name=get_test_profile_name()) as test_mgr:
+    try:
+        test_arguments = request.param
+    except AttributeError:
+        test_arguments = AttributeDict()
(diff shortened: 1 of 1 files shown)


espenfl:
In the second commit a factory function have been used, which makes it possible to pass arguments. However, when passing e.g. `{'pgtest': {'pg_ctl': '/usr/pgsql-9.6/bin/pg_ctl'}` into `test_arguments` of `test_manager` we get the errors reported in #3482. However, that is a separate issue.

Is this something we can run with? I have heard rumors that factory functions obfuscate the test printouts in case there is something amiss, but have not yet seen this myself (or cared to much about it).

ltalirz:
It seems @greschd solved this type of problem by using a configuration file for tests instead.

If you ask me, I would lean towards the configuration file approach (see relevant links [here](https://github.com/aiidateam/aiida-core/issues/3473#issue-513416914)) - adding arguments makes both the definition and the usage of these fixtures rather complicated.

Perhaps @greschd would be willing to come up with a minimal PR that adds support for such a configuration file?

ltalirz:
Anyhow, it's not clear to me what the best solution is - we want to keep these fixtures easy to use but at the same time we don't want to make them too complicated.

Another approach is to say: we try to keep the `test_manager` context manager highly configurable, and simply let developers who have particular needs write their own fixtures using the manager (after all, the fixtures themselves are very little code).

Perhaps @chrisjsewell or @sphuber also have some ideas.

espenfl:
Happy to take a look at your solution @greschd for instance for the `pgtest` parsing.

With respect to complexity I do not think that is a problem as long as we support `aiida_profile()`. Passing whatever arguments down (well, basically right now it is just `pgtest` in addition to `backend` and `profile_name` as far as I can see) does not increase complexity.

Having said that, I am happy with other solutions and this was just a suggestion following the lines we had before and the newly introduced concept. I think in principle I would like to avoid factory functions if we can, but if this is the only solution to offer parameters to `pgtest`, so be it.

greschd:
@espenfl In your use case, would different developers use different ``pg_ctl`` executables? In that case it would probably make sense to have some config file approach. But we would still need to decide at which point in the code to inject the config values - this might work nicely with @ltalirz's proposal to have a configurable base with sugar on top for "casual" use. The top-layer fixtures could use the configuration, with little extra effort required to go around that.

This particular use case seems somewhat different from the things I proposed in #3473 (computer & code setup) in that it determines how the DB itself is set up, whereas the other things happen after that initialization step. That doesn't necessarily mean that we have to treat it in a different way, though.

espenfl:
@greschd Yes and no. One could also envision testing on different Postgresql versions, but let us leave that option for now. Also such a test belongs to `aiida_core` and not the plugins I believe, so fixture nesting would be less of a problem. 

Also, opening up for a config file is certainly maybe the best option. I totally missed that this was considered. Sure, let us just open for a `pgtest` section there and then we can parse whatever `pgtest` eats in the json, following the formatting expected from `pgtest`.

Closing this now.

PR #3486, opened 2019-10-30
add test for `pgtest` argument of TemporaryProfileManager

ltalirz:
this adds a test (running only on travis) for the `pgtest` argument of
the `TemporaryProfileManager`.

Furthermore, the `pgtest` argument is removed from internal functions of
the `TemporaryProfileManager` and treated as an instance variable instead.


ltalirz:
For reviewers: the new test is actually in `test_test_manager.py` (the old content of `test_test_manager.py` has been moved to `test_profile_manager.py`, since that is what it was actually testing).

.ci/test_test_manager.py
```diff
@@ -11,79 +11,41 @@
...
 
-        temp_dir = self.profile_manager.root_dir
-        self.profile_manager.destroy_all()
-        with self.assertRaises(Exception):
-            self.profile_manager.postgres.db_exists(self.profile_manager.dbinfo['db_name'])
-        self.assertFalse(os.path.exists(temp_dir))
-        self.assertIsNone(self.profile_manager.root_dir)
-        self.assertIsNone(self.profile_manager.pg_cluster)
+        # pg_ctl is what PGTest also looks for (although it might be more clever)
+        pgtest = {'pg_ctl': which('pg_ctl')}
```
espenfl: Maybe we should use the `which` from `pgtest`? And/or add another test that tests what we expect with `which` from AiiDA is the same that `pgtest` expects. There is a tight connection between AiiDA and `pgtest` now, but that might change.

ltalirz: > Maybe we should use the which from pgtest? 

Since pgtest is a bit more clever than the standard "which", we could do that.
I'll make the change.

> And/or add another test that tests what we expect with which from AiiDA is the same that pgtest expects.

I don't think that needs to be the case, and it's not what I'm testing here - I'm just trying to see of the `TemporaryProfileManager` works if I give it *some* valid path to `pg_ctl` (could be a different one than the one autodiscovered by pgtest).

ltalirz: My main question for you was to check whether you are ok with me removing the `pgtest` argument from the internal functions of the `TemporaryProfileManager` (since I guess you were the main user) - I guess as long as it is still accessible, it is fine with you, right?

espenfl: Yes, absolutely. I was using it from the `test_manager`, which run the constructor I think, so for me this would work nicely.

ltalirz:
This is ready for review

ltalirz:
@espenfl Please have another look through this when you find the time. Cheers!

PR #3497, opened 2019-10-31
Backport database schema generation check

### LINKED ISSUE #3496: Backport schema generation check for release `1.0.0`

sphuber:
Fixes #3496 

This check will be implemented in the code after the schema has been
reset, but to make sure that people with a database schema generation
that is newer than the current one, cannot trigger the database, we
backport the check here. Note that this will only affect the release of
`v1.0.0 which will include this commit. Older versions will just attempt
the migration, which will fail and except. So at least it should not
touch and corrupt the database.

### DIFF
--- a/aiida/backends/utils.py
+++ b/aiida/backends/utils.py
@@ -21,6 +21,9 @@ from aiida.manage import configuration
 
 AIIDA_ATTRIBUTE_SEP = '.'
 
+SCHEMA_GENERATION_KEY = 'schema_generation'
+SCHEMA_GENERATION_VALUE = '1'
+
 
 Setting = collections.namedtuple('Setting', ['key', 'value', 'description', 'time'])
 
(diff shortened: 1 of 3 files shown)


aiida/backends/utils.py
```diff
@@ -68,6 +71,22 @@ def get_settings_manager():
...
+        raise ConfigurationError(
+            'The schema generation of your database is newer than that of the code and is therefore incompatible.'
+        )
+
+
+def get_db_schema_generation():
+    """Get the schema generation of the current database."""
+    from aiida.backends.utils import get_settings_manager
+    manager = get_settings_manager()
+    return manager.get(SCHEMA_GENERATION_KEY)
```
giovannipizzi: Thinking better about this, I would do `manager.get(SCHEMA_GENERATION_KEY, 1)` (does this `.get` accept default values?) because otherwise this will always fail for anything before a few commits ago, when the key was added. Do you agree?

sphuber: It does support a default, but I explicitly didn't. The reason is that "a few commits before" this code doesn't exist, so it cannot except

sphuber: In fact the `get` does not have a default, but I just catch the exception and use the default

aiida/backends/utils.py
```diff
@@ -68,6 +71,29 @@ def get_settings_manager():
     return manager
 
 
+def validate_schema_generation():
+    """Verify that the database schema generation is compatible with the current code schema generation."""
+    from aiida.common.exceptions import ConfigurationError, NotExistent
+    try:
+        schema_generation_database = get_db_schema_generation().value
+    except NotExistent:
+        schema_generation_database = SCHEMA_GENERATION_VALUE
```
giovannipizzi: I would have hardcoded `1` here - if in the future we reuse the code (I know we shouldn't but maybe someone tried to merge) and then up the SCHEMA_GENERATION_VALUE, still here the value should remain 1

PR #3500, opened 2019-10-31
reorganize backup docs

ltalirz:
Backup docs now start with an overview of the data to back up;
then go to the file repository, and finally to the database.

Added a `verdi devel configure-backup` command so that people don't have to manually find the script in the aiida-core folder.

Happy to discuss the name/placement of the command.


ltalirz:
Also ready for review - perhaps @CasperWA might want to have a look?

docs/source/backup/index.rst
```diff
@@ -1,272 +1,255 @@
...
-  Remember also to give it the right permissions (read and write): ``chmod u=rw .pgpass``.
+        .. code-block:: shell
 
-To dump the database in a file automatically everyday, you can add the following script :download:`backup-aiidadb-USERNAME <backup-aiidadb-USERNAME>` in the folder ``/etc/cron.daily/``, which will launch the previous script once per day:
+            sshfs -o idmap=user -o rw backup_user@server_2:/home/backup_user/backup_destination_dir/ /home/aiida_user/remote_backup_dir/
 
-.. literalinclude:: backup-aiidadb-USERNAME
-   :language: shell
+        Use ``gnome-session-properties`` in the terminal to add this line to the actions performed at start-up.
+        Do **not** add it to your shell's startup file (e.g. ``.bashrc``) or your computer will complain that the mount point is not empty whenever you open a new terminal...
```
CasperWA: I guess the `...` can be left out?
```suggestion
        Do **not** add it to your shell's startup file (e.g. ``.bashrc``) or your computer will complain that the mount point is not empty whenever you open a new terminal.
```

docs/source/backup/index.rst
```diff
@@ -1,272 +1,255 @@
...
-Where all instances of ``USERNAME`` must replaced by your actual user name.
-The ``su USERNAME`` makes the dumped file be owned by you rather than by ``root``.
-Remember to give the script the right permissions::
 
-  sudo chmod +x /etc/cron.daily/backup-aiidadb-USERNAME
+ 2. Run the ``start_backup.py`` script in the "backup folder" to start the backup.
 
-Finally make sure your database folder (``/home/USERNAME/.aiida/``) containing this dump file and the ``repository`` directory, is properly backed up by your backup software (under Ubuntu, 12.04: Backup -> check the "Folders" tab, 18.04: Backups -> check the "Folder to save" tab).
+    This will back up all data added after the ``oldest_object_backedup`` date.
+    It will only carry out a new backup every ``periodicity`` days, until a certain end date if specified (using ``end_date_of_backup`` or ``days_to_backup``).
```
CasperWA: Since these terms have not yet been introduced, the place in the docs for their definitions should probably be mentioned, e.g.:
```suggestion
    It will only carry out a new backup every ``periodicity`` days, until a certain end date if specified (using ``end_date_of_backup`` or ``days_to_backup``), see :ref:`backup_configuration_options` below.
```
Also adding the tag `_backup_configuration_options` below, prior to the section "Configuration options".

CasperWA: This goes for the previous line as well, although the reference to configuration options should probably only be done once.

docs/source/backup/index.rst
```diff
@@ -1,272 +1,255 @@
...
 
-where ``SOURCE_DIRECTORY`` is the directory you got from the ``SHOW data_directory;`` command, and ``DESTINATION_DIRECTORY`` is the new directory for the database files.
+Add the following script :download:`backup-aiidadb-USERNAME <backup-aiidadb-USERNAME>` to the folder ``/etc/cron.daily/``:
 
-.. note:: The behaviour of the ``cp -a`` command is to create a directory within ``DESTINATION_DIRECTORY``, e.g.::
+.. literalinclude:: backup-aiidadb-USERNAME
+   :language: shell
 
-    cp -a OLD_DIR/main/ NEW_DIR/
+and replace all instances of ``USERNAME`` by your UNIX user name.
```
CasperWA:
```suggestion
and replace all instances of ``USERNAME`` with your UNIX user name.
```

docs/source/backup/index.rst
```diff
@@ -1,272 +1,255 @@
...
-    sudo crontab -u USERNAME -e
 
-  It will open an editor, where you can add a line of the form::
+    After supplying your database password, the database should be restored.
 
-    00 03 * * * /home/USERNAME/.aiida/backup/start_backup.py 2>&1 | mail -s "Incremental backup of the repository" USER_EMAIL@domain.net
+    .. note::
 
-  or (if you need to backup a different profile than the default one)::
+        On Ubuntu, type ``sudo su - postgres`` to become the postgres UNIX user.
```
CasperWA: This should probably be shifted upwards to come before the first time one needs to do this.

docs/source/backup/index.rst
ltalirz: I think both versions are equivalent, see [english SE](https://english.stackexchange.com/a/151352).

docs/source/backup/index.rst
ltalirz: Ah yes - that would be in the section on "moving the database".

In my opinion, the "moving the database" section should either be moved out to some "recipe collection section" or removed entirely:
 * it is not really about backup
 * it contradicts what we say in the backup section (that you should not use the postgres files themselves)
 * it is only about postgres, no AiiDA specifics (`moving postgres data directory` gives lots of results on Google)
 * the same effect could usually be achieved with a proper backup + restore

I guess it can save you a bit of time if you indeed need to move the database, so we could move it to "Troubleshooting" under "Working with AiiDA" and link to it. Or just get rid of it.

@giovannipizzi What do you say?

docs/source/backup/index.rst
```diff
@@ -1,272 +1,255 @@
 .. _backup:
 
-On this page you will find useful information on how to backup your database, how to move it to a different location, and how to backup your repository.
+AiiDA stores information in:
 
+ * the ``.aiida`` folder, typically located at ``~/.aiida`` (see also :ref:`configure_aiida`).
+   This contains the configuration of the profiles as well as their associated file repositories.
```
sphuber: Just wanted to point out that technically the file repositories do not have to live within `.aiida`. I think it is probably beter to make the repositories a separate entry in this itemization and then maybe put in parentheses that by default they often are located in the `.aiida` folder.

docs/source/backup/index.rst
CasperWA: > Also adding the tag `_backup_configuration_options` below, prior to the section "Configuration options".

Did you miss this ^^^ ? :)

docs/source/backup/index.rst
giovannipizzi: I think we should keep the section on moving the database, but where it's now probably is not useful to users.

The goal, to me, would be to create a section on "how to configure your computer for a performant use of AiiDA" and we give there all the tricks, including things like:
- put the DB in a fast disk, ideally on a big partition (and we link to the section on 'moving the db' to say that, if you already have the db, you can move it to a different partition following those instructions)
- perform incremental backups (and link to section)
- increase the number of workers
- anything else that we do in our production DBs but it's not really documented, or documented in a scattered way (e.g. remove Ubuntu's indexing of the folders etc.)

Feel free to open an issue with my message above if you don't want to address it here.


@ltalirz Note that I'm not sure that the same effect can be obtained with dump+restore - it's not just about moving a single DB, but the whole folder where Postgres stores *all* DBs.

sphuber:
> Concerning the `verdi` command, I suggest creating `verdi backup` wherein this can live, and possibly more to come, if needed?

I would rather go for @ltalirz suggestion and keep it under `devel` for now. I have the feeling the backup functionality is still in early stages of usability and so might change quite a bit with future improvements. Once we release it under `verdi backup` it will become more difficult to adapt it

docs/source/backup/index.rst
ltalirz: thanks, yes I overlooked that this still needs to be added

docs/source/backup/index.rst
ltalirz: adapted

docs/source/backup/index.rst
ltalirz: The "performant AiiDA usage" section sounds like a good idea to me.
It is actually already pretty much exists where I suggested to move things (Working with AiiDA => Troubleshooting).

I've added it now.
@sphuber : Would you like to add some more wisdom concerning the number of workers?
E.g. how to roughly estimate it and what number not to exceed (#cores? 4x#cores?)

ltalirz: I've also added a "tips&tricks" section, starting from the one that was already there.

ltalirz:
@sphuber by the way. this is ready for review since quite some time

PR #3504, opened 2019-11-01
Reset the database schema generation

### LINKED ISSUE #3011: Reset migrations after release of `v1.0.0`
Currently there are 37 migrations for the Django backend and a bit less for SqlAlchemy. Aside from adding time to each time a database is created (not so much a real problem for new profiles, but time consuming for testing with the fixture manager), it adds a lot of complexity and overhead on for example testing and maintenance. It would be good if at some point we can "reset" the initial schema. Ideally this is done directly after the release of `v1.0.0`.

sphuber:
Fixes #3011


sphuber:
Most likely we will not go through with this, so closing for now

PR #3506, opened 2019-11-01
minor update of export docs

ltalirz:
* suggest putting nodes into a group more explicitly (some people struggle with this)
 * move info on cli options up so that people are more likely to see it
 * minor rewordings


docs/source/import_export/main.rst
```diff
@@ -9,30 +9,29 @@ Export
...
-  By default, comments and logs are also added for each exported node.
-  These may be excluded using the proper flags.
-  There are several other options to export specific parts of the provenance graph, starting from the selection.
-  For more information about this please read the :ref:`corresponding subsection<consistency>` of the :ref:`Provenance section<concepts_provenance>`.
+See ``verdi export create -h`` for a full list of available options.
+
+* **Selection:** Nodes can be selected via identifier (PK, UUID, label).
+  It is often convenient to first put the nodes you would like to export into a :py:class:`~aiida.orm.Group` and then to export the group using ``verdi export create -G <group-name>``.
+* **Augmentation:** By default, the export function automatically augments your selection of nodes by their logical and data provenance (and calculations by their direct outputs).
+  You can customize this behavior using command line options.
```
ramirezfranciscof: If the default behavior is still to export comments and logs and you can disable this, then I would maybe keep the line clarifying this (maybe not in the **augmentation** section but in the **file content** section below? unless you think it is obvious that this "metadata" would be included) because the data consistency section only discusses what nodes are included or not, and doesn't deal with what kind of information gets exported for each node.

docs/source/import_export/main.rst
```diff
@@ -9,30 +9,29 @@ Export
 
 Use ``verdi export create`` to export a selection of nodes, codes, groups, users, or computers.
```
CasperWA: I guess, for completeness, this could be and/or as well?
```suggestion
Use ``verdi export create`` to export a selection of nodes, codes, groups, users, and/or computers.
```

docs/source/import_export/main.rst
```diff
@@ -9,30 +9,29 @@ Export
...
-* **Augmentation:** By default, the export function augments the selection of nodes by their parents in order to preserve the provenance.
-  For calculation nodes, the direct outputs are also added.
-  By default, comments and logs are also added for each exported node.
-  These may be excluded using the proper flags.
-  There are several other options to export specific parts of the provenance graph, starting from the selection.
-  For more information about this please read the :ref:`corresponding subsection<consistency>` of the :ref:`Provenance section<concepts_provenance>`.
+See ``verdi export create -h`` for a full list of available options.
+
+* **Selection:** Nodes can be selected via identifier (PK, UUID, label).
+  It is often convenient to first put the nodes you would like to export into a :py:class:`~aiida.orm.Group` and then to export the group using ``verdi export create -G <group-name>``.
```
CasperWA: I am not sure this will work as a link to the Group class? Unless it has been fixed, the link path should be absolute, i.e.:
```suggestion
  It is often convenient to first put the nodes you would like to export into a :py:class:`~aiida.orm.groups.Group` and then to export the group using ``verdi export create -G <group-name>``.
```

docs/source/import_export/main.rst
```diff
@@ -9,30 +9,29 @@ Export
...
+* **Augmentation:** By default, the export function automatically augments your selection of nodes by their logical and data provenance (and calculations by their direct outputs).
```
CasperWA: Maybe just provenance, instead of listing both kinds? Although, if you want both kinds provenance should be in plural, right? In any case, my suggestion is to just write provenance:
```suggestion
* **Augmentation:** By default, the export function automatically augments your selection of nodes by their provenance (and calculations by their direct outputs).
```

docs/source/import_export/main.rst
CasperWA: I agree, this should be left in as a comment, since it may result in a lot of extra exported data if one has a lot of Logs (or really use Comments a lot - in which case one would probably _want_ to export them... :sweat_smile: )

ltalirz: My thinking was that users should expect this to be included, and that they will see the cli option when they run the `-h` command:
```
 --include-logs / --exclude-logs
                                  Include or exclude logs for node(s) in
                                  export.  [default: True]
  --include-comments / --exclude-comments
                                  Include or exclude comments for node(s) in
                                  export. (Will also export extra users who
                                  commented).  [default: True]
```

docs/source/import_export/main.rst
```diff
@@ -9,30 +9,29 @@ Export
...
+  You can customize this behavior using command line options.
+  For details on how the provenance graph is traversed, see the section on :ref:`data consistency<consistency>`.
 * **File content:** The export file contains all information pertaining to the exported nodes: both the information stored in the database and files stored in the AiiDA repository.
-* **Compression:** By default, the export file is compressed using gzip, creating a tar ball (tar.gz).
-  It is also possible to create a compressed or uncompressed zip file.
+* **Compression:** By default, the export file is compressed using ``gzip``, creating a tar ball (``tar.gz``).
```
CasperWA: This is not true. By default the export file is a compressed ``zip`` file, so the content of these lines should be switched, leaving the uncompressed ``zip`` file in its place :)

docs/source/import_export/main.rst
ltalirz: We might want to mention if certain data is *not* included by default - to my knowledge it's only the `Authinfo` of the computers (?).

docs/source/import_export/main.rst
```diff
@@ -9,30 +9,29 @@ Export
...
+  Use the command line options to compress using ``zip`` instead or create an uncompressed ``zip`` archive.
 
-See ``verdi export create -h`` for a full list of available options.
 
 Import
 ++++++
 
-Use ``verdi import`` to import an AiiDA export archive generated by ``verdi export create``, either a local file or an URL.
+Use ``verdi import`` to import an AiiDA export archive generated by ``verdi export create``, either from a local file or from a URL.
```
CasperWA:
```suggestion
Use ``verdi import`` to import an AiiDA export archive generated by ``verdi export create``, either passing a local file or a URL.
```

CasperWA: It may or may not be prudent to add the notion of lists of URLs or local file paths as well (which is supported).

docs/source/import_export/main.rst
ltalirz: I just checked and it works :) I would also consider it a sphinx bug, if it did not issue a warning if I provided a path it does not understand.

docs/source/import_export/main.rst
ltalirz: thanks, will adapt!

docs/source/import_export/main.rst
ltalirz: will switch to plural.

docs/source/import_export/main.rst
CasperWA: Since people shouldn't directly interact with `Authinfo` anyway (I believe?), I don't see that it needs to be written here - the more technical users can in any case refer to the `metadata.json` example to see what is and what isn't exported?

docs/source/import_export/main.rst
```diff
@@ -7,32 +7,31 @@ AiiDA offers tools for exporting selected parts of an AiiDA graph to a file for
...
+* **File content:** The export file contains all information pertaining to the exported nodes: both the information stored in the database and files stored in the AiiDA repository.
+* **Compression:** By default, the export file is a compressed ``zip`` archive.
+  Use the command line options to disable compression or to switch to a different format (tar ball).
+
+
 Import
 ++++++
 
-Use ``verdi import`` to import an AiiDA export archive generated by ``verdi export create``, either a local file or an URL.
+Use ``verdi import`` to import AiiDA export archives generated by ``verdi export create``, either passing local paths or a URLs.
```
CasperWA:
```suggestion
Use ``verdi import`` to import AiiDA export archives generated by ``verdi export create``, either passing local paths or URLs.
```

docs/source/import_export/main.rst
```diff
@@ -7,32 +7,31 @@ AiiDA offers tools for exporting selected parts of an AiiDA graph to a file for
...
 See ``verdi export create -h`` for a full list of available options.
 
+* **Selection:** Nodes can be selected via identifier (PK, UUID, label).
+  It is often convenient to first put the nodes you would like to export into a :py:class:`~aiida.orm.Group` and then to export the group using ``verdi export create -G <group-name>``.
+* **Augmentation:** By default, the export function automatically augments your selection of nodes by their provenance (and calculations by their direct outputs).
+  You can customize this behavior using command line options.
+  For details on how the provenance graph is traversed, see the section on :ref:`data consistency<consistency>`.
+* **File content:** The export file contains all information pertaining to the exported nodes: both the information stored in the database and files stored in the AiiDA repository.
+* **Compression:** By default, the export file is a compressed ``zip`` archive.
+  Use the command line options to disable compression or to switch to a different format (tar ball).
```
CasperWA:
```suggestion
  Use the command line options to disable compression or to switch to a different format (gzipped tar ball or uncompressed zip).
```

docs/source/import_export/main.rst
ltalirz: I agree.

docs/source/import_export/main.rst
```diff
@@ -7,32 +7,31 @@ AiiDA offers tools for exporting selected parts of an AiiDA graph to a file for
...
+* **File content:** The export file contains all information pertaining to the exported nodes: both the information stored in the database and files stored in the AiiDA repository.
+* **Compression:** By default, the export file is a compressed ``zip`` archive.
+  Use the command line options to disable compression or to switch to a different format (gzipped tar ball or uncompressed zip).
+
+
 Import
 ++++++
 
-Use ``verdi import`` to import an AiiDA export archive generated by ``verdi export create``, either a local file or an URL.
+Use ``verdi import`` to import AiiDA export archives generated by ``verdi export create``, either passing local paths or URLs.
```
CasperWA:
```suggestion
Use ``verdi import`` to import AiiDA export archives generated by ``verdi export create``, either passing local paths and/or URLs.
```

ltalirz:
Done ;-)

PR #3509, opened 2019-11-04
Update the installation note for AiiDA.

sponce24:
Add possible missing postgres dep.

### DIFF
--- a/docs/source/install/installation.rst
+++ b/docs/source/install/installation.rst
@@ -115,6 +115,12 @@ which leads through the installation process and takes care of creating the corr
 For maximum control and customizability, one can use ``verdi setup``
 and set up the database manually as explained below.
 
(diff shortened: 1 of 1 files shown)


sphuber:
Thanks for the heads-up, but I am not sure if this is the actual reason for seeing that message. I have seen it before and definitely had postgres up and running. It has more to do with how our code for `quicksetup` tries to determine the way it can connect, with regards to super user privileges. Also `postgresql` is mentioned explicitly as [a dependency in the documentation](https://aiida-core.readthedocs.io/en/latest/install/prerequisites.html) before the installation part. Finally, notes like this might be better suited in the [troubleshooting section(https://aiida-core.readthedocs.io/en/latest/install/troubleshooting.html#installation-phase).

PR #3513, opened 2019-11-05
Fix bug in `upload_calculation` when running with `store_provenance=False`

### LINKED ISSUE #3511: passing unstored data node surfaces programming error in aiida-core

sphuber:
Fixes #3511 

When running a `CalcJob` in dry-run mode, it is often advised to also set
`store_provenance=False` which means nodes will not be stored, so input
nodes can be unstored. However, in `upload_calculation` to process the
`local_copy_list` the `load_node` function was used to load the
corresponding node, which would fail in this scenario. The only possible
solution is to pass the full mapping of inputs to the function and if
the loading of the node fails, try to fish out the corresponding node
from the mapping. If all fails, we emit a warning and skip the rule.


aiida/engine/daemon/execmanager.py
```diff
@@ -162,21 +162,44 @@ def upload_calculation(node, transport, calc_info, script_filename, dry_run=Fals
...
+                if data_node is not None:
+                    break
+
+            return data_node
 
         try:
             data_node = load_node(uuid=uuid)
         except exceptions.NotExistent:
-            logger.warning('failed to load Node<{}> specified in the `local_copy_list`'.format(uuid))
+            data_node = find_data_node(inputs, uuid)
```
ltalirz: do we want to allow this in general or only when `dry_run=True`?

Also, maybe good to add a comment line to both 188 and 190

sphuber: Yeah, I was wondering about that too, but I just kept the current behavior for now where it just emits a warning log. We might want to change this to actually raise. If the plugin specified that a the contents of a node need to be copied over and it is not there, that should be a problem and most likely the calculation should not work. I am not sure in what use case a mere warning is the desired alternative.

PR #3516, opened 2019-11-07
Fix namespaces hard-coded mapping

### LINKED ISSUE #3515: REST API: incorrect node_type in full_type hard-coded mapping
In the hard-coded mapping to generate the `full_type` for process nodes namespaces of external plugins, the `node_type` is incorrect since there is a dot missing.
As a consequence, the parsed `node_type` filters passed to the QueryBuilder for these namespaces are not the right ones.

asle85:
The `node_type` in the hard-coded mapping to generate the correct full types for process node namespaces of external plugins had a final dot missing.

Fixes #3515


sphuber:
Thanks a lot @asle85

PR #3517, opened 2019-11-07
add `filename` argument to `SinglefileData` constructor

### LINKED ISSUE #3520: support setting filename of SinglefileData
While it is now possible to create a `SinglefileData` directly from a string, it is then not possible to define the filename (will take the default).

This should be supported (can even be useful to store existing files under a different name).

ltalirz:
fix #3520 

 * The documentation of how to create a `SinglefileData` from a string was incorrect, and missing from the constructor (first commit).
 * While it is possible to create a SinglefileData directly from a string, it was not possible to define the filename. Fixed in second commit.
 * there was no way to get the content of a Singlefiledata on the cli. fixed in 3rd commit.

### DIFF
--- a/aiida/cmdline/commands/cmd_data/cmd_singlefile.py
+++ b/aiida/cmdline/commands/cmd_data/cmd_singlefile.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""`verdi data singlefile` command."""
+from __future__ import division
+from __future__ import print_function
+from __future__ import absolute_import
+
+from aiida.cmdline.commands.cmd_data import verdi_data
+from aiida.cmdline.params import arguments, types
+from aiida.cmdline.utils import decorators, echo
+
+
+@verdi_data.group('singlefile')
+def singlefile():
+    """Work with SinglefileData nodes."""
+
+
+@singlefile.command('content')
+@arguments.DATA(type=types.DataParamType(sub_classes=('aiida.data:singlefile',)))
+@decorators.with_dbenv()
+def singlefile_content(data):
+    """Show the content of the file."""
+    for node in data:
+        try:
+            echo.echo(node.get_content())
+        except IOError as exception:
+            echo.echo_warning('could not read the content for SinglefileData<{}>: {}'.format(node.pk, str(exception)))
--- a/aiida/cmdline/commands/cmd_data/cmd_singlefile.py
+++ b/aiida/cmdline/commands/cmd_data/cmd_singlefile.py
@@ -23,12 +23,11 @@ def singlefile():
 
 
 @singlefile.command('content')
-@arguments.DATA(type=types.DataParamType(sub_classes=('aiida.data:singlefile',)))
+@arguments.DATUM(type=types.DataParamType(sub_classes=('aiida.data:singlefile',)))
 @decorators.with_dbenv()
-def singlefile_content(data):
+def singlefile_content(datum):
     """Show the content of the file."""
-    for node in data:
-        try:
-            echo.echo(node.get_content())
-        except IOError as exception:
-            echo.echo_warning('could not read the content for SinglefileData<{}>: {}'.format(node.pk, str(exception)))
+    try:
+        echo.echo(datum.get_content())
+    except (IOError, OSError) as exception:
+        echo.echo_critical('could not read the content for SinglefileData<{}>: {}'.format(datum.pk, str(exception)))
--- a/aiida/orm/nodes/data/data.py
+++ b/aiida/orm/nodes/data/data.py
@@ -25,20 +25,15 @@ __all__ = ('Data',)
 
 class Data(Node):
     """
-    This class is base class for all data objects.
+    The base class for all Data nodes.
 
-    Specifications of the Data class:
-    AiiDA Data objects are subclasses of Node and should have
-
-    Multiple inheritance must be supported, i.e. Data should have methods for
-    querying and be able to inherit other library objects such as ASE for
+    AiiDA Data classes are subclasses of Node and must support multiple inheritance,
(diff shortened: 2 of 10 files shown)


aiida/cmdline/commands/cmd_data/cmd_singlefile.py
```diff
@@ -0,0 +1,34 @@
...
+from aiida.cmdline.utils import decorators, echo
+
+
+@verdi_data.group('singlefile')
+def singlefile():
+    """Work with SinglefileData nodes."""
+
+
+@singlefile.command('content')
+@arguments.DATA(type=types.DataParamType(sub_classes=('aiida.data:singlefile',)))
```
sphuber: Do we really want to support passing multiple nodes here?
How useful is it to have all content concatenated. There won't be an easy way to distinguish where one ends and the next starts, correct?

aiida/cmdline/commands/cmd_data/cmd_singlefile.py
```diff
@@ -0,0 +1,34 @@
...
+@decorators.with_dbenv()
+def singlefile_content(data):
+    """Show the content of the file."""
+    for node in data:
+        try:
+            echo.echo(node.get_content())
+        except IOError as exception:
```
sphuber: For the current repository implementation we should also catch `OSError`

aiida/cmdline/commands/cmd_data/cmd_singlefile.py
```diff
@@ -0,0 +1,34 @@
...
+            echo.echo_warning('could not read the content for SinglefileData<{}>: {}'.format(node.pk, str(exception)))
```
sphuber: if we change interface to only expect single node, this should become `echo_critical`

aiida/orm/nodes/data/data.py
```diff
@@ -25,20 +25,15 @@
...
     """
-    This class is base class for all data objects.
+    The base class for all Data nodes.
 
-    Specifications of the Data class:
-    AiiDA Data objects are subclasses of Node and should have
-
-    Multiple inheritance must be supported, i.e. Data should have methods for
-    querying and be able to inherit other library objects such as ASE for
+    AiiDA Data classes are subclasses of Node and must support multiple inheritance,
```
sphuber: I realize that you just rephrased the old sentence, but to me it does not make a lot sense still. I don't see how "multiple inheritance" has anything to do with being queryable or support construction of other library definitions. I think definitely the latter part should not be here. It should be up to the plugin developer to decide what its interface supports.

aiida/orm/nodes/data/upf.py
```diff
@@ -269,6 +269,11 @@ class UpfData(SinglefileData):
     """`Data` sub class to represent a pseudopotential single file in UPF format."""
 
     def __init__(self, file=None, source=None, **kwargs):
```
sphuber: Shouldn't we also then expose `filename` on the constructor of this sub class? Same goes for `CifData`.

aiida/orm/nodes/data/singlefile.py
```diff
@@ -26,15 +26,21 @@ class SinglefileData(Data):
...
+        :param file: an absolute filepath or filelike object whose contents to copy.
+            Hint: Pass io.BytesIO(b"my string") to construct the SinglefileData directly from a string.
+        :param filename: specify filename to use (defaults to name of provided file).
         """
         # pylint: disable=redefined-builtin
         super(SinglefileData, self).__init__(**kwargs)
         if file is not None:
-            self.set_file(file)
+            if filename is None:
+                # for backwards compatibility don't assume that set_file has a 'filename' argument
```
sphuber: I don't think this accomplishes what you intend it to do. What if the sub class does _not_ override the constructor and a user passes `filename`, it will then go to line 43 and except on `set_file` if the overridden version does not add `filename`. Doing the switch on the value to guard against the signature seems fragile. Shouldn't we just call with the filename argument and catch `TypeError`. Even that is non-ideal, but the only remaining option would be to inspect the signature of the method which is also fragile I think

ltalirz: >  What if the sub class does not override the constructor and a user passes filename

I would consider it ok if an exception is raised in this case - since the user is using the class in a way it is not designed by the author.
What is not ok is to assume that `set_file` supports the `filename` argument as there may be many subclasses of `SinglefileData` out there that don't.

> Shouldn't we just call with the filename argument and catch TypeError. Even that is non-ideal, but the only remaining option would be to inspect the signature of the method which is also fragile I think

Hm... I would prefer the guard based on the argument over catching any `TypeError` thrown by the function call.

We could do inspection of the function arguments (`if 'filename' not in inspect.getargspec(self.f)[0]`) but what do you if we see that `filename` is missing?
If it is missing, then perhaps the plugin author should update it - we could raise a warning.
But we would still need to check the user-provided value - if it is not `None`, we cannot just silently ignore it.

aiida/orm/nodes/data/data.py
ltalirz: hehe, similar questions here - I just started to modify this in order to get some feedback on this.

So, I propose to keep only `AiiDA Data classes are subclasses of Node and must support multiple inheritance.`

aiida/cmdline/commands/cmd_data/cmd_singlefile.py
ltalirz: thanks for catching this - didn't realize there was `DATUM`

aiida/orm/nodes/data/upf.py
ltalirz: oops - forgotten.
I got sidetracked by this `source` variable that I had never encountered before.

sphuber: also add it for `CifData` please

aiida/cmdline/commands/cmd_data/cmd_singlefile.py
sphuber: You can now also get rid of the for loop

aiida/orm/nodes/data/singlefile.py
sphuber: Ok fine to keep it. Is there a path that we can remove this clause in a future version though for simplification? I realize a "deprecation" warning does not really work here. So how would we go about warning people that the new `set_file` needs a new argument?

ltalirz: hm, I think we can add the deprecation warning (we just need to keep the logic based on the value of the filename argument as well).
done in the last commit.

sphuber:
Thanks a lot @ltalirz

PR #3524, opened 2019-11-07
add `queryhelp` property to QueryBuilder

### LINKED ISSUE #3522: how to copy a querybuilder instance?
It would be very useful to be able to copy a querybuilder instance, for example in cases where you want to build up multiple queries that start the same, but then take different paths.

Currently it seems impossible to make a copy of the QueryBuilder
 * `copy.copy` is not enough
 * `copy.deepcopy` somehow ends in an "infinite recursion" issue
 * while you can construct a QueryBuilder using a `queryhelp` dictionary, I don't see a way to get such a dictionary from an existing QueryBuilder(?)

ltalirz:
fix #3522 

 * add `queryhelp` property to QueryBuilder
   (currently pipes through to get_json_compatible_queryhelp)
 * add __deepcopy__ function to allow direct copies of QueryBuilder
   instances
 * test that QueryBuilder can be re-created from queryhelp
 * remove `test_query.py` and merge with `orm/test_querybuilder.py`


ltalirz:
The test I added is here: https://github.com/aiidateam/aiida-core/pull/3524/files#diff-b9a67a0878b5bdee86b7a596950bc95fR678

aiida/orm/querybuilder.py
```diff
@@ -1766,6 +1766,23 @@ def get_json_compatible_queryhelp(self):
...
+        The queryhelp can be used to create a copy of the QueryBuilder instance like so::
+
+            qb = QueryBuilder(limit=3).append(StructureData, project='id').order_by({StructureData:'id'})
+            qb2=QueryBuilder(**qb.queryhelp)
+
+        :return: a queryhelp dictionary
+        """
+        return self.get_json_compatible_queryhelp()
+
+    def __deepcopy__(self, memodict={}):
```
sphuber: signature should be `def __deepcopy__(self, memo)`

aiida/orm/querybuilder.py
```diff
@@ -1766,6 +1766,23 @@ def get_json_compatible_queryhelp(self):
...
+        qb = type(self)(**self.queryhelp)
```
sphuber: no need for intermediate variable assignment

aiida/orm/querybuilder.py
sphuber: Why only `__deepcopy__` implementation. Wouldn't a normal copy also not work? Since the queryhelp won't contain any references to other objects a shallow copy would also just work when implemented like this no?

sphuber:
Thanks @ltalirz just some minor comments

aiida/orm/querybuilder.py
ltalirz: fixed!

aiida/orm/querybuilder.py
ltalirz: I had it in here as a bit of a help that what is being returned here is actually a querybuilder.
debatable though whether this is helpful - I'll get rid of it

aiida/orm/querybuilder.py
ltalirz: Well, I chose this because the implementation of the queryhelp uses `deepcopy`.
https://github.com/aiidateam/aiida-core/blob/5a1f18b67599ce25766cffea9a00c7e420141f6d/aiida/orm/querybuilder.py#L1760-L1767

And the deepcopy is actually necessary since it contains nested dictionaries:
```
In [1]: qb=QueryBuilder()

In [2]: qb.append(Data)
Out[2]: <aiida.orm.querybuilder.QueryBuilder at 0x11ea02208>

In [3]: qb.queryhelp
Out[3]:
{'path': [{'entity_type': 'data.Data.',
   'tag': 'Data_1',
   'joining_keyword': None,
   'joining_value': None,
   'outerjoin': False,
   'edge_tag': None}],
 'filters': {'Data_1': {'node_type': {'like': 'data.%'}}},
 'project': {'Data_1': []},
 'order_by': {},
 'limit': None,
 'offset': None}
```

sphuber:
Thanks leo

PR #3526, opened 2019-11-08
Remove unused QB methods/functions

### LINKED ISSUE #3518: Possibly unused QueryBuilder code
Poking around in the front-end QueryBuilder code, it seems a function is never used.
I found this since it also calls the now obsolete attribute `self._imp` in it.

The function is `except_if_input_to` and can be found [here](https://github.com/aiidateam/aiida-core/blob/develop/aiida/orm/querybuilder.py#L1924).
A search through the whole repository for the function name only reveals a single hit - the definition of it.

@lekah is this function a leftover from a previous rewrite of the class, or the beginning of an addition to the QueryBuilder? And are there more of these dead-code-methods around that either need a purpose in a grander scheme or should simply be sent to oblivion?

CasperWA:
Closes #3518.

Also deals with the method mentioned in #3522.

Feel free to find and remove more QueryBuilder code that is _completely_ unused in the code base of AiiDA, if you believe it should not be updated and used.
(Please post here if you want/need writing permissions to my fork).


ltalirz:
@lekah  What about `inputs`, `outputs`, `children`, `parents`?

lekah:
> @lekah What about `inputs`, `outputs`, `children`, `parents`?

I will create a new issue for them. They contain(ed) functionality that was often discussed as important, namely to do graph traversals using the querybuilder, as in node.qb.input().outputs() -> get all outputs of all inputs of a certain node. It looks like some of this functionality was removed, but not the functions within the QueryBuilder...

lekah:
Ok for me as soon as @ltalirz approves

aiida/orm/querybuilder.py
```diff
@@ -1710,32 +1710,6 @@ def _get_connecting_node(self, index, joining_keyword=None, joining_value=None,
                                                    self.tag_to_alias_map.keys()))
         return returnval
 
-    def _get_json_compatible(self, inp):
```
ltalirz: Since we are now removing this function, could you please:

* Move the implementation from `get_json_compatible_queryhelp` into the `queryhelp` property
* Use `self.queryhelp` in `get_json_compatible_queryhelp` and add a deprecation warning (will be removed in AiiDA 2.0)?

CasperWA: Hereby done. Tried to assimilate the doc string as well a bit.

aiida/orm/querybuilder.py
```diff
@@ -1773,11 +1746,25 @@ def queryhelp(self):
         The queryhelp can be used to create a copy of the QueryBuilder instance like so::
 
             qb = QueryBuilder(limit=3).append(StructureData, project='id').order_by({StructureData:'id'})
-            qb2=QueryBuilder(**qb.queryhelp)
+            qb2 = QueryBuilder(**qb.queryhelp)
 
-        :return: a queryhelp dictionary
+        This is a json-comptible dictionary.
```
ltalirz: Ah, so is the dictionary json-compatible by default, despite we are not using the corresponding method?
@lekah do you want to comment?

aiida/orm/querybuilder.py
```diff
@@ -1773,11 +1746,25 @@ def queryhelp(self):
...
+        In this way, the queryhelp can be stored in the database or a json-object, retrieved or shared and used later.
+        The following is True if no change has been made to the database::
```
ltalirz: why not simply append the line `qb.all() == qb2.all()  # true` to the code above?

aiida/orm/querybuilder.py
CasperWA: Since the property `queryhelp` was simply a proxy for `get_json_compatible_queryhelp` before, I guess so?

aiida/orm/querybuilder.py
CasperWA: Could've done that, sure. I just preferred the separation of "concepts", but if you prefer them together I can update it?

aiida/orm/querybuilder.py
CasperWA: > (..) despite us not using the corresponding method?

I am not sure, but I think you mixed the methods `get_json_compatible_queryhelp` and `_get_json_compatible`? The latter is completely unused in the code base, while the first generates the current `queryhelp` dictionary - i.e., the `queryhelp` property was a proxy for the first method.

lekah: If I understood correctly, this is just a renaming: the instance method get_json_compatible_queryhelp is moved to the attribute .queryhelp, but it returns the same thing. However, whether the dictionary can be stored as JSON is indeed not being checked, so I can't answer whether it's the default (it was back in the day). I feel that this is outside the scope of this PR, though. I'd suggest creating a new issue for adding tests whether the queryhelp obtained is JSON compatible, and put in proper tests.

CasperWA: So I should remove the line specifying that `queryhelp` returns a json-compatible dictionary at the moment?

aiida/orm/querybuilder.py
CasperWA: It is done.

aiida/orm/querybuilder.py
CasperWA: It has been removed.

aiida/orm/querybuilder.py
```diff
@@ -1773,11 +1746,22 @@ def queryhelp(self):
...
 
             qb = QueryBuilder(limit=3).append(StructureData, project='id').order_by({StructureData:'id'})
-            qb2=QueryBuilder(**qb.queryhelp)
+            qb2 = QueryBuilder(**qb.queryhelp)
+
+            # The following is True if no change has been made to the database.
+            # Note that such a comparison can only be True if the order of results is enforced
+            qb.all() == qb2.all()
 
-        :return: a queryhelp dictionary
```
ltalirz: still one json-compatible left here, I'm afraid ;-)

ltalirz: otherwise looks good to merge for me

CasperWA: done :) I'll squash and merge once the tests go through (and it is approved - again)

PR #3537, opened 2019-11-12
Patching export for unicode issue

### LINKED ISSUE #3492: Error exporting some files (another Unicode issue)
Here is the error, reported by @AntimoMarrazzo:
```
STARTING EXPORT...
RETRIEVING LINKED NODES AND STORING LINKS...
STORING DATABASE ENTRIES...
Exporting a total of 342751 db entries, of which 227939 nodes.
STORING NODE ATTRIBUTES AND EXTRAS...
STORING GROUP ELEMENTS...
STORING DATA...
  File "/home/marrazzo/.virtualenvs/migrate_scdm/bin/verdi", line 10, in <module>
    sys.exit(verdi())
  File "/home/marrazzo/.virtualenvs/migrate_scdm/local/lib/python2.7/site-packages/click/core.py", line 764, in __call__
  File "/home/marrazzo/.virtualenvs/migrate_scdm/local/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/marrazzo/.virtualenvs/migrate_scdm/local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
  File "/home/marrazzo/.virtualenvs/migrate_scdm/local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
  File "/home/marrazzo/.virtualenvs/migrate_scdm/local/lib/python2.7/site-packages/click/core.py", line 956, in invoke
  File "/home/marrazzo/.virtualenvs/migrate_scdm/local/lib/python2.7/site-packages/click/core.py", line 555, in invoke
  File "/home/marrazzo/.virtualenvs/migrate_scdm/aiida-core/aiida/cmdline/utils/decorators.py", line 68, in wrapper
  File "/home/marrazzo/.virtualenvs/migrate_scdm/aiida-core/aiida/cmdline/commands/cmd_export.py", line 138, in create
    export_function(entities, outfile=output_file, **kwargs)
  File "/home/marrazzo/.virtualenvs/migrate_scdm/aiida-core/aiida/tools/importexport/dbexport/__init__.py", line 92, in export_zip
    export_tree(what, folder=folder, silent=silent, **kwargs)
  File "/home/marrazzo/.virtualenvs/migrate_scdm/aiida-core/aiida/tools/importexport/dbexport/__init__.py", line 417, in export_tree
    fhandle.write(json.dumps(data))
  File "/home/marrazzo/.virtualenvs/migrate_scdm/aiida-core/aiida/tools/importexport/dbexport/zip.py", line 47, in __exit__
    self.close()
  File "/home/marrazzo/.virtualenvs/migrate_scdm/aiida-core/aiida/tools/importexport/dbexport/zip.py", line 39, in close
    self._zipfile.writestr(self._fname, self._buffer.read())
  File "/usr/lib/python2.7/zipfile.py", line 1236, in writestr
    zinfo.CRC = crc32(bytes) & 0xffffffff       # CRC-32 checksum
UnicodeEncodeError: 'ascii' codec can't encode character u'\xc9' in position 281806205: ordinal not in range(128)
```

@sphuber @CasperWA do you agree that this would be a correct fix?
```python
    # N.B. We're really calling zipfolder.open
    with folder.open('data.json', mode='wb') as fhandle:
        # fhandle.write(json.dumps(data, cls=UUIDEncoder))
        fhandle.write(json.dumps(data).encode('utf-8'))
```
Changes: mode from `w` to `wb`, and adding `.encode('utf-8')`. I think this should work both in py2 and py3?

CasperWA:
Fixes #3492

This PR attemps to accommodate the issue described in #3492 by adding encoding info when opening the json files for writing during export and further adding some encoding info in our own `ZipFolder`.

@AntimoMarrazzo, could you possibly try and export using this branch and see if it fixes your usecase?

### DIFF
--- a/aiida/tools/importexport/dbexport/__init__.py
+++ b/aiida/tools/importexport/dbexport/__init__.py
@@ -674,7 +674,7 @@ def export_tree(
     }
 
     # N.B. We're really calling zipfolder.open (if exporting a zipfile)
-    with folder.open('data.json', mode='w') as fhandle:
+    with folder.open('data.json', mode='w', encoding='utf-8') as fhandle:
         # fhandle.write(json.dumps(data, cls=UUIDEncoder))
         fhandle.write(json.dumps(data))
 
@@ -695,7 +695,7 @@ def export_tree(
         }
     }
 
-    with folder.open('metadata.json', 'w') as fhandle:
+    with folder.open('metadata.json', 'w', encoding='utf-8') as fhandle:
         fhandle.write(json.dumps(metadata))
 
     EXPORT_LOGGER.debug('ADDING REPOSITORY FILES TO EXPORT ARCHIVE...')
--- a/aiida/tools/importexport/dbexport/__init__.py
+++ b/aiida/tools/importexport/dbexport/__init__.py
@@ -673,10 +673,9 @@ def export_tree(
         'groups_uuid': groups_uuid
     }
 
-    # N.B. We're really calling zipfolder.open (if exporting a zipfile)
-    with folder.open('data.json', mode='w', encoding='utf-8') as fhandle:
-        # fhandle.write(json.dumps(data, cls=UUIDEncoder))
-        fhandle.write(json.dumps(data))
+    # NOTE: We're really calling zipfolder.open (if exporting a zipfile)
+    with folder.open('data.json', mode='w') as fhandle:
+        fhandle.write(json.dumps(data))  # json.dumps will always return a str
 
     # Turn sets into lists to be able to export them as JSON metadata.
     for entity, entity_set in entities_starting_set.items():
@@ -695,8 +694,8 @@ def export_tree(
         }
     }
 
-    with folder.open('metadata.json', 'w', encoding='utf-8') as fhandle:
-        fhandle.write(json.dumps(metadata))
+    with folder.open('metadata.json', 'w') as fhandle:
+        fhandle.write(json.dumps(metadata))  # json.dumps will always return a str
 
     EXPORT_LOGGER.debug('ADDING REPOSITORY FILES TO EXPORT ARCHIVE...')
 
--- a/aiida/tools/importexport/dbexport/zip.py
+++ b/aiida/tools/importexport/dbexport/zip.py
@@ -16,10 +16,11 @@ import zipfile
 
 class MyWritingZipFile:
 
-    def __init__(self, zip_file, fname):
+    def __init__(self, zip_file, fname, encoding):
         self._zipfile = zip_file
         self._fname = fname
         self._buffer = None
+        self._encoding = encoding
 
     def open(self):
         if self._buffer is not None:
@@ -27,7 +28,10 @@ class MyWritingZipFile:
         self._buffer = io.StringIO()
 
     def write(self, data):
-        self._buffer.write(data)
+        if self._encoding:
+            self._buffer.write(str(data.encode(self._encoding)))
+        else:
+            self._buffer.write(data)
 
     def close(self):
         self._buffer.seek(0)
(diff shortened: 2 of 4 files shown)


aiida/tools/importexport/dbexport/zip.py
```diff
@@ -21,18 +21,22 @@
...
 
     def open(self):
         if self._buffer is not None:
             raise IOError('Cannot open again!')
         self._buffer = six.moves.StringIO()
 
     def write(self, data):
-        self._buffer.write(data)
+        if self._encoding:
+            self._buffer.write(str(data.encode(self._encoding)))
```
sphuber: why the casting to `str`?

aiida/tools/importexport/dbexport/zip.py
```diff
@@ -99,9 +103,12 @@ def close(self):
     def pwd(self):
         return self._pwd
 
-    def open(self, fname, mode='r'):
+    def open(self, fname, mode='r', encoding='utf8'):
+        if 'b' in mode:
+            encoding = None
+
         if mode == 'w':
```
sphuber: This should rather be `if 'w' in mode` because now the `wb` mode will go to the else

aiida/tools/importexport/dbexport/zip.py
sphuber: should the buffer type not change to `io.BytesIO` if the user is going to pass bytes to `write`?

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -412,7 +412,7 @@ def export_tree(
     }
 
     # N.B. We're really calling zipfolder.open (if exporting a zipfile)
-    with folder.open('data.json', mode='w') as fhandle:
+    with folder.open('data.json', mode='w', encoding='utf-8') as fhandle:
```
sphuber: The main question is under what conditions did @AntimoMarrazzo case fail and why was this not caught by the tests. I would first want to fully understand the failure and then add a test.

ramirezfranciscof:
@CasperWA what is the status of this? are you waiting for @AntimoMarrazzo to test it? did you discuss @sphuber comment in person? (in which case shouldn't we mark them as resolved?)

sphuber:
Since this has been open for very long and needs more work, is it ok to close this @CasperWA ?

CasperWA:
> Since this has been open for very long and needs more work, is it ok to close this @CasperWA ?

Please wait a bit longer. I will get around to it soonish - at least to fully review what needs to be done and at the very least update the original issue / update the PR.

CasperWA:
The original issue pertains to discrepancies between Py2 and Py3. A good explanation can be found in [this StackOverflow](https://stackoverflow.com/questions/18034272/python-str-vs-unicode-types) accepted answer.
In short, in Py2 `str` was stored as a bytes object, while `unicode` was stored as a "text" object. In Py3 `unicode` was changed to `str`, i.e., `str` is now a "text" object and a new class `bytes` replaced the old style `str` class.
So the issue is the `str` class.

In the current version of AiiDA we do not support Py2, so this is not an issue.
However, to be more clear, I have aligned the custom `ZipFolder` class with the custom `Folder` class, so that `open` and `write` works in the same way, i.e., with a `BytesIO` stream buffer when writing.
Also, I have added a check for the type of data when writing, which will raise `TypeError` if it's incorrect, e.g., if you try to write a `str` object when having opened it in simple `'w'` mode, and vice versa when opening in `'wb'` mode.

CasperWA:
Thinking further about this, the issue is actually that we're using the custom `MyWritingZipFile` class, since `zipfile.ZipFile.writestr` works fine with either `bytes` or `str` - it will simply convert a `str` to a UTF-8 encoded `bytes` object behind the scenes ... So if we don't go through the buffer in `MyWritingZipFile`...

CasperWA:
I think perhaps this PR should be closed, and issue #3492 should be directed towards the mixed Py2/Py3 versions.

I will create another PR for the changes I have introduced in this PR, but they have nothing to do with the original issue, but are simply a streamlining of the `aiida.tools.importexport.dbexport.zip` module instead (I have further local changes than what has been shown here).

PR #3543, opened 2019-11-14
Fix `UpfData` constructor not forwarding `filename` argument

### LINKED ISSUE #3542: Constructor of `UpfData` does not properly pass `filename` argument along

sphuber:
Fixes #3542 

The feature to set the desired filename of `SinglefileData` nodes and
its subclasses was recently added, but the constructor of `UpfData` was
not properly forwarding the new argument, causing it to be ignored.


sphuber:
> Sorry about this, and thanks for the fix!

ba, I did the review so equally guilty :)

PR #3544, opened 2019-11-14
update computer test output

### LINKED ISSUE #3415: Parse UNLIMITED and NOT_SET time in SLURM scheduler
When running things on a SLURM scheduler, my log is usually cluttered with 

```
10/01/2019 03:58:01 PM <69426> aiida.scheduler.slurm: [WARNING] Unrecognized format for time string 'UNLIMITED'
10/01/2019 03:58:01 PM <69426> aiida.scheduler.slurm: [WARNING] Error parsing the time limit for job id 622565
```

This should be easy to fix: In the ``_convert_time`` method of the SLURM plugin, return ``float('inf')`` if the string matches ``'UNLIMITED'`` (and ``None`` for ``'NOT_SET'``, which is also possible according to SLURM docs).

Does anyone know if this could have consequences down the line because the places where this is used (``this_job.requested_wallclock_time_seconds`` and `` this_job.wallclock_time_seconds``) are assumed to be ``int``? They both go into the ``JobInfo``, and ``None`` and ``float('inf')`` are JSON-serializable.

ltalirz:
This PR updates the output of `verdi computer test`

Old output:
Screenshot 2019-11-14 at 14 05 01

New output:
Screenshot 2019-11-14 at 14 13 02

### DIFF
--- a/aiida/cmdline/commands/cmd_computer.py
+++ b/aiida/cmdline/commands/cmd_computer.py
@@ -63,7 +63,8 @@ def _computer_test_get_jobs(transport, scheduler, authinfo):  # pylint: disable=
     """
     echo.echo('> Getting job list...')
     found_jobs = scheduler.get_jobs(as_dict=True)
-    echo.echo('  `-> OK, {} jobs found in the queue.'.format(len(found_jobs)))
+    echo.echo('  [{} jobs found in the queue]'.format(len(found_jobs)))
+    click.secho('  [OK]', fg=echo.COLORS['success'], bold=echo.BOLD)
     return True
 
 
@@ -119,7 +120,7 @@ https://github.com/aiidateam/aiida-core/issues/1890
         )
         return False
 
-    echo.echo('      [OK]')
+    click.secho('  [OK]', fg=echo.COLORS['success'], bold=echo.BOLD)
     return True
 
 
@@ -141,11 +142,11 @@ def _computer_create_temp_file(transport, scheduler, authinfo):  # pylint: disab
 
     file_content = "Test from 'verdi computer test' on {}".format(datetime.datetime.now().isoformat())
     echo.echo('> Creating a temporary file in the work directory...')
-    echo.echo('  `-> Getting the remote user name...')
+    echo.echo('  -> Getting the remote user name...')
     remote_user = transport.whoami()
-    echo.echo('      [remote username: {}]'.format(remote_user))
+    echo.echo('     [remote username: {}]'.format(remote_user))
     workdir = authinfo.get_workdir().format(username=remote_user)
-    echo.echo('      [Checking/creating work directory: {}]'.format(workdir))
+    echo.echo('  -> Checking/creating work directory: {}'.format(workdir))
 
     try:
         transport.chdir(workdir)
@@ -155,18 +156,18 @@ def _computer_create_temp_file(transport, scheduler, authinfo):  # pylint: disab
 
     with tempfile.NamedTemporaryFile(mode='w+') as tempf:
         fname = os.path.split(tempf.name)[1]
-        echo.echo('  `-> Creating the file {}...'.format(fname))
+        echo.echo('  -> Creating the file {}...'.format(fname))
         remote_file_path = os.path.join(workdir, fname)
         tempf.write(file_content)
         tempf.flush()
         transport.putfile(tempf.name, remote_file_path)
-    echo.echo('  `-> Checking if the file has been created...')
+    echo.echo('  -> Checking if the file has been created...')
     if not transport.path_exists(remote_file_path):
         echo.echo_error('* ERROR! The file was not found!')
         return False
 
-    echo.echo('      [OK]')
-    echo.echo('  `-> Retrieving the file and checking its content...')
+    echo.echo('     [OK]')
+    echo.echo('  -> Retrieving the file and checking its content...')
 
     handle, destfile = tempfile.mkstemp()
     os.close(handle)
@@ -174,7 +175,7 @@ def _computer_create_temp_file(transport, scheduler, authinfo):  # pylint: disab
         transport.getfile(remote_file_path, destfile)
         with io.open(destfile, encoding='utf8') as dfile:
             read_string = dfile.read()
-        echo.echo('      [Retrieved]')
+        echo.echo('     [Retrieved]')
         if read_string != file_content:
             echo.echo_error('* ERROR! The file content is different from what was expected!')
             echo.echo('** Expected:')
@@ -183,13 +184,14 @@ def _computer_create_temp_file(transport, scheduler, authinfo):  # pylint: disab
             echo.echo(read_string)
             return False
 
-        echo.echo('      [Content OK]')
+        echo.echo('     [Content OK]')
     finally:
         os.remove(destfile)
 
-    echo.echo('  `-> Removing the file...')
+    echo.echo('  -> Removing the file...')
     transport.remove(remote_file_path)
-    echo.echo('  [Deleted successfully]')
+    echo.echo('     [Deleted successfully]')
+    click.secho('  [OK]', fg=echo.COLORS['success'], bold=echo.BOLD)
     return True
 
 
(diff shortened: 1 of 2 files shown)


aiida/schedulers/plugins/test_slurm.py
```diff
@@ -46,43 +54,46 @@ def test_parse_common_joblist_output(self):
...
+        self.assertEquals(get_job('863553').submission_time, datetime.datetime(2013, 5, 23, 10, 42, 11))
+
+        self.assertEquals(get_job('863100').annotation, 'Resources')
+        self.assertEquals(get_job('863100').num_machines, 32)
+        self.assertEquals(get_job('863100').num_mpiprocs, 1024)
+        self.assertEquals(get_job('863100').queue_name, 'normal')
+
+        self.assertEquals(get_job('861352').title, 'Pressure_PBEsol_0')
+
+        self.assertEquals(get_job('863554').requested_wallclock_time_seconds, None)  # not set
```
ltalirz: One thing I noticed is that this time will also be `None`, if the value could not be parsed.
Perhaps ok, but it means that finding a value `None` does no longer mean that parsing failed.

sphuber:
I feel the change you are making to the scheduler parsing is too important to be muffled away in a commit on changing the output format of a computer test function. Would you be willing to separate that in another PR?

ltalirz:
> Would you be willing to separate that in another PR? 

Done https://github.com/aiidateam/aiida-core/pull/3586

aiida/cmdline/commands/cmd_computer.py
```diff
@@ -141,11 +142,11 @@ def _computer_create_temp_file(transport, scheduler, authinfo):  # pylint: disab
 
     file_content = "Test from 'verdi computer test' on {}".format(datetime.datetime.now().isoformat())
     echo.echo('> Creating a temporary file in the work directory...')
-    echo.echo('  `-> Getting the remote user name...')
+    echo.echo('  -> Getting the remote user name...')
```
sphuber: Maybe it is nice to put the `[OK]` on the same line as the first line.
This would go for all other lines as well where there is no other intermediate result but just an ok.
Put the message with ellipses and OK on the same line

aiida/cmdline/commands/cmd_computer.py
```diff
@@ -155,26 +156,26 @@ def _computer_create_temp_file(transport, scheduler, authinfo):  # pylint: disab
...
+    echo.echo('  -> Retrieving the file and checking its content...')
 
     handle, destfile = tempfile.mkstemp()
     os.close(handle)
     try:
         transport.getfile(remote_file_path, destfile)
         with io.open(destfile, encoding='utf8') as dfile:
             read_string = dfile.read()
-        echo.echo('      [Retrieved]')
+        echo.echo('     [Retrieved]')
```
sphuber: maybe remove this one and just put `[OK]` instead of `[Content OK]` on same line.

aiida/cmdline/commands/cmd_computer.py
```diff
@@ -183,13 +184,14 @@ def _computer_create_temp_file(transport, scheduler, authinfo):  # pylint: disab
...
-        echo.echo('      [Content OK]')
+        echo.echo('     [Content OK]')
     finally:
         os.remove(destfile)
 
-    echo.echo('  `-> Removing the file...')
+    echo.echo('  -> Removing the file...')
     transport.remove(remote_file_path)
-    echo.echo('  [Deleted successfully]')
+    echo.echo('     [Deleted successfully]')
```
sphuber: same here, just `[OK]` would suffice?

aiida/cmdline/commands/cmd_computer.py
```diff
@@ -526,9 +530,9 @@ def computer_test(user, print_traceback, computer):
                     num_failures += 1
 
         if num_failures:
-            echo.echo('Some tests failed! ({} out of {} failed)'.format(num_failures, num_tests))
+            echo.echo_warning('Some tests failed! ({} out of {} failed)'.format(num_failures, num_tests))
         else:
-            echo.echo('Test completed (all {} tests succeeded)'.format(num_tests))
+            echo.echo_success('All {} tests succeeded)'.format(num_tests))
```
sphuber: the string contains unbalanced closing parens

aiida/cmdline/commands/cmd_computer.py
ltalirz: I agree this looks nicer and I wanted to do this at first. It's nasty though - what do you do if the test fails?

During test failures, some commands print random stuff, which will then inevitably land on the same line.
For the "echo" statements originating from the command itself one could handle it, but also there you would need to do quite a bit of if/else in order to avoid having stuff printed on the same line that shouldn't be.

I've made the changes here https://github.com/ltalirz/aiida-core/tree/computer-test-echo-nl-false
Let me know what you prefer.

sphuber: If it fails, we can still print the error after a `[FAILED]` right? Or even on the next line if it is multiple lines. But I would imagine something like this:
```
* Getting job list... [OK]: Found 6 jobs in the queue
* Checking that no spurious output is present... [FAILED]
* Getting the remote user name... [OK]: <ltalirz>
* Retrieving the file and checking its content... [FAILED]: file not found
```

ltalirz: well, that is if you control everything that happens between printing "..." and the failure.
currently, many of the methods being called here just randomly print stuff (and not just the ones defined inside this file).
I'd have to start capturing all of this - can be done but is quite some work

sphuber: we can just put the calling blocks in `with Capturing()` no? If you are ok with it, I will gladly update your other branch to do it.

ltalirz: sure, please go ahead!
you have write acces to my fork already

sphuber:
@ltalirz I have updated the branch. I intended to push this to your second "test" branch but accidentally did it to the PR branch. Anyway, I didn't touch your commit, just rebased it so it is up-to-date. The additional commit will make it look as follows:
# Success
verdi_computer_test_success
# Failed
verdi_computer_test_failed
# With traceback
verdi_computer_test_traceback

aiida/cmdline/commands/cmd_computer.py
```diff
@@ -501,43 +491,70 @@ def computer_test(user, print_traceback, computer):
...
+        _computer_test_get_jobs: 'Getting number of jobs from scheduler',
+        _computer_get_remote_username: 'Determining remote user name',
+        _computer_create_temp_file: 'Creating and deleting temporary file'
+    }
+
     try:
-        echo.echo('> Testing connection...')
+        echo.echo('* Opening connection... ', nl=False)
+
         with transport:
```
ltalirz: What if an error happens here (e.g. configure a wrong username).
It will start printing stuff...

aiida/cmdline/commands/cmd_computer.py
```diff
@@ -501,43 +491,70 @@ def computer_test(user, print_traceback, computer):
...
-            scheduler.set_transport(transport)
             num_tests += 1
-            for test in [_computer_test_no_unexpected_output, _computer_test_get_jobs, _computer_create_temp_file]:
+
+            click.secho('[OK]', fg=echo.COLORS['success'], bold=echo.BOLD)
```
ltalirz: in my 2nd PR I introduced an `echo.echo_highlight` command that saves one from specifying the `bold` parameter + one can pass directly `'success'`.
Can save some text...

aiida/cmdline/commands/cmd_computer.py
sphuber: I don't think so, it will be caught in the outer try-except and that also just prints the error or including the traceback if the option is specified

aiida/cmdline/commands/cmd_computer.py
sphuber: I have added this to your first commit and then used it in mine.

sphuber:
> For me it's anyhow an improvement and good to merge.
> 
> @sphuber Since it's my PR, you will need to approve yourself ;-)

I included `echo_highlight` to your commit and then used it in mine. If you can confirm once more that you are happy with this I will merge it. Thanks

PR #3546, opened 2019-11-14
add --force option for verdi code delete

### LINKED ISSUE #3545: allow to delete nodes starting from code
Currently, this is what happens when trying to delete a code that has associated calculations:
```
 $ verdi code delete 19
Error: cannot delete Node<19> because it has output links

$ verdi node delete 19
I will delete 4 nodes
Warning: YOU ARE ABOUT TO DELETE 4 NODES! THIS CANNOT BE UNDONE!
Shall I continue? [y/N]: y
Starting node deletion...
Nodes deleted from database, deleting fil
```

I suggest we add a ``--force`` option to `verdi code delete` that simply pipes through to `verdi node delete`.

ltalirz:
fix #3545 

verdi code delete was preventing users from deleting codes with
associated calculations.
While it's good to warn users, preventing them from doing it entirely
seems unnecessary, also given that they could simply use verdi node
delete (which most users won't realize).

Thus adding a --force option that pipes `verdi code delete` through to
`verdi node delete`.


aiida/cmdline/commands/cmd_code.py
```diff
@@ -172,25 +172,32 @@ def show(code, verbose):
 
 @verdi_code.command()
 @arguments.CODES()
+@options.FORCE(help='Force deletion of code and all calculations using it.')
```
ltalirz: I could add a `NON_INTERACTIVE`  option as well.

One small issue is that for `verdi node delete`, the option to suppress prompts is called `--force`.
I think this should actually be called `--non-interactive` as well.

ltalirz: Another option is: I don't modify `verdi code delete` at all, except for adding the info message:
`echo.echo_info("Use 'verdi node delete <pk>' to delete the code and all of its calculations")`.

ramirezfranciscof:
I'm not sure I understand. Why was it preventing the deletion of those codes? were they trying to delete the codes (which are data nodes) without deleting the calculations associated? If this is so, passing the force keyword from code delete to node delete wouldn't help because it is only used to avoid asking for confirmation, it does not allow you to override fixed traversal rules which would be the issue in that case.

ramirezfranciscof:
Now, although adding the force keyword could be useful, I don't see why for verdi code delete you get an error instead of the information and confirmation prompt you get when you call verdi node delete.

Also, mildly unrelated: is `verdi code delete` checking that the provided pks are actually codes instead of just any node?

ltalirz:
>Now, although adding the force keyword could be useful, I don't see why for verdi code delete you get an error instead of the information and confirmation prompt you get when you call verdi node delete.

Well, one could pipe through directly to `verdi node delete` - but this would be a larger step from the current status (impossible to delete a code with associated calculations).
People might argue that it is dangerous to make it too easy to delete a code.
Since there is a prompt, however, I would be perfectly fine with that.

@sphuber @giovannipizzi What's your take on this?

> is verdi code delete checking that the provided pks are actually codes instead of just any node?

Yes.

ramirezfranciscof:
> Well, one could pipe through directly to `verdi node delete` - but this would be a larger step from the current status (impossible to delete a code with associated calculations).

But I mean, isn't that what you wanted with this PR? (or at least one of the options proposed) To be able to have 'verdi code delete' work when deleting codes that have calculations attached?

ltalirz:
@ramirezfranciscof I just updated the PR to indeed simply pipe through to the function used by `verdi node delete`.
I agree that this is the most straightforward solution.

aiida/backends/djsite/utils.py
```diff
@@ -162,16 +162,23 @@ def check_schema_version(profile_name):
         set_db_schema_version(code_schema_version)
         db_schema_version = get_db_schema_version()
 
-    if code_schema_version != db_schema_version:
+    if code_schema_version > db_schema_version:
```
sphuber: This code also exists for SqlAlchemy, maybe you want to update it there as well. Although this is not as easy as there you cannot simply compare revisions by their value. Anyway this code is touched heavily by PR #3582

aiida/cmdline/commands/cmd_code.py
```diff
@@ -172,25 +172,25 @@ def show(code, verbose):
...
-        else:
-            echo.echo_success('Code<{}> {} deleted'.format(pk, full_label))
+    verbosity = 1
+    if force:
+        verbosity = 0
+    elif verbose:
+        verbosity = 2
+
+    node_pks_to_delete = [code.pk for code in codes]
+    delete_nodes(node_pks_to_delete, dry_run=dry_run, verbosity=verbosity, force=force)
```
sphuber: What will the message now be here for a code with calculations and `force=False`? I guess since it pipes through to `delete_nodes` it will not explicitly mention code or its calculations but just nodes in general. Would it maybe be better to do the check in this command and raise a critical error unless `force=True`. Then the message can be made more specific.

aiida/backends/tests/cmdline/commands/test_code.py
```diff
@@ -192,14 +192,23 @@ def test_relabel_code_full_bad(self):
...
-    def test_delete_one(self):
+    def test_code_delete_one(self):
         result = self.cli_runner.invoke(delete, [str(self.code.pk)])
         self.assertIsNone(result.exception, result.output)
 
         with self.assertRaises(NotExistent):
             from aiida.orm import Code
             Code.get_from_string('code')
 
+    def test_code_delete_one_force(self):
```
sphuber: If we make the check in the command itself (see my comment below) then we can also add a test that when deleting a code with at least one calculation without the force flag will exit with a critical message

aiida/backends/djsite/utils.py
ltalirz: Ok, so how would one do it?
If there's no good way to do it, I guess we can leave the sqlalchemy version untouched.

aiida/cmdline/commands/cmd_code.py
ltalirz: An earlier version of the PR did exactly this - but then I also agree with @ramirezfranciscof that I don't really see a reason to be more strict with codes than with other nodes.
When one calls  `verdi code delete ...` without `--force`, it will simply prompt yes/no if there are other nodes attached.
To me this seems fine.

aiida/backends/djsite/utils.py
sphuber: It's fine for now, but these changes will probably become irrelevant with PR #3582 . But we can merge this now.

aiida/cmdline/commands/cmd_code.py
sphuber: Fair enough

sphuber:
Thanks @ltalirz

PR #3556, opened 2019-11-18
Fix REST API `attributes_filter` and `extras_filter`

### LINKED ISSUE #3555: REST API: attributes_filter and extras_filter filters not working properly with pagination
In the REST API `/nodes` endpoint we have the `extras_filter` and `attributes_filter` filters. When `attributes` and `extras` are set to true in the query string and `attributes_filter` and/or `extras_filter` are specified, the query returns the nodes with the attributes and/or extras specified there. Attributes and extras are returned in the `attributes` and `extras` dictionaries, where the keys are the ones specified in `attributes_filter` and `extras_filter`. 
This is not working correctly when pagination is requested, e.g. the request to the endpoint `/nodes/page/1?attributes=true&attributes_filter=cell` returns for each node dictionary an `attributes.cell` key instead of the expected `attributes: {cell: ... }` dictionary.

asle85:
Fixes #3555.

The REST API `attributes_filter` and `extras_filter` were not working correctly when pagination is requested.
Without pagination they return attributes and extras dictionaries with the keys specified in `attributes_filter` and `extras_filter` as expected.
When pagination is added they return instead the attributes keys as `attributes.key`, which is not consistent with the expected behaviour.

Changed REST API version to 4.0.1.


giovannipizzi:
Thanks!

Could you add a test to check this please?

sphuber:
@asle85 could you please still add the additional tests that we discussed? Then I can merge this and prepare the release. Thanks

asle85:
Improved tests on attributes_filter and extras_filter and added the two additional tests.

aiida/backends/tests/test_restapi.py
```diff
@@ -876,6 +876,76 @@ def test_structure_attributes_filter(self):
...
+                self.assertNotIn('attributes.resources', node)
+                self.assertNotIn('attributes.cell', node)
+                self.assertEqual(len(node['attributes']), len(expected_attributes))
+                for attr in expected_attributes:
+                    self.assertIn(attr, node['attributes'])
+
+    ############### node get one attributes_filter with pagination #############
+    def test_node_single_attributes_filter(self):
+        """
+        Check that when only one node extra is specified in extras_filter
```
sphuber: You still have "extra" twice in this docstring but should be "attribute".

asle85: Sorry, fixed it.

sphuber:
Thanks @asle85

PR #3561, opened 2019-11-21
Migrations for legacy and now illegal default link label `_return`

### LINKED ISSUE #3527: verdi import chokes on `_return` link label
Both @danieleongari and me independently have AiiDA databases, potentially migrated along from 0.12 that contain links labeled `_return`:

```
# select * from db_dblink WHERE label LIKE '\_%';
 id  |  label  | input_id | output_id |  type
-----+---------+----------+-----------+--------
  21 | _return |       63 |        64 | return
  33 | _return |      107 |       108 | return
  35 | _return |      245 |       246 | return
 227 | _return |      272 |       273 | return
 235 | _return |      112 |       113 | return
 271 | _return |      114 |       115 | return
 307 | _return |      270 |       271 | return
 337 | _return |       66 |        67 | return
 362 | _return |      240 |       119 | return
 438 | _return |      158 |       159 | return
 478 | _return |      190 |       191 | return
 512 | _return |       60 |        61 | return
 593 | _return |      222 |       223 | return
```

These end up in the AiiDA export file; and when one then tries to import the export file you get:

```
...
NEW Group: 063bf003-ed66-4a94-927a-f31129ba1824 (333->309)
NEW Group: 35db6bab-b93c-4bce-b44d-7989b86d27de (334->310)
STORING NODE LINKS...
Error: an exception occurred while importing the archive export_v1.aiida
  File "/home/daniele/aiida1/aiida_core/aiida/tools/importexport/dbimport/backends/django/__init__.py", line 553, in import_data_dj
    validate_link_label(link['label'])
  File "/home/daniele/aiida1/aiida_core/aiida/common/links.py", line 108, in validate_link_label
    raise ValueError('cannot start with an underscore')
ValueError: cannot start with an underscore
  File "/home/daniele/aiida1/aiida_core/aiida/cmdline/commands/cmd_import.py", line 67, in _try_import
    import_data(file_to_import, group, **kwargs)
  File "/home/daniele/aiida1/aiida_core/aiida/tools/importexport/dbimport/__init__.py", line 76, in import_data
    return import_data_dj(in_path, group=group, silent=silent, **kwargs)
  File "/home/daniele/aiida1/aiida_core/aiida/tools/importexport/dbimport/backends/django/__init__.py", line 555, in import_data_dj
    raise exceptions.ImportValidationError('Error during Link label validation: {}'.format(why))
aiida.tools.importexport.common.exceptions.ImportValidationError: Error during Link label validation: cannot start with an underscore
do you want to continue? [y/N]:
```

We checked that the link label in the error message is `_return`.

Is this a bug in the migrations?

Please advise on how to proceed as we need to produce working export files.
@giovannipizzi @sphuber

sphuber:
Fixes #3527 

The second commit adds the migration for export archives, however, it does not include the respective tests. @CasperWA would be willing to add those?


CasperWA:
For future reference, I refer to: https://github.com/aiidateam/aiida-core/wiki/Checklist-when-updating-the-export-version

And I can add the files, sure :+1:

ltalirz:
Don't wait for me - fine to go ahead

sphuber:
Would be good if @giovannipizzi could still have a look to see if the migrations are ok.

giovannipizzi:
I gave - I'm not fully comfortable with changing *all* link labels being `_return` (Maybe I would have limited to calcfunction and work function CREATE/RETURN links) - on the other hand there shouldn't be other people using this name, and now it's illegal so maybe ok?

This on the other hand raises a different question - should we migrate link labels of people that used an invalid one before validation is in place? Or it's ok to keep as is, and they just cannot create new ones? (I guess these are typically created by plugins, so plugin developers will anyway have to upgrade their plugins if using invalid names - the main question is if anything 'crashes' or raises if an invalid link is in the DB

sphuber:
> there shouldn't be other people using this name, and now it's illegal so maybe ok?

This was my thinking exactly. I don't think there is actual exceptions that can occur when just loading the node or links, but it will certainly prevent from importing archives that contain them, since then the link validation will stop the link from being created.

> This on the other hand raises a different question - should we migrate link labels of people that used an invalid one before validation is in place? Or it's ok to keep as is, and they just cannot create new ones? 

I don't think we can start to add data migrations for things done by plugins. We could remove leading and trailing underscores from link labels as that is what is now illegal, but not sure that we should. So far we have never considered plugin controlled data in data migrations and not sure if we should start now.

Issue #3547, opened 2019-11-14
Lots of dependencies are pinned to an exact (patch) version


csadorf:
Looking at the [long list of requirements](https://github.com/aiidateam/aiida-core/blob/develop/setup.json#L21), it seems that the vast majority of dependencies for aiida-core are pinned to a specific version, usually to a specific *patch* version. This makes it very difficult to setup a Python environment that is actually compliant with all of these dependencies.

I am wondering whether a lot of these requirements could be relaxed, because this makes it very difficult to manage a Python environment that depends on aiida-core.

For example, the [numpy package](https://pypi.org/project/numpy/#history) is currently pinned to version 1.16.4. Does that mean that aiida-core will not work with version 1.16.3 or 1.16.5? Or with version 1.17.4 (the currently latest release) for that matter?

I propose a strong relaxation of these requirements with liberal use of the [compatible release](https://www.python.org/dev/peps/pep-0440/#compatible-release) version specifier. If pinning versions for testing is required/desired, then these dependencies should be managed separately.

ramirezfranciscof:
> The pinning of numpy is a perfect example. Moving to `1.17` does indeed break our builds, even though it shouldn't as it is a minor version change.
> (...)
> I think starting to use compatible release on patch version like you propose would however already be a good change.

Probably I'm missing something, but from what I understood of compatible releases from @csadorf 's [link](https://www.python.org/dev/peps/pep-0440/#compatible-release), if we were using them the numpy update would have broken our builds anyways and we still would have needed to patch it, correct? If that was the case, you still think it would be a net positive to implement?

sphuber:
You can use the compatible release notation to specify that all patch releases should be considered. For example `numpy~=1.16.1` will install all `1.16.*` versions but not `1.17` etc. Technically it is still possible that patch releases break our code, but that is a lot rarer.

csadorf:
It is of course possible that even minor or patch releases breaks our builds due to poorly implemented upstream changes, however I'd argue that we should absolutely make it a priority to fix those builds (including patching upstream dependencies) instead of pinning to a particular patch version **unless there is absolutely no alternative**.

I understand that you interpret `aiida-core` to be an end-user application rather than a library, however I'd argue that since the package is supposed to support a multitude of workflows in combination with other applications, we should make it as easy as possible to install the package in conjunction with other applications.

[...]

PR #3566, opened 2019-11-21
Drop python 2

### LINKED ISSUE #3562: Drop python 2 support

sphuber:
Fixes #3562


setup.json
```diff
@@ -84,15 +80,12 @@
...
+      "sphinxcontrib-contentui==0.2.4",
+      "sphinxcontrib-details-directive==0.1.0"
     ],
     "atomic_tools": [
-      "PyCifRW==4.2.1; python_version < '3'",
-      "PyCifRW==4.4.1; python_version >= '3'",
+      "PyCifRW==4.4.1",
       "ase==3.17.0",
       "monty==2.0.4",
       "pymatgen<=2018.12.12",
```
greschd: I think for ``pymatgen`` and ``monty`` the main reason to use the pre-2019 version was also py2 compatibility.

aiida/sphinxext/__init__.py
```diff
@@ -11,10 +11,6 @@
 Defines reStructuredText directives to simplify documenting AiiDA and its plugins.
 """
 
-from __future__ import division
-from __future__ import print_function
-from __future__ import absolute_import
-
 from sphinx.errors import ExtensionError
```
greschd: The try / except just below this is also a py2 workaround (for some reason I can't comment there :upside_down_face:)

aiida/transports/transport.py
```diff
@@ -41,8 +35,7 @@ def validate_positive_number(ctx, param, value):  # pylint: disable=unused-argum
         raise BadParameter('{} is not a valid positive number'.format(value))
 
 
-@six.add_metaclass(ABCMeta)
-class Transport(object):
```
greschd: Same comment about the metaclass as above.

aiida/backends/general/abstractqueries.py
```diff
@@ -7,16 +7,11 @@
...
-from __future__ import print_function
-from __future__ import absolute_import
 
 from abc import ABCMeta
-import six
 
 
-@six.add_metaclass(ABCMeta)
-class AbstractQueryManager(object):
+class AbstractQueryManager:
```
greschd: Should the metaclass actually be removed, or just put with new syntax? Could also inherit from ``ABC`` instead.

aiida/backends/testimplbase.py
```diff
@@ -7,19 +7,14 @@
...
-from __future__ import absolute_import
 from abc import ABCMeta, abstractmethod
 
-import six
 
 from aiida import orm
 from aiida.common import exceptions
 
 
-@six.add_metaclass(ABCMeta)
```
greschd: Same as above: should the metaclass be removed?

aiida/backends/tests/cmdline/commands/test_data.py
```diff
@@ -19,7 +16,6 @@
 import numpy as np
 import subprocess as sp
 
-from six.moves import cStringIO as StringIO
```
greschd: Is there a ``from io import StringIO`` missing? Or maybe just put it as ``io.StringIO`` below?

aiida/backends/tests/test_nodes.py
```diff
@@ -1593,16 +1585,8 @@ def _check_existence(self, uuids_check_existence, uuids_check_deleted):
...
-            import sys
-            six.reraise(
-                type(exc),
-                exc.__class__((
-                    str(exc) + '\nCurrent UUID being processed: {}' + '\nFull uuids_check_existence: {}' +
-                    '\nFull uuids_check_deleted: {}'
-                ).format(uuid, uuids_check_existence, uuids_check_deleted)),
-                sys.exc_info()[2]
-            )
+        except Exception:  # pylint: disable=broad-except
```
greschd: This try/except doesn't do anything, right?

aiida/manage/backup/backup_base.py
```diff
@@ -8,27 +8,21 @@
...
 from dateutil.parser import parse
 from pytz import timezone as ptimezone
 
 from aiida.common import json
 from aiida.common import timezone as dtimezone
 
 
-@six.add_metaclass(ABCMeta)  # pylint: disable=useless-object-inheritance
-class AbstractBackup(object):
+class AbstractBackup:
```
greschd: Same here about the metaclass.

aiida/orm/implementation/backends.py
```diff
@@ -8,21 +8,16 @@
...
-import six
 
 __all__ = ('Backend', 'BackendEntity', 'BackendCollection', 'EntityType')
 
 EntityType = typing.TypeVar('EntityType')  # pylint: disable=invalid-name
 
 
-@six.add_metaclass(abc.ABCMeta)
-class Backend(object):
+class Backend:
```
greschd: Same here, is the metaclass needed?

aiida/orm/implementation/backends.py
```diff
@@ -121,8 +116,7 @@ def transaction(self):
         """
 
 
-@six.add_metaclass(abc.ABCMeta)
-class BackendEntity(object):
+class BackendEntity:
```
greschd: Same here, is the metaclass needed?

aiida/parsers/parser.py
```diff
@@ -25,8 +20,7 @@
 __all__ = ('Parser',)
 
 
-@six.add_metaclass(ABCMeta)
```
greschd: metaclass?

aiida/tools/dbimporters/plugins/materialsproject.py
```diff
@@ -45,13 +41,11 @@ def setup_db(self, **kwargs):
...
                 api_key = os.environ['PMG_MAPI_KEY']
-            except KeyError as exc:
-                raise_from(
-                    KeyError(
-                        'API key not supplied and PMG_MAPI_KEY environment '
-                        'variable not set. Either pass it when initializing the class, '
-                        'or set the environment variable PMG_MAPI_KEY to your API key.'
-                    ), exc
+            except KeyError:
+                KeyError(
```
greschd: This should probably be 
```
except KeyError as exc:
    raise KeyError(...) from exc
```

That way you get the proper "error chaining" (``The above exception was the direct cause of the following exception:`` instead of ``During handling of the above exception, another exception occurred:``).

setup.json
greschd: Although we can of course update these later as well.

sphuber: I was leaving all those dependencies to be dealt with in #3547 separately. We intend to relax the pinning as much as possible but that will require quite some testing. For now I just removed those that are explicitly for python<3.5 since that is no longer supported

aiida/transports/transport.py
sphuber: Good point. I put it for the `Node` class which used the `AbstractNodeMeta` but not sure if it is actually _needed_ for any of the others. Do you know when using ABC as metaclass is crucial as opposed to just making a class with abstract methods? Will the latter fail if class is not explicitly marked as abstract as well?

aiida/backends/tests/cmdline/commands/test_data.py
sphuber: no, but there is `import io` at the top and just used `io.StringIO`

aiida/transports/transport.py
sphuber: I think I understand now why nothing complained and everything still works when I just removed it. Subclassing from `ABC` is the same as setting the meta to `ABCMeta`. It is just a shortcut. Having `ABCMeta` as meta class is _necessary_ if you want to mark methods as abstract with `@abstractmethod`. Looking at the code where I removed it though, which is mostly very old code, it seems that the original writers never used the `@abstractmethod` decorators, but just raised `NotImplementedError`. This is not statically checked at launch time, unlike the decorators. Instantiating a sub class with an abstract method will raise, but if you don't override the method of the base class, no-one will notice until you call it.

In conclusion, I think these should for the most part be abstract classes but then their methods should be appropriately made abstract as well. The question is now, do I do this in this PR? Or shall I for now just make them subclass `ABC` directly and leave the fixing of the methods for another time?

aiida/backends/tests/test_nodes.py
sphuber: Correct, removed

aiida/transports/transport.py
greschd: I see, that makes sense.

For the sake of making it a bit easier to track the change, maybe it's best to just subclass ``ABC`` now, and then make a separate PR with the abstractmethods.

aiida/tools/dbimporters/plugins/materialsproject.py
sphuber: Done

aiida/transports/transport.py
sphuber: Done

aiida/backends/tests/cmdline/commands/test_data.py
greschd: Isn't there a bare ``StringIO`` [here](https://github.com/sphuber/aiida_core/blob/fix_3562_drop_python_2/aiida/backends/tests/cmdline/commands/test_data.py#L33)? I _think_ I'm looking at the right commit 😅

sphuber: yep you were right, but the function that it was in was never used :sweat_smile: the calls were commented out, so I will just remove the function

greschd:
I think the only thing left to do now is telling Jenkins to use py3 as well.

sphuber:
Yes, but I also wanted to merge some final things that should go into `v1.0.1` that I want to release soon. Then we can merge this in `develop` which is then py3 only and this will become `v1.1.0`. We are about to accept the AEP that defines that version as the one where we drop py2 support.

greschd:
@sphuber I've just stumbled across a *very important* little detail: In order for the "graceful fallback" on ``pip install`` to work for older versions, we need to *first* release another version that also specifies ``python_requires``. No idea why, but 

So that means, for the ``1.0.1`` release we should add ``python_requires=">=2.7"`` in ``setup.json``. 

No idea why that's the case, but just tested this with a different package. It's described [here (last paragraph)](https://packaging.python.org/guides/dropping-older-python-versions/#dropping-a-python-release), although you could easily miss it.

sphuber:
Good catch! I will open a PR now

sphuber:
@greschd if you'd like, I added three more commits on top since last time you reviewed. If you have the time you can just look at those commits.

aiida/common/hashing.py
```diff
@@ -160,19 +160,13 @@ def _single_digest(obj_type, obj_bytes=b''):
 
 @_make_hash.register(bytes)
 def _(bytes_obj, **kwargs):
-    """
-    Hash arbitrary binary strings (str in Python 2, bytes in Python 3).
-    For compat reason between Python 2 and 3, this gets the same hash-type
-    as for unicode in Python 2, resp. str in Python 3."""
+    """Hash arbitrary byte strings."""
     return [_single_digest('str', bytes_obj)]
```
greschd: Hmm, I'm not sure if we might want to change the "type specifier" here to be bytes - specific, now that py2 / py3 compatibility is no longer a concern.

PR #3568, opened 2019-11-23
Create backup of configuration file before migrating

### LINKED ISSUE #3510: Warn when configuration is migrated and create unique backup
Currently the configuration file is automatically migrated without any messages or asking for a confirmation. On top of that, the backup is written to the same file as all other backups. The next (silent) update, which means any change to the configuration file, will overwrite it yet again, causing the old structure to be lost forever.

sphuber:
Fixes #3510 

This allows users to revert in case of problems with the migration or if
the migration happens by accident. Since the migration is done
automatically without prompt and is not reversible, having a backup is
crucial if users checkout a version of the code by accident.

P.S.: tests not included.

### DIFF
--- a/aiida/manage/configuration/config.py
+++ b/aiida/manage/configuration/config.py
@@ -18,7 +18,6 @@ import shutil
 
 from aiida.common import json
 
-from .migrations import CURRENT_CONFIG_VERSION, OLDEST_COMPATIBLE_CONFIG_VERSION
 from .options import get_option, parse_option, NO_DEFAULT
 from .profile import Profile
 from .settings import DEFAULT_UMASK, DEFAULT_CONFIG_INDENT_SIZE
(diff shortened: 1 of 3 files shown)


sphuber:
Wanted to include this in a `v1.0.0` release with everything else that is currently on `develop` and then merge PR #3566

giovannipizzi:
I guess you meant `v1.0.1`, right?
I'm not sure anymore, didn't we say to ask explicitly to migrate instead of doing it silently? Maybe I'm wrong.
Probably OK to do `v1.0.1` (@asle85 could you confirm that you don't need at this moment additional changes to the REST API?), in case we can do a `v1.0.2` later (even if a bit more complex because we have to backport to py2).

sphuber:
Yes, I meant `v1.0.1`. I think the whole point of making a separate backup, with timestamp in case it is performed in quick succession, is to prevent having to add the complexity of a prompt. In most cases one actually does just want to migrate it. This is mostly for developers who are switching version often. @ltalirz what do you think?

aiida/manage/configuration/config.py
```diff
@@ -18,12 +18,34 @@
...
+    :param config_dictionary: contents of the configuration file
+    :param filepath: absolute path to the current configuration file
+    """
+    from aiida.common import timezone
+    from aiida.cmdline.utils import echo
+
+    version = get_current_version(config_dictionary)
+    filepath_backup = '{}.v{}.{}'.format(filepath, version, timezone.now().strftime('%Y%m%d-%H%M%S'))
+
+    echo.echo_warning('current configuration file `{}` is outdated and will be migrated'.format(filepath))
```
ltalirz: To me, these two echo statements don't belong in this function - the function should just create the backup, right?
I would move them to where you do the migration

aiida/manage/configuration/config.py
```diff
@@ -18,12 +18,34 @@
...
+    """Create backup of configuration.
+
+    :param config_dictionary: contents of the configuration file
+    :param filepath: absolute path to the current configuration file
+    """
+    from aiida.common import timezone
+    from aiida.cmdline.utils import echo
+
+    version = get_current_version(config_dictionary)
+    filepath_backup = '{}.v{}.{}'.format(filepath, version, timezone.now().strftime('%Y%m%d-%H%M%S'))
```
ltalirz: Maybe switch the order of version and date so that the files will be ordered by date.

Also, we already have a backup function that creates backups of the config file ending in `.json~`
https://github.com/aiidateam/aiida-core/blob/cd916288509f31ad088120f24193bb2c42c6afc2/aiida/manage/configuration/config.py#L336

Perhaps we should join with this function.

sphuber: Can't do because that is a member function of `Config` which in order to be constructed requires a fully migrated config dictionary, so at that point creating a backup will be too late. It would be possible if we move the logic that determines whether migration is necessary in the constructor of the `Config` class. I am not sure if that is better. The code there is already a bit complicated and might make it even more tightly coupled. On the other hand maybe it makes it more self-contained and therefore better. I will take a look what that would look like.

sphuber: Also re the ordering, with the current order they will still be ordered by data, within each version group, which should also be ordered, no?

aiida/manage/configuration/config.py
sphuber: I have looked at the code a bit better and think I have found a solution that is a lot better and aligns with your suggestion. Construction of `Config` from file is now done through `Config.from_file` class method which is now responsible for migrating and creating of backup. This bundles these two operations more closely, and for the better I feel.

aiida/manage/configuration/config.py
sphuber: Decided to omit the version as it is in the files themselves.

sphuber:
> thanks @sphuber , looks good to me.
> 
> now we can even do concurrent backups in the same microsecond ;-)

It is actually necessary without concurrent backups. If you look in the constructor there are multiple checks for inconsistencies and for each one it will create a backup. These can follow one another pretty quickly so having a granularity of seconds ran the risk of the first backup with the original file being overridden.

PR #3571, opened 2019-11-25
Enable CI on Github

### LINKED ISSUE #3553: Run tests on python 3.7

sphuber:
Fixes #3553 

I did not touch the files for the Jenkins tests yet. I want to see what it is exactly that we want to run there and if possible simplify the scripts and remove what we no longer need. The current setup also does not include coverage yet.

To move the CI to GitHub actions, I created a workflow `.github/workflows/ci.yml`.
This defines 5 different actions:

1. `conda`: will run `.github/workflows/conda.sh` to install conda and create the environment from `environment.yml`
2. `docs`: will build the documentation
3. `pre-commit`: runs pre-commit on all the files
4. `tests`: runs `.github/workflows/tests.sh` to run all tests for both backends, for now only for python 3.7 but other versions can easily be added in the matrix. The tests include the entire unit test suite through `verdi devel tests` as well as the various stand alone tests in `.ci`
5. `verdi`: tests the loading speed of verdi

The reason I put the actions in separate actions is that many of them do not require the same setup and like this they will run in parallel making the total run time shorter. I also feel that the files are clearer this way and it is easier to understand what is going on and what actions is responsible for what. The only downside of this approach is if we would run out of actions slots, but it seems there is only a limit on number of concurrent _workflows_, which at 20 per repository I don't think we will hit very often.

Finally, to simplify the configuration of profiles, computer and codes for the `tests` action, I have created `.yaml` configuration files in `.github/config` which are called in `.github/workflows/setup.sh`.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,145 @@
...
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+
+    - name: Install python dependencies
+      run:
+        pip install -e .
+
+    - name: Run verdi
```
ltalirz: just wondering: why is this a separate test?
Wouldn't we save a lot of setup time if we made this part of the rest of the tests?

sphuber:
@ltalirz just added an overview, let me know if something is still missing.

.ci/test_process.py
```diff
@@ -0,0 +1,407 @@
+# -*- coding: utf-8 -*-
```
ltalirz: Do I need to review this whole file - is it entirely new?
I guess not but I don't see a corresponding deletions...

.github/config/doubler.yaml
```diff
@@ -0,0 +1,9 @@
+---
```
ltalirz: why is this in `.github` while `test_process.py` remains in `.ci`?
I guess we keep general stuff in `.ci` and move only the github-specific things under `.github`?

.github/workflows/conda.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# Workaround for https://github.com/conda/conda/issues/9337
+pip uninstall -y setuptools
+conda install setuptools
+
+conda update -q conda
+conda info -a
+
+conda env create -f environment.yml -n test-environment python=$PYTHON_VERSION
+source activate test-environment
```
ltalirz: this last line could then also be removed

.github/workflows/ci.yml
```diff
@@ -0,0 +1,145 @@
...
+      run:
+        pre-commit run --all-files || ( git status --short ; git diff ; exit 1 )
+
+  tests:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        backend: ['django', 'sqlalchemy']
```
ltalirz: if the number of concurrent jobs is indeed not a problem, I would suggest we test at least django also on python 3.5

.github/workflows/tests.sh
```diff
@@ -0,0 +1,17 @@
...
+verdi -p test_${AIIDA_TEST_BACKEND} run .ci/test_process.py
+verdi daemon stop
+
+verdi -p test_$AIIDA_TEST_BACKEND devel tests
+pytest aiida/sphinxext/tests
+pytest .ci/test_test_manager.py
+pytest .ci/test_profile_manager.py
+python .ci/test_plugin_testcase.py
+pytest .ci/pytest
+TEST_AIIDA_PROFILE=test_$AIIDA_TEST_BACKEND pytest .ci/pytest
```
ltalirz:
```suggestion
AIIDA_TEST_PROFILE=test_$AIIDA_TEST_BACKEND pytest .ci/pytest
```

setup.json
```diff
@@ -120,7 +120,7 @@
     "dev_precommit": [
       "astroid==1.6.6; python_version<'3.0'",
       "astroid==2.2.5; python_version>='3.0'",
-      "pep8-naming==0.8.2",
+      "pep8-naming==0.4.1",
```
ltalirz: why the downgrade?

.github/workflows/tests.sh
sphuber: Done

.ci/test_process.py
sphuber: It was a copy of `test_daemon.py` because I initially though I was going to have to change quite a bit to get working and wanted to keep the original working on Travis/Jenkins before dropping those. Turns out it wasn't that big a change. I removed it and just fixed linter issues in `test_daemon.py`.

.github/workflows/conda.sh
sphuber: done

.github/workflows/ci.yml
sphuber: I have added python 3.6 to the matrix strategy for now. Is there a reason you prefer 3.5 over 3.6? We officially support both though but a matrix of 6 is maybe overdoing it a bit?

setup.json
sphuber: This was necessary because otherwise the pre-commit failed because `prospector` has this requirement. We can tackle these in the other issue about relaxing dependency requirements

.github/config/doubler.yaml
sphuber: Everything that is under `.github` now is GitHub specific though. None of it is being used elsewhere. I kept the `test_` scripts in `.ci` because those are still being run by Travis/Jenkins. Once we are happy with the current changes, I will remove Travis and trim down the requirements for Jenkins. This will allow me to clean up those files significantly. What is used by both I will keep in `.ci`

ltalirz: Sounds good.

.github/workflows/ci.yml
ltalirz: The thinking was that if it is working on 3.5 and 3.7 it's likely to work on 3.6 as well.
And I think it probably suffices to test sqlalchemy on one python version... anyhow up to you.

sphuber: OK, that makes sense. I will switch it to python 3.5

sphuber:
@ltalirz I think the current state then is ready. I will then add a commit to remove Travis and simplify the remaining code necessary to run on Jenkins

sphuber:
This is essentially ready. Once we have released `1.0.1` and then merged PR #3566 into `develop` this can be rebased. We can then turn of Travis as required checks and enable the new GitHub workflow. The only remaining differences with the current situation:

* Coverage has been removed: I vote that we try to move our unit testing to `pytest` soon and then use the coverage plugin. This should solve quite a few issues that are currently open related to coverage.
* Transifex: this deploy trigger is not yet ported
* PyPI release: this deploy trigger is also not yet ported, but I want to do this separately to make sure it works well.

sphuber:
@ltalirz with `v1.0.1` released, python 2 dropped, this PR rebased, it is now ready to go!

sphuber:
GitHub workflow run can be found here:https://github.com/sphuber/aiida_core/commit/64c63ad75dd3821c02ed2ad4b813140f7bad68e8/checks?check_suite_id=338130505

.github/workflows/ci.yml
```diff
@@ -0,0 +1,147 @@
...
+      run:
+        .github/workflows/setup.sh
+
+    - name: Run test suite
+      env:
+        AIIDA_TEST_BACKEND: ${{ matrix.backend }}
+      run:
+        .github/workflows/tests.sh
+
+  verdi:
```
ltalirz: still seems like a bit of a waste to do all of the installation just to run one test...
anyhow, it's fine for now

sphuber: Let's bask in the superfluous wealth of practically unlimited parallel builds :)

PR #3577, opened 2019-11-26
updates to AUTHORS.txt

ltalirz:
as proposed in #3498 (but postponed due to 1.0 release).


AUTHORS.txt
```diff
@@ -30,16 +31,15 @@ Former members of the AiiDA team
...
 
-Contributors
-############
-We would like to thank the following people for their major contibutions:
+## Contributors
+
+We would like to thank
 
 * Dominik Gresch (ETHZ): for the caching mechanism and contributions to the workflow engine
 * Tiziano Müller (ETHZ, Switzerland): for the first implementation of
```
sphuber:
```suggestion
* Tiziano Müller (ETHZ): for the first implementation of the plugin system and making the original code python 3 compatible
```

greschd: Last I've known @dev-zero was at UZH, not ETHZ.

AUTHORS.txt
```diff
@@ -30,16 +31,15 @@ Former members of the AiiDA team
...
 
 
-Contributors
-############
-We would like to thank the following people for their major contibutions:
+## Contributors
+
+We would like to thank
 
 * Dominik Gresch (ETHZ): for the caching mechanism and contributions to the workflow engine
```
greschd: Might as well change it to ``(Microsoft Quantum)`` if we're already updating this file.

ltalirz: Hm... since you are the only two with affiliation in the file, I think it's better to remove it entirely (otherwise we have to keep updating it)

greschd: Works for me 👍

PR #3579, opened 2019-11-26
Fix problem with dummy configuration file on ReadTheDocs

### LINKED ISSUE #3578: documentation build failures
The documentation on RTD recently started failing with

```
PermissionError: [Errno 13] Permission denied: '/dev/null.20191126-175659.723679'
```
https://readthedocs.org/projects/aiida-core/builds/10024937/

@sphuber This looks a bit like the format of the new configuration file backup?

sphuber:
Fixes #3578 

The recent change in the backup behavior of the configuration file leads
to problems on ReadTheDocs. The dummy config returned by `load_config`
when on RTD uses `/dev/null` as dummy filepath, but the new backup
mechanism will try to write a backup using that filepath as a prefix,
leading to a filepath that can't be written to. The temporary fix until
we get rid of this hacky dummy profile concept is to simply use the
default location of the config file.

### DIFF
--- a/aiida/manage/configuration/__init__.py
+++ b/aiida/manage/configuration/__init__.py
@@ -86,12 +86,15 @@ def load_config(create=False):
     from .config import Config
     from .settings import AIIDA_CONFIG_FOLDER, DEFAULT_CONFIG_FILE_NAME
 
+    filepath = os.path.join(AIIDA_CONFIG_FOLDER, DEFAULT_CONFIG_FILE_NAME)
+
     if IN_RT_DOC_MODE:
         # The following is a dummy config.json configuration that it is used for the
         # proper compilation of the documentation on readthedocs.
         from aiida.manage.external.postgres import DEFAULT_DBINFO
+        import tempfile
         return Config(
-            '/dev/null', {
+            tempfile.mkstemp()[1], {
                 'default_profile': 'default',
                 'profiles': {
                     'default': {
@@ -109,8 +112,6 @@ def load_config(create=False):
                 }
             }
         )
-
-    filepath = os.path.join(AIIDA_CONFIG_FOLDER, DEFAULT_CONFIG_FILE_NAME)
 
     if not os.path.isfile(filepath) and not create:
         raise exceptions.MissingConfigurationError('configuration file {} does not exist'.format(filepath))


sphuber:
With this fix it seems to build again: https://readthedocs.org/projects/aiida-core/builds/10026000/

aiida/manage/configuration/__init__.py
```diff
@@ -86,12 +86,14 @@ def load_config(create=False):
...
 
+    filepath = os.path.join(AIIDA_CONFIG_FOLDER, DEFAULT_CONFIG_FILE_NAME)
+
     if IN_RT_DOC_MODE:
         # The following is a dummy config.json configuration that it is used for the
         # proper compilation of the documentation on readthedocs.
         from aiida.manage.external.postgres import DEFAULT_DBINFO
         return Config(
-            '/dev/null', {
+            filepath, {
```
ltalirz: Ah hm - I wonder:  What if now someone sets the environment variable to run in RTD mode on their local laptop. Will this overwrite the existing config file?

Perhaps it would be safer to simply use some temporary file path?
Could we perhaps reuse some of the code from here:
https://github.com/aiidateam/aiida-core/blob/d750d513f7adf5126d43a230c9fb378de02f0772/aiida/backends/tests/utils/configuration.py#L50

sphuber: That's what I originally was trying as well, but then it is difficult to control the lifetime of it. I will have to create the tempfile in the `load_config` function, but then cannot control the lifetime. This way we will end up with uncleaned temp files. Frankly, I wanted to get rid of these hacks just for RTD for a long time. I remember you also opened an issue recently with some official API to do what we currently do in the `docs/source/conf.py` for I believe plugins to be able to do the same. I'd rather have this go in now and then fix it properly soon

ltalirz: Ok, fine not to merge the functionality.
I still think, though, that we should be sure not to overwrite an existing config file (creating a temporary one without deleting it is preferable).

Are we sure your approach does not overwrite an existing config file?

sphuber:
@ltalirz OK I replaced it with a real temporary file, but it is not cleaning it up.

greschd:
Hacky solution: why not put it in the sphinx BUILDDIR? Then ``make clean`` should take care of it.

Probably doesn't matter much if it will be fixed properly soon anyway, though.

PR #3581, opened 2019-11-27
Improve the backup mechanism of the configuration file

### LINKED ISSUE #3580: Only write configuration file backup when contents have changed
Now it is done on every `store` as well, even if there have been no changes and with the recent change to unique backup names, this will create way too many backups

sphuber:
Fixes #3580 

The `Config` class was creating a new backup each time `store` was
called, which also happens if nothing really changed to the configuration
content. With the recently introduced change that backup files are now
made unique through a timestamp, this led to a lot of backups being
created that were essentially clones.

To improve this, the `Config.store` method now only writes the file to
disk if the contents in memory have changed. This is done by comparing
the checksum of the on disk file and that memory contents as written to
a temporary file. If the check sums differ a backup is created of the
existing file and the new contents written to the temporary file are
copied to the actual configuration file location. This latter new
approach also guarantees that the backup is always created before
overwriting the original one.

### DIFF
--- a/aiida/manage/configuration/config.py
+++ b/aiida/manage/configuration/config.py
@@ -46,29 +46,20 @@ class Config(object):  # pylint: disable=too-many-public-methods
         from aiida.cmdline.utils import echo
         from .migrations import check_and_migrate_config, config_needs_migrating
 
-        created = False
-
         try:
             with io.open(filepath, 'r', encoding='utf8') as handle:
                 config = json.load(handle)
         except (IOError, OSError):
-            # The file does not exist so "touch" it and create an empty configuration
-            with io.open(filepath, 'ab'):
-                pass
-            config = {}
-            created = True
-
-        # If the configuration file needs to be migrated, first create a specific backup so it can easily be reverted
-        if config_needs_migrating(config):
-            echo.echo_warning('current configuration file `{}` is outdated and will be migrated'.format(filepath))
-            filepath_backup = cls._backup(filepath)
-            echo.echo_warning('original backed up to `{}`'.format(filepath_backup))
-
-        config = check_and_migrate_config(config)
-        config = Config(filepath, config)
-
-        if created:
+            config = Config(filepath, check_and_migrate_config({}))
             config.store()
+        else:
+            # If the configuration file needs to be migrated first create a specific backup so it can easily be reverted
+            if config_needs_migrating(config):
+                echo.echo_warning('current configuration file `{}` is outdated and will be migrated'.format(filepath))
+                filepath_backup = cls._backup(filepath)
+                echo.echo_warning('original backed up to `{}`'.format(filepath_backup))
+
+            config = Config(filepath, check_and_migrate_config(config))
 
         return config
 
--- a/aiida/manage/configuration/config.py
+++ b/aiida/manage/configuration/config.py
@@ -38,15 +38,15 @@ class Config(object):  # pylint: disable=too-many-public-methods
     def from_file(cls, filepath):
         """Instantiate a configuration object from the contents of a given file.
 
-        .. note:: if the filepath does not exist an empty file will be created and a `Config` instance with the default
-            configuration will be returned. Note that its contents are *not* yet written to disk but are only kept in
-            memory until `Config.store` is called.
+        .. note:: if the filepath does not exist an empty file will be created with the default configuration.
 
         :param filepath: the absolute path to the configuration file
         :return: `Config` instance
         """
         from aiida.cmdline.utils import echo
         from .migrations import check_and_migrate_config, config_needs_migrating
+
+        created = False
 
         try:
             with io.open(filepath, 'r', encoding='utf8') as handle:
@@ -56,6 +56,7 @@ class Config(object):  # pylint: disable=too-many-public-methods
             with io.open(filepath, 'ab'):
                 pass
             config = {}
+            created = True
 
         # If the configuration file needs to be migrated, first create a specific backup so it can easily be reverted
         if config_needs_migrating(config):
@@ -64,8 +65,12 @@ class Config(object):  # pylint: disable=too-many-public-methods
             echo.echo_warning('original backed up to `{}`'.format(filepath_backup))
 
         config = check_and_migrate_config(config)
+        config = Config(filepath, config)
 
-        return Config(filepath, config)
+        if created:
+            config.store()
+
+        return config
 
     @classmethod
     def _backup(cls, filepath):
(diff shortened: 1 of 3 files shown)


sphuber:
@ltalirz this should also be merged before the `v1.0.1` release because it was generating loads of backup files. Better to merge PR #3579 first though

aiida/manage/configuration/config.py
```diff
@@ -39,14 +38,22 @@ class Config(object):  # pylint: disable=too-many-public-methods
     def from_file(cls, filepath):
         """Instantiate a configuration object from the contents of a given file.
 
+        .. note:: if the filepath does not exist it will be created with the default configuration.
```
ltalirz: Where in the function is the "default content" written to the file?

aiida/manage/configuration/config.py
```diff
@@ -366,16 +369,44 @@ def get_option(self, option_name, scope=None, default=True):
         return value
 
     def store(self):
-        """Write the current config to file."""
-        if os.path.isfile(self.filepath):
-            self._backup(self.filepath)
+        """Write the current config to file.
+
+        .. note:: if the configuration file already exists on disk it will only be overwritten if there are changes
+            made to the configuration in memory. In that case, a backup of the original file on disk will be created.
```
ltalirz: Is that the behavior we want? If I call "conf.store()" I expect the content to be written to disk (i.e. the timestamp of the config file should be updated).

I completely agree, of course, that a backup is only necessary when the content changes.

sphuber: Fair enough, I can take the actual write call out of the conditional

aiida/manage/configuration/config.py
sphuber: It is not. The file is touched but is initially empty. The `Config` instance is then created with the default configuration but is not stored explicitly. This is done by the caller of `from_file` if needed. In our case this is done in `load_config` that calls `Config.from_file`. If I add to the docstring that the file is touched but the content is not written, is that clear enough?

sphuber:
> Thanks @sphuber
> One question: if I understand correctly, you are saying there are many calls to `store()` that are, essentially, unnecessary (since the file content actually isn't changing).
> Would it be possible to simply remove those?

This was mostly happening during running the unittests.

ltalirz:
> This was mostly happening during running the unittests.

Do you know whether this is an unintended consequence of `with_temporary_config_instance` or is it tests that *should* be using `with_temporary_config_onstance` but aren't ?

sphuber:
> Do you know whether this is an unintended consequence of `with_temporary_config_instance` or is it tests that _should_ be using `with_temporary_config_instance` but aren't ?

I think I found the problem and it is exactly because `from_file` delegates the storing of the file in case it is new to the caller. That's why `load_config` _always_ called `config.store` even if it was just loaded from file and nothing changed. It seems like it is better to put the store-if-created logic in the `from_file` method and then we can remove it from `load_config`. This should prevent `store` from being called all the time when running the tests.

aiida/manage/configuration/config.py
ltalirz: Taking a step back, would it not be cleaner if `.from_file` simply excepts if the filepath being passed does not exist?
If one wants to create a default configuration, shouldn't one simply all the constructor, followed by `.store()`?

I find this intermediate step confusing, where the configuration file is created but is empty.

sphuber: This design followed more or less directly from your other request (with which I agree) to not have warning messages being printed in the `Config` constructor. So the detection of whether the configuration has to be migrated as well as the actual migration (including the warning messages and backup generation) has to be done elsewhere. I think putting this in the `from_file` class method makes a lot of sense. Then the question was just how to deal with non-existing files. I think it is fine to put this here as well. This way one can always just call `from_file` also in all the unittest utils that don't have to worry about creating the file first or creating an initial default profile

ltalirz:
> I think I found the problem and it is exactly because from_file delegates the storing of the file in case it is new to the caller. That's why load_config always called config.store even if it was just loaded from file and nothing changed. It seems like it is better to put the store-if-created logic in the from_file method and then we can remove it from load_config. This should prevent store from being called all the time when running the tests. Would you agree?

I would certainly agree that `load_config` should not result in storing of the config file, if possible.

I would say that `.from_file` should actually also not result in storing of the config file if we can avoid it (see comment above).

sphuber:
> I would say that `.from_file` should actually also not result in storing of the config file if we can avoid it (see comment above).

That _only_ happens though when the specified `filepath` did not yet exist. In that case I think it is perfectly reasonable to create the default config, store it to disk and return the constructed `Config` object, no?

aiida/manage/configuration/config.py
ltalirz: > Then the question was just how to deal with non-existing files. I think it is fine to put this here as well. This way one can always just call from_file also in all the unittest utils that don't have to worry about creating the file first or creating an initial default profile

I agree that it is a little bit less code to write, and so I'm not totally against it (but I would still say that this stretches the meaning of what a`.from_file` method is supposed to do).
Let's keep it for the moment - in that case, the question still remains whether this intermediate stage is necessary, where the configuration file is simply "touched" but without the actual content. 

If the configuration file does not exist, can we not simply write the default content to it?

sphuber: No you are right, the indirect is not necessary. I will make that final simplification.

sphuber:
> Thanks for the patience @sphuber ;-)
> Looks good to me now!

Not at all, code ended up way better as a result of your review. Seems the system works :) thanks a lot

aiida/manage/configuration/config.py
```diff
@@ -39,24 +38,30 @@ class Config(object):  # pylint: disable=too-many-public-methods
...
-            echo.echo_warning('original backed up to `{}`'.format(filepath_backup))
+        try:
+            with io.open(filepath, 'r', encoding='utf8') as handle:
+                config = json.load(handle)
+        except (IOError, OSError):
+            config = Config(filepath, check_and_migrate_config({}))
+            config.store()
+        else:
+            # If the configuration file needs to be migrated first create a specific backup so it can easily be reverted
+            if config_needs_migrating(config):
```
ltalirz: sorry, one last thing I noticed: if the config file needs migrating, will this now result in two backups?
Won't the `.store` method take care of this backup automatically?

ltalirz: ah no, because there is no storing going on here...
one could potentially replace the call to `_backup` with a `.store` though..

sphuber: We could have, but for the reporting it is nice to write the filename of the created backup, which is only returned by `_backup`. So I think it is fine to keep it like this

PR #3586, opened 2019-11-27
support 'UNLIMITED' and 'NOT_SET' slurm times

### LINKED ISSUE #3415: Parse UNLIMITED and NOT_SET time in SLURM scheduler

ltalirz:
Fixes #3415 

requested walltimes in slurm can be 'UNLIMITED' and 'NOT_SET'.
now supporting these in the slurm plugin (& added tests).

### DIFF
--- a/aiida/schedulers/plugins/test_slurm.py
+++ b/aiida/schedulers/plugins/test_slurm.py
@@ -7,7 +7,8 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-
+"""Tests for the SLURM scheduler plugin.
+"""
 from __future__ import division
 from __future__ import print_function
 from __future__ import absolute_import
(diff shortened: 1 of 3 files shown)


aiida/schedulers/plugins/test_slurm.py
```diff
@@ -46,43 +54,46 @@ def test_parse_common_joblist_output(self):
 
         job_list = scheduler._parse_joblist_output(retval, stdout, stderr)
 
+        def get_job(id):
```
sphuber: `id` is a reserved keyword but I guess this file is in pre-commit exclude. Maybe `job_id`is better

aiida/schedulers/plugins/test_slurm.py
```diff
@@ -46,43 +54,46 @@ def test_parse_common_joblist_output(self):
...
+            try:
```
sphuber: Probably be a bit easier and more efficient to just turn `job_list` into a mapping on `job_id` and then index that in the tests instead of doing `get_job`

ltalirz: done

greschd:
Looks good to me, too. Thanks for implementing this!

PR #3587, opened 2019-11-28
print path to config file in verdi status

ltalirz:
From time to time, the need arises to have a look inside the currently
active AiiDA configuration file.
While the path to the file repository is printed in the output of `verdi
status` and `verdi profile show`, one needs to edit this path to get to
the config file (and, in principle, the repository can even be located
comepletely elsewhere).

Printing the location of the config file is helpful for daily use, also
as a reminder which configuration file is currently active.


sphuber:
I am not sure what you mean with "active" configuration file.
For each AiiDA instance there is only one configuration file which is always in the same place, namely the `.aiida` folder and its location is already printed using `verdi profile list`. In principle a user should never have to look in the configuration file manually. If they do, then there is probably a problem that should be fixed. I am not saying we should fully hide it (which is why it is already in `verdi profile list`) but I am not sure we should start adding all sorts of places where this info is printed.

ltalirz:
> I am not sure what you mean with "active" configuration file.
For each AiiDA instance there is only one configuration file which is always in the same place, namely the .aiida folder and its location is already printed using verdi profile list

I meant "active configuration file" as in "active configuration folder".
I think printing which configuration folder (or configuration file) is active is quite important, since it can easily lead to unintended consequences - e.g. you might not realize that `AIIDA_PATH` is set.
It is good that `verdi profile list` prints it, but I would suggest that  `verdi status` and probably `verdi profile show` should display it as well.

> In principle a user should never have to look in the configuration file manually. If they do, then there is probably a problem that should be fixed.

I would say in practice we are still quite some way from this goal - even with new AiiDA instances for newbies I often find myself looking at the configuration file, but I agree with the sentiment of directing people more towards `verdi config`. 
Still, I would say, printing the path to the configuration folder is meaningful.

Do you agree?

ramirezfranciscof:
@sphuber are you still unconvinced of this? If so, we can add it for discussion in the next aiida meeting. Else if @ltalirz rebases it I can review and merge it.

sphuber:
I'd be fine with adding it to `verdi status`, but maybe we don't have to also add it to `verdi profile show`.

ramirezfranciscof:
This path seems to be the same for all profiles, right? That being the case I agree that it doesn't need to be on the `verdi profile show` and `verdi status` would suffice. Do you agree @ltalirz ? If so, can you make the changes an update the PR? I could then review it.

PR #3590, opened 2019-11-29
Set default for the `safe_interval` option of `verdi computer configure`

### LINKED ISSUE #2880: connection cooldown time for local computers?

sphuber:
Fixes #2880 

The `safe_interval` is a common option for all `Transport` types but its
default value is class dependent. Since it is a common option it was
defined in the `Transport` base class, but since it is defined as a
class attribute, the default cannot be set yet. Keeping this design, the
only option to insert the class specific default is in `auth_options`
the class property that returns the options for the CLI. By adding the
default here, `verdi computer configure local localhost -n` will now
work without prompting for its only option `safe_interval` as it will
simply use the default.

### DIFF
--- a/aiida/transports/cli.py
+++ b/aiida/transports/cli.py
@@ -111,10 +111,14 @@ def create_option(name, spec):
     if spec.pop('switch', False):
         option_name = '--{name}/--no-{name}'.format(name=name_dashed)
     kwargs = {}
-    if 'default' not in spec:
+
+    if 'default' in spec:
+        kwargs['show_default'] = True
+    else:
         kwargs['contextual_default'] = interactive_default(
             'ssh', name, also_noninteractive=spec.pop('non_interactive_default', False)
         )
+
     kwargs['cls'] = InteractiveOption
     kwargs.update(spec)
     if existing_option:
--- a/aiida/transports/transport.py
+++ b/aiida/transports/transport.py
@@ -49,6 +49,8 @@ class Transport(object):
     """
     # pylint: disable=too-many-public-methods
 
+    _DEFAULT_SAFE_OPEN_INTERVAL = DEFAULT_TRANSPORT_INTERVAL
+
     # To be defined in the subclass
     # See the ssh or local plugin to see the format
     _valid_auth_params = None
@@ -68,12 +70,11 @@ class Transport(object):
         __init__ method of the Transport base class.
         """
         from aiida.common import AIIDA_LOGGER
-
+        self._safe_open_interval = kwargs.pop('safe_interval', self._DEFAULT_SAFE_OPEN_INTERVAL)
         self._logger = AIIDA_LOGGER.getChild('transport').getChild(self.__class__.__name__)
         self._logger_extra = None
         self._is_open = False
         self._enters = 0
-        self._safe_open_interval = DEFAULT_TRANSPORT_INTERVAL
 
     def __enter__(self):
         """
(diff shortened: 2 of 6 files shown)


aiida/transports/transport.py
```diff
@@ -186,6 +187,16 @@ def get_valid_auth_params(cls):
...
     @classproperty
     def auth_options(cls):  # pylint: disable=no-self-argument
+        """Return the authentication options to be used for building the CLI.
+
+        :return: `OrderedDict` of tuples, with first element option name and second dictionary of kwargs
+        """
+        # The common auth options are currently defined as class members, but the default for `safe_interval` is sub
+        # class specific. With the current design the default cannot already be specified directly but has to be added
+        # manually here.
+        for option in cls._common_auth_options:
```
ltalirz: hm, nasty...
how about making ` _common_auth_options` and `_valid_auth_options` `OrderedDict`s themselves. that would make accessing them easier

sphuber: That will make it impossible to set the correct default though, see the comment.

aiida/transports/cli.py
```diff
@@ -111,10 +111,15 @@ def create_option(name, spec):
     if spec.pop('switch', False):
         option_name = '--{name}/--no-{name}'.format(name=name_dashed)
     kwargs = {}
+
     if 'default' not in spec:
         kwargs['contextual_default'] = interactive_default(
             'ssh', name, also_noninteractive=spec.pop('non_interactive_default', False)
         )
+
+    if 'default' in spec:
```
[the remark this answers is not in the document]
sphuber: Done

sphuber:
> man, this is nasty code to fix...
> 
> Given that the `..._options_...` lists already contain click-specific code, it seems to me that rather than trying to synthesize the click commands from some custom format, the much more

That will remove the functionality that the `verdi computer configure` commands are created automatically for all registered transports. Once you remove this, people will have to define these commands manually and explicitly.

aiida/transports/transport.py
ltalirz: hm... I don't quite follow. How does `_common_auth_options` being an `OrderedDict` instead of a list of tuples change this?
Anyhow it would just be a minor improvement, not important.

ltalirz:
> That will remove the functionality that the verdi computer configure commands are created automatically for all registered transports. Once you remove this, people will have to define these commands manually and explicitly.

Right - it would mean you need to write the click command and the transport class (that inspects it) together, which might be less effort than writing just the transport class with the current machinery.

Anyhow, it would also not be idea.
For me this PR is good to merge!

PR #3591, opened 2019-11-29
Fix false positive for `verdi database integrity detect-invalid-links`

### LINKED ISSUE #3539: Database integrity issue
Recently, I had an issue resulted from power shortage and after restarting the `rabbitmq-server` and `postgres`, out of curiosity I checked the integrity of my database using `verdi database integrity detect-invalid-links` where I got a huge log which looks like:
```
Warning: detected calculation nodes with outgoing `call` links:
    ID  Input node                            Output node                           Type       Label
------  ------------------------------------  ------------------------------------  ---------  -------
    43  3eae25c2-8212-4192-aeed-7508831ba771  f79320bd-7c16-484e-b433-6ec0a6e62129  call_work  CALL
   152  5eaff866-0543-4967-b2c0-c3c2f1590bf7  1b6c7a90-1161-4de4-b9f5-a11fdab774dd  call_work  CALL
   486  7ce0002d-ed44-4bcb-bf4a-a1d5bda52c97  fcaacf07-21bd-4995-8bd2-7ce8ace7466e  call_work  CALL
   590  df0ce225-343a-4713-b521-c13794871043  fb1d1487-e784-4cb3-bc61-bc5c0694638e  call_work  CALL
  1384  ae00c8b1-5488-4986-8662-5f9eee7e902d  370c6661-6fdd-4825-84fa-cce34e3f8d00  call_work  CALL
.
.
.
Critical: one or more integrity violations detected
```
We (@yakutovicha and I) checked few things and so far realized that in the aforementioned log,  `Input node` is a parent workchain and `Output node` is the called `BaseWorkChain` within the parent. For instance, `Cp2kMultiStageWorkChain` and `Cp2kBaseWorkChain`. The same behavior is observed for our other workchains where the highest level work chain calls the `BaseRestartWorkChain`. 
We also have tested direct submission of `BaseRestartWorkChain` to the daemon and we have not observed similar integrity violation report. 
At this point, I am not sure if:
1. It is an issue with the workchain design combined with a bug in `aiida-core` which cannot catch it at the spot. 
2. It is a bug in `aiida-core` that wrongly flags these links as the violation of database integrity. 

Therefore, I thought to open an issue for further discussion.

sphuber:
Fixes #3539 

The query `SELECT_CALCULATIONS_WITH_OUTGOING_CALL` defined in the module
`aiida.manage.database.integrity.sql.links` contained an error causing
the command to return false positives. The problem was not detected by
the tests because they operated on an empty database and only checked
for true positives. A basic graph is added to now detect the false ones.

### DIFF
--- a/aiida/backends/tests/cmdline/commands/test_database.py
+++ b/aiida/backends/tests/cmdline/commands/test_database.py
@@ -26,6 +26,31 @@ from aiida.orm import Data, Node, CalculationNode, WorkflowNode
 class TestVerdiDatabasaIntegrity(AiidaTestCase):
     """Tests for `verdi database integrity`."""
 
+    @classmethod
+    def setUpClass(cls, *args, **kwargs):
+        """Create a basic valid graph that should help detect false positives."""
+        super(TestVerdiDatabasaIntegrity, cls).setUpClass(*args, **kwargs)
+        data_input = Data().store()
+        data_output = Data().store()
+        calculation = CalculationNode()
+        workflow_parent = WorkflowNode()
+        workflow_child = WorkflowNode()
+
+        workflow_parent.add_incoming(data_input, link_label='input', link_type=LinkType.INPUT_WORK)
+        workflow_parent.store()
+
+        workflow_child.add_incoming(data_input, link_label='input', link_type=LinkType.INPUT_WORK)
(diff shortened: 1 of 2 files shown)


sphuber:
> thanks @sphuber !
> So, if I understand correctly, the old integrity check was flagging any kind of node with a `call_work` output link?
> That might be good to add to the commit message

Done!

PR #3593, opened 2019-11-30
fix gotocomputer when key_filename missing

### LINKED ISSUE #3401: `verdi calcjob gotocomputer` will except if corresponding `AuthInfo` does not exist
Discovered by @csadorf and first reported [here](https://github.com/aiidateam/aiida-tutorials/issues/186)

### LINKED ISSUE #3588: verdi calcjob gotocomputer fails, if key filename empty
When using the cli options of `verdi computer configure ssh`, it is possible to leave the `key-filename` empty (with `look-for-keys` True).

E.g. 
```
verdi computer configure ssh fidis --username talirs --safe-interval 10 --look-for-keys --key-policy AutoAddPolicy --non-interactive
```

This works, and `verdi computer test fidis` finds that all 4 tests complete successfully.

However, as discovered by @danieleongari, when one runs calculations on fidis, `verdi calcjob gotocomputer` fails to go to the correct directory:
```
Warning: Identity file if [ -d '/scratch/ongari/aiida_run/98/68/7475-f879-42b6-8cd3-88db6d3e771a' ] ; then cd '/scratch/ongari/aiida_run/98/68/7475-f879-42b6-8cd3-88db6d3e771a' ; bash -l ; else echo '  ** The directory' ; echo '  ** /scratch/ongari/aiida_run/98/68/7475-f879-42b6-8cd3-88db6d3e771a' ; echo '  ** seems to have been deleted, I logout...' ; fi not accessible: No such file or directory.
###########################################################
# Welcome to FIDIS: https://scitas.epfl.ch/hardware/fidis #
###########################################################
# System announcements:                                   #
#                                                         #
# Jul-2019: New Operating System and Slurm versions!      #
#           Before submitting jobs read about the changes #
#           at: https://go.epfl.ch/scitas-maint-2019      #
#                                                         #
# Nov-2019: New Slurm version (bugfixes)                  #
#                                                         #
###########################################################
# Software announcements:                                 #
#                                                         #
# Jul-2019: NEW software release deployed, with new       #
#           versions of packages. Recompile your code!    #
#                                                         #
# 14-Aug-2019:                                            #
#         - gromacs rebuilt with CPU optimizations        #
#         - new intel compiled packages: R, foam-extend   #
#         gromacs/2019.2, neuron, abaqus                  #
#                                                         #
###########################################################
# Live usage report:    https://scitas-data.epfl.ch/live  #
# Maintenance schedule: https://go.epfl.ch/scitas-down    #
# Usage statistics:     https://go.epfl.ch/scitas-stats   #
# Need Help? Questions? Comments?     mailto:1234@epfl.ch #
###########################################################
# Last SCITAS MOTD update - Tue Nov 26 16:25:03 CET 2019 #
Last login: Thu Nov 28 15:43:37 2019 from i17dhcp-1-232.epfl.ch
[ongari@fidis ~]$ cd /scratch/ongari/aiida_run/98/68/7475-f879-42b6-8cd3-88db6d3e771a
[ongari@fidis 7475-f879-42b6-8cd3-88db6d3e771a]$ ls
```

For some reason, the bash command to be executed is misinterpreted as the "identity file".

ltalirz:
fix #3588 
fix #3401 

* the "verdi calcjob gotocomputer" command was broken when used on computers that were configured without a value for the ssh key_filename.
* "verdi calcjob gotocomputer" obviously does not work when no authinfo is
present for the computer to go to. There was a bug in printing the exception, resulting in an exception bubbling up to user space (instead of a critical error with a helpful message).


ltalirz:
I noticed that there does not seem to be any test of the `gotocomputer` command.
Perhaps not that easy to test but I guess also not impossible. Should we open an issue?

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -41,7 +41,7 @@ def calcjob_gotocomputer(calcjob):
     try:
         transport = calcjob.get_transport()
     except NotExistent as exception:
-        echo.echo_critical(exception)
+        echo.echo_critical(repr(exception))
```
giovannipizzi: Why `repr` and not `str`?

giovannipizzi:
It would be great to have a test for this, but I'm not sure how... One should have a configured computer and a calculation run in there (one could use the `test_daemon.py` file that actually runs simulations, but I'm not sure this is over SSH - actually maybe @sphuber is changing to local in #3571?). But then one needs to test if an interactive shell is opened... Not sure how to do it exactly.

Maybe one could instead just get internally the command that is generated `ssh ...` and append something execute (e.g. create an folder with a given file in `/tmp/XXX`, then append `ls /tmp/XXX` to the ssh command, and check if the result shows correctly the folder content? Even if I'm not sure how robust this test would be).

aiida/cmdline/commands/cmd_calcjob.py
ltalirz: It just includes information about the error as well:
```
In [1]: try:
   ...:     raise ValueError("abc")
   ...: except ValueError as er:
   ...:     print(str(er))
   ...:
abc

In [2]: try:
   ...:     raise ValueError("abc")
   ...: except ValueError as er:
   ...:     print(repr(er))
   ...:
   ...:
ValueError('abc')
```

giovannipizzi: Ok - I'm not sure this is the best way to achieve this (I would have printed `str` and the type of the exception) but I think it serves the purpose here, for now

PR #3597, opened 2019-12-03
Cleanup dependencies

### LINKED ISSUE #2903: CodDbImporter throws warning about pybtex when importing CIF file
Example:
```python
from aiida.tools.dbimporters.plugins.cod import CodDbImporter
importer = CodDbImporter()
structure_Al = importer.query(id='9008460')[0].get_aiida_structure()
```
Outputs:
```
/home/aiidatests/venvs/aiida_py3.6/lib/python3.6/site-packages/pymatgen/io/cif.py:44: UserWarning: Please install optional dependency pybtex if youwant to extract references from CIF files.
  warnings.warn("Please install optional dependency pybtex if you"
```

### LINKED ISSUE #3528: REST API dependencies: are all of them needed?
There are quite a few dependencies in the `rest` extra. I'm not sure I understand, however, if they are all used and how. Maybe some of them are just "plugins" for flask and just need to be there, but could you:
- check if there are any dependencies that are not really needed, and remove them
- report here which one are used, how (i.e. if they are imported or used implicitly) and, especially in the latter case, what are they needed for?

Just to make two examples, if I do `git grep marshmallow` I don't get any hit in python files, but there are the dependencies `flask-marshmallow==0.10.1` and `marshmallow-sqlalchemy==0.19.0` - are they used?

### LINKED ISSUE #3547: Lots of dependencies are pinned to an exact (patch) version

sphuber:
Fixes #3547 and fixes #2903 and fixes #3528 

This PR releases the strict pinning of dependency versions.
In addition, it gets rid of some unnecessary dependencies:

 * `mock`
 * `passlib`
 * `uritools`


setup.json
```diff
@@ -19,105 +19,87 @@
...
-      "spglib==1.14.1.post0"
+      "PyCifRW~=4.4",
+      "ase~=3.18",
+      "pymatgen>=2019.7.2",
+      "pymysql~=0.9.3",
+      "seekpath~=1.9",
+      "spglib~=1.14"
     ],
     "notebook": [
       "jupyter==1.0.0",
```
giovannipizzi: I think these requirements were to avoid collisions with py2 and maybe now we can just put `jupyter` and `notebook`, without version specification?

setup.json
```diff
@@ -19,105 +19,87 @@
...
-      "toml==0.10.0",
-      "yapf==0.28.0"
+      "pre-commit~=1.19",
+      "prospector~=1.1",
+      "toml~=0.10.0",
+      "yapf~=0.28.0"
     ],
     "bpython": [
-      "bpython==0.17.1"
+      "bpython~=0.18.0"
```
giovannipizzi: Do we want to drop this? Is any user really using this? I guess it's more for convenience of users, but then maybe it's ok to remove?

setup.json
sphuber: No unfortunately we can't. It's true that `notebook>=6` does not support python 2, but it also requires `pyzmq>=17` and `tornado>=5` and we, because of `circus` (and `plumpy` and `kiwipy`) have exactly `pyzmq<17` and `tornado<5`. I had made a start into seeing whether I could fix circus and make it compatible with the newer versions and made some headway but it is not trivial. Hopefully I can look into this more with @muhrin during the coding week

setup.json
sphuber: I am not sure if anyone uses it, so that's why I left it in for now.

sphuber:
@asle85 and @waychal I have removed quite a few dependencies in the `rest` extras. The tests are running so apparently they weren't used internally, at least not in the tested paths. Can you please check that this makes sense?

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
-ete3==3.1.1
-flask-cache==0.13.1
-flask-cors==3.0.8
-flask-httpauth==3.3.0
-flask-marshmallow==0.10.1
-flask-restful==0.3.7
-flask-sqlalchemy==2.4.1
-flask==1.1.1
-graphviz==0.13
+ete3~=3.1
```
ltalirz:
```suggestion
ete3~=3.1.1
```

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
+flask-cors~=3.0
```
ltalirz:
```suggestion
flask-cors~=3.0.8
```

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
-sphinx-rtd-theme==0.4.3
-sphinx==2.2.0
-sphinxcontrib-contentui==0.2.4
-sphinxcontrib-details-directive==0.1.0
-sqlalchemy-diff==0.1.3
-sqlalchemy-migrate==0.12.0
-sqlalchemy-utils==0.34.2
-sqlalchemy==1.3.10
-tabulate==0.8.5
+jinja2~=2.10
```
ltalirz:
```suggestion
jinja2~=2.10.3
```

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
+kiwipy[rmq]~=0.5.1
+numpy~=1.17
+paramiko~=2.6
+pg8000~=1.13
+pgtest~=1.3
```
ltalirz:
```suggestion
pgtest~=1.3.1
```

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
+pika~=1.1
+plumpy~=0.14.3
+psutil~=5.6
```
ltalirz:
```suggestion
psutil~=5.6.3
```

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
+psycopg2-binary~=2.8
```
ltalirz:
```suggestion
psycopg2-binary~=2.8.3
```

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
+pyblake2~=1.1; python_version<'3.6'
```
ltalirz:
```suggestion
pyblake2~=1.1.2; python_version<'3.6'
```

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
+pygments~=2.5
+pymatgen>=2019.7.2
+pymysql~=0.9.3
+pyparsing~=2.4
```
ltalirz:
```suggestion
pyparsing~=2.4.2
```

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
+pytest~=5.3
+python-dateutil~=2.8
+python-memcached~=1.59
+pytz~=2019.3
+pyyaml~=3.13
+reentry~=1.3
+seekpath~=1.9
```
ltalirz:
```suggestion
seekpath~=1.9.3
```

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
+simplejson~=3.16
+spglib~=1.14
+sphinx-rtd-theme~=0.4.3
+sphinxcontrib-contentui~=0.2.4
+sphinxcontrib-details-directive~=0.1.0
+sphinx~=2.2
+sqlalchemy-diff~=0.1.3
+sqlalchemy-utils~=0.34.2
+sqlalchemy~=1.3
```
ltalirz:
```suggestion
sqlalchemy~=1.3.10
```

docs/requirements_for_rtd.txt
```diff
@@ -1,69 +1,54 @@
...
+tabulate~=0.8.5
 tornado<5.0
-tzlocal==2.0.0
-upf_to_json==0.9.2
-uritools==2.2.0
-wrapt==1.11.2
\ No newline at end of file
+tzlocal~=2.0
+upf_to_json~=0.9.2
+wrapt~=1.11
```
ltalirz:
```suggestion
wrapt~=1.11.2
```

setup.json
```diff
@@ -19,105 +19,87 @@
...
     "dev_precommit": [
-      "astroid==2.2.5",
-      "pep8-naming==0.4.1",
-      "pre-commit==1.18.3",
-      "prospector==1.1.7",
-      "pylint==2.3.1",
-      "toml==0.10.0",
-      "yapf==0.28.0"
+      "pre-commit~=1.19",
+      "prospector~=1.1",
```
ltalirz:
```suggestion
      "prospector~=1.1.7",
```

ltalirz:
install on macos seems fine

csadorf:
Hi, I have not reviewed all changes to the dependencies in detail yet, but I have had prepared a [draft for a related AEP](https://github.com/aiidateam/AEP/pull/10) that I just had not submitted yet. Maybe we can consider this in the context of this PR?

sphuber:
@ltalirz thanks for the review. The changes you proposed are not necessary. The changes I made will not result in adowngrade. The compatible operator `~=` will install the latest version compatible with the spec. For example, for `ete3` the latest is version `3.1.1`. When doing `pip install ete3~=3.1` it will install `3.1.1`. This is the whole reason I adopted this specifier and for dependencies with a major release `>=1` we will only ever specify the minor version. In this way the latest minor version will always be installed. See the explanation in the first commit for more details.

ltalirz:
> @ltalirz thanks for the review. The changes you proposed are not necessary. The changes I made will not result in adowngrade. The compatible operator ~= will install the latest version compatible with the spec.

Sure, but if you install AiIDA in an environment that already has an older version of the package installed, then it will not upgrade.
And we have not tested AiiDA with the older version of the package

sphuber:
I see, but by specifying the patch version, we strictly reduce the accepted range of versions to that specific minor version, a limitation the removal of which was the whole point of this PR. So there is a direct conflict between those two requirements. We could just tell people in the installation documentation that when installing in an existing environment, the should run `pip` with the `--upgrade` flag. This will do the right thing. And for new environments there is no problem.

asle85:
Both I and @waychal have checked the dependencies removed in the rest extras and it looks fine to us. Given that the tests are passing it's ok to remove them.

ltalirz:
> I see, but by specifying the patch version, we strictly reduce the accepted range of versions to that specific minor version, a limitation the removal of which was the whole point of this PR. 

That is not how the `~=` operator works:
```
$ pip install wrapt~=1.11.1
Collecting wrapt~=1.11.1
aiida-core 1.0.1 requires reentry~=1.3, which is not installed.
Installing collected packages: wrapt
  Found existing installation: wrapt 1.10.11
    Uninstalling wrapt-1.10.11:
      Successfully uninstalled wrapt-1.10.11
Successfully installed wrapt-1.11.2
```

sphuber:
> That is not how the `~=` operator works:

I am afraid it does. Consider the following two examples:
```
(aiida_test) sphuber@theos:~/code/aiida/env/dev$ pip install numpy~=1.16.1 --no-cache-dir
Collecting numpy~=1.16.1
  Downloading https://files.pythonhosted.org/packages/98/87/41283370f942f647422581eed16df4b653a744a3e9d5cfbb9aee0440f6eb/numpy-1.16.5-cp36-cp36m-manylinux1_x86_64.whl (17.4MB)
     |████████████████████████████████| 17.4MB 3.3MB/s 
Installing collected packages: numpy
Successfully installed numpy-1.16.5
```
versus
```
Collecting numpy~=1.16
  Downloading https://files.pythonhosted.org/packages/d2/ab/43e678759326f728de861edbef34b8e2ad1b1490505f20e0d1f0716c3bf4/numpy-1.17.4-cp36-cp36m-manylinux1_x86_64.whl (20.0MB)
     |████████████████████████████████| 20.0MB 3.5MB/s 
Installing collected packages: numpy
Successfully installed numpy-1.17.4
```

When you specify a patch version, it will always remain within the corresponding minor version, as shown by the first example. Doing `numpy~=1.16.1` will install `numpy==1.16.5` as that is the last patch of the `1.16` minor series. Doing `numpy~=1.16` (without explicit patch version) instead will install `numpy==1.17.4`. It is the latter that we want, which is why I removed the explicit patch versions for dependencies with a major version `>=1`.

ltalirz:
You're right - and apologies for not reading your commit message/comments properly (I didn't realize we are only fixing the major version).

I think we should retain the ability to force upgrades of the patch level of certain packages, if we know that they contain important bug fixes (such as pgtest 1.3.1).
Otherwise, at some point we will have cases where users will upgrade their AiiDA instance and be stuck with an old, broken dependency.

`pip install --upgrade` would indeed solve the problem... but will people remember it?
And I think it will upgrade all dependencies of dependencies as well, which may be something the user does not want.

How about something like `pgtest~=1.3,>=1.3.1`? That would work, no?

sphuber:
> `pip install --upgrade` would indeed solve the problem... but will people remember it?
> And I think it will upgrade all dependencies of dependencies as well, which may be something the user does not want.

Probably not in some cases, but that is the case with many things. I agree that we should try to make things as robust as possible but this is not always possible in the case of competing and directly conflicting goals.

> How about something like `pgtest~=1.3,>=1.3.1`? That would work, no?

I think this does indeed work and would be a possible solution and certainly makes a lot of sense for dependencies where we no a particular patch version contains important bug fixes. I don't think we should blindly apply this to all our current versions though. For the majority of dependencies, the specific version we have currently were just picked because they were the latest version that was available the last time I updated all the dependencies en masse. Do you know of any other dependency that might have a specifically required patch version?

ltalirz:
> Do you know of any other dependency that might have a specifically required patch version?

pgtest is the only one I know for sure - as a "gut feeling" I would also add `sqlalchemy` (very large patch version), `seekpath` (probably want the latest version here),  `prospector` (also here we want the latest version), `psycopg2-binary` (not sure, would just be bad if we have connection issues because of some old version)

setup.json
```diff
@@ -19,105 +19,87 @@
...
-      "pgtest==1.3.1",
-      "pytest==4.6.6",
-      "pytest-cov==2.8.1",
-      "sqlalchemy-diff==0.1.3"
+      "pg8000~=1.13",
+      "pgtest~=1.3",
+      "pytest~=5.3",
+      "sqlalchemy-diff~=0.1.3"
     ],
     "dev_precommit": [
```
greschd: I think for the pre-commit dependencies it actually makes sense to pin them exactly. The issue with these is that even a "backwards compatible" change in behavior can cause the build to fail. 

For example, when ``pylint`` has a false positive, we add a ``# pylint: disable=...`` comment. If a subsequent patch fixes the false positive, it will trigger ``useless-suppression`` errors.

greschd: Also, they are presumably used when developing ``aiida-core`` - so there really isn't much of a concern for being compatible with other environments / packages.

sphuber: We have the `useless-suppression` warning disabled globally, but I see your point. This can also just happen with `pre-commit` and `yapf` that introduce new formatting rules. However, if that happens, since these are only for developers anyway, the dev just upgrades his dependencies in his development environment and commits the changes no?

ltalirz: Just imagine your PR fails because in just that moment some new yapf/pylint version has been released...
I would prefer to avoid this ;-)

sphuber:
@ltalirz I have added explicit lower required patch versions for `psycopg-binary`, `seekpath`, `sqlalchemy` and `pgtest`. Only question then remains what we do with the pre-commit extras. Do we pin those to the version that we had for now?

ltalirz:
>  Do we pin those to the version that we had for now?

Yes, I think @greschd has a good point here.

setup.json
greschd: Exactly, it can be quite an annoying thing. Also, if we have scheduled CI on older branches they will just suddenly fail with pre-commit errors, for no reason.

sphuber:
@greschd and @ltalirz ok I pinned the packages of `dev_precommit` to an explicit version, with the exception of `toml`. Since that one is not used for actual linting or formatting, I think there we still automatically want updates in patch versions.

sphuber:
Seems like a little bit more patience is required :sweat_smile: . The `pre-commit` step now fails because for some reason `astroid=2.3.3` is installed but `prospector==1.1.7` requires `2.2.5`. Maybe because I relaxed the requirements before it installed `2.3.3` and now it is using the cache?

ltalirz:
That's very strange... With your previous version of the branch (all dependencies with `~=`) that I have locally, it is installing astroid 2.2.5.

The only main difference I can see is that you were using pre-commit 1.19.

Perhaps 1.17 had some astroid dependency that pip is somehow not properly resolving?

sphuber:
> Perhaps 1.17 had some astroid dependency that pip is somehow not properly resolving?

I know that `prospector` demands an explicit version, see their [`setup.py`](https://github.com/PyCQA/prospector/blob/1.1.7/setup.py). What I don't understand who required `astroid==2.3.3` and why prospectors requirement does not get respected during setup.

sphuber:
Seems like it was a problem with the cache. I added a commit to run `pip install` without cache and now it properly installed `astroid==2.2.5`.

PR #3599, opened 2019-12-03
Import/export progress bar

### LINKED ISSUE #2776: progress indicator for verdi export/import
Since exports can take a long time, having a kind of progress bar (or at least an idea of the size of hte export) would be very valuable.

CasperWA:
Fixes #2776

This PR does a multitude of things/is a bit complex. Therefore, I have tried to keep each "thing" kept in its own commit. Upon merging, these may be squashed in a way that makes sense.
I will try to keep this commit segregation throughout the lifetime of the PR, meaning if I commit changes, they will be squashed into the respective commit, where they belong, or a new commit will be made, following a complete rebase. _EDIT_: I failed to do this, but at least the history is more clear now.

Some of the things introduced in this PR:

- ~`debug` parameter for import/export.~
  This is now logged at the debug level instead.
- Meta data ~header~ summary for import/export.
- Minor optimizations of import/export. This is done for the `verdi` commands, but also for the backend functions.

But most importantly, this PR creates a progress bar for import, export, and extracting export archives.

Possible updates (either in this PR or in the future):
- [ ] Progress bar for export archive migrations.
- [X] Minimize waiting time "between" progress bars (to the degree it is possible in the scope of this PR).


CasperWA:
The main thing I am not convinced about with this PR, is the design of the headers of meta data printed when im- and exporting. Feedback on this would be great, contributing commits or PR's against my branch are even more welcome.

CasperWA:
> The main thing I am not convinced about with this PR, is the design of the headers of meta data printed when im- and exporting. Feedback on this would be great, contributing commits or PR's against my branch are even more welcome.

This has now been fixed and brought into line with the rest of AiiDA, using `tabulate`.

CasperWA:
@ltalirz @sphuber @dev-zero @yakutovicha @chrisjsewell I consider this ready for review. Since this is very much a UX addition, it will probably be needed for you to checkout the branch and test ex- and importing.

ltalirz:
So far tested it only on a small export - on first impression it looks great and seems to provide very useful information, well done!
I plan to test it on a larger export later this evening.
We should decide how to handle the code review as well.

I've also checked the new dependency [tqdm](https://github.com/tqdm/tqdm).
It's used by 50k other repos and has 70 contributors, but only the maintainer had >3 commits during the last year, i.e. it is de facto a 1-person project (named Casper... coincidence? ;-) ).

I would therefore ask @CasperWA to motivate which features of tqdm warrant the addition of a new package (vs the [various 50ish lines implementations](https://stackoverflow.com/questions/3160699/python-progress-bar) in pure python).
This explanation should be recorded in the eventual commit message (if we keep it).

CasperWA:
> I would therefore ask @CasperWA to motivate which features of tqdm warrant the addition of a new dependency (vs the [various 50ish lines implementations](https://stackoverflow.com/questions/3160699/python-progress-bar) in pure python).
> This explanation should be recorded in the eventual commit message of the PR (if we keep it).

The features I expect from the progress bar:

1. Easily implemented and manipulated to quickly show the exact information I want the user to see.
2. Have automated statistics for an estimated time of success.
3. Be non-invasive, i.e., it should be removable from stdout.

How `tqdm` does for these "requirements":

1. It is easily manipulated, but depending on the use-case, the implementation can turn out to be more or less extensive.
   - As an example here, compare the "simple" implementation for the `extract` functions [here](https://github.com/aiidateam/aiida-core/pull/3599/files#diff-ef9f06c30877d6829eedd1e26c294827R280), where the progress bar updates are handled by `tqdm`, with the more involved implementation in the general export function [here](https://github.com/aiidateam/aiida-core/pull/3599/files#diff-4ad07fbf155835880f2ae12c22a7989cR344), where the progress bar needs to be updated for each iteration.
2. The statistical estimations of `tqdm` are calculated on-the-run without a lot of overhead. The logic for these numerical calculations seem to be "cleverly" handled by `tqdm` based on the inputs given: The less `tqdm` is left to itself, the less accurate the estimates, as a rule-of-thumb. That does not, however, render the estimates useless, I think.
   To re-implement these calculations would probably take more than what is suggested by the solutions in the [stackoverflow thread](https://stackoverflow.com/questions/3160699/python-progress-bar), but they would make up a great start.
3. This should be the case for all progress bars that rely on `/r` and `/n`, I guess?

The justification for `tqdm` is not stone-solid. It is also not without merit. I could definitely use some time to elaborate on the suggested implementations from [the stackoverflow thread](https://stackoverflow.com/questions/3160699/python-progress-bar) referenced by @ltalirz and come up with a solution that meets my requirements. But in the end, I value my time more precious, and can therefore not deny that this has influenced my addition of `tqdm` here.
I hope either that the dependency will find uses in other areas of the base code or a similarly fast and reliable implementation may be found in existing dependencies. It should here be said that I have looked at the one from `click` and found it wanting.

ltalirz:
Some comments from a production test.

 * What is the "Initializing" phase? Would perhaps be good to make clear what the items are that you go through here
 * Second is the "Loading group" phase.
    Here, it seems to show one bar per group. In my case, however, we are dealing with tens of thousands of groups, and the subset of the label you are showing is always the same. I think it would be better to replace this progress bar with an overall progress over all groups.

The group loading phase seems extremely slow to me - perhaps this is a good point to look into optimizing the code.

CasperWA:
> Some comments from a production test.
> 
>     * What is the "Initializing" phase? Would perhaps be good to make clear what the items are that you go through here

As far as I remember, it is indeed just this: An initialization. Nothing quantifiable is truly happening, other than some memory-caching/-reserving, I believe?

>     * Second is the "Loading group" phase.
>       Here, it seems to show one bar per group. In my case, however, we are dealing with tens of thousands of groups, and the subset of the label you are showing is always the same. I think it would be better to replace this progress bar with an overall progress over all groups.

Right!
So the idea (as it is now) was to go through each group to-be-exported and show progress for each node in the group.
I guess the choice to either do this, or take a "step back" and go through all groups instead, is subject to personal preference as much as it is _what_ one is exporting?

In the beginning I played around with having multiple progress bars as well: An overarching one and then sub-progress bars for each "part". This may be the solution for this. What do you think?

> The group loading phase seems extremely slow to me - perhaps this is a good point to look into optimizing the code.

It indeed is. The code section looks like this:

```python
    # Add all the nodes contained within the specified groups
    for group in given_groups:
        if not silent:
            group_count = group.count()
            if group_count:
                progress_bar.reset(total=group_count)
                progress_bar.set_description_str('Loading Group - LABEL={}'.format(group.label))

        for entry in group.nodes:
            if not silent:
                progress_bar.update()
                progress_bar.refresh()

            entities_starting_set[NODE_ENTITY_NAME].add(entry.uuid)
            if issubclass(entry.__class__, orm.Data):
                given_data_entry_ids.add(entry.pk)
            elif issubclass(entry.__class__, orm.ProcessNode):
                given_calculation_entry_ids.add(entry.pk)
```

It may be that it is the `count()` that is slow or otherwise?
If you provide me with an export file I can use to test these cases, I may be able to pinpoint the issue and optimize a bit.

On another note, I must say that putting a progress bar on all of this truly exposes the slow parts of both the import and export functions - something that is quite neat for optimization purposes, but may be bothersome for users. Although I still think it is nicer than the current "print a bunch of stuff" approach :)

For future reference, these are comments concerning the export part only.

ltalirz:
I'm working on fixing this now.

ltalirz:
I've opened a PR for fixing making the search over groups more performant (for the moment on top of develop).

Furthermore, here is a branch that contains this fix + some further changes to the progress bar - in particular I have removed the "refresh" from a couple of places since refreshing is costly and already handled by tqdm automatically (it has an internal timeout set to 100ms).
Note that updating the description triggers a refresh by default; there might be a couple more places where this is not what you want.

https://github.com/ltalirz/aiida-core/tree/fix-progress-bar-rebased

CasperWA:
> I've opened a PR for fixing making the search over groups more performant (for the moment on top of develop).

Cheers. I see @lekah has already come up with some improvements.

> Furthermore, here is a branch that contains this fix + some further changes to the progress bar - in particular I have removed the "refresh" from a couple of places since constant refreshing is costly and already handled in a sensible way by tqdm (it has an internal timeout set to 100ms).

Right! I knew there was a reason why I chose `tqdm` (I actually also mentioned this reason above).

> Note that updating the description triggers a refresh by default; there might be a couple more places where this is not what you want and you might want to pass `refresh=False`.

I knew this, and I am also not doing this some places, but mostly I am, since I "forgot" about the `refresh()` cost.

I will take a look at it, thanks!

CasperWA:
@ltalirz I have incorporated some changes, especially concerning the `refresh` parameter for the `set_description_str` method for the progress bar.

Furthermore, I have gone through the export function `export_tree` again, after the updated graph traversal tool. Some places I have optimized the code accordingly and tried to remove some older concepts that are no longer necessary (like the many different `given_*_ids` sets. This also means most of the `QueryBuilder` filtering is done on the 'UUID` column instead of `PK`/`id`.

ltalirz:
As a first step, I'm now running the benchmark export I used for the AGE with current `develop` + this PR to get some timings.

CasperWA:
> As a first step, I'm now running the benchmark export I used for the AGE with current `develop` + this PR to get some timings.

Great!
I would however like to say that this PR has _nothing_ to do with optimizing the import/export - it only deals with adding a progress bar to the two - and then there are some code tweaks as well, but these are considered minor and only done because they stood out.
Since the whole import/export is going to be updated (most likely from the bottom up) over the coming year, I see no reason to continue optimizing on it - again, other than what stands out (and it "easily" optimized).

Though to that effect, I applaud the benchmark tests you are currently running, since they will act as a "before" to the import/export remake.

ltalirz:
> I would however like to say that this PR has nothing to do with optimizing the import/export

Sure, I just wanted to make sure the updates of the progress bar no longer slow down the export as they did in the first iteration of this PR.
As it turns out, however, on the machine where I wanted to test, almost all time is spent in the file repo part. I'll first need to have a quick look into what is going on there.

CasperWA:
Currently experiencing the issue mentioned by @csadorf [here](https://github.com/aiidateam/aiida-core/pull/3839#issuecomment-597097943). He has promised to look into it, possibly triggering the workflows via "bot" commands in PR comments...

csadorf:
> Currently experiencing the issue mentioned by @csadorf [here](https://github.com/aiidateam/aiida-core/pull/3839#issuecomment-597097943). He has promised to look into it, possibly triggering the workflows via "bot" commands in PR comments...

I'll to address this as soon as possible. In the meantime, if you would like to work-around this issue, I suggest to make an empty commit on this branch.

aiida/tools/importexport/dbexport/utils.py
```diff
@@ -271,3 +271,33 @@ def check_process_nodes_sealed(nodes):
...
+
+    include_comments = kwargs.get('include_comments', True)
+    include_logs = kwargs.get('include_logs', True)
+    input_forward = kwargs.get('input_forward', False)
+    create_reversed = kwargs.get('create_reversed', True)
+    return_reversed = kwargs.get('return_reversed', False)
+    call_reversed = kwargs.get('call_reversed', False)
+
+    inclusions = [['Include Comments', include_comments], ['Include Logs', include_logs]]
+    result += '\n\n{}'.format(tabulate(inclusions, headers=['Included', '']))
```
ltalirz, line 295:
```suggestion
    result += '\n\n{}'.format(tabulate(inclusions, headers=['Inclusion rules', '']))
```

ltalirz:
Comments while testing:

Very quickly got this
```
STARTING EXPORT...
Getting provenance and storing links ... 100.0%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▉| 66855/66856
```
but then no further output for a very long time

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -325,108 +338,128 @@ def export_tree(
...
         # Skip those nodes where the license is not set (this is the standard behavior with Django)
         node_licenses = list((a, b) for [a, b] in builder.all() if b is not None)
         check_licenses(node_licenses, allowed_licenses, forbidden_licenses)
 
     ############################################################
     ##### Start automatic recursive export data generation #####
     ############################################################
-    if not silent:
+    if debug:
         print('STORING DATABASE ENTRIES...')
```
ltalirz, line 354:
```suggestion
        print('GATHERING DATABASE ENTRIES...')
```

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -325,108 +338,128 @@ def export_tree(
...
-                sum(len(model_data) for model_data in export_data.values()), len(all_nodes_pk)
-            )
-        )
+        # Instantiate new progress bar
+        progress_bar = tqdm(total=1, bar_format=BAR_FORMAT, leave=True)
 
     # ATTRIBUTES and EXTRAS
-    if not silent:
+    if debug:
         print('STORING NODE ATTRIBUTES AND EXTRAS...')
```
ltalirz, line 416:
```suggestion
        print('GATHERING NODE ATTRIBUTES AND EXTRAS...')
```

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -325,108 +338,128 @@ def export_tree(
...
+
+        for node_pk, attributes, extras in all_nodes_query.iterall():
+            if not silent:
+                progress_bar.update()
+
+            node_attributes[str(node_pk)] = attributes
+            node_extras[str(node_pk)] = extras
+
+    if debug:
         print('STORING GROUP ELEMENTS...')
```
ltalirz, line 440:
```suggestion
        print('GATHERING GROUP ELEMENTS...')
```

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -439,12 +472,12 @@ def export_tree(
...
     ######################################
-    # Now I store
+    # Now collecting and storing
     ######################################
     # subfolder inside the export package
     nodesubfolder = folder.get_subfolder(NODES_EXPORT_SUBFOLDER, create=True, reset_limit=True)
 
-    if not silent:
+    if debug:
         print('STORING DATA...')
```
ltalirz, line 481:
```suggestion
        print('WRITING DATA TO DISK...')
```

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -479,18 +513,24 @@ def export_tree(
     with folder.open('metadata.json', 'w') as fhandle:
         fhandle.write(json.dumps(metadata))
 
-    if silent is not True:
+    if debug:
         print('STORING REPOSITORY FILES...')
```
ltalirz, line 517:
```suggestion
        print('ADDING REPOSITORY FILES TO ARCHIVE...')
```

ltalirz:
Ok, so here the results from timing the export of a DB with 1.4M nodes, 67k groups (for the moment excluding the copying of the file repository, which would have taken too long):
 * `develop`: 43 minutes
 * this branch: 37 minutes
I don't know what the uncertainty bar is on these timings so I don't know whether there is indeed a speed up , but I certainly have no more concerns about overhead introduced by the progress bar.

I did notice that there are a couple of times where the export hangs for a long time in between progress bars - I'll try to make a short movie for a subset of the DB.

ltalirz:
And here is the recording: https://asciinema.org/a/ToXULwqQBXZQ6hJOgx2Dzl0gb

I think this will help you identify remaining issues / things to improve

CasperWA:
> Comments while testing:
> 
> * Very quickly got this
> 
> ```
> STARTING EXPORT...
> Getting provenance and storing links ... 100.0%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▉| 66855/66856
> ```
> 
> but then no further output for a very long time

Yup - with the new addition of the AGE I don't have a say anymore in the specifics of this particular part. There used to be much more information here.
I can perhaps try and add it, but it would undoubtedly step on the toes of the AGE module... So I didn't think it wise.

> * Progress bar seems not to mix well with `debug=True` (if possible, make sure to print debug statements before progress bar, e.g. `STORING NODE ATTRIBUTES AND EXTRAS...`)

It's not really meant to, in my design. The debug is only for developers - as such, it doesn't need to mix well. But maybe it could be _a bit_ better in some areas?

CasperWA:
> Ok, so here the results from timing the export of a DB with 1.4M nodes, 67k groups (for the moment excluding the copying of the file repository, which would have taken too long):
> 
> * `develop`: 43 minutes
> * this branch: 37 minutes
> 
> I don't know what the uncertainty bar is on these timings so I don't know whether there is indeed a speed up , but I certainly have no more concerns about overhead introduced by the progress bar.
> 
> I did notice that there are a couple of times where the export hangs for a long time in between progress bars - I'll try to make a short movie for a subset of the DB.

Thanks for the testing @ltalirz! Truly helpful (especially the video).
Seeing the video, I am sure the debug mode is as it should be, concerning the progress bar mixing. To me, I don't want to remove the snapshot progress of how the progress bar fared during debug mode, rather, I would like to keep it there between the debug print statements, to make sure all went as expected with the progress bar.

Concerning the timings, I am sure there is a bit of caching or other involved, not completely sure of which level it's at, but I agree with your assessment that the progress bar will not cause the import/export module extra trouble :) (at least not more than the issues it already has).

CasperWA:
@sphuber @ltalirz This should be ready for "final" review.
I have implemented @ltalirz's suggestions from the latest review, as well as aligned the import logic slightly so that the two backend-dependent functions are even more similar.

ltalirz:
fyi, this is how the export now looks like (with debug enabled)
https://asciinema.org/a/UGbRXIOWVxbbgkUbdJMm9Y1c0

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -325,108 +343,128 @@ def export_tree(
...
-        if not silent:
-            print('No nodes to store, exiting...')
+    model_data = sum(len(model_data) for model_data in export_data.values())
+    if not model_data:
+        if not silent or debug:
+            print('Nothing to store, exiting...')
         return
 
+    if not silent or debug:
+        print('Exporting a total of {} database entries, of which {} are Nodes.'.format(model_data, len(all_node_pks)))
```
ltalirz, line 413:
```suggestion
        print('Exporting {} database entries ({} Nodes).'.format(model_data, len(all_node_pks)))
```
(otherwise doesn't fit left of the bar)

aiida/cmdline/commands/cmd_export.py
```diff
@@ -79,10 +79,11 @@ def inspect(archive, version, data, meta_data):
     show_default=True,
     help='Include or exclude comments for node(s) in export. (Will also export extra users who commented).'
 )
+@click.option('--debug', is_flag=True, default=False, flag_value=True, help='Show debug messages.', hidden=True)
```
ltalirz, line 82: I suggest using the existing `params.options.VERBOSE` option instead (also since the extra print statements are not excessive).

aiida/cmdline/commands/cmd_import.py
```diff
@@ -44,6 +91,7 @@ def _try_import(migration_performed, file_to_import, archive, group, migration,
     :param group: AiiDA Group into which the import will be associated.
     :param migration: Whether or not to force migration of archive, if needed.
     :param non_interactive: Whether or not the user should be asked for input for any reason.
+    :param more_archives: Whether or not there are more archives to be imported.
```
ltalirz, line 94: I'm not this function needs to know about other archives.
I think this would better be handled outside this function.

aiida/cmdline/commands/cmd_import.py
```diff
@@ -34,7 +33,55 @@ class ExtrasImportCode(Enum):
...
+    # Close progress bar, if it exists
+    if progress_bar:
+        if not debug:
+            progress_bar.leave = False
+        progress_bar.close()
+
+    if debug:
+        exception = traceback.format_exc()
+    else:
+        exception = traceback.format_exc(limit=0).splitlines()[-1]
```
ltalirz, line 60: I think what you do below can simply be handled by something like `str(e)` or `e.msg` where `e` is the `Exception` instance, instead of parsing from the traceback.
You may need to pass the exception to the `_echo_error` function explicitly, which is perhaps anyhow a bit cleaner.

aiida/tools/importexport/common/archive.py
```diff
@@ -79,9 +81,9 @@ def unpack(self):
         if os.path.isdir(self.filepath):
             extract_tree(self.filepath, self.folder)
         elif tarfile.is_tarfile(self.filepath):
-            extract_tar(self.filepath, self.folder, silent=True, nodes_export_subfolder=NODES_EXPORT_SUBFOLDER)
+            extract_tar(self.filepath, self.folder, silent=False, nodes_export_subfolder=NODES_EXPORT_SUBFOLDER)
```
ltalirz, line 84: why change the default here?

aiida/tools/importexport/common/archive.py
```diff
@@ -218,7 +220,23 @@ def _read_json_file(self, filename):
...
+    :type path: str
+
+    :param progress_bar: `tqdm` progress bar to have its description updated
+    """
+    (path, description) = os.path.split(path)
+    while description == '':
+        (path, description) = os.path.split(path)
+    description = 'EXTRACTING: {}'.format(description)
+
+    progress_bar.set_description_str(description)
```
ltalirz, line 236: Note: by default this forces a refresh, see https://tqdm.github.io/docs/tqdm/#set_description_str
I don't think we want to do this hundreds of thousands of times.

aiida/tools/importexport/common/archive.py
```diff
@@ -252,33 +269,44 @@ def extract_zip(infile, folder, nodes_export_subfolder=None, silent=False):
         with zipfile.ZipFile(infile, 'r', allowZip64=True) as handle:
 
             if not handle.namelist():
-                raise CorruptArchive('no files detected')
+                raise CorruptArchive('no files detected in archive')
+
+            if silent:
+                p_bar = json_files
```
ltalirz, line 275: what does this mean?
the progress bar is a set of json files?
I hope there is a less dirty way of achieving the behavior you desire.

tests/tools/importexport/orm/test_extras.py
```diff
@@ -89,7 +89,7 @@ def test_absence_of_extras(self):
             self.imported_node.get_extra('c')
 
     def test_extras_import_mode_keep_existing(self):
-        """Check if old extras are not modified in case of name collision"""
+        """Check if old extras are not modified in case of name collision (keep org, create new, leave org)"""
```
ltalirz, line 92: what does this mean?
what is org?

aiida/tools/importexport/common/archive.py
ltalirz: perhaps rather call it "file_iterator" or similar

CasperWA:
> fyi, this is how the export now looks like (with debug enabled)
> https://asciinema.org/a/UGbRXIOWVxbbgkUbdJMm9Y1c0

This is very acceptable for me (could you possibly also create (a) short movie(s) for non-debug modes?)

aiida/tools/importexport/dbexport/__init__.py
CasperWA: This is not supposed to be in the progress bar. This is a permanent print.

Check the export/import in normal mode.

aiida/tools/importexport/common/archive.py
```diff
@@ -311,37 +338,51 @@ def extract_tar(infile, folder, nodes_export_subfolder=None, silent=False):
...
         with tarfile.open(infile, 'r:*', format=tarfile.PAX_FORMAT) as handle:
 
-            try:
-                handle.extract(path=folder.abspath, member=handle.getmember('metadata.json'))
-            except KeyError:
-                raise CorruptArchive('required file `metadata.json` is not included')
+            if len(handle.getmembers()) == 1 and handle.getmembers()[0].size == 0:
+                raise CorruptArchive('no files detected in archive')
+
+            if silent:
```
ltalirz, line 344: since you are repeating yourself here, could this code be moved outside?

aiida/tools/importexport/dbimport/__init__.py
```diff
@@ -64,13 +64,15 @@ def import_data(in_path, group=None, silent=False, **kwargs):
...
 
-    if configuration.PROFILE.database_backend == BACKEND_SQLA:
+    backend = configuration.PROFILE.database_backend
+
+    if backend == BACKEND_SQLA:
         from aiida.tools.importexport.dbimport.backends.sqla import import_data_sqla
         return import_data_sqla(in_path, group=group, silent=silent, **kwargs)
 
-    if configuration.PROFILE.database_backend == BACKEND_DJANGO:
+    if backend == BACKEND_DJANGO:
```
ltalirz, line 73:
```suggestion
    elif backend == BACKEND_DJANGO:
```

aiida/tools/importexport/dbimport/utils.py
```diff
@@ -241,3 +243,48 @@ def deserialize_field(key, value, fields_info, import_unique_ids_mappings, forei
...
+    """Print header for import"""
+    import os
+    from aiida.cmdline.utils import echo
+
+    archive = os.path.basename(archive)
+    title = 'IMPORT - !!! DEBUG MODE !!!' if debug else 'IMPORT'
+    result = '\n{}'.format(tabulate([['Archive', archive]], headers=[title, '']))
+
+    parameters = [
+        ['Comment mode', comment_mode],
```
ltalirz, line 258: Why call it "rules" in export and "mode" in import?
Perhaps use "rules" everywhere?

aiida/tools/importexport/dbimport/utils.py
```diff
@@ -241,3 +243,48 @@ def deserialize_field(key, value, fields_info, import_unique_ids_mappings, forei
 
     # else
     return ('{}_id'.format(key), None)
+
+
+def start_header(archive, comment_mode, extras_mode_new, extras_mode_existing, debug):
+    """Print header for import"""
+    import os
+    from aiida.cmdline.utils import echo
```
ltalirz, line 251: move imports to top of module.
I'm surprised pylint doesn't complain

aiida/tools/importexport/dbimport/utils.py
```diff
@@ -241,3 +243,48 @@ def deserialize_field(key, value, fields_info, import_unique_ids_mappings, forei
...
+        ['Comment mode', comment_mode],
+        ['New Node Extras mode', extras_mode_new],
+        ['Existing Node Extras mode', extras_mode_existing],
+    ]
+    result += '\n\n{}'.format(tabulate(parameters, headers=['Parameters', '']))
+
+    echo.echo(result)
+
+
+def finish_header(results, import_group_label):
```
ltalirz, line 267: is this still a "header"?
maybe `def summarize_results(...)`?

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -84,10 +89,17 @@ def export_zip(what, outfile='testzip', overwrite=False, silent=False, use_compr
...
         raise exceptions.ArchiveExportError("the output file '{}' already exists".format(outfile))
 
-    time_start = time.time()
-    with ZipFolder(outfile, mode='w', use_compression=use_compression) as folder:
-        export_tree(what, folder=folder, silent=silent, **kwargs)
     if not silent:
+        file_format = 'Zip (compressed)' if use_compression else 'Zip (uncompressed)'
+        print_header(file_format, outfile, debug, **kwargs)
+
+    if debug:
```
ltalirz, line 96: how about moving anything that is not specific to export_zip outside?
e.g. timing should probably be handled at the top level, same as "File written"...

aiida/cmdline/commands/cmd_export.py
CasperWA: I don't like this, since I consider the extra print statements a developer/debug thing. Not for the common user. "Verbose" smells more like a valid flag a user might use, where "debug" is clearly _only_ for developers, or people who are interested in the inner workings.

tests/tools/importexport/orm/test_extras.py
CasperWA: original. This was to make the description of @yakutovicha's tests more explicit. Will make them even more explicit :)

aiida/tools/importexport/dbimport/__init__.py
CasperWA: No need, since the previous statement is a `return`.

aiida/tools/importexport/dbimport/utils.py
CasperWA: I don't think any of it is a header in the end - only in the sense that it's above the progress bar, which will always be the case, even for this. I think maybe using "summary" instead of header is indeed more appropriate throughout.

aiida/tools/importexport/dbexport/__init__.py
CasperWA: Not really related to the progress bar - but sure 👍 I think several of your comments are not related to the progress bar, but since many of my changes to the code is also not directly related to this, it seems completely fair :)

aiida/cmdline/commands/cmd_export.py
CasperWA: In the end, I consider the progress bar (including summary/header) to be a complete exchange for the current print statements. The old print statements are still useful for debugging however, since they are static in the output and show up at given places.

CasperWA:
> Hi @CasperWA
> I've started reviewing the PR - I still have some of the `dbexport/__init__` to go + I still need to look at
> the two `__init__.py`s in dbimport/backends.
> 
> Apart from that I went through everything; now need to go to a meeting

Cheers. There are lots to go through! So take your time. And remember that there are many places in the import/export files that one can pinpoint things that could be optimized, but that is _not_ the point of this PR. However, it is naturally also not my point to add more "bad" code ;)

aiida/tools/importexport/dbexport/__init__.py
ltalirz: I see, so I guess what is wrong here is the if statement: this is printed in non-debug mode!

ltalirz:
Cheers, I'll try this evening to go through all your replies + through the remaining files.

aiida/cmdline/commands/cmd_export.py
ltalirz: Ok. In that case I suggest you add it in params.options.DEBUG so that other commands can reuse it (e.g. the restapi one probably should)

aiida/tools/importexport/dbimport/__init__.py
ltalirz: It's not super important of course, but my change saves the evaluation of the second if condition if the first one is met.
It is always the right thing to do if you have multiple mutually exclusive options.

aiida/tools/importexport/dbexport/__init__.py
CasperWA: It's supposed to be present if we're _not_ in silent mode (default) OR in debug mode.

aiida/cmdline/commands/cmd_import.py
CasperWA: It seems to only be used to transfer the knowledge on to `_echo_error`. I could tuck it into `kwargs`? But it is _needed_ for the `_echo_error` function to determine what next action to take...

aiida/cmdline/commands/cmd_import.py
CasperWA: I think you're right. My original reason, if I remember correctly, for doing this was that to limit the traceback stack to the immediate exception thrown - for a cleaner output.

aiida/tools/importexport/common/archive.py
CasperWA: It's not the default. It's the actual call to `extract_tar` and `extract_zip`.
I think what would be better is to add `silent` as an ´Archive` property - but then again, it doesn't _really_ make sense to have an `Archive` be `silent` - only the action of unpacking it makes sense to be silent ... But maybe that's too philosphical, and we need a more practical approach? 😅

aiida/tools/importexport/common/archive.py
CasperWA: It doesn't seem to be a problem in my tests with big export archives - but I agree and will implement the `update` parameter in the `update_description()` utility function, with a default of `False`.

aiida/tools/importexport/common/archive.py
CasperWA: A lot of this I left as is, since I have another PR to completely redo all this code (#3242). Instead I have kept it as explicit as possible, but if you want I can try and minimize the code duplication here until #3242 will be implemented?

aiida/tools/importexport/dbimport/__init__.py
CasperWA: > It's not super important of course, but my change saves the evaluation of the second if condition if the first one is met.

Are you sure?
I think Python reads from the top down, so no matter what, the first `if`-statement will always be evaluated first, and if True, the block will be executed, and first _then_ the second `if`-statement will be evaluated, however, it never will, since the first `if`-block will have returned out of the evaluation process.. right?

ltalirz:
I thought a bit more about the `debug` variable and the progress bar and how to make things more elegant.

I think the canonical way of handling logging is this (you can apply it directly to the `debug` messages):
```python
import logging

# create logger with 'spam_application'
EXPORT_LOGGER = logging.getLogger(__name__)
EXPORT_LOGGER.setLevel(logging.DEBUG)

# create formatter
formatter = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')

# the following you do inside a function, only when `debug=True`
ch = logging.StreamHandler()
ch.setLevel(logging.DEBUG)
ch.setFormatter(formatter)
EXPORT_LOGGER.addHandler(ch)

# inside the functions, you then always do (no if statements needed)
EXPORT_LOGGER.debug("msg")
```

By analogy, I would do the same for the progress bar.

 * You have one singleton instance of the progress bar that other modules can import.
 * When `silent=True` you simply set this to invisible / disable it 
 * You simply do your progress bar updates always - they will have no effect when things are disabled

ltalirz:
Just some further comments on the logger from discussing with @sphuber : 
We may want to reuse the settings for the existing AiiDA logger and get a "child" - e.g.
```
from aiida.common.log import AIIDA_LOGGER
import logging
EXPORT_LOGGER = AIIDA_LOGGER.getChild('export')
EXPORT_LOGGER.setLevel(logging.DEBUG)
```
and then make sure that the `EXPORT_LOGGER.handlers` is according to our specifications

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -140,149 +156,153 @@ def export_tree(
...
             )
 
     # Add all the nodes contained within the specified groups
-    if given_group_entry_ids:
+    if GROUP_ENTITY_NAME in entities_starting_set:
 
-        if not silent:
-            print('RETRIEVING NODES FROM GROUPS...')
+        if debug or not silent:
+            progress_bar.set_description_str('Retrieving Nodes from Groups ...', refresh=True)
```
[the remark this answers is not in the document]
CasperWA: I know - just want to be explicit about these refreshments, since a future code-reader may not know this immediately.

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -325,108 +343,128 @@ def export_tree(
...
+                    'in': export_data[GROUP_ENTITY_NAME]
+                }
+            }, project='uuid', tag='groups'
+        ).append(orm.Node, project='uuid', with_group='groups')
+
+        if not silent:
+            total_node_uuids_for_groups = group_uuids_with_node_uuids.count()
+            if total_node_uuids_for_groups:
+                progress_bar.reset(total=total_node_uuids_for_groups)
+                progress_bar.set_description_str('Exporting Groups ...', refresh=False)
```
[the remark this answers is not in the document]
CasperWA: Sure? I am not sure what you mean by this?
I call `progress_bar.update()` after this, hence the description will also update at that stage, hence I avoid one refresh by setting it to `False` here.

ltalirz:
Just wanted to say the I find the progress bar very informative, e.g. during file extraction etc. 
It's definitely a substantial improvement!

One thing I noticed is that, after the "Creating import group" bar was full, I waited for a couple of minutes with nothing happening and no indication of what was going on.
Looking at the code, I believe what is taking the time is the removal of the sandbox folder.
It would be good to somehow indicate that this is what is going on.

CasperWA:
> One thing I noticed is that, after the "Creating import group" bar was full, I waited for a couple of minutes with nothing happening and no indication of what was going on.
> Looking at the code, I believe what is taking the time is the removal of the sandbox folder.
> It would be good to somehow indicate that this is what is going on.

It may also be the actual addition of the nodes to the `Group`. Since this is done after the progress bar has reached 100 % (it's a bit sneaky, but the "Creating import Group" is actually counting up the retrieval of all the newly imported `Node`s, which will then subsequently be added to the import `Group`).

I have added a `"Done"` after the addition of `Node`s to the `Group`, prior to anything else happening, so it should be clear then if it's the addition of `Node`s or something else that happens afterwards, i.e., whether the `"Done"` is there for some minutes or it's still `"Creating import Group"`.

CasperWA:
@ltalirz thanks for reviews!
I have been through them and given the whole thing an overhaul, in terms of exchanging `debug` with log messages and having a single progress bar that can be retrieved by anyone (instead of passing it around).

The latter, I think, also naturally divides the import function into smaller segments (whenever the `get_progress_bar` function is called). This may be useful if we in the future wish to modularize or unify this further, e.g., if the new import/export system seems too far down the road.

It should be ready for another review.
Note that I kept the `DEBUG` cmdline parameter in place, but since it is not being used now (only by the REST API) it may as well be removed again. But I kept it in for now.

ltalirz:
Just wanted to say that I'll be looking into this as soon as I have the
time, hopefully on Friday.

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -29,31 +29,34 @@
...
 
 from .zip import ZipFolder
 
-__all__ = ('export', 'export_zip')
+__all__ = ('export',)
 
 
-def export_zip(what, outfile='testzip', overwrite=False, silent=False, use_compression=True, **kwargs):
-    """Export in a zipped folder
+def export(entities, filename=None, file_format=None, overwrite=False, silent=False, use_compression=True, **kwargs):
```
sphuber, line 40: Note that you are breaking backwards compatibility here by removing the `export_zip` function and by changing the argument `outfile` to `filename`. I checked the [list of public resources](https://aiida.readthedocs.io/projects/aiida-core/en/latest/working_with_aiida/python_api.html) on the documentation, and neither `export` nor `export_zip` are on there to my surprise, so technically we are in the clear here. Just something we may want to think about a bit

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -81,14 +84,92 @@ def export_zip(what, outfile='testzip', overwrite=False, silent=False, use_compr
         exporting.
     :raises `~aiida.common.exceptions.LicensingException`: if any node is licensed under forbidden license.
     """
-    if not overwrite and os.path.exists(outfile):
-        raise exceptions.ArchiveExportError("the output file '{}' already exists".format(outfile))
+    if file_format is None:
```
sphuber, line 87: Why not put `zip` as the default in the function definition?

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -81,14 +84,92 @@ def export_zip(what, outfile='testzip', overwrite=False, silent=False, use_compr
...
+
+    if not overwrite and os.path.exists(filename):
+        raise exceptions.ArchiveExportError("The output file '{}' already exists".format(filename))
 
-    time_start = time.time()
-    with ZipFolder(outfile, mode='w', use_compression=use_compression) as folder:
-        export_tree(what, folder=folder, silent=silent, **kwargs)
     if not silent:
-        print('File written in {:10.3g} s.'.format(time.time() - time_start))
+        if file_format == 'tar':
```
sphuber, line 100: Note that `tar` is not the same as `tar.gz`. Tar is just an archive, no compression whatsoever, which is clearly not the case for a gzipped tarball. I really wouldn't use `tar` to actually mean `tar.gz`

aiida/tools/importexport/common/archive.py
```diff
@@ -218,9 +222,70 @@ def _read_json_file(self, filename):
...
+    :type file_handle: `tarfile.TarFile`, `zipfile.ZipFile`
+
+    :param folderpath: Path to folder.
+    :type folderpath: str
+
+    :param silent: suppress progress bar.
+    :type silent: bool
+
+    :return: List of filenames in the archive.
+    :rtype: list
```
sphuber, line 255: Is this correct? I can only see `return get_progress_bar` and `get_progress_bar` should return an instance of progress bar, not a list?

aiida/tools/importexport/common/archive.py
```diff
@@ -311,37 +363,33 @@ def extract_tar(infile, folder, nodes_export_subfolder=None, silent=False):
...
                 if not member.name.startswith(nodes_export_subfolder + os.sep):
                     continue
+
+                update_description(member.name, file_iterator)
+
                 handle.extract(path=folder.abspath, member=member)
     except tarfile.ReadError:
-        raise ValueError('The input file format for import is not valid (1)')
+        raise ValueError('The input file format for import is not valid (not a tar file)')
+    close_progress_bar(leave=False)
```
sphuber, line 392: I appreciate that this is not an easy problem, but I am really a bit worried about the tight coupling of all these various components to the progress bar. You may have removed the necessity of passing an instance around, but just hiding this dependency by going through a global singleton is not actually decoupling it. Why do you need the progress bar to be updated here inside the archive class? Why can't this be done in the caller?

Is it really not possible to have the top import method instantiate a progress bar and update it when necessary. If it needs information that are hidden in lower lying functions/classes, rather than passing the progress bar in there, one could maybe implement a method to get the information needed. The import method could call that first, update the progress bar and then call the actual method. Can you point me to the code why you think it is crucial to have the `Archive` class do it?

aiida/tools/importexport/dbexport/__init__.py
CasperWA: I know. And I can change it back if needed. But when/if we update this module, this will all change anyway.
And while it looks like I'm removing `export_zip` here, it's not true. I am merely renaming the former function `export` to `export_tar` and creating a wrapper function `export` to call either `export_zip` or `export_tar`, giving it a proper public API - instead of the weird setup that was the case before, where there were three major function one could call:

- `export_tree`: In reality not a small function that tries to export into a system folder, but the actual main export function.
- `export_zip`: A wrapper for `export_tree` to stream it all into a home-backed `ZipFolder` class (and subsequently a zip file).
- `export`: A wrapper for `export_tree` to run it and later tar and gzip it, resulting in a `tar.gz` file.

Now we have instead the following:

- `export_tree`: No changes, still the major export function.
- `export_zip`: No changes, other than it now focuses on what it should do, i.e., run main export function and zip the result.
- `export_tar`: Renamed `export`, which, like `export_zip`, now focuses on what it sohuld do, i.e., run main export function and tar and gzip the result.
- ´export`: Wrapper function with logging and timings for `export_tar` and `export_zip`.

aiida/tools/importexport/dbexport/__init__.py
CasperWA: I could, I guess. I just have it in my head that setting defaults like that, instead of `None` can cause a usage issue - but that may be in different scenarios - like when setting a class attribute, and not in this case?

aiida/tools/importexport/dbexport/__init__.py
CasperWA: Since we do not support any other `tar` format than the g-zipped tarball, I thought it of no import. But since we are calling it the same (`tar.gz`) in the click options, I will align with that.

aiida/tools/importexport/common/archive.py
CasperWA: True. This is a failure to update the doc string after the last big re-write.
But as information: The `tqdm` progress bar can act as a replacement for the iterable one wants to traverse, at the same time updating the progress bar for each run of the traversal. So it's more helpful to think of it here as an iterable. However, the instance that is returned, is indeed still `tqdm.tqdm`.

CasperWA:
> @CasperWA thanks for all the work. I haven't tried the code myself, but I trust @ltalirz when he says it is both more efficient and also provides a better UX. Still I have a few comments. They are all minor and easily addressable, except one, and that is the manner in which the progress bar is integrated in the code. I don't want to dismiss all the work you have done here, but I think it is important to discuss. Would you maybe have time to discuss in person? Maybe you can try to explain why you think this is the best way of doing it and if so, then I will be fine with it.

To see what it looks like is easy.
Just find yourself a nice export archive. Create a dummy profile, import the archive, and export it again.

Yeah, the progress bar is a bit tricky. And again, I didn't want to do too much work in implementing it, since I expect this whole code section to be rebuild from the bottom up.

However, if there is an easier way of doing it, I would happily hear of it.
I'll write you on slack.

sphuber:
As discussed offline with @CasperWA just now: I still am not a fan of the global singleton solution of the progress bar. It might seem more decoupled than the original solution of explicitly passing an instance around in method calls, but that is mere appearance. The `Archive` class which should be independent of the progress bar is no longer, and worse, the dependency is hidden. Only because it calls `extract_zip`, which calls some standalone utility which then ultimate fetches the global instance.

Nonetheless, apparently having this additional level of granularity in the unpacking of the archives is important for the UX and it is difficult to come up with a solution that does not enter the `Archive` class somehow. So for now I am ok with this solution.

aiida/tools/importexport/dbexport/__init__.py
ltalirz: Thanks for the clarification. Since people may indeed have used the `export` function in scripts (I certainly know that we have used `export_zip` in scripts), could you make it such that the change is still backwards compatible?
I.e. that a call to `export` (e.g. using `outfile`) with the old parameters will continue to work (I don't really care whether it produces a tar or a zip; feel free to decide the default).

aiida/tools/importexport/dbexport/utils.py
```diff
@@ -270,3 +281,32 @@ def check_process_nodes_sealed(nodes):
             'All ProcessNodes must be sealed before they can be exported. '
             'Node(s) with PK(s): {} is/are not sealed.'.format(', '.join(str(pk) for pk in nodes - sealed_nodes))
         )
+
+
+def summary(file_format, outfile, **kwargs):
```
ltalirz, line 286: to be consistent, this should also be printed to the logger, right?

aiida/tools/importexport/dbimport/utils.py
```diff
@@ -241,3 +249,42 @@ def deserialize_field(key, value, fields_info, import_unique_ids_mappings, forei
...
+        ['Comment rules', comment_mode],
+        ['New Node Extras rules', extras_mode_new],
+        ['Existing Node Extras rules', extras_mode_existing],
+    ]
+    result += '\n\n{}'.format(tabulate(parameters, headers=['Parameters', '']))
+
+    echo.echo(result)
+
+
+def result_summary(results, import_group_label):
```
ltalirz, line 269: also this should go to the logger I guess

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -81,14 +92,93 @@ def export_zip(what, outfile='testzip', overwrite=False, silent=False, use_compr
...
+
+    if not overwrite and os.path.exists(filename):
+        raise exceptions.ArchiveExportError("The output file '{}' already exists".format(filename))
 
-    time_start = time.time()
-    with ZipFolder(outfile, mode='w', use_compression=use_compression) as folder:
-        export_tree(what, folder=folder, silent=silent, **kwargs)
     if not silent:
-        print('File written in {:10.3g} s.'.format(time.time() - time_start))
+        if file_format == ExportFileFormat.TAR_GZIPPED:
```
ltalirz, line 108: this if condition would then go away as well

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -325,108 +407,122 @@ def export_tree(
...
+    # Pointer. Renaming, since Nodes have now technically been retrieved and "stored"
+    all_node_pks = node_ids_to_be_exported
 
-    if sum(len(model_data) for model_data in export_data.values()) == 0:
+    model_data = sum(len(model_data) for model_data in export_data.values())
+    if not model_data:
+        msg = 'Nothing to store, exiting...'
         if not silent:
-            print('No nodes to store, exiting...')
+            print(msg)
```
ltalirz, line 470: I guess this was left over

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -325,108 +407,122 @@ def export_tree(
...
         return
 
+    msg = 'Exporting a total of {} database entries, of which {} are Nodes.'.format(model_data, len(all_node_pks))
     if not silent:
-        print(
-            'Exporting a total of {} db entries, of which {} nodes.'.format(
-                sum(len(model_data) for model_data in export_data.values()), len(all_nodes_pk)
-            )
-        )
+        print(msg)
```
ltalirz, line 476: same here

ltalirz:
P.S. Please also let me know whether this should be merged in one commit or whether you are going to separate changes into topical commits.

CasperWA:
> P.S. Please also let me know whether this should be merged in one commit or whether you are going to separate changes into topical commits.

Yup. I hope that it can be squashed, but I'll have to go through it and think it over.
Thanks for going through it yet again, again! 👍

aiida/tools/importexport/dbexport/__init__.py
CasperWA: Kind of not. It was intended that this message should also show up in the terminal.

aiida/tools/importexport/dbexport/utils.py
CasperWA: It would be good to have this information in there as well, yes.

aiida/tools/importexport/dbexport/__init__.py
sphuber: Writing to a logger does not necessarily mean it won't end up on the console. In that case one just needs to configure a handler to directs output to stdout.

CasperWA: I am aware of that, but then I would need to setup a whole other handler for just these things. But maybe it will be time well spent, when considering also the summaries... Although, I was thinking to have a more condensed version of these go to the logger...

CasperWA: So I have made a new handler to simply print all stuff to `sys.stdout`.
However, now is my question - how is this different from `print` (other than it also prints the whole thing to the log), and wouldn't this still have to be under a `if not silent` "clause"? Note, I am logging at the `INFO` level for all the things that should go to the console.
Should I instead use the custom `REPORT` level, or? How would toggling these messages now actually work?

CasperWA: Maybe I missed to clarify that I consider `silent` messages and `debug` messages quite different here. The first includes things that would "normally" go to the console to give a nice overview of what's happening, this including the actual progress bar. The latter is for debugging, to be able to see how far in the given `export` function we got until some error occurred, i.e., it's messages for us, not the users.

ltalirz: Hi @CasperWA , sorry for the late reply.
My basic idea was: the `silent` flag toggles the log level of the logger - if it is not set, use the default level; otherwise e.g. `ERROR`.

The logger would have a handler that writes to the stdout (and, if desired, also one writing to the AiiDA log file).

I'll now have another look at the code to check your implementation.

CasperWA: > Should I instead use the custom `REPORT` level, or? How would toggling these messages now actually work?

Maybe I evaluate the supplied `silent` parameter and set the logging level to `WARNING` if `silent=True`, and otherwise set it to `INFO` - caching the original logging level to be able to reinstate it at the end?

I don't like this though, since it momentarily overrides any level the user/developer may be otherwise expecting.

CasperWA: > My basic idea was: the `silent` flag toggles the log level of the logger - if it is not set, use the default level; otherwise e.g. `ERROR`.

See my other comment for a consideration on this.

> The logger would have a handler that writes to the stdout (and, if desired, also one writing to the AiiDA log file).

Yup, set this up already (except the logging to the AiiDA log file, I expected this to be there already since I am using a child Logger of the AiiDA logger).

> I'll now have another look at the code to check your implementation.

I havne't pushed any changes yet, I have it only locally for now.

ltalirz: > However, now is my question - how is this different from print (other than it also prints the whole thing to the log)

The difference to `print` is that logging is filtered by the log level.

> , and wouldn't this still have to be under a if not silent "clause"? Note, I am logging at the INFO level for all the things that should go to the console.
Should I instead use the custom REPORT level, or? How would toggling these messages now actually work?

As mentioned above, I suggest you use the `silent` flag to filter out all but the most important log messages (e.g. by setting the log level to `ERROR`).
If the flag isn't set, the default log level should then be `INFO` (I think?), i.e. anything logged with `logger.info` will show up, and anything logged with `logger.debug` won't.

As this example shows, what one would actually want is more than a boolean flag to control the log level. 
My plan is to introduce a verbosity flag in https://github.com/aiidateam/aiida-core/pull/3896 which could then also be used here.

CasperWA: Ah right! We just set the handler's logging level based on the `silent` parameters. Done. It doesn't change anything else.

CasperWA: > As mentioned above, I suggest you use the `silent` flag to filter out all but the most important log messages (e.g. by setting the log level to `ERROR`).
> If the flag isn't set, the default log level should then be `INFO` (I think?), i.e. anything logged with `logger.info` will show up, and anything logged with `logger.debug` won't.
> 
> As this example shows, what one would actually want is more than a boolean flag to control the log level.
> My plan is to introduce a verbosity flag in #3896 which could then also be used here.

Sure. But all of this can only be respected properly (I think) if we then set the level for the console stream handler according to the `silent` parameter. And even then, it's a bit off, since the target level (if `silent`) is arbitrary _and_ hard-coded.

And while the `verbosity` flag may solve that if implemented here as well, it doesn't solve the cases when not going through the CLI, but instead the Python API directly ...

sphuber: When this thread started a few minutes ago, I wrote a simple example script to demonstrate how the logging can (and I think should) be used. We should not be using `print` for debug or informational messages and we should definitely not be mixing it with logging. I think we should only use logging and it is already set up in AiiDA to be used. I don't know why you think that you should add more handlers, because that is not True. They are already defined and automatically configured and loaded when you load a profile. So if you call a `verdi` command, the logging is enabled. Here is the example script that I think demonstrates how this can be done:
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
import click

from aiida.cmdline.params import options, types


@click.command()
@options.PROFILE(type=types.ProfileParamType(load_profile=True))
@click.option('--silent', is_flag=True)
def launch(profile, silent):
    """Test CLI script to demonstrate logging."""
    import logging
    from aiida.common.log import AIIDA_LOGGER

    logger = AIIDA_LOGGER.getChild('cli')

    if silent:
        logging.disable(level=logging.CRITICAL)

    logger.debug('this is a debug level message')
    logger.info('this is an info level message')
    logger.warning('this is a warning level message')
    logger.critical('this is a critical level message')


if __name__ == '__main__':
    launch()
```

If you call this with default AiiDA config, you will see:
```
./test_logging.py
05/13/2020 03:01:31 PM <15932> aiida.cli: [WARNING] this is a warning level message
05/13/2020 03:01:31 PM <15932> aiida.cli: [CRITICAL] this is a critical level message
```
If you change the loglevel that changes:
```
verdi config logging.aiida_loglevel DEBUG
05/13/2020 03:06:52 PM <16162> aiida.cli: [DEBUG] this is a debug level message
05/13/2020 03:06:52 PM <16162> aiida.cli: [INFO] this is an info level message
05/13/2020 03:06:52 PM <16162> aiida.cli: [WARNING] this is a warning level message
05/13/2020 03:06:52 PM <16162> aiida.cli: [CRITICAL] this is a critical level message
```
If the `--silent` flag is intended to suppress _everything_ the current example does just that:
```
./test_logging.py --silent
```
now won't print anything.
The only decision then is whether `--silent` should still show errors and or critical messages.

ltalirz: > And while the verbosity flag may solve that if implemented here as well, it doesn't solve the cases when not going through the CLI, but instead the Python API directly

You can set the verbosity of the AiiDA logger through the python API of course (this is what the verbosity flag will do).
We just need to document how to do this.

CasperWA: > If you call this with default AiiDA config, you will see:
> 
> ```
> ./test_logging.py
> 05/13/2020 03:01:31 PM <15932> aiida.cli: [WARNING] this is a warning level message
> 05/13/2020 03:01:31 PM <15932> aiida.cli: [CRITICAL] this is a critical level message
> ```

Why doesn't this print the `INFO` statement as well?

> The only decision then is whether `--silent` should still show errors and or critical messages.

Fine.

I would need a different formatter though to get the desired output in the console (just `'%(message)s'`).

sphuber: > Why doesn't this print the `INFO` statement as well?

Because the default Python level is `WARNING` and the default in AiiDA is `REPORT` which is a level that we invented and does not even exist in normal PYTHON. Since `REPORT` is in between `INFO` and `WARNING` you don't see info logs by default

CasperWA: As I expected then and also referenced in earlier comments, fine.
Then how can I set another formatter without setting up a new handler as well?

CasperWA: I basically want to use `echo.echo()` without having to import `echo`, I guess, since this should only be imported and used in the `cmdline` module.

sphuber: Well the following would work:
```
    for handler in AIIDA_LOGGER.handlers:
        handler.setFormatter(logging.Formatter('%(message)s'))
```
but you should reset the changes at the end of the method. When called through `verdi` it is fine, because at the end of the command, the changed formatters don't matter anymore. But if the export methods are called from within a normal interpreter, the logging of all other code that is executed thereafter is also affected. So you should make sure to reset the changes to the logger at the end.

The real underlying problem is that you are affecting normal module code such that its output is such as the CLI interface would like it. But to be honest, this is a bit of a code smell. The CLI is not the only caller and making the code do stuff particularly for that end point is not what we should be doing. In the interest of getting this merged, let's finish this though and just undo the formatting changes at the end of your methods.

CasperWA: > The real underlying problem is that you are affecting normal module code such that its output is such as the CLI interface would like it. But to be honest, this is a bit of a code smell. The CLI is not the only caller and making the code do stuff particularly for that end point is not what we should be doing. (...)

But then I would need to set this up specifically in the `log` module, as you have done with certain exceptions for the daemons, etc.

CasperWA:
@ltalirz and @sphuber Please see the latest commits for implementation of logging as well as how to deprecate the `what` and `outfile` `export()` parameters.

CasperWA:
@sphuber NOTE: I have very slightly altered the `return` of `aiida.common.lang:type_check()`.

CasperWA:
This logging stuff is seriously messing up the tests now.

I think this continuous re-integration of "new" updates in an existing "old" PR is not at all desireable and should be discouraged as much as possible (in short, I'm tired of having to fix this PR due to seemingly pointless change requests).

CasperWA:
Locally, only a `verdi import` test fails, since it's checking something from the `output`, which is logged and so it apparently can not capture it.

CasperWA:
I think this PR and I are properly angry with each other at this point. Please when reviewing, keep in mind that non-essential changes can be put into an issue and addressed later.

PR #3605, opened 2019-12-04
Set job poll interval to zero in localhost pytest fixture.

greschd:
In the ``aiida_localhost`` fixture, set the "minimum job poll interval" to zero. This should in general make tests using this computer quicker.


ltalirz:
This is definitely how it should be - I'm just wondering why this is necessary?
Shouldn't 0 be the default for local computers?

greschd:
It's definitely not the default now - note that this is different from the "safe interval" defined in the transport.

As to why the default is different @sphuber probably knows more. Maybe it's to avoid idling with high CPU load? I can see this as being an ok default for production loads if there is such a tradeoff.

ltalirz:
Hm... for running on your laptop or workstation I think you will very likely want to set it to 0.
Of course if you install AiiDA directly on the cluster, then a finite number makes sense.

PR #3607, opened 2019-12-05
make it easy to load dummy profile

### LINKED ISSUE #3390: interface for loading dummy profile
So far, loading a dummy profile (no connection to a DB) has been necessary only for building the documentation on "readthedocs". The code necessary to do this therefore currently is a bit hackish and contains variables specific for this use case, see

https://github.com/aiidateam/aiida-core/pull/3374/files#diff-1ec61223727a10e5e4c74c68336809fcR261-R274

Loading a dummy profile could be useful not just on read the docs but in other contexts as well (e.g. for a quick test that doesn't need the DB... perhaps more).
But even if it is just used on RTD for the moment, it would be nice to provide an easier way for developers to load a dummy profile in their sphinx `conf.py`:

I propose to either
 1. add a function `load_dummy_profile()` or
 2. add an optional argument `dummy_profile` (or similar) to`load_profile` 

In the first case, the function could live not at the top level but only under `aiida.manage.configuration` to allow for some time to converge behavior and user interface.

mentioning @sphuber @giovannipizzi for comment

ltalirz:
fix #3390 

 - [x] do not create config file upon AiiDa import

So far, importing aiida would directly create a config file.
Such side effects can have problematic consequences e.g. if the default
path that AiiDA chooses is not writeable. It also means that you pollute
your file system just by importing the package.
This PR moves the creation of the config file (if necessary) into the
`load_profile` function, which is anyhow called e.g. by the `verdi`
command line.

The same goes for the configuration of the "logger".

 - [x] remove dummy sqlite setting for AiiDA database on RTD

 - [x] make loading a dummy profile as simple as setting `aiida.manage.configuration.RT_DOC_MODE=True` before calling `load_profile`


aiida/__init__.py
```diff
@@ -42,18 +42,6 @@
 )
 __paper_short__ = 'G. Pizzi et al., Comp. Mat. Sci 111, 218 (2016).'
 
-# Configure the default logging
-configure_logging()
```
sphuber: By removing this the logging will now not be configured until a profile is loaded. This is probably OK in most cases, but since we allow importing AiiDA and running parts of the code without a profile having been loaded, it might happen that logging is needed before the profile is loaded. This is the reason I already put this statement here and made the profile dependent part optional. I think we should keep it

aiida/manage/configuration/__init__.py
```diff
@@ -62,13 +64,22 @@ def load_profile(profile=None):
...
+        get_manager()._load_backend(schema_check=False)  # pylint: disable=protected-access
+
     return PROFILE
 
 
 def load_config(create=False):
-    """Instantiate the Config object representing the configuration file of the current AiiDA instance.
+    """Instantiate Config object representing an AiiDA configuration file.
+
+    Warning: Contrary to :func:`~aiida.manage.configuration.get_config`, this function is uncached and will always
```
sphuber: could use the `.. warning:` marker

docs/source/conf.py
```diff
@@ -234,41 +233,16 @@
...
-# on_rtd is whether we are on readthedocs.org, this line of code grabbed
-# from docs.readthedocs.org
-# NOTE: it is needed to have these lines before load_dbenv()
+# on_rtd is whether we are on readthedocs.org
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 
-if not on_rtd:  # only import and set the theme if we're building docs locally
-    try:
-        import sphinx_rtd_theme
-        html_theme = 'sphinx_rtd_theme'
```
sphuber: You are also removing the setting of the theme I think. Is this ok? And if so, maybe we can also remove the `sphinx-rtd-theme` dependency?

ltalirz:
>  I was also wondering if it is not possible to get completely rid of that ugly dummy profile.

I think there is definitely the opportunity to remove some code duplication, since there are other places (`temporary_config_instance`, test manager) where "dummy profiles" are needed as well.

When you say "get rid of it completely", what exactly do you have in mind?

aiida/__init__.py
ltalirz: At first I left it in but it turns out that `configure_logging()` needs the profile (currently).
Of course one could change the implementation, but I wonder (I really don't know): what is the logging used for, if no profile is loaded?

docs/source/conf.py
ltalirz: Ah, good catch - my intention was not to switch to another theme (which I don't think is a good idea since you want to compare to how it looks on rtd)
I actually should have left `html_theme = 'sphinx_rtd_theme'` in (without condition, since the dependency should always be present).

sphuber:
> When you say "get rid of it completely", what exactly do you have in mind?

That we no longer need this special switch just to build the documentation. I guess it all boils down to Django being complicated and requiring the settings to be loaded if you import the model classes.

sphuber:
Superseded by #3669

PR #3608, opened 2019-12-05
fix error message when repo not found

ltalirz:
discovered with @asle85 

When the repository folder was not found, AiiDA was throwing an
ImportError without any additional information.
Replacing with NotADirectoryError + path to directory

PR #3612, opened 2019-12-08
Temporarily fix `numpy` and remove unnecessary `aiida.common.utils.get_repository_path`

sphuber:
Would supersede PR #3608


sphuber:
> thanks @sphuber !
> 
> it's really a pity that we need to fix build requirements for pymatgen again.
> I wonder whether we can suggest to them a way where at least the build requirements are resolved via pip (respecting the same semantics for pre-releases) and not by easybuild...

I agree, this has bitten us now multiple times. We can always ask them to change anything. One additional problem, even if they might be willing to change it, we are currently fixed to an older version because after that they dropped python 3.5 support. It is unlikely they would want to backport their fix, so we would anyway have to do this until we drop python 3.5 ourselves.

PR #3613, opened 2019-12-09
GroupPaths: A class to provide attribute/key access to Groups, with delimited labels.

### LINKED ISSUE #3589: Possibility to bundle Groups
Having the possibility to bundle Groups would be convenient.

chrisjsewell:
Fixes #3589 

Heya, this is some code that I've been using for a while, to manage my groups as a pseudo folder system, which has been super useful. From the tests you should be able to understand its functionality. For example, if you have three groups labelled `['f1/f2/f3a', 'f1/f2/f3b', 'f1/f2/f3-c/f4a']`, you can do:

```python
grouppaths = GroupPaths()
sub_group = grouppaths.f1.f2
for node in sub_group["f3a"].nodes:
    # do something
```

It appears to relate closely to 'Group Nesting' in the aiidacw2019-plan 😀

### DIFF
--- a/aiida/cmdline/commands/cmd_group.py
+++ b/aiida/cmdline/commands/cmd_group.py
@@ -361,3 +361,72 @@ def group_copy(source_group, destination_group):
     # Copy nodes
     dest_group.add_nodes(list(source_group.nodes))
     echo.echo_success('Nodes copied from group<{}> to group<{}>'.format(source_group.label, dest_group.label))
+
+
+@verdi_group.group('path')
+def verdi_group_path():
+    """Inspect groups of nodes, with delimited label paths."""
+
+
+@verdi_group_path.command('ls')
+@click.argument('path', type=click.STRING, required=False)
+@click.option('-R', '--recursive', is_flag=True, default=False, help='Recursively list sub-paths encountered')
+@click.option('-l', '--long', 'as_table', is_flag=True, default=False, help='List as a table, with sub-group count')
+@click.option(
+    '-d', '--with-description', 'with_description', is_flag=True, default=False, help='Show also the group description'
+)
+@click.option(
+    '--no-virtual',
+    'no_virtual',
+    is_flag=True,
+    default=False,
+    help='Only show paths that fully correspond to an existing group'
+)
+@click.option(
+    '-t',
+    '--type',
+    'group_type',
+    type=types.LazyChoice(valid_group_type_strings),
+    default=user_defined_group,
+    help='Show groups of a specific type, instead of user-defined groups. Start with semicolumn if you want to '
+    'specify aiida-internal type'
+)
+@click.option('--no-warn', is_flag=True, default=False, help='Do not issue a warning if any paths are invalid.')
+@with_dbenv()
+def group_path_ls(path, recursive, as_table, no_virtual, group_type, with_description, no_warn):
+    # pylint: disable=too-many-arguments
+    """Show a list of existing group paths."""
+    from aiida.tools.groups.paths import GroupPath, InvalidPath
+
+    try:
+        path = GroupPath(path or '', type_string=group_type, warn_invalid_child=not no_warn)
+    except InvalidPath as err:
+        echo.echo_critical(str(err))
+
+    if recursive:
+        children = path.walk()
+    else:
+        children = path.children
+
+    if as_table or with_description:
+        from tabulate import tabulate
+        headers = ['Path', 'Sub-Groups']
+        if with_description:
+            headers.append('Description')
+        rows = []
+        for child in sorted(children):
+            if no_virtual and child.is_virtual:
+                continue
+            row = [
+                child.path if child.is_virtual else click.style(child.path, bold=True),
+                len([c for c in child.walk() if not c.is_virtual])
+            ]
+            if with_description:
+                row.append('-' if child.is_virtual else child.get_group().description)
+            rows.append(row)
+        echo.echo(tabulate(rows, headers=headers))
+    else:
+        for child in sorted(children):
+            if no_virtual and child.is_virtual:
+                continue
+            echo.echo(child.path, bold=not child.is_virtual)
--- a/aiida/tools/groups/__init__.py
+++ b/aiida/tools/groups/__init__.py
@@ -1,0 +1,11 @@
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+# pylint: disable=wildcard-import,undefined-variable
+"""Provides tools for interacting with AiiDA Groups."""
+from .paths import *
+
+__all__ = paths.__all__
--- a/aiida/tools/groups/paths.py
+++ b/aiida/tools/groups/paths.py
@@ -1,0 +1,352 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Provides functionality for managing large numbers of AiiDA Groups, via label delimitation."""
+from collections import namedtuple
+from functools import total_ordering
+import re
+from typing import Any, Iterable, List, Optional  # pylint: disable=unused-import
+import warnings
+
+from aiida import orm
+from aiida.common.exceptions import NotExistent
+
+__all__ = ('GroupPath', 'InvalidPath')
+
+REGEX_ATTR = re.compile('^[a-zA-Z][\\_a-zA-Z0-9]*$')
+
+
+class InvalidPath(Exception):
+    """An exception to indicate that a path is not valid."""
+
+
+class GroupNotFoundError(Exception):
+    """An exception raised when a path does not have an associated group."""
+
+    def __init__(self, grouppath):
+        msg = 'No such group: {}'.format(grouppath.path)
+        super().__init__(msg)
+
+
+class GroupNotUniqueError(Exception):
+    """An exception raised when a path has multiple associated groups."""
+
+    def __init__(self, grouppath):
+        msg = 'The path is not unique: {}'.format(grouppath.path)
+        super().__init__(msg)
+
+
+class NoGroupsInPathError(Exception):
+    """An exception raised when a path has multiple associated groups."""
+
+    def __init__(self, grouppath):
+        msg = 'The path does not contain any descendant groups: {}'.format(grouppath.path)
+        super().__init__(msg)
+
+
+WalkNodeResult = namedtuple('WalkNodeResult', ['group_path', 'node'])
+
+
+@total_ordering
+class GroupPath:
+    """A class to provide label delimited access to groups.
+
+    See tests for usage examples.
+    """
+
+    def __init__(self, path='', type_string=orm.GroupTypeString.USER.value, warn_invalid_child=True):
+        # type: (str, Optional[str], Optional[GroupPath])
+        """Instantiate the class.
+
+        :param path: The initial path of the group.
+        :param type_string: Used to query for and instantiate a ``Group`` with.
+        :param warn_invalid_child: Issue a warning, when iterating children, if a child path is invalid.
+
+        """
+        self._delimiter = '/'
+        if not isinstance(type_string, str):
+            raise TypeError('type_string must a str: {}'.format(type_string))
+        self._type_string = type_string
+        self._path_string = self._validate_path(path)
+        self._path_list = self._path_string.split(self._delimiter) if path else []
+        self._warn_invalid_child = warn_invalid_child
+
+    def _validate_path(self, path):
+        """Validate the supplied path."""
+        if path == self._delimiter:
+            return ''
+        if self._delimiter * 2 in path:
+            raise InvalidPath("The path may not contain a duplicate delimiter '{}': {}".format(self._delimiter, path))
+        if (path.startswith(self._delimiter) or path.endswith(self._delimiter)):
+            raise InvalidPath("The path may not start/end with the delimiter '{}': {}".format(self._delimiter, path))
+        return path
+
+    def __repr__(self):
+        # type: () -> str
+        """Represent the instantiated class."""
+        return "{}('{}', type='{}')".format(self.__class__.__name__, self.path, self.type_string)
+
+    def __eq__(self, other):
+        # type: (Any) -> bool
+        """Compare equality of path and type string to another ``GroupPath`` object."""
+        if not isinstance(other, GroupPath):
+            return NotImplemented
+        return (self.path, self.type_string) == (other.path, other.type_string)
+
+    def __lt__(self, other):
+        # type: (Any) -> bool
+        """Compare less-than operator of path and type string to another ``GroupPath`` object."""
+        if not isinstance(other, GroupPath):
+            return NotImplemented
+        return (self.path, self.type_string) < (other.path, other.type_string)
+
+    @property
+    def path(self):
+        # type: () -> str
+        """Return the path string."""
+        return self._path_string
+
+    @property
+    def path_list(self):
+        # type: () -> List[str]
+        """Return a list of the path components."""
+        return self._path_list[:]
+
+    @property
+    def key(self):
+        # type: () -> str
+        """Return the final component of the the path."""
+        if self._path_list:
+            return self._path_list[-1]
+        return None
+
+    @property
+    def delimiter(self):
+        # type: () -> str
+        """Return the delimiter used to split path into components."""
+        return self._delimiter
+
+    @property
+    def type_string(self):
+        # type: () -> str
+        """Return the type_string used to query for and instantiate a ``Group`` with."""
+        return self._type_string
+
+    @property
+    def parent(self):
+        # type: () -> Optional[GroupPath]
+        """Return the parent path."""
+        if self.path_list:
+            return GroupPath(
+                self.delimiter.join(self.path_list[:-1]),
+                type_string=self.type_string,
+                warn_invalid_child=self._warn_invalid_child
+            )
+        return None
+
+    def __truediv__(self, path):
+        # type: (str) -> GroupPath
+        """Return a child ``GroupPath``, with a new path formed by appending ``path`` to the current path."""
+        if not isinstance(path, str):
+            raise TypeError('path is not a string: {}'.format(path))
+        path = self._validate_path(path)
+        child = GroupPath(
+            path=self.path + self.delimiter + path if self.path else path,
+            type_string=self.type_string,
+            warn_invalid_child=self._warn_invalid_child
+        )
+        return child
+
+    def __getitem__(self, path):
+        # type: (str) -> GroupPath
+        """Return a child ``GroupPath``, with a new path formed by appending ``path`` to the current path."""
+        return self.__truediv__(path)
+
+    def get_group(self):
+        # type: () -> Optional[orm.Group]
+        """Return the concrete group associated with this path."""
+        try:
+            return orm.Group.objects.get(label=self.path, type_string=self.type_string)
+        except NotExistent:
+            return None
+
+    @property
+    def group_ids(self):
+        # type: () -> List[int]
+        """Return all the UUID associated with this GroupPath.
+
+        :returns: and empty list, if no group associated with this label,
+            or can be multiple if type_string was None
+
+        This is an efficient method for checking existence,
+        which does not require the (slow) loading of the ORM entity.
+        """
+        query = orm.QueryBuilder()
+        filters = {'label': self.path}
+        if self.type_string is not None:
+            filters['type_string'] = self.type_string
+        query.append(orm.Group, filters=filters, project='id')
+        return [r[0] for r in query.all()]
+
+    @property
+    def is_virtual(self):
+        # type: () -> bool
+        """Return whether there is one or more concrete groups associated with this path."""
+        return len(self.group_ids) == 0
+
+    def get_or_create_group(self):
+        # type: () -> (orm.Group, bool)
+        """Return the concrete group associated with this path or, create it, if it does not already exist."""
+        if self.type_string is not None:
+            return orm.Group.objects.get_or_create(label=self.path, type_string=self.type_string)
+        return orm.Group.objects.get_or_create(label=self.path)
+
+    def delete_group(self):
+        """Delete the concrete group associated with this path.
+
+        :raises: GroupNotFoundError, GroupNotUniqueError
+        """
+        ids = self.group_ids
+        if not ids:
+            raise GroupNotFoundError(self)
+        if len(ids) > 1:
+            raise GroupNotUniqueError(self)
+        orm.Group.objects.delete(ids[0])
+
+    @property
+    def children(self):
+        # type: () -> Iterable[GroupPath]
+        """Iterate through all (direct) children of this path."""
+        query = orm.QueryBuilder()
+        filters = {}
+        if self.path:
+            filters['label'] = {'like': self.path + self.delimiter + '%'}
+        if self.type_string is not None:
+            filters['type_string'] = self.type_string
+        query.append(orm.Group, filters=filters, project='label')
+        if query.count() == 0 and self.is_virtual:
+            raise NoGroupsInPathError(self)
+
+        yielded = []
+        for (label,) in query.iterall():
+            path = label.split(self._delimiter)
+            if len(path) <= len(self._path_list):
+                continue
+            path_string = self._delimiter.join(path[:len(self._path_list) + 1])
+            if (path_string not in yielded and path[:len(self._path_list)] == self._path_list):
+                yielded.append(path_string)
+                try:
+                    yield GroupPath(
+                        path=path_string, type_string=self.type_string, warn_invalid_child=self._warn_invalid_child
+                    )
+                except InvalidPath:
+                    if self._warn_invalid_child:
+                        warnings.warn('invalid path encountered: {}'.format(path_string))  # pylint: disable=no-member
+
+    def __iter__(self):
+        # type: () -> Iterable[GroupPath]
+        """Iterate through all (direct) children of this path."""
+        return self.children
+
+    def __len__(self):
+        # type: () -> int
+        """Return the number of children for this path."""
+        return sum(1 for _ in self.children)
+
+    def __contains__(self, key):
+        # type: (str) -> bool
+        """Return whether a child exists for this key."""
+        for child in self.children:
+            if child.path_list[-1] == key:
+                return True
+        return False
+
+    def walk(self, return_virtual=True):
+        # type: () -> Iterable[GroupPath]
+        """Recursively iterate through all children of this path."""
+        for child in self:
+            if return_virtual or not child.is_virtual:
+                yield child
+            for sub_child in child.walk(return_virtual=return_virtual):
+                if return_virtual or not sub_child.is_virtual:
+                    yield sub_child
+
+    def walk_nodes(self, filters=None, node_class=None, query_batch=None):
+        # type: () -> Iterable[WalkNodeResult]
+        """Recursively iterate through all nodes of this path and its children.
+
+        :param filters: filters to apply to the node query
+        :param node_class: return only nodes of a certain class (or list of classes)
+        :param int batch_size: The size of the batches to ask the backend to batch results in subcollections.
+            You can optimize the speed of the query by tuning this parameter.
+            Be aware though that is only safe if no commit will take place during this transaction.
+        """
+        query = orm.QueryBuilder()
+        group_filters = {}
+        if self.path:
+            group_filters['label'] = {'or': [{'==': self.path}, {'like': self.path + self.delimiter + '%'}]}
+        if self.type_string is not None:
+            group_filters['type_string'] = self.type_string
+        query.append(orm.Group, filters=group_filters, project='label', tag='group')
+        query.append(
+            orm.Node if node_class is None else node_class,
+            with_group='group',
+            filters=filters,
+            project=['*'],
+        )
+        for (label, node) in query.iterall(query_batch) if query_batch else query.all():
+            yield WalkNodeResult(GroupPath(label, type_string=self.type_string), node)
+
+    @property
+    def browse(self):
+        """Return a ``GroupAttr`` instance, for attribute access to children."""
+        return GroupAttr(self)
+
+
+class GroupAttr:
+    """A class to provide attribute access to a ``GroupPath`` children.
+
+    The only public attributes on this class are dynamically created from the ``GroupPath`` child keys.
+    NOTE: any child keys that do not conform to an acceptable (public) attribute string will be ignored.
+    The ``GroupPath`` can be retrieved *via* a function call, e.g.::
+
+        group_path = GroupPath()
+        group_attr = GroupAttr(group_path)
+        group_attr.a.b.c() == GroupPath("a/b/c")
+
+    """
+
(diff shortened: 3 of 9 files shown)


giovannipizzi:
We discussed some API changes with @chrisjsewell and @greschd that now @chrisjsewell will implement

giovannipizzi:
Two questions that came to my mind:
- are different group types differentiated by the API? Or just the name is used, so users might unexpectedly see different things for different group types?
- shall we rename the auto-group labels from `Verdi autogroup on 2017-11-06 12:26:18` to something like `autogroup/20171106_122518`?
- shall we display the concept of group types instead as a top-level folder? or this is confusing?

Any feedback is welcome

greschd:
>     * shall we rename the auto-group labels from `Verdi autogroup on 2017-11-06 12:26:18` to something like `autogroup/20171106_122518`?

Seems nice. However, is this a backwards-incompatible change? We could create both, although that is also kind of ugly.

chrisjsewell:
Morning! 
8329485 implements pretty much all the discussed `GroupPath` API, except some points for discussion, which are denoted with `# TODO` in the source code.
If you look in the test module, it covers all the methods (I think), and so should give you an idea of how things are working. At the moment its written in pytest, as its quicker for development, and I expect you to have migrated to pytest by the end of this week 😜

aiida/cmdline/commands/cmd_group.py
```diff
@@ -353,3 +353,22 @@ def group_copy(source_group, destination_group):
     # Copy nodes
     dest_group.add_nodes(list(source_group.nodes))
     echo.echo_success('Nodes copied from group<{}> to group<{}>'.format(source_group.label, dest_group.label))
+
+
+@verdi_group.command('path')
+@click.option('-l', 'show_all', is_flag=True, default=False, help='Show all group paths')
```
giovannipizzi: I think:
- by default just show the Path, with bold for non-virtual, without any table (like `ls` without options)
- `-l` should be `--long` like in `ls` and show the table. Also here, use bold for non-virtual, but ok to keep the column (but maybe move the `Virtual` column as the third one). I would also rename `Children` (that means `len(self.children)` -> `Groups contained`
- what now it's `-ls/show_all` should become `-R/--recursive` like `ls`

chrisjsewell: Implemented these options, and some additional ones (see tests). Also, instead of `Groups contained` I have added `Sub-Groups`, which is a recursive count of all groups under that path. I think that will be more helpful.

chrisjsewell:
@giovannipizzi and @sphuber this is ready for 'final' review.
The only change that is definitely needed is for the tests. As you'll see, I'm currently using pytest, with the `clear_database` fixture. This works on my computer, but looks to be failing above due to no User being set the first time `clear_database` is called (but ok for subsequent tests!?). @ltalirz is this something that needs fixing within the fixture manager? Otherwise, I can just rewrite them in unittest.

chrisjsewell:
One other thing I forgot. It would be ideal to include tab completion of paths. I know you have dynamic autocompletion for `verdi data` sub-commands, but I don't think it is currently possible to do this for arguments, considering I have an open PR on this very issue! click-contrib/click-completion#27

ramirezfranciscof:
> @giovannipizzi and @sphuber this is ready for 'final' review.
> The only change that is definitely needed is for the tests. As you'll see, I'm currently using pytest, with the `clear_database` fixture. This works on my computer, but looks to be failing above due to no User being set the first time `clear_database` is called (but ok for subsequent tests!?). @ltalirz is this something that needs fixing within the fixture manager? Otherwise, I can just rewrite them in unittest.

@chrisjsewell @ltalirz 

Why have the tests failed? Is it because of the problem we solved last week or is something else related with the quoted comment above? If all is worked out now and you can rebase, I can then review the PR.

ltalirz:
Ah, this is a known issue - since the AiiDA core tests aren't yet fully migrated to pytest, there can be cases where the test database is completely empty, even without user.

@greschd has just added a `clear_database_before_test` fixture that you can use to circumvent this. https://github.com/aiidateam/aiida-core/pull/3783

chrisjsewell:
Yep @ramirezfranciscof, this is ready for review now 👍

aiida/tools/groups/grouppaths.py
```diff
@@ -0,0 +1,321 @@
...
+                return orm.Group.objects.get(label=self.path)
+            return orm.Group.objects.get(label=self.path, type_string=self.type_string)
+        except NotExistent:
+            return None
+
+    @property
+    def is_virtual(self):
+        # type: () -> bool
+        """Return whether there is a concrete group associated with this path."""
+        # Note we don't test (self.group is None) here, to improve efficiency,
```
ramirezfranciscof: I get that you don't want to load nodes for efficiency reasons, but is this really a desired behavior? If the user checks for non existent groups it will not say they don't exist, it will just think these are virtual. For example:

<pre><code>
In [3]: orm.Group.objects.get_or_create(label= 'a', type_string=orm.GroupTypeString.USER.value )
In [4]: orm.Group.objects.get_or_create(label= 'b/c', type_string=orm.GroupTypeString.USER.value )

In [5]: group_path['a'].is_virtual
Out[5]: False

In [6]: group_path['b'].is_virtual
Out[6]: True

In [7]: group_path['d'].is_virtual
Out[7]: True
</code></pre>

This..."lack of sensitivity"? then propagates to other functions, so for example:

<pre><code>
In [10]: list( group_path['d'].children )
Out[10]: []

In [11]: list( group_path['d/e'].children )
Out[11]: []

In [12]: list( group_path['d/e/f'].children )
Out[12]: []

In [13]: group_path['d/e/f'].path_list
Out[13]: ['d', 'e', 'f']

In [14]: group_path['d/e']['f']
Out[14]: GroupPath('d/e/f', type='user')
</code></pre>

I could find no direct way to check if a group actually exists or use methods in a way that will let you know if they are trying to do stuff in non-existent groups. The user has to do this manually by, for example, checking if each group label  is among its parent's children:

```
In [21]: group_path['b'].path in list( child.path for child in group_path['b'].parent.children )
Out[21]: True

In [22]: group_path['d/e/f'].path in list( child.path for child in group_path['d/e/f'].parent.children )
Out[22]: False
```

I think there would be value in at least wrapping this procedure around a 'group_exists' method, and I would argue that we should consider doing this check (by default?) for (almost?) all of the public methods of the class. Maybe it would require a slightly different implementation that uses private methods/properties if this way turns out to be cyclical (i.e. can't use methods inside the 'check' that also need to 'check' the path themselves), but I would think there is a way to do this without actually loading nodes or doing expensive queries.

Again, I don't remember if this was explicitly discussed during the coding week, I am just very wary of functions that have this "_I'm **technically** doing what you asked for_" kind of "malicious compliance" behavior.

aiida/tools/groups/grouppaths.py
```diff
@@ -0,0 +1,321 @@
...
+    @property
+    def browse(self):
+        """Return a ``GroupAttr`` instance, for attribute access to children."""
+        return GroupAttr(self)
+
+
+class GroupAttr:
+    """A class to provide attribute access to a ``GroupPath`` children.
+
+    The only public attributes on this class are dynamically created from the ``GroupPath`` child keys.
```
ramirezfranciscof: So, yeah, these are dynamically created, however I would maybe clarify somewhere that these get frozen the moment you create the object. So, whereas the object `group_path` created by  `group_path = GroupPath()` will always reflect the current state of the database, the object `group_attr` created by `group_attr = group_path.browse` will only reflect the state of the database at the moment of its creation.

On the other hand, I am no expert in this, but is this the expected/desired behavior? Isn't it possible to run `c.path_list[-1]: c for c in self._group_path.children if REGEX_ATTR.match(c.path_list[-1])` inside of `__getattr__`? Or is this much more expensive / incompatible with autocomplete features?

aiida/cmdline/commands/cmd_group.py
```diff
@@ -353,3 +353,66 @@ def group_copy(source_group, destination_group):
...
+    help='Show groups of a specific type, instead of user-defined groups. Start with semicolumn if you want to '
+    'specify aiida-internal type'
+)
+@with_dbenv()
+def group_path_ls(path, recursive, as_table, groups_only, group_type, with_description):
+    """Show a list of existing group paths."""
+    from aiida.tools.groups.grouppaths import GroupPath, InvalidPath
+
+    try:
+        path = GroupPath(path or '', type_string=group_type)
```
ramirezfranciscof: So, this is just a question from ignorance, but how does that "or" within arguments work? Perhaps I'm misunderstanding something but I think I never saw it before.

chrisjsewell: This is something that @ltalirz is fond of lol. It just means that if `path` evaluates to False, use `''`, and is used here as a shorthand for `'' if path is None else path`

aiida/tools/groups/grouppaths.py
chrisjsewell: > will always reflect the current state of the database

Good point, yeh I think it's mainly on how this would affect autocompletion speed.
Probably need a bit of a manual test on that.

aiida/tools/groups/grouppaths.py
chrisjsewell: > I could find no direct way to check if a group actually exists

Can you not just do `group_path['a'].group is not None`? I could add an `exists` method, but that is all it would do.

> I would argue that we should consider doing this check (by default?) for (almost?) all of the public methods of the class

I disagree. This class is intended to work like `pathlib.Path`, in that there is no explicit checks for group/file existence until you actually need the group. I think would be very costly to be making database calls for every operation.

chrisjsewell:
Thanks @ramirezfranciscof, I will work through these (it may take a little while though!)

aiida/tools/groups/grouppaths.py
chrisjsewell: Maybe the children one is a little counter-intuitive, and could have an existence check (without loading the node)

aiida/backends/tests/tools/groups/test_grouppath.py
```diff
@@ -0,0 +1,171 @@
...
+    assert group_path.is_virtual
+    assert group_path.group is None
+    assert (group_path / 'a').path == 'a'
+    assert (group_path / 'a' / 'b').path == 'a/b'
+    assert (group_path / 'a/b').path == 'a/b'
+    assert group_path['a/b'].path == 'a/b'
+    assert 'a' in group_path
+    assert 'x' not in group_path
+    assert not group_path['a'].is_virtual
+    assert group_path.group is None
```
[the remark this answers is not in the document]
chrisjsewell: Yeh probably just an artefact of an earlier version, that can be removed

aiida/tools/groups/grouppaths.py
```diff
@@ -0,0 +1,321 @@
...
+        self._delimiter = '/'
+        if type_string is not None and not isinstance(type_string, str):
+            raise AssertionError('type_string must be None or str: {}'.format(type_string))
+        self._type_string = type_string
+        self._path_string = self._validate_path(path)
+        self._path_list = self._path_string.split(self._delimiter) if path else []
+        self._warn_invalid_child = warn_invalid_child
+
+    def _validate_path(self, path):
+        """Validate the supplied path."""
```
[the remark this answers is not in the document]
chrisjsewell: You should aready be able to initiate the `GroupPath` with GroupPath(warn_invalid_child=False), to produce the 'ignore' behaviour you describe (but this shouldn't be the default).
Probably though there should be an option on the CLI to use this API feature.

aiida/tools/groups/grouppaths.py
ramirezfranciscof: > Can you not just do group_path['a'].group is not None? I could add an exists method, but that is all it would do.

Ah, ok, yes that looks shorter. It does require you to "load the group", right? Is this as expensive as loading nodes? I mean, probably doesn't matter for one check but maybe if you are doing some intensive processing the farfetched method could be faster?

> I disagree. This class is intended to work like `pathlib.Path`, in that there is no explicit checks for group/file existence until you actually need the group. I think it would be very costly to be making database calls for every operation.

So, I just checked `pathlib` and I think this is not exactly the case. It is true that it doesn't seem to check when you set the path and it does perform some methods without checking, but it also has an `exist` method and it is not letting me do iterations over an inexistent path:

```
In [1]: from pathlib import Path
In [2]: p = Path('.')
In [3]: p = Path('aloha')
In [4]: p.name
Out[4]: 'aloha'

In [5]: p.is_absolute()
Out[5]: False

In [6]: list( p.iterdir() )
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-7-ede0ddf8d362> in <module>()
----> 1 list( p.iterdir() )

/usr/lib/python3.6/pathlib.py in iterdir(self)
   1079         if self._closed:
   1080             self._raise_closed()
-> 1081         for name in self._accessor.listdir(self):
   1082             if name in {'.', '..'}:
   1083                 # Yielding a path object for these makes little sense

/usr/lib/python3.6/pathlib.py in wrapped(pathobj, *args)
    385         @functools.wraps(strfunc)
    386         def wrapped(pathobj, *args):
--> 387             return strfunc(str(pathobj), *args)
    388         return staticmethod(wrapped)
    389 

FileNotFoundError: [Errno 2] No such file or directory: 'aloha'
```

It also returns the same error when using the `rmdir` method. I don't think this is true for the `walk` and `delete_group`, no?

aiida/tools/groups/grouppaths.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+class InvalidPath(Exception):
+    pass
+
+
+WalkNodeResult = namedtuple('WalkNodeResult', ['group_path', 'node'])
+
+
+@total_ordering
```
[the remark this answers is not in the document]
chrisjsewell: It's purely a path string comparison, so that you can use `sorted(group_path.children)` and guarantee the order. As I've mentioned below, it should work generally the same as `pathlib.Path`, which in this case it does:

```python

In [0]: from pathlib import Path

In [1]: (Path('')/'a/f' ) > (Path('')/'a')
Out[1]: True

In [2]: (Path('')/'a/f' ) > (Path('')/'a')
Out[2]: True

In [3]: (Path('')/'a' ) < (Path('')/'a/c/e')
Out[3]: True

In [4]: (Path('')/'a/f' ) > (Path('')/'a/c/e')
Out[4]: True
```

aiida/backends/tests/tools/groups/test_grouppath.py
```diff
@@ -0,0 +1,171 @@
...
+
+from click.testing import CliRunner
+import pytest
+
+from aiida import orm
+from aiida.cmdline.commands.cmd_group import group_path_ls
+from aiida.tools.groups.grouppaths import GroupAttr, GroupPath, InvalidPath
+
+
+def test_basic(clear_database):
```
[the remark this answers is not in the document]
chrisjsewell: Yep, just me being lazy lol. Although obviously it's helpful to have 'progressive' tests, as you mention below.

aiida/backends/tests/tools/groups/test_grouppath.py
```diff
@@ -0,0 +1,171 @@
...
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Tests for GroupPath"""
+# pylint: disable=redefined-outer-name,unused-argument
+from __future__ import division
+from __future__ import print_function
+from __future__ import absolute_import
```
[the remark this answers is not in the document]
chrisjsewell: Yep 👍 , I guess these have all been remove during the course of me writing this module

aiida/backends/tests/tools/groups/test_grouppath.py
```diff
@@ -0,0 +1,171 @@
...
+    assert isinstance(group_path.browse.a.c.d, GroupAttr)
+    assert isinstance(group_path.browse.a.c.d(), GroupPath)
+    assert group_path.browse.a.c.d().path == 'a/c/d'
+    assert not set(group_path.browse.__dir__()).intersection(['bad space', 'bad@char', '_badstart'])
+    with pytest.raises(AttributeError):
+        group_path.browse.a.c.x  # pylint: disable=pointless-statement
+
+
+def test_cmdline_group_path_ls(clear_database):
+    """Test ``verdi group path ls``"""
```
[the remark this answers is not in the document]
chrisjsewell: 👍

aiida/cmdline/commands/cmd_group.py
```diff
@@ -353,3 +353,66 @@ def group_copy(source_group, destination_group):
...
+
+@verdi_group_path.command('ls')
+@click.argument('path', type=click.STRING, required=False)
+@click.option('-R', '--recursive', is_flag=True, default=False, help='Recursively list sub-paths encountered')
+@click.option('-l', '--long', 'as_table', is_flag=True, default=False, help='List in long format')
+@click.option(
+    '-d', '--with-description', 'with_description', is_flag=True, default=False, help='Show also the group description'
+)
+@click.option(
+    '-g', '--groups-only', 'groups_only', is_flag=True, default=False, help='Only show paths that contain a group'
```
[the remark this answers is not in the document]
chrisjsewell: 👍

aiida/tools/groups/grouppaths.py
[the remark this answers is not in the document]
ramirezfranciscof: Ahh, I see, so it is just for the purpose of listing it in a user / performance friendly way. Ok then, forget I mentioned it.

aiida/tools/groups/grouppaths.py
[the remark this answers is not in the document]
ramirezfranciscof: But in what case would you want a "broken" list? I mean, as I see it you can either get an already clean list which you can iterate without problems (although might be incomplete, depending on what you aim to do) or you want it to return an error telling you there are groups that can't be parsed this way. The broken list just pushes the error one step further so when I try to use `bad` as a group_path I may get an error saying it doesn't exist, yet the direct resources within this tool will tell me it does (I will only notice if I check for contradictions such that is both virtual and has no children or by using external tools such as manually querying for the group). I see no case where this doesn't just obfuscate the understanding of the situation.

aiida/tools/groups/grouppaths.py
chrisjsewell: Yep as I've mentioned in the 2nd comment, an existence check at the start of `def children` should clear this up (and raise an error on `delete_group`). 

Thats the mindset though; where possible mirror `pathlib.Path`. To that end, maybe a specific `GroupNotFoundError` exception class would also be a good idea?

aiida/tools/groups/grouppaths.py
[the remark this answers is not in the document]
chrisjsewell: I'll have a think on it 😁

aiida/tools/groups/grouppaths.py
ramirezfranciscof: Ah, yeah, sorry, I missed the second comment as it got posted after I started writing the reply. Yes, I think it would be important at least for the `children`, `walk` and `delete` methods, and I would explicitly create a method that directly checks for this. Maybe also consider the check for the `__getitem__` and the `group` methods? I'm not sure, personally I would do most of them but perhaps a third opinion would be good here.

And yeah, the specific error sounds good, although again I'm not very experienced in when it is convenient to create new error classes so also open to suggestions.

aiida/tools/groups/__init__.py
```diff
@@ -0,0 +1,14 @@
+# This file is part of the AiiDA code.                                    #
```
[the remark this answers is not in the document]
sphuber: This is not checked by the pre-commit. We have a script that does this when we create a release. We could think about adding this to the pre-commit, but for now it is not even necessary to manually add the headers for new files.

danieleongari:
Hi, 
what is the status of this PR?
We are looking it forward because it would be very helpful for our user case! Thanks!

chrisjsewell:
I’ll finish it off next week 👍

aiida/backends/tests/tools/groups/test_grouppath.py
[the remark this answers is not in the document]
chrisjsewell: fixed in 1f6dc82

aiida/backends/tests/tools/groups/test_grouppath.py
```diff
@@ -0,0 +1,171 @@
...
+    assert group_path['a'].group is not None
+    assert group_path['a'].get_or_create_group()[1] is False
+    assert len(group_path) == 1
+    assert [(c.path, c.is_virtual) for c in group_path.children] == [('a', False)]
+    child = next(group_path.children)
+    assert child.parent == group_path
+    assert len(child) == 3
+    assert [(c.path, c.is_virtual) for c in sorted(child)] == [('a/b', False), ('a/c', True), ('a/f', False)]
+    assert [c.path for c in sorted(group_path.walk())] == ['a', 'a/b', 'a/c', 'a/c/d', 'a/c/e', 'a/c/e/g', 'a/f']
+    assert not group_path['a'].is_virtual
```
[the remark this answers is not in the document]
chrisjsewell: Split up tests in 1f6dc82

aiida/backends/tests/tools/groups/test_grouppath.py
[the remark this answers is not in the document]
chrisjsewell: Split up tests in 1f6dc82

aiida/backends/tests/tools/groups/test_grouppath.py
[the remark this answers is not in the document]
chrisjsewell: fixed in 1f6dc82

aiida/tools/groups/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+# pylint: disable=wildcard-import,undefined-variable
+"""Provides tools for interacting with AiiDA Groups."""
+from __future__ import division
+from __future__ import print_function
+from __future__ import absolute_import
```
sphuber, line 11: remove lines

aiida/tools/groups/grouppaths.py
```diff
@@ -0,0 +1,321 @@
...
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Provides functionality for managing large numbers of AiiDA Groups, via label delimitation."""
+from __future__ import division
+from __future__ import print_function
+from __future__ import absolute_import
+
```
sphuber, line 14: can be removed

chrisjsewell: done in 1e9caab0217ba9eaa43153d7b47a56999d4dfebc

aiida/tools/groups/__init__.py
chrisjsewell: done in 1e9caab0217ba9eaa43153d7b47a56999d4dfebc

aiida/tools/groups/grouppaths.py
chrisjsewell: > It does require you to "load the group", right? Is this as expensive as loading nodes?

Yes and loading any database ORM entity is relatively slow, because of the additional logic that django/sqlalchemy does to load it into the session

In d6e5f08, `GroupNotFoundError`/`GroupNotUniqueError` are now raised.

> I think there would be value in at least wrapping this procedure around a 'group_exists' method

You actually missed an important point here; most paths won't have an associated group. For example, if you have paths `a/b/c` and `a/d/e`, I don't want `GroupPath('a').children` failing because it is not directly associated with a group.

To that effect, in 73183a1e5dae5352d71a37d32014a4b08aa15f9b, for anything involving `children` (`__iter__`, `__len__`, `walk`), a `NoGroupsInPathError` exception will be raised if the group itself is not associated with a group, or there are no descendant paths that are contain groups. Personally I think this is added complication to just checking if the path has zero children, but I'm not that bothered.

For `__getitem__`/`__truediv__` this should definitely not raise an existence error though, otherwise this would raise an error `(GroupPath() / 'x' / 'y' / 'z').get_or_create_group()`

aiida/tools/groups/grouppaths.py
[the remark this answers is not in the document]
chrisjsewell: In 9088575 there is now a `verdi group path ls --no-warn` option which will not omit the warning.

But I'm not implemented the other behaviour, because I think it overcomplicates the situation. For example what if you have `group\\bad` and `group\good`, you're not going to exclude `group` then. So now you need to do some complex look ahead for every child, to decide if they have any valid children of their own and thus whether to exclude them.

I'm pretty user that users that want to use this feature will make all their groups compliant, and the warning is sufficient. But fell free to make a PR of your own after this 😁 if you want to implement it

aiida/tools/groups/grouppaths.py
chrisjsewell: Fixed in c8203d1

chrisjsewell:
All changes made, over to you guys...

aiida/tools/groups/grouppaths.py
```diff
@@ -0,0 +1,351 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
chrisjsewell: Fixed in 6a19a13

aiida/cmdline/commands/cmd_group.py
```diff
@@ -361,3 +361,73 @@ def group_copy(source_group, destination_group):
...
+@verdi_group_path.command('ls')
+@click.argument('path', type=click.STRING, required=False)
+@click.option('-R', '--recursive', is_flag=True, default=False, help='Recursively list sub-paths encountered')
+@click.option('-l', '--long', 'as_table', is_flag=True, default=False, help='List in long format')
+@click.option(
+    '-d', '--with-description', 'with_description', is_flag=True, default=False, help='Show also the group description'
+)
+@click.option(
+    '-g',
+    '--groups-only',
```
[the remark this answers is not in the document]
chrisjsewell: Fixed in 6a19a13

aiida/backends/tests/tools/groups/test_grouppath.py
```diff
@@ -0,0 +1,161 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
chrisjsewell: Fixed in 6a19a13

chrisjsewell:
As discussed, CLI improvements to open in a separate "enhancements" issue, when this PR is merged:

- Add `--depth` CLI option to work with `--recursive`
- Add node count option; parent paths will return a count of all descendant nodes, e.g.

```console
$ verdi group path ls --recursive --count
Path       Nodes
-------  ------------
a                   10
a/b                 4
a/c                 6
``` 

- Option to report paths relative to base path, e.g.

```console
$ verdi group path ls a
a/b
a/c
$ verdi group path ls a --relative
b
c
```

sphuber:
Another thing I just realized @chrisjsewell , in terms of enhancements for the future, is to maybe make the currently hardcoded delimiter in `GroupPath` configurable. Or would this make implementation/behavior to complicated to allow this flexibility?

chrisjsewell:
Nope thats fine, its already a private variable, just needs to be added to the `__init__` args: https://github.com/aiidateam/aiida-core/blob/72e5923876f534a08619202215a7063db1f150ae/aiida/tools/groups/paths.py#L72

chrisjsewell:
Introduced a cyclic import my moving the files??

```
Messages
========

aiida/cmdline/params/options/conditional.py
  Line: 1
    pylint: cyclic-import / Cyclic import (aiida.engine.processes -> aiida.engine.processes.workchains -> aiida.engine.processes.workchains.restart)
```

sphuber:
No I think you just got unlucky and this happened in a previous commit that you merged in. These cyclic-imports are not always trustworthy and are notoriously difficult to track down.

chrisjsewell:
whats the course of action here 😕 ? Do I mess around with imports until it disappears, or do we turn a blind eye 🙈

aiida/tools/groups/paths.py
```diff
@@ -0,0 +1,351 @@
...
+        """Return a child ``GroupPath``, with a new path formed by appending ``path`` to the current path."""
+        return self.__truediv__(path)
+
+    @property
+    def group(self):
+        # type: () -> Optional[orm.Group]
+        """Return the concrete group associated with this path."""
+        try:
+            if self.type_string is not None:
+                return orm.Group.objects.get(label=self.path)
```
sphuber, line 177: I just noticed a potential problem with this implementation. Consider the following
```
Group(label='a', type_string='user').store()
Group(label='a', type_string='auto.import').store()
path = GroupPath('a', type_string=None)
path.group  # This will except because the filters do not uniquely determine a single group.
```
You catch the `NotExistent` case but not the `MultipleObjectsError` that will be thrown in my example. Did you think about this and made the decision to ignore it, or is this something that wasn't discussed during the lifetime of the PR? I am pointing this out here, because quite a bit of the rest of the functionality may depend on this.

To be honest, I am not quite sure what should happen here. Maybe we should make the `type_string` required. In this way, one can only use the `GroupPath` to iterate over a specific type of groups, but at least you know that the `label` will guarantee uniqueness then.

chrisjsewell: No I wouldn't say it has been specifically addressed. I agree that we err on the side of disallowing a `None` `type_string` for now, and this can always be addressed in a later PR.

On a related note, how does your group sub-classing PR discriminate group classes; *via* the type string?

chrisjsewell: see 5141b0d

sphuber: Yes, exactly, the `type_string` is what is used to load the correct sub class. I am just now actually working on updating the `QueryBuilder` such that querying with subclasses works properly and this is also based on the `type_string` which itself is determined by the entry point name. However, working on this made me realize that using `core` as the entry point name for the base `Group` class, is not ideal, since this will never match any groups whose entry point that does not start with `core.`, which is to say all groups from external plugins... Not sure what the best approach is here. Setting the entry point for `Group` base class to empty string is not really possible I think, so I will just have to make an exception for it in the query builder filtering logic.

chrisjsewell:
@ltalirz I note that your `pgsu` package is not yet on Conda. Should this have not been picked up before the merge, with `.github/workflows/test-install.yml`?

aiida/tools/groups/paths.py
chrisjsewell: 👍 
This is the use case, where I could envisage you might want to start using `GroupPath` with multiple `type_string`. But yeh, I don't yet have a concrete solution for how to deal with path clashes in that instance.

sphuber:
> @ltalirz I note that your `pgsu` package is not yet on Conda. Should this have not been picked up before the merge, with `.github/workflows/test-install.yml`?

I think the `test-install.yml` is currently only triggered on merge into `develop`. Leo has already submitted the request for the recipe to be added.

chrisjsewell:
> I think the test-install.yml is currently only triggered on merge into develop

#3892 was a merge into to develop though, wasn't it?

sphuber:
> #3892 was a merge into to develop though, wasn't it?

Yes what I meant was that it the workflow only runs on new commits on `develop`. They are not run for PRs.

chrisjsewell:
ah ok fair enough

aiida/tools/groups/paths.py
chrisjsewell: so in 22ee75a, as discussed, I have changed the `group` property to the `get_group` method. This will allow for the later extension of replacing `type_string` to `type_cls`, at which point the association of a path to only 0 or 1 groups cannot be guaranteed.
Within this extension, all "iterative" methods would still work fine (like `children` and `walk`), but for methods where you directly access a group; `get_group`, delete_group`, `get_or_create_group`, you would then want to add arguments/logic to ensure only one group is selected.

greschd:
Great stuff, thanks @chrisjsewell!

chrisjsewell:
thanks 😁

PR #3616, opened 2019-12-09
Refactor `Computer` and add `--format` option to `verdi computer show`

### LINKED ISSUE #2135: orm.Computer define class constants for all properties
Currently a couple of computer properties have class constants that store the corresponding name e.g. `PROPERTY_WORKDIR = 'workdir'`.  This should be extended for all properties e.g. `mpirun_command`.

### LINKED ISSUE #3521: dump computer/codes in YAML format
Sometimes when using another environment, or setting up a new aiida for a new user, we have to setup the same computers/codes over and over again. For this, loading from YAML is quite practical. 

But I can't find a way to generate these YAML files from an existing installation directly. 
"verdi computer show" output is quite close to yaml and can be converted rather simply, but a "dump" command, or a --yaml switch for show would be useful (maybe I missed something and it's already possible, apologies in this case).

sphuber:
Fixes #2135 and fixes #3521


aiida/cmdline/commands/cmd_code.py
```diff
@@ -245,7 +245,7 @@ def code_list(computer, input_plugin, all_entries, all_users, show_owner):
 
     qb_computer_filters = dict()
     if computer is not None:
-        qb_computer_filters['name'] = computer.name
+        qb_computer_filters['name'] = computer.lbael
```
ltalirz:
```suggestion
        qb_computer_filters['name'] = computer.label
```

ltalirz: not sure whether this is also in other places (I guess not?) - a quick search should show

aiida/orm/authinfos.py
```diff
@@ -130,21 +128,13 @@ def get_workdir(self):
         try:
             return self.get_metadata()[self.PROPERTY_WORKDIR]
         except KeyError:
-            return self.computer.get_workdir()
+            return self.computer.get_property('work_dir')
```
ltalirz: The property seems to be called `workdir` - at least on profile (on your branch) I get:

```
In [1]: c=load_computer(1)

In [2]: c.get_metadata()
Out[2]:
{'shebang': '#!/bin/bash',
 'workdir': '/scratch/{username}/aiida_run/',
 'append_text': ' ',
 'prepend_text': ' ',
 'mpirun_command': ['srun', '-n', '{tot_num_mpiprocs}'],
 'default_mpiprocs_per_machine': 28}
```
would this be broken now?

ramirezfranciscof:
@ltalirz is this your review or just a few comments? Just to know if after @sphuber answers this would be ready to be merged or further reviewing is needed.

ltalirz:
This was my review of the PR (but it does contain a couple of questions).

ramirezfranciscof:
Yes yes, I was just checking if after @sphuber addressed them it could be merged or further full reviewing was needed.

sphuber:
This needs quite a bit more discussion. The goal was to homogenize the interface of the `Computer` class, to reduce the number of defined methods and to provide a way to programmatically determine what "properties" define a `Computer`. This is necessary to implement a sort of export functionality that can export a computer in `yaml` format that can then be consumed again by `verdi computer setup --config`. However, I am not happy with the current solution. Still not sure what the best way is to do this

ramirezfranciscof:
Would it help to bring it for discussion during the next weekly meeting? If it needs more than that and we are not sure when we'll be able to deal with it maybe we should close the PR?

sphuber:
Since I think this requires quite a bit more discussion and work, I will close this for now

ltalirz:
Just one comment in case someone wants to pick this up again: I think fixing https://github.com/aiidateam/aiida-core/issues/3521 (i.e. adding the yaml export) is quite straightforward.
What needs discussion is which methods etc. to expose on Computers.

sphuber:
> Just one comment in case someone wants to pick this up again: I think fixing #3521 (i.e. adding the yaml export) is quite straightforward.
> What needs discussion is which methods etc. to expose on Computers.

An ad-hoc implementation would indeed be straightforward, but this would most likely hard code which properties need to be dumped, how to fetch them from a computer instance and to what keys they should be written.

The whole reason why my attempt became so much more involved was because I was exactly trying to find a way to avoid this, for which a significant redesign of these ORM classes is necessary

PR #3618, opened 2019-12-09
Re-add support for Coverage reports through codecov.io

dev-zero:


.github/workflows/tests.sh
```diff
@@ -8,6 +8,14 @@ verdi daemon start 4
 verdi -p test_${AIIDA_TEST_BACKEND} run .ci/test_daemon.py
 verdi daemon stop
 
+# Make sure
+# - omissions are taken from the top-level .coveragerc
+# - coverage is reported as XML and in terminal including
+#   the numbers/ranges of lines which are not covered,
+# - coverage results of multiple tests are collected,
```
CasperWA: Are you sure this is the case?

dev-zero: this is a description of the specified flags and they seem to be all respected based on the test output

CasperWA:
Info: I added the secret CODECOV_TOKEN.

CasperWA:
@dev-zero will you create an issue in codecov/codecov-action for the failure or maybe reopen codecov/codecov-action#22 ?

.github/workflows/tests.sh
```diff
@@ -8,6 +8,14 @@ verdi daemon start 4
...
+# - the base directory for files to consider is aiida/
+export PYTEST_ADDOPTS="--cov-config=.coveragerc --cov-report xml --cov-report term-missing --cov-append --cov=aiida"
+
 verdi -p test_$AIIDA_TEST_BACKEND devel tests
```
ltalirz: is this at all included in the coverage report?
This is not using pytest.
Please check how this was run before (probably using `coverage run`)

CasperWA: These settings are specific for the `pytest-cov` extension for `pytest` as far as I know?

CasperWA:
It seems we need to update to `codecov-action` v1.0.5 according to [this issue](https://github.com/codecov/codecov-action/issues/43).

dev-zero:
@CasperWA https://community.codecov.io/t/whitelist-github-action-servers-to-upload-without-a-token/491/5 ... Travis doesn't need a token because Codecov/Coveralls can verify the request via an API, but for GHA there's no API yet

dev-zero:
Short summary: Coverage through codecov or coveralls is not possible until GitHub releases on API to query GitHub Actions such that codecov/coveralls can verify the uploader. Supposed to happen in Q1 2020. The secret token is for obvious reasons not available in a PR, only in builds initiated by direct commits on the repo (currently restricted to `develop` and `master`).

We have two options for now: disable coverage upload for pull requests or not fail the CI when uploading to coverage report fails.

ltalirz:
> We have two options for now: disable coverage upload for pull requests or not fail the CI when uploading to coverage report fails.

I think it doesn't really matter.
If you go for the second option, then things might start to "magically" work at some point, so perhaps that's an argument for the second option, if there is an easy way to implement it.

CasperWA:
Testing this in the branch [`feature/codecov-integration-PR-3618`](https://github.com/aiidateam/aiida-core/tree/feature/codecov-integration-PR-3618), [this](https://codecov.io/gh/aiidateam/aiida-core/branch/feature%2Fcodecov-integration-PR-3618) is the uploaded coverage report.

As seen in the codecov bot comment here, the total coverage is 70.87 % - it doesn't seem to catch a lot of the database migration files/tests.

Furthermore, anything that isn't in the realm of `pytest`, i.e. `.ci/test_daemon.py`, as well as `.ci/test_plugin_testcase.py`, will not be covered (I believe).

ltalirz:
It would be great if we could move forward with this.
Does anyone have an idea why `coverage` would suddenly stop noticing migration-related tests?

giovannipizzi:
I think it's just because we are asking coverage to omit the whole folder (together with some more) in the `.coveragerc`:
https://github.com/aiidateam/aiida-core/blob/557a6a82935d1d9c93d7bf6c5cc986a2c01d18e6/.coveragerc#L3

I'm not sure anymore why this was introduced in the first place (in any case, it was introduced when migrations were untested).

Maybe we don't need to omit anything? You might want to just push a commit removing the omit line and check how things change.

ltalirz:
Great, thanks a lot - will look into this now

ltalirz:
Removing the omissions specified in .coveragerc [brings back coverage on migrations](https://codecov.io/gh/aiidateam/aiida-core/branch/codecov-integration-PR-3618).
Test coverage now reported at 77%.

To me, this PR is good to merge - does anybody still want to give another look?
One question to me is whether we really need the [coverage output](https://github.com/aiidateam/aiida-core/runs/482200036#step:9:1985) in the travis logs (it's about 500 lines)...
Anyhow, perhaps it's useful to keep at least until codecov starts working on PRs.

Finally, Casper is right that there are still two tests that don't report coverage: 
.ci/test_daemon.py, as well as .ci/test_plugin_testcase.py

I'll have a look into moving the daemon test to pytest (in a separate PR) - that should not be a big deal.
The plugin testcase is explicitly testing a unittest test runner, and so there is not much point in moving it - we could still run it via `coverage run ...` but then we need to make sure to enable the same options. Doable as well, but not very important.

CasperWA:
> Removing the omissions specified in .coveragerc [brings back coverage on migrations](https://codecov.io/gh/aiidateam/aiida-core/branch/codecov-integration-PR-3618).
> Test coverage now reported at 77%.

Good catch on this @giovannipizzi and @ltalirz !

> Finally, Casper is right that there are still two tests that don't report coverage:
> .ci/test_daemon.py, as well as .ci/test_plugin_testcase.py
> 
> I'll have a look into moving the daemon test to pytest (in a separate PR) - that should not be a big deal.
> The plugin testcase is explicitly testing a unittest test runner, and so there is not much point in moving it - we could still run it via `coverage run ...` but then we need to make sure to enable the same options. Doable as well, but not very important.

If it's easily done, it wouldn't hurt? But if it doesn't change the coverage, there is no reason to do this.

sphuber:
> If it's easily done, it wouldn't hurt? But if it doesn't change the coverage, there is no reason to do this.

It _should_ increase the coverage because for example the `BaseRestartWorkChain` class is only tested in `test_daemon.py`. This goes for quite some other classes like `ArithmeticAddCalculation` and `TemplatereplacerCalculation` although they are less critical since they mostly exist for testing and demonstration purposes.

CasperWA:
> > If it's easily done, it wouldn't hurt? But if it doesn't change the coverage, there is no reason to do this.
> 
> It _should_ increase the coverage because for example the `BaseRestartWorkChain` class is only tested in `test_daemon.py`. This goes for quite some other classes like `ArithmeticAddCalculation` and `TemplatereplacerCalculation` although they are less critical since they mostly exist for testing and demonstration purposes.

All right - I'll look into doing this - unless anyone else is already on top of it? (@ltalirz, @dev-zero ...)

ltalirz:
@sphuber Not sure whether what I wrote was clear - I meant I will move the `test_daemon` to pytest, while the `test_plugin_testcase` can also be run via coverage (but is not very important).

@CasperWA If you prefer to do it instead, go ahead!
I'd still suggest to do it in a separate PR.

CasperWA:
> @CasperWA If you prefer to do it instead, go ahead!
> I'd still suggest to do it in a separate PR.

Ah right! Yeah, sure, okay. I've already started some work on it now, so I'll continue, but can surely make a new PR for it. I'll just squash our contributions into 3 separate commits and then you're free to re-approve and merge.

CasperWA:
To note: Since codecov still does not support token-less uploads of coverage reports via GitHub Actions CI, the coverage will only be updated for a PR when it is merged (unless it's coming from an `aiidateam` branch already).

ltalirz:
Yes, that's ok. I let you merge - not sure whether you preferred not to squash here...

CasperWA:
> Yes, that's ok. I let you merge - not sure whether you preferred not to squash here...

I already squashed, but for multiple contributions we usually keep a commit per contributor, right? (Calling @sphuber).

sphuber:
> > Yes, that's ok. I let you merge - not sure whether you preferred not to squash here...
> 
> I already squashed, but for multiple contributions we usually keep a commit per contributor, right? (Calling @sphuber).

If the contributions are significant, yes, I typically keep them. Sometimes, if an additional commit is really small though, one can opt to use `Co-Authored-By: John Doe <email@address>` for the contributors at the end of the commit message of the squashed commit

CasperWA:
I went through them, and they all add the right amount of sugar - so they all stay :smiley:

PR #3622, opened 2019-12-09
Cache entrypoints in group

ltalirz:
Entry point loading is already cached at the `reentry` level but getting
all entry points within a group can still take significant amount of
time.
This commit introduces a simple cache at the AiiDA level. An alternative
would be to add a cache at the reentry level.

To provide some context: The timings on a query for 300 Dict nodes are
as follows:

 * No cache: ~110 ms
 * Cache: 67ms
 * Cache at `load_node_class` level: 58ms

@muhrin I would say the savings are significant and probably worth reaping. Of course, one could also try to do this at the reentry side...
Let me know. If you think we keep the cache here, I'll add the proper docstrings etc.

### DIFF
--- a/aiida/plugins/entry_point.py
+++ b/aiida/plugins/entry_point.py
@@ -10,6 +10,7 @@
 
 import enum
 import traceback
+import functools
 
 try:
     from reentry.default_manager import PluginManager
@@ -185,7 +186,6 @@ def load_entry_point_from_string(entry_point_string):
     group, name = parse_entry_point_string(entry_point_string)
     return load_entry_point(group, name)
 
-
 def load_entry_point(group, name):
     """
     Load the class registered under the entry point for a given name and group
@@ -234,6 +234,7 @@ def get_entry_point_names(group, sort=True):
     return entry_point_names
 
 
+@functools.lru_cache(maxsize=None)
 def get_entry_points(group):
     """
     Return a list of all the entry points within a specific group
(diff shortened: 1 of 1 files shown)


aiida/plugins/entry_point.py
```diff
@@ -50,6 +50,28 @@ class EntryPointFormat(enum.Enum):
     MINIMAL = 3
 
 
+class EntryPointCache():
```
greschd: I think this might be equivalent to adding the ``@functools.lru_cache(maxsize=None)`` decorator to the ``get_entry_points`` function.

muhrin: Good point @greschd , yes this would be a better way to do it

ltalirz: Thanks, that seems to be much more elegant

ltalirz:
The `lru_cache` seems to be a bit slower for some reason
```
In [4]: %timeit qb2.all()  # with project='*'
10 loops, best of 5: 82 ms per loop
```
There is a C implementation of the cache: https://pypi.org/project/fastcache/
Anyhow, I think it doesn't really matter that much here - we have some speedup from the `lru_cache` which adds two lines of code. I'd be fine with that.

muhrin:
So you want to go with the `lru_cache`?  If not there are is a small change I would make to the PR as it stands

PR #3623, opened 2019-12-09
Add 'verdi node repo dump' command.

### LINKED ISSUE #1121: Add 'verdi node copyfiles' command
For inspecting results from the command-line, it would be nice to have an easy way to see output files, especially for ``SinglefileData`` or ``FolderData``. What do you think about adding a ``verdi node gotorepository`` command, which takes you to the node's repository folder. 

Or does this collide with the plans to add an object store backend for the repository? I could also see simply printing something like "This feature is not supported with the object-store repository." in that case.

greschd:
Fixes #1121.

The command takes a node pk / uuid, zero or more file names, and an output directory as arguments. By default, it will create a sub-directory with the node uuid in the output directory, and copy the files with the given names from the repository to that directory. If no file names are given, all objects in the repository are copied.

The command has two optional flags:
 - ``--no-uuid``: copies the files directly into the output directory, without
            uuid prefix
- ``--force``: Copy files even if they already exist at the target destination

Changed from the discussion with @ltalirz: For specifying the source file names, I am now not using a separate flag, since that mimics how regular ``cp`` works. For example:

```bash
verdi node repo cp 2323 _scheduler-stderr.txt _scheduler-stdout.txt output_dir 
```

Click help:
```
Usage: verdi node repo cp [OPTIONS] NODE [SRC_FILES]... OUTPUT_DIRECTORY

  Copy the repository files of a node to an output directory.

Options:
  --no-uuid    Do not prefix the output paths with the node uuid.
  -f, --force  Do not ask for confirmation.
  -h, --help   Show this message and exit.
```


greschd:
Following the feedback session, made the following changes:

- Changed the name of the command to ``verdi node repo dump``. 
- The option to pass specific file names is removed
- The UUID prefix is removed (might be added later as an option, if need arises)
- When a file or directory exists already, prompt the user with the following options:
  - ``[a]`` abort the command
  - ``[s]`` skip the current file / directory
  - ``[m]`` (for directories only) merge contents
  - ``[o]`` overwrite content - in the case of directories removing the existing content
- A ``--force`` option can be passed to always overwrite.

Current click help:
```
Usage: verdi node repo dump [OPTIONS] NODE OUTPUT_DIRECTORY

  Copy the repository files of a node to an output directory.

Options:
  -f, --force  Overwrite existing directories and files without prompting for
               confirmation.
  -h, --help   Show this message and exit.
```

greschd:
Mentioning @giovannipizzi @ltalirz for feedback.

giovannipizzi:
The described behaviour looks fine to me thanks!

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,87 @@ def repo_ls(node, relative_path, color):
...
+
+                _copy_tree(key=file_path, output_dir=new_out_dir)
+            else:
+                out_file_path = (output_dir / file.name)
+                if out_file_path.exists():
+                    if force:
+                        prompt_res = 'o'
+                    else:
+                        prompt_res = click.prompt(
+                            'File {} exists. Do you wish to abort [a], skip this file [s], or overwrite the file [o]?'.
```
ltalirz: This line is rather long, perhaps better like this?
```suggestion
                            "File '{}' exists. Abort [a], skip [s], or replace [r]?".
```

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,87 @@ def repo_ls(node, relative_path, color):
...
+            else:
+                file_path = key + '/' + file.name
+            if file.type == FileType.DIRECTORY:
+                new_out_dir = output_dir / file.name
+                if new_out_dir.exists():
+                    if force:
+                        prompt_res = 'o'
+                    else:
+                        prompt_res = click.prompt(
+                            "Directory '{}' exists. Do you wish to abort [a], "
```
ltalirz:
```suggestion
                            "Directory '{}' exists. Abort [a], "
```

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,87 @@ def repo_ls(node, relative_path, color):
...
+                            'skip this directory [s], merge directory contents [m], '
```
ltalirz:
```suggestion
                            'skip [s], merge contents [m], '
```

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,87 @@ def repo_ls(node, relative_path, color):
...
+                            'or overwrite the directory (delete current contents) [o]?'.format(new_out_dir),
```
ltalirz:
```suggestion
                            'or replace [r]?'.format(new_out_dir),
```
If we use "replace [r]" instead of "overwrite" we don't need the clarification about the contents.

greschd:
Since there was quite some discussion, maybe we should show the updated interface again?

ltalirz:
Sure, we can do it briefly this evening. I think the others will approve.

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,86 @@ def repo_ls(node, relative_path, color):
...
+    required=True,
+)
+@with_dbenv()
+def repo_dump(node, output_directory, force):
+    """Copy the repository files of a node to an output directory."""
+    from aiida.orm.utils.repository import FileType
+
+    output_directory = pathlib.Path(output_directory)
+
+    output_directory.mkdir(exist_ok=True)
```
ltalirz: so, does it need to exist or not? In the option I see `exists=True`
I'm not sure we want to be creating new base directories...

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,86 @@ def repo_ls(node, relative_path, color):
...
+
+    def _copy_tree(key, output_dir):  # pylint: disable=too-many-branches
```
ltalirz: perhaps add a basic docstring

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,86 @@ def repo_ls(node, relative_path, color):
...
+                        new_out_dir.mkdir()
+                    else:
+                        assert prompt_res == 'm'
+                else:
+                    new_out_dir.mkdir()
+
+                _copy_tree(key=file_path, output_dir=new_out_dir)
+            else:
+                assert file.type == FileType.FILE
+                out_file_path = (output_dir / file.name)
```
ltalirz: remove brackets?

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,86 @@ def repo_ls(node, relative_path, color):
...
+                    if prompt_res == 'a':
+                        raise click.Abort
+                    elif prompt_res == 's':
+                        continue
+                    else:
+                        assert prompt_res == 'r'
+                        out_file_path.unlink()
+                with node.open(file_path, 'rb') as in_file:
+                    with out_file_path.open('wb') as out_file:
+                        out_file.write(in_file.read())
```
ltalirz: does `read()` first load the whole `in_file` into memory?

greschd: fixed

aiida/cmdline/commands/cmd_node.py
greschd: yes, fixed

aiida/cmdline/commands/cmd_node.py
greschd: done

aiida/cmdline/commands/cmd_node.py
greschd: removed the ``mkdir``, indeed this is checked by click.

greschd:
@giovannipizzi, do you want to have a final look?

giovannipizzi:
Thanks @greschd ! @sphuber was a bit doubtful on whether to keep the complex logic for merging, or just limit the possibility to create a new folder, specified on the cmd line by the user. Did you already discuss with him?

greschd:
I did not, but I would be ok with removing the "merge" option and only leave abort / skip / overwrite.

sphuber:
Sorry, I did not take this up yet with @greschd . Apologies for coming with this after you have already implemented it but I was wondering if it is not a simpler/safer to make the target directory into which the files should be dumped a required argument _and_ require that it cannot exist. If the goal of the command is to be as simple as possible and just dump the entire content of a node's repo this seems in line with that goal. I think for the same reason we decided not to provide flags to include/exclude certain paths. @ltalirz pointed out that it prevents one from dumping in the current directory, but my argument would be that doing `mkdir some_folder` just before is a lot easier then potentially responding to a bunch of prompts. I am afraid that the current solution provides unnecessary functionality that introduces quite some code complexity which is difficult to fully test and may provide surprising behavior. @giovannipizzi agreed with my suggestion to simplify the code/interface when we discussed this on Friday @Fiesch. I realize that you have already invested quite some time in writing the code and it would be unfortunate if we decide not to include it. Let me know what you think about simplifying the command

greschd:
>[...] I was wondering if it is not a simpler/safer to make the target directory into which the files should be dumped a required argument _and_ require that it cannot exist. 

So the proposed syntax would be ``verdi node repo dump <ID> <PATH>`` where ``PATH`` does not exist, but its parent does, correct? Or would you also auto-create the parents?

> [...] that doing `mkdir some_folder` 

which means the user _wouldn't_ have to `mkdir` the folder, right?

> I realize that you have already invested quite some time in writing the code and it would be unfortunate if we decide not to include it. Let me know what you think about simplifying the command

No worries, I'm not particularly attached to my code - that also doesn't make a particularly strong technical argument 😄 I'm fine with making this as simple as possible.

sphuber:
I think it is fine to automatically create a nested folder, as long as the leaf folder does not already exist. This is just to prevent copying files by accident into an already existing folder. So interface:
```
verdi node repo dump <ID> <PATH>
```
with `<ID>` and `<PATH>` both required and `<PATH>` can be a relative nested path where the leaf folder should _not_ exist but any sub paths can exist and if not will be created on the fly. The contents of the node repository are then copied wholesale into the leaf folder of `<PATH>` (omitting the still present but soon to be removed `path` or `raw_input` sub paths in the node repo).

But before we go ahead, lets get an OK on this proposal from @giovannipizzi and @ltalirz and yourself :)

greschd:
Good for me. The `path` and `raw_input` are omitted by virtue of going through the ``list_objects`` interface.

giovannipizzi:
I would not have created intermediate paths, but I do not see a major problem with this, so I'm ok with Sebastiaan's proposal.
Just to be clear, I agree that the simplest is to stop if the provided path already exists.

greschd:
> Just to be clear, I agree that the simplest is to stop if the provided path already exists.

Yes, that's the intended behavior. In principle this could be handled by ``click``, by setting ``type=click.Path(exists=False, file_okay=False, dir_okay=False)``. However, the resulting error message
```
Usage: verdi node repo dump [OPTIONS] NODE OUTPUT_DIRECTORY
Try "verdi node repo dump -h" for help.

Error: Invalid value for "OUTPUT_DIRECTORY": Path "." is a directory.
```
is not very clear IMO, so I'll instead handle it directly and write 
```
Error: Invalid value for "OUTPUT_DIRECTORY": Path "." exists.
Aborted!
```

greschd:
Pre-commit failure seems to be unrelated?
```
Messages
========

aiida/schedulers/plugins/slurm.py
  Line: 586
    pylint: invalid-name / Attribute name "requested_wallclock_time_seconds" doesn't conform to '(([a-z][a-z0-9_]{2,30})|(_[a-z0-9_]*))$' pattern (col 16)
```

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,52 @@ def repo_ls(node, relative_path, color):
...
+
+    def _copy_tree(key, output_dir):  # pylint: disable=too-many-branches
+        """
+        Recursively copy the content at the ``key`` path in the given node to
+        the ``output_dir``.
+        """
+        for file in node.list_objects(key=key):
+            # Not using os.path.join here, because this is the "path"
+            # in the AiiDA node, not an actual OS - level path.
+            file_key = file.name if not key else key + '/' + file.name
```
sphuber: This is a good point and I now start to wonder whether we actually handle this correctly and consistently everywhere in the repositiory interface. Anyhow, this is good for now.

greschd: Well, that's only going to be a problem if ``os.sep`` is a backslash.. not the trickiest part for Windows support, I suppose 😄

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,52 @@ def repo_ls(node, relative_path, color):
...
+@verdi_node_repo.command('dump')
+@arguments.NODE()
+@click.argument(
+    'output_directory',
+    type=click.Path(),
+    required=True,
+)
+@with_dbenv()
+def repo_dump(node, output_directory):
+    """Copy the repository files of a node to an output directory."""
```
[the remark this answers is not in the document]
greschd: Done! I did notice though that this doesn't appear anywhere in the documentation, only the top-level command help is shown. That should probably be improved, we could show the help for all command groups.

aiida/cmdline/commands/cmd_node.py
```diff
@@ -61,6 +64,52 @@ def repo_ls(node, relative_path, color):
...
+    from aiida.orm.utils.repository import FileType
+
+    output_directory = pathlib.Path(output_directory)
+
+    try:
+        output_directory.mkdir(parents=True, exist_ok=False)
+    except FileExistsError:
+        click.echo('Error: Invalid value for "OUTPUT_DIRECTORY": Path "{}" exists.'.format(output_directory))
```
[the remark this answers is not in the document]
greschd: Done!

PR #3627, opened 2019-12-10
Timeout a CI job after 30 min. (instead of 360)

### LINKED ISSUE #3626: Change the Actions "wall time"
A GH Action job's timout interval can be set using `timeout-minutes` (see [here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes)). The default is 360 minutes (6 hours).

CasperWA:
FIxes #3626

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,6 @@
 name: aiida-core
 
-on:
-  pull_request:
-  push:
-  - master
-  - develop
+on: [push, pull_request]
 
 jobs:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 name: aiida-core
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+  - master
+  - develop
 
 jobs:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,11 +2,15 @@ name: aiida-core
 
 on: [push, pull_request]
 
+env:
+  TIMEOUT_MIN: 30
+
 jobs:
 
   conda:
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


ltalirz:
Thanks @CasperWA !
Just one question: since you're setting the same timeout for each job, is there a way to set it globally?

CasperWA:
> Just one question: since you're setting the same timeout for each job, is there a way to set it globally?

I spent the first ... 3 ... commits attempting to do this. Unfortunately, it doesn't seem to be the case. The only tool I could find to set this, was the global `env`, but it doesn't seem to be callable in GH contexts (even though it is mentioned [here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#contexts)).

CasperWA:
> > Just one question: since you're setting the same timeout for each job, is there a way to set it globally?
> 
> I spent the first ... 3 ... commits attempting to do this. Unfortunately, it doesn't seem to be the case. The only tool I could find to set this, was the global `env`, but it doesn't seem to be callable in GH contexts (even though it is mentioned [here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#contexts)).

Since we cannot (at the moment) set a global time-out variable, it might make sense to utilize it to set more appropriate time-out values for each separate job? E.g., the test-jobs take on average little less than 15 min., whereas other jobs take much less time. So it might make sense to lower the time-out minutes for them to, e.g., 15 min.?

ltalirz:
> E.g., the test-jobs take on average little less than 15 min., whereas other jobs take much less time. So it might make sense to lower the time-out minutes for them to, e.g., 15 min.?

You can, but in practice it's anyhow just the test jobs that are prone to getting stuck.
I think it's fine as it is.

PR #3629, opened 2019-12-10
Adding documentation on performance tips

### LINKED ISSUE #3603: Write documentation with tips on optimizing interaction with clusters
There are many ways too overload the scheduler of a cluster and AiiDA provides a number of ways to reduce this. We should add a page to the docs detailing how these can be tuned.

giovannipizzi:
Some parts of the documentation on how to maximise performance
of AiiDA were already there. I am now adding additional useful
information originating from our experience, that is going to be
useful to users.

I am also adding suggestions from the CSCS staff on how to optimise/tune
the performance of SLURM, that might be relevant information that AiiDA
users might forward to cluster administrators.

Fixes #3603


sphuber (approved):
Muchas gracies @giovannipizzi

Issue #2207, opened 2018-11-16
Reduce/remove files from the exclusion list of pre-commit, prospector, yapf, pylint


giovannipizzi:
We now still have a quite long blacklist of files not to be checked by pre-commit hooks (yapf, prospector).
Most of them only have 1/2 things to fix.
A simple script to check how many errors there are is the following (run it in a py3 venv with prospector installed):
```python
import yaml, json, os
from subprocess import run, DEVNULL, PIPE

with open('.pre-commit-config.yaml') as f:
    data = yaml.load(f)

assert data[0]['hooks'][0]['id'] == 'yapf'
files = [_.strip() for _ in data[0]['hooks'][0]['exclude'].split('(')[2].split(')')[0].replace('\n', '').split("|") if '*' not in _]

for file in files:
    if not os.path.isfile(file):
        print("REMOVE ->", file)

print("*"*72)
for file in files[:2]:
    response = run(['prospector', '-o', 'json', file], stdout=PIPE, stderr=DEVNULL, encoding='utf8')
    response_data = json.loads(response.stdout)
    print("{} {}".format(file, response_data['summary']['message_count']))
```
(of course it can be changed to also show the specific errors etc.)

The idea is that we can already remove and fix most of those with few problems.
As once put in there they will always be checked and must be ok, it's better to first try to remove as many as possible of those that are simple to fix, and leave the complicated ones at the end.

giovannipizzi:
Quick statistics out of my current branch (columns: line number, filename, number of messages);
```
1 aiida/backends/djsite/db/subtests/djangomigrations.py 
2 aiida/backends/tests/control/test_computer_ctrl.py 
3 aiida/backends/tests/work/utils.py 
4 aiida/orm/authinfo.py 
5 aiida/orm/backend.py 
6 aiida/orm/implementation/calculation.py 
7 aiida/orm/node.py 
8 utils/create_requirements.py 
9 .ci/test_fixtures.py 0
10 aiida/backends/tests/cmdline/commands/test_devel.py 0
11 aiida/backends/tests/cmdline/commands/test_export.py 0
12 aiida/backends/tests/cmdline/commands/test_work.py 0
13 aiida/calculations/plugins/__init__.py 0
14 aiida/control/tests/test_postgres.py 0
15 aiida/parsers/plugins/templatereplacer/__init__.py 0
16 aiida/scheduler/__init__.py 0
17 setup.py 0
18 aiida/backends/__init__.py 1
[...]

PR #3636, opened 2019-12-10
Fix two issues in `verdi computer configure ssh`

### LINKED ISSUE #3633: Empty key filename is not accepted if `verdi computer configure ssh`
This should be ok as it should not be properly dealt with (see e.g. #3588)

### LINKED ISSUE #3634: Default value for cooldown time incorrectly shown in the default interactive prompt
The value that is shown when reconfiguring a computer is always the class default rather than the value set in the authinfo. So that reconfiguring a computer and just pressing enter would reset the default value.
The current value should be instead be shown.

giovannipizzi:
The interactive SSH was not accepting an empty
ssh key_filename, while this should be acceptable.
I am also adding a test for the interactive setup of
SSH computers.

Moreover, when reconfiguring a computer, the value
of the cooldown time wasn't resued from the authinfo;
instead, the class default was also reused due to some
hardcoding. Code now behaves correctly and is simplified
avoid duplication in base classes and subclasses.

Fixes #3633 and fixes #3634

### DIFF
--- a/aiida/transports/transport.py
+++ b/aiida/transports/transport.py
@@ -18,8 +18,6 @@ from collections import OrderedDict
 from aiida.common.exceptions import InternalError
 from aiida.common.lang import classproperty
 
-DEFAULT_TRANSPORT_INTERVAL = 30.
-
 __all__ = ('Transport',)
 
 
@@ -40,7 +38,9 @@ class Transport(abc.ABC):
     """Abstract class for a generic transport (ssh, local, ...) ontains the set of minimal methods."""
     # pylint: disable=too-many-public-methods
 
-    _DEFAULT_SAFE_OPEN_INTERVAL = DEFAULT_TRANSPORT_INTERVAL
+    # This is used as a global default in case subclasses don't redefine this,
+    # but this should  be redefined in plugins where appropriate
+    _DEFAULT_SAFE_OPEN_INTERVAL = 30.
 
     # To be defined in the subclass
     # See the ssh or local plugin to see the format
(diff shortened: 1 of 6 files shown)


giovannipizzi:
Note for reviewers: there are two commits - the first one does the job, the second is just fixing issues in pylint, so you might want to check them independently.

Note for @yakutovicha who's working on #2207: I'm removing two more files (actually two complex ones) from the pylint exclude list.

aiida/transports/transport.py
```diff
@@ -185,24 +185,16 @@ def auth_options(cls):  # pylint: disable=no-self-argument
...
-        for option in cls._common_auth_options:
-            if option[0] == 'safe_interval':
-                option[1]['default'] = cls._DEFAULT_SAFE_OPEN_INTERVAL
         return OrderedDict(cls._valid_auth_options + cls._common_auth_options)
 
     @classmethod
     def _get_safe_interval_suggestion_string(cls, computer):  # pylint: disable=unused-argument
         """
-        Default time in seconds between consecutive checks.
+        Return the currently set safe interval.
```
sphuber: I find this confusing because it suggests this is set for a transport instance, but instead it is just for the class. Maybe something like: "Return the default safe interval for this transport class." The second phrase then is not necessary and also a bit misleading. The value is also used in constructing a transport instance if `safe_interval` is not passed in explicitly

aiida/transports/transport.py
```diff
@@ -185,24 +185,16 @@ def auth_options(cls):  # pylint: disable=no-self-argument
         # The common auth options are currently defined as class members, but the default for `safe_interval` is sub
         # class specific. With the current design the default cannot already be specified directly but has to be added
         # manually here.
-        for option in cls._common_auth_options:
```
sphuber: The comment block above this can now then also be removed

sphuber (changes requested):
Thanks @giovannipizzi just some minor changes

aiida/transports/transport.py
giovannipizzi: I agree on the first part, it was a leftover of a first rework of the code I was doing - I realised later that I could have simplified it more. 
Instead, I don't agree on the second - this is *only* used as a suggestion string, while the default in the init is set directly from the `self._DEFAULT_SAFE_OPEN_INTERVAL`:
https://github.com/aiidateam/aiida-core/blob/f68b304b0c7b04245582f8a6f61413dc73929d29/aiida/transports/transport.py#L64

giovannipizzi:
Fixed the two comments by @sphuber in the first of the two commits, then rebased all on top of the most recent develop and force-pushed. This is now ready for (re)review.

sphuber (approved):

PR #3637, opened 2019-12-10
Add hook and exit_code keyword for more nuanced cache validation.

### LINKED ISSUE #3368: Limit caching to successful calculations?
In the previous AiiDA version, caching checked that a calculation was in FINISHED state before using it as a basis for caching -- the purpose of this being that it allows failed calculations to be re-run.

With the updates to 1.0, the concept of calculation states was overhauled, and now a calculation can be finished even if it failed (with non-zero exit codes).

I believe the caching implementation should be adapted so that it's only triggered when the exit status is zero.

@sphuber @giovannipizzi do you agree with this?

greschd:
Fixes #3368.

Adds a method `is_valid_cache` to the `Process` class, which is called with a `ProcessNode` instance as an argument. This method is called by `ProcessNode.is_valid_cache`. It allows implementing
custom behavior in `Process` sub-classes, such as the `CalcJob` classes implemented in plugins.

As a default, the `Process.is_valid_cache` checks whether the exit code returned from the process has a - newly introduced - `invalidates_cache` attribute set to `True`. This is a simple way of marking the as cache invalid for a specific exit code only.

Question: Should all the documentation go into the "Caching: implementation details" section, or do we want to mention the ``spec.exit_code`` option in the "main" caching section?

### DIFF
--- a/aiida/engine/processes/exit_code.py
+++ b/aiida/engine/processes/exit_code.py
@@ -15,8 +15,8 @@ from aiida.common.extendeddicts import AttributeDict
 
 __all__ = ('ExitCode', 'ExitCodesNamespace')
 
-ExitCode = namedtuple('ExitCode', 'status message')
-ExitCode.__new__.__defaults__ = (0, None)
+ExitCode = namedtuple('ExitCode', ['status', 'message', 'invalidates_cache'])
+ExitCode.__new__.__defaults__ = (0, None, False)
 """
 A namedtuple to define an exit code for a :class:`~aiida.engine.processes.process.Process`.
 
@@ -29,6 +29,9 @@ corresponding attributes of the node.
 
 :param message: optional message with more details about the failure mode
 :type message: str
+
+:param invalidates_cache: optional flag, indicating that a process should not be used in caching
+:type invalidates_cache: bool
 """
 
 
--- a/aiida/engine/processes/process.py
+++ b/aiida/engine/processes/process.py
@@ -894,6 +894,22 @@ class Process(plumpy.Process):
             namespace_list.extend(['.'.join(split_ns[:i]) for i in range(1, len(split_ns) + 1)])
         return namespace_list
 
+    @classmethod
+    def is_valid_cache(cls, node):
+        """Check if the given node can be cached from.
+
+        .. warning :: When overriding this method, make sure to call
+            super().is_valid_cache(node) and respect its output. Otherwise,
+            the 'invalidates_cache' keyword on exit codes will not work.
+
+        This method allows extending the behavior of `ProcessNode.is_valid_cache`
+        from `Process` sub-classes, for example in plug-ins.
+        """
+        try:
+            return not cls.spec().exit_codes(node.exit_status).invalidates_cache
+        except ValueError:
+            return True
+
 
 def get_query_string_from_process_type_string(process_type_string):  # pylint: disable=invalid-name
     """
--- a/aiida/orm/nodes/process/process.py
+++ b/aiida/orm/nodes/process/process.py
@@ -479,7 +479,23 @@ class ProcessNode(Sealable, Node):
 
         :returns: True if this process node is valid to be used for caching, False otherwise
         """
-        return super().is_valid_cache and self.is_finished
+        if not (super().is_valid_cache and self.is_finished):
+            return False
+        try:
+            process_class = self.process_class
+        except Exception as exc:  # pylint: disable=broad-except
+            self.logger.warning(
+                "Not considering {} for caching, '{!r}' when accessing its process class.".format(self, exc)
+            )
+            return False
+        # If the process class does not have an 'is_valid_cache'
+        # method, we do not need to perform this check.
+        try:
+            is_valid_cache_func = process_class.is_valid_cache
+        except AttributeError:
+            return True
+
+        return is_valid_cache_func(self)
 
     def _get_objects_to_hash(self):
         """
--- a/aiida/orm/nodes/process/process.py
+++ b/aiida/orm/nodes/process/process.py
@@ -483,13 +483,13 @@ class ProcessNode(Sealable, Node):
             return False
         try:
             process_class = self.process_class
-        except Exception as exc:  # pylint: disable=broad-except
+        except ValueError as exc:
             self.logger.warning(
                 "Not considering {} for caching, '{!r}' when accessing its process class.".format(self, exc)
             )
             return False
-        # If the process class does not have an 'is_valid_cache'
-        # method, we do not need to perform this check.
+        # For process functions, the `process_class` does not have an
+        # is_valid_cache attribute
         try:
             is_valid_cache_func = process_class.is_valid_cache
         except AttributeError:
--- a/docs/source/developer_guide/core/caching.rst
+++ b/docs/source/developer_guide/core/caching.rst
@@ -23,10 +23,18 @@ Below are some methods you can use to control how the hashes of calculation and 
 Controlling caching
 -------------------
 
-There are two methods you can use to disable caching for particular nodes:
+There are several methods you can use to disable caching for particular nodes:
+
+On the level of generic :class:`aiida.orm.nodes.Node`:
 
 * The :meth:`~aiida.orm.nodes.Node.is_valid_cache` property determines whether a particular node can be used as a cache. This is used for example to disable caching from failed calculations.
 * Node classes have a ``_cachable`` attribute, which can be set to ``False`` to completely switch off caching for nodes of that class. This avoids performing queries for the hash altogether.
+
+On the level of :class:`aiida.engine.processes.process.Process` and :class:`aiida.orm.nodes.process.ProcessNode`:
+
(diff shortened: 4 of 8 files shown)


greschd:
Mentioning also @sphuber, who helped me with the design.

ltalirz:
> Should all the documentation go into the "Caching: implementation details" section, or do we want to mention the spec.exit_code option in the "main" caching section?

Let's put it this way - since the documentation is going to be restructured into tutorials/howtos/topics/reference, this information is certainly not going to go into the "how to enable caching".

So, I would keep it in the implementation section for the moment. We'll decide what to do with this section later.

greschd:
Ready for review now.

aiida/orm/nodes/process/process.py
```diff
@@ -479,7 +479,15 @@ def is_valid_cache(self):
 
         :returns: True if this process node is valid to be used for caching, False otherwise
         """
-        return super().is_valid_cache and self.is_finished
+        if not (super().is_valid_cache and self.is_finished):
+            return False
+        # Catch errors when getting the process_class, or when it does not have
+        # a 'is_valid_cache' method.
+        try:
+            is_valid_cache_func = self.process_class.is_valid_cache
```
sphuber: Can't this also throw other errors, specifically something like `EntryPointError` if the class cannot be loaded?

docs/source/developer_guide/core/caching.rst
```diff
@@ -23,11 +23,19 @@ Below are some methods you can use to control how the hashes of calculation and
...
+There are several methods you can use to disable caching for particular nodes:
+
+On the level of generic :class:`aiida.orm.nodes.Node`:
 
 * The :meth:`~aiida.orm.nodes.Node.is_valid_cache` property determines whether a particular node can be used as a cache. This is used for example to disable caching from failed calculations.
 * Node classes have a ``_cachable`` attribute, which can be set to ``False`` to completely switch off caching for nodes of that class. This avoids performing queries for the hash altogether.
 
+On the level of :class:`aiida.engine.processes.process.Process` and :class:`aiida.orm.nodes.process.ProcessNode`:
+
+* The :meth:`ProcessNode.is_valid_cache <aiida.orm.nodes.process.ProcessNode.is_valid_cache>` calls :meth:`Process.is_valid_cache <aiida.engine.processes.process.Process.is_valid_cache>`, passing the node itself. This can be used in :class:`~aiida.engine.processes.process.Process` subclasses (e.g. in calculation plugins) to implement custom ways of invalidating the cache.
```
sphuber: you can shorthen these `:meth:ProcessNode.is_valid_cache <aiida.orm.nodes.process.ProcessNode.is_valid_cache>` to `:meth:~aiida.orm.nodes.process.ProcessNode.is_valid_cache` which is what we use elsewhere in the documentation

aiida/orm/nodes/process/process.py
```diff
@@ -479,7 +479,15 @@ def is_valid_cache(self):
 
         :returns: True if this process node is valid to be used for caching, False otherwise
         """
-        return super().is_valid_cache and self.is_finished
+        if not (super().is_valid_cache and self.is_finished):
+            return False
```
sphuber: why not
```
if super().is_valid_cache and self.is_finished:
    return True
```

aiida/engine/processes/exit_code.py
```diff
@@ -15,8 +15,8 @@
 
 __all__ = ('ExitCode', 'ExitCodesNamespace')
 
-ExitCode = namedtuple('ExitCode', 'status message')
-ExitCode.__new__.__defaults__ = (0, None)
+ExitCode = namedtuple('ExitCode', ['status', 'message', 'invalidates_cache'])
+ExitCode.__new__.__defaults__ = (0, None, False)
```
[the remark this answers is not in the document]
greschd: Indeed, that is left over from a previous iteration.

aiida/engine/processes/process.py
```diff
@@ -894,6 +894,22 @@ def _get_namespace_list(namespace=None, agglomerate=True):
             namespace_list.extend(['.'.join(split_ns[:i]) for i in range(1, len(split_ns) + 1)])
         return namespace_list
 
+    @classmethod
+    def is_valid_cache(cls, node):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
greschd: Indeed, that is left over from a previous iteration.

docs/source/developer_guide/core/caching.rst
greschd: Tried that initially, but then it just writes "``is_valid_cache()`` calls `is_valid_cache()`` (i.e., the class name is dropped completely. In this case, that doesn't convey the actual meaning.

aiida/orm/nodes/process/process.py
greschd: That's not the same logic -- in the case where the two statements are true, we want to _also_ perform the checks implemented after that.

aiida/orm/nodes/process/process.py
greschd: Hmm yes, indeed it can. It's not quite clear to me what the correct response to that is, though. If the process class can not be loaded due to an entry point error, essentially we can no longer perform the checks that _might_ be implemented in that class. Should that fall back to being a valid cache, or invalid one?

aiida/engine/processes/exit_code.py
```diff
@@ -15,8 +15,8 @@
 
 __all__ = ('ExitCode', 'ExitCodesNamespace')
 
-ExitCode = namedtuple('ExitCode', 'status message')
-ExitCode.__new__.__defaults__ = (0, None)
+ExitCode = namedtuple('ExitCode', ['status', 'message', 'invalidates_cache'])
```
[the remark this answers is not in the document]
greschd: In any case, this should probably be documented in a changelog / breaking changes list for the next release.

What's the correct place to put this information?

aiida/engine/processes/exit_code.py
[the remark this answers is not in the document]
greschd: Done.

aiida/engine/processes/process.py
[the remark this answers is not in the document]
greschd: Done.

aiida/engine/processes/exit_code.py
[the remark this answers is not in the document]
sphuber: Discusses with @giovannipizzi and agreed that it is very unlikely that people would explode the tuple and so for this change to break existing code. So we can go through with it and just make a mention in the commit and then later we add it to the changelog

docs/source/developer_guide/core/caching.rst
sphuber: I see, fine to keep as is then

aiida/engine/processes/exit_code.py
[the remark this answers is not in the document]
greschd: Ok, I updated the commit message.

aiida/orm/nodes/process/process.py
sphuber: good point

aiida/orm/nodes/process/process.py
sphuber: Should we perhaps err on the safe side and consider it to be invalid. If the user really want them to be cached they should simply install the relevant plugin. I am just wondering if in this case we should emit a warning otherwise these calculations may be silently skipped for caching and might be difficult to track down

greschd: So we keep catching ``ValueError`` and return ``True`` (because that also happens in calcfunctions, I think), and for the rest log a warning and return ``False``?

Side note: This seems like a pretty weird edge case -- you're just launching a new process, but then its process class somehow isn't available. How does this even happen? Not saying we shouldn't handle it, though.

greschd: Updated the code now.

aiida/orm/nodes/process/process.py
```diff
@@ -479,7 +479,23 @@ def is_valid_cache(self):
...
+        try:
+            process_class = self.process_class
+        except Exception as exc:  # pylint: disable=broad-except
```
sphuber: Should we not make this a `ValueError`?

greschd: This is the case where the process class might not be loadable (as discussed above). I don't think we know all possible causes for this, do we?

aiida/orm/nodes/process/process.py
greschd: Seems I have misunderstood this before: The ``ValueError`` happens whenever the ``process_class`` can not be loaded.

This clashes with the existing calcfunctions tests not because calcfunctions are inherently problematic, but because those tests define new calcfunctions inside the class. Since these can not be imported, they are not allowed to cache. That could be fixed by moving all calcfunctions to the module level.

Since calcfunctions should anyway be importable for the daemon to find them, I think it is a safe choice to disable caching when they are not importable. It is definitely a change in behavior though, so if you want to be strict on backwards-compatibility, it should be enabled instead.

aiida/orm/nodes/process/process.py
sphuber: But the `process_class` property already catches all these different potential reasons and converts them in a `ValueError`. Of course `process_class` implementation may have missed one but then it is a bug and I'd rather have it shown and fixed then hidden by broad except.

greschd: Never mind, you are right.

greschd:
I've implemented the changes in the ``calcfunctions`` tests - for me this is good to go.

Let me know if you'd rather have the version where inaccessible ``process_class`` returns ``True``, and I'll revert the test changes and implement that instead.

sphuber:
Thanks a lot @greschd

PR #3639, opened 2019-12-10
Move getting completed job accounting to `retrieve` transport task

### LINKED ISSUE #3133: Detailed job info for completed `CalcJobs` gets set to `None`

sphuber:
Fixes #3133 

The task of retrieving the detailed accounting info from the scheduler
for completed jobs was erroneously placed within the scheduler status
update cycle of the `JobsList`. This would have requested the detailed
job info each update cycle where it not for a conditional that checked
that the scheduler status was `DONE`. However, within the `JobsList`
loop, completed jobs would not even appear in the result and hence the
conditional would never be hit. This caused the detailed accounting
never to be retrieved.

A logically better place for this functionality is in the `retrieve`
transport task after the `UPDATE` task, i.e. when the scheduler reports
that the job has terminated. The job accounting will query the scheduler
for the full accounting which, when implemented for the given scheduler
plugin, will be stored as an attribute on the calculation job node.

### DIFF
--- a/aiida/schedulers/scheduler.py
+++ b/aiida/schedulers/scheduler.py
@@ -272,7 +272,7 @@ class Scheduler:
         """
         raise NotImplementedError
 
-    def _get_detailed_jobinfo_command(self, jobid):
+    def _get_detailed_job_info_command(self, job_id):
         """
         Return the command to run to get the detailed information on a job.
         This is typically called after the job has finished, to retrieve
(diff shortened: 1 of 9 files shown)


aiida/schedulers/scheduler.py
```diff
@@ -283,20 +283,43 @@ def _get_detailed_jobinfo_command(self, jobid):
...
+            'stderr': stderr,
+        }
+
+        return detailed_job_info
+
     def get_detailed_jobinfo(self, jobid):
         """
         Return a string with the output of the detailed_jobinfo command.
 
+        .. deprecated:: use the `get_detailed_job_info` instead
```
giovannipizzi: Can we declare when we drop this (2.0) and issue a warning in the code?

giovannipizzi:
Actually, a second comment - here we're writing job_info with an underscore since it's a new key anyway. What to do we do with the old one? Am I allowed to just replace it with an underscore when I change from string to dict?

aiida/schedulers/scheduler.py
sphuber: done

sphuber:
> Actually, a second comment - here we're writing job_info with an underscore since it's a new key anyway. What to do we do with the old one? Am I allowed to just replace it with an underscore when I change from string to dict?

I think that if you change the attribute key, we should add a migration. I am not sure if it is worth the effort but feel free to add it if you prefer the consistency in the attribute keys

PR #3647, opened 2019-12-12
Initial prototype for scheduler parsing

espenfl:
Here is an initial approach which enable basic scheduler parsing. The Slurm scheduler is used as an example. For Slurm, the `State` entry in the `sacct` contains for instance `OUT_OF_MEMORY` when the process breach the set memory limit. As such it makes sense to parse the `sacct` information and in case needed, map that to generic error codes.

There is also a method to parse the `stdout` and `stderr` of the parser, which in this commit does nothing. Currently the error code with the highest `exit_status` number is set (in case there are differences between the `sacct` and file approach.

Currently, exit codes are set on `JobCalc`. We might, in the future consider to move the same framework onto the scheduler framework to allow for specific handling of scheduler error codes without relying on specific definitions in a given `JobCalc`.

A test is also included, which sets up a node and tests the parsing for the `sacct`. We should, when functionality added, also add tests the file parsing.

### DIFF
--- a/aiida/engine/processes/calcjobs/calcjob.py
+++ b/aiida/engine/processes/calcjobs/calcjob.py
@@ -169,6 +169,8 @@ class CalcJob(Process):
         spec.output(cls.link_label_retrieved, valid_type=orm.FolderData, pass_to_parser=True,
             help='Files that are retrieved by the daemon will be stored in this node. By default the stdout and stderr '
                  'of the scheduler will be added, but one can add more by specifying them in `CalcInfo.retrieve_list`.')
+        spec.exit_code(0, 'NO_ERROR', message='no errors')
+        spec.exit_code(50, 'OUT_OF_MEMORY', message='scheduler reports out of memory')
 
     @classproperty
     def spec_options(cls):  # pylint: disable=no-self-argument
--- a/aiida/engine/processes/calcjobs/calcjob.py
+++ b/aiida/engine/processes/calcjobs/calcjob.py
@@ -169,8 +169,6 @@ class CalcJob(Process):
         spec.output(cls.link_label_retrieved, valid_type=orm.FolderData, pass_to_parser=True,
             help='Files that are retrieved by the daemon will be stored in this node. By default the stdout and stderr '
                  'of the scheduler will be added, but one can add more by specifying them in `CalcInfo.retrieve_list`.')
-        spec.exit_code(0, 'NO_ERROR', message='no errors')
-        spec.exit_code(50, 'OUT_OF_MEMORY', message='scheduler reports out of memory')
 
     @classproperty
     def spec_options(cls):  # pylint: disable=no-self-argument
--- a/aiida/orm/nodes/process/process.py
+++ b/aiida/orm/nodes/process/process.py
@@ -296,19 +296,6 @@ class ProcessNode(Sealable, Node):
         """
         return self.get_attribute(self.EXIT_STATUS_KEY, None)
 
-    def set_exit_code(self, exit_code):
-        """
-        Set the exit status and message via the exit code framework.
-
-        :param exit_code: a valid tuple set from an ExitCode instance
-        """
-
-        if isinstance(exit_code, tuple):
-            self.set_exit_status(exit_code[0])
-            self.set_exit_message(exit_code[1])
-        else:
-            raise ValueError('exit code is not a tuple set with ExitCode')
-
     def set_exit_status(self, status):
         """
         Set the exit status of the process
--- a/aiida/orm/nodes/process/process.py
+++ b/aiida/orm/nodes/process/process.py
@@ -296,6 +296,19 @@ class ProcessNode(Sealable, Node):
         """
         return self.get_attribute(self.EXIT_STATUS_KEY, None)
 
+    def set_exit_code(self, exit_code):
+        """
+        Set the exit status and message via the exit code framework.
+
+        :param exit_code: a valid tuple set from an ExitCode instance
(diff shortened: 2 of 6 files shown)


aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -169,6 +169,8 @@ def define(cls, spec):
         spec.output(cls.link_label_retrieved, valid_type=orm.FolderData, pass_to_parser=True,
             help='Files that are retrieved by the daemon will be stored in this node. By default the stdout and stderr '
                  'of the scheduler will be added, but one can add more by specifying them in `CalcInfo.retrieve_list`.')
+        spec.exit_code(0, 'NO_ERROR', message='no errors')
```
sphuber: The `0` is implicit and not necessary to define explicilty

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -169,6 +169,8 @@ def define(cls, spec):
...
+        spec.exit_code(50, 'OUT_OF_MEMORY', message='scheduler reports out of memory')
```
sphuber: The range `100-199` is reserved for generic scheduler/node errors. Let's reserve `100` for the most generic value (if exact reason for failure is unknown) and so maybe use `101` for OOM?

aiida/orm/nodes/process/process.py
```diff
@@ -296,6 +296,19 @@ def exit_status(self):
...
         return self.get_attribute(self.EXIT_STATUS_KEY, None)
 
+    def set_exit_code(self, exit_code):
+        """
+        Set the exit status and message via the exit code framework.
+
+        :param exit_code: a valid tuple set from an ExitCode instance
+        """
+
+        if isinstance(exit_code, tuple):
```
sphuber: please change to `if isinstance(exit_code, ExitCode)`

aiida/orm/nodes/process/process.py
```diff
@@ -296,6 +296,19 @@ def exit_status(self):
...
+            self.set_exit_status(exit_code[0])
```
sphuber: and here use the actual attributes of the namedtuple i.e.
```
self.set_exit_status(exit_code.status)
self.set_exit_message(exit_code.message)
```

aiida/orm/nodes/process/process.py
```diff
@@ -296,6 +296,19 @@ def exit_status(self):
...
+            self.set_exit_message(exit_code[1])
+        else:
+            raise ValueError('exit code is not a tuple set with ExitCode')
```
sphuber: this should be a `TypeError('exit code should be an instance of ExitCode')`
Actually, it is best to put this check at the top of the method using `aiida.common.lang.type_check`

aiida/schedulers/plugins/slurm.py
```diff
@@ -236,6 +236,65 @@ def _get_detailed_job_info_command(self, job_id):
...
+        # Check the detailed job information and parse that
+        detailed_job_info = node.get_detailed_job_info()
+        exit_codes = node.process_class.exit_codes
+        if detailed_job_info is not None:
+            exit_code_info = self._parse_detailed_job_info(detailed_job_info['stdout'], exit_codes)
+        # Locate scheduler files and parse those
+        scheduler_stdout = node.get_scheduler_stdout()
+        scheduler_stderr = node.get_scheduler_stderr()
+        exit_code_file = self._parse_scheduler_files(scheduler_stdout, scheduler_stderr, exit_codes)
+        # For now we check which has the highest exit_status, assuming the higher the number, the
```
sphuber: Not sure about this logic, let's discuss

aiida/schedulers/plugins/slurm.py
```diff
@@ -236,6 +236,65 @@ def _get_detailed_job_info_command(self, job_id):
...
+        else:
+            raise ValueError('The supplied job id for the saact command does not return '
+                             'any valid entries.')
+
+        # Compose detailed job info dictionary
+        detailed_info = dict(zip(self._detailed_job_info_fields, info))
+        exit_status = int(detailed_info['ExitCode'].split(':')[0])
+        state = detailed_info['State']
+        reason = detailed_info['Reason']
+        exit_code = exit_codes.NO_ERROR
```
sphuber: this exit code will not exist, the `0` is implied. Simply return nothing from the function if no problem is found

aiida/schedulers/plugins/slurm.py
```diff
@@ -236,6 +236,65 @@ def _get_detailed_job_info_command(self, job_id):
...
+        if 'OUT_OF_MEMORY' in state:
+            exit_code = exit_codes.OUT_OF_MEMORY
+        return exit_code
+
+    def _parse_scheduler_files(self, scheduler_stdout, scheduler_stderr, exit_codes):
```
sphuber: Same here, return `None` or simply pass in this case if there is no error to return

aiida/engine/processes/calcjobs/calcjob.py
sphuber: nitpick-alert: Let's keep the convention to start exit code messages with capital and terminate with full-stop

ramirezfranciscof:
Hey @espenfl , could you briefly explain to me the relationship between this PR #3261? They would seem to be related (the other one seems more broad but also seems to be dealing with parsing the scheduler). Which is the current status of each?

sphuber:
Superseded by PR #3906 and #3931

Issue #1550, opened 2018-05-17
verdi computer enable/disabled ignored


giovannipizzi:
Now, even if I run `verdi computer disable <COMPUTERNAME>` and then I submit there, the simulation is submitted anyway.
Instead, no remote connection should be done while the computer is disabled (also for local transports).

giovannipizzi:
If we want to maintain the old behaviour, there should be an `if` check right before trying to ssh/connect to the machine, and just silently skip all calculations there if the computer is disabled (note: check the old logic, one should check both the global disabled state, and the one in the uauthinfo for the current user). One should also skip (or maybe in that case complain?) if the computer exists but it's not configured (again, the logic was there earlier, I believe). This is e.g. the case when you submit to a code that you imported.

As a bonus, I would also add a message to the bottom warnings (where we e.g. say that the daemon is off) if there is at least a calculation on a disabled computer).

Issue #664, opened 2017-08-23
Show warning when submitting calculation to disabled computer


sphuber:
When submitting a calculation to a computer that is currently disabled, there is no message, nor is it directly obvious from the output of `verdi calculation list`. If you have two computers `localhost` and `localhost1`, like me for example, where `localhost` is disabled and you accidentally submit to `localhost`, you will be scratching your head why your calculation seems to be stuck in the `TOSUBMIT` state for longer than you would care to admit...

PR #3648, opened 2019-12-12
Respect disabled computer and set processes to pause

dev-zero:
simple fix for #1550

### DIFF
--- a/aiida/backends/tests/engine/test_launch.py
+++ b/aiida/backends/tests/engine/test_launch.py
@@ -8,10 +8,14 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Module to test processess launch."""
+
+from tornado import gen
+
 from aiida import orm
 from aiida.backends.testbase import AiidaTestCase
 from aiida.common import exceptions
 from aiida.engine import launch, Process, CalcJob, WorkChain, calcfunction
+from aiida.manage.manager import get_manager
 
 
 @calcfunction
@@ -288,3 +292,60 @@ class TestLaunchersDryRun(AiidaTestCase):
         self.assertIn('folder', node.dry_run_info)
         for filename in ['single_file', 'file_one', 'file_two']:
             self.assertIn(filename, os.listdir(node.dry_run_info['folder']))
+
+
+class TestLauncherDisabledComputer(AiidaTestCase):
+    """Test the launcher when computer is disabled."""
+
+    @classmethod
+    def setUpClass(cls, *args, **kwargs):
+        """Set up a simple authinfo and for later use."""
+        super().setUpClass(*args, **kwargs)
+        cls.authinfo = orm.AuthInfo(computer=cls.computer, user=orm.User.objects.get_default())
+        cls.authinfo.store()
+
+    def setUp(self):
+        super().setUp()
+        self.assertIsNone(Process.current())
+        self.authinfo.enabled = False  # disable the computer
+
+    def tearDown(self):
+        super().tearDown()
+        self.assertIsNone(Process.current())
+        self.authinfo.enabled = True  # re-enable again
+
+    def test_calcjob_disabled_computer(self):
+        """Test that running a calcjob on a disabled computer will set the job to pause.
+
+        """
+        from aiida.plugins import CalculationFactory
+
+        ArithmeticAddCalculation = CalculationFactory('arithmetic.add')  # pylint: disable=invalid-name
+
+        code = orm.Code(input_plugin_name='arithmetic.add', remote_computer_exec=[self.computer, '/bin/true']).store()
+
+        inputs = {
+            'code': code,
+            'x': orm.Int(1),
+            'y': orm.Int(1),
+            'metadata': {
+                'options': {
+                    'resources': {
+                        'num_machines': 1,
+                        'num_mpiprocs_per_machine': 1
+                    }
+                }
+            }
+        }
+
+        timeout = 1  # Give the runner sometime to pause the calculation after submission
+        runner = get_manager().get_runner()
+
+        @gen.coroutine
+        def run_with_disabled():
+            calc_node = runner.submit(ArithmeticAddCalculation, **inputs)
+            self.assertFalse(calc_node.paused)
+            yield gen.sleep(timeout)
+            self.assertTrue(calc_node.paused)
+
+        runner.loop.run_sync(run_with_disabled)
--- a/aiida/cmdline/commands/cmd_computer.py
+++ b/aiida/cmdline/commands/cmd_computer.py
@@ -335,6 +335,8 @@ def computer_duplicate(ctx, computer, non_interactive, **kwargs):
 def computer_enable(computer, user):
     """Enable the computer for the given user."""
     from aiida.common.exceptions import NotExistent
+    from aiida.orm.querybuilder import QueryBuilder
+    from aiida.orm import CalcJobNode, Computer
 
     try:
         authinfo = computer.get_authinfo(user)
@@ -346,6 +348,17 @@ def computer_enable(computer, user):
     if not authinfo.enabled:
         authinfo.enabled = True
         echo.echo_info("Computer '{}' enabled for user {}.".format(computer.name, user.get_full_name()))
+
+        qb = QueryBuilder()
+        qb.append(CalcJobNode, filters={'attributes.paused': True}, tag='proc')
+        qb.append(Computer, with_node='proc', filters={'id': {'==': computer.id}})
+
+        if qb.first():
+            echo.echo_info(
+                'Found paused processes for Computer<{}>'
+                ', use `verdi process play` to resume them'.format(computer.name)
+            )
+
     else:
         echo.echo_info(
             "Computer '{}' was already enabled for user {} {}.".format(computer.name, user.first_name, user.last_name)
--- a/aiida/engine/processes/calcjobs/tasks.py
+++ b/aiida/engine/processes/calcjobs/tasks.py
@@ -12,6 +12,7 @@ import logging
 import tempfile
 
 from tornado.gen import coroutine, Return
+from wrapt import decorator
 
 import plumpy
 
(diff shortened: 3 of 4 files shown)


dev-zero:
ok, it seems to be a bit trickier: when submitting a job using `verdi run submit.py`, the process will hang. When trying to kill it, it will hang again and does not terminate on `ctrl+c` or `SIGTERM`.

```
$ verdi run submit.py 
^C12/12/2019 10:20:43 PM <404578> aiida.engine.runners: [CRITICAL] runner received interrupt, killing process 29
^C12/12/2019 10:21:58 PM <404578> aiida.engine.runners: [CRITICAL] runner received interrupt, killing process 29
[...]
```

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -35,7 +36,22 @@
...
 
 
+@decorator
+def pause_task_for_disabled_computer(wrapped, _, args, kwargs):
+    node = args[0]
+    authinfo = node.computer.get_authinfo(node.user)
+
+    if not authinfo.enabled:  # the target computer is disabled for this user
+        raise plumpy.PauseInterruption(
+            ('pausing CalcJob<{node.pk}> due to Computer<{node.computer.name}>'
```
sphuber: Two things:
1. The tuple around the string is not necessary because it is already in the `PauseInterruption` constructor, right?
2. No need to incude the calcjob pk in the message since this gets sets on the node itself so it is clear to which one it applies. I would just say `'Paused because Computer<{node.computer.name}> is disabled for User<{node.user.email}>'`

aiida/orm/computers.py
```diff
@@ -131,6 +133,18 @@ def full_text_info(self):
         else:
             ret_lines.append('   # No append text.')
 
+        ret_lines += [' * Configured users:']
+        all_users = users.User.objects(self.backend).all()
```
sphuber: No need to explicitly pass `self.backend`

aiida/orm/computers.py
```diff
@@ -131,6 +133,18 @@ def full_text_info(self):
...
+        for user in all_users:
+            try:
+                authinfo = self.get_authinfo(user)
+            except NotExistent:
+                ret_lines += ['   ✗ {}']
```
sphuber: unmatched `{}` in the string

sphuber:
> ok, it seems to be a bit trickier: when submitting a job using `verdi run submit.py`, the process will hang. When trying to kill it, it will hang again and does not terminate on `ctrl+c` or `SIGTERM`.
> 

And the same script _can_ be ran and then killed with CTRL+C in `develop` so without your changes?

aiida/engine/processes/calcjobs/tasks.py
dev-zero: true. ok, done.

aiida/orm/computers.py
dev-zero: done, thanks

aiida/orm/computers.py
dev-zero: ok

dev-zero:
> > ok, it seems to be a bit trickier: when submitting a job using `verdi run submit.py`, the process will hang. When trying to kill it, it will hang again and does not terminate on `ctrl+c` or `SIGTERM`.
> 
> And the same script _can_ be ran and then killed with CTRL+C in `develop` so without your changes?

yes

aiida/backends/tests/engine/test_launch.py
```diff
@@ -273,3 +276,61 @@ def test_calcjob_dry_run_no_provenance(self):
...
+                }
+            }
+        }
+
+        runner = get_manager().get_runner()
+
+        @gen.coroutine
+        def run_with_disabled():
+            calc_node = runner.submit(ArithmeticAddCalculation, **inputs)
+            for _ in range(1000):
```
sphuber: Is this loop just there to give the runner the chance to pause the process? In that case it might be better to use a small sleep, e.g.:
```
        timeout = 1  # Give the runner sometime to pause the calculation after submission
        runner = get_manager().get_runner()

        @gen.coroutine
        def run_with_disabled():
            calc_node = runner.submit(ArithmeticAddCalculation, **inputs)
            self.assertFalse(calc_node.paused)
            yield gen.sleep(timeout)
            self.assertTrue(calc_node.paused)

        runner.loop.run_sync(run_with_disabled)
```

dev-zero: Yes, it is. No, I don't think that's an appropriate solution: while I can assume that we have a finite and sane amount of yields/process interrupts in-between I can't make any assumptions about how much time they take, a sufficiently slow worker might get back from the sleep first, no?

aiida/cmdline/commands/cmd_computer.py
```diff
@@ -346,6 +348,17 @@ def computer_enable(computer, user):
     if not authinfo.enabled:
         authinfo.enabled = True
         echo.echo_info("Computer '{}' enabled for user {}.".format(computer.name, user.get_full_name()))
+
+        qb = QueryBuilder()
+        qb.append(ProcessNode, filters={'attributes.paused': True}, tag='proc')
```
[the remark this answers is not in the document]
dev-zero: ok, done

aiida/cmdline/commands/cmd_computer.py
```diff
@@ -346,6 +348,17 @@ def computer_enable(computer, user):
...
+        qb.append(Computer, with_node='proc', filters={'id': {'==': computer.id}})
+
+        if qb.limit(1):
+            echo.echo_info(
+                'Found paused processes for Computer<{}>'
+                ', use `verdi process play` to continue them'.format(computer.name)
```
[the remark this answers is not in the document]
dev-zero: done

aiida/backends/tests/engine/test_launch.py
sphuber: That is my problem with this solution though, I have no idea how long a 1000 yields take. I prefer to have a sane explicit timeout in seconds so we now how long we are waiting before giving up. If you think 1 second is cutting it a bit short and runs the risks of giving up too early we can increase it.

dev-zero: I am somewhat reluctant to build-in a race condition, but it's just a test, so ok (btw, would be max 1000 yields, in tests it takes less than 10). 1s should be more than enough, unless of course your system suddenly starts to swap, mines bitcoins or is encrypting your files ;-)

sphuber:
This looks ready to go, thanks a lot @dev-zero . Also thanks for adapting the doc string of the upload task that I sloppily forgot about :) I will try today to test it out in the wild as well. The only thing that still worries me a bit is your report that you could not kill a running process properly with these changes. I have not been able to reproduce this. Could you maybe still share what you ran in that `submit.py` script? Or have the problems gone away since?

dev-zero:
> Some minor changes. I checked out your branch locally and launched a calculation job with `run` and then interrupted it and in my case it works just fine. The process ends up in the killed state. Can you share the `submit.py` contents?

The following runs without problems without the patch (well, obvious) and hangs with it:

```python
from aiida.engine import run
from aiida import orm

code = load_code(14)

inputs = {
        'code': code,
        'x': orm.Int(1),
        'y': orm.Int(1),
        'metadata': {
            'options': {
                'resources': {'num_machines': 1, 'num_mpiprocs_per_machine': 1}
                }
            }
        }

result = run(CalculationFactory('arithmetic.add'), **inputs)
```

what I get:

```console
(aiida) tiziano@scarlet ~/work/aiida $ verdi run submit.py
(aiida) tiziano@scarlet ~/work/aiida $ verdi process list -a
  PK  Created    Process label             Process State    Process status
----  ---------  ------------------------  ---------------  -----------------------------------------------------------
  [...]
  57  10s ago    ArithmeticAddCalculation  ⏹ Finished [0]
Info: last time an entry changed state: 9s ago (at 09:14:20 on 2019-12-18)
(aiida) tiziano@scarlet ~/work/aiida $ verdi computer disable localhost tiziano.mueller@chem.uzh.ch
Info: Computer 'localhost' disabled for user Tiziano Müller (tiziano.mueller@chem.uzh.ch).
(aiida) tiziano@scarlet ~/work/aiida $ verdi run submit.py 
|
^C12/18/2019 10:25:46 AM <512974> aiida.engine.runners: [CRITICAL] runner received interrupt, killing process 63
12/18/2019 10:27:26 AM <512974> aiida.engine.runners: [CRITICAL] runner received interrupt, killing process 63
Killed
```

... first `CRITICAL`is from Ctrl+c, second one from sending `SIGTERM`, then `SIGKILL`.

sphuber:
When you say that the script runs normally without and hangs with patch, is that when you just run it and let it run? And is that with the computer disabled? Or, finally, do you only see the difference if you CTRL+C it?

If you see the hanging with the patch when you disable the computer, I think I know why. In that case the process is paused and even for the engine to kill it, it needs to be running first. If this is the case, we might need to have a kill action automatically resume a paused process first. However, I also notice that you decorated the `task_kill_job`. I realize we decided this because the disable should also be used by users to indicate that the machine is unreachable and the engine should not even try, but this then makes it obvious that a kill command will hang in a blocking `run` until the computer is enabled and the process is resumed. I am not sure yet where to "fix" this impasse.

dev-zero:
> When you say that the script runs normally without and hangs with patch, is that when you just run it and let it run? And is that with the computer disabled? Or, finally, do you only see the difference if you CTRL+C it?

What I try:

1. run the script on AiiDA develop with computer enabled. Works.
2. run the script on AiiDA develop with computer disabled. Works.
3. run the script on AiiDA this branch with computer enabled. Works.
4. run the script on AiiDA this branch with computer disabled. Hangs. Meaning the `verdi run` command does not terminate and can't be terminated by Ctrl+C or `SIGTERM`.

> If you see the hanging with the patch when you disable the computer, I think I know why. In that case the process is paused and even for the engine to kill it, it needs to be running first. If this is the case, we might need to have a kill action automatically resume a paused process first. However, I also notice that you decorated the `task_kill_job`. I realize we decided this because the disable should also be used by users to indicate that the machine is unreachable and the engine should not even try, but this then makes it obvious that a kill command will hang in a blocking `run` until the computer is enabled and the process is resumed. I am not sure yet where to "fix" this impasse.

Yes, I came to the same conclusion. Intuitively I would have said that `PAUSED-UPLOAD/SUBMIT/...` need to be full (and visible) states in the state machine rather than some kind of metastate (paused meaning: awaiting state change in the state machine) as it seems to be now.
And the reason why you need multiple states (or a hierarchical) state machine is because you have to remember where to go to on resume.

Submitting jobs to a disabled computer should generate a warning as requested in #664 but still end up in the queue and `verdi run` should return.

sphuber:
> 4. run the script on AiiDA this branch with computer disabled. Hangs. Meaning the verdi run command does not terminate and can't be terminated by Ctrl+C or SIGTERM.

This is a tricky one. With the current design and intention of `run` it is supposed to hang. It is not supposed to return control to the interpreter until the process has finished. Since it is paused it "hangs". Now, I do think that it should still be interruptable so Ctrl+C should just kill the process and terminate the interpreter, so that is something to fix. But I do not think that if a process that is launched through `run` is paused, the interpreter should just close, because that process is lost forever. It won't ever be resumed by anything.

sphuber:
We have put this to draft because although the current code is useful, it is currently broken for unknown reasons. The problem is low priority however, so we will keep this open for a later time.

sphuber:
I am going to close this since it has been open for so long and the code has changed significantly in between.

PR #3651, opened 2019-12-12
Move last_jobinfo from JSON-serialized string to dictionary

### LINKED ISSUE #3649: Move last_jobinfo from JSON-serialized string to dictionary
This is a follow-up of #3133 

In particular, add a `get_dict()` and a `load_from_dict()` in `scheduler.dataclasses.JobInfo` instead of only a serialiser, so we can store in the DB a dictionary and not a JSON string for the `last_jobinfo` in [this part of the code](https://github.com/aiidateam/aiida-core/blob/0644e7d9ff492fdb2c94d608c575c697daee6d32/aiida/orm/nodes/process/calculation/calcjob.py#L423), i.e. the cache of the latest time the engine saw the job (we will do this without DB migrations). (also, rename `last_jobinfo` to `last_job_info` for consistency with #3639 (`detailed_job_info`)

giovannipizzi:
For historical reasons, this field was stored as a JSON-serialized
string in the `last_jobinfo` attribute of a CalcJob. However, this
is cumbersome and makes querying very hard.

We now replace this with a dictionary, thanks to new commands to
get directly a dictionary (with serialized fields, so that the
dictionary is JSON-serializable).
These (and existing) methods of the JobInfo class are now also tested.

Finally, the attribute key has been renamed from `last_jobinfo` to
`last_job_info`, for consistency with the key `detailed_job_info`
introduced in #3639. By changing the type of the content, the field is
anyway not directly usable as before in scripts, so changing the name
is not an additional issue.
This should not give a real backward-incompatibility problem, since
this field was there mostly for debugging reasons.

Fixes #3649

### DIFF
--- a/aiida/orm/nodes/process/calculation/calcjob.py
+++ b/aiida/orm/nodes/process/calculation/calcjob.py
@@ -35,7 +35,7 @@ class CalcJobNode(CalculationNode):
     SCHEDULER_JOB_ID_KEY = 'job_id'
     SCHEDULER_STATE_KEY = 'scheduler_state'
     SCHEDULER_LAST_CHECK_TIME_KEY = 'scheduler_lastchecktime'
-    SCHEDULER_LAST_JOB_INFO_KEY = 'last_jobinfo'
+    SCHEDULER_LAST_JOB_INFO_KEY = 'last_job_info'
     SCHEDULER_DETAILED_JOB_INFO_KEY = 'detailed_job_info'
 
     # Base path within the repository where to put objects by default
@@ -436,7 +436,7 @@ class CalcJobNode(CalculationNode):
 
         :param last_job_info: a `JobInfo` object
         """
-        self.set_attribute(self.SCHEDULER_LAST_JOB_INFO_KEY, last_job_info.serialize())
+        self.set_attribute(self.SCHEDULER_LAST_JOB_INFO_KEY, last_job_info.get_dict())
 
     def get_last_job_info(self):
         """Return the last information asked to the scheduler about the status of the job.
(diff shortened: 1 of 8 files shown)


aiida/schedulers/datastructures.py
```diff
@@ -578,30 +590,42 @@ def deserialize_field(self, value, field_type):
...
-        return json.dumps(ser_data)
+    def get_dict(self):
+        """
+        Serialise the current data into a dictionary that is JSON-serializable.
+
+        :return: A dictionary
+        """
+        return {k: self.serialize_field(v, self._special_serializers.get(k, None)) for k, v in self.items()}
+
+    def load_from_dict(self, data):
```
sphuber: Wouldn't this make more sense as a classmethod?

aiida/schedulers/datastructures.py
```diff
@@ -578,30 +590,42 @@ def deserialize_field(self, value, field_type):
...
+        """
+        for key, value in data.items():
+            self[key] = self.deserialize_field(value, self._special_serializers.get(key, None))
 
     def load_from_serialized(self, data):
         """
-        Load value from serialised data
-        :param data: The data to load from
-        :return: The value after loading
+        Load value from JSON-serialised data as a string
```
sphuber: Same here, better as class method, even though that would be backwards incompatible change, but I think it is ok

aiida/orm/nodes/process/calculation/calcjob.py
```diff
@@ -445,11 +445,11 @@ def get_last_job_info(self):
...
         from aiida.schedulers.datastructures import JobInfo
 
-        last_job_info_serialized = self.get_attribute(self.SCHEDULER_LAST_JOB_INFO_KEY, None)
+        last_job_info_dictserialized = self.get_attribute(self.SCHEDULER_LAST_JOB_INFO_KEY, None)
 
-        if last_job_info_serialized is not None:
+        if last_job_info_dictserialized is not None:
             job_info = JobInfo()
-            job_info.load_from_serialized(last_job_info_serialized)
+            job_info.load_from_dict(last_job_info_dictserialized)
```
sphuber: This can become `job_info = JobInfo.load_from_dict(last_job_info_dict_serialized)`

sphuber (dismissed):
Thanks @giovannipizzi looking good, I would just turn the loaders into classmethods

sphuber (changes requested):
Accidentally approved first

aiida/schedulers/datastructures.py
giovannipizzi: Done

aiida/orm/nodes/process/calculation/calcjob.py
giovannipizzi: Done

sphuber (approved):

sphuber:
Rebasing and squashing seems to have done it, the tests passed. Or maybe it was just luck/random again. We will see. Thanks for the PR in any case @giovannipizzi !

Issue #3602, opened 2019-12-04
Reinstate `coverage` to CI when testing is migrated to `pytest`


CasperWA:
@sphuber Is it correctly understood that coverage/codecov has disappeared in the process of moving to GitHub Actions?

sphuber:
Yes as mentioned in PR #3571 I omitted it on purpose. It was often buggy so we did not actually pay any attention to what it was saying. We put a big threshold on it just to prevent it from giving false positives. In addition, it was not even correctly registering all paths we were running. The move to `pytest` will allow us to run `pytest-cov` which should fix it. Maybe we look into it then again

CasperWA:
Cool. I'll just put in a link to [codecov-action](https://github.com/codecov/codecov-action) here for the future implementation of this.

ltalirz:
we might actually want to give it a try even before - if the information is usable, then it could be very helpful for improving test coverage for the "data/tools" commands during the coding week.

CasperWA:
From short discussion at the coding week 2019 between @dev-zero, @giovannipizzi, and me:

- Move to codecov.
- Create GH Actions job to run on pushes to `master` and `develop` branches (on `aiidateam/aiida-core`) and for all PRs.
- Make sure to properly exclude test files and folders, and consider which other files/folders should be excluded.

Issue #2835, opened 2019-05-04
test coverage - worst offenders


ltalirz:
Despite the, still, rather unhelpful coveralls bot, I have the impression we're [overall making good progress in terms of test coverage](https://coveralls.io/github/aiidateam/aiida_core?branch=develop).

There aren't actually many **files with zero test coverage** left anymore. This is a complete list - perhaps we can make some progress on this together?

engine: 
- [x] https://github.com/aiidateam/aiida_core/blob/develop/aiida/engine/daemon/runner.py ~~@sphuber is this difficult to test?~~ Already tested; coverage report will be fixed automatically once we start using `pytest.cov` #2771

REST API:
 - [x] https://github.com/aiidateam/aiida_core/blob/develop/aiida/restapi/common/config.py @waychal I guess you're using a different config in the tests. One way to make sure the config is tested would be to load this config but then redefine some values
 - [x] https://github.com/aiidateam/aiida_core/blob/develop/aiida/restapi/run_api.py @waychal I think it should be possible to test at least part of this code. Also it's argparse and should be moved to click (I'll look into that)

plugins: 
There is a whole codebase for interacting with the online registry (listing plugins, installing plugins, ...) that I was unaware of and that was apparently never used. We need to decide what to do with this. On the one hand, querying the plugin registry from the cli might be convenient. On the other hand, the web site is just one click away and we should definitely avoid reimplementing pip.
 - [x] https://github.com/aiidateam/aiida_core/blob/develop/aiida/plugins/registry.py
 - [x] https://github.com/aiidateam/aiida_core/blob/develop/aiida/plugins/info.py
 - [x] https://github.com/aiidateam/aiida_core/blob/develop/aiida/plugins/utils.py
 - [x] https://github.com/aiidateam/aiida_core/blob/develop/aiida/plugins/entry.py
SPH: I also raised this point elsewhere recently and am strongly in favor of deleting the code. It does not do anything at the moment and like you say we should be careful with reimplementing our own package manager.

cmdline:
 - [x] https://github.com/aiidateam/aiida_core/blob/develop/aiida/cmdline/utils/ascii_vis.py
   @sphuber apparently this is currently only used in `verdi process status`
 - [x] https://github.com/aiidateam/aiida_core/blob/develop/aiida/cmdline/utils/repository.py
   @sphuber e.g. used in `calcjob inputls`

misc:
[...]

PR #3653, opened 2019-12-12
Move tests to pytest

### LINKED ISSUE #2771: move aiida-core tests to pytest
We are already recommending pytest for plugins because of generally increased usability & the potential for code reuse via fixtures.
At some point we should move aiida-core tests to pytest as well.

 - [ ] move all tests outside the `aiida` folder and inside a top-level `tests` folder with the same folder hierarchy. This will significantly reduce the size of the aiida package on PyPI. We can think whether it is then still necessary to rebuild the test hierarchy of `verdi devel tests` or whether it might be just simpler to do `pytest tests/the/tests/I/want/to/run`
 - [ ] get rid of the need for creating a  "test profile" by defaulting to the dynamically created postgres cluster provided by `pgtest`. Potentially give users who don't want this overhead the option to stay with their system postgres cluster by fixing #2648

as discussed with @sphuber

ltalirz:
fixes #2771 
affects #2835, #3602 

This PR starts running the tests of aiida-core through pytest.
Changes to the `AiiDATestCase` class are kept minimal.
The code for `verdi devel tests` is removed, the command is deprecated and now prints the suggestion to use `pytest`.

To do, but not in this PR:
 * Fix (py)test warnings (will touch lots of files)
 * Move test files out of aiida-core inside top-level `tests/` folder
 * Where it makes sense, start moving existing tests from the `AiiDATestCase` to pytest (e.g. where pytest fixtures are making things more effective)


aiida/backends/djsite/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Configuration file for pytest tests."""
+
+from aiida.backends import BACKEND_DJANGO
+from aiida.manage.tests import get_test_backend_name
+
+if get_test_backend_name() != BACKEND_DJANGO:
```
ltalirz: I notice there is still a minor issue here.

If you set `AIIDA_TEST_PROFILE`, the test manager will simply use that profile and its backend (ignoring the value of `AIIDA_TEST_BACKEND`).

I.e. when you set the `AIIDA_TEST_PROFILE` variable to an sqlalchemy test profile *without* also setting `AIIDA_TEST_BACKEND=sqlalchemy`, then the pytest test discovery will think you are running on the default django backend and discover the wrong tests.
This means, when running on an `sqlalchemy`

Note: This does not affect the CI setup, where the `AIIDA_TEST_BACKEND` variable is set correctly.

aiida/cmdline/params/types/test_module.py
```diff
@@ -12,35 +12,9 @@
...
-
-        prefix_db = 'db'
-        modules_db = get_db_test_names()
-        modules_base = ['aiida.cmdline', 'aiida.common', 'aiida.schedulers', 'aiida.transports']
-
-        test_modules = {}
-
-        for module in modules_base:
-            test_modules[module] = None
+    Defunct - remove when removing the "verdi devel tests" command.
```
sphuber: Maybe add the deprecation warning using `v2.0.0` because that way it is easy to search the code base when removing deprecated stuff

ltalirz: done!

aiida/backends/djsite/conftest.py
greschd: > @greschd and me briefly looked into whether pytest offers a way to hook custom code into _before_ test discovery but we didn't find a way to do that.

If I remember correctly there _are_ ways to hook custom code into pytest, but it's just not as simple as adding a ``pytest.skip`` somewhere.

Here's the API reference for pytest hooks: https://docs.pytest.org/en/latest/reference.html#hook-reference

Either ``pytest_configure`` or ``pytest_sessionstart`` seems to be the right place for that. 

I'm guessing the "wrong backend" issue can occur any time the fixtures are used, not only for the ``aiida-core`` tests (i.e., also for plugin testing).

ltalirz: > Either pytest_configure or pytest_sessionstart seems to be the right place for that code.

This looks promising. If you can find the time to look into this, that would be great.

> I'm guessing the "wrong backend" issue can occur any time the fixtures are used, not only for the aiida-core tests (i.e., also for plugin testing).

Yes, but it only affects test discovery and in practice plugins are unlikely to have backend-dependent tests.

PR #3654, opened 2019-12-12
Add the `BaseRestartWorkChain` to `aiida-core`

### LINKED ISSUE #3644: Move BaseRestartWorkChain to aiida-core
Currently, the `BaseRestartWorkChain` work chain is used in [aiida-qe](https://github.com/aiidateam/aiida-quantumespresso/blob/develop/aiida_quantumespresso/common/workchain/base/restart.py), [aiida-cp2k](https://github.com/aiidateam/aiida-cp2k/blob/develop/aiida_cp2k/workchains/aiida_base_restart.py), [aiida-vasp](https://github.com/aiida-vasp/aiida-vasp/blob/develop/aiida_vasp/workchains/restart.py), [aiida-crystal17](https://github.com/chrisjsewell/aiida-crystal17/blob/develop/aiida_crystal17/workflows/common/restart.py), [aiida-raspa](https://github.com/yakutovicha/aiida-raspa/blob/develop/aiida_raspa/workchains/aiida_base_restart.py) plugins. They've all started from the same source (aiida-qe), but then slightly diverged.

Since the functionality of the `BaseRestartWorkChain` is useful for different packages, it makes sense to move it to one central place (aiida-core).

sphuber:
Fixes #3644

### DIFF
--- a/.ci/workchains.py
+++ b/.ci/workchains.py
@@ -7,10 +7,62 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-
-from aiida.engine import calcfunction, workfunction, WorkChain, ToContext, append_
+from aiida.common import AttributeDict
+from aiida.engine import calcfunction, workfunction, WorkChain, ToContext, append_, while_
+from aiida.engine import BaseRestartWorkChain, register_process_handler, ProcessHandlerReport
 from aiida.engine.persistence import ObjectLoader
 from aiida.orm import Int, List, Str
+from aiida.plugins import CalculationFactory
+
+
+ArithmeticAddCalculation = CalculationFactory('arithmetic.add')
+
+
+class ArithmeticAddBaseWorkChain(BaseRestartWorkChain):
+    """Ridiculous workchain around `AritmethicAddCalculation` with automated sanity checks and error handling."""
+
+    _process_class = ArithmeticAddCalculation
+
+    @classmethod
+    def define(cls, spec):
+        """Define the process specification."""
+        super(ArithmeticAddBaseWorkChain, cls).define(spec)
+        spec.expose_inputs(ArithmeticAddCalculation, namespace='add')
+        spec.expose_outputs(ArithmeticAddCalculation)
+        spec.outline(
+            cls.setup,
+            while_(cls.should_run_process)(
+                cls.run_process,
+                cls.inspect_process,
+            ),
+            cls.results,
+        )
+        spec.exit_code(100, 'ERROR_TOO_BIG', message='The sum was too big.')
+
+    def setup(self):
+        """Call the `setup` of the `BaseRestartWorkChain` and then create the inputs dictionary in `self.ctx.inputs`.
+
+        This `self.ctx.inputs` dictionary will be used by the `BaseRestartWorkChain` to submit the process in the
+        internal loop.
+        """
+        super(ArithmeticAddBaseWorkChain, self).setup()
+        self.ctx.inputs = AttributeDict(self.exposed_inputs(ArithmeticAddCalculation, 'add'))
+
+
+@register_process_handler(ArithmeticAddBaseWorkChain, priority=500)
+def sanity_check_not_too_big(self, node):
+    """My puny brain cannot deal with numbers that I cannot count on my hand."""
+    if node.is_finished_ok and node.outputs.sum > 10:
+        return ProcessHandlerReport(True, self.exit_codes.ERROR_TOO_BIG)
+
+
+@register_process_handler(ArithmeticAddBaseWorkChain, priority=400)
+def error_negative_sum(self, node):
+    """What even is a negative number, how can I have minus three melons?!."""
+    if node.exit_status == ArithmeticAddCalculation.exit_codes.ERROR_NEGATIVE_NUMBER.status:
+        self.ctx.inputs.x = Int(abs(node.inputs.x.value))
+        self.ctx.inputs.y = Int(abs(node.inputs.y.value))
+        return ProcessHandlerReport(True)
 
 
 class NestedWorkChain(WorkChain):
--- a/.ci/workchains.py
+++ b/.ci/workchains.py
@@ -8,7 +8,7 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 from aiida.common import AttributeDict
-from aiida.engine import calcfunction, workfunction, WorkChain, ToContext, append_, while_
+from aiida.engine import calcfunction, workfunction, WorkChain, ToContext, append_, while_, ExitCode
 from aiida.engine import BaseRestartWorkChain, register_process_handler, ProcessHandlerReport
 from aiida.engine.persistence import ObjectLoader
 from aiida.orm import Int, List, Str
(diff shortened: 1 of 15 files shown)


aiida/engine/processes/workchains/restart.py
```diff
@@ -0,0 +1,425 @@
...
+    def load_instance_state(self, saved_state, load_context):
+        """Load the process instance from a saved state.
+
+        :param saved_state: saved state of existing process instance
+        :param load_context: context for loading instance state
+        """
+        super().load_instance_state(saved_state, load_context)
+        self._load_error_handlers()
+
+    def _load_error_handlers(self):
```
greschd: Shouldn't there also be a ``_load_sanity_checks``? Or if the two are supposed to be in the same entry points list (which I guess they might as well), maybe the ``cls._error_handler_entry_point`` should be renamed to something more generic, like ``cls._handler_entry_points``?

aiida/engine/processes/workchains/restart.py
```diff
@@ -0,0 +1,425 @@
...
+        """Load the error handlers defined through entry points, if any."""
+        if self._error_handler_entry_point is not None:
+            for entry_point_name in get_entry_point_names(self._error_handler_entry_point):
```
greschd: I think the ``_error_handler_entry_point`` attribute also needs to be concatenated with the values from base classes (same as ``_sanity_checks`` and ``_error_handlers``).

aiida/engine/processes/workchains/restart.py
sphuber: Good point, I think they can be in the same entry point, because the decorator will distinguish them. I am a bit annoyed about the asymmetry in `sanity_check` and `error_handler` and that they don't share a common suffix. Because this way `_handler_entry_points` may be mistaken to only reference error handlers.

aiida/engine/processes/workchains/restart.py
sphuber: Not sure that we need multiple entry points. Note that it really is just the name of the entry point group. The group can of course have multiple entry points. So a sub class can always register additional entry points in the same group that is declared by the base class. No?

aiida/engine/processes/workchains/restart.py
```diff
@@ -0,0 +1,425 @@
...
+        if calculation.is_finished_ok:
+
+            # The calculation was successful so we reset the unhandled failure flag
+            self.ctx.unhandled_failure = False
+
+            exit_code = self._call_sanity_checks(calculation)
+
+            if isinstance(exit_code, ExitCode):
+                if exit_code.status == 0:
+                    self.report('{}<{}> finished successfully, but sanity check failed, restarting'.format(
```
yakutovicha: I am a bit puzzled here. Shouldn't it be the opposite? I thought `exit_code.status == 0:` means sucesfull termination.

aiida/engine/processes/workchains/restart.py
```diff
@@ -0,0 +1,425 @@
...
+                        self.ctx.calc_name, calculation.pk))
+                else:
+                    self.report('{}<{}> finished successfully, but sanity check detected unrecoverable problem'.format(
+                        self.ctx.calc_name, calculation.pk))
+
+        else:
+
+            try:
```
yakutovicha: I would propose to align the logic here to the one of sanity checks.

sphuber: As discussed in person, but here for reference. There is a fundamental difference in handling successful and failed calculations. For a successful calculation, if no sanity checks return a report that is fine, if for a _failed_ calculation no reports are returned, that _is_ a problem and is called an "unhandled failure". Here the base workchain then implements the logic that it aborts if there are two unhandled failed calculations consecutively.

yakutovicha: Okay, we had an offline discussion with @sphuber where he explained the logic behind this design. Unlike sanity checks, if error handlers did not do anything - that means the problem remains (returns `UnhandledCalculationFailure`).

aiida/engine/processes/workchains/restart.py
greschd: Maybe call it ``_registration_entry_points`` instead?

aiida/engine/processes/workchains/restart.py
greschd: Hmm, it depends how you think about "ownership" of the entry point namespace. In principle, there could be clashes across subclasses from different plugins - although that seems kind of unlikely.

aiida/engine/processes/workchains/restart.py
```diff
@@ -0,0 +1,423 @@
...
+                    self.report('{}<{}> finished successfully, but sanity check failed, restarting'.format(
+                        self.ctx.process_name, node.pk))
+                else:
+                    self.report('{}<{}> finished successfully, but sanity check detected unrecoverable problem'.format(
+                        self.ctx.process_name, node.pk))
+
+        else:
+
+            try:
+                exit_code = self._call_error_handlers(node)
```
greschd: Shouldn't this happen _before_ the sanity checks? If I understand this code right, if the error handler succeeds (i.e., it becomes _no longer an error_), the sanity checks will not be called.

aiida/engine/processes/workchains/restart.py
```diff
@@ -0,0 +1,332 @@
...
+        if node.is_killed:
+            return self.exit_codes.ERROR_SUB_PROCESS_KILLED  # pylint: disable=no-member
+
+        # Sort the handlers with a priority defined, based on their priority in reverse order
+        handlers = [handler for handler in self._handlers if handler.priority]
+        handlers = sorted(handlers, key=lambda x: x.priority, reverse=True)
+
+        last_report = None
+
+        for handler in handlers:
```
yakutovicha: Just a remark that we need to be very careful to what logic we stick here. Because it will shape the handlers to a very large extent. The difficult decision needs to be made when at least one handler returned a non-zero report and at least one returned a zero status.

The possibilities that I see here are the following:

1. If at least one handler returned a non-zero status - consider that the work chain failed, so it should be stopped.

1. If at least one handler returned a zero status - consider that the error was fixed, so the work chain should be restarted.

1. **(current approach)** Decide whether to stop or to continue the work chain based on the last meaningful handler report. Add `do_break` attribute to the ProcessHandlerReport to make the handler **the last one**.

1. Decide whether to stop or to continue the work chain based on the first meaningful handler report. Add `do_continue` attribute to continue iterating the handlers.

1. Add `stop_workchain` attribute to the ErrorHandlerReport that explicitly says whether to stop or not the work chain if the handler failed.

yakutovicha: Thinking a bit more on this, I would propose the combination of 2. and 5.:

If at least one handler returned a zero-status - restart the work chain. If at least one handler returned non-zero status and has `stop_workchain` set to `True` - stop the work chain. If all meaningful reports have non-zero exit status - stop the work chain.

.ci/workchains.py
```diff
@@ -7,10 +7,62 @@
...
+def sanity_check_not_too_big(self, node):
+    """My puny brain cannot deal with numbers that I cannot count on my hand."""
+    if node.is_finished_ok and node.outputs.sum > 10:
+        return ProcessHandlerReport(True, self.exit_codes.ERROR_TOO_BIG)
+
+
+@register_process_handler(ArithmeticAddBaseWorkChain, priority=400)
+def error_negative_sum(self, node):
+    """What even is a negative number, how can I have minus three melons?!."""
+    if node.exit_status == ArithmeticAddCalculation.exit_codes.ERROR_NEGATIVE_NUMBER.status:
```
yakutovicha: I think this error isn't defined.

sphuber: I added it in the [first commit](e22ad2bdafb8d12eaa4431381f47dc800214efa3) of this PR

aiida/engine/processes/workchains/restart.py
```diff
@@ -0,0 +1,332 @@
...
+                last_report = report
+
+            # After certain handlers, we may want to skip all other handlers
+            if report and report.do_break:
+                break
+
+        report_args = (self.ctx.process_name, node.pk)
+
+        # If the process failed and no handler returned a report we consider it an unhandled failure
+        if node.is_failed and not last_report:
```
yakutovicha: I think this condition can make problems if the user does not register a handler for some exit code. Another handler may handle a completely different problem and will provide a report for that handling. However, the actual problem wasn't resolved, so the work chain should fail.

I think one should stop

sphuber:
Update on the progress:

 * I have merged the original commit introducing the `BaseRestartWorkChain` with error handler and sanity checks separate, with the follow up commit that merged the concepts.
 * I have implemented a way to register process handlers through the inputs. The `process_handlers` input namespace takes a mapping of `ProcessHandler` tuples. See the code documentation for further details and commit for design justification.

I fear that although this last feature can be useful, it potentially badly affects reproducibility. Since these inputs are non-storable, there currently is no real record of them. We should think how/where we store a record of all handlers that were registered when the work chain was launched. This should be the bare minimum I would say. The same goes for the other mechanism of registering additional handlers through the entry points.

That just leaves the feature to disable certain handlers dynamically through the inputs.

sphuber:
Superseded by PR #3748 that implements a minimal working version of the base restart work chain infrastructure. Two requested features have been omitted for the time being for which feature request issues have been opened:
* #3750 dynamically registering process handlers
* #3751 disabling registered process handlers at launch time

Issue #3645, opened 2019-12-12
Sessions not handled correctly in QB for asynchronous function calls


CasperWA:
I have found in my [`aiida-optimade`](https://github.com/aiidateam/aiida-optimade) REST API implementation exposing an AiiDA DB in the OPTiMaDe spec format that I get the [SQLA QueuePool error](https://docs.sqlalchemy.org/en/13/errors.html#queuepool-limit-of-size-x-overflow-y-reached-connection-timed-out-timeout-z) when making more than ~15 HTTP GET requests.

I have remedied this in my implementation by an AiiDA hack that creates a new DB session per request and closes it at the end.

It seems to me the native AiiDA REST API will have a similar problem for too many (concurrent) requests.

CasperWA:
Using the branch [szoupanos/aiida_core:fix_issue_3645](https://github.com/szoupanos/aiida_core/tree/fix_issue_3645) I have created a test branch for `aiida-optimade` [here](https://github.com/CasperWA/aiida-optimade/tree/change_for_aiida-core_fix_3645).

Currently, my tests are failing with the error:

```shell
self = <sqlalchemy.pool.impl.QueuePool object at 0x7f77894c1d10>

    def _do_get(self):
        use_overflow = self._max_overflow > -1

        try:
            wait = use_overflow and self._overflow >= self._max_overflow
            return self._pool.get(wait, self._timeout)
        except sqla_queue.Empty:
            # don't do things inside of "except Empty", because when we say
            # we timed out or can't connect and raise, Python 3 tells
            # people the real error is queue.Empty which it isn't.
            pass
        if use_overflow and self._overflow >= self._max_overflow:
            if not wait:
                return self._do_get()
            else:
                raise exc.TimeoutError(
                    "QueuePool limit of size %d overflow %d reached, "
                    "connection timed out, timeout %d"
                    % (self.size(), self.overflow(), self._timeout),
>                   code="3o7r",
                )
E               sqlalchemy.exc.TimeoutError: QueuePool limit of size 2 overflow 1 reached, connection timed out, timeout 30 (Background on this error at: http://sqlalche.me/e/3o7r)
```

This fail occurs for a test that tries to iterate through the result pages until there are no more results returning under an entry resource endpoint.

After this, tests do not fail but raise a pytest error. In the output it reveals the same error as shown above, meaning a new session is not provided to the new requests, instead it seems to reuse the same session, where the QueuePool error has occurred and a time-out has been set.

[...]

PR #3655, opened 2019-12-12
Fix of the multithreading connection problem

szoupanos:
This pull request addresses the problem described at issue #3645

The problem was isolated and seems to appear when multiple threads issue a query. They obtain thread local sessions from the scoped session registry (aiida.backends.sqlalchemy - https://docs.sqlalchemy.org/en/13/orm/contextual.html) and subsequently they obtain connections to the database from a connection pool.

The size of this pool can be defined at the construction of the engine (aiida.backends.sqlalchemy.reset_session). Unfortunately, when a thread ends its life, the connection is not released and therefore in the end, all the connections of the connection pull are used leading to rejection of new query requests and timeouts.

The solution to this problem was to pass a flag during the construction of the QueryBuilder that instructs the QueryBuilder to close the session and subsequently the connection to the database. This will leave any obtained objects without any connection/link to the database data (so the user will not be able to update the received object and store it back to the database) but it is a solution that:
a) works for both backends
b) it works with multuthreaded QueryBuilder usage
c) It works with mutithreaded REST usage
d) It works with Optimade (@CasperWA)

Moreover, we keep the scoped session registry that we consider important.

It was decided to use to pass a flag to the construction of the QueryBuilder (to instruct it to close the session) instead of a context manager to keep the changes to the REST api (aiida.restapi.translator.base.BaseTranslator) to a minimum - please note that in BaseTranslator the QueryBuilder query is built gradually by several methods.

### DIFF
--- a/aiida/backends/sqlalchemy/__init__.py
+++ b/aiida/backends/sqlalchemy/__init__.py
@@ -74,29 +74,6 @@ def reset_session(profile=None):
         name=profile.database_name
     )
 
-    ENGINE = create_engine(
-        engine_url,
-        json_serializer=json.dumps,
-        json_deserializer=json.loads,
-        encoding='utf-8'
-        # pool_size=2,
-        # max_overflow=1
-    )
+    ENGINE = create_engine(engine_url, json_serializer=json.dumps, json_deserializer=json.loads, encoding='utf-8')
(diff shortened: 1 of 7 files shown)


sphuber:
Superseded by #3708

Issue #3541, opened 2019-11-13
add cli for setting minimum poll interval


ltalirz:
it should be possible to set the minimum poll interval for a computer from the command line

```
In [1]: c = load_computer(1)

In [2]: c.get_minimum_job_poll_interval()
Out[2]: 10.0

In [3]: c.set_minimum_job_poll_interval(30)
```

E.g. `verdi computer configure ssh mycomp --job-poll-interval 40` (I think we should leave out the "minimal" - it's also not used in `safety_interval`).

giovannipizzi:
See discussion in #3630 to move this to the authinfo (this will make it possible to ask in `verdi computer configure`)

giovannipizzi:
Please check the discussion of the first tentative implementation in PR #3663, now closed - this contains a tentative, partial implementation, and some discussions, that should be helpful when addressing this issue.
To be addressed probably together with #3630

PR #3663, opened 2019-12-13
Migrating configuration options to AuthInfo and redesigning the interface

### LINKED ISSUE #3630: Move the `minimum_job_poll_interval` from Computer to AuthInfo
Currently, the `minimum_job_poll_interval` is set in the Computer metadata (this is the interval between running `squeue` commands).

I think this is a user setting more than a global computer feature, so this should go to the AuthInfo and be asked during the `computer configure` phase.

This might be partially related also the restructuring of Computer (#3616).

Also, when this is changed, please also update the corresponding documentation that is being merged in #3629

giovannipizzi:
I am creating this PR mainly as a discussion place for the design.

This work originally started to fix #3630 (and #3541). However, while working on it, I found a number of design issues that might benefit for a general redesign.

This code shouldn't be merged yet - it starts some moving around, but after further discussions with @muhrin and @sphuber we might want to change properly the current interaction between transport/AuthInfo/scheduler/... classes (and click).

Here are some points after the discussion - once we agree, we can start doing the actual work applying these decisions.

Design issues:
- `AuthInfo` it's actually a `ComputerConfig` object. We don't necessarily need to rename it, but we should think to it that way (all the configuration for a given computer and aiida user pair).
- It contains a set of properties that can be queried by name. This includes generic things, like the safe open interval and the scheduler poll interval, plus one special property being the `auth_params`, that are just piped through to the transport to generate a new instance.
- The transport should *not* know about the generic properties. Now done in the class by the following code:
  ```
  _common_auth_options = [(		
        'safe_interval', {		
            'type': float,		
            'prompt': 'Connection cooldown time (s)',		
            'help': 'Minimum time interval in seconds between consecutive connection openings',		
            'callback': validate_positive_number		
        }		
    )]
  ```
  Moreover, these properties are sent to the `__init__` of the transport, and then just set in an internal variable to then return it back when asked. Since there is no functionality, these variables should never go or be known to the Transport (the transport just knows how to open a transport, the caller is in charge of knowing how often this can be done).
- These settings above (and the same for the transport-specific ones), anyway, should not be click-specific but we should have our own schema and then a way to convert them to a given UI (e.g. click) - what we need is essentially a type (and a validator), a short and a long help string, and a function to get the default
- We still need to merge the authinfo-generic keys and the auth_params (transport-specific) settings in the CLI for `verdi computer configure xxx`
- we should try to simplify the automatic generation of click commands for `verdi computer configure xxx`.
- Moreover, for the entry points, we will remove the entry point `aiida.cmdline.computer.configure` and generate also those from the `aiida.transports` entry point (this was already partially happening).
- We should not have (like I'm originally doing in this WIP code change) the `_common_auth_options` floating around, but we should have an object that knows about this schema (in a non-click-specific way). Ideally this should be in AuthInfo. The problem is, however, that this means that at import time we need to import `aiida.orm` (to get the AuthInfo and generate the click decorators to `verdi computer configure xxx`) and this is slow. We need to decide if we want to solve the speed issue first, and use the `AuthInfo` object as the main class coordinating all this, or have another class (e.g. `aiida.common.computerconfig.ComputerConfig`) that does not know about the backend. `AuthInfo` would just wrap it and provide a "proxy" for the relevant methods, but the logic will be mostly in this internal class, from which the CLI can get the schema information (currently `_common_auth_options`) without loading `aiida.orm`.
- Note: Should this `ComputerConfig` know about the user and the computer? If so, then it should know about them without needing to go to the orm/DB, otherwise it defeats its purpose. Maybe it just needs to contain the content. However, it needs to know at least the TransportClass it is bound to (maybe in the future also the scheduler?)

Other issues:
- fix the tests, make sure that everywhere where we were using the safe-interval (or setting it for efficiency in testing) this is now properly honoured
- update documentation
- Choose properly the names of options. Since this might be the most important user-facing change, let's pick names we're happy with and:
  - let's provide a data migration
  - let's make sure the old ones still work, deprecated

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -26,7 +26,7 @@ verdi -p $AIIDA_TEST_BACKEND daemon incr 4
 
 verdi -p $AIIDA_TEST_BACKEND computer setup --non-interactive --label=localhost --hostname=localhost --transport=local \
     --scheduler=direct --mpiprocs-per-machine=1 --prepend-text="" --append-text=""
-verdi -p $AIIDA_TEST_BACKEND computer configure local localhost --non-interactive --safe-interval=0
+verdi -p $AIIDA_TEST_BACKEND computer configure local localhost --non-interactive --safe-open-interval=0
 
 # Configure the 'add' code inside localhost
 verdi -p $AIIDA_TEST_BACKEND code setup -n -L add \
--- a/.github/config/localhost-config.yaml
+++ b/.github/config/localhost-config.yaml
@@ -1,2 +1,2 @@
 ---
-safe_interval: 0
+safe_open_interval: 0
--- a/aiida/backends/djsite/db/subtests/migrations/test_migrations_many.py
+++ b/aiida/backends/djsite/db/subtests/migrations/test_migrations_many.py
@@ -770,8 +770,8 @@ class TestTextFieldToJSONFieldMigration(TestMigrations):  # pylint: disable=too-
         self.computer = self.DbComputer(**self.computer_kwargs)
         self.computer.save()
 
-        self.auth_info_auth_params = {'safe_interval': 2}
-        self.auth_info_metadata = {'safe_interval': 2}
+        self.auth_info_auth_params = {'safe_open_interval': 2}
+        self.auth_info_metadata = {'safe_open_interval': 2}
         self.auth_info_kwargs = {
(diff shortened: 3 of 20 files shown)


giovannipizzi:
Closing for now, while we agree on how to fix this problem

PR #3665, opened 2019-12-13
Make command line deprecation warnings visible with test profile

sphuber:
The `deprecated_command` decorator was only printing the warning if the
profile was not a test profile. With `verdi devel tests` being
deprecated, typically run with a test profile, the warning was being
swallowed.


sphuber:
@ltalirz I had to update it a bit because the tests failed

PR #3666, opened 2019-12-13
Move `CalcJob.presubmit` call from `CalcJob.run` to `Waiting.execute`

### LINKED ISSUE #3664: Move call of `CalcJob.prepare_for_submission` to upload transport task
The current disconnect between the `presubmit` call (which calls `prepare_for_submission`) performed in the `run` method of the process and its results being used in the upload transport task are giving problems with the fleeting character of the temporary sandbox folder containing the written inputs that need to be passed between the two points. Moving this logic together in the upload task makes it less intertwined and easier to deal with problems

sphuber:
Fixes #3664 

The `presubmit` calls through to `prepare_for_submission` which is the
place where the `CalcJob` implementation writes the input files based on
the inputs nodes in a temporary sandbox folder. The contents of which
are then stored in the repository of the node before the process
transitions to the waiting state in await of the upload task. This
upload task will then copy over the contents of the node repository to
the working directory on the remote computer.

There is a use case to limit which files are copied to the node's
repository from the sandbox folder used by `prepare_for_submission`.
This means that the sandbox folder will have to be passed to the
`upload_calculation`. However, since the creation of this happens in the
`CalcJob.run` call, which is quite far removed from the eventual use,
this opens it up for a lot of problems. The time between folder
population by `prepare_for_submission` and eventual use in the upload
task can be significant, among other things because the upload task will
have to wait for transport. To make this feasible the creation of the
sandbox folder has to be moved closer to the upload task.

Here we move the `presubmit` call from `CalcJob.run` to the upload
transport task which will create the sandbox folder and pass it into the
`upload_calculation`. This limits the time frame in which there is a
chance for the contents of the sandbox to get lost. In addition, this
now gives additional freedom of deciding which files from the sandbox
are permanently stored in the node's repository for extra provenance.

### DIFF
--- a/aiida/engine/processes/calcjobs/calcjob.py
+++ b/aiida/engine/processes/calcjobs/calcjob.py
@@ -214,9 +214,21 @@ class CalcJob(Process):
         This means invoking the `presubmit` and storing the temporary folder in the node's repository. Then we move the
         process in the `Wait` state, waiting for the `UPLOAD` transport task to be started.
         """
-        from aiida.orm import Code, load_node
-        from aiida.common.folders import SandboxFolder, SubmitTestFolder
-        from aiida.common.exceptions import InputValidationError
+        if self.inputs.metadata.dry_run:
+            from aiida.common.folders import SubmitTestFolder
+            from aiida.engine.daemon.execmanager import upload_calculation
+            from aiida.transports.plugins.local import LocalTransport
+
+            with LocalTransport() as transport:
+                with SubmitTestFolder() as folder:
+                    calc_info, script_filename = self.presubmit(folder)
+                    transport.chdir(folder.abspath)
+                    upload_calculation(self.node, transport, calc_info, folder, inputs=self.inputs, dry_run=True)
+                    self.node.dry_run_info = {
+                        'folder': folder.abspath,
+                        'script_filename': script_filename
+                    }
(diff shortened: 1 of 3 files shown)


aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -315,17 +291,27 @@ def presubmit(self, folder):
...
 
+        if not self.inputs.metadata.dry_run and self.node.has_cached_links():
+            raise InvalidOperation('calculation node has unstored links in cache')
+
         codes = [_ for _ in inputs.all_nodes() if isinstance(_, Code)]
 
-        calcinfo = self.prepare_for_submission(folder)
-        scheduler = computer.get_scheduler()
+        for code in codes:
+            if not code.can_run_on(computer):
```
muhrin: is this the equivalent of the previous `is_local`/`get_local_executable`/`get_content_list` check?

sphuber: It is a different check, this one was just done in the `run` method while the one you reference was being done in `presubmit`. I just moved the former here as it makes more sense here. I realize now I can merge the two loops. I will do that

sphuber: @muhrin I merged the two checks in the same loop, so this should be good to go

PR #3669, opened 2019-12-14
Allow documentation to be build without installing and configuring AiiDA

### LINKED ISSUE #2263: Docs failed to make when the default profile has SQLA backend
When the default profile has SQLA as the backend the doc cannot be built using:
`cd docs && make`
The user will be greeted with error:
~~~~
building [mo]: targets for 0 po files that are out of date                                         
building [html]: targets for 174 source files that are out of date                                 
updating environment: 174 added, 0 changed, 0 removed                                              
/home/bonan/aiida_core/aiida/work/workfunctions.py:20: AiidaDeprecationWarning: this module has bee
n deprecated, import directly from `aiida.work` instead                                            
  warnings.warn('this module has been deprecated, import directly from `aiida.work` instead', Depre
cationWarning)  # pylint: disable=no-member                                                        
reading sources... [100%] working_with_aiida/scripting                                             
                                                                                                   
Warning, treated as error:                                                                         
autodoc: failed to import module 'backup_django' from module 'aiida.common.additions.backup_script'
; the following exception was raised:                                                              
  File "/home/bonan/miniconda3/envs/aiidadev/lib/python3.6/site-packages/sphinx/ext/autodoc/importe
r.py", line 154, in import_module                                                                  
    __import__(modname)                                                                            
  File "/home/bonan/aiida_core/aiida/common/additions/backup_script/backup_django.py", line 16, in 
<module>                                                                                           
    from aiida.backends.djsite.db.models import DbNode                                             
  File "/home/bonan/aiida_core/aiida/backends/djsite/db/models.py", line 99, in <module>           
    class DbUser(AbstractBaseUser, PermissionsMixin):                                              
  File "/home/bonan/miniconda3/envs/aiidadev/lib/python3.6/site-packages/django/db/models/base.py",
 line 118, in __new__                                                                              
    "INSTALLED_APPS." % (module, name)                                                             
RuntimeError: Model class aiida.backends.djsite.db.models.DbUser doesn't declare an explicit app_la
bel and isn't in an application in INSTALLED_APPS.                                                 
                                                                                                   
Makefile:23: recipe for target 'default' failed                                                    
make: *** [default] Error 2                                                                        
~~~~

However, the documentation can be built when the default profile has django backend.

### LINKED ISSUE #3390: interface for loading dummy profile

sphuber:
Fixes #3390 and fixes #2263 

I took PR #3607 as a base. First I reinstated the (unintentionally removed) loading of the Sphinx theme when building the documentation locally. Then I put one commit on top, see commit message for details. Effectively this provides a single function in the docs `conf.py` that fully loads a dummy environment. This is a bit more verbose than the solution proposed by @ltalirz but this removes a lot of code from the main code base that is just there to be able to build the documentation without installation. I feel this does not belong there and should instead be in the documentation configuration. Plugins can simply copy this function and call it to copy the behavior.

In addition, this approach now also fixes another issue that required a profile with a django backend to be present locally in order to build the docs. This is no longer necessary and also locally one does not even have to have a profile installed.


docs/source/conf.py
```diff
@@ -18,17 +18,65 @@
...
+
+
+# Load the dummy profile even if we are running locally, this way the documentation will succeed even if the current
+# default profile of the AiiDA installation does not use a Django backend.
+load_documentation_profile()
+
+
+# If we are not on READTHEDOCS load the Sphinx theme manually
+if not os.environ.get('READTHEDOCS', None):
+    try:
```
ltalirz: try/except should not be needed here since we have the package in the docs dependencies

ltalirz:
I would put it in `aiida.manage.configuration` and not expose it on the top level.

docs/source/conf.py
sphuber: Done

sphuber:
> I would put it in `aiida.manage.configuration` and not expose it on the top level.

Of fair enough, I have moved the function into `aiida.manage.configuration`

PR #3675, opened 2019-12-16
Update `pyyaml` to prevent arbitrary code execution 

### LINKED ISSUE #3310: consider switch form pyyaml to ruamel
See discussion [here](https://github.com/aiidateam/aiida-core/issues/2550#issuecomment-473189766) by @giovannipizzi , @sphuber , @chrisjsewell

sphuber:
Fixes #3310

Before `pyyaml==5.1` the `yaml.load` function was vulnerable to
arbitrary code execution, because it loaded the full set of YAML. There
was an alternative `safe_load` but this was not the default and could
only load a sub set of the markup language. The new version of pyyaml
deprecates the old vulnerable code and provides the `FullLoader` that
can load the full set without being vulnerable.

### DIFF
--- a/aiida/backends/tests/cmdline/commands/test_data.py
+++ b/aiida/backends/tests/cmdline/commands/test_data.py
@@ -29,7 +29,7 @@ from aiida.orm.nodes.data.cif import has_pycifrw
 from aiida.orm import Group, ArrayData, BandsData, KpointsData, CifData, Dict, RemoteData, StructureData, TrajectoryData
 
 
-class TestVerdiDataExportable:
+class DummyVerdiDataExportable:
     """Test exportable data objects."""
 
     NODE_ID_STR = 'node_id'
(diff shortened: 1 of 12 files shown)


sphuber:
@ltalirz I have marked it as blocked because it does quite a bit the same as your PR. Once yours is merged I will rebase to include the few remaining warning fixes.

We discussed a long time ago to use `ruamel.yaml` instead of sticking with `pyyaml` but the latter seems to be developed again and the vulnerability is fixed. In contrast, the former started really changing the interface where it was a big job to start adapting it all

ltalirz:
> Once yours is merged I will rebase to include the few remaining warning fixes.

Thanks, please go ahead!

> We discussed a long time ago to use ruamel.yaml instead of sticking with pyyaml but the latter seems to be developed again and the vulnerability is fixed. In contrast, the former started really changing the interface where it was a big job to start adapting it all.

Ok, no strong feelings about this. Feel free to close the corresponding issue.

sphuber:
This depends on a new version of `kiwipy` that I just released but that isn't on conda-forge yet.

sphuber:
@ltalirz this is unblocked and good to go

PR #3681, opened 2019-12-17
fix performance issue when exporting many groups

ltalirz:
When providing groups to `verdi export`, it was looping over all groups
and using the `Group.nodes` iterator to retrieve the nodes contained in
the groups. This results (at least) in one query per group, and is
therefore every inefficient for large numbers of groups. The new
implementation replaces this by a single query.

### DIFF
--- a/aiida/tools/importexport/dbexport/__init__.py
+++ b/aiida/tools/importexport/dbexport/__init__.py
@@ -181,13 +181,37 @@ def export_tree(
             )
 
     # Add all the nodes contained within the specified groups
-    for group in given_groups:
-        for entry in group.nodes:
-            entities_starting_set[NODE_ENTITY_NAME].add(entry.uuid)
-            if issubclass(entry.__class__, orm.Data):
-                given_data_entry_ids.add(entry.pk)
-            elif issubclass(entry.__class__, orm.ProcessNode):
-                given_calculation_entry_ids.add(entry.pk)
+    if given_group_entry_ids:
+
+        if not silent:
+            print('RETRIEVING NODES FROM GROUPS...')
+
+        # Use single query instead of given_group.nodes iterator for performance.
+        qh_groups = orm.QueryBuilder().append(
+            orm.Group, filters={
+                'id': {
+                    'in': given_group_entry_ids
+                }
+            }, tag='groups'
+        ).queryhelp
+
+        # Delete this import once the dbexport.zip module has been renamed
+        from builtins import zip  # pylint: disable=redefined-builtin
+
+        data_results = orm.QueryBuilder(**qh_groups).append(orm.Data, project=['id', 'uuid'], with_group='groups').all()
+        if data_results:
+            pks, uuids = map(list, zip(*data_results))
+            entities_starting_set[NODE_ENTITY_NAME].update(uuids)
+            given_data_entry_ids.update(pks)
+            del data_results, pks, uuids
+
+        calc_results = orm.QueryBuilder(**qh_groups
+                                       ).append(orm.ProcessNode, project=['id', 'uuid'], with_group='groups').all()
+        if calc_results:
+            pks, uuids = map(list, zip(*calc_results))
+            entities_starting_set[NODE_ENTITY_NAME].update(uuids)
+            given_calculation_entry_ids.update(pks)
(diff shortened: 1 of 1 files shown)


lekah:
Is there any benchmark where you compare times to get the result? It should depend on what group.nodes is exactly doing, but in the end I guess the performance bottleneck is loading the nodes, not the query. Just out of interest (since the title is "fix performance issue")

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -181,8 +181,18 @@ def export_tree(
...
+
+        if not silent:
+            print('RETRIEVING NODES FROM GROUPS...')
+
+        # Use single query instead of given_group.nodes iterator for performance.
+        group_qb = orm.QueryBuilder().append(orm.Group, filters={'id': {'in': given_group_entry_ids}}, tag='groups')
+        # Could project on ['id', 'uuid', 'node_type'] for further performance enhancement
+        group_qb.append(orm.Node, with_group='groups', project=['*'])
+
+        for row in group_qb.all():
```
lekah: Could you use group_qb.iterall() here? It might lead to further improvement because the results are fetched in batches and not all together (if the subsequent code also takes time this could speed up the time-to-solution).

ltalirz:
> Is there any benchmark where you compare times to get the result?

Here come some actual numbers.

Test set: export of 67k groups containing 5 nodes each.
Old implementation: 224s
New implementation with `qb.all()`: 64s
New implementation with `qb.iterall()`: 59s

> It should depend on what group.nodes is exactly doing, but in the end I guess the performance bottleneck is loading the nodes, not the query. 

My guess was that the bottleneck are the N queries you make for N groups and I think the benchmark results back this up.

However, as you say and as I point out in the comment, there might still be significant speedup  from avoiding to construct the ORM objects altogether, as well as memory savings.

@sphuber: Do you happen to know whether there is an easy check to replace these checks
https://github.com/aiidateam/aiida-core/blob/9172c60703897ec6686e226462943ac164663e47/aiida/tools/importexport/dbexport/__init__.py#L232-L235
by checks on the `node_type`?

sphuber:
Yes, you can select those sets of nodes directly on the `node_type`.
For `Data` nodes:
```
SELECT * FROM db_dbnode WHERE node_type LIKE = 'data.%';
```
and for `ProcessNode` nodes, i.e. all other
```
SELECT * FROM db_dbnode WHERE node_type LIKE = 'process.%';
```

ltalirz:
Thanks @sphuber - the latest implementation now takes 11s on the test set (down from 204s initially).

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -181,13 +181,34 @@ def export_tree(
...
+            orm.Group, filters={
+                'id': {
+                    'in': given_group_entry_ids
+                }
+            }, tag='groups'
+        ).queryhelp
+
+        data_results = orm.QueryBuilder(**qh_groups).append(orm.Data, project=['id', 'uuid'], with_group='groups').all()
+
+        from builtins import zip  # pylint: disable=redefined-builtin
```
ltalirz: Without this I get an error that `zip` is a module and not callable ( we have a module called `zip` in the same folder).
If you want me to rename the module instead, let me know.

ltalirz:
@sphuber This is ready for review

ltalirz:
Do you want me to write the commit message?
Or are you waiting for another review?

sphuber:
> Do you want me to write the commit message?
> Or are you waiting for another review?

No, I was just leaving the honors to yourself, commit message looks great :)

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -181,13 +181,37 @@ def export_tree(
...
+        # Use single query instead of given_group.nodes iterator for performance.
+        qh_groups = orm.QueryBuilder().append(
+            orm.Group, filters={
+                'id': {
+                    'in': given_group_entry_ids
+                }
+            }, tag='groups'
+        ).queryhelp
+
+        # Delete this import once the dbexport.zip module has been renamed
```
CasperWA: For the future: This should be changed in #3242

PR #3682, opened 2019-12-17
Improve error message in the QB

ramirezfranciscof:
Previous message was not very helpful.

### DIFF
--- a/aiida/orm/implementation/django/querybuilder.py
+++ b/aiida/orm/implementation/django/querybuilder.py
@@ -206,11 +206,11 @@ class DjangoQueryBuilder(BackendQueryBuilder):
                     ''.format(operator, value)
                 )
         elif operator == 'in':
+            if not value:
+                raise InputValidationError('Value for operator \'in\' is an empty list')
             value_type_set = set(type(i) for i in value)
             if len(value_type_set) > 1:
-                raise InputValidationError('{}  contains more than one type'.format(value))
-            elif not value_type_set:
-                raise InputValidationError('{}  contains is an empty list'.format(value))
+                raise InputValidationError('Value for operator \'in\' contains more than one type: {}'.format(value))
         elif operator in ('and', 'or'):
             expressions_for_this_path = []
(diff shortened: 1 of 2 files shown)


aiida/orm/implementation/django/querybuilder.py
```diff
@@ -206,11 +206,11 @@ def get_filter_expr(self, operator, value, attr_key, is_attribute, alias=None, c
                     ''.format(operator, value)
                 )
         elif operator == 'in':
+            if not value:
```
sphuber: what if `value` is not even an interable? This will raise a `TypeError`, so maybe it is better to first check that a list/tuple is passed.

aiida/orm/implementation/django/querybuilder.py
```diff
@@ -206,11 +206,11 @@ def get_filter_expr(self, operator, value, attr_key, is_attribute, alias=None, c
...
+                raise InputValidationError('Value for operator \'in\' is an empty list')
```
sphuber: We indeed prefer single quotes for string literals but not when that requires escaping internal quotes. In this case favor 
```
'Value for operator `in` is an empty list'
```
or double quotes
```
'Value for operator "in" is an empty list'
```
or if you really want single quotes

```
"Value for operator 'in' is an empty list"
```

aiida/orm/implementation/django/querybuilder.py
ramirezfranciscof: Change implemented. In the end I ended up using a try/except statement instead of checking for explicit types (`value` could also be a set maybe?) because apparently "checking for iterables" is kind of controversial and this is a more "pythonic" approach ([sauce](https://stackoverflow.com/a/1952481/638366)). Let me know if you still prefer explicit type checking.

sphuber: Normally that depends on your interface whether it declares a specific type, but in this case I don't think we do so I like your solution :+1:

PR #3685, opened 2019-12-18
Consider 'AIIDA_TEST_PROFILE' in 'get_test_backend_name'.

### LINKED ISSUE #3659: pytest follow-up
This is a to-do list of things to complete, following the switch to pytest for testing

 - [x] Update documentation & wiki (`verdi devel tests` => `pytest`)
 - [x] Fix (py)test warnings (will touch lots of files)
 - [ ] Move test files out of aiida-core inside top-level `tests/` folder
 - [x] Decide whether we need to improve test discovery (see [comment](https://github.com/aiidateam/aiida-core/pull/3653#discussion_r357292341))

Then, over time, where it makes sense, we should start moving existing tests from the `AiiDATestCase` to pytest (e.g. where pytest fixtures are making things more effective)

greschd:
Change the logic of `get_test_backend_name` to check (if given) the backend configured for the profile specified in `AIIDA_TEST_PROFILE`
- If only one of the two is set (`AIIDA_TEST_BACKEND` or `AIIDA_TEST_PROFILE`), it will return the corresponding backend
- If both are set, it checks that they match, raising `ValueError` otherwise
- If neither is specified, fall back to the default django backend.

Affects #3659.

### DIFF
--- a/aiida/manage/tests/__init__.py
+++ b/aiida/manage/tests/__init__.py
@@ -449,18 +449,33 @@ def test_manager(backend=BACKEND_DJANGO, profile_name=None, pgtest=None):
 
 
 def get_test_backend_name():
-    """ Read name of database backend from environment variable.
+    """ Read name of database backend from environment variable or the specified test profile.
 
-    Reads database backend ('django' or 'sqlalchemy') from 'AIIDA_TEST_BACKEND' environment variable.
+    Reads database backend ('django' or 'sqlalchemy') from 'AIIDA_TEST_BACKEND' environment variable,
+    or the backend configured for the 'AIIDA_TEST_PROFILE'.
     Defaults to django backend.
 
     :returns: content of environment variable or `BACKEND_DJANGO`
     :raises: ValueError if unknown backend name detected.
+    :raises: ValueError if both 'AIIDA_TEST_BACKEND' and 'AIIDA_TEST_PROFILE' are set, and the two
+        backends do not match.
     """
-    backend_env = os.environ.get('AIIDA_TEST_BACKEND', BACKEND_DJANGO)
-    if backend_env in (BACKEND_DJANGO, BACKEND_SQLA):
-        return backend_env
-    raise ValueError("Unknown backend '{}' read from AIIDA_TEST_BACKEND environment variable".format(backend_env))
+    test_profile_name = get_test_profile_name()
+    backend_env = os.environ.get('AIIDA_TEST_BACKEND', None)
+    if test_profile_name is not None:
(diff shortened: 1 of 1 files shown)


greschd:
Didn't end up doing anything with the `pytest` hooks - I think making `get_test_backend_name` understand that there can also be a `AIIDA_TEST_PROFILE` is a cleaner solution.

Please double-check the way I'm loading the config / profile information, since I'm unfamiliar with this part of the code.

aiida/manage/tests/__init__.py
```diff
@@ -449,17 +449,32 @@ def aiida_profile():
...
+    :raises: ValueError if both 'AIIDA_TEST_BACKEND' and 'AIIDA_TEST_PROFILE' are set, and the two
+        backends do not match.
     """
-    backend_env = os.environ.get('AIIDA_TEST_BACKEND', BACKEND_DJANGO)
-    if backend_env in (BACKEND_DJANGO, BACKEND_SQLA):
-        return backend_env
+    test_profile_name = get_test_profile_name()
+    backend_env = os.environ.get('AIIDA_TEST_BACKEND', None)
+    if test_profile_name is not None:
+        backend_profile = configuration.load_config().get_profile(test_profile_name).database_backend
```
greschd: Is this the right way to access the profile information, or is there a better way?

greschd:
No idea why pre-commit is acting up -- seems similar to what was fixed in #3668. It seems the problem doesn't always happen, though.

aiida/manage/tests/__init__.py
ltalirz: Almost, I think `get_config()` (rather than `load_config()`) would be the appropriate function here

greschd: Fixed!

greschd:
Well, I could add a test where ``os.environ`` is changed manually.. but that would end up with almost the same logic as the function itself for figuring out what the result should be, so I'm not sure if it would be useful at all.

greschd:
On a related note, should `get_test_profile_name` return the name of the automatically-created profile when `AIIDA_TEST_PROFILE` is _not_ set?

ltalirz:
> On a related note, should get_test_profile_name return the name of the automatically-created profile when AIIDA_TEST_PROFILE is not set?

I don't see how it can - what if it is called before the test manager has taken over?
And even if it is called after the test manager has taken over, should it just return the default profile?
Maybe I'm missing what you have in mind?

greschd:
I think you're right -- I was thinking of `aiida-pytest`, where I was only ever creating one automatic profile.

PR #3693, opened 2019-12-19
Update `prospector` to the latest version `v1.2.0`

### LINKED ISSUE #3688: Intermittent pre-commit failures on github actions.
The following pre-commit error seems to occur with some probability on github actions:

```
Messages
========

aiida/schedulers/plugins/slurm.py
  Line: 586
    pylint: invalid-name / Attribute name "requested_wallclock_time_seconds" doesn't conform to '(([a-z][a-z0-9_]{2,30})|(_[a-z0-9_]*))$' pattern (col 16)
```

The error is explicitly silenced with a ``# pylint: disable=invalid-name``, and I haven't seen it occur locally. Feels related to #3668, in that it could have the same root cause.

sphuber:
Fixes #3688 

This comes with new requirement of `pylint==2.4.4` which comes with a
few new warnings that have been addressed in the code base:

 * no-else-break
 * no-else-continue
 * self-assigning-variable
 * property-with-parameters
 * invalid-overridden-method
 * unnecessary-comprehension

One additional new warning has been added to the ignore list

 * import-outside-toplevel

PEP8 recommends to only have imports at the toplevel of a module and not
in function or class definitions. However, the changes required would be
significant and so we leave that for another time.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -50,7 +50,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=bad-continuation,locally-disabled,useless-suppression,django-not-available,bad-option-value,logging-format-interpolation,no-else-raise
+disable=bad-continuation,locally-disabled,useless-suppression,django-not-available,bad-option-value,logging-format-interpolation,no-else-raise,import-outside-toplevel
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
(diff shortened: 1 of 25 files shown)


sphuber:
It is difficult to say with certainty but I think the transient pylint problem is fixed. It kept complaining about a line in the `SlurmScheduler` code but it had a disable statement. I noticed that the same variable was used in the corresponding test file and there was not adorned by an inline disabling. It might have been that one triggering pylint which sadly gave the wrong source location. Still seems far fetched but I am out of other possibilities :)

ltalirz:
>  It might have been that one triggering pylint which sadly gave the wrong source location.

I did come across another case, where it was not enough to ignore issues at the definition, but also where they were used, so it's possible.

greschd:
#3688 still seems to happen - maybe it needs to be explicitly silenced in the `lsf` plugin also?

PR #3695, opened 2019-12-20
Ensure correct types for `QueryBuilder().dict()` with multiple projections

### LINKED ISSUE #3679: Projecting more than just `*` with `QueryBuilder.iterdict` results in `BackendEntity` instances being returned.
Reported by @zooks97

Projecting only `*` when using `dict` works fine:
```
In [11]: QueryBuilder().append(Node, project=['*']).limit(2).dict()
Out[11]: 
[{'node_1': {'*': <Int: uuid: 1984c02b-75c2-43ae-a065-e0a10124977b (pk: 13641) value: 2>}},
 {'node_1': {'*': <FolderData: uuid: edf5d9bf-4dcf-49fb-866f-a38fde52f488 (pk: 112)>}}]
```
However, when adding another projection, the entities are `BackendNode` instances instead of `Node` instances:
```
In [12]: QueryBuilder().append(Node, project=['*', 'id']).limit(2).dict()
Out[12]: 
[{'node_1': {'*': <aiida.orm.implementation.django.nodes.DjangoNode at 0x7ff7f15ffe48>,
   'id': 13641}},
 {'node_1': {'*': <aiida.orm.implementation.django.nodes.DjangoNode at 0x7ff7f15fff28>,
   'id': 112}}]
```
Switching to the `all` resolver works as expected:
```
In [13]: QueryBuilder().append(Node, project=['*', 'id']).limit(2).all()
Out[13]: 
[[<Int: uuid: 1984c02b-75c2-43ae-a065-e0a10124977b (pk: 13641) value: 2>,
  13641],
 [<FolderData: uuid: edf5d9bf-4dcf-49fb-866f-a38fde52f488 (pk: 112)>, 112]]
```

sphuber:
Fixes #3679 

The returned results by the backend implementation of `QueryBuilder` are
passed through `get_aiida_entity_res` to convert backend instances to
front end class instances. However, in the `iterdict` accumulator this
was not accounting for mappings and sequences. This would cause problems
when multiple projections were used, such as `project=['*', 'id']`. The
'*' notation will cause the entire node instance to be loaded, however
due to the fact that it was inside a dictionary in the response it would
not be converted.

### DIFF
--- a/aiida/orm/querybuilder.py
+++ b/aiida/orm/querybuilder.py
@@ -18,7 +18,6 @@ an interface classes which enforces the implementation of its defined methods.
 An instance of one of the implementation classes becomes a member of the :func:`QueryBuilder` instance
 when instantiated by the user.
 """
-
 # Checking for correct input with the inspect module
 from inspect import isclass as inspect_isclass
 import copy
--- a/aiida/backends/tests/orm/test_querybuilder.py
+++ b/aiida/backends/tests/orm/test_querybuilder.py
@@ -363,6 +363,21 @@ class TestQueryBuilder(AiidaTestCase):
         self.assertTrue(id(query1) != id(query2))
         self.assertTrue(id(query2) == id(query3))
 
+    def test_dict_multiple_projections(self):
+        """Test that the `.dict()` accumulator with multiple projections returns the correct types."""
+        node = orm.Data().store()
+        builder = orm.QueryBuilder().append(orm.Data, project=['*', 'id'])
+        results = builder.dict()
+
+        self.assertIsInstance(results, list)
+        self.assertTrue(all(isinstance(value, dict) for value in results))
+
+        dictionary = list(results[0].values())[0]  # `results` should have the form [{'Data_1': {'*': Node, 'id': 1}}]
+
+        self.assertIsInstance(dictionary['*'], orm.Data)
+        self.assertEqual(dictionary['*'].pk, node.pk)
+        self.assertEqual(dictionary['id'], node.pk)
+
     def test_operators_eq_lt_gt(self):
         nodes = [orm.Data() for _ in range(8)]
 
--- a/aiida/orm/convert.py
+++ b/aiida/orm/convert.py
@@ -25,7 +25,38 @@ def get_orm_entity(backend_entity):
 
 @get_orm_entity.register(Mapping)
 def _(backend_entity):
-    return {key: get_orm_entity(value) for key, value in backend_entity.items()}
+    """Convert all values of the given mapping to ORM entities if they are backend ORM instances."""
+    converted = {}
+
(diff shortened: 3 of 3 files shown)


aiida/orm/querybuilder.py
```diff
@@ -1994,15 +1995,20 @@ def get_query(self):
                 self._hash = queryhelp_hash
         return query
 
-    @staticmethod
-    def get_aiida_entity_res(value):
+    def get_aiida_entity_res(self, value):
```
ltalirz: could be a `@classmethod`, right?

aiida/orm/querybuilder.py
```diff
@@ -1994,15 +1995,20 @@ def get_query(self):
...
         """Convert a projected query result to front end class if it is an instance of a `BackendEntity`.
 
         Values that are not an `BackendEntity` instance will be returned unaltered
 
         :param value: a projected query result to convert
         :return: the converted value
         """
+        if isinstance(value, collections.abc.Mapping):
+            return {key: self.get_aiida_entity_res(sub_value) for key, sub_value in value.items()}
```
ltalirz: I wonder, should this not rather be fixed here:
https://github.com/aiidateam/aiida-core/blob/81d736c3c2f98c3f5ac02a3e6b21224e2e5bedf0/aiida/orm/convert.py#L26-L28

Also, it looks very similar, so I wonder why this is not working...
Is just the list/tuple stuff missing from there?

sphuber: You are right, it should be handled there but the problem is that there a `TypeError` can be raised if some of the elements are not node types, which is then caught by `get_aiida_entity_res` and it then returns the original value which is the original dictionary with unconverted values. I will fix it correctly

aiida/orm/querybuilder.py
sphuber: could be, but I think staticmethod makes more sense. Now that I fixed the problem in the correct place (thanks for catching this :+1:) I think we can keep this as was

sphuber:
> seems pylint tripped over again...
> can we actually fix this warning by using the proper capitalization?

I know, but I cannot reproduce it locally or in other PRs. I have been trying to fix it and already merged two more PRs in a desperate attempt to tackle this. It's infuriating. In any case it is not a problem iwth capitalization but just the fact that it is 32 characters instead of the "max allowed" 30.

ltalirz:
What if we increase the max allowed to 32?

sphuber:
The last commit I did may have done it: https://github.com/aiidateam/aiida-core/runs/358302396
But there is no certainty

sphuber:
> What if we increase the max allowed to 32?

As a last defense I just snuck this in this [last commit](https://github.com/aiidateam/aiida-core/commit/c3c9aaf4d60b58ee633b68bbffdcf40cc1f29ff2). Let's hope to all that is holy that that will do it

PR #3697, opened 2019-12-20
Remove `aiida.schedulers.plugins` files from pre-commit black list

sphuber:
Except for the test files.

I hate having to put in these minimal shitty commits into `develop` but I don't know what else to do.
As usual this works in my fork and probably this PR is going to succeed.
Why is there no throw-your-hands-in-the-air-emoji?


greschd:
That's what (╯°□°）╯︵ ┻━┻ is for

sphuber:
> That's what (╯°□°）╯︵ ┻━┻ is for

Thanks, I will be taking this now

PR #3702, opened 2019-12-24
[BLOCKED] Move `CalcJob` spec validator to corresponding namespaces

### LINKED ISSUE #3431: Move `resource` validation of `CalcJob` from global validator to port itself
The `metadata.options.resources` of a `CalcJob` are validated by a validator on the global `inputs` namespace of the process. This makes it difficult for sub classes to control this behavior. If a sub class wants to make the `resources` optional, for example to implement automatic parallelization that would set the resources automatically, the validation would always fail.

### LINKED ISSUE #3449: `Calcjob` input validation is incompatible with `expose_inputs`
As noted in aiidateam/aiida-quantumespresso#418, 9ebb8cc17e2871e196598ef5ee2d6db820a156d8 has introduced an issue if one wants to expose a subset of inputs for a `Calcjob`, within a `WorkChain.define`, since validation will fail if, for example, the `code` input is not present.

sphuber:
Fixes #3449 and fixes #3431 

The `CalcJob` process had a single validator defined on the top level
input namespace, which validated many ports scattered across the
namespace, such as the `metadata.options.parser_name` as well as the
`metadata.options.resources`. The validator assumed that all of these
ports would always be present, however, this is not necessarily true.
The expose functionality allows a wrapping process to expose only part
of the namespace but the validator remains the same.

To ammeliorate this, the signature of validators is updated to also
receive a `context` in the form of a second argument `ctx` in addition
to the value passed to the port. This `ctx` will be the instance of the
port to which the validator is assigned. This allows the validator
implementation to first check whether a specific port is present before
trying to validate the corresponding value.

Note that the `ctx` will represent the port to which the validator is
attached and so it will have no knowledge of any namespace it might be
embedded in, as it shouldn't, because that will break the portability of
the namespace.


sphuber:
This is blocked by [this PR on `plumpy`](https://github.com/aiidateam/plumpy/pull/141). Note that I could keep the exact validation behavior by splitting up the `CalcJob` inputs validator except for the `resources`. By placing it on the `metadata.options.resources` namespace itself, I no longer have access to the `scheduler` class, which is defined through the computer. If we think this validation is crucial I can move it back to validation on the top level.

PR #3703, opened 2019-12-27
Update dependeny requirement `circus~=0.16.0`

### LINKED ISSUE #3559: ipython 5.8.0 incompatible with Python 3.8
It seems that for Python 3.8  `ipython>=7.3` is required: https://ipython.org/news.html#ipython-7-3

ipython 5.8.0 is broken with Python 3.8:

```
Python 3.8.0 (default, Oct 23 2019, 18:51:26) 
Type "copyright", "credits" or "license" for more information.

IPython 5.8.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.
/usr/lib64/python3.8/codeop.py in __call__(self, source, filename, symbol)
    131 
    132     def __call__(self, source, filename, symbol):
--> 133         codeob = compile(source, filename, symbol, self.flags, 1)
    134         for feature in _features:
    135             if codeob.co_flags & feature.compiler_flag:

TypeError: required field "type_ignores" missing from Module
```

sphuber:
Fixes #3559 

This new release of `circus` makes it compatible with `pyzmq>=17`. This
in turn allows us to unpin the `ipython` requirement which required
these versions of `pyzmq`. Note that implicitly the upper limit is
`7.10` for `ipython` because there python 3.5 support is dropped which
we support for another 9 months.

### DIFF
--- a/docs/requirements_for_rtd.txt
+++ b/docs/requirements_for_rtd.txt
@@ -3,7 +3,7 @@ aiida-export-migration-tests==0.8.0
 aldjemy~=0.9.1
 alembic~=1.2
 ase~=3.18
-circus~=0.15.0
+circus~=0.16.1
 click-completion~=0.5.1
 click-config-file~=0.5.0
 click-spinner~=0.1.8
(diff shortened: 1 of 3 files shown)


sphuber:
Note that I will update this PR to require at least `0.16.1` because the tarball uploaded to PyPI for `0.16.0` was incorrect. Installing from source will fail in that case

ltalirz:
conda-forge PR merged

Issue #2579, opened 2019-03-06
Transports should provide option to put object from filelike


sphuber:
Currently the `Transport` interface only supports the creation of objects directly from paths on the local file system through `put`, `putfile` and `puttree`. However, since we are moving towards a file repository system that is not restricted to a local file system, often a local absolute path is not available but instead we only have a filelike object. Paramiko's SFTP module, which we use for the `SshTransport`, supports putting files from filelikes wit the `putfo` method. We should implement this. When this is done, be sure to replace the current use of `NamedTemporaryFile` in the `aiida.engine.daemon.execmanager.upload_calculation` function to copy files from the `local_copy_list`.

PR #3707, opened 2020-01-06
Fix bug in `upload_calculation` for `CalcJobs` with local codes

### LINKED ISSUE #3705: Error when using local code: uses inexistent 'get_folder_list' method
When trying to run a local code, I get the following error:

```
+-> ERROR at 2019-12-29 16:25:26.778537+00:00
 | Traceback (most recent call last):
 |   File "/home/a-dogres/programming/aiida/aiida-core/aiida/engine/utils.py", line 172, in exponential_backoff_retry
 |     result = yield coro()
 |   File "/home/a-dogres/.virtualenvs/aiida/lib/python3.7/site-packages/tornado/gen.py", line 1055, in run
 |     value = future.result()
 |   File "/home/a-dogres/.virtualenvs/aiida/lib/python3.7/site-packages/tornado/concurrent.py", line 238, in result
 |     raise_exc_info(self._exc_info)
 |   File "<string>", line 4, in raise_exc_info
 |   File "/home/a-dogres/.virtualenvs/aiida/lib/python3.7/site-packages/tornado/gen.py", line 1069, in run
 |     yielded = self.gen.send(value)
 |   File "/home/a-dogres/programming/aiida/aiida-core/aiida/engine/processes/calcjobs/tasks.py", line 82, in do_upload
 |     execmanager.upload_calculation(node, transport, calc_info, folder)
 |   File "/home/a-dogres/programming/aiida/aiida-core/aiida/engine/daemon/execmanager.py", line 142, in upload_calculation
 |     for f in code.get_folder_list():
 | AttributeError: 'Code' object has no attribute 'get_folder_list'
```

This comes from https://github.com/aiidateam/aiida-core/blob/develop/aiida/engine/daemon/execmanager.py#L142, which appears to be the instance of ``get_folder_list`` in the AiiDA codebase. I think this using the old node <-> repository interface (which is now removed), and should be updated to use the new one.

sphuber:
Fixes #3705 

This code path was not tested at all and so the remaining occurrence of
`get_folder_list` which is part of the old repository interface that was
removed in `v1.0.0` went unnoticed.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -47,7 +47,7 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
     link_label = 'remote_folder'
     if node.get_outgoing(RemoteData, link_label_filter=link_label).first():
         execlogger.warning('CalcJobNode<{}> already has a `{}` output: skipping upload'.format(node.pk, link_label))
-        return calc_info, script_filename
+        return calc_info
 
     computer = node.computer
 
@@ -139,8 +139,14 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
     for code in input_codes:
         if code.is_local():
             # Note: this will possibly overwrite files
-            for f in code.get_folder_list():
-                transport.put(code.get_abs_path(f), f)
+            for f in code.list_object_names():
+                # Note, once #2579 is implemented, use the `node.open` method instead of the named temporary file in
+                # combination with the new `Transport.put_object_from_filelike`
+                # Since the content of the node could potentially be binary, we read the raw bytes and pass them on
+                with NamedTemporaryFile(mode='wb+') as handle:
+                    handle.write(code.get_object_content(f, mode='rb'))
+                    handle.flush()
+                    transport.put(handle.name, f)
             transport.chmod(code.get_local_executable(), 0o755)  # rwxr-xr-x
 
     # In a dry_run, the working directory is the raw input folder, which will already contain these resources
@@ -192,7 +198,6 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
             with NamedTemporaryFile(mode='wb+') as handle:
                 handle.write(data_node.get_object_content(filename, mode='rb'))
                 handle.flush()
-                handle.seek(0)
                 transport.put(handle.name, target)
 
     if dry_run:
--- a/aiida/backends/tests/engine/test_calc_job.py
+++ b/aiida/backends/tests/engine/test_calc_job.py
@@ -10,6 +10,7 @@
 """Test for the `CalcJob` process sub class."""
 from copy import deepcopy
 from functools import partial
+import os
 from unittest.mock import patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


aiida/engine/daemon/execmanager.py
```diff
@@ -139,8 +139,12 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
     for code in input_codes:
         if code.is_local():
             # Note: this will possibly overwrite files
-            for f in code.get_folder_list():
-                transport.put(code.get_abs_path(f), f)
+            for f in code.list_object_names():
+                with NamedTemporaryFile(mode='wb+') as handle:
+                    handle.write(code.get_object_content(f, mode='rb'))
+                    handle.flush()
+                    handle.seek(0)
```
greschd: Is the `seek` really necessary here? Doesn't `transport.put` open it itself anyway? If it's necessary, might warrant a comment as to why.

aiida/engine/daemon/execmanager.py
```diff
@@ -139,8 +139,12 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
...
+                    transport.put(handle.name, f)
```
ramirezfranciscof: This might be because I'm still not familiar with `transport` and `code` objects, but from the current commit description (which only mentions that `get_folder_list` no longer exists) I don't understand why the change in the parameters of `transport.put(...)`. It seems to change the loop over folders for a loop over objects (file names, correct?), but then I don't understand why you make a temporary file and pass the name of that instead of something related to the looping objects. Also, I don't understand the process of writing content, flushing and rewinding said temporary file.

Again, I'm not sure if this merits any code comment, but maybe just a brief explanation for this changes in the commit message?

greschd: That change is because the new API doesn't expose directly an absolute path to the file to be put on the remote computer (which wouldn't be possible on an object store backend). Since `transport.put` wants a filename, the content is first copied to a temporary file.

This part was clear to me, but I have been doing a fair bit of migrating between these two.. so maybe the commit could indeed be more explicit about that.

aiida/engine/daemon/execmanager.py
```diff
@@ -139,8 +139,12 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
     for code in input_codes:
         if code.is_local():
             # Note: this will possibly overwrite files
-            for f in code.get_folder_list():
-                transport.put(code.get_abs_path(f), f)
+            for f in code.list_object_names():
```
greschd: This doesn't work if `list_object_names` contains directories, right? Is that ensured not to happen elsewhere?

sphuber:
> Looks good. I assume you checked that the test fails before the fix?

Yes, it failed.

aiida/engine/daemon/execmanager.py
sphuber: You are right, it is not necessary, I will remove it. The code is a copy of a block a few lines lower. There I already put a comment that it should be improved using `node.open` once `transport.put` supports filelike objects to be passed, which is issue #2579 . I decided not to copy the comment again.

aiida/engine/daemon/execmanager.py
sphuber: I will include the reasoning for this and a reference to #2579 which will allow this hack to be removed.

aiida/engine/daemon/execmanager.py
sphuber: Yes, it should never contain directotries. If you look at the `Code.set_files` implementation, which is called to store the files of the local code, it will copy all the files to top level without nesting. 
```
    def set_files(self, files):
        """
        Given a list of filenames (or a single filename string),
        add it to the path (all at level zero, i.e. without folders).
        Therefore, be careful for files with the same name!
```
I just noticed that I changed that code when the repository interface changed, and I am not sure it does 100% the same. The new implementation checks `os.path.isfile` which wasn't there explicitly. Old implementation:
```
for f in files:
    self.repository.add_path(f, os.path.split(f)[1])
```
However that also does not seem compatible with the docstring which says that you can pass both filenames and directories. I don't think that `add_path` was doing the correct hing either.

greschd: Well, keeping it as files-only seems like a safe thing to do for now -- we can always add directory compatibility if it's needed. Might be good to assert it in the current piece of code, so it's a bit more explicit this is not supported?

Side note: This `Therefore, be careful for files with the same name!` could easily be improved by just checking first there are no duplicate file names, right?

sphuber: Sure, but that seems a job for another time. If ever anyone actually starts even using local codes..

PR #3708, opened 2020-01-07
Improve SqlAlchemy session handling for `QueryBuilder`

### LINKED ISSUE #3372: Use Aldjemy to create the session for Django `QueryBuilder` implementation
Currently we use the one from `aiida.backends.sqlalchemy`, but this will most likely not use the session correctly and will cause problems. We should see if we can use the session managed by aldjemy instead.

In `aiida.orm.implementation.django.querybuilder` replace the `get_session` with something like:
```
def get_session():
    from aldjemy.orm import get_session
    return get_session()
```
This fixes the following test:
```
from aiida.backends.djsite.db.models import DbNode, DbUser
from django.db import transaction
with transaction.atomic():                                
    user = DbUser(email='aiida@localhost2', first_name='t', last_name='t', institution='t')
    user.save()    
    node = DbNode(user_id=user.pk)
    node.save()

    # Loading it through the Django ORM should work, even when in a transaction
    loaded = DbNode.objects.get(pk=node.pk, user_id=user.pk)

    # Try to load it through the QueryBuilder which will fail if the session of SqlAlchemy is not correctly setup
    assert QueryBuilder().append(Node, filters={'id': node.pk}).count() != 0
```

However, the unittests will still fail so the implementation of `get_session` is not enough. Probably some more configuration is necesary

### LINKED ISSUE #3645: Sessions not handled correctly in QB for asynchronous function calls

sphuber:
Fixes #3645 and fixes #3372

### DIFF
--- a/aiida/backends/djsite/manager.py
+++ b/aiida/backends/djsite/manager.py
@@ -32,7 +32,7 @@ class DjangoBackendManager(BackendManager):
 
     def reset_backend_environment(self):
         """Reset the backend environment."""
-        from aiida.orm.implementation.django.querybuilder import reset_session
+        from . import reset_session
         reset_session()
 
     def is_database_schema_ahead(self):
--- a/aiida/backends/djsite/manager.py
+++ b/aiida/backends/djsite/manager.py
@@ -32,6 +32,8 @@ class DjangoBackendManager(BackendManager):
 
     def reset_backend_environment(self):
         """Reset the backend environment."""
+        from aiida.orm.implementation.django.querybuilder import reset_session
+        reset_session()
 
     def is_database_schema_ahead(self):
         """Determine whether the database schema version is ahead of the code schema version.
--- a/aiida/backends/sqlalchemy/manager.py
+++ b/aiida/backends/sqlalchemy/manager.py
@@ -54,15 +54,12 @@ class SqlaBackendManager(BackendManager):
 
     def _load_backend_environment(self):
         """Load the backend environment."""
-        from . import reset_session
-        reset_session()
+        get_scoped_session()
 
     def reset_backend_environment(self):
         """Reset the backend environment."""
-        from aiida.backends import sqlalchemy
-        if sqlalchemy.ENGINE is not None:
-            sqlalchemy.ENGINE.dispose()
-        sqlalchemy.SCOPED_SESSION_CLASS = None
+        from . import reset_session
+        reset_session()
 
     def is_database_schema_ahead(self):
         """Determine whether the database schema version is ahead of the code schema version.
--- a/aiida/orm/implementation/querybuilder.py
+++ b/aiida/orm/implementation/querybuilder.py
@@ -7,8 +7,21 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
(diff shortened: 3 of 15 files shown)


CasperWA:
Concerning using this in a REST-API server, i.e., `aiida-optimade`, what do you suggest to do?
I have tried to retrieve the session via `get_manager().get_backend().get_session()` before handling a request and then calling `session.close()` at the end, but this still prompts a QueuePool overflow for both backends.

aiida/backends/djsite/manager.py
```diff
@@ -32,6 +32,8 @@ def _load_backend_environment(self):
 
     def reset_backend_environment(self):
         """Reset the backend environment."""
+        from aiida.orm.implementation.django.querybuilder import reset_session
```
CasperWA:
```suggestion
        from . import reset_session
```

aiida/backends/sqlalchemy/manager.py
```diff
@@ -62,7 +62,7 @@ def reset_backend_environment(self):
         from aiida.backends import sqlalchemy
         if sqlalchemy.ENGINE is not None:
             sqlalchemy.ENGINE.dispose()
-        sqlalchemy.SCOPED_SESSION_CLASS = None
+        sqlalchemy.SESSION_FACTORY = None
```
CasperWA: Shouldn't this method simply call `aiida.backends.sqlalchemy.reset_session()`?

aiida/backends/djsite/manager.py
CasperWA: I think this is still an issue?

aiida/orm/implementation/querybuilder.py
```diff
@@ -9,6 +9,12 @@
 ###########################################################################
 """Backend query implementation classes"""
```
CasperWA: Maybe note here that the general QueryBuilder here is based on the assumption that all queries are performed with SQLAlchemy, no matter the actual backend? I.e., just copy some stuff in here from your commit messages.

aiida/orm/implementation/sqlalchemy/backend.py
```diff
@@ -101,6 +98,18 @@ def transaction():
                 # Make sure to commit the outermost session
                 session.commit()
 
+    @staticmethod
+    def get_session():
+        """Return a database session that can be used by the `QueryBuilder` to perform its query.
+
+        If there is an exception within the context then the changes will be rolled back and the state will
```
CasperWA: Shouldn't this be under the description of the `transaction` method?

aiida/backends/sqlalchemy/manager.py
sphuber: Sure does :)

Issue #3712, opened 2020-01-15
verdi import exceeds maximum postgres string size


ltalirz:
On aiida-core `develop` (47cfe3400a7f5ef863e2a10b88afa6187459bb96, django backend), I've been trying to import an aiida export file with a rather large `data.json` (~2GB uncompressed):
```
$ unzip -l export_hcofs_co2.aiida data.json metadata.json
Archive:  export_hcofs_co2.aiida
error: End-of-centdir-64 signature not where expected (prepended bytes?)
  (attempting to process anyway)
warning [export_hcofs_co2.aiida]:  2718573025 extra bytes at beginning or within zipfile
  (attempting to process anyway)
  Length      Date    Time    Name
---------  ---------- -----   ----
2086189646  12-13-2019 08:56   data.json
 16216422  12-13-2019 08:57   metadata.json
---------                     -------
2102406068                     2 files
```

The export manages to go through everything until
```
...
STORING NEW NODE EXTRAS...
UPDATING EXISTING NODE EXTRAS (mode: kcl)
```
At this point, I get the the following error:
```
Error: an exception occurred while importing the archive export_hcofs_co2.aiida
Traceback (most recent call last):
  File "/Users/leopold/Applications/miniconda3/envs/discover-hcofs/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.ProgramLimitExceeded: out of memory
DETAIL:  Cannot enlarge string buffer containing 0 bytes by 1257443654 more bytes.
```
 1257443654 being ~1.2 GBytes.
Full [traceback attached](https://github.com/aiidateam/aiida-core/files/4059619/error.log).

This is actually not an out-of-memory error, but rather it seems to me that here
https://github.com/aiidateam/aiida-core/blob/a34b3566b5e98ebedb58f0f142cbf54808d146c0/aiida/tools/importexport/dbimport/backends/django/__init__.py#L474
 we are creating an INSERT string that exceeds the [max alloc size](https://github.com/postgres/postgres/blob/master/src/include/utils/memutils.h#L40) of postgresql, which defaults to 1GB.

ltalirz:
I've confirmed that this can be fixed by using  the `batch_size` of the `create_bulk` operation.
Will open a PR for this

ltalirz:
#3713 fixes this for the django backend.

Let's keep this issue open until the fix has also been applied to the sqla backend.

[...]

PR #3713, opened 2020-01-15
limit batch size for bulk_create operations

ltalirz:
partially addresses #3712 

postgresql has a "MaxAllocSize" that defaults to 1 GB [1].
If you try to insert more than that in one go (e.g. during import of a
large AiiDA export file), you encounter the error

    psycopg2.errors.ProgramLimitExceeded: out of memory
    DETAIL:  Cannot enlarge string buffer containing 0 bytes by 1257443654 more bytes.

This commit avoids this issue (for the django backend) by setting a
batch size for bulk_create operations.

[1] https://github.com/postgres/postgres/blob/master/src/include/utils/memutils.h#L40

Open questions:
 - [x] What should the batch size be? I guess it is in units of the django objects you are passing (see e..g [here](https://github.com/django/django/blob/eef3ea847e12f23db9a0af8efdc9543f24b780c4/django/db/models/query.py#L493) ), so the current value of 100k would allow each object to generate up to an average of 10 kB of strings per object (it did fix the issue for me)
 - [x] ~~Didn't look into sqla side, pointers welcome~~ let's fix this in another PR
 - [x] From the tests it seems the `BATCH_SIZE` variable should be moved somewhere else than the `settings.py`. Happy to move it wherever you guys think it should go.


ltalirz:
@sphuber I think Giovanni is quite busy now - would you mind having a look?
I think something like this should probably be in 1.1.0

sphuber:
Thanks @ltalirz good start, but couple things remain. I think we should also add this for SqlAlchemy. There is a page on [the documentation](https://docs.sqlalchemy.org/en/13/faq/performance.html#i-m-inserting-400-000-rows-with-the-orm-and-it-s-really-slow) about bulk insertions, the various available solutions and their performance. More detailed information on these methods [can be found here](https://docs.sqlalchemy.org/en/13/orm/persistence_techniques.html#bulk-operations). Currently, the SqlAlchemy import implementation just uses the `session.add` method, i.e. no bulk/batching whatsoever. I can have a look at some point if I can adapt it but not sure when I will have the time.

Since this functionality will have to be implemented for both backends, I think it makes sense to have the batch size be configured the same way. This would also allow us to make it configurable through `verdi config`. This solves your first checkbox, we simply go with that value for now and if that stops working for someone, they can simply adapt it through configuration. It also solves the second checkbox: the value will simply be taken from the configuration, for both backends.

ltalirz:
Putting it into the `config` makes sense to me. Shall we say `AIIDADB_BATCH_SIZE`?

As for adding it to sqla as well - in order to get at least the django fix into the code, I suggest I go ahead with django only for the moment and open an issue with pointers on how to proceed for sqla. Sounds good?

ltalirz:
I've updated the PR to make it a config option.
In the interest of time, I suggest we leave the sqla implementation to later.

aiida/tools/importexport/dbimport/backends/django/__init__.py
```diff
@@ -471,9 +472,9 @@ def import_data_dj(
                 if 'mtime' in [field.name for field in model._meta.local_fields]:
                     with models.suppress_auto_now([(model, ['mtime'])]):
                         # Store them all in once; however, the PK are not set in this way...
-                        model.objects.bulk_create(objects_to_create)
+                        model.objects.bulk_create(objects_to_create, batch_size=get_config_option('db.batch_size'))
```
sphuber: probably better to just fetch the value once and assign to local variable that is to be reused

aiida/manage/configuration/options.py
```diff
@@ -102,6 +38,16 @@ def parse_option(option_name, option_value):
...
     },
+    'db.batch_size': {
+        'key': 'db_batch_size',
+        'valid_type': 'int',
+        'valid_values': None,
+        'default': 100000,
+        'description':
+        'Batch size for bulk CREATE operations in the database. Avoids hitting MaxAllocSize of PostgreSQL'
+        '(1GB) when creating large numbers of database records in one go.',
+        'global_only': True,
```
sphuber: What is the reason to limit it to global option only? Maybe not the most common use case to have different batch sizes, but it is possible to have multiple profiles in one installation that connect to different databases on different machines with different memory limitations. Best to leave unconstrained I would say

aiida/tools/importexport/dbimport/backends/django/__init__.py
ltalirz: hm yes... I was thinking there might be cases where you are creating a new profile, where you don't really want to use the setting of the current profile, so I thought it might be safer to start with a global option.
Anyhow, I think making it not only global is fine

ltalirz:
Sorry for the long wait - was in meetings, meetings, meetings ;-)

aiida/tools/importexport/dbimport/backends/django/__init__.py
sphuber: I think you meant this to be a reply to the other comment ;) Anyway, that is not really what the `global_only` means. When an option is `global_only` it means it can only be set instance-wide and not per profile. Currently this only applies to the `user_*` options used for making repeated profile creation easier, so there it doesn't make sense as profile specific. The resolution order of configuration settings is `profile` -> `global` -> `default`. So if an existing profile defines a specific `db.batch_size` for itself, a new profile would still get the default, because there is nothing set globally.

sphuber:
> Sorry for the long wait - was in meetings, meetings, meetings ;-)

Not to worry, thanks a lot for the improvement :+1:

PR #3715, opened 2020-01-16
Move unit tests from `aiida` to top-level `tests`

### LINKED ISSUE #3659: pytest follow-up

sphuber:
Fixes #3659 

I have split the work in two commits to make inspecting the changes as easy as possible.
 * Move the files from `aiida` to `tests`. Besides adapting the `.pre-commit-config.yaml` this commit does not contain anything else and should not have to be inspected.
 * Adapt all imports in moved files to make things work again. This is the commit that should be looked at a bit.

Note that the solution of relative imports that I used is an ugly one I find. Most of the existing utilities can be cleaned up and be turned into pytest fixtures. However, I want to keep this for a separate PR.


tests/tools/importexport/migration/test_migration.py
```diff
@@ -13,12 +13,13 @@
 
 from aiida import orm
 from aiida.backends.testbase import AiidaTestCase
-from aiida.backends.tests.utils.archives import get_archive_file, get_json_files, migrate_archive
-from aiida.backends.tests.utils.configuration import with_temp_dir
 from aiida.tools.importexport import import_data, EXPORT_VERSION as newest_version
 from aiida.tools.importexport.migration import migrate_recursively, verify_metadata_version
 from aiida.common.utils import Capturing
 
+from ....utils.archives import get_archive_file, get_json_files, migrate_archive
```
ltalirz: that's 4 dots which makes things hard to read... is there a better way to do this?

sphuber: Yeah, as noted in the commit message this is a temporary fix in order to have minimal changes and to make it easy to track where stuff went. Most of these can/should be reimplemented as pytest fixtures which would get rid of the relative imports

sphuber:
> thanks @sphuber for making progress with this!

Thanks for the review

> If you look at the travis logs, you'll notice that it is actually not yet running the tests in `tests/`; you'll need to adapt the `pytests.ini`.

Yup, forgot to change this. I will also add an explicit path in `.github/workflows/tests.sh`

> Also, there are about a hundred tests left in `aiida`
> https://github.com/aiidateam/aiida-core/pull/3715/checks?check_run_id=392973225
> Do you want to move them in a separate PR?

No, I forgot about those and will add them here.

tests/tools/importexport/migration/test_migration.py
ltalirz: wouldn't it be possible to `from tests.path.to.module import get_archive`?

The tests should anyhow always be run from the top-level directory of the git repo

sphuber: Done

sphuber:
> Looks fine to me!
> I suggest you change the title since you move _all_ unit tests, not just the ones in `backends`.

Done

> All that remains is a minor update here: https://github.com/aiidateam/aiida-core/wiki/Writing-tests#running-existing-tests

I will see if I find sometime to adapt it. There is quite some stuff there that was already outdated even unrelated to this PR.

ltalirz:
I've adapted the wiki now to the changed test location.
We will do another update at some point for pytest suggestions.

One thing I noticed now is that you changed the folder structure for the transport tests, i.e. the plugin-specific tests are no longer in a subfolder `plugins`.
I guess the test discovery did not rely on this (?)

PR #3719, opened 2020-01-20
Add support for python 3.8

### LINKED ISSUE #3662: Python 3.8 compatibility issues
* [x] IPython 5.8.0, see #3559 
* [x] pylint 2.3.1, https://github.com/PyCQA/pylint/issues/2976 , seeing this on `functools` import

sphuber:
Fixes #3662 

Note, I have not checked the compatibility of all our (sub)dependencies manually, but the tests run.

### DIFF
--- a/aiida/tools/importexport/migration/v03_to_v04.py
+++ b/aiida/tools/importexport/migration/v03_to_v04.py
@@ -24,7 +24,7 @@ The individual SQLAlchemy database migrations may be found at:
 Where id is a SQLA id and migration-name is the name of the particular migration.
 """
 # pylint: disable=invalid-name
-
+import copy
 import os
 
 import numpy as np
@@ -312,7 +312,8 @@ def migration_calc_job_option_attribute_keys(data):
                 'jobresource_params': 'resources',
                 'parser': 'parser_name'
             }
-            for key in content:
+            # Need to loop over a clone because the `content` needs to be modified in place
+            for key in copy.deepcopy(content):
                 if key in key_mapper:
                     content[key_mapper[key]] = content.pop(key)
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.5, 3.7]
+        python-version: [3.5, 3.8]
         backend: ['django', 'sqlalchemy']
 
     steps:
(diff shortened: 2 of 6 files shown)


sphuber:
> thanks @sphuber , a smooth python version transition for once ?!

Cheered to early! Tests on my branch run, but on the PR they fail in one of the file system tests. There seems to be a problem with `shutil` handling utf-8 encoded strings. Not sure why yet

aiida/tools/importexport/migration/v03_to_v04.py
```diff
@@ -312,7 +312,8 @@ def _migration_calc_job_option_attribute_keys(attr_id, content):
                 'jobresource_params': 'resources',
                 'parser': 'parser_name'
             }
-            for key in content:
+            # Need to loop over a clone because the `content` needs to be modified in place
+            for key in copy.deepcopy(content):
```
dev-zero: As a general solution I usually prefer to create a copy for the new, modified data instead of backing up the current data to be able to provide better guarantees in case of failure: the dict will always be in either the original unmodified or the fully processed state.
With a dict this could be done by using `content.clear() ; content.update(newdata)`.

sphuber: > As a general solution I usually prefer to create a copy for the new, modified data instead of backing up the current data to be able to provide better guarantees in case of failure

I agree, except here that was not really possible because of the code outside of this function. This code is in a function defined in a local scope and the caller reuses the `content` variable after the function is invoked. The changes therefore have to be made in place to the reference and I cannot return a copy. I didn't feel like rewriting the rest of the code, so I did this for now. Your preferred approach is exactly what I used in the other case though, because there it was possible.

sphuber:
@dev-zero and @ltalirz thanks for the review. As you may have noticed, the tests on my branch ran, but on the pull request failed. Specifically it was a single test testing the `Folder` utility when handling utf-8 encoded file paths. I could not reproduce it locally either and so not figure out what was going wrong. I added a commit which gets rid of the problem. I think what was being tested was not the original intention of the author. The test now makes sense. Of course this is not wholly satisfactory because the original test _should_ be valid and run without exceptions, however, I have no clue where to start looking nor the time to do it. For now I think this is an OK solution then, but please give it a look first.

dev-zero:
@sphuber I think that's fine since the test is about unicode handling of path names rather than the logic of `Folder.insert_path` (and `shutil`s behaviour).

Issue #335, opened 2017-01-27
Implement the `RepositoryBackend` interface for Swift


sphuber:
With the new file repository implementation of `aiida-core==2.0.0` it is now possible to implement the `AbstractRepositoryBackend` interface for Swift since the repository no longer assumes a local file system solution,

Below is the original description of this issue before it was redefined to only consider implementing the repository interface for Swift.

~~Ideally, a profile can specify multiple file storage repositories and custom selection rules for which files to store in which repository. Which files and in which repository they are stored, will also be stored in the database from now on.~~

~~These changes will allow AiiDA to integrate with Swift, an object store, which will the choice of file storage system at CSCS~~

~~We should define a common API for all classes accessing files, common to directories and object store. Some methods will stay only for folders (e.g. `abspath`), others (e.g. `open` or `get_file_content`) should be common. The AiiDA code should only use the common API.~~

giovannipizzi:
See also #404

sphuber:
In converting `aiida_core` to be able to handle multiple repositories, I realized that the current repository is used for multiple things. It is not just used to store files from calculations, but it is also used as a temporary storage for things like sandbox folders and workflow pickles. It seems then, that even if we provide the option of an object store as a repository, we still may want to keep a regular filesystem repository for these temporary files. It seems a little strange that if a user configures an object store, that even these temporary files will all have to be stored in the OS. Should we start to make a distinction between the permanent repository, which can be filesystem or OS, and a temporary repository, that is what we have now and will be kept for temporary things, such as sandboxfolders and workflow related files.

giovannipizzi:
Yes, I agree the sandbox should be local, and ok to have a different configuration entry for its location/path.

giovannipizzi:
I think that before crystallising on an API and/or implementation, we should consider pyfilesystem2 [[GitHub]](https://github.com/PyFilesystem/pyfilesystem2), [[website]](https://www.pyfilesystem.org), [[blog post on why a version 2 has been written, with examples of the API]](https://www.willmcgugan.com/blog/tech/post/announcing-pyfilesystem-2/)

[...]

PR #3720, opened 2020-01-20
Add `provenance_exclude_list` attribute to `CalcInfo` data structure

### LINKED ISSUE #2956: Add support to write `CalcJob` input files that are not stored in the repository
This use case comes from the `aiida-vasp` plugin. The VASP code expects a single file named `POTCAR` with the contents of the required pseudo potentials concatenated for the various elements. The `VaspCalculation` plugin now simply creates this file in the `folder` sandbox, whose contents the engine will then copy to the working directory. However, these contents are also stored in the repository of the `CalcJobNode`. This leads to the content of the `PotcarData` nodes, which contain the pseudo potentials per element, to be copied over and over again. This is exactly why the `local_copy_list` mechanism was introduced, but since the target file is not a simple copy, but a concatenation of multiple input files, this cannot be used. The situation is made worse, because the content of the `PotcarData` nodes are proprietary and cannot be shared. The plugin already provides a mechanism to separate the actual content of the pseudo from the node representation, but all of this work is negated if their content is copied wholesale in the repository of the calculation job nodes that use them.

This is then the real problem here: neither the `local_copy_list` nor the `folder` can be used to have these files transferred to the working directory.

I see two potential solutions:
1. A richer syntax for `local_copy_list` to also instruct simple operations in addition to a literal copy of single files: e.g. concatenating of multiple files to a single
2. A second sandbox folder for files that need to be copied but *should not* be copied to the repository of the `CalcJobNode`.
3. A syntax of marking certain files or paths (with glob) in the `folder` that should not be stored.

Both of these further complicate the already rather complex `CalcJob.prepare_for_submission` interface, but it might be necessary.

**Edit:** added a third option proposed by @giovannipizzi

sphuber:
Fixes #2956 

This new attribute takes a flat list of relative filepaths, which
correspond to files in the `folder` sandbox passed to the
`prepare_for_submission` call of the `CalcJob`, that should not be
copied to the repository of the `CalcJobNode`. This functionality is
useful to avoid the content of input files, that should be copied to the
working directory of the calculation, to also be stored permanently in
the file repository. Example use cases are for very large input files or
files whose content is proprietary. Both use cases could already be
implemented using the `local_copy_list` but only in the case of files of
an input node in its entirety. The syntax of the `local_copy_list` does
not support the exclusion of arbitrary files that are written by the
calculation plugin to the sandbox folder.

Question: currently in `aiida.engine.processes.calcjobs.daemon.upload_calculation` when I delete the files in `provenance_exclude_list` from the sandbox `folder` I do not catch any errors. Do we maybe want to ignore files that do not exist?

Note: once this is tested and approved, I will add the documentation. First wanted to agree on the nomenclature and functionality so that I do not have to adapt the text every time if we need changes.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -62,9 +62,6 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
         raise ValueError('Cannot submit calculation {} because it has cached input links! If you just want to test the '
                          'submission, set `metadata.dry_run` to True in the inputs.'.format(node.pk))
 
-    # After this call, no modifications to the folder should be done
-    node.put_object_from_tree(folder.abspath, force=True)
-
     # If we are performing a dry-run, the working directory should actually be a local folder that should already exist
     if dry_run:
         workdir = transport.getcwd()
@@ -243,6 +240,23 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
             else:
                 raise IOError('It is not possible to create a symlink between two different machines for '
                               'calculation {}'.format(node.pk))
+
+    provenance_exclude_list = calc_info.provenance_exclude_list or []
+
+    # Loop recursively over content of the sandbox folder copying all that are not in `provenance_exclude_list`. Note
+    # that directories are not created explicitly. The `node.put_object_from_filelike` call will create intermediate
+    # directories for nested files automatically when needed. This means though that empty folders in the sandbox or
+    # folders that would be empty when considering the `provenance_exclude_list` will *not* be copied to the repo. The
+    # advantage of this explicit copying instead of deleting the files from `provenance_exclude_list` from the sandbox
+    # first before moving the entire remaining content to the node's repository, is that in this way we are guaranteed
+    # not to accidentally move files to the repository that should not go there at all cost.
+    for root, dirnames, filenames in os.walk(folder.abspath):
+        for filename in filenames:
+            filepath = os.path.join(root, filename)
+            relpath = os.path.relpath(filepath, folder.abspath)
+            if relpath not in provenance_exclude_list:
+                with open(filepath, 'rb') as handle:
+                    node.put_object_from_filelike(handle, relpath, 'wb', force=True)
 
     if not dry_run:
         # Make sure that attaching the `remote_folder` with a link is the last thing we do. This gives the biggest
--- a/aiida/common/datastructures.py
+++ b/aiida/common/datastructures.py
@@ -67,6 +67,12 @@ class CalcInfo(DefaultFieldsAttributeDict):
     * local_copy_list: a list of tuples with format ('node_uuid', 'filename', relativedestpath')
     * remote_copy_list: a list of tuples with format ('remotemachinename', 'remoteabspath', 'relativedestpath')
     * remote_symlink_list: a list of tuples with format ('remotemachinename', 'remoteabspath', 'relativedestpath')
+    * provenance_exclude_list: a sequence of relative paths of files in the sandbox folder of a `CalcJob` instance that
+        should not be stored permanantly in the repository folder of the corresponding `CalcJobNode` that will be
+        created, but should only be copied to the remote working directory on the target computer. This is useful for
+        input files that should be copied to the working directory but should not be copied as well to the repository
+        either, for example, because they contain proprietary information or because they are big and their content is
+        already indirectly present in the repository through one of the data nodes passed as input to the calculation.
     * codes_info: a list of dictionaries used to pass the info of the execution of a code
     * codes_run_mode: a string used to specify the order in which multi codes can be executed
     """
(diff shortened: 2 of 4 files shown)


espenfl:
@sphuber What happens if the delete operation fails?

sphuber:
> @sphuber What happens if the delete operation fails?

The logic is now this:

 1. Take the `folder` sandbox and copy contents to remote work dir
 2. Remove the `provenance_exclude_list` files from the sandbox
 3. Move remaining contents of `folder` to the repository of the `CalcJobNode`

This whole `upload_calculation` is wrapped in the exponential backoff mechanism as part of the `UPLOAD` transport task. So if the deletion of the file from the sandbox fails, it will raise an exception and the task will be retried later on. If the failure was due to a programming error by the `CalcJob` implementation, for example the file in `provenance_exclude_list` does not exist causing the delete to fail, it will try the 5 times and then pause the process. The user will then notice the programming mistake and have to kill the paused process

espenfl:
Thanks. It would be important to be sure the file is not there. If the delete operation fails (without Python catching an exception) we could possibly end up including the file anyway. One possibility is to control the copy/move side of the operation, such that if that fails, there is no way the file is present in the repository. Maybe we could introduce a second sandbox folder?

sphuber:
@giovannipizzi and @espenfl : I have added a commit that changes the method from moving the sandbox to the repository, to manually recursively copying files that are not excluded. As noted in the commit message, this is less performant _but_ once the new repository interface is there and we can no longer assume a file system solution, the move method will no longer be possible anyway. The only fundamental difference still that might be problematic is that with the current copy implementation, empty folders (or folders that would have been empty when taking the excludes into account) will _not_ be present in the repository. Is this a problem?

I know that certain solutions depend on empty folders being in the sandbox as they are uploaded to the working directory (such as the `pseudo` directory for `PwCalculation`) and that still works, except they won't be present in the repository. For efficiency reasons its obviously better to not store empty directories, but is there a situation possible where they carry important meaning for the provenance?

espenfl:
@sphuber That is formally a more safer approach so I fully support this. Thanks for submitting the update. One can use a copy operation that uses `reflinks` which basically gives you the same performance as a move operation. This works on most systems. Python support for this is another question. There is a `reflink` library as I mentioned in #2956 here: https://gitlab.com/rubdos/pyreflink. If that is something we want to include, I do not know. Also, at some point we will leave file operations altogether. The future of this also connects with possibly considering a new backend for handling repository operations as I mentioned in https://github.com/aiidateam/aiida-core/issues/335#issuecomment-573302203. So my initial view of the performance at this point is that it is less critical since we know we will at some point in the not too distant future abandon such operations. It is much better to be safe than sorry in my view when it comes to these things at this point.

sphuber:
I am not too worried about the performance hit, since anyway in the near future we will have to implement that change. So I take it you are happy with this implementation then? If @giovannipizzi is also ok, I will add the necessary documentation and squash the last commit in and then it should be ready to go in.

aiida/engine/daemon/execmanager.py
```diff
@@ -241,16 +241,22 @@ def find_data_node(inputs, uuid):
...
+    # directories for nested files automatically when needed. This means though that empty folders in the sandbox or
+    # folders that would be empty when considering the `provenance_exclude_list` will *not* be copied to the repo. The
+    # advantage of this explicit copying instead of deleting the files from `provenance_exclude_list` from the sandbox
+    # first before moving the entire remaining content to the node's repository, is that in this way we are guaranteed
+    # not to accidentally move files to the repository that should not go there at all cost.
+    for root, dirnames, filenames in os.walk(folder.abspath):
+        for filename in filenames:
+            filepath = os.path.join(root, filename)
+            relpath = os.path.relpath(filepath, folder.abspath)
+            if relpath not in provenance_exclude_list:
```
espenfl: Is there a way that `relpath` and `provenance_exclude_list` can refer to the same file, but not contain the same string? Except of course if the plugin supply something that is not correct for `provenance_exclude_list`?

sphuber: If the `provenance_exclude_list` is defined correctly, then no, this should work. The `provenance_exclude_list` should contain file paths relative to the base path of the `folder` sandbox. Since the files therein are created also with relative filepaths in the `prepare_for_submission` where the `provenance_exclude_list` is created, there is a one-to-one correspondence. So I think the risk for mistakes should be minimal.
Example would be:
```
def prepare_for_submission(self, folder):
    provenance_exclude_list = []
    for element, pseudo in self.inputs.pseudos.items():
        with pseudo.open(mode='rb') as source:
            filename = os.path.join('pseudos', element)
            folder.create_file_from_filelike(source, filename)
            provenance_exclude_list.append(filename)
```
the `provenance_exclude_list` will now for example contain:
```
['pseudos/Si', 'pseudos/Ge']
```
which will now not be stored in the repository. For your use case it will be even easier because you just specify `provenance_exclude_list = ['POSCAR']`` and you're done.

espenfl: Thanks for commenting clearly on this.

sphuber:
Since I think we are happy with naming/interface/implementation I wrote the necessary documentation and so this is now ready for final review @giovannipizzi

PR #3722, opened 2020-01-22
Build docker image based on aiida-prerequisites

### LINKED ISSUE #3725: Docker container: make user credentials configurable
For the AiiDA lab, ideally the initial AiiDA profile should be generated using user information from OAUTH2 (or however the user authenticated).

Could you please adapt the setup script such that it will check the environment variables
`AIIDA_USER_EMAIL`, `AIIDA_USER_NAME`, and `AIIDA_USER_INSTITUTION`?
If these variables are not empty, their value should take precedence over the default values.

yakutovicha:
fixes #3725


ltalirz:
For the record, this is an implementation of the [discussion on the plugin testing infrastructure](https://docs.google.com/document/d/1tOTcj-66E1ipW56ULEtzaAOJR7xhjZEGlDDyH8p-NBk/edit) from the coding week  (document now frozen).

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,65 @@
...
+        --first-name Some                      \
+        --last-name Body                       \
+        --institution XYZ                      \
+        --db-backend $aiida_backend            \
+        --db-username aiida                    \
+        --db-password aiida_db_passwd          \
+        --db-name aiidadb                      \
+        --db-host localhost                    \
+        --db-port 5432                         \
+        --repository /home/$SYSTEM_USER/aiida_repository
```
ltalirz: That's a bit of a political question - should there be a pre-configured profile by default?

The upside is that people can directly start running AiiDA commands,
the downside is they won't bother to set up their own profile and thus not learn how to do it, and when they start exporting data they may discover that the User name/email/... is not what they want.
Also, it is not needed for running tests.

I would lean towards not setting it up by default but I can see arguments for both use cases. 
I would suggest to:

 * Use an `ENV SETUP_DEFAULT_PROFILE 0` in the Dockerfile and check the value of this variable in the `configure-aiida.sh` script. One can then start a container with `docker run ... -e SETUP_DEFAULT_PROFILE=1` and it will create the default profile.

 * for the profile setup, (re)use a [config yaml file](https://github.com/aiidateam/aiida-core/blob/develop/.github/config/profile.yaml) with the `--config` option. if this yaml is present on the container, it means one can set up the profile in one line, even if it's not yet there

 * please check that `verdi quicksetup` works out of the box.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM aiidateam/aiida-prerequisites:latest
+
+USER root
+
+# Copy and install AiiDA
+COPY . aiida-core
```
ltalirz: isn't this the other way around?
```suggestion
COPY aiida-core .
```

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,65 @@
...
+
+#===============================================================================
+# environment
+reentry scan
+
+export SHELL=/bin/bash
+
+#===============================================================================
+# setup AiiDA
+aiida_backend=django
```
sphuber: Should this be configurable?

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,65 @@
...
+        --profile default                      \
+        --non-interactive                      \
+        --email some.body@xyz.com              \
+        --first-name Some                      \
+        --last-name Body                       \
+        --institution XYZ                      \
+        --db-backend $aiida_backend            \
+        --db-username aiida                    \
+        --db-password aiida_db_passwd          \
+        --db-name aiidadb                      \
```
sphuber: I take it these values correspond to what is hard-coded in `aiida-prerequisites` where a postgres DB is already installed. Seems a bit delicate, no? Would it be better to have database creation in same docker layer?

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,65 @@
...
+
+   echo 'eval "$(verdi completioncommand)"' >> /home/$SYSTEM_USER/.bashrc
+
+   verdi profile setdefault default
+fi
+
+#===============================================================================
+# perform the database migration if needed
+#
+verdi daemon start || ( verdi daemon stop && verdi database migrate --force && verdi daemon start)
```
sphuber: I think that the failing of `verdi daemon start` is not the best way of determining whether one needs to migrate the database. Maybe it is best to just run `verdi database migrate -f` before it, since if the database is up to date it is a no-op. If we really wanted, we could make a `verdi database migration-status` command that will fail if it needs to be updated and if it is up to date will return 0.

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,65 @@
...
+        --institution XYZ                      \
+        --db-backend $aiida_backend            \
+        --db-username aiida                    \
+        --db-password aiida_db_passwd          \
+        --db-name aiidadb                      \
+        --db-host localhost                    \
+        --db-port 5432                         \
+        --repository /home/$SYSTEM_USER/aiida_repository
+
+   echo 'eval "$(verdi completioncommand)"' >> /home/$SYSTEM_USER/.bashrc
```
sphuber: Maybe better to replace this with:
```
echo 'eval "$(_VERDI_COMPLETE=source verdi)"' >> /home/$SYSTEM_USER/.bashrc
```
since at some point we wanted to get rid of `verdi completioncommand`

.docker/opt/configure-aiida.sh
ltalirz: I agree - this would also also let us switch to `verdi quicksetup` here, so we'd be testing that it actually works.

Dockerfile
yakutovicha: no, I am copying the current folder `.` to the destination folder `aiida-core`

.docker/opt/configure-aiida.sh
yakutovicha: ok

Dockerfile
ltalirz: Right. Do we want the full git repo in the docker image?
The .git folder is about 100MB... but maybe you are right and we should have it. Hm.

.docker/opt/configure-aiida.sh
yakutovicha: I don't really need `verdi database migration-status` I think just run `verdi database migrate -f` is fine for my use case.

However, it would be nice to have possibility to check for an existing profile. It can also be integrated into `verdi profile list`. Something like `verdi profile list <profile name>`. Currently it doesn't work

yakutovicha: Currently I do `verdi profile show <profile name> &> /dev/null`

Dockerfile
yakutovicha: To me, for the moment it doesn't harm. If we discover problems, then look at this issue in more details.

.docker/opt/configure-aiida.sh
yakutovicha: ok

.docker/opt/configure-aiida.sh
ltalirz: How about `verdi profile show <profile name> || ...`
(or `if verdi profile show <sdsda> then ...`)?

.docker/opt/configure-aiida.sh
yakutovicha: Made all options configurable. It will also be useful for AiiDA lab when setting up a new user.

.docker/opt/configure-aiida.sh
yakutovicha: > Use an `ENV SETUP_DEFAULT_PROFILE 0`
agreed

> or the profile setup
For the moment I will still the current way, but make everything configurable through `ENV`. That way we can preconfigure things for AiiDA lab.

.docker/opt/configure-aiida.sh
yakutovicha: Made this configuration optional.

.docker/opt/configure-aiida.sh
yakutovicha: > please check that `verdi quicksetup` works out of the box.

it works.

yakutovicha:
I will now work on adding things to github workflows

yakutovicha:
Btw, I think it totally makes sense to move the whole profile setup to quicksetup. What do you guys think?

yakutovicha:
@sphuber can you help me in setting up Git Hub actions? I am not sure how to do things right. [The manual](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-a-docker-container-action) suggests making a new action file. Is it the way to go?

sphuber:
> @sphuber can you help me in setting up Git Hub actions? I am not sure how to do things right. [The manual](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-a-docker-container-action) suggests making a new action file. Is it the way to go?

What is it that needs to be run and when? Every commit? Take a look at the `.github/workflows/ci.yml`. There you will find the main job defined with the various steps that are run on each commit and PR. Depending on what you need to run, you can add one there.

ltalirz:
> What is it that needs to be run and when? Every commit?

The build of the docker image should be run on every commit

yakutovicha:
> What is it that needs to be run and when? Every commit?

It should be docker build, yes, it should be run every commit.

> Take a look at the `.github/workflows/ci.yml`. There you will find the main job defined with the various steps that are run on each commit and PR.

I did it, and the check that I've added have failed. https://github.com/aiidateam/aiida-core/pull/3722/checks?check_run_id=405089491

sphuber:
> I did it, and the check that I've added have failed. https://github.com/aiidateam/aiida-core/pull/3722/checks?check_run_id=405089491
```
##[error]Process completed with exit code 100.
Run apt-get install -y --no-install-recommends docker.io
E: Could not open lock file /var/lib/dpkg/lock-frontend - open (13: Permission denied)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), are you root?
```
You are not installing with `sudo`, so try to prepend your install statements with `sudo`. Also I noticed that in the actual docker build step you do not specify the location of the `Dockerfile`. Is it going to find it automatically?

yakutovicha:
btw, another option would be to add Docker hub status check. That requires no setup here. What do you guys think?

yakutovicha:
> You are not installing with sudo, so try to prepend your install statements with sudo. 

Let's try. I wasn't sure if it is allowed to use sudo here.

> Also I noticed that in the actual docker build step you do not specify the location of the Dockerfile. Is it going to find it automatically?

By default, it takes it from the current folder. I hope the current folder here is the package's root.

ltalirz:
> btw, another option would be to add Docker hub status check. That requires no setup here. What do you guys think?

I think that is a good idea. We should *not* require it to pass, since I believe this can take some time. But it's good to take burden away from github actions, and the build on dockerhub will be needed anyhow.

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+# This script is executed whenever the docker container is (re)started.
+
+# Debugging.
+#set -x
```
ltalirz: I've often found it useful to have this on.
Perhaps let's start with `set -x` and if in a few months everything is well converged, we can deactivate it again.

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,54 @@
...
+    verdi quicksetup                           \
+        --non-interactive                      \
+        --profile $PROFILE_NAME                \
+        --email $USER_EMAIL                    \
+        --first-name $USER_FIRST_NAME          \
+        --last-name $USER_LAST_NAME            \
+        --institution $USER_INSTITUTION        \
+        --db-backend $AIIDADB_BACKEND
+
+    # Ssetup and configure local computer.
```
ltalirz:
```suggestion
    # Setup and configure local computer.
```

.docker/opt/configure-aiida.sh
ltalirz:
```suggestion
set -x
```

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,54 @@
...
+grep _VERDI_COMPLETE .bash_profile &> /dev/null || echo 'eval "$(_VERDI_COMPLETE=source verdi)"' >> /home/$SYSTEM_USER/.bashrc
+
+# Check if profile $PROFILE_NAME needs to be setup an if it exists already
+if [[ $SETUP_DEFAULT_PROFILE == true ]] && ! verdi profile show $PROFILE_NAME &> /dev/null; then
+    NEED_SETUP_PROFILE=true;
+else
+    NEED_SETUP_PROFILE=false;
+fi
+
+# Setup AiiDA.
```
ltalirz:
```suggestion
# Setup AiiDA profile if needed.
```

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,54 @@
...
+#set -x
+
+# Environment.
+reentry scan
+export SHELL=/bin/bash
+
+# Setup AiiDA autocompletion.
+grep _VERDI_COMPLETE .bash_profile &> /dev/null || echo 'eval "$(_VERDI_COMPLETE=source verdi)"' >> /home/$SYSTEM_USER/.bashrc
+
+# Check if profile $PROFILE_NAME needs to be setup an if it exists already
```
ltalirz:
```suggestion
# Check if user requested to set up AiiDA profile (and if it exists already)
```

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,54 @@
...
+if [[ $SETUP_DEFAULT_PROFILE == true ]] && ! verdi profile show $PROFILE_NAME &> /dev/null; then
+    NEED_SETUP_PROFILE=true;
+else
+    NEED_SETUP_PROFILE=false;
+fi
+
+# Setup AiiDA.
+if [[ $NEED_SETUP_PROFILE == true ]]; then
+
+    # Create AiiDA profile if required.
```
ltalirz:
```suggestion
    # Create AiiDA profile.
```

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,54 @@
...
+        --description "this computer"                          \
+        --hostname ${computer_name}                            \
+        --transport local                                      \
+        --scheduler direct                                     \
+        --work-dir /home/aiida/aiida_run/                      \
+        --mpirun-command "mpirun -np {tot_num_mpiprocs}"       \
+        --mpiprocs-per-machine 1 &&                            \
+    verdi computer configure local ${computer_name}            \
+        --non-interactive                                      \
+    --safe-interval 0.0
```
ltalirz:
```suggestion
      --safe-interval 0.0
```

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+# Migration is run if at least one profile exists.
+# Daemon is run if the database was successfully migrated.
+( verdi profile show && verdi database migrate --force && verdi daemon start ) || echo "Daemon is not runing."
```
ltalirz:
```suggestion
( verdi profile show && verdi database migrate --force && verdi daemon start ) || echo "AiiDA daemon is not running."
```

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM aiidateam/aiida-prerequisites:latest
+
+USER root
+
+ENV SETUP_DEFAULT_PROFILE true
+
+ENV PROFILE_NAME default
+ENV USER_EMAIL some.body@xyz.com
```
ltalirz:
```suggestion
ENV USER_EMAIL aiida@localhost
```

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+ENV USER_FIRST_NAME Some
```
ltalirz:
```suggestion
ENV USER_FIRST_NAME Giuseppe
```

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+ENV USER_LAST_NAME Body
```
ltalirz:
```suggestion
ENV USER_LAST_NAME Verdi
```

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+ENV USER_INSTITUTION XYZ
```
ltalirz:
```suggestion
ENV USER_INSTITUTION Khedivial
```

yakutovicha:
I think it's ready. @ltalirz or @sphuber please review.

yakutovicha:
> thanks for the updates @yakutovicha !
> 
> I notice two issues with the actions build:
> 
>     1. It seems like [there is no profile configured](https://github.com/aiidateam/aiida-core/pull/3722/checks?check_run_id=405794933#step:5:6), although by your default there should be
> 
>     2. The docker build did not fail despite no profile being configured. I guess you should replace the `verdi profile list` by a `verdy profile show default `

I think I know why let me try to fix that.

yakutovicha:
Should be fixed with that: https://github.com/aiidateam/aiida-prerequisites/pull/6. Let's see...

.github/workflows/ci.yml
```diff
@@ -159,3 +159,30 @@ jobs:
...
+    - name: Build the aiida-core image and run it
+      run: |
+        docker build -t aiida-core .
+        docker run -d aiida-core
+
+    - name: Check 'verdi profile list' command
+      run: |
+        export DOCKERID=`docker ps -qf 'ancestor=aiida-core'`
+        docker exec --tty $DOCKERID wait-for-services
+        docker exec --tty --user aiida $DOCKERID /bin/bash -l -c 'verdy profile show default'
```
ltalirz:
```suggestion
        docker exec --tty --user aiida $DOCKERID /bin/bash -l -c 'verdi profile show default'
```

.github/workflows/ci.yml
```diff
@@ -159,3 +159,30 @@ jobs:
...
+        sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu  $(lsb_release -cs)  stable"
+        sudo apt-get update
+        sudo apt-get install docker-ce
+
+    - name: Build the aiida-core image and run it
+      run: |
+        docker build -t aiida-core .
+        docker run -d aiida-core
+
+    - name: Check 'verdi profile list' command
```
ltalirz:
```suggestion
    - name: Check that default AiiDA profile was set up
```

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,54 @@
...
+    NEED_SETUP_PROFILE=false;
+fi
+
+# Setup AiiDA profile if needed.
+if [[ $NEED_SETUP_PROFILE == true ]]; then
+
+    # Create AiiDA profile.
+    verdi quicksetup                           \
+        --non-interactive                      \
+        --profile $PROFILE_NAME                \
```
sphuber: All these variables need to be wrapped in double quotes in case they contain spaces. Best is to use `"${PROFILE_NAME}"`

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,54 @@
...
+        --description "this computer"                          \
+        --hostname ${computer_name}                            \
+        --transport local                                      \
+        --scheduler direct                                     \
+        --work-dir /home/aiida/aiida_run/                      \
+        --mpirun-command "mpirun -np {tot_num_mpiprocs}"       \
+        --mpiprocs-per-machine 1 &&                            \
+    verdi computer configure local ${computer_name}            \
+        --non-interactive                                      \
+      --safe-interval 0.0
```
sphuber: fix indentation

.docker/opt/configure-aiida.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+# Migration is run if at least one profile exists.
+# Daemon is run if the database is either up to date or it has been successfully migrated.
+( verdi profile show && verdi database migrate --force && verdi daemon start ) || echo "AiiDA daemon is not running."
```
sphuber: I am not sure if it is a good idea to chain all of these commands. If the first or second fails, it will still show that the daemon is not running. While technically correct, it is a bit misleading. Why not separate the three each which its own error message?

sphuber:
Oh and one final question: why is the `Dockerfile` on the top level? Since there is a `.docker` folder, doesn't it make more sense to put it in there? Or is that technically not possible or against convention?

ltalirz:
> Oh and one final question: why is the Dockerfile on the top level? Since there is a .docker folder, doesn't it make more sense to put it in there? Or is that technically not possible or against convention?

It is sort of a convention in that this is the place where many external tools/services will look for the `Dockerfile` by default (mybinder, heroku, Dockerhub, etc.).
It is possible to move it to the `.docker` folder but I think it's not a bad idea to have this one at the top level.

.docker/opt/configure-aiida.sh
yakutovicha: done

.docker/opt/configure-aiida.sh
yakutovicha: done

.docker/opt/configure-aiida.sh
yakutovicha: I think you are right. Migration won't run anyways if the default profile is not provided. Same true for the `daemon start`- it will work only if (a) default profile exists and (b) it does not need migration.

yakutovicha: done

yakutovicha:
@sphuber if you want to merge the PR, please let me know 5 mins in advance: I will set up automated builds on Docker Hub.

sphuber:
OK this is good to go, let me know when you have setup Docker Hub and I will merge

yakutovicha:
> OK this is good to go, let me know when you have setup Docker Hub and I will merge

ready, please merge. Docker build should now be triggered automatically at Docher Hub: https://hub.docker.com/repository/docker/aiidateam/aiida-core

sphuber:
Thanks @yakutovicha

PR #3729, opened 2020-01-24
add non-zero exit code for verdi daemon status

### LINKED ISSUE #3728: `verdi daemon status` should return non-zero exit code, if daemon is not running
currently, it always returns exit code 0.

ltalirz:
fixes #3728

`verdi daemon status` was always returning exit code 0, which makes it
difficult to use the command programmatically (e.g. in ansible).
It now returns exit code 1 if the daemon of any of the requested
profiles is not running.

### DIFF
--- a/aiida/cmdline/commands/cmd_daemon.py
+++ b/aiida/cmdline/commands/cmd_daemon.py
@@ -98,7 +98,7 @@ def status(all_profiles):
         daemons_running.append(client.is_daemon_running)
 
     if not all(daemons_running):
-        sys.exit(1)
+        sys.exit(3)
 
 
 @verdi_daemon.command()
--- a/aiida/cmdline/commands/cmd_daemon.py
+++ b/aiida/cmdline/commands/cmd_daemon.py
@@ -12,6 +12,7 @@
 import os
 import subprocess
 import time
+import sys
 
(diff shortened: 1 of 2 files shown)


ltalirz:
Thanks for paying attention - I actually planned doing what I did now, no idea what I was thinking...
Also added tests now.

aiida/cmdline/commands/cmd_daemon.py
```diff
@@ -76,7 +77,10 @@ def start(foreground, number):
 @verdi_daemon.command()
 @click.option('--all', 'all_profiles', is_flag=True, help='Show all daemons.')
 def status(all_profiles):
-    """Print the status of the current daemon or all daemons."""
+    """Print the status of the current daemon or all daemons.
+
+    Returns exit code 0 if all daemons are running, else exit code 3.
```
sphuber: Could you please clarify that "all daemons" refers to "all daemons whose status is requested" and not "daemons of all profiles", i.e., if one does not specify `--all` then 0 is returned when the current default daemon/profile is running, even if any other daemon is not running

ltalirz: sure, done!

Issue #3406, opened 2019-10-08
`Computer.default_mpiprocs_per_machine` weirdly named


sphuber:
The `default_mpiprocs_per_machine` property of the `Computer` seems to suggest that the value means the number of MPI processes to launch per machine (node). This is _not_ however what the property docstrings say; namely it is the number of processors (CPUs) per machine. The latter makes much more sense as a property of a computer, i.e. a machine. This is further corroborated by the fact that this property is set during computer setup and the label states: "Default number of cpus per machine". @giovannipizzi did the two get mixed up at some point? Is the original intended definition of this property indeed the "number of CPUs (or cores) per machine (node)"?

giovannipizzi:
This goes back in history, probably 2014 or so.
I had changed this to mpiprocs because this is the actual behaviour (what to do in `mpirun -n XXX`). I see your reasoning, though (the property that makes more sense is the number of cores per node). However, in the end this is used to choose the default number of mpiprocs per machine=node, so I would see the property name (and possibly change the docs? Even if I think it's ok to ask for the number of CPUs (in the end it's more intuitive) and then use it as a default # of MPI process

sphuber:
After some more diving it seems like the property name is correct and actually means the "default number of MPI processes to use per node", but the docstring and the label in the `verdi computer show` output is incorrect, which refer to number of CPUs per node.

What we often would need for efficient parallelization in addition to this information (which can be overridden with `number_mpiproces_per_machine` on a per-calculation basis) is to know the number of CPUs per machine, as in just its architecture. The reason is that, based on the number of MPI processes requested, we need to compute the number of CPUs per MPI proc and set it explicitly. If not, schedulers like SLURM will set the default, which is 1, running the risk of having many CPUs idling on the node. Would it be advisable to add a property to `Computer` detailing the number of CPUs each node has? Would this give problems with machines that have inhomogeneous node layouts/architectures? Maybe we can make it optional?

ltalirz:
Corresponding adding the number of CPUs per node, I think this makes sense and the discussion on this should continue in #3503 .

Staying with MPI, however, there is also the question of whether MPI (`mpirun` or similar) is available on a computer at all. I don't think this can currently be configured at the computer level, but I think it should.
This affects e.g. the AiiDA docker container, see https://github.com/aiidateam/aiida-core/pull/3732#issuecomment-579293358

ltalirz:
As of AiiDA 2.2, this is the status quo:

[...]

PR #3732, opened 2020-01-28
Fix local computer setup in the docker container

yakutovicha:
Also add test for it to GitHub actions.

### DIFF
--- a/.docker/opt/configure-aiida.sh
+++ b/.docker/opt/configure-aiida.sh
@@ -37,7 +37,7 @@ if [[ $NEED_SETUP_PROFILE == true ]]; then
     verdi computer show $computer_name || verdi computer setup \
         --non-interactive                                      \
         --label "${computer_name}"                             \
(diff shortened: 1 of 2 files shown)


ltalirz:
@yakutovicha By the way, is mpi installed?
Not sure this makes sense but I just saw mpirun mentioned in the output of `verdi computer show`
https://github.com/aiidateam/aiida-core/pull/3732/checks?check_run_id=412968790#step:5:30

yakutovicha:
Hmm, good catch. No, it is not installed.

I can add it to `aiida-prerequisites` image. Not sure if it is needed, though. If one installs a code that requires MPI - it will be installed automatically. If one runs the code that does not need MPI - one can just say  `'withmpi': False`.

what do you think?

ltalirz:
Let's not do it for the moment. I think the right way to resolve this is to be able to configure at the computer level whether MPI is available (see also #3406).

Issue #3478, opened 2019-10-29
Have a WorkChain step get the transport


giovannipizzi:
There are a few cases where it would be useful to perform operations on a remote computer during a work chain step.

Some examples:
- inspect the list of files on the output folder of a CalcJob, to decide what to do next depending on the existence of a given file (see also #3477)
- clean the workdir of calculations that you know you will not need anymore for restarts

Currently, to do this one has to open the transport "by hand", skipping all the logic for exponential backoff, connection pooling etc.
It would be good to have a way instead to get a transport "cleanly".

A few options discussed with @sphuber:
1. allow for something like this in a step:
   ```python
   with get_transport(computer, user) as request:
       tranport = yield request
       # You get back control when the transport is ready
       transport.ls()
       #... other tasks
   ```
   This probably gives max flexibility, with the risk that users can do weird stuff on the remote or close the transport (they shouldn't, it could be reused by other tasks due to the pooling mechanism).
   Also, it should be clear that the in the step the first thing to do is to get the transport, and not do complex things before, because if the daemon is killed during the yield, the whole step needs to be rerun. And we need to explain what the yield syntax means.

2. define a way to get a transport in a step (e.g. via a decorator). The value could be passed as an additional parameter to a decorated step. This clarifies that the step needs a transport, allows to put multiple decorators if more than one transport is needed at the same time, but might be a bit more limiting if the transport you want to get needs to be defined programmatically (e.g. which computer, which user, ...). There might be workarounds, like 
   ```python
   @with_transport(computer='comp1')
   @with_transport(computer='comp2')
   def my_step(self, ctx, transport1, transport2):
       pass
   ```
   where the top decorator returns `transport1`, the second `transport2`, and the computer is chosen not as `"comp1"`/`"comp2"` but as `ctx.comp1`/`ctx.comp2`

[...]

Issue #1857, opened 2018-08-08
Unprotected opening of transport in RemoteData._clean


sphuber:
The `RemoteData._clean` method directly calls `transport.open`, without going through a transport queue nor without proper error checking. The latter means that if the transport fails to open, the command will except and the former means that the limiting of opening transport connections through the engine's `TransportQueue` is circumvented.

giovannipizzi:
I also add that the utility functions `getfile()` and `listdir()` (that I added some time ago), while useful from a command line to inspect the remote content, also open a transport.
We should probably discuss if we want to keep them, and if so how to make it clear that they open (and immediately close) a transport, making users aware of this (and that, e.g., you shouldn't use them in a workflow or in a plugin).

greschd:
Besides just documenting it, we could implement a "please don't use the transport" context manager that sets a flag on the transport, with the effect of giving a warning if the transport is used.

Then, we could wrap the `prepare_for_submission` (and possibly other places) in this context manager.

giovannipizzi:
Interesting idea.
We could also combine this the other way around where these functions, if called from within an open transport "context", can just 'pick' the currently open transport? (Not sure what to do when there is more than one, though)

PR #3742, opened 2020-02-04
Fix behavior of RemoteData.getfile()

Crivella:
The :return: value in the documentation was not correct.

### DIFF
--- a/aiida/orm/nodes/data/remote.py
+++ b/aiida/orm/nodes/data/remote.py
@@ -57,9 +57,9 @@ class RemoteData(Data):
         """
         Connects to the remote folder and gets a string with the (full) content of the file.
 
-        :param relpath: The relative path of the file to show.
-        :param destpath: A path on the local computer to get the file
-        :return: a string with the file content
+        :param relpath:  The relative path of the file on the remote to retrieve.
+        :param destpath: The absolute path of the copied/retrieved file on the local machine.
+        :return: a list containing the file names in 'destpath'
         """
         authinfo = self.get_authinfo()
         t = authinfo.get_transport()
--- a/aiida/orm/nodes/data/remote.py
+++ b/aiida/orm/nodes/data/remote.py
@@ -59,7 +59,7 @@ class RemoteData(Data):
 
         :param relpath:  The relative path of the file on the remote to retrieve.
         :param destpath: The absolute path of the copied/retrieved file on the local machine.
-        :return: a list containing the file names in 'destpath'
+        :return: a string with the file content
         """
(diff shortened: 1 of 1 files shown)


aiida/orm/nodes/data/remote.py
```diff
@@ -57,9 +57,9 @@ def getfile(self, relpath, destpath):
         """
         Connects to the remote folder and gets a string with the (full) content of the file.
```
ltalirz:
```suggestion
        Connects to the remote folder and retrieves the content of a file.
```

aiida/orm/nodes/data/remote.py
```diff
@@ -57,9 +57,9 @@ def getfile(self, relpath, destpath):
...
 
-        :param relpath: The relative path of the file to show.
-        :param destpath: A path on the local computer to get the file
-        :return: a string with the file content
+        :param relpath:  The relative path of the file on the remote to retrieve.
+        :param destpath: The absolute path of the copied/retrieved file on the local machine.
+        :return: a list containing the file names in 'destpath'
```
ltalirz: Indeed, this seems to be what it is doing... 
@giovannipizzi Is there a good reason for doing that (e.g. potentially returning the names of files that have nothing to do with the one that was just transferred?)

ltalirz: Just coming back to this - this is actually a bug, right @sphuber ?

This

Crivella:
Actually as of know this method would have the double behaviour of copying the file to the local machine and returning its content as a string.

I'm not sure if it would be better to split this two into separate methods.

giovannipizzi:
Hi all.
Indeed, these were utility functions I added at some point. I think the return value is a copy-paste error, so OK to not return anything.
So for this PR I would just remove the return value (and, ideally, add a test for this and for listdir).

More generally, let's discuss in #1857 if we want to keep these functions or how to make it clear that they open an transport and you do not want to do this in a workflow (see also #3478 )

ltalirz:
@giovannipizzi 
Since I think we agree that the return value of this function was really a bug, I think it's fine to change without a deprecation warning.

Please review.

giovannipizzi:
Thanks a lot @Crivella and @ltalirz!
@sphuber can I update this branch and merge? Or are you working on merging? (Ok I you want to take care of this.

sphuber:
Thanks guys, I will merge another PR first that is already good to go and will then update this and merge it

sphuber:
Ok, the other PR has problems with the tests, so I will merge this first after all

PR #3744, opened 2020-02-04
add numfocus affiliation to README

ltalirz:
* add numfocus affiliation to README and docs
 * add twitter badge


ltalirz:
> You should however write and ask them to update their links correctly on the affiliated projects page. Currently they link to ArviZ if you want to go to AiiDA.

Good catch! will let them know immediately

PR #3745, opened 2020-02-06
Install aiida in conda environment, fix problem with .bashrc

### LINKED ISSUE #3739: Collect issues for the docker container.
There are tiny little issues that are related to the Docker container that I find here and there. Instead of making PR for every little issue I decided to collect them here and then solve all of them at once.

So the list of issues:

1. `grep _VERDI_COMPLETE .bash_profile` is done for `.bash_profile`, but should be done for `.bashrc, see [the link]( https://github.com/aiidateam/aiida-core/blob/2308176a7bb629db4d706a3697368fa02dbf97b0/.docker/opt/configure-aiida.sh#L13)

yakutovicha:
fixes #3739 

to be merged after https://github.com/aiidateam/aiida-prerequisites/pull/12

### DIFF
--- a/.docker/opt/configure-aiida.sh
+++ b/.docker/opt/configure-aiida.sh
@@ -6,11 +6,26 @@
 set -x
 
 # Environment.
-reentry scan
 export SHELL=/bin/bash
 
+# Activate conda.
+__conda_setup="$('/opt/conda/bin/conda' 'shell.bash' 'hook' 2> /dev/null)"
+if [ $? -eq 0 ]; then
+    eval "$__conda_setup"
+else
+    if [ -f "/opt/conda/etc/profile.d/conda.sh" ]; then
+        . "/opt/conda/etc/profile.d/conda.sh"
+    else
+        export PATH="/opt/conda/bin:$PATH"
+    fi
+fi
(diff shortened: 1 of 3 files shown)


yakutovicha:
the tests should fail because https://github.com/aiidateam/aiida-prerequisites/pull/12 is not yet merged

ltalirz:
actually, the tests didn't fail.
is that a problem?

yakutovicha:
I thought conda activation part should make it fail since conda wasn't present. But it didn't do so apparently.

yakutovicha:
@ltalirz, I am going to revert the previous commit, but do you have any idea why it doesn't work as expected? It puzzles me a little bit.

yakutovicha:
@ltalirz I added docker log displaying to the tests, so now I am sure that things run under conda. Please review.

ltalirz:
> @ltalirz, I am going to revert the previous commit, but do you have any idea why it doesn't work as expected?

Without looking deep, [the error is](https://github.com/aiidateam/aiida-core/runs/431242020#step:5:6), i.e. it looks like for some reason `$DOCKERID` was empty.

PR #3748, opened 2020-02-10
Minimal implementation of base restart workchain

### LINKED ISSUE #3644: Move BaseRestartWorkChain to aiida-core

sphuber:
Fixes #3644 

This is an alternative PR to #3654 that removes two features that are still contentious:

* Registering additional process handlers per workchain implementation through entry points
* Registering additional process handlers per workchain instance launched through inputs

The other PR contained a first implementation of these features but since the provenance is easily lost when using these constructs, we postpone integration until later when we have had more time to investigate them. Since they will be purely additive features they can be added later without affecting the existing interface.

### DIFF
--- a/.github/workflows/setup.sh
+++ b/.github/workflows/setup.sh
@@ -15,12 +15,14 @@ sed -i "s|PLACEHOLDER_PROFILE|test_${AIIDA_TEST_BACKEND}|" "${CONFIG}/profile.ya
 sed -i "s|PLACEHOLDER_DATABASE_NAME|test_${AIIDA_TEST_BACKEND}|" "${CONFIG}/profile.yaml"
 sed -i "s|PLACEHOLDER_REPOSITORY|/tmp/test_repository_test_${AIIDA_TEST_BACKEND}/|" "${CONFIG}/profile.yaml"
 sed -i "s|PLACEHOLDER_WORK_DIR|${GITHUB_WORKSPACE}|" "${CONFIG}/localhost.yaml"
-sed -i "s|PLACEHOLDER_REMOTE_ABS_PATH|${GITHUB_WORKSPACE}/.ci/doubler.sh|" "${CONFIG}/doubler.yaml"
+sed -i "s|PLACEHOLDER_REMOTE_ABS_PATH_ADD|${GITHUB_WORKSPACE}/.ci/add.sh|" "${CONFIG}/add.yaml"
+sed -i "s|PLACEHOLDER_REMOTE_ABS_PATH_DOUBLER|${GITHUB_WORKSPACE}/.ci/doubler.sh|" "${CONFIG}/doubler.yaml"
 
 verdi setup --config "${CONFIG}/profile.yaml"
 verdi computer setup --config "${CONFIG}/localhost.yaml"
 verdi computer configure local localhost --config "${CONFIG}/localhost-config.yaml"
 verdi code setup --config "${CONFIG}/doubler.yaml"
+verdi code setup --config "${CONFIG}/add.yaml"
 
 verdi profile setdefault test_${AIIDA_TEST_BACKEND}
 verdi config runner.poll.interval 0
--- a/aiida/engine/processes/workchains/restart.py
+++ b/aiida/engine/processes/workchains/restart.py
@@ -1,0 +1,305 @@
+# -*- coding: utf-8 -*-
+"""Base implementation of `WorkChain` class that implements a simple automated restart mechanism for sub processes."""
+from aiida import orm
+from aiida.common import AttributeDict
+from aiida.common.lang import classproperty
+
+from .context import ToContext, append_
+from .workchain import WorkChain
+from .utils import ProcessHandlerReport
+
+__all__ = ('BaseRestartWorkChain',)
+
+
+class BaseRestartWorkChain(WorkChain):
+    """Base restart work chain.
+
+    This work chain serves as the starting point for more complex work chains that will be designed to run a sub process
+    that might need multiple restarts to come to a successful end. These restarts may be necessary because a single
+    process run is not sufficient to achieve a fully converged result, or certain errors maybe encountered which
+    are recoverable.
+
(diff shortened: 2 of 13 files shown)


aiida/engine/processes/workchains/utils.py
```diff
@@ -0,0 +1,118 @@
+# -*- coding: utf-8 -*-
+"""Utilities for `WorkChain` implementations."""
+from collections import namedtuple
+from functools import wraps
+
+from ..exit_code import ExitCode
+
+__all__ = ('ProcessHandler', 'ProcessHandlerReport', 'register_process_handler')
+
+ProcessHandler = namedtuple('ProcessHandler', 'method priority exit_codes')
```
yakutovicha: Maybe replace `method` with `process_class`? I am not sure what exactly you wanted to say here, but it is definitely confusing.

aiida/engine/processes/workchains/utils.py
```diff
@@ -0,0 +1,118 @@
...
+ProcessHandlerReport.__new__.__defaults__ = (False, ExitCode())
+"""A namedtuple to define a process handler report for a :class:`aiida.engine.BaseRestartWorkChain`.
+
+This namedtuple should be returned by a process handler of a workchain instance if the condition of the handler was
+met by the completed process. If no further handling should be performed after this method the 'do_break' field should
+be set to `True`. If the handler encountered a fatal error and the work chain needs to be terminated, an `ExitCode` with
+non-zero exit status can be set. This exit code is what will be set on the work chain itself.
+
+:param do_break: boolean, set to `True` if no further process handlers should be called, default is `False`
+:param exit_code: an instance of the :class:`~aiida.engine.processes.exit_code.ExitCode` tuple
```
yakutovicha: Maybe say that by default the `ExitCode` is 0 which means succesfull error handling.

aiida/engine/processes/workchains/restart.py
```diff
@@ -0,0 +1,304 @@
...
+            Process  Handler    Handler     Action
+            result   report?    exit code
+            -----------------------------------------
+            Success      yes        == 0     Restart
+            Success      yes        != 0     Abort
+            Failed       yes        == 0     Restart
+            Failed       yes        != 0     Abort
+
+        If no handler returned a report and the process finished successfully, the work chain's work is considered done
+        and it will move on to the next step.
```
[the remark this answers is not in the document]
sphuber: I added a specification to say that it moves to the next step following the `while` block in the outline, if any is defined

.github/workflows/setup.sh
```diff
@@ -15,12 +15,14 @@ sed -i "s|PLACEHOLDER_PROFILE|test_${AIIDA_TEST_BACKEND}|" "${CONFIG}/profile.ya
...
+sed -i "s|PLACEHOLDER_REMOTE_ABS_PATH_DOUBLER|${GITHUB_WORKSPACE}/.ci/doubler.sh|" "${CONFIG}/doubler.yaml"
 
 verdi setup --config "${CONFIG}/profile.yaml"
 verdi computer setup --config "${CONFIG}/localhost.yaml"
 verdi computer configure local localhost --config "${CONFIG}/localhost-config.yaml"
 verdi code setup --config "${CONFIG}/doubler.yaml"
+verdi code setup --config "${CONFIG}/add.yaml"
 
 verdi profile setdefault test_${AIIDA_TEST_BACKEND}
 verdi config runner.poll.interval 0
```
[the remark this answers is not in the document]
sphuber: Not sure what needs to be changed here?

aiida/engine/processes/workchains/restart.py
```diff
@@ -0,0 +1,304 @@
+# -*- coding: utf-8 -*-
+"""Base implementation of `WorkChain` class that implements a simple automated restart mechanism for sub processes."""
```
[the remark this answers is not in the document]
sphuber: It _does_ in fact implement the mechanism so prefer to keep that word

aiida/engine/processes/workchains/utils.py
sphuber: The `method` argument will reference the actual function that was decorated with the `registered_process_handler` decorator. I decided to call it a method instead of function since once it has been decorated and turned into this tuple, the decorated function actually has become a method of the target process class.

.github/workflows/setup.sh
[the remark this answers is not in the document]
yakutovicha: just added a terminating empty line.

aiida/engine/processes/workchains/utils.py
sphuber: After considering renaming the `method` element to `handler` I realized that this really makes the code in `BaseRestartWorkChain.inspect_process` difficult to understand. In the loop over registered handlers, the call `handler.method(self, node)` would become `handler.handler(self, node)` which is less clear. Since the `ProcessHandler` tuple is anyway only for internal use and does not concern the user, I have left it this way. I _did_ clarify the docstring of the tuple with regards to the meaning of the `method` tuple element.

yakutovicha: looks good to me.

PR #3758, opened 2020-02-12
select python version for conda CI test

### LINKED ISSUE #3757: CI: conda check does not test on python 3.7
It turns out that when creating a conda environment from an `environment.yml` file, [it is not possible to specify or override the python version of the environment from the command line](https://github.com/conda/conda/issues/9506).

I.e. it turns out that this code
https://github.com/aiidateam/aiida-core/blob/a92c1f1bdbcc73f7910bef54700100e56928ae69/.github/workflows/conda.sh#L16
actually [sets up a python 3.6 environment](https://github.com/aiidateam/aiida-core/runs/441114729#step:4:1991) (despite `PYTHON_VERSION` being 3.7).

While I don't quite understand where the python version 3.6 comes from, I suggest for the time being we explicitly set it in the `environment.yml` file.

ltalirz:
fixes #3757 

It turns out that when creating a conda environment from an
`environment.yml` file, it is not possible to specify or override the
python version of the environment from the command line.

I.e. it turns out that

   conda env create -f environment.yml -n test-environment python=3.7

actually sets up a python 3.6 environment.

Until this functionality becomes available, I suggest we explicitly set
the python version it in the `environment.yml` file.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,4 +1,4 @@
-# Usage: conda env create -n myenvname -f environment.yml python=3.6
+# Usage: conda env create -n myenvname -f environment.yml
 ---
 name: aiida
 channels:
@@ -6,6 +6,7 @@ channels:
 - conda-forge
 - etetoolkit
 dependencies:
+- python==3.7
 - aldjemy~=0.9.1
 - alembic~=1.2
 - circus~=0.16.1
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ channels:
 - conda-forge
 - etetoolkit
 dependencies:
-- python==3.6
+- python~=3.7
 - aldjemy~=0.9.1
 - alembic~=1.2
 - circus~=0.16.1
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ channels:
 - conda-forge
 - etetoolkit
 dependencies:
-- python==3.7
+- python==3.6
 - aldjemy~=0.9.1
 - alembic~=1.2
 - circus~=0.16.1
--- a/.github/workflows/conda.sh
+++ b/.github/workflows/conda.sh
@@ -7,10 +7,6 @@ export PATH="$HOME/miniconda/bin:$PATH"
 hash -r
 conda config --set always_yes yes --set changeps1 no
 
-# Workaround for https://github.com/conda/conda/issues/9337
-pip uninstall -y setuptools
-conda install setuptools
-
 conda update -q conda
 conda info -a
 conda env create -f environment.yml -n test-environment
--- a/.github/workflows/conda.sh
+++ b/.github/workflows/conda.sh
@@ -13,4 +13,4 @@ conda install setuptools
 
 conda update -q conda
 conda info -a
-conda env create -f environment.yml -n test-environment python=$PYTHON_VERSION
+conda env create -f environment.yml -n test-environment
--- a/utils/validate_consistency.py
+++ b/utils/validate_consistency.py
@@ -291,7 +291,7 @@ def update_environment_yml():
     # python version cannot be overriden from outside environment.yml
     # (even if it is not specified at all in environment.yml)
     # https://github.com/conda/conda/issues/9506
-    conda_requires = ['python==3.6']
+    conda_requires = ['python~=3.7']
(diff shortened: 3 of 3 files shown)


sphuber:
So if I understand your description in the issue correctly, building the conda environment with python 3.7 does not work because of dependency issues? If that is the case, then why do we specify `python~=3.6` instead of `python=>3.5,<3.7` ?

ltalirz:
> So if I understand your description in the issue correctly, building the conda environment with python 3.7 does not work because of dependency issues? If that is the case, then why do we specify python~=3.6 instead of python=>3.5,<3.7 ?

Well, conda will anyhow select the right python version at the moment -- once 3.7 starts working it should pick it automatically...

sphuber:
I have no clue what is going on here I am afraid... The action is still running with python 3.7 and this is confirmed in the conda install section:
```
 conda info -a

     active environment : None
       user config file : /home/runner/.condarc
 populated config files : /home/runner/.condarc
          conda version : 4.8.2
    conda-build version : not installed
         python version : 3.7.4.final.0
       virtual packages : __glibc=2.27
```
So then why is python 3.7 here not a problem?

ltalirz:
>  So then why is python 3.7 here not a problem?

I think the python version you see there is the one that comes with the conda install by default.

conda in itself is fine with python 3.7. Also `conda create python=3.7` will create an environment with python 3.7 successfully.

However, it seems there are issues with many of the packages on conda-forge.

astamminger:
@ltalirz 

A small remark:

I was playing around with the compatible version thing and just realized that ``~=3.7`` translates to ``==3.*,>=3.7``, so, in order to limit python to ``3.7``, I think the formally correct way to specify the python version in the ``environment.yml`` would be ``python~=3.7.0`` (i.e. ``==3.7.*,>=3.7.0``)

ltalirz:
You are correct, but since conda will anyhow not install a later python version if not all dependencies are available for it, I was thinking of allowing 3.7+.
If you like, we can even go back to `~=3.6`, which should (?) now also end up installing python 3.7.

astamminger:
> I was thinking of allowing 3.7 and later.

Oh, in that case ignore my comment :)

ltalirz:
@sphuber In that case, this is also ready to merge

environment.yml
```diff
@@ -1,11 +1,12 @@
-# Usage: conda env create -n myenvname -f environment.yml python=3.6
+# Usage: conda env create -n myenvname -f environment.yml
 ---
 name: aiida
 channels:
 - defaults
 - conda-forge
 - etetoolkit
 dependencies:
+- python~=3.7
```
sphuber: Does this mean that `aiida-core` can now only be installed through conda with python>=3.7? Can someone with python 3.5 still install through conda? Because we do support 3.5 and 3.6

ltalirz: somebody using this environment.yml without changes will install aiida-core into an environment that has python>=3.7.

However, this would happen even if we were to remove this line (since - I believe - conda will choose the latest python version available in that case).

The only way to force (& test) build using other python versions would be to create copies of this file on travis that fix the minor version to 3.5, 3.6 and 3.7 and create an environment for each.

sphuber: Sorry for being a pain, but I am afraid I still don't fully understand. I thought you showed that installing `aiida-core` with conda will result in `python==3.6.10` being used, because apparently, various dependencies are not compatible with `python>=3.7`. Why then, are we here requiring `python~=3.7` when we know this should not work with conda _and_ as far as `aiida-core` is concerned, we support `python~=3.5`?

ltalirz: Sure, no problem ;-)

>  I thought you showed that installing aiida-core with conda will result in python==3.6.10 being used, because apparently, various dependencies are not compatible with python>=3.7

Correct, but then I fixed that by making plumpy available under python 3.7 (otherwise the tests on this PR would fail).

sphuber: So it was really just `plumpy` that was the only dependency not compatible with python 3.7?

ltalirz: Indeed - from conda's the error message I would have never guessed this.
Big thanks to @astamminger for figuring this one out.

sphuber: OK all clear now, thanks a lot to both!

PR #3763, opened 2020-02-16
set password for postgresql database

### LINKED ISSUE #3762: Unable to run CI tests on password-protected DB
It seems something changed in the postgresql configuration on github actions:
https://github.com/aiidateam/aiida-core/pull/3760/checks?check_run_id=448911931#step:8:32

E.g. one might expect something like this, if the configuration in the `pg_hba.conf` was changed from "trust" to "md5" (see also [here](https://stackoverflow.com/questions/17996957/fe-sendauth-no-password-supplied#) ).

ltalirz:
fix #3762


ltalirz:
for some reason, two tests are always cancelled after 30s (I already restarted once)
No idea what is going on here...

CasperWA:
Another solution would be to have our own `postgres.conf` and `pg_hba.conf` files and use the forked action [here](https://github.com/hank-cp/postgresql-action)?

CasperWA:
Furthermore, I have just forked the repo and added an `EXPOSE 5432` which solves the issue of us installing PostgreSQL via `apt` as well as using the GH Action, i.e., we can _just_ use the action.
I tested it with the GH Actions CI I have for `aiida-optimade` - I can try and either push to this PR as well or create a new one?

ltalirz:
If you have a workaround for this issue, it would be great if you can open a PR
- I started working on it but am always interrupted

Besides postgres, please also change the wrapt dependency to `~=1.11.1` to fix the pre-commit issue

ltalirz:
@CasperWA : Are you working on it now?
Otherwise I'll ask someone here

CasperWA:
> @CasperWA : Are you working on it now?
> Otherwise I'll ask someone here

I am, and have been for some time - but not specifically on the SQLA issue. More to see if I can make it work without setting a password.

ltalirz:
Ok, a workaround to get back to a state of not having to set a password will be fine for the moment so that we can get through the backlog of PRs.

PR #3765, opened 2020-02-17
suggest running `reentry scan` when EP not found

ltalirz:
Already had two users at the hackathon stumble over this.
A better error message should make this issue go away.

### DIFF
--- a/aiida/plugins/entry_point.py
+++ b/aiida/plugins/entry_point.py
@@ -258,10 +258,12 @@ def get_entry_point(group, name):
     entry_points = [ep for ep in get_entry_points(group) if ep.name == name]
 
     if not entry_points:
-        raise MissingEntryPointError("Entry point '{}' not found in group '{}'".format(name, group))
+        raise MissingEntryPointError("Entry point '{}' not found in group '{}'.".format(name, group) +
+        'Try running `reentry scan` to update the entry point cache.')
 
     if len(entry_points) > 1:
-        raise MultipleEntryPointError("Multiple entry points '{}' found in group".format(name, group))
+        raise MultipleEntryPointError("Multiple entry points '{}' found in group '{}'. ".format(name, group) +
+                                      'Try running `reentry scan` to repopulate the entry point cache.')
 
     return entry_points[0]
 


aiida/cmdline/params/types/identifier.py
```diff
@@ -59,7 +59,10 @@ def __init__(self, sub_classes=None):
                 try:
                     entry_point = get_entry_point_from_string(entry_point_string)
                 except (ValueError, exceptions.EntryPointError) as exception:
-                    raise ValueError('{} is not a valid entry point string: {}'.format(entry_point_string, exception))
+                    raise ValueError(
+                        '{} is not a valid entry point string: {}\n'.format(entry_point_string, exception) +
+                        'Try running `reentry scan` to update the entry point cache.'
```
sphuber: This is not the only place where this can happen. We should probably move this warning into the base loading function `aiida.plugins.load_plugin` or something. What do you think?

ltalirz: Thanks, yes I just put this fix in quickly during the hackathon in order not to forget.
This should be handled on a higher level - done now.

ltalirz:
Just to mention that this is ready for review since some time

csadorf:
I've updated the expected tests for the `develop` branch to match the current CI configuration. The failing tests on this PR are now only related to insufficient test coverage.

ltalirz:
This PR just needs a reapproval, since the coverage bot dismissed @giovannipizzi 's approval

aiida/plugins/entry_point.py
```diff
@@ -258,10 +258,12 @@ def get_entry_point(group, name):
...
 
     if not entry_points:
-        raise MissingEntryPointError("Entry point '{}' not found in group '{}'".format(name, group))
+        raise MissingEntryPointError("Entry point '{}' not found in group '{}'.".format(name, group) +
+        'Try running `reentry scan` to update the entry point cache.')
 
     if len(entry_points) > 1:
-        raise MultipleEntryPointError("Multiple entry points '{}' found in group".format(name, group))
+        raise MultipleEntryPointError("Multiple entry points '{}' found in group '{}'. ".format(name, group) +
+                                      'Try running `reentry clear` to reset the entry point cache.')
```
sphuber: Does `reentry clear` also do a `reentry scan`? I always do a `reentry scan` both in the case of unknown and duplicate entry points and that always works. Would it be easiest to just always advise to run `reentry scan`?

ltalirz: I think you're right. Will fix.

ltalirz:
> why are the dependencies updated in this? i presume this is due to the new dependency bot? Is this already fixed?

This was due to a bug in the GH actions setup, which has been fixed in the meanwhile.
I've removed those changes from the PR now.

PR #3771, opened 2020-02-19
Update dependency management workflow

csadorf:
Initial pass at starting to update the dependency workflow.


csadorf:
@ltalirz I'm investigating the CI issues right now.

ltalirz:
@csadorf I've started pinning down the postgres issue, see my post here
https://github.com/aiidateam/aiida-core/issues/3762

The wrapt issue in pre-commit is easy to fix - just replace the dependency by `wrapt~=1.11.1`

csadorf:
Looks like all directly PR-related issues are resolved here. I'll wait for the postgres CI issue to be resolved upstream.

.github/CODEOWNERS
```diff
@@ -0,0 +1,8 @@
+# All files related to dependency management are owned by the
+# currently active dependency manager (DM) to trigger an automatic review
+# request from the DM upon changes. Please see AEP-002 for details:
+# https://github.com/aiidateam/AEP/tree/master/002_dependency_management
+setup.*             @csadorf
+environment.yml     @csadorf
+requirements*.txt   @csadorf
+pyproject.toml      @csadorf
```
ltalirz: What is the policy for such specifications concerning the other files of aiida-core?
For which files should we aim to specify code owners?

csadorf: I do not think that such a policy exists yet, but we could of course establish one. Should we that first?

ltalirz: I think it would be a good idea to discuss this - at the next AiiDA meeting, and perhaps even before via email; also to explain what the term means in practice for us.
E.g. currently, when assigning reviewers, we've simply been following github's suggestions (based on git blame).

csadorf: The main reason that I added this was to ensure that changes that directly affect the dependency management are reviewed by the DM. I don't think we necessarily have to adopt that for the whole code base.

csadorf: Based on offline discussion, it appears we will not adopt code owners for any other parts of the code.

utils/dependency_management.py
```diff
@@ -0,0 +1,419 @@
...
+        if re.match(pattern, str(req)):
+            return Requirement.parse(re.sub(pattern, replacement, str(req)))
+    return req  # did not match any of the replacement patterns, just return original
+
+
+@click.group()
+def cli():
+    """Utility to update dependency requirements for `aiida-core`.
+
+    Since `aiida-core` fixes the versions of almost all of its dependencies, once in a while these need to be updated.
```
ltalirz: If I understand correctly, what you are referring to here is actually increasing the minimal version requirements of the dependencies.

Might be good to explain when we do this (e.g. before new minor releases / major releases / ...).

To play devil's advocate, one could also argue that such an update actually reduces compatibility of AiiDA with python packages without providing a clear benefit to the user.

utils/dependency_management.py
```diff
@@ -0,0 +1,419 @@
...
+    This is a manual process, but this CLI attempts to simplify it somewhat. The idea is to remote all explicit version
```
ltalirz:
```suggestion
    This is a manual process, but this CLI attempts to simplify it somewhat. The idea is to remove all explicit upper version
```

utils/dependency_management.py
```diff
@@ -0,0 +1,419 @@
...
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Utility CLI to update dependency version requirements of the `setup.json`."""
```
ltalirz: and the environment.yml, ...

utils/dependency_management.py
```diff
@@ -0,0 +1,419 @@
...
+def _setuptools_to_conda(req):
+    for pattern, replacement in SETUPTOOLS_CONDA_MAPPINGS.items():
+        if re.match(pattern, str(req)):
+            return Requirement.parse(re.sub(pattern, replacement, str(req)))
+    return req  # did not match any of the replacement patterns, just return original
+
+
+@click.group()
+def cli():
+    """Utility to update dependency requirements for `aiida-core`.
```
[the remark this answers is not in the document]
csadorf: I realized that I forgot to prune this particular doc-string. This was left-over from the original script, that worked quite differently.

utils/dependency_management.py
csadorf: Same here, all of this does not make any sense anymore, will revise.

sphuber:
Is it really necessary to vendor `toml`?

csadorf:
> Is it really necessary to vendor `toml`?

Of course not, but is there a strong reason against it?

sphuber:
> > Is it really necessary to vendor `toml`?
> 
> Of course not, but is there a strong reason against it?

Well it adds a lot of code to the repository that also has to then be maintained as well. Is this just necessary for the build system? If so, why can't this just be a dependency?

csadorf:
> > > Is it really necessary to vendor `toml`?
> > 
> > 
> > Of course not, but is there a strong reason against it?
> 
> Well it adds a lot of code to the repository that also has to then be maintained as well. Is this just necessary for the build system? If so, why can't this just be a dependency?

It's a dependency of the `dependency_management.py` script. I will need to install it everywhere I want to use that script. It's what other build-related packages do as well (like setuptools, etc.). 

I'll just remove it since this appears to be an issue.

csadorf:
@ltalirz This is ready for a second review. I've removed all obsolete and left-over code from the `dependency_management.py` script and updated the doc-strings. 

Furthermore, I've added a dedicated GitHub actions workflow focused on dependency management. All of this is described in detail here: https://github.com/aiidateam/aiida-core/wiki/AiiDA-Dependency-Management

I'll continue to streamline this as we move forward, but I'd like to merge this at this point.
Please note that we will need to remove the `conda` job from the expected jobs before merging this.

.github/workflows/dm.yml
```diff
@@ -0,0 +1,184 @@
...
+
+    - name: Freeze test environment
+      run: |
+        pip freeze | tee requirements-${{ matrix.backend }}-py-${{ matrix.python-version }}.txt
+
+    - name: Upload requirements.txt
+      uses: actions/upload-artifact@v1
+      with:
+        name: requirements.txt
+        path: requirements-${{ matrix.backend }}-py-${{ matrix.python-version }}.txt
```
ltalirz: Ok, so the requirements.txt will include everything -atomic_tools, docs, notebook, rest and testing - correct?

utils/dependency_management.py
```diff
@@ -0,0 +1,325 @@
...
+    ctx.invoke(validate_requirements_for_rtd)
+    ctx.invoke(validate_pyproject_toml)
+
+
+@cli.command()
+@click.argument('extras', nargs=-1)
+def pip_install_extras(extras):
+    """Install extra requirements.
+
+    Install extra requirements in isolation without triggering the installlation
```
ltalirz:
```suggestion
    Install *only* extra requirements without triggering the installation of
```

utils/dependency_management.py
```diff
@@ -0,0 +1,325 @@
...
+    of other installation requirements of the core package.
```
ltalirz:
```suggestion
    of the main installation requirements of the aiida-core package.
```

.github/workflows/dm.yml
```diff
@@ -0,0 +1,184 @@
...
+
+  update-requirements:
```
ltalirz: One thing is to check daily that aiida installs fine & tests run - that's fine.
Updating the fixed requirements.txt I see as less urgent and, as mentioned before, I feel it is almost better not to run it too often. Last time I checked, overall aiida-core had more than a hundred dependencies, i.e. I suspect this job will be triggered very often.

Since the main purpose of the requirements.txt (at least to me) is reproducibility, I would suggest to tie it to a release - so, it could e.g. be a separate workflow that only runs on PRs to `master` (and perhaps even allow failures in order not to block releases if, in between reviews, a dependency is updated).

.github/workflows/dm.yml
csadorf: Yes, I think ideally we would only include the base requirements, but since tests will currently fail if those extras are not installed, I don't think I can avoid that. We should update the tests to reliably skip tests that require extras as independent effort.

.github/workflows/dm.yml
csadorf: I think that's a good call. I'll update the workflow accordingly.

csadorf:
> One thing I noticed is that running `./dependency_management.py generate-all` results in a different `pyproject.toml` then the one that is committed.

I've fixed the command to better preserve the current file, however comments will unfortunately not be preserved. That's a current limitation of the toml package (see https://github.com/uiri/toml/issues/77).

> I think at least the validation of consistency should be done on every commit.

That's the case, because the consistency is checked as part of pre-commit hooks.

> > I'll continue to streamline this as we move forward, but I'd like to merge this at this point.
> > Please note that we will need to remove the conda job from the expected jobs before merging this.
> 
> Not quite sure what you mean here - I understand you are moving it from the build on every commit to the nightly build (which makes sense). Is there something else we need to do?

The aiida-core repository branch protection is currently configured to expect the success of a "conda" job (see as part of the checks: "_conda Expected — Waiting for status to be reported_"). I have renamed the job and further made it part of the _dm_ workflow. So it should no longer be expected to pass on every commit. I can make that change myself once this PR is approved prior to merging.

.github/workflows/dm.yml
csadorf: Done.

csadorf:
@ltalirz I've addressed all comments and change requests and split the `update-requirements` job into a separate *release* workflow.

One thing to note is that the introduction of yet another workflow that needs to run tests, creates the second duplication of our 'tests' job (I only run one Python version and backend variant on the release workflow). I've done some research, but it appears that it is currently not possible to "reuse" a job defined in one workflow in a different workflow. Since yaml does not support including other files, we are stuck with that duplication. The only thing we could do to deduplicate somewhat is to streamline the current test job. For example, we could refactor some of the steps into local actions or place more commands into a script. The latter would reduce transparency with regards to workflow execution status and failure diagnostic. I'd recommend to make eventual refactoring efforts part of a separate PR.

csadorf:
> Thanks for the clarification - I wasn't aware that with github actions we now have such fine-grained control in the branch rules.

That's not limited to GitHub actions. What I am referring to are the [branch protection rules](https://github.com/aiidateam/aiida-core/settings/branches). I've already made the adjustment for the develop branch.

> 1. continuous-integration (ci.yml): make sure that changes in aiida-core code don't break the tests. Runs on every commit

And ensure consistency.

> 2. dependency-management (dm.yml): make sure that changes in the python package ecosystem don't break the ability to install aiida & to run the tests. Runs daily.

Great. I might add "and that changes that we introduce to our specification of dependencies don't break the ability to install aiida and have all tests pass. The workflow is also executed when changes are made to `setup.json` etc.

> 3. release (release.yml): make sure that aiida-core releases are shipped with a requirements.txt that works and will continue to work. runs before making a release.

I can imagine that there might be a few other things we want to do that, but for now, that's it.

> * I notice that committing the `requirements.txt` to the git repository _and_ updating it only for releases does not work well together. E.g. when updating minimum requirements of package versions, it is strange to have an outdated (and not working) requirements.txt in the same repo.

It boils down to our definition of this file. However, updating the package requirements would trigger the dm-workflow, and thus a potential update of the file. However, we are currently not ensuring that they are consistent on the same commit. And that is because this procedure inherently requires two steps. One is updating the requirements, the next is creating a clean environment with those requirements and making sure that all tests pass. My automatic PR workflow is supposed to solve that issue.

>   If we agree that a requirements.txt should be shipped together with a release, then perhaps it makes sense not to include it in the repo, generate it _only_ in the `release` workflow, and append it (manually or otherwise) as an [asset](https://github.com/aiidateam/aiida-core/releases/tag/v1.1.1) to the release

Counter proposal: We could use the `requirements.txt` to create the environment to run the CI tests. in this way we can guarantee that a particular commit will run with that particular environment. And then nightly and whenever dependency specifications are changed we test against the ecosystem.

> * ci.yml then does not need to update the `requirements.txt` github artifact

Yes, I left that in there for convenience, because it would enable a developer to download it to recreate the exact conditions on the server. We can remove that when we use the workflow described above.

> * the current dm.yml mixes a bit purposes 1 and 2 - it checks both (daily) for changes in the ecosystem & (on every commit) for changes in the dependency specification in aiida-core. I agree that one may want to mix this (it just makes reasoning about the logic a little more tricky)

I can live with that mix.

> * dm.yml then does not need to update the `requirements.txt` github artifact

Sure, but one might want to keep it for the reason state above.

> * release.yml should (a) generate the requirements.txt, (b) use it to install & run the tests and (c) upload it as an artifact

> A general comment/question about github artifacts is that they are global (?), i.e. we should probably use them with care (e.g. imagine someone releasing a 1.0.2 version of AiiDA etc.).

I am pretty sure that their scope is one workflow instance.

ltalirz:
> Counter proposal: We could use the requirements.txt to create the environment to run the CI tests. in this way we can guarantee that a particular commit will run with that particular environment. And then nightly and whenever dependency specifications are changed we test against the ecosystem.

Could add a few more details to this counter proposal?
I.e., in particular, when is the requirements.txt (that is committed to the repo) updated in this context?

So far, I still tend to think that we are making our life unnecessarily hard by putting something into our repo (or artifacts) that is difficult to keep up to date.

> I am pretty sure that their scope is one workflow instance.

I just checked the api documentation and this looks to me like they are global:
https://developer.github.com/v3/actions/artifacts/

csadorf:
@ltalirz The one thing I forgot when I wrote my comment above is that `requirements.txt` files differs across Python versions and backends. So there is not **one** `requirements.txt` file that will necessarily work across all of these different combinations. This also means that my proposed workflow of using the `requirements.txt` file to create the test environment, wouldn't work without having multiple files.

Our general alternatives in providing a `requirements.txt` file for the purpose of reproduciblity are:

 - A) ship all of them with the repository with naming scheme `requirements-{backend}-py-{python-version}.txt`, 
 - B) [**this branch**] select one (e.g. py38, django),
 - C) [**develop**] not provide one at all, or
 - D) only add them to the GitHub release.

A) Would be most reproducible and functional, but might create some confusion.

B) That is what is currently implemented on this branch. While somewhat reasonable, it also seems a bit arbitrary and I am not sure whether we want to play favorites with our supported versions and backends.

C) That's the status quo.

D) Like A), this would make it significantly easier for users to recreate the conditions at the given release time and massively improve their chances to reproduce results with a given version of aiida-core. The risk here is that the availability of these files is bound to the survival of the GitHub organization. While that seems to be a pretty safe bet at the moment, if we think on the order of 5-10 years, it is not so likely that the service will still be available and used in the form it is today. The files are also less discoverable and also more likely to not be available, since they are not part of the repository.

In summary, I tend to favor A). While slightly awkward, it reduces the risk that the file is not available when needed, we can test it on every commit and in this way also makes those tests more reproducible.

csadorf:
> > Counter proposal: We could use the requirements.txt to create the environment to run the CI tests. in this way we can guarantee that a particular commit will run with that particular environment. And then nightly and whenever dependency specifications are changed we test against the ecosystem.
> 
> Could add a few more details to this counter proposal?
> I.e., in particular, when is the requirements.txt (that is committed to the repo) updated in this context?
> 
> So far, I still tend to think that we are making our life unnecessarily hard by putting something into our repo (or artifacts) that is difficult to keep up to date.

I was in the process of adding another comment while you wrote this. That should provide some more clarification.

> 
> > I am pretty sure that their scope is one workflow instance.
> 
> I just checked the api documentation and this looks to me like they are global:
> https://developer.github.com/v3/actions/artifacts/

It is technically true that one can access artifacts repository wide using the REST API, but artifacts downloaded with [the download-artifact action](https://github.com/actions/download-artifact/blob/master/action.yml) are limited to the workflow scope unless I am mistaken.

ltalirz:
> I was in the process of adding another comment while you wrote this. That should provide some more clarification.

Thanks for the comparison of the different approaches - I agree in general with your comments, but my original question still remains: in options A) and B), when (and how) is the (or the many) requirements.txt file in the repository updated?

Is it updated every release?
Is it updated daily?
Is it updated on every PR?
Who updates it?

csadorf:
> > I was in the process of adding another comment while you wrote this. That should provide some more clarification.
> 
> Thanks for the comparison of the different approaches - I agree in general with your comments, but my original question still remains: in options A) and B), when (and how) is the requirements.txt file (or its variants) in the repository updated?
> 
> Is it updated every release?
> Is it updated daily?
> Is it updated on every PR?
> And who updates it - is it an automatic or a manual procedure?

I am sorry that I missed your question. In case that we use the `requirements-*.txt` files to create the test environments for CI, we are independent from the ecosystem. So an update would only be needed whenever we change any of our requirements. However, it probably makes sense to also update the files with every release, so that we reduce the distance to the current ecosystem on a regular basis.

To achieve this, I suggest we apply the following changes to the workflow configuration:

 1. rename the `dependency-management` workflow to `test-install` and retain triggers,
 2. rename the `release` workflow to `update-requirements` and change the triggers to `on.push.branch:'release/*'` and `on.push.paths:['setup.json']`.

ltalirz:
Thanks for the clarification! I think what you propose is the best way forward.

Using the requirements.txt for CI (as you suggest) has the additional benefit that changes to the ecosystem won't disrupt CI on regular PRs, which is a significant plus.

In addition to the changes you propose, I guess this also means that the github artifact is no longer needed, correct?
The tests will simply use the requirements files that are committed to the repository.
In my experience, caches always make things more complicated, so it's good if we can avoid them in this case - also, relying on the committed requirements files makes it easier to reproduce issues locally.

Thank you for your patience!
I'll have another good look through the workflow setup once you've made the changes.

ltalirz:
P.S. If the requirements.txt are used in tests, we could think about putting them in `.ci/` or in `tests/` in order not to pollute the top level.
Of course they would also be a bit less visible there... up to you for them moment - after we agree I think we'll still need to briefly check with Seb or Gio.
After I give my next pass, I (or you) can update the PR description, outlining what we've agreed together.

.github/workflows/ci.yml
```diff
@@ -119,11 +100,17 @@ jobs:
...
+    - name: upgrade setuptools [py35]
+      if: matrix.python-version == 3.5
+      run: pip install -I setuptools==38.2.0
+
+    - name: Install aiida-core
       run: |
-        pip install --upgrade pip
-        pip install numpy==1.17.4
-        pip install -e .[atomic_tools,docs,notebook,rest,testing]
+        pip --version
```
csadorf: This should probably be combined with `pip install --upgrade pip`.

.github/workflows/ci.yml
```diff
@@ -91,8 +72,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.5, 3.8]
         backend: ['django', 'sqlalchemy']
+        python-version: [3.5, 3.8]
```
csadorf: This is a inconsequential change, I can revert if necessary.

.github/workflows/test-install.yml
```diff
@@ -0,0 +1,153 @@
...
+  validate-dependency-specification:
+    # Note: The specification is also validated by the pre-commit hook.
+
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python 3.7
```
csadorf: I plan on harmonizing the use of specific Python version like here in a separate PR.

requirements/requirements-django-py-3.5.txt
```diff
@@ -0,0 +1,151 @@
+aiida-export-migration-tests==0.8.0
```
csadorf: Should we leave these files in the visible `requirements/` directory or should we make them hidden?

setup.json
```diff
@@ -33,7 +33,8 @@
     "ipython~=7.0",
     "jinja2~=2.10",
     "kiwipy[rmq]~=0.5.1",
-    "numpy~=1.17,<1.18",
+    "numpy~=1.17,<1.18; python_version>='3.6'",
+    "numpy>=1.17.0,<1.17.5; python_version=='3.5'",
```
csadorf: I think we should actually revert that change, since I believe that it did not have the desired effect. I'll test that as part of the next iteration.

setup.py
```diff
@@ -10,11 +10,26 @@
...
+if (sys.version_info.major, sys.version_info.minor) == (3, 5):
+    import setuptools
+    from distutils.version import StrictVersion
+
+    REQUIRED_SETUPTOOLS_VERSION = StrictVersion('38.2.0')
+    INSTALLED_SETUPTOOLS_VERSION = StrictVersion(setuptools.__version__)
+
+    if INSTALLED_SETUPTOOLS_VERSION < REQUIRED_SETUPTOOLS_VERSION:
+        raise RuntimeError(
+            'The installation of AiiDA with Python version 3.5, requires setuptools>={}; your version: {}'.format(
```
csadorf: Should this rather be "installation of aiida-core" ?

csadorf:
@ltalirz This is ready for your review. I added a few comments and questions as well. Would appreciate if you could respond to them.

.github/workflows/ci.yml
ltalirz: sounds good

.github/workflows/ci.yml
```diff
@@ -147,6 +134,9 @@ jobs:
         file: ./coverage.xml
         fail_ci_if_error: true
 
+    - name: Freeze test environment
```
ltalirz: This could probably be combined with "Install aiida-core"

.github/workflows/test-install.yml
```diff
@@ -0,0 +1,153 @@
...
+  push:
+    branch:
+      - master
+      - develop
+      - release/*
+      - dm/*
+    paths:
+      - 'setup.*'
+      - 'environment.yml'
+      - 'requirements*.txt'
```
ltalirz: this will get stuff in `requirements/` correct?
one could make this more explicit

.github/workflows/test-install.yml
```diff
@@ -0,0 +1,153 @@
...
+      - 'pyproject.toml'
+      - '.github/workflows/*'
```
ltalirz: Is that even necessary?
I.e. do we need to try to install again, when the github actions change?
maybe just of the actual test-install workflow changes?

.github/workflows/test-install.yml
```diff
@@ -0,0 +1,153 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+
+    - name: Install dm-script dependencies
```
ltalirz: could even go into a ./utils/dependency_management_requirements.txt file.
anyhow, up to you

.github/workflows/test-install.yml
```diff
@@ -0,0 +1,153 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setup Conda
+      uses: s-weigand/setup-conda@v1
+      with:
+        update-conda: true
+        python-version: 3.7
+    - run: conda --version
```
ltalirz: these three could probably also be combined into one "Print versions" step

.github/workflows/test-install.yml
```diff
@@ -0,0 +1,153 @@
...
+    - run: python --version
+    - run: which python
+
+    - name: Create conda environment
+      run: |
+        conda env create -f environment.yml -n test-environment
+        source activate test-environment
+        python -m pip install --no-deps -e .
+
+    - name: Test package import
```
ltalirz:
```suggestion
    - name: Test importing aiida
```

.github/workflows/test-install.yml
```diff
@@ -0,0 +1,153 @@
...
+      run: pip install -I setuptools==38.2.0
+
+    - name: Install aiida-core
+      run: |
+        pip install -e .[atomic_tools,docs,notebook,rest,testing]
+        reentry scan
+
+    - run: pip freeze
+
+    - name: Setup environment
```
ltalirz:
```suggestion
    - name: Setup AiiDA environment
```

.github/workflows/update-requirements.yml
```diff
@@ -0,0 +1,100 @@
...
+
+    - name: Freeze test environment
+      run: pip freeze | sed '1d' | tee requirements-${{ matrix.backend }}-py-${{ matrix.python-version }}.txt
+
+    - uses: actions/upload-artifact@v1
+      with:
+        name: requirements.txt
+        path: requirements-${{ matrix.backend }}-py-${{ matrix.python-version }}.txt
+
+  update-requirements:
```
ltalirz: Hm... are we sure this works?
Also, will this create one PR per python version & backend?

requirements/requirements-django-py-3.5.txt
ltalirz: good question. let's leave them here for the moment, and then see how gio/seb about it.

.github/workflows/update-requirements.yml
```diff
@@ -0,0 +1,100 @@
...
+      run:
+        .github/workflows/setup.sh
+
+    - name: Run test suite
+      env:
+        AIIDA_TEST_BACKEND: ${{ matrix.backend }}
+      run:
+        .github/workflows/tests.sh
+
+    - name: Freeze test environment
```
ltalirz: Sorry, I just realized that there should be no differences in packages between django and sqla backend (we currently don't have extras specifically for either of them), so I wonder whether we need to create requirements files for each separately.

setup.json
ltalirz: ok

setup.py
ltalirz: it's fine as is

.github/workflows/test-install.yml
csadorf: No, it actually matches for all files that have a name that matches this pattern across the whole repository.

.github/workflows/test-install.yml
csadorf: I think that would be overkill.

.github/workflows/update-requirements.yml
csadorf: Agreed, only storing files for one backend now.

.github/workflows/update-requirements.yml
csadorf: No, one PR for all versions (no longer differentiating between backends). It has worked during debugging, but I will try to trigger this workflow one more time for the purpose of this PR.

csadorf:
The failing pre-commit jobs will be resolved with #3836 .

.github/workflows/test-install.yml
ltalirz: Ok, fine

.github/workflows/test-install.yml
```diff
@@ -0,0 +1,154 @@
...
+      - develop
+      - release/*
+      - dm/*
+      - misc/dependency-management  # TODO: REMOVE AFTER TESTING
+    paths:
+      - 'setup.*'
+      - 'environment.yml'
+      - 'requirements*.txt'
+      - 'pyproject.toml'
+      - '.github/workflows/test-install.sh'
```
ltalirz:
```suggestion
      - '.github/workflows/test-install.yml'
```

.github/workflows/test-install.yml
csadorf: It's actually kind of nice to be able to open each individual one with a dropdown menu. I'll leave it as-is.

csadorf:
> Alright, let's go!

@ltalirz I'll finalize and merge. :+1:

csadorf:
@ltalirz I am very sorry, I still needed to remove the trigger for running all workflows on this branch. I did not realize that we had automatic dismissal of review approvals enabled.

It appears that unfortunately the latest run is having some (unrelated?) issues. They can probably be addressed by simply restarting the workflows.

Please merge at your discretion; this is the proposed commit-message for the squashed merge:

> Revise dependency management workflow in accordance with AEP 002.
> 
>  - Use .github/CODEOWNERS file to trigger reviews from
>  dependency-manager team.
>  - All jobs related to testing the installation of aiida-core with
>  different tools (pip, conda) on different python versions and backends
>  are moved into the dedicated 'install-tests' workflow.
>  - The test jobs run as part of the continuous integration on every push
>  are executed against pinned environments (e.g.
>  'requirements/requirement-py-37.txt').
>  - The 'install-tests' workflow is triggered for the 'master', 'develop',
>  and release branches ('release/*') as well as branches related to
>  dependency-management (prefixed with 'dm/'). In addition, the workflow
>  is triggered nightly to ensure that changes within the Python ecosystem
>  that break our toolchain and possibly even tests, are automatically
>  detected within 24 hours.
>  - A new 'update-requirements' workflow is run on release branches and
>  automatically creates a pull request with revised versions of
>  requirements files.
>  - The issues caused by pymatgen's use of setup_requires and previously
>  addressed by installing numpy prior to aiida by default, are now
>  addressed by specifically checking the setuptools version for Python
>  3.5 only. We fail the installation process with a descriptive message
>  to the user in case that the installed setuptools version is
>  insufficient.
>  - All dependency-management related utility functions, such as
>  generating and validating non-authoritive dependency specification files
>  (e.g. 'environment.yml') are concentrated into the
>  'util/dependency_management.py' scripting file.
> 
> The complete workflow is described in detail as part of the Dependency
> Management page within the wiki.

csadorf:
@italirz looks like it's ready to merge.

ltalirz:
Ehm... what happened there?
Screenshot 2020-03-09 at 19 55 52

Looks like your PR is cursed ;-)

ltalirz:
Ah, I see - was just the wrong default setting

ltalirz:
Interesting - it seems like the new workflows are immediately applied on existing open PRs

Screenshot 2020-03-09 at 20 03 12

csadorf:
I think they're yellow, because I changed the expected tests. PR authors will need to either rebase or merge develop to trigger the execution of the new workflows.

PR #3781, opened 2020-02-20
Fix CI postgres issue

CasperWA:
Fixes #3763 

I have a working branch.

Forking the original postgresql-action from @Harmon758 (actually from other forks of his, since he is non-responsive for the maintenance of the action), I have updated the action with other forks' contributions as well as some of my own.
Most importantly is one for setting the auth-method when starting `initdb` (see [here](https://hub.docker.com/_/postgres), search for `POSTGRES_INITDB_ARGS`).

I have tested that it works if set to `trust` - which, obviously, it should. But I found that it then also doesn't matter if there is any inconsistencies between the db passwords set for a profile and the actual db password - which, again, makes sense.

Currently, I am testing with the auth-method set to `password` instead.
Edit: This does not seem to work (unless one sets a password, I guess).
So the question is now, do you want to have it set to `trust` so that it all runs, but then we're not sure if some security measures are tested correctly? Or should I try and find a more "secure" auth-method, where it also runs (this may take a lot more time)?

Also, this sets the `wrapt` dependency to `~=1.11.1`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,12 +96,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
-    - uses: harmon758/postgresql-action@v1
+    - uses: CasperWA/postgresql-action@v1.2
       with:
-        postgresql version: '11'
+        postgresql version: '10'
         postgresql db: test_${{ matrix.backend }}
-        postgresql user: 'postgres'
+        postgresql user: postgres
         postgresql password: ''
+        postgresql auth: trust
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v1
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -96,12 +96,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
-    - uses: harmon758/postgresql-action@v1
+    - uses: CasperWA/postgresql-action@v1.2
       with:
-        postgresql version: '11'
+        postgresql version: '10'
```
ltalirz: Can I quickly ask why the downgrade?

ltalirz:
> So the question is now, do you want to have it set to trust so that it all runs, but then we're not sure if some security measures are tested correctly? Or should I try and find a more "secure" auth-method, where it also runs (this may take a lot more time)?

For the moment, the "non-secure" fix is OK.

We will fix the underlying authentication problem later.

.github/workflows/ci.yml
CasperWA: To match the installed postgres version in the system otherwise (latest installed by default for ubuntu-18.04 is v10). Oh, and I forgot to write that we need to install postgres still (note that I have also specified the version there), since otherwise we cannot use `PGTest`.

CasperWA: However, I have (successfully) removed the explicit installation of postgres-server tools and postgres-client (but I think that one is still installed as a dependency of postgres) ... But yeah, that's life.

ltalirz:
Thanks again for working on fixing this @CasperWA, very much appreciated!

I'm restarting the builds on all the open PRs now.

Issue #3773, opened 2020-02-20
`BaseRestartWorkChain`: add enable/disable handlers input port


sphuber:
The `process_handler` should only be applicable to class methods of `BaseRestartWorkChain` sub classes, which means they will also be active by default. We should add an `enabled` keyword to the `process_handler` decorator to change this default behavior and also add an input that allows to either enable or disable specific handlers per instance.

PR #3782, opened 2020-02-20
`BaseRestartWorkChain`: require process handlers to be instance methods

### LINKED ISSUE #3780: `BaseRestartWorkChain`: require process handlers be defined as instance methods
During the hackathon at CINECA it was decided that allowing the `register_process_handler` to attach random functions to a `BaseRestartWorkChain` was too unrestrictive and allows for potential uses that easily causes provenance loss.

sphuber:
Fixes #3780 

The original implementation of the `register_process_handler` allowed an
unbound method to be bound to a subclass of `BaseRestartWorkChain`
outside of its scope. Since this also makes it possible to attach these
process handlers from outside of the module of the work chain, this will
run the risk of the loss of provenance. Here we rename the decorator to
`process_handler` and it can only be applied to instance methods. This
forces the handlers to be defined in the same module as the work chain
to which they apply.

### DIFF
--- a/aiida/engine/processes/workchains/restart.py
+++ b/aiida/engine/processes/workchains/restart.py
@@ -8,9 +8,10 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Base implementation of `WorkChain` class that implements a simple automated restart mechanism for sub processes."""
+import inspect
+
 from aiida import orm
 from aiida.common import AttributeDict
-from aiida.common.lang import classproperty
 
 from .context import ToContext, append_
 from .workchain import WorkChain
@@ -28,8 +29,9 @@ class BaseRestartWorkChain(WorkChain):
     are recoverable.
 
     This work chain implements the most basic functionality to achieve this goal. It will launch the sub process,
-    restarting until it is completed successfully or the maximum number of iterations is reached. It can recover from
-    errors through error handlers that can be registered to the class through the `register_process_handler` decorator.
+    restarting until it is completed successfully or the maximum number of iterations is reached. After completion of
+    the sub process it will be inspected, and a list of process handlers are called successively. These process handlers
+    are defined as class methods that are decorated with :meth:`~aiida.engine.process_handler`.
 
     The idea is to sub class this work chain and leverage the generic error handling that is implemented in the few
     outline methods. The minimally required outline would look something like the following::
@@ -57,13 +59,21 @@ class BaseRestartWorkChain(WorkChain):
     process will be run with those inputs.
 
     The `_process_class` attribute should be set to the `Process` class that should be run in the loop.
+    Finally, to define handlers that will be called during the `inspect_process` simply define a class method with the
+    signature `(self, node)` and decorate it with the `process_handler` decorator, for example::
+
+        @process_handler
+        def handle_problem(self, node):
+            if some_problem:
+                self.ctx.inputs = improved_inputs
+                return ProcessHandlerReport()
+
+    The `process_handler` and `ProcessHandlerReport` support various arguments to control the flow of the logic of the
+    `inspect_process`. Refer to their respective documentation for details.
     """
 
-    _verbose = False
     _process_class = None
-
-    _handler_entry_point = None
-    __handlers = tuple()
+    _considered_handlers_extra = 'considered_handlers'
 
     @classmethod
     def define(cls, spec):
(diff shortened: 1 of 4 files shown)


aiida/engine/processes/workchains/restart.py
```diff
@@ -29,7 +30,7 @@ class BaseRestartWorkChain(WorkChain):
 
     This work chain implements the most basic functionality to achieve this goal. It will launch the sub process,
     restarting until it is completed successfully or the maximum number of iterations is reached. It can recover from
-    errors through error handlers that can be registered to the class through the `register_process_handler` decorator.
+    errors through error handlers that can be registered to the class through the `process_handler` decorator.
```
yakutovicha: > handlers that can be registered to the class

I feel that 'registered to the class' is still referring to the previous approach. One should stress that handlers should be part of a class.

aiida/engine/processes/workchains/restart.py
```diff
@@ -57,13 +58,21 @@ class BaseRestartWorkChain(WorkChain):
     process will be run with those inputs.
 
     The `_process_class` attribute should be set to the `Process` class that should be run in the loop.
+    Finally, to define handlers that will be called during the `inspect_process` simply define a class method with the
+    signature `(self, node)` and decorate it with the `process_handler` decorator, for example::
+
+        @process_handler
+        def handle_problem(self, node):
```
yakutovicha: After the offline discussion yesterday I thought that the `self` argument is not needed anymore. Is it wrong?

I have no problem with either of the approaches, just got confused.

aiida/engine/processes/workchains/restart.py
```diff
@@ -113,12 +122,11 @@ def run_process(self):
         inputs = self._wrap_bare_dict_inputs(self._process_class.spec().inputs, unwrapped_inputs)
         node = self.submit(self._process_class, **inputs)
 
-        # Add a new empty list to the `called_process_handlers` extra. If any errors handled registered through the
-        # `register_process_handler` decorator return an `ProcessHandlerReport`, their name will be appended to that
-        # list.
-        called_process_handlers = self.node.get_extra('called_process_handlers', [])
-        called_process_handlers.append([])
-        self.node.set_extra('called_process_handlers', called_process_handlers)
+        # Add a new empty list to the `BaseRestartWorkChain._considered_handlers_extra` extra. This will the name and
```
yakutovicha:
```suggestion
        # Add a new empty list to the `BaseRestartWorkChain._considered_handlers_extra` extra. This will contain the name and
```

aiida/engine/processes/workchains/restart.py
```diff
@@ -245,23 +249,20 @@ def __init__(self, *args, **kwargs):
...
-        """Register a new handler to this class.
+        handlers = []
 
-        :param name: the name under which to register the handler
-        :param handler: a method with the signature `self, node`.
-        """
-        setattr(cls, name, handler)
-        cls.__handlers = cls.__handlers + (handler,)
+        for method in inspect.getmembers(self, predicate=inspect.ismethod):
+            if hasattr(method[1], 'decorator') and method[1].decorator == process_handler:  # pylint: disable=comparison-with-callable
```
yakutovicha: what about adding a check that handler is active?

aiida/engine/processes/workchains/utils.py
```diff
@@ -49,16 +35,12 @@
 """
 
 
-def register_process_handler(cls, *, priority=None, exit_codes=None):
-    """Decorator to register a function as a handler for a :class:`~aiida.engine.BaseRestartWorkChain`.
+def process_handler(wrapped=None, *, priority=None, exit_codes=None):
```
yakutovicha: what about `enabled_by_default=True` argument?

aiida/engine/processes/workchains/restart.py
sphuber: Agree, will change it

aiida/engine/processes/workchains/restart.py
sphuber: It is still needed, but simply because it is an instance method. But now it should no longer come as a surprise since you have to directly define it within the class.

aiida/engine/processes/workchains/restart.py
sphuber: This will be implemented in PR for the separate issue #3773

aiida/engine/processes/workchains/utils.py
sphuber: same as above see #3773

PR #3783, opened 2020-02-21
Add fixtures to clear the database before or after tests

greschd:
The current 'clear_database' fixture clears the DB after the test has run. Since clearing it before the tests is also a common use case, this adds a 'clear_database_before_test' fixture. For naming consistency, we rename 'clear_database' to 'clear_database_after_test', and turn 'clear_database' into an
alias for that function for backwards compatibility.


sphuber:
Maybe try to get this merged before we merge your cache PR so that one can remove the ad-hoc fix for the database reset

PR #3784, opened 2020-02-21
pyproject: add build-backend

dev-zero:
if there is a `build-system` section, it should also define
the backend, otherwise a PEP517-based build will fail:

    python -m pep517.build --source --binary --out-dir dist/ .


ltalirz:
See https://www.python.org/dev/peps/pep-0517/#build-backend-interface

PR #3785, opened 2020-02-21
Fix caching configuration

### LINKED ISSUE #3673: Caching and the configuration of caching is broken
The problem went undetected because the caching configuration and the methods that change it replaced the type from node classes to entry points. For example the `enable_caching` method deprecated that `node_class` argument in favor of `identifier` however, when `node_class` was passed it was effectively ignored. When properly used the `engine/test_calcfunctions` unit tests will fail.

greschd:
Fixes #3673 

This changes the meaning of the `identifier` used in the caching configuration from being an entry point string, to being the `process_type` of the node.

Since this change removes the ability to select process classes via subclassing (isinstance check), we instead allow the configuration to contain globbing patterns, as implemented in the `fnmatch` standard library.

Multiple globbing patterns can apply to a single process type, as long as one of them is uniquely the most specific.

Since the new configuration allows arbitrary string, the validation of the caching configuration on load is now less strict  -- we can not check if the process type exists.


greschd:
Thinking about this some more, the logic for checking which pattern is "most specific" is incorrect in corner cases. The criterion for being the most specific pattern should be that the set of identifiers that it matches is a strict subset of all other identifiers. Here's a counter example for the current logic:

With patterns
```python
pattern1 = 'a????c?`
pattern2 = 'a[ba]c*'
```
`pattern2` is identified as the most specific. However, considering the identifiers
```python
A = 'afcdecf'
B = 'abccccd'
C = 'abccccde'
```
`pattern1` matches {A, B}, while `pattern2` matches {B, C}. Clearly, it is not a strict subset.

The root of this problem lies with the `?` and `[...]` matching kinds - for the latter, a pattern will not even match itself. I have a felling restricting matches to `*` would be enough to make everything work, but I still need to convince myself fully.

Should we just restrict the patterns so that they can not contain `[`, `]`, or `?`? @sphuber what's the "grammar" of the `process_type`, could it ever contain one of these characters?

Besides this issue, we still need to update the documentation about caching configuration.

greschd:
Also, we should be using `fnmatch.fnmatchcase` instead of `fnmatch.fnmatch`.

greschd:
> Also, we should be using `fnmatch.fnmatchcase` instead of `fnmatch.fnmatch`. Or instead of using `fnmatch`, we could split the pattern by `*`, `re.escape` the parts, and then join them with `.*` again to get a regex that does _only_ `*` - matching.

I decided to go with the `re` solution, because that is pretty much exactly what `fnmatch` is doing also: https://github.com/python/cpython/blob/3.8/Lib/fnmatch.py#L74 (but it's _also_ implementing the `?` and `[...]` matching).

The documentation is now also updated, so this is ready for review.

tests/manage/test_caching_config.py
```diff
@@ -0,0 +1,224 @@
...
+        profile_name:
+            default: False
+            enabled:
+                # - aiida.calculations:templatereplacer
+
+    which is not unlikely to occurr in the wild.
+    """
+    with configure_caching(config_dict={'default': True, 'enabled': None, 'disabled': None}):
+        from aiida.manage.caching import _CONFIG
+        print(_CONFIG)
```
sphuber: a wild print appears

greschd: Good old fashioned debugging

aiida/manage/caching.py
```diff
@@ -62,32 +62,19 @@ def _get_config(config_file):
...
-            try:
-                load_entry_point_from_string(identifier)
-            except exceptions.EntryPointError as exception:
-                raise exceptions.ConfigurationError(
-                    "entry point '{}' in 'cache_config.yml' can not be loaded: {}.".format(identifier, exception)
-                )
+    try:
+        type_check(config[ConfigKeys.DEFAULT.value], bool)
+        type_check(config[ConfigKeys.ENABLED.value], list)
+        type_check(config[ConfigKeys.DISABLED.value], list)
```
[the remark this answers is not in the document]
greschd: Yes, I agree it would be good to have at least the validation that it _could_ be a valid process type string. Because of the possible wildcards, we probably have to implement the check separately and can't re-use the ``is_valid_entry_point_string``.

For the entry points, as far as I see the valid form is

``<group_name>:<something>``, where

- ``group_name`` is one of ``aiida.data``, ``aiida.calculations``, ..., as defined by the ``entry_point_group_to_module_path_map``
- ``something`` can be literally anything except it can _not_ contain another semicolon

For the fully qualified python name, each colon-separated part must fulfill
```python
name.isidentifier() and not keyword.iskeyword(name)
```
The only two restrictions I'm aware of (compared to the simple regex) is that it can't start with a number, and can't be a keyword (like ``in``, ``for``, ``else``, ...).

All this is a bit tricky to check for due to the wildcards, but I think I can come up with a reasonable solution.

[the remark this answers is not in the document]
greschd: Alright, had a crack at this in `_validate_identifier_pattern`.

aiida/manage/caching.py
```diff
@@ -163,54 +188,123 @@ def _reset_config():
...
+            _match_wildcard(string=group_name, pattern=group_pattern)
+            for group_name in entry_point_group_to_module_path_map
+        ):
+            raise ValueError(
+                common_error_msg + "Group name pattern '{}' does not match any of the AiiDA entry point group names.".
+                format(group_pattern)
+            )
+        # The group name pattern matches, and there are no further
+        # entry point string separators in the identifier, hence it is
+        # a valid pattern.
```
sphuber: I was just wondering if we should maybe still check the `tail` and see if it corresponds to an actual registered entry point, _if_ it is a fully specified entry point and if not, emit a warning. I am thinking of someone adding a fully qualified entry point string to the config but it contains a type, and then being really surprised that the corresponding class is not cached. Having the warning would help I imagine. But then we would have to cache the parsing and validation of the config as we don't want to rerun this everytime the config is retrieved. What do you think? Too much for now?

aiida/manage/caching.py
```diff
@@ -163,54 +188,123 @@ def _reset_config():
...
+    """
+    common_error_msg = "Invalid identifier pattern '{}': ".format(identifier)
+    assert ENTRY_POINT_STRING_SEPARATOR not in '.*'  # The logic of this function depends on this
+    # Check if it can be an entry point string
+    if identifier.count(ENTRY_POINT_STRING_SEPARATOR) > 1:
+        raise ValueError(
+            common_error_msg +
+            "Can contain at most one entry point string separator '{}'".format(ENTRY_POINT_STRING_SEPARATOR)
+        )
+    # If there is one separator, it must be and entry point string.
```
sphuber:
```suggestion
    # If there is one separator, it must be an entry point string.
```

aiida/manage/caching.py
```diff
@@ -163,54 +188,123 @@ def _reset_config():
...
+            )
+        # The group name pattern matches, and there are no further
+        # entry point string separators in the identifier, hence it is
+        # a valid pattern.
+        return
+    # The separator might be swallowed in a wildcard
+    if '*' in identifier:
+        group_part, _ = identifier.split('*', 1)
+        if any(group_name.startswith(group_part) for group_name in entry_point_group_to_module_path_map):
+            return
```
sphuber: is this block for identifiers like `aiida.*` and `aiida.calculations*`?

greschd: Yes

aiida/manage/caching.py
greschd: Hmm yeah, if there aren't any wildcards we could essentially do the same validation that was there before - for me it would even be fine to make it an error (warnings are very easily ignored..).

Why would it need to be cached, though? It only happens at startup, same as before. It seems difficult to _both_ check that it corresponds to a currently-available entry point _and_ cache it between interpreter restarts.

> but it contains a type
I think you mean "typo" 😄

aiida/manage/caching.py
greschd: The reason for this one is that the `:` could be swallowed by the wildcard. For example, ``aiida.calc*2foo!!`` is, technically, valid. Without this part, it would be treated as a Python name and fail due to the invalid characters after the wildcard.

aiida/manage/caching.py
sphuber: > I think you mean "typo" smile

Haha, I guess one would call that a Freudian ~~slip~~ typo :wink:
Regarding the caching: I quickly looked at the code and didn't see the parsing happening only once, but if that is the case, then there is no problem.

The reason I  think raising an error is maybe not the right choice is that then one is _forced_ to have only entry points in there that are currently installed. Imagine I have a working config and then uninstall a plugin, then all startups the config will crash, which would be quite annoying. But maybe we can just leave this for now. If it ever comes up as a problem then we can always add more heavy validation.

If you can just address the one remaining change to `allow_none=True` in the `type_check` we can merge this.

aiida/manage/caching.py
greschd: Oh no... entry point names can contain the literal wildcard character 🤦‍♂

aiida/manage/caching.py
greschd: Ok, I've fixed the ``type_check`` instances.

Parsing happens only once because the ``_with_config`` decorator only parses it if ``_CONFIG`` is empty. 

We probably don't need to handle the "`*` in the entrypoint name" case, right? The only limitation is that there's no way to put a literal / escaped `*` into your configuration pattern, or it will think it's a wildcard.

sphuber: Hmm yeah, sub ideal. But I think it is ok, because even though technically allowed in an entry point name, I don't think anyone will and we can add a section to the docs and registry that people should not use it

greschd: Well, technically, we could just replace all instances of where we use ``*`` to look for wildcards with the ``(^|[^\\])(\*)`` regex (literal `*`, preceded either by start of string or _not_ ``\``).

Might as well do things right.. but this is about as unlikely as it gets to ever be used.

greschd: But of course, then the rabbit hole really opens up.. ``foo\\*bar`` is an escaped ``\`` followed by an _unescaped_ wildcard. There's only so much you can do before implementing a full fledged grammar. I think it's fine to leave as is.

sphuber: yeah, that way madness lies. So that means this beast is ready to go, right?

greschd: yup, good to go for me

PR #3787, opened 2020-02-21
Avoid deadlocks when retrieving stdout/stderr via SSH

### LINKED ISSUE #1525: SSH transport hangs for big stdout content
If a command run through the ssh transport has a very big output (tested with a SGE plugin that, with a few thousands queued jobs, produces an XML output of ~2.5 MB), the SSH hangs forever.
Apparently, this is now mentioned also in the paramiko documentation, see the warning at http://docs.paramiko.org/en/2.4/api/channel.html#paramiko.channel.Channel.recv_exit_status.

A fix that seems to work (tested and solved the issue, giving correct content of the stdout) is the following:
```udiff
diff --git a/aiida/transport/plugins/ssh.py b/aiida/transport/plugins/ssh.py
index 383535408..d02eb8dc3 100644
--- a/aiida/transport/plugins/ssh.py
+++ b/aiida/transport/plugins/ssh.py
@@ -1290,7 +1290,7 @@ class SshTransport(aiida.transport.Transport):
 
         return stdin, stdout, stderr, channel
 
-    def exec_command_wait(self, command, stdin=None, combine_stderr=False, bufsize=-1):
+    def exec_command_wait(self, command, stdin=None, combine_stderr=False, bufsize=-1, read_bufsize=1024):
         """
         Executes the specified command and waits for it to finish.
 
@@ -1299,13 +1299,17 @@ class SshTransport(aiida.transport.Transport):
                    file-like object.
         :param combine_stderr: (optional, default=False) see docstring of
                    self._exec_command_internal()
-        :param bufsize: same meaning of paramiko.
+        :param bufsize: same meaning of paramiko. It is used as the bufsize for the channel files.
+        :param read_bufsize: size in bytes for reading/receiving data from stdout using
+            the channel recv method. This is needed to avoid hangs when there is a very
+            long output (see the warning here: 
+            http://docs.paramiko.org/en/2.4/api/channel.html#paramiko.channel.Channel.recv_exit_status).
+            It must be a positive number, if too big it will hang for big files, if too
+            small it can give performance issues.
 
         :return: a tuple with (return_value, stdout, stderr) where stdout and stderr
             are strings.
         """
-        # TODO: To see if like this it works or hangs because of buffer problems.
-
         ssh_stdin, stdout, stderr, channel = self._exec_command_internal(command, combine_stderr, bufsize=bufsize)
 
         if stdin is not None:
@@ -1325,11 +1329,23 @@ class SshTransport(aiida.transport.Transport):
         ssh_stdin.flush()
         ssh_stdin.channel.shutdown_write()
 
+        # I need to start fetching data in chunks (here I use a )
+        output_text_list = []
+        while True:
+            this_buf = channel.recv(read_bufsize)
+            if not this_buf:
+                # Length zero: it means the channel has closed
+                break
+            output_text_list.append(this_buf)
+
         # I get the return code (blocking)
         retval = channel.recv_exit_status()
 
         # needs to be after 'recv_exit_status', otherwise it might hang
-        output_text = stdout.read()
+        output_text = "".join(output_text_list)
+
+        # TODO: To see if like this it works or hangs because of buffer problems
+        #       and the same thing needs to be done as for stdout       
         stderr_text = stderr.read()
 
         return retval, output_text, stderr_text
```

Probably it needs a bit of testing, but then it can be integrated in the plugin. 
Thought is required to check if the same can happen also with big stderrs (actually I think it would, and it is potentially easy to test probably - just `cat` a big file) but I'm not sure how to use `recv` also for stderr, and how to interleave them correctly.

giovannipizzi:
Fixes #1525

The simple shift of the line `retval = channel.recv_exit_status()`
below `stdout.read()` and `stderr.read()`, while partially improving
the situation, is not enough (see discussion in #1525).
This instead reads in chunks both stdout and stderr, alternating between them.
Tests (added) show that this does not hang anymore for large files.

### DIFF
--- a/aiida/transports/plugins/local.py
+++ b/aiida/transports/plugins/local.py
@@ -759,27 +759,45 @@ class LocalTransport(Transport):
 
         return proc.stdin, proc.stdout, proc.stderr, proc
 
-    def exec_command_wait(self, command, **kwargs):
+    def exec_command_wait_bytes(self, command, stdin=None, **kwargs):
         """
         Executes the specified command and waits for it to finish.
 
         :param command: the command to execute
 
         :return: a tuple with (return_value, stdout, stderr) where stdout and
-                 stderr are strings.
+                 stderr are bytes.
         """
-        stdin = kwargs.get('stdin')
         local_stdin, _, _, local_proc = self._exec_command_internal(command)
 
         if stdin is not None:
+            # Implicitly assume that the desired encoding is 'utf-8' if I receive a string.
+            # Also, if I get a StringIO, I jsut read it all in memory and put it into a BytesIO.
+            # Clearly not memory effective - in this case do not use a StringIO, but pass directly a BytesIO
+            # that will be read line by line, if you have a huge stdin and care about memory usage.
             if isinstance(stdin, str):
-                filelike_stdin = io.StringIO(stdin)
+                filelike_stdin = io.BytesIO(stdin.encode('utf-8'))
+            elif isinstance(stdin, bytes):
+                filelike_stdin = io.BytesIO(stdin)
+            elif isinstance(stdin, io.TextIOBase):
+
+                def line_encoder(iterator, encoding='utf-8'):
+                    """Encode the iterator item by item (i.e., line by line).
+
+                    This only wraps iterating over it and not all other methods, but it's enough for its
+                    use below."""
+                    for line in iterator:
+                        yield line.encode(encoding)
+
+                filelike_stdin = line_encoder(stdin)
             else:
+                if not isinstance(stdin, io.BufferedIOBase):
+                    raise ValueError('You can only pass strings, bytes, BytesIO or StringIO objects')
                 filelike_stdin = stdin
 
             try:
-                for line in filelike_stdin.readlines():
-                    local_stdin.write(line.encode('utf-8'))  # the Popen.stdin/out/err are byte streams
+                for line in filelike_stdin:
+                    local_stdin.write(line)  # the Popen.stdin/out/err are byte streams
             except AttributeError:
                 raise ValueError('stdin can only be either a string or a file-like object!')
         else:
@@ -790,7 +808,7 @@ class LocalTransport(Transport):
 
         retval = local_proc.returncode
 
-        return retval, output_text.decode('utf-8'), stderr_text.decode('utf-8')
+        return retval, output_text, stderr_text
 
     def gotocomputer_command(self, remotedir):
         """
--- a/aiida/transports/plugins/local.py
+++ b/aiida/transports/plugins/local.py
@@ -765,14 +765,14 @@ class LocalTransport(Transport):
 
         :param command: the command to execute
 
-        :return: a tuple with (return_value, stdout, stderr) where stdout and
-                 stderr are bytes.
+        :return: a tuple with (return_value, stdout, stderr) where stdout and stderr
+            are both bytes and the return_value is an int.
         """
         local_stdin, _, _, local_proc = self._exec_command_internal(command)
 
         if stdin is not None:
             # Implicitly assume that the desired encoding is 'utf-8' if I receive a string.
-            # Also, if I get a StringIO, I jsut read it all in memory and put it into a BytesIO.
+            # Also, if I get a StringIO, I just read it all in memory and put it into a BytesIO.
             # Clearly not memory effective - in this case do not use a StringIO, but pass directly a BytesIO
             # that will be read line by line, if you have a huge stdin and care about memory usage.
             if isinstance(stdin, str):
@@ -790,10 +790,10 @@ class LocalTransport(Transport):
                         yield line.encode(encoding)
 
                 filelike_stdin = line_encoder(stdin)
-            else:
-                if not isinstance(stdin, io.BufferedIOBase):
-                    raise ValueError('You can only pass strings, bytes, BytesIO or StringIO objects')
+            elif isinstance(stdin, io.BufferedIOBase):
                 filelike_stdin = stdin
+            else:
+                raise ValueError('You can only pass strings, bytes, BytesIO or StringIO objects')
 
             try:
                 for line in filelike_stdin:
--- a/aiida/transports/plugins/ssh.py
+++ b/aiida/transports/plugins/ssh.py
@@ -1297,6 +1297,9 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         :return: a tuple with (return_value, stdout, stderr) where stdout and stderr
             are strings.
         """
+        import socket
+        import time
+
         ssh_stdin, stdout, stderr, channel = self._exec_command_internal(command, combine_stderr, bufsize=bufsize)
 
         if stdin is not None:
@@ -1316,14 +1319,68 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         ssh_stdin.flush()
         ssh_stdin.channel.shutdown_write()
 
+        # Now I get the output
+        stdout_bytes = []
+        stderr_bytes = []
+        # 10kB buffer (note that this should be smaller than the window size of paramiko)
+        # Also, apparently if the data is coming slowly, the read() command will not unlock even for
+        # times much larger than the timeout. Therefore we don't want to have very large buffers otherwise
+        # you risk that a lot of output is sent to both stdout and stderr, and stderr goes beyond the
+        # winodw size and blocks.
+        bufsize = 10 * 1024
+
+        # Set a small timeout on the channels, so that if we get data from both
+        # stderr and stdout, and the connection is slow, we interleave the receive and don't hang
+        stdout.channel.settimeout(0.1)
+        stderr.channel.settimeout(0.1)  # Maybe redundant, as this could be the same channel.
+
+        while True:
+            chunk_exists = False
+
+            if stdout.channel.recv_ready():  # True means that the next .read call will at least receive 1 byte
+                chunk_exists = True
+                try:
+                    piece = stdout.read(bufsize)
+                    stdout_bytes.append(piece)
+                except socket.timeout:
+                    # There was a timeout: I continue as there should still be data
+                    pass
+
+            if stderr.channel.recv_stderr_ready():  # True means that the next .read call will at least receive 1 byte
+                chunk_exists = True
+                try:
+                    piece = stderr.read(bufsize)
+                    stderr_bytes.append(piece)
+                except socket.timeout:
+                    # There was a timeout: I continue as there should still be data
+                    pass
+
+            # If chunk_exists, there is data (either already read and put in the std*_bytes lists, or
+            # still in the buffer because of a timeout). I need to loop.
+            # Otherwise, there is no data in the buffers, and I enter this block.
+            if not chunk_exists:
+                # Both channels have no data in the buffer
+                if channel.exit_status_ready():
+                    # The remote execution is over
+
+                    # I think that in some corner cases there might still be some data,
+                    # in case the data arrived between the previous calls and this check.
+                    # So we do a final read. Since the execution is over, I think all data is in the buffers,
+                    # so we can just read the whole buffer without loops
+                    stdout_bytes.append(stdout.read())
+                    stderr_bytes.append(stderr.read())
+                    # And we go out of the `while True` loop
+                    break
+                # The exit status is not ready:
+                # I just put a small sleep to avoid infinite fast loops when data
+                # is not available on a slow connection, and loop
+                time.sleep(0.05)
+
         # I get the return code (blocking)
+        # However, if I am here, the exit status is ready so this should be returning very quickly
         retval = channel.recv_exit_status()
 
-        # needs to be after 'recv_exit_status', otherwise it might hang
-        output_text = stdout.read().decode('utf-8')
-        stderr_text = stderr.read().decode('utf-8')
-
-        return retval, output_text, stderr_text
+        return (retval, b''.join(stdout_bytes).decode('utf-8'), b''.join(stderr_bytes).decode('utf-8'))
 
     def gotocomputer_command(self, remotedir):
         """
--- a/aiida/transports/plugins/ssh.py
+++ b/aiida/transports/plugins/ssh.py
@@ -1295,7 +1295,7 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         :param bufsize: same meaning of paramiko.
 
         :return: a tuple with (return_value, stdout, stderr) where stdout and stderr
-            are both bytes.
+            are both bytes and the return_value is an int.
         """
         import socket
         import time
@@ -1307,19 +1307,15 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
                 filelike_stdin = io.StringIO(stdin)
             elif isinstance(stdin, bytes):
                 filelike_stdin = io.BytesIO(stdin)
-            else:
+            elif isinstance(stdin, (io.BufferedIOBase, io.TextIOBase)):
                 # It seems both StringIO and BytesIO work correctly when doing ssh_stdin.write(line)?
                 # (The ChannelFile is opened with mode 'b', but until now it always has been a StringIO)
-                if not isinstance(stdin, (io.BufferedIOBase, io.TextIOBase)):
-                    raise ValueError('You can only pass strings, bytes, BytesIO or StringIO objects')
-
                 filelike_stdin = stdin
+            else:
+                raise ValueError('You can only pass strings, bytes, BytesIO or StringIO objects')
 
-            try:
-                for line in filelike_stdin:
-                    ssh_stdin.write(line)
-            except AttributeError:
-                raise ValueError('stdin can only be either a string, bytes or a file-like object!')
+            for line in filelike_stdin:
+                ssh_stdin.write(line)
 
         # I flush and close them anyway; important to call shutdown_write
         # to avoid hangouts
@@ -1329,7 +1325,7 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         # Now I get the output
         stdout_bytes = []
         stderr_bytes = []
-        # 10kB buffer (note that this should be smaller than the window size of paramiko)
+        # 100kB buffer (note that this should be smaller than the window size of paramiko)
         # Also, apparently if the data is coming slowly, the read() command will not unlock even for
         # times much larger than the timeout. Therefore we don't want to have very large buffers otherwise
         # you risk that a lot of output is sent to both stdout and stderr, and stderr goes beyond the
--- a/aiida/transports/plugins/ssh.py
+++ b/aiida/transports/plugins/ssh.py
@@ -677,7 +677,7 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
 
         command = f'{rm_exe} {rm_flags} {escape_for_bash(path)}'
 
-        retval, stdout, stderr = self.exec_command_wait(command)
+        retval, stdout, stderr = self.exec_command_wait_bytes(command)
 
         if retval == 0:
             if stderr.strip():
@@ -1133,7 +1133,7 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         # to simplify writing the above copy function
         command = f'{cp_exe} {cp_flags} {escape_for_bash(src)} {escape_for_bash(dst)}'
 
-        retval, stdout, stderr = self.exec_command_wait(command)
+        retval, stdout, stderr = self.exec_command_wait_bytes(command)
 
         if retval == 0:
             if stderr.strip():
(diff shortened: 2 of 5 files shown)


ltalirz:
Just one minor comment: since this is touching the ssh connection things, it might be good to test this in actual production runs before merging.

giovannipizzi:
Agreed - this was also my comment I gave in person to @sphuber - it this breaks, this break most of the AiiDA functionality :-D

But also if someone has the willingness to think if the logic is correct, this would be already great to at least have the feeling that we know why it is supposed to work (I am quite convinced that it's ok, but I might be missing something)

aiida/transports/plugins/ssh.py
```diff
@@ -1255,14 +1258,68 @@ def exec_command_wait(self, command, stdin=None, combine_stderr=False, bufsize=-
...
+
+        while True:
+            chunk_exists = False
+
+            if stdout.channel.recv_ready():  # True means that the next .read call will at least receive 1 byte
+                chunk_exists = True
+                try:
+                    piece = stdout.read(bufsize)
+                    stdout_bytes.append(piece)
+                except socket.timeout:
```
greschd: Is there any possibility that `socket.timeout` will be raised while `std*.channel.recv_stderr_ready()` keeps returning `True`? That could be an infinite loop.

ConradJohnston (dismissed):
Logic seems to check out, assuming that the Paramiko methods are honest and we find out the answer to @greschd's question. 

Also, did you benchmark the speed of this  approach for small stdout/stderr?

aiida/transports/plugins/ssh.py
giovannipizzi: From the [current docs](http://docs.paramiko.org/en/stable/api/channel.html):
- The documentation of `recv_ready` says that it returns True if a `recv` call on this channel would immediately return at least one byte; False otherwise.
- The documentation of `recv` says that it raises socket.timeout if no data is ready before the timeout set by settimeout.

What I gather is one of the two options:
- if a timeout is reached when only `nbytes < bufsize` bytes are received (and the channel is still open), just return those `nbytes>0`
- if a timeout is reached when only `nbytes < bufsize` bytes are received (and the channel is still open), raise. The bytes will be available in the next `recv` call if more arrive in the buffer and they go > `bufsize`, or if the channel is closed.

In both cases, this should not give an infinite loop.

Actually, from what I remember from my tests while implementing, there could even be a different possibility: I remember there was one case where no timeout was being raised, but still the data was arriving *very* slowly (with the call blocking on `recv` for much longer than the timeout, i.e., seconds, even if the bufsize was small). 
I am not sure how to interpret this. Maybe there is some kind of heartbeat in the connection, or maybe bytes were sent one at a time very slowly? However, I am not even sure I would be able to reproduce this now...

giovannipizzi:
> Also, did you benchmark the speed of this approach for small stdout/stderr?

From a quick benchmark on small files, the time seemed to be the same (see #3787: 10kB file transfer: 0.53s [new] vs 0.53s [old])

sphuber:
> > Also, did you benchmark the speed of this approach for small stdout/stderr?
> 
> From a quick benchmark on small files, the time seemed to be the same (see #3787: 10kB file transfer: 0.53s [new] vs 0.53s [old])

What happened to the big difference we were seeing originally, where the new buffer switching method was even a lot faster than the old naive implementation. Was that a measuring error?

giovannipizzi:
This is still there, again, see issue (e.g. `16MB file transfer: 1.2s vs 16s`)

giovannipizzi:
Here is a code to test out if it is working for you (remember of course to check out this branch):
```python
def get_file_md5(computer_name, remote_file_name):
    """
    Given an AiiDA computer name and a remote file name, returns the MD5 of its content as
    a string, returned

    Note that there is an issue - the current API of exec_command_wait returns unicode
    strings, we should probably change it (but it's backward incompatible?) to return bytes instead?
    The command below would fail if e.g. you pass a PDF file.
    """
    import hashlib
    from aiida import orm
    from aiida.common.escaping import escape_for_bash

    computer = orm.Computer.get(name=computer_name)
    transport = computer.get_transport()

    # this opens the SSH connection, for SSH transports
    with transport:
        retcode, stdout, stderr = transport.exec_command_wait("cat '{}'".format(escape_for_bash(remote_file_name)))
    assert retcode == 0
    assert stderr == ''
    return str(hashlib.md5(stdout.encode('utf-8')).hexdigest())

if __name__ == "__main__":
    import sys

    print(get_file_md5(sys.argv[1], sys.argv[2]))
```

Save it into a file e.g. `check_ssh.py`, then here is an example result in my case:
```
$ verdi run check_ssh.py daint rur.out
5c17d1a5ef891343517c19fd5841efe1

$ ssh daint.cscs.ch md5sum rur.out
5c17d1a5ef891343517c19fd5841efe1  rur.out
```
(note that in the first case you need to pass the AiiDA computer name).

Great if you could perform some tests on some SSH remotes you have, on large *text* (non-binary, see below) files.

**IMPORTANT** (see also comment in the function above). Currently we have the `exec_command_wait` return a string rather than bytes. I think this is a problem, e.g. if you `cat` a PDF file. See discussion in #3814 (so we don't clutter this PR with the discussion on bytes vs. strings). However note that if you test the script above, you can test on big *text* files, but it will probably crash on big binary files.

sphuber:
I have run one of my benchmark suites (558 `PwBaseWorkChains`) on this branch and I haven't noticed any problems out of the ordinary. There were some failed calculations, but those seem to have just been failures on the remote. Everything seems retrieved just fine on AiiDA's side as far as I have been able to check. The retrieved files may not have been particularly big so it may never have triggered the problem that this PR is supposed to fix, but in any case, this would be a relatively common use-case of high-throughput'ish use of AiiDA.

giovannipizzi:
ToDo: update this PR according to the comments in #3814 

Also, as a comment to @sphuber test - we should stress test in particular the part of reading from stdout/stderr; most of the AiiDA submissions will actually transfer files and this part is not touched by this PR.

giovannipizzi:
I added you as a collaborator on my fork - once you accept it, you can just push a new commit in my branch

ramirezfranciscof:
@giovannipizzi uhm, so, who do you mean by "you"?

giovannipizzi:
@ramirezfranciscof ahah I had multiple tabs open and I put the comment in wrong tab/issue :-) I had been wondering where this comment had gone, I knew I had written it but couldn't find it anymore

dev-zero:
maybe as a note for the future: for Python 3.6+ there's [an alternative to Paramiko: AsyncSSH](https://asyncssh.readthedocs.io/) which may tie in better to the current architecture of AiiDA, and avoid some of problems encountered with Paramiko (while probably  causing others, as usual).

aiida/transports/plugins/local.py
```diff
@@ -759,27 +759,34 @@ def _exec_command_internal(self, command, **kwargs):  # pylint: disable=unused-a
...
+                filelike_stdin = io.BytesIO(stdin)
+            elif isinstance(stdin, io.TextIOBase):
+                filelike_stdin = io.BytesIO(stdin.read().encode('utf-8'))
             else:
                 filelike_stdin = stdin
 
             try:
                 for line in filelike_stdin.readlines():
-                    local_stdin.write(line.encode('utf-8'))  # the Popen.stdin/out/err are byte streams
+                    local_stdin.write(line)  # the Popen.stdin/out/err are byte streams
```
dev-zero, line 789: The `for line in filelike_stdin.readlines():` should probably be a `for line filelike_stdin:` to avoid another copy of the content as list of strings.

aiida/transports/plugins/local.py
```diff
@@ -759,27 +759,34 @@ def _exec_command_internal(self, command, **kwargs):  # pylint: disable=unused-a
...
+            # Also, if I get a StringIO, I jsut read it all in memory and put it into a BytesIO.
+            # Clearly not memory effective - in this case do not use a StringIO, but pass directly a BytesIO
+            # that will be read line by line, if you have a huge stdin and care about memory usage.
             if isinstance(stdin, str):
-                filelike_stdin = io.StringIO(stdin)
+                filelike_stdin = io.BytesIO(stdin.encode('utf-8'))
+            elif isinstance(stdin, bytes):
+                filelike_stdin = io.BytesIO(stdin)
+            elif isinstance(stdin, io.TextIOBase):
+                filelike_stdin = io.BytesIO(stdin.read().encode('utf-8'))
```
dev-zero, line 783: maybe use something like this to avoid reading the complete thing in and blocking? Or is that intentional?

```
def line_encoder(iter, encoding='utf-8'):
  for line in iter:
    yield line.encode(encoding)
filelike_stdin = line_encoder(stdin)
```

aiida/transports/plugins/ssh.py
```diff
@@ -1294,35 +1294,95 @@ def exec_command_wait(self, command, stdin=None, combine_stderr=False, bufsize=-
...
+                # Both channels have no data in the buffer
+                if channel.exit_status_ready():
+                    # The remote execution is over
+
+                    # I think that in some corner cases there might still be some data,
+                    # in case the data arrived between the previous calls and this check.
+                    # So we do a final read. Since the execution is over, I think all data is in the buffers,
+                    # so we can just read the whole buffer without loops
+                    stdout_bytes.append(stdout.read())
+                    stderr_bytes.append(stderr.read())
```
dev-zero, line 1373: Is it possible that (for example on heavy load of the client machine or in a memory pressure situation/when swapping) that the channel timeout is triggered here? It's a small buffer and everything should be local and closed, but might be worth to do another try/except to avoid silent content corruption.

aiida/transports/plugins/local.py
giovannipizzi: Thanks!

aiida/transports/plugins/local.py
giovannipizzi: Fixed in the new (amended) commit

aiida/transports/plugins/ssh.py
giovannipizzi: Thanks for the review! By a look at paramiko's code, I don't think this should happen; [read()](https://github.com/paramiko/paramiko/blob/5cf145ffd3d7cd4d96a511376267a7f27cce5eed/paramiko/file.py#L172), as you say, should only use the local buffer in memory and not access the `recv` method (called within `self._read` of ChannelFile), that in turn is the one that could raise a timeout error.
Also, I fear that if I try to catch the exception and `pass`, I risk to enter an infinite loop. In the current code at least this function would raise in your scenario, and people will notice and we'll have a bug report.

As a note, at least a few people like @adegomme and @ConradJohnston have been testing this implementation and as far as I know it works fine with them, so I would be tempted not to make changes if we're not 100% sure, to avoid to introduce even more subtle bugs.

giovannipizzi: (I'm resolving this conversation but please reopen if you instead believe I should change the code, I'm not 100% sure how, though)

giovannipizzi:
While I'm doing the final tests on this, let me ping @pzarabadip (who implemented the [metavo plugin](https://aiidateam.github.io/aiida-registry/plugins/aiida-metavo-scheduler.html)) and @adegomme (who implemented the [sshonly plugin](https://aiidateam.github.io/aiida-registry/plugins/sshonly.html)).
Great if you could read the comment here and give us your feedback.

*Note: These are the only two transport plugins currently on the registry; after we agree, I'll also ping everyone on the AiiDA mailing list for people who might have developed a plugin but not contributed it.*

This is to discuss the amount of work to adapt your plugin, and the backward-compatibility of it.

1. If you don't do anything, your `exec_command_wait` implementation will override the one that is now defined in the `Transport base class, so it should *almost* work (except that now in a couple of places, in particular in `whoami`, I call the new `_bytes` method, so that would fail, and that is called often by AiiDA, IIRC.
2. The simple migration path if you only want to support new versions of AiiDA is just to rename your method from `exec_command_wait` to `exec_command_wait_bytes` and remove string decoding at the end. However, you make your plugin incompatible with AiiDA versions < 1.6.
3. If you want to support both, beside point 2, you'll need also to copy-paste the implementation of the new `exec_command_wait` (now in the main Transport class) also in your plugin, so
   - in AiiDA < 1.6, this is now defined and everything works as before
   - in AiiDA >= 1.6, you are overriding AiiDA's `exec_command_wait` function, but the implementation is the same so it shouldn't matter.

My suggestion is that you could implement point 3 "immediately" (i.e. as soon as this PR is merged, just in case there are some fixes in the `exec_command_wait` code). After that, when 1.6 is released, you have to do nothing. Only remaining thing is that, when you want to drop support for <1.6, it's better if you remove the copy-pasted implementation of `exec_command_wait` in case this changes or gets fixed in the future (and you should monitor if we apply fixes to that function in the future, so you can port those changes in your repo).

Does this seem correct and OK to do? Because we're deciding if this type of "backward-incompatibility" is ok to accept for v1.6 (users will not notice as long as you do the fix above).

giovannipizzi:
Ok, I optimised the parameters to have a reasonably good throughput (I can go to 100MB/s over SSH on the local host, a bit slower than pure SSH+md5 but still reasonably good) - see also comments and commit message in my last commit.
I also checked (by modifying a bit the script provided [in the comment above](https://github.com/aiidateam/aiida-core/pull/3787#issuecomment-592490003)) that a similar speed is obtained (and correct results are obtained) if the data is sent over the stderr rather than stdout.

Therefore, as soon as I get some feedback from @pzarabadip and @adegomme, I can convert this draft into a PR

giovannipizzi:
As a further piece of info, I tried transferring actual data over ssh between my workstation in the office and my laptop at home.
I turned off the wifi and connected by cable. I started the connection, then physically disconnected the cable for ~6-7s, then reconnected it. 
I tried this a few times and it worked fine, at the end I could get the correct MD5 of a ~330MB file.

This below is the measured network speed; the 'flat' parts are my combined network speed of ~12MB/s; the "hole" in the middle is the moment where I physically disconnected the cable (every data point is 1s):
data

pzarabadip:
Hi @giovannipizzi thanks for this. It seems quite okay with me. Unfortunately, my access to metavo machines has finished and cannot test it. I'm not aware of any other user to be honest to probably face with backward compatibility. I might get access again in future and then, I'll modify the plugin to reflect these changes (of course, if it is not to be done now).

giovannipizzi:
Thanks @pzarabadip! I suggest that in that case either you apply the fix anyway (and open an issue to say to remove the copy-pasted function at some point), or just open an issue pointing here, but add a limit on aiida-core < 1.6 already (with a comment on why, pointing to your issue). This to avoid that in the future, when someone will get back to the code, it will take quite some time to figure out this subtle change.

pzarabadip:
Thanks @giovannipizzi . I just did limit the version and opened an issue for future traceback. Cheers.

tests/transports/test_all_plugins.py
```diff
@@ -1358,10 +1362,135 @@ def test_exec_with_stdin_filelike(self, custom_transport):
             self.assertEqual(stdout, test_string)
             self.assertEqual(stderr, '')
 
+    @run_for_all_plugins
+    def test_exec_with_stdin_filelike_bytes(self, custom_transport):
+        """Test command execution with a stdin from filelike of type bytes.
+
+        I test directly the exec_command_wait_bytes function; I also pass some non-unicode
+        bytes to check that there is no internal implicit encoding/decoding in the code.
```
ramirezfranciscof, line 1370: Maybe nitpicky, but maybe there is another way to express this? I feel like "checking that there is no internal implicit encoding/decoding" is not really possible (I understand one can only unit test for behavior, what the method does internally is up to them). This also goes for the other test.

I guess the behaviour here is that we want to check that bytes that are non unicode are still accepted, and maybe add a comment that this should be true for any encoding but we can't test for all of them (and also I guess some of these can accept any byte?).

tests/transports/test_all_plugins.py
```diff
@@ -1358,10 +1362,135 @@ def test_exec_with_stdin_filelike(self, custom_transport):
...
+            location = trans.normalize(os.path.join('/', 'tmp'))
+            directory = 'temp_dir_test_transfer_big_stdout'
+            trans.chdir(location)
+
+            self.assertEqual(location, trans.getcwd())
+            while trans.isdir(directory):
+                # I append a random letter/number until it is unique
+                directory += random.choice(string.ascii_uppercase + string.digits)
+            trans.mkdir(directory)
+            trans.chdir(directory)
```
ramirezfranciscof, line 1433: This feels a bit zigzaggy, it took me a while to understand what was happening. Perhaps this is more straightforward to understand:

```python
            location = trans.normalize(os.path.join('/', 'tmp'))
            trans.chdir(location)
            self.assertEqual(location, trans.getcwd())

            directory = 'temp_dir_test_transfer_big_stdout
            while trans.isdir(directory):
                # I append a random letter/number until it is unique
                directory += random.choice(string.ascii_uppercase + string.digits)
            trans.mkdir(directory)
            trans.chdir(directory)
```

aiida/transports/plugins/ssh.py
```diff
@@ -1294,35 +1294,105 @@ def exec_command_wait(self, command, stdin=None, combine_stderr=False, bufsize=-
...
                 filelike_stdin = io.StringIO(stdin)
+            elif isinstance(stdin, bytes):
+                filelike_stdin = io.BytesIO(stdin)
             else:
+                # It seems both StringIO and BytesIO work correctly when doing ssh_stdin.write(line)?
+                # (The ChannelFile is opened with mode 'b', but until now it always has been a StringIO)
+                if not isinstance(stdin, (io.BufferedIOBase, io.TextIOBase)):
+                    raise ValueError('You can only pass strings, bytes, BytesIO or StringIO objects')
+
                 filelike_stdin = stdin
```
ramirezfranciscof, line 1315: Idem to the case in the local plugin.

aiida/transports/plugins/ssh.py
```diff
@@ -1294,35 +1294,105 @@ def exec_command_wait(self, command, stdin=None, combine_stderr=False, bufsize=-
...
 
             try:
-                for line in filelike_stdin.readlines():
+                for line in filelike_stdin:
                     ssh_stdin.write(line)
             except AttributeError:
-                raise ValueError('stdin can only be either a string of a file-like object!')
+                raise ValueError('stdin can only be either a string, bytes or a file-like object!')
```
ramirezfranciscof, line 1321: This `AttributeError` is actually raising because of `ssh_stdin`, right? Why you redirect to `ValueError` blaming `stdin`? I'm also wondering how would one test this (because codecov is complaining): the problem being caught is `self._exec_command_internal` returning an object without a `write` method, correct?

ramirezfranciscof (commented):
Very interesting work! I can probably not provide as deep technical reviews as you have already received, but I detected a couple of typos and  had a couple of comments that might be of some use.

adegomme:
Hi,
Sorry about the long delay in my answer. The reason we had our own version was actually to integrate this fix early, as it was needed. As #4363 was already merged, this override should now become redundant and be removed. So I think all is needed is to make the next version of the plugin depend on aiida>1.6 and remove it altogether.
I will try to run this this afternoon and confirm it works on this branch.

adegomme:
I can confirm that it seems to be working fine. To avoid breaking support, I will keep the ssh.py file for now, but add a version check at import, to use the aiida_core's one on for aiida > 1.6. Thanks a lot.

aiida/transports/plugins/ssh.py
giovannipizzi: You are right - this was needed before, but now we are checking the instance of all possible cases otherwise we already raise before, so this check (try/except) is not needed anymore. I'll remove it.

giovannipizzi:
Thanks @ramirezfranciscof and sorry for the long time it took! If tests pass, I think this is ready to be reviewed!

ramirezfranciscof:
Great, thanks!

> Thanks @ramirezfranciscof and sorry for the long time it took! If tests pass, I think this is ready to be reviewed!

@giovannipizzi you mean merged? Bah, unless you wanted someone else in particular to take a look at it.

Also there is the fact that the branch is a but outdated now. I'll try to merge develop into this, hopefully it will manage to do it automatically, but if it breaks then I think you might have to rebase or merge locally and manually fix any problem.

giovannipizzi:
Yeah, I meant just a final check + merge. I had rebased it, so the changes are minimal (and the merge seems to have worked).
I'm converting to a proper PR, feel free to merge!

ramirezfranciscof (approved):
All good to go then @giovannipizzi ! I don't know if you can/want to write a more descriptive commit message for this, otherwise I can merge with the message on the OP.

giovannipizzi:
Commit message clarified, and added [migration notes to the wiki](https://github.com/aiidateam/aiida-core/wiki/AiiDA-2.0-plugin-migration-guide#transport-plugins)

@pzarabadip @adegomme as discussed above, this will affect your plugins so you might want to take note of this and decide if/when to take action.

pzarabadip:
Thanks @giovannipizzi. I will adopt the plugins to reflect these changes. Cheers.

Issue #3303, opened 2019-09-09
verdi code setup should warn when setting up a code with existing label


ltalirz:
Currently, `verdi code setup` will happily set up codes with the same label multiple times.
It is fine that this is allowed in principle, but in interactive mode, a user should be warned when doing this.

The simplest might be something like:
```
$ verdi code setup
Info: enter "?" for help
Info: enter "!" to ignore the default and set no value
Label: cp
Description []: asdf
Default calculation input pluginquantumespresso.pw  
Installed on target computer? [True]: 
Computer: localhost
Warning: Code 'cp@localhost' already exists.
Remote absolute path:
```
If we want to be more fancy, we could add a prompt
```
Choose a different label [Y/n]: 
```

ltalirz:
Along the same lines: when importing an archive file, it would make sense to relabel codes in the archive if a code of the same label already exists in the database (analogous to what is done for computers).
Otherwise, imports are likely to break scripts that rely on loading codes by label (which is a common practise).

mbercx:
Just out of curiosity: Would it not be better to disallow two codes with the same label altogether?

ltalirz:
> Would it not be better to disallow two codes with the same label altogether?

Once we implement the renaming for `verdi import`, we're almost there in practice.
One problem might be that codes are represented as nodes, so I guess one cannot easily enforce this at the database schema level (while computers are not).

ramirezfranciscof:
>  One problem might be that codes are represented as nodes, so I guess one cannot easily enforce uniqueness at the database schema level (while computers are not).

But we could enforce uniqueness at the level of the `node:code` subclass no? Forcing it to check the DB for existing labels before creating (init?) and/or when changing the label property.

greschd:
> Just out of curiosity: Would it not be better to disallow two codes with the same label altogether?

In my opinion the code label doesn't need to be unique. For context, when testing a code across variations of our infrastructure I would distinguish them only by the computer. Of course it would be possible to add a `-<computer_name>` to the code label, but that seems silly if there's already the `@<computer_name>` syntax.

ltalirz:
Yes, my comment seemed to imply that I was thinking of uniqueness at the code label only, but I agree that this discussion should focus on whether `code-label@computer-label` should be unique (not the code label alone, as @greschd rightfully points out)

janssenhenning:
Hello Aiida team,
[...]

PR #3788, opened 2020-02-21
verdi computer/code duplicate now reuses prepend_text and append_text

### LINKED ISSUE #3747: Duplication of codes and computers via 'verdi' does not duplicate prepend and append text
Using `verdi code duplicate` or `verdi computer duplicate` duplicates all the standard one-line fields, but the prepend and append texts are not pre-filled.

It would be a nice quality-of-life improvement if these too could be duplicated and then just modified when the editor is opened at the end the `verdi` command.  

Example use case - with QuantumESPRESSO one must setup many executables as codes, but on a cluster these are all typically loaded using a single module. The prepend text, for example, might be `module load qe-6.5.0` and is the same in every case, but is necessary to ensure that the linked libraries are available.

giovannipizzi:
This required adding a new ContextualDefaultOption class
to get a default callable that also receives the context.
Moreover, a but was fixed in the function that prompts
for the prepend/append_text, as if the file was not touched/modified,
instead of reusing the text provided in input, an empty text was used instead.

Fixes #3747


giovannipizzi:
@ConradJohnston could you check that this work as you expect? If this is the case, after adding tests this is ready to merge

ConradJohnston (dismissed):
Works great, thank you.

One curiosity - why does duplicate allow me to have identical labels for the code?
If one uses orm.Code.get_from_string(), it complains unless the code@computer label is unique.
I understand that this isn't the ID on which uniqueness is enforced, but given that there are convenience functions which do use this label, should we warn the user? Will make a separate issue if you agree that it's an oddity.

ltalirz:
> given that there are convenience functions which do use this label, should we warn the user? 

We should, and there already is an issue for it https://github.com/aiidateam/aiida-core/issues/3303

giovannipizzi:
Thanks @ConradJohnston !

Do you happen to have time to add a test, by chance? I'll be a bit busy in the next days and if we have tests we can merge this.

ConradJohnston:
@giovannipizzi , yes, I'll try to get looking at it as soon as possible.

ConradJohnston:
@giovannipizzi , have test coverage now. Can I push back to your branch?

giovannipizzi:
@ConradJohnston sure, thanks a lot! I just added you as a collaborator to my fork - not sure if this means you can push, in case you cannot I can set this after you accept the invite (GitHub might have changed this workflow recently? I remember I could set the role while inviting...)

ConradJohnston:
> @ConradJohnston sure, thanks a lot! I just added you as a collaborator to my fork - not sure if this means you can push, in case you cannot I can set this after you accept the invite (GitHub might have changed this workflow recently? I remember I could set the role while inviting...)

Hi Giovanni , I've pushed my change now, so we'll need another reviewer. 
Actually, the tests did already cover the duplication of these texts, so all I did was to ensure that the test code and computer objects have something other than an empty string for their prepend/append texts.

giovannipizzi:
Thanks a lot @ConradJohnston!
Indeed you are right, great (also better, it was an easier fix ;-) )

I cannot approve this myself - either you reprove it, or maybe @sphuber could approve and merge?

sphuber (approved):
Thanks guys

PR #3793, opened 2020-02-24
validate label string at code setup stage

unkcpz:
Prevent users from using '@' when naming the code when calling `verdi code setup`.
Not just make this validation in `verdi code relabel`


unkcpz:
It should be right now. yeah!

unkcpz:
And I search the whole package without finding `raise_error` being set, so I remove this parameter.  Hope it won't be a problem.

aiida/orm/nodes/data/code.py
```diff
@@ -131,21 +152,14 @@ def relabel(self, new_label, raise_error=True):
         if new_label.endswith(suffix):
             new_label = new_label[:-len(suffix)]
 
-        if '@' in new_label:
-            msg = "Code labels must not contain the '@' symbol"
-            if raise_error:
```
ltalirz: Ok to deprecate it - in this case please put in the docstring of the function something like
```
.. deprecated:: 1.2.0
            Will be removed in `v2.0.0`. Use `try/except` instead.
```

aiida/orm/nodes/data/code.py
```diff
@@ -131,21 +152,14 @@ def relabel(self, new_label, raise_error=True):
...
-                raise InputValidationError(msg)
-            else:
-                return InputValidationError(msg)
-
         self.label = new_label
 
     def get_description(self):
         """Return a string description of this Code instance.
```
ltalirz: While you're at it, feel free to add a similar deprecation warning for `get_description` - one should use the `.description` property instead.

aiida/orm/utils/builders/code.py
```diff
@@ -62,6 +62,8 @@ def new(self):
                 )
             )
 
+        # code.set_label(self._get_and_count('label', used))
```
ltalirz: leftover

aiida/orm/nodes/data/code.py
unkcpz: I'll remove the argument, it will make `label` setter complex otherwise.

ltalirz: What you've done here is fine (leave the argument for the moment but don't do anything with it).
Can you still put the deprecation warning about the `raise_error` argument into the docstring of `relabel`?

unkcpz:
done ;-)  thanks for reviewing @ltalirz

unkcpz:
the label validate is test in here https://github.com/aiidateam/aiida-core/blob/9fcb78a1100166a66d5d5560e6618d3e4f59b002/tests/cmdline/commands/test_code.py#L186
it call the `label` setter, so it sort of being tested I think.

ltalirz:
Right - it wasn't tested before but now it is. Great!

unkcpz:
In fact the validation only works when the whole code config is passed in. which means if user type `@` in the label interactively will see the exception at the end of config stage. 
But I think that needs to be sorted out separately.

aiida/orm/nodes/data/code.py
```diff
@@ -120,32 +120,51 @@ def full_label(self):
...
+
+    @label.setter
+    def label(self, value):
+        """Set the label.
+
+        :param value: the new value to set
+        """
+        if '@' in str(value):
+            msg = "Code labels must not contain the '@' symbol"
+            raise InputValidationError(msg)
```
sphuber: This should have been just a `ValuError`. The `InputValidationError` is intended for use in `CalcJob.prepare_for_submission`

aiida/orm/nodes/data/code.py
```diff
@@ -120,32 +120,51 @@ def full_label(self):
...
-                return InputValidationError(msg)
-
         self.label = new_label
 
     def get_description(self):
         """Return a string description of this Code instance.
 
         :return: string description of this Code instance
+
+        .. deprecated:: 1.2.0
```
sphuber: Why is this deprecated? Almost all of the ORM classes have `get_description`. If we are going for a property, we should try to be consistent and do this everywhere.

ltalirz: I would deprecate this for all classes but didn't want to ask jason to do this in this PR.
Sounds reasonable?

aiida/orm/nodes/data/code.py
ltalirz: Sorry I didn't see this - this code is just copy-pasted from where it was before.
There are a lot of places in aiida-core where this error is used that have nothing to do with prepare_for_submission:
https://github.com/aiidateam/aiida-core/blob/09cb6cb70df024143ac5a3720e4034b609536b02/aiida/orm/querybuilder.py#L141

Do all of these uses have to be replaced?

aiida/orm/nodes/data/code.py
sphuber: I am not sure if we really should replace `get_description` for `description`. The reason that both exist is that `description` is understood to return the value stored in the `description` column and `get_description` is some compound from various columns/attributes. We might be able to find another name for `get_description` to reduce confusion, but in any case I think it should be done in a consistent way across the code and not just in one place

aiida/orm/nodes/data/code.py
sphuber: I would say so yes, but that would be a breaking change. The `InputValidationError` is not a subclass of `ValueError` which is what one would expect for these kinds of errors. I will open an issue to address this for `2.0`

aiida/orm/nodes/data/code.py
ltalirz: Ok, I'll revert this deprecation at some point.

PR #3796, opened 2020-02-24
`QueryBuilder`: add support for `datetime.date` objects in filters

### LINKED ISSUE #3794: `QueryBuilder` does not support `datetime.date` types
Reproducible by the following:
```
import datetime
QueryBuilder().append(Node, filters={'ctime': {'<': datetime.date('2018-11-22')}}).all()
```
will raise the following:
```---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-bf56c15cbd41> in <module>()
----> 1 builder.all()

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/orm/querybuilder.py in all(self, batch_size)
   2202         :returns: a list of lists of all projected entities.
   2203         """
-> 2204         return list(self.iterall(batch_size=batch_size))
   2205 
   2206     def dict(self, batch_size=None):

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/orm/querybuilder.py in iterall(self, batch_size)
   2157         :returns: a generator of lists
   2158         """
-> 2159         query = self.get_query()
   2160 
   2161         for item in self._impl.iterall(query, batch_size, self._attrkeys_as_in_sql_result):

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/orm/querybuilder.py in get_query(self)
   2018         # whether the query is still valid
   2019 
-> 2020         queryhelp_hash = make_hash(self.get_json_compatible_queryhelp())
   2021         # if self._hash (which is None if this function has not been invoked
   2022         # and is a string (hash) if it has) is the same as the queryhelp

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/common/hashing.py in make_hash(object_to_hash, **kwargs)
    133     """
    134 
--> 135     hashes = _make_hash(object_to_hash, **kwargs)  # pylint: disable=assignment-from-no-return
    136 
    137     # use the Unlimited fanout hashing protocol outlined in

/home/aiida/.virtualenvs/aiida_3dd/lib/python3.7/functools.py in wrapper(*args, **kw)
    838                             '1 positional argument')
    839 
--> 840         return dispatch(args[0].__class__)(*args, **kw)
    841 
    842     funcname = getattr(func, '__name__', 'singledispatch function')

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/common/hashing.py in _(mapping, **kwargs)
    209             (k_digest + _make_hash(val, **kwargs)) for k_digest, val in sorted(hashed_key_mapping(), key=itemgetter(0))
    210         )
--> 211     ) + [_END_DIGEST]
    212 
    213 

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/common/hashing.py in <genexpr>(.0)
    207     return [_single_digest('dict(')] + list(
    208         chain.from_iterable(
--> 209             (k_digest + _make_hash(val, **kwargs)) for k_digest, val in sorted(hashed_key_mapping(), key=itemgetter(0))
    210         )
    211     ) + [_END_DIGEST]

/home/aiida/.virtualenvs/aiida_3dd/lib/python3.7/functools.py in wrapper(*args, **kw)
    838                             '1 positional argument')
    839 
--> 840         return dispatch(args[0].__class__)(*args, **kw)
    841 
    842     funcname = getattr(func, '__name__', 'singledispatch function')

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/common/hashing.py in _(mapping, **kwargs)
    209             (k_digest + _make_hash(val, **kwargs)) for k_digest, val in sorted(hashed_key_mapping(), key=itemgetter(0))
    210         )
--> 211     ) + [_END_DIGEST]
    212 
    213 

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/common/hashing.py in <genexpr>(.0)
    207     return [_single_digest('dict(')] + list(
    208         chain.from_iterable(
--> 209             (k_digest + _make_hash(val, **kwargs)) for k_digest, val in sorted(hashed_key_mapping(), key=itemgetter(0))
    210         )
    211     ) + [_END_DIGEST]

/home/aiida/.virtualenvs/aiida_3dd/lib/python3.7/functools.py in wrapper(*args, **kw)
    838                             '1 positional argument')
    839 
--> 840         return dispatch(args[0].__class__)(*args, **kw)
    841 
    842     funcname = getattr(func, '__name__', 'singledispatch function')

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/common/hashing.py in _(mapping, **kwargs)
    209             (k_digest + _make_hash(val, **kwargs)) for k_digest, val in sorted(hashed_key_mapping(), key=itemgetter(0))
    210         )
--> 211     ) + [_END_DIGEST]
    212 
    213 

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/common/hashing.py in <genexpr>(.0)
    207     return [_single_digest('dict(')] + list(
    208         chain.from_iterable(
--> 209             (k_digest + _make_hash(val, **kwargs)) for k_digest, val in sorted(hashed_key_mapping(), key=itemgetter(0))
    210         )
    211     ) + [_END_DIGEST]

/home/aiida/.virtualenvs/aiida_3dd/lib/python3.7/functools.py in wrapper(*args, **kw)
    838                             '1 positional argument')
    839 
--> 840         return dispatch(args[0].__class__)(*args, **kw)
    841 
    842     funcname = getattr(func, '__name__', 'singledispatch function')

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/common/hashing.py in _(mapping, **kwargs)
    209             (k_digest + _make_hash(val, **kwargs)) for k_digest, val in sorted(hashed_key_mapping(), key=itemgetter(0))
    210         )
--> 211     ) + [_END_DIGEST]
    212 
    213 

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/common/hashing.py in <genexpr>(.0)
    207     return [_single_digest('dict(')] + list(
    208         chain.from_iterable(
--> 209             (k_digest + _make_hash(val, **kwargs)) for k_digest, val in sorted(hashed_key_mapping(), key=itemgetter(0))
    210         )
    211     ) + [_END_DIGEST]

/home/aiida/.virtualenvs/aiida_3dd/lib/python3.7/functools.py in wrapper(*args, **kw)
    838                             '1 positional argument')
    839 
--> 840         return dispatch(args[0].__class__)(*args, **kw)
    841 
    842     funcname = getattr(func, '__name__', 'singledispatch function')

/home/aiida/code/aiida/env/3dd/aiida-core/aiida/common/hashing.py in _make_hash(object_to_hash, **_)
    154     28 byte integer, and only later converted to a string.
    155     """
--> 156     raise ValueError('Value of type {} cannot be hashed'.format(type(object_to_hash)))
    157 
    158 

ValueError: Value of type <class 'datetime.date'> cannot be hashed
```

sphuber:
Fixes #3794 

Each query is hashed for efficiency reasons, such that repeated queries
can be taken from a cache. This requires the builder instance and all
its attributes to be hashable. The filters can reference `datetime.date`
objects to express date and time conditions, but these were not
supported by the hashing implementation, unlike `datetime.datetime`
objects. Here we update the `aiida.common.hashing.make_hash` single
dispatch to also support `datetime.date` objects.

### DIFF
--- a/aiida/common/hashing.py
+++ b/aiida/common/hashing.py
@@ -8,6 +8,7 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Common password and hash generation functions."""
+import datetime
 import hashlib
 try:  # Python > 3.5
     from hashlib import blake2b
@@ -18,7 +19,6 @@ import random
 import time
 import uuid
 from collections import abc, OrderedDict
-from datetime import datetime
 from functools import singledispatch
 from itertools import chain
 from operator import itemgetter
@@ -234,15 +234,21 @@ def _(val, **kwargs):
     return [_single_digest('none')]
 
 
-@_make_hash.register(datetime)
+@_make_hash.register(datetime.datetime)
 def _(val, **kwargs):
     """hashes the little-endian rep of the float <epoch-seconds>.<subseconds>"""
     # see also https://stackoverflow.com/a/8778548 for an excellent elaboration
     if val.tzinfo is None or val.utcoffset() is None:
         val = val.replace(tzinfo=pytz.utc)
-    timestamp = val.timestamp()
 
+    timestamp = val.timestamp()
     return [_single_digest('datetime', float_to_text(timestamp, sig=AIIDA_FLOAT_PRECISION).encode('utf-8'))]
+
+
+@_make_hash.register(datetime.date)
+def _(val, **kwargs):
+    """Hashes the string representation in ISO format of the `datetime.date` object."""
+    return [_single_digest('date', val.isoformat().encode('utf-8'))]
 
 
 @_make_hash.register(uuid.UUID)
--- a/tests/orm/test_querybuilder.py
+++ b/tests/orm/test_querybuilder.py
@@ -24,6 +24,16 @@ class TestQueryBuilder(AiidaTestCase):
         self.clean_db()
         self.insert_data()
 
+    def test_date_filters_support(self):
+        """Verify that `datetime.date` is supported in filters."""
+        from datetime import datetime, date, timedelta
+
+        orm.Data(ctime=datetime.now() - timedelta(days=3)).store()
+        orm.Data(ctime=datetime.now() - timedelta(days=1)).store()
+
+        builder = orm.QueryBuilder().append(orm.Node, filters={'ctime': {'>': date.today() - timedelta(days=1)}})
+        self.assertEqual(builder.count(), 1)
+
     def test_ormclass_type_classification(self):
         """
         This tests the classifications of the QueryBuilder


sphuber:
Not a hundred percent sure about the hash implementation for `datetime.date`. It should not matter for how it is used for the `QueryBuilder` but if it is used for other purposes then it may be a problem if implemented incorrectly.

aiida/common/hashing.py
```diff
@@ -234,17 +234,35 @@ def _(val, **kwargs):
...
 
+@_make_hash.register(datetime.date)
+def _(val, **kwargs):
+    """hashes the little-endian rep of the float <epoch-seconds>.<subseconds>"""
+    import calendar
+    from aiida.manage.configuration import settings
+
+    # The method to get the timestamp from the date object depends on whether it is in UTC or a local timezeon
+    # See for an extended discussion https://stackoverflow.com/a/8778548
+    if getattr(settings, 'USE_TZ', False):
```
greschd: I'm not sure if TZ handling makes sense here. Without it, we could just use ``_single_digest('date', val.isoformat().encode('utf-8'))``.

In addition to the QueryBuilder, `_make_hash` is also used in the caching implementation. Since the `datetime.date` itself is a naive, timezone-unaware datastructure, I think the hashing should do the same. Both should be more or less trouble-free in most cases. But, if I were to import a `Node` containing a `datetime.date` somewhere (if that's even possible?) into a system using a different timezone, caching for that node would be broken.

sphuber: > Since the `datetime.date` itself is a naive, timezone-unaware datastructure, I think the hashing should do the same. Both should be more or less trouble-free in most cases. 

Fair enough

> But, if I were to import a `Node` containing a `datetime.date` somewhere (if that's even possible?) into a system using a different timezone, caching for that node would be broken.

When talking about `attributes` as to what a `Node` can contain, then it has to be serializable according to `aiida.orm.utils.node.clean_value` and neither `datetime.datetime` nor `datetime.date` objects are supported. The former used to be, but we removed that. Users are now required to serialize these themselves into a serializable type, such as a timestamp. The deserialization also has to be done manually by the caller. So I guess it is true, even for the caching mechanism this shouldn't matter because it should never get a `datetime.date` object but some string or float timestamp.

@giovannipizzi do you agree that date objects we can always treat as timezone naive and just take the isoformat string?

sphuber: Discussed in person with @giovannipizzi and he agrees that timezone for `date` object is meaningless and so just the ISO formatted string is fine

PR #3801, opened 2020-02-24
Remove unused and obsolete language extensions and utilities

sphuber:
The `aiida.common.lang` module hosts a variety of utilities that extend
the base python language. Many of these were backports for python 2,
which since now that we support python 3, have become obsolete. Some
also simply were not used:

 * `abstractclassmethod`: now available in `abc.abstractclassmethod`
 * `abstractstaticmethod`: now available in `abc.abstractstaticmethod`
 * `combomethod`: not used at all
 * `EmptyContextManager`: from py3.4 `contextlib.suppress()`
 * `protected_decorator`: only used in two places, but with `check=False`
   so it would never actually raise when the method was called from
   outside the class. What is worse, the `report` method is only useful
   when callable from the outside. All in all, it is best to remove this
   whole concept.


greschd:
Fabulous! Thanks a lot @sphuber, great to get rid of some cruft.

sphuber:
Thanks to you for the review and suggestion. I added the comment

greschd:
I've seen it.. the fastest review in the west 😉

PR #3807, opened 2020-02-25
Disable caching for data

### LINKED ISSUE #3802: Caching messes with 'verdi code duplicate'
When caching is globally enabled (``default: True``), trying to exactly duplicate a Code but changing its label results in another Code with the _same_ label being created. 

The reason for this is that the ``label`` is a mutable attribute of the ``Code``, and thus not taken into account when creating the hash. As a result, the ``Code`` is stored from cache and gets the old label.

A simple fix would just be setting ``_cachable = False`` on the ``Code`` class. Tagging @sphuber @giovannipizzi for comment.

Example output below:

```bash
(aiida) a-dogres@A-DOGRES-082219:~/programming/aiida/aiida-core$ verdi code list
# List of configured codes:
# (use 'verdi code show CODEID' to see the details)
[...]
* pk 578 - pw-6.4.1@localhost

(aiida) a-dogres@A-DOGRES-082219:~/programming/aiida/aiida-core$ verdi code duplicate pw-6.4.1
Info: enter "?" for help
Label [pw-6.4.1]: pw-new-label
Description [QE 6.4.1 pw.x]: 
Default calculation input plugin [quantumespresso.pw]: 
Installed on target computer? [True]: 
Computer [localhost]: 
Remote absolute path [/home/a-dogres/codes/qe-6.4.1/bin/pw.x]: 
Success: Code<666> pw-6.4.1@localhost created

(aiida) a-dogres@A-DOGRES-082219:~/programming/aiida/aiida-core$ verdi code list
# List of configured codes:
# (use 'verdi code show CODEID' to see the details)
[...]
* pk 578 - pw-6.4.1@localhost
* pk 666 - pw-6.4.1@localhost
```

greschd:
Fixes #3802.

Change the default for the ``_cachable`` class attribute to ``False`` in the ``Node`` class. This means it will be enabled only in the ``CalculationNode`` sub-classes.

### DIFF
--- a/tests/orm/node/test_node.py
+++ b/tests/orm/node/test_node.py
@@ -17,7 +17,6 @@ from aiida.backends.testbase import AiidaTestCase
 from aiida.common import exceptions, LinkType
 from aiida.orm import Data, Node, User, CalculationNode, WorkflowNode, load_node
 from aiida.orm.utils.links import LinkTriple
-from aiida.manage.caching import enable_caching
 
 
 class TestNode(AiidaTestCase):
@@ -749,7 +748,9 @@ def test_store_from_cache(clear_database_before_test):  # pylint: disable=unused
         data.put_object_from_tree(tmpdir)
 
     data.store()
-    with enable_caching():
-        clone = data.clone().store()
+
+    clone = data.clone()
+    clone._store_from_cache(data, with_transaction=False)  # pylint: disable=protected-access
+
     assert clone.get_cache_source() == data.uuid
     assert data.get_hash() == clone.get_hash()
--- a/tests/orm/node/test_node.py
+++ b/tests/orm/node/test_node.py
@@ -750,7 +750,8 @@ def test_store_from_cache(clear_database_before_test):  # pylint: disable=unused
     data.store()
 
     clone = data.clone()
-    clone._store_from_cache(data, with_transaction=False)  # pylint: disable=protected-access
+    clone._store_from_cache(data, with_transaction=True)  # pylint: disable=protected-access
 
+    assert clone.is_stored
     assert clone.get_cache_source() == data.uuid
     assert data.get_hash() == clone.get_hash()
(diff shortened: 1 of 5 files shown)


greschd:
I had to adapt the test for ``_store_from_cache``, because it uses a ``Data`` node. By explicitly calling ``_store_from_cache``, the ``_cachable`` and ``is_valid_cache`` checks are circumvented.

Another option (which I also tried) would be using a ``CalculationNode`` instead, but then we have to manually set ``process_state`` and ``process_type`` to ensure ``is_valid_cache`` is ``True``. Overall, it's the uglier solution IMO, with dependencies to different parts of the codebase.

greschd:
On the SQLA backend, ``_store_from_cache`` only works with ``with_transaction=True`` - is this expected?

tests/orm/node/test_node.py
```diff
@@ -749,7 +748,10 @@ def test_store_from_cache(clear_database_before_test):  # pylint: disable=unused
         data.put_object_from_tree(tmpdir)
 
     data.store()
-    with enable_caching():
-        clone = data.clone().store()
+
+    clone = data.clone()
+    clone._store_from_cache(data, with_transaction=True)  # pylint: disable=protected-access
```
sphuber: It is a bit weird that we have a caching test for a `Data` node even though they are now explicitly not cacheable. Can we change this in a `CalculationNode`? Should be straightforward change and then you can revert to using `with enable_caching()`

greschd: I did try that (see my comment above), but it becomes complicated trying to make sure that the `CalculationNode` fulfills all the requirements of the `is_valid_cache` check. I'd rather not rely on that specific behavior (as we may change the `is_valid_cache` check again).

ltalirz: What if `Node`s remain cacheable by default, but this is simply disabled at the `Data` level?

greschd: This test is not possible with `Node` because they can not be stored.

Issue #3811, opened 2020-02-27
Isolated `BandsData` will raise exception when running `verdi data bands list`


unkcpz:
```
bands = load_node(<bands_node>)
bands_clone = bands.clone()
bands_clone.store()
```
will cause `verdi data bands list` break.

```
File "/home/unkcpz/pyProject/aiida_core/aiida/backends/djsite/queries.py", line 178, in get_bands_and_parents_structure
    struc_pks = [structure_dict[pk] for pk in pks]
File "/home/unkcpz/pyProject/aiida_core/aiida/backends/djsite/queries.py", line 178, in <listcomp>
    struc_pks = [structure_dict[pk] for pk in pks]
KeyError: 230
```

Although I have no idea whether it is reasonable to have `BandsData` node without structure as its ancestor, this is still a case to consider. but I don't know which of the following measures is more reasonable

1. forbid users from creating `BandsData` instance without set a `StructureData` as its ancestor node.
2. when running `verdi data bands list` show `None` as its formula value.

I am in favor of the second personally.

ltalirz:
Thanks @unkcpz for pointing this out. I agree with your assessment and would vote for 2.

Mentioning @giovannipizzi and @sphuber for comment.

greschd:
Agree with #2, I'm fairly certain I've created `BandsData` without associated `StructureData` in some places.

sphuber:
This really depends on the original design of `BandsData`. If it really needs an associated `StructureData` then option 2 would be a bit of a workaround. But given that one can already created a separate `BandsData` object, interfaces should not except, so I would also vote for option 2

ltalirz:
Thanks for the input - @unkcpz this is the "go" signal ;-)

unkcpz:
I make a PR. but one thing need to be discussed, when a bands object is cloned, its `StructureData` should also be cloned? Maybe this behavior is what users expected when calling the `.clone()` function.

greschd:
I don't think it should clone the `StructureData`. Looking at the code, the "linked structure" is anyway just determined through a query for the closest ancestor, and `.clone()` definitely shouldn't do any node linking - the graph relationships of the clone should be up to the calling code (e.g. in the caching code).

That makes me wonder, though, if that linked structure is always the right one. It looks for the closest parent, but for example if a calculation creates both structure and bands in one step, the correct structure might be a sibling node.

PR #3798, opened 2020-02-24
plot bands with only one kpoint

unkcpz:
Fix #2462 : Support plotting `BandsData` instances with a single kpoint.

Issue #3755, opened 2020-02-12
QueryBuilder: with_ancestors does not work as expected


ltalirz:
Consider the following simple graph
![graph](https://user-images.githubusercontent.com/1053245/74344352-b84d2a80-4dac-11ea-8c7a-f0d744c63a27.png)

With the following query
```python
qb=QueryBuilder()
qb.append(Int, filters={ 'uuid': '8dba489c-6c37-49b8-9bdd-b3bbb0c19044' }, tag='top')
qb.append(Node, filters={ 'uuid': '77ffe2ac-82ae-4e6f-b972-b6604cc94671'}, with_ancestors='top')
```
I would expect exactly one result (the node at the bottom).

However, I get this:
```
In [6]: qb.all()
Out[6]:
[[<Float: uuid: 77ffe2ac-82ae-4e6f-b972-b6604cc94671 (pk: 27723) value: 256.0>],
 [<Float: uuid: 77ffe2ac-82ae-4e6f-b972-b6604cc94671 (pk: 27723) value: 256.0>]]
```

The reason is that there are two paths to arrive at the final node (one through the left, and one through the right). Discovered with @danieleongari

Mentioning @giovannipizzi and @lekah for comment.

P.S. @giovannipizzi We did test the `with_ancestors` approach and with the current implementation it is very slow. As you can imagine, you can easily get back the same node hundreds of times.

P. P.S. Python script to create the graph
```python
from aiida.engine import calcfunction
from aiida.orm import Int

@calcfunction
def divide(inp):
    return {'half_1': inp/2, 'half_2': inp/2}

@calcfunction
def sum(inp1, inp2):
    return inp1 + inp2

a = Int(256)
print(a.uuid)
b, c = divide(a).values()
d = sum(b,c)
```

CasperWA:
I guess the simple solution would be to use `set`s somewhere in the method/function? But this will not solve the speed issues, as well as the fact that all "similar" edges in the graph are probed.

lekah:
I'd say your example works exactly as expected. Same if you have 2 links connecting two Nodes (1 and 2), you'd get node2 twice traversing from node1. It's because the Cartesian product (the join) has 2 solutions, as is the case in above example. If you want unique solutions, use qb.distinct().all(), that is asking the backend to return unique solutions.

ltalirz:
> I'd say your example works exactly as expected.

[...]

PR #3817, opened 2020-02-29
isolated bandsdata list and test

unkcpz:
Fixes https://github.com/aiidateam/aiida-core/issues/3811

- when no ancestors found for `BandsData` object `verdi data bands list` returns `NOT FOUND` as formula. `<<UNKNOW>>` reserved for the situation when ancestor found but its formula cannot be read properly.
- in `abstractqueries::get_bands_and_parents_structure` query for band after the group query to prevent item listed in empty group.
- `_extrat_formula` removed from `djsite::queries` use the general function.


unkcpz:
Emm.. I think I don't have the permissions to assign reviewers, right? Could you review it?@ltalirz

aiida/backends/general/abstractqueries.py
```diff
@@ -136,7 +136,7 @@ def _extract_formula(args, akinds, asites):
 
         # We want only the StructureData that have attributes
         if akinds is None or asites is None:
-            return None
+            return '<<UNKNOWN>>'
```
ltalirz: Just to understand: why do we need to introduce this custom label?
EDIT: Oh, I see, it is already in the code... Aren't you changing the meaning of `<<UNKNOWN>>` here, though?

Anyhow, since you now already looked into this function, if you could add a few lines of documentation to its docstring that would be great.
Have you seen whether this `args` object is documented anywhere?
We're really digging in some aiida code from the "old days" here... great to clean this up a little bit.

tests/cmdline/commands/test_data.py
```diff
@@ -322,6 +323,11 @@ def test_bandlistshelp(self):
     def test_bandslist(self):
         self.data_listing_test(BandsData, 'FeO', self.ids)
 
+    def test_bandslist_isolated_node(self):
+        bands = BandsData()
+        bands.store()
+        self.data_listing_test(BandsData, '<<UNKNOWN>>')
```
ltalirz: Would it perhaps be easier to include the "lone" BandsData node simply in  `create_structure_bands()`?

I guess then you might also not have to modify the `data_listing_test` (?)

unkcpz: You are totally right! I didn't clearly understand how function `data_listing_test` works.

aiida/backends/general/abstractqueries.py
unkcpz: Yes, I think you are right. I did this because I want to make as few changes as possible.
Now, I use `NOT FOUND` for the isolated bands, and reserve `<<UNKNOW>>` for what was originally designed for (the situation when formula variable is unproperly assigned).

ltalirz:
Thanks a lot @unkcpz for working through this convoluted code!

Before I go through this line-by-line:
@giovannipizzi - in this PR, @unkcpz is removing the django-specific implementation of `_extract_formula` (I agree).

While we're at it, there is also a django-specific implementation of `get_bands_and_parents_structure`.
Do you mind giving this function a look and letting us know whether there are actual performance reasons for keeping this, or can we get rid of this as well?
To me, this code looks very complicated, so it should really be significantly faster in order to justify its existence:
https://github.com/aiidateam/aiida-core/blob/09cb6cb70df024143ac5a3720e4034b609536b02/aiida/backends/djsite/queries.py#L148-L193

 we are removing a

unkcpz:
@ltalirz @giovannipizzi 
If I understand correctly, code support for different backend comes from historical reasons and there was no efficiency promotion. This piece of code was added https://github.com/aiidateam/aiida-core/commit/e94fad187f93430522181a2a190c9bbede4e7085 , at that time, it seemed that all queries is done by `Q` function directly . 
All the band query and list code would be better to remove to `aiida/cmdline/command/cmd_data/cmd_bands.py` where the function was called. This will keep it consistent with other similar implementations such as `cmd_structure`.
 
If it is correctly, I can make a PR to try to fixes the band query function here and check if the other place have the same problem.
my two cents

giovannipizzi:
I don't think this function is a leftover - it was moved there on purpose (I have some memory of this) because it was faster. Now, I don't know if with the current changes it's faster. Note that this touched the transitive table, i.e. now the ancestors and descendants in the QueryBuilder.

I agree that that functionality is quite specific. However, before removing, I suggest testing the Django-specific implementation vs. the generic implementation, and reporting the results here (but this should be done in a big DB).
Actually, this could be a good test for @danieleongari (see #3755) to check if this (or an adaptation of this) is faster than using the QueryBuilder? (It might be, this gets the 'closest' ancestors, while the QueryBuilder always gets everything).

unkcpz:
The original purpose of this PR seems already reached. Could you merge this so I can get back to https://github.com/aiidateam/aiida-core/pull/3798 based on this PR? @ltalirz 
or we just keep this PR open and wait for the report from @danielhollas ?

I have no experience at working several interrelated issues at the same time, could you tell me how to deal with this situation? :-p

danielhollas:
> The original purpose of this PR seems already reached. Could you merge this so I can get back to #3798 based on this PR? @ltalirz
> or we just keep this PR open and wait for the report from @danielhollas ?
Too many Daniels in the world :laughing: 

@unkcpz I think you wanted to mention @danieleongari, though I seem to be unable to mention him for some reason, though Github does not suggest him while typing for some reason...

unkcpz:
forgive my careless @danielhollas . yes, Github does not suggest him while typing.

ltalirz:
@CasperWA Just to understand - should codecov now be working on PRs from forks or not?

CasperWA:
> @CasperWA Just to understand - should codecov now be working on PRs from forks or not?

I believe it should work. You can always check whether a codecov report was successfully uploaded by checking the CI run.

See, e.g., [here](https://github.com/aiidateam/aiida-core/runs/525733332?check_suite_focus=true#step:12:56).
It seems for this particular instance it tried several times to upload, and finally succeded, while for [the `django` backend](https://github.com/aiidateam/aiida-core/runs/525733324?check_suite_focus=true#step:12:36) it succeded to upload right away.

tests/cmdline/commands/test_data.py
```diff
@@ -321,6 +325,13 @@ def test_bandlistshelp(self):
 
     def test_bandslist(self):
         self.data_listing_test(BandsData, 'FeO', self.ids)
+        self.data_listing_test(BandsData, 'NOT FOUND', self.ids)
```
ltalirz, line 328:
```suggestion
        self.data_listing_test(BandsData, '<<NOT FOUND>>', self.ids)
```

tests/cmdline/commands/test_data.py
```diff
@@ -321,6 +325,13 @@ def test_bandlistshelp(self):
...
+
+    def test_bandslist_with_elements(self):
+        options = ['-e', 'Fe']
+        res = self.cli_runner.invoke(cmd_bands.bands_list, options, catch_exceptions=False)
+        self.assertIn(b'FeO', res.stdout_bytes, 'The string "FeO" was not found in the listing')
+        self.assertNotIn(b'NOT FOUND', res.stdout_bytes, 'The string "NOT FOUND" should not in the listing')
```
ltalirz, line 334:
```suggestion
        self.assertNotIn(b'<<NOT FOUND>>', res.stdout_bytes, 'The string "<<NOT FOUND>>" should not in the listing')
```

PR #3824, opened 2020-03-04
`ExitCode`: make the exit message parameterizable through templates

### LINKED ISSUE #2644: Allow `self.exit_codes` to be callable with keyword and custom message
The use case was presented where one would like to define a generic exit code for a missing file, where one would always want to use the same exit status, but have a customized exit message that includes for example the exact filename. Currently this is not possible, but maybe it would be possible to extend the syntax to be able to do both:
```
exit_code_default = self.exit_codes.ERROR_MISSING_FILE
exit_code_custom = self.exit_codes(ERROR_MISSING_FILE, 'File `aiida.out` is missing')
```

we would have to make the `exit_codes` property callable and override the behavior to construct a new `ExitCode` instance with the custom message.

Or maybe even better, we make the `ExitCode` callable which allows to change the message. This would require the `ExitCode` to be changed from a named tuple into a class, e.g.:
```
class ExitCode(object):

    def __init__(self, status, message):
        self.status = status
        self.message = message

    def __call__(self, message):
        self.message = message
        return self
```

sphuber:
Fixes #2644

Often one would like to define an exit code for a process where the
general gist of the message is known, however, the exact form might be
slightly situation dependent. Creating individual exit codes for each
slight variation is cumbersome so we introduce a way to parameterize the
exit message. To do this the `ExitCode` had to be changed from named
tuple to a proper class, such that the `__call__` method could be
implemented. This allows a mechanism where calling an exit code with a
parameterized exit message can be concretized. The keyword arguments are
used for the string interpolation and, importantly, a new exit code
instance is returned. Example:

    exit_code_template = ExitCode(450, 'the parameter {parameter} is invalid.')
    exit_code_concrete = exit_code_template(parameter='some_specific_key')

The `exit_code_concrete` is now unique to `exit_code_template` except
for the message whose parameters have been replaced.

### DIFF
--- a/aiida/engine/processes/exit_code.py
+++ b/aiida/engine/processes/exit_code.py
@@ -8,17 +8,21 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """A namedtuple and namespace for ExitCodes that can be used to exit from Processes."""
+from collections import namedtuple
 from aiida.common.extendeddicts import AttributeDict
 
 __all__ = ('ExitCode', 'ExitCodesNamespace')
 
 
-class ExitCode:
+class ExitCode(namedtuple('ExitCode', ['status', 'message', 'invalidates_cache'])):
     """A simple data class to define an exit code for a :class:`~aiida.engine.processes.process.Process`.
 
     When an instance of this clas is returned from a `Process._run()` call, it will be interpreted that the `Process`
     should be terminated and that the exit status and message of the namedtuple should be set to the corresponding
     attributes of the node.
+
+    .. note:: this class explicitly sub-classes a namedtuple to not break backwards compatibility and to have it behave
+        exactly as a tuple.
 
     :param status: positive integer exit status, where a non-zero value indicated the process failed, default is `0`
     :type status: int
@@ -29,11 +33,6 @@ class ExitCode:
     :param invalidates_cache: optional flag, indicating that a process should not be used in caching
     :type invalidates_cache: bool
     """
-
-    def __init__(self, status=0, message=None, invalidates_cache=False):
-        self.status = status
-        self.message = message
-        self.invalidates_cache = invalidates_cache
 
     def format(self, **kwargs):
         """Create a clone of this exit code where the template message is replaced by the keyword arguments.
@@ -51,6 +50,10 @@ class ExitCode:
 
     def __eq__(self, other):
         return all(getattr(self, attr) == getattr(other, attr) for attr in ['status', 'message', 'invalidates_cache'])
+
+
+# Set the defaults for the `ExitCode` attributes
+ExitCode.__new__.__defaults__ = (0, None, False)
 
 
 class ExitCodesNamespace(AttributeDict):
--- a/aiida/engine/processes/exit_code.py
+++ b/aiida/engine/processes/exit_code.py
@@ -35,7 +35,7 @@ class ExitCode:
         self.message = message
         self.invalidates_cache = invalidates_cache
 
-    def __call__(self, **kwargs):
+    def format(self, **kwargs):
         """Create a clone of this exit code where the template message is replaced by the keyword arguments.
 
         :param kwargs: replacement parameters for the template message
--- a/aiida/engine/processes/exit_code.py
+++ b/aiida/engine/processes/exit_code.py
@@ -8,49 +8,66 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """A namedtuple and namespace for ExitCodes that can be used to exit from Processes."""
-
-from collections import namedtuple
-
 from aiida.common.extendeddicts import AttributeDict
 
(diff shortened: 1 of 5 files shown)


aiida/engine/processes/exit_code.py
```diff
@@ -8,49 +8,66 @@
...
 ###########################################################################
 """A namedtuple and namespace for ExitCodes that can be used to exit from Processes."""
-
-from collections import namedtuple
-
 from aiida.common.extendeddicts import AttributeDict
 
 __all__ = ('ExitCode', 'ExitCodesNamespace')
 
-ExitCode = namedtuple('ExitCode', ['status', 'message', 'invalidates_cache'])
```
greschd: This is technically a change in interface, because `namedtuple` exposes all the methods of the underlying `tuple`. For example, one could do 

```
status, message, invalidates_cache = *ExitCode()
```

I'm not sure if any of these "accidental features" is used anywhere, though.

aiida/engine/processes/exit_code.py
```diff
@@ -8,49 +8,66 @@
...
-ExitCode.__new__.__defaults__ = (0, None, False)
-"""
-A namedtuple to define an exit code for a :class:`~aiida.engine.processes.process.Process`.
 
-When this namedtuple is returned from a Process._run() call, it will be interpreted that the Process
-should be terminated and that the exit status and message of the namedtuple should be set to the
-corresponding attributes of the node.
+class ExitCode:
```
greschd: I'm not entirely convinced we should mix the "exit code" and "exit code template" concepts in the same class. On the one hand, this makes it easy in the `spec.exit_code` call (you just pass on the message) -- but on the other hand, this makes it almost impossible for the code that uses the exit code to figure out which one it is.

Then again I can't come up with a convincing example for when you would need to distinguish this, so this might be fine.

aiida/engine/processes/exit_code.py
```diff
@@ -8,49 +8,66 @@
...
+    :param invalidates_cache: optional flag, indicating that a process should not be used in caching
+    :type invalidates_cache: bool
+    """
+
+    def __init__(self, status=0, message=None, invalidates_cache=False):
+        self.status = status
+        self.message = message
+        self.invalidates_cache = invalidates_cache
+
+    def __call__(self, **kwargs):
```
greschd: Also not completely sure about using `__call__` for this - if we want to extend the functionality of the `ExitCode` class in the future, it would be slightly arbitrary to give `__call__` to this method. Maybe `evaluate` or `format`?

aiida/engine/processes/exit_code.py
sphuber: Good point. Do you know of a way to make a simple class replicate the exact interface/behavior of a named tuple? If not, we might have to put this in `2.0`

espenfl: Indeed. Also this functionality is probably not super critical, depending on when a `2.0` release is scheduled.

aiida/engine/processes/exit_code.py
sphuber: The template part here only really applies to the message though and as the message really is just a human readable description/alternative of the exit status, I don't think this is a problem. Any machine will only ever just have to look at the status which is "immutable" and should never even look at the message.

aiida/engine/processes/exit_code.py
greschd: Behold, a thing of beauty:
```python
from collections import namedtuple

class ExitCode(namedtuple('_ExitCode', ['status', 'message', 'invalidates_cache'])):

    def __call__(self, **kwargs):
        pass

ExitCode.__new__.__defaults__ = (0, None, False)

ExitCode(0, 'bla')()
```

aiida/engine/processes/exit_code.py
greschd: Yeah, by "calling" code I meant where the exit code is returned from.. like, maybe you want to know if you should "format" the exit code or no.

greschd: But, again, I'm not super convinced on this. Since the message has to use named formatting you can always "format" with a dict that has too many keys.

aiida/engine/processes/exit_code.py
sphuber: Yer-a-wizard

aiida/engine/processes/exit_code.py
greschd: I think the latter conveys the intent a bit better - makes it clear that the parameters are there only for formatting the message, not being taken into account in any other way.

But, this might be bike-shedding (sorry for that), and I certainly don't have a _strong_ opinion either way. Both versions will work just fine.

sphuber:
@greschd this virus outbreak has made me see the light and I agree with your suggestions presented during the bike shedding. One now uses `format` instead of a call to format an exit code and its namedtupleness has been restored to former glory. This is ready for review

PR #3831, opened 2020-03-04
add append and prepend to aiida_local_code_factory

pzarabadip:
This PR aims to add the ability of adding `append_text` and `prepend_text` to the `aiida_local_code_factory` which is need in writing tests for codes where we need to have these extra fields.

### DIFF
--- a/aiida/manage/tests/pytest_fixtures.py
+++ b/aiida/manage/tests/pytest_fixtures.py
@@ -133,13 +133,15 @@ def aiida_local_code_factory(aiida_localhost):  # pylint: disable=redefined-oute
     :rtype: object
     """
 
-    def get_code(entry_point, executable, computer=aiida_localhost):
+    def get_code(entry_point, executable, computer=aiida_localhost, append=None, prepend=None):
         """Get local code.
         Sets up code for given entry point on given computer.
 
         :param entry_point: Entry point of calculation plugin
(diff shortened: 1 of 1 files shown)


greschd:
Looks useful to me, thanks! To keep consistent with the naming of the Code attribute, I would probably call the keyword arguments `prepend_text` and `append_text` (even if it's a bit more verbose).

pzarabadip:
Thanks @greschd. I just applied your suggestion and pushed the changes.

aiida/manage/tests/pytest_fixtures.py
```diff
@@ -133,13 +133,17 @@ def test_1(aiida_local_code_factory):
...
 
-    def get_code(entry_point, executable, computer=aiida_localhost):
+    def get_code(entry_point, executable, computer=aiida_localhost, prepend_text=None, append_text=None):
         """Get local code.
         Sets up code for given entry point on given computer.
 
         :param entry_point: Entry point of calculation plugin
         :param executable: name of executable; will be searched for in local system PATH.
         :param computer: (local) AiiDA computer
+        :param prepend: a string of code that will be put in the scheduler script before the
```
greschd: Please change it in the docstring also.

pzarabadip: Sorry my bad. I have overlooked it. Done.

PR #3834, opened 2020-03-09
Update coverage step

CasperWA:
Add test workflow to test the if-statement

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,10 +139,9 @@ jobs:
         .github/workflows/tests.sh
 
     - name: Upload coverage report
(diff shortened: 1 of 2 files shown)


CasperWA:
I am testing the new tokenless coverage upload - wanted to create this PR againt my fork - will create a new proper PR here shortly.

csadorf:
What is the reason that we are measuring for the almost deprecated Python version 3.5, instead of the latest supported one (3.8)?

CasperWA:
> What is the reason that we are measuring for the almost deprecated Python version 3.5, instead of the latest supported one (3.8)?

Moving this to the "correct" PR

PR #3835, opened 2020-03-09
Update coverage upload step and use actions/checkout v2

CasperWA:
Update coverage upload step.

Since it is [now possible](https://community.codecov.io/t/whitelist-github-action-servers-to-upload-without-a-token/491/17) to upload tokenless to codecov (_only_ from public GitHub repositories), we should do this for our PRs.

CI also fails if the coverage upload fails now.

The logic for when the coverage should be uploaded:
- ONLY for tests with Python 3.5
- ~If it's a PR.~
- ~If it's a push commit AND the repository is `aiidateam/aiida-core`.~
- It's a push in or PR against the repository `aiidateam/aiida-core`

Note: I wrongly created the PR #3834 to test the `if` part of when the coverage should be uploaded. It was originally meant to point to my fork - and since I later rebased, it didn't allow me to reopen the PR. So here's a new one :)

**Edit**: Also updated the checkout action to ~v2~ master (~`actions/checkout@v2`~ `actions/checkout@master`). It's more robust and performant than v1, see [here](https://github.com/actions/checkout) for more information.
It was necessary to use `master` instead of `v2`, since otherwise the codecov-action cannot retrieve the PR number, which is needed for uploading coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,14 +139,13 @@ jobs:
         .github/workflows/tests.sh
 
     - name: Upload coverage report
-      if: matrix.python-version == 3.5
-      uses: codecov/codecov-action@v1.0.5
+      if: matrix.python-version == 3.5 && github.repository == 'aiidateam/aiida-core'
+      uses: codecov/codecov-action@v1
       with:
-        token: ${{ secrets.CODECOV_TOKEN }}
         name: aiida-pytests-py3.5-${{ matrix.backend }}
         flags: ${{ matrix.backend }}
         file: ./coverage.xml
-        fail_ci_if_error: false
+        fail_ci_if_error: true
 
   verdi:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     timeout-minutes: 30
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@master
 
     - name: Set up Python 3.7
       uses: actions/setup-python@v1
@@ -29,7 +29,7 @@ jobs:
     timeout-minutes: 30
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@master
 
     - name: Set up Python 3.7
       uses: actions/setup-python@v1
@@ -60,7 +60,7 @@ jobs:
     timeout-minutes: 30
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@master
 
     - name: Set up Python 3.7
       uses: actions/setup-python@v1
@@ -95,7 +95,7 @@ jobs:
         backend: ['django', 'sqlalchemy']
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@master
     - uses: CasperWA/postgresql-action@v1.2
       with:
         postgresql version: '10'
@@ -153,7 +153,7 @@ jobs:
     timeout-minutes: 30
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@master
 
     - name: Set up Python 3.7
       uses: actions/setup-python@v1
@@ -175,7 +175,7 @@ jobs:
     timeout-minutes: 30
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@master
 
     - name: Install docker
       run: |


CasperWA:
I've already found a fault here. If it's a PR, it should be a requirement that it's towards the upstream repository, i.e., `aiidateam/aiida-core`, and not just a random PR.

CasperWA:
Note for person merging. The two commits do separate conceptual things, so please hesitate before squashing, or simply don't. One updates the coverage step, the other updates actions/checkout to v2.

.github/workflows/ci.yml
```diff
@@ -139,22 +139,21 @@ jobs:
         .github/workflows/tests.sh
 
     - name: Upload coverage report
-      if: matrix.python-version == 3.5
-      uses: codecov/codecov-action@v1.0.5
+      if: matrix.python-version == 3.5 && github.repository == 'aiidateam/aiida-core'
+      uses: codecov/codecov-action@v1
```
ltalirz: going back to older version?

.github/workflows/ci.yml
```diff
@@ -139,22 +139,21 @@ jobs:
...
       with:
-        token: ${{ secrets.CODECOV_TOKEN }}
         name: aiida-pytests-py3.5-${{ matrix.backend }}
         flags: ${{ matrix.backend }}
         file: ./coverage.xml
-        fail_ci_if_error: false
+        fail_ci_if_error: true
```
ltalirz: I guess this is the default?

CasperWA:
> Thanks @CasperWA - so, the coverage report shows that this PR decreases coverage by 7%.
> I guess this is not correct, is it?
> 
> Do you know what is going wrong here?

Yes. It's apparently the update to the checkout v2 action. I have just force pushed an update to use checkout@master instead.

See my issue in the codecov-action [here](https://github.com/codecov/codecov-action/issues/61).

.github/workflows/ci.yml
CasperWA: There is no default for this option in the action itself (see [here](https://github.com/codecov/codecov-action/blob/master/action.yml)).

Another thing is, apparently it didn't catch the latest failure, which is another error in the codecov-action action. I have reported it in the same issue as referenced below.

.github/workflows/ci.yml
CasperWA: Yes. This is what they recommend. And they have updated the `v1` tag with the updates from the "newer" versions. I.e., `v1` is synonymous with the latest v1, no matter the minor or patch version.

CasperWA:
> > Thanks @CasperWA - so, the coverage report shows that this PR decreases coverage by 7%.
> > I guess this is not correct, is it?
> > Do you know what is going wrong here?
> 
> Yes. It's apparently the update to the checkout v2 action. I have just force pushed an update to use checkout@master instead.
> 
> See my issue in the codecov-action repository [here](https://github.com/codecov/codecov-action/issues/61).

It seems to work with the newest version of the checkout action, i.e., using the `master` branch.

CasperWA:
From @csadorf:
> What is the reason that we are measuring for the almost deprecated Python version 3.5, instead of the latest supported one (3.8)?

CasperWA:
> What is the reason that we are measuring for the almost deprecated Python version 3.5, instead of the latest supported one (3.8)?

That the coverage is/should be the same, no matter the Python version - therefore it makes sense (to me) to only upload for a single Python version. That we are uploading on 3.5 is irrelevant, it could as well have been 3.8.
It _is_ however important to upload for both backends, since the coverage will be different.

Previously the coverage would be different for the two different major Python versions, hence it made sense to upload for both (indeed, we appended four different coverage reports together, varying over major Python version and backends).

ltalirz:
If there is indeed a significant difference of test coverage between python versions, we *should* be merging it together.

@csadorf Feel free to change this back in your dependency-management PR.

csadorf:
> > What is the reason that we are measuring for the almost deprecated Python version 3.5, instead of the latest supported one (3.8)?
> 
> That the coverage is/should be the same, no matter the Python version - therefore it makes sense (to me) to only upload for a single Python version. That we are uploading on 3.5 is irrelevant, it could as well have been 3.8.
> It _is_ however important to upload for both backends, since the coverage will be different.
> 
> Previously the coverage would be different for the two different major Python versions, hence it made sense to upload for both (indeed, we appended four different coverage reports together, varying over major Python version and backends).

Thank you for clarifying. While it is true that the coverage should not differ significantly between Python minor versions and it therefore appears arbitrary which version to test, I would argue that we should use the most recent supported version in all CI steps that are only concerned with a single version. The reason is that in this way we reduce the risk of any undetected incompatibilities as we slowly upgrade each part of the CI toolchain. This allows us to reduce the version instead of increase it should there be any issues. This may not be as relevant for this particular step, but I would prefer to keep things consistent across the different CI steps.

CasperWA:
> If there is indeed a significant difference of test coverage between python versions, we _should_ be merging them together.

Agreed. Although, my point was that I wouldn't expect a difference between the two Python versions in question (3.5 and 3.8).

CasperWA:
> Thank you for clarifying. While it is true that the coverage should not differ significantly between Python minor versions and it therefore appears arbitrary which version to test, I would argue that we should use the most recent supported version in all CI steps that are only concerned with a single version. The reason is that in this way we reduce the risk of any undetected incompatibilities as we slowly upgrade each part of the CI toolchain. This allows us to reduce the version instead of increase it should there be any issues. This may not be as relevant for this particular step, but I would prefer to keep things consistent across the different CI steps.

Let me rephrase a bit.
I expect the code coverage won't change at all between the two versions.
I see your point of keeping to the newest Python version. But I can't see why this argument can't be turned on its head as well, to argue for the opposite point, i.e. that we should be testing the oldest supported version, since it is more likely users are on old versions and not new, and so if something fails for the oldest supported version, we have a more serious problem than if the newest version is not completely supported.
In the ideal world, we should make sure both ends of this rope don't burn off.

In the end, this is all besides the point for this particular PR and its contents, since it's about code coverage and not testing the actual code. So the version truly doesn't matter, unless different versions would introduce the execution of different lines of code.
So if everyone agrees we should focus on the newest Python version and not the oldest supported (by AiiDA), then this should align with that philosophy as well, agreed.

csadorf:
@CasperWA I think you slightly misunderstood my argument. I am talking about compatibility within our own CI toolchain, not about compatibility with user versions. With respect to the latter, we need to run the full tests suite on all supported versions, which will be the case once #3771 is merged.

PR #3836, opened 2020-03-09
Update verdi_user_guide.rst.

csadorf:
The latest release of Click (7.1) produces a slightly different
formatting of the CLI help strings.

This patch updates our docs to match that formatting.

This patch fixes the failure of the verdi-autodocs pre-commit hook with
Click version 7.1.


csadorf:
Withdrawing in favor of #3837 , because this also requires fixing tests.

PR #3837, opened 2020-03-09
Pin the Click version to 7.0.

csadorf:
The recent update to Click (7.1) breaks our tests and this CI workflow,
because the help output formatting was slightly changed.

Until we have resolved other issues we should pin the Click version to
7.0.

Because of this incidence, I have added "Click" to the [list of problematic dependencies](https://github.com/aiidateam/aiida-core/wiki/AiiDA-Dependency-Management#problematic-dependencies).

### DIFF
--- a/docs/requirements_for_rtd.txt
+++ b/docs/requirements_for_rtd.txt
@@ -7,7 +7,7 @@ circus~=0.16.1
 click-completion~=0.5.1
 click-config-file~=0.5.0
 click-spinner~=0.1.8
-click~=7.0
+click==7.0
 coverage<5.0
 django~=2.2
 docutils==0.15.2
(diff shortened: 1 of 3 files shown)


sphuber:
Just for reference @csadorf , for most PR's, especially for single commit one's we prefer to use squash-and-merge instead of a normal merge. The linear history and absence of unnecessary merge-commits makes the history easier to follow. We still use normal merges where appropriate such as the merging of feature branches. For multi commit PRs where we don't want to squash we use rebase and merge.

csadorf:
@sphuber Thank's for letting me know. I see that these policies are documented [here](https://github.com/aiidateam/aiida-core/wiki/Git-style-requirements#pull-requests), however I did not originally see this, because I don't find the page name "Git style requirements" very descriptive. Maybe we can rename it or even have a separate page about "Merge policies" or so?

sphuber:
Agreed that the current name doesn't describe its full content very well. Sounds like a more broad title or a separate page would be a good solution

PR #3839, opened 2020-03-10
Update requirements workflow triggers

csadorf:
To prevent the GitHub actions bot to create PRs on forks and otherwise unconcerned branches.

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -2,16 +2,12 @@ name: test-install
 
 on:
   push:
-    branches:
-      - master
-      - develop
-      - release/*
-      - dm/*
     paths:
       - 'setup.*'
       - 'environment.yml'
-      - 'requirements*.txt'
+      - '**/requirements*.txt'
       - 'pyproject.toml'
+      - 'util/dependency_management.py'
       - '.github/workflows/test-install.yml'
   schedule:
     - cron: '30 02 * * *'  # nightly build
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -2,7 +2,7 @@ name: test-install
 
 on:
   push:
-    branch:
+    branches:
       - master
       - develop
       - release/*
--- a/.github/workflows/update-requirements.yml
+++ b/.github/workflows/update-requirements.yml
@@ -95,11 +95,24 @@ jobs:
           name: requirements.txt
           path: requirements
 
-      - name: Create Pull Request
-        if: github.repository == 'aiidateam/aiida-core'
+      - name: Commit requirements files
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add requirements/*
+          git commit -m "Update requirements files." -a || echo "Nothing to update."
+
+      - name: Push changes
+        uses: ad-m/github-push-action@v0.5.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: ${{ github.ref }}
+
+      - name: Create Pull Request (since update via push failed)
+        if: failure() && github.repository == 'aiidateam/aiida-core'
         uses: peter-evans/create-pull-request@v2
         with:
-          commit-message: "Update requirements.txt"
+          commit-message: "Update requirements files."
           token: ${{ secrets.GITHUB_TOKEN }}
           title: "Update requirements.txt"
           team-reviewers: dependency-manager
--- a/.github/workflows/update-requirements.yml
+++ b/.github/workflows/update-requirements.yml
@@ -2,11 +2,12 @@ name: update-requirements
 
 on:
   push:
-    branches:
-      - release/*
     paths:
-      - 'setup.json'
+      - 'setup.*'
       - '.github/workflows/update-requirements.yml'
+  pull_request:
+    branches:
+      - master
 
 jobs:
 
@@ -95,6 +96,7 @@ jobs:
           path: requirements
 
       - name: Create Pull Request
+        if: github.repository == 'aiidateam/aiida-core'
         uses: peter-evans/create-pull-request@v2
         with:
           commit-message: "Update requirements.txt"
--- a/.github/workflows/update-requirements.yml
+++ b/.github/workflows/update-requirements.yml
@@ -2,7 +2,7 @@ name: update-requirements
 
 on:
   push:
-    branch:
+    branches:
       - release/*
     paths:
       - 'setup.json'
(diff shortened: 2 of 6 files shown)


csadorf:
I don't understand why actions is running the *update-requirements* workflow for this branch. The [workflow definition](https://github.com/aiidateam/aiida-core/actions/runs/52872992/workflow) clearly states that it is only to be executed on `release/*` branches.

.github/workflows/test-install.yml
```diff
@@ -2,7 +2,7 @@ name: test-install
 
 on:
   push:
-    branch:
```
ltalirz: Do we even need to limit the scope for this workflow?

csadorf: Yes, this is a very expensive workflow. Which is why we only run it for those branches (to protect them), and nightly.

csadorf: I see what you mean now. You are saying it might be sufficient to trigger on those paths, right?

ltalirz: Yes. If someone creates a new branch, why shouldn't it run on that branch if it changes the requirements?

csadorf: Ok, I'll change that.

ltalirz:
By the way, would you mind explaining the logic of the changes?

I.e. I guess the temporary removal of the trigger is until we've updated all PRs?

And, in general, when exactly do we want this workflow to trigger (my question concerning what to do on forks was an honest question)?

I would say, the requirements*.txt files must be updated whenever somebody changes the setup.json (even when working on his or her own fork), otherwise the CI tests using the old requirements*.txt may fail.
Ideally, the workflow would directly modify the requirements*.txt files in the same branch (without PR).

Otherwise, what do you suggest as the procedure when you realize you need to modify the setup.json?

csadorf:
> By the way, would you mind explaining the logic of the changes?
> 
> I.e. I guess the temporary removal of the trigger is until we've updated all PRs?
> 
> And, in general, when exactly do we want this workflow to trigger (my question concerning what to do on forks was an honest question)?
> 
> I would say, the requirements*.txt files must be updated whenever somebody changes the setup.json (even when working on his or her own fork), otherwise the CI tests using the old requirements*.txt may fail.
> Ideally, the workflow would directly modify the requirements*.txt files in the same branch (without PR).
> 
> Otherwise, what do you suggest as the procedure when you realize you need to modify the setup.json?

I agree, the requirements files must be updated whenever the `setup.json` file is changed. I'll revert that change.

Furthermore, I would want it to be triggered on every commit for the release branch, but I don't know how to express this. It appears that these two triggers are mutually exclusive with the current syntax.

However, with this setting:
```
on:
  push:
    branches:
      - release/*
    paths:
      - setup.*
```
I believe that this will trigger the workflow exclusively for the release/ branch and only whenever a setup-file is changed.

Maybe the solution is to use PR into master trigger, like this:
```
on:
  push:
    paths:
        - 'setup.*'
  pull-request:
    branches:
        - master
```
In which case it would be triggered for all changes to the setup* files, but also on PRs into master. @ltalirz What do you think?

ltalirz:
I think triggering on PRs to master is fine. Alternatively one could even leave this out entirely.

I'm not too worried about automating the update for new releases - this update is not even strictly necessary, and it would be super easy to trigger by changing  one minimal requirement in the setup.json.

I'm more worried about responding appropriately to changes to the setup.json, which tend to happen more often than new releases (at least over the past few months).

I'm not sure you replied to my suggestion of letting the workflow apply the update directly on the branch that is being pushed rather than making a separate PR. What do you think?

csadorf:
> I think triggering on PRs to master is fine. Alternatively one could even leave this out entirely.
> 
> I'm not too worried about automating the update for new releases - this update is not even strictly necessary, and it would be super easy to trigger by changing one minimal requirement in the setup.json.
> 

OK, but should we leave it as-is anyways?

> I'm more worried about responding appropriately to changes to the setup.json, which tend to happen more often than new releases (at least over the past few months).
> 

Agreed, which is why I'll update the trigger to run on changes to `setup*`.

> I'm not sure you replied to my suggestion of letting the workflow apply the update directly on the branch that is being pushed rather than making a separate PR. What do you think?

Sorry, I somehow missed that. That's actually a good idea. I'll implement that on this branch.

ltalirz:
> OK, but should we leave it as-is anyways?

Fine to keep for the moment!

csadorf:
@ltalirz Looks like that worked: https://github.com/aiidateam/aiida-core/pull/3839/commits/36c0f8db4333181da746b057e5ccf54231813551. :)

csadorf:
So the only potential problem is that a commit created with a GitHub action can never trigger another GitHub action run. This is actually documented and I presume done to ensure that you can't accidentally create some kind of infinite loop where actions continue to trigger other actions. So in this case, we would need to trigger another run manually before being able to merge the PR. I assume that this is going to be fine in practice, but I would like to hear some perspective.

ltalirz:
> @ltalirz Looks like that worked: 36c0f8d. :)

Great, that's what we want :-)

> So the only potential problem is that a commit created with a GitHub action can never trigger another GitHub action run. This is actually documented and I presume done to ensure that you can't accidentally create some kind of infinite loop where actions continue to trigger other actions. So in this case, we would need to trigger another run manually before being able to merge the PR. I assume that this is going to be fine in practice, but I would like to hear some perspective.

I see, makes sense.

In our case, I think this is not an issue, since the requirements that are being put in are tested in the same workflow just before (which is the only change we're making).
Unless we have some mistake in our logic, this should be ok.

.github/workflows/update-requirements.yml
```diff
@@ -94,10 +95,24 @@ jobs:
           name: requirements.txt
           path: requirements
 
+      - name: Commit requirements files
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add .
```
ltalirz: actually, I would be a bit more specific here and add only the `requirements/` folder.
other things might be added to the repo when the tests are running

csadorf: Which is why this step is executed in a separate job and only the artifact is downloaded (and thus added) to the repository. We can still be more explicit here if you like.

ltalirz: You're right - in that case it's not mandatory... 
we could still add it to make the purpose explicit

.github/workflows/update-requirements.yml
```diff
@@ -94,10 +95,24 @@ jobs:
...
+          git commit -m "Update requirements files." -a || echo "Nothing to update"
+
+      - name: Push changes
+        uses: ad-m/github-push-action@v0.5.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: ${{ github.ref }}
+
       - name: Create Pull Request
```
ltalirz:
```suggestion
      - name: Create Pull Request (since update via push failed)
```
(if I understand correctly the purpose of this step)

PR #3842, opened 2020-03-10
Remove on pr trigger

ltalirz:


ltalirz:
Ok, it seems like the branch rules of the main repo don't accept/see workflows run before the PR was made.
In that case, I guess we need to keep the `pull_request` trigger.

PR #3846, opened 2020-03-12
Use actions/checkout@v2 (instead of @master)

CasperWA:
`master` was used due to `actions/checkout@v2` not working with `codecov/codecov-action@v1`.

An issue was raised at actions/checkout#180


CasperWA:
It seems to work, actions/checkout#180 can be closed.

PR #3847, opened 2020-03-13
Dm/revise update requirements workflow

csadorf:
This PR adds a command to the dependency_management script to check the requirements files and a step to the CI workflow which will execute that check. In addition to failing the CI checks, a comment will be issued on to the commit itself, that explains what the problem is and how to proceed.

The update-requirements workflow is now only manually triggered. Triggering will be achieved via  a slash-command comment on the PR which is picked up by the aiida-bot.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,8 +97,7 @@ jobs:
           If this commit is part of a pull request, you can automatically update the requirements by
           commenting with '/update-requirements'.
 
-          Click [here](https://github.com/aiidateam/aiida-core/wiki/AiiDA-Dependency-Management) for
-          more information.
+          Click [here](https://github.com/aiidateam/aiida-core/wiki/AiiDA-Dependency-Management) for more information.
 
   tests:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,42 @@ jobs:
       run:
         pre-commit run --all-files || ( git status --short ; git diff ; exit 1 )
 
+  check-requirements:
+
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+
+    - name: Install dm-script dependencies
+      run: pip install packaging==20.3 click~=7.0 pyyaml~=5.1 toml
+
+    - name: Check requirements files
+      run: python ./utils/dependency_management.py check-requirements
+
+    - name: Create commit comment
+      if: failure()
+      uses: peter-evans/commit-comment@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        path: setup.json
+        body: |
+          It appears that at least one of the environments defined in the requirements files
+          ('requirements/*.txt') is not meeting the dependencies specified in the 'setup.json' file.
+          These files are used for testing, so it is important that they are updated.
+
+          If this commit is part of a pull request, you can automatically update the requirements by
+          commenting with '/update-requirements'.
+
+          Click [here](https://github.com/aiidateam/aiida-core/wiki/AiiDA-Dependency-Management) for
+          more information.
+
   tests:
 
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
         body: |
           It appears that at least one of the environments defined in the requirements files
           ('requirements/*.txt') is not meeting the dependencies specified in the 'setup.json' file.
-          These files are used for testing, so it is important that they are updated.
+          These files define the environment for continuous integration tests, so it is important that they are updated.
 
           If this commit is part of a pull request, you can automatically update the requirements by
           commenting with '/update-requirements'.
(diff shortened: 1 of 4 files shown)


csadorf:
See https://github.com/csadorf/aiida-core/pull/12 for an example on how this is expected to work.

.github/workflows/ci.yml
```diff
@@ -64,6 +64,41 @@ jobs:
       run:
         pre-commit run --all-files || ( git status --short ; git diff ; exit 1 )
 
+  check-requirements:
```
ltalirz: Why is this not part of the general validation of the requirements?

.github/workflows/ci.yml
```diff
@@ -64,6 +64,41 @@ jobs:
...
+    - name: Create commit comment
+      if: failure()
+      uses: peter-evans/commit-comment@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        path: setup.json
+        body: |
+          It appears that at least one of the environments defined in the requirements files
+          ('requirements/*.txt') is not meeting the dependencies specified in the 'setup.json' file.
+          These files are used for testing, so it is important that they are updated.
```
ltalirz:
```suggestion
          These files define the environment for continuous integration tests, so it is important that they are updated.
```

utils/dependency_management.py
```diff
@@ -77,6 +78,24 @@ def _setuptools_to_conda(req):
     return Requirement.parse(str(req))
 
 
+class _Entry:
```
ltalirz: instead of all these underscores, one could also just use  `__all__` to expose just the classes/functions you want to expose

utils/dependency_management.py
```diff
@@ -311,6 +330,48 @@ def validate_all(ctx):
...
+@click.argument('extras', nargs=-1)
+def check_requirements(extras):
+    """Check the 'requirements/*.txt' files.
+
+    Checks that the environments specified in the requirements files
+    match all the dependencies specified in 'setup.json.
+
+    The arguments allow to specify which 'extra' requirements to expect.
+    Use DEFAULT to expect all extra requirements that are to be expected
+    by default.
```
ltalirz:
```suggestion
    Use 'DEFAULT' to select 'atomic_tools', 'docs', 'notebook', 'rest', and 'testing'.
```

utils/dependency_management.py
```diff
@@ -311,6 +330,48 @@ def validate_all(ctx):
...
+
+        with open(fn_req) as req_file:
+            working_set = list(_parse_working_set(req_file))
+            installed = {req for req in required for entry in working_set if entry.fulfills(req)}
+
+        not_installed = required.difference(installed)
+        if not_installed:  # switch to assignment expression after yapf supports 3.8
+            raise DependencySpecificationError(
+                f"Environment specified in '{fn_req.relative_to(ROOT)}' misses matches for:\n" +
+                '\n'.join(' - ' + str(f) for f in not_installed)
```
ltalirz: interesting... didn't know list comprehensions worked without brackets!

utils/dependency_management.py
csadorf: This is a stand-alone script, so there is not anything exposed. The underscores are merely used to guide the reader's eye. Rather than introducing `__all__`, we can also remove all underscores.

utils/dependency_management.py
csadorf: :)

csadorf:
> One question: Why is there no checkmark on the last commit after you did `/run-ci`?

That's because I disabled the `/run-ci` slash-command for now, because it would run on the default branch instead of the desired branch. I am able to run the `update-requirements` workflow on the correct branch, because I add the correct [HEAD reference as a payload](https://github.com/aiidateam/aiida-bot-app/blob/master/index.js#L26) to the repository-dispatch command. The update-requirements branch then picks that up and [checks it out](https://github.com/csadorf/aiida-core/blob/9b78a01ac10822b83b3a5413a9ffa34dd67ac86c/.github/workflows/update-requirements.yml#L21-L23).

However, our CI workflow uses checkout without any arguments. That means it will checkout the correct branch if triggered by a `push` or `pull_request` event, however for a `repository_dispatch` event the workflow's default branch is the repository's default branch (`develop`).

If we wanted to enable this, we would need to either find some way to use the payload conditionally or alternatively duplicate the workflow so that there is one triggered by push/pr and one triggered only by repository_dispatch.

ltalirz:
Ok, so let me try to walk through the basic process from a developer point of view:

 1. developers push to their fork a commit changing setup.json, such that the requirements/ files are now out of date 
 2. the tests fail + they get a comment notifying them to update the requirements
    * where does this comment appear?
    * is there a way for them to fix their branch before making a PR?
 3. they make a PR, the tests fail and same comment is added to the PR
 4. they write `/update-requirements`; a new commit appears with updated requirements
 5. ? Do they pull the changes, make some trivial change and push again?

This is just for me to understand, so that I can explain it to the next guy.

csadorf:
>    * where does this comment appear?

The comment is attached to the commit, specifically to the `setup.json` file of the commit. It will appear on the GitHub commit page, as part of a PR conversation, and the author is notified (on the website and probably also via email, but depends on the notification settings).

>    * is there a way for them to fix their branch before making a PR?

If they update the requirements files in a way that it satisfies constraints of `setup.json` and all tests pass, then yes, they can fix it beforehand. For that, the developer would likely need to install aiida-core into a clean environment for each supported Python version and make sure that tests pass. That is not a recommended workflow. 

> 3. they make a PR, the tests fail and same comment is added to the PR

Right, GitHub is just smart enough to add it to the PR conversation.

>  4. they write `/update-requirements`; a new commit appears with updated requirements

Yes.

>  5. ? Do they pull the changes, make some trivial change and push again?

Unless we find a better way to trigger the re-run of the CI workflow.

> This is just for me to understand, so that I can explain it to the next guy.

I will also update the wiki once this is through.

PR #3849, opened 2020-03-16
Do not restrict numpy to '<1.18'.

csadorf:
Previous issues with the broken release should be resolved now.

### DIFF
--- a/docs/requirements_for_rtd.txt
+++ b/docs/requirements_for_rtd.txt
@@ -19,7 +19,7 @@ graphviz~=0.13
 ipython~=7.0
 jinja2~=2.10
 kiwipy[rmq]~=0.5.1
-numpy<1.18,~=1.17
+numpy~=1.17
 paramiko~=2.6
 pg8000~=1.13
 pgtest>=1.3.1,~=1.3
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
 - ipython~=7.0
 - jinja2~=2.10
 - kiwipy[rmq]~=0.5.1
-- numpy<1.18,~=1.17
+- numpy~=1.17
 - paramiko~=2.6
 - pika~=1.1
 - plumpy~=0.14.5
--- a/setup.json
+++ b/setup.json
@@ -33,7 +33,7 @@
     "ipython~=7.0",
     "jinja2~=2.10",
     "kiwipy[rmq]~=0.5.1",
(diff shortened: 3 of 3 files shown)


csadorf:
@CasperWA Is there something wrong with our coverage setup? Why is it complaining about decreased coverage for this particular PR?

CasperWA:
> @CasperWA Is there something wrong with our coverage setup? Why is it complaining about decreased coverage for this particular PR?

Well it's not really a huge complain. The coverage change 0.00% ...
We can maybe set it otherwise so the "patch change" doesn't complain.

In any case, you can see [here](https://codecov.io/gh/aiidateam/aiida-core/pull/3849/changes) what has changed (link grabbed from the `codecov` comment above.
It seems it is suddenly not testing 5 lines in `aiida/orm/utils/log.py`.

csadorf:
> > @CasperWA Is there something wrong with our coverage setup? Why is it complaining about decreased coverage for this particular PR?
> 
> Well it's not really a huge complain. The coverage changes 0.00% ...
> We can maybe set it otherwise so the ~"patch change"~ (sorry, project change) doesn't complain.
> 
> In any case, you can see [here](https://codecov.io/gh/aiidateam/aiida-core/pull/3849/changes) what has changed (link grabbed from the `codecov` comment above).
> It seems it is suddenly not testing 5 lines in `aiida/orm/utils/log.py`.

Well, I think it *should* complain if the coverage decreases. I just don't understand why it does in this particular case.

CasperWA:
> Well, I think it _should_ complain if the coverage decreases. I just don't understand why it does in this particular case.

To a degree, sure.

In any case, I would guess that the newer version of `numpy` affects the tests to not run those lines of code in `log.py`?

csadorf:
> r version of `numpy` affects the tests to not run those lines

As I was able to show with #3850 , this unexpected change in coverage reporting is not related to any particular changes in this specific commit. But even if it was, the change of the numpy dependency specification is actually not reflected in the coverage report, because the tests are executed for the environments specified in the requirements-files, which have not been updated (yet).

csadorf:
/update-requirements (this will probably not work on the fork, but we can try...)

ltalirz:
@csadorf Is this ready for review?

csadorf:
> @csadorf Is this ready for review?

@ltalirz Unfortunately, we currently have no way to run the update-requirements workflow on forks, so I'll close this PR and probably create a new one.

PR #3850, opened 2020-03-16
Inconsequential 'test' commit to test the codecov coverage checks.

csadorf:


csadorf:
@CasperWA Something seems to be off since this particular commit should not display any change in coverage. Maybe this is related to our recent changes to the CI configuration? I have tried to look into the codecov report, but couldn't find the origin of this problem. Maybe we can "reset" codecov somehow?

CasperWA:
> (...) Maybe we can "reset" codecov somehow?

The only way to do this, would be to re-run the CI workflows on the latest `develop`.
But it may be something from the cron-jobs as well?

I tried to go poking into the latest runs, but found that GitHub is crashing a bit on me at the moment. But I wanted to see if the cronjobs successfully upload coverages and what they look like. And also see if I could re-run the CI for the latest one on `develop`, which would be the one that @ltalirz merged into `develop` from your previous PR.

csadorf:
> > (...) Maybe we can "reset" codecov somehow?
> 
> The only way to do this, would be to re-run the CI workflows on the latest `develop`.
> But it may be something from the cron-jobs as well?
> 
> I tried to go poking into the latest runs, but found that GitHub is crashing a bit on me at the moment. But I wanted to see if the cronjobs successfully upload coverages and what they look like. 

The cron jobs are executed for the ["test-install"](https://github.com/aiidateam/aiida-core/blob/develop/.github/workflows/test-install.yml) workflow, which does not upload any coverage reports.

csadorf:
It looks like this is what codecov calls an ["unexpected coverage change"](https://docs.codecov.io/docs/unexpected-coverage-changes#section-reasons-for-unexpected-changes). Looking at their list of possible reasons I suspect that this is either related to some failure to upload some reports or indeed some change in execution paths due our dependencies. Either way, probably not something to be too concerned about, it's just very annoying that it fails some of our builds.

ltalirz:
Can this PR be closed?

PR #3854, opened 2020-03-20
test using Computer uuid in Code setup

ltalirz:
Test that Codes can be set up while specifying the Computer via its uuid
instead of its label.

### DIFF
--- a/tests/cmdline/commands/test_code.py
+++ b/tests/cmdline/commands/test_code.py
@@ -66,18 +66,23 @@ class TestVerdiCodeSetup(AiidaTestCase):
         self.assertIsInstance(Code.get_from_string('{}'.format(label)), Code)
 
     def test_noninteractive_remote(self):
-        """Test non-interactive remote code setup."""
+        """Test non-interactive remote code setup.
+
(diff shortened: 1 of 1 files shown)


ltalirz:
@giovannipizzi Turns out this works already, even via the verdi cli

sphuber:
> @giovannipizzi Turns out this works already, even via the verdi cli

Ahhh, the beauty of the reusable `verdi` parameters :smirk: 

It is of course always good to test more, but I am not sure that this addition is necessary. The reason it already works is because the computer option in `verdi code setup` is defined through `@options_code.COMPUTER()`. This option has the type `IdentifierParamType` which therefore automatically implements support for label, UUID _and_ pk. So if we really wanted to be complete, we should also test that the pk can be used in `verdi code setup`. But this is exactly why I don't think we have to, because this is rather something that should be tested on the param type. Otherwise we would have to start testing that label/UUID/pk works for every command parameter that currently accepts it. Feels a bit redundant to be honest.

ltalirz:
I think you're right - it was more of a test for myself since we're going to need this for the code/computer registry.
Since this does not seem to increase coverage, no need to merge this.

PR #3857, opened 2020-03-21
Add the `-l/--limit` option to `verdi group show`

### LINKED ISSUE #3855: Add `limit` option to `verdi group show`
This used to be there at some point but is no longer the case, however, it is a very useful flag to have.

sphuber:
Fixes #3855

### DIFF
--- a/aiida/cmdline/commands/cmd_group.py
+++ b/aiida/cmdline/commands/cmd_group.py
@@ -123,6 +123,7 @@ def group_description(group, description):
 
 @verdi_group.command('show')
 @options.RAW(help='Show only a space-separated list of PKs of the calculations in the group')
+@options.LIMIT()
 @click.option(
     '-u',
     '--uuid',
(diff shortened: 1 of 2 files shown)


sphuber:
Can @CasperWA or @csadorf maybe explain why codecov reports that the coverage decreases? Why do these changes affect `aiida.orm.utils.log` at all?

csadorf:
@sphuber I am puzzled by that myself and already inquired about that as part of this test PR: https://github.com/aiidateam/aiida-core/pull/3850

sphuber:
> @sphuber I am puzzled by that myself and already inquired about that as part of this test PR: #3850

I see, thanks for the comment. If I understand your conclusion in that PR correctly, it means that this can happen "randomly" and there is no way to prevent this systematically? I guess that is also why the `codecov` actions are not yet marked as required for passing a PR, correct? I take it that will have to remain if the chance remains that it can fail arbitrarily, which would block an otherwise perfectly fine PR from being merged.

csadorf:
@sphuber I don't think that this really happens randomly. At least I have not yet observed this behavior in other projects that used codecov. I presume that something went wrong during the initial setup and that this weird behavior will eventually go away. We should revisit in a few weeks if it doesn't.

PR #3875, opened 2020-03-28
fix interface of run_api for running REST API

ltalirz:
The current `run_api` interface for running the AiiDA REST API had
unnecessarily many required arguments, making it complicated to use in WSGI
scripts, e.g.:

```python
from aiida.restapi import api
from aiida.restapi.run_api import run_api
import aiida.restapi

CONFIG_DIR = os.path.join(os.path.split(
                  os.path.abspath(aiida.restapi.__file__))[0], 'common')

(app, api) = run_api(
    api.App,
    api.AiidaApi,
    hostname="localhost",
    port=5000,
    config=CONFIG_DIR,
    debug=False,
    wsgi_profile=False,
    hookup=False,
    catch_internal_server=False
)
```
While all but the first two parameters are keyword arguments, the code
would actually crash if some of the keyword arguments are not provided.

In reality, there is no reason to force people to specify *any* parameters
whatsoever. One should simply be able to call `run_api()`.
This commit accomplishes this by defining the appropriate default
values (without introducing breaking changes).

### DIFF
--- a/aiida/cmdline/commands/cmd_restapi.py
+++ b/aiida/cmdline/commands/cmd_restapi.py
@@ -12,60 +12,54 @@ This allows to hook-up the AiiDA built-in RESTful API.
 Main advantage of doing this by means of a verdi command is that different
 profiles can be selected at hook-up (-p flag).
 """
-import os
 
 import click
 
-import aiida.restapi
 from aiida.cmdline.commands.cmd_verdi import verdi
 from aiida.cmdline.params.options import HOSTNAME, PORT
-
-CONFIG_DIR = os.path.join(os.path.split(os.path.abspath(aiida.restapi.__file__))[0], 'common')
+from aiida.restapi.common import config
 
 
 @verdi.command('restapi')
-@HOSTNAME(default='127.0.0.1')
-@PORT(default=5000)
+@HOSTNAME(default=config.CLI_DEFAULTS['HOST_NAME'])
+@PORT(default=config.CLI_DEFAULTS['PORT'])
 @click.option(
     '-c',
     '--config-dir',
     type=click.Path(exists=True),
-    default=CONFIG_DIR,
-    help='the path of the configuration directory'
+    default=config.CLI_DEFAULTS['CONFIG_DIR'],
+    help='Path to the configuration directory'
 )
-@click.option('--debug', 'debug', is_flag=True, default=False, help='run app in debug mode')
+@click.option('--debug', 'debug', is_flag=True, default=config.APP_CONFIG['DEBUG'], help='Enable debugging')
 @click.option(
     '--wsgi-profile',
-    'wsgi_profile',
     is_flag=True,
-    default=False,
-    help='to use WSGI profiler middleware for finding bottlenecks in web application'
+    default=config.CLI_DEFAULTS['WSGI_PROFILE'],
+    help='Whether to enable WSGI profiler middleware for finding bottlenecks'
 )
-@click.option('--hookup/--no-hookup', 'hookup', is_flag=True, default=True, help='to hookup app')
+@click.option(
+    '--hookup/--no-hookup',
+    'hookup',
+    is_flag=True,
+    default=config.CLI_DEFAULTS['HOOKUP_APP'],
+    help='Hookup app to flask server'
+)
 def restapi(hostname, port, config_dir, debug, wsgi_profile, hookup):
     """
     Run the AiiDA REST API server.
 
     Example Usage:
 
-        \b
-        verdi -p <profile_name> restapi --hostname 127.0.0.5 --port 6789 --config-dir <location of the config.py file>
-        --debug --wsgi-profile --hookup
+        verdi -p <profile_name> restapi --hostname 127.0.0.5 --port 6789
     """
-    from aiida.restapi.api import App, AiidaApi
     from aiida.restapi.run_api import run_api
 
-    # Construct parameter dictionary
-    kwargs = dict(
-        prog_name='verdi-restapi',
+    # Invoke the runner
+    run_api(
         hostname=hostname,
         port=port,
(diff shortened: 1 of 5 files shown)


aiida/cmdline/commands/cmd_restapi.py
```diff
@@ -12,60 +12,54 @@
...
     '-c',
     '--config-dir',
     type=click.Path(exists=True),
-    default=CONFIG_DIR,
-    help='the path of the configuration directory'
+    default=config.CLI_DEFAULTS['CONFIG_DIR'],
+    help='Path to the configuration directory'
 )
-@click.option('--debug', 'debug', is_flag=True, default=False, help='run app in debug mode')
+@click.option('--debug', 'debug', is_flag=True, default=config.APP_CONFIG['DEBUG'], help='Enable debugging')
```
sphuber: Should this default maybe also come from `CLI_DEFAULTS`?

.ci/workchains.py
```diff
@@ -68,8 +68,8 @@ def a_magic_unicorn_appeared(self, node):
     @process_handler(priority=400, exit_codes=ArithmeticAddCalculation.exit_codes.ERROR_NEGATIVE_NUMBER)
     def error_negative_sum(self, node):
         """What even is a negative number, how can I have minus three melons?!."""
-        self.ctx.inputs.x = Int(abs(node.inputs.x.value))
-        self.ctx.inputs.y = Int(abs(node.inputs.y.value))
+        self.ctx.inputs.x = Int(abs(node.inputs.x.value))  # pylint: disable=invalid-name
+        self.ctx.inputs.y = Int(abs(node.inputs.y.value))  # pylint: disable=invalid-name
```
sphuber: Sure this is fine, just wondering how this came up in this PR? I don't think you've removed this file from the pre-commit ignore right?

docs/source/conf.py
```diff
@@ -49,7 +49,10 @@
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['sphinx.ext.autodoc', 'sphinx.ext.doctest', 'sphinx.ext.todo', 'sphinx.ext.coverage', 'sphinx.ext.imgmath', 'sphinx.ext.ifconfig', 'sphinx.ext.viewcode', 'IPython.sphinxext.ipython_console_highlighting', 'IPython.sphinxext.ipython_directive', 'sphinxcontrib.contentui', 'aiida.sphinxext']
+extensions = ['sphinx.ext.intersphinx', 'sphinx.ext.autodoc', 'sphinx.ext.doctest', 'sphinx.ext.todo',
+              'sphinx.ext.coverage', 'sphinx.ext.imgmath', 'sphinx.ext.ifconfig', 'sphinx.ext.viewcode',
+              'IPython.sphinxext.ipython_console_highlighting', 'IPython.sphinxext.ipython_directive',
+              'sphinxcontrib.contentui', 'aiida.sphinxext']
 ipython_mplbackend = ''
```
sphuber: I see some of the changes that I think you wanted to address in #3876 have snuck in. Maybe it is best to remove these here and only apply this in #3876

aiida/cmdline/commands/cmd_restapi.py
ltalirz: Well, the variable was already in `APP_CONFIG`, so I did not want to change this.

.ci/workchains.py
ltalirz: That's a very good question.
Notice that I introduced this in my second commit because the pre-commit hook failed.
I suspect this is because some dependency of dependency changed...

docs/source/conf.py
ltalirz: Well, not quite - I added some new references to `flask`, which then caused the docs build to fail, which then caused me to add an intersphinx pointer to the flask docs here.
Which then caused me to think about adding also the docs of the python standard library (which is #3876)

sphuber: But then don't we just want to first merge #3876 which should be fine without this PR and then you rebase _this_ PR on top of that, which then only includes relevant changes, no?

aiida/cmdline/commands/cmd_restapi.py
sphuber: Sure, but it seems to be a different variable then. It is not actually a default for the cli, you are just co-opting it. If you want to keep in in `APP_CONFIG` then I think you should simply add a separate one to `CLI_DEFAULT`. Or should these two values always be the same anyway?

docs/source/conf.py
ltalirz: well, it's just extra effort and I think it's ok to keep the flask intersphinx mapping in this PR - otherwise I need to add another nitpick here and remove it again.
if you want me to do this, let me know.

aiida/cmdline/commands/cmd_restapi.py
ltalirz: > Or should these two values always be the same anyway?

There should only be one default value.

Perhaps the issue is rather that `CLI_DEFAULT` is not the best name.
I chose it to keep the default values together that I needed to add for the cli + for run_api, so they aren't really just meant for the cli.
I could also call them `RUN_API_DEFAULTS` or something but I'm not sure that's really better.
Let me know.

sphuber: I'm not familiar enough with the REST API to make that call. Fine to keep it as is

docs/source/conf.py
sphuber: I played around a bit with the intersphinx and nitpicks, and rebased your other PR and merged it. This PR is now also rebased and you already addressed my comments, so will also merge this. Thanks a lot!

ltalirz:
Great, thanks a lot Seb!

PR #3876, opened 2020-03-28
Add python intersphinx mapping

ltalirz:
Add intersphinx mapping to python standard library documentation, i.e.
things like :py:exc:`ValueError` now link directly there.
This also allows to cut down the "nitpick exceptions" quite
dramatically.

Note: I will rebase this PR after #3875 is merged

### DIFF
--- a/aiida/cmdline/params/options/interactive.py
+++ b/aiida/cmdline/params/options/interactive.py
@@ -106,7 +106,7 @@ class InteractiveOption(ConditionalOption):
         return None
 
     def _get_default(self, ctx):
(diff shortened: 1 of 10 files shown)


sphuber:
This is great @ltalirz , thanks. I think it would maybe also be good to add `plumpy` and `kiwipy` or is there something preventing us from doing that?

ltalirz:
> I think it would maybe also be good to add plumpy and kiwipy or is there something preventing us from doing that?

Right - there was no docs for plumpy on RTD (fixed it just now), and the one from kiwipy was not linked from the readme (opened PRs).
I'll have a go at adding them now as well.

Issue #3051, opened 2019-06-19
Randomly failing test (test_daemon_start_number)


borellim:
```
======================================================================
FAIL: test_daemon_start_number (aiida.backends.tests.cmdline.commands.test_daemon.TestVerdiDaemon)
Test `verdi daemon start` with a specific number of workers.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/aiidateam/aiida_core/aiida/backends/tests/cmdline/commands/test_daemon.py", line 67, in test_daemon_start_number
    self.assertIn('info', worker_response, worker_response)
[...]

PR #3883, opened 2020-04-01
Skip unit test 'test_daemon_start_number' temporarily.

csadorf:
This unit test currently fails non-deterministically.

Addresses, but does not resolve issue #3051.


csadorf:
Coverage obviously decreases, but otherwise this should be ready to merge.

PR #3886, opened 2020-04-01
Prevent nodes without registered entry points from being stored

### LINKED ISSUE #3457: Decide behavior when storing `Data` sub classes within `__main__`
Consider the following:
```
class SubData(Data):
    pass

data = SubData().store()
```
This will store the node just fine, but its `node_type` will be `__main__.SubData.` if performed in a script or interactive shell. Loading the node will fail though, without a fallback:
```
load_node(data.pk)
```
will raise
```
EntryPointError                           Traceback (most recent call last)
<ipython-input-33-0a20738a6046> in <module>()
----> 1 load_node(d.pk)

/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/utils/__init__.py in load_node(identifier, pk, uuid, label, sub_classes, query_with_dashes)
    207         label=label,
    208         sub_classes=sub_classes,
--> 209         query_with_dashes=query_with_dashes
    210     )

/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/utils/__init__.py in load_entity(entity_loader, identifier, pk, uuid, label, sub_classes, query_with_dashes)
     81 
     82     return entity_loader.load_entity(
---> 83         identifier, identifier_type, sub_classes=sub_classes, query_with_dashes=query_with_dashes
     84     )
     85 

/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/utils/loaders.py in load_entity(cls, identifier, identifier_type, sub_classes, query_with_dashes)
    218 
    219         try:
--> 220             entity = builder.one()[0]
    221         except MultipleObjectsError:
    222             error = 'multiple {} entries found with {}<{}>'.format(classes, identifier_type, identifier)

/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/querybuilder.py in one(self)
   2127         from aiida.common.exceptions import MultipleObjectsError, NotExistent
   2128         self.limit(2)
-> 2129         res = self.all()
   2130         if len(res) > 1:
   2131             raise MultipleObjectsError('More than one result was found')

/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/querybuilder.py in all(self, batch_size)
   2202         :returns: a list of lists of all projected entities.
   2203         """
-> 2204         return list(self.iterall(batch_size=batch_size))
   2205 
   2206     def dict(self, batch_size=None):

/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/querybuilder.py in iterall(self, batch_size)
   2162             # Convert to AiiDA frontend entities (if they are such)
   2163             for i, item_entry in enumerate(item):
-> 2164                 item[i] = self.get_aiida_entity_res(item_entry)
   2165 
   2166             yield item

/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/querybuilder.py in get_aiida_entity_res(value)
   2054         """
   2055         try:
-> 2056             return convert.get_orm_entity(value)
   2057         except TypeError:
   2058             return value

/home/sphuber/.virtualenvs/aiida_dev/lib/python3.6/functools.py in wrapper(*args, **kw)
    805                             '1 positional argument')
    806 
--> 807         return dispatch(args[0].__class__)(*args, **kw)
    808 
    809     funcname = getattr(func, '__name__', 'singledispatch function')

/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/convert.py in _(backend_entity)
     81 def _(backend_entity):
     82     from .utils.node import load_node_class
---> 83     node_class = load_node_class(backend_entity.node_type)
     84     return node_class.from_backend_entity(backend_entity)
     85 

/home/sphuber/code/aiida/env/dev/aiida-core/aiida/orm/utils/node.py in load_node_class(type_string)
     81         return load_entry_point('aiida.node', entry_point_name)
     82 
---> 83     raise exceptions.EntryPointError('unknown type string {}'.format(type_string))
     84 
     85 

EntryPointError: unknown type string __main__.SubData.
```

sphuber:
Fixes #3457

Up until now it was possible to store instances of `Node` subclasses
that do not have a registered entry point. Imagine for example the
definition of a `Data` subclass called `SubClass` in a shell and an
instance being stored. The node type string will be:

  `__main__.SubClass.`

When trying to load this node at a later point in time, the type string
can of course not be resolved to an importable class and the loader
would raise an exception.

The first change is that this exception is now turned into a warning and
the loader falls back onto the `Data` class. Note that we do not use the
`Node` class for this as this base class is also not storable and so the
ORM logic is potentially ill-defined for instances of the base `Node`.

Secondly, we now add a check to `Node.store` to make sure that the class
corresponds to a registered entry point. If this is not the case, the
store method will raise `StoringNotAllowed`.

One unit test was deleted because its implementation was incorrect and
was not actually testing what it was intending to test. Besides intended
functionality is now covered by other tests.

### DIFF
--- a/aiida/orm/utils/node.py
+++ b/aiida/orm/utils/node.py
@@ -12,6 +12,7 @@ from abc import ABCMeta
 import logging
 import math
 import numbers
+import warnings
 from collections.abc import Iterable, Mapping
 
 from aiida.common import exceptions
@@ -70,7 +71,14 @@ def load_node_class(type_string):
         entry_point_name = strip_prefix(base_path, 'nodes.')
         return load_entry_point('aiida.node', entry_point_name)
 
-    raise exceptions.EntryPointError('unknown type string {}'.format(type_string))
+    # At this point we really have an anomalous type string. At some point, storing nodes with unresolvable type strings
+    # was allowed, for example by creating a sub class in a shell and then storing an instance. Attempting to load the
+    # node then would fail miserably. This is now no longer allowed, but we need a fallback for existing cases, which
+    # should be rare. We fallback on `Data` and not `Node` because bare node instances are also not storable and so the
+    # logic of the ORM is not well defined for a loaded instance of the base `Node` class.
+    warnings.warn('unknown type string `{}`, falling back onto `Data` class'.format(type_string))  # pylint: disable=no-member
+
+    return Data
 
 
 def get_type_string_from_class(class_module, class_name):
@@ -247,13 +255,9 @@ def clean_value(value):
 
 
 class AbstractNodeMeta(ABCMeta):  # pylint: disable=too-few-public-methods
-    """
-    Some python black magic to set correctly the logger also in subclasses.
-    """
+    """Some python black magic to set correctly the logger also in subclasses."""
 
-    # pylint: disable=arguments-differ,protected-access,too-many-function-args
-
-    def __new__(mcs, name, bases, namespace):
+    def __new__(mcs, name, bases, namespace):  # pylint: disable=arguments-differ,protected-access,too-many-function-args
         newcls = ABCMeta.__new__(mcs, name, bases, namespace)
         newcls._logger = logging.getLogger('{}.{}'.format(namespace['__module__'], name))
 
--- a/aiida/plugins/entry_point.py
+++ b/aiida/plugins/entry_point.py
@@ -186,6 +186,7 @@ def load_entry_point_from_string(entry_point_string):
     group, name = parse_entry_point_string(entry_point_string)
     return load_entry_point(group, name)
 
+
 def load_entry_point(group, name):
     """
     Load the class registered under the entry point for a given name and group
@@ -244,6 +245,7 @@ def get_entry_points(group):
     """
     return [ep for ep in ENTRYPOINT_MANAGER.iter_entry_points(group=group)]
 
+
 @functools.lru_cache(maxsize=None)
 def get_entry_point(group, name):
     """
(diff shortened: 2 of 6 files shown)


aiida/orm/utils/node.py
```diff
@@ -70,7 +71,14 @@ def load_node_class(type_string):
         entry_point_name = strip_prefix(base_path, 'nodes.')
         return load_entry_point('aiida.node', entry_point_name)
 
-    raise exceptions.EntryPointError('unknown type string {}'.format(type_string))
+    # At this point we really have an anomolous type string. At this point, storing nodes with unresolvable type strings
```
greschd, line 74: There are two typos here:
- "anomalous" type string (nit-pick, I know)
- probably, it should be "At _one_ point, [...] was allowed"

aiida/plugins/entry_point.py
```diff
@@ -332,3 +334,27 @@ def is_valid_entry_point_string(entry_point_string):
...
+    :param groups: optionally consider only these entry point groups to look for the class
+    :return: boolean, True if the class is a registered entry point, False otherwise.
+    """
+    if groups is None:
+        groups = list(entry_point_group_to_module_path_map.keys())
+
+    registered_entry_points = []
+
+    for group in groups:
+        for entry_point in ENTRYPOINT_MANAGER.iter_entry_points(group):
```
greschd, line 357: Why do we build up all the registered entry points first, instead of just returning `True` as soon as we encounter a matching one (and `False` at the end)? 

Since the total number of entry points is not gigantic it doesn't make much of a difference, though.

sphuber: I have changed it now, but I had the original implementation thinking that this would help with the caching added by the decorator. But to be honest I am not sure how this works. Since we are calling this function now at every `Node.store()` I want to make sure that the overhead is acceptable and we don't impact the performance too much. I will try and run some actual benchmarks.

greschd: The `functools.lru_cache` just checks if the input arguments are the same, and then returns an existing result (if any). Since `registered_entry_points` was just a local variable inside the function here, that's not preserved in any way.

To cache that part, we would need to put the part setting up the registered entry points into a separate function, and decorate _that_ with `lru_cache`.

sphuber: But that wouldn't matter, right. If we call the function for the same plugin class, so with the same function arguments, it will instantly return the same boolean result, without having to go through the function body again? So essentially that is exactly what we want. I was indeed trying to prevent from having to implement our own caching of the registered entry point mapping, even though it would be not that much work.

greschd: Yes, I think this is fine - my comment was more a hypothetical, if you _also_ wanted to cache the full list of AiiDA-related entry points.

To be honest I don't think the performance difference to the previous implementation is even that big now, but (to me at least) this seems like the more straightforward way of doing it.

sphuber:
I have performed the following performance test on the branch of this PR and `develop`.
In a `verdi shell`:
```
%timeit [Int().store() for _ in range(100)
```
* branch: 1.72 s ± 85.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
* develop: 1.69 s ± 101 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

I feel this is acceptable. @giovannipizzi and @greschd this is ready for review

PR #3891, opened 2020-04-03
Fix the `RotatingFileHandler` configuration of the daemon logger

### LINKED ISSUE #3557: Add rotating log file handlers for the daemon log

sphuber:
Fixes #3557

The logging configuration for the daemon did already include the
definition of a `RotatingFileHandler`, however, the logs were not
actually being rolled over when the maximum size was hit. The problem
was because the argument `backupCount` was not defined. As the python
documentation of the `logging` module states:

  but if either of `maxBytes` or `backupCount` is zero, rollover never
  occurs, so you generally want to set `backupCount` to at least 1,
  and have a non-zero `maxBytes`.

Setting `backupCount` to 10, fixes the problem. The size of each file is
set to 10 MB. This should then keep at most 100 MB of log files for each
profile.

### DIFF
--- a/aiida/common/log.py
+++ b/aiida/common/log.py
@@ -175,7 +175,8 @@ def configure_logging(with_orm=False, daemon=False, daemon_log_file=None):
             'class': 'logging.handlers.RotatingFileHandler',
             'filename': daemon_log_file,
             'encoding': 'utf8',
-            'maxBytes': 100000,
+            'maxBytes': 10000000,  # 10 MB
+            'backupCount': 10,
         }
 
         for logger in config.get('loggers', {}).values():


ltalirz:
Right - one question in general would be what should determine the maximum size of a file.
If I understand correctly, then the reason we do this here is because it is needed to have a file size limit at all.

In that sense, a backup count of 1 would do the job as well.

greschd:
> 
> Right - one question in general would be what should determine the maximum size of a file.
> If I understand correctly, then the reason we introduce backup files here is because it is needed to have a file size limit at all.
> 

I think it is to have some granularity in how the backup is rolled. More, smaller, files mean that the amount of history deleted at a time is smaller for the same total log size. So the fluctuations in how much history is currently available at any given time are smaller.

sphuber:
I will merge this now and if there really is a need to make these values configurable, we can open a feature request issue to have support added for this through `verdi config`

PR #3899, opened 2020-04-06
enable loading config.yml files from URL

ltalirz:
In the view of the plans of starting online repositories that maintain
yaml configuration files for computers and codes, it becomes very useful
to be able to directly import a configuration file from a URL.


ltalirz:
This requires a minor modification of `click-config-file`, see https://github.com/phha/click_config_file/pull/19
Closing until this is figured out.

PR #3901, opened 2020-04-07
Use GitHub Actions services (for PostgreSQL and RabbitMQ)

CasperWA:
Move over to a more "clean" setup of system services in CI tests, here PostgreSQL and RabbitMQ.

This utilizes [GitHub Actions services](https://help.github.com/en/actions/configuring-and-managing-workflows/using-databases-and-service-containers), specifically integrating the Docker images for PostgreSQL and RabbitMQ.

Hence, some oblique lines in the workflow can be removed under the "Install system dependencies" step(s) as well as the PostgreSQL action.

**Note**: This does _not_ solve the issue of having a PostgreSQL super user with an empty password; the authentication method is kept to `"trust"`, i.e., allowing all connections no matter the database user password.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,18 +139,9 @@ jobs:
 
     - name: Install system dependencies
       run: |
+        sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
         sudo apt update
         sudo apt install postgresql-10 graphviz
-
-    # - name: Install system dependencies
-    #   run: |
-    #     wget -O - "https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc" | sudo apt-key add -
-    #     echo 'deb https://dl.bintray.com/rabbitmq-erlang/debian bionic erlang' | sudo tee -a /etc/apt/sources.list.d/bintray.rabbitmq.list
-    #     echo 'deb https://dl.bintray.com/rabbitmq/debian bionic main' | sudo tee -a /etc/apt/sources.list.d/bintray.rabbitmq.list
-    #     sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
-    #     sudo apt update
-    #     sudo apt install postgresql-10 rabbitmq-server graphviz
-    #     sudo systemctl status rabbitmq-server.service
 
     - name: Upgrade pip
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,13 +127,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    # - uses: CasperWA/postgresql-action@v1.2
-    #   with:
-    #     postgresql version: '10'
-    #     postgresql db: test_${{ matrix.backend }}
-    #     postgresql user: postgres
-    #     postgresql password: ''
-    #     postgresql auth: trust
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,15 +110,29 @@ jobs:
         backend: ['django', 'sqlalchemy']
         python-version: [3.5, 3.8]
 
+    services:
+      postgres:
+        image: postgres:10
+        env:
+          POSTGRES_DB: test_${{ matrix.backend }}
+          POSTGRES_PASSWORD: ''
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -127,13 +139,9 @@ jobs:
...
     - name: Install system dependencies
       run: |
-        wget -O - "https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc" | sudo apt-key add -
-        echo 'deb https://dl.bintray.com/rabbitmq-erlang/debian bionic erlang' | sudo tee -a /etc/apt/sources.list.d/bintray.rabbitmq.list
-        echo 'deb https://dl.bintray.com/rabbitmq/debian bionic main' | sudo tee -a /etc/apt/sources.list.d/bintray.rabbitmq.list
         sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
         sudo apt update
-        sudo apt install postgresql-10 rabbitmq-server graphviz
-        sudo systemctl status rabbitmq-server.service
+        sudo apt install postgresql-10 graphviz
```
sphuber, line 144: why do we still need to install postgresql-10 manually here?

CasperWA: Due to the PostgreSQL tests. It may be that we do not need to after @ltalirz has moved these tests out of the package, but I am not sure?

CasperWA: More specifically, all tests using [this class](https://github.com/aiidateam/aiida-core/blob/develop/aiida/backends/testbase.py#L207) will not run properly.

sphuber: If you are mentioning the moving out of the `PGSU` class, then that already happened. The PR was merged yesterday. Did you try to remove it and see if the tests still run? If so what was the failure?

sphuber: ah that makes sense, because there it relies on `pg_test` to create a postgres cluster on the localhost. So if the service just exposes a postgres database on some other machine, then we still have to install it manually

CasperWA: Didn't try it this time, will do that now.

CasperWA: > ah that makes sense, because there it relies on `pg_test` to create a postgres cluster on the localhost. So if the service just exposes a postgres database on some other machine, then we still have to install it manually

Yeah. My understanding was also that the `postgresql-10` package came with some tools that were needed for the tests. I originally tried to install _only_ the tools, but it was more work than just doing this and be done with it - but let me just test it quickly anyway.

csadorf: In the actions that ran for this PR we again see sporadic failure to install dependencies using apt. Since it seems that we can't get rid of those even with the improvements introduced with this PR, I think we should probably looking into either providing a custom image or [caching](https://stackoverflow.com/a/59277514). Either way, this would probably be out of scope for this PR.

CasperWA: Concerning the removal of `postgresql-10`: It is indeed not possible, see [here](https://github.com/aiidateam/aiida-core/pull/3901/checks?check_run_id=567968831#step:10:182).

CasperWA: > In the actions that ran for this PR we again see sporadic failure to install dependencies using apt. Since it seems that we can't get rid of those even with the improvements introduced with this PR, I think we should probably looking into either providing a custom image or [caching](https://stackoverflow.com/a/59277514). Either way, this would probably be out of scope for this PR.

Yeah - there's some weird stuff happening. Hmm, also since the GitHub Pages were having some maintenance, the Docs CI is failing (I'm guessing that's why).

CasperWA:
> LGTM!

Thanks! But just removed the testing commit that removed installing `postgresql-10`, automatically dismissing your review. You're welcome to re-accept 😄

.github/workflows/ci.yml
CasperWA: > In the actions that ran for this PR we again see sporadic failure to install dependencies using apt. Since it seems that we can't get rid of those even with the improvements introduced with this PR, I think we should probably looking into either providing a custom image or [caching](https://stackoverflow.com/a/59277514). Either way, this would probably be out of scope for this PR.

Since there are "only" 3 packages left (although there may be a lot of dependencies) we could download the packages and keep them under `/.ci/` or similar and install them from there? I don't know if this is a viable solution, but it would avoid the fragile `sudo apt update`.

sphuber: > Since there are "only" 3 packages left (although there may be a lot of dependencies) we could download the packages and keep them under `/.ci/` or similar and install them from there? I don't know if this is a viable solution, but it would avoid the fragile `sudo apt update`.

The `aiida-core` repository is already over 100MB, so I would be really hesitant to start vendoring entire system software packages just to make the CI work.

CasperWA: I found [this page](https://github.community/t5/GitHub-Actions/Install-dependencies-in-Ubuntu-flakes-several-times-a-wekk/td-p/51785) commenting on the issue.

PR #3903, opened 2020-04-07
Add support for subclassing in `Groups` to the `QueryBuilder`

### LINKED ISSUE #3902: Add support for querying of subclasses of `Group` in the `QueryBuilder`
PR #3882 is looking to make the `Group` class subclassable. It would be nice if the `QueryBuilder` would support the same functionality to query for subclasses, just as is implemented for the `Node` class and its subclasses.

sphuber:
Fixes #3902

This now allows to search for instances of `Group` and all its
subclasses through the `QueryBuilder` just as implemented for the `Node`
class. The subclassing is based on the `type_string` of the group
instances, which in turn is defined by the entry point name. When
querying for `Group` instances through the `QueryBuilder`:

    QueryBuilder().append(Group).all()

by default subclassing is enabled and should return all instances of a
subclass of `Group` as well. The same goes for subclasses of `Group`
themselves. Imagine a plugin defines the following entry points:

    'plugin.sub = aiida_plugin.groups:SubGroup'
    'plugin.sub.a = aiida_plugin.groups:SubAGroup'
    'plugin.sub.b = aiida_plugin.groups:SubBGroup'

The following query:

    QueryBuilder().append(SubGroup).all()

will return all instances of `SubGroup` but also of `SubAGroup` and
`SubBGroup`. This is because their entry point names both start with the
entry point name `plugin.sub` of the class that is being appended.

Note that in order for all group instances to be returned when querying
for the `Group` base class, a small hack had to be applied in the query
builder code building the filters. Note that the entry point name of
`Group` is `core` and so with the rules described above, the plugin
subclasses would never be matched since their entry point names do not
start with `core`. This is a problem and could be fixed by making the
type string of base group instances an empty string, however, this would
require using an empty string for the `Group` base class which is not
allowed in Python. To still provide the desired behavior, when querying
for `Group` instances the `type_string` filter is set to match anything.

### DIFF
--- a/tests/tools/groups/test_paths.py
+++ b/tests/tools/groups/test_paths.py
@@ -125,7 +125,7 @@ def test_walk_with_invalid_path(clear_database_before_test):
     assert [c.path for c in sorted(group_path.walk())] == expected
 
 
-def test_walk_nodes(clear_database):
+def test_walk_nodes(clear_database_before_test):
     """Test the ``GroupPath.walk_nodes()`` function."""
     group, _ = orm.Group.objects.get_or_create('a')
     node = orm.Data()
@@ -136,7 +136,6 @@ def test_walk_nodes(clear_database):
     assert [(r.group_path.path, r.node.attributes) for r in group_path.walk_nodes()] == [('a', {'i': 1, 'j': 2})]
 
 
-@pytest.mark.skip('Reenable when subclassing in the query builder is implemented (#3902)')
 def test_cls(clear_database_before_test):
     """Test that only instances of `cls` or its subclasses are matched by ``GroupPath``."""
     for label in ['a', 'a/b', 'a/c/d', 'a/c/e/g']:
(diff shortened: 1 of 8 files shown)


tests/tools/groups/test_paths.py
```diff
@@ -117,16 +117,17 @@ def test_walk(setup_groups):
...
-        orm.Group.objects.get_or_create(label, type_string=orm.GroupTypeString.USER.value)
+        orm.Group.objects.get_or_create(label)
     group_path = GroupPath()
-    assert [c.path for c in sorted(group_path.walk())
-           ] == ['a', 'a/b', 'a/c', 'a/c/d', 'a/c/e', 'a/c/e/g', 'a/f', 'bad', 'bad/other']
+    expected = ['a', 'a/b', 'a/c', 'a/c/d', 'a/c/e', 'a/c/e/g', 'a/f', 'bad', 'bad/other']
+    assert [c.path for c in sorted(group_path.walk())] == expected
 
 
 def test_walk_nodes(clear_database):
```
chrisjsewell, line 128:
```suggestion
def test_walk_nodes(clear_database_before_test):
```

Unrelated to this PR, but should the default for `clear_database` be `clear_database_before_test`? I can't really think of a use case where it is ever good to be doing `clear_database_after_test`?

tests/tools/groups/test_paths.py
```diff
@@ -135,17 +136,17 @@ def test_walk_nodes(clear_database):
...
     for label in ['a/c/e', 'a/f']:
-        orm.Group.objects.get_or_create(label, type_string=orm.GroupTypeString.UPFGROUP_TYPE.value)
+        orm.UpfFamily.objects.get_or_create(label)
     group_path = GroupPath()
     assert sorted([c.path for c in group_path.walk()]) == ['a', 'a/b', 'a/c', 'a/c/d', 'a/c/e', 'a/c/e/g']
-    group_path = GroupPath(type_string=orm.GroupTypeString.UPFGROUP_TYPE.value)
+    group_path = GroupPath(cls=orm.UpfFamily)
     assert sorted([c.path for c in group_path.walk()]) == ['a', 'a/c', 'a/c/e', 'a/f']
-    assert GroupPath('a/b/c') != GroupPath('a/b/c', type_string=orm.GroupTypeString.UPFGROUP_TYPE.value)
+    assert GroupPath('a/b/c') != GroupPath('a/b/c', cls=orm.UpfFamily)
```
chrisjsewell, line 151:
```suggestion

def test_cls_label_clashes(clear_database_before_test):
    """Test behaviour when multiple group classes have the same label."""

    grp1, _ = orm.Group.objects.get_or_create("a")
    node1 = orm.Data()
    node1.store()
    grp1.add_nodes(node1)

    grp2, _ = orm.UpfFamily.objects.get_or_create("a")
    node2 = orm.Data()
    node2.store()
    grp2.add_nodes(node2)

    assert GroupPath("a").group_ids == [grp1.id]
    with pytest.raises(MultipleObjectsError):
        GroupPath("a").get_group()
    assert [(r.group_path.path, r.node.id) for r in GroupPath("a").walk_nodes()] == [("a", node1.id)]

    assert GroupPath("a", cls=orm.UpfFamily).group_ids == [grp2.id]
    assert GroupPath("a", cls=orm.UpfFamily).get_group().id == grp2.id
    assert [(r.group_path.path, r.node.id) for r in GroupPath("a", cls=orm.UpfFamily).walk_nodes()] == [("a", node2.id)]
```

This is the main thing I wanted to check. I'm not sure if it is desired behaviour that `GroupPath("a").get_group()` raises an exception. Should it act the same as `group_ids`, such that it does not allow sub-classing. Also should there be a `get_groups` that does allow sub-classing?

PR #3905, opened 2020-04-07
[ON HOLD] Use apt-spy2 for CI

### LINKED ISSUE #3904: sudo apt calls failing regularly in CI
`sudo apt` is failing regularly to connect to `azure.archive.ubuntu.com`. See actions/virtual-environments#675 for more information.

The suggestion there is to implement a work-around using [`apt-spy2`](https://github.com/lagged/apt-spy2).

CasperWA:
Fixes #3904

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,9 @@ jobs:
       # remove occasionally problematic repositories we don't use anyway
       run: |
         sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
+        sudo gem install apt-spy2
+        sudo apt-spy2 check
+        sudo apt-spy2 fix
         sudo apt update
         sudo apt install texlive-base texlive-generic-recommended texlive-fonts-recommended texlive-latex-base texlive-latex-recommended texlive-latex-extra dvipng dvidvi
 
@@ -51,6 +54,9 @@ jobs:
     - name: Install system dependencies
       run: |
         sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
+        sudo gem install apt-spy2
+        sudo apt-spy2 check
+        sudo apt-spy2 fix
         sudo apt update
         sudo apt install libkrb5-dev ruby ruby-dev
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,7 @@ jobs:
       run: |
         sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
         sudo gem install apt-spy2
-        sudo apt-spy2 check
-        sudo apt-spy2 fix
+        sudo apt-spy2 fix --commit
         sudo apt-get update
         sudo apt-get install texlive-base texlive-generic-recommended texlive-fonts-recommended texlive-latex-base texlive-latex-recommended texlive-latex-extra dvipng dvidvi
 
@@ -55,8 +54,7 @@ jobs:
       run: |
         sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
         sudo gem install apt-spy2
-        sudo apt-spy2 check
-        sudo apt-spy2 fix
+        sudo apt-spy2 fix --commit
         sudo apt-get update
         sudo apt-get install libkrb5-dev ruby ruby-dev
 
@@ -147,8 +145,7 @@ jobs:
       run: |
         sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
         sudo gem install apt-spy2
-        sudo apt-spy2 check
-        sudo apt-spy2 fix
+        sudo apt-spy2 fix --commit
         sudo apt-get update
         sudo apt-get install postgresql-10 graphviz
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,8 @@ jobs:
         sudo gem install apt-spy2
         sudo apt-spy2 check
         sudo apt-spy2 fix
-        sudo apt update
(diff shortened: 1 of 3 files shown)


sphuber:
Can you please provide some explanation and analysis why this should fix the issue as mentioned?

CasperWA:
> Can you please provide some explanation and analysis why this should fix the issue as mentioned?

[`apt-spy2`](https://github.com/lagged/apt-spy2) is a package that adds mirror sites that can be used by `apt`, `apt-get`, etc. if the list of providers should not be online.
By running `apt-spy2 check` and `apt-spy2 fix` it first checks the current list of `apt` providers and then adds the mirrors (as far as I understand). `apt` does _not_ support mirrors by default.

To learn more about `apt-spy2` I suggest going to [their repository page](https://github.com/lagged/apt-spy2).

CasperWA:
It should be noted that it adds some minutes to each workflow job that uses it, since it takes some time to go through the list and check whether they are UP or not.

CasperWA:
The latest commit uses `apt-get` instead of `apt` in the CI workflows, since `apt-get` is more stable (according to the `apt` package itself).

CasperWA:
It seems to work 😮

sphuber:
> It seems to work open_mouth

That it is kind of a hard statement to make given that it is solving a transient problem, isn't it? :wink: 

Since this issue is quite recent and there might be other less heavy solutions, is it okay to hold off a bit with merging this? Is it always the azure mirror that is timing out? If so we could maybe just remove that mirror from the apt sources list

sphuber:
It seems like they [put in a fix](https://github.com/actions/virtual-environments/issues/675#issuecomment-610718578) and we cannot at least hold on a bit with putting this in

MCOfficer:
> > Can you please provide some explanation and analysis why this should fix the issue as mentioned?
> 
> [`apt-spy2`](https://github.com/lagged/apt-spy2) is a package that adds mirror sites that can be used by `apt`, `apt-get`, etc. if the list of providers should not be online.
> By running `apt-spy2 check` and `apt-spy2 fix` it first checks the current list of `apt` providers and then adds the mirrors (as far as I understand). `apt` does _not_ support mirrors by default.
> 
> To learn more about `apt-spy2` I suggest going to [their repository page](https://github.com/lagged/apt-spy2).

for the record, it doesn't add mirrors, it just replaces the current one. It does check for availability before doing so, so (unlike the preconfigured one) the new mirror is all but guaranteed to work.

CasperWA:
@MCOfficer wrote:
> for the record, it doesn't add mirrors, it just replaces the current one. It does check for availability before doing so, so (unlike the preconfigured one) the new mirror is all but guaranteed to work.

Thanks for the clarification :)
Could you clarify maybe also which mirror hosts you have chosen and why, or direct to a page where it's written and explained?

@sphuber wrote:
> It seems like they [put in a fix](https://github.com/actions/virtual-environments/issues/675#issuecomment-610718578) and we cannot at least hold on a bit with putting this in

"_Cannot_", I think, is correct. I will rebase.

sphuber:
> "_Cannot_", I think, is correct. I will rebase.

obvious typo :) but I haven't seen the problem for a while now

CasperWA:
> > "_Cannot_", I think, is correct. I will rebase.
> 
> obvious typo :) but I haven't seen the problem for a while now

I choose to flat out ignore your typo, hehe. No, but _if_ they succeeded in fixing the issue, then great! If not - I think this is a viable temporary solution, and therefore think it should be left open a little while longer.

MCOfficer:
> @MCOfficer wrote:
> 
> > for the record, it doesn't add mirrors, it just replaces the current one. It does check for availability before doing so, so (unlike the preconfigured one) the new mirror is all but guaranteed to work.
> 
> Thanks for the clarification :)
> Could you clarify maybe also which mirror hosts you have chosen and why, or direct to a page where it's written and explained?

I don't think it's fully documented, but judging by the source code, it will get the list of mirrors from http://mirrors.ubuntu.com/mirrors.txt. That list changes based on the client's geolocation, so the runners (iirc they are part of azure's us-east cluster) will likely get http://mirrors.ubuntu.com/US.txt.

These mirrors are then checked for availability, and presumably sorted by ping. The best mirror is chosen, if `--fix` was provided it is written to `/etc/apt/source.list`.

.github/workflows/ci.yml
```diff
@@ -21,8 +21,11 @@ jobs:
       # remove occasionally problematic repositories we don't use anyway
       run: |
         sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
-        sudo apt update
-        sudo apt install texlive-base texlive-generic-recommended texlive-fonts-recommended texlive-latex-base texlive-latex-recommended texlive-latex-extra dvipng dvidvi
+        sudo gem install apt-spy2
+        sudo apt-spy2 check
+        sudo apt-spy2 fix
```
MCOfficer, line 26:
```suggestion
        sudo apt-spy2 fix --commit
```

these lines are redundant - `check` apparently only exists for convenience, to output the best mirror. `fix` does the same, but can also apply that change to `/etc/apt/sources.list`.

To actually apply the change, you need to provide the `--commit` argument:
```bash
florian@florian-vm-lubuntu1910:~$ sudo apt-spy2 fix
[sudo] password for florian: 
The closest mirror is: http://ftp.rrzn.uni-hannover.de/pub/mirror/linux/ubuntu/
Run with --commit to adjust /etc/apt/sources.list
```

The same applies to all other usages in this changeset.

CasperWA: So for the previous runs, `apt-spy2` didn't actually apply a change, but we were indeed still running on the original addresses? Hmm.

MCOfficer: Yes, you can see apt fetching from `azure.archive.ubuntu.com` just below the log line i linked.

CasperWA: Sure. My point was just that the temporary fix was not needed from the get go then.
I have applied your suggestion, thanks! :) But think I will advise to let this PR lie around for some time, to see if the issue persists and only merge it if it does.

sphuber:
Closing this for now, since we haven't seen the problem in a week

PR #3910, opened 2020-04-08
Add the `-v/--version` option to `verdi export migrate`

### LINKED ISSUE #3909: Add `--version` option to `verdi export migrate`
This use case might be rare for normal users, although it is not unthinkable that someone may want to migrate to a particular version instead of to the latest. However, it does come very handy for developers when having to create archives for testing. I just had to do this and recreate archives of all version starting from `v0.1`. Without this option this is very tedious.

sphuber:
Fixes #3909 

The default behavior remains the same and if not specified the export
archive will be migrated to the latest version. However, with the flag
any other version can be chosen, as long as it constitutes a forward
migration as backward migrations are not supported.

### DIFF
--- a/aiida/tools/importexport/migration/__init__.py
+++ b/aiida/tools/importexport/migration/__init__.py
@@ -8,9 +8,9 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Migration export files from old export versions to the newest, used by `verdi export migrate` command."""
-
-from aiida.cmdline.utils import echo
-from aiida.tools.importexport.common.exceptions import DanglingLinkError
+from aiida.common.lang import type_check
+from aiida.tools.importexport import EXPORT_VERSION
+from aiida.tools.importexport.common.exceptions import DanglingLinkError, ArchiveMigrationError
 
 from .utils import verify_metadata_version
 from .v01_to_v02 import migrate_v1_to_v2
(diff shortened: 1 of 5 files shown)


aiida/tools/importexport/migration/__init__.py
```diff
@@ -34,34 +33,35 @@
...
-    Recursive migration of export files from v0.1 to newest version,
+def migrate_recursively(metadata, data, folder, version=EXPORT_VERSION):
+    """Recursive migration of export files from v0.1 to a newer version.
+
     See specific migration functions for detailed descriptions.
 
     :param metadata: the content of an export archive metadata.json file
     :param data: the content of an export archive data.json file
     :param folder: SandboxFolder in which the archive has been unpacked (workdir)
+    :param version: the version to migrate to, by default is the current export version
```
CasperWA, line 44:
```suggestion
    :param version: the version to migrate to, by default the current export version
```
or `it is`.

CasperWA:
Since you're adding functionality to `migrate_recursively` could you also make a test for the new feature, directly testing the function?

Furthermore, the failing test (that I see you have now fixed) was originally a request by @ltalirz that upon failure the exit code should be non-zero. Can you confirm that this is still the case? (Since we're raising in all cases, I guess this is still the case, but just to be sure.) There should probably already be another test for this?

sphuber:
> Since you're adding functionality to `migrate_recursively` could you also make a test for the new feature, directly testing the function?

Good point, will add it

> Furthermore, the failing test (that I see you have now fixed) was originally a request by @ltalirz that upon failure the exit code should be non-zero. Can you confirm that this is still the case? (Since we're raising in all cases, I guess this is still the case, but just to be sure.) There should probably already be another test for this?

The behavior is still the same, I simply made the utility function raise instead of echo'ing directly. The CLI now catches this exception and turns it into a critical echo. We should not include `echo` functionality in non-CLI code.

CasperWA:
> > Furthermore, the failing test (that I see you have now fixed) was originally a request by @ltalirz that upon failure the exit code should be non-zero. Can you confirm that this is still the case? (Since we're raising in all cases, I guess this is still the case, but just to be sure.) There should probably already be another test for this?
> 
> The behavior is still the same, I simply made the utility function raise instead of echo'ing directly. The CLI now catches this exception and turns it into a critical echo. We should not include `echo` functionality in non-CLI code.

I completely agree, as per my original review comment.
I just wanted to mention it (and tag @ltalirz) and make doubly sure it still satisfies his specific use case :)

tests/tools/importexport/migration/test_migration.py
```diff
@@ -158,43 +175,22 @@ def test_wrong_versions(self):
 
         # Make sure migrate_recursively throws a critical message and raises SystemExit
```
CasperWA, line 176:
```suggestion
        # Make sure migrate_recursively raises ArchiveMigrationError
```

tests/tools/importexport/migration/test_migration.py
```diff
@@ -102,6 +100,27 @@ def test_migrate_recursively(self):
...
+            # Backward migrations are not supported
+            with self.assertRaises(ArchiveMigrationError):
+                migrate_recursively(archive.meta_data, archive.data, None, version='0.2')
+
+            # Same version will also raise
+            with self.assertRaises(ArchiveMigrationError):
+                migrate_recursively(archive.meta_data, archive.data, None, version='0.3')
+
+            version = migrate_recursively(archive.meta_data, archive.data, None, version='0.5')
+            self.assertEqual(version, '0.5')
```
CasperWA, line 122:
```suggestion
            migrated_version = '0.5'
            version = migrate_recursively(archive.meta_data, archive.data, None, version=migrated_version)
            self.assertEqual(version, migrated_version)
```

tests/tools/importexport/migration/test_migration.py
CasperWA: The comment here can alternatively be completely removed.

ltalirz:
I'm sure raising an exception/echo_critical will result in a nonzero exit code, so no complaints from my side.

sphuber:
@CasperWA done and done

PR #3912, opened 2020-04-08
Add export archive migration for `Group` type strings 

### LINKED ISSUE #3908: Add migration for export archives to account for change in group type strings
PR #3882 implemented the functionality to subclass `Group` and register it through entry points. The entry point string would be stored in the `type_string` column. This required a database migration and likewise we should migrate existing export archives.

sphuber:
Fixes #3908

The `Group` entity class is now subclassable, for which the type string
of existing entries in the database had to be changed through a
migration. Here we add the corresponding migration for existing export
archives. The only required change is to map the type string of groups.

To test this migration we use the existing export archives in the module
`tests/fixtures/export/migrate`. They did not contain any group entities
so `export_v0.1_simple.aiida` was updated first and had four groups
added, one for each of the migrated type strings. This initial archive
was then migrated to each subsequent version, step by step, using the
command `verdi export migrate --version`.


sphuber:
@CasperWA this PR is blocked because it depends on #3882 and #3903 need to be merged first. But I wanted to already open it to discuss it. We only have to look at the last commit. The migration is relatively simple, just updating the `type_string` of `Group` entities in archives. To test this, I use the existing export archives in the `tests/fixtures/export/migrate`. They did not contain any group entities, so I updated `v0.1` to add 4 groups, one of each currently existing type strings, and then migrated it step by step to create each archive for each version (hence my previous PR for `verdi export migrate --version :wink: ).

Question: is this sufficient, or do we need an update in the archives that are hosted in the external repository as well?

sphuber:
@CasperWA do you have an idea when you could give a look at this? Because this PR is blocking the release

CasperWA:
> @CasperWA do you have an idea when you could give a look at this? Because this PR is blocking the release

I can see what time I can spare over Easter and get it done. Otherwise it will not be until after, I'm afraid.

aiida/tools/importexport/migration/v08_to_v09.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def migration_dbgroup_type_string(data):
+    """Apply migration 0044 - REV. 1.0.44
+
+    Rename the `type_string` columns of all `Group` instances.
+    """
+    mapping = {
+        'user': 'core',
+        'data.upf.family': 'core.upf',
```
CasperWA, line 38: In the database migrations for both Django and SQLA this is simply called `data.upf`.

aiida/tools/importexport/migration/v08_to_v09.py
```diff
@@ -0,0 +1,58 @@
...
+        'auto.import': 'core.import',
+        'auto.run': 'core.auto',
+    }
+
+    for _, attributes in data.get('export_data', {}).get('Group', {}).items():
```
CasperWA, line 43:
```suggestion
    for attributes in data.get('export_data', {}).get('Group', {}).values():
```

tests/tools/importexport/migration/test_v08_to_v09.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def test_migrate_complete(self):
+        """Test specific data migrations"""
+        metadata, data = get_json_files('export_v0.8_manual.aiida', **self.external_archive)
+        verify_metadata_version(metadata, version='0.8')
+
+        migrate_v8_to_v9(metadata, data)
+        verify_metadata_version(metadata, version='0.9')
+
+        for _, attributes in data.get('export_data', {}).get('Group', {}).items():
```
CasperWA, line 77:
```suggestion
        for attributes in data.get('export_data', {}).get('Group', {}).values():
```

tests/tools/importexport/migration/test_v08_to_v09.py
```diff
@@ -0,0 +1,79 @@
...
+###########################################################################
+"""Test export file migration from export version 0.8 to 0.9"""
+from aiida.backends.testbase import AiidaTestCase
+from aiida.tools.importexport.migration.utils import verify_metadata_version
+from aiida.tools.importexport.migration.v08_to_v09 import migrate_v8_to_v9
+
+from tests.utils.archives import get_json_files
+
+
+class TestMigrate(AiidaTestCase):
```
CasperWA, line 18:
```suggestion
class TestMigrateV08toV09(AiidaTestCase):
```
Not crucial - simply to keep in step with the other Class names.

tests/tools/importexport/migration/test_v08_to_v09.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @classmethod
+    def setUpClass(cls, *args, **kwargs):
+        super().setUpClass(*args, **kwargs)
+
+        # Utility helpers
+        cls.external_archive = {'filepath': 'archives', 'external_module': 'aiida-export-migration-tests'}
+        cls.core_archive = {'filepath': 'export/migrate'}
+
+    def test_migrate(self):
```
CasperWA, line 29:
```suggestion
    def test_migrate_v8_to_v9(self):
```
Same as the comment above.

tests/tools/importexport/migration/test_v08_to_v09.py
sphuber: I agree the inconsistency with the existing test files is not nice, but I decided to do this on purpose to simplify creating these test files. Each file contains a lot of repetition of the two versions concerned making for a lot of work to make sure all occurrences are properly replaced. The two changes you have highlighted are not the only occurrences. Do you prefer I revert those changes or I simplify all other test files as well to remove references to version numbers. As I see it, since the module name itself already includes the versions, that is more than enough

aiida/tools/importexport/migration/v08_to_v09.py
sphuber: well-spotted! :+1:

sphuber:
Thanks a lot for the review @CasperWA , especially during this easter weekend.

> An initial question:
> 
>     * What changed in the `export_v0.X_simple.aiida` archives? They shouldn't change, I believe, right?

As explained in the commit and OP:
>  To test this migration we use the existing export archives in the module
`tests/fixtures/export/migrate`. They did not contain any group entities
so `export_v0.1_simple.aiida` was updated first and had four groups
added, one for each of the migrated type strings. This initial archive
was then migrated to each subsequent version, step by step, using the
command `verdi export migrate --version`.


> 
> And then, let's go through the [dedicated wiki-page](https://github.com/aiidateam/aiida-core/wiki/Checklist-when-updating-the-export-version).

Ah yes, apologies, forgot that that existed.

>     3. Tests
>        a. Export/Import tests for changes introduced to export/import
>        You may or may not choose to add a small test here (at `tests/tools/importexport/orm/test_groups.py`) making sure the `type_string` is as expected. Perhaps a quick look through the tests may also find that some may need updating?
>        b. Archive migration tests
>        No PR has been made for the [`aiida-export-migration-tests` repository](https://github.com/aiidateam/aiida-export-migration-tests) in order to update it to v0.9, with all that it may entail.
>        Also, you should add a `test_v0.8_to_newest` in `tests.tools.importexport.migration.text_migration:TestExportFileMigration`
>        And finally, you're missing a test specifically for the `migration_dbgroup_type_string` function.

This last part 3b was exactly what I was wondering. Since the changes are really minor (just a changing of the `type_string`) and those changes are already tested through the simple archives that are included in `aiida-core` itself, is it really necessary to create an additional archive on `aiida-export-migration-tests`?

CasperWA:
> > To test this migration we use the existing export archives in the module
> > `tests/fixtures/export/migrate`. They did not contain any group entities
> > so `export_v0.1_simple.aiida` was updated first and had four groups
> > added, one for each of the migrated type strings. This initial archive
> > was then migrated to each subsequent version, step by step, using the
> > command `verdi export migrate --version`.

Cool, thanks! 👍 

> > ```
> > 3. Tests
> >    a. Export/Import tests for changes introduced to export/import
> >    You may or may not choose to add a small test here (at `tests/tools/importexport/orm/test_groups.py`) making sure the `type_string` is as expected. Perhaps a quick look through the tests may also find that some may need updating?
> >    b. Archive migration tests
> >    No PR has been made for the [`aiida-export-migration-tests` repository](https://github.com/aiidateam/aiida-export-migration-tests) in order to update it to v0.9, with all that it may entail.
> >    Also, you should add a `test_v0.8_to_newest` in `tests.tools.importexport.migration.text_migration:TestExportFileMigration`
> >    And finally, you're missing a test specifically for the `migration_dbgroup_type_string` function.
> > ```
> 
> This last part 3b was exactly what I was wondering. Since the changes are really minor (just a changing of the `type_string`) and those changes are already tested through the simple archives that are included in `aiida-core` itself, is it really necessary to create an additional archive on `aiida-export-migration-tests`?

You can definitely argue that it's overkill.
However, I suspect that skipping a version (no matter the minority of it) will end badly for future migration testing.
I just saw that the `aiida-fleur` plug-in also relies on the `aiida-export-migration-tests` package - for testing, I assume - so skipping a version may present unknown problems (in testing), both for `aiida-core` and others.

In any case, I have created a PR in the external repo that you can go through. Concerning the updated archive, I have only changed the `type_string` of the only group that was in there and updated the `metadata.json` file to reflect the new version.

tests/tools/importexport/migration/test_v08_to_v09.py
CasperWA: I vote for simplification in the name of hindsight 🗳️

CasperWA:
You should update to `aiida-export-migration-tests==0.9.0` in this PR (just released).

sphuber:
@CasperWA this is ready for review

tests/tools/importexport/migration/test_v08_to_v09.py
```diff
@@ -0,0 +1,25 @@
...
+"""Test export file migration from export version 0.8 to 0.9"""
+from aiida.tools.importexport.migration.v08_to_v09 import migrate_v8_to_v9
+
+from . import ArchiveMigrationTest
+
+
+class TestMigrate(ArchiveMigrationTest):
+    """Tests specific for this archive migration."""
+
+    def test_migrate_external(self):
```
CasperWA, line 19: It may be useful to add a specific test for the `migration_dbgroup_type_string` function, to assert that you get the expected 1:1 conversion here. However, it may also be overkill, I'm not entirely sure :)

sphuber:
> Edit: Also, you may want to make a test under `tests/tools/importexport/orm/test_groups.py` to make sure the `type_strings` are correct in the created `data.json` file.
> Again, this is in line with my comment here, and may be overkill and a bit too specific a test?

I don't think really understand why you think there should be a test of the migration function in the `orm` subsection. The migration functionality should be tested in the migration tests. I do agree with your other suggestion though. Even though I made sure that the test archives now contain one group of each type, that may be changed (by mistake) in the future and then it won't be tested anymore. I will add an explicit test for the function

tests/tools/importexport/migration/test_v08_to_v09.py
sphuber: Done

CasperWA:
> > Edit: Also, you may want to make a test under `tests/tools/importexport/orm/test_groups.py` to make sure the `type_strings` are correct in the created `data.json` file.
> > Again, this is in line with my comment here, and may be overkill and a bit too specific a test?
> 
> I don't think really understand why you think there should be a test of the migration function in the `orm` subsection. The migration functionality should be tested in the migration tests. (...)

It is in the importexport `orm` subsection. And it's not to test the migration, but to test that the export function is writing the correct data for `Group`s.

sphuber:
> It is in the importexport `orm` subsection. And it's not to test the migration, but to test that the export function is writing the correct data for `Group`s.

I see, except this code has not been touched nor changed, so don't think we should add a test

sphuber:
@CasperWA this is ready for review

CasperWA:
> > It is in the importexport `orm` subsection. And it's not to test the migration, but to test that the export function is writing the correct data for `Group`s.
> 
> I see, except this code has not been touched nor changed, so don't think we should add a test

No, it is very implicit indeed. I've accepted the PR on the basis of the current changes, but left this open for you to decide upon yourself. I consider it good enough.

sphuber:
Thanks for the review @CasperWA . I hold off a bit with merging still because technically the tests failed, but that is due to a non-related issue. The `pgsu` package still is not available on conda. If this won't be resolved today, I will temporarily disable those tests until it is fixed.

> In any case, I'll approve this PR in its current state. The extra and extremely specific test under `importexport/orm/test_groups.py` that we have discussed is still open for you to do, but I do not consider it essential, as we should already have simple tests that makes sure we can export and import `Group`s, which is enough for me, and can be argued is enough for making sure the `type_string`s are valid, right?

The `type_string` is literally stored as a column on the `DbGroup` table. There is no logic between getting that value to storing it in the export file. Like all other properties of entities, this is a one-to-one dump, so it would really be overkill to test this. That would be like testing that a pk or UUID is exported correctly

CasperWA:
> > In any case, I'll approve this PR in its current state. The extra and extremely specific test under `importexport/orm/test_groups.py` that we have discussed is still open for you to do, but I do not consider it essential, as we should already have simple tests that makes sure we can export and import `Group`s, which is enough for me, and can be argued is enough for making sure the `type_string`s are valid, right?
> 
> The `type_string` is literally stored as a column on the `DbGroup` table. There is no logic between getting that value to storing it in the export file. Like all other properties of entities, this is a one-to-one dump, so it would really be overkill to test this. That would be like testing that a pk or UUID is exported correctly

This was my understanding and thought as well, concerning the "overkill". Great to have it confirmed.

PR #3918, opened 2020-04-09
Trigger test-install workflow for pull requests.

csadorf:
Replace `push` with `pull_request` trigger.

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -1,7 +1,7 @@
 name: test-install
 
 on:
-  push:
+  pull_request:
     paths:
       - 'setup.*'
       - 'environment.yml'


sphuber:
Thanks @csadorf . Just to make sure: the fact that the `test-install / install-with-conda` fails is what we expect and exactly why it is there, correct? So we could merge this and then once the package really is available on conda-forge, we can make the job required for PRs.

csadorf:
> Thanks @csadorf . Just to make sure: the fact that the `test-install / install-with-conda` fails is what we expect and exactly why it is there, correct? So we could merge this and then once the package really is available on conda-forge, we can make the job required for PRs.

Yes, correct. The `test-install` job is expected to fail for this PR, however we should not mark it as _required_, because it will only run when the dependencies are actually changed or when the workflow is changed (like here). See the `paths` list in the trigger definition for all the files that trigger the run.

PR #3894, opened 2020-04-06
additional validation for cli arguments

ltalirz:
fixes #3301 
fixes #3596 
fixes #3897 
fixes #3770 

* Start validating
    * ~~labels (#3596)~~
    * hostnames (#3301)
    * emails (still allow `aiida@localhost`)
    * entry point names
    * profile names
    * First name, last name and institution need to be non-empty (#3301)
* Don't skip validation if user enters `!`: The `InteractiveOption` of the CLI always allowed users to pass `!` in order to avoid setting a value, bypassing any further validation that may be provided by the `type`. Change this behavior such that entering `!` results in an empty string being passed further to the `type`. Only if conversion by the type passes as well should a `None` value be returned.
 * add test for #3770
 * reduce pytest output on travis to make it more readable (coverage + test timings)
 * update `verdi setup` docs for authentication via Unix sockets
 * increase code reuse in click options (specifically: verdi setup/quicksetup)

Note: Since existing databases may contain data that does not validate under the new rules, we have not introduced validations for CLI options that would make it impossible to work with existing data (e.g. for the `label` of nodes, etc.).
Those should be made in a separate PR.

Todo:
 - [x] Open issue https://github.com/aiidateam/aiida-core/issues/3927 for introduction of validation of labels

PR #3921, opened 2020-04-11
Unpin the `click` dependency

### LINKED ISSUE #3810: aiida_local_code_factory: search for existing test code is fragile
The `aiidaLlocal_code_factory` fixture uses a rather fragile method for identifying whether a test code is already in the DB:
https://github.com/aiidateam/aiida-core/blob/928c2fc969af60f125006b5245684d44156b4cc0/aiida/manage/tests/pytest_fixtures.py#L148-L150

It would be better to:
 * Filter both by label and entry point
 * make sure the computer of the code equals the provided one as well

### LINKED ISSUE #3838: Click version 7.1 breaks test and verdi-autodocs pre-commit hook
The click[ version 7.1](https://pypi.org/project/click/7.1/) (released a few hours ago) breaks test and verdi-autodocs pre-commit hook.

As an immediate measure, I recommend to pin the Click version to 7.0 (#3837 ).

sphuber:
Fixes #3838 
Fixes #3810 

The breaking of the tests was caused by two separate commits:

  * 718485be48263056e7036ea9a60ce11b47e2fc26
  * 37d897069f58f7f2a016c14b0620c6d387430b4b

The first one changes the format of the output of a command if a
required option is missing. The double quotes around the option have
been changed to single quotes.

The second commit is more pernicious: the `Editor.edit_file` method was
updated to escape the editor and filename parameters passed to the sub
process call. Our tests were abusing the absence of escaping to pass
arguments to the editor command, in our case `vim`, as to make the
normally interactive command a non-interactive one for testing purposes.
Now that the editor command is escaped, the arguments are understood by
bash to be part of the command which of course then cannot be found and
the test fails. We "fix" this problem by patching the changed method in
`click` to undo the escaping.

### DIFF
--- a/aiida/manage/tests/pytest_fixtures.py
+++ b/aiida/manage/tests/pytest_fixtures.py
@@ -7,6 +7,7 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
+# pylint: disable=redefined-outer-name,unused-argument
 """
 Collection of pytest fixtures using the TestManager for easy testing of AiiDA plugins.
 
@@ -16,8 +17,8 @@ Collection of pytest fixtures using the TestManager for easy testing of AiiDA pl
  * aiida_local_code_factory
 
 """
-import tempfile
 import shutil
+import tempfile
 import pytest
 
 from aiida.manage.tests import test_manager, get_test_backend_name, get_test_profile_name
@@ -30,14 +31,13 @@ def aiida_profile():
     Note: scope='session' limits this fixture to run once per session. Thanks to ``autouse=True``, you don't actually
      need to depend on it explicitly - it will activate as soon as you import it in your ``conftest.py``.
     """
-    # create new TestManager instance
-    with test_manager(backend=get_test_backend_name(), profile_name=get_test_profile_name()) as test_mgr:
-        yield test_mgr
-    # here, the TestManager instance has already been destroyed
+    with test_manager(backend=get_test_backend_name(), profile_name=get_test_profile_name()) as manager:
+        yield manager
+    # Leaving the context manager will automatically cause the `TestManager` instance to be destroyed
 
 
 @pytest.fixture(scope='function')
-def clear_database(clear_database_after_test):  # pylint: disable=redefined-outer-name,unused-argument
+def clear_database(clear_database_after_test):
     """Alias for 'clear_database_after_test'.
 
     Clears the database after each test. Use of the explicit
@@ -46,18 +46,15 @@ def clear_database(clear_database_after_test):  # pylint: disable=redefined-oute
 
 
 @pytest.fixture(scope='function')
-def clear_database_after_test(aiida_profile):  # pylint: disable=redefined-outer-name
-    """Clear the database after each test.
-    """
+def clear_database_after_test(aiida_profile):
+    """Clear the database after the test."""
     yield
-    # after the test function has completed, reset the database
     aiida_profile.reset_db()
 
 
 @pytest.fixture(scope='function')
-def clear_database_before_test(aiida_profile):  # pylint: disable=redefined-outer-name
-    """Clear the database before each test.
-    """
+def clear_database_before_test(aiida_profile):
+    """Clear the database before the test."""
     aiida_profile.reset_db()
     yield
 
@@ -81,7 +78,7 @@ def temp_dir():
 
 
 @pytest.fixture(scope='function')
-def aiida_localhost(temp_dir):  # pylint: disable=redefined-outer-name
+def aiida_localhost(temp_dir):
     """Get an AiiDA computer for localhost.
 
     Usage::
@@ -118,7 +115,7 @@ def aiida_localhost(temp_dir):  # pylint: disable=redefined-outer-name
 
 
 @pytest.fixture(scope='function')
-def aiida_local_code_factory(aiida_localhost):  # pylint: disable=redefined-outer-name
+def aiida_local_code_factory(aiida_localhost):
     """Get an AiiDA code on localhost.
 
     Searches in the PATH for a given executable and creates an AiiDA code with provided entry point.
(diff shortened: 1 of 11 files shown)


sphuber:
@ltalirz @greschd and @csadorf note that in the first commit I fix the version of `click` to be equal higher `7.1`. This is just temporarily to force the tests to run on that version, since they were broken. After review, when you think this is good to be merged, I can revert the requirement to `click~=7.0` since it should now be compatible with both minor versions.

tests/conftest.py
```diff
@@ -11,3 +11,48 @@
...
+
+
+@pytest.fixture()
+def non_interactive_editor(request):
+    """Fixture to patch click's `Editor.edit_file`.
+
+    In `click==7.1` the `Editor.edit_file` command was changed to escape the `os.environ['EDITOR']` command. This was
+    breaking our tests since were abusing this system to define an automated vim command in order to make an interactive
+    command non-interactive. However, the escaping made the command and arguments understood as a single command and
+    so bash could not find it. Here we patch the method to remove the escaping of the editor command.
```
ltalirz, line 23:
```suggestion
    In `click==7.1` the `Editor.edit_file` command was changed to escape the `os.environ['EDITOR']` command.
    We are currently abusing this variable to define an automated vim command in order to make an interactive
    command non-interactive, and escaping it makes bash interpret the command and its arguments as a single command instead.
    Here we patch the method to remove the escaping of the editor command.
```

ltalirz: Why is it enough to define this as a fixture? Is this relevant just for testing?
Perhaps I haven't fully grasped the problem here...

tests/cmdline/commands/test_code.py
```diff
@@ -276,3 +231,54 @@ def setUp(self):
     def test_code_list_no_codes_error_message(self):
         result = self.cli_runner.invoke(code_list)
         self.assertEqual(1, result.output.count('# No codes found matching the specified criteria.'))
+
+
+@pytest.mark.parametrize('non_interactive_editor', ('sleep 1; vim -cwq',), indirect=True)
```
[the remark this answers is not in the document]
ltalirz: ok, I guess it has to do with this monkey-patching fixture...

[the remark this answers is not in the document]
sphuber: Yes, this is the way to paramterize a pytest fixture. In this case it parameterizes the command of the `non_interactive_editor` fixture

tests/conftest.py
sphuber: I am pretty sure yeah, because I don't think we explicitly pass arguments to a file editor anywhere else in the code. But cannot be 100% sure

tests/cmdline/params/options/test_conditional.py
```diff
@@ -66,7 +66,10 @@ def test_switch_on(self):
         runner = CliRunner()
         result = runner.invoke(cmd, ['--on'])
         self.assertIsNotNone(result.exception)
-        self.assertIn('Error: Missing option "--opt".', result.output)
+        # The quoting changed in `click==7.1` so we have to check for both cases if we want to support `click<7.1`
```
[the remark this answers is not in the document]
sphuber: in any case, the condition would require two clauses, but the alternative would maybe indeed be shorter and more legible. Something like:
```
assert "Error: Missing option" in result.output and "--opt" in result.output
```

tests/cmdline/commands/test_code.py
```diff
@@ -276,3 +231,54 @@ def setUp(self):
...
+    from aiida.orm import Code
+    label = 'mixed_remote'
+    options = ['--description=description', '--on-computer', '--remote-abs-path=/remote/abs/path']
+    user_input = '\n'.join([label, 'arithmetic.add', aiida_localhost.name])
+    result = CliRunner().invoke(setup_code, options, input=user_input)
+    assert result.exception is None
+    assert isinstance(Code.get_from_string('{}@{}'.format(label, aiida_localhost.name)), Code)
+
+
+@pytest.mark.skip('This test is hanging for some reason.')
```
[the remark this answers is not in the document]
sphuber: I tried fixing it but I had even trouble understanding what the problem was. The problem was that the test hangs and you get no output. By changing the `click` source and setting default of `resilient_parsing=False` for the `Context` at least it printed a warning. Apparently the `CodeBuilder` was flagging an issue saying that properties were specified for a "remote" code even though it thinks that it concerns a local one. This is weird, because in the test what we are duplicating is a "remote" code. Anyway, I find the code of the `CodeBuilder` a bit hard to follow and I could not figure out why now all of a sudden it was complaining when I am just switching to a different manner of running the test and the test itself remaining essentially unchanged. Would be great if you may have some pointers, since I think you are the original author of the `CodeBuilder`.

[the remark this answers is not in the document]
ltalirz: > since I think you are the original author of the CodeBuilder

I think this honor belongs to Rico, in 6873af9 .
However perhaps the issue is indeed in the code I added for code duplication?
E.g. the `Code.get_code_spec` function?
https://github.com/aiidateam/aiida-core/blob/0644e7d9ff492fdb2c94d608c575c697daee6d32/aiida/orm/utils/builders/code.py#L88-L115

Anyhow, since Codes are now Nodes, perhaps this function is basically obsolete and one could simply clone the Code?=

sphuber:
@ltalirz I fixed the hanging test and in doing so tackled another issue #3810 . If you are happy with the changes let me know, then I will rebase and revert the requirements to `click~=7.0`. It would maybe be good to also merge your PR #3894 before this, as it touches the test that was hanging and I now fixed.

aiida/manage/tests/pytest_fixtures.py
```diff
@@ -118,55 +115,75 @@ def test_1(aiida_localhost):
...
 
-    def get_code(entry_point, executable, computer=aiida_localhost, prepend_text=None, append_text=None):
+    def get_code(
+        entry_point,
+        executable,
+        computer=aiida_localhost,
+        label=None,
+        prepend_text=None,
+        append_text=None,
+        check_executable=True
```
ltalirz, line 140: My only comment is that we should be a bit careful not to let the parameters of this function explode.
Is it really necessary to use this fixture to set up a code pointing to a non-existing path?

tests/cmdline/commands/test_code.py
```diff
@@ -276,3 +231,52 @@ def setUp(self):
...
+    assert result.exception is None
+    assert isinstance(Code.get_from_string('{}@{}'.format(label, aiida_localhost.name)), Code)
+
+
+@pytest.mark.parametrize('non_interactive_editor', ('sleep 1; vim -cwq',), indirect=True)
+def test_code_duplicate_interactive(clear_database_before_test, aiida_local_code_factory, non_interactive_editor):
+    """Test code duplication interactive."""
+    label = 'code_duplicate_interactive_teor'
+    user_input = label + '\n\n\n\n\n\n'
+    code = aiida_local_code_factory('arithmetic.add', '/remote/abs/path', label='code', check_executable=False)
```
ltalirz, line 275: e.g. here, you could point to an existing executable, say `python` or `verdi` or ...

tests/conftest.py
ltalirz: Ok. Just checking whether you are still planning to accept the changes, which are meant describe the status quo rather than what changed.

tests/cmdline/commands/test_code.py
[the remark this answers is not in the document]
ltalirz: Replacing this with a `clone` is probably still a valid point, perhaps open an issue for this if you agree?

aiida/manage/tests/pytest_fixtures.py
sphuber: I would say yes, in most test case scenarios, the executable is never actually run so who cares if it can actually be found. So to be honest I am not sure why the check is there, but in order to not break backwards compatibility I put in a flag.

sphuber: I have reverted the change for now

tests/cmdline/commands/test_code.py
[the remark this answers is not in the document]
sphuber: I was going to, but it does quite a bit more then just providing the functionality that `clone` does. It also provides all the validation and a way for the contextual defaults to be set in the CLI, so I am not sure that we can actually get rid of it.

PR #3924, opened 2020-04-11
Deduplicate code for tests of archive migration code

### LINKED ISSUE #3923: Remove code duplication in export archive migration tests
This came to light in PR #3912

sphuber:
Fixes #3923 

Each archive migration was being tested on a archives included in this
repo in `tools/fixtures/export/migrate` and the test was always the
same. Simply take an archive, migrate it to the next version with the
appropriate method and check that both metadata and data dictionaries of
the migrated data match that of the reference archive. This same test
was implemented for each migration method, but has now been centralized
in a single test using pytest to parametrize the versions.

The testing on external archives from `aiida-export-migration-testing`
has also been streamlined a bit, by defining a new base class called
`ArchiveMigrationTest` that provides some utility methods.


aiida/tools/importexport/migration/v03_to_v04.py
```diff
@@ -457,7 +457,7 @@ def migrate_v3_to_v4(metadata, data, folder, *args):  # pylint: disable=unused-a
     migration_dbgroup_type_string_change_content(data)
     migration_calc_job_option_attribute_keys(data)
     migration_move_data_within_node_module(data)
-    migration_trajectory_symbols_to_attribute(data, folder)
+    migration_trajectory_symbols_to_attribute(data, args[0])
```
CasperWA, line 460: Wouldn't it be more clear to then define `folder = args[0]` above and pass it here? I know it isn't stringently necessary, it just seems like a small effort with large gain for future code readers?

tests/tools/importexport/migration/test_migrations.py
```diff
@@ -0,0 +1,62 @@
...
+    filepath_archive = get_archive_file('export_v{}_simple.aiida'.format(version_old), filepath='export/migrate')
+
+    with Archive(filepath_archive) as archive:
+        metadata_old = copy.deepcopy(archive.meta_data)
+        data_old = copy.deepcopy(archive.data)
+
+        migration_method(metadata_old, data_old, archive.folder)
+        verify_metadata_version(metadata_old, version=version_new)
+
+    yield (version_old, version_new, metadata_old, metadata_new, data_old, data_new)
```
CasperWA, line 38:
```suggestion
    yield version_old, version_new, metadata_old, metadata_new, data_old, data_new
```

tests/tools/importexport/migration/test_migrations.py
```diff
@@ -0,0 +1,62 @@
...
+    metadata_old.pop('aiida_version')
+    metadata_new.pop('aiida_version')
+
+    # Assert conversion message in `metadata.json` is correct and then remove it for later assertions
+    metadata_new.pop('conversion_info')
+    message = 'Converted from version {} to {} with AiiDA v{}'.format(version_old, version_new, get_version())
+    assert metadata_old.pop('conversion_info')[-1] == message, 'Conversion message after migration is wrong'
+
+    assert metadata_old == metadata_new
+    assert data_old == data_new
```
CasperWA, line 62: The `cls.maxDiff` that is set to `None` during the class set up of `ArchiveMigrationTest`, is meant to also work here, or rather, `maxDiff` should be set to `None` so that any differences between the JSON data can be seen. I don't know if this is the case here or not.

CasperWA: Looking around a bit, it seems `pytest` does not have the same "issue". Rather, it seems to be related to the passed verbosity level, where one can set the `max_lines` of verbosity level 1 somewhere, the default seems to be 10 lines.
Maybe this is not an issue in `pytest`, I simply wanted to highlight that the `maxDiff` is set to `None` for these two specific tests/assertions.
I think the best thing to do, is keep as is, and if we in the future see that this test fails (for whatever reason) and the printed diff is too small, we address it then.

tests/tools/importexport/migration/test_v02_to_v03.py
```diff
@@ -137,7 +79,7 @@ def test_compare_migration_with_aiida_made(self):
         metadata_v3.pop('aiida_version')
         self.assertDictEqual(metadata_v2, metadata_v3)
 
-        self.maxDiff = None
+        self.maxDiff = None  # pylint: disable=invalid-name
```
CasperWA, line 82: Since it's set in the `setUpClass` of the parent class, it shouldn't need to be set again.
```suggestion
```

tests/tools/importexport/migration/test_v02_to_v03.py
```diff
@@ -7,79 +7,21 @@
...
-        )
-
-    def test_migrate_v2_to_v3_complete(self):
-        """Test migration for file containing complete v0.2 era possibilities"""
-
-        # Get metadata.json and data.json as dicts from v0.2 file archive
-        metadata, data = get_json_files('export_v0.2.aiida', **self.external_archive)
-        verify_metadata_version(metadata, version='0.2')
+    def test_migrate_external(self):
+        """Test the migration on the test archive provided by the external test package."""
```
CasperWA, line 22: Either make this method title the same for all subsequent migration version test classes or keep it as it was, like you did with the others.

tests/tools/importexport/migration/test_v03_to_v04.py
```diff
@@ -22,77 +20,11 @@
...
-            metadata_v4,
-            msg='After migration, metadata.json should equal intended metadata.json from archives'
-        )
-        self.assertDictEqual(
-            data_v3, data_v4, msg='After migration, data.json should equal intended data.json from archives'
-        )
+class TestMigrate(ArchiveMigrationTest):
+    """Tests specific for this archive migration."""
 
     def test_migrate_v3_to_v4_complete(self):
```
CasperWA, line 29: Should either be updated according to the updates in `test_migrate_v2_to_v3_complete` or vice versa.

tests/tools/importexport/migration/test_v02_to_v03.py
CasperWA: Hmm, it seems only `v3_to_v4` is the odd-one-out.

tests/tools/importexport/migration/test_migrations.py
sphuber: `maxDiff` is indeed `unittest` specific and no longer applicable. With `pytest` you can adjust this through a command line option

tests/tools/importexport/migration/test_v02_to_v03.py
sphuber: Yes, I have left only that one the way it was, because it needs to do more than all the other checks.

tests/tools/importexport/migration/test_v03_to_v04.py
sphuber: I will also just rename this method to `test_migrate_external` just like all the others.

sphuber:
@CasperWA I addressed your comments

aiida/tools/importexport/migration/v03_to_v04.py
```diff
@@ -432,7 +432,7 @@ def add_extras(data):
     data.update({'node_extras': node_extras, 'node_extras_conversion': node_extras_conversion})
 
 
-def migrate_v3_to_v4(metadata, data, folder, *args):  # pylint: disable=unused-argument
+def migrate_v3_to_v4(metadata, data, *args):  # pylint: disable=unused-argument
```
CasperWA, line 435:
```suggestion
def migrate_v3_to_v4(metadata, data, *args):
```

sphuber:
> Thanks for the changes!

Thanks for the review :+1:

> I found a single pylint disable comment that can be removed,

Done

> and to round it off I think there may be an `a` too much in your first commit message.

I might be blind to double a's as a result of my name, because I could not spot it.
If you could be more specific I can fix it when I merge

CasperWA:
> > and to round it off I think there may be an `a` too much in your first commit message.
> 
> I might be blind to double a's as a result of my name, because I could not spot it.
> If you could be more specific I can fix it when I merge

So very true. And so very tragic... 😆 

Anyway, there is something funky with your first commit in the first sentence, around `(...) tested on a archives included in this repo in (...)` specifically the `a archives`-part.

sphuber:
Thanks @CasperWA I will rebase PR #3912 now which should then also be ready

PR #3935, opened 2020-04-15
Do not ignore `type_string` in constructor of `Group`

sphuber:
Doing so would actually break backwards compatibility. Code that creates
groups with explicit custom type strings, would no longer be able to
query for them as the type string was silently converted to `core`. Even
though accepting the passed `type_string` will cause warnings when
loading them from the database, that is preferable then breaking
existing code.

### DIFF
--- a/aiida/orm/groups.py
+++ b/aiida/orm/groups.py
@@ -142,8 +142,13 @@ class Group(entities.Entity, metaclass=GroupMeta):
             raise ValueError('Group label must be provided')
 
         if type_string is not None:
-            message = '`type_string` is deprecated because it is determined automatically, using default `core`'
+            message = '`type_string` is deprecated because it is determined automatically'
             warnings.warn(message)  # pylint: disable=no-member
+
+        # If `type_string` is explicitly defined, override automatically determined `self._type_string`. This is
+        # necessary for backwards compatibility.
+        if type_string is not None:
+            self._type_string = type_string
 
         type_string = self._type_string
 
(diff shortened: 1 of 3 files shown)


greschd:
There's another (less serious) backwards incompatibility: If someone just _happened_ to use a `type_string` that now matches a valid one, their groups will be automatically "promoted" to be of that class. Not much we can do about that one..

sphuber:
It was a tricky one. The problem with this is that if people now use some of the existing type strings that ship with `aiida-core`, e.g. `data.upf`, the existing ones will be migrated to `core.upf` once they upgrade, but any new ones they create will then still get `data.upf`. They then have multiple groups that should be the same but have different type strings. We decided to not allow this, but I just found out that this breaks the `aiida-vasp` plugin. I made a PR that would fix it and make it compatible with `aiida-core==1.2.0` but then they could not also support `aiida-core<=1.2` at the same time. Even though that is not a problem for them as they do not have a released version yet compatible with `aiida-core>=1.0` and so they can go straight to `aiida-core>=1.2`, the fact that we break backwards compatibility is something we should try to avoid at all cost. Guess this change is the lesser of two evils

greschd:
Ah, I see.. yeah that's tricky indeed. For just the querying problem, wouldn't it be possible to write cross-compatible code by just matching either of the two type strings?

For the "built-in" type strings we _could_ translate them automatically in the `Group` constructor.. but that's also pretty hacky, and then querying is still broken. Probably the change in this PR is actually the least-bad behavior.

sphuber:
@greschd if you could reapprove please. I had to fix a test, but now it is good to go

PR #3940, opened 2020-04-17
Remove the equality operator of `ExitCode`

### LINKED ISSUE #3939: Process serialization will fail if it contains an instance of `ExitCode`
This is because the `ExitCode` was changed from simple tuple to a class in `v1.2.0` which now causes `yaml.dump` to except
```
import yaml
from aiida.engine import ExitCode
exit_code = ExitCode()
yaml.dump(exit_code)
```
will raise
```
<ipython-input-4-33b699827b3a> in <module>
----> 1 yaml.dump(exit_code)

~/.virtualenvs/aiida_dev/lib/python3.7/site-packages/yaml/__init__.py in dump(data, stream, Dumper, **kwds)
    288     If stream is None, return the produced string instead.
    289     """
--> 290     return dump_all([data], stream, Dumper=Dumper, **kwds)
    291 
    292 def safe_dump_all(documents, stream=None, **kwds):

~/.virtualenvs/aiida_dev/lib/python3.7/site-packages/yaml/__init__.py in dump_all(documents, stream, Dumper, default_style, default_flow_style, canonical, indent, width, allow_unicode, line_break, encoding, explicit_start, explicit_end, version, tags, sort_keys)
    276         dumper.open()
    277         for data in documents:
--> 278             dumper.represent(data)
    279         dumper.close()
    280     finally:

~/.virtualenvs/aiida_dev/lib/python3.7/site-packages/yaml/representer.py in represent(self, data)
     25 
     26     def represent(self, data):
---> 27         node = self.represent_data(data)
     28         self.serialize(node)
     29         self.represented_objects = {}

~/.virtualenvs/aiida_dev/lib/python3.7/site-packages/yaml/representer.py in represent_data(self, data)
     32 
     33     def represent_data(self, data):
---> 34         if self.ignore_aliases(data):
     35             self.alias_key = None
     36         else:

~/.virtualenvs/aiida_dev/lib/python3.7/site-packages/yaml/representer.py in ignore_aliases(self, data)
    137         if data is None:
    138             return True
--> 139         if isinstance(data, tuple) and data == ():
    140             return True
    141         if isinstance(data, (str, bytes, bool, int, float)):

~/code/aiida/env/dev/aiida-core/aiida/engine/processes/exit_code.py in __eq__(self, other)
     50 
     51     def __eq__(self, other):
---> 52         return all(getattr(self, attr) == getattr(other, attr) for attr in ['status', 'message', 'invalidates_cache'])
     53 
     54 

~/code/aiida/env/dev/aiida-core/aiida/engine/processes/exit_code.py in <genexpr>(.0)
     50 
     51     def __eq__(self, other):
---> 52         return all(getattr(self, attr) == getattr(other, attr) for attr in ['status', 'message', 'invalidates_cache'])
     53 
     54 

AttributeError: 'tuple' object has no attribute 'status'

```

sphuber:
Fixes #3939 

The operator implementation assumed that `other` would always be an
instance of `ExitCode` as well and so therefore was guaranteed to
contain the same attributes. However, comparing it to a bare tuple would
also invoke the `ExitCode.__eq__` method and would raise and
`AttributeError` as a result. The desired behavior for the equality
operator is that it returns True for any other tuple of the same length
and content. Since this is exactly what is implemented by the tuple
base type, we do not have to override it at all.

### DIFF
--- a/tests/engine/processes/text_exit_code.py
+++ b/tests/engine/processes/text_exit_code.py
@@ -33,6 +33,18 @@ def test_exit_code_construct():
     assert exit_code.invalidates_cache == invalidates_cache
 
 
+def test_exit_code_serializability():
+    """Test that an `ExitCode` instance can be serialized and deserialized with `yaml`."""
+    import yaml
+
+    exit_code = ExitCode()
+    serialized = yaml.dump(exit_code)
+    deserialized = yaml.full_load(serialized)
+
+    assert deserialized == exit_code
+    assert isinstance(deserialized, ExitCode)
+
+
 def test_exit_code_equality():
     """Test that the equality operator works properly."""
     exit_code_origin = ExitCode(1, 'message', True)
@@ -41,6 +53,24 @@ def test_exit_code_equality():
 
     assert exit_code_origin == exit_code_clone
     assert exit_code_clone != exit_code_different
+
+    # Check default `ExitCode` also matches normal tuples
+    exit_code = ExitCode()
+    assert exit_code == (0, None, False)
+
+    assert exit_code != {}
+    assert exit_code != []
+    assert exit_code != ()
+    assert exit_code != (0)
+    assert exit_code != (None,)
+    assert exit_code != (0, None)
+    assert exit_code != [0, None, False]
+    assert exit_code != [0, None, False, 'test']
+
+    # `ExitCode` instances should match bare tuples as long as the content is the same
+    exit_code = ExitCode(1, 'message', True)
+    assert exit_code == (1, 'message', True)
+    assert exit_code != ()
 
 
 def test_exit_code_template_message():
--- a/aiida/engine/processes/exit_code.py
+++ b/aiida/engine/processes/exit_code.py
@@ -48,9 +48,6 @@ class ExitCode(namedtuple('ExitCode', ['status', 'message', 'invalidates_cache']
 
         return ExitCode(self.status, message, self.invalidates_cache)
 
-    def __eq__(self, other):
-        return all(getattr(self, attr) == getattr(other, attr) for attr in ['status', 'message', 'invalidates_cache'])
-
 
 # Set the defaults for the `ExitCode` attributes
 ExitCode.__new__.__defaults__ = (0, None, False)


greschd:
Why do we need to implement our own `__eq__` in the first place? By inheriting from the `namedtuple`, it already has an `__eq__`, no?

The difference I can see to the current implementation is that tuples also compare equal to lists (and maybe other containers) with the same content.

sphuber:
> Why do we need to implement our own `__eq__` in the first place? By inheriting from the `namedtuple`, `ExitCode` already has an `__eq__`, no?
> 
> The difference I can see to the current implementation is that tuples also compare equal to lists (and maybe other containers) with the same content.

Uhm, good question :sweat_smile:  I just tried and removed the implementation and the tests I added still pass. That means it does distinguish between type, because the list with the same content does not match (which is what we want I think). So I think you are right and the fix is simply to remove it? Huh

greschd:
> 
> That means it does distinguish between type, because the list with the same content does not match (which is what we want I think). So I think you are right and the fix is simply to remove it? Huh

Right, not sure what made me believe it didn't distinguish list / tuple :dizzy_face:

Indeed, simply removing it seems like the easiest fix.

tests/engine/processes/text_exit_code.py
```diff
@@ -33,6 +33,15 @@ def test_exit_code_construct():
...
 
 
+def test_exit_code_serializability():
+    """Test that an `ExitCode` instance can be serialized and deserialized with `yaml`."""
+    import yaml
+    exit_code = ExitCode()
+    serialized = yaml.dump(exit_code)
+    deserialized = yaml.full_load(serialized)
+    assert deserialized == exit_code
+
```
greschd, line 43: Maybe we should check that the result is still an `ExitCode` instance here?

sphuber:
> Nice! Since this is a pretty significant bug, are you planning a hotfix release for this?

We should release it soon I agree, but I think I will hold off a bit longer to see if anything else props up. If nothing is reported over the weekend I will release `1.2.1` on Monday I think. Sound reasonable?

PR #3945, opened 2020-04-19
Add the `flat` argument to `QueryBuilder.all()` method

sphuber:
This allows to simplify the commonly found form:

    results = [entry[0] for entry in builder.all()]

to the more compact and readable form:

    results = builder.all(flat=True)

The default is set to `flat=False` such that this change is fully
backwards compatible. Occurrences in the code base itself have been
updated as much as possible.

### DIFF
--- a/aiida/orm/groups.py
+++ b/aiida/orm/groups.py
@@ -340,21 +340,12 @@ class Group(entities.Entity, metaclass=GroupMeta):
         """
         from aiida.orm import QueryBuilder
 
-        filters = {}
         if 'type_string' in kwargs:
+            message = '`type_string` is deprecated because it is determined automatically'
+            warnings.warn(message)  # pylint: disable=no-member
             type_check(kwargs['type_string'], str)
 
-        query = QueryBuilder()
-        for key, val in kwargs.items():
-            filters[key] = val
-
-        query.append(cls, filters=filters)
-        results = query.all()
-        if len(results) > 1:
-            raise exceptions.MultipleObjectsError("Found {} groups matching criteria '{}'".format(len(results), kwargs))
-        if not results:
-            raise exceptions.NotExistent("No group found matching criteria '{}'".format(kwargs))
-        return results[0][0]
+        return QueryBuilder().append(cls, filters=kwargs).one()[0]
 
     def is_user_defined(self):
         """
--- a/aiida/orm/querybuilder.py
+++ b/aiida/orm/querybuilder.py
@@ -2189,21 +2189,24 @@ class QueryBuilder:
 
             yield item
 
-    def all(self, batch_size=None):
-        """
-        Executes the full query with the order of the rows as returned by the backend.
-        the order inside each row is given by the order of the vertices in the path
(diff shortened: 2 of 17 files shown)


aiida/orm/groups.py
```diff
@@ -340,21 +340,12 @@ def get(cls, **kwargs):
...
-        for key, val in kwargs.items():
-            filters[key] = val
-
-        query.append(cls, filters=filters)
-        results = query.all()
-        if len(results) > 1:
-            raise exceptions.MultipleObjectsError("Found {} groups matching criteria '{}'".format(len(results), kwargs))
-        if not results:
-            raise exceptions.NotExistent("No group found matching criteria '{}'".format(kwargs))
-        return results[0][0]
```
ramirezfranciscof, line 357: Are these modifications related to the topic? Is it even doing the same thing? At least it would seem that some exceptions would be missed, no?

tests/orm/test_querybuilder.py
```diff
@@ -629,6 +629,32 @@ def test_direction_keyword(self):
         res2 = {item[1] for item in qb.all()}
         self.assertEqual(res2, {d2.id, d4.id})
 
+    @staticmethod
+    def test_flat():
+        """Test the `flat` keyword for the `.iterall()`, `.all()` and `.one()` methods."""
```
ramirezfranciscof, line 634: Mmm how is this testing `.iterall()` and `.one()`? Is this some copy/paste leftover?

aiida/orm/groups.py
sphuber: It would be, except that if you look at the behavior of `QueryBuilder.one()` already has exactly the same behavior, so this code that I removed was just redundant. The behavior should remain exactly the same. If you think it doesn't, let me know, because then I messed up ;)

tests/orm/test_querybuilder.py
sphuber: Yeah, this is just a mistake that I left in. I originally thought I was going to implement it for all, but for `one` it would break backwards-compatibility and for `iterall` it does not make sense

aiida/orm/groups.py
ramirezfranciscof: Ah, I see, yes, thats true. The only difference is in the second check `if not results` vs `elif len(res) == 0`, but I can't think of a case where this would cause a divergent behaviour, let alone one which would actually be problematic. I mean, if res/result was a float let's say, it would have problems when evaluating `elif len(res) == 0` but would pass `if not results`, although in that case it would already had problems with the first `if len(results) > 1`. So no, I guess no complaints here.

aiida/orm/querybuilder.py
```diff
@@ -2203,7 +2203,12 @@ def all(self, batch_size=None):
```
chrisjsewell, line 2203:
```suggestion
        :param bool flat: 
            Return the result as a flat list of projected entities,
            with no sub-lists.

```

chrisjsewell: Apart from that, it looks good to me 👍

PR #3948, opened 2020-04-20
Make the job submission script configurable through `CalcJob` options

### LINKED ISSUE #3947: Idempotency shortcut in `task_upload_job` returns incorrect value
The `task_upload_job` checks the state of the `CalcJob` to make sure it has not already completed that step before. If that is the case, it returns straight away. Unfortunately, the return type is incorrect. https://github.com/aiidateam/aiida-core/blob/e33504cdd5a490ddf82076fc79986de862a15479/aiida/engine/processes/calcjobs/tasks.py#L63
It returns `True` while it should return the tuple of `calc_info, script_filename`.

sphuber:
Fixes #3947 

The name of the job submission script was hard coded to `_aiidasubmit.sh`
which not only makes it less flexible but it also required the value to
be passed in a long series of calls from `CalcJob.presubmit` all the way
through the transport tasks to the `submit_calculation` method in the
`aiida.engine.daemon.execmanager` module.

This value is now configurable through `metadata.options.script_filename`
with the default being the same. This allows to remove the value from
the return value of `CalcJob.presubmit` and the call-chain through all
the transport tasks, because `submit_calculation` can now simply
retrieve the correct value directly from the `CalcJobNode`.

This change also fixes a bug in the `task_upload_job` that returned the
wrong value if the idempotency check was hit to see if the method had
already been executed. This can happen if the daemon is stopped before
the next transport task is reached but after the upload has already been
executed. This check prevents it from being executed again by simply
returning. However, it was supposed to return the `calc_info` and
`script_filename` normally returned by the `presubmit` call, but it
would no longer have these variables and so returned `True`. The caller
`Waiting.execute` would expect a tuple and except.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -269,13 +269,11 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
         remotedata.store()
 
 
-def submit_calculation(calculation, transport, calc_info, script_filename):
+def submit_calculation(calculation, transport):
     """Submit a previously uploaded `CalcJob` to the scheduler.
 
     :param calculation: the instance of CalcJobNode to submit.
     :param transport: an already opened transport to use to submit the calculation.
-    :param calc_info: the calculation info datastructure returned by `CalcJobNode._presubmit`
-    :param script_filename: the job launch script returned by `CalcJobNode._presubmit`
     :return: the job id as returned by the scheduler `submit_from_script` call
     """
     job_id = calculation.get_job_id()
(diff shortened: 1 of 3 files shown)


sphuber:
@ltalirz if we can get this merged, then I can release `v1.2.1`. The other open PRs #3949 and #3945 are features and are not crucial

sphuber:
Thanks @ltalirz . I actually agree with the more specific name `submit_script_filename` so I have applied that change (even though internally the name was actually `script_filename` but that should not affect the user). Not that the name of the variable returned by a dry-run I have kept as `script_filename` as that would be a breaking change.

PR #3949, opened 2020-04-20
Add config option `daemon.default_workers` and  `daemon.worker_process_slots` 

### LINKED ISSUE #3677: Make RMQ prefetch count a profile option
At the moment the number of worker slots, as defined by the RMQ prefetch count, is only configurable in the code at `aiida.manage.external.rmq._RMQ_TASK_PREFETCH_COUNT`.  This would be better placed as an advanced configuration option in the profile.

@giovannipizzi , @sphuber , @dev-zero, @greschd  Are we happy with using the name 'slots' such that our two tuning parameters would be num. slots and num. workers?

sphuber:
Fixes #3677 

This option will determine the default number of workers that are
started by the `verdi daemon start` command. Specifying an explicit
value as argument will of course still override this.

This option is used in the construction of the RabbitMQ communicator and
will determine the maximum number of process tasks that any one daemon
runner can operate on concurrently. If this number is set too high, the
runner can get overwhelmed as there is no logic for it to prioritize
finishing active processes over taking on new processes that have
arrived in the process task queue.

### DIFF
--- a/aiida/manage/configuration/options.py
+++ b/aiida/manage/configuration/options.py
@@ -14,6 +14,7 @@ import collections
 __all__ = ('get_option', 'get_option_names', 'parse_option')
 
 NO_DEFAULT = ()
+DEFAULT_DAEMON_WORKERS = 1
 DEFAULT_DAEMON_TIMEOUT = 20  # Default timeout in seconds for circus client calls
 VALID_LOG_LEVELS = ['CRITICAL', 'ERROR', 'WARNING', 'REPORT', 'INFO', 'DEBUG']
 
@@ -28,6 +29,14 @@ CONFIG_OPTIONS = {
         'valid_values': None,
         'default': 1,
         'description': 'The polling interval in seconds to be used by process runners',
+        'global_only': False,
+    },
+    'daemon.default_workers': {
+        'key': 'daemon_default_workers',
+        'valid_type': 'int',
+        'valid_values': None,
+        'default': DEFAULT_DAEMON_WORKERS,
+        'description': 'The default number of workers to be launched by `verdi daemon start`',
         'global_only': False,
     },
     'daemon.timeout': {
--- a/aiida/manage/configuration/options.py
+++ b/aiida/manage/configuration/options.py
@@ -16,6 +16,7 @@ __all__ = ('get_option', 'get_option_names', 'parse_option')
 NO_DEFAULT = ()
 DEFAULT_DAEMON_WORKERS = 1
 DEFAULT_DAEMON_TIMEOUT = 20  # Default timeout in seconds for circus client calls
+DEFAULT_DAEMON_WORKER_PROCESS_SLOTS = 200
 VALID_LOG_LEVELS = ['CRITICAL', 'ERROR', 'WARNING', 'REPORT', 'INFO', 'DEBUG']
 
 Option = collections.namedtuple(
@@ -45,6 +46,14 @@ CONFIG_OPTIONS = {
         'valid_values': None,
         'default': DEFAULT_DAEMON_TIMEOUT,
         'description': 'The timeout in seconds for calls to the circus client',
+        'global_only': False,
+    },
+    'daemon.worker_process_slots': {
+        'key': 'daemon_worker_process_slots',
+        'valid_type': 'int',
+        'valid_values': None,
+        'default': DEFAULT_DAEMON_WORKER_PROCESS_SLOTS,
+        'description': 'The maximum number of concurrent process tasks that each daemon worker can handle',
         'global_only': False,
     },
     'db.batch_size': {
--- a/aiida/manage/manager.py
+++ b/aiida/manage/manager.py
@@ -9,7 +9,6 @@
 ###########################################################################
 # pylint: disable=cyclic-import
 """AiiDA manager for global settings"""
-
 import functools
 
 __all__ = ('get_manager', 'reset_manager')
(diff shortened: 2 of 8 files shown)


aiida/manage/configuration/options.py
```diff
@@ -30,6 +32,14 @@
         'description': 'The polling interval in seconds to be used by process runners',
         'global_only': False,
     },
+    'daemon.default_workers': {
+        'key': 'daemon_default_workers',
+        'valid_type': 'int',
+        'valid_values': None,
+        'default': DEFAULT_DAEMON_WORKERS,
+        'description': 'The default number of workers to be launched by `verdi daemon start`',
```
ltalirz, line 40: Just to understand - e.g. if I use `incr` does this value change?
In which sense is it a default rather than the actual number of daemon workers?

aiida/manage/manager.py
```diff
@@ -33,6 +32,16 @@ class Manager:
       * reset manager cache when loading a new profile
     """
 
+    @staticmethod
+    def get_config():
```
ltalirz, line 36: Since `get_config` is used in many places in the code, I'd just like to understand whether this is now the one we suggest for users.

aiida/manage/manager.py
```diff
@@ -33,6 +32,16 @@ class Manager:
...
+        """Return the current loaded config, if any
+
+        :return: current loaded config instance
+        :rtype: :class:`~aiida.manage.configuration.config.Config` or None
+        """
```
ltalirz, line 41: Looking at the docstring of `get_config`, it looks like this function will raise, if the config does not exist (however, it likely gets called in `get_manager` already, to check).
I.e. perhaps the "if any" is not needed here ?

aiida/manage/manager.py
sphuber: I actually just wrote this as a utility function for the `Manager` class itself and did not really think about it being used as the main way of retrieving the config for users. That is currently `from aiida.manage.configuration import get_config`. You'll notice that we haven't currently exposed it on `aiida.manage` level because I had to rewrite the config/profile code for `v1.0.0` and didn't know how stable/safe it was for random user use.

aiida/manage/manager.py
sphuber: The `get_manager` will just return the manager global singleton, creating it if it doesn't exist. Doing so does not require a config nor a loaded profile. i.e. you can do the following in ipython without problem
```
In [2]: from aiida.manage.manager import get_manager                                                                                                                                                             

In [3]: get_manager()                                                                                                                                                                                            
Out[3]: <aiida.manage.manager.Manager at 0x7fd32b9a1f50>

In [6]: get_manager().get_profile() == None                                                                                                                                                                      
Out[6]: True
```
This is why the `get_profile` docstring has the "if any". It is true however, that the `get_config` will raise if it cannot find or parse it, and otherwise will load the config and return it.

I guess updating the docstring of `get_config` to mention the raise, would be sufficient, correct?

aiida/manage/configuration/options.py
sphuber: `verdi daemon incr` is an independent command. It still takes an argument to tell how many workers should be added, `verdi daemon incr N` will add `N` workers with default being 1. The new option only applies how many workers are started automatically when invoking `verdi daemon start`. Already you can do `verdi daemon start 10` to start 10 workers straight away, but now you can configure it to have it start 10 by default. Saves an entire three key strokes ;)

aiida/manage/manager.py
ltalirz: > I guess updating the docstring of get_config to mention the raise, would be sufficient, correct?

Yep!

aiida/manage/manager.py
ltalirz: Ok.

aiida/manage/manager.py
sphuber: Done

aiida/manage/configuration/options.py
ltalirz: Ok, I just wanted to make sure.
I think some users might assume that if you do `verdi daemon incr 1` and then restart the daemon that you would keep the increased number of workers rather than resetting to this default.

It might make sense to  mention explicitly in the docstring of `verdi daemon incr` that these changes are reset whenever you stop the daemon, and that you should use this new config option to make permanent changes.

sphuber: It is a bit more subtle than that even. Doing `verdi daemon restart` will restart the workers but not the daemon itself, i.e. you will keep how every many workers that were running before the restart command. Doing `verdi daemon restart --reset` literally does `verdi daemon stop && verdi daemon start` so there it will start whatever is the default for `verdi daemon start`, which can now be configured through this new option

sphuber: This was already explained in the docstring of `verdi daemon restart`:
```
    By default will only reset the workers of the running daemon. After the restart the same amount of workers will be
    running. If the `--reset` flag is passed, however, the full circus daemon will be stopped and restarted with just
    a single worker.
```
I just noticed that we just have to update the last bit with respect to the new default behavior.

sphuber: I have updated the docstrings of `verdi daemon start` and `verdi daemon restart` in the first commit to reflect the new behavior and give more detail

sphuber:
Thanks for the reviews @ltalirz !

PR #3951, opened 2020-04-20
reintroduce CORS headers to REST API

### LINKED ISSUE #3950: rest API in aiida 1.2.0 missing CORS headers
the AiiDA REST API in the 1.2.0 release is missing the CORS headers (more specifically, it is sending nonsensical CORS headers).

This is a bug introduced by me in https://github.com/aiidateam/aiida-core/pull/3875 
Will take care of fixing this.

ltalirz:
fixes #3950 

Due to a recently introduced bug, the CORS headers sent by the REST API
did not pertain to the correct prefix but to a python module path.

Also, there was a leftover print statement in the HOSTNAME CLI
parameter, introduced in another PR (shame on me!).

### DIFF
--- a/aiida/restapi/run_api.py
+++ b/aiida/restapi/run_api.py
@@ -65,7 +65,7 @@ def run_api(flask_app=api_classes.App, flask_api=api_classes.AiidaApi, **kwargs)
     app.config.update(**config_module.APP_CONFIG)
 
     # Allow cross-origin resource sharing
-    cors_prefix = r'{}/*'.format(config_module)
+    cors_prefix = r'{}/*'.format(config_module.API_CONFIG['PREFIX'])
     CORS(app, resources={cors_prefix: {'origins': '*'}})
 
     # Configure the serializer
--- a/aiida/restapi/run_api.py
+++ b/aiida/restapi/run_api.py
@@ -21,6 +21,8 @@ from aiida.common.warnings import AiidaDeprecationWarning
 from .common.config import CLI_DEFAULTS, APP_CONFIG, API_CONFIG
 from . import api as api_classes
 
+__all__ = ('run_api', 'configure_api')
+
 
 def run_api(flask_app=api_classes.App, flask_api=api_classes.AiidaApi, **kwargs):
     """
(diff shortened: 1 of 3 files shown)


aiida/restapi/run_api.py
```diff
@@ -37,18 +39,53 @@ def run_api(flask_app=api_classes.App, flask_api=api_classes.AiidaApi, **kwargs)
     :param debug: enable debugging
     :param wsgi_profile: use WSGI profiler middleware for finding bottlenecks in web application
     :param hookup: If true, hook up application to built-in server - else just return it
-    """
-    # pylint: disable=too-many-locals
 
-    # Unpack parameters
+    :returns: tuple (app, api) if hookup==False or runs app if hookup==True
```
sphuber, line 43: Good to also add deprecation message of `hookup` to the docstring

aiida/restapi/run_api.py
```diff
@@ -37,18 +39,53 @@ def run_api(flask_app=api_classes.App, flask_api=api_classes.AiidaApi, **kwargs)
...
+            'Using `run_api` with `hookup==False` is deprecated since `v1.2.1` and will stop working in `v2.0.0`. '
+            'Use `configure_api` instead.', AiidaDeprecationWarning
+        )
+
+        # Return the app & api without specifying port/host to be handled by an external server (e.g. apache).
+        # Some of the user-defined configuration of the app is ineffective (only affects built-in server).
+        return (app, api)
+
+
+def configure_api(flask_app=api_classes.App, flask_api=api_classes.AiidaApi, **kwargs):
```
sphuber, line 68: if this is supposed to be used by users of the package, may want to add it to the `__all__` at the top of the file

ltalirz:
Thanks for the review, done!

As for the reason for CORS: without this, you cannot embed results of the REST API in the provenance browser on Materials Cloud.
E.g. this breaks things we've been doing in many tutorials...

sphuber:
Ok, that is a pretty significant bug then. I already fixed another one that was introduced in `1.2.0` so I will merge this and release `1.2.1` today or tomorrow. There are some other open PRs opened yesterday that we may want to include

sphuber:
[pre-commit tests fail with the usual](https://www.youtube.com/watch?v=h_D3VFfhvs4&t=2m52s)
But anyway you have to push a fix for the docs

ltalirz:
thanks, should be fixed now (let's see).
I've also changed the deprecation - you're right that what we want to do is deprecate the entire argument, not just a value.

sphuber:
pre-commit failed again but with the same false alert, so merging anyway

PR #3952, opened 2020-04-21
Docs: add documentation for `GroupPath`

### LINKED ISSUE #3913: Document New Group Utilities
Once the new group based capabilities are finalized, principally #3613, #3882, #3903,
we should update the documentation, e.g. `docs/source/working_with_aiida/groups.rst`,
with a guide for users.

chrisjsewell:
closes #3913


docs/source/working_with_aiida/groups.rst
```diff
@@ -177,3 +177,108 @@ This is what AiiDA uses to load the correct class when reloading the group from
...
+
+    In [12]: path["base1/sub_group1"].delete_group()
+    In [13]: path["base1/sub_group1"].is_virtual
+    Out[13]: True
+    In [14]: path["base1/sub_group1"].get_or_create_group()
+    Out[14]: (<Group: "base1/sub_group1" [type core], of user user@email.com>, True)
+    In [15]: path["base1/sub_group1"].is_virtual
+    Out[15]: False
+
+To traverse paths, the `children` attribute can be used, or for recursive traversal::
```
ltalirz, line 256:
```suggestion
To traverse paths, use the `children` attribute - for recursive traversal, use `walk`::
```

docs/source/working_with_aiida/groups.rst
```diff
@@ -177,3 +177,108 @@ This is what AiiDA uses to load the correct class when reloading the group from
...
+or the group can be retrieved with the `get_group()` method::
+
+    In [8]: path.is_virtual
+    Out[8]: True
+    In [9]: path.get_group() is None
+    Out[9]: True
+    In [10]: path["base1/sub_group1"].get_group() is None
+    Out[10]: True
+    In [11]: path["base1/sub_group1"].get_group()
+    Out[11]: <Group: "base1/sub_group1" [type core], of user user@email.com>
```
sphuber, line 244: As @mbercx highlighted, these two results seen contradicting no?

chrisjsewell: yep copy-paste error 😬 fixed in 9b2f222

chrisjsewell:
> Please maintain one sentence per line for text.

Do you mean converting this:

```
You can also traverse directly through the nodes of a path,
optionally filtering by node class and any other filters allowed by the `QueryBuilder`::
```

to this:

```
You can also traverse directly through the nodes of a path, optionally filtering by node class and any other filters allowed by the `QueryBuilder`::
```

because you end up with quite long lines, which is why I'd added a few breaks on commas.

chrisjsewell:
> Could you please add a sentence or two on the function of cls in the GroupPath and the current limitations?

Added in 506a6c6

CasperWA:
Should this be added before revamping the docs?

chrisjsewell:
I think so. Its ready for merging IMO

sphuber:
> because you end up with quite long lines, which is why I'd added a few breaks on commas.

Yes, that is what I mean, because that is the [style rule that was decided](https://github.com/aiidateam/aiida-core/wiki/Writing-documentation#style-guide). We should not add arbitrary line breaks but just let the editor use soft line-wrapping.

Regarding the merging. Since anyway this will have to be moved once we have the new skeleton, there is no rush right? Since it is purely adding new text, there won't be any merge conflicts. So I actually would hold off with merging this until the skeleton is inplace, and then we can add it in the correct place straight away. Doing the same with my open docs PR #3966

chrisjsewell:
> Please maintain one sentence per line for text.

Fixed in 3cc9a03

CasperWA:
> Regarding the merging. Since anyway this will have to be moved once we have the new skeleton, there is no rush right? Since it is purely adding new text, there won't be any merge conflicts. So I actually would hold off with merging this until the skeleton is inplace, and then we can add it in the correct place straight away. Doing the same with my open docs PR #3966

Fine. Then I'll add it to the project instead.

csadorf:
@chrisjsewell Could you adapt this PR so that we can merge it into docs-revamp?

chrisjsewell:
> Could you adapt this PR so that we can merge it into docs-revamp?

On it now 👍

chrisjsewell:
rebased to #4126

PR #3953, opened 2020-04-21
restapi: silence deprecation warning

ltalirz:
The deprecation warning of the `hookup` argument was triggered also
during regular use since the REST API CLI was setting it to a default
value.
Now setting it to `None` in order to avoid the deprecation warning.

### DIFF
--- a/aiida/cmdline/commands/cmd_restapi.py
+++ b/aiida/cmdline/commands/cmd_restapi.py
@@ -37,13 +37,7 @@ from aiida.restapi.common import config
     default=config.CLI_DEFAULTS['WSGI_PROFILE'],
     help='Whether to enable WSGI profiler middleware for finding bottlenecks'
 )
-@click.option(
-    '--hookup/--no-hookup',
-    'hookup',
-    is_flag=True,
-    default=config.CLI_DEFAULTS['HOOKUP_APP'],
-    help='Hookup app to flask server'
-)
+@click.option('--hookup/--no-hookup', 'hookup', is_flag=True, default=None, help='Hookup app to flask server')
 def restapi(hostname, port, config_dir, debug, wsgi_profile, hookup):
     """
     Run the AiiDA REST API server.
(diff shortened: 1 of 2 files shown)


ltalirz:
@sphuber Sorry for the many things on this - I'm juggling a few things at the moment..

It turns out that the deprecation warning would now have always been printed and we'd like to avoid that.
As for the corresponding cli option, is there a specific way to mark it as deprecated?
otherwise, using it will now anyhow result in the deprecation warning being shown.

sphuber:
> Sorry for the many issues with this - I'm juggling a few things at the moment..

No worries, same here, and I also missed it in the review.

> As for the corresponding cli option, is there a specific way to mark it as deprecated?
> otherwise, using it will now anyhow result in the deprecation warning being shown.
There is no standard way that I am aware off. We could think of making a custom `Option` that takes care of displaying the message, but I think this is overengineering it. Simply set the default to `None` as you did and then checking it when passed in the command body is fine. I have done this elsewhere as well

ltalirz:
Ok, I tested it and it works - i.e. the warning is shown only if the cli option is explicitly used (and otherwise it works fine).

To be honest, I really don't understand what the point of the `--hookup` option in the cli is.
If you don't start a server, the python function will return the `(app, api)` tuple but there is no way for a user to do anything with it.
Perhaps it's better to remove it altogether; I really can't imagine who could have used this in the cli.

ltalirz:
Ok, I see now - it may have been introduced in order to be able test the cli... hm.

ltalirz:
> I would suggest to add a separate deprecation message in the CLI saying that the option will be removed and then just not even pass the value to run_api.

I think it's ok as is - this is a case where I really can't imagine anybody having ever used this option and if they do they will see the deprecation message.

I do wonder a bit how we are going to test it after we remove it... do we start the server, check stdout and then send ctrl+c?

sphuber:
Since the command line is literally just forwarding the options to the `run_api` call, I think we can get away with very light testing. Just check that the command is reachable with `-h` would already be fine. Then we could always just test the options themselves independently of the command.

PR #3954, opened 2020-04-21
Parse complex chemical formula strings

lorisercole:
The "parse_formula" function can now parse more complex formulas, including elements with partial occupancies and groups of elements enclosed in parentheses. 
Spaces separators are no more needed. 
Elements are counted and a dictionary is returned.
This can be useful to parse the formulas included in cif files (for example to debug a cif parser), or those generated by pymatgen.

e.g.  C[NH2]3NO3  -->  {'C': 1, 'N': 4, 'H': 6, 'O': 3}


aiida/orm/nodes/data/cif.py
```diff
@@ -196,27 +196,48 @@ def pycifrw_from_cif(datablocks, loops=None, names=None):
...
     import re
 
+    def chemcount_str_to_number(string):
+        if (string is None) or (len(string) == 0):
+            quantity = 1
+        else:
+            if re.match(r'^\d+$', string):  # check if it is an integer number
+                quantity = int(string)
+            else:
+                quantity = float(string)
```
CasperWA, line 213:
```suggestion
        elif re.match(r'^\d+$', string):  # check if it is an integer number
            quantity = int(string)
        else:
            quantity = float(string)
```

CasperWA: One could also simply first make it a `float` and then use the `float` method `is_integer()` to determine whether it should be cast to an `int` or not.

aiida/orm/nodes/data/cif.py
```diff
@@ -196,27 +196,48 @@ def pycifrw_from_cif(datablocks, loops=None, names=None):
...
+    for block in re.split(r'(\([^\)]*\)[^A-Z\(\[\{]*|\[[^\]]*\][^A-Z\(\[\{]*|\{[^\}]*\}[^A-Z\(\[\{]*)', formula):
+        if not block:  # block is void
             continue
 
-        specie = m.group(1)
-        quantity = m.group(2)
-        if quantity is None:
-            quantity = 1
+        # get molecular formula (within parentheses) & count
+        g = re.search(r'[\{\[\(](.+)[\}\]\)]([\.\d]*)', block)  # pylint: disable=invalid-name
```
CasperWA, line 224:
```suggestion
        group = re.search(r'[\{\[\(](.+)[\}\]\)]([\.\d]*)', block)
```

aiida/orm/nodes/data/cif.py
```diff
@@ -196,27 +196,48 @@ def pycifrw_from_cif(datablocks, loops=None, names=None):
...
+        if g is None:  # block does not contain parentheses
```
CasperWA, line 225:
```suggestion
        if group is None:  # block does not contain parentheses
```

aiida/orm/nodes/data/cif.py
```diff
@@ -196,27 +196,48 @@ def pycifrw_from_cif(datablocks, loops=None, names=None):
...
+            molformula = block
+            molcount = 1
         else:
-            if re.match(r'^\d+$', quantity):
-                quantity = int(quantity)
-            else:
-                quantity = float(quantity)
-        contents[specie] = quantity
+            molformula = g.group(1)
+            molcount = chemcount_str_to_number(g.group(2))
```
CasperWA, line 230:
```suggestion
            molformula = group.group(1)
            molcount = chemcount_str_to_number(group.group(2))
```

aiida/orm/nodes/data/cif.py
```diff
@@ -196,27 +196,48 @@ def pycifrw_from_cif(datablocks, loops=None, names=None):
...
+
+        for part in re.findall(r'[A-Z][^A-Z\s]*', molformula.replace(' ', '')):  # split at uppercase letters
+            m = re.match(r'(\D+)([\.\d]+)?', part)  # separates element and count  # pylint: disable=invalid-name
```
CasperWA, line 233:
```suggestion
            match = re.match(r'(\D+)([\.\d]+)?', part)  # separates element and count
```

aiida/orm/nodes/data/cif.py
```diff
@@ -196,27 +196,48 @@ def pycifrw_from_cif(datablocks, loops=None, names=None):
...
+
+            if m is None:
```
CasperWA, line 235:
```suggestion
            if match is None:
```

aiida/orm/nodes/data/cif.py
```diff
@@ -196,27 +196,48 @@ def pycifrw_from_cif(datablocks, loops=None, names=None):
...
+                continue
+
+            specie = m.group(1)
+            quantity = chemcount_str_to_number(m.group(2)) * molcount
+            contents[specie] = contents.get(specie, 0) + quantity
```
CasperWA, line 240: While pymatgen as well as the old version of this function may use "specie", it is grammatically incorrect.

```suggestion
            species = match.group(1)
            quantity = chemcount_str_to_number(match.group(2)) * molcount
            contents[species] = contents.get(species, 0) + quantity
```

aiida/orm/nodes/data/cif.py
```diff
@@ -196,27 +196,48 @@ def pycifrw_from_cif(datablocks, loops=None, names=None):
 
 def parse_formula(formula):
     """
-    Parses the Hill formulae, written with spaces for separators.
+    Parses the Hill formulae. Does not need spaces as separators.
+    Works also for partial occupancies and for chemical groups enclosed in round/square/curly brackets.
+    Elements are counted and a dictionary is returned.
+    e.g.  'C[NH2]3NO3'  -->  {'C': 1, 'N': 4, 'H': 6, 'O': 3}
     """
     import re
```
CasperWA, line 204: Move this import out of this function to the top of the file.

aiida/orm/nodes/data/cif.py
```diff
@@ -196,27 +196,48 @@ def pycifrw_from_cif(datablocks, loops=None, names=None):
...
 
+    def chemcount_str_to_number(string):
+        if (string is None) or (len(string) == 0):
```
CasperWA, line 207:
```suggestion
        if not string:
```

lorisercole:
OK, thanks guys.
I implemented the suggestions by @CasperWA and added some tests.
The parser should work for the chemical formulas that can be read from COD, ICSD, MPDS cif files, as well as those generated by the many modes of `StructureData.get_formula`, by pymatgen, and ase.

In the commit e0c6704ee35d7820d8084a10fe807f964b514f0b I also edited the function `CifData.get_formulae` to be able to read the chemical formulas of MPDS cif files, which use a different tag (only available for v1.0.5+ files, recently updated).

aiida/orm/nodes/data/cif.py
```diff
@@ -196,27 +197,46 @@ def pycifrw_from_cif(datablocks, loops=None, names=None):
...
+                quantity = int(quantity)
+        return quantity
 
     contents = {}
-    for part in re.split(r'\s+', formula):
-        m = re.match(r'(\D+)([\.\d]+)?', part)
 
-        if m is None:
+    # split blocks with parentheses
+    for block in re.split(r'(\([^\)]*\)[^A-Z\(\[\{]*|\[[^\]]*\][^A-Z\(\[\{]*|\{[^\}]*\}[^A-Z\(\[\{]*)', formula):
```
sphuber, line 218: I see that you are deviating from a number of syntax rules that are [defined rigorously for the `_chemical_formula_*` tag](https://www.iucr.org/__data/iucr/cifdic_html/1/cif_core.dic/Cchemical_formula.html) of the CIF format. For example, the spaces between elements and element groups are required in the spec. Also, only parentheses are allowed to group elements, not curly brackets. So in a sense, you are now parsing chemical formula tags that are technically invalid CIF format. I am not sure whether we should do this.

aiida/orm/nodes/data/cif.py
```diff
@@ -536,12 +556,14 @@ def get_formulae(self, mode='sum'):
         """
         # note: If formulae are not None, they could be returned
         # directly (but the function is very cheap anyhow).
-        formula_tag = '_chemical_formula_{}'.format(mode)
+        formula_tags = ('_chemical_formula_{}'.format(mode), '_pauling_file_chemical_formula')
```
sphuber, line 559: I am also not sure if we should build in support in such a hardcoded way for a tag that is custom to just one particular external database. I am not sure why they decided to have their own tag and if that respects the same spec as the official CIF `_chemica_formula_*` tag.

tests/orm/data/test_cif.py
```diff
@@ -0,0 +1,56 @@
...
+        'N': 8,
+        'Cl': 8.0,
+        'O': 8
+    }
+    assert parse_formula('Na1.28[NH]0.28{N H2}0.72') == {'Na': 1.28, 'N': 1.0, 'H': 1.72}
+
+    with pytest.raises(ValueError):
+        parse_formula('H0.5.2 O')
+        parse_formula('Fe2.05Ni0.05.4')
+        parse_formula('Na1.28[NH]0.28.3{NH2}0.72')
```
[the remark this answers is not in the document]
CasperWA: E.g., in a `for`-loop, iterating over a list, tuple, or set containing each of these formula strings.
```suggestion
    for test_formula in ['H0.5.2 O', 'Fe2.05Ni0.05.4', 'Na1.28[NH]0.28.3{NH2}0.72']:
        with pytest.raises(ValueError):
            parse_formula(test_formula)
```

aiida/orm/nodes/data/cif.py
CasperWA: There are (unfortunately) many of these "custom" CIF tags around. It's very good to include them and be flexible in this way, however, it should not be "hidden" within a general method like this, I think.
It would be good to dedicate special utility functions, dictionaries, methods or other to each _special_ list of tags, so they can be retrieved in this general method (or not).
Other than the `pauling` and `MPDS` specials, I know of `PDB` (Protein Data Bank) who have a very large number of special CIF tags (and even a dedicated format (PDBx/mmCIF), still using the `.cif` file type. These are listed in a special library/dictionary at [mmcif.wwpdb.org](http://mmcif.wwpdb.org/pdbx-mmcif-home-page.html).

aiida/orm/nodes/data/cif.py
lorisercole: Yeah, this regex is quite more general than what one should (in principle) find in CIF files.
Nevertheless, I found that chemical formulas in MPDS cif files do not have spaces between elements and may use square brackets (e.g. `LiSc[SeO3]2`).

Also, I wanted this function to be able to parse formulas generated by `StructureData.get_formula()`, by pymtagen, and ase. These may not contain spaces.
In the case of partial occupancies, `StructureData.get_formula()` will return a formula with curly brackets.
I thought it would be useful to have one function that does all of this: for example I am using it to detect possible errors in the CIF parsing (I will make a PR to `aiida-codtools` with this soon). If you think it is useful, maybe it can be moved somewhere else, as it is not limited to CIF chemical formulas only. Otherwise we can specialize it, and one can write a more general parser for other applications.

tests/orm/data/test_cif.py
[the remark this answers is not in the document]
lorisercole: Ok, I wasn't sure of the behavior, I'll update it. Thanks

aiida/orm/nodes/data/cif.py
lorisercole: I guess one could identify the database from the CIF file and subsequently read the relevant tags, e.g. `_pauling_file_chemical_formula` in the case of MPDS.

aiida/orm/nodes/data/cif.py
lorisercole: It is also true that one could avoid parsing the formula of a `StructureData` or a pymatgen structure, as he could obtain the set of elements and their quantities directly from the list of sites...

sphuber: As you stated yourself afterwards, it would be a bit of a merry-go-round to parse the string returned by `StructureData.get_formula()` when all information can be gotten directly from the class itself in other more efficient ways. I do see now, that even though the `parse_formula` is in the `cif` module, it was not explicitly marked as only parsing the CIF `_chemical_formula` tag, but merely Hill notation formulae in general. So I guess it is ok to keep this more flexible version.

aiida/orm/nodes/data/cif.py
sphuber: Where I think making the `parsing_formula` method more general is OK, since here we are talking strictly about parsing the CIF format, I would not hardcode these tags like this, because then we would potentially have to keep adding loads of them. How about, we add an optional argument `custom_tags=None` which can take a list of optional formula tags that should also be searched for?

lorisercole: OK, this seems to be a good solution.

lorisercole:
Thanks. I updated the branch following your suggestions.

ltalirz:
Is there somewhere a specification of the "more complex" formula format we are supporting here?

lorisercole:
@ltalirz see 9ead2b1162d7ec2d034f451ade22d482b3f58e5b:
```
The "parse_formula" function can now parse more complex formulas,
including elements with partial occupancies and groups of elements enclosed
in parentheses. Spaces separators are no more needed.
Elements are counted and a dictionary is returned.
This can be useful to parse the formulas included in cif files, or those
generated by pymatgen.

e.g.  C[NH2]3NO3  -->  {'C': 1, 'N': 4, 'H': 6, 'O': 3}
```
More (weird) examples in the tests:
https://github.com/aiidateam/aiida-core/blob/862fa2085094f975d6db4c70357bd74b1c3ea56c/tests/orm/data/test_cif.py#L17-L54

ltalirz:
Thanks, so I see it includes square brackets, curly braces as well as round braces, with and without spaces; integers, floating point numbers etc.
Is this following a standard specified somewhere or is this something you "invented"?

I'm just mentioning this because it can be very useful to tell people that this field is guaranteed to support formulas of format X (ideally X is fully specified in itself; a second-best option would be X = specification A + features B,C)

lorisercole:
Yes it supports round, square, and curly brackets.
- round brackets are used in the chemical formula tags of some CIF files, such as from ICSD
- square brackets are found in the chemical formula tags of some CIF files, such as from MPDS
- curly brackets are used in the formulas returned by `StructureData.get_formula()`
- Spaces are sometimes used, sometimes not.

The tests contain intentionally more general/complex cases than one should ever find (this would even allow one to parse formulas written "by hand" in weird ways, for whatever reason).
The parser counts all the elements and returns a dictionary of them. If an element is present in several bracketed groups, the sum will be returned, e.g.
```
parse_formula('Na1.28[NH]0.28{N H2}0.72')  -->  {'Na': 1.28, 'N': 1.0, 'H': 1.72}
```

PR #3958, opened 2020-04-22
add github action of transifex upload

unkcpz:
Migrate transifex integration from travis.ci to github action. @csadorf 

I am not sure whether `TRANSIFEX_USER` and `TRANSIFEX_PASSWORD` set previously still available. You may need to set them if not.


.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
...
+    - uses: actions/checkout@v2
+
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+
+    - name: Install python dependencies
+      run: |
+        pip install numpy==1.17.4
```
unkcpz, line 25: Not sure about this, just copy and paste from other jobs.

csadorf: There are some issues installing AiiDA when numpy is not already installed. Those are related to broken build setups of dependencies and should be mostly resolved, please see discussion on #3944 for example. In this case I suggest that we follow the [installation steps of the docs job](https://github.com/aiidateam/aiida-core/blob/d6e8c7d10537b754d1cb6334689ba2edf4347d43/.github/workflows/ci.yml#L7-L30) as part of the `continuous-integration` workflow (I assume that's what you did). At some point we can try to simplify the installation for the docs-related jobs, but that can be a separate PR.

.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
+name: release
+
+on:
+  release:
+    types: [published, edited]
+
+jobs:
+
+  upload-transifex:
```
csadorf, line 9: Can you add a brief comment here that describes the purpose of this job a little bit? Maybe even add a link to the portal where the translation is happening?

.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    name: Upload pot files to transifex
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
```
csadorf, line 15: Can you add a comment here that states that the installation steps are modeled after the `docs` job as part of the `continuous-integration` workflow?

.github/workflows/release.yml
unkcpz: Thanks for explanation.

.github/workflows/release.yml
unkcpz: done

unkcpz:
There seems no way to manually trigger the action. So have to wait until next release to make sure this workflow is working properly. :crossed_fingers:

unkcpz:
I recall that the `TRANSIFEX_USER` shall be set as `api` and `TRANSIFEX_PASSWORD` is the token generated from transifex. fyi, https://docs.transifex.com/api/introduction#authentication

And API token: https://www.transifex.com/user/settings/api/

As a reference, the context of my `transifexrc` file is:
```
[https://www.transifex.com]
api_hostname = https://api.transifex.com
hostname = https://www.transifex.com
password = 1/980000000000dfdfdfdfdf
username = api
```

csadorf:
> I recall that the `TRANSIFEX_USER` shall be set as `api` and `TRANSIFEX_PASSWORD` is the token generated from transifex. fyi, https://docs.transifex.com/api/introduction#authentication
> 
> And API token: https://www.transifex.com/user/settings/api/
> 
> As a reference, the context of my `transifexrc` file is:
> 
> ```
> [https://www.transifex.com]
> api_hostname = https://api.transifex.com
> hostname = https://www.transifex.com
> password = ****
> username = api
> ```

@unkcpz  Did you post the actual credentials? In this case, please reset this password **immediately** and send the new password to my personal email. Thank you.

unkcpz:
@csadorf no is just a example, I don't have the privilege to log in aiida account in transifex. @ltalirz has the actual credentials.

csadorf:
> I trigger it in a hack way and it failed, maybe since the secret keys are not set.

@unkcpz It might be a good idea for us to debug this workflow before we merge it by simply running it temporarily on `push`. It would only be executed for this branch in this case. We would then remove that trigger prior to merge after we have verified that it works. What do you think?

@ltalirz Can you verify that the secrets have been added to the repository?

ltalirz:
Sorry guys, I've added them now (user: api, password: the api token).
Please try again

unkcpz:
It is strange. I can successfully upload the pot to my account, but with all similar processes, still failed when setting and uploading to aiida account.
 
I got: `tx ERROR: EOF when reading a line` . This is most likely caused by incorrectly setting configuration file.

@ltalirz Can you reconfirm you add and set secrets:
```
TRANSIFEX_PASSWORD -> token from transifex
TRANSIFEX_USER -> api
```
Thanks!

ltalirz:
Hi @unkcpz , sorry about this, I believe the token I had stored was missing a few characters.
I've generated a new one and updated the github secrets.

I've also changed your status in the aiida docs to project maintainer in case it is useful at any point.

unkcpz:
emmm... unfortunately, same error here. 

Maybe, it is not allowed to using secrets keys in pull request actions? It cause anyone can access and use the secret token by just making a pull request, which is unsafe.

> I've also changed your status in the aiida docs to project maintainer in case it is useful at any point.

Thanks! :smile:

csadorf:
> Maybe, it is not allowed to using secrets keys in pull request actions? It cause anyone can access and use the secret token by just making a pull request, which is unsafe. In this case, could you help to make a push for test?

Ah sorry that I missed that. The secret is indeed not available for pull requests. We will need to push directly to the repository. I'll do that.

csadorf:
@unkcpz Just pushed to `ci-test-transifex-action` branch, action running here: https://github.com/aiidateam/aiida-core/actions/runs/86005424

ltalirz:
Seems to have worked, right?

unkcpz:
It works now. Cheers!  And change trigger event back to release, ready to be merged. 

Thank you guys! :smiley:

unkcpz:
However, I concerned if `release` event can successfully run the action as `push`? I didn't find information about which event is able to use secrets.  @csadorf

csadorf:
> However, I concerned if `release` event can successfully run the action as `push`? I didn't find information about which event is able to use secrets. @csadorf

The _release_ event is going to trigger the workflow with the appropriate permissions. So don't worry about that.

PR #3962, opened 2020-04-24
Test update dependency flow click 7.1

csadorf:
To test the revised 'update-dependency' workflow. Do not merge.


csadorf:
Successfully tested the `update-dependency` flow. Seems to be working as expected.

PR #3968, opened 2020-04-25
Fix exception handling in commands calling `list_repository_contents`

### LINKED ISSUE #3967: Commands `verdi node repo ls` and `verdi calcjob inputls/outputls` will not catch correct exception
They are all catching `ValueError` from the `aiida.cmdline.utils.repository.list_repository_contents` utility function, but I think this will at best throw a `FileNotFoundError` if the path is invalid and never a `ValueError`. What's more, the error handling will incur another exception. The exception is passed raw to the `echo_critical` function, which will then also except. The exception should first be cast to a string, for example through `echo_critical(str(exception))`. There are multiple occurrences of this error across the CLI.

sphuber:
Fixes #3967 

The commands `verdi node repo ls` and `verdi calcjob inputls/outputls`
were passing the caught exception object straight into the echo call
without casting it to a string first, which would incur another
exception by itself. On top of that, they were catching `ValueError`
exceptions, which, as far as I can tell, would never be thrown. Instead
if the path does not exist a `FileNotFoundError` will be thrown. The
commands now catch this instead and print a useful message. This failure
path of the commands was not tested, so the bug went unnoticed.

The problem of raw exception objects being passed to `echo_critical`
were present in some other commands as well, which have now been fixed.

### DIFF
--- a/aiida/cmdline/commands/cmd_calcjob.py
+++ b/aiida/cmdline/commands/cmd_calcjob.py
@@ -176,8 +176,8 @@ def calcjob_inputls(calcjob, path, color):
 
     try:
         list_repository_contents(calcjob, path, color)
-    except ValueError as exception:
-        echo.echo_critical(exception)
+    except FileNotFoundError:
+        echo.echo_critical('the path `{}` does not exist for the given node'.format(path))
 
 
 @verdi_calcjob.command('outputls')
@@ -203,8 +203,8 @@ def calcjob_outputls(calcjob, path, color):
 
     try:
         list_repository_contents(retrieved, path, color)
-    except ValueError as exception:
-        echo.echo_critical(exception)
+    except FileNotFoundError:
+        echo.echo_critical('the path `{}` does not exist for the given node'.format(path))
 
 
 @verdi_calcjob.command('cleanworkdir')
--- a/aiida/cmdline/commands/cmd_daemon.py
+++ b/aiida/cmdline/commands/cmd_daemon.py
@@ -71,7 +71,7 @@ def start(foreground, number):
         subprocess.check_output(command, env=currenv, stderr=subprocess.STDOUT)  # pylint: disable=unexpected-keyword-arg
     except subprocess.CalledProcessError as exception:
         click.secho('FAILED', fg='red', bold=True)
-        echo.echo_critical(exception.output)
+        echo.echo_critical(str(exception))
 
     # We add a small timeout to give the pid-file a chance to be created
     with spinner():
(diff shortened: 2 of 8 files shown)


aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -176,8 +176,8 @@ def calcjob_inputls(calcjob, path, color):
 
     try:
         list_repository_contents(calcjob, path, color)
-    except ValueError as exception:
-        echo.echo_critical(exception)
+    except FileNotFoundError:
```
ltalirz, line 179: the only comment I have is that this might become a different exception when we move to the new repository implementation.
is there perhaps a higher-level exception that we could catch here?

aiida/cmdline/commands/cmd_daemon.py
```diff
@@ -71,7 +71,7 @@ def start(foreground, number):
         subprocess.check_output(command, env=currenv, stderr=subprocess.STDOUT)  # pylint: disable=unexpected-keyword-arg
     except subprocess.CalledProcessError as exception:
         click.secho('FAILED', fg='red', bold=True)
-        echo.echo_critical(exception.output)
+        echo.echo_critical(str(exception))
```
ltalirz, line 74: Or use `repr`, see https://stackoverflow.com/a/4308203/1069467

sphuber: I used `str` since that is what is intended for normal users, whereas `repr` is for debugging purposes. I think in this case, showing the exception type for the user is not necessary. But if you think this would actually be better, happy to change it

aiida/cmdline/commands/cmd_calcjob.py
sphuber: The tests I have added will then exactly catch that change and we can adapt it. Don't see the point in trying to anticipate the type of exception that this might throw in the future. Or did you mean to catch a more generic exception like `OSError` which is the [base class](https://docs.python.org/3/library/exceptions.html#FileNotFoundError) of `FileNotFoundError`?

ltalirz: right, I meant more generic. 
Anyhow, if you think we should just use the one we have currently it's also ok. the tests will fail if this changes, so it will be fixed then.

aiida/cmdline/commands/cmd_daemon.py
ltalirz: I see... up to you.

ltalirz:
good to go from me as well

PR #3738, opened 2020-01-31
Clean up top level directory of repository

ltalirz:
fix #3730

PR #3974, opened 2020-04-27
Properly handle multiple requests to threaded REST API

### LINKED ISSUE #3743: Make sure threads are handled correctly for the REST API

CasperWA:
Fixes #3743 

This PR implements the same fix as was used for [`aiida-optimade`](https://github.com/aiidateam/aiida-optimade/pull/34), wherein it closes the SQLAlchemy session used by `QueryBuilder` after having run a method/function to handle a `GET` request to the REST API.

If this is not done, the session should(?) still eventually close, releasing connections made in the SQLA connection pool. However, this may take such a long time that when a new `GET` request comes in (or more precisely, whenever the `QueryBuilder` is used), the newly created thread that handles the request cannot use the singleton session that AiiDA has, since the connections have not been released, and so it fails with the following message:

```sh
sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30 (Background on this error at: http://sqlalche.me/e/3o7r)
```

At least that is how I remember the explanation @sphuber and I came up with some time ago. Hence @sphuber also cleaned up the whole session handling code in AiiDA and provided a public interface for closing the given session.

The major issue here is threading; the AiiDA SQLA session (that the `QueryBuilder` uses no matter the actual profile's database backend) is a singleton, and only meant to be used within the same (main) thread. Since the REST API runs threaded, as it should to accommodate multiple incoming requests effectively, the singleton AiiDA SQLA sessions design breaks down. However, by closing the session at the end of every request, the connections are immediately released and only in extreme cases will the `QueuePool` ever reach is limit. In fact, one may argue that in those cases, it _should_ complain, since something is probably wrong/you're flooding the database with too many requests.

This issue was never actually tested with the current tests, since a threaded server is never started.
I have added a new test that starts the REST API server in a new `Thread`, issues 100 requests in serial, and finally shuts the server down.
This resulted in a reproduction of the `QueuePool` error, which was then fixed after implementing my fix.

A side note: Since the REST API (as far as I know, please back me up here @waychal) only serves data from the database, and never alters or creates data in the database, there is no risk of corrupting the database by bombarding the REST API with requests.

Last thing: @sphuber, should the `close_session` decorator function be placed somewhere else in the code? I think I would like to perhaps use this directly in `aiida-optimade` as well, instead of re-implementing it.
Also, would I then have to move the `from aiida.manage.manager import get_manager` import line into the `finally:` section of the `close_session()` function again, where it was originally placed, in order to not unintentionally load the AiiDA profile before it's truly time?

### DIFF
--- a/aiida/backends/djsite/__init__.py
+++ b/aiida/backends/djsite/__init__.py
@@ -31,7 +31,7 @@ def reset_session():
     SESSION_FACTORY = None
 
 
-def get_scoped_session():
+def get_scoped_session(**kwargs):
     """Return a scoped session for the given profile that is exclusively to be used for the `QueryBuilder`.
 
     Since the `QueryBuilder` implementation uses SqlAlchemy to map the query onto the models in order to generate the
@@ -43,7 +43,11 @@ def get_scoped_session():
     Django implementation of the `QueryBuilder` should keep its own SqlAlchemy engine and scoped session factory
     instances that are used to provide the query builder with a session.
 
-    :param profile: :class:`aiida.manage.configuration.profile.Profile` for which to configure the engine.
+    :param kwargs: keyword arguments that will be passed on to :py:func:`aiida.backends.utils.create_sqlalchemy_engine`,
+        opening the possibility to change QueuePool time outs and more.
+        See https://docs.sqlalchemy.org/en/13/core/engines.html?highlight=create_engine#sqlalchemy.create_engine for
+        more info.
+
     :return: :class:`sqlalchemy.orm.session.Session` instance with engine configured for the given profile.
     """
     from aiida.manage.configuration import get_profile
@@ -56,7 +60,7 @@ def get_scoped_session():
         return session
 
     if ENGINE is None:
-        ENGINE = create_sqlalchemy_engine(get_profile())
+        ENGINE = create_sqlalchemy_engine(get_profile(), **kwargs)
 
     SESSION_FACTORY = create_scoped_session_factory(ENGINE)
 
--- a/aiida/restapi/common/identifiers.py
+++ b/aiida/restapi/common/identifiers.py
@@ -31,8 +31,7 @@ Examples of invalid full types:
     'process.calculation%.calcfunction.%|aiida.calculations:arithmetic.add'  # More than one operator in segment
 
 """
-
-import collections
+from collections.abc import MutableMapping
 
 from aiida.common.escaping import escape_for_sql_like
 
@@ -163,7 +162,7 @@ def load_entry_point_from_full_type(full_type):
     raise EntryPointError('entry point of the given full type cannot be loaded')
 
 
-class Namespace(collections.MutableMapping):
+class Namespace(MutableMapping):
     """Namespace that can be used to map the node class hierarchy."""
 
     namespace_separator = '.'
--- a/aiida/restapi/common/utils.py
+++ b/aiida/restapi/common/utils.py
@@ -8,13 +8,15 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """ Util methods """
-import urllib.parse
 from datetime import datetime, timedelta
+import urllib.parse
 
 from flask import jsonify
 from flask.json import JSONEncoder
+from wrapt import decorator
 
 from aiida.common.exceptions import InputValidationError, ValidationError
+from aiida.manage.manager import get_manager
 from aiida.restapi.common.exceptions import RestValidationError, \
     RestInputValidationError
 
@@ -845,3 +847,17 @@ def list_routes():
         output.append(line)
 
     return sorted(set(output))
+
+
+@decorator
+def close_session(wrapped, _, args, kwargs):
+    """Close AiiDA SQLAlchemy (QueryBuilder) session
+
+    This decorator can be used for router endpoints to close the SQLAlchemy global scoped session after the response
+    has been created. This is needed, since the QueryBuilder uses a SQLAlchemy global scoped session no matter the
+    profile's database backend.
+    """
+    try:
+        return wrapped(*args, **kwargs)
+    finally:
+        get_manager().get_backend().get_session().close()
--- a/aiida/restapi/resources.py
+++ b/aiida/restapi/resources.py
@@ -15,12 +15,11 @@ from flask_restful import Resource
 
 from aiida.common.lang import classproperty
 from aiida.restapi.common.exceptions import RestInputValidationError
-from aiida.restapi.common.utils import Utils
+from aiida.restapi.common.utils import Utils, close_session
 
 
 class ServerInfo(Resource):
-    # pylint: disable=fixme
-    """Endpointd to return general server info"""
+    """Endpoint to return general server info"""
 
     def __init__(self, **kwargs):
         # Configure utils
@@ -97,6 +96,8 @@ class BaseResource(Resource):
     _translator_class = BaseTranslator
     _parse_pk_uuid = None  # Flag to tell the path parser whether to expect a pk or a uuid pattern
 
+    method_decorators = [close_session]  # Close SQLA session after any method call
+
     ## TODO add the caching support. I cache total count, results, and possibly
 
     def __init__(self, **kwargs):
@@ -106,11 +107,13 @@ class BaseResource(Resource):
         utils_conf_keys = ('PREFIX', 'PERPAGE_DEFAULT', 'LIMIT_DEFAULT')
         self.utils_confs = {k: kwargs[k] for k in utils_conf_keys if k in kwargs}
         self.utils = Utils(**self.utils_confs)
-        self.method_decorators = {'get': kwargs.get('get_decorators', [])}
+
+        # HTTP Request method decorators
+        if 'get_decorators' in kwargs and isinstance(kwargs['get_decorators'], (tuple, list, set)):
+            self.method_decorators = {'get': list(kwargs['get_decorators'])}
 
     @classproperty
-    def parse_pk_uuid(cls):
-        # pylint: disable=no-self-argument
+    def parse_pk_uuid(cls):  # pylint: disable=no-self-argument
         return cls._parse_pk_uuid
 
     def _load_and_verify(self, node_id=None):
@@ -211,17 +214,6 @@ class Node(BaseResource):
 
     _translator_class = NodeTranslator
     _parse_pk_uuid = 'uuid'  # Parse a uuid pattern in the URL path (not a pk)
-
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        from aiida.orm import Node as tNode
-        self.tclass = tNode
-
-        # Configure utils
-        utils_conf_keys = ('PREFIX', 'PERPAGE_DEFAULT', 'LIMIT_DEFAULT')
-        self.utils_confs = {k: kwargs[k] for k in utils_conf_keys if k in kwargs}
-        self.utils = Utils(**self.utils_confs)
-        self.method_decorators = {'get': kwargs.get('get_decorators', [])}
 
     def get(self, id=None, page=None):  # pylint: disable=redefined-builtin,invalid-name,unused-argument
         # pylint: disable=too-many-locals,too-many-statements,too-many-branches,fixme,unused-variable
--- a/tests/restapi/test_threaded_restapi.py
+++ b/tests/restapi/test_threaded_restapi.py
@@ -1,0 +1,121 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Tests for the `aiida.restapi` module, using it in threaded mode.
+
+Threaded mode is the default (and only) way to run the AiiDA REST API (see `aiida.restapi.run_api:run_api()`).
+This test file's layout is inspired by https://gist.github.com/prschmid/4643738
+"""
+import time
+from threading import Thread
+
+import requests
+import pytest
+
+NO_OF_REQUESTS = 100
+
+
+@pytest.mark.usefixtures('clear_database_before_test', 'restrict_sqlalchemy_queuepool')
+def test_run_threaded_server(restapi_server, server_url, aiida_localhost):
+    """Run AiiDA REST API threaded in a separate thread and perform many sequential requests"""
+
+    server = restapi_server()
+    computer_id = aiida_localhost.uuid
+
+    # Create a thread that will contain the running server,
+    # since we do not wish to block the main thread
+    server_thread = Thread(target=server.serve_forever)
+
+    try:
+        server_thread.start()
+
+        for _ in range(NO_OF_REQUESTS):
+            response = requests.get(server_url + '/computers/{}'.format(computer_id), timeout=10)
+
+            assert response.status_code == 200
+
+            try:
+                response_json = response.json()
+            except ValueError:
+                pytest.fail('Could not turn response into JSON. Response: {}'.format(response.raw))
+            else:
+                assert 'data' in response_json
+
+    except Exception as exc:  # pylint: disable=broad-except
+        pytest.fail('Something went terribly wrong! Exception: {}'.format(repr(exc)))
+    finally:
+        server.shutdown()
+
+        # Wait a total of 1 min (100 x 0.6 s) for the Thread to close/join, else fail
+        for _ in range(100):
+            if server_thread.is_alive():
+                time.sleep(0.6)
+            else:
+                break
+        else:
+            pytest.fail('Thread did not close/join within 1 min after REST API server was called to shutdown')
+
+
+@pytest.mark.usefixtures('clear_database_before_test', 'restrict_sqlalchemy_queuepool')
+def test_run_without_close_session(restapi_server, server_url, aiida_localhost, capfd):
+    """Run AiiDA REST API threaded in a separate thread and perform many sequential requests"""
+    from aiida.restapi.api import AiidaApi
+    from aiida.restapi.resources import Computer
+
+    class NoCloseSessionApi(AiidaApi):
+        """Add Computer to this API (again) with a new endpoint, but pass an empty list for `get_decorators`"""
+
+        def __init__(self, app=None, **kwargs):
+            super().__init__(app=app, **kwargs)
+
+            # This is a copy of adding the `Computer` resource,
+            # but only a few URLs are added, and `get_decorators` is passed with an empty list.
+            extra_kwargs = kwargs.copy()
+            extra_kwargs.update({'get_decorators': []})
+            self.add_resource(
+                Computer,
+                '/computers_no_close_session/',
+                '/computers_no_close_session/<id>/',
+                endpoint='computers_no_close_session',
+                strict_slashes=False,
+                resource_class_kwargs=extra_kwargs,
+            )
+
+    server = restapi_server(NoCloseSessionApi)
+    computer_id = aiida_localhost.uuid
+
+    # Create a thread that will contain the running server,
+    # since we do not wish to block the main thread
+    server_thread = Thread(target=server.serve_forever)
+
+    try:
+        server_thread.start()
+
+        for _ in range(NO_OF_REQUESTS):
+            requests.get(server_url + '/computers_no_close_session/{}'.format(computer_id), timeout=10)
+        pytest.fail('{} requests were not enough to raise a SQLAlchemy TimeoutError!'.format(NO_OF_REQUESTS))
+
+    except (requests.exceptions.ConnectionError, OSError):
+        pass
+    except Exception as exc:  # pylint: disable=broad-except
+        pytest.fail('Something went terribly wrong! Exception: {}'.format(repr(exc)))
+    finally:
+        server.shutdown()
+
+        # Wait a total of 1 min (100 x 0.6 s) for the Thread to close/join, else fail
+        for _ in range(100):
+            if server_thread.is_alive():
+                time.sleep(0.6)
+            else:
+                break
+        else:
+            pytest.fail('Thread did not close/join within 1 min after REST API server was called to shutdown')
+
+    captured = capfd.readouterr()
+    assert 'sqlalchemy.exc.TimeoutError: QueuePool limit of size ' in captured.err
--- a/aiida/backends/djsite/manager.py
+++ b/aiida/backends/djsite/manager.py
@@ -33,10 +33,19 @@ class DjangoBackendManager(BackendManager):
 
         return self._settings_manager
 
-    def _load_backend_environment(self):
-        """Load the backend environment."""
+    def _load_backend_environment(self, **kwargs):
+        """Load the backend environment.
+
+        The scoped session is needed for the QueryBuilder only.
+
+        :param kwargs: keyword arguments that will be passed on to :py:func:`aiida.backends.djsite.get_scoped_session`.
+        """
         os.environ['DJANGO_SETTINGS_MODULE'] = 'aiida.backends.djsite.settings'
         django.setup()  # pylint: disable=no-member
+
+        # For QueryBuilder only
+        from . import get_scoped_session
+        get_scoped_session(**kwargs)
 
     def reset_backend_environment(self):
         """Reset the backend environment."""
--- a/aiida/backends/manager.py
+++ b/aiida/backends/manager.py
@@ -106,20 +106,24 @@ class BackendManager:
         :return: `SettingsManager`
         """
 
-    def load_backend_environment(self, profile, validate_schema=True):
+    def load_backend_environment(self, profile, validate_schema=True, **kwargs):
         """Load the backend environment.
 
         :param profile: the profile whose backend environment to load
         :param validate_schema: boolean, if True, validate the schema first before loading the environment.
+        :param kwargs: keyword arguments that will be passed on to the backend specific scoped session getter function.
         """
-        self._load_backend_environment()
+        self._load_backend_environment(**kwargs)
 
         if validate_schema:
             self.validate_schema(profile)
 
     @abc.abstractmethod
-    def _load_backend_environment(self):
-        """Load the backend environment."""
+    def _load_backend_environment(self, **kwargs):
+        """Load the backend environment.
+
+        :param kwargs: keyword arguments that will be passed on to the backend specific scoped session getter function.
+        """
 
     @abc.abstractmethod
     def reset_backend_environment(self):
--- a/aiida/backends/sqlalchemy/__init__.py
+++ b/aiida/backends/sqlalchemy/__init__.py
@@ -31,11 +31,16 @@ def reset_session():
     SESSION_FACTORY = None
 
 
-def get_scoped_session():
+def get_scoped_session(**kwargs):
     """Return a scoped session
 
     According to SQLAlchemy docs, this returns always the same object within a thread, and a different object in a
     different thread. Moreover, since we update the session class upon forking, different session objects will be used.
+
+    :param kwargs: keyword argument that will be passed on to :py:func:`aiida.backends.utils.create_sqlalchemy_engine`,
+        opening the possibility to change QueuePool time outs and more.
+        See https://docs.sqlalchemy.org/en/13/core/engines.html?highlight=create_engine#sqlalchemy.create_engine for
+        more info.
     """
     from aiida.manage.configuration import get_profile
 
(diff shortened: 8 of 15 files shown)


tests/restapi/test_threaded_restapi.py
```diff
@@ -0,0 +1,87 @@
...
+        super(ThreadedRESTApiTestCase, self).setUp()
+
+        # Use the factory method to create the app.
+        _, self.api = configure_api(App, AiidaApi)
+
+        # Create server
+        self.server = make_server(
+            host=CLI_DEFAULTS['HOST_NAME'],
+            port=int(CLI_DEFAULTS['PORT']),
+            app=self.api.app,
```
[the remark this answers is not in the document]
ltalirz: > On a side note (not a problem of your PR, but maybe it can be simplified), since apparently the app is an attribute of the api, why is configure_api returning the tuple and not just the api. If someone needs the app they can get it through api.app. Returning the tuple is a bit pointless.

Good point. Since `configure_api` was introduced very recently (in 1.2.1), it might still be ok to change the return value since I doubt anybody has start using it outside aiida-core.
The return value of `run_api` should not be changed.

CasperWA:
> > Last thing: @sphuber, should the close_session decorator function be placed somewhere else in the code?
> 
> Any code path that will cause the `QueryBuilder` to be constructed by a thread should have this. I have the feeling that the `get` method would be a good place to catch all of these, but I am not sure. We should check with @waychal and @asle85

I didn't mean where to use the decorater, I believe this is already the best place where it is now.
I meant more the actual decorator code, i.e., the `close_session()` function. Should it be moved to `aiida.backends.sqlalchemy` instead, together with the rest of the session-related code?

> > Also, would I then have to move the from aiida.manage.manager import get_manager import line into the finally: section of the close_session() function again, where it was originally placed, in order to not unintentionally load the AiiDA profile before it's truly time?
> 
> You can import the `get_manager` function safely without having to have loaded a profile.

Great! Thanks 👍

aiida/restapi/common/identifiers.py
```diff
@@ -31,8 +31,7 @@
     'process.calculation%.calcfunction.%|aiida.calculations:arithmetic.add'  # More than one operator in segment
 
 """
-
-import collections
+import collections.abc as collections
```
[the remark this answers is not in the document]
CasperWA: 👍 I thought this would be the least obstructive way, but will change the file a bit further than a one-liner

tests/restapi/test_threaded_restapi.py
```diff
@@ -0,0 +1,87 @@
...
+from aiida.restapi.run_api import configure_api
+
+# The URL to the REST API that we will start up. This is the default location
+# where Flask apps start when using the built in web server.
+SERVER_URL = 'http://{hostname}:{port}{api}'.format(
+    hostname=CLI_DEFAULTS['HOST_NAME'], port=CLI_DEFAULTS['PORT'], api=API_CONFIG['PREFIX']
+)
+
+
+class ThreadedRESTApiTestCase(AiidaTestCase):
```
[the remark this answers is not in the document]
CasperWA: Okay - and there will not be an issue with the automagically created test profile, database, and computer?

aiida/restapi/resources.py
```diff
@@ -125,6 +125,7 @@ def _load_and_verify(self, node_id=None):
 
         return node
 
+    @close_session
```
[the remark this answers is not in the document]
CasperWA: Sure. Was looking for the most "upstream" way of introducing this session-decorator as well, so perhaps this is the best way. Can you confirm that perhaps @waychal?

tests/restapi/test_threaded_restapi.py
```diff
@@ -0,0 +1,87 @@
...
+    """A TestCase to test the thread safety of the AiiDA REST API"""
+
+    def setUp(self):
+        """Setup to run before each test case."""
+        super(ThreadedRESTApiTestCase, self).setUp()
+
+        # Use the factory method to create the app.
+        _, self.api = configure_api(App, AiidaApi)
```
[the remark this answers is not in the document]
CasperWA: I just wanted it to be as close to the "real" REST API that is started when running `verdi restapi` as possible.

tests/restapi/test_threaded_restapi.py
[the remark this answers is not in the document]
CasperWA: Agreed, I wondered this myself as well.
And concerning your comment of using `api.app`, again, it is to be as close to the way `verdi restapi` runs the REST API as possible. In the code it runs the `app` exactly this way, i.e., `self.api.app.run()`.

sphuber:
> I didn't mean where to use the decorater, I believe this is already the best place where it is now.
> I meant more the actual decorator code, i.e., the `close_session()` function. Should it be moved to `aiida.backends.sqlalchemy` instead, together with the rest of the session-related code?

Ah I see, no it shouldn't go to `aiida.backends.sqlalchemy` because it is also need when you run with a django backend. This is why I created the `get_session` on the `Backend` class, such that it can be gotten in a backend independent way. Since this is the offical API to dealing with the API, we don't need to add additional code.

CasperWA:
> > I didn't mean where to use the decorater, I believe this is already the best place where it is now.
> > I meant more the actual decorator code, i.e., the `close_session()` function. Should it be moved to `aiida.backends.sqlalchemy` instead, together with the rest of the session-related code?
> 
> Ah I see, no it shouldn't go to `aiida.backends.sqlalchemy` because it is also need when you run with a django backend. This is why I created the `get_session` on the `Backend` class, such that it can be gotten in a backend independent way. Since this is the offical API to dealing with the API, we don't need to add additional code.

I'll try and ask again then: Do you think the `close_session()` should be moved somewhere else in the code than where it is now? (And if yes, where?) Because I would _not_ consider it an intrinsic part of the REST API, but rather belonging to AiiDA core and its session handling.

sphuber:
> I'll try and ask again then: Do you think the `close_session()` should be moved somewhere else in the code than where it is now? (And if yes, where?) Because I would _not_ consider it an intrinsic part of the REST API, but rather belonging to AiiDA core and its session handling.

I have answered this in the PR comments themselves. I suggested defining the decorator on the `BaseResource` class itself, as it is there where it is used. When I refactored the session handling code, I decided to provide `get_session` and not just `close_session` because maybe the caller would need to do have access to the session. Since closing the session is as simple as `get_session().close()` I didn't feel it was necessary to have an explicit `close_session` method. Since the context manager you wrote is just one of the many potential ways to deal with the problem at hand, I don't really see why this has to be some standard utility in AiiDA.

tests/restapi/test_threaded_restapi.py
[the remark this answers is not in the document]
CasperWA: Also, I intended to add more tests here, but found I didn't need to, i.e., the `self` is a leftover of this idea.
I will have to redo all this if moving to the `pytest` test design in any case.

CasperWA:
> I have answered this in the PR comments themselves. (...)

LazyMan (aka. me ...) strikes again!

LazyMan strikes again!

CasperWA:
With the latest commits I have moved to the `pytest` way of testing, as well as addressed all other review comments by @sphuber.
Furthermore, an extra commit adds another test - I thought this was relevant to actually validate the result of the "first" test, in the sense that it tests a REST API endpoint/resource _without_ the `close_session` decorator.
The issue with this test is the default time-out for `QueuePool` connections (30 s), since it means the test will at minimum _always_ take this long ... I haven't found a way to set the time-out without also getting a setup that doesn't properly reflect how the REST API will actually run.

aiida/restapi/resources.py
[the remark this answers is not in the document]
CasperWA: I have gone for a compromise, since I discovered the `Resource` attribute `method_decorators`, which I have initialized in the `BaseResource` to always include `close_session`.
This makes for "easy" testing as well of a REST API without the use of `close_session`, by simply sub-classing an existing `Resource` and setting the attribute to `None` (or equivalent).

CasperWA:
@sphuber, it seems the fixture `aiida_localhost` doesn't work as expected?

ltalirz:
This is likely related to the discussion [here](https://github.com/aiidateam/aiida-core/pull/3738#issuecomment-607687961).

In essence, I think the issue is rooted in interference between the pytest fixtures and the remaining AiiDATestCase implementation.
In the PR I linked we decided to remove the change again since we wanted to release. I wrote

> Note, however, that we will likely run into this issue again soon, namely when people start to use pytest fixtures in the test suite more actively.

aiida/restapi/resources.py
```diff
@@ -96,6 +96,7 @@ class BaseResource(Resource):
 
     _translator_class = BaseTranslator
     _parse_pk_uuid = None  # Flag to tell the path parser whether to expect a pk or a uuid pattern
+    _method_decorators = [close_session]  # Close SQLA session after any method call
```
sphuber, line 99: That is a nifty feature of `flask` and a nice solution. Since we want this decorator to apply to all resource methods and all subclasses, why not just define it straight as class attribute `method_decorators = [close_session]` here instead of applying this private attribute in the constructor? This way there is also no problems with subclasses forgetting to call `super` if they override the constructor

aiida/restapi/common/utils.py
```diff
@@ -845,3 +847,17 @@ def list_routes():
         output.append(line)
 
     return sorted(set(output))
+
+
+@decorator
+def close_session(wrapped, _, args, kwargs):
```
sphuber, line 853: I guess you are dead-set on keeping this here as a utility then? :)

aiida/restapi/common/utils.py
```diff
@@ -8,13 +8,15 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """ Util methods """
-import urllib.parse
 from datetime import datetime, timedelta
+import urllib.parse
+from wrapt import decorator
```
sphuber, line 13: minor nitpick: this is third-party library and so should go after `from flask.json` (surprised this is not picked up by `pylint`)

aiida/restapi/resources.py
```diff
@@ -212,17 +215,6 @@ class Node(BaseResource):
     _translator_class = NodeTranslator
     _parse_pk_uuid = 'uuid'  # Parse a uuid pattern in the URL path (not a pk)
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        from aiida.orm import Node as tNode
-        self.tclass = tNode
```
sphuber, line 218: I cannot see why it is ok to remove these lines? Where is this being set for the `Node` base resource?

tests/restapi/test_threaded_restapi.py
```diff
@@ -0,0 +1,140 @@
...
+from aiida.restapi.run_api import configure_api
+
+# The URL to the REST API we will start up.
+# This is the default location where Flask apps start when using the built in web server.
+SERVER_URL = 'http://{hostname}:{port}{api}'.format(
+    hostname=CLI_DEFAULTS['HOST_NAME'], port=CLI_DEFAULTS['PORT'], api=API_CONFIG['PREFIX']
+)
+
+NO_OF_REQUESTS = 100
+
```
sphuber, line 32: These tests look good, there is just another opportunity to leverage pytest's amazing fixture system. You could turn the threaded REST API server into a fixture and reuse it. Something like
```
@pytest_fixture
def restapi_server():
    from aiida.restapi.common.config import CLI_DEFAULTS, API_CONFIG

    server_url = 'http://{hostname}:{port}{api}'.format(
        hostname=CLI_DEFAULTS['HOST_NAME'], port=CLI_DEFAULTS['PORT'], api=API_CONFIG['PREFIX']
    )
    aiida_restapi = configure_api()
    server = make_server(
        host=CLI_DEFAULTS['HOST_NAME'],
        port=int(CLI_DEFAULTS['PORT']),
        app=aiida_restapi.app,
        threaded=True,
        processes=1,
        request_handler=None,
        passthrough_errors=False,
        ssl_context=None,
        fd=None
    )

    server_thread = Thread(target=server.serve_forever)

    try:
        server_thread.start()
        yield server_url
    except Exception as exc:  # pylint: disable=broad-except
        pytest.fail('Something went horribly wrong! {}'.format(repr(exc)))
    finally:
        server.shutdown()


def test_run_threaded_server(aiida_localhost, restapi_server):
    server_url = restapi_server

    for _ in range(NO_OF_REQUESTS):
        response = requests.get(server_url + '/computers/{}'.format(computer_id))

        assert response.status_code == 200

        try:
            response_json = response.json()
        except ValueError:
            pytest.fail('Could not turn response into JSON. Response: {}'.format(response.raw))
        else:
            assert 'data' in response_json
```

You can then reuse the same fixture in the second test

aiida/restapi/resources.py
CasperWA: It is to keep the current API (sort of) still there. It seemed one could pass the keyword argument `get_decorators` when adding a resource to the REST API (using `.add_resource()` in `aiida.restapi.api:AiidaApi`). To keep this functionality, and potentially add more decorators from there, I have renamed the keyword argument to simply `decorators`, since it doesn't conform to simply to HTTP `GET` request any more.
I wanted to align the two ways of passing decorators to the HTTP request methods, but unfortunately that's not possible. Either `method_decorators` is a list with callables, and they apply to _all_ HTTP request methods OR `method_decorators` is a dictionary/mapping, where the keys are the HTTP request method names and the values a list of callable decorators.

aiida/restapi/common/utils.py
CasperWA: I don't see the reason to make it a class static method in any case. However, I would _really_ like to move it some place else, if you have a good argument :)

aiida/restapi/common/utils.py
CasperWA: It wasn't picked up by either `pylint` nor the pre-commit `prospector` (which usually _always_ catches these things).

CasperWA: And `flask` is a third-party library as well ... right?

aiida/restapi/resources.py
CasperWA: I have tried to explain this in my commit message of 41ad616.

It seems these lines were "missed" to be removed when moving to the REST API v4 (AiiDA v1). It was also present for all the sub-classes, but was removed in the transition, and so I guess this acted as a "meta" class in a way for these sub-resources, but is now not relevant, since this information is retrieved in a different way.
However, I am not _completely_ sure and would like a confirmation from @waychal, @asle85 (or @ltalirz who originally wrote these lines some years ago).

CasperWA: > (...) Where is this being set for the `Node` base resource?

It is not - but it is also (apparently) never used.

tests/restapi/test_threaded_restapi.py
CasperWA: This was the incentive of making it a class originally ...
And yeah, there are definitely improvements to be done here, and this seems a good one.
For now, I simply want the tests to work with the CI. They work fine locally ...

CasperWA:
> This is likely related to the discussion [here](https://github.com/aiidateam/aiida-core/pull/3738#issuecomment-607687961).
> 
> In essence, I think the issue is rooted in interference between the pytest fixtures and the remaining AiiDATestCase implementation.
> In the PR I linked we decided to remove the change again since we wanted to release. I wrote
> 
> > Note, however, that we will likely run into this issue again soon, namely when people start to use pytest fixtures in the test suite more actively.

Well that's crap.

So what now? Should I undo my conversion to `pytest` or should I make and store a `Computer` "manually" or what is the solution, do you think?

aiida/restapi/common/utils.py
sphuber: yes, but the `w` comes after the `f` in the alphabet :) As I said, nitpick ...

CasperWA: 🙄 ...

aiida/restapi/resources.py
CasperWA: Hmm.. re-reading your comment here, it seems my answer is based on a misunderstanding. And yeah sure.
However, it also means that if a sub-class "forgets" to call `super().__init__()` it will not get the benefit of the `decorators` keyword argument.

greschd:
> So what now? Should I undo my conversion to `pytest` or should I make and store a `Computer` "manually" or what is the solution, do you think?

You could try using the `clear_database_before_test` fixture. My understanding is that some of the "old-style" tests are trashing the database (specifically here, the user). The "good" solution here would be finding and fixing those tests, but the workaround is to reset the database before running "new-style" tests.

aiida/restapi/resources.py
```diff
@@ -106,11 +107,15 @@ def __init__(self, **kwargs):
         utils_conf_keys = ('PREFIX', 'PERPAGE_DEFAULT', 'LIMIT_DEFAULT')
         self.utils_confs = {k: kwargs[k] for k in utils_conf_keys if k in kwargs}
         self.utils = Utils(**self.utils_confs)
-        self.method_decorators = {'get': kwargs.get('get_decorators', [])}
+
+        # By making `method_decorators` a list, it will run the list of decorators for any HTTP request
+        # TODO: Consider using/allowing `decorators` to be a dict or similar instead ({method: [decorators]})
+        decorators = kwargs.get('decorators', [])
```
CasperWA, line 113: Change back: If `get_decorators` is set overwrite `method_decorators`

CasperWA:
> > So what now?
> 
> You could try using the `clear_database_before_test` fixture. My understanding is that some of the "old-style" tests are trashing the database (specifically here, the user). The "good" solution here would be finding and fixing those tests, but the workaround is to reset the database before running "new-style" tests.

I think it seemed to fix it - only one test is failing now; my newest test that checks whether the fix is needed AND solves the issue :) Threading is hard - especially across several Python versions.

tests/restapi/test_threaded_restapi.py
```diff
@@ -0,0 +1,147 @@
...
+# This is the default location where Flask apps start when using the built in web server.
+SERVER_URL = 'http://{hostname}:{port}{api}'.format(
+    hostname=CLI_DEFAULTS['HOST_NAME'], port=CLI_DEFAULTS['PORT'], api=API_CONFIG['PREFIX']
+)
+
+NO_OF_REQUESTS = 1000
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_run_threaded_server(aiida_localhost):
```
sphuber, line 36: Not sure if there is a particular reason why you opted for this form, but to "use" a fixture, you just add it as an argument to the test signature. I know ... pytest magic :eyes: 
So these two lines could simply become
```
def test_run_threaded_server(clear_database_before_test, aiida_localhost):
```

greschd: Actually kinda like this, at least `pylint` stops complaining about unused function arguments 😄 

But yeah, I think `usefixtures` is meant mostly for defining a fixture use for an entire module or class.

CasperWA: It was to avoid adding a `pylint` disable as well ...

CasperWA:
Hmm.. it seems one Py3.8 worked (django), while the other failed with the error message from the previous CI run as well (sqlalchemy), while the Py3.5 are stuck in an infinite `while` loop waiting on the `server_thread` to kill itself, which it should, since the server _should_ be shut down ... But I guess threading works differently in Py3.5....?

CasperWA:
Hmm, I'm out of ideas for this last test. I have looked into whether it may be a GitHub Actions thing (and it may still be), but I can't seem to pinpoint exactly where that issue should be, since each job should get its own separate "virtual environment", and even then I think it should still work, right? Unless each job is a different process, since threading and multiprocessing is not really compatible, right, or?

In any case, I'm considering removing the last test and simply keeping the one testing that the fix works - and simply set the number of times a request is done to a very, very high number - what do you think @greschd and @sphuber?

greschd:
I don't really have the background to know what might be going wrong here.. just a quick comment: it seems these two tests run for almost a minute combined (even when they pass). That seems awfully slow, is it the same locally?

Also, is it always the same run that succeeds, or is it random? If it's consistently the same one, I'd guess it's at least not interaction between the runs.

greschd:
Just a random guess: Is the port `5000` used in the test available, or might that be occupied by something else?

CasperWA:
> Just a random guess: Is the port `5000` used in the test available, or might that be occupied by something else?

Port 5000 is indeed used in the test, since it's the default REST API port to be used.

CasperWA:
> I don't really have the background to know what might be going wrong here.. just a quick comment: it seems these two tests run for almost a minute combined (even when they pass). That seems awfully slow, is it the same locally?

Yeah, one test will at minimum take 30 s, since that is the timeout period set by AiiDA for the SQLAlchemy engine's QueuePool (or rather, it's the default and AiiDA doesn't change it).

> Also, is it always the same job that succeeds, or is it random? If it's consistently the same one, I'd guess it's at least not interaction between the jobs.

I'm guessing it's always the same job, _however_ for the latest run, it's probably different, since I put in a `while` loop, which I think the py3.5 tests got stuck in (no matter which of the two tests it's in), but I'm not sure.

greschd:
> Yeah, one test will at minimum take 30 s, since that is the timeout period set by AiiDA for the SQLAlchemy engine's QueuePool (or rather, it's the default and AiiDA doesn't change it).

Yikes, is there any way to change this for the test?

> But also, it may indeed be the port issue. I could make it a random number within a valid range instead?

Maybe random isn't the best choice, you could use something like this to check if it's available https://codereview.stackexchange.com/questions/116450/find-available-ports-on-localhost

Regarding the threading, we're talking about threads on _different_ processes (from the different jobs) interacting, right? In that case, the only interaction I can think of is competing for system resources (e.g., any one thread/process might be suspended by the kernel while a different one runs) - but that shouldn't be different from running in a resource-constrained scenario in general. 

Threads of the _same_ Python process are more closely linked - they share memory (i.e., Python objects are accessible from multiple threads) and, for that reason, also the global interpreter lock (GIL) can only be held by one thread at a time.

CasperWA:
> > Yeah, one test will at minimum take 30 s, since that is the timeout period set by AiiDA for the SQLAlchemy engine's QueuePool (or rather, it's the default and AiiDA doesn't change it).
> 
> Yikes, is there any way to change this for the test? If there's no way around it it's still better than _not_ having the test, but we should keep the tests short whenever possible.

Sure - but there is no way around this as far as I know.

> > But also, it may indeed be the port issue. I could make it a random number within a valid range instead?
> 
> Maybe random isn't the best choice, you could use something like this to check if it's available https://codereview.stackexchange.com/questions/116450/find-available-ports-on-localhost

Right. Thinking more about this, it shouldn't actually be a problem, and it should complain when starting the servers, right? Saying that the port is already in use...?

> Regarding the threading, we're talking about threads on _different_ processes (from the different jobs) interacting, right? In that case, the only interaction I can think of is competing for system resources (e.g., any one thread/process might be suspended by the kernel while a different one runs) - but that shouldn't be different from running in a resource-constrained scenario in general.
> 
> Threads of the _same_ Python process are more closely linked - they share memory (i.e., Python objects are accessible from multiple threads) and, for that reason, also the global interpreter lock (GIL) can only be held by one thread at a time.

Hmm, I wonder if they block each other then, and time out in a different way?

greschd:
Given that both client and server run on the same Python process, I would guess it's _possible_ to create a deadlock situation - not sure if it's _likely_ here.

Can you maybe print out the whole `captured.err` traceback here? https://github.com/aiidateam/aiida-core/runs/626243863#step:10:432

sphuber:
Is the only call `server.shutdown()` in the test cleanup sufficient to properly shut down the server and close the thread? Should we manually terminate the thread as well, or is the `shutdown` method supposed to take care of everything.

Also, regarding the runtime, we could change the default settings of the SqlAlchemy engine to throw the error more quickly.These settings can be set in the [`create_engine`](https://docs.sqlalchemy.org/en/13/core/engines.html?highlight=create_engine#sqlalchemy.create_engine) factory. We call this in `aiida.backends.utils.create_sqlalchemy_engine`. If we update this method to take `**kwargs` that it forwards to `create_engine` than we can update `aiida.backends.sqlalchemy.get_scoped_session` method to _also_ take `**kwargs` which it in turn forwards to `create_sqlalchemy_engine`. In your test, you can then call `reset_session` and then `get_session` passing the correct keywords, i.e. `pool_timeout=1, pool_size=1, max_overflow=0`. This way you should only have to run just a few queries and it will timeout almost instantly

CasperWA:
> Is the only call `server.shutdown()` in the test cleanup sufficient to properly shut down the server and close the thread? Should we manually terminate the thread as well, or is the `shutdown` method supposed to take care of everything.

As far as I understand, you shouldn't manually shutdown a thread (or join it), but rather end whatever it runs, which is what `server.shutdown()` does.

> Also, regarding the runtime, we could change the default settings of the SqlAlchemy engine to throw the error more quickly.These settings can be set in the [`create_engine`](https://docs.sqlalchemy.org/en/13/core/engines.html?highlight=create_engine#sqlalchemy.create_engine) factory. We call this in `aiida.backends.utils.create_sqlalchemy_engine`. If we update this method to take `**kwargs` that it forwards to `create_engine` than we can update `aiida.backends.sqlalchemy.get_scoped_session` method to _also_ take `**kwargs` which it in turn forwards to `create_sqlalchemy_engine`. In your test, you can then call `reset_session` and then `get_session` passing the correct keywords, i.e. `pool_timeout=1, pool_size=1, max_overflow=0`. This way you should only have to run just a few queries and it will timeout almost instantly

But isn't this a bit too hacky?

CasperWA:
> Given that both client and server run on the same Python process, I would guess it's _possible_ to create a deadlock situation - not sure if it's _likely_ here.
> 
> Can you maybe print out the whole `captured.err` traceback here? https://github.com/aiidateam/aiida-core/runs/626243863#step:10:432

I'm afraid this is all it catches, but I will print it out, and maybe also invoke the `traceback` lib to print, which will be visible only when/if it fails.

sphuber:
> But isn't this a bit too hacky?

Not at all. There is a reason that SqlAlchemy provides a way to tweak these parameters. We just haven't exposed the same interface in AiiDA because so far it had not been necessary yet. It is fine to add this for the purposes of the tests. It would of course also open it up to users changing, but the `get_scoped_session` is not part of the public API so this should serve as enough of a warning.

aiida/restapi/resources.py
CasperWA: I expect by now, since there haven't been any outcry from either the tests (of this) or @waychal @asle85 or @ltalirz that it's fine to remove.

ltalirz: I'm not an expert here - I trust Casper's assessment :-)

CasperWA:
All right!
[It seems](https://github.com/aiidateam/aiida-core/actions/runs/92227466) that this is now (finally) working "reliably" for Py3.8 (only run once).
However, for Py3.5, shutting down the server, doesn't seem to close the `Thread`.

Also, I am unsure whether it was the switch to check the `capfd` instead of `capsys` or have a small `sleep` period before shutting down the server that made sure the correct output could be found in the traceback thrown for the test that doesn't apply the fix.
I want to test this by switching back to using `capsys`, run the tests again, and then I guess I'll have to setup a local Py3.5 environment to adapt the tests to work with older py3 version.

CasperWA:
Update:
I have set up a local Py3.5 environment and could reproduce the never-ending tests also seen here in the GH Actions CI.
Doing a bit of research I found out that the socket was stuck in an infinite loop trying to read.
I then found [this](https://github.com/psf/requests/issues/3353) and simply setting a `timeout` together with the `get` requests, seemed to work and get the socket out of the loop, and catching the subsequent exception (and `pass`ing on it) made the tests run locally.

Now to see if this is also the case with GH Actions ...

CasperWA:
https://github.com/aiidateam/aiida-core/runs/636184358

🤩 ❤️ 😍 🎉 🥳

CasperWA:
> https://github.com/aiidateam/aiida-core/runs/636184358
> 
> 🤩 ❤️ 😍 🎉 🥳

However, it does seem that the whole traceback sometimes _still_ doesn't get registered and captured in `capfd`: https://github.com/CasperWA/aiida_core/runs/636183960

aiida/backends/djsite/__init__.py
```diff
@@ -43,7 +43,11 @@ def get_scoped_session():
     Django implementation of the `QueryBuilder` should keep its own SqlAlchemy engine and scoped session factory
     instances that are used to provide the query builder with a session.
 
-    :param profile: :class:`aiida.manage.configuration.profile.Profile` for which to configure the engine.
+    Pass on kwargs to :py:func:`aiida.backends.utils.create_sqlalchemy_engine`, opening the possibility to change
```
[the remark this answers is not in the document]
sphuber: Please apply this similar change in all other docstring as well

CasperWA:
@sphuber Wait a second with merging - I just realized I should use our "new" `get_decorators` argument when initiating the mock REST API class in one of my tests, instead of also creating a mock resource.
This way the `get_decorators` argument will be tested also.

CasperWA:
@sphuber All right, it seems to be all fixed with docs building etc. now. Also, I have updated the test and it seems to work.
It was a good thing as well, I found out the _actual_ way to pass this kwarg to the `Resource`.
Aaaand out-of-date ....

sphuber:
Thanks a lot @CasperWA !

PR #3976, opened 2020-04-27
Add codecov.yml codecov configuration file

### LINKED ISSUE #3975: Update codecov tolerance and notification time
> Unfortunately codecov is complaining, no idea why.

Is there a way that we can put a bigger tolerance to prevent these false positives? It is a bit annoying to have build officially fail just because a false positive in the coverage. Not only do PRs look like they are failing, even though everything is fine, but even when merging, it could mark the `develop` or `master` branch as failing, which will even influence the `build status` badge on the README in a negative way. Having a failing build status badge because of this is not the nicest. I have no experience with these coverage tools, so thought I'd ask before starting to investigate. @CasperWA do you have an idea?

_Originally posted by @sphuber in https://github.com/aiidateam/aiida-core/pull/3957#issuecomment-619902790_

The coverage reporting should be a bit more friendly.

CasperWA:
Closes #3975 

A Codecov configuration file is added in the package root (`codecov.yml`).
In it, the following is specified:
- Upload _only_ if CI passes (I guess this is only true for the CI jobs that run `pytest` in Python 3.5).
- Add PR comment/status report when two builds have been received, i.e., when the Python 3.5 `pytest` jobs have run for both backends.
  This is to avoid confusion when only one of the CI jobs have finished and the codecov report is suddenly reporting missing lines.
- Use default range for coloration (70-100 %), where coverage will be shaded from red to green in this range, and be solid red and green below and above this range (not relevant), respectively.
- Round percentages up (instead of down).
- Set threshold for failing `codecov/project` CI status to 0.02 % (instead of the default 0 %).
  ~It seems this can not be set as a float (see [details](https://docs.codecov.io/docs/commit-status#threshold)).~

### DIFF
--- a/.github/workflows/tests.sh
+++ b/.github/workflows/tests.sh
@@ -9,9 +9,13 @@ export PYTHONPATH="${PYTHONPATH}:${GITHUB_WORKSPACE}/.ci"
 # - pytest-cov configuration taken from top-level .coveragerc
 # - coverage is reported as XML and in terminal,
 #   including the numbers/ranges of lines which are not covered
-# - coverage results of multiple tests are collected
+# - coverage results of multiple tests are collected (this is always the case with Codecov)
 # - coverage is reported on files in aiida/
+export PYTEST_ADDOPTS="${PYTEST_ADDOPTS} --durations=50"
+export PYTEST_ADDOPTS="${PYTEST_ADDOPTS} --cov-config=${GITHUB_WORKSPACE}/.coveragerc"
+export PYTEST_ADDOPTS="${PYTEST_ADDOPTS} --cov-report xml"
 export PYTEST_ADDOPTS="${PYTEST_ADDOPTS} --cov-append"
+export PYTEST_ADDOPTS="${PYTEST_ADDOPTS} --cov=aiida"
 
 # daemon tests
 verdi daemon start 4
--- a/.github/workflows/tests.sh
+++ b/.github/workflows/tests.sh
@@ -9,7 +9,7 @@ export PYTHONPATH="${PYTHONPATH}:${GITHUB_WORKSPACE}/.ci"
 # - pytest-cov configuration taken from top-level .coveragerc
 # - coverage is reported as XML and in terminal,
 #   including the numbers/ranges of lines which are not covered
-# - coverage results of multiple tests are collected (this is always the case with Codecov)
+# - coverage results of multiple tests (within a single GH Actions CI job) are collected
 # - coverage is reported on files in aiida/
 export PYTEST_ADDOPTS="${PYTEST_ADDOPTS} --durations=50"
 export PYTEST_ADDOPTS="${PYTEST_ADDOPTS} --cov-config=${GITHUB_WORKSPACE}/.coveragerc"
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,4 @@
 [pytest]
-addopts = --durations=50 --cov-config=.coveragerc --cov-report xml --cov=aiida
 testpaths = tests
 filterwarnings =
     ignore::DeprecationWarning:babel:
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,5 @@
 [pytest]
-addopts= --durations=50 --cov-config=.coveragerc --cov-report xml --cov=aiida
+addopts = --durations=50 --cov-config=.coveragerc --cov-report xml --cov=aiida
 testpaths = tests
 filterwarnings =
     ignore::DeprecationWarning:babel:
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,10 @@
+codecov:
+  notify:
+    after_n_builds: 2
+    wait_for_ci: yes
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: up
+  range: "60...90"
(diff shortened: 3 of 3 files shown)


pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
-addopts= --durations=50 --cov-config=.coveragerc --cov-report xml --cov=aiida
+addopts = --durations=50 --cov-config=.coveragerc --cov-report xml --cov=aiida
```
sphuber, line 2: While we are here (I know this wasn't added by this PR, but it was added recently) can we have these options just enabled on github? These are now also run everytime I run the tests locally and it takes a lot of extra time and outputs a lot of noise. This really disturbs me when developing locally, so I have to remove these every time. @ltalirz I think you added these. Can we just move these to the CI configuration?

CasperWA: I agree that this would be better to _not_ have locally. I can look into changing this back.

ltalirz: >  it takes a lot of extra time

I was not aware of that... is it the coverage check?
In that case I agree that the two `--cov` options can probably be moved to github...

> outputs a lot of noise

What it should be outputting in addition is test timings.
So far, nobody has been paying attention to them (we have "unit tests" that run several seconds), and so I think it is a good idea to show them when you run tests locally.

CasperWA: > > outputs a lot of noise
> 
> What it should be outputting in addition is test timings.
> So far, nobody has been paying attention to them (we have "unit tests" that run several seconds), and so I think it is a good idea to show them when you run tests locally.

It would be better to be able to toggle this locally. It _is_ a lot of unneccesary noise if you're just trying to see why a test is failing. Or even better would be to filter this and only show timings above a certain threshold.

sphuber: We don't need to see the timings all the time though. Of course when there are new tests, it would be good to see that they are not super slow. But that is why we have the timings run on the CI so they can be checked before merging.

sphuber: > Or even better would be to filter this and only show timings above a certain threshold.

that would be ok if there currently aren't any tests that surpass it, or at the very least are then also addressed immediately to reduce their time, otherwise we would still be spammed every time with the same information

.github/workflows/tests.sh
```diff
@@ -9,9 +9,13 @@ export PYTHONPATH="${PYTHONPATH}:${GITHUB_WORKSPACE}/.ci"
 # - pytest-cov configuration taken from top-level .coveragerc
 # - coverage is reported as XML and in terminal,
 #   including the numbers/ranges of lines which are not covered
-# - coverage results of multiple tests are collected
+# - coverage results of multiple tests are collected (this is always the case with Codecov)
```
greschd, line 12: I think the comment here refers to coverage results within a _single_ workflow step (when `pytest` is run more than once). That's what ``--cov-append`` does, otherwise the coverage file would be overwritten (AFAIK, _before_ it's uploaded to codecov).

CasperWA: For sure.
I also read, however, that one could simply upload it (using `coverage` CLI) to Codecov, as soon as you have a "new" report, and Codecov would _always_ merge these reports on their side.
However, I much prefer adding this here and not continuously upload to Codecov ...
But maybe I should rewrite my addition to the comment here to represent the intention better.

pytest.ini
ltalirz: > Or even better would be to filter this and only show timings above a certain threshold.

This is already the case. It shows the 50 slowest tests.

> It is a lot of unneccesary noise if you're just trying to see why a test is failing.

Ok, perhaps I'm using local tests differently - I usually just run a handful of them.

> that is why we have the timings run on the CI so they can be checked

They *can* be, but I guess we agree that what matter is whether they actually *are*.

Perhaps that is better done by sensibilizing developers on this topic, though. I didn't want the timings to be annoying, of course.

ltalirz: > Or even better would be to filter this and only show timings above a certain threshold.

Ok, I guess you meant a cutoff in ms, which would be even better.
Unfortunately, I don't think this is currently supported by pytest https://docs.pytest.org/en/latest/usage.html#profiling-test-execution-duration

CasperWA: 50 is a bit of a high number as well, I think. At least for the local test runs. Just 5 would probably be better. But if it's a contested issue, I think we should decide on this within the developer team and not deal with it in this PR (or deal with it after a quick decision)?

CasperWA:
Testing if threshold can be set with float values instead of integers.

- When setting the threshold to 0.05 % and skipping all REST API tests, leaving the coverage with a diff of -2.97 % (or -2.96 % for each backend) the codecov/project status fails (as expected) ✔️ 
- When setting the threshold to 3.55 % and still skipping all REST API tests, the codecov/project status should succeed ... (to-be-updated)

pytest.ini
ltalirz: Feel free to adapt to your preferences. I agree it is very important not to annoy people :-)
You could set it to 5 for local testing and 50 on github actions; this will already have the effect of people getting used to looking at timings.

Feel free to proceed without another review from my side. Cheers!

sphuber:
> I will set it to 0.02 % or should it be 0.01 % ?

I think 0.02 for now is fine. We can adapt it later if we find it is too large. So I guess this is ready to got correct?

CasperWA:
Note, I have also "reset" the color range to the default 70-100 %.

PR #3980, opened 2020-04-30
Docs: flesh out skeleton of "How-to Guides"

sphuber:
This section is subdivided in 6 main categories:

 * How to run external codes
 * How to create and run workflows
 * How to import and manage data
 * How to query and visualize data
 * How to manage your installation
 * How to write a plugin package

The naming has been purposefully kept free of AiiDA specific jargon as
these guides need to be quickly findable by relatively new users that
after having run through a tutorial want to quickly know how to
accomplish certain specific goals.

Each section has its own document that will be presented as its own
page with a number of subsections that present guides for specific
goals. The sections and subsections are all created with the
corresponding link labels for cross-referencing.


docs/source/howto/data_manage.rst
```diff
@@ -0,0 +1,41 @@
...
+
+.. _how-to:data-manage:load-external:
+
+How to load external data
+=========================
+
+
+.. _how-to:data-manage:plugins:
+
+How to create custom data types
```
giovannipizzi, line 16: I would put this at the end of the section, as this is advanced (for "plugin developers"?).

docs/source/howto/plugins.rst
```diff
@@ -0,0 +1,41 @@
...
+===============================
+
+
+.. _how-to:plugins:publish:
+
+How to publish a plugin package
+===============================
+
+
+.. _how-to:plugins:scheduler:
```
giovannipizzi, line 32: Maybe you want to put these in a subsection in specific types of plugins? Otherwise it's a bit weird that we isolate only these two out of all the other types of plugins we have?

docs/source/howto/workflows.rst
```diff
@@ -0,0 +1,29 @@
...
+.. _how-to:workflows:run:
+
+How to run an existing workflow
+===============================
+
+
+.. _how-to:workflows:write-simple:
+
+How to write simple workflows
+=============================
```
giovannipizzi, line 17: Should we have an 'how-to' section on how to choose between a WorkChain vs workfunction? (at least implicitly, maybe inside 'how to write simple workflows', that shows both syntaxes and explains - by example - when you pick one or another?)

Or it's really for "topics"? I still think there should be a pointer here, avoiding the assumption that there are only WorkChains (sorry to write this here, it's just not to forget)

sphuber: I agree that there should be a pointer. My idea would be to at the end of the "simple workflow" how-to, quickly explain the limitation of work functions being blocking and non-restartable and that for that reason you might want to use "more complex" workflows which are workchains and then link to that how-to (besides all the other advantages that the work chain gives). I would also add a link to a section in topics that goes into more depth about the difference between work function and work chain, as well as their respective advantages/disadvantages.

docs/source/howto/plugins.rst
sphuber: yeah it's a good point, it is a bit awkward, but I am not sure where else to put it. I thought it was not a good idea to have a category "How to write a plugin" with individual sections for each possible plugin category, because that is not in the spirit of the how-to. The goal is not to write a plugin, the goal is to do something, and then we tell them in order to do that, what plugin they need to write. So the `CalcJob` manual will be in "How to run external codes".

Thinking about it now, since transport and scheduler really only applies to computers, which in itself only really applies (in most cases) to calculation jobs, maybe I can put these two sections at the end of the "How to run external codes" category? What do you think?

docs/source/howto/data_manage.rst
sphuber: Fine to place it towards the end. Just maybe in the future, once people from other domains start to use AiIDA, the need to create custom data types becomes more common. Anyway, there is a link at the end of the section of using existing types, so it is fine to move to the end for now

docs/source/howto/plugins.rst
sphuber: These have been added to the end of "How to run external codes" since that is where they are really needed logically

sphuber:
@giovannipizzi just for your information, I discussed with @csadorf and @ltalirz and they had quite some comments and we changed quite a bit.

PR #3985, opened 2020-05-01
Fix autocompletion for LinkManager and AttributeManager

### LINKED ISSUE #3984: Auto-completion of managers broken in verdi shell (calc.inputs, calc.outputs, ...)
Reported by Hongyi Zhao in the mailing list.

The reason is that internally, in the managers, after their rewrite, exceptions are not caught properly, so `getattr(calc.inputs, 'value')` does not return an `AttributeError` if `value` does not exist, and as a consequence `getattr(calc.inputs, 'value', None)` raises instead of returning None.

giovannipizzi:
In the managers, exceptions were not caught properly.
For instance, `getattr(calc.inputs, 'value')` was not return
an AttributeError if `'value'` does not exist, and as a
consequence `getattr(calc.inputs, 'value', None)` was raising
instead of returning `None`.

Similarly, I fixed `calc.inputs['value']` to raise a KeyError
for not-existing `value`.

The AttributeManager was not having Tab completion for a different
reason, `__dir__` was returning tuples of dict items, rather than
just the keys.

Now these are fixed and tests are added (I cannot really test
the TAB-completion functionality, but at least I test that the
values of `dir()` and the exceptions raised are the correct ones).

Fixes #3984

### DIFF
--- a/aiida/orm/utils/managers.py
+++ b/aiida/orm/utils/managers.py
@@ -14,6 +14,7 @@ to access members of other classes via TAB-completable attributes
 """
 
 from aiida.common.links import LinkType
+from aiida.common.exceptions import NotExistent, NotExistentAttributeError, NotExistentKeyError
 
 __all__ = ('NodeLinksManager', 'AttributeManager')
 
@@ -80,8 +81,15 @@ class NodeLinksManager:  # pylint: disable=too-few-public-methods
         """
         try:
             return self._get_node_by_link_label(label=name)
-        except KeyError:
-            raise AttributeError("Node '{}' does not have an input with link '{}'".format(self._node.pk, name))
+        except NotExistent:
+            # Note: in order for TAB-completion to work, we need to raise an
+            # exception that also inherits from AttributeError, so that
+            # `getattr(node.inputs, 'some_label', some_default)` returns
+            # `some_default`. Otherwise, the exception is not caught by
+            # `getattr` and is just propagated, instead of returning the default.
+            raise NotExistentAttributeError(
+                "Node '{}' does not have an input with link '{}'".format(self._node.pk, name)
+            )
 
     def __getitem__(self, name):
         """
@@ -91,8 +99,12 @@ class NodeLinksManager:  # pylint: disable=too-few-public-methods
         """
         try:
             return self._get_node_by_link_label(label=name)
-        except KeyError:
-            raise KeyError("Node '{}' does not have an input with link '{}'".format(self._node.pk, name))
+        except NotExistent:
+            # Note: in order for this class to behave as a dictionary, we raise
+            # an exception that also inherits from KeyError - in this way, users
+            # can use the standard construct `try/except KeyError` and this will
+            # behave like a standard dictionary.
+            raise NotExistentKeyError("Node '{}' does not have an input with link '{}'".format(self._node.pk, name))
 
     def __str__(self):
         """Return a string representation of the manager"""
@@ -125,7 +137,7 @@ class AttributeManager:  # pylint: disable=too-few-public-methods
         """
         Allow to list the keys of the dictionary
         """
-        return sorted(self._node.attributes_items())
+        return sorted(self._node.attributes_keys())
 
     def __iter__(self):
         """
--- a/tests/orm/utils/test_managers.py
+++ b/tests/orm/utils/test_managers.py
@@ -1,0 +1,137 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Tests for the various node managers (.inputs, .outputs, .dict, ...)."""
+# pylint: disable=unused-argument
+
+import pytest
+
+from aiida import orm
+from aiida.common.exceptions import NotExistent, NotExistentAttributeError, NotExistentKeyError
+from aiida.common import LinkType
+
+
+def test_dot_dict_manager(clear_database_before_test):
+    """Verify that the Dict.dict manager behaves as intended."""
+    dict_content = {'a': True, 'b': 1, 'c': 'Some string'}
+    dict_node = orm.Dict(dict=dict_content)
+
+    # Check that dir() return all keys and nothing else, important
+    # for tab competion
+    assert len(dir(dict_node.dict)) == len(dict_content)
+    assert set(dir(dict_node.dict)) == set(dict_content)
+    # Check that it works also as an iterator
+    assert len(list(dict_node.dict)) == len(dict_content)
+    assert set(dict_node.dict) == set(dict_content)
+
+    for key, val in dict_content.items():
+        # dict_node.dict.a == True, ...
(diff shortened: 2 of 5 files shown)


giovannipizzi:
I'm realising now that plugin developers might have made assumptions on the exception returned.
However, I'm convinced that it's not a good idea not to raise something different than a AttributeError because it's something unexpected and e.g. tab completion will not work.

I don't know if this is something we accept and we alert all plugin developers that in 1.3 this behaviour is changing, or there is something else we can do that is more backward-compatible.
But I would consider this a bug, and not wait for a 2.0 to be out to re-enable TAB-completion.

One option would be to create for now a combined class NotExistentAttributeError that inherits both from AttributeError and NotExistent? Would this be working and tenable also in the future when releasing a 2.0?

Pinging @sphuber @ltalirz for feedback

greschd:
I totally agree that raising `NotExistent` on `__getattr__` and `__getitem__` is not great -- for example, the `getattr(obj, 'key', default)` idiom also doesn't work, because it relies on `AttributeError` being raised when `'key'` is not present.

However, I'm also convinced that we need a backwards-compatible solution for this. The `NotExistent` exception is caught in many places. Across the plugins that I happen to have cloned on my machine, there are 23 instances of `except NotExistent`:
```
$ grep -inR 'except NotExistent' . | grep -v core | wc -l
23
```

I'd favor adding a `NotExistentAttributeError` and `NotExistentKeyError`, but it would also be nice to have a way to eventually deprecate this. Not sure _how_ we could know if the user code catches it as `AttributeError` or `NotExistent`, though.

giovannipizzi:
>  Not sure how we could know if the user code catches it as AttributeError or NotExistent, though.

Indeed, this is my main concern... It would be great if there was a clever way to check if they are catching NotExistent to warn them.
But maybe we just make it backwards-compatible, and inform all plugin developers, and adapt the documentation...

greschd:
Yeah, it's a difficult one.. if we were to phase out the `NotExistent` exception completely, we could warn on import. But I'm guessing there still is a valid reason for this exception, right?

giovannipizzi:
Yeah. And actually, thinking better to it, indeed, we don't really need to deprecate the old behaviour.
I think we can keep all three catching behaviours acceptable: `AttributeError`, `NotExistent` and `NotExistentAttributeError`.

I just pushed the backwards-compatible fix to the code, and I think it's now OK (I tested and tab completion works) and shouldn't break backwards-compatibility

sphuber:
> Yeah, it's a difficult one.. if we were to phase out the `NotExistent` exception completely, we could warn on import. But I'm guessing there still is a valid use case for this exception, right?

Absolutely. It may have been misused in the managers, but I don't think we should get rid off it entirely. It is used in various methods where it is actually just fine. For example in the `QueryBuilder.one` will return `NotExistent`, which makes perfect sense and more so than an `AttributeError` or `KeyError`.

Couldn't we also make `NotExistent` a subclass of `AttributeError`. That way, we can change the managers to throw `AttributeError` and existing code will continue to work. But not sure if this has undesirable effects elsewhere

giovannipizzi:
> Absolutely. It may have been misused in the managers, but I don't think we should get rid off it entirely. It is used in various methods where it is actually just fine. For example in the QueryBuilder.one will return NotExistent, which makes perfect sense and more so than an AttributeError or KeyError.

Sure, the plan was not to drop the `NotExistent` exception.

> Couldn't we also make NotExistent a subclass of AttributeError. That way, we can change the managers to throw AttributeError and existing code will continue to work. But not sure if this has undesirable effects elsewhere

I think the current approach is better. It's backwards-compatible, and we don't risk that people catch a AttributeError and end up catching a NotExistent, instead. Also, we would need to subclass NotExistent also both from AttributeError and KeyError, that I don't think it's a great idea. For me the current code in the PR is a good final solution.

greschd:
I think the problem with subclassing `NotExistent` from `AttributeError` is that it would _also_ need to subclass from `KeyError` - because it can be the cause of either one of them.

I'm fine with the current solution, but would move the `NotExistentAttributeError` and `NotExistentKeyError` to be internal - i.e., remove it from the `__all__` and maybe even add a leading underscore. If people start catching these explicitly, we will be in even more of a backwards-compatibility mess.
I think, _eventually_, we might get to a point where we can drop these and just raise the bare `AttributeError` and `KeyError`.

giovannipizzi:
Ok, let's wait for another opinion before I change the code.
I was thinking that in the end *all* the exceptions are OK, even to be kept as public: they are specific extensions of very well defined exceptions (python's or AiiDA's) and, now that I think to I,  I think it's ok if people catch any of these. 

It's a bit like now in Python3 you have a `FileNotFoundError` that actually is a subclass both of `IOError` and `OSError`, while in python2 you would e.g. get `IOError` if calling `open` on a non-existing file.

And actually it's ok if you catch the more specific exception (if you want to). No?

tests/orm/utils/test_managers.py
```diff
@@ -0,0 +1,138 @@
...
+"""Tests for the various node managers (.inputs, .outputs, .dict, ...)."""
+import pytest
+
+from aiida.backends.testbase import AiidaTestCase
+from aiida import orm
+from aiida.common import exceptions as exc
+from aiida.common import LinkType
+
+
+class TestAttributeManager(AiidaTestCase):
```
greschd, line 19: Since (I believe) we want to slowly phase out the `AiidaTestCase`, could you try if this "just works" as a pure pytest test? Only if you feel like it, of course.

greschd:
Hmm, I think you're right - but it should definitely be carefully considered before we commit to this. As we've seen, exception types are kind of hard to change in a backwards-compatible way.

tests/orm/utils/test_managers.py
giovannipizzi: Done. @ramirezfranciscof for me this PR is ready for a final review

aiida/orm/utils/managers.py
```diff
@@ -14,6 +14,7 @@
 """
 
 from aiida.common.links import LinkType
+from aiida.common import exceptions as exc
```
sphuber, line 17: If you don't mind, please remove the alias. Prefer the long version and we don't have the abbreviation `exc` in the code base

aiida/orm/utils/managers.py
```diff
@@ -80,8 +81,10 @@ def __getattr__(self, name):
         """
         try:
             return self._get_node_by_link_label(label=name)
-        except KeyError:
-            raise AttributeError("Node '{}' does not have an input with link '{}'".format(self._node.pk, name))
+        except exc.NotExistent:
+            raise exc.NotExistentAttributeError(
```
sphuber, line 85: Was the idea that we are going to deprecate this at some point? I.e. for AiIDA 2.0 this will just raise an `AttributeError`? Not sure how to do that, because I don't think we can inspect when people catch a `NotExistent` exception. But if we don't we will always have to keep this composed exception just for this case. Same goes for the `NotExistentKeyError`.

If we have to keep it, can we add a quick comment here, where it is being raised why we need this special exception?

aiida/orm/utils/managers.py
giovannipizzi: Sure, for some reason I was sure we were using this pattern, not sure why...

aiida/orm/utils/managers.py
giovannipizzi: I think we don't need to deprecate, and that using this compound exception is safe to keep also in future versions. I have added more comments to explain the reason for this. Now I also rebased, and squashed all in 1 commit, so it's ready to merge (unless I broke tests in the meantime ;-) ).

PR #4025, opened 2020-05-01
Docs: scaffolding for the "Tutorials" section

mbercx:
Initial scaffolding for the tutorial section.


docs/source/tutorial/short.rst
```diff
@@ -0,0 +1,9 @@
+.. _tutorial:short:
```
csadorf, line 1: I'm wondering a little bit about the name `short` here, since it will appear in the url. Maybe `index.rst` is better?

PR #4031, opened 2020-05-01
Remove the top-level plugins/ section from scaffolding.

csadorf:


csadorf:
Will be removed as part of #4013 .

PR #4044, opened 2020-05-02
Docs: add the "Topics - Command line interface" section

### LINKED ISSUE #4015: Docs: Topic command line interface

sphuber:
Fixes #4015 

This content already existed and has been largely moved as is. Just the
style was adapted to match that of the new documentation, such as link
labels, indentation and header styles.


docs/source/topics/cli.rst
```diff
@@ -4,7 +4,160 @@
...
+This section explains the basic concepts that apply to all ``verdi`` commands.
 
 
-.. _#4015: https://github.com/aiidateam/aiida-core/issues/4015
\ No newline at end of file
+.. _topics:cli:parameters:
+
+Parameters
+==========
+Parameters to ``verdi`` commands come in two flavors:
```
csadorf, line 15: This particular description always struck me as a bit odd. Maybe we should at least acknowledge that this is UNIX standard?
```suggestion
Like many other command line utilities, ``verdi`` accepts both regular and optional arguments:
```

sphuber: It is just a collective term to refer to "arguments" and "options" which are POSIX naming conventions. Of course one can also see options as a special case of command line program arguments. I am not sure if your suggestion would help, because it is not the "optionality" that makes an argument an option. It is the way it is specified. An argument is positional and an option is defined through a flag. An argument can be optional just as well.

Before I merge this: what is it that you find odd about the description? The name parameters? Or the fact that it is not clear that this is a POSIX standard and not just something that AiiDA invented?

sphuber:
Thanks for the review @csadorf  I am putting this on hold because I think it is better to first merge the last remaining scaffolding PR. Also I would like to try to improve the description of parameters if you think it is odd.

docs/source/topics/cli.rst
csadorf: > Before I merge this: what is it that you find odd about the description? The name parameters? Or the fact that it is not clear that this is a POSIX standard and not just something that AiiDA invented?

The fact that it's not made clear that this is POSIX standard and not something that AiiDA invented. I also find the word _flavors_ very odd in this context.

Since we already being pedantic about naming, as a far as I know, I don't think we are using the POSIX terms correctly. 

1. The word "parameter" refers to actual values provided as argument.
2. The "positional arguments" should actually be referred to as "operands".
3. A "flag" is an historical term for an option argument, it does not necessarily refer to the convention of prefixing an argument with one or two hyphens.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html
https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap12.html

I'm happy to defer that discussion to a separate issue / PR and just merge this _as is_.

sphuber: I love a good bit of pedantry :) point well taken. I have to admit that the terminology has also been adopted from `click`. Fine to merge this now and discuss this at a later point.

PR #4048, opened 2020-05-04
Docs: move in section "Topics - Processes"

### LINKED ISSUE #4022: Docs: Topic processes

sphuber:
Fixes #4022 

The content that was envisioned for this section was originally divided
over the "Concepts" and "Working" sections, which were itself roughly
divided in information on "Processes", "Calculations" and "Workflows".
The original idea was to nest the material on calculations and workflows
within the global processes topic, since in the code the former are also
sub classes of the latter. However, since there are also documentation
on processes in general, I was not able to make this hierarchy work.

Instead, "Calculations" and "Workflows" are now individul topics on the
same level as "Processes". Each of these topics are roughly subdivided
in a "Concepts" and "Usage" section, where the content from the latter
came from the original "Working" directory. The concepts explains the
concepts of the topic, where the usage really exposes and explains the
API with all the nitty gritty details.


docs/source/topics/processes/index.rst
```diff
@@ -0,0 +1,29 @@
+.. _topics:engine:processes:
+
+*********
+Processes
+*********
+
+Before you start working with processes, make sure you have read and understood the :ref:`basic concept<topics:processes:concepts>`.
```
csadorf, line 7: I don't think this is in the spirit of a background section that we imply that users need to have read to be able to work with processes.

I am also not a fan of this slightly aggressive language that is telling users what to do. I think it's better to use slightly more friendly language that is providing hints and suggestions instead or simply states facts like this:
```suggestion
This section assumes knowledge of the :ref:`basic concepts<topics:processes:concepts>` .
```
We could also add this as a `note::`.

docs/source/topics/processes/index.rst
```diff
@@ -0,0 +1,29 @@
...
+
+    * :ref:`calculation functions<topics:calculations:usage:calcfunctions>`
+    * :ref:`calculation jobs<topics:calculations:usage:calcjobs>`
+    * :ref:`work functions<topics:workflows:usage:workfunctions>`
+    * :ref:`work chains<topics:workflows:usage:workchains>`
+
+Since all of these are types of processes, everything that will be explained in this section, will apply to each and everyone of them.
+That makes it very useful to read and understand this section well, as the concepts apply so broadly.
+However, for the same reason, at times this section may feel a bit abstract.
+It may therefore be advisable to start reading a section on one of the more specific processes listed above first, to get a more concrete example, and then simply refer back here for a more extensive explanation of the details.
```
csadorf, line 19: I appreciate providing guidance to the reader, but the language feels a bit imposing. I'd prefer to stick to more factual statements instead and then let the reader draw their own conclusions:
```suggestion
All of the types listed above are processes, meaning they are concrete implementations of the more abstract concepts described in this section.
```
Again, might be worthwhile to format all of the "guidance for readers" notes as `notes::`. We could even create a distinct sphinx directive for that, but I'm trying to keep things as simple as possible during the transition, so probably better to just stick to `notes::` for now.

docs/source/concepts/provenance.rst
```diff
@@ -65,20 +65,20 @@ In particular, **process nodes** are divided into two sub categories:
 
     - **calculation nodes** (``CalculationNode``): Represent code execution that creates new data. These are further subdivided in two subclasses:
 
-        - :py:class:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode`: Represents the execution of a calculation external to AiiDA, typically via a job batch scheduler (see the concept of :ref:`calculation jobs<concepts_calcjobs>`).
-        - :py:class:`~aiida.orm.nodes.process.calculation.calcfunction.CalcFunctionNode`: Represents the execution of a python function (see the concept of :ref:`calculation functions<concepts_calcfunctions>`).
+        - :py:class:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode`: Represents the execution of a calculation external to AiiDA, typically via a job batch scheduler (see the concept of :ref:`calculation jobs<topics:calculations:concepts:calcjobs>`).
+        - :py:class:`~aiida.orm.nodes.process.calculation.calcfunction.CalcFunctionNode`: Represents the execution of a python function (see the concept of :ref:`calculation functions<topics:calculations:concepts:calcfunctions>`).
```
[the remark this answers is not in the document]
sphuber: yes, I agree

sphuber:
I have pushed a commit with the changes you suggested. I am fine with either form. I kept it as a separate commit so it is easier to review the changes, but when we merge this it should be squashed, getting rid of the commit message.

csadorf:
It's possible that I misinterpreted the intention of the earlier version and my suggested edits introduced some confusion.

When I click on the top-level entry "Processes" it takes me to the "Processes" overview page that now immediately states that this section requires knowledge of the basic concepts, which doesn't make any sense:
Screen Shot 2020-05-07 at 5 18 15 PM

There is also still a strong warning at the beginning of the ["Calculations::Usage" section](https://github.com/aiidateam/aiida-core/blob/bdbe0527875db14fe90da4f17eab1b334b1726f9/docs/source/topics/calculations/usage.rst#usage) that is similar to the one that we converted earlier. Maybe that is actually the right place for the note that I suggested?

I think the best way to untangle this may be to just add a note to each of the "usage" sections that states that the section expects knowledge of the _associated concepts_ section which in turn assumes knowledge of the _process concepts_ section. If that is correct then we can maybe get rid of those notes altogether, because within the topics section we can actually expect a linear flow, so it reads like a book.

sphuber:
> I think the best way to untangle this may be to just add a note to each of the "usage" sections that states that the section expects knowledge of the _associated concepts_ section which in turn assumes knowledge of the _process concepts_ section. If that is correct then we can maybe get rid of those notes altogether, because within the topics section we can actually expect a linear flow, so it reads like a book.

This is exactly what it is supposed to do and what the original docs did, before the reorganization. Apologies for the sloppiness. I will correct it.

csadorf:
> A user may click on "workflows" and see "Usage" and go straight there thinking, "concepts" is boring and I don't need it.

I think that's a risk but it may not be our problem. ;)

sphuber:
I have removed all of the "user-guidance" text, except for the usage sections, that now include `note::` in the beginning that disclaims that knowledge of the preceding "concepts" section is assumed. I see your point that in principle this can already be deduced based on the order of the sections, but I am sure people will skip to "usage" and not understand things. At least in this way, we can't be blamed. Since I also made the language less imperative as you suggested, I think it is fine to keep it now. Let me know if you want to discuss, I am available all day .

sphuber:
@csadorf as discussed during the meeting, I have added a few lines to each index page before the TOC. For calculations and workflows it refers to the corresponding how-to page for a lower barrier introduction. For processes, I mention that this is general information that applies to both calculations and workflows, with a link to each respective topic section

docs/source/topics/processes/index.rst
```diff
@@ -0,0 +1,17 @@
+.. _topics:processes:
+
+*********
+Processes
+*********
+
+This topic section provides extensive and detailed information on the concept of processes in AiiDA and how to work with them.
+Since the concepts explained here apply to processes in general, they also apply to its subtypes, calculations and workflows.
```
csadorf, line 8:
```suggestion
Since the concepts explained here apply to processes in general, they also apply to its subtypes: calculations and workflows.
```

PR #4052, opened 2020-05-06
Include PYTHONPATH in `get_env_with_venv_bin` for use in subprocesses

### LINKED ISSUE #4051: Timeout in unit test `tests.cmdline.commands.test_processes.py:test_pause_play_kill` when running locally
The test will timeout:
```
(aiida_dev) sph@bastion:~/code/aiida/env/dev/aiida-core$ pytest tests/cmdline/commands/test_process.py  -sv -k test_pause_play_kill
============================================================================================== test session starts ==============================================================================================
platform linux -- Python 3.7.6, pytest-5.4.1, py-1.5.4, pluggy-0.13.1 -- /home/sph/.virtualenvs/aiida_dev/bin/python3.7
cachedir: .pytest_cache
rootdir: /home/sph/code/aiida/env/dev/aiida-core, inifile: pytest.ini
plugins: regressions-1.0.6, timeout-1.3.4, asyncio-0.10.0, datadir-1.3.1, cov-2.8.1
collected 10 items / 9 deselected / 1 selected                                                                                                                                                                  

tests/cmdline/commands/test_process.py::TestVerdiProcessDaemon::test_pause_play_kill ['/home/sph/.virtualenvs/aiida_dev/bin/verdi', '-p', 'test_profile', 'devel', 'run_daemon']
FAILED

=================================================================================================== FAILURES ====================================================================================================
__________________________________________________________________________________ TestVerdiProcessDaemon.test_pause_play_kill __________________________________________________________________________________

self = <tests.cmdline.commands.test_process.TestVerdiProcessDaemon testMethod=test_pause_play_kill>

    def test_pause_play_kill(self):
        """
        Test the pause/play/kill commands
        """
        # pylint: disable=no-member
        from aiida.orm import load_node
    
        calc = self.runner.submit(test_processes.WaitProcess)
        start_time = time.time()
        while calc.process_state is not plumpy.ProcessState.WAITING:
            if time.time() - start_time >= self.TEST_TIMEOUT:
>               self.fail('Timed out waiting for process to enter waiting state')
E               AssertionError: Timed out waiting for process to enter waiting state
```

Note that this happens both when specifying an explicit test profile as well as when letting the test manager generate one on the fly.

sphuber:
Fixes #4051 

The test `tests.cmdline.commands.test_process:test_pause_play_kill`
was failing in local setups with a time out. The problem was that the
daemon, launched as a subprocess in the `setUp`, would fail to load the
submitted process, because the class could not be imported. This is
because it is defined in the `tests.utils` module, which would not be
part of the Python path of the subprocess.

The `tests` module is automatically added to the path by `pytest`,
however, it was not being forwarded to the subprocess. Passing the
current environment through the `env` argument solves this. The only
additional change is to include the `PYTHONPATH` variable in the env
copy created by `aiida.cmdline.utils.common.get_env_with_venv_bin`.
These two changes guarantee that the test process classes are importable
by the daemon that is running in the subprocess.

### DIFF
--- a/tests/cmdline/commands/test_process.py
+++ b/tests/cmdline/commands/test_process.py
@@ -8,7 +8,6 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Tests for `verdi process`."""
-
 import datetime
 import subprocess
 import sys
@@ -41,13 +40,19 @@ class TestVerdiProcessDaemon(AiidaTestCase):
 
     def setUp(self):
         super().setUp()
-        from aiida.manage.configuration import get_config
+        from aiida.cmdline.utils.common import get_env_with_venv_bin
         from aiida.engine.daemon.client import DaemonClient
+        from aiida.manage.configuration import get_config
+
+        # Add the current python path to the environment that will be used for the daemon sub process. This is necessary
+        # to guarantee the daemon can also import all the classes that are defined in this `tests` module.
+        env = get_env_with_venv_bin()
+        env['PYTHONPATH'] = ':'.join(sys.path)
 
         profile = get_config().current_profile
         self.daemon_client = DaemonClient(profile)
         self.daemon_pid = subprocess.Popen(
-            self.daemon_client.cmd_string.split(), stderr=sys.stderr, stdout=sys.stdout
+            self.daemon_client.cmd_string.split(), stderr=sys.stderr, stdout=sys.stdout, env=env
         ).pid
         self.runner = get_manager().create_runner(rmq_submit=True)
         self.cli_runner = CliRunner()


sphuber:
Only thing that I am not sure about is that this change will also be used in non-test code, since we use the `get_env_with_venv_bin` function also in normal use. Should we put a switch in to only include the python path when running tests? Normally, these switches tend to be very dirty. Trying to think if it is bad to set the PYTHONPATH for the subprocess from the main. In principle that is the goal of this `get_env_with_venv_bin` function. We want to make sure that daemon runners have the same environment. So I don't think this should be problematic

sphuber:
After discussion with @giovannipizzi we opt to only apply the change to the test itself, until we properly understand the full effect of always including the `PYTHONPATH` when copying the `env`.

PR #4055, opened 2020-05-06
Improving SSH configuration defaults

### LINKED ISSUE #4054: Improve default ssh configuration of computers
Currently, during `computer configure`, the `look_for_keys` is set to False by default. Better to set it to True (paramiko's default). Also, I would set `allow_agent` to True (on the shell, if you have a SSH agent, you expect that it is automatically used by `ssh`).
Finally, if no key is specified (empty string), this should just not be passed to paramiko. Currently, during a `verdi computer test`, I get an error
```
[SshTransport] [Errno 2] No such file or directory: '', connect_args were: {
'gss_host': 'daint.cscs.ch', 
'port': 22, 
'look_for_keys': True, 
'timeout': 60, 
'gss_deleg_creds': False, 
'compress': True, 
'username': 'pizzi', 
'sock': <aiida.transports.util._DetachedProxyCommand object at 0x7f0251adaa20>, 
'gss_auth': False, 
'key_filename': '', 
'allow_agent': True, 
'gss_kex': False
}
```

giovannipizzi:
- `look_for_keys` is now set to `True` by default
  (it's also paramiko's default)
- `allow_agent` is now set to `True` by default (on the shell,
  if you have a SSH agent, you expect that it is automatically
  used by ssh)
- If an empty string is passed as the key_filename, this is not passed
  to paramiko, so it will search for it rather than complaining that
  file `''` does not exist.

Fixes #4054

### DIFF
--- a/aiida/transports/plugins/ssh.py
+++ b/aiida/transports/plugins/ssh.py
@@ -254,7 +254,7 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         Return a suggestion for the specific field.
         """
         config = parse_sshconfig(computer.hostname)
-        return convert_to_bool(str(config.get('allow_agent', 'no')))
+        return convert_to_bool(str(config.get('allow_agent', 'yes')))
 
     @classmethod
     def _get_look_for_keys_suggestion_string(cls, computer):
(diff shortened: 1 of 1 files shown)


giovannipizzi:
Not sure how to write a test for this - this would need to make many assumptions on how the SSH keys are setup in the test machine (and this can be different on each person's computer), etc. But I think that the change makes sense so maybe it's ok to merge even without explicit (additional) tests, w.r.t. those we already have?

sphuber (approved):
Thanks @giovannipizzi . I agree that this doesn't require a test

PR #4058, opened 2020-05-08
Docs: move in section "Topics - Provenance"

### LINKED ISSUE #4017: Docs: Topic provenance

sphuber:
Fixes #4017 

The content is largely untouched and mostly moved in to its new place in
the scaffolding as is. There were slight changes in link labels and
header characters to adhere to new style guide rules.


sphuber:
Applied the comma-colon change cap'n!

sphuber:
I tried the split to see what it would look like. I'm in principle always for more consistency, but maybe in this case the implementation and especially the concept page look a bit light weight. Then again, I agree that everything in a single page was getting a bit bulky. I would be fine with the split, especially given that in the future we might probably want to add more information.

csadorf:
> @csadorf I tried the split to see what it would look like. I'm in principle always for more consistency, but maybe in this case the implementation and especially the concept page look a bit light weight. Then again, I agree that everything in a single page was getting a bit bulky. I would be fine with the split, especially given that in the future we might probably want to add more information.

I actually think that the length of the pages is perfect now. I'll accept and then we can merge this.

PR #4066, opened 2020-05-12
Docs: add "How to manage your installation - Managing profiles"

sphuber:
Fixes #4000 

This is a new section that is intended to describe the function of
profiles and the various commands that can be used to work with them.
Note that the explanation on the purpose and details of profiles is kept
to a minimum as this is a How-to guide and is merely intended to help
users with goal oriented problems.

The section quickly explains the various profile commands:

    * verdi profile list
    * verdi profile setdefault
    * verdi profile show
    * verdi profile delete

Information on how further configuration through profiles is reserved
for the next how-to section "Configuring your installation".

PR #4068, opened 2020-05-13
Docs: add "How to - Configuring your installation"

### LINKED ISSUE #4001: Docs: How to configure your installation

sphuber:
Fixes #4001

This section describes how multiple instances of AiiDA can be isolated
on the same machine, how tab-completion can be activated and how the
`verdi config` command can be used to configure profile options.

Except for the `verdi config` section, which is completely new, the
other content has been adapted from existing documentation.

### DIFF
--- a/docs/source/howto/installation.rst
+++ b/docs/source/howto/installation.rst
@@ -74,7 +74,148 @@ This behavior can be changed using the ``--skip-repository`` and ``--skip-db`` o
 Configuring your installation
 =============================
 
-`#4001`_
+.. _how-to:installation:configure:tab-completion:
+
+Activating tab-completion
+-------------------------
+The ``verdi`` command line interface has many commands and parameters, which can be tab-completed to simplify its use.
+To enable tab-completion, the following shell command should be executed:
+
+.. code:: bash
+
+    $ eval "$(_VERDI_COMPLETE=source verdi)"
+
+Place this command in your shell or virtual environment activation script to automatically enable tab completion when opening a new shell or activating an environment.
+This file is shell specific, but likely one of the following:
+
+    * the startup file of your shell (``.bashrc``, ``.zsh``, ...), if aiida is installed system-wide
+    * the `activation script <https://virtualenv.pypa.io/en/latest/userguide/#activate-script>`_ of your virtual environment
+    * a `startup file <https://conda.io/docs/user-guide/tasks/manage-environments.html#saving-environment-variables>`_ for your conda environment
+
+
+.. important::
+
+    After you have added the line to the start up script, make sure to restart the terminal or source the script for the changes to take effect.
+
+
+.. _how-to:installation:configure:options:
+
+Configuring profile options
+---------------------------
+AiiDA provides various configurational options for profiles, which can be controlled with the :ref:`verdi config<reference:command-line:verdi-config>` command.
+To set a configurational option, simply pass the name of the option and the value to set ``verdi config OPTION_NAME OPTION_VALUE``.
+The available options are tab-completed, so simply type ``verdi config`` and thit <TAB> twice to list them.
+
+For example, if you want to change the default number of workers that are created when you start the daemon, you can run:
+
+.. code:: bash
+
+    $ verdi config daemon.default_workers 5
+    Success: daemon.default_workers set to 5 for profile-one
+
+You can check the currently defined value of any option by simply calling the command without specifying a value, for example:
+
+.. code:: bash
+
+    $ verdi config daemon.default_workers
+    5
+
+If no value is displayed, it means that no value has ever explicitly been set for this particular option and the default will always be used.
+By default any option set through ``verdi config`` will be applied to the current default profile.
+To change the profile you can use the :ref:`profile option<topics:cli:profile>`.
+
+To undo the configuration of a particular option and reset it so the default value is used, you can use the ``--unset`` option:
+
+.. code:: bash
+
+    $ verdi config daemon.default_workers --unset
+    Success: daemon.default_workers unset for profile-one
+
+If you want to set a particular option that should be applied to all profiles, you can use the ``--global`` flag:
+
+.. code:: bash
+
+    $ verdi config daemon.default_workers 5 --global
+    Success: daemon.default_workers set to 5 globally
+
+and just as on a per-profile basis, this can be undone with the ``--unset`` flag:
+
+.. code:: bash
+
+    $ verdi config daemon.default_workers --unset --global
+    Success: daemon.default_workers unset globally
+
+.. important::
+
+    Changes that affect the daemon (e.g. ``logging.aiida_loglevel``) will only take affect after restarting the daemon.
+
+
+.. _how-to:installation:configure:instance-isolation:
+
+Isolating multiple instances
+----------------------------
+An AiiDA instance is defined as the installed source code plus the configuration folder that stores the configuration files with all the configured profiles.
+It is possible to run multiple AiiDA instances on a single machine, simply by isolating the code and configuration in a virtual environment.
+
+To isolate the code, simply create a virtual environment, e.g., with conda or venv, and then follow the instructions for :ref:`installation<intro/install_advanced>` after activation.
+Whenever you activate this particular environment, you will be running the particular version of AiiDA (and all the plugins) that you installed specifically for it.
+
+This is separate from the configuration of AiiDA, which is stored in the configuration directory which is always named ``.aiida`` and by default is stored in the home directory.
+Therefore, the default path of the configuration directory is ``~/.aiida``.
+By default, each AiiDA instance (each installation) will store associated profiles in this folder.
+A best practice is to always separate the profiles together with the code to which they belong.
+The typical approach is to place the configuration folder in the virtual environment itself and have it automatically selected whenever the environment is activated.
+
+The location of the AiiDA configuration folder, can be controlled with the ``AIIDA_PATH`` environment variable.
+This allows us to change the configuration folder automatically, by adding the following lines to the activation script of a virtual environment.
+For example, if the path of your virtual environment is ``/home/user/.virtualenvs/aiida``, add the following line:
+
+.. code:: bash
+
+    $ export AIIDA_PATH='/home/user/.virtualenvs/aiida'
+
+Make sure to reactivate the virtual environment, if it was already active, for the changes to take effect.
+
+.. note::
+
+   For ``conda``, create a directory structure ``etc/conda/activate.d`` in the root folder of your conda environment (e.g. ``/home/user/miniconda/envs/aiida``), and place a file ``aiida-init.sh`` in that folder which exports the ``AIIDA_PATH``.
+
+You can test that everything works by first echoing the environment variable with ``echo $AIIDA_PATH`` to confirm it prints the correct path.
+Finally, you can check that AiiDA know also properly realizes the new location for the configuration folder by calling ``verdi profile list``.
+This should display the current location of the configuration directory:
+
+.. code:: bash
+
+    Info: configuration folder: /home/user/.virtualenvs/aiida/.aiida
+    Critical: configuration file /home/user/.virtualenvs/aiida/.aiida/config.json does not exist
+
+The second line you will only see if you haven't yet setup a profile for this AiiDA instance.
+For information on setting up a profile, refer to :ref:`creating profiles<how-to:installation:profile>`.
+
+Besides a single path, the value of ``AIIDA_PATH`` can also be a colon-separated list of paths.
+AiiDA will go through each of the paths and check whether they contain a configuration directory, i.e., a folder with the name ``.aiida``.
+The first configuration directory that is encountered will be used as the configuration directory.
+If no configuration directory is found, one will be created in the last path that was considered.
+For example, the directory structure in your home folder ``~/`` might look like this::
+
+    .
+    ├── .aiida
+    └── project_a
+        ├── .aiida
+        └── subfolder
+
+If you leave the ``AIIDA_PATH`` variable unset, the default location ``~/.aiida`` will be used.
+However, if you set:
+
+.. code:: bash
+
+    $ export AIIDA_PATH='~/project_a:'
+
+the configuration directory ``~/project_a/.aiida`` will be used.
+
+.. warning::
+
+    If there was no ``.aiida`` directory in ``~/project_a``, AiiDA would have created it for you, so make sure to set the ``AIIDA_PATH`` correctly.
 
 
 .. _how-to:installation:performance:
@@ -110,7 +251,6 @@ Managing multiple users
 
 
 
-.. _#4001: https://github.com/aiidateam/aiida-core/issues/4001
 .. _#4002: https://github.com/aiidateam/aiida-core/issues/4002
 .. _#4003: https://github.com/aiidateam/aiida-core/issues/4003
 .. _#4004: https://github.com/aiidateam/aiida-core/issues/4004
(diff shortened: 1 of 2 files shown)


sphuber:
Two things here:
1. In the "Isolating multiple instances" section, should there be a link to more detailed information on how to work with virtual environments? I feel that we should not be responsible for explaining this concept, and certainly not in this how-to section, but on the other hand I fear this section maybe difficult to understand for people that have never even heard of a virtual environment. Then again, would those be the kind of people to read this section?
2. I have the slight feeling that this how-to should actually come before the "Managing profiles" of PR #4066

greschd:
I think it should be mentioned somewhere that using the same code (virtualenv) for different ``.aiida`` folders works perfectly fine, while the opposite (multiple code versions running on the same ``.aiida`` folder) is bound to lead to problems.

sphuber:
> I think it should be mentioned somewhere that using the same code (virtualenv) for different `.aiida` folders works perfectly fine, while the opposite (multiple code versions running on the same `.aiida` folder) is bound to lead to problems.

Good point. This is more or less implied by the later detailed explanation that the `AIIDA_PATH` can support multiple paths, correct. Should I add your comment there in a `.. note::` perhaps?

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
+It is possible to run multiple AiiDA instances on a single machine, simply by isolating the code and configuration in a virtual environment.
+This section will explain how to accomplish this.
+
+To isolate the code, simply create a virtual environment with the tool of your choosing :ref:`SHOULD THERE BE A LINK HERE?<>` and :ref:`run the installation<intro/install_advanced>` while it is activated.
+Whenever you activate this particular environment, you will be running the particular version of AiiDA (and all the plugins) that you installed specifically for it.
+
+Then there is the configuration of AiiDA, which is stored in the configuration directory which is always named ``.aiida`` and by default is stored in the home directory.
+Therefore, the default path of the configuration directory is ``~/.aiida``.
+When unchanged, each AiiDA instance (each installation in a virtual environment) will store the profiles that are configured for it in this same folder.
+However, often one would like to keep the profile for each instance separate.
```
greschd, line 35: Maybe, we should just use stronger language here? It should be clear that separating the ``.aiida`` folders is the "default" thing to do, and you need to be careful if you don't. 

For my understanding, if the ``.aiida`` folders are _not_ split, the user needs to make sure to only ever use the "right" profiles with a given virtualenv, right?

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
-`#4001`_
+.. _how-to:installation:configure:instance-isolation:
+
+Isolating multiple instances
+----------------------------
+An AiiDA instance is defined as the installed source code plus the configuration folder that stores the configuration files with all the configured profiles.
+It is possible to run multiple AiiDA instances on a single machine, simply by isolating the code and configuration in a virtual environment.
+This section will explain how to accomplish this.
+
+To isolate the code, simply create a virtual environment with the tool of your choosing :ref:`SHOULD THERE BE A LINK HERE?<>` and :ref:`run the installation<intro/install_advanced>` while it is activated.
```
csadorf, line 29:
```suggestion
To isolate the code, simply create a virtual environment, e.g., with conda or venv, and then follow the instructions for :ref:`installation<intro/install_advanced>` after activation.
```

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
 =============================
 
-`#4001`_
+.. _how-to:installation:configure:instance-isolation:
+
+Isolating multiple instances
+----------------------------
+An AiiDA instance is defined as the installed source code plus the configuration folder that stores the configuration files with all the configured profiles.
+It is possible to run multiple AiiDA instances on a single machine, simply by isolating the code and configuration in a virtual environment.
+This section will explain how to accomplish this.
```
csadorf, line 27: I'd drop that line, self-evident.

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
+
+To isolate the code, simply create a virtual environment with the tool of your choosing :ref:`SHOULD THERE BE A LINK HERE?<>` and :ref:`run the installation<intro/install_advanced>` while it is activated.
+Whenever you activate this particular environment, you will be running the particular version of AiiDA (and all the plugins) that you installed specifically for it.
+
+Then there is the configuration of AiiDA, which is stored in the configuration directory which is always named ``.aiida`` and by default is stored in the home directory.
```
csadorf, line 32:
```suggestion
This is separate from the configuration of AiiDA, which is stored in the configuration directory which is always named ``.aiida`` and by default is stored in the home directory.
```

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
+Therefore, the default path of the configuration directory is ``~/.aiida``.
+When unchanged, each AiiDA instance (each installation in a virtual environment) will store the profiles that are configured for it in this same folder.
```
csadorf, line 34:
```suggestion
By default, each AiiDA instance (each installation) will store associated profiles in this folder.
```

docs/source/howto/installation.rst
csadorf: I'm ok with using stronger language or leaving as is. It seems to me like the possibility of a version mismatch should actually be addressed programatically.

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
+However, often one would like to keep the profile for each instance separate.
+This can be accomplished by simply changing the location of the AiiDA configuration folder, which can be controlled with the ``AIIDA_PATH`` environment variable.
```
csadorf, line 36:
```suggestion
The location of the AiiDA configuration folder, can be controlled with the ``AIIDA_PATH`` environment variable.
```

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
+
+To change the ``AIIDA_PATH`` variable, simply add it to the activation script of your virtual environment.
+For example, if the path of your virtual environment is ``/home/user/.virtualenvs/aiida``, add the following line:
```
csadorf, line 39:
```suggestion
This allows us to change the configuration folder automatically, by adding the following lines to the activation script of a virtual environment.
For example, if the path of your virtual environment is ``/home/user/.virtualenvs/aiida``, add the following line:
```

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
+-------------------------
+The ``verdi`` command line interface has many commands and parameters, which can be tab-completed to simplify its use.
+To enable tab-completion, the following shell command should be executed:
+
+.. code:: bash
+
+    eval "$(_VERDI_COMPLETE=source verdi)"
+
+Note that this *only* enables tab-completion in the *current shell*.
+In order to make this change permanent and have it automatically applied to any shell instance, you can add the command to the start up script.
```
csadorf, line 100:
```suggestion
Place this command in your shell or virtual environment activation script to automatically enable tab completion when opening a new shell or activating an environment.
```

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
+This file is shell specific, but likely one of the following:
+
+    * the startup file of your shell (``.bashrc``, ``.zsh``, ...), if aiida is installed system-wide
+    * the `activation script <https://virtualenv.pypa.io/en/latest/userguide/#activate-script>`_ of your virtual environment
+    * a `startup file <https://conda.io/docs/user-guide/tasks/manage-environments.html#saving-environment-variables>`_ for your conda environment
+
+
+.. important::
+
+    After you have added the line to the start up script, make sure to restart the terminal or source the script for the changes to take effect
```
csadorf, line 110: Does the `.. important::` directive require two empty lines surrounding it?

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
+.. _how-to:installation:configure:options:
+
+Configuring profile options
+---------------------------
+In addition to the configuration of AiiDA instances described in the previous sections, AiiDA also provides various configurational options for profiles.
+To control the configuration options of profiles use the :ref:`verdi config<reference:command-line:verdi-config>` command.
+To set a configurational option, simply pass the name of the option and the value to set ``verdi config OPTION_NAME OPTION_VALUE``.
+The available options are tab-completed, so simply type ``verdi config`` and thit <TAB> twice to list them.
+
+For example, if you want to change the default number of workers that are created when you start the daemon, you can run::
```
csadorf, line 122:
```suggestion
For example, if you want to change the default number of workers that are created when you start the daemon, you can run:
```

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
+
+.. code:: bash
+
+    (aiida): verdi config daemon.default_workers 5
```
csadorf, line 126: Do we already have a convention as to how to present a shell prompt? I believe that for the tutorial we have decided to use `$`.

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,149 @@ Managing profiles
...
+and just as on a per-profile basis, this can be undone with the ``--unset`` flag:
+
+.. code:: bash
+
+    (aiida): verdi config daemon.default_workers --unset --global
+    Success: daemon.default_workers unset globally
+
+.. important::
+
+    Remember that if you change options that affect the daemon as well (e.g. the `logging.aiida_loglevel``), the changes will not take effect until *after* you have restarted the daemon.
```
csadorf, line 163:
```suggestion
    Changes that affect the daemon (e.g. ``logging.aiida_loglevel``) will only take affect after restarting the daemon.
```
It preferable to use affirmative language for these kinds of notes, see [here](https://devguide.python.org/documenting/#affirmative-tone) for an explanation.

docs/source/howto/installation.rst
sphuber: Good point. Using the `$` is fine. In some places prepending a virtual environment is useful to remind people they should load it first. Then again, it may look differently for different virtual environment types, people might intentionally decide to install system wide. On top of that, we would have to use this everywhere in the documentation, even in places where this might be considered self-evident. I guess I am just making the point for myself that I will switch to just the `$` as prompt ^^

docs/source/howto/installation.rst
sphuber: Thanks for the link, very useful. Will adapt it

docs/source/howto/installation.rst
sphuber: It doesn't require anything different from the other markup marks such as `.. note::` but I see just a single empty line. Did you mean two linebreaks? I personally like the empty line for readability, but happy to change. What would the standard be? Directly following the mark, or a single linebreak? I guess the former would become problematic if the note contains multiple sentences. So I would vote for either 1 or 2 linebreaks

sphuber:
@csadorf what do you think about my question regarding the order of this section and that of my other PR. Would you agree that it makes sense to swap the current order and put this one first and the "Managing profiles" second?

docs/source/howto/installation.rst
greschd: Hmm, not sure if there is a good way to catch this programmatically. The problems I can see happening are:
- Using two different AiiDA versions: The newer AiiDA version can trigger a DB migration, after which the older version will complain. The question here is how to distinguish it from a "normal" upgrade.
- Different plugin / dependency versions: This is problematic when the daemon and the launching process run on different versions. The persisted process can probably  be "wrong" in countless ways.

One way I could see of handling this would be pinning either ``which verdi`` or ``which python`` in the profile - if it changes, flash a big warning and ask for confirmation to change it.

Nevertheless, I think it should be made clear in the documentation - ideally the programmatic catch is there only as a safety net, and users are led to do the right thing.

csadorf:
> @csadorf what do you think about my question regarding the order of this section and that of my other PR. Would you agree that it makes sense to swap the current order and put this one first and the "Managing profiles" second?

I think it depends on what we consider a more widely applicable scenario:
1. Having multiple software installations.
2. Using multiple profiles.

I'm assuming it's #1 for the expert user and #2 for the average user, what do you think?

sphuber:
> I'm assuming it's #1 for the expert user and #2 for the average user, what do you think?

This is the problem though, #1 is not necessarily just for expert users. Since we always recommend the use of virtual environments, the information of the `AIIDA_PATH` is also pertinent to those situations, even though there is just a single instance of AiiDA.

There is even a comment from a user on #4001 that suggest that this information should be close to the installation. I am not convinced that we should, but it does show that it might be valuable information also for starting users with a single instance.

Of course, not knowing about `AIIDA_PATH` will just use the standard location, so thinking about it, a beginning user does not _need_ to know about it, so maybe we shouldn't. I guess the current order is fine then.

csadorf:
@sphuber I think you misunderstood my comment. I did not want to imply that #1 is only applicable for expert aiida users nor that #2 is only applicable to average aiida users. What I'm asking is: which scenario you are more likely to encounter as the average user? That should drive the order of these two sections.

sphuber:
@csadorf  I see. Well, I guess parts of part 2 (enabling tab-completion and `verdi config`) are just as useful for beginning users as part 1. But the part on having isolating multiple environments is actually more for experienced users. So maybe there is even a problem in structure _within_ part 2. I guess I can just move those two sub parts to the beginning since they don't depend at all on the isolation of environments part.

docs/source/howto/installation.rst
sphuber: I tend to agree with @greschd that we should guide the user to do the right thing and just have a safety net. We already have a check for `aiida-core` that will check the version of the database schema against that of the code to check that they are compatible. However, this does not (and cannot) check the code of the plugins. Moreover, there is currently a way that the check can be circumvented (by accident). If you install a new version of AiiDA and configure a profile for an existing database using `verdi setup` the command will automatically migrate it. Since new databases always have to be migrated, this is done automatically and unprompted. This can be unintended and therefore unexpected, especially given the reverse migration is not possible. Now this is of course a bug of sorts that we should perhaps fix (only automatically migrate with `verdi setup` if the database is "empty"), but my point is that there might be more subtle problems like this that we might not have thought off. TLDR: I think it is indeed wise to make the distinction that Dominik described, just for completeness and clarity.

greschd: I think this point is resolved with the changes made, I'll let @csadorf review the rest.

docs/source/howto/installation.rst
```diff
@@ -18,7 +18,148 @@ Managing profiles
...
+
+.. code:: bash
+
+    $ export AIIDA_PATH='~/project_a:'
+
+the configuration directory ``~/project_a/.aiida`` will be used.
+
+.. warning::
+
+    If there was no ``.aiida`` directory in ``~/project_a``, AiiDA would have created it for you, so make sure to set the ``AIIDA_PATH`` correctly.
```
csadorf, line 162: I'm a bit confused by this warning. Why do I need to worry about this?

Would the following warning be correct?
```suggestion
    AiiDA will create the ``.aiida`` directory in ``~/project_a`` in case that it does not exist yet.
```

Or is the warning related to the fact that it will create the _complete path_ and not only the `.aiida/` leaf?

sphuber: The "problem" is that AiiDA will create the folder silently if it does not exist and will not raise an error. If a user, by accident, specifies the wrong path, the folder will be created in a different path then expected. Not that this will have grave consequences but it might be unexpected.

Issue #3981, opened 2020-04-30
Docs: Create a new tutorial for first-time users


mbercx:
Create a short (~30 min) tutorial to give beginning users a first introduction on how to use AiiDA.

giovannipizzi:
Here a possible short code that can be used as CalcJob, avoiding to have to create a custom plugin.

1. Make sure they start with a `localhost` computer (probably with `direct` scheduler and `local` transport)

2. Ask them to setup a `bash@localhost` code (good excuse to explain what a code is), with the following one-liner:
   ```bash
   verdi code setup -L bash -P templatereplacer --on-computer --computer=localhost --remote-abs-path=/bin/bash --non-interactive
   ```
   Optionally, you can explain them `verdi code list` here.

3. Ask them to run a calculation with the following script:
   ```python
   from aiida.orm import Dict, load_code
   from aiida.engine import run_get_node
   
   code = load_code('bash@localhost')
   
   builder = code.get_builder()
   builder.parameters = Dict(dict={'x': 3, 'y': 4})
   builder.template = Dict(
       dict={
           'input_file_name': 'script.sh',
           'input_file_template': "echo $(({x} * {y}))",
           'cmdline_params': ['script.sh'],
           'output_file_name': 'stdout.txt'
       }
   )
   builder.metadata.options = {
       'resources': {'num_machines': 1},
       'parser_name': 'templatereplacer.doubler',
   }
   
   results, calcjobnode = run_get_node(builder)
   
   print(calcjobnode.res.value)
   ```

Possible things to adapt:
- define a better parser (e.g. `generic.integer_parser`, `generic.cvs_parser`, `generic.json_parser`) that could become part of AiiDA
- use `print(results['output_parameters'].get_dict()['value'])` instead of `calcjobnode.res.value` (a bit more explicit, but maybe they understand what is going on.
- should we extend the `template_replacer` so that if an `input_file_template` is provided, the `input_file_name` is set by default (maybe using the metadata option that now AiiDA has), and also the cmdline_params becomes by default just the filename? This would remove two lines from above

We should then think if instead of multiplying two numbers we want to do something more fancy, but the snippet above should be enough to extend the example.

Note also that the advantage of the input `parameters` dict is that the values can then be queried.
And, even if slightly more complex than the add calculation, this makes it obvious how this can be extended to any code without the need to write any new calculation plugin.

Issue #4062, opened 2020-05-08
Docs: Consider adding the sphinx-copybutton extension


chrisjsewell:
See: https://sphinx-copybutton.readthedocs.io

Very simple, you just add `sphinx_copybutton` to the extension list in `conf.py`

csadorf:
@chrisjsewell In favor, can you create the PR?

chrisjsewell:
Will do

mbercx:
I tried out the `sphinx-copybutton` extension for the tutorial, and was wondering if you can somehow configure for which code blocks the copy button is added. The issue is that some of the code blocks also include the output, so for these I wouldn't add the copy button. I also was only able to configure one prompt with the `copybutton_prompt_text` configuration, which wasn't ideal because the tutorial has both Python and terminal code blocks.

chrisjsewell:
> if you can somehow configure for which code blocks the copy button is added

@mbercx yes you can, see https://sphinx-copybutton.readthedocs.io/en/latest/#configure-the-css-selector-used-to-add-copy-buttons.
In the PR I can try something like:

```python
copybutton_selector = "div.highlight:not(.no-copy) pre"
```

Such that you "should" be able to do

```rst
.. code-block:: python
    :class: no-copy
```

FYI in #4060 I am removing the `accordion` directive I recall you had added to your tutorial, since I've moved the functionality to `dropdown` in https://sphinx-panels.readthedocs.io/en/latest/#dropdown-usage 😄

mbercx:
@chrisjsewell Nice! I saw the section on the selector, but my css knowledge was too limited to put it into practise. 😄 

Good to know about the switch from `accordion` to `dropdown`! I'll try and remember to fix that in the final version of the tutorial.

PR #4070, opened 2020-05-13
Docs: Add first draft for tutorial

mbercx:
First draft of the AiiDA tutorial (See Issue #3981). Next to adding the tutorial itself, I've also:
- Changed the ArithmeticAddCalculation to set defaults for the resources and use a basic bash execution on the input files.
- Added the MultiplyAddWorkChain for demonstration purposes.
- Set the background-color of bash code snippets to `aliceblue`.
- Added the `sphinx-copybutton` extension. 

Some outstanding issues/questions:
- The `sphinx-copybutton` extension works fine, but perhaps it would be better if we can specify which code snippets the copy button is added to. Now using the copy button for many verdi shell snippets also copies the prompt. You can specify _one_ prompt that is not included when copying the snippet contents, which I’ve currently set to ‘`$ `‘ for the terminal snippets.  
- The work directory of the `tutor` computer is now set up under `$PWD/work`. I figured it might be nice to have them explore the contents of this directory at the end of the tutorial, so they can see how AiiDA uses the UUID to organise the directories. 
- Currently, the tutorial does not include querying. 

I have gone through the tutorial myself in 13 minutes (reading the text aloud to slow me down a bit). I think the size is ok at this point, but we might want to cut some parts (if possible) and still include an example on querying.


sphuber:
Thanks @mbercx . Since this change also changes existing code and adds the workflow, I wonder if we should add this in a separate commit straight onto `develop`. Since there are no tests in the `docs-revamp` that will depend on these changes, there are no problems to then have this only `develop` until we merge `docs-revamp`.

mbercx:
I have moved the changes to the `ArithmeticAddCalculation`, as well as the new `MultiplyAddWorkChain`, to a different branch. Once I've adjusted the tests and am able to run them successfully locally, I'll rebase this branch onto the `develop` branch and do the PR there.

I have also removed the `sphinx-copybutton` extension for now, as @chrisjsewell is working on this (see #4062). I'll also leave the issue I had regarding the prompt there.

csadorf:
@mbercx Sorry, this fell off my radar. Is this ready for my review?

mbercx:
> @mbercx Sorry, this fell off my radar. Is this ready for my review?

No worries! The text is ready for review, yes, but note that not all of the code snippets will work as e.g. the `MultiplyAddWorkChain` has been removed from this PR to commit it straight into `develop`.

I have also used the `code-block` directive for Python code snippets here, but later noticed that many other parts of the documentation use a double colon `::`at the end of the preceding paragraph. Have we established a convention for this?

csadorf:
> > @mbercx Sorry, this fell off my radar. Is this ready for my review?
> 
> No worries! The text is ready for review, yes, but note that not all of the code snippets will work as e.g. the `MultiplyAddWorkChain` has been removed from this PR to commit it straight into `develop`.

Ok, that's not a problem, because I did not plan on testing the tutorial before we are further down the line anyways.

> I have also used the `code-block` directive for Python code snippets here, but later noticed that many other parts of the documentation use a double colon `::`at the end of the preceding paragraph. Have we established a convention for this?

Good point, we had not established a convention for that yet. I suggest that we just use a double colon. Unless there are objections I will put that into the [wiki page](https://github.com/aiidateam/aiida-core/wiki/Writing-documentation#style-guide) and ensure consistency prior to merging of the docs-revamp branch.

sphuber:
> Good point, we had not established a convention for that yet. I suggest that we just use a double colon. Unless there are objections I will put that into the [wiki page](https://github.com/aiidateam/aiida-core/wiki/Writing-documentation#style-guide) and ensure consistency prior to merging of the docs-revamp branch.

The only question I have there is weather the double colon will robustly determine the correct language for the highlighting. Even if that is guaranteed, having an explicit `.. code-block::` with explicit language might also be a lot clearer for the reader of the raw docs.

csadorf:
> > Good point, we had not established a convention for that yet. I suggest that we just use a double colon. Unless there are objections I will put that into the [wiki page](https://github.com/aiidateam/aiida-core/wiki/Writing-documentation#style-guide) and ensure consistency prior to merging of the docs-revamp branch.
> 
> The only question I have there is weather the double colon will robustly determine the correct language for the highlighting. Even if that is guaranteed, having an explicit `.. code-block::` with explicit language might also be a lot clearer for the reader of the raw docs.

[It is assumed to be Python](https://www.sphinx-doc.org/en/1.5/markup/code.html#showing-code-examples). If I understand this correctly then one would use `.. highlight::` directive to control the syntax highlighting and other things like line numbers.

On the other hand, `.. codeblock::` requires one to [explicitly set the language every time](https://github.com/marshallward/vim-restructuredtext/issues/17), right? I'm honestly not sure what is better here.

sphuber:
Since we sometimes also have `bash` blocks, I would opt for consistency and clarity and use `.. code-block:: python`.

mbercx:
> Even if that is guaranteed, having an explicit `.. code-block::` with explicit language might also be a lot clearer for the reader of the raw docs.

As someone who had little experience with reStructuredText before we started working on the docs revamp, I second this notion!

Also, I just remembered that if we are using the `sphinx-copybutton` extension, as per @chrisjsewell's suggestion, you can use a CSS selector to indicate that a code block should not have a copybutton, see issue #4062.

csadorf:
Ok, then it's settled `.. code-block::` it is, I'll update the wiki.

chrisjsewell:
While we are at it, did you want to add in the wiki, some extra clarification over writing code blocks,
i.e. IMO there is probably 4 distonct types of code block that we use a lot:

- Bash shell scripts should use `bash`
  ```rst
  .. code-block:: bash

      verdi node show 1
  ```
- Bash shell **sessions** (i.e. interactive) should use `console`
  ```rst
  .. code-block:: console

      $ verdi node show 1
  ```
- Python scripts should use `python`
  ```rst
  .. code-block:: python

      from aiida import orm
  ```
- Python sessions should use `ipython`
  ```rst
  .. code-block:: ipython

      In  [1]: print("a")
      Out [1]: "a"
  ```

Also in the guidelines

> Use the `*` symbol for itemized lists

This seems like a weird symbol to use for itemized, since it is also used for *emphasis*/**bold**. Would `-` not be better?

csadorf:
> While we are at it, perhaps in the wiki, we want to add some extra clarification over writing code blocks,

Yes, this actually came up in another PR as well. I would just go ahead and copy your examples if you don't mind.

> 
> Also in the guidelines
> 
> > Use the `*` symbol for itemized lists
> 
> This seems like a weird symbol to use for itemized, since it is also used for _emphasis_/**bold**. Would `-` not be better?

I just made a decision, I believe based on prevalence. I don't really think that using `*` is a problem.

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,564 @@
...
+The following Python function:
+
+.. code-block:: python
+
+    def multiply(x, y):
+        return x * y
+
+will give the desired result when applied to two ``Int`` nodes, but the calculation will not be stored in the provenance.
+However, AiiDA ships with a handy ``decorator`` that you can add to any Python function so its execution will be stored in the provenance.
+Start up the AiiDA IPython shell again using ``verdi shell`` and execute the following code snippet:
```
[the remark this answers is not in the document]
mbercx: Agreed! I originally left out the prompts because they were also copied with the `sphinx-copybutton`, but @chrisjsewell is working on fixing this with regexes.

mbercx:
Thanks a lot for the review, @csadorf! Although I'm not sure what connotation of handy you are referring to. ^^

> * [x]  Make sure that the code-blocks are marked up as shown [here](https://github.com/aiidateam/aiida-core/wiki/Writing-documentation#code-segments).
> * [x]  Make sure that the user always knows whether code is supposed to be executed in a script or within a Python interpreter. I found some ambiguity with respect to that.

As all code snippets are supposed to be run in the `verdi shell`, I adjusted them to make this clear to the user. However, copying code snippets with multiple lines is now very tedious manually, so we should probably add a note to point out the `sphinx-copybutton`'s to make sure users don't miss them, once they have been integrated with regexes for the prompts.

csadorf:
@mbercx Is this ready for another round of review?

mbercx:
> @mbercx Is this ready for another round of review?

Yes! Sorry, I hadn't discovered the re-request review button yet. ^^

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+This is the directory in which the calculations running on the ``tutor`` computer will be executed.
+
+Let's now start up the ``verdi shell`` again and load the ``add@tutor`` code using its label:
+
+.. code-block:: ipython
 
-See `issue #3981 <https://github.com/aiidateam/aiida-core/issues/3981>`_.
+    In [1]: code = load_code(label='add')
 
+Every code has a convenient tool for setting up the required input, called the builder.
```
csadorf, line 293:
```suggestion
Every code has a convenient tool for setting up the required input, called the *builder*.
```

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+It can be obtained by using the ``get_builder`` method:
+
+.. code-block:: ipython
+
+    In [2]: builder = code.get_builder()
+
+Using the builder, you can easily set up the calculation by directly providing the input arguments.
+Let's use the ``Int`` node that was created by our previous ``calcfunction`` as one of the inputs:
```
csadorf, line 301:
```suggestion
Let's use the ``Int`` node that was created by our previous ``calcfunction`` as one of the inputs and a new node as the second input:
```

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+
+.. code-block:: ipython
+
+    In [3]: builder.x = load_node(pk=4)
+       ...: builder.y = Int(5)
```
csadorf, line 306: The prompt doesn't look right to me.
```suggestion
    In [3]: builder.x = load_node(pk=4)
    In [4]: builder.y = Int(5)
```

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+.. code-block:: ipython
+
+    In [2]: builder = code.get_builder()
+
+Using the builder, you can easily set up the calculation by directly providing the input arguments.
+Let's use the ``Int`` node that was created by our previous ``calcfunction`` as one of the inputs:
+
+.. code-block:: ipython
+
+    In [3]: builder.x = load_node(pk=4)
```
csadorf, line 305: Is there a good way to figure out which PK to use here in case that mine differs from the tutorial?

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+
+    One nifty feature of the builder is the ability to use tab completion for the inputs.
+    Try it out by typing ``builder.`` + ``<TAB>`` in the verdi shell.
+
+To execute the ``CalcJob``, we use the ``run`` function provided by the AiiDA engine:
+
+.. code-block:: ipython
+
+    In [4]: from aiida.engine import run
+       ...: run(builder)
```
csadorf, line 318: This command failed for me with:
```
ValueError: Error occurred validating port 'inputs.metadata.options.resources': required value was not provided for 'resources'
```

csadorf, line 318:
```suggestion
    In [5]: run(builder)
```

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+
+Wait for the process to complete.
+Once it is done, it will return a dictionary with the output nodes.
```
csadorf, line 321: Show?

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+.. code-block:: ipython
+
+    In [1]: from aiida.engine import submit
+       ...:
+       ...: code = load_code(label='add')
+       ...: builder = code.get_builder()
+       ...: builder.x = load_node(pk=4)
+       ...: builder.y = Int(5)
+       ...:
+       ...: submit(builder)
```
csadorf, line 394: Same question as before, why not execute those commands on separate code lines?

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+
+Note that the submission finishes very quickly, and that it returns the ``CalcJob`` that was just submitted:
```
csadorf, line 396: Why should we note that the submission finishes very quickly?

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+                self.out('result', self.ctx['addition'].get_outgoing().get_node_by_label('sum'))
+
+    First, we recognize the ``multiply`` function we have used earlier, decorated as a ``calcfunction``.
+    The ``define`` class method specifies the ``input`` and ``output`` of the ``WorkChain``, as well as the ``outline``, which are the steps of the workflow.
+    These steps are provided as methods of the ``MultiplyAddWorkChain`` class.
+
+.. note::
+
+    Besides WorkChain's, workflows can also be implemented as *work functions*.
+    These are ideal for workflows that are not very computationally intensive and can be easily implemented in a Python function.
```
csadorf, line 506: I think we need to add some space after the dropdown:
Screen Shot 2020-05-25 at 2 35 48 PM 1

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+      20  16s ago    multiply                  ⏹ Finished [0]
+      22  15s ago    ArithmeticAddCalculation  ⏵ Waiting        Waiting for transport task: retrieve
+
+    Total results: 6
+
+    Info: last time an entry changed state: 0s ago (at 09:08:59 on 2020-05-13)
+
+We can see that the ``MultiplyAddWorkChain`` is currently waiting for its *child process*, the ``ArithmeticAddCalculation``, to finish.
+Check the process list again for *all* processes (You should know how by now!).
+Now all the processes should be in the ``Finished`` state.
```
csadorf, line 557: Meaning we should wait until they all finish or should they be all finished after checking the processes?

csadorf:
Please also resolve the merge conflicts.

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+An AiiDA database does not only contain the results of your calculations, but also their inputs and each step that was executed to obtain them.
+All of this information is stored in the form of a *directed acyclic graph* (DAG).
+As an example, :numref:`fig_intro_workchain_graph` shows the provenance of the calculations of this tutorial.
+
+.. _fig_intro_workchain_graph:
+.. figure:: include/workchain_graph.png
+    :scale: 30
+    :align: center
+
+    Provenance Graph of a basic AiiDA WorkChain.
```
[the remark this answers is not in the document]
mbercx: Good point! As clicking on the image opens the `.png` file, I think the only way to fix this is by adding the same background color to the images as that of the docs (#FCFCFC). I've fixed the backgrounds with some image magick. :)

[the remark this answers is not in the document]
chrisjsewell: Yeh I was going to say that you could try adding a CSS class, that sets the background color:

```rst
.. figure:: include/workchain_graph.png
    :scale: 30
    :align: center
    :class: bg-white
```

But I guess this won't work when opening it

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+
+    .. code-block:: console
+
+        $ verdi code setup -L add --on-computer --computer=tutor -P arithmetic.add --remote-abs-path=/bin/bash -n
+
+    This command sets up a code with *label* ``add`` on the *computer* ``tutor``, using the *plugin* ``arithmetic.add``.
+
+.. note::
+    A typical real-world example of a computer is a remote supercomputing facility.
+    Codes can be anything from a Python script to powerful *ab initio* codes such as Quantum Espresso or machine learning tools like Tensorflow.
```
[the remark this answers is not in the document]
mbercx: Fair point! I've changed the note to a paragraph.

Regarding the spacing, I now use a pipe `|` to add some spacing, but it seems like it adds a little too much. Is there a better way of doing this?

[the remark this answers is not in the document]
chrisjsewell: when you change from `accordion` to `dropdown` you won't have this problem. You can even go wild and style them with any of the bootstrap classes: 

- https://sphinx-panels.readthedocs.io/en/latest/#dropdown-usage
- https://sphinx-panels.readthedocs.io/en/latest/#card-styling

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,586 @@
...
+
+You can see a single code ``add@tutor``, with PK = 5, in the printed list.
+This code allows us to add two integers together.
+The ``add@tutor`` identifier indicates that the code with label ``add`` is run on the computer with label ``tutor``.
+To see more details about the computer, you can use the following ``verdi`` command:
+
+.. code:: console
+
+    $ verdi computer show tutor
+
```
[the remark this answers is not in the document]
mbercx: Alright, I've added the output. It's probably a good idea to also explain the fact that the `tutor` computer has PK = 1, same as the ``Int`` node that they made at the start of the tutorial. I know that nodes, computers and groups are different types of entities in the database. Is that why they can have the same PK? How would I best explain this in the tutorial?

docs/source/conf.py
```diff
@@ -170,7 +170,7 @@
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
-html_css_files = ['link-box.css']
+html_css_files = ['link-box.css', 'custom.css']
```
chrisjsewell, line 173: If you remove this change, it should resolve your merge conflict. Then move the `custom.css` content to `docs/source/_static/aiida-custom.css` that will now be present

docs/source/tutorial/basic.rst
mbercx: This is because the changes I've made to the `ArithmeticAddCalculation` calculation job aren't implemented in this branch.

I'll be doing a PR for the demo processes into `develop` once we're entirely settled on them, and the tests have been updated. However, apparently implementing a test for a calculation job is not a trivial task, so @sphuber is going to help me with that. ^^

docs/source/tutorial/basic.rst
mbercx: This is what the prompt looks like when you copy-paste a code snippet into the IPython shell. Once the `sphinx-copybutton` extension has been supplied with regexes for the various prompts by @chrisjsewell, I think this is what the users will see if they use the button to copy the code block and paste it into the shell. I'm happy to change it though, if you think this prompt is confusing for the user.

docs/source/tutorial/basic.rst
mbercx: Perhaps a good way to help the user load the output node from the calculation function in case they have forgotten the PK is by having them copy the UUID from the provenance chart they have generated and use e.g.

```
In [3]: builder.x = load_node(uuid='42541d38')
```

That way they can immediately see that you can also use the UUID to load a node. I suppose it might be confusing that you don't use the _whole_ UUID, but we can explain that as well.

Maybe there is a much easier way that I'm missing though.

docs/source/conf.py
mbercx: Great, thanks for figuring that out for me, @chrisjsewell! 😄

docs/source/tutorial/basic.rst
mbercx: I've rephrased this using one of the sentences that @sphuber suggested for the explanation on the difference between the ``run`` and ``submit`` functions for the how-to on running workflows.

docs/source/tutorial/basic.rst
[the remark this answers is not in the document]
csadorf: I actually didn't know that and cannot provide any guidance on this. Maybe @sphuber can help? So far we've been promoting the message that PKs are unique within the profile space.

docs/source/tutorial/basic.rst
csadorf: No, that's ok. As long as it looks exactly like it would if copy & pasted, everything is fine.

docs/source/tutorial/basic.rst
csadorf: No, I think it's worth expounding on the relationship between PKs and UUIDs just a little bit. So I am in favor of this change.

docs/source/tutorial/basic.rst
csadorf: Ok, sounds good. However, we will have to run a final test on a released AiiDA version before we can publish the revised documentation.

docs/source/tutorial/basic.rst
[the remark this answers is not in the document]
sphuber: Yes, the PKs are unique for each profile for each "entity" as long as they are stored in a separate database table, simply because they are the unique identifiers in said table. The following entities have their own table, AuthInfo, Comment, Computer, Group, Node, Link, Log, Node, Setting and User, and so they have their own unique set of pks. Note that sublcasses of each entity are stored in the same table.

Of course we should not explain it in this way or with this amount of details to the user. The most relevant entities for a user would be Computer, Group and Node. So maybe we can add a quick note where we say that PKs are unique to each database, that they are even unique within a database for different entities.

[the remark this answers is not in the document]
csadorf: For those users familiar with databases, it will make immediate sense that the primary key (PK) is unique for a given table and that computers and nodes, etc. are stored in different tables and can therefore have overlapping PKs.

So let's make sure that we mention this such that "expert" users can contextualize this and for all other users we will just say that PKs are unique to each profile for each "class".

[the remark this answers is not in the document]
mbercx: I was thinking of adding a note after they see the output of `verdi computer show tutor`, as they will probably wonder about the PK at that point. Something along the lines of:

```
.. note:: 

    You may have noticed that the PK of the ``tutor`` computer is 1, same as the ``Int`` node we created at the start of this tutorial.
    This is because different entities, such as nodes, computers and groups, are stored in different tables of the database.
    So, the PKs for each entity type are unique for each database, but entities of different types can have the same PK within one database.
```

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,629 @@
...
+    Out[3]: <Int: uuid: eac48d2b-ae20-438b-aeab-2d02b69eb6a8 (pk: 1) value: 2>
+
+As you can see, the data node has now been assigned a *primary key* (**PK**), a number that identifies the node in your database ``(pk: 1)``.
+The PK and UUID both reference the node with the only difference that the PK is unique *for your local database only*, whereas the UUID is a globally unique identifier and can therefore be used between *different* databases.
+Use the PK only if you are working within a single database, i.e. in an interactive session and the UUID in all other cases.
+
+.. important::
+
+    The PK numbers shown throughout this tutorial assume that you start from a completely empty database.
+    It is possible that the nodes' PKs will be different for your database!
```
csadorf, line 95:
```suggestion
    It is possible that the nodes' PKs will be different for your database!

    The UUIDs are generated randomly and are therefore **guaranteed** to be different.
```
I guess users could also be confused as to why the UUIDs are different...

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,629 @@
...
+    -----------  ------------------------------------
+    type         Int
+    pk           1
+    uuid         eac48d2b-ae20-438b-aeab-2d02b69eb6a8
+    label
+    description
+    ctime        2020-05-13 08:58:15.193421+00:00
+    mtime        2020-05-13 08:58:40.976821+00:00
+
+Once again, we can see that the node is of type ``Int``, as PK = 1, and UUID = ``eac48d2b-ae20-438b-aeab-2d02b69eb6a8``.
```
csadorf, line 118:
```suggestion
Once again, we can see that the node is of type ``Int``, has PK = 1, and UUID = ``eac48d2b-ae20-438b-aeab-2d02b69eb6a8``.
```

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,629 @@
...
+
+.. note:: Remember that the PK of the ``CalcJob`` can be different for your database.
+
+CalcJobs
+========
+
+When running calculations that require an external code or run on a remote machine, a simple calculation function is no longer sufficient.
+For this purpose, AiiDA provides the ``CalcJob`` process class.
+
+To run a ``CalcJob``, you need to set up two things: a ``computer`` for the calculation to run on, and the ``code`` that you want the ``CalcJob`` to run.
```
csadorf, line 235:
```suggestion
To run a ``CalcJob``, you need to set up two things: a ``code`` that is going to implement the desired calculation and a ``computer`` for the calculation to run on.
```
Not 100% sure about my suggested formulation, but I found the previous sentence a bit convoluted and thus confusing.

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,629 @@
...
 
+Using the builder, you can easily set up the calculation by directly providing the input arguments.
+Let's use the ``Int`` node that was created by our previous ``calcfunction`` as one of the inputs and a new node as the second input:
+
+.. code-block:: ipython
+
+    In [3]: builder.x = load_node(pk=4)
+       ...: builder.y = Int(5)
+
+In case your nodes' PKs are different, and you don't remember the PK of the output node from the previous calculation, you can also check the provenance graph you have generated and use the start of the UUID of the output node:
```
csadorf, line 330:
```suggestion
In case that your nodes' PKs are different and you don't remember the PK of the output node from the previous calculation, check the provenance graph you generated earlier and use the UUID of the output node instead:
```

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,629 @@
...
+
+.. code-block:: ipython
+
+    Out[6]:
+    {'sum': <Int: uuid: 7d5d781e-8f17-498a-b3d5-dbbd3488b935 (pk: 8) value: 11>,
+    'remote_folder': <RemoteData: uuid: 888d654a-65fb-4da0-b3bc-d63f0374f274 (pk: 9)>,
+    'retrieved': <FolderData: uuid: 4733aa78-2e2f-4aeb-8e09-c5cfb58553db (pk: 10s)>}
+
+Besides the sum of the two ``Int`` nodes, the calculation function also returns two other outputs: one of type ``RemoteData`` and one of type ``FolderData``.
+See the `topics section on calculation jobs <topics:calculations:usage:calcfunctions>` for more details.
```
csadorf, line 363: This link is broken.

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,629 @@
...
+
+    $ verdi daemon start
+
+Now you can either use ``verdi process list`` to follow the execution of the ``CalcJob``, or ``watch`` its progress:
+
+.. code-block:: console
+
+    $ verdi process watch 12
+
+Let's wait for the ``CalcJob` to complete and then use ``verdi process list -a`` to see all processes we have run so far:
```
csadorf, line 476:
```suggestion
Let's wait for the ``CalcJob`` to complete and then use ``verdi process list -a`` to see all processes we have run so far:
```

docs/source/tutorial/basic.rst
```diff
@@ -2,21 +2,629 @@
...
+
+    Besides WorkChain's, workflows can also be implemented as *work functions*.
+    These are ideal for workflows that are not very computationally intensive and can be easily implemented in a Python function.
+
+Let's run the ``WorkChain`` above!
+Start up the ``verdi shell`` and import the ``MultiplyAddWorkChain``:
+
+.. code-block:: ipython
+
+    In [1]: from aiida.workflows.multiplyadd import MultiplyAddWorkChain
```
csadorf, line 556: Shouldn't we use a factory function here instead?

mbercx: Good point! I didn't know how to add an entry point when I first wrote the tutorial. ^^ I've also added a short sentence on the ``WorkflowFactory``. Have a look and see if you think it's clear/sufficient.

PR #4071, opened 2020-05-14
Allow profile selection ipython magic %aiida 

zhubonan:
A small change in the ipython `%aiida` magic to allow selection of the profile. The name of the profile is echoed back so the user knows what they are using.


aiida/tools/ipython/ipython_magics.py
```diff
@@ -81,8 +81,13 @@ def aiida(self, line='', local_ns=None):
         from aiida.cmdline.utils.shell import get_start_namespace
 
         self.is_warning = False
-        load_profile()
-        self.current_state = 'Loaded AiiDA DB environment.'
+        lcontent = line.strip()
+        if lcontent:
```
CasperWA, line 85: Does it make sense to do a try/except instead, if the provided profile name cannot be loaded?

zhubonan: At the moment it will raise an exception, which I think is verbose enough? 

If I do `%aiida bla`, then it gives:

```
    234         if name not in self.profile_names:
--> 235             raise exceptions.ProfileConfigurationError('profile `{}` does not exist'.format(name))
    236
    237     def get_profile(self, name=None):

ProfileConfigurationError: profile `bla` does not exist
```

ltalirz: Seems fine to me

docs/source/install/configuration.rst
```diff
@@ -135,7 +135,8 @@ The current ipython profile folder can be located using::
 
 After this, if you open a Jupyter notebook as explained above and type in a cell::
 
-    %aiida
+    %aiida <profile_name>
 
-followed by ``Shift-Enter``. You should receive the message "Loaded AiiDA DB environment."
+followed by ``Shift-Enter``. The ``<profile_name>`` argument is optional, and the line magic selects the default profile if it is omitted.
```
ltalirz, line 140: to be super-picky - you could put the new sentence on a new line

docs/source/install/configuration.rst
```diff
@@ -135,7 +135,8 @@ The current ipython profile folder can be located using::
...
+You should receive the message "Loaded AiiDA DB environment - profile name: <profile_name>."
```
ltalirz, line 141: I'm not sure this extra sentence is necessary - if people see the line, it's self-explanatory. If they don't... I guess that should never happen?

zhubonan:
@ltalirz I have just tweaked the documentation accordingly.

PR #4072, opened 2020-05-14
Update ZipFolder (in importexport module)

CasperWA:
The outcome of the now closed PR #3537.

It streamlines the `aiida.tools.importexport.dbexport.zip` module, renaming and minimizing the `MyWritingZipFile` to it's necessary parts, and further cements the desire for `ZipFolder`'s methods to be similar to those of `aiida.common.folders.Folder`.
I have chosen not to sub-class `Folder` at this point, simply because I would need to override almost all methods, most of which are not needed. However, this is still an opportunity, if desired.

Concerning handling `bytes` vs. `str` for writing, `zipfile.ZipFile.writestr()` doesn't care, it converts a `str` to `bytes` encoded in UTF-8. Thus, opening `ZipFolder` (keeping the parameters from `Folder`) `encoding` has no effect, nor does adding `b` in the `mode` parameter.
If `encoding` is specifically supplied (or 'b' is found in `mode`) when calling `ZipFolder.open()`, a warning is emitted.

Other things in this PR (in separate commits):
- I add a top-level `AiidaWarning`, similar to `AiidaException`.
  All AiiDA-specific `Warning` classes are now sub-classes of this one.
  I needed a general warning for the `open` method.
- Remove skipping a test in `tests.tools.importexport.test_specific_import`, since the related issue was fixed with a previous PR, but never re-introduced.

### DIFF
--- a/aiida/common/warnings.py
+++ b/aiida/common/warnings.py
@@ -10,7 +10,11 @@
 """Define warnings that can be thrown by AiiDA."""
 
 
-class AiidaDeprecationWarning(Warning):
+class AiidaWarning(Warning):
+    """Generic AiiDA warning."""
+
+
+class AiidaDeprecationWarning(AiidaWarning):
     """
     Class for AiiDA deprecations.
 
@@ -22,13 +26,13 @@ class AiidaDeprecationWarning(Warning):
     """
 
 
-class AiidaEntryPointWarning(Warning):
+class AiidaEntryPointWarning(AiidaWarning):
     """
     Class for warnings concerning AiiDA entry points.
     """
 
 
-class AiidaTestWarning(Warning):
+class AiidaTestWarning(AiidaWarning):
     """
     Class for warnings concerning the AiiDA testing infrastructure.
     """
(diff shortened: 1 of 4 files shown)


sphuber:
Hi @CasperWA , we were discussing the open PRs with @ramirezfranciscof and @chrisjsewell and came across this one. We think that since we are going to overhaul the export code very soon for the new repo, it may not make a lot of sense to touch this code now. Would it be ok to close this PR?

CasperWA:
> Hi @CasperWA , we were discussing the open PRs with @ramirezfranciscof and @chrisjsewell and came across this one. We think that since we are going to overhaul the export code very soon for the new repo, it may not make a lot of sense to touch this code now. Would it be ok to close this PR?

I guess.
The extra point in the OP might still be viable to get in though.

sphuber:
> The extra points in the OP might still be viable to get in though.

That's fine. Make sure to bring them up during the design discussion. Or better, already add them to the google doc with the user stories to explain which parts are important and what we should take into account

CasperWA:
> > The extra points in the OP might still be viable to get in though.
> 
> That's fine. Make sure to bring them up during the design discussion. Or better, already add them to the google doc with the user stories to explain which parts are important and what we should take into account

Sure. But the extra points are not directly related to the export archive situation. One has to do with adding an "umbrella" top-level `Warning` to easier have a catch-'em-all for AiiDA `Warning`s. And the other is simply re-adding a test...

PR #4077, opened 2020-05-17
Allow `verdi calcjob outputcat` work with binary files

### LINKED ISSUE #4086: The `verdi` commands that cat file content break for binary files

zhubonan:
Fixes #4086 

This PR allows `verdi calcjob outputcat` to work with binary files such as those compressed by `gzip`. Previously, only plain text files can be printed because of the missing parameter `mode` in `get_object_content`, which defaults to `r`.

Edit:

Now just open the file as it is in bytes, and use `shutil.copyfileobj` to do the copying. This function copies the content in chunks and avoids high memory usage. The command should now work with both text files and binary data.

### DIFF
--- a/aiida/cmdline/commands/cmd_calcjob.py
+++ b/aiida/cmdline/commands/cmd_calcjob.py
@@ -117,8 +117,9 @@ def calcjob_inputcat(calcjob, path):
 @verdi_calcjob.command('outputcat')
 @arguments.CALCULATION('calcjob', type=CalculationParamType(sub_classes=('aiida.node:process.calculation.calcjob',)))
 @click.argument('path', type=click.STRING, required=False)
+@click.option('--byte', '-b', is_flag=True)
 @decorators.with_dbenv()
-def calcjob_outputcat(calcjob, path):
+def calcjob_outputcat(calcjob, path, byte):
     """
     Show the contents of one of the calcjob retrieved outputs.
 
@@ -127,6 +128,7 @@ def calcjob_outputcat(calcjob, path):
     If PATH is not specified, the default output file path will be used, if defined by the calcjob plugin class.
     Content can only be shown after the daemon has retrieved the remote files.
     """
+    mode = 'rb' if byte else 'r'
     try:
         retrieved = calcjob.outputs.retrieved
     except AttributeError:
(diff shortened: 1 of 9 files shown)


aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -151,12 +154,15 @@ def calcjob_outputcat(calcjob, path):
         )
 
     try:
-        content = retrieved.get_object_content(path)
+        # When we `cat`, it makes sense to directly send the output to stdout as it is
+        with retrieved.open(path, mode='rb') as fhandle:
+            copyfileobj(fhandle, sys.stdout.buffer)
     except (IOError, OSError) as exception:
```
giovannipizzi, line 160: We can just catch OSError in python 3, as IOError is just an alias:
https://docs.python.org/3/library/exceptions.html#IOError

(see a few lines above the link)

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -151,12 +154,15 @@ def calcjob_outputcat(calcjob, path):
...
-        # Incorrect path or file not readable
-        echo.echo_critical('Could not open output path "{}". Exception: {}'.format(path, exception))
-    else:
-        echo.echo(content)
+        # The sepcial case is error No. 32 breakon pipe, which is usually OK.
+        # It can happen if the output is redirected, for example, to `head`.
+        if exception.args[0] != 32:
```
giovannipizzi, line 163: Probably you want to do
```python
import errno

exception.errno != errno.EPIPE
```
See reference [here](https://docs.python.org/3/library/exceptions.html#OSError.errno) for the `errno` of OSError, and you can check the errno module to verify `EPIPE` is 32

aiida/cmdline/commands/cmd_calcjob.py
zhubonan: Yes, I was wondering why... Now py2 is dropped we can just catch OSError

aiida/cmdline/commands/cmd_calcjob.py
zhubonan: good point

giovannipizzi:
BTW, since you are at it, probably you might want to:
- fix also inputcat, and the `cat` command of `verdi node repo`, if it has the same issue?
- add some tests?

aiida/cmdline/commands/cmd_calcjob.py
giovannipizzi: Yes, maybe just fix it here, but I opened #4082 to fix it in the whole codebase

zhubonan:
@giovannipizzi Done! I have also added the tests.

sphuber:
Looking good @zhubonan , thanks a lot. There is also `verdi node repo cat` that suffers from the same problem. Would you be willing to add your fix there as well?

zhubonan:
@sphuber i thinks that has been fixed as well in the latest pushes (please do double check). It is really to merge now.

PR #4083, opened 2020-05-19
Fix bug in `Code.get_full_text_info`

### LINKED ISSUE #4079: code show crashes with 'local' code
I did setup a test script as a 'local' code to be copied on the remote systems. It appears on the list and works, but `verdi code show testscript` fails with trunk aiida_core

at line https://github.com/aiidateam/aiida-core/blob/develop/aiida/orm/nodes/data/code.py#L512
```
  File "lib/python3.8/site-packages/click/core.py", line 956, in invoke
  File "lib/python3.8/site-packages/click/core.py", line 555, in invoke
  File "aiida_core/aiida/cmdline/utils/decorators.py", line 65, in wrapper
  File "aiida_core/aiida/cmdline/commands/cmd_code.py", line 171, in show
    click.echo(tabulate.tabulate(code.get_full_text_info(verbose)))
  File "aiida_core/aiida/orm/nodes/data/code.py", line 512, in get_full_text_info
    if self._repository._get_folder_pathsubfolder.isdir(fname):
AttributeError: 'Repository' object has no attribute '_get_folder_pathsubfolder' 
```

sphuber:
Fixes #4079 

This method was not tested and so it went unnoticed that it was still
using the `_get_folder_pathsubfolder` repository method that has been
removed in `aiida-core==1.0.0`. Direct access to the folder underlying
the node repository is no longer allowed and instead the API should be
used to inspect the available objects, in this case `list_objects`.

### DIFF
--- a/aiida/orm/nodes/data/code.py
+++ b/aiida/orm/nodes/data/code.py
@@ -7,14 +7,10 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-
 import os
 
 from aiida.common.exceptions import ValidationError, EntryPointError, InputValidationError
-
 from .data import Data
-
-DEPRECATION_DOCS_URL = 'http://aiida-core.readthedocs.io/en/latest/concepts/processes.html#the-process-builder'
 
 __all__ = ('Code',)
 
@@ -491,9 +487,12 @@ class Code(Data):
         return builder
 
     def get_full_text_info(self, verbose=False):
+        """Return a list of lists with a human-readable detailed information on this code.
+
+        :return: list of lists where each entry consists of two elements: a key and a value
         """
-        Return a (multiline) string with a human-readable detailed information on this computer
-        """
+        from aiida.orm.utils.repository import FileType
+
         result = []
         result.append(['PK', self.pk])
         result.append(['UUID', self.uuid])
@@ -508,11 +507,11 @@ class Code(Data):
             result.append(['Type', 'local'])
             result.append(['Exec name', self.get_execname()])
             result.append(['List of files/folders:', ''])
-            for fname in self.list_object_names():
-                if self._repository._get_folder_pathsubfolder.isdir(fname):
-                    result.append(['directory', fname])
+            for obj in self.list_objects():
+                if obj.type == FileType.DIRECTORY:
+                    result.append(['directory', obj.name])
                 else:
-                    result.append(['file', fname])
+                    result.append(['file', obj.name])
         else:
             result.append(['Type', 'remote'])
             result.append(['Remote machine', self.get_remote_computer().name])
@@ -536,7 +535,6 @@ class Code(Data):
 
     @classmethod
     def setup(cls, **kwargs):
-        # raise NotImplementedError
         from aiida.cmdline.commands.code import CodeInputValidationClass
         code = CodeInputValidationClass().set_and_validate_from_code(kwargs)
 
--- a/tests/orm/data/test_code.py
+++ b/tests/orm/data/test_code.py
@@ -1,0 +1,46 @@
+"""Tests for the `Code` class."""
+# pylint: disable=redefined-outer-name
+import pytest
+
+from aiida.orm import Code
+
+
+@pytest.fixture
+def create_codes(tmpdir, aiida_localhost):
+    """Create a local and remote `Code` instance."""
+    filename = 'add.sh'
+    filepath = str(tmpdir / filename)  # Cast the filepath to str as Python 3.5 does not support Path objects for `open`
+
+    with open(filepath, 'w'):
+        pass
+
+    code_local = Code(local_executable=filename, files=[filepath]).store()
+    code_remote = Code(remote_computer_exec=(aiida_localhost, '/bin/true')).store()
+
+    return code_local, code_remote
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_get_full_text_info(create_codes):
+    """Test the `Code.get_full_text_info` method."""
+    for code in create_codes:
+        full_text_info = code.get_full_text_info()
+
+        assert isinstance(full_text_info, list)
+        assert ['PK', code.pk] in full_text_info
+        assert ['UUID', code.uuid] in full_text_info
+        assert ['Label', code.label] in full_text_info
+        assert ['Description', code.description] in full_text_info
+
+        if code.is_local():
+            assert ['Type', 'local'] in full_text_info
+            assert ['Exec name', code.get_execname()] in full_text_info
+            assert ['List of files/folders:', ''] in full_text_info
+        else:
+            assert ['Type', 'remote'] in full_text_info
+            assert ['Remote machine', code.computer.name] in full_text_info
+            assert ['Remote absolute path', code.get_remote_exec_path()] in full_text_info
+
+    for code in create_codes:
+        full_text_info = code.get_full_text_info(verbose=True)
+        assert ['Calculations', 0] in full_text_info


aiida/orm/nodes/data/code.py
```diff
@@ -494,6 +494,8 @@ def get_full_text_info(self, verbose=False):
         """
         Return a (multiline) string with a human-readable detailed information on this computer
```
greschd, line 495: The docstring also seems outdated - the function returns a nested list of strings, not a multiline string.

sphuber: Thanks, have updated it.

tests/orm/data/test_code.py
```diff
@@ -0,0 +1,48 @@
...
+    """Create a local and remote `Code` instance."""
+    import os
+
+    filename = 'add.sh'
+    filepath = os.path.join(tmpdir, filename)
+
+    with open(filepath, 'w'):
+        pass
+
+    code_local = Code(local_executable=filename, files=[filepath]).store()
```
ramirezfranciscof, line 19: Nitpick: So, what is the purpose of the parameter `files` when initializing codes? It seems a bit strange to me having to manually add the executable there. If this could have been any other file then maybe it would be less confusing to just make two files for this? (or maybe also use `bin/true` as a local code).

sphuber: This is due to the nature of "local" codes. Remote codes point to a binary that is already preinstalled on the target computer. Local codes, in contrast, are "codes" or "scripts" you have on the machine where AiiDA is running and you want to run on a target machine. This might include multiple files: think some Python package or modules and a single script that actually runs the calculation. When running a "local" code, AiiDA will first upload all these files to the remote machine and then call the script. This is the interface as it was coded at the time and it requires all files to be passed in `files` and it should explicitly tell which one is the "executable"

sphuber:
> (1) Why a list of lists? I mean, I understand we don't want to change the interface now, but is there another reason no to have used a dictionary (since this list of lists seems like a "hacked" dict) or even a list of tupples (since this return being info on the node should perhaps not have mutable "info-pieces")?

This is indeed a remnant of old (sub-ideal) design. In old versions of AiiDA, the ORM classes also included many methods to do a lot of formatting of information such that it is easier, for example, for CLI commands to call them. I think this is not a good design as the formatting of data is not the task of an ORM class but exactly that of a CLI. I have been removing a lot of them, or at the very least, return them in a form where they can still be manipulated somewhat for formatting (i.e. do not return a preformatted string). I totally agree that this current form also makes little sense, so maybe we can deprecate this and remove it for `2.0`. I would maybe do this in another PR with all other comparable commands in the ORM interface that should be removed.

> (2) This was not just the `get_full_text_info` method that was not being tested, but nothing from the code class correct? Then maybe it could be good to open an issue to discuss what other tests should be making for it?
Yes, please go ahead

tests/orm/data/test_code.py
ramirezfranciscof: Ah, ok, so it is not a "local code" in the sense of a code you execute locally, but just in the sense that the executable has to be copied. So can a code be initialized with both `local_executable' AND `remote_computer_exec`? Does it make any sense to specify `files` when using `remote_computer_exec`?

sphuber: > Ah, ok, so it is not a "local code" in the sense of a code you execute locally, but just in the sense that the executable has to be copied.

Exactly, but not just the executable, also any other file that it needs to execute it remotely, think module files etc.

> So can a code be initialized with both `local_executable' AND `remote_computer_exec`? Does it make any sense to specify `files`when using`remote_computer_exec`?

No, these are mutually exclusive.

sphuber:
Ok this is super annoying: the test failed for Py3.5, but I fixed it and pushed the changes. The change shows up here on the PR changes: `    filepath = os.path.join(tmpdir.realpath(), filename)  # Python 3.5 does not support Path objects for `join` yet`
But on GHA, the test still fails and it still seems to be running the old code:
```
==================================== ERRORS ====================================
__________________ ERROR at setup of test_get_full_text_info ___________________

tmpdir = local('/tmp/pytest-of-runner/pytest-0/test_get_full_text_info0')
aiida_localhost = <Computer: localhost-test (localhost-test), pk: 274>

    @pytest.fixture
    def create_codes(tmpdir, aiida_localhost):
        """Create a local and remote `Code` instance."""
        import os
    
        filename = 'add.sh'
>       filepath = os.path.join(tmpdir, filename)
```

@greschd do you maybe have an idea? Is GHA caching the code? I think I have force pushed very often to fix open PRs and it has never been a problem.

greschd:
Wow, that's a curious one - I doubt GHA is _intentionally_ caching the code, that would be a curious choice to lay the least.

But looking at the log, it indeed showed 

```
Successfully installed aiida-core
-e git+https://github.com/aiidateam/aiida-core@7dbd51ac88b9a30bf645b224524fd7c411aaaad0#egg=aiida_core
```

which is (see https://github.com/aiidateam/aiida-core/commit/7dbd51ac88b9a30bf645b224524fd7c411aaaad0) the merge-commit of `37e18bd` (your previous commit) into `develop`.

Maybe there's an internal API somewhere that should tell GHA which commit this PR points to and failed. I've noticed minor glitches on GitHub quite frequently the last few weeks.

How about you make an insignificant edit and force-push again?

tests/orm/data/test_code.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_get_full_text_info(create_codes):
+    """Test the `Code.get_full_text_info` method."""
+    for code in create_codes:
+        full_text_info = code.get_full_text_info()
+
+        assert isinstance(full_text_info, list)
+        assert ['PK', code.pk] in full_text_info
```
greschd, line 32: Since the result is a list of lists with consistent sorting, wouldn't it be simpler to just test

```python
assert full_text_info == [['PK', code.pk], ['UUID', code.uuid], ...]
```

greschd: Ah never mind, I realized it's because of the local / remote distinction below.

greschd:
Seems like it picked the right commit this time around 🎉

sphuber:
> Seems like GHA picked the right commit this time around tada

Except turns out my "fix" was worth exactly nothing :sweat_smile:

tests/orm/data/test_code.py
```diff
@@ -0,0 +1,51 @@
...
+@pytest.fixture
+def create_codes(tmpdir, aiida_localhost):
+    """Create a local and remote `Code` instance."""
+    import os
+
+    filename = 'add.sh'
+
+    # Cast `tmpdir` to str as it is a Path object and Python 3.5 does not support Path objects for `join` yet
+    dirpath = str(tmpdir)
+    filepath = os.path.join(dirpath, filename)
```
greschd, line 17: You could also cast to ``str`` here, and use the glorious ``/`` operator for joining:

```python
filepath = str(dirpath / filename)
```

sphuber: nice! It is clear I have not yet worked with `Path`s at all. Was going to merge because I am sick of waiting for the builds of this PR, but this is too nice to pass on

sphuber: But, then I don't even need the cast right? Or does `open` also not support `Path` in 3.5?

greschd: You still need the cast, `pathlib` support across the board only came in 3.6.

greschd: I guess we're back to waiting for another build 😅

sphuber:
I guess that's a no ...

PR #4087, opened 2020-05-20
Docs: Add HowTo on running workflows

### LINKED ISSUE #3992: Docs: How to run a predefined workflow

mbercx:
Fixes #3992 

Add first draft of the HowTo section on running predefined workflows.

### DIFF
--- a/docs/source/howto/workflows.rst
+++ b/docs/source/howto/workflows.rst
@@ -123,9 +123,7 @@ So, it is advisable to *submit* more complex or longer work chains to the daemon
 Note that when using ``submit`` the work chain is not run in the local interpreter but is sent off to the daemon and you get back control instantly.
 This allows you to submit multiple work chains at the same time and the daemon will start working on them in parallel.
 Once the ``submit`` call returns, you will not get the result as with ``run``, but you will get the **node** that represents the work chain.
-This is because at that point in time, the results do not exist yet, because the work chain has not finished.
-In addition to making it easy to execute multiple work chains in parallel, submitting work chains also allows you to shut down your computer.
-The daemon will always save the progress of the work chains that it is managing such that if it is stopped, it can restart where it left off.
+Submitting a work chain instead of directly running it not only makes it easier to execute multiple work chains in parallel, but also ensures that the progress of a workchain is not lost when you restart your computer.
 
 .. important::
 
--- a/docs/source/howto/workflows.rst
+++ b/docs/source/howto/workflows.rst
@@ -15,22 +15,20 @@ Writing workflows
 
 .. _how-to:workflows:run:
 
-Running a predefined workflow
-=============================
+Launching a predefined workflow
+===============================
 
-The first step to running a predefined workflow is loading the work function or work chain class that defines the workflow you want to run.
+The first step to launching a predefined workflow is loading the work function or work chain class that defines the workflow you want to run.
 The recommended method for loading a workflow is using the ``WorkflowFactory``, for example:
 
 .. code-block:: python
 
     from aiida.plugins import WorkflowFactory
     add_and_multiply = WorkflowFactory('arithmetic.add_multiply')
-    MultiplyAndAddWorkChain = WorkflowFactory('arithmetic.multiply_add')
+    MultiplyAddWorkChain = WorkflowFactory('arithmetic.multiply_add')
 
-This is essentially the same as importing the workflow from its respective module, but using the ``WorkflowFactory`` has the advantage that the entry point will not change when the packages or plugins are reorganised.
+This is essentially the same as importing the workflow from its respective module, but using the ``WorkflowFactory`` has the advantage that the so called *entry point* (e.g. ``'arithmetic.multiply_add'``) will not change when the packages or plugins are reorganised.
 This means your code is less likely to break when updating AiiDA or the plugin that supplies the workflow.
-The input argument of the ``WorkflowFactory`` is the so-called *entry point*.
-For each workflow, the entry point usually starts with the name of the plugin package (here ``arithmetic``), followed by a dot and the string identifying the workflow (here ``multiplyadd``).
 
 The list of installed plugins can be easily accessed via the verdi CLI:
 
@@ -38,7 +36,7 @@ The list of installed plugins can be easily accessed via the verdi CLI:
 
     $ verdi plugin list
 
-To see the list of ``WorkChain`` entry points, simply use:
+To see the list of workflow entry points, simply use:
 
 .. code-block:: console
 
@@ -48,7 +46,7 @@ By further specifying the entry point of the workflow, you can see its descripti
 
 .. code-block:: console
 
-    $ verdi plugin list aiida.workflows arithmetic.multiplyadd
+    $ verdi plugin list aiida.workflows arithmetic.multiply_add
 
 Work functions
 --------------
@@ -57,6 +55,7 @@ Running a work function is as simple as calling a typical Python function: simpl
 
 .. code-block:: python
 
+    from aiida.plugins import WorkflowFactory, DataFactory
     add_and_multiply = WorkflowFactory('arithmetic.add_multiply')
     Int = DataFactory('int')
 
@@ -75,7 +74,7 @@ Just calling the ``add_and_multiply`` function with regular integers will result
 Work chains
 -----------
 
-To launch a work chain, you can either use the ``run`` or ``submit`` functions from the AiiDA engine.
+To launch a work chain, you can either use the ``run`` or ``submit`` functions.
 For either function, you need to provide the class of the work chain as the first argument, followed by the inputs as keyword arguments.
 Using the ``run`` function, or "running", a work chain means it is executed in the same system process as the interpreter in which it is launched:
 
@@ -84,18 +83,18 @@ Using the ``run`` function, or "running", a work chain means it is executed in t
     from aiida.engine import run
     from aiida.plugins import WorkflowFactory, DataFactory
     Int = DataFactory('int')
-    MultiplyAndAddWorkChain = WorkflowFactory('arithmetic.multiply_add')
+    MultiplyAddWorkChain = WorkflowFactory('arithmetic.multiply_add')
 
     add_code = load_code(label='add')
 
-    results = run(MultiplyAndAddWorkChain, x=Int(2), y=Int(3), z=Int(5), code=add_code)
+    results = run(MultiplyAddWorkChain, x=Int(2), y=Int(3), z=Int(5), code=add_code)
 
 Alternatively, you can first construct a dictionary of the inputs, and pass it to the ``run`` function by taking advantage of `Python's automatic keyword expansion <https://docs.python.org/3/tutorial/controlflow.html#unpacking-argument-lists>`_:
 
 .. code-block:: python
 
     inputs = {'x': Int(1), 'y': Int(2), 'z': Int(3), 'code': add_code}
-    results = run(MultiplyAndAddWorkChain, **inputs)
+    results = run(MultiplyAddWorkChain, **inputs)
 
 This is particularly useful in case you have a workflow with a lot of inputs.
 In both cases, running the ``MultiplyAddWorkChain`` workflow returns the **results** of the workflow, i.e. a dictionary of the nodes that are produced as outputs, where the keys of the dictionary correspond to the labels of each respective output.
@@ -112,13 +111,14 @@ So, it is advisable to *submit* more complex or longer work chains to the daemon
 .. code-block:: python
 
     from aiida.engine import submit
+    from aiida.plugins import WorkflowFactory, DataFactory
     Int = DataFactory('int')
-    MultiplyAndAddWorkChain = WorkflowFactory('arithmetic.multiplyadd')
+    MultiplyAddWorkChain = WorkflowFactory('arithmetic.multiply_add')
 
     add_code = load_code(label='add')
     inputs = {'x': Int(1), 'y': Int(2), 'z': Int(3), 'code': add_code}
 
-    workchain_node = submit(MultiplyAndAddWorkChain, **inputs)
+    workchain_node = submit(MultiplyAddWorkChain, **inputs)
 
 Note that when using ``submit`` the work chain is not run in the local interpreter but is sent off to the daemon and you get back control instantly.
 This allows you to submit multiple work chains at the same time and the daemon will start working on them in parallel.
@@ -132,17 +132,19 @@ Submitting a work chain instead of directly running it not only makes it easier 
 If you are unfamiliar with the inputs of a particular ``WorkChain``, a convenient tool for setting up the work chain is the :ref:`process builder<topics:processes:usage:builder>`.
 This can be obtained by using the ``get_builder()`` method, which is implemented for every ``CalcJob`` and ``WorkChain``:
 
-.. code-block:: python
+.. code-block:: ipython
 
-    MultiplyAddWorkChain = WorkflowFactory('arithmetic.multiplyadd')
-    builder = MultiplyAddWorkChain.get_builder()
+    In [1]: from aiida.plugins import WorkflowFactory, DataFactory
+       ...: Int = DataFactory('int')
+       ...: MultiplyAddWorkChain = WorkflowFactory('arithmetic.multiply_add')
+       ...: builder = MultiplyAddWorkChain.get_builder()
 
 To explore the inputs of the work chain, you can use tab autocompletion by typing ``builder.`` and then hitting ``TAB``.
 If you want to get more details on a specific input, you can simply add a ``?`` and press enter:
 
 .. code-block:: ipython
 
-    In [3]: builder.x?
+    In [2]: builder.x?
     Type:        property
     String form: <property object at 0x119ad2dd0>
     Docstring:   {"name": "x", "required": "True", "valid_type": "<class 'aiida.orm.nodes.data.int.Int'>", "non_db": "False"}
@@ -151,20 +153,19 @@ Here you can see that the ``x`` input is required, needs to be of the ``Int`` ty
 
 Using the builder, the inputs of the ``WorkChain`` can be provided one by one:
 
-.. code-block:: python
+.. code-block:: ipython
 
-    builder.code = load_code(label='add')
-    builder.x = Int(2)
-    builder.y = Int(3)
-    builder.z = Int(5)
+    In [3]: builder.code = load_code(label='add')
+       ...: builder.x = Int(2)
+       ...: builder.y = Int(3)
+       ...: builder.z = Int(5)
 
 Once the *required* inputs of the workflow have been provided to the builder, you can either run the work chain or submit it to the daemon:
 
-.. code-block:: python
+.. code-block:: ipython
 
-    from aiida.engine import submit
-
-    workchain_node = submit(builder)
+    In [4]: from aiida.engine import submit
+       ...: workchain_node = submit(builder)
 
 .. note::
 
--- a/docs/source/howto/workflows.rst
+++ b/docs/source/howto/workflows.rst
@@ -18,7 +18,159 @@ Writing workflows
 Running a predefined workflow
 =============================
 
-`#3992`_
+The first step to running a predefined workflow is loading the work function or work chain class that defines the workflow you want to run.
+The recommended method for loading a workflow is using the ``WorkflowFactory``, for example:
+
+.. code-block:: python
+
+    from aiida.plugins import WorkflowFactory
+    add_and_multiply = WorkflowFactory('arithmetic.add_multiply')
+    MultiplyAndAddWorkChain = WorkflowFactory('arithmetic.multiply_add')
+
+This is essentially the same as importing the workflow from its respective module, but using the ``WorkflowFactory`` has the advantage that the entry point will not change when the packages or plugins are reorganised.
+This means your code is less likely to break when updating AiiDA or the plugin that supplies the workflow.
+The input argument of the ``WorkflowFactory`` is the so-called *entry point*.
+For each workflow, the entry point usually starts with the name of the plugin package (here ``arithmetic``), followed by a dot and the string identifying the workflow (here ``multiplyadd``).
+
+The list of installed plugins can be easily accessed via the verdi CLI:
+
+.. code-block:: console
+
+    $ verdi plugin list
+
+To see the list of ``WorkChain`` entry points, simply use:
+
+.. code-block:: console
+
+    $ verdi plugin list aiida.workflows
+
+By further specifying the entry point of the workflow, you can see its description, inputs, outputs and exit codes:
+
+.. code-block:: console
+
+    $ verdi plugin list aiida.workflows arithmetic.multiplyadd
+
+Work functions
+--------------
+
+Running a work function is as simple as calling a typical Python function: simply call it with the required input arguments:
+
+.. code-block:: python
+
+    add_and_multiply = WorkflowFactory('arithmetic.add_multiply')
+    Int = DataFactory('int')
+
+    result = add_and_multiply(Int(2), Int(3), Int(5))
+
+Here, the ``add_and_multiply`` work function returns the output ``Int`` node and we assign it to the variable ``result``.
+Note that the input arguments of a work function must be an instance of ``Data`` node, or any of its subclasses.
+Just calling the ``add_and_multiply`` function with regular integers will result in a ``ValueError``, as these cannot be stored in the provenance graph.
+
+.. note::
+
+    Although the example above shows the most straightforward way to run the ``add_and_multiply`` work function, there are several other ways of running processes that can return more than just the result.
+    For example, the ``run_get_node`` function from the AiiDA engine returns both the result of the workflow and the work function node.
+    See the :ref:`corresponding topics section for more details <topics:processes:usage:launching>`.
+
+Work chains
+-----------
+
+To launch a work chain, you can either use the ``run`` or ``submit`` functions from the AiiDA engine.
+For either function, you need to provide the class of the work chain as the first argument, followed by the inputs as keyword arguments.
+Using the ``run`` function, or "running", a work chain means it is executed in the same system process as the interpreter in which it is launched:
+
+.. code-block:: python
+
+    from aiida.engine import run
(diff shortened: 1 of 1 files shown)


docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,159 @@ Running a predefined workflow
...
+    results = run(MultiplyAndAddWorkChain, **inputs)
+
+This is particularly useful in case you have a workflow with a lot of inputs.
+In both cases, running the ``MultiplyAddWorkChain`` workflow returns the **results** of the workflow, i.e. a dictionary of the nodes that are produced as outputs, where the keys of the dictionary correspond to the labels of each respective output.
+
+.. note::
+
+    Similar to other processes, there are multiple functions for running a work chain. See the section on :ref:`launching processes for more details<topics:processes:usage:launching>`.
+
+Since *running* a workflow will block the interpreter, it is advisable to submit more complex work chains to the daemon:
```
sphuber, line 117: I would introduce this slightly different and first have the user realize the "problem" of running themselves. Only then we give the solution. So something like:

> You may have noticed that when you "run" a workflow, the interpreter is blocked, and you have to wait until the workflow is finished before you get back control. This might not be a problem for these trivial examples that only take a small amount of time to complete, however, when you start running longer workflows this will become problematic as you won't be able to turn your computer or even your terminal off until the workflow has fully terminated. Note that with this approach it is also difficult to run multiple workflows in parallel. The solution to these problems, is to, instead of running, submit the workflows to the daemon.

Then you show the code snippet and after you can continue with

> Note that when using `submit` the workflow is not run in the local interpreter but is sent off to the daemon and you get back control instantly. This allows you to submit multiple workflows at the same time and the daemon will start working on them in parallel. Once the `submit` call returns, you will not get the result as with `run`, but you will get the node that represents the workflow. This is because at that point in time, the results do not exist yet, because the workflow has not finished yet. In addition to allowing you to easily run multiple workflows in parallel, submitting workflows also allows you to shutdown your computer. The daemon will always save the progress of the workflows that is managing such that if it is stopped, it can restart where it left off


I see now that the text is quite long, but I think this is crucial for users to understand. See how you can fit it in and maybe shorten it a bit. Like I mentioned before that we will have the same problem in the calculation how-to, we might actually want to take part of this description and add it to a topic section and just reference that here, so we can reduce the amount of text here a bit. I would add it maybe to the same section that you [already referenced about the different launchers](https://aiida.readthedocs.io/projects/aiida-core/en/docs-revamp/topics/processes/usage.html#process-launch)

This reminds, we should also not forget that work functions *cannot* be submitted, but only work chains

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,159 @@ Running a predefined workflow
...
+
+However, a more robust way of loading a ``WorkChain`` is using the ``WorkflowFactory``, for example:
+
+.. code-block:: python
+
+    MultiplyAndAddWorkChain = WorkflowFactory('arithmetic.multiplyadd')
+
+The input argument of the ``WorkflowFactory`` is a string called the *entry point*.
+For each work chain, the entry point usually starts with with the name of the plugin package (``arithmetic``), followed by a dot and the string identifying the work chain (``multiplyadd``).
+Using the ``WorkflowFactory`` has the advantage that the entry point will not change when the packages or plugins are reorganised, and hence your code is less likely to break when updating AiiDA or the plugin that supplies the work chain.
```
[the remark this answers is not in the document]
csadorf: Naive question: why is using a "factory" liked this better than ensuring that our import paths remain stable.

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,159 @@ Running a predefined workflow
...
+    Although the example above shows the most straightforward way to run the ``add_and_multiply`` work function, there are several other ways of running processes that can return more than just the result.
+    For example, the ``run_get_node`` function from the AiiDA engine returns both the result of the workflow and the work function node.
+    See the :ref:`corresponding topics section for more details <topics:processes:usage:launching>`.
+
+Work chains
+-----------
+
+To launch a work chain, you can either use the ``run`` or ``submit`` functions from the AiiDA engine.
+For either function, you need to provide the class of the work chain as the first argument, followed by the inputs as keyword arguments.
+Running a work chain means it is executed in the same interpreter in which it is launched:
```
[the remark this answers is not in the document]
csadorf: I'd even suggest to be more explicit: "in the same process as the interpreter in which it is launched".

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,159 @@ Running a predefined workflow
...
+
+    from aiida.engine import run
+    Int = DataFactory('int')
+    MultiplyAndAddWorkChain = WorkflowFactory('arithmetic.multiplyadd')
+
+    add_code = load_code(label='add')
+
+    results = run(MultiplyAndAddWorkChain, x=Int(2), y=Int(3), z=Int(5), code=add_code)
+
+Alternatively, you can first construct a dictionary of the inputs, and pass it to the ``run`` function as an expanded dictionary:
```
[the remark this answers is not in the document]
csadorf: Yes, something akin to "take advantage of Python's automatic keyword expansion" or so. With a link to the Python docs.

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,159 @@ Running a predefined workflow
...
+
+    inputs = {'x': Int(1), 'y': Int(2), 'z': Int(3), 'code': add_code}
+    results = run(MultiplyAndAddWorkChain, **inputs)
+
+This is particularly useful in case you have a workflow with a lot of inputs.
+In both cases, running the ``MultiplyAddWorkChain`` workflow returns the **results** of the workflow, i.e. a dictionary of the nodes that are produced as outputs, where the keys of the dictionary correspond to the labels of each respective output.
+
+.. note::
+
+    Similar to other processes, there are multiple functions for running a work chain. See the section on :ref:`launching processes for more details<topics:processes:usage:launching>`.
```
csadorf, line 115: new line missing

docs/source/howto/workflows.rst
csadorf: Be careful with the language here:

> You may have noticed that when you "run" a workflow, the interpreter is blocked

This sounds like a tutorial, however this is a how-to guide, not a tutorial. Users are not expected to actually execute this code to understand what is going on.

sphuber: I think I agree. Like I already alluded in the postscript, I think this should probably be explained once, well and in detail, probably in the "Topics - Processes" section. In the how-to, you can then just say that instead of running , we can also submit and then reference the section if you want to know when to use which and what are the pro's and con's of each. Would that work?

docs/source/howto/workflows.rst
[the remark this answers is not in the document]
sphuber: Because the same content also applies to other plugins and not just `aiida-core`, so it is a good paradigm to get used to. In addition, the entry point names are unique identifiers that are supported in other parts of the API. For example, in CLI commands such as `verdi plugin list` and many other. Also in the Python API there are places where these strings come back. This is not the case with the module path. So I think for users in general it is very useful to start thinking about the entry point names, instead of the full module path of the resource

docs/source/howto/workflows.rst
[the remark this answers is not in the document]
sphuber: Fine with being more explicit. We should just be careful with the ambiguity of the word "process" in these contexts. You mean the system process in this case, but in a context of AiiDA it may easily be misunderstood for the `Process` class.

docs/source/howto/workflows.rst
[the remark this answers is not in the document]
csadorf: @sphuber Your explanation had the missing piece about the entry points from plugins so I think we should better motivate that in the docs.

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,159 @@ Running a predefined workflow
 
 `#3992`_
 
+The first step to running a predefined workflow is loading the work function or work chain class that defines the workflow you want to run.
+This can of course be achieved by importing the function or class from its respective module:
+
+.. code-block:: python
+
+    from aiida.workflows.plugins.arithmetic.addmultiply import add_and_multiply              # Import the addmultiply work function
+    from aiida.workflows.plugins.arithmetic.multiplyadd import MultiplyAndAddWorkChain       # Import the MultiplyAndAddWorkChain WorkChain
```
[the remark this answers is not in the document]
mbercx: As we discussed earlier, the motivation for doing the addition after the multiplication in the `MultiplyAndAddWorkChain` was twofold:

* The tutorial first shows the calculation function and then the `ArithmeticAddCalculation` calculation job, using the output of the calculation function. In order to be consistent when explaining the work chain, I used the same order. Of course, we can also write a `ArithmeticMultiplyCalculation` calculation job to solve this.
* `AddAndMultiplyWorkChain` is already used in other parts of the documentation (e.g. [here](https://aiida-core.readthedocs.io/en/latest/working/workflows.html#work-chains)), and is defined as a work chain that only calls calculation functions. 

Perhaps it would be good if I opened an issue to discuss the *arithmetic* processes, where we can settle on which set to use and that I can refer to when I do the PR into `develop`?

On the second point: agreed!  I'll refactor the code and leave mirrors with a deprecation warning.

csadorf:
What is the status here?

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,159 @@ Running a predefined workflow
...
+
+    $ verdi plugin list
+
+To see the list of ``WorkChain`` entry points, simply use:
+
+.. code-block:: console
+
+    $ verdi plugin list aiida.workflows
+
+By further specifying the entry point of the work chain, you can see its description, inputs, outputs and exit codes:
```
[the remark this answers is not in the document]
mbercx: Listing the entry point works fine:

```
$ verdi plugin list aiida.workflows 
Registered entry points for aiida.workflows:
* arithmetic.add_multiply
* arithmetic.multiply_add
```

But trying to get more details on the work function fails:

```
$ verdi plugin list aiida.workflows arithmetic.add_multiply
Error: No description available for arithmetic.add_multiply
```

Should I add a note on this in the how-to, or open an issue?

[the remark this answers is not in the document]
mbercx: As a sidenote: Having two entry points, one for the `add_and_multiply` work function and one for the `MultiplyAddWorkChain` is perhaps another good reason to not use `AddMultiplyWorkChain`, to clearly differentiate the two.

docs/source/howto/workflows.rst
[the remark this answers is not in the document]
mbercx: Maybe explicitly say it is a system process?

"Using the `run` function, or "running", a work chain means it is executed in the same system process as the interpreter in which it is launched"

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,159 @@ Running a predefined workflow
...
+    result = add_and_multiply(Int(2), Int(3), Int(5))
+
+Here, the ``add_and_multiply`` work function returns the output ``Int`` node and stores it in the variable ``result``.
+Note that the input arguments of a work function must be a subclass of the ``Data`` class.
+Just calling the ``add_and_multiply`` function with regular integers will result in a ``ValueError``, as these cannot be stored in the provenance.
+
+.. note::
+
+    Although the example above shows the most straightforward way to run the ``add_and_multiply`` work function, there are several other ways of running processes that can return more than just the result.
+    For example, the ``run_get_node`` function from the AiiDA engine returns both the result of the workflow and the work function node.
```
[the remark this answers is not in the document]
mbercx: I agree it's more practical, but it can be a bit confusing to the novice user. I for one wasn't familiar with using function attributes. So, here `get_node` is a function that is an `attribute` of the `run` function?

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,159 @@ Running a predefined workflow
...
+    from aiida.workflows.plugins.arithmetic.multiplyadd import MultiplyAndAddWorkChain       # Import the MultiplyAndAddWorkChain WorkChain
+
+However, a more robust way of loading a ``WorkChain`` is using the ``WorkflowFactory``, for example:
+
+.. code-block:: python
+
+    MultiplyAndAddWorkChain = WorkflowFactory('arithmetic.multiplyadd')
+
+The input argument of the ``WorkflowFactory`` is a string called the *entry point*.
+For each work chain, the entry point usually starts with with the name of the plugin package (``arithmetic``), followed by a dot and the string identifying the work chain (``multiplyadd``).
```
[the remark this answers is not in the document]
mbercx: Agreed! As may have been clear from the rest of the how-to, I simply wasn't aware that you could also add an entry point for a work function and load it with the `WorkFlowFactory`.

docs/source/howto/workflows.rst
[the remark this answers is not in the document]
mbercx: > @sphuber Your explanation had the missing piece about the entry points from plugins so I think we should better motivate that in the docs.

Should I add another sentence or two in the note at the end of this section, e.g.:

"Using the ``WorkflowFactory`` is a good paradigm to get used to as the entry point names are unique identifiers that are supported in many parts of AiiDA's API, such as the `verdi plugin list` command shown above."

docs/source/howto/workflows.rst
mbercx: I've shortened the pre-snippet paragraph and merged it with the text that was there. The post-snippet paragraph isn't that much longer than the text that was there, and expresses the important points more clearly, so I've largely replaced the text with your suggestion.

I think it's fine to leave it in the how-to, but in case you think it tries to explain too much, I can still shorten it.

mbercx:
@sphuber and @csadorf: many thanks for the suggestions/corrections! I've tried to implement the discussions as best as possible, but have left some conversations as unresolved as they are not fully settled. Maybe quickly comment in case you still would like to see changes on these issues so I can correct them before a second round of review.

docs/source/howto/workflows.rst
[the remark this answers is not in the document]
sphuber: > Listing the entry point works fine:
> 
> ```
> $ verdi plugin list aiida.workflows 
> Registered entry points for aiida.workflows:
> * arithmetic.add_multiply
> * arithmetic.multiply_add
> ```
> 
> But trying to get more details on the work function fails:
> 
> ```
> $ verdi plugin list aiida.workflows arithmetic.add_multiply
> Error: No description available for arithmetic.add_multiply
> ```
> 
> Should I add a note on this in the how-to, or open an issue?

I will open an issue and fix it. It should simply show the docstring which it currently doesn't.

docs/source/howto/workflows.rst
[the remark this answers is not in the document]
sphuber: Exactly. But since they are just copy pasting, they don't really have to understand, it no? @csadorf what do you think?

docs/source/howto/workflows.rst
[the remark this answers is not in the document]
sphuber: As discussed, `MultiplyAddWorkChain` and `add_multiply` for the workfunction with `multipy_add` and `add_multiply` as the two entry points names is good for me :+1:

docs/source/howto/workflows.rst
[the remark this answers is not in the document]
sphuber: See #4115

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,172 @@ Running a predefined workflow
 
 `#3992`_
 
+The first step to running a predefined workflow is loading the work function or work chain class that defines the workflow you want to run.
+This can of course be achieved by importing the function or class from its respective module:
```
csadorf, line 24:
```suggestion
This can be achieved by importing the function or class from its respective module:
```

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,172 @@ Running a predefined workflow
...
+
+.. code-block:: python
+
+    from aiida.workflows.arithmetic.add_multiply import add_and_multiply              # Import the add_and_multiply work function
+    from aiida.workflows.arithmetic.multiply_add import MultiplyAddWorkChain       # Import the MultiplyAddWorkChain WorkChain
+
+However, a more robust way of loading a workflow is using the ``WorkflowFactory``, for example:
```
csadorf, line 31: Reading this now, I think it would be better to just show what is best practice (factory) and mention that this is basically equivalent to an import, but more robust because of the plugin/entrypoint system. Starting the how-to with *what not to do* is more confusing than helpful and also runs counter the idea that a how-to should show the solution to the problem with only minimal context and additional explanations.

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,172 @@ Running a predefined workflow
...
+
+.. code-block:: python
+
+    from aiida.plugins import WorkflowFactory
+    add_and_multiply = WorkflowFactory('arithmetic.add_multiply')
+    MultiplyAndAddWorkChain = WorkflowFactory('arithmetic.multiply_add')
+
+The input argument of the ``WorkflowFactory`` is a string called the *entry point*.
+For each workflow, the entry point usually starts with with the name of the plugin package (``arithmetic``), followed by a dot and the string identifying the workflow (``multiplyadd``).
```
csadorf, line 40:
```suggestion
For each workflow, the entry point usually starts with the name of the plugin package (here ``arithmetic``), followed by a dot and the string identifying the workflow (here ``multiplyadd``).
```

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,172 @@ Running a predefined workflow
...
+
+However, a more robust way of loading a workflow is using the ``WorkflowFactory``, for example:
+
+.. code-block:: python
+
+    from aiida.plugins import WorkflowFactory
+    add_and_multiply = WorkflowFactory('arithmetic.add_multiply')
+    MultiplyAndAddWorkChain = WorkflowFactory('arithmetic.multiply_add')
+
+The input argument of the ``WorkflowFactory`` is a string called the *entry point*.
```
csadorf, line 39:
```suggestion
The input argument of the ``WorkflowFactory`` is the so called *entry point*.
```

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,172 @@ Running a predefined workflow
...
+
+By further specifying the entry point of the workflow, you can see its description, inputs, outputs and exit codes:
+
+.. code-block:: console
+
+    $ verdi plugin list aiida.workflows arithmetic.multiplyadd
+
+.. note::
+
+    Using the ``WorkflowFactory`` is a good paradigm to get used to as the entry point names are unique identifiers that are supported in many parts of AiiDA's API, such as the ``verdi plugin list`` command shown above.
```
csadorf, line 63: Again, the documented way to "import" workflows is using the ``WorkflowFactory``. As long as we properly motivate it once, I don't think there is much more *convincing* that we need to do. If users decide to use direct imports instead, they are employing a non-documented way of using AiiDA and may need to live with potential downsides of that approach. That means we should completely remove this sentence.

docs/source/howto/workflows.rst
```diff
@@ -20,6 +20,172 @@ Running a predefined workflow
...
+.. code-block:: python
+
+    add_and_multiply = WorkflowFactory('arithmetic.add_multiply')
+    Int = DataFactory('int')
+
+    result = add_and_multiply(Int(2), Int(3), Int(5))
+
+Here, the ``add_and_multiply`` work function returns the output ``Int`` node and we assign it to the variable ``result``.
+Note that the input arguments of a work function must be an instance of ``Data`` node, or any of its subclasses.
+Just calling the ``add_and_multiply`` function with regular integers will result in a ``ValueError``, as these cannot be stored in the provenance.
```
csadorf, line 81:
```suggestion
Just calling the ``add_and_multiply`` function with regular integers will result in a ``ValueError``, as these cannot be stored in the provenance graph.
```

Correct me if I'm wrong, but the term [*provenance*](https://en.wikipedia.org/wiki/Provenance) for me basically translates to "origin" or "history". I don't think that you can "store anything within the provenance".

docs/source/howto/workflows.rst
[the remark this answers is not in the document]
csadorf: I agree with @sphuber that we should show best practice, but I also agree that with @mbercx that function attributes, while legal, are a bit of a more obscure use of the Python language and typically avoided for public API, because of the high potential for confusion. Especially for a reader of the coder.

I think that clarity trumps convenience in this case, and I would prefer  to show `run_get_node` instead.

docs/source/howto/workflows.rst
mbercx: I agree, starting with what not to do is probably not a great idea. I've removed the examples that imports the workflows, and reorganised the text.

mbercx:
Thanks again for the suggestions, @csadorf! This is ready for another round of review.

docs/source/howto/workflows.rst
```diff
@@ -18,7 +18,159 @@ Writing workflows
...
+    inputs = {'x': Int(1), 'y': Int(2), 'z': Int(3), 'code': add_code}
+
+    workchain_node = submit(MultiplyAndAddWorkChain, **inputs)
+
+Note that when using ``submit`` the work chain is not run in the local interpreter but is sent off to the daemon and you get back control instantly.
+This allows you to submit multiple work chains at the same time and the daemon will start working on them in parallel.
+Once the ``submit`` call returns, you will not get the result as with ``run``, but you will get the **node** that represents the work chain.
+This is because at that point in time, the results do not exist yet, because the work chain has not finished.
+In addition to making it easy to execute multiple work chains in parallel, submitting work chains also allows you to shut down your computer.
+The daemon will always save the progress of the work chains that it is managing such that if it is stopped, it can restart where it left off.
```
csadorf, line 128:
```suggestion
Submitting a work chain instead of directly running it not only makes it easier to execute multiple work chains in parallel, but also ensures that the progress of a workchain is not lost when you restart your computer.
```

docs/source/howto/workflows.rst
```diff
@@ -18,7 +18,157 @@ Writing workflows
 Running a predefined workflow
```
sphuber, line 18: Sorry to come back on this now, but I was just wondering whether we should renaming this to "Launching a predefined workflow". I have been trying to strictly use "launch" when talking in the generic sense, since "run" and "submit" will quickly get specific meaning when talking in the context of AiiDA, given that the two launchers  `run` and `submit` fundamtenally change the behavior.

I think "Launching" is a common enough term for novice users to understand straight away what it means (instead of running) and it will already keep consistency with the language that will follow

docs/source/howto/workflows.rst
```diff
@@ -18,7 +18,157 @@ Writing workflows
...
 =============================
 
-`#3992`_
+The first step to running a predefined workflow is loading the work function or work chain class that defines the workflow you want to run.
```
sphuber, line 21:
```suggestion
The first step to launching a predefined workflow is loading the work function or work chain class that defines the workflow you want to run.
```

docs/source/howto/workflows.rst
```diff
@@ -18,7 +18,157 @@ Writing workflows
...
+.. code-block:: python
+
+    from aiida.plugins import WorkflowFactory
+    add_and_multiply = WorkflowFactory('arithmetic.add_multiply')
+    MultiplyAndAddWorkChain = WorkflowFactory('arithmetic.multiply_add')
+
+This is essentially the same as importing the workflow from its respective module, but using the ``WorkflowFactory`` has the advantage that the entry point will not change when the packages or plugins are reorganised.
+This means your code is less likely to break when updating AiiDA or the plugin that supplies the workflow.
+The input argument of the ``WorkflowFactory`` is the so-called *entry point*.
+For each workflow, the entry point usually starts with the name of the plugin package (here ``arithmetic``), followed by a dot and the string identifying the workflow (here ``multiplyadd``).
```
sphuber, line 33: Not something to address now, but a point for the future: this is quite a lot of info, that is maybe necessary, but only tangentially related. In the future, we might want to reduce this a bit perhaps and reference a topic section on entry points and the plugin system.

docs/source/howto/workflows.rst
```diff
@@ -18,7 +18,157 @@ Writing workflows
...
+
+The list of installed plugins can be easily accessed via the verdi CLI:
+
+.. code-block:: console
+
+    $ verdi plugin list
+
+To see the list of ``WorkChain`` entry points, simply use:
```
sphuber, line 41:
```suggestion
To see the list of workflow entry points, simply use:
```

docs/source/howto/workflows.rst
```diff
@@ -18,7 +18,157 @@ Writing workflows
...
+.. note::
+
+    Although the example above shows the most straightforward way to run the ``add_and_multiply`` work function, there are several other ways of running processes that can return more than just the result.
+    For example, the ``run_get_node`` function from the AiiDA engine returns both the result of the workflow and the work function node.
+    See the :ref:`corresponding topics section for more details <topics:processes:usage:launching>`.
+
+Work chains
+-----------
+
+To launch a work chain, you can either use the ``run`` or ``submit`` functions from the AiiDA engine.
```
sphuber, line 78:
```suggestion
To launch a work chain, you can either use the ``run`` or ``submit`` functions.
```

docs/source/howto/workflows.rst
mbercx: Good point! Everyone should understand what "launching" means, and this will make the terminology in the docs more consistent. Do you agree, @csadorf?

docs/source/howto/workflows.rst
mbercx: I also just noticed that the *entry point* is used before it is defined due to a restructuring of the text. Let me fix that and maybe slim down this paragraph a bit.

csadorf:

docs/source/howto/workflows.rst
csadorf: I think that's a good idea! 👍

mbercx:
Thanks for the suggestions @sphuber and @csadorf! I've made the requested changes, and adapted the arithmetic examples to the latest version. I've also switched the code-blocks of the process builder piece to `ipython`, as these show outputs and are supposed to be executed in the shell, e.g. when exploring an unknown workflow.

PR #4092, opened 2020-05-20
Ensure `verdi group show --limit` respects limit even in raw mode

### LINKED ISSUE #4091: `verdi group show` does not respect the limit if `raw=False`

sphuber:
Fixes #4091 

The code is affected by the `--limit` flag both where the `--raw` flag
is enabled and if it is disabled. However, in the disabled case the
limit was not respected. This was not detected because the unit test
only tested the code path when `--raw` is enabled.

### DIFF
--- a/tests/cmdline/commands/test_group.py
+++ b/tests/cmdline/commands/test_group.py
@@ -192,6 +192,14 @@ class TestVerdiGroup(AiidaTestCase):
         self.assertEqual(len(result.output.strip().split('\n')), 1)
         self.assertTrue(str(nodes[0].pk) in result.output or str(nodes[1].pk) in result.output)
 
+        # Repeat test with `limit=1` but without the `--raw` flag as it has a different code path that is affected
+        result = self.cli_runner.invoke(cmd_group.group_show, [label, '--limit', '1'])
+        self.assertClickResultNoException(result)
+
+        # Check that one, and only one pk appears in the output
+        self.assertTrue(str(nodes[0].pk) in result.output or str(nodes[1].pk) in result.output)
+        self.assertTrue(not (str(nodes[0].pk) in result.output and str(nodes[1].pk) in result.output))
+
     def test_description(self):
         """Test `verdi group description` command."""
         description = 'It is a new description'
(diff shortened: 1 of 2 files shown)


tests/cmdline/commands/test_group.py
```diff
@@ -192,6 +192,14 @@ def test_show_limit(self):
...
         self.assertTrue(str(nodes[0].pk) in result.output or str(nodes[1].pk) in result.output)
 
+        # Repeat test with `limit=1` but without the `--raw` flag as it has a different code path that is affected
+        result = self.cli_runner.invoke(cmd_group.group_show, [label, '--limit', '1'])
+        self.assertClickResultNoException(result)
+
+        # Check that one, and only one pk appears in the output
+        self.assertTrue(str(nodes[0].pk) in result.output or str(nodes[1].pk) in result.output)
+        self.assertTrue(not (str(nodes[0].pk) in result.output and str(nodes[1].pk) in result.output))
+
```
ramirezfranciscof, line 202: Perhaps the second assert could be replaced by checking the length of the output to be 1? Feels a bit more straighforward and easy to read than the negative of the intersection (although less logically-fancy). Also it feels a bit more resistant to the DB leackeages that sometimes ocurr with the tests. Idk, just an idea, up to you.

sphuber: I would agree that that would be easier, except, unfortunately, in this case the output is not a list. It is a single string. That is the reason I used `--raw` before in the original test, so I would only get the nodes as rows, so I could exactly use the count. But that caused the hiding of the bug that this very PR resolves. So I cannot use `--raw` and so I have to "parse" the full string, or at best the lines of the result. Since I don't want to make the parsing dependent on the number of header and footer lines, I solved it like this.

ramirezfranciscof: Uhm, I see, I didn't notice the string casting. Question: although I guess very unlikely, but this way of checking with strings could raise a false fail if the nodes were assigned, say, pk=1 and pk=12, because both "12" and "1" are in "12", correct?

sphuber: Well they should get adjacent identifiers, but that is besides the point. You are correct that this test is vulnerable to false positives, but also to false negatives.

PR #4093, opened 2020-05-20
Docs: How to interface external codes

### LINKED ISSUE #3986: Docs: How to interface external codes

ramirezfranciscof:
Fixes #3986

This how-to explains how to implement the most basics features of the `CalcJob` and `Parser` classes in order to be able to interface an external code. The example is taken from a slightly simplified version of the `Arithmetic.add`, already introduced in the tutorial. The sections is aimed at providing a fully working example with the minimum features required and without going deep into the explanations. However, the task of interfacing a code is still quite daunting and thus this will probably be the longest of the how-tos. If you still thing it is too long, advice on what can be cut out is welcome.

The most relevant features shown here include:

* Defining new inputs/outputs/exit codes.
* Modifying parameters of existing inputs of the base `CalcJon`.
* Accessing the inputs provided by the user.
* Writing files in the scratch folder.
* Setting the minimal `CodeInfo`(code id, command line parameters, standard output)
* Setting the minimal `CalcInfo` (code info and retrieve list)
* Catching errors during parsing.
* Reading the files in the retrieved folder.
* Creating the output nodes.

### DIFF
--- a/docs/source/howto/codes.rst
+++ b/docs/source/howto/codes.rst
@@ -32,7 +32,7 @@ Some general guidelines to keep in mind are:
      #. store in the file repository for safe-keeping (:py:class:`~aiida.orm.nodes.data.singlefile.SinglefileData`, ...)
      #. leave on the computer where the calculation ran (:py:class:`~aiida.orm.nodes.data.remote.RemoteData`, ...)
 
-As the underlying example for these we will use the process of using the `bash` interpreter (``/bin/bash``) to add up two numbers by running the command: ``echo $(( numx + numy ))``.
+To demonstrate how to create a plugin for an external code, we will use the trivial example of using the `bash` shell (``/bin/bash``) to sum two numbers by running the command: ``echo $(( numx + numy ))``.
 Here, the `bash` binary will be effectively acting as our |Code| executable, the input (``aiida.in``) will then be a file containing the command with the numbers provided by the user replaced, and the output (``aiida.out``) will be caught through the standard output.
 The final recipe to run this code will then be:
 
@@ -45,7 +45,7 @@ The final recipe to run this code will then be:
 Interfacing external codes
 ==========================
 
-The way you provide AiiDA with the set of instructions required to run a code is by extending the |CalcJob| class, which has the following two key methods:
+To provide AiiDA with the set of instructions, required to run a code, one should implement the |CalcJob| class, which has the following two key methods:
 
 .. code-block:: python
 
@@ -64,13 +64,13 @@ The way you provide AiiDA with the set of instructions required to run a code is
             # Preparation of input files and instructions for engine
             return calcinfo
 
-We will now indicate how to deal with each of them separatedly.
+We will now show how each of these can be implemented.
 
 Defining the specifications
 ---------------------------
 
 As the comment in the code above indicates, the first method (|define|) is where one specifies the different inputs that the user of the |CalcJob| will have to provide in order to run the code, as well as the outputs that will be produced (exit codes are discussed in the respective section).
-This is done through the |spec| object, which, as can be seen, is passed as an input to the method.
+This is done through the |spec| object, which, as can be seen, is passed as an argument to the method.
 For the code that adds up two numbers, we will need to define those numbers as inputs (lets call them ``x`` and ``y`` to label them) and the result as an output (``sum``).
 
 .. code-block:: python
@@ -88,19 +88,19 @@ For the code that adds up two numbers, we will need to define those numbers as i
         spec.inputs['metadata']['options']['input_filename'].default = 'aiida.in'
         spec.inputs['metadata']['options']['output_filename'].default = 'aiida.out'
 
-The first line after the import just runs the |define| method of the parent base |CalcJob| class, which will do some set-ups and define some basic `inputs` and `outputs` (we will see some of them shortly).
-The second "block of code" (lines 6-8), is where we are defining the specific inputs and outputs for the code, specifying their valid type (in this case AiiDA nodes of type |Int|) and a help message for the users of the class.
+The first line after the import just runs the |define| method of the parent base |CalcJob| class, which will define `inputs` and `outputs` that are common to all |CalcJob|'s (we will see some of them shortly).
+The second "block of code" (lines 6-8), is where we define the inputs and outputs that are specific to this implementation, specifying their valid type (in this case AiiDA nodes of type |Int|) and a description.
 
 The last block (lines 10+11) seems similar to the previous one, but has a subtle difference: it is not defining new `inputs`, but modifying some properties of the base `inputs` that are already defined in the parent |CalcJob| class.
-You can spot the difference in that the definition of new `inputs` uses the ``spec.input`` method, singular, whereas accessing is achieved via the ``spec.inputs`` method, plural.
+You can spot the difference in that the definition of new `inputs` uses the ``spec.input`` method (note `input` is singular here), whereas existing inputs are accessed through the ``spec.inputs`` property (where `inputs` is now plural).
 
-You can check the Topics section about :ref:`defining processes <topics:processes:usage:defining>` if you want more information about setting up your `inputs` and `outputs` (covering validation, dynamic number of inputs, etc).
+You can check the Topics section about :ref:`defining processes <topics:processes:usage:defining>` if you want more information about setting up your `inputs` and `outputs` (covering validation, dynamic number of inputs, etc.).
 
 Preparing for submission
 ------------------------
 
-There are two main tasks to take care of in this method: preparing the folder in which the code will be run (so that all required input files are correctly built or copied there) and setting up the configuration of the engine.
-The first one is achieved by manipulating the |Folder| object that the method receives as an input, whereas the second one requires the construction of the |CalcInfo| object (which is then returned by the method) and the |CodeInfo| object (which will be included in the |CalcInfo| one, see line 18 in the following code snippet).
+There are two main tasks to take care of in this method: writing the required input files into a folder in which the code will be run and defining additional options on how the code should be run.
+The first one is achieved by using the |Folder| object that the method receives as an argument, whereas the second one requires the construction of a |CalcInfo| object, which should be returned by the method.
 
 .. code-block:: python
    :linenos:
@@ -127,15 +127,15 @@ The first one is achieved by manipulating the |Folder| object that the method re
 
         return calcinfo
 
-The first block of code (lines 3-7) is just the unloading of the information stored in the ``spec`` into local variables.
+The first block of code (lines 3-7) is just unpacking the inputs that have been passed, which can be accessed through ``self.inputs``, into local variables.
 Note that this information is not accessed via the ``spec.inputs`` anymore, but by ``self.inputs``: by the time this method is executed the specs will have become properties of the |CalcJob| and now should contain the actual inputs provided by the user.
 
 The input required by the addition "code" is just a `bashscript` line with the value of the input nodes replaced appropriately.
-This is being created on line 9, using the |folder.open| method to get a handle to the file and simply wirting in it.
+This is being created on line 9, using the |folder.open| method to get a handle to the file and simply writing in it.
 This directory represented by the |Folder| object (along with all the files created in it) will not only be copied to the remote machine for the code to be run there, but will also be stored in the local repository of the calculation node.
 
 Next in lines 12-15 we are creating and setting up the ``codeinfo = CodeInfo()`` object.
-The ``code_uuid`` being passed in line 13 is necessary for the engine to get the base information from the |Code| node (in which computer the code is placed, what is the location of the executable, etc).
+The ``code_uuid`` that is passed in line 13 is necessary for the engine to get the required information from the |Code| node (such as the full path of the executable, etc.).
 Note that this was taken in the unloading block from ``input_code = self.inputs['code']``, which we never specified in the |define| method: this is one of the inputs defined in the base |CalcJob| class that we mentioned earlier when discussing the ``super().define(spec)``:
 
 .. code-block:: python
@@ -157,8 +157,8 @@ The specific combination presented here, together with the information inside of
 Through the |CodeInfo| object you can also pass more flags (by adding them as string elements to the list in ``codeinfo.cmdline_params``), configure what to pass through the standard input (just as it is shown for the standard output), add commands to be run before and after the execution line, etc.
 
 Finally, the last block remaining in lines 17-19 creates the |CalcInfo| object, passes to it the |CodeInfo| object, and adds the output to the ``retrieve_list``.
-This is a list of all files that the engine needs to copy back from the computer where the code is located, either for immediate parsing or local storage in an output node labeled ``retrieved`` for future post-processing.
-The ``retrieved`` node is a ``output`` defined in the base |CalcJob| class.
+This is a list of all files that the code will produce that the engine should copy from the computer where the code ran into an output node labeled ``retrieved``.
+The ``retrieved`` node is an ``output`` defined in the base |CalcJob| class.
 There are other lists available that allow you to easily customize how to move files to and from the remote working directory in order to prevent the creation of unnecessary copies.
 
 In the Topics section on :ref:`defining calculations <topics:calculations:usage>` you will find more information on available settings of the |CalcInfo| and |CodeInfo|, such as available copy lists, running script options, etc.
@@ -170,7 +170,7 @@ Parsing the outputs
 
 The parsing step occurs after the calculation has finished running and all the relevant outputs have been retrieved.
 It is an optional step that allows you to extract relevant information from the output files and store it into AiiDA nodes in formats that are easier and quicker to query and analyze.
-The way to implement this is by extending the base |Parser| class as you can see below:
+To parse retrieved files into nodes that can be stored in the database, one should implement the |Parser| class:
 
 .. code-block:: python
    :linenos:
@@ -183,22 +183,21 @@ The way to implement this is by extending the base |Parser| class as you can see
         def parse(self, **kwargs):
 
             output_folder = self.retrieved
-            output_filename = self.node.get_attribute('output_filename')
+            output_filename = self.node.get_option('output_filename')
 
             with output_folder.open(output_filename, 'r') as handle:
                 result = int(handle.read())
 
             self.out('sum', orm.Int(result))
 
-The first command in the example (line 8) shows the way to get the ``retrieve`` folder that was generated by the associated |CalcJob| and that, by this point, should contain the files included in the ``retrieve_list`` (as specified in the |prepare_for_submission| method).
-The second one (line 9) is also accessing a parameter of the |CalcJob|: in this case, the name of the output.
-After unloading this information into local variables, these are then used to open said output file located in the obtained retrieved folder and read the single integer value that was written there by the addition |Code| (lines 11 and 12).
+The first command in the example (line 8) shows how to get the ``retrieved`` folder that was generated by the associated |CalcJob|, which contains the files included in the ``retrieve_list`` (as specified in the |prepare_for_submission| method).
+The second one (line 9) retrieves the name of the output file that was defined in the inputs when the |CalcJob| was launched.
+Lines 11 and 12 show how the content of the output file in the output folder is read, which should be the sum as written by the code, and cast to an integer.
 
-Finally, it uses the ``out`` method to provide the ``sum`` output of the calculation its final value (of type AiiDA integer, as was specified in the |define| section of the associated |CalcJob|).
-It is important to note here that there is no return statement: the output is provided to the ``self.out`` method instead (any returned value is interpreted as an error signal) as an unstored node, and the engine will be in charge of performing the storing process.
+Finally, the parsed sum is wrapped into an |Int| node, which allows it to be registered as the ``sum`` output through the ``out`` method.
 
-In order to use this ``ArithmeticAddParser`` inside an appropriate |CalcJob| (such as the one described in the :ref:`previous section <how-to:codes:interfacing>`), one needs to add it as a ``metadata.options.parser_name`` input.
-You can set a parser as the default option in the |define| method, but note that this choice can be overriden when instantiating the |CalcJob|.
+To trigger the parsing using a |Parser| after a |CalcJob| has finished (such as the one described in the :ref:`previous section <how-to:codes:interfacing>`) its entry point name needs to be passed as the ``metadata.options.parser_name`` input.
+If a particular parser should always be used by default for a given |CalcJob|, it can be defined as the default in the |define| method.
 
 .. code-block:: python
 
@@ -206,15 +205,22 @@ You can set a parser as the default option in the |define| method, but note that
     def define(cls, spec):
         (...)
         spec.inputs['metadata']['options']['parser_name'].default = 'arithmetic.add'
-
-As can be seen in the previous line, the way to do this is not by passing the |Parser| class directly, but by providing the string label that identifies the |Parser| as a registered plugin in your working environment.
+Note that this default can be overridden through the inputs when launching the calculation job.
+To define the parser that should be used, one should not pass the |Parser| class itself, but rather the corresponding entry point name under which it is registered as a plugin.
 In other words, in order to use a |Parser| you will need to register it as explained in the following how-to section on :ref:`registering plugins <how-to:plugins>`.
 
 Handling parsing errors
 -----------------------
 
-In order for you to be able to provide the user with information regarding the errors that can ocur after the calculation has finished (so, mostly during parsing), you have the option of defining ``exit_codes`` for cases when something goes wrong.
-The typical way to implement these is to use ``try``/``except`` clauses to wrap the lines of code that might raise some typical errors, that can then be replaced by these ``exit_codes``:
+So far we have assumed in the implementation of the |Parser| that the code executed nominally and produced the correct output.
+For this trivial example this is likely the case, but for many codes there can be a variety of errors that prevent it from producing the desired result.
+These exit codes can be defined through the |spec| of the |CalcJob| that is used for that code, just as the inputs and output are defined
+The parser can be used to detect these problems and communicate them to the caller by returning an `exit code`.
+An `exit code` is a positive integer that corresponds to a particular known and well-defined error mode of a code.
+To signal a particular problem, you should just return the corresponding exit code.
+These can be easily referenced in the parser through the ``self.exit_codes`` property.
+For example, if the |CalcJob| defines an exit code with the label ``ERROR_READING_OUTPUT_FILE``, you can retrieve it as ``self.exit_codes.ERROR_READING_OUTPUT_FILE``.
+If an exit code is returned from the parser, the engine will set it on the corresponding calculation job node.
 
 .. code-block:: python
    :linenos:
@@ -227,19 +233,17 @@ The typical way to implement these is to use ``try``/``except`` clauses to wrap 
         except exceptions.NotExistent:
             return self.exit_codes.ERROR_NO_RETRIEVED_FOLDER
 
-        output_filename = self.node.get_attribute('output_filename')
+        output_filename = self.node.get_option('output_filename')
 
         try:
             with output_folder.open(output_filename, 'r') as handle:
                 try:
                     result = int(handle.read())
                 except ValueError:
-                    result = None
-        except (OSError, IOError):
+                    return self.exit_codes.ERROR_INVALID_OUTPUT
+        except OSError:
             return self.exit_codes.ERROR_READING_OUTPUT_FILE
 
-        if result is None:
-            return self.exit_codes.ERROR_INVALID_OUTPUT
 
         self.out('sum', orm.Int(result))
 
@@ -255,8 +259,8 @@ You then have to introduce all of these `exit_codes` inside of the |define| meth
         spec.exit_code(310, 'ERROR_READING_OUTPUT_FILE', message='The output file could not be read from the retrieved folder.')
         spec.exit_code(320, 'ERROR_INVALID_OUTPUT', message='The output file contains invalid output.')
 
-As you can see, for each ``exit_code`` you need to provide an ID number that will be used to identify it, a label you can then use to reference the code in the |parse| method (``self.exit_codes.LABEL``), and a message that will give the user more information on the problem.
-The Topics section on :ref:`defining processes <topics:processes:usage:defining>` also contains more information on how to use exit codes.
+As you can see, for each ``exit_code`` you need to provide a exit status (which should be a positive integer), a label that can be used to reference the code in the |parse| method (``self.exit_codes.LABEL``), and a message that provides a more detailed information on the problem.
+The Topics section on :ref:`defining processes <topics:processes:usage:defining>` provides additional information on how to use exit codes.
 
 Design guidelines
 -----------------
--- a/docs/source/howto/codes.rst
+++ b/docs/source/howto/codes.rst
@@ -4,40 +4,264 @@
 How to run external codes
 *************************
 
+To run an external code with AiiDA, you will need to use an appropriate :ref:`calculation plugin <topics:plugins>`.
+Its job is to orchestrate the whole process that starts with transforming the input nodes to the input files that the code expects, then copy everything to the machine where the code is located (using a :ref:`transport plugin <how-to:codes:transport>`), run the actual code (using a :ref:`scheduler plugin <how-to:codes:scheduler>`), and finally get back the outputs and, optionally, parse them into queryable AiiDA nodes (using a :ref:`parser plugin <how-to:codes:parser>`).
 
-.. _how-to:codes:plugin:
+The following subsections will take you through the process of :ref:`creating the calculation plugin<how-to:codes:interfacing>` and its acompanying :ref:`parser<how-to:codes:parsing>`, and then using these to actually :ref:`run the code<how-to:codes:run>`.
+Other supporting sections are also included.
+
+As the underlying example for these we will use the process of using the `bash` interpreter (``/bin/bash``) to add up two numbers by running the command: ``echo $(( numx + numy ))``.
+Here, the `bash` binary will be effectively acting as our |Code| executable, the input (``aiida.in``) will then be a file containing the command with the numbers provided by the user replaced, and the output (``aiida.out``) will be caught through the standard output.
+The final recipe to run this code will then be:
+
+.. code-block:: bash
+
+    /bin/bash aiida.in > aiida.out
+
+.. _how-to:codes:interfacing:
 
 Interfacing external codes
 ==========================
 
-`#3986`_
+The way you provide AiiDA with the set of instructions required to run a code is by extending the |CalcJob| class, which has the following two key methods:
 
-In order to work with an external simulation code in AiiDA, you need a calculation input plugin (subclassing the :py:class:`~aiida.engine.CalcJob` class) and an output parser plugin (subclassing the :py:class:`~aiida.parsers.Parser` class):
+.. code-block:: python
 
-Before starting to write a plugin, check on the `aiida plugin registry <https://aiidateam.github.io/aiida-registry/>`_ whether a plugin for your code is already available.
+    from aiida.engine import CalcJob
+
+    class ArithmeticAddCalculation(CalcJob):
+
+        @classmethod
+        def define(cls, spec):
+            super().define(spec)
+            # Specifications of the inputs and outputs (and exit codes).
+            # no return statement
+
+        def prepare_for_submission(self, folder):
+            # No super call to parent class method
+            # Preparation of input files and instructions for engine
+            return calcinfo
+
+We will now indicate how to deal with each of them separatedly.
+
+Defining the specifications
+---------------------------
+
+As the comment in the code above indicates, the first method (|define|) is where one specifies the different inputs that the user of the |CalcJob| will have to provide in order to run the code, as well as the outputs that will be produced (exit codes are discussed in the respective section).
+This is done through the |spec| object, which, as can be seen, is passed as an input to the method.
+For the code that adds up two numbers, we will need to define those numbers as inputs (lets call them ``x`` and ``y`` to label them) and the result as an output (``sum``).
+
+.. code-block:: python
+   :linenos:
+
+    @classmethod
+    def define(cls, spec):
+        from aiida import orm
+        super().define(spec)
+
+        spec.output('sum', valid_type=orm.Int, help='The sum of the left and right operand.')
+        spec.input('x', valid_type=orm.Int, help='The left operand.')
+        spec.input('y', valid_type=orm.Int, help='The right operand.')
+
+        spec.inputs['metadata']['options']['input_filename'].default = 'aiida.in'
+        spec.inputs['metadata']['options']['output_filename'].default = 'aiida.out'
+
+The first line after the import just runs the |define| method of the parent base |CalcJob| class, which will do some set-ups and define some basic `inputs` and `outputs` (we will see some of them shortly).
+The second "block of code" (lines 6-8), is where we are defining the specific inputs and outputs for the code, specifying their valid type (in this case AiiDA nodes of type |Int|) and a help message for the users of the class.
+
+The last block (lines 10+11) seems similar to the previous one, but has a subtle difference: it is not defining new `inputs`, but modifying some properties of the base `inputs` that are already defined in the parent |CalcJob| class.
+You can spot the difference in that the definition of new `inputs` uses the ``spec.input`` method, singular, whereas accessing is achieved via the ``spec.inputs`` method, plural.
+
+You can check the Topics section about :ref:`defining processes <topics:processes:usage:defining>` if you want more information about setting up your `inputs` and `outputs` (covering validation, dynamic number of inputs, etc).
+
+Preparing for submission
+------------------------
+
+There are two main tasks to take care of in this method: preparing the folder in which the code will be run (so that all required input files are correctly built or copied there) and setting up the configuration of the engine.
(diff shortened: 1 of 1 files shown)


ramirezfranciscof:
So...anybody sees why is the pre-commit failing on a file that I didn't touch?

csadorf:
@ramirezfranciscof Please explicitly request my review once you think this PR is ready for review. Feel free to request review before the content is polished, I'm happy to provide input on a draft.

ramirezfranciscof:
@csadorf Yes, go ahead please. I mean probably there is a lot to polish still but I have already brewed and iterated this by myself for way too long, so now I need to get some external feedback.

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
+.. _how-to:codes:plugin:
+
+Interfacing external codes
+==========================
+
+In order to run an external code, you will first have to explain to AiiDA how to run it: how to create inputs files from the content of the input node, how the program can read those input files, where to find and how to read the output files, etc.
+The way you transmit these instructions to AiiDA is by extending the |CalcJob| and the |Parser| classes:
```
csadorf, line 7:
```suggestion
The way you provide these instructions to AiiDA is by extending the |CalcJob| and the |Parser| classes:
```

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
...
+    from aiida.parsers.parser import Parser
+
+    class ArithmeticAddParser(Parser):
+
+        def parse(self, **kwargs):
+            # Instructions on how to process the output files
+            # no return statement
+
+As you can guess from the name of the child-classes in the previous code snippet, we will illustrate the implementation of this code wrapping procedure with a possible design for the example introduced in the :ref:`tutorial<tutorial:basic>` of a code that sums two numbers.
+To have a clearer idea of how to interact with this code, we can assume it consists of the following bash script which basically reads two consecutive integers in a single line of an input file provided through a command line argument and write to standard output the result of its sum (so one would run it as ``$ add.sh -in [input_file] > [output_file]``).
```
csadorf, line 38: I think you should introduce the example code with its purpose, inputs and outputs first. That will make it easier to follow how you wrap it.

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
...
+
+.. code:: bash
```
csadorf, line 40: Consider placing the code for the `add.sh` script in  dropdown or possibly don't provide it at all. The how-to section is not a tutorial, so you can use examples, but there is no expectation that users can follow them step by step.

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
...
+        @classmethod
+        def define(cls, spec):
+            super().define(spec)
+            # Specifications of inputs, outputs and exit codes.
+            # no return statement
+
+        def prepare_for_submission(self, folder):
+            # Instructions on how to prepare the input files
+            return calcinfo
+
```
csadorf, line 25: You should place all code segments into context. Just a brief sentence before the segment about what you are showing.

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
...
+    # -*- coding: utf-8 -*-
+    from aiida.parsers.parser import Parser
+
+    class ArithmeticAddParser(Parser):
+
+        def parse(self, **kwargs):
+            # Instructions on how to process the output files
+            # no return statement
+
+As you can guess from the name of the child-classes in the previous code snippet, we will illustrate the implementation of this code wrapping procedure with a possible design for the example introduced in the :ref:`tutorial<tutorial:basic>` of a code that sums two numbers.
```
csadorf, line 37: I think it would be better to not directly reference the tutorial. For the how-to's, we make the assumption that people know the things we teach in the tutorial, but the how-to's are not a direct extension of the tutorial.

Think of it this way, an AiiDA expert might still want to look some things up among the how-to guides, but may never have gone through the tutorial.

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
...
+
+        spec.input(
+            'metadata.options.input_filename',
+            valid_type=str,
+            required=False,
+            help='Filename to which the input for the code that is to be run is written.'
+        )
+
+You can see that this is input is defined with ``required=False``, so the user can ``run`` the code without specifying it in its instantiated |CalcJob|.
+This is not the case for the specific inputs we defined (since ``required=True`` by default), so if the user tries to ``run`` or ``submit`` the |CalcJob| without providing these, the code will raise.
```
csadorf, line 105:
```suggestion
This is not the case for the specific inputs we defined (since ``required=True`` by default), so if the user tries to ``run`` or ``submit`` the |CalcJob| without providing these, the code will raise an exception.
```

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
...
+For more information on how to provide these different kind of inputs when runnig a code, you can see the :ref:`next how-to section<tutorial:basic>` which covers exactly this.
```
csadorf, line 106: I'm confused about the reference here. Why are we linking to the tutorial?

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
...
+        codeinfo.cmdline_params = ['-in', input_filename]
+
+        calcinfo = CalcInfo()
+        calcinfo.codes_info = [codeinfo]
+        calcinfo.retrieve_list = [output_filename]
+
+        return calcinfo
+
+The second block corresponds to our previously described first step (preparing the inputs to run): here we are using the ``folder.open`` method to get a handle to the file and write in an input file (in the format in which the code will read it) the values stored in our input nodes.
+This folder (along with all the files created in it) will not only be copied to the remote machine for the code to be ran there, but will also be stored in the local repository of the calculation node.
```
csadorf, line 149:
```suggestion
This folder (along with all the files created in it) will not only be copied to the remote machine for the code to be run there, but will also be stored in the local repository of the calculation node.
```

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
...
+
+The last two blocks correspond to the second step (instructions for the engine).
+In the first one we are creating the ``codeinfo = CodeInfo()`` object, that will be then part of the ``calcinfo = CalcInfo()`` object, as one can see in ``calcinfo.codes_info = [codeinfo]``.
+The ``code_uuid`` is necessary for the engine to get the base information set up in the |Code| node (in which machine the code is placed, what is the location of the executable, etc).
+Note that this was taken from ``input_code = self.inputs['code']``, which we never specified in the ``define`` method: this is one of the two mandatory inputs defined in the base |CalcJob| that will need to be always provided in the instance.
+Moreover, in the case of the ``inputs['code']``, the designer of the |CalcJob| sub-class will always have to take care to pass its uuid into ``codeinfo.code_uuid`` as shown here.
```
csadorf, line 155:
```suggestion
Moreover, in the case of the ``inputs['code']``, the designer of the |CalcJob| sub-class will always have to take care to pass its UUID into ``codeinfo.code_uuid`` as shown here.
```
The name UUID must be block capitalized.

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
...
+    # Scheduler commands
+
+    # Pre-running setups
+
+    '${executable_from_code_uuid}' '-in' '${input_filename}' > ${output_filename}
+
+    # Post-running commands
+
+Which will default to ``'add.sh' '-in' 'aiida.in' > 'aiida.out'``.
+Through the |CodeInfo| instanse it is also posible to configure what to pass through the standard input, as well as what commands to run before and after the execution line.
```
csadorf, line 180:
```suggestion
Through the |CodeInfo| instance it is also possible to configure what to pass through the standard input, as well as what commands to run before and after the execution line.
```

docs/source/howto/codes/interfacing.rst
```diff
@@ -0,0 +1,282 @@
...
+
+The final line remaining (``calcinfo.retrieve_list = [output_filename]``) just indicates to the engine which files should be retrieved (either for immediate parsing or local storage for future post-processing).
```
csadorf, line 182: Sphinx allows you to turn on line numbers for code segments. I'd suggest you turn those on an then reference the lines by number.

ramirezfranciscof:
> It is great to provide an example in the how-to's, but we need to be very careful that it doesn't become a tutorial. You can show and demonstrate, but users are not expected to follow step by step.

Oh, ok, now then I'm operating under a different understanding of what a How-to is, or well, this depends on the exact meaning of "users are not **expected** to follow step by step". I mean, I would not expect them to in the sense that a how-to can work as a quick reference for just the part one forgot or is not familiar with, but I thought the idea is that the how-to is a step-by-step guide on how to do something. Is this not so? What would be a better description?

docs/source/howto/codes/interfacing.rst
ramirezfranciscof: Noise.

docs/source/howto/codes/interfacing.rst
ramirezfranciscof: Sorry, I meant to reference the how-to run codes section, I copy/pasted and forgot to replace, he-he.

docs/source/howto/codes/interfacing.rst
sphuber: Here it might be better to go with the changes of Marnik, which does not require a separate script at all, but just uses `bash`. Since this can typically be found always at the same place, it makes it very easy to configure the code (and even provide a complete configuration yaml file) and easier to explain, as we are not executing some custom script, but just bash.

csadorf:
> > It is great to provide an example in the how-to's, but we need to be very careful that it doesn't become a tutorial. You can show and demonstrate, but users are not expected to follow step by step.
> 
> Oh, ok, now then I'm operating under a different understanding of what a How-to is, or well, this depends on the exact meaning of "users are not **expected** to follow step by step". I mean, I would not expect them to in the sense that a how-to can work as a quick reference for just the part one forgot or is not familiar with, but I thought the idea is that the how-to is a step-by-step guide on how to do something. Is this not so? What would be a better description?

I think the difference is subtle. Yes, a how-to guide should show concrete steps, meaning all code examples should be correct and can in principle be copy-pasted. However, they don't have to be comprehensive. You can leave things out, make some assumptions about the user being able to place them into context. Here is the description from the wiki:

> The how-to guides should provide concrete steps, but do not have to be as detailed as a tutorial and further are not required to demonstrate a specific recipe from start to finish, meaning that there is some expectation that users are generally familiar with AiiDA.

docs/source/howto/codes/interfacing.rst
ramirezfranciscof: Ok; I think it still may be relevant if someone who went through the tutorial to place themselves more easily, but I reduced the presence to a "btw" comment that can be just ignored by more expert users:

> As you may have guessed from the name of the child-classes in the previous code snippet, we will illustrate this code wrapping procedure by designing a calculation that will add up two numbers (note that this is the same example used in the `tutorial`).

Would this be better or do you still think is best to remove it completely?

csadorf: Yes, I think that's good.

docs/source/howto/codes/interfacing.rst
ramirezfranciscof: Not super sure what you mean since I think this code is preceeded by an explanation of what is going to contain, but I modified it a bit with the hope that it is clearer now.

csadorf:
@ramirezfranciscof Please re-request review when this is ready for a second round, thank you!

ramirezfranciscof:
@csadorf Yes, I was finishing some checks and writing this text. So, I already took care of all the specific comments, and changed the code version used to bash as suggested by @sphuber. I think a lot got simplified, but please check if now this is better addressed:

> A how-to should be problem-oriented and limited to showing how to achieve a specific goal. It doesn't have to be comprehensive and detailed explanations are not needed. The current content reads more like a comprehensive topic section that tries to teach every detail. We could consider to move much of this content into the "Topics" chapter and then show a stripped down version in the how-to's? I know that is very difficult so feel free to ask for help on how to achieve that concretely.

If not, the only card I have left is removing the `exit_codes`, but I am a bit afraid of people starting to code CalcJobs without doing proper handling of exceptions (I am not sure how the engine reacts to these: do they get a generic code or are these the reason why some jobs just hang and stay running forever and similar problems?). Any advice on this or other stuff to take out is welcome.

sphuber:
> If not, the only card I have left is removing the `exit_codes`, but I am a bit afraid of people starting to code CalcJobs without doing proper handling of exceptions (I am not sure how the engine reacts to these: do they get a generic code or are these the reason why some jobs just hang and stay running forever and similar problems?). Any advice on this or other stuff to take out is welcome.

There is no risk here whatsoever. Since exit codes are only set in the parser (the plugin specific ones that is, the engine can also set a select few), and parsing is completely optional, the default exit status for a calculation job is simply `Finished [0]`. A non-zero exit status is the non-default behavior and requires the parser to actually return it. Exit codes are not there to handle exceptions, that is what the engine does.

With that in mind, I think it might be perhaps even best to first describe an example that does not mention exit codes at all. This will make it very simple and clear. Straight after, you can then mention "but what if the output is incorrect, how can I signal there was a problem", and you start a next section "How to deal with errors" or something to that effect.

I am also thinking whether it would be better to describe the parsing completely. For a kick-off, it is implemented separately from the calculation job class in AiiDA, but it is also completely optional. It is a perfectly valid use case to not parse. All outputs are still accessible as they are stored in the `retrieved` node. This will simplify the first explanation even further. At the end of the section you can say 
>The output files are stored in the `retrieved` folder, which you can inspect using ... <with link>. If you want to make the results queryable and easily reusable, you want to parse the results and turn (part of) it into output nodes. Parsing the output files of a calculation job is done by the `Parser` class.

ramirezfranciscof:
Ok, yeah, I actually like that structure of dividing it into 3 ("how to interface/parse/handle errors") a lot better. Unless someone raises any issue, I'll start working on this (so hold on on the review @csadorf ).

BTW: should I maybe take the description to the idea of the code (bash running the command in the input) as an introduction to the whole section (so, under the first "how to run external codes" header)? It would be relevant but not strictly necessary for almost all sub-sections, and repeating it at the begining of all of them seems a bit...redundant?

sphuber:
> BTW: should I maybe take the description of the idea of the underlying code (bash running the command in the input) as an introduction to the whole section?

I think that might make sense. In the first section, you can say something to the effect:
> For the purposes of an example, we will launch an `ArithmeticAddCalculation` which wraps `bash` as a code to sum two integers.

Then in the next section, you can reference it again and explain how the concept generalizes.
> In principle, any code that has a command line interface can be run through AiIDA, just as the bash example from the previous section.
> The calculation job plugin functions to translate the inputs, which are AiiDA nodes, to the input files and scripts that the code expects.
> Here we will demonstrate how this works at the hand of the `ArithmeticAddCalculation`

csadorf:
@ramirezfranciscof I also like the idea of splitting this up into the mentioned sub-sections. Please ping me here with a mention when this is ready for my review.

ramirezfranciscof:
Ok, so I finished the restructure. There possibly be some things to correct, but I think that overall this has a much balanced size and distribution now. @csadorf @sphuber dig in.

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,264 @@
...
+        spec.input('x', valid_type=orm.Int, help='The left operand.')
+        spec.input('y', valid_type=orm.Int, help='The right operand.')
+
+        spec.inputs['metadata']['options']['input_filename'].default = 'aiida.in'
+        spec.inputs['metadata']['options']['output_filename'].default = 'aiida.out'
+
+The first line after the import just runs the ``define`` method of the parent base |CalcJob| class, which will do some set-ups and define some basic `inputs` and `outputs` (we will see some of them shortly).
+The second "block of code" (lines 6-8), is where we are defining the specific inputs and outputs for the code, specifying their valid type (in this case AiiDA nodes of type |Int|) and a help message for the users of the class.
+
+The last block (lines 10-12) seems similar to the previous one, but has a subtle difference: it is not defining new `inputs`, but modifying some properties of the base `inputs` that are already defined in the parent |CalcJob| class.
```
csadorf, line 72:
```suggestion
The last block (lines 10+11) seems similar to the previous one, but has a subtle difference: it is not defining new `inputs`, but modifying some properties of the base `inputs` that are already defined in the parent |CalcJob| class.
```

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,264 @@
...
+
+The first block of code (lines 3-7) is just the unloading of the information stored in the ``spec`` into local variables.
+Note that this information is not accessed via the ``spec.inputs`` anymore, but by ``self.inputs``: by the time this method is executed the specs will have become properties of the ``CalcJob`` and now should contain the actual inputs provided by the user.
+
+The input required by the addition "code" is just a `bashscript` line with the value of the input nodes replaced appropriately.
+This is being created on line 9, using the ``folder.open`` method to get a handle to the file and simply wirting in it.
+This directory represented by the |Folder| object (along with all the files created in it) will not only be copied to the remote machine for the code to be run there, but will also be stored in the local repository of the calculation node.
+
+Next in lines 12-15 we are creating and setting up the ``codeinfo = CodeInfo()`` object.
+The ``code_uuid`` being passed in line 13 is necessary for the engine to get the base information from the |Code| node (in which computer the code is placed, what is the location of the executable, etc).
```
csadorf, line 117: Somehow the vertical bars are showing up in the [formatted output here](https://36-77234579-gh.circle-artifacts.com/0/html/howto/codes.html).

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,264 @@
...
+        calcinfo.retrieve_list = [output_filename]
+
+        return calcinfo
+
+
+The first block of code (lines 3-7) is just the unloading of the information stored in the ``spec`` into local variables.
+Note that this information is not accessed via the ``spec.inputs`` anymore, but by ``self.inputs``: by the time this method is executed the specs will have become properties of the ``CalcJob`` and now should contain the actual inputs provided by the user.
+
+The input required by the addition "code" is just a `bashscript` line with the value of the input nodes replaced appropriately.
+This is being created on line 9, using the ``folder.open`` method to get a handle to the file and simply wirting in it.
```
csadorf, line 113: I think it would be good to link to API reference here. The same goes for many other code elements named here.

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,264 @@
...
+This directory represented by the |Folder| object (along with all the files created in it) will not only be copied to the remote machine for the code to be run there, but will also be stored in the local repository of the calculation node.
+
+Next in lines 12-15 we are creating and setting up the ``codeinfo = CodeInfo()`` object.
+The ``code_uuid`` being passed in line 13 is necessary for the engine to get the base information from the |Code| node (in which computer the code is placed, what is the location of the executable, etc).
+Note that this was taken in the unloading block from ``input_code = self.inputs['code']``, which we never specified in the ``define`` method: this is one of the inputs defined in the base |CalcJob| class that we mentioned earlier when discussing the ``super().define(spec)``:
```
csadorf, line 118: I think ``define`` should also be linked to the API reference.

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,264 @@
...
+.. code-block:: bash
+
+    #!/bin/bash
+
+    '[executable path in code node]' '[input_filename]' > '[output_filename]'
+
+Through the |CodeInfo| object you can also pass more flags (by adding them as string elements to the list in ``codeinfo.cmdline_params``), configure what to pass through the standard input (just as it is shown for the standard output), add commands to be run before and after the execution line, etc.
+
+Finally, the last block remaining in lines 17-19 creates the |CalcInfo| object, passes to it the |CodeInfo| object, and adds the output to the ``retrieve_list``.
+This is a list of all files that the engine needs to copy back from the computer where the code is located, either for immediate parsing or local storage in an output node labeled ``retrieved`` for future post-processing (again, this is another base ``output`` defined in the base |CalcJob| class).
```
csadorf, line 139:
```suggestion
This is a list of all files that the engine needs to copy back from the computer where the code is located, either for immediate parsing or local storage in an output node labeled ``retrieved`` for future post-processing.
The ``retrieved`` node is a ``output`` defined in the base |CalcJob| class.
```

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,264 @@
...
+You can expand this list to include more files, and there are other lists available that allow you to easily specify which files to make available in the remote folder without having to manually copy them during this step and thus having to store unnecessary copies of them in the calculation node.
```
csadorf, line 140: This is a *very long* sentence and I find it quite confusing. You are mentioning this and other lists (without actually specifying them) to let the user know how to control what is copied and what is not copied? I'd appreciate if you could try to reword this sentence. It may be warranted its own paragraph.

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,264 @@
...
+
+In the Topics section on :ref:`defining calculations <topics:calculations:usage>` you will find more information on available settings of the |CalcInfo| and |CodeInfo|, such as available copy lists, running script options, etc.
 
 Design guidelines
-------------------
+-----------------
```
csadorf, line 145: Something is off here. Why is this followed by the "Design guidelines" completely out of context? Did something go wrong during a merge/rebase?

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,264 @@
...
    | If the code can do it, there should be *some* way to do it with your plugin.
-
  * | **Don't rely on AiiDA internals.**
    | AiiDA's :ref:`public python API<python_api_public_list>` includes anything that you can import via ``from aiida.module import thing``.
    | Functionality at deeper nesting levels is not considered part of the public API and may change between minor AiiDA releases, breaking your plugin.
 
+.. _how-to:codes:parsing:
+
+Parsing the outputs
+===================
```
csadorf, line 164: Shouldn't this be a sub-section of "Interfacing external code" ?

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,264 @@
...
+
+            output_folder = self.retrieved
+            output_filename = self.node.get_attribute('output_filename')
+
+            with output_folder.open(output_filename, 'r') as handle:
+                result = int(handle.read())
+
+            self.out('sum', orm.Int(result))
+
+The first command in the example (line 8) shows the way to get the ``retrieve`` folder that was generated by the associated ``CalcJob`` and contains the files included in the ``retrieve_list`` during the ``prepare_for_submission``.
```
csadorf, line 188: > during the ``prepare_for_submission``.

What does that mean?

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,264 @@
...
+After unloading this information into local variables, these are then used to open said output file located in the obtained retrieved folder and read the single integer value that was written there by the addition ``code`` (lines 11 and 12).
+
+Finally, it uses the ``out`` method to provide the ``sum`` output of the calculation its final value (of type AiiDA integer, as was specified in the ``define`` section of the associated |CalcJob|).
+It is important to note here that there is no return statement: the output is provided to the ``self.out`` method instead (any returned value is interpreted as an error signal) as an unstored node, and the engine will be in charge of performing the storing process.
+
+In order to use this ``ArithmeticAddParser`` inside an appropriate |CalcJob| (such as the one described in the :ref:`previous section <how-to:codes:interfacing>`), one needs to add it as a `metadata.options.parser_name` input.
+You can leave it for the user to do so when instantiating the desired calculation plugin, or, if you are developing the calculation plugin, you can add it as the default option in the ``define`` method:
```
csadorf, line 196: This how-to is written for users, so this sentence is confusing.

ramirezfranciscof:
Ok, I see. I will adress the particular comments later, but first the general structure: from the comments of @sphuber I understood that was being suggested was to separate "how to parse" and "how to handle errors" as completely separate sections of "How to run external codes". Afterwards I saw how the "how to handle errors" turned out and thought it might be actually better to have it be a sub-section of parsing. I asked @sphuber and he agreed, but now I'm afraid we might have different contexts in our heads because, re-reading his comments, I realize the whole idea might have been a pretty liberal interpretation on my part.

Anyways, I do believe from what we discussed and what came out that separating "how-to parse outputs" from "how-to interface codes" does lead to a more balanced structure. If you disagree, however, it shouldn't be too difficult to re-incorporate it as a sub-section (trying to minimize the changes in content).

csadorf:
@ramirezfranciscof Given the context of the overall section "How to run external codes", splitting "how to parse" and "how to handle errors" into distinct sections may actually make sense. So, just find a structure that makes sense to you and we can adjust later if needed.

docs/source/howto/codes.rst
ramirezfranciscof: I added most of them to the docs. The only ones that were left without an API reference were the `spec`, `retrieve_list`(because I couldn't find a specific useful method to link to) and one `ArithmeticAddParser` mention (because the exposed implementation is a simplified version from the current one shipped with aiida so I think it could be confusing).

docs/source/howto/codes.rst
ramirezfranciscof: Ok,, removed the clarification for expansion (it is probably self evident) and shortened a bit the phrase about the other lists:


This is a list of all files that the engine needs to copy back from the computer where the code is located, either for immediate parsing or local storage in an output node labeled ``retrieved`` for future post-processing.
The ``retrieved`` node is a ``output`` defined in the base |CalcJob| class.
There are other lists available that allow you to easily customize how to move files to and from the remote working directory in order to prevent the creation of unnecessary copies.

docs/source/howto/codes.rst
ramirezfranciscof: "Design guidelines" was something that @ltalirz  put here in one of his recen PR's, I think the plugin ones. I was not sure how to incorporate them into the text of the how-to so I left it out as a final sub-section (I only separated the item that referred to parsing and moved it at the end of the respective section). I'm not really sure what to do with it other than this but I also didn't want to just delete it.

docs/source/howto/codes.rst
ramirezfranciscof: Changed it, clearer?:

The first command in the example (line 8) shows the way to get the ``retrieve`` folder that was generated by the associated |CalcJob| and that, by this point, should contain the files included in the ``retrieve_list`` (as specified in the |prepare_for_submission| method).

docs/source/howto/codes.rst
ramirezfranciscof: Changed it, clearer?

In order to use this ``ArithmeticAddParser`` inside an appropriate |CalcJob| (such as the one described in the :ref:`previous section <how-to:codes:interfacing>`), one needs to add it as a `metadata.options.parser_name` input.
You can set a parser as the default option in the |define| method, but note that this choice can be overriden when instantiating the |CalcJob|.

docs/source/howto/codes.rst
csadorf: Yes, that's clearer.

docs/source/howto/codes.rst
csadorf: Better!

ramirezfranciscof:
@csadorf btw: I already changed all I could, I'm just waiting for responses in the unresolved conversations in order to make any final change before re-requesting review.

docs/source/howto/codes.rst
csadorf: Ok.

docs/source/howto/codes.rst
csadorf: Ok, then let's just leave it there and I'll try to figure it out later.

csadorf:
> @csadorf btw: I already changed all I could, I'm just waiting for responses in the unresolved conversations in order to make any final change before re-requesting review.

@ramirezfranciscof Just responded, I think we're fine now. If you could just resolve the merge conflict we are probably good to go.

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,265 @@
 How to run external codes
 *************************
 
+To run an external code with AiiDA, you will need to use an appropriate :ref:`calculation plugin <topics:plugins>`.
+Its job is to orchestrate the whole process that starts with transforming the input nodes to the input files that the code expects, then copy everything to the machine where the code is located (using a :ref:`transport plugin <how-to:codes:transport>`), run the actual code (using a :ref:`scheduler plugin <how-to:codes:scheduler>`), and finally get back the outputs and, optionally, parse them into queryable AiiDA nodes (using a :ref:`parser plugin <how-to:codes:parser>`).
 
-.. _how-to:codes:plugin:
+The following subsections will take you through the process of :ref:`creating the calculation plugin<how-to:codes:interfacing>` and its acompanying :ref:`parser<how-to:codes:parsing>`, and then using these to actually :ref:`run the code<how-to:codes:run>`.
+Other supporting sections are also included.
```
csadorf, line 11: After reviewing this once more, I think the "Design guidelines" from line 144 should go right after this paragraph.

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,265 @@
...
+
+            output_folder = self.retrieved
+            output_filename = self.node.get_attribute('output_filename')
+
+            with output_folder.open(output_filename, 'r') as handle:
+                result = int(handle.read())
+
+            self.out('sum', orm.Int(result))
+
+The first command in the example (line 8) shows the way to get the ``retrieve`` folder that was generated by the associated |CalcJob| and that, by this point, should contain the files included in the ``retrieve_list`` (as specified in the |prepare_for_submission| method).
```
csadorf, line 188: The `| prepare_for_submission|` reference does not resolve properly.
https://88-77234579-gh.circle-artifacts.com/0/html/howto/codes.html#parsing-the-outputs

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,265 @@
...
+The second one (line 9) is also accessing a parameter of the |CalcJob|: in this case, the name of the output.
+After unloading this information into local variables, these are then used to open said output file located in the obtained retrieved folder and read the single integer value that was written there by the addition |Code| (lines 11 and 12).
+
+Finally, it uses the ``out`` method to provide the ``sum`` output of the calculation its final value (of type AiiDA integer, as was specified in the |define| section of the associated |CalcJob|).
+It is important to note here that there is no return statement: the output is provided to the ``self.out`` method instead (any returned value is interpreted as an error signal) as an unstored node, and the engine will be in charge of performing the storing process.
+
+In order to use this ``ArithmeticAddParser`` inside an appropriate |CalcJob| (such as the one described in the :ref:`previous section <how-to:codes:interfacing>`), one needs to add it as a `metadata.options.parser_name` input.
```
csadorf, line 195:
```suggestion
In order to use this ``ArithmeticAddParser`` inside an appropriate |CalcJob| (such as the one described in the :ref:`previous section <how-to:codes:interfacing>`), one needs to add it as a ``metadata.options.parser_name`` input.
```
Code quotes  must be placed within two backticks for reStructuredText.

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,265 @@
...
+
+Design guidelines
+-----------------
+
+ * | **Parse what you want to query for.**
+   | Make a list of which information to:
+
+     #. parse into the database for querying (:py:class:`~aiida.orm.nodes.data.dict.Dict`, ...)
+     #. store in the file repository for safe-keeping (:py:class:`~aiida.orm.nodes.data.singlefile.SinglefileData`, ...)
+     #. leave on the computer where the calculation ran (:py:class:`~aiida.orm.nodes.data.remote.RemoteData`, ...)
```
csadorf, line 264: It's not clear to me what the meaning of the ellipses is here.

docs/source/howto/codes.rst
```diff
@@ -4,41 +4,265 @@
...
+        spec.exit_code(320, 'ERROR_INVALID_OUTPUT', message='The output file contains invalid output.')
+
+As you can see, for each ``exit_code`` you need to provide an ID number that will be used to identify it, a label you can then use to reference the code in the |parse| method (``self.exit_codes.LABEL``), and a message that will give the user more information on the problem.
+The Topics section on :ref:`defining processes <topics:processes:usage:defining>` also contains more information on how to use exit codes.
+
+Design guidelines
+-----------------
+
+ * | **Parse what you want to query for.**
+   | Make a list of which information to:
```
csadorf, line 260: This is a list with only one bullet point. Is there supposed to be another bullet point?

docs/source/howto/codes.rst
```diff
@@ -162,6 +386,19 @@ Adding support for a custom transport
...
+.. |Parser| replace:: :py:class:`~aiida.parsers.parser.Parser`
+.. |parse| replace:: :py:class:`~aiida.parsers.parser.Parser.parse`
+.. |folder| replace:: :py:class:`~aiida.common.folders.Folder`
+.. |folder.open| replace:: :py:class:`~aiida.common.folders.Folder.open`
+.. |CalcJob| replace:: :py:class:`~aiida.engine.processes.calcjobs.calcjob.CalcJob`
+.. |CalcInfo| replace:: :py:class:`~aiida.common.CalcInfo`
+.. |CodeInfo| replace:: :py:class:`~aiida.common.CodeInfo`
+.. |spec| replace:: ``spec``
+.. |define| replace:: :py:class:`~aiida.engine.processes.calcjobs.CalcJob.define`
+.. |prepare_for_submission| :py:class:`~aiida.engine.processes.calcjobs.CalcJob.prepare_for_submission`
```
csadorf, line 400: I think sphinx might not like the underscore, but not sure.

ramirezfranciscof: Nah, it was the fact that I was not telling it to actually `replace::` the thing =P.

docs/source/howto/codes.rst
ramirezfranciscof: The list of all guidelines was only one and I took out the point that was about parsing to put it in the parse section. I thought it would be better to keep the same formet even though that meant having a one item list, but now I tend to agree that its not so good. Anyways, if we go with your other suggestion I will just put all guidelines back together, otherwise I may remove this subsection and just add the content as a paragraph at the endo of the parse section.

docs/source/howto/codes.rst
ramirezfranciscof: So, I didn't write this part so this is just a guess, but I think it is suggesting one posibility of other options? Like, to store queryable data you can do it as a `Dict` but maybe some data would be more convenient to store as an array.

docs/source/howto/codes.rst
ramirezfranciscof: Ok, I can change it. The only caveat is that this are designing guidelines, and although most sections are about designing (calcjobs, parsers, adding caching, schedulers, transports), there are some that are not (setting up computer and running), and furthermore the whole section is called "How to **run** external codes", so it might be weird to read about **design** guidelines in the begining. Please confirm if this is ok.

sphuber:
Thanks for the review @csadorf , if you don't mind I would also like to give a pass. I will try to do it today and keep your past conversations in mind as much as possible.

csadorf:
> Thanks for the review @csadorf , if you don't mind I would also like to give a pass. I will try to do it today and keep your past conversations in mind as much as possible.

Sounds good!

@ramirezfranciscof Please re-request review after addressing both @sphuber and my comments. Thank you.

ramirezfranciscof:
@csadorf Sure! Do check the last responses for some of the comments when you get the time =D.

docs/source/howto/codes.rst
csadorf: Yeah, let's just go with it for now.

docs/source/howto/codes.rst
csadorf: Ok, sorry the provenance of those lines got lost. Nevermind then.

docs/source/howto/codes.rst
csadorf: Let's keep it all together.

csadorf:
> @csadorf Sure! Do check the last responses for some of the comments when you get the time =D.

Done! :)

csadorf:

docs/source/howto/codes.rst
```diff
@@ -4,40 +4,263 @@
...
+        spec.inputs['metadata']['options']['parser_name'].default = 'arithmetic.add'
+
+As can be seen in the previous line, the way to do this is not by passing the |Parser| class directly, but by providing the string label that identifies the |Parser| as a registered plugin in your working environment.
+In other words, in order to use a |Parser| you will need to register it as explained in the following how-to section on :ref:`registering plugins <how-to:plugins>`.
+
+Handling parsing errors
+-----------------------
+
+In order for you to be able to provide the user with information regarding the errors that can ocur after the calculation has finished (so, mostly during parsing), you have the option of defining ``exit_codes`` for cases when something goes wrong.
+The typical way to implement these is to use ``try``/``except`` clauses to wrap the lines of code that might raise some typical errors, that can then be replaced by these ``exit_codes``:
```
sphuber, line 217: I would not frame it as try/except because that is not all necessary or the idea behind. There don't need to be exceptions for one to want to return an exit code. For example you can check the output of the file, and all works just fine, but you conclude the output is not as desired, so you return an exit code. The try/except is not at all linked to in such a direct way. I would rather propose something like the following:
```suggestion
To signal a particular problem, you should just return the corresponding exit code.
These can be easily referenced in the parser through the ``self.exit_codes`` property.
For example, if the |CalcJob| defines an exit code with the label ``ERROR_READING_OUTPUT_FILE``, you can retrieve it as ``self.exit_codes.ERROR_READING_OUTPUT_FILE``.
If an exit code is returned from the parser, the engine will set it on the corresponding calculation job node.
```

docs/source/howto/codes.rst
```diff
@@ -4,40 +4,263 @@
...
+
+        output_filename = self.node.get_attribute('output_filename')
+
+        try:
+            with output_folder.open(output_filename, 'r') as handle:
+                try:
+                    result = int(handle.read())
+                except ValueError:
+                    result = None
+        except (OSError, IOError):
```
sphuber, line 238:
```suggestion
        except OSError:
```
In Python 3 we can safely only catch `OSError`

docs/source/howto/codes.rst
```diff
@@ -4,40 +4,263 @@
...
+            return self.exit_codes.ERROR_NO_RETRIEVED_FOLDER
+
+        output_filename = self.node.get_attribute('output_filename')
+
+        try:
+            with output_folder.open(output_filename, 'r') as handle:
+                try:
+                    result = int(handle.read())
+                except ValueError:
+                    result = None
```
sphuber, line 237:
```suggestion
                    return self.exit_codes.ERROR_INVALID_OUTPUT
```

docs/source/howto/codes.rst
```diff
@@ -4,40 +4,263 @@
...
+        except (OSError, IOError):
+            return self.exit_codes.ERROR_READING_OUTPUT_FILE
+
+        if result is None:
+            return self.exit_codes.ERROR_INVALID_OUTPUT
```
sphuber, line 242:
```suggestion
```

docs/source/howto/codes.rst
```diff
@@ -4,40 +4,263 @@
...
+.. code-block:: python
+
+    @classmethod
+    def define(cls, spec):
+        (...)
+        spec.exit_code(300, 'ERROR_NO_RETRIEVED_FOLDER', message='The retrieved folder data node could not be accessed.')
+        spec.exit_code(310, 'ERROR_READING_OUTPUT_FILE', message='The output file could not be read from the retrieved folder.')
+        spec.exit_code(320, 'ERROR_INVALID_OUTPUT', message='The output file contains invalid output.')
+
+As you can see, for each ``exit_code`` you need to provide an ID number that will be used to identify it, a label you can then use to reference the code in the |parse| method (``self.exit_codes.LABEL``), and a message that will give the user more information on the problem.
```
sphuber, line 258:
```suggestion
As you can see, for each ``exit_code`` you need to provide a exit status (which should be a positive integer), a label that can be used to reference the code in the |parse| method (``self.exit_codes.LABEL``), and a message that provides a more detailed information on the problem.
```

docs/source/howto/codes.rst
```diff
@@ -4,40 +4,263 @@
...
+The Topics section on :ref:`defining processes <topics:processes:usage:defining>` also contains more information on how to use exit codes.
```
sphuber, line 259:
```suggestion
The Topics section on :ref:`defining processes <topics:processes:usage:defining>` also contains more information on how to use exit codes.
```
```suggestion
The Topics section on :ref:`defining processes <topics:processes:usage:defining>` provides additional information on how to use exit codes.
```

docs/source/howto/codes.rst
```diff
@@ -4,40 +4,263 @@
...
+For the plugin user this means they will have to provide it when instantiating the calculation (as can be seen in the :ref:`respective section<how-to:codes:run>` below), whereas for the plugin developer this means they will have to make sure to manually pass its UUID from the ``self.inputs['code']`` node to the ``codeinfo.code_uuid`` property, as shown here.
+
+The other two lines are configuring how to build the running script: line 14 indicates where to redirect the standard output, whereas line 15 lists the command line arguments to be passed to the code.
+The specific combination presented here, together with the information inside of the |Code| node, results in the following script:
+
+.. code-block:: bash
+
+    #!/bin/bash
+
+    '[executable path in code node]' '[input_filename]' > '[output_filename]'
```
sphuber, line 155:
```suggestion
    '[executable path in code node]' < '[input_filename]' > '[output_filename]'
```

sphuber:
@ramirezfranciscof thanks for the update. I notice that you have resolved a number of change requests that are actually not addressed. Are you still working on these? Or was this ready for final review?

ramirezfranciscof:
Nono, I'm almost done, I'm just updating referenced line numbers and then I need to change the first paragraph (you see that's the only unresolved comment) and will push and ping you.

PR #4094, opened 2020-05-20
Increase the efficiency of the `SqlaGroup.nodes` iterator

### LINKED ISSUE #4090: Iterator of the `SqlaGroup` implementation incorrectly preloads all nodes when slicing
This makes calls like `group.nodes[:10]` which should only fetch 10 nodes and therefore run quickly, take forever in the case of big groups as it incorrectly preloads all nodes before returning only the requested slice.

sphuber:
Fixes #4090

The iterator was calling `dbnodes.__iter__()` which was violting
the lazy loading of the nodes as they are needed during iteration
leading to a huge load time for big groups even when only a slice
is demanded. For reference, on a group with roughly half a million
nodes, getting a single one through `group.nodes[0]` went from
taking a minute to a few milliseconds.

P.S.: not sure how to test this. Could think of a dedicated performance test on Jenkins, but for a regression I am not sure it is worth effort

### DIFF
--- a/aiida/orm/implementation/sqlalchemy/groups.py
+++ b/aiida/orm/implementation/sqlalchemy/groups.py
@@ -146,11 +146,10 @@ class SqlaGroup(entities.SqlaModelEntity[DbGroup], BackendGroup):  # pylint: dis
             def __init__(self, dbnodes, backend):
                 self._backend = backend
                 self._dbnodes = dbnodes
-                self._iter = dbnodes.__iter__()
                 self.generator = self._genfunction()
 
             def _genfunction(self):
-                for node in self._iter:
+                for node in self._dbnodes:
                     yield self._backend.get_backend_entity(node)
 
             def __iter__(self):


ramirezfranciscof:
So, I don't understand how this changes so much in the performance. It was after all only passing the iterator to an intermediate variable, right? Not the whole list. What am I missing?

sphuber:
I did some profiling to fully understand but it is quite complex and I also couldn't completely understand it. Calling the  `__iter__()` on the `dbnodes` attribute, which is a `sqlalchemy.orm.dynamic.AppenderQuery`, was causing additional database each time `_genfunction` was being called. In the end it does not matter much, because `dbnodes` already behaves like an iterator and so there is no need to create another wrapping generator.

ramirezfranciscof:
> was causing additional database each time

Additional database...what? Did you miss a word there or I'm not understanding something.

Also, re-testing:

> P.S.: not sure how to test this. Could think of a dedicated performance test on Jenkins, but for a regression I am not sure it is worth effort

What do you mean "for a regression"? In any case, is there a way to set up maybe a separate suite of automatic tests that runs on develop on a nightly basis (so say, from 2-6 am) instead of with every PR/commit? Maybe we could build some performance tests that way (so, the test itself doesn't seem to be the slow part but building the DB, correct? so this would have to be designed so that for this suite the DB is built only once and is used in all performance tests).

Anyways, yes, this is a two:one line change, it doesn't get any more specific than this, change aproved.

sphuber:
> Is ascii art allowed?

```
 _______  __    _  _______  _______  __   __  ______    _______  _______  _______  ______     _______  __   __  _______  __    _ 
|       ||  |  | ||       ||       ||  | |  ||    _ |  |   _   ||       ||       ||      |   |       ||  | |  ||       ||  |  | |
|    ___||   |_| ||       ||   _   ||  | |  ||   | ||  |  |_|  ||    ___||    ___||  _    |  |    ___||  |_|  ||    ___||   |_| |
|   |___ |       ||       ||  | |  ||  |_|  ||   |_||_ |       ||   | __ |   |___ | | |   |  |   |___ |       ||   |___ |       |
|    ___||  _    ||      _||  |_|  ||       ||    __  ||       ||   ||  ||    ___|| |_|   |  |    ___||       ||    ___||  _    |
|   |___ | | |   ||     |_ |       ||       ||   |  | ||   _   ||   |_| ||   |___ |       |  |   |___  |     | |   |___ | | |   |
|_______||_|  |__||_______||_______||_______||___|  |_||__| |__||_______||_______||______|   |_______|  |___|  |_______||_|  |__|
```

PR #4060, opened 2020-05-08
Docs: revamp Intro/Install 

chrisjsewell:
closes #4026
closes #4056

Issue #4050, opened 2020-05-05
Docs: Enable general CI checks


csadorf:
Reminder to revert PR #4047 before merging the `docs-revamp` branch.

sphuber:
We should also reenable the `-W` flag on the docs build to turn warnings into errors.

chrisjsewell:
Note the command line in .circleci/config.yml would then be: `sphinx-build -b html -nW --keep-going -d docs/_build/doctrees docs/source docs/_build/html`

When re-enabled, these were the noted reasons for possible failures: https://github.com/aiidateam/aiida-core/pull/4099#issuecomment-633469208

PR #4099, opened 2020-05-20
Move to CircleCI for CI Documentation builds

### LINKED ISSUE #4028: Docs: Consider using CircleCI for documentation CI
One thing that I wanted to briefly mention, although it is most probably not something I expect us to implement, is that currently using CircleCI for documentation integration testing has a big benefit over GitHub Actions, because it can provide HTML build artefacts. For example in: https://github.com/executablebooks/sphinx-panels/pull/3, you get a link that directly opens the newly built documentation for review:

image

> I'm wondering whether it would be possible to implement something similar in GitHub actions as well though

Unfortunately at the moment no. I would refer you to this issue, roughly tracking the status of this feature in Github actions: https://github.com/spatialaudio/nbsphinx/issues/361

_Originally posted by @chrisjsewell in https://github.com/aiidateam/aiida-core/pull/4011#issuecomment-622323342_

chrisjsewell:
closes #4028 

As discussed there, the main reason for this is to allow for the built documentation to be viewed during PR reviews.

This is already working on #4060, under my CircleCI account: https://app.circleci.com/pipelines/github/chrisjsewell/aiida_core?branch=docs-revamp-intro-install2, but I think I need someone else to set up an account for aiidateam.

Note you can also use this action to create a shortcut in the PR checks list, to open the `index.html`, but it's not super necessary:

```yaml
on: [status]

jobs:
  circleci_artifacts_redirector_job:
    runs-on: ubuntu-latest
    name: Run CircleCI artifacts redirector
    steps:
      - name: GitHub Action step
        uses: larsoner/circleci-artifacts-redirector-action@master
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
          artifact-path: 0/html/index.html
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  build_docs:
+  docs:
     docker:
       # see: https://circleci.com/docs/2.0/circleci-images/#python
       - image: circleci/python:3.7-stretch
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,14 +27,8 @@ jobs:
           name: Build docs to store
           # nit-picky mode, turn warnings into errors,
           # but do not stop the build on errors (so we can still inspect the doc artifacts)
-          # this needs to re re-instated when doc-revamp merged
-          # command: |
-          #   sphinx-build -b html -nW --keep-going -d docs/_build/doctrees docs/source docs/_build/html
           command: |
             sphinx-build -b html -n --keep-going -d docs/_build/doctrees docs/source docs/_build/html
-          # environment:
-            # READTHEDOCS: 'True'
-            # RUN_APIDOC: 'False'
       - store_artifacts:
           path: docs/_build/html/
           destination: html
@@ -43,4 +37,4 @@ workflows:
   version: 2
   default:
     jobs:
-      - build_docs
+      - docs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
       - run:
           name: Build docs to store
           # nit-picky mode, turn warnings into errors,
-          # but don't stop the build on errors (so we can still inspect the doc artifacts)
+          # but do not stop the build on errors (so we can still inspect the doc artifacts)
           command: |
             sphinx-build -b html -nW --keep-going -d docs/_build/doctrees docs/source docs/_build/html
           # environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,8 +27,11 @@ jobs:
           name: Build docs to store
           # nit-picky mode, turn warnings into errors,
           # but do not stop the build on errors (so we can still inspect the doc artifacts)
+          # this needs to re re-instated when doc-revamp merged
+          # command: |
+          #   sphinx-build -b html -nW --keep-going -d docs/_build/doctrees docs/source docs/_build/html
           command: |
-            sphinx-build -b html -nW --keep-going -d docs/_build/doctrees docs/source docs/_build/html
+            sphinx-build -b html -n --keep-going -d docs/_build/doctrees docs/source docs/_build/html
           # environment:
             # READTHEDOCS: 'True'
             # RUN_APIDOC: 'False'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,43 @@
+version: 2
+jobs:
+  build_docs:
+    docker:
+      # see: https://circleci.com/docs/2.0/circleci-images/#python
+      - image: circleci/python:3.7-stretch
+    steps:
+      # Get our data and merge with upstream
+      - run: sudo apt-get update
+      - checkout
+
+      - restore_cache:
+          keys:
+            - cache-pip
+
+      - run: |
+          pip install numpy==1.17.4
+          pip install --user .[docs,testing]
+
+      - save_cache:
+          key: cache-pip
+          paths:
+            - ~/.cache/pip
+
+      # Build the docs
+      - run:
+          name: Build docs to store
+          # nit-picky mode, turn warnings into errors,
+          # but don't stop the build on errors (so we can still inspect the doc artifacts)
+          command: |
+            sphinx-build -b html -nW --keep-going -d docs/_build/doctrees docs/source docs/_build/html
+          # environment:
+            # READTHEDOCS: 'True'
+            # RUN_APIDOC: 'False'
+      - store_artifacts:
+          path: docs/_build/html/
+          destination: html
+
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -4,36 +4,38 @@ on: [push, pull_request]
...
-      run: |
-        sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
-        sudo apt update
-        sudo apt install texlive-base texlive-generic-recommended texlive-fonts-recommended texlive-latex-base texlive-latex-recommended texlive-latex-extra dvipng dvidvi
+  #   - name: Install system dependencies
+  #     # remove occasionally problematic repositories we don't use anyway
+  #     run: |
+  #       sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
+  #       sudo apt update
+  #       sudo apt install texlive-base texlive-generic-recommended texlive-fonts-recommended texlive-latex-base texlive-latex-recommended texlive-latex-extra dvipng dvidvi
```
chrisjsewell, line 27: Note here that the packages for a PDF build are installed. But I'm pretty sure this action doesn't actually build latex, only html? So I haven't added them to the circleci config

chrisjsewell: Or unless there is another reason they are required?

csadorf: I think this is left-over and probably not needed, but I suggest that we just ignore that as part of this change.

sphuber:
@chrisjsewell I have just authorized CircleCI to access the `aiidateam` organization. Can you check if that did the trick for what you needed?

chrisjsewell:
> @chrisjsewell I have just authorized CircleCI to access the `aiidateam` organization. Can you check if that did the trick for what you needed?

Looks lie you also need to enable it for aiida-core as well (obviously just skip anytime it asks you to create a config.yml)

image

.github/workflows/ci.yml
chrisjsewell: removed in 832600a

csadorf:
@chrisjsewell Is it actually running the CC job? It's not showing up among the checks.

chrisjsewell:
> @chrisjsewell Is it actually running the CC job? It's not showing up among the checks.

No its not yet activated for the repo (see my comment above)

chrisjsewell:
link: https://app.circleci.com/projects/project-dashboard/github/aiidateam

csadorf:
I just activated aiida-core, but I'm not sure how to trigger the build for this particular branch.

chrisjsewell:
Ok the circleci is now triggering. Two things that cause it to fail:

1. I have setup the sphinx-build to fail on warnings, e.g. missing references, which it wasn't before. This is certainly the correct way to test the documentation, but will mean that some docs-revamp PRs will likely fail, if they are referencing other parts of the documentation that has not yet been added.
2. I note we are using [sphinx.ext.imgmath](https://www.sphinx-doc.org/en/master/usage/extensions/math.html#module-sphinx.ext.imgmath) extension, which is why actually the latex installs are necessary. I am wondering though if there is any reasoning for not using the [sphinx.ext.mathjax](https://www.sphinx-doc.org/en/master/usage/extensions/math.html#module-sphinx.ext.mathjax), which is what I normally use and means you don't need to install latex

chrisjsewell:
FYI, to see the docs, you just click on the `details` link, and select the `index.html` here:

image

sphuber:
>     1. I have setup the sphinx-build to fail on warnings, e.g. missing references, which it wasn't before. This is certainly the correct way to test the documentation, but will mean that some docs-revamp PRs will likely fail, if they are referencing other parts of the documentation that have not yet been added.

We explicitly removed the convert-warning-to-error flag for the `docs-revamp` exactly for this reason. During development there might some minor warnings that we don't want to have block PR's. Just before merging we will reinstate it and require there are no warnings as we normally do. For now we just check the warnings manually to see they are not related to the changes and will be fixed in another PR at some point.

chrisjsewell:
Ah ok, disabled warnings-to-errors in 95cb95a

.circleci/config.yml
```diff
@@ -0,0 +1,46 @@
...
+            - ~/.cache/pip
+
+      # Build the docs
+      - run:
+          name: Build docs to store
+          # nit-picky mode, turn warnings into errors,
+          # but do not stop the build on errors (so we can still inspect the doc artifacts)
+          # this needs to re re-instated when doc-revamp merged
+          # command: |
+          #   sphinx-build -b html -nW --keep-going -d docs/_build/doctrees docs/source docs/_build/html
```
sphuber, line 32: Can remove this. I have made a note in the issue that will take care of this before merging `docs-revamp` into `develop`: https://github.com/aiidateam/aiida-core/issues/4050

.circleci/config.yml
```diff
@@ -0,0 +1,46 @@
+version: 2
+jobs:
+  build_docs:
```
csadorf, line 3:
```suggestion
  docs:
```
So that we don't have to change the name of the expected check.

.circleci/config.yml
```diff
@@ -0,0 +1,46 @@
...
+          # nit-picky mode, turn warnings into errors,
+          # but do not stop the build on errors (so we can still inspect the doc artifacts)
+          # this needs to re re-instated when doc-revamp merged
+          # command: |
+          #   sphinx-build -b html -nW --keep-going -d docs/_build/doctrees docs/source docs/_build/html
+          command: |
+            sphinx-build -b html -n --keep-going -d docs/_build/doctrees docs/source docs/_build/html
+          # environment:
+            # READTHEDOCS: 'True'
+            # RUN_APIDOC: 'False'
```
sphuber, line 37: Should this be enabled? If not, best to remove it. But I guess we need `RUN_APIDOC = True` for the full build to succeed at some point no?

csadorf: I think that's the default anyways, setting it to 'False' will disable it. No setting means it will run it.

chrisjsewell: Yeh it's enabled by default, I just had these there for testing but can remove now if you want?

csadorf: I don't think it really matters, fine with me to leave this.

.circleci/config.yml
chrisjsewell: Fixed in 9bebcb7

.circleci/config.yml
chrisjsewell: fixed in 9bebcb7

chrisjsewell:
@csadorf you still need to change the expected check I'm afraid

PR #4107, opened 2020-05-25
Update sphinx extensions

chrisjsewell:
This PR adds the sphinx extensions; [sphinx-panels](sphinx-panels.readthedocs.io) and [sphinx-copybutton](https://sphinx-copybutton.readthedocs.io), and deprecates the internal ``accordion`` directive. It also adds some custom CSS, used in #4060, (a) to add specific icons to admonition blocks:

```rst
.. admonition:: Having this issue?
   :class: title-icon-troubleshoot

   Here's the fix....
```

and (b) to improve the layout of `dropdown` directive "groups"

```rst
... div:: dropdown-group

   .. dropdown:: Title

      content

   .. dropdown:: Title

      content
```


chrisjsewell:
Note, the selector for the copy-button means, by default on will be added to all code blocks, but you can turn this off for a particular blobk by:

```rst
.. code-block: console
   :class: no-copy

   $ echo "hi"
   hi
```

 `$ ` signs will also be stripped from the copy and output lines will not be included.
Also this should/will be changed when https://github.com/executablebooks/sphinx-copybutton/issues/52#issuecomment-631376338 is added, to better handle ipython code blocks (perhaps I should make an issue for that)

chrisjsewell:
The errors for the `test-install` are for pymatgen, I'm guessing that is a separate issue

PR #4109, opened 2020-05-25
Adding short note on multiple users' usage

### LINKED ISSUE #4005: Docs: How to manage multiple users

giovannipizzi:
This will be extended in the future when better multiuser support will be available

Fixes #4005


sphuber (changes requested):
Thanks @giovannipizzi . The way you describe it here, we are essentially saying that we don't support multi-user configurations. Having each user install their own instance does not count. If that is what we want to do, I think we should make this more explicit. Something like the following as a first sentence. "Setups with multiple users for a single AiiDA instance is currently not supported".

That being said, even though we don't have a formal multi-user system in place, there are certain ways to have some kind of multiple user setups. I feel it might still be useful to note these here, albeit with a fat disclaimer of the relevant caveats. Would you agree?

docs/source/howto/installation.rst
```diff
@@ -246,12 +246,18 @@ Backing up your installation
...
+Under this account it is possible to configure all the credentials to connect to remote computers.
+Using independent accounts will ensure that, for instance, the SSH credentials to connect to supercomputers are not shared with others.
 
-`#4005`_
+When data needs to be shared with colleagues, you can share data using :ref:`AiiDA's sharing functionality <how-to:data:share>`. Sharing (subsets of) the AiiDA graph can be done as often as needed.
 
+.. note::
+
+   Remember that, when sharing, the PK of your nodes will change, while instead UUIDs will always be the same.
+   Therefore, always use UUIDs when you are referring to AiiDA nodes that you are not going to use only in your AiiDA installation.
```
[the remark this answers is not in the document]
giovannipizzi: Ok

giovannipizzi:
I have accepted the suggestion, and removed the note on UUIDs vs PKs.
Instead, as discussed, for now I will not expand on ways to have multiple user-setup as I never tested them, or some of the other options might not guarantee enough security. We can always extend this section later, if we want.

This is ready to be reviewed again (and merged - I have rebased).

docs/source/howto/installation.rst
```diff
@@ -246,12 +246,15 @@ Backing up your installation
 
 Managing multiple users
 =======================
+Setups with multiple users for a single AiiDA instance is currently not supported.
```
sphuber, line 249:
```suggestion
Setups with multiple users for a single AiiDA instance are currently not supported.
```

docs/source/howto/installation.rst
```diff
@@ -246,12 +246,15 @@ Backing up your installation
...
+Instead, each AiiDA user should install AiiDA in an Unix/Windows account on their own computer or workstation.
```
sphuber, line 250:
```suggestion
Instead, each AiiDA user should install AiiDA in a Unix/Windows account on their own computer.
```

docs/source/howto/installation.rst
```diff
@@ -246,12 +246,15 @@ Backing up your installation
...
+Under this account it is possible to configure all the credentials to connect to remote computers.
```
sphuber, line 251:
```suggestion
Under this account it is possible to configure all the credentials necessary to connect to remote computers.
```

docs/source/howto/installation.rst
```diff
@@ -246,12 +246,15 @@ Backing up your installation
...
+Using independent accounts will ensure that, for instance, the SSH credentials to connect to supercomputers are not shared with others.
 
-`#4005`_
-
+Data can be shared using :ref:`AiiDA's export and import functionality <how-to:data:share>`.
```
sphuber, line 254:
```suggestion
Data can be shared between instances using :ref:`AiiDA's export and import functionality <how-to:data:share>`.
```

sphuber (changes requested):
Thanks @giovannipizzi some minor changes and then I will merge this

csadorf (commented):
This looks fine and I agree with @sphuber 's comments. I'll approve once those comments are addressed and the merge conflict is resolved.

giovannipizzi:
Should be done now

csadorf (approved):

csadorf:

sphuber (approved):

Issue #3987, opened 2020-05-01
Docs: How to run external codes


sphuber:

sphuber:
Fixed in PR #4114

PR #4114, opened 2020-05-25
Docs: How to run external codes

ramirezfranciscof:
Added content to the section on how to run codes. Indeed this ended up being much more direct than the other section. Here I am actually worried you might ask me to add more stuff to it.


docs/source/howto/codes.rst
```diff
@@ -45,8 +45,92 @@ Design guidelines
 Running external codes
 ======================
 
-`#3987`_
+The process of running an external code through AiiDA involves loading and instantiating the appropriate ``CalcJob`` wrapping class for the code, filling the inputs necessary, and then running or submitting the constructed object.
```
sphuber, line 48: "The process of running an external code .... involves" is pretty verbose language for a how-to.
How about something more direct and to the point:
> To run an external code with AiiDA, you need a :ref:`plugin <correct link to more info>` for it.
> The plugin's job is to transform the input *nodes* to the input files that the code expects.
> If the plugin for your code already exists (see the `plugin registry <http://registry>`_ what is available) and you have installed the :ref:`plugin package <ref>`, you can load it using the ``CalculationFactory``.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,92 @@ Design guidelines
...
 
+If the ``CalcJob`` has been registered in AiiDA (see section :ref:`on how to register plugins<how-to:plugins:entrypoints>`) you can use the ``CalculationFactory``:
+
+.. code-block:: python
+
+    from aiida.plugins import CalculationFactory
+    calculation_builder = CalculationFactory('arithmetic.add').get_builder()
```
sphuber, line 55:
```suggestion
    from aiida.plugins import CalculationFactory
    calculation_class = CalculationFactory('arithmetic.add')
```
I would assign the class explicitly to make it clear that that is what the factory returns.
I would also not use the builder straight away but go for the simple dictionary syntax

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,92 @@ Design guidelines
...
+
+Otherwise, you can just load the class and get the builder directly:
+
+.. code-block:: python
+
+    from aiida.calculations.plugins.arithmetic.add import ArithmeticAddCalculation
+    calculation_builder = ArithmeticAddCalculation.get_builder()
```
sphuber, line 62: This will very often lead to problems when they start submitting to the daemon and I would not mention it here as it is an unnecessary alternative.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,92 @@ Design guidelines
...
+
+Then you need to start filling out the builder with all the inputs that you need to specify to run your calculation.
```
sphuber, line 64: Should add a line here saying:
> For the purpose of this example, we will use the `arithmetic.add` plugin, which is a plugin for `bash` to sum two integers

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,92 @@ Design guidelines
...
+    # Specific for ArithmeticAdd
+    spec.input('x', valid_type=(orm.Int, orm.Float), help='The left operand.')
+    spec.input('y', valid_type=(orm.Int, orm.Float), help='The right operand.')
+
+    # General for all CalcJobs
+    spec.input('code', valid_type=orm.Code, help='The `Code` to use for this job.')
+    spec.input('metadata.options.resources', valid_type=dict, required=True, validator=validate_resources,
+        help='Set the dictionary of resources to be used by the scheduler plugin, like the number of nodes, '
+             'cpus etc. This dictionary is scheduler-plugin dependent. Look at the documentation of the '
+             'scheduler for more details.')
```
sphuber, line 79: Instead of this I would suggest to use `verdi plugin list aiida.calculations arithmetic.add`
This way we don't have to worry about updating this when it changes, and it will also apply to any other calculation plugin

sphuber: Would also remove the mention of resources, which will be simplified by Marnik's changes that will soon be merged in `develop`.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,92 @@ Design guidelines
...
+
+    code_node = orm.load_node(100)
+    num1_node = orm.Int(17)
+    num2_node = orm.Int(11)
+
+    calculation_builder.code = code_node
+    calculation_builder.x = num1_node
+    calculation_builder.y = num2_node
+
+The last input (line 7) is not a node but a dictionary containing information about the resources required from the remote machine to run the code.
```
sphuber, line 93: This line is incorrect I think. You are not adding the metadata here, but in the next block

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,92 @@ Design guidelines
...
+It will be stored as a property of the ``CalcJob`` node, and should at least contain the following two keys:
+
+.. code-block:: python
+
+    calculation_builder.metadata.options.resources = {
+        'num_machines': 1,
+        'tot_num_mpiprocs': 1,
+    }
```
sphuber, line 101: as said before, I think we should remove this. These are not the correct settings for all schedulers and soon this will be set automatically. This will simplify it a lot.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,92 @@ Design guidelines
...
+    from aiida.engine import submit
+    calculation_node = submit(calculation_builder)
+
+As the variable name suggests, the return value for the submit method is the calculation node that is stored in the database.
+In this case the underlying calculation is not guaranteed to have finished, so it is often convenient to keep track of it using the verdi command line interface:
+
+.. code-block:: bash
+
+    $ verdi process list
+
```
sphuber, line 125: Should probably create a subsection for "Performing a dry-run"

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,92 @@ Design guidelines
...
+Additionally, you might want to check and verify your inputs before actually running or submitting a calculation.
+You can do so by specifying to use a ``dry_run``, which will create the running folder in the current directory (``submit_test/[date]-0000[x]``) so the user can inspect the integrity of the inputs generated:
```
sphuber, line 127:
```suggestion
You can do so by specifying to use a ``dry_run``, which will create all the input files in a local directory (``submit_test/[date]-0000[x]``) so you can inspect them before actually launching the calculation:
```

docs/source/howto/codes.rst
ramirezfranciscof: Ok, so, the drawback of removing the `resources` is that in this example only nodes would be provided, which might make people think you can only pass nodes to the calculations. Are we ok with this? Should I maybe add a change to one of the optional parameters (like the `output_filename`)?

docs/source/howto/codes.rst
ramirezfranciscof: I guess it is no longer relevant if I'm removing the metadata, but the line "The last input (line 7)" refers not to the last code block, but the previous one. I was hoping it would have been a bit more clear looking at the compiled doc because it was the only code block with numbers in the section, but it is true that it is not ideally phrased.

docs/source/howto/codes.rst
sphuber: I think that is actually ok for this introductory how-to. We could make another on in the future, something like "How to change scheduler specific inputs", which will then provide a natural way to introduce `metadata` inputs.

csadorf:
Resolves #3987 .

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,84 @@ Design guidelines
 Running external codes
 ======================
 
-`#3987`_
+To run an external code with AiiDA, you will need to use an appropriate :ref:`calculation plugin <topics:plugins>` that knows how to take transform the input nodes into the input files that the code expects, copy everything in the code's machine, run the calculation and retrieve the results.
```
csadorf, line 48:
```suggestion
To run an external code with AiiDA, you will need to use an appropriate :ref:`calculation plugin <topics:plugins>` that knows how to transform the input nodes into the input files that the code expects, copy everything in the code's machine, run the calculation and retrieve the results.
```

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,84 @@ Design guidelines
...
+This plugin would have already been created, either by you (see section on :ref:`interfacing external codes <how-to:codes:plugin>`) or by other AiiDA plugin-developers (available packages are listed in the `plugin registry <https://aiidateam.github.io/aiida-registry/>`_), and has been properly installed in your environment.
+You can check which ones you have currently available by running the ``verdi plugin`` command:
```
csadorf, line 50:
```suggestion
You can check which ones you have currently available:
```

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,84 @@ Design guidelines
...
 
+.. code-block:: bash
+
+    $ verdi plugin list aiida.calculations
+
+For the purpose of this example, we will use the ``arithmetic.add`` plugin, which is a plugin for bash to sum two integers that comes pre-installed with ``aiida-core``.
```
csadorf, line 56:
```suggestion
As an example, we will show how to use the ``arithmetic.add`` plugin, which is a pre-installed plugin implemented in `bash` to sum two integers.
```
The how-to is not an example, but it can utilize examples.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,84 @@ Design guidelines
...
+You can access it through the use of the ``CalculationFactory``:
```
csadorf, line 57:
```suggestion
You can access it with the ``CalculationFactory``:
```

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,84 @@ Design guidelines
...
+
+.. code-block:: python
+
+    from aiida.plugins import CalculationFactory
+    calculation_class = CalculationFactory('arithmetic.add')
+
+We will obviously also need to provide the inputs for the code when running the calculation.
```
csadorf, line 64:
```suggestion
Next, we provide the inputs for the code when running the calculation.
```

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,84 @@ Design guidelines
...
+You can also use ``verdi plugin`` to check this and other relevant information for specific plugins:
```
csadorf, line 65:
```suggestion
Use ``verdi plugin`` to determine what inputs a specific plugin expects:
```

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,84 @@ Design guidelines
...
+
+.. code-block:: bash
+
+    $ verdi plugin list aiida.calculations arithmetic.add
```
csadorf, line 69: Please add the output here as well. We don't expect users to run these commands to be able to follow the example.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,84 @@ Design guidelines
...
+
+You will see that 3 inputs nodes are required: two containing the values to add up (``x``, ``y``) and one containing information about the specific code to execute (``code``).
+If you already have these nodes in your database, you can get them by :ref:`querying for them <how-to:data:finding-data>` or using ``orm.load_node(<PK>)``.
+Otherwise, you will need to create them as shown below (note that you `will` need to already have the ``localhost`` computer configured):
```
csadorf, line 73: We will need to provide a link where it is described how to do that or add a *dropdown* that shows this.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,84 @@ Design guidelines
...
+To provide these as inputs to the calculations, we will now use the ``builder`` object that we can get from the class:
+
+.. code-block:: python
+
+    calculation_builder = calculation_class.get_builder()
+    calculation_builder.code = code_node
+    calculation_builder.x = numx_node
+    calculation_builder.y = numy_node
+
+Now everything is in place and ready to perform the calculation, which can be done in two different ways.
```
csadorf, line 91: I'm wondering if we can hand-off to a different section here. How to run and submit calculations is already described in a different how-to.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,84 @@ Design guidelines
...
+Performing a dry-run
+--------------------
+
+Additionally, you might want to check and verify your inputs before actually running or submitting a calculation.
+You can do so by specifying to use a ``dry_run``, which will create all the input files in a local directory (``submit_test/[date]-0000[x]``) so you can inspect them before actually launching the calculation:
+
+.. code-block:: python
+
+    calculation_builder.metadata.dry_run = True
+    calculation_builder.metadata.store_provenance = False
```
csadorf, line 124: Quick question: Does `dry_run = True` not imply `store_provenance = False`?

docs/source/howto/codes.rst
ramirezfranciscof: Mmm, but from what I understood from @sphuber 's [previous review](https://github.com/aiidateam/aiida-core/pull/4114#discussion_r430023980), the idea was to not include the actual information in order for it to be easier to maintain. Maybe I misunderstood that?

docs/source/howto/codes.rst
ramirezfranciscof: You mean setting the localhost? Ok, yes, I agree. Would this be on the "Advanced Installation" section? I remember that when we were discussin the tutorial part of the assumptions was that they would have already had this set up, that's why I assumed it here.

docs/source/howto/codes.rst
ramirezfranciscof: Wait, what? Where? I thougth this was the how-to on running calculations.

docs/source/howto/codes.rst
ramirezfranciscof: @sphuber I do not know this, honestly I just used the same two lines that were present in the aiida tutorial of 2019.

sphuber: No, dry run means the job does not actually get submitted and starts running, but simply creates the input files in a local folder. It will still create a node in the database and link the inputs up. If you do not want this (which is likely if you are running a dry run) you need to set it

docs/source/howto/codes.rst
csadorf: That's not how I understand that comment and I was missing the output when I read through this how-to.

docs/source/howto/codes.rst
csadorf: I think it hasn't been merged yet. So feel free to just ignore this for now. I'll try to add meaningful links to the relevant section after everything has been merged.

docs/source/howto/codes.rst
csadorf: Seems like "dry-run" is a bit of a misnomer then, but anyways, doesn't really affect the documentation.

docs/source/howto/codes.rst
csadorf: Honestly, I might have gotten confused here. Let's leave this and I'll double check any potential duplication after everything is merged.

csadorf:
@ramirezfranciscof Is this ready for another round of review? If yes, please re-request it, thx.

ramirezfranciscof:
@csadorf Hey, not yet, I was waiting on the replies for the comments, there is one left I need to address.

docs/source/howto/codes.rst
ramirezfranciscof: Should we add a reminder of this...somewhere?

csadorf: I'll put it on the project board.

ramirezfranciscof:
Now it is ready =D

docs/source/howto/codes.rst
sphuber: why do you think it is a misnomer?

docs/source/howto/codes.rst
sphuber: The idea was indeed to not have a copy of the spec here in case the implementation changes and the documentation goes out of date. Pasting the output of `verdi plugin list` has the same downside, but I can see why it is important to show it, so fine for me to add it. Maybe don't show the exit codes as they are not necessary and just do description, inputs and outputs.

Anyway, using this over copying source code is still better as they can then execute the command themselves to try and they can apply it to any other plugin, for example the one they would actually like to launch.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,90 @@ Design guidelines
...
+Now everything is in place and ready to perform the calculation, which can be done in two different ways.
+The first one is blocking and will return a dictionary containing all the output nodes (keyed after their label, so in this case these should be: "remote_folder", "retrieved" and "sum") that you can safely inspect and work with:
+
+.. code-block:: python
+
+    from aiida.engine import run
+    output_dict = run(calculation_builder)
+    sum_node = output_dict['sum']
+
+The second one is non blocking, as you will be submitting it to the daemon and immediately getting control in the interpreter.
```
csadorf, line 106:
```suggestion
The second one is non blocking, as you will be submitting it to the daemon and control is immediately returned to the interpreter.
```

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,90 @@ Design guidelines
...
+The return value in this case is the actual calculation node that is stored in the database.
```
csadorf, line 107:
```suggestion
The return value in this case is the calculation node that is stored in the database.
```
I'm not sure what we try to convey here with "actual".

docs/source/howto/codes.rst
csadorf: Because "dry-run" to me implies that nothing is actually going to happen, i.e., it is perfectly safe to execute without any effects. But I see that in this context it might make sense.

ramirezfranciscof:
@csadorf So, not sure if with the re-request suffices or I also needed to comment-ping you, so I'll do it just in case: changes applied, ready to review / squash in.

sphuber:
@ramirezfranciscof I will try to give a review this afternoon

csadorf:
> @csadorf So, not sure if with the re-request suffices or I also needed to comment-ping you, so I'll do it just in case: changes applied, ready to review / squash in.

Just re-requesting is sufficient.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,90 @@ Design guidelines
...
+
+    $ verdi plugin list aiida.calculations
+
+As an example, we will show how to use the ``arithmetic.add`` plugin, which is a pre-installed plugin implemented in `bash` to sum two integers.
+You can access it with the ``CalculationFactory``:
+
+.. code-block:: python
+
+    from aiida.plugins import CalculationFactory
+    calculation_class = CalculationFactory('arithmetic.add')
```
csadorf, line 62: Just making sure, this is actually a class, right? And not an instance of a class, in which case it should be:
```suggestion
    calculation = CalculationFactory('arithmetic.add')
```

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,90 @@ Design guidelines
...
+You will see that 3 inputs nodes are required: two containing the values to add up (``x``, ``y``) and one containing information about the specific code to execute (``code``).
+If you already have these nodes in your database, you can get them by :ref:`querying for them <how-to:data:finding-data>` or using ``orm.load_node(<PK>)``.
+Otherwise, you will need to create them as shown below (note that you `will` need to already have the ``localhost`` computer configured):
+
+.. code-block:: python
+
+    from aiida import orm
+    code_node = orm.Code(remote_computer_exec=[localhost, '/bin/bash'])
+    numx_node = orm.Int(17)
+    numy_node = orm.Int(11)
```
csadorf, line 86: In other sections we have not used the suffix `_node` so far. I'd prefer to drop that everywhere.

docs/source/howto/codes.rst
ramirezfranciscof: Tis a class. Proof:
```
In [1]: from aiida.plugins import CalculationFactory
In [2]: import inspect
In [3]: calculation_class = CalculationFactory('arithmetic.add') 
In [4]: inspect.isclass(calculation_class)                                                                              
Out[4]: True
```

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,90 @@ Design guidelines
...
-`#3987`_
+To run an external code with AiiDA, you will need to use an appropriate :ref:`calculation plugin <topics:plugins>` that knows how to transform the input nodes into the input files that the code expects, copy everything in the code's machine, run the calculation and retrieve the results.
+This plugin would have already been created, either by you (see section on :ref:`interfacing external codes <how-to:codes:plugin>`) or by other AiiDA plugin-developers (available packages are listed in the `plugin registry <https://aiidateam.github.io/aiida-registry/>`_), and has been properly installed in your environment.
+You can check which ones you have currently available:
 
+.. code-block:: bash
+
+    $ verdi plugin list aiida.calculations
+
+As an example, we will show how to use the ``arithmetic.add`` plugin, which is a pre-installed plugin implemented in `bash` to sum two integers.
```
sphuber, line 56:
```suggestion
As an example, we will show how to use the ``arithmetic.add`` plugin, which is a pre-installed plugin that uses the `bash shell<https://www.gnu.org/software/bash/>`_ to sum two integers.
```

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,90 @@ Design guidelines
 Running external codes
 ======================
 
-`#3987`_
+To run an external code with AiiDA, you will need to use an appropriate :ref:`calculation plugin <topics:plugins>` that knows how to transform the input nodes into the input files that the code expects, copy everything in the code's machine, run the calculation and retrieve the results.
+This plugin would have already been created, either by you (see section on :ref:`interfacing external codes <how-to:codes:plugin>`) or by other AiiDA plugin-developers (available packages are listed in the `plugin registry <https://aiidateam.github.io/aiida-registry/>`_), and has been properly installed in your environment.
```
sphuber, line 49:
```suggestion
You can check the `plugin registry <https://aiidateam.github.io/aiida-registry/>`_ to see if a plugin already exists for the code that you would like to run.
If that is not the case, you can :ref:`develop your own <how-to:codes:plugin>`.
After you have installed the plugin, you can start running the code through AiiDA.
```
Just wondering if we need a quick link for "installed the plugin" to a section on how to install plugins. Not sure if we already have it.

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,90 @@ Design guidelines
...
+You can check which ones you have currently available:
```
sphuber, line 50:
```suggestion
To check which calculation plugins you have currently installed, run:
```

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,90 @@ Design guidelines
...
+    (...)
+        Inputs:
+               code:  required  Code        The `Code` to use for this job.
+                  x:  required  Int, Float  The left operand.
+                  y:  required  Int, Float  The right operand.
+    (...)
+
+You will see that 3 inputs nodes are required: two containing the values to add up (``x``, ``y``) and one containing information about the specific code to execute (``code``).
+If you already have these nodes in your database, you can get them by :ref:`querying for them <how-to:data:finding-data>` or using ``orm.load_node(<PK>)``.
+Otherwise, you will need to create them as shown below (note that you `will` need to already have the ``localhost`` computer configured):
```
sphuber, line 79: I think we really need at a minimum a link to a How-To configure a computer. I don't think it exists yet, but you can already create the link reference and stub, and open the corresponding issue

docs/source/howto/codes.rst
```diff
@@ -45,8 +45,90 @@ Design guidelines
...
+
+    from aiida.engine import submit
+    calculation = submit(calculation_builder)
+
+Note that, although you have access to the node, the underlying calculation `process` is not guaranteed to have finished when you get back control in the interpreter.
+In order to keep track of it you can use the verdi command line interface:
+
+.. code-block:: bash
+
+    $ verdi process list
```
sphuber, line 119: Maybe a link here to the topic on monitoring processes: https://aiida.readthedocs.io/projects/aiida-core/en/docs-revamp/topics/processes/usage.html#monitoring-processes

docs/source/howto/codes.rst
ramirezfranciscof: Yeah, I thougth the same, but there is currently no such section. Should it be part of "How to manage your installation" maybe?

docs/source/howto/codes.rst
ramirezfranciscof: Should it be part of "How to manage your installation"?

sphuber: I would more be tempted to put it in this same How to, because really the only reason you need to configure a computer is to run `CalcJobs`. Not sure where, maybe after "Interfacing external codes" for now? Something like "Configuring a remote computer"

docs/source/howto/codes.rst
sphuber: Yes make a section stub there called "Installing plugins" and create the corresponding issue, if you will please. Put it before the "Tuning performance" section, for now

docs/source/howto/codes.rst
ramirezfranciscof: ["Configuring remote resources"](https://github.com/aiidateam/aiida-core/issues/4123) sounds good?

sphuber: I think "resources" is too broad in this case, because it does not necessarily convey that we are talking about computers. We could specify "Configuring remote computing resources" but in that case I think "Configuring remote computers" is better, as it is clearer and more concise, while saying the exact thing

ramirezfranciscof: Ok, changed. Ready for final review now.

PR #4118, opened 2020-05-26
Make message broker server settings configurable

### LINKED ISSUE #1074: Add authentication to RMQ setup
Currently we use the default user (guest:guest) for RMQ connections, this means that  someone could put any jobs in the queue which would be picked up by whichever daemon is listening to that queue.

We should add a user and password to the RMQ account, see:

https://www.rabbitmq.com/rabbitmqctl.8.html#User_Management

This could probably just piggy back off the database user and password but will need to be done during (quick)setup.

sphuber:
Fixes #1074 

This PR makes the configuration of the message broker (currently RabbitMQ) configurable.
So far I have implemented:

 * Create `Communicator` URL from current profile instead of hardcoded
 * Extend the `Profile` class to contain message broker information
 * Add options to `verdi quick/setup` to configure the message broker
 * Add configuration migration to 

Things that remain to be done:

 - [ ] Update the documentation
 - [x] Verify that current configuration options are sufficient for full TLS/SSL setup
 - [ ] Do docker images need to be updated?
 - [ ] Should we do additional testing, beyond our unit tests, that using multiple users works? I tried myself locally to create another user with password, but not yet over SSH with SSL.


sphuber:
Pinging @muhrin for comments and ideas, and @yakutovicha @chrisjsewell for comment on what should be done on the Docker images

chrisjsewell:
Yes that looks good cheers. The main thing with respect to a docker compose / multi-container system, is the ability to set the `broker_host`, which this fixes. For the standard (monolithic) Docker image nothing should need changing, since it will just use the defaults.

aiida/manage/external/rmq.py
```diff
@@ -16,48 +16,57 @@
...
-    url = _RMQ_URL
-
-    if heartbeat_timeout is None:
-        heartbeat_timeout = _RMQ_HEARTBEAT_TIMEOUT
-
-    if heartbeat_timeout is not None:
-        url += '?heartbeat={}'.format(heartbeat_timeout)
-
-    return url
+    return '{protocol}://{username}:{password}@{host}:{port}?heartbeat={heartbeat_timeout}'.format(
```
muhrin, line 58: You might want to consider using [urlparse](https://docs.python.org/3.5/library/urllib.parse.html?highlight=urlparse#urllib.parse.urlunparse) for this.

sphuber: Changed it to use `urlparse`. Should note though that the `{username}:{password}@{host}:{port}` part is not formatted by it though. It understands that entire part as the `netloc` and so it us up to us to format it. So there I still use string interpolation as before

sphuber:
Superseded by PR #4341

Issue #4019, opened 2020-05-01
Docs: Topic database


sphuber:

mbercx:
In order to slim down the how-to on finding data (#3996), we're moving the more advanced functionality of the `QueryBuilder` to the topics section. I think it would make sense to make a section on this page on "Advanced Queries", which includes more advanced features of the appender method and information on the `queryhelp`.

I'd also add a more extensive discussion on certain database related terminology like *entities*, *vertices*, *query path* etc.

@csadorf what do you think?

sphuber:
Another option would be to put it as a sub section of the "Provenance" topic? There are already quite a few top-level sections there, and maybe we should be a bit careful of letting this number grow uncontrolledly

mbercx:
I agree with limiting the number of top-level sections. What I meant was adding the advanced queries as a sub-section in the "Database" topic, sorry if that wasn't clear.

sphuber:
> I agree with limiting the number of top-level sections. What I meant was adding the advanced queries as a sub-section in the "Database" topic, sorry if that wasn't clear.

Haha, actually, it is me who is the idiot here. Clearly didn't read the context properly and realize in what issue this was being posted. I think both Database or Provenance could work.

mbercx:
I think the "Database" topic makes sense as this would also be a good place to explain the database terminology used by the `QueryBuilder`. In the "Provenance" topic, concepts are explained via nodes, links, etc. So it might be confusing to suddenly end with a sub-section that talks about entities, vertices, edges, etc.

sphuber:
I see, but the concepts of the provenance graph very much apply to querying. In order to effectively use the joining of entities in the query builder, one should understand how these are linked through edges in the provenance graph. The way I see it, the `QueryBuilder` is exactly built to explore the provenance graph. A lot of its functionality and terminology only exists because the data it is querying is represented by a graph. There are plenty of other data structures that you can store in a database that are not necessarily best represented by a graph. The only connotation is that the `QueryBuilder` also allows to query for data that is not explicitly represented in the graph by nodes, but tangentially, such as the `User` of a node, or its `Computer`.

@csadorf @giovannipizzi what do you think?
Just to be clear, I don't want to make this a hugely contentious point, as I would be fine with both, but think it is interesting to discuss it a bit

csadorf:
[...]

PR #4132, opened 2020-05-29
Docs: Add advanced querying section to database topics

mbercx:
See #4019.

* Adds the more advanced elements of the QueryBuilder to the topics section
on databases.

* Adds the reference tables for the operators, relationships and cast types.

* Adds the explanation of the queryhelp.

### DIFF
--- a/docs/source/howto/data.rst
+++ b/docs/source/howto/data.rst
@@ -285,7 +285,7 @@ Suppose you want to query for calculation job nodes in your database:
 
 .. code-block:: python
 
-    from aiida.orm.querybuilder import QueryBuilder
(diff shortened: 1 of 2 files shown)


mbercx:
Note that this PR does not complete the topics section on databases, it just moves in the advanced sections on the appender method and the `queryhelp`. 

I've adjusted the start of the `queryhelp` section a bit to highlight one of its most important use cases (at least for me): the ability to store and reload your query into/from a JSON file. The rest of the section could still do with some improvements I think, but those can wait until we add the rest of the database topics section (if we leave the `QueryBuilder` sections here).

PR #4138, opened 2020-05-29
Docs: move visualizing graphs to howto

chrisjsewell:


chrisjsewell:
Mostly a move, but...
I had to make a small "hotfix" to code 😬 because @ramirezfranciscof broke it in 47cfe3400a7f5ef863e2a10b88afa6187459bb96 😛 
Also I had to copy the export file in to the docs, so people can download it, as it has been moved to the test folder (was previously within the package)

PR #4144, opened 2020-06-03
`QueryBuilder`: fix type string filter generation for `Group` subclasses

### LINKED ISSUE #4143: `QueryBuilder.append` applies incorrect type string filter for `Group` subclasses
The logic that defines the `type_string` filter based on the `Group` subclass is incorrect and for example, `type_strings` that should be `pseudo.family` will become `seudo.family`. The culprit is: https://github.com/aiidateam/aiida-core/blob/2a0d11e70577c4921edd5f72a0b286f0932b0abb/aiida/orm/querybuilder.py#L285

sphuber:
Fixes #4143 

The `aiida.orm.querybuilder.get_group_type_filter` function that is
responsible for generating the correct filters for the `type_string`
when a `Group` or subclass of it is appended, contained a bug. To
generate the correct type string, it should strip the `group.` prefix
from the provided ORM classifier. This was incorrectly done with the
string method `lstrip`, which does not strip an entire prefix as is, but
will strip any characters that are matched starting from the left. This
would cause extra characters to be erroneously stripped if the rest of
the type string started with any letters in the sequence `group.`. For
example `group.pseudo.family` would be stripped to `seudo.family`. The
solution is to use the first N characters where N is the length of the
prefix.

### DIFF
--- a/aiida/orm/querybuilder.py
+++ b/aiida/orm/querybuilder.py
@@ -282,7 +282,7 @@ def get_group_type_filter(classifiers, subclassing):
     """
     from aiida.common.escaping import escape_for_sql_like
 
-    value = classifiers['ormclass_type_string'].lstrip(GROUP_ENTITY_TYPE_PREFIX)
+    value = classifiers['ormclass_type_string'][len(GROUP_ENTITY_TYPE_PREFIX):]
 
     if not subclassing:
         filters = {'==': value}
--- a/tests/orm/test_querybuilder.py
+++ b/tests/orm/test_querybuilder.py
@@ -114,6 +114,22 @@ class TestQueryBuilder(AiidaTestCase):
         self.assertEqual(classifiers['ormclass_type_string'], 'process.calculation.calcjob.CalcJobNode.')
         self.assertEqual(classifiers['process_type_string'], 'aiida.calculations:arithmetic.add')
 
+    def test_get_group_type_filter(self):
+        """Test the `aiida.orm.querybuilder.get_group_type_filter` function."""
+        from aiida.orm.querybuilder import get_group_type_filter
+
+        classifiers = {'ormclass_type_string': 'group.core'}
+        self.assertEqual(get_group_type_filter(classifiers, False), {'==': 'core'})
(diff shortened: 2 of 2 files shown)


ramirezfranciscof:
My only question would be if by this point in the code is it guaranteed that the string will start with that prefix and you are not just stripping part of the type away? And this is backwards compatible with older versions of this label?

sphuber:
> My only question would be if by this point in the code is it guaranteed that the string will start with that prefix and you are not just stripping part of the type away?

Yes, this should be guaranteed because the `QueryBuilder` is also responsible for prepending the prefix elsewhere:
https://github.com/aiidateam/aiida-core/blob/690025e8fd77d76e17cd7bbe41a3672c2b7b938f/aiida/orm/querybuilder.py#L97
The code is a mess overall, but it will require a lot of work to refactor it all in to something more comprehensible and less error prone.

> And this is backwards compatible with older versions of this label?

There is no question of backwards compatibility here. This was an internal bug in the `QueryBuilder` logic that was introduced in `1.2.0` when support for `Group` subclassing was added. This should not affect any user code, except solve a bug

ramirezfranciscof:
> There is no question of backwards compatibility here. This was an internal bug in the QueryBuilder logic that was introduced in 1.2.0 when support for Group subclassing was added. This should not affect any user code, except solve a bug

No, what I meant is that perhaps this was using `lstrip` to take into account the possibility of the user having older labels that may not include the `group.` prefix, but I guess that if you say this is not intrinsecal to the group but added by the querybuilder then the point is moot.

sphuber:
> No, what I meant is that perhaps this was using `lstrip` to take into account the possibility of the user having older labels that may not include the `group.`

It was using `lstrip` because I was being a Python-amateur :sweat_smile: .

Thanks for the review :+1:

greschd:
> 
> It was using `lstrip` because I was being a Python-amateur 😅 .

That is common enough that it got its own PEP: https://www.python.org/dev/peps/pep-0616/ 😂

sphuber:
Python 3.9 here we go ! :rocket:

Issue #3764, opened 2020-02-17
Import / Export without empty directories


PhilippRue:
For some nodes AiiDA generates empty directories in the repository.
In my case I removed some of these to reduce the number of inodes I have. The database export and import now poses a problem since it runs into these (false positive) exception:
https://github.com/aiidateam/aiida-core/blob/12f96415e48a43fa43f6b2493101cb5e8b27108a/aiida/tools/importexport/dbexport/__init__.py#L500
and 
https://github.com/aiidateam/aiida-core/blob/12f96415e48a43fa43f6b2493101cb5e8b27108a/aiida/tools/importexport/dbimport/backends/django/__init__.py#L392

It would probably be good to not require empty directories to be included in import/export files.

ltalirz:
@sphuber Let's discuss whether it perhaps makes sense to - until the repo migration - add a "lazy" node folder creation that only creates node folders when something should be put inside.

CasperWA:
I agree - and it should be a fairly easy(TM) fix in the end. I can try and knock it out.

ltalirz:
One comment by @sphuber : Since we are currently not tracking the content of folders, this change would mean that we are no longer able to detect whether the folder was (e.g. accidentally) deleted.

I.e. we should weigh whether the reduction in inodes is worth giving up this consistency check (which is of course quite basic - e.g. missing files are anyhow not detected).

Perhaps @giovannipizzi  can comment

ltalirz:
This will be addressed by https://github.com/aiidateam/AEP/pull/21

chrisjsewell:
Heya, I've marked this a being closed by: #3445.
This is because with this new repository system there will actually no longer be a folder system on disk; files are stored in a compressed format (indexed by hash), with the node's database entry containing a virtual mapping of the actual file/folder structure.

> Let's discuss whether it perhaps makes sense to - until the repo migration - add a "lazy" node folder creation that only creates node folders when something should be put inside.

I very much doubt anyone's going to have time to do this now, it'll just be wasted effort

PR #4148, opened 2020-06-04
Add repository issue templates

### LINKED ISSUE #4121: Add Issue Templates for GitHub
Along the lines of: https://github.com/executablebooks/MyST-Parser/tree/master/.github/ISSUE_TEMPLATE

chrisjsewell:
As the title suggests! I have also copied them to this repository, so you can see how they look: https://github.com/chrisjsewell/issues_test/issues/new/choose

closes #4121

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,42 @@
+---
+name: Bug report
+about: Report a bug you have encountered 
+title: ''
+labels: type/bug
+assignees: ''
+
+---
+
+<!-- Before raising an issue, it is suggested that you first check out the: -->
+
+- [ ] [AiiDA Troubleshooting Documentation](https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/troubleshooting.html)
+- [ ] [AiiDA Users Forum](https://groups.google.com/forum/#!forum/aiidausers)
+
+### Describe the bug
+
+<!-- A clear and concise description of what the bug is. -->
+
+### Steps to reproduce
+
+Steps to reproduce the behavior:
+
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+### Expected behavior
+
+<!-- A clear and concise description of what you expected to happen. -->
+
+### Your environment
+
+- Operating system [e.g. Linux]:
+- Python version [e.g. 3.7.1]:
+- aiida-core version [e.g. 1.2.1]:
+
+Other relevant software versions, e.g. Postres & RabbitMQ
+
+### Additional context
+
+<!-- Add any other context about the problem here. -->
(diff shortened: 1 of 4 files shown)


greschd:
I'd be in favor of adding a "blank" template, for cases which don't neatly fall into one of the categories.

For example, #3764 is a bug report / feature request where the user has already identified the root cause of the issue. In that case, I think adding all the environment / steps to reproduce / etc. of the bug report is just unnecessary noise.

Of course you can just delete the template once selected, but it creates a mental barrier to doing so.

chrisjsewell:
> I'd be in favor of adding a "blank" template, for cases which don't neatly fall into one of the categories.

Note, there is the option for this at the bottom of the form:

image

greschd:
Ah, yeah.. guess I missed it because it doesn't get that much screen real estate. Anyway, in that case I think it's fine as is.

chrisjsewell:
Yes I imagine when any of us open an issue, we will probably just use the blank template, in which case this unnecessary boiler plate. But for new users I think it could be quite useful, and allows us to also point them towards other places they can look first; like the troubleshooting documentation and the mailing list.
But anyway, I will leave it to @giovannipizzi and @sphuber to decide if they think it is worthwhile 😁

sphuber:
Is there anyway to test and see what this looks like? In principle I am not opposed but so far we have had very little issues from external users that are not clear enough, so currently I don't feel this is really necessary. This might of course change in the future

chrisjsewell:
> Is there anyway to test and see what this looks like?

I don't know of a way to preview directly on this repo, but as I mention above, this link is a direct replica of what you will see:
https://github.com/chrisjsewell/issues_test/issues/new/choose

sphuber:
So for us to create an issue we would have an intermediate page where we need to click on one button and then remove the context. This is not too much of a hassle, but I doubt that the benefits really outweigh the downsides for now, simply since it is mostly us creating the majority of issues and we don't really need to be that specific oftentimes

chrisjsewell:
> click on one button and then remove the context. 

Yes it's essentially an extra click, but you don't need to remove anything just click on the bottom blank issue link:

image

> but I doubt that the benefits really outweigh the downsides for now

I'm not fussed either way, just wanted to bring it up as a suggestion

sphuber:
> I'm not too fussed either way, just wanted to bring it up as a suggestion

Cheers for that. I think it is not necessary right now, but I don't mind adding it either if others think it is a useful addition.

sphuber:
OK, in that case we will add it for now. If we get reports that is disturbs more than that it helps, we can remove it

chrisjsewell:
> OK, in that case we will add it for now. If we get reports that is disturbs more than that it helps, we can remove it

👍 
I've personally found it helpful on repos, especially when getting lots of "new users" raising issues. But if it becomes a hindrance to developers, then by all means drop it

PR #4154, opened 2020-06-05
Reinstate broadcast subscription of workchain for child processes

### LINKED ISSUE #4151: Processes do not subscribe to state transition broadcasts of child processes
In this case, this only concerns the `WorkChain` as it is the only process type that can call other processes that are not run blockingly in the same runner. They currently rely on the backup polling mechanism of the runner to trigger them to continue running once the child processes have finished. The whole gimmick of using RabbitMQ was to support enabling this behavior in an event-based way. It is currently unclear if this concept never made it to the code, or it got accidentally removed at some point. The latter seems unlikely, however.

sphuber:
Fixes #4151

When a `WorkChain` step submits sub processes, awaitables will be
created for them. Only when these awaitables have been resolved, meaning
the subprocesses, have terminated, can the workchain continue to the
next step.

The original concept was, for each awaitable, to schedule a callback
once the process had reached a terminal state. The callback was supposed
to be triggered by having the runner add a broadcast subscriber that would
listen for state changes of the sub process. As a fail-safe, a polling
mechanism would also check periodically just in case the broadcast
message would be missed and prevent the caller from waiting indefinitely.

However, the broadcast subscriber was never added and so the system
relied solely on the polling mechanism. This completely undermines the
benefits of having an event-based mechanism, so in this commit the
`Runner.call_on_process_finish` now also registers the broadcast
subscriber.

Note that the affected code had some references to `calculation` which
has been generalized to `process`, since this also applies to workflows
that might waited upon. The `CalculationFuture` has been renamed to
`ProcessFuture` in similar vein. It is currently not used, but it could
have been used for the problem that this commit solves, so it has been
decided to leave it in for now and not remove it entirely.

### DIFF
--- a/aiida/engine/runners.py
+++ b/aiida/engine/runners.py
@@ -9,19 +9,22 @@
 ###########################################################################
 # pylint: disable=global-statement
 """Runners that can run and submit processes."""
-
 import collections
+import functools
 import logging
 import signal
-import tornado.ioloop
+import threading
+import uuid
 
+import kiwipy
 import plumpy
+import tornado.ioloop
 
 from aiida.common import exceptions
 from aiida.orm import load_node
 from aiida.plugins.utils import PluginVersionProvider
 
-from .processes import futures
+from .processes import futures, ProcessState
 from .processes.calcjobs import manager
 from . import transports
 from . import utils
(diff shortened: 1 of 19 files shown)


sphuber:
This PR is blocked because it needs a new release of `plumpy`. Currently I am just manually installing the `develop` version in the CI. Once we are happy with the code, I will release `plumpy==0.15.0` and update the PR.

aiida/engine/runners.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, poll_interval=0, loop=None, communicator=None, rmq_submit=Fal
         self._plugin_version_provider = PluginVersionProvider()
 
         if communicator is not None:
-            self._communicator = communicator
+            self._communicator = plumpy.wrap_communicator(communicator, self._loop)
```
unkcpz, line 70: hi @sphuber , as a reminder, since now you wrap the communicator here to `LoopCommnicator`, then there is no need to add task subscriber with scheduling task_receiver callback to the process event loop by `create_task`. 
In file `aiida/manage/manager.py::Manager::create_daemon_runner`  .

sphuber: Thanks a lot for the heads-up @unkcpz . I will have a look and update the PR.

sphuber: @unkcpz  I added this simplification in the first commit of this PR

unkcpz: all good. Thanks!

aiida/engine/runners.py
```diff
@@ -263,27 +264,65 @@ def run_get_pk(self, process, *args, **inputs):
...
+                callback()
+            finally:
+                resolved = True
+                try:
+                    self._communicator.remove_broadcast_subscriber(subscriber_identifier)
+                except ValueError:
+                    # The broadcast subscriber has already removed
+                    pass
+
+        broadcast_filter = kiwipy.BroadcastFilter(functools.partial(inline_callback, resolved), sender=pk)
```
muhrin, line 300: This, unfortunately, doesn't work.  Python has this sneaky thing where all primitives are passed by value other types by reference meaning that your `resolved` in the callback is actually always a new (`False`) boolean each time the function is invoked.

You can test it with this:
```
res = False

def do(res):
    if res:
        print("Already resolved")
        return
    res = True
    print("Now resolved")



do(res)
do(res)
```


As ugly as it is, this is one solution:
```
res = [False]

def do(res):
    if res[0]:
        print("Already resolved")
        return
    res[0] = True
    print("Now resolved")



do(res)
do(res)
```

But if you don't like it feel free to create anything where an non-primitive is used as a flag.  I think at some point I may have created a special `Flag` class just for this in some of my code.  A `threading.Event` is also a decent choice.

PR #4155, opened 2020-06-06
Fix the unhandled failure mechanism of the `BaseRestartWorkChain`

### LINKED ISSUE #4153: `BaseRestartWorkChain`: `unhandled_failure` flag reset
Hello,
I feel the `unhandled_failure` flag should be reset to `False` once a calculation process finished successfully or it was handled correctly, like the comment says. Am I wrong?
@sphuber 
https://github.com/aiidateam/aiida-core/blob/f112cd9d43553bad129550c1331e5208a5362d83/aiida/engine/processes/workchains/restart.py#L241-L243

sphuber:
Fixes #4153 

Any `BaseRestartWorkChain` will fail if the sub process fails twice in a
row without being handled by a registered process handler. To monitor
this the `unhandled_failure` context variable is used, which is set to
`True` once a failed process was not handled. It should be unset as soon
as the next process finishes successfully, or the failed process is
handled. The reset was there, but was incorrectly setting the context
var to `True` instead of `False`, which went unnoticed due to a lack of
tests. Tests are now added, as well as for the case where the sub
process excepted or was killed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -91,3 +91,56 @@ def generate_calc_job():
         return calc_info
 
     return _generate_calc_job
+
+
+@pytest.fixture
+def generate_work_chain():
+    """Generate an instance of a `WorkChain`."""
+
+    def _generate_work_chain(entry_point, inputs=None):
+        """Generate an instance of a `WorkChain` with the given entry point and inputs.
+
+        :param entry_point: entry point name of the work chain subclass.
+        :param inputs: inputs to be passed to process construction.
+        :return: a `WorkChain` instance.
+        """
+        from aiida.engine.utils import instantiate_process
+        from aiida.manage.manager import get_manager
+        from aiida.plugins import WorkflowFactory
+
+        inputs = inputs or {}
+        process_class = WorkflowFactory(entry_point) if isinstance(entry_point, str) else entry_point
+        runner = get_manager().get_runner()
+        process = instantiate_process(runner, process_class, **inputs)
+
+        return process
+
+    return _generate_work_chain
+
+
+@pytest.fixture
+def generate_calculation_node():
+    """Generate an instance of a `CalculationNode`."""
+    from aiida.engine import ProcessState
+
+    def _generate_calculation_node(process_state=ProcessState.FINISHED, exit_status=None):
+        """Generate an instance of a `CalculationNode`..
+
+        :param process_state: state to set
+        :param exit_status: optional exit status, will be set to `0` if `process_state` is `ProcessState.FINISHED`
+        :return: a `CalculationNode` instance.
+        """
+        from aiida.orm import CalculationNode
+
+        if process_state is ProcessState.FINISHED and exit_status is None:
+            exit_status = 0
+
+        node = CalculationNode()
+        node.set_process_state(process_state)
+
+        if exit_status is not None:
+            node.set_exit_status(exit_status)
+
+        return node
+
+    return _generate_calculation_node
--- a/tests/engine/processes/workchains/test_restart.py
+++ b/tests/engine/processes/workchains/test_restart.py
@@ -8,44 +8,99 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Tests for `aiida.engine.processes.workchains.restart` module."""
-from aiida.backends.testbase import AiidaTestCase
-from aiida.engine.processes.workchains.restart import BaseRestartWorkChain
-from aiida.engine.processes.workchains.utils import process_handler
+# pylint: disable=invalid-name
+import pytest
+
+from aiida.engine import CalcJob, BaseRestartWorkChain, process_handler, ProcessState, ProcessHandlerReport, ExitCode
 
 
-class TestBaseRestartWorkChain(AiidaTestCase):
-    """Tests for the `BaseRestartWorkChain` class."""
+class SomeWorkChain(BaseRestartWorkChain):
+    """Dummy class."""
 
-    @staticmethod
-    def test_is_process_handler():
-        """Test the `BaseRestartWorkChain.is_process_handler` class method."""
+    _process_class = CalcJob
 
-        class SomeWorkChain(BaseRestartWorkChain):
-            """Dummy class."""
+    @process_handler()
+    def handler_a(self, node):  # pylint: disable=inconsistent-return-statements,no-self-use
+        if node.exit_status == 400:
+            return ProcessHandlerReport()
 
-            @process_handler()
-            def handler_a(self, node):
-                pass
+    def not_a_handler(self, node):
+        pass
 
-            def not_a_handler(self, node):
-                pass
 
-        assert SomeWorkChain.is_process_handler('handler_a')
-        assert not SomeWorkChain.is_process_handler('not_a_handler')
-        assert not SomeWorkChain.is_process_handler('unexisting_method')
+def test_is_process_handler():
+    """Test the `BaseRestartWorkChain.is_process_handler` class method."""
+    assert SomeWorkChain.is_process_handler('handler_a')
+    assert not SomeWorkChain.is_process_handler('not_a_handler')
+    assert not SomeWorkChain.is_process_handler('unexisting_method')
 
-    @staticmethod
-    def test_get_process_handler():
-        """Test the `BaseRestartWorkChain.get_process_handlers` class method."""
 
-        class SomeWorkChain(BaseRestartWorkChain):
-            """Dummy class."""
+def test_get_process_handler():
+    """Test the `BaseRestartWorkChain.get_process_handlers` class method."""
+    assert [handler.__name__ for handler in SomeWorkChain.get_process_handlers()] == ['handler_a']
 
-            @process_handler
-            def handler_a(self, node):
-                pass
 
-            def not_a_handler(self, node):
-                pass
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_excepted_process(generate_work_chain, generate_calculation_node):
+    """Test that the workchain aborts if the sub process was excepted."""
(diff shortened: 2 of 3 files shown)


tests/engine/processes/workchains/test_restart.py
```diff
@@ -8,44 +8,94 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Tests for `aiida.engine.processes.workchains.restart` module."""
-from aiida.backends.testbase import AiidaTestCase
-from aiida.engine.processes.workchains.restart import BaseRestartWorkChain
-from aiida.engine.processes.workchains.utils import process_handler
+# pylint: disable=invalid-name
+import pytest
 
+from aiida.engine import CalcJob, BaseRestartWorkChain, process_handler, ProcessState, ProcessHandlerReport, ExitCode
```
ramirezfranciscof, line 14: So, this is a more general question.

I understand that from a core-user perspective, having the important methods available through the higher level modules insted of the specific ones (I mean using `from aiida.engine import process_handler` instead of using the complete `from aiida.engine.processes.workchains.utils import process_handler`) is helpful because they don't need to know the internal organization to have access to these.

However, wouldn't still be better for us to internally keep using the most specific path possible when importing inside aiida-core? I mean, I feel that if we need to debug or stuff like that it would be easier to know the exact location of the functions and classes being imported, and although I know most GUI's have some sort of automatic seach engine for this, but perhaps sometimes you need to use some simplified interface or maybe you are just grepping and it helps to have the information there.

tests/engine/processes/workchains/test_restart.py
```diff
@@ -8,44 +8,94 @@
...
+    The workchain should be aborted if there are two consecutive failed sub processes that went unhandled. We simulate
+    it by setting `ctx.unhandled_failure` to True and append two failed process nodes in `ctx.children`.
+    """
+    process = generate_work_chain(SomeWorkChain, {})
+    process.setup()
+    process.ctx.unhandled_failure = True
+    process.ctx.iteration = 2
+    process.ctx.children = [generate_calculation_node(exit_status=100) for _ in range(2)]
+    assert process.inspect_process() == BaseRestartWorkChain.exit_codes.ERROR_SECOND_CONSECUTIVE_UNHANDLED_FAILURE  # pylint: disable=no-member
+
```
ramirezfranciscof, line 74: Could I ask you how these tests work? It seems like you are getting the `work_chain` process and then manually modifying the context of the workchain from outside of it. As a side note, I don't know why, but I was assuming the context was something that was only modifiable from inside the workchain methods: anyways this is just for testing and that should be the intended way to use it, right?

On the other hand, if you are just modifying things manually I'm not sure how does the workchain error code gets set because I don't see any run or submits, so I only see you setting up the workchain. Is it being sent to run when you `instantiate_process` with the `runner` in `generate_work_chain`? If that is the case, how are you sure that when you modify it the workchain has not already finished? I am not familiar with this way of running stuff...

tests/engine/processes/workchains/test_restart.py
```diff
@@ -8,44 +8,94 @@
...
 
-    @staticmethod
-    def test_is_process_handler():
-        """Test the `BaseRestartWorkChain.is_process_handler` class method."""
+class SomeWorkChain(BaseRestartWorkChain):
+    """Dummy class."""
 
-        class SomeWorkChain(BaseRestartWorkChain):
-            """Dummy class."""
+    _process_class = CalcJob
```
ramirezfranciscof, line 20: What is this? I don't remember needing to set a `_process_class` property inside workchains...is this new?

tests/engine/processes/workchains/test_restart.py
```diff
@@ -8,44 +8,94 @@
...
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_unhandled_reset_after_handled(generate_work_chain, generate_calculation_node):
+    """Test `ctx.unhandled_failure` is reset to `False` in `inspect_process` after a handled failed process."""
+    process = generate_work_chain(SomeWorkChain, {})
+    process.setup()
+    process.ctx.unhandled_failure = True
+    process.ctx.iteration = 2
+    process.ctx.children = [generate_calculation_node(exit_status=0), generate_calculation_node(exit_status=400)]
```
ramirezfranciscof, line 95: If you are just testing that the handler catches the `exit_status=400` here, do you need the previous calculation with `exit_status=0`?

tests/conftest.py
```diff
@@ -91,3 +91,56 @@ def _generate_calc_job(folder, entry_point_name, inputs=None):
...
+        process = instantiate_process(runner, process_class, **inputs)
+
+        return process
+
+    return _generate_work_chain
+
+
+@pytest.fixture
+def generate_calculation_node():
+    """Generate an instance of a `WorkChain`."""
```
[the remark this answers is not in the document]
sphuber: Yes, will adapt it. Thanks

tests/conftest.py
```diff
@@ -91,3 +91,56 @@ def _generate_calc_job(folder, entry_point_name, inputs=None):
...
+        :return: a `WorkChain` instance.
+        """
+        from aiida.engine.utils import instantiate_process
+        from aiida.manage.manager import get_manager
+        from aiida.plugins import WorkflowFactory
+
+        inputs = inputs or {}
+        process_class = WorkflowFactory(entry_point) if isinstance(entry_point, str) else entry_point
+        runner = get_manager().get_runner()
+        process = instantiate_process(runner, process_class, **inputs)
```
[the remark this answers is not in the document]
sphuber: This is how a `Process` instance is created. Normally this is done automatically by the engine when you call `run` or `submit`, but in this case I need more control. Because instead of the engine actually starting to run it, for these tests I just want the process instance and "run" parts of it myself.

tests/engine/processes/workchains/test_restart.py
sphuber: As you said, most proper IDE's can automatically jump to the source and the increased brevity is a big plus in this case. Exception, in my opinion, is when you are importing something from a submodule. In that case the `from .module import something` is better than the absolute form

tests/engine/processes/workchains/test_restart.py
sphuber: This is just a concept of the `BaseRestartWorkChain`. Note that this is an implementation of a `WorkChain` but not a concrete one, i.e., you cannot run a `BaseRestartWorkChain`. It is merely a baseclass for a very common type of `WorkChain`: one that runs a `CalcJob` and deals with errors. So to run an *actual* base restart workchain, one should subclass it and define this `_process_class` which will be the `CalcJob` that the workchain will launch. Look at the `BaseRestartWorkChain` docstring for more information

tests/engine/processes/workchains/test_restart.py
sphuber: The whole point of these tests is exactly that really "running" them is very difficult. This is because it often requires a lot of pre-requisites. In the case of the `BaseRestartWorkChain`, which I am testing here, it needs to run some `CalcJob`, which in turn needs a `Code` that then requires the corresponding binary to be installed. Since I don't want to test all of this in a big integration test, but want to write smaller unit tests of the individual methods of the `BaseRestartWorkChain` I have to mock certain things. Since the methods I am testing are class methods, I first need to have an instance. This is what the `generate_work_chain` is doing. It just gets me a `WorkChain` instance. There I can then call the methods I want to test. In this case, I want to test the `inspect_process` method. This is normally not the first method that is called, so I have to make sure to correctly "mock" the state of the workchain instance that it would normally have, when run fully automatically by the engine. This includes defining the `ctx` correctly, which is an internal dictionary of the `WorkChain` where it stores temporary data. I populate it such that it should trigger a certain code path in `inspect_process` and check that it behaves correctly.

tests/engine/processes/workchains/test_restart.py
sphuber: Technically no, since the code currently won't check the first, and I don't run `inspect_process` manually twice, after having appended a calculation to the ctx first each time. I will maybe change the test to do this and be more explicit, instead of setting the `iteration` manually

tests/engine/processes/workchains/test_restart.py
greschd: Additional point, it's important for us to use the "public" interface as much as possible in the tests - if only to make sure that we don't accidentally break it.

tests/engine/processes/workchains/test_restart.py
ramirezfranciscof: Uhm, I see. I think that my main confussion was that I thought `inspect_process` just got you a result that was set by the engine when running the process, but now I understand this is not so and the `inspect_process` itself looks at the status of the calculations in the context to decide its output. Correct?

sphuber: The `inspect_process` is just a custom `WorkChain` method, that is one of the steps that is called in its outline:
https://github.com/aiidateam/aiida-core/blob/a0e39c69ba7fb15eb36cc400789a172b3670af6d/aiida/engine/processes/workchains/restart.py#L173

So I am just testing the code that is in there.

tests/engine/processes/workchains/test_restart.py
ramirezfranciscof: Ah, ok. Simultaneously checking in the tests that you are not breaking the public interface is a good point, I didn't thought about that. I still think there is some advantages of knowing the exact location of the code you are using, and (at least for the non-test internals) it is not clear to me that the brevity in the "header" of the file (where there is not much logic to obfuscate) outweights this.

sphuber: I really don't think using `from aiida.orm.nodes.data.int import Int` every time instead of `from aiida.orm import Int` does the understanding of the code any favors

PR #4156, opened 2020-06-06
`WorkChain`: guarantee to maintain order of appended awaitables 

### LINKED ISSUE #2961: Guarantee that order is preserved when using `append_` in `WorkChains`
Currently, the `append_` implementation does *not* guarantee that the completed futures will be added to the target list in the same order as the futures are added in the `to_context` call. This may be counter-intuitive and users actually may want to be able to rely on this. We should see if this is possible to implement.

sphuber:
Fixes #2961

Intuitively, one would expect the order of awaitables in the context
once they have been resolved, to maintain the same order with which they
had been inserted using the `append_` construct in the `to_context`
call. However, this was not the case, as the resolved awaitables were
just appended to the target container in the order that they resolved.
To correct this behavior, once the awaitable is inserted, it is already
inserted in the context, where its resolved value should be placed in
the future. This allows the `resolve_awaitable` method to find the
correct place and simply swap out the awaitable, that has been serving
as a placeholder, with the resolved value.

### DIFF
--- a/aiida/engine/processes/workchains/workchain.py
+++ b/aiida/engine/processes/workchains/workchain.py
@@ -8,7 +8,6 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Components for the WorkChain concept of the workflow engine."""
-
 import collections
 import functools
 
@@ -25,7 +24,7 @@ from aiida.orm.utils import load_node
 from ..exit_code import ExitCode
 from ..process_spec import ProcessSpec
 from ..process import Process, ProcessState
-from .awaitable import AwaitableTarget, AwaitableAction, construct_awaitable
+from .awaitable import Awaitable, AwaitableTarget, AwaitableAction, construct_awaitable
 
 __all__ = ('WorkChain', 'if_', 'while_', 'return_')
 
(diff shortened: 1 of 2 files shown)


aiida/engine/processes/workchains/workchain.py
```diff
@@ -124,17 +124,51 @@ def insert_awaitable(self, awaitable):
...
         """
         self._awaitables.append(awaitable)
+
+        # Already assign the awaitable itself to the location in the context container where it is supposed to end up
+        # once it is resolved. This is especially important for the `APPEND` action, since it needs to maintain the
+        # order, but the awaitables will not necessarily be resolved in the order in which they are added. By using the
+        # awaitable as a placeholder, in the `resolve_awaitable`, it can be found and replaced by the resolved value.
+        if awaitable.action == AwaitableAction.ASSIGN:
+            self.ctx[awaitable.key] = awaitable
+        elif awaitable.action == AwaitableAction.APPEND:
```
unkcpz, line 134: I have a question here, it seems the functionality assign_ can be replaced by the append_, is there any special user scenario for `assign_`?

aiida/engine/processes/workchains/awaitable.py
```diff
@@ -62,6 +61,7 @@ def construct_awaitable(target):
             'action': AwaitableAction.ASSIGN,
             'target': awaitable_target,
             'outputs': False,
+            'resolved': False,
```
unkcpz, line 64: Not fully sure why `resolved` used here, I guess it is used to prevent repeatedly resolve the awaitable, since there are two mechanism now. Could you add an entry in docstring for explanation?

aiida/engine/processes/workchains/workchain.py
unkcpz: What is the attribute 'key' of awaitable ? I can't find where it is initialized.

sphuber: The key is set in `construct_awaitables`:https://github.com/aiidateam/aiida-core/blob/c817021b73cc874a87348948b04eace2ba0900b6/aiida/engine/processes/workchains/workchain.py#L147
Basically it is the key that is used in `to_context` or `ToContext` and so is the key that should be used to store the resolved value in the `ctx`

sphuber: > I have a question here, it seems the functionality assign_ can be replaced by the append_, is there any special user scenario for `assign_`?

Why do you say so? If you use assign, it will be assigned to that key in the `ctx` overriding any value that may have been there. If you use `append` it will create a list or simply append the value if the key already contained a list.

aiida/engine/processes/workchains/awaitable.py
sphuber: I based this PR on another branch that is used for PR #4154 . Since that one touched quite a bit of the same code, I didn't want to have to deal with the merge conflicts. So I just based this PR of off that branch. That is why there are 3 commits. Only the last commit is relevant to solve this issue. The other commits reinstate the second mechanism of checking awaitables, as you mentioned. That solves one problem, but added another problem with the awaitables that now, sometimes, it can be "actioned" twice, putting the same node in the context twice, which is unwanted. So I needed some kind of "guard" to make sure it is done only once. The `resolved` attribute is my solution. It has its own problems and possibilities for failure but was the best solution I could come up with for now.

sphuber:
> I think I am not capable enough to review this PR, but interesting in this feature and the relevant #4154 and look forward it to be merged soon, so I make the try. Do you want to have a look? @ltalirz

Thanks for the review @unkcpz always useful to have more eyes on the code. Anyway this PR will have to wait to be merged before PR #4154 is merged

aiida/engine/processes/workchains/workchain.py
```diff
@@ -124,16 +124,45 @@ def insert_awaitable(self, awaitable):
...
+            self.ctx.setdefault(awaitable.key, []).append(awaitable)
```
muhrin, line 135: Might be worth having a quick check in here just to make sure it hasn't been appended before.  Just something like
```
assert awaitable.pk not in (await.pk for await in awaitables)
```

sphuber: Why? I don't see the use case, but if a user _wants_ to add the same subprocess twice, why not let them? When asserting, we are mostly doing this for internal consistency, no? Here we would be basically checking whatever the user returned from a step, because the `insert_awaitable` is only called when `to_context` or `ToContext` is used with a child process

muhrin: Yup, agreed they might want to be able to do it but this PR doesn't support it.   Unless I've misunderstood, when resolving it looks like you interate the placeholders container from start to end and so will find the _frist_ matching entry which will be the corresponding node the second time around and just be overwritten leaving the second occurrence unresolved.

Agreed re: assert, this should really be some kind of timetime error, just didn't want to push an additional line of code change on you.

sphuber: ah damn you're right. Well that is just a straight up bug! When checking the placeholders I should also check that it actually is still an awaitable and not a node, in addition to checking the pk, right?

muhrin: Yeah, I was thinking of a type check as well...

PR #4164, opened 2020-06-08
Various docs improvements

ltalirz:


docs/source/intro/get_started.rst
```diff
@@ -4,11 +4,11 @@
...
 ****************
 
-A working AiiDA installation consists of three core components, plus any external codes you wish to run:
+An AiiDA installation consists of three core components (plus any external codes you wish to run):
 
-* aiida-core: The main Python package and the associated command line tool: ``verdi``.
-* |PostgreSQL|: A service which manages the database that AiiDA uses to store generated data.
-* |RabbitMQ|: A service which manages communication with the processes that AiiDA spawns.
+* ``aiida-core``: The main Python package and the associated ``verdi`` command line interface
+* |PostgreSQL|: The service which manages the database that AiiDA uses to store data.
```
sphuber, line 10:
```suggestion
* |PostgreSQL|: The service that manages the database that AiiDA uses to store data.
```

docs/source/intro/get_started.rst
```diff
@@ -4,11 +4,11 @@
...
+* |RabbitMQ|: The service which manages communication with the processes that AiiDA spawns.
```
sphuber, line 11:
```suggestion
* |RabbitMQ|: The message broker that enables communication with an between AiiDA's daemon workers.
```

docs/source/index.rst
```diff
@@ -114,7 +112,10 @@ How to cite
 
 If you use AiiDA for your research, please cite the following work:
 
-.. highlights:: Giovanni Pizzi, Andrea Cepellotti, Riccardo Sabatini, Nicola Marzari,
+.. highlights:: **AiiDA 1.0:** Sebastiaan. P. Huber, Spyros Zoupanos, Martin Uhrin, Leopold Talirz, Leonid Kahle, Rico Häuselmann, Dominik Gresch, Tiziano Müller, Aliaksandr V. Yakutovich, Casper W. Andersen, Francisco F. Ramirez, Carl S. Adorf, Fernando Gargiulo, Snehal Kumbhar, Elsa Passaro, Conrad Johnston, Andrius Merkys, Andrea Cepellotti, Nicolas Mounet, Nicola Marzari, Boris Kozinsky, Giovanni Pizzi, *AiiDA 1.0, a scalable computational infrastructure for automated reproducible workflows and data provenance*, arXiv:2003.12476 (2020);
+    http://www.aiida.net.
```
sphuber, line 116: This does not show up as a hyperlink, but the one in the second reference does. Do you know why?

docs/source/index.rst
```diff
@@ -114,7 +112,10 @@ How to cite
 
 If you use AiiDA for your research, please cite the following work:
 
-.. highlights:: Giovanni Pizzi, Andrea Cepellotti, Riccardo Sabatini, Nicola Marzari,
+.. highlights:: **AiiDA 1.0:** Sebastiaan. P. Huber, Spyros Zoupanos, Martin Uhrin, Leopold Talirz, Leonid Kahle, Rico Häuselmann, Dominik Gresch, Tiziano Müller, Aliaksandr V. Yakutovich, Casper W. Andersen, Francisco F. Ramirez, Carl S. Adorf, Fernando Gargiulo, Snehal Kumbhar, Elsa Passaro, Conrad Johnston, Andrius Merkys, Andrea Cepellotti, Nicolas Mounet, Nicola Marzari, Boris Kozinsky, Giovanni Pizzi, *AiiDA 1.0, a scalable computational infrastructure for automated reproducible workflows and data provenance*, arXiv:2003.12476 (2020);
```
sphuber, line 115: Can we turn the `arXiv` URI into a hyperlink?

docs/source/index.rst
```diff
@@ -133,15 +134,15 @@ This work is supported by:
...
 
 .. image:: images/MaX.png
-   :height: 80px
+   :height: 65px
    :target: `MaX European Centre of Excellence`_
 
 .. image:: images/swissuniversities.png
-   :height: 35px
+   :height: 30px
    :target: `swissuniversities P-5 project "Materials Cloud"`_
```
sphuber, line 146: Do you know of a way to make these vertically centered (preferably equal height) and equally spaced?

docs/source/index.rst
ltalirz: no idea why (I don't know why the other one works).
anyhow, I can put an explicit link

docs/source/index.rst
ltalirz: sure

docs/source/index.rst
ltalirz: You can use
```
.. raw:: html

   <img src="...">
```
but I would like to avoid that if possible

docs/source/intro/get_started.rst
ltalirz: I modified your suggestion a bit, since people at that point may not know what daemon workers are

docs/source/conf.py
```diff
@@ -143,17 +145,29 @@
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-# html_theme = 'default'
+html_theme = 'pydata_sphinx_theme'
```
csadorf, line 148: I'm confused, didn't we use that already?

docs/source/conf.py
```diff
@@ -227,10 +222,10 @@
 #html_show_sourcelink = True
 
 # If true, "Created using Sphinx" is shown in the HTML footer. Default is True.
-#html_show_sphinx = True
+html_show_sphinx = False
 
 # If true, "(C) Copyright ..." is shown in the HTML footer. Default is True.
-#html_show_copyright = True
+html_show_copyright = False
```
csadorf, line 228: Why not?

docs/source/intro/get_started.rst
```diff
@@ -4,11 +4,11 @@
 Getting Started
 ****************
 
-A working AiiDA installation consists of three core components, plus any external codes you wish to run:
+An AiiDA installation consists of three core components (plus any external codes you wish to run):
 
-* aiida-core: The main Python package and the associated command line tool: ``verdi``.
-* |PostgreSQL|: A service which manages the database that AiiDA uses to store generated data.
-* |RabbitMQ|: A service which manages communication with the processes that AiiDA spawns.
+* ``aiida-core``: The main Python package and the associated ``verdi`` command line interface
```
csadorf, line 9: Why would we typeset `aiida-core` as code, when it just represents a name in this context like all the other components?

docs/source/intro/get_started.rst
```diff
@@ -4,11 +4,11 @@
...
+* |PostgreSQL|: The service that manages the database that AiiDA uses to store data.
+* |RabbitMQ|: The message broker used by the AiiDA daemon to manage your simulations.
```
csadorf, line 11: We don't just manage "simulations".

sphuber: It is also not correct. I get that we don't want to make it too technical, but we also don't want to give the wrong impression. How about:
> The message broker that enables communication with and between AiiDA's process runners
Concise, correct and understandable, no? Even if "process runner" may not be fully understood, I think the majority of people will also not understand "message broker" but that is not important here I would say

docs/source/conf.py
ltalirz: I've just moved it to the location suggested by the config file layout

docs/source/conf.py
ltalirz: Because we use our own footer (see changes in this PR and discussion in "general" channel)

docs/source/intro/get_started.rst
ltalirz: I think you're right, I'll revert that

docs/source/intro/get_started.rst
ltalirz: > We don't just manage "simulations".

Happy to improve - what would you suggest instead? Workflows?
Please note that "simulations" are actually mentioned twice in the abstract https://arxiv.org/abs/2003.12476

> Even if "process runner" may not be fully understood,

I agree that "process runner" will not be understood, and I therefore don't think we should use the term here, which is the entry point for every new user installing AiiDA.
In my opinion, a new user should not even have to care about the fact that there are different runners, until the point where they need to do performance tuning.
All they need to know here is that rabbitmq is used by the daemon to keep track of what it does.

>  I think the majority of people will also not understand "message broker" but that is not important here I would say

I accepted your suggestion to replace "service" by "message broker" because I felt it was a more descriptive term, and one that is used outside AiiDA as well.
But if we think people will not understand what it means, we should go back to "service".

greschd: "The message broker used for communication within AiiDA"?

sphuber: > In my opinion, a new user should not even have to care about the fact that there are different runners, until the point where they need to do performance tuning.

I agree that they don't need to understand exactly what it does at this point, so a strongly reduced description is fine. Changing "service" with "message broker" was also not the real point, just that the rest of the statement was incorrect...

> All they need to know here is that rabbitmq is used by the daemon to keep track of what it does.

... as is this. It is not at all used "by the daemon to keep track of what it does". Brevity is fine, but not at the cost of correctness.

But I am fine with the suggestion of @greschd

docs/source/conf.py
csadorf: Ok.

docs/source/conf.py
csadorf: Got it.

ltalirz:
Ready for re-review

docs/source/index.rst
```diff
@@ -114,7 +110,10 @@ How to cite
 
 If you use AiiDA for your research, please cite the following work:
 
-.. highlights:: Giovanni Pizzi, Andrea Cepellotti, Riccardo Sabatini, Nicola Marzari,
+.. highlights:: **AiiDA 1.0:** Sebastiaan. P. Huber, Spyros Zoupanos, Martin Uhrin, Leopold Talirz, Leonid Kahle, Rico Häuselmann, Dominik Gresch, Tiziano Müller, Aliaksandr V. Yakutovich, Casper W. Andersen, Francisco F. Ramirez, Carl S. Adorf, Fernando Gargiulo, Snehal Kumbhar, Elsa Passaro, Conrad Johnston, Andrius Merkys, Andrea Cepellotti, Nicolas Mounet, Nicola Marzari, Boris Kozinsky, Giovanni Pizzi, *AiiDA 1.0, a scalable computational infrastructure for automated reproducible workflows and data provenance*, `arXiv:2003.12476 (2020) <https://arxiv.org/abs/2003.12476>`_;
```
csadorf, line 113: Is the intention that you should cite this paper when using AiiDA 1.0+ and the other one for < 1.0? If yes, I don't think that's clear enough. If that's not the intention, it's perfectly fine.

giovannipizzi: Probably yes, but I think it's ok to merge this for now, also considering that the paper is not really published; we will have anyway to fix this when the paper is out, and also there will be an AiiDA engine paper out soon and we need to clarify this further on what to cite when.

sphuber:
> I would proceed merging this for now unless there are more issues.
> Also, I am going to make a PR with further improvements on the text after discussion with Nicola, that are going to create git conflicts.

OK to merge this. If you are going to add additional changes to the docs, could you maybe include the one letter fix in the PR that was just opened?

giovannipizzi:
Yes I will

PR #4166, opened 2020-06-08
Fix bug in `BaseRestartWorkChain.inspect_process`

### LINKED ISSUE #4165: `BaseRestartWorkChain.inspect_process` erroneously references `report` instead of `last_report` for final exit code
Originally reported by @lorisercole in #4153 
https://github.com/aiidateam/aiida-core/blob/a67a625f930b4649b09ba63a75d1c184fa5b9e94/aiida/engine/processes/workchains/restart.py#L258

This should fail if `last_report` is set and the last handler returns `None`. In this case it should trigger an `AttributeError` as `None` does not have `exit_code`. This can happen if at least one handler return a process report with `do_break=False` and then at least one other handler is called that returns `None`.

sphuber:
Fixes #4165 

If at the end of the method, `last_report` is defined, meaning at least
one handler returned a report, it should return the corresponding exit
code, except it was erroneously accessing that of the `report` variable,
which could very well be `None` if the last called handler returned
nothing.

### DIFF
--- a/tests/engine/processes/workchains/test_restart.py
+++ b/tests/engine/processes/workchains/test_restart.py
@@ -8,7 +8,7 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Tests for `aiida.engine.processes.workchains.restart` module."""
-# pylint: disable=invalid-name
+# pylint: disable=invalid-name,inconsistent-return-statements,no-self-use,no-member
 import pytest
 
 from aiida.engine import CalcJob, BaseRestartWorkChain, process_handler, ProcessState, ProcessHandlerReport, ExitCode
@@ -19,12 +19,16 @@ class SomeWorkChain(BaseRestartWorkChain):
 
     _process_class = CalcJob
 
-    @process_handler()
-    def handler_a(self, node):  # pylint: disable=inconsistent-return-statements,no-self-use
+    @process_handler(priority=200)
+    def handler_a(self, node):
         if node.exit_status == 400:
-            return ProcessHandlerReport()
+            return ProcessHandlerReport(do_break=False, exit_code=ExitCode(418, 'IMATEAPOT'))
 
-    def not_a_handler(self, node):
+    @process_handler(priority=100)
+    def handler_b(self, _):
+        return
+
+    def not_a_handler(self, _):
         pass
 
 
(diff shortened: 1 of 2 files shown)


tests/engine/processes/workchains/test_restart.py
```diff
@@ -101,6 +105,10 @@ def test_unhandled_reset_after_handled(generate_work_chain, generate_calculation
...
     process.ctx.children.append(generate_calculation_node(exit_status=400))
     result = process.inspect_process()
+
+    # Even though `handler_a` was followed by `handler_b`, we should retrieve the exit_code from `handler_a` because
+    # `handler_b` returned `None` which should not overwrite the last report.
     assert isinstance(result, ExitCode)
-    assert result.status == 0
+    assert result.status == 418
+    assert result.message == 'IMATEAPOT'
     assert process.ctx.unhandled_failure is False
```
ramirezfranciscof, line 114: If we are testing the reset, and although it might be redundant because you already check that a calculation with an exit status != 0 sets off the `process.ctx.unhandled_failure`, shouldn't it be better to check that the status actually change and was set back? (so perhaps `assert process.ctx.unhandled_failure is True` just before calling `result = process.inspect_process()`)

ramirezfranciscof, line 114: If we are testing the reset, and although it might be redundant because you already check that a calculation with an exit status != 0 sets off the `process.ctx.unhandled_failure`, shouldn't it be anyways better to check that the status actually change and was set back? (so perhaps `assert process.ctx.unhandled_failure is True` just before calling `result = process.inspect_process()`)

sphuber: I may misunderstand your point, but the `unhandled_failure` is only changed by the workchain in the `inspect_process` call. So adding the `assert process.ctx.unhandled_failure is True` check before the call will fail.

ramirezfranciscof: No, I think you understood correctly. But then maybe the docstring of the test should say "Test `ctx.unhandled_failure` is set to `False` in `inspect_process` when a failed process is handled." or something like this? I mean, this is probably a nitpick, but using "reset" makes me think that at some point the `ctx.unhandled_failure` is set to `True`, which is not the case from what I am seeing in the code right?

sphuber: Ah I see now, you are right. There is a mistake. The idea of the test is that there is first a failed calculation, which sets `unhandled_failure=True` followed by another failed calculation, but that one returns a report, so it is not considered unhandled and so should reset the flag. The first calculation shouldn't be `0` then but some non-zero code. Will push a fix.

ramirezfranciscof: Ah, ok, that makes more sense, haaha. This was why I was asking why you needed this previous calculation in the other PR. Good catch (Y)

sphuber: > Good catch (Y)

Right back at ya. Thanks for the review :+1:

PR #4169, opened 2020-06-09
Update tutorial and how-to guides to literal-include for code snippets 

### LINKED ISSUE #4128: Docs: update all process code snippets in how-to and tutorial to literal include
This can only be done once `docs-revamp` is merged into `develop`, because the correct versions of those process plugins `add_multiply`, `ArithmeticAddCalculation` and `MultiplyAddWorkChain`.

sphuber:
Fixes #4128 

The basic tutorial and how-to guides on implementing `CalcJob` and
`Parser` plugins, as well as how-to write work functions and work chains
have been updated to `literal-include` the implementation of the actual
plugins that ship with `aiida-core`.

This has the advantage that the code that is shown in these crucial
parts of the documentation is guaranteed to work with each release as
the plugins themselves are tested in the unit test suite. The down side,
however, is that we do not always want to show everything from the
entire implementation at any point, so we have to resort to specifying
line numbers.

This is fine, except that it runs the risk that the line numbers run out
of sync if the implementation is changed and automatically detecting
this is difficult. Nonetheless, the other solution of just duplicating
the code in the documentation has the same problem, and in addition the
code is not easily tested. Tests *could* be added, but given that
testing these plugins is not trivial, the required additional code and
infrastructure would be more of a burden than the current solution.


aiida/calculations/arithmetic/add.py
```diff
@@ -7,26 +7,36 @@
...
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-"""`CalcJob` implementation to add two numbers using bash for testing and demonstration purposes."""
+# yapf: disable
+"""`CalcJob` implementation to add two numbers using bash for testing and demonstration purposes.
+
+.. warning::
+
+    This implementation is used directly in the documentation as a literal-include, which means that if any part of this
+    code is changed, the snippets in the file `docs/source/howto/codes.rst` have to be checked for consistency.
```
csadorf, line 16: I don't think this should be part of the doc-string, but instead an Python comment. Otherwise it shows up [here](https://258-77234579-gh.circle-artifacts.com/0/html/reference/apidoc/aiida.calculations.arithmetic.html#module-aiida.calculations.arithmetic.add) as part of the reference documentation, which I don't think is intended.

aiida/calculations/arithmetic/add.py
```diff
@@ -35,45 +45,29 @@ def define(cls, spec):
...
-        example, on what files to copy to the remote machine, what files to retrieve once it has completed, specific
-        scheduler settings and more.
-        :param folder: an `aiida.common.folders.Folder` to write the input files to
-        :returns: `aiida.common.datastructures.CalcInfo` instance
-        """
-        input_x = self.inputs.x
-        input_y = self.inputs.y
-        input_code = self.inputs.code
+        Convert the input nodes into the corresponding input files in the format that the code will expect. In addition,
+        define and return a `CalcInfo` instance, which is a simple data structure that contains  information for the
```
csadorf, line 54:
```suggestion
        define and return a `CalcInfo` instance, which is a simple data structure that contains information for the
```

aiida/parsers/plugins/arithmetic/add.py
```diff
@@ -7,50 +7,37 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
+# pylint: disable=inconsistent-return-statements
+"""Parser for an `ArithmeticAddCalculation` job.
+
+.. warning::
+
+    This implementation is used directly in the documentation as a literal-include, which means that if any part of this
+    code is changed, the snippets in the file `docs/source/howto/codes.rst` have to be checked for consistency.
```
csadorf, line 16: See comment on module `calculations/arithmetic/add.py`.

aiida/parsers/plugins/arithmetic/add.py
```diff
@@ -7,50 +7,37 @@
...
+                result = int(handle.read())
+        except OSError:
             return self.exit_codes.ERROR_READING_OUTPUT_FILE
-
-        if result is None:
+        except ValueError:
             return self.exit_codes.ERROR_INVALID_OUTPUT
 
-        try:
-            allow_negative = self.node.inputs.settings.get_attribute('allow_negative', True)
```
csadorf, line 34: Was this attribute deprecated?

aiida/workflows/arithmetic/add_multiply.py
```diff
@@ -7,7 +7,14 @@
...
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-"""Basic calcfunction-based workflows for demonstration purposes."""
+"""Basic calcfunction-based workflows for demonstration purposes.
+
+.. warning::
+
+    This implementation is used directly in the documentation as a literal-include, which means that if any part of this
+    code is changed, the snippets in the file `docs/source/howto/workflows.rst` and `docs/source/intro.tutorial.rst`
+    have to be checked for consistency.
```
csadorf, line 16: See comment on module `calculations/arithmetic/add.py`.

docs/source/intro/tutorial.rst
```diff
@@ -499,62 +499,9 @@ You can see the code below:
...
-
-                if result.value < 0:
-                    return self.exit_codes.ERROR_NEGATIVE_NUMBER  # pylint: disable=no-member
-
-            def result(self):
-                """Add the result to the outputs."""
-                self.out('result', self.ctx.addition.outputs.sum)
+    .. literalinclude:: ../../../aiida/workflows/arithmetic/multiply_add.py
+        :language: python
+        :lines: 20-72
```
csadorf, line 504: It is  possible to place-in code markers (could be a comment) that are used to delimit the included code instead. Those markers could also include the developer-directed warnings about the fact that changes to the enclosed code segment will directly affect the documentation. I think that is something we should at least evaluate as an alternative to hard-coded line numbers.

sphuber: If that is possible, that would indeed be worth considering. I have looked around online, but have not been able to find this functionality in the Sphinx docs. Only `:pyobject:` to include certain methods, which is already useful and I will apply where possible. Can you point me to the syntax that supports these markers?

aiida/parsers/plugins/arithmetic/add.py
sphuber: It was not, but since this class is here purely for demonstration and internal testing purposes, I think we can remove it without going through a proper deprecation path.

docs/source/intro/tutorial.rst
sphuber: Nvm, I found them.

sphuber:
@csadorf I used the `:pyobject:` option and the markers you mention to get rid of depending on explicit line numbers. The only exception is for a snippet from the parser implementation. This file therefore is the only one that still has a warning, which is now formatted as a normal Python comment instead of a module docstring.

aiida/calculations/arithmetic/add.py
```diff
@@ -7,73 +7,63 @@
...
-        # yapf: disable
         spec.inputs['metadata']['options']['parser_name'].default = 'arithmetic.add'
         spec.inputs['metadata']['options']['input_filename'].default = 'aiida.in'
         spec.inputs['metadata']['options']['output_filename'].default = 'aiida.out'
         spec.inputs['metadata']['options']['resources'].default = {'num_machines': 1, 'num_mpiprocs_per_machine': 1}
         spec.input('x', valid_type=(orm.Int, orm.Float), help='The left operand.')
         spec.input('y', valid_type=(orm.Int, orm.Float), help='The right operand.')
-        spec.input('settings', required=False, valid_type=orm.Dict, help='Optional settings.')
         spec.output('sum', valid_type=(orm.Int, orm.Float), help='The sum of the left and right operand.')
+        # start exit codes
```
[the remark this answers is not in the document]
csadorf: I can see this comment to be removed at some point, because it is not clear why it is there.

aiida/calculations/arithmetic/add.py
csadorf: This should still be corrected.

aiida/parsers/plugins/arithmetic/add.py
csadorf: Ok, no problem then.

aiida/calculations/arithmetic/add.py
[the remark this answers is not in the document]
sphuber: The nice thing about the new approach with the markers though is that then the docs build will fail.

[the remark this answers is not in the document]
csadorf: Good point, but the comment might not hurt. Up to you.

sphuber:
@mbercx @ramirezfranciscof do you still want to give a look at the changes? If not, then I will clean up the commits (and address the final open changes requested by @csadorf) and we can merge it

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -111,8 +113,12 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     @classmethod
-    def define(cls, spec):
+    def define(cls, spec: CalcJobProcessSpec):
```
ramirezfranciscof, line 116: So, this does not change behavior, it is just for documentation's purposes, right? If so, maybe we should do the same in `aiida/workflows/arithmetic/multiply_add.py`(line 30).

ramirezfranciscof:
@sphuber yes, for me its fine. I took a general glance at it to see anything that popped out but just had the one comment above.

aiida/engine/processes/calcjobs/calcjob.py
sphuber: yes, this is a feature called type hinting that was introduced in Python 3.5. You can directly specify the type of arguments and return values. This is not just for documentation, but also enables some form of static type analysis which can help IDE's and linting tools

greschd: If we're going to start adding type hints, we should probably run `mypy` (with the least restrictive options, to start) in the pre-commit hooks. Having _wrong_ type hints is definitely worse than having none at all.

greschd: Will give at try momentarily.

greschd: ``Found 664 errors in 240 files (checked 724 source files)`` ooh boy 🥴

sphuber: Huh? What are these errors? Are these type violations detected by `mypy`? How is that possible if we don't even specify types yet?

greschd: So yeah, enabling it on the whole codebase at once may be a tad too ambitious. Here's a guide how to do it step-by-step: https://mypy.readthedocs.io/en/stable/existing_code.html

greschd: It'll also detect type violations for things it can infer automatically.. like overwriting a variable with something of a different type. And there's also a bunch of errors from things it doesn't know how/where to import, because it hasn't been annotated yet.

aiida/calculations/arithmetic/add.py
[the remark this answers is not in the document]
sphuber: Added the comment

sphuber:
Btw, this should probably be rebase-merged and not squash merged. Think it is good to keep these commits separate

mbercx:
Sorry for the delay! I had a look at the changes, did another testrun of the tutorial, and everything looks fine to me.

PR #4171, opened 2020-06-09
Fix bug in `TrajectoryData.get_step_data`

### LINKED ISSUE #4170: `TrajectoryData.get_step_data` throws error if trajectory does not contain `cells` array
[`TrajectoryData.get_step_data`](https://github.com/aiidateam/aiida-core/blob/c817021b73cc874a87348948b04eace2ba0900b6/aiida/orm/nodes/data/array/trajectory.py#L320) throws an error if the trajectory does not contain an array `cells`.
This is due to a typo here:
https://github.com/aiidateam/aiida-core/blob/c817021b73cc874a87348948b04eace2ba0900b6/aiida/orm/nodes/data/array/trajectory.py#L354-L357
`cell` will not be defined if `cells` is `None`.

**Side note**:
once the bug above if fixed, if a trajectory does not contain `cells`, the method [`TrajectoryData.get_step_structure`](https://github.com/aiidateam/aiida-core/blob/c817021b73cc874a87348948b04eace2ba0900b6/aiida/orm/nodes/data/array/trajectory.py#L359) will return a `StructureData` with the default cell matrix
```
[[1.0, 0.0, 0.0],
[0.0, 1.0, 0.0],
[0.0, 0.0, 1.0]]
```
Should the user be warned about this (perhaps not intended) behavior? or is it ok, since it is the default cell value of `StructureData`?

lorisercole:
Fixes #4170 

If a `TrajectoryData` does not contain an array `cells` the
`get_step_data` method should define and return `cell = None`.
There was a typo that caused an error in this case.

### DIFF
--- a/aiida/orm/nodes/data/array/trajectory.py
+++ b/aiida/orm/nodes/data/array/trajectory.py
@@ -351,9 +351,9 @@ class TrajectoryData(ArrayData):
         time = self.get_times()
         if time is not None:
             time = time[index]
-        cells = self.get_cells()
-        if cells is not None:
-            cell = cells[index, :, :]
+        cell = self.get_cells()
+        if cell is not None:
+            cell = cell[index, :, :]
         return (self.get_stepids()[index], time, cell, self.symbols, self.get_positions()[index, :, :], vel)
 
     def get_step_structure(self, index, custom_kinds=None):
--- a/aiida/orm/nodes/data/array/trajectory.py
+++ b/aiida/orm/nodes/data/array/trajectory.py
@@ -12,6 +12,7 @@ AiiDA class to deal with crystal structure trajectories.
 """
 
 import collections
+import warnings
 
 from .array import ArrayData
 
@@ -318,14 +319,14 @@ class TrajectoryData(ArrayData):
             raise ValueError('{} not among the stepids'.format(stepid))
 
     def get_step_data(self, index):
-        r"""
-        Return a tuple with all information concerning
-        the stepid with given index (0 is the first step, 1 the second step
-        and so on). If you know only the step value, use the
-        :py:meth:`.get_index_from_stepid`  method to get the
-        corresponding index.
+        """
+        Return a tuple with all information concerning the stepid with given
+        index (0 is the first step, 1 the second step and so on). If you know
+        only the step value, use the :py:meth:`.get_index_from_stepid` method
+        to get the corresponding index.
 
-        If no velocities were specified, None is returned as the last element.
+        If no velocities, cells, or times were specified, None is returned as
+        the corresponding element.
 
         :return: A tuple in the format
           ``(stepid, time, cell, symbols, positions, velocities)``,
@@ -341,8 +342,7 @@ class TrajectoryData(ArrayData):
         """
         if index >= self.numsteps:
             raise IndexError(
-                'You have only {} steps, but you are looking beyond'
-                ' (index={})'.format(self.numsteps, index)
+                'You have only {} steps, but you are looking beyond (index={})'.format(self.numsteps, index)
             )
 
         vel = self.get_velocities()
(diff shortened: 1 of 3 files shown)


sphuber:
Thanks @lorisercole . Could you rebase your branch on the latest `develop` such that the build is run correctly and maybe add a regression tests? It should go in `tests.orm.data.test_trajectory.py` which doesn't exist yet, but you can create it.

lorisercole:
Sure @sphuber. I've done the rebase.
What kind of tests should I add? I see there is already something in [tests.orm.data.test_data.py](https://github.com/aiidateam/aiida-core/blob/c817021b73cc874a87348948b04eace2ba0900b6/tests/orm/data/test_data.py).
Maybe we should test these methods to extract data or a structure from a particular step of the trajectory. Anything else?

sphuber:
Those tests are only to test the export functionality. What would be best is to create the file `tests.orm.data.test_trajectory` and add some basic tests. At the very least create a test that would trigger the exception that you uncovered. So something like:
```
def test_trajectory_get_step_data_empty():
    trajectory = TrajectoryData()  # Create without cells, times, etc.
    result = trajectory.get_step_data()  # Check that this doesn't raises
    assert result[2] == None  # Or whatever the index of cells is in the index. Just check the return value is what you expect
```

lorisercole:
I added the optional parameter `custom_cell` to `TrajectoryData.get_step_structure`, that allows one to override the cell of
the structure that is be extracted from a `TrajectoryData`.
If omitted, the cell will be read from the trajectory, if present.
If a cell is not found, the `StructureData` will use the default cell, and an user warning will be emitted.
I added a test for this functionality. Ready for review.

aiida/orm/nodes/data/array/trajectory.py
```diff
@@ -318,14 +319,14 @@ def get_index_from_stepid(self, stepid):
             raise ValueError('{} not among the stepids'.format(stepid))
 
     def get_step_data(self, index):
-        r"""
-        Return a tuple with all information concerning
-        the stepid with given index (0 is the first step, 1 the second step
-        and so on). If you know only the step value, use the
-        :py:meth:`.get_index_from_stepid`  method to get the
-        corresponding index.
+        """
```
giovannipizzi, line 322: you need the 'r' in front, since there are some backslashes in the string

giovannipizzi (changes requested):
Hi @lorisercole sorry it took so long to reply, this thread got hidden among the others...
I think your implemented behaviour  (custom_cell) might be a good option.
Only one question: do you see a case in which you really want to override the trajectory cell if provided and meaningful?

Otherwise, if not provided, probably it's OK to provide the default cell, *but* we should also set the pbc1/2/3 flags to False to indicate a non-periodic molecule? And in this case, maybe it's ok not to raise a Warning? In the sense that it's ok to provide a trajectory of a molecule MD, I don't think this should be considered wrong?

Was your original use case with no cell specified a molecule or a periodic system?

Finally, even when providing a custom_cell: is this because you still want to consider the system as non-periodic, but with a meaningful cell? Because then probably we should also have a custom_pbc command? (I.e., when you provide a cell, you should also be able to decide if you still want to consider the system as periodic or not)

sphuber:
Since this PR has been open for a long time, I have taken the liberty to rebase it on the current latest `main` and just keep the fix for #4170 in a new PR #5734 . I have only kept the immediate bug fix and not included the added feature of the `custom_cell` for the `get_step_structure` method. The reason is that there were open questions on its use and if a user doesn't want the default cell, it can be easily changed after the return:
```
structure = trajectory.get_step_structure(1)
structure.cell = some_other_cell
```
The `get_step_structure` returns an unstored node, so this is possible without any issues. @lorisercole if you or anyone else still thinks that the argument is a better/required solution, please open a separate PR.

PR #4175, opened 2020-06-15
`SshTransport.gettree`: allow non-existing nested target directories

### LINKED ISSUE #4174: `SshTransport.gettree` fails when asking to create nested directory
When trying to copy relative path that is nested `sub/path` into a nested location `/base/sub/path` and `sub` does not yet exist, the `gettree` call will except.

sphuber:
Fixes #4174 

The `gettree` method would raise an `OSError` for the `SshTransport` if
the target local path contains intermediate subdirectories that are not
created beforehand. The same would work without problem for the
`LocalTransport` plugin. It makes sense to just create non-existing
directories along the way, so the `os.mkdir` command that was used
originally to create the target directory is replaced with the method
`os.makedirs` setting `exists_ok=True`.


sphuber:
It is interesting that this went unnoticed. This came up during development of the `aiida-quantumespresso-hp` plugin, but I definitely would have thought that retrieving a folder to a nested local folder would not be such an exotic use case.

ramirezfranciscof:
The only thing is that when I try to test the transport locally I keep running into problems. This is clearly my issue because the test pass on the CI and this also happened before when I was trying to do some modifications in the transport, but do you have any idea of what may it be? I think all of the tests get the same error:
```
paramiko.ssh_exception.NoValidConnectionsError: [Errno None] Unable to connect to port 22 on 127.0.0.1 or ::1
```

sphuber:
Seems like you might have a problem with the configuration of the openssh-server on your localhost. We are testing the `SshTransport` by connecting to localhost over ssh. You may have blocked the port 22 or maybe the server is not even running?

sphuber:
Thanks for the review. Let's wait with merging until @giovannipizzi has had a look to make sure I am not patching a problem that has its origin elsewhere

PR #4176, opened 2020-06-15
Add minimal 'mypy' run to the pre-commit hooks.

greschd:
Currently, only two files are checked: ``aiida/tools/groups/paths.py`` and ``aiida/engine/processes/calcjobs/calcjob.py``. The option ``follow_imports`` is set to ``skip``, because there are a lot of issues in other files, mostly because these lack type hints.
Even on the two checked files, the strictness level of mypy is minimal (see the commented options in ``mypy.ini``). 

From here, we can continue expanding the type hints one file at
a time:
- add the file to ``files`` in the ``mypy`` pre-commit hook
- fix all issues that mypy reports

@chrisjsewell I had to change some of the type hints in `groups/paths.py` - please check that these are correct.


chrisjsewell:
Yeh but you know we can't add these proper type annotations, until we drop python 3.5 support in October: https://devguide.python.org/#status-of-python-branches

sphuber:
> Yeh but you know we can't add these proper type annotations, until we drop python 3.5 support in October: https://devguide.python.org/#status-of-python-branches

These are only method argument and return value annotations, which is supported by Python 3.5. We just cannot add variable annotations because that was only added in Python 3.6.

Anyway, I really cannot wait for `pyproject.toml` to catch on and we can do away with the hundreds of different configuration files in the repo root for all the different tools we run...

greschd:
Only variable annotations and lazy annotations are not supported, right? Need to fix the lazy thing, though..

chrisjsewell:
> We just cannot add variable annotations because that was only added in Python 3.6

Ah fair enough brilliant, for some reason I thought it was 3.5, otherwise I would have already added them!

chrisjsewell:
> Only variable annotations and lazy annotations are not supported, right?

Well the other big thing in the pipeline is [TypedDict](https://www.python.org/dev/peps/pep-0589/) (introduced in 3.8). Having worked with interfaces in TypeScript, which are similar, these will be so helpful.

Other than that LGTM 👍

aiida/tools/groups/paths.py
```diff
@@ -88,69 +87,59 @@ def _validate_path(self, path):
...
         """Return the parent path."""
         if self.path_list:
             return GroupPath(
                 self.delimiter.join(self.path_list[:-1]), cls=self.cls, warn_invalid_child=self._warn_invalid_child
             )
         return None
 
-    def __truediv__(self, path):
-        # type: (str) -> GroupPath
+    def __truediv__(self, path) -> 'GroupPath':
```
chrisjsewell, line 142:
```suggestion
    def __truediv__(self, path: str) -> 'GroupPath':
```

aiida/tools/groups/paths.py
```diff
@@ -240,26 +223,22 @@ def children(self):
...
 
-    def __len__(self):
-        # type: () -> int
+    def __len__(self) -> int:
         """Return the number of children for this path."""
         return sum(1 for _ in self.children)
 
-    def __contains__(self, key):
-        # type: (str) -> bool
+    def __contains__(self, key) -> bool:
```
chrisjsewell, line 234:
```suggestion
    def __contains__(self, key: str) -> bool:
```

greschd:
@chrisjsewell just changed something more: ``get_or_create_group`` returns ``Tuple[orm.Group, bool]``, not ``Tuple['GroupPath', bool]``, right?

greschd:
Ah, and 

```python
def walk_nodes(self, filters=None, node_class=None, query_batch=None) -> Iterable[WalkNodeResult]:
```

to
```python
def walk_nodes(self, filters=None, node_class=None, query_batch=None) -> Iterator[WalkNodeResult]:
```

aiida/tools/groups/paths.py
```diff
@@ -268,8 +247,7 @@ def walk(self, return_virtual=True):
                 if return_virtual or not sub_child.is_virtual:
                     yield sub_child
 
-    def walk_nodes(self, filters=None, node_class=None, query_batch=None):
-        # type: () -> Iterable[WalkNodeResult]
+    def walk_nodes(self, filters=None, node_class=None, query_batch=None) -> Iterable[WalkNodeResult]:
```
chrisjsewell, line 250:
```suggestion
    def walk_nodes(self, filters: Optional[dict]=None, node_class: Optional[orm.Node]=None, query_batch: Optional[int]=None) -> Iterable[WalkNodeResult]:
```

so is this not allowed in 3.5?

greschd: That.. should all be fine I think. What's not allowed is hints directly on a _variable_ (not function argument), and classes which are not yet defined must be given as string.

greschd: Of course, you could make it an `Optional[Dict[key_type, value_type]]` if you wanna be cute.

chrisjsewell: Well what would be super cute would be using `TypeDicts` to create a `filters` interface, which would in turn give much improved autocompletion for query writing. But that's for next year!

chrisjsewell: (btw I didn't change `Iterable` to `Iterator` here)

greschd:
```python
 sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 0 reached, connection timed out, timeout 1 (Background on this error at: http://sqlalche.me/e/3o7r)
```
Is this a known flaky test?

aiida/tools/groups/paths.py
```diff
@@ -268,8 +247,9 @@ def walk(self, return_virtual=True):
                 if return_virtual or not sub_child.is_virtual:
                     yield sub_child
 
-    def walk_nodes(self, filters=None, node_class=None, query_batch=None):
-        # type: () -> Iterable[WalkNodeResult]
+    def walk_nodes(
+        self, filters: Optional[dict] = None, node_class: Optional[orm.Node] = None, query_batch: Optional[int] = None
+    ) -> Iterable[WalkNodeResult]:
```
chrisjsewell, line 252:
```suggestion
    ) -> Iterator[WalkNodeResult]:
```

sphuber:
> ```
>  sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 0 reached, connection timed out, timeout 1 (Background on this error at: http://sqlalche.me/e/3o7r)
> ```
> 
> Is this a known flaky test?

I haven't seen it fail yet, but it is relatively new. This looks like it comes from a test that was put in to check the REST API cleanly closing the SqlAlchemy sessions so we don't get a timeout because the queuepool is exhausted. Clearly this was triggered here where we run with reduced pool size and overflow, but don't see how your changes should have affected it. So I would just chalk this up to bad luck for now and ignore it until we start to see it more often

setup.json
```diff
@@ -107,7 +107,8 @@
       "prospector==1.2.0",
       "pylint==2.4.4",
       "toml~=0.10.0",
-      "yapf==0.28.0"
+      "yapf==0.28.0",
+      "mypy==0.780"
```
sphuber, line 111: We don't need this, right? Since we install in the `pre-commit` from the repo and this shouldn't rely on the `system`? At least you haven't marked it as such in the `pre-commit-config.yaml`. Or does this depend on being able to import resources from other dependencies in the environment?

sphuber:
BTW, I have just been reading up on a 2-year long discussion on `mypy` adding support for `pyproject.toml`. Seems it might be getting close to getting there: https://github.com/python/mypy/pull/5208
`pytest` also just merged it and should be releasing support with `v6.0`

setup.json
greschd: Yup, you're right. I'm just used to also putting it here so that my editor picks up the right version - but that sounds like a me problem 😄

greschd:
FYI, there is a potential for confusion because our ``orm`` and the stdlib ``typing`` share some names, specifically ``Dict`` and ``List``.

In my own packages I solved this by 
```python
import typing as ty
<...>
ty.Dict
```

I'm open for better suggestions, but since this is a problem that's bound to happen in many places in our code base, it's probably a good idea to have a consistent style.

mypy.ini
```diff
@@ -0,0 +1,41 @@
...
+; no_strict_optional = False
+
+; Enable all warnings
+; warn_redundant_casts = True
+; warn_unused_ignores = True
+; warn_return_any = True
+; warn_unreachable = True
+
+; allow_untyped_globals = False
+; strict_equality = True
```
sphuber, line 31: can we get rid of all these if we are not using them?

mypy.ini
```diff
@@ -0,0 +1,41 @@
...
+
+[mypy-numpy.*]
+ignore_missing_imports = True
+
+[mypy-plumpy.*]
+ignore_missing_imports = True
```
sphuber, line 37: Is this similar to `pylint` issuing missing import warnings if it is not being run in the "system" environment? If so, would it make sense to also run `mypy` under system?

mypy.ini
greschd: Yeah, although it might be nice seeing them listed here since (I think) the goal would be to slowly enable some of them.

mypy.ini
greschd: Mypy.. has its own "import system" - the problem here is that when dependencies don't have type annotations, that will be reported as a failure.

For now I'm not sure this makes a difference since we _also_ use ``skip`` for the ``follow_imports`` option, but we definitely want to get rid of that one eventually.

The import mechanism is described here: https://mypy.readthedocs.io/en/stable/running_mypy.html#how-mypy-handles-imports

sphuber:
Thanks @greschd

PR #4177, opened 2020-06-16
Typo: Keberos -> Kerberos

merkys:
I assume this to be a typo.

PR #4179, opened 2020-06-16
Updating the acknowledgments on the docs

giovannipizzi:
Also:
- fixing a typo for the work `Kerberos` (this replaces #4177)
- Changing the top-left logo from the AiiDA logo to a new
  AiiDA documentation logo (with the same Futura font, and the
  open `D` of AiiDA). As Nicola pointed out, the logo is confusing
  there because one would expect to just to the AiiDA homepage,
  rather than the docs homepage, by clicking on it. This change
  fixes the issue.
- Updating the description of AiiDA (homepage + about page) to
  be in sync with the recent updates on the aiida.net website,
  with the text revisited by Nicola


giovannipizzi:
Question: I've already removed the logo, as putting all of them becomes hard to maintain in a nice form.

Should we just remove the list completely and just link to aiida.net (link already there)?
(We can also merge this as is, and remove the list later if it evolves too quickly)

docs/source/index.rst
```diff
@@ -4,7 +4,7 @@ Welcome to AiiDA's documentation!
 
 **aiida-core version:** |release|
 
-`AiiDA`_ is an open-source framework implemented in Python that aims to help researchers with managing simple to complex workflows and making them fully reproducible (see :ref:`features<intro:about>`).
+`AiiDA`_ is an open-source Python infrastructure to automate, manage, persist, search and reproduce the complex workflows associated with modern computational science, and their associated data. (see :ref:`features<intro:about>`).
```
sphuber, line 7:
```suggestion
`AiiDA`_ is an open-source Python infrastructure to automate, manage, persist, search and reproduce the complex workflows and their associated data associated with modern computational science. (see :ref:`features<intro:about>`).
```

docs/source/intro/about.rst
```diff
@@ -4,17 +4,17 @@
...
-*  **Open Science:** Export subsets of your provenance graph and share them with peers or make them available online for everyone on the `Materials Cloud <https://www.materialscloud.org>`__.
-*  **Open source:** AiiDA is released under the `MIT open source license <LICENSE.txt>`__
+*  **Workflows:** AiiDA allows to build and execute complex, auto-documenting workflows linked to multiple codes on local and remote computers.
+*  **High-throughput:** AiiDA’s event-based workflow engine supports tens of thousands of processes per hour with full check-pointing.
+*  **Data provenance:** AiiDA automatically tracks and records inputs, outputs and metadata of all calculations in extensive provenance graphs that preserve and persist the full lineage of all parent and child data and calculations.
+*  **Advanced queries:** AiiDA’s query language enables fast graph-traversal queries on millions of nodes.
+*  **Plugin interface:** AiiDA can support any computational code and data analytics tool, data type, scheduler, transport mode, … via plugins (see `public plugin repository <https://aiidateam.github.io/aiida-registry/>`__).
```
chrisjsewell, line 15: is the `...` meant to be there, or a placeholder?

docs/source/conf.py
```diff
@@ -186,7 +186,7 @@
 
 # The name of an image file (relative to this directory) to place at the top
 # of the sidebar.
-html_logo = 'images/AiiDA_transparent_logo.png'
```
sphuber, line 189: We can then delete `images/AiiDA_transparent_logo.png` because it is no longer used

docs/source/conf.py
```diff
@@ -72,7 +72,7 @@
 
 # General information about the project.
 project = 'AiiDA'
-copyright = '2019, ECOLE POLYTECHNIQUE FEDERALE DE LAUSANNE (Theory and Simulation of Materials (THEOS) and National Centre for Computational Design and Discovery of Novel Materials (NCCR MARVEL)), Switzerland and ROBERT BOSCH LLC, USA. All rights reserved'
+copyright = '2020, ECOLE POLYTECHNIQUE FEDERALE DE LAUSANNE (Theory and Simulation of Materials (THEOS) and National Centre for Computational Design and Discovery of Novel Materials (NCCR MARVEL)), Switzerland and ROBERT BOSCH LLC, USA. All rights reserved'
```
sphuber, line 75: can you also update `LICENSE.txt` please?

docs/source/index.rst
sphuber: Do we need both "automate" and "manage". I have the feeling that one of the two should accurately cover the meaning of both

docs/source/index.rst
```diff
@@ -123,29 +123,36 @@ If you use AiiDA for your research, please cite the following work:
...
-   :height: 75px
-   :target: `MARVEL National Centre of Competence in Research`_
-
-.. image:: images/MaX.png
-   :height: 65px
-   :target: `MaX European Centre of Excellence`_
-
-.. image:: images/swissuniversities.png
-   :height: 30px
-   :target: `swissuniversities P-5 project "Materials Cloud"`_
```
sphuber, line 144: we can then also delete the corresponding images

docs/source/intro/about.rst
```diff
@@ -4,17 +4,17 @@
 What is AiiDA
 *************
 
-Implement data-driven and fully automated workflows with AiiDA.
+AiiDA is an open-source Python infrastructure to automate, manage, persist, search and reproduce the complex workflows associated with modern computational science, and their associated data.
```
sphuber, line 7: If changed in `index.rst` we should adopt this accordingly

docs/source/intro/about.rst
```diff
@@ -4,17 +4,17 @@
...
-*  **Plugin interface:** Extend AiiDA with `plugins <https://aiidateam.github.io/aiida-registry/>`__ for new simulation codes (input generation & parsing), data types, schedulers, transport modes and more.
-*  **Open Science:** Export subsets of your provenance graph and share them with peers or make them available online for everyone on the `Materials Cloud <https://www.materialscloud.org>`__.
-*  **Open source:** AiiDA is released under the `MIT open source license <LICENSE.txt>`__
+*  **Workflows:** AiiDA allows to build and execute complex, auto-documenting workflows linked to multiple codes on local and remote computers.
+*  **High-throughput:** AiiDA’s event-based workflow engine supports tens of thousands of processes per hour with full check-pointing.
+*  **Data provenance:** AiiDA automatically tracks and records inputs, outputs and metadata of all calculations in extensive provenance graphs that preserve and persist the full lineage of all parent and child data and calculations.
```
sphuber, line 13: what is the difference between "preserve" and "persist"? Feels a bit like a tautology to me. 
Also, we should probably change "calculations" for "workflows and calculations" or just "workflows" since that is described in the first point.

Also: "lineage of all parent and child data and calculations." I don't find very clear. I think "full lineage of all data" captures exactly the same and is a lot clearer.

docs/source/intro/about.rst
sphuber, line 15: would not use the ellipsis, but rather an explicit "etc." if necessary

docs/source/intro/about.rst
```diff
@@ -4,17 +4,17 @@
...
+*  **Advanced queries:** AiiDA’s query language enables fast graph-traversal queries on millions of nodes.
```
sphuber, line 14: strictly it is not just for "graph-traversal". Can we get rid off the "traversal"?

sphuber (changes requested):

docs/source/conf.py
csadorf: I believe that it is generally not necessary to update the year in these kind of notices, but do we have some kind of guideline for this? Should it say `2014-2020` instead?

docs/source/index.rst
csadorf: I generally don't like the style of the rephrased version, but if we do change it I'd propose `and all associated` instead of `and their associated data`.

docs/source/index.rst
```diff
@@ -123,29 +123,36 @@ If you use AiiDA for your research, please cite the following work:
...
+* the `INTERSECT`_ project, funded by the Horizon 2020 H2020-NMBP-TO-IND-2018 call, Grant No. 814487;
+* the `PASC`_ Swiss platform for Advanced Scientific Computing;
+* the `swissuniversities P-5 project "Materials Cloud"`_, project ID 182-008;
+* the `NFFA`_ project, funded by the Horizon 2020 programm, Grant No. 654360;
+* the BIG-MAP Battery 2030+ initiative;
+* `EPFL`_ (École Polytechnique Fédérale de Lausanne);
+* `Harvard University`_;
+* the `Bosch Research and Technology Center`_ (USA).
+
+A full list of supporting partners and institutions, with the respective logos, is available on the `AiiDA acknowledgements page`_.
```
csadorf, line 140: Since we are already linking to the acknowledgements page on the homepage, should we maybe just list all funding sources their and restrict it here to only the most important ones? Those lists are likely going to get out of sync and then we avoid the need to maintain both of them.

docs/source/intro/about.rst
```diff
@@ -4,17 +4,17 @@
 What is AiiDA
 *************
 
-Implement data-driven and fully automated workflows with AiiDA.
+AiiDA is an open-source Python infrastructure to automate, manage, persist, search and reproduce the complex workflows associated with modern computational science, and their associated data.
 
-AiiDA's primary features include:
+AiiDA is built to support and streamline its four core pillars of Automation, Data, Environment, and Sharing (the ADES model, described `here <https://arxiv.org/abs/1504.01163>`__). Some of the key features of AiiDA include:
```
csadorf, line 9:
```suggestion
AiiDA is built to support and streamline the four core pillars of the ADES model: Automation, Data, Environment, and Sharing (described `here <https://arxiv.org/abs/1504.01163>`__).
Some of the key features of AiiDA include:
```

csadorf (changes requested):
I think the changes are generally fine, however I think it would be better to keep unrelated changes in separate PRs in the future.

We also should be careful that the feature list does not get too long and that we make sure it remains user/problem-oriented. I think the length is ok as it is, but it should not get any longer.

docs/source/conf.py
giovannipizzi: LICENSE.txt updated. I agree that it's better to put the starting date as well, so I've now added it (set to 2014 both here and in LICENSE.txt)

docs/source/conf.py
giovannipizzi: Done

docs/source/index.rst
giovannipizzi: I think manage conveys more the meaning of "organise". I applied Simon's suggestion, and rephrased a bit to be "midway" between the old and the new. Let me know if it is now ok.

docs/source/index.rst
giovannipizzi: Indeed, this was one of my comments/questions. I agree, and I have now simplified.

docs/source/index.rst
giovannipizzi: I was waiting for the review to know if it was ok to remove. Now done.

docs/source/intro/about.rst
giovannipizzi: Done

docs/source/intro/about.rst
giovannipizzi: Done

docs/source/intro/about.rst
giovannipizzi: Ok, I tried to adapt the text to address the various comments.

docs/source/intro/about.rst
giovannipizzi: Done

docs/source/intro/about.rst
giovannipizzi: It was supposed to be there, see my reply to Sebastiaan's comment as well

giovannipizzi: OK

giovannipizzi (commented):
Thanks for the review. I have addressed all points and it's ready to re-review. 

Regarding the PR, I have put the various fixes together because they originated from a single discussion/request by Nicola (+ the fix of the typo for Kerberos, that I was explicitly asked to include in this PR).

csadorf (approved):

sphuber (approved):

sphuber:
Thanks a lot @giovannipizzi

PR #4187, opened 2020-06-19
`CalcJobResultManager`: fix bug that broke tab completion

### LINKED ISSUE #4184: Tab completion not working for `CalcJob.res`
### Describe the bug

`calcjob.res.<TAB>` should complete the keys of the output dictionary, but it doesn't.

### Steps to reproduce

1. load a CalcJobNode that uses a plugin that supports the `.res` method and has parsed results
2. Try to type `calcjob.res.` and press TAB twice. Nothing happens.

### Expected behavior

Tab completion should work

### Additional context

I addressed this in #3985 for `.inputs`, `.outputs` but I fear I forgot about `.res`.

sphuber:
Fixes #4184 

For tab-completion to work properly, the `__getattr__` method need to
properly return an `AttributeError` if the corresponding attribute does
not exist, however, it was accidentally raising a `KeyError` since the
implementation caught an `AttributeError` but the dereferencing was
using `__getitem__` so would never raise an `AttributeError`.

Likewise, the `__getitem__` implementation was incorrectly returning an
`AttributeError` instead of a `KeyError` making it behave different from
any other mapping-like data structure.


aiida/orm/utils/calcjob.py
```diff
@@ -97,17 +97,17 @@ def __getattr__(self, name):
         """
         try:
             return self.get_results()[name]
-        except AttributeError:
+        except KeyError:
```
giovannipizzi, line 100: I think we need to guard also for any exception that can happen in `get_results`, both here and below.

I found at least two possibilities, both a `ValueError`:
```
ValueError: cannot load results because process class cannot be loaded: no process type for CalcJobNode<None>: cannot recreate process class

ValueError: cannot load results as the default node could not be retrieved: no neighbor with the label output_parameters found
```
I am not sure if there is still some case if an AttributeError can still be raised.

aiida/orm/utils/calcjob.py
```diff
@@ -97,17 +97,17 @@ def __getattr__(self, name):
...
         :param name: name of the result return
         :return: value of the attribute
-        :raises AttributeError: if the results dictionary does not contain an attribute with the given name
+        :raises KeyError: if the results dictionary does not contain an attribute with the given name
         """
         try:
             return self.get_results()[name]
-        except AttributeError:
-            raise AttributeError("Default result node<{}> does not contain key '{}'".format(self._result_node.pk, name))
+        except KeyError:
```
giovannipizzi, line 112: Also here

aiida/orm/utils/calcjob.py
giovannipizzi: For tests, in the first case I just created an unstored `CalcJobNode()`; in the second one, a calculation that excepted and hand no output node.

Issue #2205, opened 2018-11-15
pylint and absolute imports


ltalirz:
The `provenance-design` branch flagged up a couple of prospector warnings such as these:
```
aiida/orm/implementation/querybuilder.py
  Line: 143
    pylint: import-error / Unable to import 'sqlalchemy.sql.elements' (col 8)
    pylint: no-name-in-module / No name 'sql' in module 'sqlalchemy' (col 8)
  Line: 144
    pylint: import-error / Unable to import 'sqlalchemy.orm.attributes' (col 8)
    pylint: no-name-in-module / No name 'orm' in module 'sqlalchemy' (col 8)
  Line: 145
    pylint: import-error / Unable to import 'sqlalchemy.sql.expression' (col 8)
    pylint: no-name-in-module / No name 'sql' in module 'sqlalchemy' (col 8)
  Line: 146
    pylint: import-error / Unable to import 'sqlalchemy.types' (col 8)
    pylint: no-name-in-module / No name 'types' in module 'sqlalchemy' (col 8)
```

The reason they show up is because pylint (somewhat [1]) incorrectly treats these as relative imports (there are `django` and `sqlalchemy` subfolders in `aiida/orm/implementation`).
There is an open issue for this (https://github.com/PyCQA/pylint/issues/1931) but no workaround as of now.

Since this is a pylint problem, I will disable the corresponding warning in affected files, together with a link to the bug report, with the idea that the pylint directives can be removed again in the future.

Mentioning @muhrin @sphuber @giovannipizzi 

[1] "Somewhat" because trying to run the file individually (i.e. `python aiida/orm/implementation/querybuilder.py`) will fail. Only running it as part of a package (`python -m aiida.orm.implementation.querybuilder` will cause python to honour the absolute import. One can argue, however, that python submodules of packages are not meant to be executed individually.

giovannipizzi:
The simplest probably is to refactor the code at this stage, renaming our django + sqlalchemy folders to something else, like `backend_django`, `backend_sqlalchemy`?

ltalirz:
While it might be safer, I don't find this simpler. Why should a package like aiida that contains backends for sqlalchemy and django not container subfolders named sqlalchemy and django?

Also, keep in mind that this applies to all dependencies, not just django and sqla.
What if someone decides we need a package named `utils` ?

giovannipizzi:
Actually this is a prospector problem and not a pylint problem, reopening

giovannipizzi:
[...]

PR #4190, opened 2020-06-19
Update the pre-commit configuration 

sphuber:
* Drop `prospector`: it was only being used to run `pylint` but in that
   case it is much simpler to just run it directly as `prospector` has
   often given issues with incompatible dependencies
 * Add more hooks from `pre-commit` itself:
   - end-of-file-fixer    automatically add new lines to end of files
   - fix-encoding-pragma  automatically add the utf-8 encoding pragma
   - mixed-line-ending    automatically deal with line break characters
 * Run `pylint` and `yapf` directly from the remote repository
   Since `pylint` still needs to be run with `system` such that it can
   see all dependencies and not throw false import errors, we still need
   to keep it in the `pre-commit` extras requirements.
 * Renamed extra requires categories to match other projects
   dev_precommit -> pre-commit
   testing -> tests


chrisjsewell:
> prospector has often given issues with incompatible dependencies

Yes! It's always been a pain making it compatible with `flake8` that I use for other packages.

I was just having a scan down https://pre-commit.com/hooks.html, to see if there are any other good ones. There is a few for checking the order of imports, but that may be overkill.
What we don't currently do is any specific linting of documentation, maybe:

- `rst-backticks` - Detect common mistake of using single backticks when writing rst
- or https://github.com/myint/rstcheck#use-as-a-pre-commit-hook, is pretty good

sphuber:
> What we don't currently do is any specific linting of documentation, maybe:
> 
>     * `rst-backticks` - Detect common mistake of using single backticks when writing rst
> 
>     * or https://github.com/myint/rstcheck#use-as-a-pre-commit-hook, is pretty good

Sounds good, but I explicitly didn't add this here. In some of the other repos I already added the `pydocstyle` linter to check the docstrings, but that would be a lot of work here. Wanted to postpone that for a bit. But once we add this, we can look at the backtick fixer as well

ltalirz:
Hi guys, sorry for being a bit late to the game but I find the change of the extras a bit problematic.
Please note that plugins/CI workflows may rely on these and will now break silently (because pip does not complain for incorrect extra names).

One possible solution is to add the old extras back in the `setup.py` and to add a deprecation warning to the CHANGELOG

sphuber:
> Hi guys, sorry for being a bit late to the game but I find the change of the extras a bit problematic.
> Please note that plugins/CI workflows may rely on these and will now break silently (because pip does not complain for incorrect extra names).

Sorry if this change is causing issues, that is of course not intended. I figured that since these are merely extras for development, it would be fine to change them. Can you tell me what other packages depend on it?
Also, maybe we should clarify what `tests` is intended for. I was under the impression that this is just for development of `aiida-core` and not necessarily to be used by plugins. They should define their own testing requirements I feel.

ltalirz:
>  I was under the impression that this is just for development of aiida-core and not necessarily to be used by plugins. They should define their own testing requirements I feel.

I agree with you here, and since pip does not (or at least did not, when I last checked) properly support specifying a dependency on the same package twice (e.g. `aiida-core` under `install_requires` and `aiida-core[testing]` under `extras_require['testing']`), all plugins I've now checked actually follow your advice.
I do think this is a "lucky case", though - if pip did support this I bet there would be a number of plugins reusing the aiida-core extras.

I've looked around a bit more - instances that will need to be changed are:
* https://github.com/marvel-nccr/ansible-role-aiida/blob/9ee58f9cc59f7850d1d5d67fb5b7ca707481c205/defaults/main.yml#L4-L9 @chrisjsewell 
 * https://github.com/aiidateam/aiida-core/wiki/Development-environment

greschd:
Relying on a dependency to pull in packages that are used _directly_ in the plugin code is just plain bad practice. There's no way we can guarantee that some dependency we're pulling in will always be there.

There is some nuance to this, though: the pytest fixtures and sphinx extension are _features of AiiDA_, not just internal dev utilities. So IMO the requirements we use there should be in a "feature" extra, just like e.g `atomic-tools`. In this case the user shouldn't have to care about exactly which dependencies are needed to implement these features, because they _don't_ use the dependencies directly.

giovannipizzi:
Just FYI, prospector just made a new release that addresses our original issue #2205, in case you are using prospector in another project.
See [here](https://github.com/PyCQA/prospector/issues/294)

PR #4191, opened 2020-06-20
Process on finish call without using LoopCommunicator

unkcpz:
In PR #4154, the LoopCommunicator was used as the runner's communicator.
This makes it easy to schecule callbacks into the correct event loops,
however, the `Process` has its own mechanism to do this. In object
method `_schedule_rpc` process schedule the callback into its event
loop. Using `LoopCommunitor` lead to a problem that when controller send
the controlling message by rpc for example when calling
`controller.pause_process` the `pause` subscriber is run by
`plumpy.futures.create_task`, which is not necessary. 
I didn't find using `LoopCommunicator` cause any exceptions, however, it is not expected in design correctly? Therefore, I propose to conservatively not use `LoopCommunicator` as communicator one for all in `runner`. 

`call_on_process_finish` receive a callback which not expected to be a
coroutine (otherwise it is supposed to be wait there and `inline_callback` also should be a coroutine. Through, this not cause problem for `tornado`.), however, in `action_awaitables` we schedule `_run_task` which
is a corotinefunction to it. The `call_soon` method of Process object is
more proper here.

### DIFF
--- a/aiida/engine/processes/futures.py
+++ b/aiida/engine/processes/futures.py
@@ -9,6 +9,8 @@
 ###########################################################################
 # pylint: disable=cyclic-import
 """Futures that can poll or receive broadcasted messages while waiting for a task to be completed."""
+import functools
+
 import tornado.gen
 
 import plumpy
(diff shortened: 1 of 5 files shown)


muhrin:
Thanks Jason!  So, in my opinion this is not the right solution.

Let me outline some of the design decisions that I think we should stick to and see if we can figure out a solution that's compatible with all of them.  I'm happy to have a video chat if it helps.

In this PR you've changed it so that it's the clients (i.e. the Process) that is responsible for scheduling the callback on the event loop.  This, to me, is highly undesirable as now every time someone wants to subscribe themselves for a callback from the communicator they have to remember to schedule it on the loop which is a violation of the DRY principle and what's more it creates an unnecessary coupling.  The process shouldn't know about this 'implementation detail' that the callbacks come from another thread and are scheduled on the event loop.  Only the runner and manager should know this because they create the `LoopCommunicator`.  Imagine tomorrow we decide to remove the event loops an and use another solution, well now we have to change all the communicator subscription calls throughout the code (because there is this artifical coupling).  Before, with the loop communicator, those calls wouldn't change.

unkcpz:
Thanks, I think both two aspects of the PR are useless, so I'd better close this.

Let me try to recap here:
1. callback in function `call_on_process_finished` can be either and function or a coroutine. `_run_task` is also supposed to be used here. No reason to think of it as a regular function.

2. It is true that callback invoked by the communicator(rpc or a new process to be launched) scheduled to the runner's event loop twice (one in `_schedule_rpc` while other in `plumpy.create_task`). But it is necessary to have the `LoopCommunicator` to hide the specific callback schedule implementation(wrap a new callback with `plumpy.create_task`) .  And Process should make sure everything it does (like pausing) is scheduled on its own event loop so it has its own `_schedule_rpc` method.

PR #4192, opened 2020-06-21
`CalcJob`: move job resource validation to the `Scheduler` class

### LINKED ISSUE #3887: Lsf scheduler wrong with `num_machines`
aiida-core version: 1.1.1

1. When I set `metadata.options.resources` without `num_machines`, it raises "aiida.common.exceptions.InputValidationError: the `resources` input has to at least include `num_machines`."
2. When I set  `metadata.options.resources` with `num_machines: None`, it raises "aiida.common.exceptions.InputValidationError: the input `resources.num_machines` shoud be an integer."
3. When I set `metadata.options.resources` with `num_machines: 1`, it raises "aiida.common.exceptions.ConfigurationError: num_machines cannot be set for LSF scheduler"

btw, when I use aiida-core v1.0.1 and set `metadata.options.resources` without `num_machines`, aiida can work properly.

sphuber:
Fixes #3887 

This is necessary because the resource validation is scheduler
dependent. Up till now, the `CalcJob` defined a validator on the entire
input namespace that validates the `metadata.options.resources` input.
Specifically, it demanded that the `num_machines` keyword was defined,
which is indeed a requirement for the `NodeNumberJobResource`, used for
schedulers like SLURM, however, this field doesn't even make sense for
schedulers like LSF and SGE that use the `ParEnvJobResource` type of job
resources.

The solution is to delegate the validation of the resources to the
scheduler which in turn delegates it to the `JobResource` class that it
uses. The validation used to happend on construction of the job resource
instance, but is factored out to the `validate_resources` classmethod.
This allows it to be called without having to construct an instance
which is more efficient.

Finally, the signature of the `CalcJob` input validators is changed and
no longer raise an `InputValidationError` but instead return an error
message as the interface of `Port.validator` requires. The port itself
will detect if an error message is returned and raise a `ValueError`
with all the relevant error messages.

### DIFF
--- a/aiida/engine/processes/calcjobs/calcjob.py
+++ b/aiida/engine/processes/calcjobs/calcjob.py
@@ -81,8 +81,15 @@ def validate_calc_job(inputs, ctx):  # pylint: disable=inconsistent-return-state
     except KeyError:
         return 'input `metadata.options.resources` is required but is not specified'
 
+    default_mpiprocs_per_machine = computer.get_default_mpiprocs_per_machine()
+    num_mpiprocs_per_machine = resources.get('num_mpiprocs_per_machine', None)
+    tot_num_mpiprocs = resources.get('tot_num_mpiprocs', None)
+    if num_mpiprocs_per_machine is None and tot_num_mpiprocs is None and default_mpiprocs_per_machine is not None:
+        # Only set the default value if tot_num_mpiprocs is not provided. Otherwise, it means that the user provided
+        # both `num_machines` and `tot_num_mpiprocs`, and we have to ignore the `num_mpiprocs_per_machine` default.
+        resources['num_mpiprocs_per_machine'] = default_mpiprocs_per_machine
+
     try:
-        resources['default_mpiprocs_per_machine'] = computer.get_default_mpiprocs_per_machine()
         scheduler.validate_resources(**resources)
     except (ValueError, TypeError) as exception:
         return 'input `metadata.options.resources` is not valid for the {} scheduler: {}'.format(scheduler, exception)
@@ -398,9 +405,13 @@ class CalcJob(Process):
 
         # Set resources, also with get_default_mpiprocs_per_machine
         resources = self.node.get_option('resources')
-        def_cpus_machine = computer.get_default_mpiprocs_per_machine()
-        if def_cpus_machine is not None:
-            resources['default_mpiprocs_per_machine'] = def_cpus_machine
+        default_mpiprocs_per_machine = computer.get_default_mpiprocs_per_machine()
+        num_mpiprocs_per_machine = resources.get('num_mpiprocs_per_machine', None)
+        tot_num_mpiprocs = resources.get('tot_num_mpiprocs', None)
+        if num_mpiprocs_per_machine is None and tot_num_mpiprocs is None and default_mpiprocs_per_machine is not None:
+            # Only set the default value if tot_num_mpiprocs is not provided. Otherwise, it means that the user provided
+            # both `num_machines` and `tot_num_mpiprocs`, and we have to ignore the `num_mpiprocs_per_machine` default.
+            resources['num_mpiprocs_per_machine'] = default_mpiprocs_per_machine
         job_tmpl.job_resource = scheduler.create_job_resource(**resources)
 
         subst_dict = {'tot_num_mpiprocs': job_tmpl.job_resource.get_tot_num_mpiprocs()}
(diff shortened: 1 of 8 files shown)


sphuber:
N.B.: this PR assumes that there are no schedulers out there in the wild that subclass their own `JobResource`. If that is the case, the code will break, because the `validate_resources` is an `abstractclassmethod` and therefore the custom subclasses won't be instantiable until they implement it. If we think this is a problem, we could make it concrete and just `pass`. We do need to be able to notify subclasses that they need to start implementing this method at some point in that case.

Second important point: the `ArithmeticAddCalculation` implementation specifies `{'num_machines': 1, 'num_mpiprocs_per_machine': 1}` as default for the `metadata.options.resources`. This was done because this class is used often for demonstrations and tutorials, and not having to specify the resources simplifies the examples. However, we now know that this default is incorrect as it only caters to schedulers with job resources of the type `NodeNumberJobResource`, which is not always the case. So really we have to remove this default, but this will impact the simplicity of the tutorial and documentation. What do we do here?

sphuber:
@giovannipizzi it would probably be good to ship this with 1.3.0 given that we have had multiple users face this problem. I have added a second commit on top of the original one, which contains just some cleaning of the scheduler base class and job resource implementations. Especially the `validate_resources` of the node number sub class I have significantly refactored in an attempt to make it more readable. I found the original nested if-blocks very difficult to grok.

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -398,9 +405,13 @@ def presubmit(self, folder):
...
         resources = self.node.get_option('resources')
-        def_cpus_machine = computer.get_default_mpiprocs_per_machine()
-        if def_cpus_machine is not None:
-            resources['default_mpiprocs_per_machine'] = def_cpus_machine
+        default_mpiprocs_per_machine = computer.get_default_mpiprocs_per_machine()
+        num_mpiprocs_per_machine = resources.get('num_mpiprocs_per_machine', None)
+        tot_num_mpiprocs = resources.get('tot_num_mpiprocs', None)
+        if num_mpiprocs_per_machine is None and tot_num_mpiprocs is None and default_mpiprocs_per_machine is not None:
+            # Only set the default value if tot_num_mpiprocs is not provided. Otherwise, it means that the user provided
+            # both `num_machines` and `tot_num_mpiprocs`, and we have to ignore the default value of `tot_num_mpiprocs`.
```
giovannipizzi, line 413:
```suggestion
            # both `num_machines` and `tot_num_mpiprocs`, and we have to ignore the default value of `num_mpiprocs_per_machine`.
```

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -398,9 +405,13 @@ def presubmit(self, folder):
 
         # Set resources, also with get_default_mpiprocs_per_machine
         resources = self.node.get_option('resources')
-        def_cpus_machine = computer.get_default_mpiprocs_per_machine()
-        if def_cpus_machine is not None:
-            resources['default_mpiprocs_per_machine'] = def_cpus_machine
+        default_mpiprocs_per_machine = computer.get_default_mpiprocs_per_machine()
```
giovannipizzi, line 408: The general doubt I have here is if this is a general logic, or still specific to some scheduler types. Do we have tests for a parenv job resource, or can we ask someone (@zhubonan, @atztogo, ...) to check if this new logic now works also for them?

sphuber: I think you are right and this will fail. Even if the user does not explicitly specify `num_mpiprocs_per_machine` and `tot_num_mpiprocs`, if the computer has `default_mpiprocs_per_machine` defined (which I think currently is almost always the case, or at least, `verdi computer setup` suggests to set it), then `num_mpiprocs_per_machine` will be added to the `resources` and since I changed `ParEnvJobResource.validate_resources` to raise when it gets unsupported keywords, this will fail.

I am not sure where and how to fix it. Easy hack would be to simply ignore this keyword in the `ParEnvJobResource` but is not the *correct* solution. If specifying a number of machines for a `ParEnvJobResource` based scheduler never makes sense, maybe we should not even have computers that are configured with such a scheduler, define a `num_mpiprocs_per_machine`.

sphuber:
@giovannipizzi I have a added a new commit that I think should address the problem you mentioned. I added a test that launches a calculation job using a computer with SGE scheduler, and I confirmed that it failed as I explained in my response to your question. The fix in the last commit addresses the problem. I am not sure if there are still other failure points hiding out there. @atztogo and @zhubonan if you have the time to give this branch a spin with a real calculation on a LSF or SGE cluster, that would be great!

atztogo:
Thank you very much @sphuber . I ran some calculations using 80fef5ef , and went OK without num_machines with sge.

PR #4195, opened 2020-06-22
Only color directories in `verdi node repo ls --color`

### LINKED ISSUE #4183: verdi node repo ls --color uses a single color
### Describe the bug

The `--color` option of `verdi node repo ls` and related commands just shows everything in blue. This is not very useful.

### Steps to reproduce

1. Run `verdi node repo ls <ID> --color`

### Expected behavior

According to the docs, the color should help distinguish directories from files.

The line to fix is here: 
https://github.com/aiidateam/aiida-core/blob/98e21dac650fc1e7e79cb69eeebfadaf90c21c24/aiida/cmdline/utils/repository.py#L28

sphuber:
Fixes #4183 

The help string says that the `--color` flag can be used to color
directories different from files to easily distinguish them, except the
implementation colored all files, regardless of type.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -145,7 +145,7 @@ function-name-hint=(([a-z][a-z0-9_]{2,40})|(_[a-z0-9_]*))$
 function-rgx=(([a-z][a-z0-9_]{2,30})|(_[a-z0-9_]*))$
 
 # Good variable names which should always be accepted, separated by a comma
-good-names=x,y,z,i,j,k,ex,Run,_,pk,TemplatereplacerCalculation,ArithmeticAddCalculation,MultiplyAddWorkChain
+good-names=x,y,z,i,j,k,ex,Run,_,pk,fg,TemplatereplacerCalculation,ArithmeticAddCalculation,MultiplyAddWorkChain
 
 # Include a hint for the correct naming format with invalid-name
 include-naming-hint=no
--- a/aiida/cmdline/utils/repository.py
+++ b/aiida/cmdline/utils/repository.py
@@ -8,7 +8,6 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Utility functions for command line commands operating on the repository."""
-
 import click
 
 
@@ -23,8 +22,5 @@ def list_repository_contents(node, path, color):
 
     for entry in node.list_objects(path):
         bold = bool(entry.type == FileType.DIRECTORY)
-
-        if color:
-            click.secho(entry.name, bold=bold, fg='blue')
-        else:
-            click.secho(entry.name, bold=bold)
+        fg = 'blue' if color and entry.type == FileType.DIRECTORY else None
+        click.secho(entry.name, bold=bold, fg=fg)
--- a/tests/cmdline/utils/test_repository.py
+++ b/tests/cmdline/utils/test_repository.py
@@ -1,0 +1,62 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+# pylint: disable=redefined-outer-name,invalid-name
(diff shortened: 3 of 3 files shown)


sphuber:
@giovannipizzi I realize that you assigned this issue to yourself, but given that this is the [last open issue for the `v1.2.2` milestone](https://github.com/aiidateam/aiida-core/milestones/v1.2.2) I wanted to get those done as soon as possible. This way I can already prepare the release and start focusing on other responsibilities. Hope you don't mind.

sphuber:
Pinging @greschd and @chrisjsewell for question about `pytest`: I wanted to test whether the function also correctly colored the output when `color=True` but could not figure out how to do it. I know that to change the color or make the text bold, `click.secho` uses ANSI escape sequences like `\033[1m` for bold, but not sure how to compare those. Using the `capsys` fixture I don't see them when printing. I though using the `capsysbinary` instead, but even there when printing the captured bytes, the escape sequences were not printed. Maybe they are just stripped when printing that captured output through `pytest` when using the `-s` flag? I.e. if I were to just check for those bytes it should maybe work? Or is there another way that you know of?

greschd:
I think it's `click` doing some auto-detection and stripping the escape codes when the output goes to a file. Reason I'm thinking this is this comment: https://github.com/pallets/click/blob/db06b61fd0399c63bbdb26d8d47b84b29aa94ecf/src/click/termui.py#L531

chrisjsewell:
Yep you just beat me to it! ANSI codes are automatically hidden if the destination file is not a
terminal: https://github.com/pallets/click/blob/db06b61fd0399c63bbdb26d8d47b84b29aa94ecf/src/click/utils.py#L188-L189

greschd:
Yup, here it is: https://github.com/pallets/click/blob/db06b61fd0399c63bbdb26d8d47b84b29aa94ecf/src/click/utils.py#L254

The `CliRunner.invoke` apparently lets you keep the color: https://github.com/pallets/click/blob/db06b61fd0399c63bbdb26d8d47b84b29aa94ecf/src/click/testing.py#L255

sphuber:
Thanks for the pointers guys. So now the trick is to figure out how to get `click` *not* to strip those things without going through the runner, since I am testing a free standing function and not a `Command`

greschd:
In the end this is about a command though, right? Couldn't you switch to using the runner, or do both?

greschd:
As for how to convince `click` it's writing to a file: wrap the file object you're writing to in a proxy that returns `True` from `.isatty()`:
https://stackoverflow.com/questions/8389155/how-to-convince-python-stream-objects-to-return-true-from-isatty

sphuber:
> In the end this is about a command though, right? Couldn't you switch to using the runner, or test both?

Well the command is really just calling a utility function `aiida.cmdline.utils.repository.list_repository_contents` so I want to just test that and not the whole command invocation. Anyway, I found that you can use the `CliRunner.isolation` context manager and set `color=True` to stop stripping the ANSI sequences. This should have worked, but it still kept stripping the color code, but the bold sequence remained. I scratched my idiot head for a full hour, until it finally dawned on me: maybe the test may actually be doing its job and the implementation might be broken ?? :no_mouth:

Sure enough, the code was just broken and I was overriding the `color` variable within the loop :man_facepalming: 

Should be all fixed now

greschd:
> 
> Anyway, I found that you can use the `CliRunner.isolation` context manager and set `color=True` to stop stripping the ANSI sequences.

Nice! Here's the ugly "manual" solution I came up with in the meantime 😅

```python
import io
import click
from contextlib import redirect_stdout

def test_secho(capsys):
    with capsys.disabled():
        click.secho('without_redirect', fg='red', bold=True)
        with io.StringIO() as capt:
            capt.isatty = lambda: True
            with redirect_stdout(capt):
                click.secho('with_redirect', fg='red', bold=True)
            capt.seek(0)
            res = capt.read()
            print(res)
```

sphuber:
> All good, except that I fear that the order of the output in both tests might not be determined? and we need to do `sorted(values.splitlines())1`? (Not sure if `splitlines` exists for `bytes`, in case you can probably do `.split('\n')`?)

This would be true if we rely solely on the order of `os.listdir` which is [determined by the operating system](https://docs.python.org/3/library/os.html#os.listdir) and so i "random", however, in `Repository.list_objects` we call `sorted` for the results returned by `os.listdir` so for this utility function the results will always be sorted. If we ever implement a feature in `verdi node repo ls` to change the ordering, the test will have to be adapted, but it will fail, so we will notice. I would leave this like this for now

PR #4196, opened 2020-06-23
Remove all files from the pre-commit exclude list

### LINKED ISSUE #2207: Reduce/remove files from the exclusion list of pre-commit, prospector, yapf, pylint

### LINKED ISSUE #3885: seemingly random changes in pre-commit
Something strange happened in PR https://github.com/aiidateam/aiida-core/pull/3875

After some time working on the PR, pylint suddenly started complaining about invalid variable names in `.ci/workchains.py` (a file untouched by the PR):
https://github.com/aiidateam/aiida-core/blob/035a9daa4e3b9282419eb2e5c58621d23cee1792/.ci/workchains.py#L71-L72

I added a commit that added `  # pylint: disable=invalid-name` comments and the problem went away.

@sphuber in the end rebased my PR, somehow getting rid of these additional changes in the process. So far, so good.

The strange thing is: The pre-commit check on the final commit of the PR *passed*, despite the fact the comments were no longer there.

It seems we are witnessing the same again in PR https://github.com/aiidateam/aiida-core/pull/3738 - pre-commit now started complaining about these two lines again.

@csadorf Do you have any idea what could be going on?

sphuber:
Fixes #2207 and fixes #3885

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,104 +8,16 @@ repos:
     -   id: mixed-line-ending
     -   id: trailing-whitespace
 
--   repo: https://github.com/PyCQA/pylint
-    rev: pylint-2.5.2
-    hooks:
-    -   id: pylint
-        language: system
-        exclude: &exclude_files >
-            (?x)^(
-                .ci/workchains.py|
-                aiida/backends/djsite/queries.py|
-                aiida/backends/djsite/db/models.py|
-                aiida/backends/djsite/db/migrations/0001_initial.py|
-                aiida/backends/djsite/db/migrations/0002_db_state_change.py|
-                aiida/backends/djsite/db/migrations/0003_add_link_type.py|
-                aiida/backends/djsite/db/migrations/0004_add_daemon_and_uuid_indices.py|
-                aiida/backends/djsite/db/migrations/0005_add_cmtime_indices.py|
-                aiida/backends/djsite/db/migrations/0006_delete_dbpath.py|
-                aiida/backends/djsite/db/migrations/0007_update_linktypes.py|
-                aiida/backends/djsite/db/migrations/0008_code_hidden_to_extra.py|
-                aiida/backends/djsite/db/migrations/0009_base_data_plugin_type_string.py|
-                aiida/backends/djsite/db/migrations/0010_process_type.py|
-                aiida/backends/djsite/db/migrations/0011_delete_kombu_tables.py|
-                aiida/backends/djsite/db/migrations/0012_drop_dblock.py|
-                aiida/backends/djsite/db/migrations/0013_django_1_8.py|
-                aiida/backends/djsite/db/migrations/0014_add_node_uuid_unique_constraint.py|
-                aiida/backends/djsite/db/migrations/0016_code_sub_class_of_data.py|
-                aiida/backends/djsite/db/migrations/0017_drop_dbcalcstate.py|
-                aiida/backends/sqlalchemy/migrations/versions/0aebbeab274d_base_data_plugin_type_string.py|
-                aiida/backends/sqlalchemy/migrations/versions/35d4ee9a1b0e_code_hidden_attr_to_extra.py|
-                aiida/backends/sqlalchemy/migrations/versions/59edaf8a8b79_adding_indexes_and_constraints_to_the_.py|
-                aiida/backends/sqlalchemy/migrations/versions/6c629c886f84_process_type.py|
-                aiida/backends/sqlalchemy/migrations/versions/70c7d732f1b2_delete_dbpath.py|
-                aiida/backends/sqlalchemy/migrations/versions/89176227b25_add_indexes_to_dbworkflowdata_table.py|
-                aiida/backends/sqlalchemy/migrations/versions/a514d673c163_drop_dblock.py|
-                aiida/backends/sqlalchemy/migrations/versions/a6048f0ffca8_update_linktypes.py|
-                aiida/backends/sqlalchemy/migrations/versions/e15ef2630a1b_initial_schema.py|
-                aiida/backends/sqlalchemy/migrations/versions/f9a69de76a9a_delete_kombu_tables.py|
-                aiida/backends/sqlalchemy/migrations/versions/62fe0d36de90_add_node_uuid_unique_constraint.py|
-                aiida/backends/sqlalchemy/migrations/versions/a603da2cc809_code_sub_class_of_data.py|
-                aiida/backends/sqlalchemy/migrations/versions/162b99bca4a2_drop_dbcalcstate.py|
-                aiida/backends/sqlalchemy/models/computer.py|
-                aiida/backends/sqlalchemy/models/settings.py|
-                aiida/backends/sqlalchemy/models/node.py|
-                aiida/backends/utils.py|
-                aiida/common/datastructures.py|
-                aiida/engine/daemon/execmanager.py|
-                aiida/engine/processes/calcjobs/tasks.py|
-                aiida/orm/querybuilder.py|
-                aiida/orm/nodes/data/array/bands.py|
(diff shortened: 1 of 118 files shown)


sphuber:
@ramirezfranciscof and @chrisjsewell this bad boy is ready to go!

sphuber:
To anyone reading this **Note that this should not be merged!**

ltalirz:
That's fantastic @sphuber  :-)

One small question: I notice that you moved (in a previous PR) the pylint hook from `local` to `repo: https://github.com/PyCQA/pylint`- but then of course you are still using `language: system` since it needs to see AiiDA and its dependencies.
Just wondering: does this not install a separate virtual environment with pylint inside that is never used?
I.e. would it not be better to move the `pylint` hook back under the local ones?

sphuber:
> That's fantastic @sphuber :-)

Combination of OCD and procrastination is one heck of a drug ;)

> One small question: I notice that you moved (in a previous PR) the pylint hook from `local` to `repo: https://github.com/PyCQA/pylint`- but then of course you are still using `language: system` since it needs to see AiiDA and its dependencies.
> Just wondering: what are the benefits of this? Does this not install a separate virtual environment with pylint inside that is never used?
> I.e. would it not be better to move the `pylint` hook back under the local ones?

As you said, we need to run this with `language: system` such that we don't get spurious `import-error` warnings. However, we also want to run with a very specific version of `pylint` because changes in that version can cause differences between local and CI environment and therefore annoying CI failures. We used to "solve" this by pinning the requirement in the requirements. However, this would cause problems when developing multiple packages in the same virtual environment. Often there would be version conflicts. So you would run pre-commit with `aiida-core` and so that would require a specific `pylint` version, but then `aiida-quantumespresso` would be still on a different version and so when you commit, `pre-commit` would fail. By using the remote hook, the separate virtual environment solves these dependency incompatibilities. I think this is worth the little overhead, because anyway the environment is cached.

> P.S. I'm curious - did removing the exclude list surface any bugs?

Yes, plenty! Mostly in the data plugins. They are under tested and are some of the oldest code in `aiida-core`.

ltalirz:
>  However, we also want to run with a very specific version of pylint because changes in that version can cause differences between local and CI environment and therefore annoying CI failures. 

Perhaps I'm wrong, but I believe the version (and executable) of pylint that you are using is the one installed by the `pre-commit` extra of `aiida-core`, not the version installed by the pylint repo.
At least that's how I understand what is written here: https://pre-commit.com/#system

Otherwise, how are they mixing the two python environments?

>  However, this would cause problems when developing multiple packages in the same virtual environment. Often there would be version conflicts

This point I fully support, i.e. I'm all for moving linters, formatters etc. into separate environments if we can.

chrisjsewell:
> not the version installed by the pylint repo

yeh pretty sure thats the case

> This point I fully support, i.e. I'm all for moving linters, formatters etc. into separate environments if we can.

Just in case you didn't know, you can get pre-commit to install additional packages into its environment with: https://pre-commit.com/#config-additional_dependencies. But thats probably not an adequate cure for the `import-error` warnings

.pre-commit-config.yaml
```diff
@@ -8,105 +8,24 @@ repos:
...
-                tests/transports/test_local.py|
-                tests/transports/test_ssh.py|
-                tests/test_dataclasses.py|
             )$
+        args: ['-i']
 
--   repo: https://github.com/pre-commit/mirrors-yapf
-    rev: v0.30.0
+-   repo: https://github.com/PyCQA/pylint
+    rev: pylint-2.5.2
```
chrisjsewell, line 24: were we going to move this into the local repo, as suggested by @ltalirz ?

sphuber: But then we are back to the situation where we have to pin an exact version which will collide with any other projects you might have in the same virtual env. Is there no way to address this? I am not sure if this current solution actually allowed for this or if it had the same limitation and I just never noticed the collision

ltalirz: I think we still are in this situation - the only difference being that now we create an additional virtual environment that is never used

chrisjsewell: but as we discussed, I'm under the impression that, because of `language: system`, it is already using the virtual env version of the package **not** the one that is installed in the pre-commit env

chrisjsewell: to use the one in the pre-commit env, you would have to either ignore errors regarding dependency installs, or also install them dependencies into the pre-commit environment, using https://pre-commit.com/#config-additional_dependencies

ltalirz: @chrisjsewell not sure whether you replied to my comment or seb's - anyhow I did mean we create an additional *pre-commit* environment that is never used, i.e. we are saying the same thing.

As for your following comment, I don't think either of those are viable solutions, i.e. we can just move the hook back under `local`

sphuber: Moved it back to `local`

chrisjsewell: > not sure whether you replied to my comment or seb's

seb's

> I don't think either of those are viable solutions

yep agree, just outlining the possible option 👍

.pre-commit-config.yaml
```diff
@@ -120,6 +32,13 @@ repos:
...
 
 -   repo: local
+
+    hooks:
+    -   id: pylint
+        name: pylint
+        language: system
+        exclude: *exclude_files
+
     hooks:
```
ltalirz, line 42: @sphuber I didn't review this PR but I think you might have just disabled pylint there...

ltalirz: The configured hook doesn't actually work - you'll need to add
```
entry: pylint
```

sphuber: Thanks for the heads-up @ltalirz ! That is indeed a required key but normally that should raise a an error when the file is parsed
```
An error has occurred: InvalidConfigError: 
==> File .pre-commit-config.yaml
==> At Config()
==> At key: repos
==> At Repository(repo='local')
==> At key: hooks
==> At Hook(id='pylint')
=====> Missing required key: entry
```
The bigger problem is that I accidentally defined the `hooks` key twice. The first one, containing the `pylint` config simply gets overridden, which is why `pylint` is never actually run. I will create a PR to fix it.

sphuber: See PR #4258

ltalirz: > The bigger problem is that I accidentally defined the hooks key twice.

Right. That's what I wrote above, no? ;-)

sphuber: You did, but not in the original post. I read it on my email, which only ever send the first version. So you edit didn't show up Guess I could have noticed when responding here though :)

PR #4198, opened 2020-06-24
Update dependency requirement `kiwipy~=0.5.5`

sphuber:
Note that `kiwipy==0.5.4` will break existing installations.


csadorf:
> Note that `kiwipy==0.5.4` will break existing installations.

What exactly do you mean by that?

sphuber:
> > Note that `kiwipy==0.5.4` will break existing installations.
> 
> What exactly do you mean by that?

That release contained a change that when installed will prevent the installation from working. That is why I released `0.5.5` straight after to revert the change. Installing `0.5.4` will not pose a threat to existing data, it just that nothing can be run.

Issue #4104, opened 2020-05-25
Move doc builds back to GitHub Actions (when artefacts supported)


chrisjsewell:
In #4099, the docs action will be replaced by a build with CircleCI.
When/if GitHub Actions allows for storage of (unzipped) artefacts, we can move back to this action

chrisjsewell:
Note this essentially waiting on https://github.com/actions/upload-artifact/issues/14

chrisjsewell:
Note there is another option that I've just been made aware of, to use RTD as the CI runner: https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html

Possible cons are that I'm not sure about build limits, and also you would need to make RTD fail on errors, as I discuss here:
https://github.com/executablebooks/MyST-Parser/issues/177

csadorf:
> Note there is another option that I've just been made aware of, to use RTD directly as the CI runner: https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html

That's interesting and might be worthwhile to test.

> Possible cons are that I'm not sure about build limits, and also you would need to make RTD fail on errors, as I discuss here:
> [executablebooks/MyST-Parser#177](https://github.com/executablebooks/MyST-Parser/issues/177)

That is an issue that maybe could be fixed on the side of RTD? Surely this would be a problem for many users of that service.

chrisjsewell:
Well I'm implementing in https://github.com/aiidateam/aiida-tutorials/pull/228, so we can use that as a testbed 😄

chrisjsewell:
> maybe could be fixed on RTD side?


https://github.com/readthedocs/readthedocs.org/issues/7240

PR #4201, opened 2020-06-26
Add reentry scan to sphinx build

### LINKED ISSUE #4200: ReadTheDocs autodoc requires `reentry scan`
Currently RTD has a bunch of autodoc errors, which cause the affected classes to be missing from the API documentation:
https://readthedocs.org/projects/aiida-core/builds/11321660/

Hopefully running `reentry scan` (somewhere?) should fix this

chrisjsewell:
closes #4200

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -27,6 +27,10 @@ sys.path.append(os.path.join(os.path.split(__file__)[0], os.pardir, os.pardir))
 import aiida
 from aiida.manage.configuration import load_documentation_profile
 
+# a reentry scan is required in RTD, to ensure all the API can be imported
+import reentry
+reentry.manager.scan()
+
 # Load the dummy profile even if we are running locally, this way the documentation will succeed even if the current
 # default profile of the AiiDA installation does not use a Django backend.
 load_documentation_profile()


docs/source/conf.py
```diff
@@ -27,6 +27,10 @@
 import aiida
 from aiida.manage.configuration import load_documentation_profile
 
+# a reentry scan is required in RTD, to ensure all the API can be imported
+import reentry
```
sphuber, line 31: Since this is only needed on RTD (correct me if I'm wrong) should we couch it in the `if ON_RTD` conditional like the other RTD specific code? The `ON_RTD` is defined on line 38

chrisjsewell: Well actually there is no RTD specific code now after `docs_revamp` 😉 
But anyway, my thinking was that if we only run it on RTD, then it won't get tested in CI

chrisjsewell: BTW on a related note: https://github.com/aiidateam/aiida-core/issues/4104#issuecomment-650200606

sphuber: Ah, I didn't realize that got removed in the docs-revamp. Well in that case, we can also remove that environment variable declaration in the conf. But no need to do that here.

PR #4203, opened 2020-06-26
Docs: update ReadTheDocs configuration file

### LINKED ISSUE #4200: ReadTheDocs autodoc requires `reentry scan`

sphuber:
Fixes #4200 

This forces RTD to install the `aiida-core` package such that the entry
points are installed and it will now fail the build if warnings are
emitted.

Also remove the changes of the previous commit to manually call reentry
scan and the `ON_RTD` variable, which was a remnant of the old
documentation build.

Finally, delete folder `aiida/sphinxext/tests` which became obsolete
when all tests were moved to the `tests` folder.


docs/source/conf.py
```diff
@@ -27,16 +27,10 @@
 import aiida
```
chrisjsewell, line 27: I believe you could also remove the addition of aiida to the path, above this, now if you want?

greschd:
I think we can also delete the `requirements_on_rtd.txt` file.

chrisjsewell:
a quick thing while we are at it @sphuber : `` :meth:`~aiida.engine.processes.process_spec.ProcessSpec.define` `` in https://github.com/aiidateam/aiida-tutorials/blob/2020-intro-week/docs/pages/2020_Intro_Week/sections/workflows_adv.rst is currently failing, because we don't include `inherited-members` in the api-doc settings. So either we find a way here of doing that, or remove the reference?

sphuber:
I will get back to these additional things tomorrow. Might as well clean everything up in one go.

sphuber:
> a quick thing while we are at it @sphuber : `` :meth:`~aiida.engine.processes.process_spec.ProcessSpec.define` `` in https://github.com/aiidateam/aiida-tutorials/blob/2020-intro-week/docs/pages/2020_Intro_Week/sections/workflows_adv.rst is currently failing, because we don't include `inherited-members` in the api-doc settings:
> 
> https://github.com/aiidateam/aiida-core/blob/bb2098e7035eda1526960897dec02fbd0a1a6a14/docs/source/conf.py#L325
> 
> So either we find a way here of doing that, or remove/change the reference?

Well for a kick-off, it is incorrect :sweat_smile: It should be ``:meth:`~aiida.engine.processes.process.Process.define` ``.  So maybe we should fix that.

But I think I have seen this problem before, where inherited member methods are not properly found. Are you suggesting that by changing some setting, we can enable this? Because if that is the case, that would be great

chrisjsewell:
>  So maybe we should fix that

Ah yeh perfect I'll change that, it's not an inherited member so won't require a fix here.

> Are you suggesting that by changing some setting, we can enable this?

Yes, for every object (class/method/function/...) `autodoc` creates, it adds a reference to `docs/build/html/objects.inv`, then intersphinx downloads/reads this file to find references. You can look whats in this file using https://sphobjinv.readthedocs.io. 
The methods then have to be actually in the documentation; intersphinx can't infer inheritance. As mentioned above you could get apidoc to add the `:inherited-members:` to every module, but for aiida-core thats going to be heck of a lot of methods; so I don't know if thats a good idea, in terms of build times, search indexing, etc

chrisjsewell:
I'm guessing in the same sentence, `` :meth:`~aiida.engine.processes.process_spec.ProcessSpec.exit_code` ``, also need to be changed to `` :meth:`~aiida.engine.processes.process.Process.exit_code` ``?

sphuber:
> I'm guessing in the same sentence, `` :meth:`~aiida.engine.processes.process_spec.ProcessSpec.exit_code` ``, also need to be changed to `` :meth:`~aiida.engine.processes.process.Process.exit_code` ``?

Nope, that one is correct as it is a method of the spec and not the process itself

docs/source/conf.py
sphuber: Done

sphuber:
> As mentioned above you could get `apidoc` to add the `:inherited-members:` to every `automodule` directive it creates, but for aiida-core thats going to be heck of a lot of methods; so I don't know if thats a good idea, in terms of build times, search indexing, etc

So I guess we won't do this here and now. In that case, I think I have addressed all comments and this can be reviewed

PR #4206, opened 2020-06-29
do not install tests/ and util/ packages

ltalirz:
The previous generic "find_packages()" call in setup.py  would also
catch the tests/ and utils/ subfolders, installing them into the user's
python environment.  This not only affected installs from github, but
also source releases on PyPI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ if __name__ == '__main__':
     EXTRAS_REQUIRE['all'] = list({item for sublist in EXTRAS_REQUIRE.values() for item in sublist if item != 'bpython'})
 
     setup(
-        packages=find_packages(),
+        packages=find_packages(include=['aiida', 'aiida.*']),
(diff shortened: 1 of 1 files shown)


ltalirz:
@sphuber This should go into the 1.3.0 release - let me know whether this means I should make the PR against a different branch

ltalirz:
This is something to keep track of for a next plugin developer meeting - I believe many plugins (certainly the ones templated via the cookiecutter) are making this mistake.

greschd:
Heh, see #4204. AFAICT the change is exactly the same, so doesn't matter which one we merge.

greschd:
Note that this is a problem _only_ if the other packages contain an ``__init__.py`` - so some plugin packages may not be affected (but it's still better to limit explicitly, of course).

ltalirz:
hehe ;-) Sorry - I just checked on the issue tracker, not in the PRs.

Issue #4123, opened 2020-05-27
Docs: How to configure remote computers


ramirezfranciscof:
"Configuring remote computers" should be a subsection of "How to run external codes".

ltalirz:
We just had a question on the AiiDA mailing list on how to configure a code on a remote machine, and it appears there is not a single place in the documentation where we explain how to set up a remote computer.

This needs to be expedited.

ltalirz:
I've assigned myself for the moment, since I am anyhow revising the "How to run external codes" section, but since this howto already has substantial length, I'm actually not convinced we should explain in this section how to set up remote computers and codes.

How about creating a "How to run on remote computers" (we can always link there from "how to run external codes").
@csadorf @sphuber @ramirezfranciscof 

Happy if anybody would be willing to help with this

csadorf:
> Since this is such a major use case, how about creating a "How to run on remote computers" (we can always link there from "how to run external codes")?

I see some risk for duplication (which we can tolerate to some extent as long as there is a different focus), but I'm generally in favor of that. What sub-sections would you envision for it to contain?

- Configuring a remote computer
- Configuring a remote code
- Configuring a transport *?*
- *What else?*

sphuber:
Just wanted to note that we should be a bit careful with the word "remote". Configuring a code on your localhost is just as valid a use-case and actually shares the vast majority of steps. Just the `verdi computer configure` step is different. Whatever we write, I think it is important to note that from AiiDA's perspective, the localhost, is also "just" a computer.

> Configuring a transport ?

What do you mean with this? Using existing transport plugins is trivial as it is just one field to fill in during `verdi computer setup`. Or did you mean "how to create a custom transport plugin"?

sphuber:
> Good point, but I think the idea is to keep the top-level section as goal- and problem-oriented as possible so from a user perspective it might make sense.

That's fair enough, but we should also not start to fill in use-cases. We might think that the majority want to run on a remote computer, but maybe a lot of people just want to run on their own machine where AiiDA is running. Removing the "remote" from the titles caters to both and doesn't make it less clear I feel

ltalirz:
> Removing the "remote" from the titles caters to both and doesn't make it less clear I feel

The reason I put it there is that the tutorial already explains how to set up the localhost computer.
However, I think it's fine to duplicate a bit of information here (which will anyhow be the case).

ltalirz:
[...]

PR #4208, opened 2020-06-30
docs: revise "how to run external codes"

ltalirz:
Give a pass to this relatively new section.

### DIFF
--- a/docs/source/howto/plugin_codes.rst
+++ b/docs/source/howto/plugin_codes.rst
@@ -71,7 +71,11 @@ The first line of the method calls the |define| method of the |CalcJob| parent c
 This necessary step defines the `inputs` and `outputs` that are common to all |CalcJob|'s.
 
 Next, we use the :py:meth:`~plumpy.process_spec.ProcessSpec.input` method in order to define our two input numbers ``x`` and ``y`` (we support integers and floating point numbers), and we use :py:meth:`~plumpy.process_spec.ProcessSpec.output` to define the only output of the calculation with the label ``sum``.
-Once a calculation finishes, any output node specified here will be linked to the calculation node with the specified link label.
+AiiDA will attach the outputs defined here to a (successfully) finished calculation using the link label provided.
+
+.. note::
+    This holds for *required* outputs (the default behaviour).
+    Use ``required=False`` in order to mark an output as optional.
 
 .. tip::
 
@@ -108,7 +112,7 @@ For example:
 
 .. note:: Unlike the |define| method, the ``prepare_for_submission`` method is implemented from scratch and so there is no super call.
 
-The first step is writing the simple bash script mentioned in the beginning: summing the numbers ``x`` and ``y``, using Python's string interpolation to replace the ``x`` and ``y`` placeholders with the actual values ``self.inputs.x`` and ``self.inputs.y`` that were passed by the user.
+The first step is writing the simple bash script mentioned in the beginning: summing the numbers ``x`` and ``y``, using Python's string interpolation to replace the ``x`` and ``y`` placeholders with the actual values ``self.inputs.x`` and ``self.inputs.y`` that were provided as inputs by the caller.
 
 All inputs provided to the calculation are validated against the ``spec`` *before* |prepare_for_submission| is called.
 Therefore, when accessing the :py:attr:`~plumpy.processes.Process.inputs` attribute, you can safely assume that all required inputs have been set and that all inputs have a valid type.
@@ -176,7 +180,8 @@ Before the ``parse()`` method is called, two important attributes are set on the
 
   2. ``self.node``: The :py:class:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode` representing the finished calculation, which, among other things, provides access to all of its inputs (``self.node.inputs``).
 
-We start by opening the output file, whose filename we get from the ``self.node`` attribute via the :py:meth:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode.get_option` convenience method.
+The :py:meth:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode.get_option` convenience method is used to get the filename of the output file.
+Its content is cast to an integer, since the output file should contain the sum produced by the ``aiida.in`` bash script.
 
 We read the content of the file and cast it to an integer, which should contain the sum that was produced by the ``bash`` code.
 
--- a/docs/source/howto/plugin_codes.rst
+++ b/docs/source/howto/plugin_codes.rst
@@ -11,8 +11,8 @@ How to write a plugin for an external code
 
 To run an external code with AiiDA, you need a corresponding *calculation* plugin, which tells AiiDA how to:
 
-1. Prepare the required input files
-2. Run the code with the correct command line parameters
+1. Prepare the required input files.
+2. Run the code with the correct command line parameters.
 
 Finally, you will probably want a *parser* plugin, which tells AiiDA how to:
 
@@ -45,7 +45,7 @@ thus writing the sum of the two numbers ``numx`` and ``numy`` (provided by the u
 Interfacing external codes
 ==========================
 
-We start our ``calcjob.py`` script by subclassing the |CalcJob| class:
+Start your ``calcjob.py`` script by subclassing the |CalcJob| class:
 
 .. literalinclude:: ../../../aiida/calculations/arithmetic/add.py
     :language: python
@@ -351,7 +351,6 @@ You can do so by specifying to use a ``dry_run``, which will create all the inpu
     run(calculation_builder)
 
 
-This marks the end of this how-to, and you should now be fully equipped to wrap your own code in an AiiDA calculation plugin.
 For further reading, you may want to consult the :ref:`guidelines on plugin design <topics:plugins:guidelines>` or the how to on :ref:`packaging plugins <how-to:plugins>`.
 
 
--- a/docs/source/howto/plugin_codes.rst
+++ b/docs/source/howto/plugin_codes.rst
@@ -70,8 +70,8 @@ For example:
 The first line of the method calls the |define| method of the |CalcJob| parent class.
 This necessary step defines the `inputs` and `outputs` that are common to all |CalcJob|'s.
 
-Next, we use the :py:meth:`~plumpy.process_spec.ProcessSpec.input` method in order to define our two input numbers ``x`` and ``y`` (we support integers and floating point numbers), and we use :py:meth:`~plumpy.process_spec.ProcessSpec.output` to define the ``sum`` as an output of the calculation.
-Once a calculation finishes, a |FolderData| node will be attached as an output to the calculation node with the label ``retrieved``.
+Next, we use the :py:meth:`~plumpy.process_spec.ProcessSpec.input` method in order to define our two input numbers ``x`` and ``y`` (we support integers and floating point numbers), and we use :py:meth:`~plumpy.process_spec.ProcessSpec.output` to define an output of the calculation with label ``'sum'``.
+Once a calculation finishes, any output node specified here will be linked to the calculation node with the specified link label.
 
 .. tip::
 
--- a/docs/source/howto/plugin_codes.rst
+++ b/docs/source/howto/plugin_codes.rst
@@ -45,7 +45,7 @@ thus writing the sum of the two numbers ``numx`` and ``numy`` (provided by the u
 Interfacing external codes
 ==========================
 
-Start your ``calcjob.py`` script by subclassing the |CalcJob| class:
+Start by creating a file ``calcjob.py`` and subclass the |CalcJob| class:
 
 .. literalinclude:: ../../../aiida/calculations/arithmetic/add.py
     :language: python
@@ -70,7 +70,7 @@ For example:
 The first line of the method calls the |define| method of the |CalcJob| parent class.
 This necessary step defines the `inputs` and `outputs` that are common to all |CalcJob|'s.
 
-Next, we use the :py:meth:`~plumpy.process_spec.ProcessSpec.input` method in order to define our two input numbers ``x`` and ``y`` (we support integers and floating point numbers), and we use :py:meth:`~plumpy.process_spec.ProcessSpec.output` to define an output of the calculation with label ``'sum'``.
+Next, we use the :py:meth:`~plumpy.process_spec.ProcessSpec.input` method in order to define our two input numbers ``x`` and ``y`` (we support integers and floating point numbers), and we use :py:meth:`~plumpy.process_spec.ProcessSpec.output` to define the only output of the calculation with the label ``sum``.
 Once a calculation finishes, any output node specified here will be linked to the calculation node with the specified link label.
 
 .. tip::
@@ -85,9 +85,10 @@ These ``options`` have already been defined on the |spec| by the ``super().defin
     One more important input required by any |CalcJob| is which external executable to use.
     External executables are represented by |Code|  instances that contain information about the computer they reside on, their path in the file system and more.
 
-    They are passed to a |CalcJob| via the ``code`` input (defined in the |CalcJob| base class, so you don't have to)::
+    They are passed to a |CalcJob| via the ``code`` input, which is defined in the |CalcJob| base class, so you don't have to:
 
-        spec.input('code', valid_type=orm.Code, help='The `Code` to use for this job.')
+        .. code-block:: python
+            spec.input('code', valid_type=orm.Code, help='The `Code` to use for this job.')
 
 
 
@@ -98,8 +99,7 @@ Preparing for submission
 ------------------------
 
 The :py:meth:`~aiida.engine.processes.calcjobs.calcjob.CalcJob.prepare_for_submission` method has two jobs:
-Firstly, it creates the input files in the format the external code expects.
-And secondly, it returns a :py:class:`~aiida.common.datastructures.CalcInfo` object that contains instructions for the AiiDA engine on how the code should be run.
+Creating the input files in the format the external code expects and returning a :py:class:`~aiida.common.datastructures.CalcInfo` object that contains instructions for the AiiDA engine on how the code should be run.
 For example:
 
 .. literalinclude:: ../../../aiida/calculations/arithmetic/add.py
@@ -111,7 +111,7 @@ For example:
 The first step is writing the simple bash script mentioned in the beginning: summing the numbers ``x`` and ``y``, using Python's string interpolation to replace the ``x`` and ``y`` placeholders with the actual values ``self.inputs.x`` and ``self.inputs.y`` that were passed by the user.
 
 All inputs provided to the calculation are validated against the ``spec`` *before* |prepare_for_submission| is called.
-When accessing the :py:attr:`~plumpy.processes.Process.inputs` attribute, you can therefore safely assume that all required inputs have been set and that all inputs have a valid type.
+Therefore, when accessing the :py:attr:`~plumpy.processes.Process.inputs` attribute, you can safely assume that all required inputs have been set and that all inputs have a valid type.
 
 The ``folder`` argument (a |Folder| instance) allows us to write the input file to a sandbox folder, whose contents will be transferred to the compute resource where the actual calculation takes place.
 In this example, we only create a single input file, but you can create as many as you need, including subfolders if required.
@@ -146,7 +146,7 @@ We want to pass our input file to the executable via standard input, and record 
 
 .. tip::
 
-    ``self.inputs.input_filename`` is just a shorthand for ``self.inputs.metadata['options']['input_filename']``.
+    ``self.options.input_filename`` is just a shorthand for ``self.inputs.metadata['options']['input_filename']``.
 
 Finally, we pass the |CodeInfo| to a |CalcInfo| object (one calculation job can involve more than one executable, so ``codes_info`` is a list).
 We define the ``retrieve_list`` of filenames that the engine should retrieve from the directory where the job ran after it has finished.
@@ -161,9 +161,10 @@ For more details on the |CalcJob| class, refer to the Topics section on :ref:`de
 Parsing the outputs
 ===================
 
-Parsing the output files produced by your code into AiiDA nodes is optional, but it can make your data queryable and therefore easier to access and analyze.
+Parsing the output files produced by a code into AiiDA nodes is optional, but it can make your data queryable and therefore easier to access and analyze.
 
-We start our ``parser.py`` plugin by subclassing the |Parser| class and implementing its :py:meth:`~aiida.parsers.parser.Parser.parse` method.
+To create a parser plugin, subclass the |Parser| class (for example in a file called ``parser.py``) and implement its :py:meth:`~aiida.parsers.parser.Parser.parse` method.
+The following is an example of a simple implementation:
 
 .. literalinclude:: ../../../aiida/parsers/plugins/arithmetic/add.py
     :language: python
@@ -171,21 +172,21 @@ We start our ``parser.py`` plugin by subclassing the |Parser| class and implemen
 
 Before the ``parse()`` method is called, two important attributes are set on the |Parser|  instance:
 
-  1. ``self.retrieved``: An instance of |FolderData|, which points to the folder containing all output files produced by the code and provides the means to :py:meth:`~aiida.orm.nodes.node.Node.open` any file it contains.
+  1. ``self.retrieved``: An instance of |FolderData|, which points to the folder containing all output files that the |CalcJob| instructed to retrieve, and provides the means to :py:meth:`~aiida.orm.nodes.node.Node.open` any file it contains.
 
-  2. ``self.node``: The :py:class:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode` representing the finished calculation, which e.g. provides access to all of its inputs (``self.node.inputs``).
+  2. ``self.node``: The :py:class:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode` representing the finished calculation, which, among other things, provides access to all of its inputs (``self.node.inputs``).
 
 We start by opening the output file, whose filename we get from the ``self.node`` attribute via the :py:meth:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode.get_option` convenience method.
 
 We read the content of the file and cast it to an integer, which should contain the sum that was produced by the ``bash`` code.
 
-Finally, we use the :py:meth:`~aiida.parsers.parser.Parser.out` method to link the parsed sum as an output of the calculation.
+Finally, the :py:meth:`~aiida.parsers.parser.Parser.out` method is used to link the parsed sum as an output of the calculation.
 The first argument is the name of the output, which will be used as the label for the link that connects the calculation and data node, and the second is the node that should be recorded as an output.
 Note that the type of the output should match the type that is specified by the process specification of the corresponding |CalcJob|.
 If any of the registered outputs do not match the specification, the calculation will be marked as failed.
 
 In order to request automatic parsing of a |CalcJob| (once it has finished), users can set the ``metadata.options.parser_name`` input when launching the job.
-If a particular parser should *always* be used by default, the |CalcJob| ``define`` method can set a default value for the parser name as we did in the :ref:`previous section <how-to:plugin-codes:interfacing>`:
+If a particular parser should be used by default, the |CalcJob| ``define`` method can set a default value for the parser name as was done in the :ref:`previous section <how-to:plugin-codes:interfacing>`:
 
 .. code-block:: python
 
@@ -194,8 +195,8 @@ If a particular parser should *always* be used by default, the |CalcJob| ``defin
         ...
         spec.inputs['metadata']['options']['parser_name'].default = 'arithmetic.add'
 
-Note, that we are not passing the |Parser| class itself, but an *entry point string* under which the parser class is registered.
-How to register your parser class as an entry point is explained in the how-to section on :ref:`registering plugins <how-to:plugins>`.
+Note, that the default is not set to the |Parser| class itself, but the *entry point string* under which the parser class is registered.
+How to register a parser class through an entry point is explained in the how-to section on :ref:`registering plugins <how-to:plugins>`.
 
 
 .. _how-to:plugin-codes:parsing:errors:
@@ -205,10 +206,10 @@ Handling parsing errors
 
 So far, we have not spent much attention on dealing with potential errors that can arise when running external codes.
 However, there are lots of ways in which codes can fail to execute nominally.
-A |Parser| can play an important role in detecting such errors and reporting them to AiiDA, where :ref:`workflows <how-to:workflows>` can then decide how to proceed, e.g. by modifying input parameters and resubmitting the calculation.
+A |Parser| can play an important role in detecting and communicating such errors, where :ref:`workflows <how-to:workflows>` can then decide how to proceed, e.g., by modifying input parameters and resubmitting the calculation.
 
 Parsers communicate errors through :ref:`exit codes<topics:processes:concepts:exit_codes>`, which are defined in the |spec| of the |CalcJob| they parse.
-Our :py:class:`~aiida.calculations.arithmetic.add.ArithmeticAddCalculation` example, defines the following exit codes:
+The :py:class:`~aiida.calculations.arithmetic.add.ArithmeticAddCalculation` example, defines the following exit codes:
 
 .. literalinclude:: ../../../aiida/calculations/arithmetic/add.py
     :language: python
@@ -216,14 +217,14 @@ Our :py:class:`~aiida.calculations.arithmetic.add.ArithmeticAddCalculation` exam
     :end-before: end exit codes
     :dedent: 8
 
-Each ``exit_code`` defines
+Each ``exit_code`` defines:
 
  * an exit status (a positive integer),
  * a label that can be used to reference the code in the |parse| method (through the ``self.exit_codes`` property, as shown below), and
  * a message that provides a more detailed description of the problem.
 
 In order to inform AiiDA about a failed calculation, simply return from the ``parse`` method the exit code that corresponds to the detected issue.
-Here is a more complete version of our |Parser|:
+Here is a more complete version of the example |Parser| presented in the previous section:
 
 .. literalinclude:: ../../../aiida/parsers/plugins/arithmetic/add.py
     :language: python
@@ -236,7 +237,7 @@ It checks:
  3. Whether the output file contains an integer.
  4. Whether the sum is negative.
 
-AiiDA stored the exit code returned by the |parse| method on the calculation node that is being parsed, from where it can then be inspected further down the line.
+AiiDA stores the exit code returned by the |parse| method on the calculation node that is being parsed, from where it can then be inspected further down the line.
 The Topics section on :ref:`defining processes <topics:processes:usage:defining>` provides more details on exit codes.
 
 
@@ -250,8 +251,8 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
 
 .. _how-to:plugin-codes:run:
 
-Running your calculation
-========================
+Running a calculation
+=====================
 
 With your ``calcjob.py`` and ``parser.py`` files at hand, you can launch your first calculation:
 
@@ -283,7 +284,7 @@ With your ``calcjob.py`` and ``parser.py`` files at hand, you can launch your fi
             'x': Int(4),
             'y': Int(5),
             'metadata': {
-                'options': { 'resources': resources},
+                'options': {'resources': resources},
                 'description': "My first calculation.",
             },
         }
@@ -294,7 +295,7 @@ With your ``calcjob.py`` and ``parser.py`` files at hand, you can launch your fi
 
         # Parsing its results
         output_dict, node = ArithmeticAddParser.parse_from_node(node)
-        print("Parsing completed. Result: {}".format(int(output_dict['sum'])))
+        print("Parsing completed. Result: {}".format(output_dict['sum'].value))
 
     .. note::
 
@@ -306,7 +307,7 @@ With your ``calcjob.py`` and ``parser.py`` files at hand, you can launch your fi
 
         $ verdi run launch.py
 
-    If everything goes well, this should print the results of your calculations, something like:
+    If everything goes well, this should print the results of your calculation, something like:
 
     .. code-block:: console
 
@@ -319,13 +320,13 @@ With your ``calcjob.py`` and ``parser.py`` files at hand, you can launch your fi
     If you encountered a parsing error, it can be helpful to make a :ref:`topics:calculations:usage:calcjobs:dry_run`, which allows you to inspect the input folder generated by AiiDA before any calculation is launched.
 
 
-So far, we have deliberately avoided to touch the concept of :ref:`entry points <how-to:plugins:entrypoints>`, which are the preferred method of announcing new calculations, parsers and other plugins to AiiDA.
+So far, we have deliberately avoided to touch the concept of :ref:`entry points <how-to:plugins:entrypoints>`, which are the preferred method of registering new calculation, parser and other plugins with AiiDA.
 Using entry points simplifies your life in a number of ways:
 
  * You can use ``verdi plugin list`` to inspect which plugins are available as well as their internal documentation.
  * You can specify a default |Parser| for a |CalcJob|, which will run automatically.
  * You can specify a default |CalcJob| for a |Code|, making it easy to set the inputs for a code.
- * You can submit calcualtions to the AiiDA daemon without having to worry about how to make sure that your calculation and parser plugins can be imported from the daemon's python environment.
+ * You can submit calculations to the AiiDA daemon without having to worry about how to make sure that your calculation and parser plugins can be imported from the daemon's Python environment.
 
 Registering entry points
 ------------------------
@@ -394,7 +395,7 @@ Launching your calculation now becomes easier:
 
         # Running the calculation & parsing results
         output_dict, node = engine.run_get_node(builder)
-        print("Parsing completed. Result: {}".format(int(output_dict['sum'])))
+        print("Parsing completed. Result: {}".format(output_dict['sum'].value))
 
     .. note::
 
@@ -410,11 +411,11 @@ Launching your calculation now becomes easier:
 
 Finally instead of running your calculation in the current shell, you can submit your calculation to the AiiDA daemon:
 
- * (Re)start the daemon to update its python environment:
+ * (Re)start the daemon to update its Python environment:
 
    .. code-block:: console
 
-      $ verdi daemon restart
+      $ verdi daemon restart --reset
 
  * Update your launch script to use:
 
@@ -443,9 +444,9 @@ You can use the verdi command line interface to :ref:`monitor<topics:processes:u
     $ verdi process list
 
 
-This marks the end of this howto.
+This marks the end of this how-to.
 
-The |CalcJob|  and |Parser| plugins are still rather basic and the ``aiida-add`` plugin package is missing a number of useful gimmicks (package metadata, documentation, tests, CI, ...).
+The |CalcJob| and |Parser| plugins are still rather basic and the ``aiida-add`` plugin package is missing a number of useful features, such as package metadata, documentation, tests, CI, etc.
 Continue with :ref:`how-to:plugins` in order to learn how to quickly create a feature-rich new plugin package from scratch.
 
 .. todo::
--- a/docs/source/howto/plugin_codes.rst
+++ b/docs/source/howto/plugin_codes.rst
@@ -45,11 +45,19 @@ thus writing the sum of the two numbers ``numx`` and ``numy`` (provided by the u
 Interfacing external codes
 ==========================
 
-Start by creating a file ``calcjob.py`` and subclass the |CalcJob| class:
+Start by creating a file ``calculations.py`` and subclass the |CalcJob| class:
 
-.. literalinclude:: ../../../aiida/calculations/arithmetic/add.py
-    :language: python
-    :lines: 10-18
+.. code-block:: python
+
+    from aiida import orm
+    from aiida.common.datastructures import CalcInfo, CodeInfo
+    from aiida.common.folders import Folder
+    from aiida.engine import CalcJob, CalcJobProcessSpec
+
+
+    class ArithmeticAddCalculation(CalcJob):
+        """`CalcJob` implementation to add two numbers using bash for testing and demonstration purposes."""
+
 
 In the following, we will tell AiiDA how to run our code by implementing two key methods:
 
@@ -91,8 +99,9 @@ These ``options`` have already been defined on the |spec| by the ``super().defin
 
     They are passed to a |CalcJob| via the ``code`` input, which is defined in the |CalcJob| base class, so you don't have to:
 
-        .. code-block:: python
-            spec.input('code', valid_type=orm.Code, help='The `Code` to use for this job.')
+    .. code-block:: python
+
+        spec.input('code', valid_type=orm.Code, help='The `Code` to use for this job.')
 
 
 
@@ -167,7 +176,7 @@ Parsing the outputs
 
 Parsing the output files produced by a code into AiiDA nodes is optional, but it can make your data queryable and therefore easier to access and analyze.
 
-To create a parser plugin, subclass the |Parser| class (for example in a file called ``parser.py``) and implement its :py:meth:`~aiida.parsers.parser.Parser.parse` method.
+To create a parser plugin, subclass the |Parser| class (for example in a file called ``parsers.py``) and implement its :py:meth:`~aiida.parsers.parser.Parser.parse` method.
 The following is an example of a simple implementation:
 
 .. literalinclude:: ../../../aiida/parsers/plugins/arithmetic/add.py
@@ -182,8 +191,6 @@ Before the ``parse()`` method is called, two important attributes are set on the
 
 The :py:meth:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode.get_option` convenience method is used to get the filename of the output file.
 Its content is cast to an integer, since the output file should contain the sum produced by the ``aiida.in`` bash script.
-
-We read the content of the file and cast it to an integer, which should contain the sum that was produced by the ``bash`` code.
 
 Finally, the :py:meth:`~aiida.parsers.parser.Parser.out` method is used to link the parsed sum as an output of the calculation.
 The first argument is the name of the output, which will be used as the label for the link that connects the calculation and data node, and the second is the node that should be recorded as an output.
@@ -254,91 +261,25 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
 
     `#4123`_
 
-.. _how-to:plugin-codes:run:
-
-Running a calculation
-=====================
-
-With your ``calcjob.py`` and ``parser.py`` files at hand, you can launch your first calculation:
-
-
- 1. If you haven't already done so, set up your localhost computer:
-
-    .. code-block:: console
-
-        $ verdi computer setup -L localhost -H localhost -T local -S direct -w `echo $PWD/work` -n
-        $ verdi computer configure local tutor --safe-interval 5 -n
-
- 2. Write a first ``launch.py`` script:
-
-    .. code-block:: python
-
-        from aiida import orm, engine
-        from calcjob import ArithmeticAddCalculation
-        from parser import ArithmeticAddParser
-
-        # Setting up inputs
-        computer = orm.load_computer('localhost')
-        code = orm.Code(remote_computer_exec=[computer, '/bin/bash'])
-        resources = {
-            'num_machines': 1,
-            'num_mpiprocs_per_machine': 1,
-        }
-        inputs = {
-            'code': code,
-            'x': Int(4),
-            'y': Int(5),
-            'metadata': {
-                'options': {'resources': resources},
-                'description': "My first calculation.",
-            },
-        }
-
-        # Running the calculation
-        _ , node = engine.run_get_node( ArithmeticAddCalculation, **inputs )
-        print("Calculation completed: {}".format(node))
-
-        # Parsing its results
-        output_dict, node = ArithmeticAddParser.parse_from_node(node)
-        print("Parsing completed. Result: {}".format(output_dict['sum'].value))
-
-    .. note::
-
-        If you already have your inputs in the database, load them using ``orm.load_node()`` or find them using the :ref:`QueryBuilder <how-to:data:find>`.
-
- 3. Launch the calculation:
-
-    .. code-block:: console
-
-        $ verdi run launch.py
-
-    If everything goes well, this should print the results of your calculation, something like:
-
-    .. code-block:: console
-
-        $ verdi run launch.py
-        Calculation completed: uuid: 607d50ba-5396-411c-8bac-563b71dbaff4 (pk: 229) (calcjob.ArithmeticAddCalculation)
-        Parsing completed. Result: 9
-
-.. tip::
-
-    If you encountered a parsing error, it can be helpful to make a :ref:`topics:calculations:usage:calcjobs:dry_run`, which allows you to inspect the input folder generated by AiiDA before any calculation is launched.
-
-
-So far, we have deliberately avoided to touch the concept of :ref:`entry points <how-to:plugins:entrypoints>`, which are the preferred method of registering new calculation, parser and other plugins with AiiDA.
-Using entry points simplifies your life in a number of ways:
-
- * You can use ``verdi plugin list`` to inspect which plugins are available as well as their internal documentation.
- * You can specify a default |Parser| for a |CalcJob|, which will run automatically.
- * You can specify a default |CalcJob| for a |Code|, making it easy to set the inputs for a code.
- * You can submit calculations to the AiiDA daemon without having to worry about how to make sure that your calculation and parser plugins can be imported from the daemon's Python environment.
+.. _how-to:plugin-codes:entry-points:
 
 Registering entry points
-------------------------
+========================
 
-Registering entry points is simple:
+:ref:`Entry points <how-to:plugins:entrypoints>` are the preferred method of registering new calculation, parser and other plugins with AiiDA.
 
- * Write a minimalistic ``setup.py`` script:
+With your ``calculations.py`` and ``parsers.py`` files at hand, let's register entry points for the plugins they contain:
+
+ * Move your two scripts into a subfolder ``aiida_add``:
+
+   .. code-block:: console
+
+      mkdir aiida_add
+      mv calculations.py parsers.py aiida_add/
+
+   You have just created an ``aiida_add`` Python *package*!
+
+ * Write a minimalistic ``setup.py`` script for your new package:
 
     .. code-block:: python
 
@@ -353,12 +294,10 @@ Registering entry points is simple:
             }
         )
 
- * Move your two scripts into a subfolder ``aiida_add``:
+    .. note::
+        Strictly speaking, ``aiida-add`` is the name of the *distribution*, while ``aiida_add`` is the name of the *package*.
+        The aiida-core documentation uses the term *package* a bit more loosely.
 
-   .. code-block:: console
-
-      mkdir aiida_add
-      mv calcjob.py parser.py aiida_add/
 
  * Install your new ``aiida-add`` plugin package:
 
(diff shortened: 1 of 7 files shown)


ltalirz:
@sphuber I've given a first pass to the section. 

There are still a couple of open questions from my side:


A) This howto is rather long. I can see why keeping the CalcJob and the Parser howto in the same place makes sense, but I would suggest splitting off the section on actually running a code (which could then even include the computer setup if we want).
The reasoning is that many people will actually want to run code using existing plugins. Having the section on running codes buried at the bottom of this howto (and specific to the example above) is therefore not ideal.
I.e., I propose two howtos:

   1. How to run external codes: only contains what is needed to actually run them if a plugin already exists, including a word on installing plugins, computer setup, code setup and code submission
   2. How to write a plugin for an external code (or similar): Contains CalcJob and Parser plugin

Do you agree?

B) Since parsers need to be specified via entry points, at the CalcJob/Parser howto, people will still need to learn how to register entry points in order to get any usable result out of it.
It would be great if we could provide a fix for this - either by providing a setup.py template (or other hack) to register the entry point, or perhaps even add support for passing parser classes directly.

C) This is a minor comment - the Parser currently checks for an `OSError` thrown by `read()`, but if the file does not exist, `open()` should actually already through a `FileNotFoundError` before:
```
OSError: if the file does not exist, it will fail even earlier:

>>> h= open('abc','r')
  File "<stdin>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: 'abc'
```
Do you agree that I fix this?

ltalirz:
@csadorf Since Seb has already too many things on his hands a the moment, could you please have a look at this - if not the PR itself, at least points A) and B) in my comment above?

Also pinging @chrisjsewell for feedback

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,200 @@
...
+     In that case, use the |CodeInfo| ``cmdline_params`` attribute:
+
+     .. code-block:: python
+
+         codeinfo.cmdline_params = [ '--input', self.inputs.input_filename ]
+
+.. tip::
+
+   The dot-notation ``self.options.input_filename`` and the key-notation ``self.options['input_filename']`` are equivalent for *accessing* the value of an option, but the key-notation provides more control when *setting* values (e.g. overriding the default value while preserving the ``valid_type``).
+   That is why we used the key-notation in the ``define`` method.
```
sphuber, line 150: I might have misunderstood what you are trying to say here, but the second part is incorrect. Note that the return type of `self.options` is an attribute dictionary of actual option values passed to the process and is completely different from ``spec.inputs['metadata']['options']`` which is a `PortNamespace`. The reason that we use key-notation in define is simply because the `PortNamespace` is not an attribute dict, unlike the return value of `self.options`. Note that there is *nothing to set* also in `prepare_for_submission`.

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,200 @@
...
+
+Finally, we pass the |CodeInfo| to a |CalcInfo| object (one CalcJob can involve more than one executable, so ``codes_info`` is a list).
```
sphuber, line 152:
```suggestion
Finally, we pass the |CodeInfo| to a |CalcInfo| object (one calculation job can involve more than one executable, so ``codes_info`` is a list).
```

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,200 @@
...
+        """Parser for an `ArithmeticAddCalculation` job."""
+
+        def parse(self, **kwargs):
+            """Parse the contents of the output files stored in the `retrieved` output node."""
+            output_folder = self.retrieved
+
+            with output_folder.open(self.node.get_option('output_filename'), 'r') as handle:
+                result = int(handle.read())
+
+            self.out('sum', Int(result))
```
sphuber, line 181: I see the point that you want to first show the minimal version without exit codes, but the downside of this method is that now the code is not tested. If we change anything to the code, this might easily get broken without us noticing. We should consider if this is worth the risk and if it is really that bad to not have the full version of the code. I think people are capable of overlooking the exit codes until we get to them a bit further on

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,200 @@
...
+Finally, we use the :py:meth:`~aiida.parsers.parser.Parser.out` method to link the parsed sum as an output of the calculation.
 The first argument is the name of the output, which will be used as the label for the link that connects the calculation and data node, and the second is the node that should be recorded as an output.
 Note that the type of the output should match the type that is specified by the process specification of the corresponding |CalcJob|.
 If any of the registered outputs do not match the specification, the calculation will be marked as failed.
 
-To trigger the parsing using a |Parser| after a |CalcJob| has finished (such as the one described in the :ref:`previous section <how-to:codes:interfacing>`) it should be defined in the ``metadata.options.parser_name`` input.
-If a particular parser should always be used by default for a given |CalcJob|, it can be defined as the default in the |define| method, for example:
+In order to request automatic parsing of a |CalcJob| (once it has finished), users can set the ``metadata.options.parser_name`` input when submitting the job.
```
sphuber, line 198:
```suggestion
In order to request automatic parsing of a |CalcJob| (once it has finished), users can set the ``metadata.options.parser_name`` input when launching the job.
```

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,200 @@
...
+If a particular parser should *always* be used by default, the |CalcJob| ``define`` method can set a default value for the parser name as we did above:
```
sphuber, line 199:
```suggestion
If a particular parser should *always* be used by default, the |CalcJob| ``define`` method can set a default value for the parser name as we did in the :ref:`previous section <how-to:codes:interfacing>`:
```

docs/source/howto/codes.rst
```diff
@@ -178,40 +205,52 @@ If a particular parser should always be used by default for a given |CalcJob|, i
...
 
 Handling parsing errors
 -----------------------
 
-So far we have not spent too much attention on dealing with potential errors that might arise when running external codes.
-However, for many codes, there are lots of ways in which it can fail to execute nominally and produced the correct output.
-A |Parser| is the solution to detect these errors and report them to the caller through :ref:`exit codes<topics:processes:concepts:exit_codes>`.
-These exit codes can be defined through the |spec| of the |CalcJob| that is used for that code, just as the inputs and output are defined.
-For example, the :py:class:`~aiida.calculations.arithmetic.add.ArithmeticAddCalculation` introduced in :ref:`"Interfacing external codes"<how-to:codes:interfacing>`, defines the following exit codes:
+So far, our we have not spent much attention on dealing with potential errors that can arise when running external codes.
```
sphuber, line 217:
```suggestion
So far, we have not spent much attention on dealing with potential errors that can arise when running external codes.
```

docs/source/howto/codes.rst
```diff
@@ -178,40 +205,52 @@ If a particular parser should always be used by default for a given |CalcJob|, i
...
+However, there are lots of ways in which codes can fail to execute nominally.
+A |Parser| can play an important role in detecting such errors and reporting them to AiiDA, where :ref:`workflows <how-to:workflows>` can then decide how to proceed further, e.g. by modifying input parameters and resubmitting the calculation.
```
sphuber, line 219:
```suggestion
A |Parser| can play an important role in detecting such errors and reporting them to AiiDA, where :ref:`workflows <how-to:workflows>` can then decide how to proceed, e.g. by modifying input parameters and resubmitting the calculation.
```

docs/source/howto/codes.rst
```diff
@@ -178,40 +205,52 @@ If a particular parser should always be used by default for a given |CalcJob|, i
...
-Each ``exit_code`` defines an exit status (a positive integer), a label that can be used to reference the code in the |parse| method (through the ``self.exit_codes`` property, as seen below), and a message that provides a more detailed description of the problem.
-To use these in the |parse| method, you just need to return the corresponding exit code which instructs the engine to store it on the node of the calculation that is being parsed.
-The snippet of the previous section on :ref:`parsing the outputs<how-to:codes:parsing>` already showed two problems that are detected and are communicated by returning the corresponding the exit code:
+Each ``exit_code`` defines
+
+ * an exit status (a positive integer),
+ * a label that can be used to reference the code in the |parse| method (through the ``self.exit_codes`` property, as shown below), and
+ * a message that provides a more detailed description of the problem.
+
+In order to inform AiiDA about a failed calculation, simply return the corresponding exit codes in the ``parse`` method, when you detect the corresponding issue.
```
sphuber, line 236:
```suggestion
In order to inform AiiDA about a failed calculation, simply return from the ``parse`` method the exit code that corresponds to the detected issue.
```

docs/source/howto/codes.rst
```diff
@@ -178,40 +205,52 @@ If a particular parser should always be used by default for a given |CalcJob|, i
...
     :language: python
-    :lines: 28-34
-    :dedent: 8
+    :pyobject: ArithmeticAddParser
+
+It checks:
+
+ 1. Whether a retrieved folder is present.
+ 2. Whether the output file can be read (whether ``read()`` will throw an ``OSError``).
+
```
sphuber, line 247: empty line?

docs/source/howto/codes.rst
```diff
@@ -313,6 +352,11 @@ You can do so by specifying to use a ``dry_run``, which will create all the inpu
     calculation_builder.metadata.store_provenance = False
     run(calculation_builder)
 
+
+This marks the end of this how-to, and you should now be fully equipped to wrap your own code in an AiiDA plugin.
```
sphuber, line 356:
```suggestion
This marks the end of this how-to, and you should now be fully equipped to wrap your own code in an AiiDA calculation plugin.
```

docs/source/howto/codes.rst
```diff
@@ -313,6 +352,11 @@ You can do so by specifying to use a ``dry_run``, which will create all the inpu
...
+For further reading, you may want to consult the :ref:`guidelines on plugin design <topics:plugins:guidelines>` or the how to on `packaging plugins <how-to:plugins>`.
```
sphuber, line 357:
```suggestion
For further reading, you may want to consult the :ref:`guidelines on plugin design <topics:plugins:guidelines>` or the how to on :ref:`packaging plugins <how-to:plugins>`.
```

ltalirz:
> I definitely see the point though and would actually also prefer with "Running" first (including computer and code setup) and "Interfacing" second. I am not sure if we should split this in two different how-tos.

Ok, glad you agree. Whether or not to split into two is probably less important than the order, but running a code using an existing plugin and writing a calcjob/parser for one are really quite different purposes, so I think it makes sense to have a separate howto for each (apart from the fact that, with adding info on installing plugins/setting up computers + codes, the howto would get very long).

Thanks also for the suggestion to "parse manually", this looks like a good workaround. 
I'll look into it later today.

As for the `OSError` - I wasn't aware of it being a subclass.
Your solution seems perfect, I will just adapt the textual description to include the fact it will also catch errors in `open`.

docs/source/topics/plugins.rst
```diff
@@ -34,6 +34,30 @@ If you find yourself in a situation where you feel like you need to do any of th
 .. _core: https://github.com/aiidateam/aiida-core
 .. _registry: https://github.com/aiidateam/aiida-registry
 
+.. _topics:plugins:guidelines:
+
+Guidelines for plugin design
```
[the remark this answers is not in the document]
ltalirz: Thanks, I see your point - I checked the place you suggested, and I see the relation, but I feel it would be a bit lost in this very detailed explanation of the API. It would be good to have a place with simple considerations for plugin design - not just parsers and calcjobs, but eventually more.

I suggest to keep it in the plugins topic for the moment, but with this hierarchy

```
Guidelines for plugin design
============================

CalcJob & Parser plugins
--------------------------
```
Sounds good?

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,200 @@
...
+
 Interfacing external codes
 ==========================
 
-To provide AiiDA with the set of instructions, required to run a code, one should subclass the |CalcJob| class and implement the following two key methods:
+We start our ``calcjob.py`` script by subclassing the |CalcJob| class:
+
+.. literalinclude:: ../../../aiida/calculations/arithmetic/add.py
+    :language: python
+    :lines: 10-18
```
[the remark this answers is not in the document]
ltalirz: About the line numbers, I can also just copy-paste the code - just let me know.

However, I do disagree about the imports.
The purpose of this section is to get people started with their first calcjob and parser plugin.
In this context, figuring out predictable import errors is not time well spent, and we should save people this time when we can.

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,200 @@
...
-For the code that adds up two numbers, we will need to define those numbers as inputs (let's call them ``x`` and ``y`` to label them) and the result as an output (``sum``).
-The snippet below shows one potential implementation, as it is included in ``aiida-core``:
+The |define| method tells AiiDA which inputs the |CalcJob| expects and which outputs it produces (exit codes will be :ref:`discussed later<how-to:codes:parsing:errors>`).
+This is done through an instance of the :py:class:`~aiida.engine.processes.process_spec.CalcJobProcessSpec` class, which is passed as the |spec| argument to the |define| method.
+For example:
 
 .. literalinclude:: ../../../aiida/calculations/arithmetic/add.py
     :language: python
     :pyobject: ArithmeticAddCalculation.define
-    :dedent: 4
```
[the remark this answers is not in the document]
ltalirz: Because people will (and should be able to) copy-paste it into their code.
Let's not make life harder unnecessarily

docs/source/howto/codes.rst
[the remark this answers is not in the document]
sphuber: Copy-pasting code has the same problem in that it won't be tested and so can break. I do see your point about the imports, but I would then prefer to move them in the methods where they are needed. This way we can use the current direct includes that are fully tested and they include all imports. For the first subclassing, we can then literally say: "Create a sublcass of `CalcJob` which you can import from `aiida.engine`."

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,200 @@
...
 
-The first line of the |define| implementation calls the method of the parent class |CalcJob|.
-This step is crucial as it will define `inputs` and `outputs` that are common to all |CalcJob|'s and failing to do so will leave the implementation broken.
-After the super call, we modify the default values for some of these inputs that are defined by the base class.
-Inputs that have already been defined can be accessed from the |spec| through the :py:attr:`~plumpy.process_spec.ProcessSpec.inputs` attribute, which behaves like a normal dictionary.
+The first line of the method calls the |define| method of the |CalcJob| parent class.
+This necessary step defines the `inputs` and `outputs` that are common to all |CalcJob|'s.
+
+Next, we use the :py:meth:`~plumpy.process_spec.ProcessSpec.input` method in order to define our two input numbers ``x`` and ``y`` (we support integers and floating point numbers), and we use :py:meth:`~plumpy.process_spec.ProcessSpec.output` to define the ``sum`` as an output of the calculation.
+Once a calculation finishes, any output node specified here will be attached to the calculation node with the label ``retrieved``.
```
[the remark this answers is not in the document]
ltalirz: Thanks for catching this one (it was a bit too late when I did this revision ;-) ).
Will fix.

docs/source/howto/codes.rst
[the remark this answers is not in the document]
sphuber: Well it makes it harder to read because it is unnecessarily indented. I would argue that is more of a burden than indenting the code once after you have copy pasted the snippet. Anyway, you will have to adapt the code and indenting all code once is a single click in any half-decent editor

docs/source/howto/codes.rst
ltalirz: Right, I now checked the old version of this and it just mentioned a short-hand notation; not the detail about ports etc. (which indeed should be explained in the `define` section, if any).
I will fix this.

docs/source/howto/codes.rst
ltalirz: The alternative is that I also add this simple parser to the codebase and test it.

docs/source/howto/codes.rst
ltalirz: fixed!

docs/source/howto/codes.rst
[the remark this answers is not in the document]
ltalirz: From my previous tutorial experience, the issue is that not everybody is actually using a decent editor - my main goal here is that I would like to receive zero questions about indentation of this code snippet during the tutorial.

However, you are of course right that indentation ends up hiding a few more characters.
Let's let a 3rd person decide @csadorf What do you say?

Just for reference, here is

NO INDENTATION
image

INDENTATION
image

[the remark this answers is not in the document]
csadorf: I vote against dedenting, because it is more correct and if folks have issues copy & pasting dedented code they will also have issues with figuring out the right detention. But might be worthwhile to record what caused more issues during the actual tutorial.

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,200 @@
...
-An example implementation, as shipped with ``aiida-core`` can be seen in the following snippet:
+The :py:meth:`~aiida.engine.processes.calcjobs.calcjob.CalcJob.prepare_for_submission` method has two jobs:
+Firstly, it creates the input files in the format the external code expects.
+And secondly, it returns a :py:class:`~aiida.common.datastructures.CalcInfo` object that contains instructions for the AiiDA engine on how the code should be run.
+For example:
 
 .. literalinclude:: ../../../aiida/calculations/arithmetic/add.py
     :language: python
     :pyobject: ArithmeticAddCalculation.prepare_for_submission
-    :dedent: 4
```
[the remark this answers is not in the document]
csadorf: see my comment above

csadorf:
> @sphuber I've given a first pass to the section.
> 
> There are still a couple of open questions from my side:
> 
> A) This howto is rather long. I can see why keeping the CalcJob and the Parser howto in the same place makes sense, but I would suggest splitting off the section on actually running a code (which could then even include the computer setup if we want).
> The reasoning is that many people will actually want to run code using existing plugins. Having the section on running codes buried at the bottom of this howto (and specific to the example above) is therefore not ideal.
> I.e., I propose two howtos:
> 
>     1. How to run external codes: only contains what is needed to actually run them if a plugin already exists, including a word on installing plugins, computer setup, code setup and code submission (also addressing #4123)
> 
>     2. How to write a plugin for an external code (or similar): Contains CalcJob and Parser plugin
> 
> 
> Do you agree?
> 

Yes.

> B) Since parsers need to be specified via entry points, at the end of the CalcJob/Parser howto, people still need to learn how to register entry points in order to get any usable result out of it.
> It would be great if we could provide a fix for this - either by providing a setup.py template (or other hack) to register the entry point, or perhaps even add support for passing parser classes directly into CalcJobs.
> 

I do not know enough about / have sufficient experience with the entry point system to judge this.

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,199 @@
...
 
-The following subsections will not only take you through the process of :ref:`creating the calculation plugin<how-to:codes:interfacing>` and then using these to actually :ref:`run the code<how-to:codes:run>`.
-It will also show examples on how to implement tools that are commonly coupled with the running of a calculation, such as :ref:`the parsing of outputs<how-to:codes:parsing>`.
+1. Prepare the required input files
+2. Run the code with the correct command line parameters
 
-.. todo::
+Finally, you will probably want a *parser* plugin, which tells AiiDA how to:
+
+3. Parse the output of the code.
```
csadorf, line 18: Inconsistent use of the period in the listing: either everywhere or nowhere

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,199 @@
...
-   | Make a list of which information to:
-
-     #. parse into the database for querying (:py:class:`~aiida.orm.nodes.data.dict.Dict`, ...)
-     #. store in the file repository for safe-keeping (:py:class:`~aiida.orm.nodes.data.singlefile.SinglefileData`, ...)
-     #. leave on the computer where the calculation ran (:py:class:`~aiida.orm.nodes.data.remote.RemoteData`, ...)
-
-To demonstrate how to create a plugin for an external code, we will use the trivial example of using the `bash` shell (``/bin/bash``) to sum two numbers by running the command: ``echo $(( numx + numy ))``.
-Here, the `bash` binary will be effectively acting as our |Code| executable, the input (``aiida.in``) will then be a file containing the command with the numbers provided by the user replaced, and the output (``aiida.out``) will be caught through the standard output.
-The final recipe to run this code will then be:
+This how-to takes you through the process of :ref:`creating a calculation plugin<how-to:codes:interfacing>` for a dummy executable that sums two numbers, using it to :ref:`run the code<how-to:codes:run>`, and :ref:`writing a parser <how-to:codes:parsing>` for its outputs.
```
csadorf, line 21:
```suggestion
This how-to takes you through the process of :ref:`creating a calculation plugin<how-to:codes:interfacing>` for a very simple executable that sums two numbers, using it to :ref:`run the code<how-to:codes:run>`, and :ref:`writing a parser <how-to:codes:parsing>` for its outputs.
```
It is not actually a dummy executable since it performs the desired calculation.

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,199 @@
...
-Inputs that have already been defined can be accessed from the |spec| through the :py:attr:`~plumpy.process_spec.ProcessSpec.inputs` attribute, which behaves like a normal dictionary.
+The first line of the method calls the |define| method of the |CalcJob| parent class.
+This necessary step defines the `inputs` and `outputs` that are common to all |CalcJob|'s.
+
+Next, we use the :py:meth:`~plumpy.process_spec.ProcessSpec.input` method in order to define our two input numbers ``x`` and ``y`` (we support integers and floating point numbers), and we use :py:meth:`~plumpy.process_spec.ProcessSpec.output` to define an output of the calculation with label ``'sum'``.
+Once a calculation finishes, any output node specified here will be linked to the calculation node with the specified link label.
+
+.. tip::
+
+    In real-world plugins, popular types of input nodes are Python dictionaries (:py:class:`~aiida.orm.nodes.data.dict.Dict`) and files (:py:class:`~aiida.orm.nodes.data.singlefile.SinglefileData`).
```
csadorf, line 78: I'm not quite sure what this tip is trying to tell me. Should I prefer those types?

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,199 @@
...
 
-After modifying the existing inputs, we define the inputs that are specific to this code.
-For this purpose we use the :py:meth:`~plumpy.process_spec.ProcessSpec.input` method, which does not modify the existing `inputs`, accessed through :py:attr:`~plumpy.process_spec.ProcessSpec.inputs`, but defines new ones that will be specific to this implementation.
-You can also see that the definitions do not involve the assignment of a value, but only the passing of parameters to the method: a label to identify it, their valid types (in this case nodes of type |Int|) and a description.
-Finally, note that there is no return statement: this method does not need to return anything, since all modifications are made directly into the received |spec| object.
-You can check the Topics section about :ref:`defining processes <topics:processes:usage:defining>` if you want more information about setting up your `inputs` and `outputs` (covering validation, dynamic number of inputs, etc.).
+
+
+Note that there is no ``return`` statement in ``define``: the ``define`` method directly modifies the |spec| object it receives.
```
csadorf, line 94:
```suggestion
There is no ``return`` statement in ``define``: the ``define`` method directly modifies the |spec| object it receives.
```

docs/source/howto/codes.rst
```diff
@@ -3,173 +3,199 @@
...
-    :dedent: 4
 
-Note that, unlike the |define| method, this one is implemented from scratch and so there is no super call.
-The external code that we are running with this |CalcJob| is ``bash`` and so to sum the input numbers ``x`` and ``y``, we should write a bash input file that performs the summation, for example ``echo $((x + y))``, where one of course has to replace ``x`` and ``y`` with the actual numbers.
-You can see how the snippet uses the ``folder`` argument, which is a |Folder| instance that represents a temporary folder on disk, to write the input file with the bash summation.
-It uses Python's string interpolation to replace the ``x`` and ``y`` placeholders with the actual values that were passed as input, ``self.inputs.x`` and ``self.inputs.y``, respectively.
+.. note:: Unlike the |define| method, the ``prepare_for_submission`` method is implemented from scratch and so there is no super call.
 
-.. note::
+We start by writing our simple bash script that sums the two numbers ``x`` and ``y``, using Python's string interpolation to replace the ``x`` and ``y`` placeholders with the actual values ``self.inputs.x`` and ``self.inputs.y`` that were passed by the user.
```
csadorf, line 111: I would prefer to avoid "tutorial-style" language in the how-tos. Any instructions that explicitly prompt the reader to follow along ("We start by...") should be reserved for the tutorial.

docs/source/howto/codes.rst
```diff
@@ -313,6 +350,11 @@ You can do so by specifying to use a ``dry_run``, which will create all the inpu
     calculation_builder.metadata.store_provenance = False
     run(calculation_builder)
 
+
+This marks the end of this how-to, and you should now be fully equipped to wrap your own code in an AiiDA calculation plugin.
```
csadorf, line 354:
```suggestion
```
Self-evident.

docs/source/howto/codes.rst
ltalirz: People go to this howto with the purpose of wrapping their own code, for which they will likely need Dict or SinglefileData input nodes.
This is just to give them an idea, happy to reformulate to make this more clear

docs/source/howto/codes.rst
ltalirz: So, replace by "Start by"?

ltalirz: Anyhow, to some degree I think we do want the reader to follow along here, i.e. to write their first calcjob and parser using the example we provide.
I.e. this howto certainly does have some tutorial character.
However, we currently don't have a "tutorial" category in the docs, just an "introduction" category that contains the basic tutorial. 
I think we should probably leave this under "howto"

docs/source/howto/codes.rst
csadorf: I think a good formulation would be along the lines of "Consider to use x y, and z as input parameter types, that are used in many real world ..."

docs/source/howto/codes.rst
csadorf: This is from the wiki about the how-to sections:
> The how-to guides should provide concrete steps, but do not have to be as detailed as a tutorial and further are not required to demonstrate a specific recipe from start to finish, meaning that there is some expectation that users are generally familiar with AiiDA.

So, yes, they can have some tutorial-character, but I would prefer to not give the impression that users are *expected* to follow along as they read them. That's obviously a very nuanced distinction, so happy to discuss, and we can even consider to change our guidelines regarding this if there is consensus.

docs/source/howto/codes.rst
ltalirz: I'm adding the code into the codebase + adding a test

ltalirz:
@sphuber I'm now done with the update of this section - I believe it is now self-contained, with a useful result at the end

 * add SimpleArithmeticAddParser to codebase (+ add one test)
 * add complete example of how to run the created calculation and parser
plugins
 * add complete example of how to register entry points for these

Please have a look!

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
 Preparing for submission
 ------------------------
 
-The :py:meth:`~aiida.engine.processes.calcjobs.calcjob.CalcJob.prepare_for_submission` method is used for two purposes.
-Firstly, it should create the input files, based on the input nodes passed to the calculation, in the format that the external code will expect.
-Secondly, the method should create and return a :py:class:`~aiida.common.datastructures.CalcInfo` instance that contains various instructions for the engine on how the code should be run.
-An example implementation, as shipped with ``aiida-core`` can be seen in the following snippet:
+The :py:meth:`~aiida.engine.processes.calcjobs.calcjob.CalcJob.prepare_for_submission` method has two jobs:
+Firstly, it creates the input files in the format the external code expects.
+And secondly, it returns a :py:class:`~aiida.common.datastructures.CalcInfo` object that contains instructions for the AiiDA engine on how the code should be run.
```
sphuber, line 102:
```suggestion
Creating the input files in the format the external code expects and returning a :py:class:`~aiida.common.datastructures.CalcInfo` object that contains instructions for the AiiDA engine on how the code should be run.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
-The external code that we are running with this |CalcJob| is ``bash`` and so to sum the input numbers ``x`` and ``y``, we should write a bash input file that performs the summation, for example ``echo $((x + y))``, where one of course has to replace ``x`` and ``y`` with the actual numbers.
-You can see how the snippet uses the ``folder`` argument, which is a |Folder| instance that represents a temporary folder on disk, to write the input file with the bash summation.
-It uses Python's string interpolation to replace the ``x`` and ``y`` placeholders with the actual values that were passed as input, ``self.inputs.x`` and ``self.inputs.y``, respectively.
+.. note:: Unlike the |define| method, the ``prepare_for_submission`` method is implemented from scratch and so there is no super call.
 
-.. note::
+The first step is writing the simple bash script mentioned in the beginning: summing the numbers ``x`` and ``y``, using Python's string interpolation to replace the ``x`` and ``y`` placeholders with the actual values ``self.inputs.x`` and ``self.inputs.y`` that were passed by the user.
```
sphuber, line 111:
```suggestion
The first step is writing the simple bash script mentioned in the beginning: summing the numbers ``x`` and ``y``, using Python's string interpolation to replace the ``x`` and ``y`` placeholders with the actual values ``self.inputs.x`` and ``self.inputs.y`` that were passed in the inputs by the caller.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
 
-    When the |prepare_for_submission| is called, the inputs that have been passed will have been validated against the specification defined in the |define| method and they can be accessed through the :py:attr:`~plumpy.processes.Process.inputs` attribute.
-    This means that if a particular input is required according to the spec, you can safely assume that it will have been set and you do not need to check explicitly for its existence.
+All inputs provided to the calculation are validated against the ``spec`` *before* |prepare_for_submission| is called.
+When accessing the :py:attr:`~plumpy.processes.Process.inputs` attribute, you can therefore safely assume that all required inputs have been set and that all inputs have a valid type.
```
sphuber, line 114:
```suggestion
Therefore, when accessing the :py:attr:`~plumpy.processes.Process.inputs` attribute, you can safely assume that all required inputs have been set and that all inputs have a valid type.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
+     Many executables don't read from standard input but instead require the path to an input file to be passed via command line parameters (potentially including further configuration options).
+     In that case, use the |CodeInfo| ``cmdline_params`` attribute:
+
+     .. code-block:: python
+
+         codeinfo.cmdline_params = ['--input', self.inputs.input_filename]
+
+.. tip::
+
+    ``self.inputs.input_filename`` is just a shorthand for ``self.inputs.metadata['options']['input_filename']``.
```
sphuber, line 149:
```suggestion
    ``self.options.input_filename`` is just a shorthand for ``self.inputs.metadata['options']['input_filename']``.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
 
 .. _how-to:plugin-codes:parsing:
 
 Parsing the outputs
 ===================
 
-The parsing of the output files generated by a |CalcJob| is optional and can be used to store (part of) their information as AiiDA nodes, which makes the data queryable and therefore easier to access and analyze.
-To enable |CalcJob| output file parsing, one should subclass the |Parser| class and implement the :py:meth:`~aiida.parsers.parser.Parser.parse` method.
-The following is an example implementation, as shipped with ``aiida-core``, to parse the outputs of the :py:class:`~aiida.calculations.arithmetic.add.ArithmeticAddCalculation` discussed in the previous section:
+Parsing the output files produced by your code into AiiDA nodes is optional, but it can make your data queryable and therefore easier to access and analyze.
```
sphuber, line 164:
```suggestion
Parsing the output files produced by a code into AiiDA nodes is optional, but it can make your data queryable and therefore easier to access and analyze.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
+We start our ``parser.py`` plugin by subclassing the |Parser| class and implementing its :py:meth:`~aiida.parsers.parser.Parser.parse` method.
 
 .. literalinclude:: ../../../aiida/parsers/plugins/arithmetic/add.py
     :language: python
-    :pyobject: ArithmeticAddParser
+    :pyobject: SimpleArithmeticAddParser
+
+Before the ``parse()`` method is called, two important attributes are set on the |Parser|  instance:
+
+  1. ``self.retrieved``: An instance of |FolderData|, which points to the folder containing all output files produced by the code and provides the means to :py:meth:`~aiida.orm.nodes.node.Node.open` any file it contains.
```
sphuber, line 174:
```suggestion
  1. ``self.retrieved``: An instance of |FolderData|, which points to the folder containing all output files that the |CalcJob| instructed to retrieve, and provides the means to :py:meth:`~aiida.orm.nodes.node.Node.open` any file it contains.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
 
 Parsing the outputs
 ===================
 
-The parsing of the output files generated by a |CalcJob| is optional and can be used to store (part of) their information as AiiDA nodes, which makes the data queryable and therefore easier to access and analyze.
-To enable |CalcJob| output file parsing, one should subclass the |Parser| class and implement the :py:meth:`~aiida.parsers.parser.Parser.parse` method.
-The following is an example implementation, as shipped with ``aiida-core``, to parse the outputs of the :py:class:`~aiida.calculations.arithmetic.add.ArithmeticAddCalculation` discussed in the previous section:
+Parsing the output files produced by your code into AiiDA nodes is optional, but it can make your data queryable and therefore easier to access and analyze.
+
+We start our ``parser.py`` plugin by subclassing the |Parser| class and implementing its :py:meth:`~aiida.parsers.parser.Parser.parse` method.
```
sphuber, line 166:
```suggestion
To create a parser plugin, subclass the |Parser| class (for example in a file called ``parser.py``) and implement its :py:meth:`~aiida.parsers.parser.Parser.parse` method.
The following is an example of a simple implementation:
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
 .. literalinclude:: ../../../aiida/parsers/plugins/arithmetic/add.py
     :language: python
-    :pyobject: ArithmeticAddParser
+    :pyobject: SimpleArithmeticAddParser
+
+Before the ``parse()`` method is called, two important attributes are set on the |Parser|  instance:
+
+  1. ``self.retrieved``: An instance of |FolderData|, which points to the folder containing all output files produced by the code and provides the means to :py:meth:`~aiida.orm.nodes.node.Node.open` any file it contains.
+
+  2. ``self.node``: The :py:class:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode` representing the finished calculation, which e.g. provides access to all of its inputs (``self.node.inputs``).
```
sphuber, line 176:
```suggestion
  2. ``self.node``: The :py:class:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode` representing the finished calculation, which, among other things, provides access to all of its inputs (``self.node.inputs``).
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
+
+We start by opening the output file, whose filename we get from the ``self.node`` attribute via the :py:meth:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode.get_option` convenience method.
```
sphuber, line 178:
```suggestion
The output of the code is obtained by reading the content of the output file, whose filename we get from the ``self.node`` attribute via the :py:meth:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode.get_option` convenience method.
The read content is cast to an integer, as it should contain the sum that was produced by the ``bash`` code.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
 
-The output files generated by the completed calculation can be accessed from the ``retrieved`` output folder, which can be accessed through the :py:attr:`~aiida.parsers.parser.Parser.retrieved` property.
-It is an instance of :py:class:`~aiida.orm.nodes.data.folder.FolderData` and so provides, among other things, the :py:meth:`~aiida.orm.nodes.node.Node.open` method to open any file it contains.
-In this example implementation, we use it to open the output file, whose filename we get through the :py:meth:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode.get_option` method of the corresponding calculation node, which we obtain through the :py:attr:`~aiida.parsers.parser.Parser.node` property of the ``Parser``.
 We read the content of the file and cast it to an integer, which should contain the sum that was produced by the ``bash`` code.
```
sphuber, line 180: delete this line

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
-We catch any exceptions that might be thrown, for example when the file cannot be read, or if its content cannot be interpreted as an integer, and return an exit code.
-This method of dealing with potential errors of external codes is discussed in the section on :ref:`handling parsing errors<how-to:plugin-codes:parsing:errors>`.
 
-To attach the parsed sum as an output, use the :py:meth:`~aiida.parsers.parser.Parser.out` method.
+Finally, we use the :py:meth:`~aiida.parsers.parser.Parser.out` method to link the parsed sum as an output of the calculation.
```
sphuber, line 182:
```suggestion
Finally, the :py:meth:`~aiida.parsers.parser.Parser.out` method is used to link the parsed sum as an output of the calculation.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
 Note that the type of the output should match the type that is specified by the process specification of the corresponding |CalcJob|.
 If any of the registered outputs do not match the specification, the calculation will be marked as failed.
 
-To trigger the parsing using a |Parser| after a |CalcJob| has finished (such as the one described in the :ref:`previous section <how-to:plugin-codes:interfacing>`) it should be defined in the ``metadata.options.parser_name`` input.
-If a particular parser should always be used by default for a given |CalcJob|, it can be defined as the default in the |define| method, for example:
+In order to request automatic parsing of a |CalcJob| (once it has finished), users can set the ``metadata.options.parser_name`` input when launching the job.
+If a particular parser should *always* be used by default, the |CalcJob| ``define`` method can set a default value for the parser name as we did in the :ref:`previous section <how-to:plugin-codes:interfacing>`:
```
sphuber, line 188:
```suggestion
If a particular parser should be used by default, the |CalcJob| ``define`` method can set a default value for the parser name as was done in the :ref:`previous section <how-to:plugin-codes:interfacing>`:
```

docs/source/howto/plugin_codes.rst
```diff
@@ -178,40 +194,51 @@ If a particular parser should always be used by default for a given |CalcJob|, i
         ...
         spec.inputs['metadata']['options']['parser_name'].default = 'arithmetic.add'
 
-The default can be overridden through the inputs when launching the calculation job.
-Note, that one should not pass the |Parser| class itself, but rather the corresponding entry point name under which it is registered as a plugin.
-In other words, in order to use a |Parser| you will need to register it as explained in the how-to section on :ref:`registering plugins <how-to:plugins>`.
+Note, that we are not passing the |Parser| class itself, but an *entry point string* under which the parser class is registered.
```
sphuber, line 197:
```suggestion
Note, that the default is not set to the |Parser| class itself, but the *entry point string* under which the parser class is registered.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -178,40 +194,51 @@ If a particular parser should always be used by default for a given |CalcJob|, i
...
+How to register your parser class as an entry point is explained in the how-to section on :ref:`registering plugins <how-to:plugins>`.
```
sphuber, line 198:
```suggestion
How to register a parser class through an entry point is explained in the how-to section on :ref:`registering plugins <how-to:plugins>`.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -178,40 +194,51 @@ If a particular parser should always be used by default for a given |CalcJob|, i
...
 
-So far we have not spent too much attention on dealing with potential errors that might arise when running external codes.
-However, for many codes, there are lots of ways in which it can fail to execute nominally and produced the correct output.
-A |Parser| is the solution to detect these errors and report them to the caller through :ref:`exit codes<topics:processes:concepts:exit_codes>`.
-These exit codes can be defined through the |spec| of the |CalcJob| that is used for that code, just as the inputs and output are defined.
-For example, the :py:class:`~aiida.calculations.arithmetic.add.ArithmeticAddCalculation` introduced in :ref:`"Interfacing external codes"<how-to:plugin-codes:interfacing>`, defines the following exit codes:
+So far, we have not spent much attention on dealing with potential errors that can arise when running external codes.
+However, there are lots of ways in which codes can fail to execute nominally.
+A |Parser| can play an important role in detecting such errors and reporting them to AiiDA, where :ref:`workflows <how-to:workflows>` can then decide how to proceed, e.g. by modifying input parameters and resubmitting the calculation.
```
sphuber, line 208:
```suggestion
A |Parser| can play an important role in detecting and communicating such errors, where :ref:`workflows <how-to:workflows>` can then decide how to proceed, e.g., by modifying input parameters and resubmitting the calculation.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -178,40 +194,51 @@ If a particular parser should always be used by default for a given |CalcJob|, i
...
+
+Parsers communicate errors through :ref:`exit codes<topics:processes:concepts:exit_codes>`, which are defined in the |spec| of the |CalcJob| they parse.
+Our :py:class:`~aiida.calculations.arithmetic.add.ArithmeticAddCalculation` example, defines the following exit codes:
```
sphuber, line 211:
```suggestion
The :py:class:`~aiida.calculations.arithmetic.add.ArithmeticAddCalculation` example, defines the following exit codes:
```

docs/source/howto/plugin_codes.rst
```diff
@@ -178,40 +194,51 @@ If a particular parser should always be used by default for a given |CalcJob|, i
...
 .. literalinclude:: ../../../aiida/calculations/arithmetic/add.py
     :language: python
     :start-after: start exit codes
     :end-before: end exit codes
     :dedent: 8
 
-Each ``exit_code`` defines an exit status (a positive integer), a label that can be used to reference the code in the |parse| method (through the ``self.exit_codes`` property, as seen below), and a message that provides a more detailed description of the problem.
-To use these in the |parse| method, you just need to return the corresponding exit code which instructs the engine to store it on the node of the calculation that is being parsed.
-The snippet of the previous section on :ref:`parsing the outputs<how-to:plugin-codes:parsing>` already showed two problems that are detected and are communicated by returning the corresponding the exit code:
+Each ``exit_code`` defines
```
sphuber, line 219:
```suggestion
Each ``exit_code`` defines:
```

docs/source/howto/plugin_codes.rst
```diff
@@ -178,40 +194,51 @@ If a particular parser should always be used by default for a given |CalcJob|, i
...
+
+ * an exit status (a positive integer),
+ * a label that can be used to reference the code in the |parse| method (through the ``self.exit_codes`` property, as shown below), and
+ * a message that provides a more detailed description of the problem.
+
+In order to inform AiiDA about a failed calculation, simply return from the ``parse`` method the exit code that corresponds to the detected issue.
+Here is a more complete version of our |Parser|:
```
sphuber, line 226:
```suggestion
Here is a more complete version of the example |Parser| presented in the previous section:
```

docs/source/howto/plugin_codes.rst
```diff
@@ -178,40 +194,51 @@ If a particular parser should always be used by default for a given |CalcJob|, i
...
+    :pyobject: ArithmeticAddParser
+
+It checks:
+
+ 1. Whether a retrieved folder is present.
+ 2. Whether the output file can be read (whether ``open()`` or ``read()`` will throw an ``OSError``).
+ 3. Whether the output file contains an integer.
+ 4. Whether the sum is negative.
+
+AiiDA stored the exit code returned by the |parse| method on the calculation node that is being parsed, from where it can then be inspected further down the line.
```
sphuber, line 239:
```suggestion
AiiDA stores the exit code returned by the |parse| method on the calculation node that is being parsed, from where it can then be inspected further down the line.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
+        resources = {
+            'num_machines': 1,
+            'num_mpiprocs_per_machine': 1,
+        }
+        inputs = {
+            'code': code,
+            'x': Int(4),
+            'y': Int(5),
+            'metadata': {
+                'options': { 'resources': resources},
```
sphuber, line 286:
```suggestion
                'options': {'resources': resources},
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
 
 .. _how-to:plugin-codes:run:
 
-Running external codes
-======================
+Running your calculation
+========================
```
sphuber, line 254:
```suggestion
Running a calculation
=====================
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
-    $ verdi plugin list aiida.calculations
+ 1. If you haven't already done so, set up your localhost computer:
 
-As an example, we will show how to use the ``arithmetic.add`` plugin, which is a pre-installed plugin that uses the `bash shell <https://www.gnu.org/software/bash/>`_ to sum two integers.
-You can access it with the ``CalculationFactory``:
+    .. code-block:: console
 
-.. code-block:: python
+        $ verdi computer setup -L localhost -H localhost -T local -S direct -w `echo $PWD/work` -n
+        $ verdi computer configure local tutor --safe-interval 5 -n
```
sphuber, line 264: have you tested these commands? I don't think `tutor` exists. Think that should be `localhost`.
Shouldn't this also just refer to a section that explains how to setup a computer?

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
-If you already have these nodes in your database, you can get them by :ref:`querying for them <how-to:data:find>` or using ``orm.load_node(<PK>)``.
-Otherwise, you will need to create them as shown below (note that you `will` need to already have the ``localhost`` computer configured, as explained in the :ref:`previous how-to<how-to:plugin-codes:computers>`):
+        # Running the calculation
+        _ , node = engine.run_get_node( ArithmeticAddCalculation, **inputs )
+        print("Calculation completed: {}".format(node))
 
-.. code-block:: python
+        # Parsing its results
+        output_dict, node = ArithmeticAddParser.parse_from_node(node)
+        print("Parsing completed. Result: {}".format(int(output_dict['sum'])))
```
sphuber, line 297:
```suggestion
        print("Parsing completed. Result: {}".format(output_dict['sum'].value))
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
-    calculation_builder.x = number_x
-    calculation_builder.y = number_y
+    .. code-block:: console
 
-Now everything is in place and ready to perform the calculation, which can be done in two different ways.
-The first one is blocking and will return a dictionary containing all the output nodes (keyed after their label, so in this case these should be: "remote_folder", "retrieved" and "sum") that you can safely inspect and work with:
+        $ verdi run launch.py
 
-.. code-block:: python
+    If everything goes well, this should print the results of your calculations, something like:
```
sphuber, line 309:
```suggestion
    If everything goes well, this should print the results of your calculation, something like:
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
 
-    from aiida.engine import submit
-    calculation = submit(calculation_builder)
+    If you encountered a parsing error, it can be helpful to make a :ref:`topics:calculations:usage:calcjobs:dry_run`, which allows you to inspect the input folder generated by AiiDA before any calculation is launched.
 
-Note that, although you have access to the node, the underlying calculation `process` is not guaranteed to have finished when you get back control in the interpreter.
-You can use the verdi command line interface to :ref:`monitor<topics:processes:usage:monitoring>` these processes:
 
-.. code-block:: bash
+So far, we have deliberately avoided to touch the concept of :ref:`entry points <how-to:plugins:entrypoints>`, which are the preferred method of announcing new calculations, parsers and other plugins to AiiDA.
```
sphuber, line 322:
```suggestion
So far, we have deliberately avoided to touch the concept of :ref:`entry points <how-to:plugins:entrypoints>`, which are the preferred method of registering new calculation, parser and other plugins with AiiDA.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
+Using entry points simplifies your life in a number of ways:
 
-    $ verdi process list
+ * You can use ``verdi plugin list`` to inspect which plugins are available as well as their internal documentation.
+ * You can specify a default |Parser| for a |CalcJob|, which will run automatically.
+ * You can specify a default |CalcJob| for a |Code|, making it easy to set the inputs for a code.
+ * You can submit calcualtions to the AiiDA daemon without having to worry about how to make sure that your calculation and parser plugins can be imported from the daemon's python environment.
```
sphuber, line 328:
```suggestion
 * You can submit calculations to the AiiDA daemon without having to worry about how to make sure that your calculation and parser plugins can be imported from the daemon's Python environment.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
+ * Simplify your ``launch.py`` script:
+
+    .. code-block:: python
+
+        from aiida import orm, engine
 
-    .. _how-to:plugin-codes:caching:
+        # Setting up inputs
+        computer = orm.load_computer('localhost')
+        code = orm.Code(remote_computer_exec=[computer, '/bin/bash'], input_plugin_name='add')
```
sphuber, line 385: This is actually adding complexity, adding the `input_plugin_name` right? And there is no benefit to it here.

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
+        builder.metadata.options.withmpi = False
+        builder.metadata.options.resources = {
+            'num_machines': 1,
+            'num_mpiprocs_per_machine': 1,
+        }
 
-    title: Using caching to save computational resources
+        # Running the calculation & parsing results
+        output_dict, node = engine.run_get_node(builder)
+        print("Parsing completed. Result: {}".format(int(output_dict['sum'])))
```
sphuber, line 397:
```suggestion
        print("Parsing completed. Result: {}".format(output_dict['sum'].value))
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
 
-    `#3988`_
+    .. note::
 
+        ``output_dict`` is a dictionary containing all the output nodes keyed after their label.
+        In this case: "remote_folder", "retrieved" and "sum".
```
sphuber, line 402: Why not put this after the first launch script?

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
+
+ * Launch the calculation:
+
+    .. code-block:: console
+
+        $ verdi run launch.py
+
+Finally instead of running your calculation in the current shell, you can submit your calculation to the AiiDA daemon:
+
+ * (Re)start the daemon to update its python environment:
```
sphuber, line 413:
```suggestion
 * (Re)start the daemon to update its Python environment:
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
+
+   .. code-block:: console
+
+      $ verdi daemon restart
```
sphuber, line 417:
```suggestion
      $ verdi daemon restart --reset
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
+
+
+You can use the verdi command line interface to :ref:`monitor<topics:processes:usage:monitoring>` this processes:
+
+.. code-block:: bash
+
+    $ verdi process list
+
+
+This marks the end of this howto.
```
sphuber, line 446:
```suggestion
This marks the end of this how-to.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -223,104 +250,205 @@ The Topics section on :ref:`defining processes <topics:processes:usage:defining>
...
+
+The |CalcJob|  and |Parser| plugins are still rather basic and the ``aiida-add`` plugin package is missing a number of useful gimmicks (package metadata, documentation, tests, CI, ...).
```
sphuber, line 448:
```suggestion
The |CalcJob| and |Parser| plugins are still rather basic and the ``aiida-add`` plugin package is missing a number of useful features, such as package metadata, documentation, tests, CI, etc.
```

ltalirz:
Thanks @sphuber for the review!

> However, I do see that you are really trying to give a very detailed complete step-by-step instructions to create these plugins from scratch, but I think there is a lot of overlap now. For example, the whole part of setting up a computer and writing the launch script could just refer to the preceding how-to's, no?

I agree that we need to find a good balance between the howtos being self-contained (and therefore faster/easier to follow, since information on related topics is already focused/condensed) and duplication of information - I'm open for suggestions on the best approach here (also from @csadorf ).

 1) setting up a computer could certainly be replaced by a reference - on the other hand, we are taking up just 2 lines of code here (still, I'm fine with removing these two)
 2) writing a launch script: this is not as easy to remove since we are writing a quite specialized launch script here.

Point 2) could, however addressed in conjunction with another question I had:
In the current state, I let them first write a launch script without entry points, which then becomes simplified when they add entry points.
It would be possible to jump directly to the "package" with entry points and skip the intermediate step.
This would be faster; on the other hand people may learn a bit less about how CalcJobs/Parsers work. I am fine with either solution, and if we decide to remove it, then the launch script will be a lot more "standard".

Let me know, guys! I'll now go through the comments one by one.

docs/source/howto/plugin_codes.rst
ltalirz: That should have been `verdi computer configure local localhost ...` 
I agree that we can refer to the computer setup section here, though.
Will change this.

docs/source/howto/plugin_codes.rst
ltalirz: > And there is no benefit to it here.

This is not true. Adding the input plugin is what allows using the builder.

docs/source/howto/plugin_codes.rst
ltalirz: There were already other comments made there, so I decided to delay it.
But you are certainly correct that it would fit there as well.

docs/source/howto/plugin_codes.rst
ltalirz:
```suggestion
The first step is writing the simple bash script mentioned in the beginning: summing the numbers ``x`` and ``y``, using Python's string interpolation to replace the ``x`` and ``y`` placeholders with the actual values ``self.inputs.x`` and ``self.inputs.y`` that were provided as inputs by the caller.
```

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
 
-The first line of the |define| implementation calls the method of the parent class |CalcJob|.
-This step is crucial as it will define `inputs` and `outputs` that are common to all |CalcJob|'s and failing to do so will leave the implementation broken.
-After the super call, we modify the default values for some of these inputs that are defined by the base class.
-Inputs that have already been defined can be accessed from the |spec| through the :py:attr:`~plumpy.process_spec.ProcessSpec.inputs` attribute, which behaves like a normal dictionary.
+The first line of the method calls the |define| method of the |CalcJob| parent class.
+This necessary step defines the `inputs` and `outputs` that are common to all |CalcJob|'s.
+
+Next, we use the :py:meth:`~plumpy.process_spec.ProcessSpec.input` method in order to define our two input numbers ``x`` and ``y`` (we support integers and floating point numbers), and we use :py:meth:`~plumpy.process_spec.ProcessSpec.output` to define an output of the calculation with label ``'sum'``.
+Once a calculation finishes, any output node specified here will be linked to the calculation node with the specified link label.
```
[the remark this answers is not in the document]
ltalirz:
```suggestion
AiiDA will attach the outputs defined here to a (successfully) finished calculation using the link label provided.

.. note::
    This holds for *required* outputs (the default).
    Use ``required=False`` in order to mark an output as optional.
```

docs/source/howto/plugin_codes.rst
ltalirz:
```suggestion
The :py:meth:`~aiida.orm.nodes.process.calculation.calcjob.CalcJobNode.get_option` convenience method is used to get the filename of the output file.
Its content is cast to an integer, since the output file should contain the sum produced by the ``aiida.in`` bash script.
```

docs/source/howto/plugin_codes.rst
ltalirz:
```suggestion
```

sphuber:
> * setting up a computer could certainly be replaced by a reference - on the other hand, we are taking up just 2 lines of code here (still, I'm fine with removing these two)

This is fair enough and I agree that removing this won't change that much

> * writing a launch script: this is not as easy to remove since we are writing a quite specialized launch script here.

What is so specialized about his launch script though?
It consists of basic AiiDA API: loading a class, defining the inputs and calling either run or submit. This is the essence of all launch scripts.

Anyway, if @csadorf agrees with the current form and thinks this will be useful, than I am happy to go ahead with this and see how it goes.

ltalirz:
> What is so specialized about his launch script though?

We are importing the parser and calcjob directly, creating the code using the python API, parsing by hand, ...
(I'm referring to the first one, not to the second one with entry points, which is indeed quite standard)

docs/source/howto/plugin_codes.rst
```diff
@@ -4,172 +4,188 @@
...
+
 Interfacing external codes
 ==========================
 
-To provide AiiDA with the set of instructions, required to run a code, one should subclass the |CalcJob| class and implement the following two key methods:
+Start your ``calcjob.py`` script by subclassing the |CalcJob| class:
+
+.. literalinclude:: ../../../aiida/calculations/arithmetic/add.py
+    :language: python
+    :lines: 10-18
```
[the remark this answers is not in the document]
ltalirz: One problem are the type annotations - the corresponding classes all need to be imported as well.
Either we remove them for this "demo plugin" or I think we should provide the corresponding imports here as well to save time.

docs/source/howto/plugin_codes.rst
ltalirz:
```suggestion
        $ verdi computer configure local localhost --safe-interval 5 -n
```

ltalirz:
Ok @sphuber , from my side this is ready to merge - feel free to have another look at the last two commits.

ltalirz:
> I think this looks good. The only thing I was wondering is whether there should be a reference to "How to write a plugin for an external code" from "How to run external codes".

sphuber:
> By the way, I don't know what the issue with the PR run was, it was complaining about pyyaml - let's hope the aiida-core build isn't broken...

Wouldn't be the first time though. But I don't think so, because I just started building one on my fork and that seems to work just fine. Maybe just try to relaunch the workflow?

ltalirz:
> Maybe just try to relaunch the workflow?

I just tried - still getting this: https://github.com/aiidateam/aiida-core/pull/4208/checks?check_run_id=848972270#step:4:239

ltalirz:
For some reason, while both `verdi` and `tests` jobs install `pyyaml 5.1.2`,  it seems in the `verdi` job it needs to uninstall pyyaml 5.3.x for this (and this somehow fails).
It remains to figure out why this pyyaml 5.3.x is there in `verdi` but not in `tests`

ltalirz:
Ok, there are a couple of possible candiates:

`verdi` is:
 * using a different python version
 * not upgrading pip
 * installing numpy separately (despite `tests` no longer doing this)

I will check whether some of this is the source of this issue

sphuber:
It is really weird, because the build on my fork works just fine: https://github.com/sphuber/aiida_core/runs/848891500 
It is uptodate with `develop` but the PR build fails.

ltalirz:
@sphuber Can you try making a PR from your fork?

ltalirz:
Ok, I see you already did and it failed on aiida-core as well. https://github.com/aiidateam/aiida-core/pull/4228
It looks like we are somehow getting a "bad worker VM" on aiida-core (?)

Can you try restarting your build on the fork?
I have to say that, for me, the tests on the fork are failing as well 
https://github.com/ltalirz/aiida-core/commits/tutorial-running-external-codes

I.e. it does look like something we will need to fix/ work around.

ltalirz:
@sphuber Could you please approve so that I can merge this once the tests pass?

ltalirz:
@sphuber It seems the test error on django with python 3.8 is somehow reproducible.
https://github.com/aiidateam/aiida-core/pull/4208/checks?check_run_id=851172837#step:10:468

I think the test is expecting to get one (AiiDAEntryPoint) warning, but for some reason it is getting more than one (9).
Do you have an idea how this PR might affect this?
I am adding a parser that is not registered through entry points... that's the only thing I can think of here, but the query that is supposed to raise the warnings is for `PotentialFailureWorkChain`).

I can have a quick stab at trying to reproduce this issue locally but if that fails I don't know what else to do.

I'm slowly getting crazy with this PR... we need to merge this for tomorrow's plugin tutorial.

sphuber:
@ltalirz I am taking a look now

ltalirz:
Just confirmed that locally the test passes for me with python 3.8.3 & django

sphuber:
I think it may just be a fluke, even though it also happened in another build. Let's just re-run the worklow

ltalirz:
Or it is again some strange issue with the runner virtual environment... if there was an issue with django in the python 3.8 environment, this could perhaps explain the error.
However, this should then also occur in any other PI that picks up the same virtual environment version.

ltalirz:
My god, the tests are passing!

(by the way, can we disable jenkins? it's just annoying at this point, or does anybody still check it?)

giovannipizzi:
Re Jenkins, the server came back to life today after more than a month. So it's probably outdated and there might be things we didn't fix in the meantime. We have already a meeting with Chris next week to fix this. In any case they are not required - I know it's not nice but for now please just ignore them, until we fix them

PR #4213, opened 2020-07-01
Add the `--paused` flag to `verdi process list`

### LINKED ISSUE #4212: Add the flag `--paused` to `verdi process list`
This should filter processes that are paused which can be useful to find those that have been paused due to the exponential backoff mechanism

sphuber:
Fixes #4212 

This flag will filter for processes that are currently paused which is
useful to find calculation jobs that may have hit the exponential
backoff mechanism, among other things.

### DIFF
--- a/aiida/cmdline/params/options/__init__.py
+++ b/aiida/cmdline/params/options/__init__.py
@@ -324,6 +324,8 @@ PROCESS_STATE = OverridableOption(
     help='Only include entries with this process state.'
 )
 
+PAUSED = OverridableOption('--paused', 'paused', is_flag=True, help='Only include entries that are paused.')
+
 PROCESS_LABEL = OverridableOption(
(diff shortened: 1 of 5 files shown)


aiida/cmdline/params/options/__init__.py
```diff
@@ -324,6 +324,8 @@ def decorator(command):
     help='Only include entries with this process state.'
 )
 
+PAUSED = OverridableOption('--paused', 'paused', is_flag=True, help='Only include entries that are paused.')
```
chrisjsewell, line 327: Running out of letters lol; I see `p` and `P` are already taken

sphuber: I figured this might not be used _that_ often so it's okay to only have the long form, which is not that long anyway

PR #4217, opened 2020-07-04
docs: use new page rank feature

ltalirz:
ReadTheDocs just introduced a way to control the search rank of
documentation pages, allowing us to push hits in the autogenerated API
docs further down in the list of search results.

See https://github.com/readthedocs/readthedocs.org/issues/7082#issuecomment-653120310

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -20,3 +20,7 @@ python:
 sphinx:
     builder: html
     fail_on_warning: true
+
+search:
+  ranking:
+    reference/apidoc/*: -5
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -23,4 +23,4 @@ sphinx:
 
 search:
   ranking:
-    reference/apidoc/*: -5
+    reference/apidoc/*: -7


ltalirz:
@chrisjsewell do you think it would make sense to add the nice "rtd build" for the aiida-core docs as well?

Obviously, the focus is less on the documentation but every now and then there are PRs like this one where it would come in handy.

chrisjsewell:
As in move from circleci to rtd PR builds?
this is an outstanding issue for that which is currently lacking: https://github.com/readthedocs/readthedocs.org/issues/7240
My other question was regarding the RTD server capacity; it allows for 2 concurrent builds, but I'm not sure if that means others during that time are just cancelled, which would be a bit of a pain

ltalirz:
@sphuber This is mainly in response to your comment [here](https://github.com/aiidateam/aiida-core/pull/4111#issuecomment-637809927) - 

The rank feature is now live on RTD 
 * original docs: https://aiida.readthedocs.io/projects/aiida-core/en/latest/search.html?q=workflows
 * new docs: https://aiida.readthedocs.io/projects/aiida-core/en/fix-search-rank/search.html?q=workflows

Try also other searches, e.g. CalcJob, configuration, daemon, ... and let me know whether to adjust the rank (e.g. we could also choose something less negative than -5).

ltalirz:
@csadorf @chrisjsewell Feel free to review as well - this is a small but potentially very useful change.
All that remains to be decided is whether we should further tune the "ranking".

sphuber:
Thanks @ltalirz already a lot better, but I still think we might want to further derank the reference. I think that if you want to use the reference, you will know how to find it. Most people using the search really are just looking for the how-to's or topics. As an example, when [searching for the querybuilder](https://aiida.readthedocs.io/projects/aiida-core/en/fix-search-rank/search.html?q=querybuilder) we still mostly get reference and the actual how-to's are the 6th entry.

chrisjsewell:
> As an example, when searching for the querybuilder 

With respect to this, I would note also that a complimentary approach is to use: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#html-metadata. Since the term `querybuilder`, as an example, doesn't show up specifically in any of the how-to headings, its difficult for RTD to determine a rank for it. But if you added:

```rst
.. meta::
   :keywords: querybuilder
```

at the start of the query builder section, then this *should* move it up the rankings

ltalirz:
Thanks for the input, guys!

> Thanks @ltalirz already a lot better, but I still think we might want to further derank the reference.

Ok, I've lowered it a bit more to -6. 

>  I think that if you want to use the reference, you will know how to find it.

ltalirz:
Ok, further reduced to -7 
Let's merge this and once more elaborate processing of meta tags becomes available we can start using that.

PR #4218, opened 2020-07-06
Gsoc: replace tornado with asyncio

unkcpz:
This pr constructed by three parts of refactoring:

1. https://github.com/aiidateam/aiida-core/commit/2844b86eea688cd4b7a9f75f6a4599cc6d8838e3 Totally drop the `tornado` and replace it with `asyncio` for modules and function which are independent of `plumpy`, such as `exponential_backoff_retry`.
2.  https://github.com/aiidateam/aiida-core/commit/5e461b71509de4ff8cb2e5f18791f6080b33a3be For `aiida.engine.runner` and process which involved with plumpy. 
3. https://github.com/aiidateam/aiida-core/commit/8c29be32a0168e6fda6311c36301c72f68b8082e fix the issues after using `LoopCommunicator` as runner's communicator for all broker communication.


unkcpz:
hi @sphuber @muhrin , after some code ironing it is now ready to be reviewed. Meanwhile, as @ltalirz suggest, I will also test the code with new version jupyter notebook.

sphuber:
Thanks a lot @unkcpz . I will try to find the time to review this soon, but I think it would be best that we finalize the outstanding PR in `plumpy` first. It seems like you could make all the tests run (aside from the pre-commit, docs and installation, but that is ok) on `aiida-core` with that version of `plumpy` so I think we should review and merge that first. I will try to do it soon.

unkcpz:
There is a critical issue prevent this version to being used in production. The issue is due to new circus(I guess) reported https://github.com/circus-tent/circus/issues/1131 . Seems that endpoint can not be found with `--waiting` option on.

PR #4220, opened 2020-07-06
add --in-place flag to verdi export migrate

### LINKED ISSUE #4233: Add an `--in-place` flag for `verdi export migrate`

### LINKED ISSUE #4322: Calling `verdi export migrate` on an up-to-date file should return exit code 0
Currently, when trying to migrate an export file that is already at the latest version, you get a critical error:
```bash
$ verdi export migrate old.aiida new.aiida
Critical: Your export file is already at the version 0.9
```

This makes it unnecessarily difficult to migrate a bunch of archives to the latest version (e.g. needed on Materials Cloud) - currently, you need to first check what version an archive has, and only if it is not already at the latest version, perform the migration.

The much simpler solution is to make the migration of an archive that is already at the latest version a no-op and return exit code 0.

ltalirz:
Fixes #4233 
Fixes #4322

When export migration is required, you often want to get rid of the outdated
export file.
The `--in-place` flag saves users from specifying a temporary filename,
removing the old file and renaming the temporary one.

This commit also changes the exit code for migrating an export file that
already is up to date from an error to success (0).


tests/cmdline/commands/test_export.py
```diff
@@ -213,6 +201,27 @@ def test_migrate_force(self):
...
                 self.assertEqual(zipfile.ZipFile(filename_output).testzip(), None)
 
+    def test_migrate_in_place(self):
+        """Test that passing the -i/--in-place option will overwrite the passed file."""
+        archive = 'export_v0.1_simple.aiida'
+        target_version = '0.2'
+        filename_input = get_archive_file(archive, filepath=self.fixture_archive)
+        filename_tmp = next(tempfile._get_candidate_names())
+
+        try:
```
sphuber, line 211: why not use:
```
with tempfile.NamedTemporaryFile() as handle:
    filename = handle.name
```
That way no need to manually remove nor use protected `_get_candidate_names`

aiida/tools/importexport/common/archive.py
```diff
@@ -147,8 +147,8 @@ def get_info(self):
         :return: a dictionary with basic details
         """
         info = {
-            'version aiida': self.version_aiida,
-            'version format': self.version_format,
+            'aiida version': self.version_aiida,
+            'format version': self.version_format,
```
sphuber, line 151: attributes are also called `version_aiida` so why not keep this the same?

tests/cmdline/commands/test_export.py
```diff
@@ -213,6 +201,27 @@ def test_migrate_force(self):
                 self.assertTrue(os.path.isfile(filename_output))
                 self.assertEqual(zipfile.ZipFile(filename_output).testzip(), None)
 
+    def test_migrate_in_place(self):
```
sphuber, line 204: Please add another test that not specifying an output file and also not the `-i` flag, exist with non-zero

tests/cmdline/commands/test_export.py
```diff
@@ -213,6 +201,27 @@ def test_migrate_force(self):
...
+
+        try:
+            # copy file (don't want to overwrite test data)
+            shutil.copy(filename_input, filename_tmp)
+
+            options = [filename_tmp, '--in-place', '--version', target_version]
+            result = self.cli_runner.invoke(cmd_export.migrate, options)
+            self.assertIsNone(result.exception)
+            self.assertTrue(os.path.isfile(filename_tmp))
+            self.assertEqual(zipfile.ZipFile(filename_tmp).testzip(), None)
```
sphuber, line 219: What is this line doing? I don't understand why you are comparing to `None`

aiida/cmdline/commands/cmd_export.py
```diff
@@ -141,19 +142,20 @@ def create(
...
+@click.option('-i', '--in-place', is_flag=True, help='Migrate archive in placePrint the data contents and exit.')
 @options.SILENT()
 @click.option(
     '-v',
     '--version',
     type=click.STRING,
     required=False,
     metavar='VERSION',
-    help='Specify an exact archive version to migrate to. By default the most recent version is taken.'
+    help='Archive format version to migrate to (default: latest version).'
```
sphuber, line 156: Maybe better to use an actual default? As in, import the latest version from `aiida.tools.importexport`, set that as the default and set `show_default=True`.

aiida/cmdline/commands/cmd_export.py
```diff
@@ -141,19 +142,20 @@ def create(
 
 @verdi_export.command('migrate')
 @arguments.INPUT_FILE()
-@arguments.OUTPUT_FILE()
+@arguments.OUTPUT_FILE(required=False)
 @options.ARCHIVE_FORMAT()
 @options.FORCE(help='overwrite output file if it already exists')
+@click.option('-i', '--in-place', is_flag=True, help='Migrate archive in placePrint the data contents and exit.')
```
sphuber, line 148:
```suggestion
@click.option('-i', '--in-place', is_flag=True, help='Migrate the archive in place, overwriting the original file.')
```

aiida/tools/importexport/common/archive.py
ltalirz: It's just not the colloquial use (you would say "AiiDA version") but if that mirrors the attributes, it's fine to keep.

tests/cmdline/commands/test_export.py
ltalirz: I saw in the [docs](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) that 
> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms

In any case, I was a bit worried about opening a handle to a file that is then copied over, read, deleted and copied over again (perhaps it would be fine as well...)

tests/cmdline/commands/test_export.py
ltalirz: Good question - I copied this from the test above.
See https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.testzip

tests/cmdline/commands/test_export.py
ltalirz: Done!

aiida/cmdline/commands/cmd_export.py
```diff
@@ -21,6 +22,7 @@
 from aiida.cmdline.utils import decorators
 from aiida.cmdline.utils import echo
 from aiida.common.links import GraphTraversalRules
+from aiida.tools.importexport import EXPORT_VERSION
```
CasperWA, line 25: I think there was a reason behind moving all `importexport` module imports into the functions. I believe it had to do with  verdi's loading time? But if this works (all tests pass) then all good for me :+1:

aiida/tools/importexport/migration/__init__.py
```diff
@@ -52,8 +52,8 @@ def migrate_recursively(metadata, data, folder, version=EXPORT_VERSION):
 
     try:
         if old_version == version:
-            raise ArchiveMigrationError('Your export file is already at the version {}'.format(version))
-        elif old_version > version:
+            return old_version
```
CasperWA, line 55: This is changing the user-expected behaviour of the migration functionality, something quite out-of-scope of this PR.

tests/cmdline/commands/test_export.py
```diff
@@ -181,18 +181,6 @@ def test_migrate_version_specific(self):
         finally:
             delete_temporary_file(filename_output)
 
-    def test_migrate_versions_recent(self):
```
CasperWA, line 184: Related to my other comment about changing the migration functionality behaviour.

tests/cmdline/commands/test_export.py
greschd: Since we're only supporting Unix-y platforms, that shouldn't be an issue though, right?

tests/cmdline/commands/test_export.py
```diff
@@ -213,6 +201,39 @@ def test_migrate_force(self):
...
+        filename_tmp = next(tempfile._get_candidate_names())  # pylint: disable=protected-access
+
+        try:
+            # copy file (don't want to overwrite test data)
+            shutil.copy(filename_input, filename_tmp)
+
+            # specifying both output and in-place should except
+            options = [filename_tmp, '--in-place', '--output-file', 'test.aiida']
+            result = self.cli_runner.invoke(cmd_export.migrate, options)
+            self.assertIsNotNone(result.exception)
```
CasperWA, line 218: It's nice to print `stdout` if this fails, and vice versa when assert something in `stdout`, it's nice to print `stderr`.
If you find it confusing like this to read, you can also pass `result.output` in by using the `msg` parameter, i.e. `assertTrue(something, msg="this is shown if assertion fails")`.
```suggestion
            self.assertIsNotNone(result.exception, result.output)
```

tests/cmdline/commands/test_export.py
```diff
@@ -213,6 +201,39 @@ def test_migrate_force(self):
...
+
+            # specifying neither output nor in-place should except
+            options = [filename_tmp]
+            result = self.cli_runner.invoke(cmd_export.migrate, options)
+            self.assertIsNotNone(result.exception)
```
CasperWA, line 223:
```suggestion
            self.assertIsNotNone(result.exception, result.output)
```

tests/cmdline/commands/test_export.py
```diff
@@ -213,6 +201,39 @@ def test_migrate_force(self):
...
+
+            # check that in-place migration produces a valid archive in place of the old file
+            options = [filename_tmp, '--in-place', '--version', target_version]
+            result = self.cli_runner.invoke(cmd_export.migrate, options)
+            self.assertIsNone(result.exception)
```
CasperWA, line 228:
```suggestion
            self.assertIsNone(result.exception, result.output)
```

tests/cmdline/commands/test_export.py
ltalirz: > Since we're only supporting Unix-y platforms, that shouldn't be an issue though, right?

So far ;-) 
Maybe it never happens, but I believe circus is our only dependency currently that is not available on windows (at least it was some time ago), and fixes for windows are underway.

Perhaps it never happens... but let's keep the option open I would say.

greschd: Not sure about the deleting and re-creating part, either.

aiida/cmdline/commands/cmd_export.py
```diff
@@ -212,5 +227,8 @@ def migrate(input_file, output_file, force, silent, archive_format, version):
             with tarfile.open(output_file, 'w:gz', format=tarfile.PAX_FORMAT, dereference=True) as archive:
                 archive.add(folder.abspath, arcname='')
 
+        if in_place:
+            os.rename(output_file, input_file)
```
CasperWA, line 231: Will all this work if the file is not in the current working directory and such? I guess it should, since `rename` works with relative paths...

CasperWA (commented):
Seems fine to me.

I have some suggested updates to parts of the tests, nothing serious.
What's more serious is that you are effectively implementing/changing two things in this PR, one is adding the `--in-place` flag option, but then you're also changing expected behaviour of the migration functionality.
The latter might be fine, but where has this change been discussed?

aiida/cmdline/commands/cmd_export.py
ltalirz: Ok, I see. Unfortunately, keeping the methods in that is not possible if we want to use the `EXPORT_VERSION` as the default value of one of the options.

However, there is an issue in the tests because it seems that importing `aiida.tools.importexport` mandates that `seekpath` is installed (which I consider a bug).
We'll need to find a solution for this.

aiida/tools/importexport/migration/__init__.py
ltalirz: While I would consider the new behavior the expected one, I agree that I am changing it here and that there was no issue opened for it where it was discussed. I will open it now.

ltalirz: Opened #4322

tests/cmdline/commands/test_export.py
greschd: Anyway, if you want to avoid manually getting the name, you could create a `TemporaryDirectory` and then just name the file whatever.

I think the `_get_candidate_names()` only generates a random string.. so there might not be any protection against filename clashes.

ltalirz: Done!

aiida/cmdline/commands/cmd_export.py
ltalirz: Fixing this in https://github.com/aiidateam/aiida-core/pull/4324

aiida/cmdline/commands/cmd_export.py
ltalirz: I think so. Click checks that the input file is a valid path; and the output file should be an absolute path in this case.

aiida/cmdline/commands/cmd_export.py
CasperWA: > Ok, I see. Unfortunately, keeping the imports inside the methods is not possible if we want to use the `EXPORT_VERSION` as the default value of one of the click options.

Agreed, but since it's also not a necessity, it can be argued that it can be reverted again to simply say "latest version"?

> However, there is an issue in the tests because it seems that importing `aiida.tools.importexport` mandates that `seekpath` is installed (which I consider a bug).
> We'll need to find a solution for this.

Yes. I am unsure where this comes from exactly. What series of imports in the `importexport` module leads to the import of `seekpath`?

ltalirz: Ok, after #4324 is merged, this now indeed triggers the load-time checker
```
Run verdi devel check-load-time
Critical: potential `verdi` speed problem: `aiida.orm` module is imported.
##[error]Process completed with exit code 1.
```
There are of course different ways of fixing this
 1. make sure that importing aiida.tools does not lead to importing aiida.orm 
 1. move the EXPORT_VERSION to a place from where it can safely be imported
 1. revert back the change to pass the EXPORT_VERSION as a default value of the option (change requested by @sphuber )

Given that tools are operating on AiiDA ORM classes and the docstring of the subpackage states that modules in this sub package may require the database environment to be loaded, I'm not sure option 1. makes much sense.

I would tend towards option 2., moving the EXPORT_VERSION to some high-level place (perhaps `aiida.common`?).
Let me know @sphuber @CasperWA

CasperWA: I agree mostly with your sentiment here, although I would strongly advice against moving `EXPORT_VERSION` outside the `importexport` module.
However, we could create another file that simply stores this (and maybe other) hard-coded variables, which can then be imported directly from said file (i.e. one cannot do `from aiida.tools.importexport import EXPORT_VERSION` without also importing `aiida.orm`, but one can maybe do ` from aiida.tools.importexport.common.config import EXPORT_VERSION` or something?)

ltalirz: > i.e. one cannot do from aiida.tools.importexport import EXPORT_VERSION without also importing aiida.orm, but one can maybe do from aiida.tools.importexport.common.config import EXPORT_VERSION or something?

Sure, that is what I am suggesting.

CasperWA: Ah, I understood that you were suggesting `aiida.common.config`, i.e., not under the `importexport` module still, but rather the top-level `common`.

CasperWA: So if I look closer, I fint that `EXPORT_VERSION` already is at `aiida.tools.importexport.common.config`. The mappings that depend on `orm` objects could then be moved from this to another file, maybe `utils` instead? That should then solve the problem (if you import `EXPORT_VERSION` from `aiida.tools.importexport.common.config`).

ltalirz: Ah, sorry, I misread your comment (was a bit late ;-) ).

I think there is no way to prevent the import of `aiida.tools` when importing a submodule of it, i.e. the version constant would need to be moved out.

One thing I would note is that this issue is caused by the fact that we import the functionality from all tools directly into the name space `aiida.tools`:
https://github.com/aiidateam/aiida-core/blob/5f50f0335761386ddacbb5de4a6934b659b20e2d/aiida/tools/__init__.py#L24-L27

I get that it makes the import path for tools shorter, but it will also lead to nameclashes if one tool wants to provide a function with the same name as another tool (i.e. it forces us to encode the namespacing into function names).
Perhaps something worth reconsidering.

CasperWA: What you're hinting at is done all over the code base, in all modules. It may be good to start moving away from wildcard imports, but I don't think I will be the one to make that change all throughout...

sphuber: In many places, I have done this intentionally though and I think is not necessarily bad practice. In the main modules (especially during the ORM rewrite of last year) I have used the approach of adding public facing interface in each module to the `__all__`. This then allows in the `__init__.py` of the corresponding module to use wildcard imports to import those and only those resources in the `__all__`. By doing this recursively all the way up to the second level module, e.g. `aiida.orm` all the public facing interface will be directly importable from there.

ltalirz: Yes, I agree that it is not bad per se (e.g. when importing class names of node types that are meant to be unique), I just wonder whether we should be doing it in `aiida.tools`. 

Anyhow, to get back to the original problem: we need to decide - do we move out the EXPORT_VERSION constant or do we stop importing it?

sphuber: There is another alternative where you make a "lazy-default" which is just a callback that will only be called when the actual command is called. Only downside may be that I am not sure if this makes it possible to have the value show up in the help string. If it does, then that would also require loading the ORM and so incur the slow down, although that is then isolated to just that command which might be ok.

ltalirz: I'd be fine with that.

ltalirz: @sphuber Ah hm... I tried but it doesn't quite do what we had in mind.

When providing a callable to `default`, it ends up being shown as
```
  -v, --version VERSION           Archive format version to migrate to.
                                  [default: (dynamic)]
```
One can use `show_default='latest version'` to get 
```
  -v, --version VERSION           Archive format version to migrate to.
                                  [default: (latest version)]
```
but then we're almost back to square one anyhow.

I also tried the `contextual_default` but there the default value isn't shown in the `-h` output at all.

Given this, I think the less convoluted approach is not to set a default and to go back to the original logic.
Do you agree?

aiida/cmdline/commands/cmd_export.py
```diff
@@ -141,31 +143,45 @@ def create(
...
-    from aiida.tools.importexport import EXPORT_VERSION, migration, extract_zip, extract_tar, ArchiveMigrationError
+    from aiida.tools.importexport import migration, extract_zip, extract_tar, ArchiveMigrationError
 
     if version is None:
         version = EXPORT_VERSION
 
+    if in_place:
+        if output_file:
+            echo.echo_critical('output file specified together with --in-place flag')
+        tempdir = tempfile.mkdtemp()
```
greschd, line 178: Minor nit, but I think `tempfile.TemporaryDirectory()` would be better here, even if used without the context manager.

The functional difference is that `tempfile.TemporaryDirectory()` gets cleaned when collected by the garbage collector, for example, if there is an error somewhere in the code between here and the `shutil.rmtree`.

ltalirz: Oh, nice, I wasn't aware of this feature. Thanks, will fix!

ltalirz: There might be a couple of other places in the AiiDA codebase where we may want to switch from `mkdtemp` to `TemporaryDirectory`. Feel free to open an issue about this.

greschd: Done, see https://github.com/aiidateam/aiida-core/issues/4336. 

Note that you'll have to remove it using `.cleanup()` instead of `shutil.rmtree` - otherwise it'll try to delete it twice (and fail).

aiida/cmdline/commands/cmd_export.py
CasperWA: For me this is fine (going back). I think I tried going through all this originally as well and thought I didn't want to bother untangling it for a feature that most people will not care about anyway. The main message is simply: "It will be the latest archive version" ... If one then wants to check that an archive indeed _is_ that latest version one could use the inspection functionality in verdi.

In the end, I agree that `aiida.tools` should maybe not include imports from submodules, but rather one should specify the individual tools. That makes more sense conceptually, I think.

sphuber: Agreed that the more convoluted method is not worth the trouble if it still doesn't get the fully desired result. So fine to revert. Please make a note in the commit explaining why this is and what we tried and why that didn't work. To hopefully prevent us from going on this wild goose chase once more in the future :)

ltalirz: Done!

ltalirz:
@sphuber This should be ready for review

aiida/cmdline/commands/cmd_export.py
```diff
@@ -141,31 +142,45 @@ def create(
...
-    from aiida.tools.importexport import EXPORT_VERSION, migration, extract_zip, extract_tar, ArchiveMigrationError
+    from aiida.tools.importexport import migration, extract_zip, extract_tar, ArchiveMigrationError, EXPORT_VERSION
 
     if version is None:
         version = EXPORT_VERSION
 
+    if in_place:
+        if output_file:
+            echo.echo_critical('output file specified together with --in-place flag')
+        tempdir = tempfile.TemporaryDirectory()
```
sphuber, line 177: not having this in a context manager, will not trigger the automatic cleanup. It would be good to try and use `with tempfile.TemporaryDirectory as tempdir` and not depend on your own `tempdir.cleanup` call. I realize that this would mean moving all code below within the context, but maybe you can refactor/reshuffle the code a bit to have the opening of the context as short as possible.

aiida/cmdline/commands/cmd_export.py
```diff
@@ -212,5 +227,12 @@ def migrate(input_file, output_file, force, silent, archive_format, version):
...
+        if in_place:
+            os.rename(output_file, input_file)
+            tempdir.cleanup()
+
         if not silent:
-            echo.echo_success('migrated the archive from version {} to {}'.format(old_version, new_version))
+            if new_version > old_version:
+                echo.echo_success('migrated the archive from version {} to {}'.format(old_version, new_version))
+            else:
+                echo.echo_success('archive was already at latest version {}'.format(new_version))
```
sphuber, line 238: I think I would move this check to the beginning of the command. With the `Archive` context manager util class, you can relatively easily get the version number and if the same you can skip all the unpackking/migrating/repacking logic. By checking the version number, the archive will have to be unpacked first, so that cost cannot be skipped but at least we are not repacking it again for nothing. Once we have the new export format that should even make the introspection cheap, even the unnecessary unpacking will no longer be necessary

aiida/cmdline/commands/cmd_export.py
greschd: I think the trouble with using a context manager is that it's only needed in the `in_place` branch, and not the other branch.

In my (admittedly limited) testing, the temporary directory will be destroyed when the Python interpreter exits at the latest.

sphuber: > In my (admittedly limited) testing, the temporary directory will be destroyed when the Python interpreter exits at the latest.

I still think that it might be nicer to use a context manager, despite it only being used in a particular branch of code. Part of the reason why it would be cumbersome in the current implementation is because the code is a bit messy. The whole code about migration (unpacking, checking, migration, repacking) should be done in some utility. Once that is done, having the context manager would not be so problematic.

But as long as the folder is cleaned after interpreter, I think it is ok for now

greschd: > But as long as the folder is cleaned after interpreter, I think it is ok for now

That's observed behavior, not sure if it's _guaranteed_ behavior, since "It is not guaranteed that ``__del__()`` methods are called for objects that still exist when the interpreter exits." (https://docs.python.org/3/reference/datamodel.html#object.__del__).

greschd: Another option would be using a `contextlib.nullcontext` for the `in_place=False` branch - creating the context manager object conditionally, and then entering it later.

Of course, if the code can be cleaned up to not need that, even better.

greschd: After some prodding: the cleanup is implemented with [`weakref.finalize`](https://docs.python.org/3/library/weakref.html#weakref.finalize), not a `__del__` method. I'm still not 100% sure about it, but I think unless things go absolutely haywire it should always be cleaned.

aiida/cmdline/commands/cmd_export.py
ltalirz: Right. It does change the user interface a bit: this change means you can no longer count on the new archive being present in the `outfile` in all cases (unless you use `--in-place`).
I agree however that we may want to save the repackaging step...

ltalirz: Change made!

aiida/cmdline/commands/cmd_export.py
sphuber: Hmm, still not ideal, but since this would require more refactoring to be effective, I will leave this for now.

sphuber (approved):

PR #4222, opened 2020-07-06
update paramiko to 2.7

ltalirz:
Version 2.7 of paramiko finally brings support for the OpenSSH private
key format, which has been the default on MacOS for some time.

See https://github.com/paramiko/paramiko/pull/1343

### DIFF
--- a/.github/workflows/setup.sh
+++ b/.github/workflows/setup.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -ev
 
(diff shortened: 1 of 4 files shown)


sphuber:
I am a bit confused. Shouldn't there also be an automated bot/runner that checks the `requirements/requirements*.txt` files and update the `paramiko` version in there? @csadorf how does this work again?

ltalirz:
> Shouldn't there also be an automated bot/runner that checks the requirements/requirements*.txt files

There is, but the paramiko version there is already 2.7.x

sphuber (approved):
Good to go, thanks

PR #4223, opened 2020-07-06
Show the default for switch options in transport CLI

sphuber:
The autogenerated command line interfaces for transport plugins were not
setting `show_default` to `True`, even if a `non_interactive_default` is
specified, which means that the help string would not say which of the
two switch values is the default.

### DIFF
--- a/aiida/transports/cli.py
+++ b/aiida/transports/cli.py
@@ -71,25 +71,33 @@ def transport_option_default(name, computer):
     return default
 
 
-def interactive_default(transport_type, key, also_noninteractive=False):
-    """Create a contextual_default value callback for an auth_param key."""
+def interactive_default(key, also_non_interactive=False):
+    """Create a contextual_default value callback for an auth_param key.
+
+    :param key: the name of the option.
+    :param also_non_interactive: indicates whether this option should provide a default also in non-interactive mode. If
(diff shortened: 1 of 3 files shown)


aiida/transports/cli.py
```diff
@@ -110,9 +110,10 @@ def create_option(name, spec):
     if 'default' in spec:
         kwargs['show_default'] = True
     else:
-        kwargs['contextual_default'] = interactive_default(
-            'ssh', name, also_noninteractive=spec.pop('non_interactive_default', False)
-        )
+        non_interactive_default = spec.pop('non_interactive_default', False)
+        kwargs['default'] = non_interactive_default
+        kwargs['show_default'] = True
```
ltalirz, line 115: could this then not just move outside the if/else?

sphuber:
> I can't say I fully understand what's going on with the different defaults here but I've checked that it fixes the help strings (and that interactive configuration still works as well).

I agree that the `non_interactive_default` key was not very clear, it also took me quite a while to figure out the original intention. I have updated the code and docstrings a bit to hopefully clarify its functioning. I have dismissed your first review since the code changed quite a bit. Would be good if you could test again that help strings are correct and that interactive works.

PR #4228, opened 2020-07-08
Docs: move in the documentation on caching

### LINKED ISSUE #3988: Docs: How to use caching to save computational resources

sphuber:
Fixes #3988 

The content is mostly moved in from existing content, just adapting the
style to the new guidelines.


docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+Maybe you run a great number of complex workflows in high-throughput that each may repeat the same calculation, or you may have to restart an entire workflow that failed somewhere half-way through.
+Since AiiDA stores the full provenance of each calculation, it can detect whether a calculation has been run before and, instead of running it again, simply reuse its outputs, thereby saving valuable computational resources.
+This is what we mean by **caching** in AiiDA.
+
+.. _how-to:run-codes:caching:enable:
+
+How to enable caching
+---------------------
+
+Caching is **not enabled by default**.
```
csadorf, line 385: Should we provide a very brief explanation as to why not or possibly just link to the "Limitations"?

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+
+.. note:: Caching is **not** implemented at the WorkChain/workfunction level (see :ref:`how-to:run-codes:caching:limitations` for details).
+
+
+.. _how-to:run-codes:caching:hashing:
+
+How are nodes hashed
+--------------------
+
+*Hashing* is turned on by default, i.e. all nodes in AiiDA are hashed.
```
csadorf, line 419:
```suggestion
*Hashing* is turned on by default, i.e., all nodes in AiiDA are hashed.
```

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+This means that even when you enable caching once you have already completed a number of calculations, those calculations can still be used retro-actively by the caching mechanism since their hashes have been computed.
+
+The hash of a ``Data`` node is computed from:
+
+* all attributes of the node, except the ``_updatable_attributes`` and ``_hash_ignored_attributes``
+* the ``__version__`` of the package which defined the node class
+* the content of the repository folder of the node
+* the UUID of the computer, if the node is associated with one
```
csadorf, line 427: Since this is part of the How-Tos, I'm not sure that this level of detail concerning how the hash is computed is necessary. The important piece of information is the fact that the hash is unique to the node, no?

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+------------------------
+
+Class level
+...........
+
+Besides an on/off switch per profile, the ``.aiida/cache_config.yml`` provides control over caching at the level of specific calculations using their corresponding entry point strings (see the output of ``verdi plugin list aiida.calculations``):
+
+.. code-block:: yaml
+
+    profile-name:
```
csadorf, line 475: Is `profile-name` a key-word or a placeholder for the *profile name*? I assume it's the latter, but this could be confusing.

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+        enabled:
+            - aiida.calculations:quantumespresso.pw
+        disabled:
+            - aiida.calculations:templatereplacer
+
+In this example, caching is disabled by default, but explicitly enabled for calculaions of the ``PwCalculation`` class, identified by its corresponding ``aiida.calculations:quantumespresso.pw`` entry point string.
+It also shows how to disable caching for particular calculations (which has no effect here due to the profile-wide default).
+
+For calculations which do not have an entry point, you need to specify the fully qualified Python name instead.
+For example, the ``seekpath_structure_analysis`` calcfunction defined in ``aiida_quantumespresso.workflows.functions.seekpath_structure_analysis`` is labelled as ``aiida_quantumespresso.workflows.functions.seekpath_structure_analysis.seekpath_structure_analysis``.
```
csadorf, line 486:
```suggestion
For example, the ``seekpath_structure_analysis`` calcfunction defined in ``aiida_quantumespresso.workflows.functions.seekpath_structure_analysis`` is labeled as ``aiida_quantumespresso.workflows.functions.seekpath_structure_analysis.seekpath_structure_analysis``.
```

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+
+.. code-block:: yaml
+
+    profile-name:
+        default: False
+        enabled:
+            - aiida.calculations:quantumespresso.*
+            - '*.seekpath_structure_analysis'
+
+You can even override a wildcard with a more specific entry.
```
csadorf, line 501:
```suggestion
Any entry with a wildcard is overridden by a more specific entry.
```

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+        enabled:
+            - aiida.calculations:*
+        disabled:
+            - aiida.calculations:quantumespresso.*
+
+
+Instance level
+..............
+
+Even when caching is turned off for a given calculation type, you can enable it on a case-by-case basis by using the :class:`~aiida.manage.caching.enable_caching` context manager for testing purposes:
```
csadorf, line 517:
```suggestion
You can enable caching on a case-by-case basis by using the :class:`~aiida.manage.caching.enable_caching` context manager regardless of the profile settings:
```

csadorf, line 517: Is the use of this context manager limited to testing purposes or could it be used "in production"?

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+.. code-block:: python
+
+    from aiida.engine import run
+    from aiida.manage.caching import enable_caching
+    with enable_caching(identifier='aiida.calculations:templatereplacer'):
+       run(...)
+
+.. warning::
+
+    This affects only the current python interpreter and won't change the behavior of the daemon workers.
```
csadorf, line 528:
```suggestion
    This affects only the current Python interpreter and won't change the behavior of the daemon workers.
```

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+
+#. Workflow nodes are not cached.
+   In the current design this follows from the requirement that the provenance graph be independent of whether caching is enabled or not:
+
+   * **Calculation nodes:** Calculation nodes can have data inputs and create new data nodes as outputs.
+     In order to make it look as if a cloned calculation produced its own outputs, the output nodes are copied and linked as well.
+   * **Workflow nodes:** Workflows differ from calculations in that they can *return* an input node or an output node created by a calculation.
+     Since caching does not care about the *identity* of input nodes but only their *content*, it is not straightforward to figure out which node to return in a cached workflow.
+
+   For the moment, this limitation is acceptable since the runtime of AiiDA WorkChains is usually dominated by expensive calculations, which are covered by the current caching mechanism.
```
csadorf, line 559:
```suggestion
   This limitation has typically no significant impact since the runtime of AiiDA WorkChains is commonly dominated by expensive calculations, which are covered by the current caching mechanism.
```
Let's state this more factual and leave it up to the user/reader to decide whether that is acceptable.

csadorf, line 559:
```suggestion
   For the moment, this limitation is acceptable since the runtime of AiiDA WorkChains is usually dominated by expensive calculations.
```

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+
+#. The caching mechanism for calculations *should* trigger only when the inputs and the calculation to be performed are exactly the same.
+   While AiiDA's hashes include the version of the python package containing the calculation/data classes, it cannot detect cases where the underlying python code was changed without increasing the version number.
```
csadorf, line 562:
```suggestion
   While AiiDA's hashes include the version of the Python package containing the calculation/data classes, it cannot detect cases where the underlying Python code was changed without increasing the version number.
```

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+   Another edge case would be if the parser lives in a different python package than the calculation (calculation nodes store the name of the parser used but not the version of the package containing the parser).
```
csadorf, line 563:
```suggestion
   Another edge case would be if the parser lives in a different Python package than the calculation (calculation nodes store the name of the parser used but not the version of the package containing the parser).
```

csadorf, line 563:
```suggestion
   Another scenario that can lead to an erroneous cache hit is if the parser and calculation are not implemented as part of the same Python package, because the calculation nodes store only the name, but not the version of the used parser.
```

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
 
-    `#3988`_
+Finally, while caching saves unnecessary computations, it does not save disk space: The output nodes of the cached calculation are full copies of the original outputs.
+The plan is to add data deduplication as a global feature at the repository and database level (independent of caching).
```
csadorf, line 566:
```suggestion
```
That should go into a roadmap, not into the documentation.

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
+     Since caching does not care about the *identity* of input nodes but only their *content*, it is not straightforward to figure out which node to return in a cached workflow.
+
+   For the moment, this limitation is acceptable since the runtime of AiiDA WorkChains is usually dominated by expensive calculations, which are covered by the current caching mechanism.
+
+#. The caching mechanism for calculations *should* trigger only when the inputs and the calculation to be performed are exactly the same.
+   While AiiDA's hashes include the version of the python package containing the calculation/data classes, it cannot detect cases where the underlying python code was changed without increasing the version number.
+   Another edge case would be if the parser lives in a different python package than the calculation (calculation nodes store the name of the parser used but not the version of the package containing the parser).
 
-    `#3988`_
+Finally, while caching saves unnecessary computations, it does not save disk space: The output nodes of the cached calculation are full copies of the original outputs.
```
csadorf, line 565: Is there any reason that this is shown as  separate paragraph instead of just another bullet point?

docs/source/howto/run_codes.rst
```diff
@@ -357,24 +357,215 @@ After :ref:`setting up your computer <how-to:run-codes:computer>` and :ref:`sett
...
-    title: Using caching to save computational resources
+.. _how-to:run-codes:caching:
+
+How to save computational resources using caching
+=================================================
+
+There are numerous reasons why you may need to re-run calculations you’ve already done before.
+Maybe you run a great number of complex workflows in high-throughput that each may repeat the same calculation, or you may have to restart an entire workflow that failed somewhere half-way through.
+Since AiiDA stores the full provenance of each calculation, it can detect whether a calculation has been run before and, instead of running it again, simply reuse its outputs, thereby saving valuable computational resources.
+This is what we mean by **caching** in AiiDA.
```
csadorf, line 378: I think there should be a fat warning (disclaimer) with a link to the limitations here, because – unless I misunderstood – there are some cases in which AiiDA can have a false cache hit, which would pose a significant risk to use caching. Since caching is not 100% safe, users need to know about this right from the start. I assume that's also the reason why it's not enabled by default?

docs/source/howto/run_codes.rst
sphuber: it should be the actual profile name. I assumed this was going to be clear because I explained this in "How to enable caching". I can repeat the extra clarify I added there as well

docs/source/howto/run_codes.rst
sphuber: Shouldn't we here not also rather maintain the style of your preceding comment? I.e., there you changed "You can even override.." into "Any entry with ... is overridden". I actually also prefer that style. Should we change the sentence here rather in something like:
> Caching can be enabled or disabled on a case-by-case basis by using the ....  regardless of the profile settings

sphuber: It can be used in production but with the given caveat that it only applies to the interpreter in which it is open. That means the daemon will ignore it. Since most production work will actually include the daemon, practically it won't be very useful and so it is mostly intended for quick work/testing

docs/source/howto/run_codes.rst
sphuber: It is already there, so will just delete the line

docs/source/howto/run_codes.rst
sphuber: Don't think so. Will include as bullet point

docs/source/howto/run_codes.rst
sphuber: I agree. I can move this to a Topics section and just link it here. Would that work?

csadorf: Yes, I think that would be ideal.

docs/source/howto/run_codes.rst
csadorf: Since users are expected to jump around in the how-tos, I think that duplication is warranted. Also, I missed it, and I probably already read this more thoroughly then most readers would. ;)

docs/source/howto/run_codes.rst
sphuber: There are two reasons, but not sure if we want to mention (any) of them
1. It is quite an unusual feature and if you are not aware of it, it can surprise users a lot.
2. _Especially_, if it does not always do the "correct" thing. This is not even talking about bugs, which would be bad, but for example, if you have a calculation that finished with a non-zero exit status, it _is_ a valid cache source and so will be used. But a plugin may have intended this to really not be cached from. So long story short, it is a bit complicated and we feel currently that users should be aware of what they are enabling.

docs/source/howto/run_codes.rst
sphuber: depends on what you mean by "false" cache hit. If it is really due to two calculations _not_ being equal, but AiiDA erroneously thinking them to be equal, that is bad. However, I think these cases are unlikely, although I cannot exclude them of course. The bigger question though is what I wrote in point 2) of my response to your next comment.

docs/source/howto/run_codes.rst
csadorf: I understand your explanation, but I strongly disagree with your conclusion that we should not mention this, see also my comment on the disclaimer below. The fact that caching is not reliable and false cache hits are expected, makes it even more important to mention this here.

sphuber: Fair, enough. I will add an explanation.

docs/source/howto/run_codes.rst
csadorf: Good call.

csadorf: Ok.

sphuber:
Thanks for the review @csadorf . I have addressed the simple comments and split off the details to a Topics section, and added a disclaimer. See commit message for reasoning of location of new section.

docs/source/howto/run_codes.rst
csadorf: A false cache hit in this context would be the code using a cached result when in fact the result would be different on recalculation. This is what I consider the worst case scenario and what I think we have to warn the user about.

PR #4229, opened 2020-07-08
remove explicit numpy install

ltalirz:
seems no longer needed... and possibly broke CI

Note: Creating a draft PR for this since @sphuber mentioned that tests were passing fine on his fork (so better test that things actually work on the official repo as well).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,10 +12,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Set up Python 3.7
+    - name: Set up Python 3.8
       uses: actions/setup-python@v1
       with:
-        python-version: 3.7
+        python-version: 3.8
 
     - name: Install system dependencies
       run: |
@@ -154,10 +154,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Set up Python 3.7
+    - name: Set up Python 3.8
       uses: actions/setup-python@v1
       with:
-        python-version: 3.7
+        python-version: 3.8
 
     - name: Install python dependencies
       run: |


sphuber:
Note that this was put in because `pymatgen` will installs `numpy` during its build and it ignores our version requirements in the `setup.json`. This means that always the latest version of `numpy` will be installed if we install `pymatgen`, which has broken our builds multiple times in the past. Even though it works now then, I would not remove it, because as soon as `numpy` releases another release candidate, it will most likely break our builds.

ltalirz:
@sphuber It looks like the issue is with the python 3.7 environment.
I can make this the only change for the moment

sphuber:
> It looks like the issue is with the python 3.7 environment.

Do we know why? As in, are we sure that this is not a random error that occurs and this time we got lucky and the tests passed? Reason I am asking is that the build on my fork ran without problems, so it looks like a transient problem no?

ltalirz:
It may be transient, but tests failed on my fork, too.
I don't know the reason but we need to merge at least the docs PR; i.e. it is either ignore the failure or switch to python 3.8 env for the moment.

sphuber:
> It may be transient, but tests failed on my fork, too.

Yeah, but so the point is that even if we merge this, the next build may just fail again. Or am I missing something?

ltalirz:
Well, of course it can start failing again (and I'm sure at some point in the future it will again ;-) )
But for the time being it's fixed.

ltalirz:
I think at the moment it may be specific to the python 3.7 env

ltalirz:
We could open a bug report here: https://github.com/actions/setup-python/issues

ltalirz:
There is an open issue mentioning issues with the installation of dependencies
https://github.com/actions/setup-python/issues/105

The ubuntu-latest runner has python 3.8 installed but not python 3.7:
https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu2004-README.md

sphuber:
> I think at the moment it may be specific to the python 3.7 env

What is that based on though? Is there anything you saw that hints at this? Or is it just a guess?

sphuber:
> you can check whether it fails now when you restart the workflow on your fork)

Just reran it and the `verdi` job ran fine even with Python 3.7: https://github.com/sphuber/aiida_core/runs/850405963

ltalirz:
Then I guess it is just "the python 3.7 environment for the runners that are being used aiidateam"?

ltalirz:
As expected, after restarting the runs of the old PR, they still fail.
I say we merge this now. We can figure out later what exactly caused this.

ltalirz:
Ok, on your fork, you have an older version of the runner virtual environment:
https://github.com/sphuber/aiida_core/runs/850405963#step:1:8
https://github.com/aiidateam/aiida-core/pull/4229/checks?check_run_id=850252814#step:1:8

Probably the new one is broken

ltalirz:
Opened issue on the corresponding repo https://github.com/actions/virtual-environments/issues/1202

PR #4234, opened 2020-07-09
fixes for plugin howto

ltalirz:


docs/source/howto/plugin_codes.rst
```diff
@@ -346,7 +346,7 @@ With the entry points set up, you are ready to launch your first calculation wit
           code = load_code('add@localhost')
         except NotExistent:
           # Setting up code via python API (or use "verdi code setup")
-          code = orm.Code(label='add', remote_computer_exec=[computer, '/bin/bash'])
+          code = orm.Code(label='add', remote_computer_exec=[computer, '/bin/bash'], input_plugin_name='add')
```
sphuber, line 349: This is also incorrectly indented. Should be four spaces. Same goes for in the try block itself

ltalirz: thanks, in this case it actually did not raise an error because it was the only indentation in the code.
anyhow, I fixed it now

Issue #4199, opened 2020-06-25
ssh without sftp ?


adegomme:
I recently stumbled on a supercomputing center where SFTP server is not running and actually forbidden (executing the server manually is technically possible, but not really an option), while SSH access is otherwise fully functional. I asked, and the underlying idea is to prevent users from using SSHFS and sharing their remote home/work folders without any control. No comment on that, but maybe this policy will be applied by other supercomputing centers at some point.

This prevents Aiida from connecting to the system, as it uses paramiko's sftp client automatically. This raises a few questions about how to proceed in this case.

- Has anyone been affected by this case and found a workaround ?
- Would a "pure" SSH+command mockup of sftp commands used in aiida be possible/a good idea when sftp fails to initialize ? I will begin to toy with this to see if it can work.
- Would it be in another transport plugin, or in the default ssh one, and used as a fallback?
- verdi computer test output was a bit confusing in this case, and could maybe provide a specific test for sftp connexion (obviously system is under maintenance right now and I can't provide the output)? Maybe checking the output of open_sftp() call and returning an explicit error could do the trick to assist the user (I know now that in the doc it's already written to check sftp access).

adegomme:
update:
I've just managed to get "Success: all 5 tests succeeded" on a cluster without sftp, with pure SSH + shell to replace sftp in all calls that need it (cat+echo to perform the put/get actually do the trick, no idea about performance). 

I will try to go further and launch some computation soon, to see if it's robust enough (not yet, I guess).

greschd:
For moving the files, couldn't you use `scp`? I guess that shouldn't be blocked, and performance should be acceptable.

In general, the only downside I could see of adding this as a fallback is that users who _have_ SFTP access (but misconfigured for some reason) would not notice, and end up with potentially degraded performance.
In any case I think it's a good idea to start testing this separately (which I guess is what you're doing now), and then we can decide how to integrate it.

giovannipizzi:
Indeed, it would be probably a wise idea to use this as a different transport plugin (maybe with common code in a base class 'SshBase' and then reimplementing the differences in the two implementations).

Note that with the current implementation, if you try to transfer a file (via STDOUT/STDERR as you will be doing) larger than a few MB, it will probably hang. This does not pop out so often in AiiDA because we typically big transfer files via SFTP ;-)
But you should definitely see it.

I think I fixed this in this PR: #3787 but never managed to properly test it.

[...]

PR #4235, opened 2020-07-09
Ssh only transport

adegomme:
As discussed in #4199  here is a preliminary implementation of a SSH transport without SFTP.
It's been tested and works to submit jobs and gather results on two different systems with SFTP disabled/forbidden (one with slurm, one direct).
It has not yet been tested with large files yet ( see #3787 ) so this should not be merged. I will do this as soon as possible.

I switched from the earlier mockup of an SFTP client (in order to keep all self.sftp.xxx calls as is), to a different transport. This loses the ability to dynamically default to sshonly when sftp is found to not work, but is generally cleaner.

Only changes in ssh.py were:
- split initialization of file transport client
- add an explicit message when sftp opening fails. 
- split calls to separate self.sftp calls in their own methods (overriden in the sshonly transport).
- when logging remote commands to execute, limit the output to 1k chars, as full files are copied through this mechanism in sshonly mode.


ltalirz:
@adegomme just to let you know that in case you would like someone to review this draft PR, you'll need to ping them

adegomme:
closing - this is mainly moved to a plugin, and I will open another PR for the ssh.py parts which could need changing

PR #4236, opened 2020-07-09
docs: fix in plugins intro

ltalirz:
The information printed by the launch script was not described
correctly.

### DIFF
--- a/docs/source/howto/plugin_codes.rst
+++ b/docs/source/howto/plugin_codes.rst
@@ -379,7 +379,6 @@ With the entry points set up, you are ready to launch your first calculation wit
     .. code-block:: console
 
         $ verdi run launch.py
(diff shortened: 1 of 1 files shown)


ltalirz:
@sphuber Thanks for the help with merging the first fix!

Here comes another one that was discovered by a participant.

ltalirz:
@mbercx Would you mind reviewing this?

sphuber:
seems like one test failed due to random thread error but the other due to the uploading of the coverage report failing. I think it would be better to not have the test fail if the coverage upload fails. Have seen this one quite a bit and we anyway don't really respond if coverage decreases.

ltalirz:
> I think it would be better to not have the test fail if the coverage upload fails.

Indeed, if this happens regularly, let's change this.
Happy to open a PR

PR #4237, opened 2020-07-09
Pin Python to version 3.7.7 for test-install.

csadorf:


csadorf:
The runner has only the latest Python version installed.

PR #4238, opened 2020-07-09
Update the setup-python step to v2.

csadorf:
For test-install CI workflow.


csadorf:
Does not resolve issue.

PR #4239, opened 2020-07-09
ci: don't fail job if coverage upload fails

ltalirz:
The upload of test coverage reports to codecov sometimes fails,
currently causing the job to fail.

Since maintaining test coverage is currently not a strict requirement,
we can live with not getting it right some of the time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,7 +144,7 @@ jobs:
         name: aiida-pytests-py3.5-${{ matrix.backend }}
         flags: ${{ matrix.backend }}
         file: ./coverage.xml
-        fail_ci_if_error: true
+        fail_ci_if_error: false  # don't fail job, if coverage upload fails
 
   verdi:
 


ltalirz:
Following up on @sphuber 's comment https://github.com/aiidateam/aiida-core/pull/4236#issuecomment-656189249

.github/workflows/ci.yml
```diff
@@ -140,6 +140,7 @@ jobs:
     - name: Upload coverage report
       if: matrix.python-version == 3.5 && github.repository == 'aiidateam/aiida-core'
       uses: codecov/codecov-action@v1
+      continue-on-error: true  # don't fail job, if coverage upload fails
```
sphuber, line 143: Shouldn't we rather set line `147` to `false`?

csadorf: I think one refers to the upload process itself and the other one allows us to fail the CI if the coverage is insufficient.

sphuber: From [their documentation](https://github.com/codecov/codecov-action#arguments):
> `fail_ci_if_error`: Specify if CI pipeline should fail when Codecov runs into errors during upload. Defaults to false

I interpret this to mean that us setting this to `True` is exactly what is failing the pipeline when the upload false. Or am I misreading this?

ltalirz: @sphuber , thanks for pointing this out - indeed the action has already intrinsic support for this.
Will fix

csadorf: Should have look it up instead of just guessing, sorry about that.

Issue #4243, opened 2020-07-10
Provide cli for modifying existing profiles


ltalirz:
Both `verdi setup` and `verdi quicksetup` are currently limited to creating new profiles.

There is no interface for modifying the settings of an existing profile (other than editing the `config.json` file directly, but such an interface has been requested at least twice during the 2020 July tutorial.

I think we should consider introducing such an API, with a clear statement on the scope: it *only* changes the setting and does nothing to "migrate" the profile data behind.
I.e. it is almost the same as editing the config.json file by hand, except it may do some additional validation on inputs (e.g. does the path to the new file repo exist, ...).
 
Copying the list of settings from the `verdi setup` help for context:
```
  --profile PROFILE               The name of the new profile.  [required]
  --email EMAIL                   Email address associated with the data you
                                  generate. The email address is exported
                                  along with the data, when sharing it.
                                  [required]

  --first-name NONEMPTYSTRING     First name of the user.  [required]
  --last-name NONEMPTYSTRING      Last name of the user.  [required]
  --institution NONEMPTYSTRING    Institution of the user.  [required]
  --db-engine [postgresql_psycopg2]
                                  Engine to use to connect to the database.
  --db-backend [django|sqlalchemy]
                                  Database backend to use.
  --db-host HOSTNAME              Database server host. Leave empty for "peer"
                                  authentication.

  --db-port INTEGER               Database server port.
  --db-name NONEMPTYSTRING        Name of the database to create.  [required]
  --db-username NONEMPTYSTRING    Name of the database user to create.
                                  [required]

  --db-password TEXT              Password of the database user.  [required]
  --repository DIRECTORY          Absolute path to the file repository.
```

sphuber:
I see the point of the feature, but I think the behavior you propose is actually dangerous and counter productive. If users want to change something, they most likely want everything to work as expected afterwards. If this is not guaranteed, but all the command does is just update the `config.json`, why do we need a command for this? If users are anyway expected to understand the inner workings of AiiDA to make these kinds of changes work, why not just let them manually update the config file? Its location is not hidden after all, it is shown in multiple places.

ltalirz:
I did mention editing the configuration file as an alternative, in particular after we've switched to the yaml format which then allows to add comments etc. 
I think this is a viable alternative - it just means there will be no validation of inputs, which can be a downside to having a dedicated API.

In any case, however, I do think we will need a (little) bit of documentation on how to do this.

PR #4245, opened 2020-07-10
Add the `verdi profile caching` command

### LINKED ISSUE #4240: Add `verdi` command that parses and prints the caching configuration
This is useful to test whether the caching config file is found correctly and correctly parsed

sphuber:
Fixes #4240 

This command will try to load the caching configuration from the caching
configuration file, for any given profile. This is useful for users to
make sure that the file is correctly named and contains correct syntax,
before actually starting to run new calculations.

To test this, a new pytest fixture is created that creates a completely
new and independent configuration folder, along with fixtures to create
profiles to add to the config and caching configuration files. Similar
code already exists for normal unittests, but this can be removed once
those have been refactored to pytests.

### DIFF
--- a/aiida/cmdline/commands/cmd_profile.py
+++ b/aiida/cmdline/commands/cmd_profile.py
@@ -54,11 +54,38 @@ def profile_show(profile):
     """Show details for a profile."""
 
     if profile is None:
-        echo.echo_critical('no profile to show')
+        echo.echo_critical('no default profile is defined')
 
     echo.echo_info('Profile: {}'.format(profile.name))
     data = sorted([(k.lower(), v) for k, v in profile.dictionary.items()])
     echo.echo(tabulate.tabulate(data))
+
+
+@verdi_profile.command('caching')
+@arguments.PROFILE(default=defaults.get_default_profile)
+def profile_caching(profile):
+    """Show the caching configuration for a profile."""
+    from aiida.manage.caching import get_configuration, get_configuration_filepath
+
+    if profile is None:
+        echo.echo_critical('no default profile is defined')
+
+    try:
+        config = get_configuration()
+    except FileNotFoundError:
+        echo.echo_critical('caching configuration file `{}` does not exist'.format(get_configuration_filepath()))
+    except exceptions.ConfigurationError as exception:
+        echo.echo_critical('caching configuration is invalid: {}'.format(str(exception)))
+
+    print('PROFILE', profile)
+
+    try:
+        profile_config = config[profile.name]
+    except KeyError:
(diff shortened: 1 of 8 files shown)


sphuber:
Note that this is missing documentation, but I will add this once PR #4228 has been merged.

ltalirz:
Thanks @sphuber !

Two questions from browsing through quickly would be:
 * why not include the caching configuration in the output of `verdi profile show` as we are anyhow planning to merge the caching configuration into the configuration file?
 * for the new fixtures for creating profiles: can we try to reuse somehow this context manager?
https://github.com/aiidateam/aiida-core/blob/e4c7fb51c44cb260c4f983f3bdd0a54688659901/tests/utils/configuration.py#L47

I'll give a proper review next week

sphuber:
>     * why not include the caching configuration in the output of `verdi profile show` as we are anyhow planning to merge the caching configuration into the configuration file?

My thinking is that at some point we want to add an endpoint to also modify the caching from the interface. I though of `verdi config` but its interface doesn't support what we'd need for the config, so I thought the best place was `verdi profile caching`. Glad to have input if you think of a better place. Didn't think it deserved a top-level command group in any case. Could be wrong though.

>     * for the new fixtures for creating profiles: can we try to reuse somehow this context manager?
>       https://github.com/aiidateam/aiida-core/blob/e4c7fb51c44cb260c4f983f3bdd0a54688659901/tests/utils/configuration.py#L47
> 

See the commit and PR message. The tests using it should actually be refactored to `pytest` in which case they should use this new fixture instead. Had already opened [an issue for this refactoring](https://github.com/aiidateam/aiida-core/issues/4249).

> I'll give a proper review next week

Thanks a lot!

chrisjsewell:
> I though of verdi config but its interface doesn't support what we'd need for the config

I think ideally this should go in `verdi config`. As mentioned in https://github.com/aiidateam/aiida-core/issues/4260#issuecomment-657520302, I think we are arbitrarily limiting the config interface by trying to mirror exactly `git config`.

ltalirz:
I agree with @chrisjsewell that it would be nice to come to an agreement on what the role of the `verdi profile` vs `verdi config` commands should be going forward (this also ties into https://github.com/aiidateam/aiida-core/issues/4243 ).

I could well imagine having `verdi config` as an API for modifying the config.json file (possibly starting just with things like caching etc.), while `verdi profile` as the commands that perform additional actions, such as populating the database with tables or creating the db in the first place.

chrisjsewell:
note I've made an issue for discussing `verdi config` (with a proposal) #4584

chrisjsewell:
FYI this PR is superceded by #4712 😜

ltalirz:
Then let's close this

PR #4251, opened 2020-07-10
Use the full host name when using a proxy

giovannipizzi:


sphuber:
Is this absolutely necessary, or will it just help the `verdi computer configure` to automatically detect the settings from the SSH config? Not saying we should not merge it, but if that is the case, we might want to mention it. Does this also hold for all types of connections or just for SSH over proxy?

giovannipizzi:
> Is this absolutely necessary, or will it just help the verdi computer configure to automatically detect the settings from the SSH config? Not saying we should not merge it, but if that is the case, we might want to mention it. 

It is just to help verdi computer configure. I have already written this, is this not clear enough?

"You need to use the ``FULLHOSTNAME_TARGET`` also in the ``Host`` line to make sure that AiiDA can automatically read this data and use it in the ``verdi computer configure`` step."

> Does this also hold for all types of connections or just for SSH over proxy?

For anything we want `verdi computer configure ssh` to pick automatically. Since there are many things asked, I think that we ask people to use the full hostname, and `ssh FULLHOSTNAME_TARGET` works on the command line, we have a high chance and just pressing enter at any prompt of `verdi computer configure ssh` will just work. And if `ssh` does not work users will try to debug that first.

I have had more than one user that I tried to help, who didn't configure the ``FULLHOSTNAME_TARGET`` but the short version, and it became much harder to remember and specify the correct settings by hand in `configure`, and then debug and fix if they weren't correct.

sphuber:
We were discussing with @ramirezfranciscof and @chrisjsewell and we think their might actually be an improvement to be made in the code. If we understand correctly, the situation currently is as follows.

1. `verdi computer setup` will ask for the fully qualified hostname, e.g. `daint.cscs.ch`.
2. With paramiko, we scan the `.ssh/config` if it exists to see if there is an entry where `Host` equals that value and then use the configuration values in that.

The problem now is that people often have a configuration that looks like the following:
```
Host daint
    Hostname daint.cscs.ch
    User sphuber
```
that is, they use a shorthand for the `Host` which is just a label. This allows one to do `ssh daint` instead of being forced to type the longer `daint.cscs.ch`. With the current code in `aiida-core` one is force to either a) not profit from automatic configuration detection by paramiko or b) adapt their ssh config, forcing them to also type the long form outside of AiiDA.

It looks like this limitation was "chosen" in AiiDA because `verdi computer setup` adds for the "hostname", which should be the fully qualified one, and just happens to double as a label to fetch a configuration entry. Wouldn't it be possible to allow `hostname` to refer to a `Host` label in the SSH config, and if it matches, then it just replaces `hostname` with the actual value specified in the config. We have to deal with the case where that is not defined in the config, but I think anyway that config would be invalid without a FQHN.

@giovannipizzi what do you think?

giovannipizzi:
Probably what you suggest is a good idea - we have to implement to check if there are strange corner cases but I agree that covering that use case would be good

sphuber:
> Probably what you suggest is a good idea - we have to implement to check if there are strange corner cases but I agree that covering that use case would be good

OK, I opened issue #4387 for this

Issue #2926, opened 2019-05-27
Rename `name` column of `Computer` to `label`


sphuber:
This to be consistent with its function compared to all other ORM models, such as `Node` and `Group`

CasperWA:
Note for import/export:
This will (obviously) change the import/export, since "name" of a `Computer` is (and was) exported.

sphuber:
@ltalirz @giovannipizzi given that this would break user facing API, we also don't change this for `v1.0.0`? If so, just like with not merging `User.first_name` and `User.last_name`, what milestone *would* we plan this for?

ltalirz:
strictly speaking 2.0

giovannipizzi:
Note that however we are already changing the API e.g. for Users, where we are dropping a number of columns. So I still believe that in this case a discussion is in order... If we do the change, since we are already in the beta phase, we could already put in 'label' along with 'name', but keep both working with a deprecation warning and drop it in 1.1. If we agree, we just need to evaluate how much work this requires and if we manage to do it before 1.0 (it's possible that we don't...). If we want to try to discuss this before 1.0, I'd suggest to open another issue pointing to this (and to the first_name/last_name one?) and put it in the 1.0 project, that says clearly that we do it only if we have enough time (and we first need to solve the rest).

sphuber:
@giovannipizzi what makes this even more problematic, is that applying these changes also need to change the export archive version. During the PRs we merged yesterday I *did* update the export schema, to make export/import work, but (as @CasperWA correctly pointed out) I did not up the archive version. Since it is in develop, people now can theoretically export archives which will have `v0.4` but with different content than those archives with the same version but created before the changes of the referenced PRs. 

I am about the create a PR to up the version. But again at that point, as soon as we introduce another change in the schema, we are strictly forced to up the archive version as well.

This being said, if we want to discuss whether these proposed changes should be done for v1.0, we should do it now rather than later. Otherwise we should maybe just decide to push it to `v2.0`.

PR #4252, opened 2020-07-10
Deprecate getter and setter methods of `Computer` properties

sphuber:
Prepares for #2926 

From `aiida-core==1.0.0`, we have started using properties for getters
and setters of the basic attributes of ORM entities. Rule of thumb here
is that if the attribute corresponds to a column on the underlying
database model, a property is used.

In addition, `label` is the preferred name for the third entity
identifier, alongside the ID and UUID. This is already the case for most
entities, except for `Computer` which is still using `name`. Here,
`name` is deprecated and replaced for `label`. The changes are not yet
propagated to the backend, which will be done once the deprecated
resources are fully removed. This is fine because the backend is not
part of the public API so doesn't have to go through a deprecation path.
The `name` keyword in `Computer.objects.get` is also deprecated and
replaced by `label`.

### DIFF
--- a/aiida/orm/computers.py
+++ b/aiida/orm/computers.py
@@ -8,7 +8,6 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Module for Computer entities"""
-
 import logging
 import os
 import warnings
@@ -17,6 +16,7 @@ from aiida import transports, schedulers
 from aiida.common import exceptions
 from aiida.common.warnings import AiidaDeprecationWarning
 from aiida.manage.manager import get_manager
+from aiida.orm.implementation import Backend
 from aiida.plugins import SchedulerFactory, TransportFactory
 
 from . import entities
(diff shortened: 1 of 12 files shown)


aiida/orm/computers.py
```diff
@@ -51,7 +51,33 @@ class Computer(entities.Entity):
...
         def list_names(self):
+            """Return a list with all the names of the computers in the DB.
+
+            .. deprecated:: 1.4.0
+                Will be removed in `v2.0.0`, use `list_labels` instead.
+            """
+            return self._backend.computers.list_names()
+
+        def list_labels(self):
             """Return a list with all the names of the computers in the DB."""
```
ltalirz, line 81:
```suggestion
            """Return a list with all the labels of the computers in the DB."""
```

aiida/orm/computers.py
```diff
@@ -60,24 +86,37 @@ def delete(self, id):  # pylint: disable=redefined-builtin,invalid-name
...
-        :type hostname: str
-        :type description: str
-        :type transport_type: str
-        :type scheduler_type: str
-        :type workdir: str
-        :type backend: :class:`aiida.orm.implementation.Backend`
-
-        :rtype: :class:`aiida.orm.Computer`
+            .. deprecated:: 1.4.0
+                The keyword name will be removed in `v2.0.0`, use `label` instead.
```
ltalirz, line 102:
```suggestion
                The `name` keyword will be removed in `v2.0.0`, use `label` instead.
```

aiida/orm/computers.py
```diff
@@ -297,49 +334,100 @@ def store(self):
...
+        """
         return self._backend_entity.name
 
+    @label.setter
+    def label(self, value: str):
+        """Set the computer label.
+
+        :param value: the label to set.
+        """
+        self._backend_entity.set_name(value)
```
ltalirz, line 350: interesting - why doesn't the backend entity have a setter as well?

aiida/orm/computers.py
```diff
@@ -297,49 +334,100 @@ def store(self):
...
-    def get_metadata(self):
-        return self._backend_entity.get_metadata()
+        :return: the transport_type.
+        """
+        return self._backend_entity.get_transport_type()
 
-    def set_metadata(self, metadata):
+    @transport_type.setter
+    def transport_type(self, value: str):
+        """Set the computer transport_type.
```
ltalirz, line 410:
```suggestion
        """Set the computer transport type.
```

aiida/orm/computers.py
```diff
@@ -297,49 +334,100 @@ def store(self):
...
+
+        :param value: the transport_type to set.
```
ltalirz, line 412:
```suggestion
        :param value: the transport type to set.
```

aiida/orm/computers.py
```diff
@@ -720,6 +731,158 @@ def get_configuration(self, user=None):
...
 
+    @property
+    def name(self):
+        """Return the computer name.
+
+        .. deprecated:: 1.4.0
+            Will be removed in `v2.0.0`, use the `label` property instead.
+        """
+        warnings.warn('this property is deprecated, use `label` instead', AiidaDeprecationWarning)  # pylint: disable=no-member
+        return self._backend_entity.name
```
ltalirz, line 742: you might want to use the  `label` property instead (in case we need to change something there)

aiida/orm/computers.py
```diff
@@ -720,6 +731,158 @@ def get_configuration(self, user=None):
...
+
+    def get_name(self):
+        """Return the computer name.
+
+        .. deprecated:: 1.4.0
+            Will be removed in `v2.0.0`, use the `label` property instead.
+        """
+        warnings.warn('this property is deprecated, use `label` instead', AiidaDeprecationWarning)  # pylint: disable=no-member
+        return self._backend_entity.get_name()
```
ltalirz, line 751: similar here, you might just want to return `self.label`; also below

aiida/orm/computers.py
```diff
@@ -720,6 +731,158 @@ def get_configuration(self, user=None):
...
+
+    def get_hostname(self):
+        """Get this computer hostname
+
+        .. deprecated:: 1.4.0
+            Will be removed in `v2.0.0`, use the `hostname` property instead.
+
+        :rtype: str
+        """
+        warnings.warn('this method is deprecated, use `hostname` instead', AiidaDeprecationWarning)  # pylint: disable=no-member
```
ltalirz, line 770:
```suggestion
        warnings.warn('this method is deprecated, use the `hostname` property instead', AiidaDeprecationWarning)  # pylint: disable=no-member
```

tests/cmdline/commands/test_computer.py
```diff
@@ -146,13 +146,13 @@ def test_mixed(self):
...
 
         self.assertEqual(new_computer.description, options_dict_full['description'])
         self.assertEqual(new_computer.hostname, options_dict_full['hostname'])
-        self.assertEqual(new_computer.get_transport_type(), options_dict_full['transport'])
-        self.assertEqual(new_computer.get_scheduler_type(), options_dict_full['scheduler'])
+        self.assertEqual(new_computer.transport_type, options_dict_full['transport'])
+        self.assertEqual(new_computer.scheduler_type, options_dict_full['scheduler'])
         self.assertEqual(new_computer.get_mpirun_command(), options_dict_full['mpirun-command'].split())
         self.assertEqual(new_computer.get_shebang(), options_dict_full['shebang'])
         self.assertEqual(new_computer.get_workdir(), options_dict_full['work-dir'])
```
ltalirz, line 158: What about these - shouldn't these also become properties?

aiida/orm/computers.py
sphuber: No particular reason other than it was never added. I haven't touched the backend yet since I was planning to do so when I would remove all deprecations and tests. Since it is the backend and so not public API I can then change whatever I need and clean things up.

tests/cmdline/commands/test_computer.py
sphuber: When making these changes I have followed the following:
> Rule of thumb here is that if the attribute corresponds to a column on the underlying
database model, a property is used.

Since these properties here are stored in the `metadata` column, and in a sense are comparable to a node's attributes, I am not sure if we should give properties for all of them. This is an arbitrary choice ultimately, but one that at least tries to keep consistency with the other entities such as the `Node` class. The database columns are exposed through properties, while anything else gets normal methods. In the latter, we should even ask ourselves to what level we want to give individual methods to all these sub attributes. Giving too many clutters the interface, however, it does make it tab-discoverable. Putting them behind a generic `Computer.get_property('shebang')` makes the interface cleaner and more maintainable but maybe less discoverable. I am not sure yet what the best solution is here

aiida/orm/computers.py
sphuber: the backend entity doesn't have a `label` property though. The plan was to correct this once all the `name` related code is ripped out just before `v2.0`. Since it is not public facing API we can change that at any time

ltalirz: I meant. self.label

sphuber: I see, will do

ltalirz: Not sure I was clear - I meant to simply replace this by `return self.label`, since we're repeating ourselves here which might (in principle) lead to divergence between `self.label` and `self.get_label()`.

In any case, since this is just one line and is going to be removed eventually, it's just minor and I'll approve now.

PR #4253, opened 2020-07-10
`verdi status`: do not except when no profile is configured

### LINKED ISSUE #4250: `verdi status` shows error message if no profile is configured
```
$ verdi status
 ✔ config dir:  /Users/leopold/.aiida
 ✘ profile:     Unable to read AiiDA profile
'NoneType' object has no attribute 'name'
```

sphuber:
Fixes #4250 

Instead, print that no profile could be found and suggest that one is
setup with `verdi quicksetup` or `verdi setup`.

To test this, a new pytest fixture is created that creates a completely
new and independent configuration folder, along with fixtures to create
profiles to add to the config and caching configuration files. Similar
code already exists for normal unittests, but this can be removed once
those have been refactored to pytests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,7 +8,11 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Configuration file for pytest tests."""
-import pytest  # pylint: disable=unused-import
+import os
+
+import pytest
+
+from aiida.manage.configuration import Config, Profile, get_config
 
 pytest_plugins = ['aiida.manage.tests.pytest_fixtures']  # pylint: disable=invalid-name
 
@@ -24,7 +28,6 @@ def non_interactive_editor(request):
 
     :param request: the command to set for the editor that is to be called
     """
-    import os
     from unittest.mock import patch
     from click._termui_impl import Editor
 
@@ -32,7 +35,6 @@ def non_interactive_editor(request):
     os.environ['VISUAL'] = request.param
 
     def edit_file(self, filename):
-        import os
         import subprocess
         import click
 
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -144,3 +148,101 @@ def _generate_calculation_node(process_state=ProcessState.FINISHED, exit_status=
...
 
     return _generate_calculation_node
+
+
+@pytest.fixture
+def create_empty_config_instance() -> Config:
+    """Create a temporary configuration instance.
+
+    This creates a temporary directory with a clean `.aiida` folder and basic configuration file. The currently loaded
+    configuration and profile are stored in memory and at are automatically restored at the end of this context manager.
```
chrisjsewell, line 158:
```suggestion
    configuration and profile are stored in memory and are automatically restored at the end of this context manager.
```

tests/conftest.py
```diff
@@ -144,3 +148,101 @@ def _generate_calculation_node(process_state=ProcessState.FINISHED, exit_status=
...
+        # Create the instance base directory structure, the config file and a dummy profile
+        settings.create_instance_directories()
+
+        # The constructor of `Config` called by `load_config` will print warning messages about migrating it
+        with Capturing():
+            configuration.CONFIG = configuration.load_config(create=True)
+
+        try:
+            yield get_config()
+        finally:
```
chrisjsewell, line 190: Its not generally necessary to wrap `yield` in a `try/finally` block with pytest fixtures: https://docs.pytest.org/en/5.4.3/fixture.html#fixture-finalization-executing-teardown-code

tests/conftest.py
```diff
@@ -144,3 +148,101 @@ def _generate_calculation_node(process_state=ProcessState.FINISHED, exit_status=
...
+
+    :return: a new empty config instance.
+    """
+    import tempfile
+
+    from aiida.common.utils import Capturing
+    from aiida.manage import configuration
+    from aiida.manage.configuration import settings, load_profile, reset_profile
+
+    with tempfile.TemporaryDirectory() as dirpath:
```
chrisjsewell, line 168: probably best to just use the `tmp_path` fixture here: https://docs.pytest.org/en/5.4.3/reference.html#std:fixture-tmp_path

sphuber:
Thanks @chrisjsewell . All comments are addressed.

PR #4258, opened 2020-07-12
Fix pre-commit configuration to reinstate `pylint` running

sphuber:
The `pylint` hook was ignored because the `hook` key in the `local`
category was defined twice. In this case `pre-commit` won't complain
that the file is invalid but the second just overwrites the first
declaration. In addition, the hook was missing the keys `entry`, which
is required, and the `types` key which restricts it to running only on
Python file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,10 +36,11 @@ repos:
     hooks:
     -   id: pylint
         name: pylint
+        entry: pylint
+        types: [python]
         language: system
         exclude: *exclude_files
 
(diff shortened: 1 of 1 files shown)


sphuber:
> that is the right fix, thanks @sphuber 

Thanks again for noticing.

> A bit surprised that we didn't introduce any pylint complaints in the meanwhile ;-)

Indeed, although all commits were essentially CI or Docs.
I had two code commits, but after all this time, I guess I have internalized the majority of pylint's gripes :)

PR #4261, opened 2020-07-13
DOC: pip install package groups for zsh

unkcpz:
quatation mark the name of package groups so the command also works for `zsh` .

### DIFF
--- a/docs/source/intro/installation.rst
+++ b/docs/source/intro/installation.rst
@@ -462,7 +462,7 @@ In order to install any of these package groups, simply append them as a comma s
 
 .. code-block:: console
 
-    $ pip install -e aiida-core[atomic_tools,docs]
+    $ pip install -e "aiida-core[atomic_tools,docs]"
 
 .. admonition:: Kerberos on Ubuntu
    :class: note title-icon-troubleshoot


ltalirz:
@unkcpz Does the rest work fine with zsh?
I guess whenever AiiDA calls a terminal command, it should specify the shell explicitly, so perhaps that solves things..

unkcpz:
For the current version aiida_core>1.0 I only find this issue on zsh. I'll report if I find more. @ltalirz 
zsh is basically compatible with bash command, so not need to explicitly specify the shell for zsh I guess. Not sure about other shell, but I guess they are not used very often outside the computer science community, true? I also propose to add a note at the beginning part of documentation about this say all shell commands are tested in bash.

chrisjsewell:
@ltalirz yeh I been using it with zsh for last ~month in my new macbook (if you didn't know zsh is now the default terminal on Macs) with mainly no issues. For the most part zsh is a strict superset of bash, except you configure with `.zshrc` instead of `bashrc`/`bash_profile`.
Quoting of `[` is a known "issue" I'd come across before on other projects, and the only other thing I just encountered last week, is that to use `eval "$(_VERDI_COMPLETE=source verdi)"` I first have to use something like `autoload -Uz compinit && compinit`.
I'm planning to add a PR about this soon.

chrisjsewell:
> Not sure about other shell, but I guess they are not used very often outside the computer science community, true? I also propose to add a note at the beginning part of documentation about this say all shell commands are tested in bash.

I'd note that in [installing-prerequisites](https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/installation.html#installing-prerequisites), it does specify that the shell should be bash or zsh

Issue #4256, opened 2020-07-11
failing builds on python 3.7


ltalirz:
We already reported in a couple of places that installing AiiDA in the latest Github action runners with python 3.7.8 fails https://github.com/actions/virtual-environments/issues/1202

The reason is that, for some reason, the downgrade from pyyaml 5.3.2 to pyyaml 5.1.2 fails in that environment.

While we've temporarily switched to the python 3.8 environments for the CI tests that run on every commit, the nightly test install runs are still failing.

There are a couple of options for working around this issue:

1) keep using python 3.7.8 but uninstall pyyaml before installing aiida-core
```
        pip uninstall -y pyyaml
```
2) use the `setup-python` action `v2` and fix the version of the python environment to 3.7.7
3) Upgrade pyyaml; see also [this comment](https://github.com/aiidateam/aiida-core/issues/3709#issuecomment-655660612)

Since 1) is a bit of a hack, and this issue isn't really our fault, I would suggest 2).
3) might still take a bit of time since this update apparently first has to happen in plympy.

@csadorf @sphuber @chrisjsewell Sounds good?

sphuber:
I would say 1 is just as much a hack as number 2. I was going to propose to just switch from Python 3.7 to 3.8 entirely, as we did for the CI workflow, but then I saw that we actually run the `test_install.yml` for all Python versions, so there we would anyway have to find a solution for 3.7.

I still have some questions though:

#. You say the problem is with uninstalling `pyyaml` on 3.7.8, but then you propose doing exactly that as a fix. Does it really work if we just uninstall before running `aiida-core`? So somehow the problem only arises when `pip` tries to uninstall it as part of a bigger install? Seems weird.
#. Do we know if this is just a problem with 3.7.8 and `pyyaml==5.1.2` on the GHA runner, or can we expect the same to happen on an actual machine that has Python 3.7.8. If so, having just a workaound defeats the purpose of having an automated test that checks the software can be installed on all Python versions.

Finally, just wanted to mention that 3 might be even more work than one would think. It is not just `plumpy` that will have to be updated, but `kiwipy` as well. However, doing so would be relatively simple, if it weren't for the fact that upgrading will break the checkpointing of processes. This is the very reason I pinned the version in the first place. It is not exactly clear what the implications are, but we have to be a bit careful.

ltalirz:
[...]

PR #4265, opened 2020-07-14
Pin Python version for test-install:pip workflow to 3.7.7.

csadorf:
As a stop-gap solution for issue #4256.

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -24,7 +24,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Set up Python 3.8
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: 3.8
 
@@ -185,7 +185,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Set up Python 3.8
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: 3.8
 
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -43,9 +43,11 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Set up Python 3.7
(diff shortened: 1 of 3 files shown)


.github/workflows/test-install.yml
```diff
@@ -43,9 +43,11 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Set up Python 3.7
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
```
sphuber, line 46: Since here we do not necessarily need Python 3.7 but just need _a_ version, why not switch to Python 3.8? We are anyway testing Python 3.7 in the `tests` job where it really matters.

Fine to keep the updated version of the `setup-python` action, but I would update it everywhere in all workflows.

csadorf: Done. 👍

csadorf: Done. 👍

sphuber:
Thanks @csadorf . There are two more uses of the `setup-python` action in the `test_install.yml` workflow. Could you please update these as well. The same goes for all occurrences in the `.github/workflows/ci.yml` and `.github/workflows/release.yml` workflows

sphuber:
Sorry to be stickler @csadorf , but as mentioned in my previous comment, there are two more uses of `setup-python@v1` in `test_install.yml` as well as in the `release.yml`. Could you please update those as well?

csadorf:
> Sorry to be stickler @csadorf , but as mentioned in my previous comment, there are two more uses of `setup-python@v1` in `test_install.yml` as well as in the `release.yml`. Could you please update those as well?

Sorry about that. However, I think we should generally avoid introducing unrelated changes into a PR (change set).

sphuber:
I agree with that in general, but don't think that these changes are unrelated. We are using an action in our CI workflows, which was outdated. In one specific case this was causing problems, so we need to update it, but I think it is perfectly justified to then also update the other usages. No need to use multiple and outdated versions

PR #4270, opened 2020-07-15
Add benchmark workflow

chrisjsewell:
As discussed earlier I want to trial this workflow in the `benchmark-test-cjs` branch.
However, after some head-scratching, it appears that the file has to be added to the default branch before it is recognised by Github Action: https://github.community/t/workflow-files-only-picked-up-from-master/16129/12
As you can see this workflow only triggers for the `benchmark-test-cjs` branch for now, so will not actually affect anything just yet

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,6 +1,5 @@
-## Run benchmarks on the three OSs
-## We run only on one python version (a recent one)
-## I think that different python versions might share hardware resources so results might have a much larger fluctuation
+# Note: this workflow is currently being tested,
+# and so currently is only activated for the benchmark-test-cjs branch 
 
 name: Performance benchmarks
 
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Install python dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -e .[tests]
+        pip install -r requirements/requirements-py-3.8.txt
     - name: Run benchmarks
       run: pytest --benchmark-only --benchmark-json output.json
     - name: Store benchmark result
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,5 +1,5 @@
 # Note: this workflow is currently being tested,
-# and so is only activated for the benchmark-test-cjs branch 
+# and so is only activated for the benchmark-test-cjs branch
 
 name: Performance benchmarks
 
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,5 +1,5 @@
 # Note: this workflow is currently being tested,
-# and so currently is only activated for the benchmark-test-cjs branch 
+# and so is only activated for the benchmark-test-cjs branch 
 
 name: Performance benchmarks
 
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,49 @@
+## Run benchmarks on the three OSs
+## We run only on one python version (a recent one)
+## I think that different python versions might share hardware resources so results might have a much larger fluctuation
+
+name: Performance benchmarks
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v1
```
csadorf, line 24:
```suggestion
      uses: actions/setup-python@v2
```
We switched all other workflows to v2.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,48 @@
...
+      with:
+        python-version: 3.8
+    - name: Install python dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -e .[tests]
```
csadorf, line 30: I think we should install pinned dependencies using the `requirements-*.txt` files for the benchmark runs. Our CI builds should be as predictable as possible. We are explicitly installing from the unpinned dependencies in the `test-install` workflow to test whether there are any issues with this, everywhere else we should use the pinned dependencies.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,48 @@
...
+      #if: "github.event_name == 'push' && github.ref == 'refs/heads/develop'"
+      uses: rhysd/github-action-benchmark@v1
+      with:
+        name: "Benchmark on ${{ matrix.os }}"
+        tool: "pytest"
+        output-file-path: output.json
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        auto-push: true
+        # Show alert with commit comment on detecting possible performance regression
+        alert-threshold: '200%'
```
csadorf, line 46: That seems a bit high, doesn't it? Are we really ok with a performance regression of that magnitude? I would recommend to warn at about 10% performance regression:
```suggestion
        alert-threshold: '110%'
```
We might need to adjust that to higher values in case that the benchmark runs are more volatile than that.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,48 @@
...
+        comment-on-alert: false
+        fail-on-alert: false
```
csadorf, line 48: So what will happen if the threshold is exceeded? It seems to me that at least `comment-on-alert` should be `true`?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install python dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -e .[tests]
```
chrisjsewell, line 30:
```suggestion
        pip install -r requirements/requirements-py-3.8.txt
```

.github/workflows/benchmark.yml
chrisjsewell: Well for now it is turned off, until I have had a chance to assess some initial data as to how variable the data is.
Then the intention is that when I make a follow up PR, with some benchmarking tests implemented, I will set one/both to true with an informed threshold.

.github/workflows/benchmark.yml
chrisjsewell: makes sense thanks, now changed

chrisjsewell:
> And just to double check, this workflow did not run yet on this PR, did it? (just in case I missed it somehow)

Thanks @csadorf, no as mentioned currently it is only configured to trigger on pushes to the `benchmark-test-cjs` branch. Then once I have implemented and tested a number of benchmarking tests, I will make a follow-up PR and change this configuration

.github/workflows/benchmark.yml
csadorf: Sounds like a good plan. 👍

Issue #2978, opened 2019-06-07
Exhaustive printing by remote frontend


pzarabadip:
I am trying to configure and use a remote front end in our HPC facility which prints exhaustive log to the screen upon connection. We tried to use the solution provided under the issue #1890 but apparently putting either of guards (both in .bashrc and .bash_profile) does not solve the issue. 

The non-interactive shell by executing `ssh <host> " bash --login -c 'ls' " `still prints out the welcome message:
```
                   __                             _                       cesnet
 __ __ ___    ___ |  |_  ___ _   ___  ___  _ __  | |_  _ __  _   _  _ __ ___
|  \_ ` _ \ /  _ \|  __|/  _` | / __|/ _ \| \_ \ | __|| \__|| | | || \_ ` _ \
|  | | | | |   __/|  |_|  (_| || (__|  __/| | | || |_ | |   | |_| || | | | | |
|__| |_| |_|\____| \___|\___,_| \___|\___||_| |_| \__||_|    \__,_||_| |_| |_|
 _   _       _   _       _
|_| |_|  _  |_| |_|  _  |_|
        |_|         |_|


 Your Home Directories
 ----------------------
  Size  Avail  S.Quota   Used Directory
  175T   142T       1G    48k /storage/brno11-elixir/home/pezhman
  1.8P   789T     none    46G /storage/brno1-cerit/home/pezhman
  110T    37T      31G    18G /storage/brno2/home/pezhman
  932T   102T     none    38G /storage/brno3-cerit/home/pezhman
  262T   100T     209G    46G /storage/brno6/home/pezhman
   88T    32T     102G   232k /storage/brno8/home/pezhman
  262T   111T       3T   897G /storage/brno9-ceitec/home/pezhman
   60T    26T     102G   204k /storage/budejovice1/home/pezhman
   30T    25T      10G    64k /storage/liberec3-tul/home/pezhman
  352T    60T       3T   160k /storage/plzen1/home/pezhman
  100T    37T     524G   224k /storage/praha1/home/pezhman
  142T    90T       1G    68k /storage/praha5-elixir/home/pezhman

 CPUs
 -----
  Free   Used    Total        Centre
  1021   4469     5490        CERIT
   873   8101     8974        META


 Contact
 --------
  E-mail:	meta@cesnet.cz
  Web:		http://www.metacentrum.cz/

[...]

PR #4271, opened 2020-07-16
`verdi computer` option to switch off login shell

unkcpz:
Issue closing https://github.com/aiidateam/aiida-core/issues/2978 happened to me when connecting to [TianHe-2](https://en.wikipedia.org/wiki/Tianhe-2) . The messages output to the login shell was set in file `/etc/profile`. I make a rough fix and let it set in `verdi computer configuration ssh` . 

EDIT:
local transport plugin should also configurable with on/off login shell. As well as `gotocomputer_command` of all transport plugins. To avoid repeating define `bash_command` and repeating gotocomputer command string, encapsulating them in `Tranport` class.

### DIFF
--- a/aiida/transports/plugins/ssh.py
+++ b/aiida/transports/plugins/ssh.py
@@ -311,13 +311,6 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         return 'True'
 
     @classmethod
-    def _get_use_login_shell_suggestion_string(cls, computer):  # pylint: disable=unused-argument
-        """
-        Return a suggestion for the specific field.
-        """
-        return 'True'
-
-    @classmethod
     def _get_load_system_host_keys_suggestion_string(cls, computer):  # pylint: disable=unused-argument
         """
         Return a suggestion for the specific field.
@@ -368,8 +361,6 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         Initialize the SshTransport class.
 
         :param machine: the machine to connect to
-        :param use_login_shell: (optional, default True)
-           if False, do not use a login shell when executing command
         :param load_system_host_keys: (optional, default False)
            if False, do not load the system host keys
         :param key_policy: (optional, default = paramiko.RejectPolicy())
@@ -1241,10 +1232,7 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
 
         # Note: The default shell will eat one level of escaping, while
         # 'bash -l -c ...' will eat another. Thus, we need to escape again.
-        if self._use_login_shell:
-            bash_commmand = 'bash -l -c '
-        else:
-            bash_commmand = 'bash -c '
+        bash_commmand = self._bash_command_str + '-c '
 
         channel.exec_command(bash_commmand + escape_for_bash(command_to_execute))
 
@@ -1312,21 +1300,14 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
             further_params.append('-i {}'.format(escape_for_bash(self._connect_args['key_filename'])))
 
         further_params_str = ' '.join(further_params)
-        # I use triple strings because I both have single and double quotes, but I still want everything in
-        # a single line
-        connect_string = (
-            """ssh -t {machine} {further_params} "if [ -d {escaped_remotedir} ] ;"""
-            """ then cd {escaped_remotedir} ; bash -l ; else echo '  ** The directory' ; """
-            """echo '  ** {remotedir}' ; echo '  ** seems to have been deleted, I logout...' ; fi" """.format(
-                further_params=further_params_str,
-                machine=self._machine,
-                escaped_remotedir="'{}'".format(remotedir),
-                remotedir=remotedir
-            )
-        )
 
-        # print connect_string
-        return connect_string
+        connect_string = self._gotocomputer_string(remotedir)
+        cmd = 'ssh -t {machine} {further_params} {connect_string}'.format(
+            further_params=further_params_str,
+            machine=self._machine,
+            connect_string=connect_string,
+        )
+        return cmd
 
     def symlink(self, remotesource, remotedestination):
         """
--- a/aiida/transports/plugins/ssh.py
+++ b/aiida/transports/plugins/ssh.py
@@ -311,6 +311,13 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         return 'True'
 
     @classmethod
+    def _get_use_login_shell_suggestion_string(cls, computer):  # pylint: disable=unused-argument
+        """
+        Return a suggestion for the specific field.
+        """
+        return 'True'
+
+    @classmethod
     def _get_load_system_host_keys_suggestion_string(cls, computer):  # pylint: disable=unused-argument
         """
         Return a suggestion for the specific field.
@@ -361,6 +368,8 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         Initialize the SshTransport class.
 
         :param machine: the machine to connect to
+        :param use_login_shell: (optional, default True)
+           if False, do not use a login shell when executing command
         :param load_system_host_keys: (optional, default False)
            if False, do not load the system host keys
         :param key_policy: (optional, default = paramiko.RejectPolicy())
@@ -1232,7 +1241,12 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
 
         # Note: The default shell will eat one level of escaping, while
         # 'bash -l -c ...' will eat another. Thus, we need to escape again.
-        channel.exec_command('bash -l -c ' + escape_for_bash(command_to_execute))
+        if self._use_login_shell:
+            bash_commmand = 'bash -l -c '
+        else:
+            bash_commmand = 'bash -c '
+
+        channel.exec_command(bash_commmand + escape_for_bash(command_to_execute))
 
         stdin = channel.makefile('wb', bufsize)
         stdout = channel.makefile('rb', bufsize)
--- a/aiida/transports/transport.py
+++ b/aiida/transports/transport.py
@@ -77,10 +77,20 @@ class Transport(abc.ABC):
     def __init__(self, *args, **kwargs):  # pylint: disable=unused-argument
         """
         __init__ method of the Transport base class.
+
+        :param safe_interval: (optional, default self._DEFAULT_SAFE_OPEN_INTERVAL)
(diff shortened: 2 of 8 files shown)


aiida/transports/plugins/ssh.py
```diff
@@ -361,6 +377,8 @@ def __init__(self, *args, **kwargs):
         Initialize the SshTransport class.
 
         :param machine: the machine to connect to
+        :param use_login_shell: (optional, default False)
```
ltalirz, line 380:
```suggestion
        :param use_login_shell: (optional, default True)
```

aiida/transports/plugins/ssh.py
```diff
@@ -380,6 +398,7 @@ def __init__(self, *args, **kwargs):
         self._machine = kwargs.pop('machine')
 
         self._client = paramiko.SSHClient()
+        self._use_login_shell = kwargs.pop('use_login_shell', False)
```
ltalirz, line 401:
```suggestion
        self._use_login_shell = kwargs.pop('use_login_shell', True)
```

aiida/transports/plugins/ssh.py
ltalirz: I guess you don't want to change the default, right?

aiida/transports/plugins/ssh.py
```diff
@@ -184,6 +184,15 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
     # to return a suggestion; it must accept only one parameter, being a Computer
     # instance
     _valid_auth_options = _valid_connect_options + [
+        (
+            'use_login_shell', {
+                'default': True,
+                'switch': True,
+                'prompt': 'Use login shell',
+                'help': 'Whether use login shell when executing command.',
```
ltalirz, line 192:
```suggestion
                'help': 'Whether use login shell when executing command. Avoiding a login shell can help suppress spurious text output but may result in some startup files (.profile) not being sourced.',
```

aiida/transports/plugins/ssh.py
unkcpz: Sure, a copy-paste mistake.

aiida/transports/plugins/ssh.py
```diff
@@ -184,6 +184,16 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
     # to return a suggestion; it must accept only one parameter, being a Computer
     # instance
     _valid_auth_options = _valid_connect_options + [
+        (
+            'use_login_shell', {
+                'default': True,
+                'switch': True,
+                'prompt': 'Use login shell when executing command',
+                'help': ' Avoiding a login shell can help suppress spurious text output'
```
sphuber, line 192:
```suggestion
                'help': 'Not using a login shell can help suppress any potential spurious text output that can interfere with AiiDA executing commands, '
```
I guess it might not be immediately clear why suppressing spurious output is something desirable, per se.

aiida/transports/plugins/ssh.py
```diff
@@ -361,6 +378,8 @@ def __init__(self, *args, **kwargs):
         Initialize the SshTransport class.
 
         :param machine: the machine to connect to
+        :param use_login_shell: (optional, default True)
+           if False, do not use login shell to execute command
```
sphuber, line 382:
```suggestion
           if False, do not use a login shell when executing commands
```

sphuber:
I just realized that this can also apply to `local` transports, although more unlikely probably. However, for completeness, should we also add this switch there?

And finally, if I understand correctly, this would then solve #2978 ? Actually, that made me realize that probably we will want to add a small note about this new switch in [this part of the documentation troubleshooting](https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/troubleshooting.html?highlight=spurious#output-from-bashrc-and-or-bash-profile-on-remote-computers).
I would propose to add something like the following at the end of that section:

```
.. note::

    If, for whatever reason, the methods explained above do not work, you can also opt to change the computer configuration to not use a login shell.
    During `verdi computer configure`, set `-no-use-login-shell` or when asked to use a login shell, set it to `False`.
    This will tell AiiDA to not use a login shell when connecting, which may potentially prevent the spurious output from being printed.
    Note, however, that this may also prevent other code from being executed, such as certain environment variables being set.
```

ltalirz:
It's true that this does make a difference for local transports as well.
A practical use case might be if someone wants to install AiiDA directly on the login node of the "communicative" cluster.

unkcpz:
> A practical use case might be if someone wants to install AiiDA directly on the login node of the "communicative" cluster.

emm.. that's true. So, put this option to setup?

giovannipizzi:
Thanks @unkcpz !
Indeed, I think it's ok to add this as an option keeping the old default - it might create other problems, but being optional it's ok. Also yes, I think it fixes #2978?

A few final points:
- Is this automatically configurable now with `verdi computer configure ssh`? Can we have a test maybe?
- Indeed, it would be good to have the same logic for `local` transports

giovannipizzi:
I think `configure` is still the right place instead of `setup`, but done for both backends. I had tried an attempt to simplify how the logic is managed in #3663 but it turned out to be more complex than I thought and I didn't have the time to finish that one. Maybe for now we just follow the same thing that is done for the cool down time that is defined for both plugins.

unkcpz:
ok, I'll also add it to local transports. 

> Is this automatically configurable now with verdi computer configure ssh? Can we have a test maybe?

I add a simple test to check that autoinfo getting set with this option. @giovannipizzi 
https://github.com/aiidateam/aiida-core/blob/91d1e845878816acc8c2e1e77d94bf8bae5d706a/tests/cmdline/commands/test_computer.py#L414

unkcpz:
The `use_login_shell` option is now added to local transports. 

- To avoid repeating this setting code in two files, I add it to abstract `Transport` class will take effect for all transports plugins. 
- Also add test to make sure computer's auto info is set in `test_local_interactive`.
- DOCS: note is added in troubleshooting part as @sphuber suggested.

unkcpz:
@ltalirz I find one more place where login shell is used, here at `gotocomputer_command`. https://github.com/aiidateam/aiida-core/blob/0d7baa5bfef430c1d8364a375219d863b51463ff/aiida/transports/plugins/ssh.py#L1305

why force user to bash login shell when they goto the specific computer, is that necessary?

ltalirz:
> why force user to bash login shell when they goto the specific computer, is that necessary?

I think it's the expected behavior - a user will expect the same environment variables as when connecting manually and changing to that directory, no?

>  If there are spurious output being printed the login shell here will also cause problem.

If it causes problems, then we may need to make it configurable here as well, but I'm not aware that AiiDA is processing the output of this command.
Can you check whether it indeed causes a problem (and why)?

aiida/transports/transport.py
```diff
@@ -47,21 +47,40 @@ class Transport(abc.ABC):
...
+        (
+            'use_login_shell', {
+                'default':
+                True,
+                'switch':
+                True,
+                'prompt':
+                'Use login shell when executing command',
+                'help':
+                ' Not using a login shell can help suppress any potential'
```
ltalirz, line 60:
```suggestion
                ' Not using a login shell can help suppress potential'
```

aiida/transports/transport.py
```diff
@@ -47,21 +47,40 @@ class Transport(abc.ABC):
...
+                ' spurious text output that can interfere with AiiDA executing commands,'
```
ltalirz, line 61:
```suggestion
                ' spurious text output that can prevent AiiDA from parsing the output of commands,'
```

docs/source/intro/troubleshooting.rst
```diff
@@ -269,6 +269,13 @@ To test if a the computer does not produce spurious output, run (after configuri
 
 which checks and, in case of problems, suggests how to solve the problem.
 
+.. note::
+
+    If, for whatever reason, the methods explained above do not work, you can also opt to change the computer configuration to not use a login shell.
+    During `verdi computer configure`, set `-no-use-login-shell` or when asked to use a login shell, set it to `False`.
+    This will tell AiiDA to not use a login shell when connecting, which may potentially prevent the spurious output from being printed.
+    Note, however, that this may also prevent other code from being executed, such as certain environment variables being set.
```
ltalirz, line 277:
```suggestion
    If the methods explained above do not work, you can configure AiiDA to not use a login shell when connecting to your computer, which may prevent the spurious output from being printed:
    During ``verdi computer configure``, set ``-no-use-login-shell`` or when asked to use a login shell, set it to ``False``.
    Note, however, that this may result in a slightly different environment, since `certain startup files are only sourced for login shells <https://unix.stackexchange.com/a/46856/155909>`_.
```

unkcpz:
Ah, the problem comes from the `bashrc` config of remote cluster, it somehow execute `cd` to the home directory after login. So the `gotocomputer command` should keep it the way it is.

unkcpz:
@ltalirz thanks! looks better now.

ltalirz:
> Ah, the problem comes from the bashrc config of remote cluster, it somehow execute cd to the home directory after login. So the gotocomputer command should keep it the way it is.

Just so that I understand: some system-level bashrc file is being sourced only for login shells, and this breaks the `gotocomputer` command? Strange...

In any case, reading back through what I wrote, I think you were right after all.
It may be the "expected behavior" by default, but for people who explicitly configured AiiDA not to use a login shell, it is probably rather unexpected that AiiDA would use continue to use a login shell for `gotocomputer`.

Given the issue you encountered, I think it's best to make it configurable here as we..
Since this switching logic will now be present in three places, I suggest you encapsulate it, e.g. in a `_bash_command` property of the `Transport`.

unkcpz:
I am not very familiar with login/non-login and interactive/non-interactive shell stuff, correct me if I understand it wrong. 

> Just so that I understand: some system-level bashrc file is being sourced only for login shells, and this breaks the gotocomputer command? Strange...

My fault for not clarifying this, not the system-level bashrc but the user's `~/.bashrc` which did the `cd` action in the remote cluster. Using `gotocomputer` will always activate the `~/.bashrc` no matter login shell is on or off, correct? So the problem is definitely not from AiiDA side, and can be solved by modifying the ~/.bashrc file which user always have permission to edit. 

 > It may be the "expected behavior" by default, but for people who explicitly configured AiiDA not to use a login shell, it is probably rather unexpected that AiiDA would use continue using a login shell for gotocomputer.

Agree! If the `cd` action is setting in the system wide profile which users have no permission to edit, the weird `cd` action should be suppressed by turn off the login shell. That is when the login shell is set to false, we also expect gotocomputer is non-login shell.

PR #4272, opened 2020-07-16
DOCS: stop rabbitmq service before deactivate

unkcpz:
When rabbitmq and postgresql are both installed by conda in a isolated virtual environment, it suppose to stop rabbitmq service when deactivate from the environment.


unkcpz:
Oh, hard to create and modify pr with github web API:cry:, will close this first.

PR #4275, opened 2020-07-20
Fix bug in `aiida.engine.daemon.execmanager.retrieve_files_from_list`

### LINKED ISSUE #4273: Bug in `aiida.engine.daemon.execmanager.retrieve_files_from_list`
In the main loop of the function the `local_names` and `remote_names` are set, however, if the remote path does *not* contain globbing characters, the code goes in a branch where `remote_names` is not redefined. This means the definition of the previous loop iteration will be used erroneously.

sphuber:
Fixes #4273 

The `retrieve_files_from_list` would loop over the instructions of the
`retrieve_list` attribute of a calculation job and for each entry define
the variables `remote_names` and `local_names` which contain the
filenames of the remote files that are to be retrieved and with what
name locally.

However, for the code path where the element of `retrieve_list` is a
list or tuple and the first element contains no wildcard characters, the
`remote_names` variable is not defined, meaning that the value of the
previous iteration would be used. This was never detected because this
code path was not actually tested.

This bug would only affect `CalcJob`s that specified a `retrieve_list`
that contained an entry of the form:

    ['some/path', 'some/path', 0]

where the entry is a list and the first element does not contain a
wildcard.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -546,12 +546,12 @@ def retrieve_files_from_list(calculation, transport, folder, retrieve_list):
     treated as the work directory of the folder and the depth integer determines
     upto what level of the original remotepath nesting the files will be copied.
 
-    :param transport: the Transport instance
-    :param folder: an absolute path to a folder to copy files in
-    :param retrieve_list: the list of files to retrieve
+    :param transport: the Transport instance.
+    :param folder: an absolute path to a folder that contains the files to copy.
+    :param retrieve_list: the list of files to retrieve.
     """
     for item in retrieve_list:
-        if isinstance(item, list):
+        if isinstance(item, (list, tuple)):
             tmp_rname, tmp_lname, depth = item
             # if there are more than one file I do something differently
             if transport.has_magic(tmp_rname):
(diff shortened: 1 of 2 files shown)


ramirezfranciscof:
Since this creates the tests for this function, shouldn't we maybe add checks to make sure that you can (1) change the name of the file/folder while copying and (2) create dirs in the target directory (i.e. `['filename.txt', 'new/path/filename.txt', 0]`)? These seem like relevant cases that could break without noticing. Other than that I don't have many more comments, so let me know what you think of this and I'll aprove / wait for the changes and then aprove.

sphuber:
> Since this creates the tests for this function, shouldn't we maybe add checks to make sure that you can (1) change the name of the file/folder while copying and (2) create dirs in the target directory (i.e. `['filename.txt', 'new/path/filename.txt', 0]`)? These seem like relevant cases that could break without noticing. Other than that I don't have many more comments, so let me know what you think of this and I'll aprove / wait for the changes and then aprove.

It's true that I only tested a small percentage of the pathways of this code, specifically the part that concerned the bug I am fixing there. Parts of it are tested indirectly through the integration tests that actually run some entire `CalcJob`s. There is no particular reason I did not add more tests other than lack of time, but I will see if I can add some more to test the other pathways as well.

ramirezfranciscof:
Yes, I agree it might not make sense to use this bugfix PR to try design the perfect coverage test for this function. I just thougth these two extra tests were particularly related to the functionality being fixed and might be worth including. Anyways, its up to you, I will aprove the PR so you can merge it if you prefer to leave these for later (or you can ping me to take a look if you decide to include them).

sphuber:
Thanks @ramirezfranciscof . I think I will leave it as is for now. If another bug crops up, the test can be expanded

PR #4290, opened 2020-08-06
ci: unpin python version

ltalirz:
The virtual environment of runners on Github Actions had an issue, where
installing a different pyyaml version than the one present resulted in
an error.
The issue has been resolved in the latest image, and so the workaround
of specifically requesting the (outdated) python version 3.7.7 can be
dropped.

See https://github.com/actions/virtual-environments/issues/1202#issuecomment-670157812

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -98,9 +98,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


csadorf:
Let's merge this after #4295 is fixed.

csadorf:
Thank you @ltalirz !

Issue #4281, opened 2020-07-25
Frequent CI failures due to unreliable test results


ltalirz:
In the last couple of weeks, we are experiencing frequent CI failures (the daily `test-install` workflow fails about every 2nd or 3rd time).

Many of these failures seem to stem from a couple of tests with unreliable outcomes, some of which might be quite easy to fix.

Let's collect a list of culprits here:
 - [ ] REST API: test_run_without_close_session (`tests/restapi/test_threaded_restapi.py:121: AssertionError`)
        CI log: https://github.com/aiidateam/aiida-core/runs/908939811#step:11:445
        Issue: the exception encountered is not the expected one
- [ ] QueryBuilder: test_process_query(`tests/orm/test_querybuilder.py:211: AssertionError`)
       CI log: https://github.com/aiidateam/aiida-core/runs/888226193#step:11:468
      Issue: more warnings encountered than expected
 - [x] Data: test_data_exporters (`tests/orm/data/test_data.py`)
        CI log: https://github.com/aiidateam/aiida-core/runs/955848737?check_suite_focus=true#step:11:608
        Issue: pika.exceptions.ChannelClosedByBroker: (404, "NOT_FOUND - no exchange 'aiida-b876a05457ef4f3f86e8e1284dffd85f.tasks' in vhost '/'") when running a `convert` function for a CifData node

Pinging @CasperWA @sphuber @csadorf for info

csadorf:
I don't think it's an option to make the test-install workflow more giving, because otherwise we might miss actual issues. The only option I see is to make sure all of our tests are deterministic.

ltalirz:
> The only option I see is to make sure all of our tests are deterministic.

Sure, that's what I'm proposing here

ltalirz:
While the two tests above seem relatively straightforward to fix (by adapting the assertion statement), I'm not sure that I'm the right person to do this - it should be the "code owners" of the REST API / QueryBuilder (I wonder who they would be?).

In the meanwhile, a mitigation strategy would be to [mark the corresponding tests as flaky](https://github.com/pytest-dev/pytest-rerunfailures#re-run-individual-failures), and let the `pytest-rerunfailures` plugin automatically re-run them up to, say, 3 times before counting them as failed.

I think, having such a mitigation strategy in place is anyhow very useful, so I will go ahead and make a PR for this.

ltalirz:
Daemon test still failing with restart:
[...]

Issue #2134, opened 2018-10-26
Randomly failing test (in the backup_script tests)


giovannipizzi:
https://travis-ci.org/aiidateam/aiida_core/jobs/446675963
```
======================================================================
FAIL: test_integration (aiida.backends.tests.backup_script.TestBackupScriptIntegration)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/aiidateam/aiida_core/aiida/backends/tests/backup_script.py", line 340, in test_integration
    "The backed-up repository has differences "
AssertionError: False is not true : The backed-up repository has differences to the original one.
----------------------------------------------------------------------
```
Any idea why this happens? Could you critically check the test code to see if there are cases in which can happen but it's not a failed test? Can you maybe also improve the output of the test to tell exactly what changed?

szoupanos:
I ran it 16 times locally without being able to reproduce the problem. With PR #2141, I improve the messages of aiida.common.utils.are_dir_trees_equal function in order to have more info the next time that it fails.

sphuber:
The error message has been changed to give more detail in PR #2141 
If you encountered another failure and was directed to this issue, please
post the error message and reopen this issue, such that the origin of the
failing test can be determined and fixed.

giovannipizzi:
This happened again: https://travis-ci.org/aiidateam/aiida_core/jobs/452782213

sphuber:
```
======================================================================
FAIL: test_integration (aiida.backends.tests.backup_script.TestBackupScriptIntegration)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/aiidateam/aiida_core/aiida/backends/tests/backup_script.py", line 341, in test_integration
    + ". If the test fails, report it in issue #2134.")
AssertionError: False is not true : The backed-up repository has differences to the original one. Left directory: /tmp/test_repository_test_sqlalchemy/repository/node, right directory: /tmp/tmp87nq149k/.aiida/backup/backup_dest/repository/node, files only in left directory: ['5d', 'cd'], files only in right directory: [], not comparable files: []. If the test fails, report it in issue #2134.
```

CasperWA:
For PR #2478.

```shell
======================================================================
FAIL: test_integration (aiida.backends.tests.test_backup_script.TestBackupScriptIntegration)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #4291, opened 2020-08-06
rerun flaky tests

ltalirz:
Over time, we've accumulated a handful of tests that pass just fine most
of the time, but fail every now and then when the bits align with the
wrong star.
While the right thing to do would be a deep dive into astrology, there's
always so many other things to do!

As a mitigation strategy, this PR marks those tests as 'flaky' and uses
the pytest-rerunfailures plugin to automatically try re-running those
tests a few times before considering them failed.

Note that this also makes it easier to identify flaky tests, a simple
`git grep mark.flaky` will do the trick.

P.S. Mentioning issues of tests that are marked as flaky: #2134 #3051 #4281

### DIFF
--- a/tests/cmdline/commands/test_daemon.py
+++ b/tests/cmdline/commands/test_daemon.py
@@ -86,7 +86,8 @@ class TestVerdiDaemon(AiidaTestCase):
         finally:
             self.daemon_client.stop_daemon(wait=True)
 
-    @pytest.mark.skip(reason='Test fails non-deterministically; see issue #3051.')
+    # Tracked in issue #3051
+    @pytest.mark.flaky(reruns=2)
     def test_daemon_start_number(self):
         """Test `verdi daemon start` with a specific number of workers."""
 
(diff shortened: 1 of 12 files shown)


ltalirz:
@csadorf I ran `./util/dependency_management.py generate-all` but nothing changed in `requirements/`.
Is there a bug in the script?

tests/cmdline/commands/test_daemon.py
```diff
@@ -86,7 +86,7 @@ def test_daemon_restart(self):
         finally:
             self.daemon_client.stop_daemon(wait=True)
 
-    @pytest.mark.skip(reason='Test fails non-deterministically; see issue #3051.')
+    @pytest.mark.flaky(reruns=2)
```
greschd, line 89: Maybe we should keep the reference to the issue number in a comment?

ltalirz: I agree - I've now added this info to all flaky tests

ltalirz:
@greschd Sorry to keep dismissing you reviews ;-) I discovered a few other skipped tests that could be unskipped.

Of course, one question that these CI tests don't answer is whether restarting the tests actually has a chance to make them work or whether some of these tests will continue to fail because the thing in the state that makes them fail does not change.
We'll know over time ;-)

greschd:
> We'll know over time ;-)

May the stars be in our favor.

ltalirz:
I'll leave this open for a day in case @sphuber would like to comment as well - otherwise I'll go ahead and merge this since we're experiencing frequent failures of the daily test runs.

PR #4294, opened 2020-08-08
Docs: fixed few typos/issues in docs

pzarabadip:
This PR fixes few issues in documentations:

1. In `docs/source/topics/calculations/usage.rst` definition of code snippets related to `include/snippets/calcjobs/arithmetic_add_parser.py` needs to be iincremented by one line for propoer rendering.

2. In `docs/source/topics/workflows/usage.rst` there is an extra `:` that needs to be deleted for proper rendering of following code blocks.


greschd:
I think the line number issue should be fixed "permanently" by using markers in the code, such as here: https://github.com/aiidateam/aiida-core/blob/develop/aiida/calculations/arithmetic/add.py#L35

In the doc, the markers are referenced by using `start-after` and `end-before`, see https://github.com/aiidateam/aiida-core/blob/develop/docs/source/howto/plugin_codes.rst#handling-parsing-errors

ltalirz:
That would of course be even better :-)

pzarabadip:
Thanks @ltalirz and @greschd 
I have not done it before but surely it is the better way to do it. As it is an urgent PR, I can give it try.

sphuber:
> The line number issue can be fixed "permanently" by using markers in the code, such as here: https://github.com/aiidateam/aiida-core/blob/develop/aiida/calculations/arithmetic/add.py#L35

The problem with this approach here, is that this snippet is shown in its entirety at the beginning of the paragraph. This snippet was created in order to present one that was a simple as possible, without any noise. @ltalirz actually created this because he thought using the actual `ArithmeticAddParser` (as I did originally when I wrote this part of the docs) was too noisy because it includes catching exceptions. Having these vague comment lines that will mark starting lines of the docs will be counterproductive I fear.

greschd:
Aah, I see... bit of a trade-off, but I agree with that assessment. 

I guess this PR is good to go then!

Stupid idea time: what if we put the actual code of the previous line into `start-after`? Does it raise a warning if it can't find that code?

pzarabadip:
So then, shall we merge this PR as it is?

ltalirz:
> Stupid idea time: what if we put the actual code of the previous line into start-after? Does it raise a warning if it can't find that code?

I guess that would work as long as that line of code does not change and is unique in the file.
Not sure whether this is "safer" than the current solution, though.

From my side, this PR can be merged.

greschd:
Yeah let's just leave it as is - @ltalirz I think you're the only one here who can do the honors.

PR #4299, opened 2020-08-12
Querybuilder init now accepts entity_type = ''

### LINKED ISSUE #4248: Issue with the queryhelp when appending a `Node`
### Describe the bug
When reloading a query from a `queryhelp`, if in one of the appends we used simply `Node` rather than a more specific class, this crashes.

### Steps to reproduce

```python
In [45]: q = QueryBuilder()                                                     

In [46]: q.append(Node, filters={'label': 'a'}, project='pk')                   
Out[46]: <aiida.orm.querybuilder.QueryBuilder at 0x15156fd867b8>

In [47]: q2 = QueryBuilder(**q.queryhelp)                                       
InputValidationError                      Traceback (most recent call last)
<ipython-input-47-e6794d35f662> in <module>
----> 1 q2 = QueryBuilder(**q.queryhelp)

~/codes/aiida-core/aiida/orm/querybuilder.py in __init__(self, backend, **kwargs)
    414         for path_spec in path:
    415             if isinstance(path_spec, dict):
--> 416                 self.append(**path_spec)
    417             # ~ except TypeError as e:
    418             elif isinstance(path_spec, str):

~/codes/aiida-core/aiida/orm/querybuilder.py in append(self, cls, entity_type, tag, filters, project, subclassing, edge_tag, edge_filters, edge_project, outerjoin, **kwargs)
    648 
    649         if not (cls or entity_type):
--> 650             raise InputValidationError('You need to specify at least a class or a entity_type')
    651 
    652         # Let's check if it is a valid class or type

InputValidationError: You need to specify at least a class or a entity_type
```

### Expected behavior

With anything else than `Node` it works:
```python
In [48]: q = QueryBuilder()                                                     

In [49]: q.append(Data, filters={'label': 'a'}, project='pk')                   
Out[49]: <aiida.orm.querybuilder.QueryBuilder at 0x15156fc92550>

In [50]: q2 = QueryBuilder(**q.queryhelp)           
```

### Your environment

AiiDA 1.3.0
### Additional context

I think the problem is that now appending a `Node` does not put an entity_type since putting a `Node` is not really restricting as a filter. So simply this should be accepted when reloading?

```
In [53]: q = QueryBuilder()                                                     

In [54]: q.append(Node, filters={'label': 'a'}, project='pk')                   
Out[54]: <aiida.orm.querybuilder.QueryBuilder at 0x15156f4a7160>

In [55]: q.queryhelp                                                            
Out[55]: 
{'path': [{'entity_type': '',
   'tag': 'node_1',
   'joining_keyword': None,
   'joining_value': None,
   'outerjoin': False,
   'edge_tag': None}],
 'filters': {'node_1': {'node_type': {'like': '%'}, 'label': 'a'}},
 'project': {'node_1': [{'pk': {}}]},
 'order_by': {},
 'limit': None,
 'offset': None}
```

ramirezfranciscof:
There was a bug when using the queryhelp of a previously defined
QueryBuilder which was appended a single query for the generic "Node"
class as a template in the initialization for a new query.

The underlying problem was that the append method was doing the wrong
checks to detect missing cls/entity_type.

Also added a test for this new specific case.

Fixes #4248


aiida/orm/querybuilder.py
```diff
@@ -658,12 +658,15 @@ def append(
         # input now.
         # First of all, let's make sure the specified
         # the class or the type (not both)
-        if cls and entity_type:
+
+        if (cls is not None) and (entity_type is not None):
+            #if cls and entity_type:
```
sphuber, line 663:
```suggestion
```

aiida/orm/querybuilder.py
```diff
@@ -658,12 +658,15 @@ def append(
...
             raise InputValidationError(
                 'You cannot specify both a class ({}) and a entity_type ({})'.format(cls, entity_type)
             )
 
-        if not (cls or entity_type):
+        if (cls is None) and (entity_type is None):
+            #if not (cls or entity_type):
```
sphuber, line 669:
```suggestion
```

aiida/orm/querybuilder.py
```diff
@@ -675,7 +678,8 @@ def append(
             else:
                 if not inspect_isclass(cls):
                     raise InputValidationError("{} was passed with kw 'cls', but is not a class".format(cls))
-        elif entity_type:
+        elif entity_type is not None:
+            #elif entity_type:
```
sphuber, line 682:
```suggestion
```

aiida/orm/querybuilder.py
```diff
@@ -658,12 +658,15 @@ def append(
         # input now.
         # First of all, let's make sure the specified
         # the class or the type (not both)
-        if cls and entity_type:
+
+        if (cls is not None) and (entity_type is not None):
```
sphuber, line 662:
```suggestion
        if cls is not None and entity_type is not None:
```
the parentheses are not needed here and for such simple conditionals do not really contribute to clarity either.

aiida/orm/querybuilder.py
```diff
@@ -658,12 +658,15 @@ def append(
...
+            #if cls and entity_type:
             raise InputValidationError(
                 'You cannot specify both a class ({}) and a entity_type ({})'.format(cls, entity_type)
             )
 
-        if not (cls or entity_type):
+        if (cls is None) and (entity_type is None):
```
sphuber, line 668:
```suggestion
        if cls is None and entity_type is None:
```

tests/orm/test_querybuilder.py
```diff
@@ -740,7 +740,11 @@ def test_queryhelp(self):
         self.assertEqual(qb.count(), 1)
 
     def test_recreate_from_queryhelp(self):
-        """Test recreating a QueryBuilder from the Query Help"""
+        """
+        Test recreating a QueryBuilder from the Query Help
```
sphuber, line 744:
```suggestion
        """Test recreating a QueryBuilder from the Query Help.
        
```

tests/orm/test_querybuilder.py
```diff
@@ -753,6 +757,11 @@ def test_recreate_from_queryhelp(self):
         qb3 = copy.deepcopy(qb1)
         self.assertDictEqual(qb1.queryhelp, qb3.queryhelp)
 
+        qb1 = orm.QueryBuilder()
+        qb1.append(orm.Node)
+        qb2 = orm.QueryBuilder(**qb1.queryhelp)
+        self.assertDictEqual(qb1.queryhelp, qb2.queryhelp)
```
sphuber, line 763: Might as well just add an `append(Node)` statement to the very first querybuilder. Just add a comment that the `Node` one is important as it will indirectly test that `append` accepts `entity_type=''`

ramirezfranciscof: Yeah, you are right. Since the original example was with the single append of `Node`, I was unsure that maybe when more appends where included then the internal logic may take a different path and not manifest this behavior, but I just checked this doesn't make a difference.

tests/orm/test_querybuilder.py
ramirezfranciscof: Ok, I will change this, but I thought that the convention was to use

````
"""Single line header"""

---

"""
Multiline
Header
"""
````

Did I misunderstand this?

**Comment by ramirezfranciscof** [comment]
@sphuber very sorry for the commented leftovers, I had to take a break between making it work and preparing the PR and in the middle I forgot to take those out. Now it should be ready, I will re-request your final review once all the CI tests pass.

File: tests/orm/test_querybuilder.py (hunk 6 above)
Comment (sphuber) [reply tests/orm/test_querybuilder.py]: Yes, the first line of a docstring should always stay on a single line, be a single short description in the imperative mood and end with a period. If just a single line, the closing quotes are also on the same line. For multiline string, an empty line is created after the first subject line and the closing quotes are on a separate line. Examples:
```
"""Do something with the given inputs."""
```
Multiline
```
"""Do something with the given inputs.

This function does something particular that you may not expect.
"""
```
These rules are following the [PEP 257](https://www.python.org/dev/peps/pep-0257/) conventions.

PR #4307, opened 2020-08-15
explicitly close runner loop of function process

### LINKED ISSUE #2386: Too many open files error (file leak when creating runners/communicators)
When running tests, I get `[Errno 24] Too many open files` on Mac. 
This can be temporarily worked around using `ulimit -n 1024` to increase the limit to 1024 (as in linux) instead of the Mac OS X limit of 256. However, this pointed out a leak in the current code.

By a bit of inspection (printing the number of open FDs with `psutil.Process().num_fds()`) I realised that the source of the problem can be tracked down as follows:
- call to `run_get_node` (in process_function.py)
- this calls `self.create_runner(with_persistence=False)` 
  - internally, this goes down to kiwipi to the `RmqThreadCommunicator.connect` class method, that creates a new loop as this is not passed as a kwargs: `loop = loop or ioloop.IOLoop()` (this opens 3 FDs)
  - the following call to `loop.run_sync(lambda: connection_factory(**connection_params))` in the same method opens one more FD
  - when returning a runner (`return runners.Runner(**settings)` in `create_runner()`, file `manager.py`, as we don't pass a loop, a new loop is created in the __init__ of `Runner`: 3 more FDs
- occasionally, `proc = process_class(inputs=inputs, runner=runner)` inside `run_get_node` will clean up many processes - but this does not seem to be consistent, only happen in 1 of the 8 tests of `db.work.calcfunctions`, not clear why... also in other places of the code sometimes some FDs disappear but it's not clear why.
- the call to `runner.close()` in `run_get_node()` does not clean up any of these open FDs (only one eventually gets cleaned up of all the open FDs)

I think this might be due to the default creation of loops when entering functions such as `.connect()`, `create_runner`, or in the communicator, that are never properly stopped and closed.

I imagine this might even occur in production runs and not only on the test environment, even if I didn't debug this (neither I debugged if this happens also on Linux and not only on Mac).

Note: #1923 maybe was related to this?

### LINKED ISSUE #2895: Too many files open during tutorial
Reporting here an issue encountered during the May tutorial.

A user had two processes open (`verdi restapi` and a jupyter notebook).

- While trying to get data from the TCOD, for some reason the `rabbitmq` server died and we started getting errors from tornado saying the connection was closed (we tried to run a few commands)
- we restarted rabbitmq with `sudo /etc/init.d/rabbitmq-server restart`
- after this, in the jupyter notebook, we started getting `too many files open`
- we fixed this by restarting the jupyter kernel

While I think the death of rabbitmq is independent of AiiDA, I'm reporting this here because next time we should check why there was a "leak" of open files.

My bet: the rabbitmq server was not up, and at every command that tried to connect to it, one (or more) files were opened, hitting the limit. But we couldn't test this. Next time we could try to save a dump of `sudo lsof` or some similar command to see which files are open.

unkcpz:
Fixes #2386 and fixes #2895

This is the tentative fix of issue https://github.com/aiidateam/aiida-core/issues/2386. See the issue for more information.

### DIFF
--- a/aiida/engine/runners.py
+++ b/aiida/engine/runners.py
@@ -60,7 +60,14 @@ class Runner:  # pylint: disable=too-many-public-methods
         assert not (rmq_submit and persister is None), \
             'Must supply a persister if you want to submit using communicator'
 
-        self._loop = loop if loop is not None else tornado.ioloop.IOLoop()
+        # Runner take responsibility to clear up loop only if the loop was created by Runner
+        self._do_close_loop = False
+        if loop is not None:
+            self._loop = loop
+        else:
+            self._loop = tornado.ioloop.IOLoop()
+            self._do_close_loop = True
+
         self._poll_interval = poll_interval
         self._rmq_submit = rmq_submit
         self._transport = transports.TransportQueue(self._loop)
@@ -150,6 +157,8 @@ class Runner:  # pylint: disable=too-many-public-methods
         """Close the runner by stopping the loop."""
         assert not self._closed
         self.stop()
+        if self._do_close_loop:
+            self._loop.close()
         self._closed = True
 
     def instantiate_process(self, process, *args, **inputs):


ltalirz:
Thanks @unkcpz ! 
Can the others confirm that this is the proper way to address #2386 ?

sphuber:
> Can the others confirm that this is the proper way to address #2386 ?

Well, this could certainly be one of potential file leaks. However, it is impossible to say whether this was the only one. Could you or Daniele check out this branch (or apply the fix) and run the test again to see if now the file handles remain more or less constant?

aiida/engine/processes/functions.py
```diff
@@ -156,6 +156,7 @@ def kill_process(_num, _frame):
                 if original_handler:
                     signal.signal(signal.SIGINT, original_handler)
                 runner.close()
+                runner.loop.close()
```
sphuber, line 159: Looking at the `close` method of the `Runner`, I see that it calls `stop`:
https://github.com/aiidateam/aiida-core/blob/f2d1e9492ee33304fd197b0c8788896f76abb17d/aiida/engine/runners.py#L152
Which in turn calls `stop` on its `loop`:
https://github.com/aiidateam/aiida-core/blob/f2d1e9492ee33304fd197b0c8788896f76abb17d/aiida/engine/runners.py#L142
However, apparently it does not call `close` on the `loop`. Is it known that closing a loop is necessary to free all resources, including open file handles?

The question arises why `Runner.close` should not also call `loop.close()`? My guess is because the `loop` could be shared since it can be passed into the constructor of the `Runner`. Would it make sense though to have `Runner.close()` also close the loop if and only if the loop was constructed by the runner itself? @muhrin what was the original design here? Would this difference in behavior I propose be more confusing in the end?

ltalirz:
> Could you or Daniele check out this branch (or apply the fix) and run the test again to see if now the file handles remain more or less constant?

Sure, I can confirm that the basic file descriptor leak when running WorkChains is fixed (as @unkcpz mentioned).

My question was more whether closing the loop is the right thing to do since I vaguely remembered some discussion on a related question on slack - if it is, then please go ahead with this change!

As for releasing this, it seems to me this warrants a bug fix release.

sphuber:
>  (as @unkcpz mentioned).
> 
> My question was more whether closing the loop is the right thing to do since I vaguely remembered some discussion on a related question on slack

I understand that that was what you were asking, and I realize that @unkcpz checked that this fixes _a_ file leak, but not specifically the one that you experienced. There could be more sources of leaks and without repeating the test that showed the initial leak, that is impossible to say if this PR fixes it.

> As for releasing this, it seems to me this warrants a bug fix release.

I agree, and if we confirm that we see no more leaks in normal usage, I will make a patch release.

aiida/engine/processes/functions.py
muhrin: I've had a chance to look through this now and here are my thoughts.

* Yes, it is expected that `close()` frees file descriptors, there's even a parameter to explicitly free all those created from the scope of the loop by client code (https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.close).
* I agree that the `Runner` should take responsibility for cleaning it up if it created it.  Personally, I would find that _less_ confusing.  Just drop a flag into `Runner.__init__` that keeps track.

One thing that will have to be born in mind is that this will have to change once @unkcpz 's work gets merged as there will only be one loop (used in a re-entrant manner).  For this, I suggest we keep track of how many files are open before and after the `run_get_node` execution so at least we know if it's leaking.

ltalirz:
@unkcpz Is it clear to you what @muhrin is asking?
It would be great to get this PR merged since it fixes a major issue in 1.3.0 and we should probably make a separate patch release before moving to asyncio.

sphuber:
To summarize, all that remains to be done before we merge this: close the loop if and only if the loop was created by the runner itself in its constructor. If it was passed in as an argument, we do not claim ownership and will not close it.

unkcpz:
Sorry I didn't read Martin's reply carefully. I thought we were waiting for asyncio version to fix the issue.
I push the changes, is this the expected implementation?

sphuber:
@unkcpz implementation looks good to me. The idea is to release this as a hotfix in `v1.3.1`. I am just thinking of how to approach this since current `develop` also contains quite some new features which should really go with `v1.4.0`, however, this we cannot release right now as we are waiting for some additional features to be finalized. Tentative timeline for `v1.4.0` is in two/three weeks.

sphuber (approved):
Thanks @unkcpz

PR #4308, opened 2020-08-18
Bump base aiida-prerequisites docker image to v0.2.0

yakutovicha:


yakutovicha:
> Why is the change needed? Does `latest` not already point to `0.2.0` given that that is the latest?

This is for consistency. We would like to pin the concrete version of the base image rather than just pointing to some random commit in the `develop` branch (to which `latest` build corresponds to).

PR #4310, opened 2020-08-19
Cleanup of BackendNode/ModelEntities internals

### LINKED ISSUE #2495: Implement remaining abstract methods of the `BackendNode` interface class
In PR #2481 the `Node` class was moved to the new backend interface. For this a new abstract interface `BackendNode` was designed. Especially the interface to retrieve and set attributes/extras was redesigned. Existing functionality was migrated but the interface also defined some new methods that have not yet been implemented. These should be implemented and tested independent of the front end ORM.

ramirezfranciscof:
Set up BackendNode as a metaclass after checking that all methods were
implemented in the child classes.

Removed the duplicated pk methods from the specific classes of model
entities (SqlaModelEntity and DjangoModelEntity), since it is already
implemented in the class BackendEntity (from which nodes also inherit
through the intermediate class BackendNode).

Fix #2495 (or at least as far as we could interpret what the issue referred to)

### DIFF
--- a/aiida/orm/implementation/django/entities.py
+++ b/aiida/orm/implementation/django/entities.py
@@ -77,15 +77,6 @@ class DjangoModelEntity(typing.Generic[ModelType]):
         return self._dbmodel.pk
 
     @property
-    def pk(self):
-        """
-        Get the principal key for this entry
-
-        :return: the principal key
-        """
-        return self._dbmodel.id
-
-    @property
     def is_stored(self):
         """
         Is this entity stored?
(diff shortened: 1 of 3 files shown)


sphuber:
@ramirezfranciscof the way you have written your PR message, the linked issue won't be automatically closed. You need to prefix the issue with any of the words "fixes", "fix", "close" or "closes" (and there might be some others that work). If you have the correct syntax, the keyword will be underlined to hint that Github detected the additional meaning. I advise you to just always put the following at the top of the PR message:

Fixes #XXXX

Here the rest of the message....

ramirezfranciscof:
Yes, there, thanks for pointing it out.

Yeah, I was going to add that and also a comment that we may want to re-thing the distribution of general methods in the parent classes and more specific methods in the child classes, I was just going to do that once this was closed at the same time I added it to the next milestone. I can do all those things now I guess.

sphuber (approved):
Good to go, thanks @ramirezfranciscof

PR #4316, opened 2020-08-20
`verdi computer test`: fix bug in spurious output test

### LINKED ISSUE #4314: Exception in `verdi computer test` if spurious output in stderr

sphuber:
Fixes #4314 

The test that checks for spurious output when executing a normal command
on a computer over the transport had a bug in it that went unnoticed
because the code path was not tested. If the `stderr` of the command
contained any output the command would raise because the test that is
called `_computer_test_no_unexpected_output` would incorrectly return a
tuple of length one instead of two in that case.

In addition to adding tests to hit this code path, the message that is
printed in the case of non-empty stdout or stderr is deduplicated and
adapted to be bit clearer and refer directly to the documentation
instead of through a Github issue.

### DIFF
--- a/aiida/cmdline/commands/cmd_computer.py
+++ b/aiida/cmdline/commands/cmd_computer.py
@@ -76,35 +76,20 @@ def _computer_test_no_unexpected_output(transport, scheduler, authinfo):  # pyli
     if retval != 0:
         return False, 'The command `echo -n` returned a non-zero return code ({})'.format(retval)
 
-    if stdout:
-        return False, u"""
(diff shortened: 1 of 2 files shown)


giovannipizzi:
Thanks!
I'd be ready to approve, but maybe in the template can you please add another `{}` and say if the string originates from the stdout or the stderr?
Something like `We detected some spurious output in {} ...` and you set that to the string `stderr` or `stdout` accordingly

sphuber:
@giovannipizzi done!

giovannipizzi (approved):

PR #4318, opened 2020-08-22
Make `--prepend-text` and `--append-text` options properly interactive 

### LINKED ISSUE #4300: Can't provide empty string for the append text in a yaml config file.
### Describe the bug

If the `append_text/prepend_text` parameters are set to an empty string `''` (without space) `verdi code setup` would request to provide them and wait for user's input. As a temporarily solution one can use a string with one space.

### Steps to reproduce

Try to setup a computer using the follwing config file:
```yaml
---
label: daint-mc
hostname: daint.cscs.ch
description: Piz Daint supercomputer at CSCS Lugano, Switzerland, multicore partition.
transport: ssh
scheduler: slurm
work_dir: /scratch/snx3000/{username}/aiida_run/
shebang: '#!/bin/bash'
mpirun_command: srun -n {tot_num_mpiprocs}
mpiprocs_per_machine: '36'
prepend_text: ''
append_text: ''
```

### Expected behavior

An empty string should be acceptable.

sphuber:
Fixes #4300 

The `--prepend-text` and `--append-text` options, used for both the
`verdi computer setup` and `verdi code setup` commands were normal
options as opposed to all other options that are interactive options.
The `InteractiveOption` is a custom option class type we developed that
will present the user with a prompt if it was not explicitly provided on
the command line. The `--non-interactive` flag can be used by the user
to prevent prompting, as long as a default value is defined.

The two options in question were implemented differently, presumably
because they take a potentially multiline string, which is not easily
defined on a prompt and therefore instead these options would be defined
through a text editor. The prompting of this text editor, if necessary,
was however not performed in the option itself, but in the command body.
At this point, the prompt cycle of the parameter cycle controlled by
click is already over.

Additionally, since the function checking whether the option had been
defined also considered an empty string as undefined, despite it being
the default, a specified empty string would still lead to the user to
be prompted, even when specifying `--non-interactive`.

The fix is to make both options proper interactive options just like the
rest. To this end, we create the `TemplateInteractiveOption` that works
just like the `InteractiveOption` with the only difference being that it
uses a file editor instead of an inline prompt. This change does force
us to have both options pop up their own file editor, whereas before
they were joined in a single file and both files were specified in the
same file, separated by a header that we defined.


sphuber:
> After checking out your changes, 2. works without prompting as well, but as you mention, it disables prompting for the pre/postpend scripts altogether (which is not what we want).

I forgot about this and agree, this is definitely not what we want. Before trying to continue to add patches, I think we first need to make absolutely sure how we want the commands with the `--config` and `--non-interactive` options to behave. Because I think, aside from the `prepend_text` and `append_text`, other options are working as expected. To sketch out the scenarios we have, I created the following example command:
```
#!/usr/bin/env python
import click

from aiida.cmdline.params import options
from aiida.cmdline.params.options.interactive import InteractiveOption


@click.command()
@click.option('--label', prompt='Label', cls=InteractiveOption, required=True)
@click.option('--hostname', prompt='Hostname', cls=InteractiveOption, default='localhost')
@options.NON_INTERACTIVE()
@options.CONFIG_FILE()
def command(label, hostname, non_interactive):
    print(label, hostname)


if __name__ == '__main__':
    command()
```
As you can see, it has two options, both required, but one with a default and one without. In addition, there are the `--config` and `--non-interactive` options to allow specifying from config file and to prevent all prompting. Then I create two config files, `config-complete.yml`:
```
label: 'daint'
hostname: daint.cscs.ch
```
and `config-partial.yml`:
```
label: 'daint'
```
We now have the following (meaningful) permutations of possible commands to use:

| Command                                                                      | Result                             |
|------------------------------------------------------------------------------|------------------------------------|
| ./test_cli.py --label daint --hostname daint.cscs.ch                         | Succeeds: {daint, daint.cscs.ch}   |
| ./test_cli.py --label daint                                                  | Prompts: for hostname              |
| ./test_cli.py --label daint -n                                               | Succeeds: {daint, localhost}       |
| ./test_cli.py -n                                                             | Errors: Missing option '--label'.  |
| ./test_cli.py --config config-complete.yml                                   | Succeeds: {daint, daint.cscs.ch}   |
| ./test_cli.py --config config-partial.yml                                    | Prompts: for hostname              |
| ./test_cli.py --config config-partial.yml -n                                 | Succeeds: {daint, localhost}       |
| ./test_cli.py --config config-partial.yml --hostname daint.cscs.ch           | Succeeds: {daint, daint.cscs.ch}   |

To me, the results are actually what I would expect, as long as the `--non-interactive` options is defined as follows:

> Use the `--non-interactive` flag if the command should never prompt and simply use the default if defined. Note that the command will fail if any default-less option is not specified explicitly.

If we agree on this definition and the current behavior, then the current problem is simply with the implementation of the `prepend_text` and `append_text`. They, unlike all other options, are not of type `InteractiveOption`. Instead, their value is checked _inside_ the command and a separate control loop is used to prompt for values. I think the fix would be to adapt these options to be individual prompters, just like the others. The default can then still remain the empty string and it would simply be accepted as an actual value.

Let me know if you agree with this assessment @ltalirz and I will update this PR.

ltalirz:
Thanks @sphuber for the detailed writeup - I agree with the description of the desired behavior entirely (perhaps we could put it somewhere into the documentation, e.g. starting a "command line interface" section under "internal architecture).

If that behavior can be attained, then great!

yakutovicha:
> After checking out your changes, 2. works without prompting as well, but as you mention, it disables prompting for the pre/postpend scripts altogether (which is not what we want).

@ltalirz, maybe I misunderstood something, but I believe this is exactly what we want. In the `yml` file you provide both prepend and append text are there. Or do you mean that user won't be asked to provide prepend/append text even if they are absent in the `yml` file?

ltalirz:
> Or do you mean that user won't be asked to provide prepend/append text even if they are absent in the yml file?

Correct, with the current code, the user would never be prompted for append/prepend texts.

yakutovicha:
Alright, thanks indeed @sphuber for the detailed explanation. I completely agree with the your proposed solution.

sphuber:
@ltalirz and @yakutovicha I have implemented the discussed changes and this should be good to go for a new review

aiida/cmdline/params/options/commands/code.py
```diff
@@ -97,3 +90,29 @@ def is_not_on_computer(ctx):
...
+
+APPEND_TEXT = OverridableOption(
+    '--append-text',
+    cls=TemplateInteractiveOption,
+    prompt='Append script',
+    type=click.STRING,
+    default='',
+    help='Bash code that should be appended after the executable call in all submit scripts for this code.',
+    extension='.bash',
+    header='**APPEND_TEXT**\nIf there is any bash code that should be appended after the executable call in all '
```
ltalirz, line 115: Careful, this is not prefixed by #= (same for prepend_text)
```
#==========================================================================#
#= **APPEND_TEXT**
If there is any bash code that should be appended after
#= the executable call in all submit scripts for this computer, type that
#= between the equal signs below and save the file.
#==========================================================================#


#==========================================================================#
#= All lines that start with `#=` will be ignored.
#==========================================================================#
```

It might be good to have a test that checks that the stored append_text, if not provided by the user, is actually an empty string.

aiida/cmdline/params/options/commands/code.py
```diff
@@ -97,3 +90,29 @@ def is_not_on_computer(ctx):
...
     help='Name of the computer, on which the code is installed.'
 )
+
+PREPEND_TEXT = OverridableOption(
+    '--prepend-text',
+    cls=TemplateInteractiveOption,
+    prompt='Prepend script',
+    type=click.STRING,
+    default='',
+    help='Bash code that should be prepended before the executable call in all submit scripts for this code.',
```
ltalirz, line 100: While "bash code" is correct as well, I would lean slightly towards "bash script" (or "bash commands"), just because code has such a special meaning in AiiDA.

And, I guess, strictly speaking it would be "prepended to".

I leave the decision up to you (both are minor suggestions, not required).

ltalirz (changes requested):
thanks @sphuber - I think this is exactly the behavior we want, and the code has become cleaner as well.

I noticed a minor issue, then it should be good to merge.

ltalirz:
P.S. As mentioned, you may want to consider adding what you wrote [here](https://github.com/aiidateam/aiida-core/pull/4318#issuecomment-679021426) somewhere into the documentation, e.g. starting a "command line interface" section under "internal architecture".

sphuber:
Thanks for the review, I will address the changes.

> P.S. As mentioned, you may want to consider adding what you wrote [here](https://github.com/aiidateam/aiida-core/pull/4318#issuecomment-679021426) somewhere into the documentation, e.g. starting a "command line interface" section under "internal architecture".

I know that is what I suggested, but really that was just to hash out where exactly the error in the code was. I don't see what information is there that is still worth having in terms of describing of how things should work. The difference between defining parameters inline or via config file ultimately had nothing to do with this issue. The only thing that we could describe is the intended working of the `--non-interactive` option:

> Use the --non-interactive flag if the command should never prompt and simply use the default if defined. Note that the command will fail if any default-less option is not specified explicitly.

but that is more for the user manual than the internal workings. What exactly do you think would have to go in the internal architecture?

aiida/cmdline/params/options/commands/code.py
sphuber: There already is such a test:
https://github.com/aiidateam/aiida-core/blob/f23078e0367aacf7e4e3319ab4ee8e1c48ed6745/tests/cmdline/commands/test_computer.py#L728-L755
It pops the options and runs interactively. The editor just saves the default. You can see that the test checks that the actual values will be the empty string.

aiida/cmdline/params/options/commands/code.py
sphuber: Agree with the suggestions. Have adapted them.

ltalirz:
If you think it's self-explanatory enough, perhaps it's indeed not needed to add to the documentation.
It would just have been an illustration of how the prompting mechanism is supposed to work and what the `--non-interactive` option does.

> Use the --non-interactive flag if the command should never prompt and simply use the default if defined. Note that the command will fail if any default-less option is not specified explicitly.

Yes, perhaps we can just expand the help string of the option a bit
https://github.com/aiidateam/aiida-core/blob/5e5d5e03a410c9986317f2badbccf242ade18c6f/aiida/cmdline/params/options/__init__.py#L208

My suggestion would be to focus a bit more on what the user can do
> In non-interactive mode the cli never prompts for input and uses default values, if defined. Provide your values via the respective command line options or via a configuration file.

aiida/cmdline/params/options/commands/code.py
ltalirz: So why did the test not fail? Or did it fail?

sphuber: It didn't fail. How did you generate the new header? Did you check the code out locally and run `verdi computer setup`? Because when I do it locally (Ubuntu) the newline is properly prefixed with the comment sequence. Might this be a problem of Mac OS? Anyway, to be safe, I have removed the newline in the `header` keyword while updating according to your suggestions. New header looks like this:
```
#==========================================================================#
#= PREPEND_TEXT: if there is any bash commands that should be prepended to the
#= executable call in all submit scripts for this computer, type that
#= between the equal signs below and save the file.
#==========================================================================#
```

look good?

sphuber:
> My suggestion would be to focus a bit more on what the user can do
> 
> > In non-interactive mode the cli never prompts for input and uses default values, if defined. Provide your values via the respective command line options or via a configuration file.

I took your suggestion but adapted it a bit and removed the mention of the configuration file. Although usable together, the `--non-interactive` and `--config` are not mutably dependent and so I think it does not belong in the description of the non-interactive flag. I have made it:

> In non-interactive mode, the CLI never prompts but simply uses default values for options that define one.

aiida/cmdline/params/options/commands/code.py
ltalirz: > It didn't fail. How did you generate the new header? Did you check the code out locally and run verdi computer setup? 

Yes. I used this yaml
```yaml
label: "daint-mc23"
hostname: "daint.cscs.ch"
description: Piz Daint supercomputer at CSCS Lugano, Switzerland, multicore partition.
transport: ssh
scheduler: "slurm"
work_dir: "/scratch/snx3000/{username}/aiida_run/"
shebang: "#!/bin/bash"
mpirun_command: "srun -n {tot_num_mpiprocs}"
mpiprocs_per_machine: "36"
prepend_text: |
    ### computer prepend_text start ###
    #SBATCH --partition=normal
    #SBATCH --account=mr0
    #SBATCH --constraint=mc
    #SBATCH --cpus-per-task=1
    export OMP_NUM_THREADS=$SLURM_CPUS_PER_TASK
    source $MODULESHOME/init/bash
    module load daint-mc
    ulimit -s unlimited
    ### computer prepend_text end ###
#append_text: ""
```
(which correctly prompted me to enter the append text. I added "asda")
```
$ verdi computer setup --config computer.yml
Info: enter "?" for help
Success: Computer<9> daint-mc23 created
Info: Note: before the computer can be used, it has to be configured with the command:
Info:   verdi computer configure ssh daint-mc23
$ verdi computer show daint-mc23
--------------  -------------------------------------------------------------------------
Label           daint-mc23
PK              9
UUID            39e1450d-d174-4bf2-87e2-538a055e9c66
Description     Piz Daint supercomputer at CSCS Lugano, Switzerland, multicore partition.
Hostname        daint.cscs.ch
Transport type  ssh
Scheduler type  slurm
Work directory  /scratch/snx3000/{username}/aiida_run/
Shebang         #!/bin/bash
Mpirun command  srun -n {tot_num_mpiprocs}
Prepend text    ### computer prepend_text start ###
                #SBATCH --partition=normal
                #SBATCH --account=mr0
                #SBATCH --constraint=mc
                #SBATCH --cpus-per-task=1
                export OMP_NUM_THREADS=$SLURM_CPUS_PER_TASK
                source $MODULESHOME/init/bash
                module load daint-mc
                ulimit -s unlimited
                ### computer prepend_text end ###
Append text     If there is any bash code that should be appended after
                asda
--------------  -------------------------------------------------------------------------
```
> Because when I do it locally (Ubuntu) the newline is properly prefixed with the comment sequence.

ltalirz: Anyhow, with your latest version, it works fine.

ltalirz (approved):
thanks @sphuber , latest version works fine

aiida/cmdline/params/options/commands/code.py
sphuber: I have the suspicion this might be a bug in `click` on MacOS in combination with `jinja2` when using the wordwrap filter. It was due to the literal newline I put in the `{value}` placeholder. Apparently this works on Ubuntu but not on MacOS. Anyway, I removed the newline such that it works everywhere. We'll merge this

Issue #3806, opened 2020-02-25
remove pip's --editable flag on CI tests


ltalirz:
During the hackathon, @yakutovicha suggested to remove the --editable flag for pip installs during CI tests.
The reasoning: with a regular install you also test whether all files are included (which otherwise might be present just in the git repo but not in the distribution that is later uploaded to PyPI).

ltalirz:
Unfortunately, there is a problem with this in practice - the setup to measure test coverage becomes complicated, since we need to point coverage to directories *outside* the git repo where the python files of the package are installed.

I'm closing this for the moment - happy to move further in this direction if the coverage issue can be addressed.

greschd:
Would creating two separate jobs for the different installs be feasible, or too much additional run-time for the build?

ltalirz:
Running another job just to test the regular install would be possible but I'm wondering a bit whether it's worth the effort...

Anyhow, we should re-enable coverage reports ;-)

greschd:
Yeah, might not be. I've gotten tired of manually testing the regular install before release, so in some plugins I've added that.. but the ration of builds to releases is a very different one there 😄

By the way, do you know if there's a difference between
```bash
python setup.py sdist
pip install dist/<package>.tar.gz
```
and
```
pip install .
```

I've always done the former to check that the package I'm about to upload is valid, but now I'm wondering if ``pip install .`` might be just a shortcut to do the same.

ltalirz:
Good question - I don't know

PR #4320, opened 2020-08-26
Fix/issue 4286 pip 2020 resolver

### LINKED ISSUE #4286: pip's 2020-10 resolver

The `pip` installer is updating their resolver for version `20.3` in October 2020 (see [this](https://pip.pypa.io/en/stable/user_guide/#changes-to-the-pip-dependency-resolver-in-20-2-2020) for more information).

Since `aiida-core` is fixed to some old versions for various larger python packages (like `tornado`, `numpy`, ...) this will make it difficult to depend on `aiida-core` for a novel package.

Furthermore, it may even mess with `aiida-core`'s `pip install`. This can be tested by using the `--use-feature=2020-resolver` option from `pip` version `20.2`.

Update `aiida-core` and it's main dependency packages (like `plumpy`, `circus`, ...) to support newest versions of other or related dependencies.

### Describe alternatives you've considered

Support newest versions of the most major dependencies, like `numpy`.

csadorf:
* Test whether our installation paths are compatible with the pip version 20.2 `2020-resolver` feature flag.
 * Ensure that tests also cover _extras_.

Resolves #4286 .

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -41,6 +41,10 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
 
+    strategy:
+      matrix:
+        pip-feature-flag: [ '', '--use-feature=2020-resolver' ]
+
     steps:
     - uses: actions/checkout@v2
 
@@ -51,7 +55,8 @@ jobs:
 
     - name: Pip install
       run: |
-        python -m pip install -e .
+        python -m pip --version
+        python -m pip install -e . ${{ matrix.pip-feature-flag }}
         python -m pip freeze
 
     - name: Test importing aiida
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -57,7 +57,7 @@ jobs:
     - name: Pip install
       run: |
         python -m pip --version
-        python -m pip install -e .${{ matrix.extras }} ${{ matrix.pip-feature-flag }}
+        python -m pip install .${{ matrix.extras }} ${{ matrix.pip-feature-flag }}
         python -m pip freeze
 
     - name: Test importing aiida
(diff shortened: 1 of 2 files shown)


.github/workflows/test-install.yml
```diff
@@ -51,7 +55,8 @@ jobs:
 
     - name: Pip install
       run: |
-        python -m pip --use-feature=2020-resolver install -e .
+        python -m pip --version
+        python -m pip install -e . ${{ matrix.pip-feature-flag }}
```
greschd, line 59: Unrelated to this change, but is there a reason for using `-e` here? Editable installs just add an "egg-link" to the environment, instead of "properly" installing it. Not sure if there might be some issues that we could catch without `-e` that we won't see now.

csadorf: Good point. There is no need for the `-e` here.

csadorf:
@CasperWA @ltalirz This should be ready. Everything looks fine, however the `codecov/project` check has failed, I don't know why.

CasperWA (approved):
Thanks @csadorf !

The addition of `pip --version` seems like a good choice as well to bring in some context.

ltalirz:
@csadorf since you switched away from editable installs, I'm a bit puzzled why you did not encounter the test coverage issue https://github.com/aiidateam/aiida-core/issues/3806 here ...
Is it possible that you are installing aiida-core but then still continue using the files from the local folder when running the test?

There is a small decrease in test coverage reported; can you check whether this is a fluke?

csadorf:
> @csadorf since you switched away from editable installs, I'm a bit puzzled why you did not encounter the test coverage issue #3806 here ...
> Is it possible that you are installing aiida-core but then still continue using the files from the local folder when running the test?
> 
> There is a small decrease in test coverage reported; can you check whether this is a fluke?
> 
> Perhaps it's wiser to keep the `--editable` option after all (?)

I *did* encounter the test coverage issue and that also explains that then. :) I'll revert that change.

greschd:
Ha, I guess I could've remember that one 😅

Sorry for the troubles!

csadorf:
@ltalirz Everything should be fine now.
@CasperWA Unfortunately your review was automatically dismissed.

ltalirz (approved):
thanks @csadorf , looks good to me!

PR #4324, opened 2020-08-26
move seekpath import inside function

### LINKED ISSUE #4323: importing aiida.tools imports seekpath
```python
In [1]: import aiida.tools
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-c2ba1a6f79ed> in <module>
----> 1 import aiida.tools

~/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/tools/__init__.py in <module>
     23
     24 from .calculations import *
---> 25 from .data.array.kpoints import *
     26 from .data.structure import *
     27 from .dbimporters import *

~/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/tools/data/array/kpoints/__init__.py in <module>
     14 from aiida.orm import KpointsData, Dict
     15 from aiida.tools.data.array.kpoints import legacy
---> 16 from aiida.tools.data.array.kpoints import seekpath
     17
     18 __all__ = ('get_kpoints_path', 'get_explicit_kpoints_path')

~/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/tools/data/array/kpoints/seekpath.py in <module>
      9 ###########################################################################
     10 """Tool to automatically determine k-points for a given structure using SeeK-path."""
---> 11 import seekpath
     12
     13 from aiida.orm import KpointsData, Dict

ModuleNotFoundError: No module named 'seekpath'
```

This prevents a user who has not installed the `atomic_tools` extra from using any functionality present under `aiida.tools`, including for example any import/export-related functionality.

This seems problematic to me, as well as unnecessary. 
The seekpath dependency should only be required when actually using data structures that depend on it (e.g. KpointsData)

ltalirz:
Fixes #4323 

Prior to this commit, importing aiida.tools caused an import of the
seekpath package, which is not a required dependency of aiida-core (only
of the 'atomic_tools' extra).

### DIFF
--- a/.github/workflows/verdi.sh
+++ b/.github/workflows/verdi.sh
@@ -34,3 +34,6 @@ while true; do
     fi
 
 done
+
+$VERDI devel check-load-time
+$VERDI devel check-undesired-imports
--- a/aiida/cmdline/commands/cmd_devel.py
+++ b/aiida/cmdline/commands/cmd_devel.py
@@ -48,6 +48,24 @@ def devel_check_load_time():
     echo.echo_success('no issues detected')
 
 
+@verdi_devel.command('check-undesired-imports')
+def devel_check_undesired_imports():
+    """Check that verdi does not import python modules it shouldn't.
+
+    Note: The blacklist was taken from the list of packages in the 'atomic_tools' extra but can be extended.
+    """
+    loaded_modules = 0
+
+    for modulename in ['seekpath', 'CifFile', 'ase', 'pymatgen', 'spglib', 'pymysql']:
+        if modulename in sys.modules:
+            echo.echo_warning('Detected loaded module "{}"'.format(modulename))
+            loaded_modules += 1
+
+    if loaded_modules > 0:
+        echo.echo_critical('Detected {} unwanted modules'.format(loaded_modules))
+    echo.echo_success('no issues detected')
+
+
 @verdi_devel.command('run_daemon')
 @decorators.with_dbenv()
 def devel_run_daemon():
--- a/aiida/tools/data/array/kpoints/__init__.py
+++ b/aiida/tools/data/array/kpoints/__init__.py
@@ -12,8 +12,6 @@ Various utilities to deal with KpointsData instances or create new ones
 (e.g. band paths, kpoints from a parsed input text file, ...)
 """
 from aiida.orm import KpointsData, Dict
-from aiida.tools.data.array.kpoints import legacy
-from aiida.tools.data.array.kpoints import seekpath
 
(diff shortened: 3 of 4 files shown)


aiida/cmdline/commands/cmd_devel.py
```diff
@@ -48,6 +48,24 @@ def devel_check_load_time():
     echo.echo_success('no issues detected')
 
 
+@verdi_devel.command('check-undesired-imports')
+def devel_check_undesired_imports():
+    """Check that verdi does not import python modules it shouldn't.
+
+    Note: The blacklist was taken from the list of packages in the 'atomic_tools' extra but can be extended.
```
CasperWA, line 55: Should we use this more dynamically instead by reading it in from `setup.json`, or is it preferable to keep it explicit and hard-coded here?

.github/workflows/verdi.sh
```diff
@@ -34,3 +34,6 @@ while true; do
     fi
 
 done
+
+$VERDI devel check-load-time
+$VERDI devel check-undesired-imports
```
CasperWA, line 39: Won't this just run `which verdi ...`? Will this actually call the verdi commands?

CasperWA (commented):
Thanks @ltalirz!

And well found, concerning the `seekpath` dependency.

I have just a few questions.

.github/workflows/verdi.sh
sphuber: `which verdi` will just resolve to the full path of the verdi binary, and it should then just be executed normally. Note that we don't really need to call `check-load-time` because that is exactly already done in the bash code in this file. The command is named a bit misleading because it just checks that the orm backend is not loaded and the orm module is not imported. These were two of the leading causes of slow down but not the exclusive ones. That's why I created the bash timing code. We can keep the `check-load-time` call I guess, doesn't hurt

CasperWA: > `which verdi` will just resolve to the full path of the verdi binary, and it should then just be executed normally. (...)

Ah, I see the `` now in the `VERDI` parameter definition. Makes sense.

aiida/cmdline/commands/cmd_devel.py
ltalirz: Up for discussion of course. However, the setup.json handling code is currently under `utils/` i.e. not even in aiida-core.
Not sure it's worth the effort.

ltalirz:
Let me know whether there is still something I need to change here.
I have no idea why test coverage went down dramatically here; is it a fluke?

giovannipizzi (approved):

PR #4325, opened 2020-08-26
Opt-out for creating auto-import group

### LINKED ISSUE #3680: Make importing into a `Group` completely optional
Currently both the API `import_data` as well as the CLI `verdi import` do not provide an option to not have the imported nodes put into a group at all. I would like this to be completely optional.

CasperWA:
Closes #3680 

When importing an export archive, the `create_autogroup` can be set to `False` to _not_ create the auto-import Group that contains all the imported Nodes. By default it is `True`.

Via `verdi import`, this option is available through the `--create-import-group / --no-create-import-group` flag option.


sphuber:
Before we add another option to the interface, I would like to poll who would be in favor of turning the creation of the group off by default, and just have `-G` to create a group, where the user specifies the label. Is there anyone that relies on this automatic group generation?

Issue #2231, opened 2018-11-22
AiiDA extras


giovannipizzi:
Currently all extras are accessible to users. If AiiDA has to store custom data, that we don't necessarily want the user to change without noticing, we don't currently have a namespacing scheme to allow this.
Two current occurrences of this are for instance the `hidden` flag of a code and the `hash` of a Node (currently stored in extras but the user might inadvertently replace everything).

## Proposed solution
- namespace AiiDA-reserved keys in extras (suggestion: prefix starts with `_aiida_`
- possibly leave this functionality available also to plugins, reserving `_aiida_core_` to AiiDA-core, and suggesting plugins to use `_aiida_entrypoint_` for plugins (e.g. `_aiida_quantumespresso_`).
- migrate current flags under this namespace where appropriate
- have `get_extras(also_private=False)` where the new `also_private` flag, `False` by default, establishes if key/value pairs starting with `_aiida_` should be shown or not. (I chose `also_private` instead of, e.g., `show_private` because then the same name could be used also in the set methods, see below)
- TBD: if `get_extra()` should have the flag as well or not (my idea: not)
- adapt also other extra methods (e.g. `set_extra(s)` would complain with an exception when setting a key starting with `_aiida_` unless a flag is set, `reset_extras` by default resets only extras without prefix, unless a flag is set)
  - also, ensure that this is the only current method to change extras, otherwise adapt also the other ones
- decide if the querybuilder should hide these results (probably not, to be documented - the other option is to specify a flag in the append of a node to a QueryBuilder object, but anyway this would apply only to projecting all extras, all other cases shouldn't have a flag I believe)
- As a bonus, in the tests for extras we wouldn't need (as we do now) to ignore all extras starting with `_aiida_` (that is currently done because the hashing mechasism would add "unexpected" extras).

## List of known internal extras
* `Code.hidden`: this is being used in import/export logic
* `Node._aiida_hash`
* `Node._aiida_cached_from`

szoupanos:
This is not urgent but there should be a discussion on what should happen when extras are exported (and what is the behaviour of private extras).
Issue #1761

yakutovicha:
After today's discussion we agreed:
1. Instead of *private extras* call them *aiida extras* as it reflects better the idea of keeping them internally managed by aiida.
2. Create a separate api for the aiida extras and keep it hidden.
    2.1.  `get_aiida_extra() / get_aiida_extras()`
[...]

PR #4328, opened 2020-08-27
Add 'extras' column to Group database class

### LINKED ISSUE #4303: Add `metadata` column to the `Group` database class
For certain use cases, it would be very useful to be able to store "metadata" on a `Group` instance. Especially now that `Groups` are pluginnable, a lot of power can be given to the user through this simple addition. One example use case, is the creation of a pseudo potential family, by subclassing `Group` as `PseudoFamily`. This group would store pseudo potential data nodes, one for each element. Often these pseudo potentials come with recommended metadata, such as wave function cutoffs in the case of `UpfData` nodes. These are family dependent and so should not be stored on the data node but belong to the "family". It would be perfect to be able to store this in the metadata. Now we are forced to come up with more complex solutions, such as using a `Dict` node that also contains a back reference to the `Group` to which it belongs.

As of yet, I do not really see any disadvantages of this feature, except that maybe it could encourage users to store too much information in the group metadata and "abuse" it in ways that end up losing provenance. However, with proper documentation and example use cases, this should be ok. @giovannipizzi @ltalirz @greschd comments?

mbercx:
Fixes #4303 

The changes in this PR can be summarized by two points:
* Move all methods of the `Node` backend entity classes related to `attributes` and `extras` to two mixin classes: `AttributesBackendEntity` and `ExtrasBackendEntity`. Add these as parent classes to the `BackendNode` class.
* Add an `extras` column to the `db_dbgroup` table, which contains mutable  information that can be stored for `Group` instances. Provide the methods for the frontend class (`Group`) and mix in the methods for the backend class (`BackendGroup`).

Notes: 
* After taking a closer look at the idea of setting the `extras` column to `null` by default in the database, I think it's not really necessary. For nodes, the extras are almost never empty, because [when the node is stored the AiiDA hash is written to the extras](https://github.com/aiidateam/aiida-core/blob/develop/aiida/orm/nodes/node.py#L1075). Groups can of course have empty extras, but usually a user won't have that many groups for it to make a difference, so I don't think it's worth adding extra logic to the backend entities do deal with this. I did find that the `attributes` of `FolderData` are actually empty, so I suppose we could set these to `null` in the database. But then we's still have to adjust the `AttributesBackendEntity` methods to deal with this.

Questions:
* ~~Why is the `_aiida_hash` stored as an `extra` and not as an `attribute`? I.e.: why do we need the functionality to re-hash a `Node`?~~ See #2231
* ~~Might it be better to split this into two PR's? The first one would be: "Add mix-in classes for the attributes and extras methods" and include the first two commits.~~ Yes! Done, see #4376 

List of TODO's:
- [x] ~~Add unit tests~~
- [x] ~~Consider changing the default value for extras stored in the database to `null` instead of an empty dictionary `'{}'`. This is currently also the default for the `Node` extras, so if we make this change for both it might meaningfully reduce the size of the database. This will require some changes in the extras methods to deal with the `null` value stored in the database, however.~~
- [x] ~~Consider writing two "mix-in" classes for the extras methods of the `Node`/`Group` class and the corresponding Django and SqlAlchemy backend classes, as they are pretty much the same at each layer.~~
- [x] ~~Consider adding an `extras` mixing class for the frontend classes.~~
- [x] ~~(**Nitpicking alert**) Homogenize the directory structures of the `backend` subpackage. Why use `djsite` instead of `django`? Why are the models of django in `backends.djsite.db.models.py`, whereas for SqlAlchemy they are spread over different modules in `backends.sqlalchemy.models`?~~

### DIFF
--- a/aiida/backends/djsite/db/migrations/0045_dbgroup_extras.py
+++ b/aiida/backends/djsite/db/migrations/0045_dbgroup_extras.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Migration to add the `extras` JSONB column to the `DbGroup` model."""
+# pylint: disable=invalid-name
+import django.contrib.postgres.fields.jsonb
+from django.db import migrations
+from aiida.backends.djsite.db.migrations import upgrade_schema_version
+
+REVISION = '1.0.45'
+DOWN_REVISION = '1.0.44'
+
+
+class Migration(migrations.Migration):
+    """Migrate to add the extras column to the dbgroup table."""
+    dependencies = [
+        ('db', '0044_dbgroup_type_string'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='dbgroup',
+            name='extras',
+            field=django.contrib.postgres.fields.jsonb.JSONField(default=dict, null=False),
+        ),
+        upgrade_schema_version(REVISION, DOWN_REVISION),
+    ]
--- a/aiida/backends/sqlalchemy/migrations/versions/0edcdd5a30f0_add_extras_to_group.py
+++ b/aiida/backends/sqlalchemy/migrations/versions/0edcdd5a30f0_add_extras_to_group.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+# pylint: disable=no-member,invalid-name
+"""Migration to add the `extras` JSONB column to the `DbGroup` model.
+
+Revision ID: 0edcdd5a30f0
+Revises: bf591f31dd12
+Create Date: 2019-04-03 14:38:50.585639
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+from sqlalchemy.sql import text
+
+# revision identifiers, used by Alembic.
+revision = '0edcdd5a30f0'
+down_revision = 'bf591f31dd12'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    """Upgrade: Add the extras column to the 'db_dbgroup' table"""
+    op.add_column('db_dbgroup', sa.Column('extras', postgresql.JSONB(astext_type=sa.Text())))
+    op.execute(text("""
+    UPDATE db_dbgroup
+    SET extras='{}'
+    """))
+    op.alter_column('db_dbgroup', 'extras', nullable=False)
+
+
+def downgrade():
+    """Downgrade: Drop the extras column from the 'db_dbgroup' table"""
+    op.drop_column('db_dbgroup', 'extras')
--- a/aiida/backends/sqlalchemy/models/group.py
+++ b/aiida/backends/sqlalchemy/models/group.py
@@ -15,7 +15,7 @@ from sqlalchemy.orm import relationship, backref
 from sqlalchemy.schema import Column, Table, UniqueConstraint, Index
 from sqlalchemy.types import Integer, String, DateTime, Text
 
-from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.dialects.postgresql import UUID, JSONB
 
 from aiida.common import timezone
 from aiida.common.utils import get_new_uuid
@@ -46,6 +46,8 @@ class DbGroup(Base):
 
     time = Column(DateTime(timezone=True), default=timezone.now)
     description = Column(Text, nullable=True)
+
+    extras = Column(JSONB, default=dict, nullable=False)
 
     user_id = Column(Integer, ForeignKey('db_dbuser.id', ondelete='CASCADE', deferrable=True, initially='DEFERRED'))
     user = relationship('DbUser', backref=backref('dbgroups', cascade='merge'))
--- a/aiida/backends/djsite/db/migrations/__init__.py
+++ b/aiida/backends/djsite/db/migrations/__init__.py
@@ -21,7 +21,7 @@ class DeserializationException(AiidaException):
     pass
 
 
-LATEST_MIGRATION = '0044_dbgroup_type_string'
+LATEST_MIGRATION = '0045_dbgroup_extras'
 
 
 def _update_schema_version(version, apps, _):
@@ -273,7 +273,7 @@ def _deserialize_attribute(mainitem, subitems, sep, original_class=None, origina
         try:
             return json.loads(mainitem['tval'])
         except ValueError:
-            raise DeserializationException('Error in the content of the json field')
+            raise DeserializationException('Error in the content of the json field') from ValueError
     else:
         raise DeserializationException("The type field '{}' is not recognized".format(mainitem['datatype']))
 
@@ -426,7 +426,8 @@ class ModelModifierV0025:
         try:
             attr = cls.objects.get(dbnode=dbnode_node, key=key)
         except ObjectDoesNotExist:
-            raise AttributeError('{} with key {} for node {} not found in db'.format(cls.__name__, key, dbnode.pk))
+            raise AttributeError('{} with key {} for node {} not found in db'.format(cls.__name__, key, dbnode.pk)) \
+                from ObjectDoesNotExist
 
         return self.getvalue(attr)
 
@@ -674,7 +675,7 @@ class ModelModifierV0025:
                     'another entry already exists and the creation would '
                     'violate an uniqueness constraint.\nFurther details: '
                     '{}'.format(cls.__name__, exc)
-                )
+                ) from exc
             raise
 
     @staticmethod
@@ -806,7 +807,7 @@ class ModelModifierV0025:
                 raise ValueError(
                     'Unable to store the value: it must be either a basic datatype, or json-serializable: {}'.
                     format(value)
-                )
+                ) from TypeError
 
             new_entry.datatype = 'json'
             new_entry.tval = jsondata
--- a/aiida/backends/djsite/db/models.py
+++ b/aiida/backends/djsite/db/models.py
@@ -254,6 +254,8 @@ class DbGroup(m.Model):
     # On user deletion, remove his/her groups too (not the calcuations, only
     # the groups
     user = m.ForeignKey(DbUser, on_delete=m.CASCADE, related_name='dbgroups')
+    # JSON Extras
+    extras = JSONField(default=dict, null=False)
 
     class Meta:  # pylint: disable=too-few-public-methods
         unique_together = (('label', 'type_string'),)
--- a/aiida/backends/sqlalchemy/migrations/versions/0edcdd5a30f0_dbgroup_extras.py
+++ b/aiida/backends/sqlalchemy/migrations/versions/0edcdd5a30f0_dbgroup_extras.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+# pylint: disable=no-member,invalid-name
+"""Migration to add the `extras` JSONB column to the `DbGroup` model.
+
+Revision ID: 0edcdd5a30f0
+Revises: bf591f31dd12
+Create Date: 2019-04-03 14:38:50.585639
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+from sqlalchemy.sql import text
+
+# revision identifiers, used by Alembic.
+revision = '0edcdd5a30f0'
+down_revision = 'bf591f31dd12'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    """Upgrade: Add the extras column to the 'db_dbgroup' table"""
+    op.add_column('db_dbgroup', sa.Column('extras', postgresql.JSONB(astext_type=sa.Text())))
+    op.execute(text("""
+    UPDATE db_dbgroup
+    SET extras='{}'
+    """))
+    op.alter_column('db_dbgroup', 'extras', nullable=False)
+
+
+def downgrade():
+    """Downgrade: Drop the extras column from the 'db_dbgroup' table"""
+    op.drop_column('db_dbgroup', 'extras')
--- a/aiida/orm/groups.py
+++ b/aiida/orm/groups.py
@@ -7,7 +7,7 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-""" AiiDA Group entites"""
+"""AiiDA Group entites"""
 from abc import ABCMeta
 from enum import Enum
 import warnings
@@ -77,7 +77,7 @@ class GroupTypeString(Enum):
     USER = 'user'
 
 
-class Group(entities.Entity, metaclass=GroupMeta):
+class Group(entities.Entity, entities.EntityExtrasMixin, metaclass=GroupMeta):
     """An AiiDA ORM implementation of group of nodes."""
 
     class Collection(entities.Collection):
--- a/aiida/orm/implementation/django/utils.py
+++ b/aiida/orm/implementation/django/utils.py
@@ -114,9 +114,9 @@ class ModelWrapper:
     def _flush(self, fields=None):
         """Flush the fields of the model to the database.
 
-        .. note:: If the wrapped model is not actually save in the database yet, this method is a no-op.
+        .. note:: If the wrapped model is not actually saved in the database yet, this method is a no-op.
 
-        :param fields: the model fields whose currently value to flush to the database
+        :param fields: the model fields whose current value to flush to the database
         """
         if self.is_saved():
             try:
--- a/aiida/orm/implementation/sqlalchemy/utils.py
+++ b/aiida/orm/implementation/sqlalchemy/utils.py
@@ -120,9 +120,9 @@ class ModelWrapper:
     def _flush(self, fields=()):
         """Flush the fields of the model to the database.
 
-        .. note:: If the wrapped model is not actually save in the database yet, this method is a no-op.
+        .. note:: If the wrapped model is not actually saved in the database yet, this method is a no-op.
 
-        :param fields: the model fields whose currently value to flush to the database
+        :param fields: the model fields whose current value to flush to the database
         """
         if self.is_saved():
             for field in fields:
--- a/aiida/orm/nodes/data/remote.py
+++ b/aiida/orm/nodes/data/remote.py
@@ -34,7 +34,7 @@ class RemoteData(Data):
         .. deprecated:: 1.4.0
             Will be removed in `v2.0.0`, use the `self.computer.label` property instead.
         """
-        return self.computer.label
+        return self.computer.label  # pylint: disable=no-member
 
     def get_remote_path(self):
         return self.get_attribute('remote_path')
@@ -76,7 +76,8 @@ class RemoteData(Data):
                 if exception.errno == 2:  # file does not exist
                     raise IOError(
                         'The required remote file {} on {} does not exist or has been deleted.'.format(
-                            full_path, self.computer.label
+                            full_path,
+                            self.computer.label  # pylint: disable=no-member
                         )
                     )
                 raise
@@ -98,7 +99,7 @@ class RemoteData(Data):
                 if exception.errno == 2 or exception.errno == 20:  # directory not existing or not a directory
                     exc = IOError(
                         'The required remote folder {} on {} does not exist, is not a directory or has been deleted.'.
-                        format(full_path, self.computer.label)
+                        format(full_path, self.computer.label)  # pylint: disable=no-member
                     )
                     exc.errno = exception.errno
                     raise exc
@@ -111,7 +112,7 @@ class RemoteData(Data):
                 if exception.errno == 2 or exception.errno == 20:  # directory not existing or not a directory
                     exc = IOError(
                         'The required remote folder {} on {} does not exist, is not a directory or has been deleted.'.
-                        format(full_path, self.computer.label)
+                        format(full_path, self.computer.label)  # pylint: disable=no-member
                     )
                     exc.errno = exception.errno
                     raise exc
@@ -135,7 +136,7 @@ class RemoteData(Data):
                 if exception.errno == 2 or exception.errno == 20:  # directory not existing or not a directory
                     exc = IOError(
                         'The required remote folder {} on {} does not exist, is not a directory or has been deleted.'.
-                        format(full_path, self.computer.label)
+                        format(full_path, self.computer.label)  # pylint: disable=no-member
                     )
                     exc.errno = exception.errno
                     raise exc
@@ -148,7 +149,7 @@ class RemoteData(Data):
                 if exception.errno == 2 or exception.errno == 20:  # directory not existing or not a directory
                     exc = IOError(
                         'The required remote folder {} on {} does not exist, is not a directory or has been deleted.'.
-                        format(full_path, self.computer.label)
+                        format(full_path, self.computer.label)  # pylint: disable=no-member
                     )
                     exc.errno = exception.errno
                     raise exc
--- a/tests/backends/aiida_django/migrations/test_migrations_0045_dbgroup_extras.py
+++ b/tests/backends/aiida_django/migrations/test_migrations_0045_dbgroup_extras.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+# pylint: disable=import-error,no-name-in-module,invalid-name
+"""Test migration to add the `extras` JSONB column to the `DbGroup` model."""
+
+from .test_migrations_common import TestMigrations
+
+
+class TestGroupExtrasMigration(TestMigrations):
+    """Test migration to add the `extras` JSONB column to the `DbGroup` model."""
+
+    migrate_from = '0044_dbgroup_type_string'
+    migrate_to = '0045_dbgroup_extras'
+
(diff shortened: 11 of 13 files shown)


mbercx:
The `extras` methods for the `Group` and `DjangoGroup` class are largely copied from their corresponding `Node` and `DjangoNode` versions. 

I've already tested the functionality manually for a new profile and django-backend database. All of the methods seem to be working as expected.

The migration seems to work fine when I setup a new profile with an empty database. However, when I try setting up a profile with the current schema (1.0.44) and migrating it after switching to my feature branch, this gets stuck at the 'Running migrations' step. This isn't super surprising considering my inexperience with databases and migrations. 😅  It would be great if someone (@sphuber, @ltalirz, @giovannipizzi ...?) could have a look and see what I've missed.

Of course, other comments/suggestions are also most welcome!

sphuber:
I think you generated the migration file yourself, correct? If you introduce a schema change, as is the case here, you can let Django and SqlAlchemy create the migration file itself. At least, this will create the basic migration code, which you then have to update a bit with our own requirements, such as upping the schema version (in the case of Django, for SqlA this is automatic). To run the command to automatically create migrations you can run:
```
python aiida/backends/djsite/manage.py -p PROFILE makemigrations
python aiida/backends/sqlalchemy/manage.py -p PROFILE revision 'Add extras as JSONB column to the Group model.'
```

mbercx:
> I think you generated the migration file yourself, correct? If you introduce a schema change, as is the case here, you can let Django and SqlAlchemy create the migration file itself. 

Good to know! 😅 I've used the commands you provided to generate the migration files for both the Django and SqlAlchemy backends, and added the extras property and methods to the `Group` class and its backend classes. The methods and migrations seem to work fine based on some manual testing, but I'll start working on proper unit tests. If you can have a look at the current implementation in the meanwhile, that would great. :)

I'll add a little checklist with TODO's in the first message for the improvements we already discussed.

aiida/backends/sqlalchemy/models/group.py
```diff
@@ -57,6 +59,13 @@ class DbGroup(Base):
     Index('db_dbgroup_dbnodes_dbnode_id_idx', table_groups_nodes.c.dbnode_id)
     Index('db_dbgroup_dbnodes_dbgroup_id_idx', table_groups_nodes.c.dbgroup_id)
 
+    def __init__(self, *args, **kwargs):
+        """Add a default for the extras column."""
+        super().__init__(*args, **kwargs)
+
+        if self.extras is None:
+            self.extras = dict()
+
```
sphuber, line 68: I wonder if this is necessary. I know this is what the `Node` class does, but what happens if we remove it? Couldn't our backend layer not just return `{}` if it gets `None`? Also, what is the default value in the database? Is it null? Or should we add `nullable` explicitly to the column declaration?

Note that this is asymmetric with the Django solution. There the default empty dict is set on the column instead of in the logic. We may want to homogenize it.

sphuber (commented):
Will wait with further review until you have added tests and attempted to deduplicate the code through a mixin class

sphuber:
@mbercx what is the status on this PR. When do you think you will have the time to finish it? Ideally, I would like to ship it with v1.4 since we will need this functionality in `aiida-pseudo`.

mbercx:
> @mbercx what is the status on this PR. When do you think you will have the time to finish it? Ideally, I would like to ship it with v1.4 since we will need this functionality in `aiida-pseudo`.

Sorry about the delay. I've been working on and testing this PR on my work station @ EPFL, but at the moment this is difficult due to the fact the wifi connection at the hostel keeps interrupting every 5-10 mins... I was going to continue working on this when my internet situation was improved, but I'll move the branch and test notebook to my local machine and prioritise it.

sphuber:
> Sorry about the delay. I've been working on and testing this PR on my work station @ EPFL, but at the moment this is difficult due to the fact the wifi connection at the hostel keeps interrupting every 5-10 mins... I was going to continue working on this when my internet situation was improved, but I'll move the branch and test notebook to my local machine and prioritise it.

No worries, actually, there is not much of a rush, since I will be AFK next week, so the release will at the earliest be the week after. Just wanted to know a bit what your schedule was and how things were going. Is there any reason you were running this on a workstation? You don't need much performance for this. Or did you simply not yet have a dev env installed on your laptop?

mbercx:
>  Is there any reason you were running this on a workstation? You don't need much performance for this. Or did you simply not yet have a dev env installed on your laptop?

Mainly to keep things a little organised in one place. If I start developing on too many different locations I get confused. 😅

aiida/backends/sqlalchemy/models/group.py
mbercx: I copied the logic from `DbNode` indeed, but I agree that it makes more sense to simply set a default in the column declaration. I've done so now for both the `DbNode` and `DbGroup` model.

After taking a closer look at the idea of setting the `extras` column to `null` by default in the database, I think it's not really necessary. For nodes, the extras are almost never empty, because [when the node is stored the AiiDA hash is written to the extras](https://github.com/aiidateam/aiida-core/blob/develop/aiida/orm/nodes/node.py#L1075). Groups can of course have empty extras, but usually a user won't have that many groups for it to make a difference, so I don't think it's worth adding extra logic to the backend entities do deal with this.

mbercx:
Ok, I've added the mixin classes for `extras` as well as `attributes`, and have fixed the issues I was having with the migrations. The commits are now neatly organised so it's easier to spot how the `attributes`/`extras` methods were homogenized. See the first post and the commit messages for more details.

I'll start working on the tests for the `Group` extras next.

aiida/orm/implementation/groups.py
```diff
@@ -12,14 +12,15 @@
 import abc
 
 from aiida.common import exceptions
+from aiida.orm.implementation.entities import ExtrasBackendEntity
```
sphuber, line 15: Since this comes from the same package, it is nice to use relative import
```suggestion
from .entities import ExtrasBackendEntity
```
Makes it easier to read and easier to refactor should the package ever be moved

aiida/orm/implementation/entities.py
```diff
@@ -0,0 +1,354 @@
...
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Classes and methods for backend non-specific entities"""
+
+import abc
+
+from aiida.orm.utils.node import clean_value, validate_attribute_extra_key
+
+
+class AttributesBackendEntity(abc.ABC):
```
sphuber, line 17: I would be tempted to name this differently. First of all, I think it is important to suffix it with `Mixin`. This suggests that the class is not supposed to be used on its own, but can only be functional when it is mixed in with another class. So I would suggest, `BackendEntityAttributesMixin`. Same for `BackendEntityExtrasMixin`.

aiida/orm/implementation/nodes.py
```diff
@@ -11,12 +11,13 @@
 
 import abc
 
+from aiida.orm.implementation.entities import AttributesBackendEntity, ExtrasBackendEntity
```
sphuber, line 14: relative imports

aiida/orm/implementation/entities.py
```diff
@@ -0,0 +1,354 @@
...
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Classes and methods for backend non-specific entities"""
+
+import abc
+
+from aiida.orm.utils.node import clean_value, validate_attribute_extra_key
```
sphuber, line 14: I am thinking, since `validate_attribute_extra_key` is now really only used in the backend implementation, maybe it makes sense to move it in this package, so move it into module `aiida.orm.implementation.utils`. The same would almost go for `clean_value`. It is just used by some modules in `aiida.tools`. However, since `aiida.orm.utils.node` is not public facing API, we can just move both. Up to our discretion. What do you think?

aiida/orm/implementation/entities.py
```diff
@@ -0,0 +1,354 @@
...
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Classes and methods for backend non-specific entities"""
+
```
sphuber, line 11:
```suggestion
```
Nitpick: I prefer not to have empty lines here. I know there are a lot of those in the codebase, but those were introduced when we automatically removed `from __future__ import print` when removing Python 2 compatibility code. They were followed with an empty line and only the line itself was removed. Writing a script to automatically remove all of them proved trickier than one would think, so I left it for now and am manually removing them whenever I come across one in a file I am editing anyway for a PR.

aiida/orm/implementation/entities.py
```diff
@@ -0,0 +1,354 @@
+# -*- coding: utf-8 -*-
```
sphuber, line 1: I think I like this new file, but maybe it makes sense to also move the `BackendEntity` class itself here, from the `backends` module?

aiida/orm/implementation/entities.py
```diff
@@ -0,0 +1,354 @@
...
+import abc
+
+from aiida.orm.utils.node import clean_value, validate_attribute_extra_key
+
+
+class AttributesBackendEntity(abc.ABC):
+    """Mixin class that adds all methods for the attributes column to an entity"""
+
+    _dbmodel = None
+    dbmodel = None
```
sphuber, line 21: Hmmm, wonder if this is how we should do it. The `_dbmodel` will indeed refer to the actual DbModel instance (wrapped in the ModelWrapper), but `dbmodel` is the property defined in `BackendEntity`. I don't think we should be setting it to `None`. This might get complaints from `pylint` but that is just because it doesn't understand mixins. So I would remove `dbmodel = None` declaration

aiida/backends/sqlalchemy/models/group.py
```diff
@@ -47,6 +47,8 @@ class DbGroup(Base):
     time = Column(DateTime(timezone=True), default=timezone.now)
     description = Column(Text, nullable=True)
 
+    extras = Column(JSONB, default=dict, server_default='{}')
```
sphuber, line 50: Why do we need the `server_default='{}'` here, but not for the `attributes` nor `extras` of `DbNode`?

aiida/backends/djsite/db/migrations/0045_dbgroup_extras.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Migration for adding the extras column to the dbgroup table."""
```
sphuber, line 10:
```suggestion
"""Migration to add the `extras` JSONB column to the `DbGroup` model."""
```

aiida/backends/sqlalchemy/migrations/versions/0edcdd5a30f0_add_extras_to_group.py
```diff
@@ -0,0 +1,38 @@
...
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+# pylint: disable=no-member,invalid-name
+"""Add extras column to the db_dbgroup table.
```
sphuber, line 11:
```suggestion
"""Migration to add the `extras` JSONB column to the `DbGroup` model.
```

aiida/backends/sqlalchemy/migrations/versions/0edcdd5a30f0_add_extras_to_group.py
```diff
@@ -0,0 +1,38 @@
...
+
+def upgrade():
+    """Upgrade: Add the extras column to the 'db_dbgroup' table"""
+    op.add_column(
+        'db_dbgroup', sa.Column('extras', postgresql.JSONB(astext_type=sa.Text()), server_default='{}', nullable=True)
+    )
+
+
+def downgrade():
+    """Downgrade: Drop the extras column to the 'db_dbgroup' table"""
```
sphuber, line 37:
```suggestion
    """Downgrade: Drop the extras column from the 'db_dbgroup' table"""
```

sphuber (changes requested):
Thanks a lot @mbercx looking really good. Just have some suggestions that shouldn't take too much time

Answering some of the questions in your OP:

> Why is the `_aiida_hash` stored as an extra and not as an attribute? I.e.: why do we need the functionality to re-hash a Node?

Good question. We didn't want to add it to the attributes, since the "attributes" are what "define" a node. However, the hash is merely some metadata that describes those attributes, and so is not actually part of it. In reality, the hash would probably have been a separate column like all other metadata, `label`, `ctime`, `mtime` etc., however, it was introduced years after the database models were first designed. We really need a new column or JSONB field to store "internal control values" like this. There are a few of these, metadata that AiiDA uses to have the behavior of the ORM change based on the values of that metadata for a given node. See the discussion in #2231 .

> Might it be better to split this into two PR's? The first one would be: "Add mix-in classes for the attributes and extras methods" and include the first two commits.

Yes! This is a great point. I looked at the first two commits, and this is exactly what we should do, very good. This way we can actually easily see the changes applied to the code before they are moved. Otherwise they get drowned out. For now we can keep this as a single PR to simplify development, but when we are done, you can create a new PR with the first two commits, we merge that and then you rebase this one so we merge the final commit that adds the extras to the `Group`.

> Homogenize the directory structures of the backend subpackage. Why use djsite instead of django? Why are the models of django in backends.djsite.db.models.py, whereas for SqlAlchemy they are spread over different modules in backends.sqlalchemy.models?

This is another artifact of the past. When AiiDA was started, it started literally from a Django default template project, simply because they wanted to use the ORM to work with the database. However, since Django is really designed to create websites, you will see a lot of old references to deploying a website. We have been slowly removing these references, but clearly haven't done everything yet. I have already refactored the `aiida.backends` module heavily over the past few years, but I have left renaming those internal file structures, mostly because they really only affect developers. It would just be (potentially a lot of work) for relatively little benefit. If you really want to do it, we can maybe quickly agree on what should be renamed/restructured and agree on the new naming/structuring an then you can open a new PR.

aiida/orm/implementation/sqlalchemy/nodes.py
```diff
@@ -441,13 +434,9 @@ def extras_keys(self):
...
             yield key
 
-    def _flag_field(self, field):
-        from aiida.backends.sqlalchemy.utils import flag_modified
-        flag_modified(self._dbmodel, field)
-
-    def _flush_if_stored(self):
+    def _flush_if_stored(self, fields):
         if self._dbmodel.is_saved():
-            self._dbmodel.save()
```
sphuber: Will this new implementation work exactly the same for the SqlAlchemy backend? I.e. will `self._dbmodel._flush(fields)` do the same thing by calling `flag_modified` on all the fields?

sphuber: Ok, I think it actually does this in the `_flush` method of the `ModelWrapper` so looks to be all good

aiida/orm/implementation/entities.py
mbercx: I agree. It's not really placed appropriately in `aiida.orm.utils.node`, and moving it shouldn't affect the user. I'll make a new module `aiida.orm.implementation.utils` and move the methods there.

aiida/orm/implementation/sqlalchemy/nodes.py
mbercx: Yes, this is also what I noticed. In both implementations the flag_modified and _dbmodel.save methods are called, so it should work exactly the same.

aiida/backends/sqlalchemy/models/group.py
mbercx: For the new `extras` of the `Group`, we need to add this column with an empty dictionary to the database in the migration. If I don't add `server_default` in the column declaration, the SqlAlchemy tests fail because the model and database schemas do not match (i.e. the `test_model_and_migration_schemas_are_the_same` test fails; see [here](https://github.com/mbercx/aiida-core/runs/1111568831?check_suite_focus=true)). 

Adding the `server_default` setting for the `DbNode` columns causes similar issues in the testing, and more importantly lead to issues migrating the database from the previous schema. The first can be fixed by manually adding `alter_column` statements to the SqlAlchemy migration file (see [here](https://github.com/mbercx/aiida-core/actions/runs/243324072)), but when migrating from the previous schema the migrate process got stuck. I'll see if I can reproduce this issue.

aiida/orm/implementation/entities.py
mbercx: How about `BackendCollection`? Does it fit better in the `entities` module as well, or should it remain in the `backends` one?

aiida/orm/implementation/entities.py
sphuber: I think you did this so we can mark this as resolved

aiida/orm/implementation/entities.py
sphuber: Given that it is a collection of entities and not backends, I think you make a fair point. But I see that you already moved it :+1:

aiida/backends/sqlalchemy/models/group.py
mbercx: Hmm, I no longer seem to be able to reproduce the issue I was having with the SqlAlchemy database migration. So, I'd vote for adding `server_default='{}'` to the `attributes` and `extras` of the `DbNode` model as well, and adding the `alter_column` statements to the SqlAlchemy migration file. Note that this *should* not have any influence for the users. Any node created with either the previous or current schema will set the `attributes` and `extras` to an empty dictionary upon initialization, so it will always be stored with this default in the database. I'd mainly change it for consistency.

Note again that removing the `server_default` everywhere is not an option, since this would mean groups from previous schemas would have a `null` value for the `extras` in the database upon migration. The current methods for the `extras` expect a dictionary in the database, however, so this leads to errors.

sphuber: Adding the `server_default` to the `DbNode` model will require a schema migration though. The `server_default` is included in the schema and so the database will be responsible for generating it. Contrastingly, the `default` is generated by the ORM in Python. By default SqlAlchemy does not populate the defaults of a model upon instantiation, but only when it is persisted. The reason for this slightly unintuitive behavior is that, depending on the nature of the default, it cannot always be _known_ before it is persisted. For consistency, it also does this for normal `default` values, which _could_ be known. There is a listener `from sqlalchemy_utils import force_instant_defaults` that can be called to force these defaults being set, which is what we actually do in `aiida.backends.sqlalchemy.model.__init__.py`. Finally, setting _both_ `default` and `server_default` is not necessary and can only cause problems.

I think then, that it is best to just keep the `default` and forget about the `server_default`. Anyway, generating an empty dictionary in Python and sending it to Postgres, instead of having PostgreSQL take care of it, should not matter much in terms of performance. Together with the force defaults listener, this allows us to have the defaults automatically set upon model instantiation. The only remaining problem you faced was the failing test, but the correct solution there is just to update the automatically generated schema and remove the `server_default` from the `add_column` statement. I think this should answer all your questions, but I may have missed something, in which case please let me know.

I also think that this would be the last open question for this PR correct? Except for the creation of a `ExtrasMixin` for the front-end ORM. I would suggest you make the new PR that refactors the backend classes with the new mixins and you also add the front-end ORM mixin. Then we can merge that, and we can rebase this PR. Sound good?

mbercx: Thanks for the explanation! I agree that setting both the `default` and `server_default` in the `Column` declaration is not necessary (I hadn't noticed the `force_instant_defaults` function though, good to know about this subtlety). However, if I do not add `server_default='{}'` to the `add_column` statement, already existing groups will get a `null` value in the `extras` column upon migration. This means `Group.extras` will return None, which creates issues when you try to e.g. use the `BackendEntityExtrasMixin.set_extra()` method, since this expects the `extras` to be a dictionary. 

One solution would be to accommodate for `null` values in the database, but doing this just because of the migration issue is probably not worth it. So perhaps it's best to just leave the `server_default` for the group extras, both in the `add_column` statement and column declaration?

> I would suggest you make the new PR that refactors the backend classes with the new mixins and you also add the front-end ORM mixin. Then we can merge that, and we can rebase this PR. Sound good?

Yes! I'll start working on a new PR that adds the mixin classes, also for the front-ends.

mbercx:
> Good question. We didn't want to add it to the attributes, since the "attributes" are what "define" a node. However, the hash is merely some metadata that describes those attributes, and so is not actually part of it. In reality, the hash would probably have been a separate column like all other metadata, `label`, `ctime`, `mtime` etc., however, it was introduced years after the database models were first designed. We really need a new column or JSONB field to store "internal control values" like this. There are a few of these, metadata that AiiDA uses to have the behavior of the ORM change based on the values of that metadata for a given node. See the discussion in #2231 .

Thanks for the explanation and context! Maybe I can open a PR at some point to set up a new column metadata column that stores the `_aiida_hash`? In case this information is no longer stored in the `extras`, it might be worthwhile to set the default value for the `extras` to `null` in the database, and adjust the methods of the `BackendEntityExtrasMixin` class to accommodate for this. 

> This is another artifact of the past. When AiiDA was started, it started literally from a Django default template project, simply because they wanted to use the ORM to work with the database. However, since Django is really designed to create websites, you will see a lot of old references to deploying a website. We have been slowly removing these references, but clearly haven't done everything yet. I have already refactored the `aiida.backends` module heavily over the past few years, but I have left renaming those internal file structures, mostly because they really only affect developers. It would just be (potentially a lot of work) for relatively little benefit. If you really want to do it, we can maybe quickly agree on what should be renamed/restructured and agree on the new naming/structuring an then you can open a new PR.

Alright, thanks for the background info! It's not really essential, so I agree that we can leave this for now. If at any point in the future it really starts to bother me, we can discuss this issue again and I'll open a PR. :)

aiida/backends/sqlalchemy/models/group.py
sphuber: > However, if I do not add server_default='{}' to the add_column statement, already existing groups will get a null value in the extras column upon migration. This means Group.extras will return None, which creates issues when you try to e.g. use the BackendEntityExtrasMixin.set_extra() method, since this expects the extras to be a dictionary.

The real fix is to have a different, slightly more complicated strategy:
1. Add the column without the `server_default`
2. Set an empty dictionary for all existing groups (either in through ORM or directly in SQL)

Note that this will work because you also mark the new column as `nullable=True`. Also this does not really make sense if you have a default. So I think we should make it non nullable. This will prevent us accidentally inserting rows with null values, despite setting a default in the Python ORM. Unfortunately, this does make the migration yet again a bit more complex

1. Add the column without the `server_default` and `nullable=True`
2. Set an empty dictionary for all existing groups (either in through ORM or directly in SQL)
3. Update the column to `nullable=False`

You can see the migration `041a79fc615f_dblog_cleaning.py` for an example where a similar problem had to be addressed.

sphuber:
> Thanks for the explanation and context! Maybe I can open a PR at some point to set up a new column metadata column that stores the `_aiida_hash`?

Sure, but the main difficulty will be agreeing on the naming. I am not sure that `metadata` is the best name, because what about other existing columns? Are `ctime`, `mtime`, `user` etc. not also metadata of a node?

> In case this information is no longer stored in the `extras`, it might be worthwhile to set the default value for the `extras` to `null` in the database, and adjust the methods of the `BackendEntityExtrasMixin` class to accommodate for this.

This will require new database schema migrations, but I agree that then this would be appropriate and we simply handle null values in the ORM layer (either backend or frontend).

ramirezfranciscof:
Sorry to barge in, but what exactly is the problem of having the hash in the extras?

mbercx:
> Sorry to barge in, but what exactly is the problem of having the hash in the extras?

I think the problem is that the extras could be inadvertently changed by the user.  See #2231.

In the context of this PR, the issue was that making the extras nullable in the database and dealing with null values in the extras methods didn't make a lot of sense, because most nodes have the `_aiida_hash` stored in the extras, so it wouldn't influence the size of the database much.

sphuber:
> Sorry to barge in, but what exactly is the problem of having the hash in the extras?

As @mbercx said, the extras are intended to be freely used to tag their data with mutable metadata. However, the hash is a property that is computed by AiiDA itself and _used_ by AiiDA to influence its behavior and how the data is treated. This means that it should not be directly changeable by a user and so storing it in the same column is dangerous. At the time we just did it because there was nothing else available. A bad choice in hindsight as we should have just bit the bullet and used a proper solution with the necessary database migrations, but that's life.

ramirezfranciscof:
Maybe a possible solution would be to use a "hiding" strategy? Like have extras that start with "_" be normally hidden from the user (unless he passes a `--all` or `show_hidden=True`).

sphuber:
> Maybe a possible solution would be to use a "hiding" strategy? Like have extras that start with "_" be normally hidden from the user (unless he passes a `--all` or `show_hidden=True`).

They already do start with an underscore, but no no, if they are not supposed to be able to touch it, they shouldn't be part of the same data structure. We should simply move them to another column. The hash is not the only property that we use like this. Also for example whether a code is hidden or not, whether a computer is enabled/disabled etc. We just need to determine the name and add it. Keeping it in extras is just asking for trouble. What happens if they call `node.set_extras_many(..)`. That will override all of them, unless we start adding complex logic to not overwrite those starting with an underscore. But what if the user stored extras starting with an underscore? You see where this is going....

ramirezfranciscof:
Yes, I see, you are right. Do you want to open a new issue to discuss this naming / structure?

sphuber:
> Yes, I see, you are right. Do you want to open a new issue to discuss this naming / structure?

[Already exists](https://github.com/aiidateam/aiida-core/issues/2231)

sphuber:
@giovannipizzi I have added you as a reviewer because although I have reviewed it, I have also added some final changes. I can summarize this PR and the discussion I have had with @mbercx as follows:

 * The goal is to add the `extras` functionality to `Groups`.
 * We define the column as not nullable and set a ORM default to an empty dict.
    o Having a default is useful to guarantee that the model instance always has a dictionary set when loaded
    o It is also to keep consistency with existing database models with similar fields. `DbNode.attributes` and `DbNode.extras` also expect a default.
    o There might be some unnecessary load in database size by storing empty dicts for groups that don't define extras, but this should be minimal. If we were to homogenize this, it should be done in a separate PR.
    o We mark the column as `nullable=False` (explicitly, just for clarity, even though that is the default). This is done because we define a default for the ORM, but this is not set on the schema itself. When making the column nullable, this makes it possible to have rows in the database that do not have a dictionary, making the logic in our ORM more complex.
  * There is some inconsistency in existing models, for example the `DbNode.attributes` and `DbNode.extras` for SqlAlchemy do _not_ define a default, but this should not be addressed in this PR.
  * The functionality is added to front-end ORM through the new `EntityExtrasMixin` that was merged recently in PR #4376 
  * The interface on the `BackendGroup` was already added in PR #4376  (I guess that should have been done in this PR, but what's done is done.)

giovannipizzi (approved):
Ok for me to merge, thanks a lot!

One important thing that is missing is the functionality to import and export also these extras.
I think it's OK to wait and do it directly in the new export implementation #4384, unless it is not ready in a few monhts - @sphuber will open an issue for this (exporting of group extras - this is also partially linked to #4383 )

sphuber (approved):

ltalirz:
> @sphuber will open an issue for this (exporting of group extras

@sphuber Did you ever open an issue for this?
I think this is actually a bit urgent, since the group extras were announced with the 1.4.0 release and so people now start using them, but have no way of sharing the group extras with someone else (if I understand correctly).
E.g. in the LSMO we would very much like to use group extras to further simplify queries but if they cannot be exported that is basically blocking us from doing so.

I would personally vote to still include this feature in the current archive implementation.

sphuber:
Just made it: #4497

PR #4332, opened 2020-08-27
slurm plugin: always raise for non-zero exit code

### LINKED ISSUE #4326: AiiDA scheduler plugin thinks job is finished while it is still running
From time to time, AiiDA will retrieve the files in a calculation folder before the calculation has finished.
When checking the calculation folder using `gotocomputer` at a later point in time, one finds that the calculation actually finished fine.

From a user perspective, this issue is very similar to https://github.com/aiidateam/aiida-core/issues/3404, but we've now established (see below) that there are two different failure modes.
In the case discussed here, AiiDA gets an incorrect job list (which no longer contains the job in question) and therefore starts retrieving it.

===========

@sphuber We ran into this issue on helvetios/fidis.

Now, following up on your suggestion to investigate the failure mode, I have looked at the historical job records of both clusters during that time using `sacct`. In particular, one can query for the job ids and work directories of past jobs as follows:

```bash
# get work dirs of jobs since 2020-07-01 
sacct -S 2020-07-01 -u ongari --format "jobid,jobname%20,workdir%70" > sacct-list
cat sacct-list | grep scratch | grep aiida_run | awk '{print $3}' > directories
cat directories | wc -l  # prints 4705
cat directories | sort -u | wc -l  # prints 4705
```

It turned out that the list of work directories is unique.
If the reason for the failure was that AiiDA submitted the same job twice (causing one or both of them to fail), then we should see the corresponding work directories appear twice in the list.
I.e. in this particular case, I believe the issue occurs at the point where AiiDA queries the scheduler for the job status and gets an incomplete/incorrect answer.

As mentioned during our discussion, one solution of this issue could be to add a fail-safe mechanism when AiiDA gets the information that a job has completed. This could be any of
 * running `squeue` again, just to be sure
 * using an alternative command, that queries the status of that job specifically, such as
   * `scontrol show job <id>` 
   * `sacct -j <id>` 

and only accepting the fact that the job is no longer running, if the output agrees.

Mentioning @danieleongari for info

_Originally posted by @ltalirz in https://github.com/aiidateam/aiida-core/issues/3404#issuecomment-672951815_

ltalirz:
potentially fixes #4326 

Prior to this change, cases where squeue would return a non-zero exit
code but an empty stderr would not lead to a SchedulerError.
This is fixed by adapting the slurm joblist command such that it is
always expected to produce exit code zero, and raising whenever a
non-zero exit code is encountered.

### DIFF
--- a/aiida/schedulers/plugins/slurm.py
+++ b/aiida/schedulers/plugins/slurm.py
@@ -228,7 +228,7 @@ class SlurmScheduler(Scheduler):
             # Verified on slurm versions 17.11.2, 19.05.3-2 and 20.02.2.
             # See also https://github.com/aiidateam/aiida-core/issues/4326
             if len(joblist) == 1:
-                joblist += joblist[0]
+                joblist += [joblist[0]]
 
             command.append('--jobs={}'.format(','.join(joblist)))
 
--- a/aiida/schedulers/plugins/slurm.py
+++ b/aiida/schedulers/plugins/slurm.py
@@ -214,6 +214,19 @@ class SlurmScheduler(Scheduler):
                 if not isinstance(jobs, (tuple, list)):
                     raise TypeError("If provided, the 'jobs' variable must be a string or a list of strings")
                 joblist = jobs
+
+            # Trick: If we are asking for a single job, we append the same job once more.
+            # We do this because squeue returns a non-zero exit code if provided an invalid jobid
+            # (stderr: "slurm_load_jobs error: Invalid job id specified"), which can happen if time has passed since
+            # the job finished and squeue no longer has knowledge of it.
+            # However, this *only* happens when providing a *single* jobid to squeue.
+            # When providing two or more jobids via `squeue --jobs=123,234`, it always returns exit code zero, and
+            # duplicating job ids has no other effect on the output.
+            # Verified on slurm versions 17.11.2, 19.05.3-2 and 20.02.2.
+            # See also https://github.com/aiidateam/aiida-core/issues/4326
+            if len(joblist) == 1:
+                joblist += joblist[0]
+
             command.append('--jobs={}'.format(','.join(joblist)))
 
         comm = ' '.join(command)
@@ -482,21 +495,18 @@ class SlurmScheduler(Scheduler):
         # pylint: disable=too-many-branches,too-many-statements
         num_fields = len(self.fields)
 
-        # I don't raise because if I pass a list of jobs,
-        # I get a non-zero status
-        # if one of the job is not in the list anymore
-        # retval should be zero
-        # if retval != 0:
-        # self.logger.warning("Error in _parse_joblist_output: retval={}; "
-        #    "stdout={}; stderr={}".format(retval, stdout, stderr))
-
-        # issue a warning if there is any stderr output and
-        # there is no line containing "Invalid job id specified", that happens
-        # when I ask for specific calculations, and they are all finished
-        if stderr.strip() and 'Invalid job id specified' not in stderr:
-            self.logger.warning("Warning in _parse_joblist_output, non-empty stderr='{}'".format(stderr.strip()))
-            if retval != 0:
-                raise SchedulerError('Error during squeue parsing (_parse_joblist_output function)')
+        if retval != 0:
+            raise SchedulerError(
+                """squeue returned exit code {} (_parse_joblist_output function)
+stdout='{}'
+stderr='{}'""".format(retval, stdout.strip(), stderr.strip())
+            )
+        if stderr.strip():
+            self.logger.warning(
+                "squeue returned exit code 0 (_parse_joblist_output function) but non-empty stderr='{}'".format(
+                    stderr.strip()
+                )
+            )
 
         # will contain raw data parsed from output: only lines with the
         # separator, and already split in fields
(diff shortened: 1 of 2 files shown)


aiida/schedulers/plugins/slurm.py
```diff
@@ -214,6 +214,19 @@ def _get_joblist_command(self, jobs=None, user=None):
...
                     raise TypeError("If provided, the 'jobs' variable must be a string or a list of strings")
                 joblist = jobs
+
+            # Trick: If we are asking for a single job, we append the same job once more.
+            # We do this because squeue returns a non-zero exit code if provided an invalid jobid
+            # (stderr: "slurm_load_jobs error: Invalid job id specified"), which can happen if time has passed since
+            # the job finished and squeue no longer has knowledge of it.
+            # However, this *only* happens when providing a *single* jobid to squeue.
+            # When providing two or more jobids via `squeue --jobs=123,234`, it always returns exit code zero, and
+            # duplicating job ids has no other effect on the output.
```
ltalirz, line 224: Perhaps clearer like this
```suggestion
            # However, this *only* applies when providing a *single* jobid to squeue.
            # When providing two or more jobids via `squeue --jobs=123,234`, squeue stops caring whether they are valid
            #  and returns exit code zero irrespectively (duplicating job ids has no other effect on the output).
```

aiida/schedulers/plugins/slurm.py
```diff
@@ -482,21 +495,18 @@ def _parse_joblist_output(self, retval, stdout, stderr):
...
-        #    "stdout={}; stderr={}".format(retval, stdout, stderr))
-
-        # issue a warning if there is any stderr output and
-        # there is no line containing "Invalid job id specified", that happens
-        # when I ask for specific calculations, and they are all finished
-        if stderr.strip() and 'Invalid job id specified' not in stderr:
-            self.logger.warning("Warning in _parse_joblist_output, non-empty stderr='{}'".format(stderr.strip()))
-            if retval != 0:
-                raise SchedulerError('Error during squeue parsing (_parse_joblist_output function)')
+        if retval != 0:
```
giovannipizzi, line 498: I would add a comment here pointing to the discussion in the comments of the other function.

aiida/schedulers/plugins/slurm.py
giovannipizzi: I haver provided a change in the suggestion above

aiida/schedulers/plugins/slurm.py
```diff
@@ -214,6 +214,19 @@ def _get_joblist_command(self, jobs=None, user=None):
                 if not isinstance(jobs, (tuple, list)):
                     raise TypeError("If provided, the 'jobs' variable must be a string or a list of strings")
                 joblist = jobs
+
+            # Trick: If we are asking for a single job, we append the same job once more.
+            # We do this because squeue returns a non-zero exit code if provided an invalid jobid
+            # (stderr: "slurm_load_jobs error: Invalid job id specified"), which can happen if time has passed since
+            # the job finished and squeue no longer has knowledge of it.
```
giovannipizzi, line 221: Is it really when squeue has no longer knowledge or it, or even just if it has finished?

aiida/schedulers/plugins/slurm.py
```diff
@@ -214,6 +214,19 @@ def _get_joblist_command(self, jobs=None, user=None):
                 if not isinstance(jobs, (tuple, list)):
                     raise TypeError("If provided, the 'jobs' variable must be a string or a list of strings")
                 joblist = jobs
+
+            # Trick: If we are asking for a single job, we append the same job once more.
+            # We do this because squeue returns a non-zero exit code if provided an invalid jobid
```
giovannipizzi, line 219:
```suggestion
            # We do this because we want to have a reliable way of knowing if the squeue command
            # failed - if it didn't fail, we will assume that jobs missing from the output are completed.
            # If, therefore, there is an error that provides an empty output, and we don't detect it, 
            # AiiDA will end up believing that all jobs are completed and will start retrieving them.
            # We therefore want to rely on the exit code of squeue.
            # However, squeue returns a non-zero exit code if provided a *single* invalid jobid
```

giovannipizzi (changes requested):
Some textual changes suggested

aiida/schedulers/plugins/slurm.py
ltalirz: Thanks Gio.

I merge in your first sentence and try to incorporate the gist of the remaining ones as well.

aiida/schedulers/plugins/slurm.py
ltalirz: You are right - i think it happens as soon as the job no longer is in the queue (will change!).

aiida/schedulers/plugins/slurm.py
ltalirz: Done!

giovannipizzi:
Thanks!
For me it could be merged, but as pointed by codecov, the new lines are not tested. Would you be willing to add one (or two?) new tests to https://github.com/aiidateam/aiida-core/blob/develop/tests/schedulers/test_slurm.py to cover the new code?

ltalirz:
@giovannipizzi fair point - added the test

giovannipizzi (approved):
Thanks!

PR #4334, opened 2020-08-27
Merge master

sphuber:


ltalirz:
Does this not violate the rules given here?
https://github.com/aiidateam/aiida-core/wiki/How-to-make-a-new-release#merge-master-back-into-develop

ltalirz:
I'm closing this just to prevent us from accidentally updating `master`

ltalirz:
ah no, it's `merge/master`

ltalirz:
@sphuber I guess you will first need to update this branch.

Also, just to understand: will the cherry-picked commits be handled gracefully?

sphuber:
> Also, just to understand: will the cherry-picked commits be handled gracefully?

Depends on what you mean by that. By cherry-picking, you are creating a new commit, because you have to. Even though the "diff" maybe the same, the base on which you place it is different and so the commit is different getting a new hash. The `cherry-pick` command is just a utility to apply a set of changes of a commit again onto another base. This means, however, that the original commit on `develop` is now duplicated on `master` with a new hash. You can see that we are merging it back into `develop`, so after merging, it will contain two commits with the same message. However, you can see from the diff of this PR that git realizes that the changes are already there. This merge is really just to make sure that `develop` branch also contains the commit of `master`. This is normal behavior though in such a branching strategy.

The duplicated commit is a small contributor to my motivation to not cherry-pick all bug fixes that are currently on `develop` on the `release/1.3.1` branch. Technically this was possible and we could have a had a bug fix release with all current bug fixes, but since they were not pressing, I didn't feel it necessary to have all the duplicated commits and more complex history. Especially since we will be releasing them soon anyway.

ltalirz (approved):
ok, thanks for the explanation @sphuber 

Looks good to me!

ltalirz:
Hi @sphuber , just a second thought on the topic of patch releases:

In a way, by being "behind" `develop`, patch releases like this one are quite similar to patch releases for old AiiDA versions. In the case of patches of old versions, we actually don't update the `master` branch and release directly from the release branch.

This could be an alternative strategy for the future, which would avoid duplicate commits in the history (the update of the changelog could be cherry-picked).
The downside would be that one can no longer rely on the fact that `master` always point to the latest release... it depends where one puts the priority.

sphuber:
Good point, except the difference I would say is that the fixes on support branches are often no longer applicable to the develop branch and so there really is no point to merging them and actually just creates unnecessary problems. In this case, the particular fix that was the hotfix actually also needed to be fixed in develop. It is true that soon we will merge the PR that replaces tornado with asyncio and the fix will be obsolete, but that should not have necessarily be the case.

PR #4344, opened 2020-09-01
Prepare the code for the new repository implementation

### LINKED ISSUE #4335: Add necessary deprecations ahead of integration of the new repository implementation

sphuber:
Fixes #4335 

In `v2.0.0`, the new repository implementation will be shipped, that
despite our best efforts, requires some slight backwards-incompatible
changes to the interface. The envisioned changes are translated as
deprecation warnings:

 * `FileType`: `aiida.orm.utils.repository` ->`aiida.repository.common`
 * `File`: `aiida.orm.utils.repository` ->`aiida.repository.common`
 * `File`: changed from namedtuple to class
 * `File`: iteration is deprecated
 * `File`: `type` attribute -> `file_type`
 * `Node.put_object_from_tree`: `path` -> `filepath`
 * `Node.put_object_from_file`: `path` -> `filepath`
 * `Node.put_object_from_tree`: `key` -> `path`
 * `Node.put_object_from_file`: `key` -> `path`
 * `Node.put_object_from_filelike`: `key` -> `path`
 * `Node.get_object`: `key` -> `path`
 * `Node.get_object_content`: `key` -> `path`
 * `Node.open`: `key` -> `path`
 * `Node.list_objects`: `key` -> `path`
 * `Node.list_object_names`: `key` -> `path`
 * `SinglefileData.open`: `key` -> `path`
 * Deprecated `contents_only` in `put_object_from_tree`
 * Deprecated `force` argument in
    o `Node.put_object_from_tree`
    o `Node.put_object_from_file`
    o `Node.put_object_from_filelike`
    o `Node.delete_object`

The special case is the `Repository` class of the internal module
`aiida.orm.utils.repository`. Even though it is not part of the public
API, plugins may have been using it. To allow deprecation warnings to be
printed when the module or class is used, we move the content to a
mirror module `aiida.orm.utils._repository`, that is then used
internally, and the original module has the deprecation warning. This
way clients will see the warning if they use it, but use in `aiida-core`
will not trigger them. Since there won't be a replacement for this class
in the new implementation, it can also not be replaced or forwarded.


sphuber:
This PR is in preparation of merging the new file repository implementation of PR #4345 . The new repository unfortunately comes with a few backwards incompatible changes. Some of these are not necessary but are very desirable and important to prevent to propagate original design mistakes and make the code more difficult to understand and maintain. However, some of them are unavoidable:

- Many plugins use `Node._repository._get_base_folder().abspath` to get the absolute filepath of a node's repository and use normal Python file system operations to create and delete files. Obviously this shouldn't have happened because it uses two protected methods, but plugin developers lent this malpractice from our own plugins. Unfortunately, there will be no backwards compatible alternative for this in the new implementation and so will break, not matter what.
- The `Node.open` can now be used without a context manager, but this will start to raise. There may be a trick we can do and provide a wrapper that will automatically create a context if called without, but will significantly complicate the code.
- The `force` keyword form the `Node.put_object_from_*` and `Node.delete_object` methods will have to be removed as this should never be usable by a user. As such, there will also not be a replacement for it.

Despite these problems, during discussions, it was decided to try and make the transition as painless as possible and try to break as few plugins as possible. The idea was to have plugin developers be notified of these upcoming changes and help them create an updated version of the plugin that would work with the new interface.

Questions: I am still not fully clear what the exact plan is. As I understood, we release the changes of this PR in `aiida-core==1.4.0` that should still work with current plugins. We ask developers to test their plugins with this version and address the deprecation warnings and maybe release a new version. Then we merge the new repository and release that in v1.5.0.

Potential problems: in the transition period, until the new plugin versions become available, people will get _a lot_ of deprecation warnings when installing v1.4.0. This is really annoying, but maybe they can be suppressed somehow. However, existing plugins that are currently usually marked as being compatible with `aiida-core~=1.0` will however break with `aiida-core==1.5.0`. Even though there may be newer versions available that _are_ compatible, this is still not really what we should be doing with minor releases.

What is the big argument against just releasing the new repository with v2.0.0. I think the changes are big enough that they merit a new major version and migration of existing code would be relatively minimal, yet to be expected since it is a major version. I really have the feeling that trying our hardest too shoehorn this in a minor version, does not really gain us anything, but has a lot of downsides. Please change my mind :)

greschd:
> What is the big argument against just releasing the new repository with v2.0.0. I think the changes are big enough that they merit a new major version and migration of existing code would be relatively minimal, yet to be expected since it is a major version. I really have the feeling that trying our hardest too shoehorn this in a minor version, does not really gain us anything, but has a lot of downsides. Please change my mind :)

This has the added benefit of setting the expectation that major version bumps may be backwards-compatibility breaking, but nowhere near as complicated to port as 0.X to 1.0. I think that's a reasonable assumption for the users to make, and a standard we should keep ourselves to.

aiida/orm/nodes/node.py
```diff
@@ -703,6 +761,9 @@ def put_object_from_file(self, path, key, mode=None, encoding=None, force=False)
         # order to detect when they were being passed such that the deprecation warning can be emitted. The defaults did
         # not make sense and so ignoring them is justified, since the side-effect of this function, a file being copied,
         # will continue working the same.
+        if force:
+            warnings.warn('the `force` keyword is deprecated and will be removed.', AiidaDeprecationWarning)  # pylint: disable=no-member
```
greschd, line 765: We should probably mention in the deprecation warning that it no longer has any effect.

greschd (commented):
To be honest, I'm not completely sold on the `key` -> `path` and `type` -> `file_type` renaming. What's the rationale behind these?

If it's just "a more appropriate name, but the old one works also" I would prefer sticking with the old names. I'd imagine there are cases where this is the only breaking change with the new interface - and forcing new plugin versions to be released just for that seems kind of silly.

sphuber:
> To be honest, I'm not completely sold on the `key` -> `path` and `type` -> `file_type` renaming. What's the rationale behind these?

The rationale was that when I originally wrote the current interface, it was looking forward exactly to this point where we would have a new "object store"-like repository. Each file stored by some backend, would be uniquely identifiable by some "key". This is why I called the argument in these functions that way. The mistake I made, is that this key is typically autogenerated by the object store and not something the AiiDA user specified. Rather the AiiDA user, through the front-end ORM, thinks in terms of each node having its own repository and the files within it having a normal file system hierarchy. The "key" in this context, is the relative filepath within this hierarchy. With the new implementation, I realized my mistake and so the backend uses "key" for the unique identifier with which the object is stored, but uses "path" as the relative path in the virtual hierarchy. Now, we could choose to keep "key" in the front-end, and just map it to the new names when it calls the backend, but this would make the code more confusing, plus I think that "path" really just makes more sense for users of the front-end, rather than "key" which doesn't make much intuitive sense.

The `type` -> `file_type` renaming is easier. The name `type` is a reserved keyword, and I don't like having this as a property. We do the same in many places, for example `link_type` instead of just `type`.

greschd:
> 
> The rationale was that when I originally wrote the current interface, it was looking forward exactly to this point where we would have a new "object store"-like repository. Each file stored by some backend, would be uniquely identifiable by some "key". This is why I called the argument in these functions that way. The mistake I made, is that this key is typically autogenerated by the object store and not something the AiiDA user specified. Rather the AiiDA user, through the front-end ORM, thinks in terms of each node having its own repository and the files within it having a normal file system hierarchy. The "key" in this context, is the relative filepath within this hierarchy. With the new implementation, I realized my mistake and so the backend uses "key" for the unique identifier with which the object is stored, but uses "path" as the relative path in the virtual hierarchy. Now, we could choose to keep "key" in the front-end, and just map it to the new names when it calls the backend, but this would make the code more confusing, plus I think that "path" really just makes more sense for users of the front-end, rather than "key" which doesn't make much intuitive sense.

Yeah, no argument on one is the better naming choice in isolation (without the existing code). I can sort of see the argument that it would be confusing if they are named differently in the interface vs. the actual implementation - it's still a trade-off, but probably acceptable on balance.

> 
> The `type` -> `file_type` renaming is easier. The name `type` is a reserved keyword, and I don't like having this as a property. We do the same in many places, for example `link_type` instead of just `type`.
> 

Just to be pedantic about language: The `type` is the name of a built-in function, but _not_ a keyword. [Keywords](https://docs.python.org/3/reference/lexical_analysis.html#keywords) _can't_ be used as identifiers, that's a `SyntaxError`. For [built-in functions](https://docs.python.org/3/library/functions.html) or [constants](https://docs.python.org/3/library/constants.html) my general rule is that they should be avoided where they shadow the builtin, but are o.k. (not good, but acceptable) e.g. for methods. 
We do have some other ones in our codebase also - e.g. `all()`, `open()`, `dict()`.

The consistency with `link_type` does make sense. The advantage I see over a bare `type` is more that seeing `some_object.file_type` is just more readable / less confusing than `some_object.type`. The latter only makes sense knowing that `some_object` is of type (hah!) `File`.

Anyway, long story short: I'm fine with the changes as is.

ltalirz:
Just my thoughts the versioning discussion:
In the end, the version numbers we assign do not really matter that much to me (and I can certainly follow the arguments you laid out above). 
What does matter to me is what we do in the code, striving for backwards compatibility and being conservative with introducing breaking changes. 

1. As an AiiDA user working with multiple plugins (as many users do), breaking changes in AiiDA mean than I cannot upgrade until every plugin I use has been updated to be compatible. I.e. we should keep in mind that for any breaking change that we introduce, we are keeping some AiiDA users stuck at older versions for some time that would otherwise be able to upgrade immediately (and reap the benefits).
2. As a plugin developer, ideally I want my plugin to be compatible with the new AiiDA version *without* breaking compatibility with the old version - otherwise I will need to start maintaining an AiiDA-1 and an AiiDA-2 branch of my plugin and create extra work for myself.
   If this backwards-compatibility layer can be written once inside aiida-core, I believe this saves a great deal of work overall compared to a situation where every plugin developer has to fend for him or herself (even if it may be a bit tedious for the core developers). 

So, what may look like shoehorning can have some real-world benefits.
I'm fine with calling the version with the repository 2.0 as long as we don't use this as an excuse to "save" efforts on providing backwards compatibility where we can.

sphuber:
> Just to be pedantic about language: The `type` is the name of a built-in function, but _not_ a keyword. [Keywords](https://docs.python.org/3/reference/lexical_analysis.html#keywords) _can't_ be used as identifiers, that's a `SyntaxError`. For [built-in functions](https://docs.python.org/3/library/functions.html) or [constants](https://docs.python.org/3/library/constants.html) my general rule is that they should be avoided where they shadow the builtin, but are o.k. (not good, but acceptable) e.g. for methods.
> We do have some other ones in our codebase also - e.g. `all()`, `open()`, `dict()`.

Thanks for the correction, that is what I should have said. All though I agree that shadowing a builtin as an argument or property is not the worst, I still really try to stay away from it and in any case it allows you to make it more specific and easier to read.

sphuber:
> In the end, the version numbers we assign do not really matter that much to me (and I can certainly follow the arguments you laid out above).

I think it _does_ make a big difference what version we use, after all the versioning scheme is a contract that we enter into. This is why we tell plugin users to declare the AiiDA dependency as `aiida-core~=1.0`. This ensures that `aiida-core==2.0.0` will not be automatically installed because the _expectation_ is that it is not backwards compatible.

Still, I take your arguments and agree that we should make a real effort to find the best solution taking everything into account. That is exactly what I am doing here. I have already spent significant effort to map out all the incompatibilities, which _can_ and which _cannot_ be avoided and present the tradeoffs, so I don't think I am just trying to get off easy.

Ultimately, the fact that the majority of current plugins will break _regardless_ of what we do, makes it moot to try and keep everything else fully backwards compatible. The way I see it: we will break things anyway, there is no way around it, so we have to just make it v2.0.0.  But that is okay, this change will be a huge change in design and performance and so if this doesn't merit a v2, what will. On top of that, it will be one year after v1, which is not to soon, and the migration path from v1 to v2 will be minimal. I have already prepared the PR on `aiida-quantumespresso`, which is one of the bigger plugins out there, and as you can see the required changes are really minimal.

So I think with that, I have made up my mind and think we should just go with releasing the new repository in v2.0. This may takes us a bit longer to release then, but given the size of the changes, that may be a good thing and give us some more time to really test it thoroughly and design a performant export import system along with it.

greschd:
>     2\. As a plugin developer, ideally I want my plugin to be compatible with the new AiiDA version _without_ breaking compatibility with the old version - otherwise I will need to start maintaining an AiiDA-1 and an AiiDA-2 branch of my plugin and create extra work for myself.
>         If this backwards-compatibility layer can be written once inside aiida-core, I believe this saves a great deal of work overall compared to a situation where every plugin developer has to fend for him or herself (even if it may be a bit more tedious for the core developers).

Regarding this point:
- For the `key` to `path` renaming, it's absolutely possible to be compatible with both versions at the same time: Simply pass the parameter as a positional arguments. Probably we should advise plugin developers to do this.
- This is actually more difficult for the `File`. You could of course do something like ``getattr(file_obj, 'file_type', getattr(file_obj, 'type', None))`` - but then you keep seeing the deprecation warning and it's quite ugly. Maybe we can just keep the (deprecated) ``__iter__`` around for a bit longer. Although I agree with @sphuber that this is not used much in plugins, so there probably isn't much impact.

Another important thing to consider is that the last 1.X release will be compatible with both the new and old interfaces - so it's not like we're forcing people into 2.X if they want a plugin that uses the new interface.

greschd:
About the release process: It might make sense to publish a release candidate for `2.0`, and ask plugin developers to test it.

greschd (approved):

sphuber:
Putting this on hold for now to still give @giovannipizzi a chance to respond. I want to make sure we are all on board with whatever final decision. Also, I want to see if we should keep the deprecation warning on the module level of `aiida.orm.utils.repository` and the `Repository` class. This warning will now be triggered a lot, also by `aiida-core` internally, because we don't have the replacement yet. I have searched on Github to see if the plugins use this class directly, but I do not think so. I would say that in that case it is better to just keep the deprecation in the docstrings but not actually emit a warning.

Also, if we decide to remove in v2.0.0, I will need to update the deprecation warnings that currently say v1.5

aiida/orm/nodes/node.py
sphuber: Actually, looking at the code, it _does_ currently still have effect. Sure, when it will be removed it no longer has an effect, but I think that would be obvious no :sweat_smile:  Is there something else you meant?

aiida/orm/nodes/data/singlefile.py
```diff
@@ -56,17 +56,24 @@ def filename(self):
         """
         return self.get_attribute('filename')
 
-    def open(self, key=None, mode='r'):
+    def open(self, path=None, mode='r', key=None):
         """Return an open file handle to the content of this data node.
 
+        :param path: the relative path of the object within the repository.
         :param key: optional key within the repository, by default is the `filename` set in the attributes
```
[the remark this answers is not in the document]
sphuber: Do you mean to add this in all docstrings of deprecated parameters or just this one? And just the `(deprecated)` or more info such as in which version it was deprecated and when it will be removed

aiida/orm/nodes/node.py
greschd: Ah, yes. I was probably thinking about the branch where the new repository is already implemented.

aiida/orm/nodes/data/singlefile.py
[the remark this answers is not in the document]
greschd: Yeah, I think it should go on all deprecated parameters. Not sure if mentioning when it will be removed is necessary - I just want to avoid people looking at the docstring when implementing new stuff to use it.

But yeah, `(deprecated since v1.4, to be removed in v2.0)` is also fine.

sphuber:
@greschd could you please tell me what you think about the deprecation of the `aiida.orm.utils.repository` module and the `Repository` class within it. Since we don't have the replacement yet, we use it a lot in `aiida-core` itself and so the user is spammed with deprecation warnings. It might be possible to shadow the module and class and use that internally and keep the deprecation warning. Except this might be a lot of work and I performed a search in the plugin repos and didn't find any uses of it. Do you think it is OK to drop the deprecation warnings? In the migration guide that I will write, I will of course mention that these resources will be removed, just in case someone is using it.

sphuber:
> As for how the module could be shadowed: Wouldn't it be enough to move the actual code into e.g. `aiida.orm.utils._repository`, and in the `aiida.orm.utils.repository` only leave the deprecation warning and `from ._repository import *`?
> Of course the internal uses would have to be switched from `.repository` to `._repository`, but that seems like a fairly straightforward search / replace.
> 
> For the `Repository` class it's a _bit_ more involved, but we could put a shim like
> 
> ```python
> from ._repository import Repository as _Repository
> 
> class Repository(_Repository):
>     def __init__(self, *args, **kwargs):
>         # deprecation warning
>         super().__init__(*args, **kwargs)
> ```
> 
> into the `aiida.orm.utils.repository` module.
This is indeed what I was thinking off as well.

> I think shadowing the module would be the ideal solution - I don't think it'd be that complicated, but if it is removing the deprecation warnings seems better.

Now that I have spent so much time on this anyway, may as well go the full length. Thanks for the input :+1:!

greschd:
Just for my understanding: What will the `Repository` class be replaced with in the new repository implementation? Isn't https://github.com/sphuber/aiida_core/blob/fix/3445/disk-object-store-repository/aiida/repository/repository.py#L19 a mostly-compatible replacement? 
I'm not sure it even makes sense to call that "deprecating" the `Repository` class - isn't the expectation that code replacing `aiida.orm.utils.repository.Repository` with `aiida.repository.Repository` should mostly just work?

sphuber:
> Just for my understanding: What will the `Repository` class be replaced with in the new repository implementation? Isn't https://github.com/sphuber/aiida_core/blob/fix/3445/disk-object-store-repository/aiida/repository/repository.py#L19 a mostly-compatible replacement?
> I'm not sure it even makes sense to call that "deprecating" the `Repository` class - isn't the expectation that code replacing `aiida.orm.utils.repository.Repository` with `aiida.repository.Repository` should mostly just work?

Not quite, because the `aiida.repository.Repository` class will not be part of the public interface, in the sense that it should not be used by users of `aiida-core` since it does not contain the mutability rules imposed on nodes. It will be used exclusively for internal purposes. The user interface , that the `Repository` class provides currently, will be provided by the `aiida.orm.nodes.repository.NodeRepository` mixin, which does implement all mutability rules and provides an interface to the entire file repository but with a virtual hierarchy on a per-node basis

greschd:
Ok, maybe we should make it a bit more obvious that it's not part of the public interface then (when it's added, not in this PR)?

The potential confusion I'm seeing is that `aiida.orm.utils.repository` tells users to change it to `aiida.repository` - if they do _only_ that, they'll get `aiida.repository.Repository`.

sphuber:
> Ok, maybe we should make it a bit more obvious that it's not part of the public interface then (when it's added, not in this PR)?

I will just not make the `aiida.repository.repository.Repository` class importable from `aiida.repository`. According to our policy, that means it is not part of the public API.

> The potential confusion I'm seeing is that `aiida.orm.utils.repository` tells users to change it to `aiida.repository` - if they do _only_ that, they'll get `aiida.repository.Repository`.

This is a good point. I put it there because this is true for `File` and `FileType`. They _have_ been moved to `aiida.repository.common` and _will_ be importable from `aiida.repository` since they are in intended to be public API. If I change the string to `this module is deprecated. `File` and `FileType` should be imported from `aiida.repository`.` that would work, right?

greschd:
Yeah, I think that'd be fine. 

As an aside: In general I'm of the opinion that any non-public API should have a leading underscore somewhere in its name, because it's way too easy to accidentally use private API otherwise. 
But that's an entirely separate discussion, and would probably be quite a pain to retroactively introduce into AiiDA.

sphuber:
> As an aside: In general I'm of the opinion that any non-public API should have a leading underscore somewhere in its name, because it's way too easy to accidentally use private API otherwise.
> But that's an entirely separate discussion, and would probably be almost impossible to retroactively introduce into AiiDA.

This is a difficult one: while I see the merit of using underscore to indicate non-publicness, I feel it is often abused. Very early AiiDA code also often used this concept, prefixing for example class methods with an underscore if they shouldn't be called by "clients", however, should be used by the code internally. This is not what the underscore is designed to mean in Python. It also leads to a lot of `pylint: disable=protected-access` as a result. (Or were you merely talking about prefixing module names that should not be public?)

To still give some kind of a handle to clients to easily know what is public, in `v1.0.0` we added the rule of thumb that "anything that can be imported from a module on the second level, is fair game. So anything that can be imported as `from aiida.something import somewhat`. I feel this is a relatively elegant and understandable scheme, that prevents peppering the code with lots of leading underscore, which I find horribly ugly. I know that code aesthetics doesn't carry the most weight for everyone, but for me it does :sweat_smile:

greschd:
Ah, I see the conundrum: we would need three levels of access, "public", "aiida-core internal" and "private" (this class / module only).
Not sure there's a good solution for this.

Isn't `aiida.repository.Repository` public by that metric?

sphuber:
> Isn't `aiida.repository.Repository` public by that metric?

Currently it is because I include `aiida.repository.repository` all in `aiida.repository.__init__` but I will remove it. Since the `Repository` class is in `aiida.repository.repository` it won't be importable from the second level.

sphuber:
@giovannipizzi are you happy with us merging this in this state? All deprecations are in and marked as being removed for `v2.0.0` conforming to the current plan of releasing the new repository with that major release.

ltalirz:
Just one last comment from my side:

Correct me if I'm getting this wrong, but by marking these new deprecations as "to be removed with 2.0", i.e. the release coinciding with the new repo (instead of a later release, say, v3.0), we are indeed limiting their use to a temporary warning mechanism.

If we remove them with 2.0, it means that no plugin will be spared from having to update because of the backwards compatibility introduced here, and when 2.0 has dropped, all plugin developers will need to implement backwards compatibility on their own inside their plugin.

To me, this goes a bit against the arguments laid out [here](https://github.com/aiidateam/aiida-core/pull/4344#issuecomment-685150912), so I tend to believe in order to be truly useful we should be keeping some of this backwards compatibility a while *after* the new repository mechanism becomes available.

sphuber:
> Correct me if I'm getting this wrong, but by marking these new deprecations as "to be removed with 2.0", i.e. the release coinciding with the new repo (instead of a later release, say, v3.0), we are indeed limiting their use to a temporary warning mechanism.

This is what we have been doing all along. We have added many deprecation warnings in v1.0 minor versions that are slated to be removed in v2.0. This is common practice. Even by adding these warnings in v1.4 now, we will most likely not be releasing v2.0 until January 2021. This will give developers at least a 4 month notice. Now we can argue if we think this is enough, but I think it does.

> If we remove them with 2.0, it means that no plugin will be spared from having to update because of the backwards compatibility introduced here, and when 2.0 has dropped, all plugin developers will need to implement backwards compatibility on their own inside their plugin.

Unfortunately, this is the case regardless of how we approach things. If we want to implement the new repository, we will _have_ to break the majority of plugins. There simply is no backwards compatible deprecation path of the `node._repository._get_base_folder().abspath` that almost all of them use. Now you could say "well, you are using a doubly protected method, you shouldn't have done that. However, that is 1) our fault because our plugins did this and people have just copied it, but most importantly 2) this changes nothing to the situation. The new repo will still break the majority of plugins. The only solution would be to abandon the new repository, which I don't think is a viable option.

> To me, this goes a bit against the arguments laid out [here](https://github.com/aiidateam/aiida-core/pull/4344#issuecomment-685150912), so I tend to believe in order to be truly useful we should strive to keep at least some of the backwards compatibility introduced here for a while _after_ the new repository mechanism becomes available.

As explained above, we are forced to break backwards compatibility. That being said, even if we could provide a backwards compatible version, and we'd drop it in v3.0 instead of v2.0, at some point the plugins will have to be updated. There is no fundamental difference between doing this in v2 or v3. The only difference is time.

greschd:
One more question about the upgrade path: Are we planning on back-porting critical fixes into 1.4 for some time after 2.0 is released, or no?

sphuber:
> One more question about the upgrade path: Are we planning on back-porting critical fixes into 1.4 for some time after 2.0 is released, or no?

This is something that I definitely am willing to do if we find that many plugins won't have migrated in a reasonable time and there are critical bugs.

aiida/orm/nodes/data/singlefile.py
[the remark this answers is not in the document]
giovannipizzi: Is this going to be done in the end?

ltalirz:
Thanks @sphuber for your comments!

>  There is no fundamental difference between doing this in v2 or v3. The only difference is time.

In the end, time is all we have on this planet, so I would say it does matter ;-) But I wasn't aware that the v2.0 release is planned for 2021, I thought it was sooner. That would make it a bit easier for plugin developers

I am of course aware of the issue with the `_repository._get_base_folder().abspath` which we cannot work around, but I'm not sure it is accurate to say that "almost all" plugins use it (we found 10 packages [1], there are >50).
My question would be: how many plugins could be spared from having to update by the backwards compatibility introduced in 1.4? If the number is too low, dropping the backwards compatibility in 2.0 makes sense. 

Then, there is the question of supporting both aiida 1.x and 2 in the same plugin. For your exemplary [PR to aiida-qe](https://github.com/aiidateam/aiida-quantumespresso/pull/549/files) - will you be able to merge it after v1.4 is released, i.e. is the idea that it will work with both aiida v1.4 and v2.0?
If that is the case - great. Being able to easily support both the 1.x and 2.x lines to me is even more important than the previous point.


[1] here the list for the record
* aiida-bigdft
* aiida-ddec
* aiida-orca
* aiida-raspa
* aiida-siesta
* aiida-quantumespresso
* aiida-gaussian
* aiida-cusp
* aiida-tcod
* aiida-porousmaterials

aiida/orm/nodes/data/singlefile.py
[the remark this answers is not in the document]
sphuber: It _is_ already done

sphuber:
> For your exemplary [PR to aiida-qe](https://github.com/aiidateam/aiida-quantumespresso/pull/549/files) - will you be able to merge it after v1.4 is released, i.e. is the idea that it will work with both aiida v1.4 and v2.0?
> If that is the case - great. Being able to easily support both the 1.x and 2.x lines to me is even more important than the previous point.

Yes, after merging those changes. The code will be compatible with 1.x and with the new repo of 2.0

giovannipizzi:
Hi, I've looked at the code and at the thread.

I think this can be merged, with the following comments:
- we should add the note that `key` is deprecated also in the `:param key:` lines, as @greschd was suggesting? I unresolved that thread. If you all agree it's not needed, I'm ok.
- we don't have a warning for `open` *not* used as a context manager. Is this possible to do in your opinion? (I'll think a bit more about this, but it would be nice to have it also in this PR as it's probably the other thing that we're going to deprecate)
- I think that in the worst plan for plugin developers if there isn't something they can write for both versions is to release one that has the dependency ">=1.4<2" and, when we will be having at least a beta of v2, to ">=1.4,<3".
- I didn't check out locally the code and tried to run it, so I don't know if warnings might still be issued (e.g. is it possible that the old `aiida.orm.repository` module gets autoimported?

sphuber:
> Hi, I've looked at the code and at the thread.

Thanks for going through it carefully.

> I think this can be merged, with the following comments:
> 
>     * we should add the note that `key` is deprecated also in the `:param key:` lines, as @greschd was suggesting? I unresolved that thread. If you all agree it's not needed, I'm ok.

It is added, just as a `.. deprecated::` marker in the docstring, instead of on the same line as the `:param:` but they should all be there. Same goes for all others `force` and `contents_only`.

>     * we don't have a warning for `open` _not_ used as a context manager. Is this possible to do in your opinion? (I'll think a bit more about this, but it would be nice to have it also in this PR as it's probably the other thing that we're going to deprecate)

This is correct. I am not sure if it is possible but haven't tried. I will try a bit and see if it is possible and if so I will open it.

>     * I didn't check out locally the code and tried to run it, so I don't know if warnings might still be issued (e.g. is it possible that the old `aiida.orm.repository` module gets autoimported?

You can check on of the builds on GHA. The warnings are reported at the end by `pytest`. I made sure that the deprecation warnings are not triggered by internal code (except by testing code, which is ok). There should in principle (barring something I have missed) be no cases of internal use of deprecated code.

giovannipizzi:
Thanks for the answers.

Regarding the deprecation for `open` not used as a context manager:
Here is a possible implementation of a temporary-phase method (warning only when calling `.read()` for now, but that should be always called, in practice, if you open something, right? - we can decide to extend to other methods like seek, tell, and other similar methods, but it's probably less critical).
And then I also show how it should become in 2.0 and that it indeed works as expected (@sphuber does the 2.0 implementation match your other branch?)

It's a bit hacky, but should do the job I think (and it works for nested context managers, even if I don't know if anybody would ever do that for our `Node.open` method):
```python
from contextlib import contextmanager
import warnings

_marker = object()

class EnsureContextMan:
    def __init__(self, fileobj):
        self._fileobj = fileobj
        self._in_context_man = 0

    def __enter__(self):
        self._in_context_man += 1
        return self._fileobj.__enter__()

    def __exit__(self, type, value, traceback):
        self._in_context_man -= 1
        return self._fileobj.__exit__(type, value, traceback)

    def read(self, *args, **kwargs):
        if not self._in_context_man:
            warnings.warn("Deprecated, use `open` in context manager (`with open() as f: ...`")
        return self._fileobj.read(*args, **kwargs)

    def __getattr__(self, name, default=_marker):
        try:
            return getattr(self._fileobj, name)
        except AttributeError:
            if default is _marker:
                raise
            return default


class NodeTransitionPeriod:
    def open(self, path):
        return EnsureContextMan(open(path))


class Node2point0:
    @contextmanager
    def open(self, path):
        yield open(path)



n = NodeTransitionPeriod()

print('AAA1')
f = n.open('test.txt')
f.seek(0)
print(f.read())
f.close()
print('AAA2')

print('BBB1')
with n.open('test.txt') as f:
    f.seek(0)
    print(f.read())

print('BBB2')

print('CCC1')
try:
    f.unknown_method
    raise AssertionError("Should have raised (unknown_method)!")
except AttributeError:
    # It is the expected behavior
    pass
print('CCC2')




n = Node2point0()

print('DDD1')
f = n.open('test.txt')
try:
    print(f.seek(0))
    raise AssertionError("Should have raised (seek)!")
except AttributeError:
    # It is the expected behavior
    pass

try:
    print(f.read())
    raise AssertionError("Should have raised (read)!")
except AttributeError:
    # It is the expected behavior
    pass
print('DDD2')

print('EEE1')
with n.open('test.txt') as f:
    f.seek(0)
    print(f.read())

print('FFF2')

print('GGG1')
try:
    f.unknown_method
    raise AssertionError("Should have raised (unknown_method)!")
except AttributeError:
    # It is the expected behavior
    pass
print('GGG2')
```
When run with a file `test.txt` containing the content `content\n`, it prints:
```
AAA1
tmp.py:21: UserWarning: Deprecated, use `open` in context manager (`with open() as f: ...`
  warnings.warn("Deprecated, use `open` in context manager (`with open() as f: ...`")
content

AAA2
BBB1
content

BBB2
CCC1
CCC2
DDD1
DDD2
EEE1
content

FFF2
GGG1
GGG2
```

P.S.: might need some more testing to make sure I'm not forgetting something important. E.g. I'm not duplicating `__setattr__` but I don't think it's needed?

greschd:
> warning only when calling `.read()` for now, but that should be always called, in practice, if you open something, right?

Not necessarily, you can also iterate directly over it, like 

```python
with some_obj.open() as file_handle:
    for line in file_handle:
        # do something
```

sphuber:
> > warning only when calling `.read()` for now, but that should be always called, in practice, if you open something, right?
> 
> Not necessarily, you can also iterate directly over it, like
> 
> ```python
> with some_obj.open() as file_handle:
>     for line in file_handle:
>         # do something
> ```
> 
> And then there's also `readline` and `readlines`.

And to add to that, people may have also been using it to `write`. In my new repo branch, there is a check on the mode and it will only accept `r` and `rb`. Thinking about this, should probably also add a warning to this for `open`.

greschd:
> And to add to that, people may have also been using it to `write`. In my new repo branch, there is a check on the mode and it will only accept `r` and `rb`. Thinking about this, should probably also add a warning to this for `open`.

So presumably the use case for this is opening e.g. an unstored `SinglefileData`, writing to it, then storing it. Is there a new way of doing this?

sphuber:
> > And to add to that, people may have also been using it to `write`. In my new repo branch, there is a check on the mode and it will only accept `r` and `rb`. Thinking about this, should probably also add a warning to this for `open`.
> 
> So presumably the use case for this is opening e.g. an unstored `SinglefileData`, writing to it, then storing it. Is there a new way of doing this?

They could have used it for any file writing, just instead of using the `put_object_from_*` methods. We simply forgot to add the check in `1.0.0`, but `open` was never supposed to be used for writing. Anyway, for `SinglefileData` these methods shouldn't be used, but rather the dedicated `set_file` method. The methods should maybe be blocked for that plugin, now that I think of it, but I don't think anyway uses it. Anyway, I think there is a check in the `store` to make sure there is only one file written to the repo and its filename matches that stored in the attributes.

greschd:
> They could have used it for any file writing, just instead of using the `put_object_from_*` methods. We simply forgot to add the check in `1.0.0`, but `open` was never supposed to be used for writing.

Yeah, I also just realized one could write to an already-stored file in that way. Not ideal 😅

sphuber:
I have added two more deprecation warnings to `open` and `get_object_content` whenever a mode is used that is not `r` or `rb`. The only open question, as far as I can see here, is to have a deprecation warning printed as soon as `open` is used without a context manager. @giovannipizzi thanks for your solution, but since it relies on a method being called on the object returned by `open` and we don't know which methods user code will call, just implementing this for `read` does not fully address the concern. Which other methods should we implement before we are happy to merge this?

greschd:
Another option to catch the case without context manager would be putting a check into the finalizer of the object returned by `open()` - if it is being destroyed without ever having been `__enter__`ed, it's probably used without context manager.

The drawback is that this would warn from an undeterministic place, and there's no _guarantee_ that `__del__` is called, either. With a good enough warning message, it could probably be a net win though.

giovannipizzi:
I like @greschd idea. As he says, the important thing is that the error message is clear enough (maybe one could print also the UUID of the node, or things like this, beside a clear error message).

Maybe, you can also put a similar check in `close()`? (there is a slightly higher probability that people call open and close, rather than open and let be garbage-collected). I would keep both, just add close as well to put the warning closer to the relevant code, rather than the closing of the verdi shell or the verdi run command

aiida/orm/nodes/data/singlefile.py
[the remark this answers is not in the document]
ramirezfranciscof: Shouldn't we also move the `by default is the 'filename' set in the attributes` from `key` to `path`?

aiida/orm/nodes/node.py
```diff
@@ -630,67 +630,187 @@ def extras_keys(self):
...
+                'keyword `key` is deprecated and will be removed in `v2.0.0`. Use `path` instead.',
+                AiidaDeprecationWarning
+            )  # pylint: disable=no-member
+            path = key
 
-    def open(self, key, mode='r'):
+        return self._repository.list_object_names(path)
+
+    def open(self, path=None, mode='r', key=None):
         """Open a file handle to an object stored under the given key.
```
ramirezfranciscof, line 677: Maybe `Open a file handle to an object stored under the given path.`? (or node local path or something like that, but changing the `key` reference to `path`)

ramirezfranciscof: So, there are a couple of instances that you changed the docstrings to add the new param and deprecation warnings but the general description of the method still makes a reference to the old keywords. I didn't want to single all of them one by one in case maybe I'm missing something and these weren't supposed to be changed, but I can keep checking for this if you agree these need to be adapted as well.

sphuber: Thanks, I will go through the docstrings once more to correct those.

sphuber:
> modifications to transparently expose the `IOWrapper` needed

Any suggestions on how to accomplish this, without breaking the usual interface? Not sure how to fully transparently forward all calls to an instance of an object through a wrapper.

greschd:
Yeah - I think we just need to add a `__getattr__` to the object returned by `open` that forwards to the `IOWrapper`, like so:

```python
import warnings

class WarnWhenNotEntered:
    def __init__(self, fileobj, name):
        self._fileobj = fileobj
        self._name = name
        self._was_entered = False

    def __del__(self):
        if not self._was_entered:
            warnings.warn(
                f"Object {self._name} used without context manager.",
                DeprecationWarning)

    def __enter__(self):
        self._was_entered = True
        return self._fileobj.__enter__()

    def __exit__(self, *args):
        self._fileobj.__exit__(*args)

    def __getattr__(self, key):
        return getattr(self._fileobj, key)


# Object that is explicitly destroyed
x = WarnWhenNotEntered(open('foo.txt', 'w'), 'foo.txt')
x.write('test1\ntest2')
x.write('test')
x.close()
del x

# Object that is used with context manager.
with WarnWhenNotEntered(open('foo.txt', 'r'), 'foo.txt') as file:
    print(file.readlines())


# Object that is used with context manager.
with WarnWhenNotEntered(open('foo.txt', 'r'), 'foo.txt') as file:
    for line in file:
        print(line)

# Not explicitly deleted, warning will show during interpreted shutdown
z = WarnWhenNotEntered(open('baz.txt', 'w'), 'baz.txt')

print('----- interpreter shutdown begins -----')

```

sphuber:
@ramirezfranciscof I have addressed your comment on the docstrings, hopefully I got them all now.

@greschd and @giovannipizzi I have added the wrapper for the `open` method.

This should now be ready for final review. I will double check that this interface will work with the interface I have in the new implementation.

aiida/orm/nodes/node.py
```diff
@@ -328,67 +368,187 @@ def mtime(self):
...
+
+        if path is None:
+            raise TypeError("open() missing 1 required positional argument: 'path'")
+
+        if mode not in ['r', 'rb']:
+            warnings.warn("from v2.0 only the modes 'r' and 'rb' will be accepted", AiidaDeprecationWarning)  # pylint: disable=no-member
 
-    def get_object(self, key):
-        """Return the object identified by key.
+        return WarnWhenNotEntered(self._repository.open(path, mode), repr(self))
```
greschd, line 439: Naively I'd have put the `WarnWhenNotEntered` mechanism at the level of `Repository.open` - but if only `Node.open` is a public interface, maybe this is fine also.
I'll leave that up to you to decide @sphuber.

aiida/orm/nodes/node.py
```diff
@@ -36,6 +36,46 @@
 _NO_DEFAULT = tuple()
 
 
+class WarnWhenNotEntered:
```
greschd, line 39: Since this should be completely internal, maybe naming it `_WarnWhenNotEntered` would be appropriate?

aiida/orm/nodes/node.py
```diff
@@ -36,6 +36,46 @@
...
+    def __getattr__(self, key):
+        if key == '_fileobj':
+            return self._fileobj
+        return getattr(self._fileobj, key)
+
+    def __del__(self):
+        self._warn_if_not_entered('del')
+
+    def read(self, *args, **kwargs):
+        self._warn_if_not_entered('read')
```
greschd, line 71: Nice, I like that it will warn as soon as it's being used (at least for the most commonly-used methods) 👍

greschd (approved):

aiida/orm/nodes/node.py
sphuber: The `Repository` is not public interface. If users do use that, they will get another deprecation warning first, so they will eventually also see the `open` warning

aiida/orm/nodes/node.py
sphuber: I am not a fan of prefixing resources with an underscore to indicate publicness of the API. We also don't maintain this currently in the code base, so I didn't for consistency. Since I didn't add it to the `__all__` by definition, according to our rules, it is not part of the public API.

ramirezfranciscof:
@sphuber yes, I now couldn't detect any more of these. Thanks!

giovannipizzi (approved):

ltalirz:
@sphuber et al. We're now getting the deprecation warnings concerning the context manager, e.g. 
```
AiidaDeprecationWarning: `read` used without context manager for <CifData: uuid: 494a631b-3e29-4a85-904d-48fc217e7e1b (unstored)>. This will raise starting from `aiida-core==2.0.0`
...
/home/daniele/aiida1/aiida_core/aiida/orm/nodes/node.py:53: AiidaDeprecationWarning: `del` used without context manager for <CifData: uuid: 494a631b-3e29-4a85-904d-48fc217e7e1b (unstored)>. This will raise starting from `aiida-core==2.0.0`
```

From these warnings alone it is quite difficult to figure out
 * where the problem is  (who called 'open' - the line mentioned is just where the warning is raised)
 * how it should be fixed (basic instructions for refactoring existing code)
 * why this is a problem (why can one not use read() outside a context manager)

Would you guys mind providing some answers to these points; I'm happy to make a PR to fix this

greschd:
> From these warnings alone it is quite difficult to figure out
> 
>     * where the problem is  (who called 'read' - the line mentioned is just where the warning is raised)

I guess here we'll have to use `inspect` to get the outer frame.

> 
>     * how it should be fixed (it should include basic instructions for refactoring existing code)

The code should be changed such that bare
```python
file_handle = node.open(...)
<...>
file_handle.read()
<...>
file_handle.close() # hopefully ;)
```
is changed to 
```python
with node.open(...):
    <...>
    file_handle.read()
    <...>
```

> 
>     * why this is a problem (why can one not use read() outside a context manager. this type of information is often not present in such warnings, but in this particular case I find it advisable to include at least some hint, since this will likely entail tedious fixes for many plugins and people should know what is causing this extra work)

In the repository implementation, `.open()` no longer returns a "readable" object - only a context manager that will give you a readable object when ``__enter__``ed.

@giovannipizzi could probably tell you more about the _why_ of this.

ltalirz:
Thanks for the quick reply @greschd ! This should be easy enough to incorporate into the warning.

greschd:
As a quick solution, I think a global search for `.open(` and checking if it's in a context manager will show you the relevant places quite quickly.

greschd:
And probably we also want to have a write-up of "how to migrate" somewhere online. We could also link to that in the warning.

sphuber:
I started a wiki page: https://github.com/aiidateam/aiida-core/pull/4344#issuecomment-705670767
We can start to add backward incompatible changes for v2.0 there as well as migration guides for the more involved changes

greschd:
@sphuber I think the link should be https://github.com/aiidateam/aiida-core/wiki/AiiDA-2.0-plugin-migration-guide

greschd:
@sphuber the migration guide looks great! One comment: I think it would make sense to add a paragraph detailing how code can be written to be compatible with both 1.x and 2.x.
The two things I can think of there is using `key` / `path` as a _positional_ argument, and always using a context manager.

sphuber:
Good point. Feel free to add things yourself if you have the time. I think this is a document that we can all maintain as we go towards 2.0

ltalirz:
Opened https://github.com/aiidateam/aiida-core/pull/4434

PR #4351, opened 2020-09-03
New item/attr setter for Dict nodes

### LINKED ISSUE #2883: Setting attributes for unstored Dict nodes fails silently
I'm confused by a behaviour of the `Dict` class. I want to update an unstored node `d = Dict()`, so I do the following:

`d.x = 1`
`d.dict.x = 1`

Neither of these commands raises an error, but the dictionary is still empty if I do `print(d.get_dict())`. I would expect them to either work or raise an error.

For context, the following gives an explicit error instead (which I'm also not sure why):

```d['x'] = 1   # TypeError: 'Dict' object does not support item assignment```

While the following actually works:

```d.update_dict({'x':1})```

### PR DESCRIPTION
**Pull request by ramirezfranciscof**
Fixes #2883 

The following two ways for setting (key:val) pairs in the dicts stored inside a Dict node were failing (the first one silently, the second one with an error message) but are now valid:

```
dict_node = Dict()
dict_node.dict.x = 'Set a value for x'
dict_node['x'] = 'Overwrite the value for x'
```

### DIFF
--- a/aiida/orm/nodes/data/dict.py
+++ b/aiida/orm/nodes/data/dict.py
@@ -19,7 +19,25 @@ __all__ = ('Dict',)
 
 
 class Dict(Data):
-    """`Data` sub class to represent a dictionary."""
+    """`Data` sub class to represent a dictionary.
+
+    The dictionary key-value pairs are stored in the database in a column named `attributes`.
+    We will then say that these are "database-attributes", or "attributes in a database sense".
+    However, these are not "python-attributes" ("attributes in the python sense"), and can only
+    be accessed with the set/get item special methods or through the `dict` special property.
+
+        # These will set a database storeable key-value pair
+        dict_node[key] = value
+        dict_node.dict.key = value
+
+    If you try to set this directly as a "python-attribute", it will only be kept in memory for
+    as long as that node object persists, but it won't appear in the `dict` property nor will
+    it be stored in the database.
+
+        #  This will just set a transient `key` attribute on the node
+        dict_node.key = value
+
+    """
 
     def __init__(self, **kwargs):
         """Store a dictionary as a `Node` instance.
@@ -37,7 +55,10 @@ class Dict(Data):
             self.set_dict(dictionary)
 
     def __getitem__(self, key):
-        return self.get_dict()[key]
+        return self.get_attribute(key)
+
+    def __setitem__(self, key, value):
+        self.set_attribute(key, value)
 
     def set_dict(self, dictionary):
         """ Replace the current dictionary with another one.
--- a/aiida/orm/nodes/data/dict.py
+++ b/aiida/orm/nodes/data/dict.py
@@ -21,22 +21,29 @@ __all__ = ('Dict',)
 class Dict(Data):
     """`Data` sub class to represent a dictionary.
 
-    The dictionary key-value pairs are stored in the database in a column named `attributes`.
-    We will then say that these are "database-attributes", or "attributes in a database sense".
-    However, these are not "python-attributes" ("attributes in the python sense"), and can only
-    be accessed with the set/get item special methods or through the `dict` special property.
+    The dictionary contents of a `Dict` node are stored in the database as attributes. The dictionary
+    can be initialized through the `dict` argument in the constructor. After construction, values can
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: aiida/orm/utils/managers.py [hunk 1]
```diff
@@ -128,7 +128,8 @@ def __init__(self, node):
         :param node: the node object.
         """
         # Possibly add checks here
-        self._node = node
+        self.__dict__['_node'] = node
+        #self._node = node
```
sphuber, line 132:
```suggestion
```

aiida/orm/utils/managers.py
```diff
@@ -128,7 +128,8 @@ def __init__(self, node):
         :param node: the node object.
         """
         # Possibly add checks here
-        self._node = node
+        self.__dict__['_node'] = node
```
sphuber, line 131: Add comment why you are doing this

tests/orm/data/test_dict.py
```diff
@@ -39,3 +39,13 @@ def test_get_item(self):
         """Test the `__getitem__` method."""
         self.assertEqual(self.node['value'], self.dictionary['value'])
         self.assertEqual(self.node['nested'], self.dictionary['nested'])
+
+    def test_set_item(self):
+        """Test the methods for setting the item:
```
sphuber, line 44:
```suggestion
        """Test the methods for setting the item.
        
```

tests/orm/data/test_dict.py
```diff
@@ -39,3 +39,13 @@ def test_get_item(self):
...
+        * `__setitem__` directly in `Dict`
+        * `__setattr__` throught the `AttributeManager`
```
sphuber, line 46:
```suggestion
        * `__setitem__` directly on the node
        * `__setattr__` through the `AttributeManager` returned by the `dict` property
```

aiida/orm/nodes/data/dict.py
```diff
@@ -19,7 +19,25 @@
 
 
 class Dict(Data):
-    """`Data` sub class to represent a dictionary."""
+    """`Data` sub class to represent a dictionary.
+
+    The dictionary key-value pairs are stored in the database in a column named `attributes`.
+    We will then say that these are "database-attributes", or "attributes in a database sense".
+    However, these are not "python-attributes" ("attributes in the python sense"), and can only
```
sphuber, line 26: Although I see the point of adding this, given that this was not clear for you either, this does not just apply to the `Dict` nodes, but to all nodes. All node specific data are stored as attributes in the database. So although this may help clear things up, it may also confuse, since it may suggest that this is behavior specific to the `Dict` plugin.

I would propose a text something like:
```
    The dictionary contents of a `Dict` node are stored in the database as attributes. The dictionary can be initialized
    through the `dict` argument in the constructor. After construction, values can be retrieved and updated through the
    item getters and setters, respectively:

        node['key'] = 'value'

    Alternatively, the `dict` property returns an instance of the `AttributeManager` that can be used to get and set
    values through attribute notation:

        node.dict.key = 'value'

    Note that trying to set dictionary values directly on the node, e.g. `node.key = value`, will not work as intended.
    It will merely set the `key` attribute on the node instance, but will not be stored in the database. As soon as the
    node goes out of scope, the value will be lost.

    Finally, all dictionary mutations will be forbidden once the node is stored.
```

sphuber (changes requested):
Thanks @ramirezfranciscof . Have some suggestions

aiida/orm/nodes/data/dict.py
ramirezfranciscof: Ok, would it still be ok to add something like this between the last two paragraphs?

```
    node goes out of scope, the value will be lost.

    Note that here it becomes apparent the difference of something being an "attribute of a node"
    (in the sense that it is stored in the "attribute" column of the database when the node is stored)
    and something being an "attribute of a python object" (in the sense of being able to modify and
    access it as if it was a property of the variable).

    Finally, all dictionary mutations will be forbidden once the node is stored.
```

sphuber (approved):

sphuber:
Thanks a lot @ramirezfranciscof !

PR #4352, opened 2020-09-03
Introduce equality concept for nodes + tests

### LINKED ISSUE #1917: Distinct instances of the same node do not compare equal
Example input:
```python
n1 = load_node(42)
n2 = load_node(42)
print n1, type(n1)
print n2, type(n2)
print n1 == n2
```
Output:
```
uuid: ddf9a6de-56bf-4783-96d2-33400df20f43 (pk: 42) <class 'aiida.orm.data.upf.UpfData'>
uuid: ddf9a6de-56bf-4783-96d2-33400df20f43 (pk: 42) <class 'aiida.orm.data.upf.UpfData'>
False
```
Tested with types `UpfData`, `WorkCalculation`, `PwCalculation`.

EDIT: also tested with more "basic" types `aiida.orm.data.str.Str` and `aiida.orm.data.int.Int`. For these, some more discussion is possibly required: should the PK/UUID be compared, or the actual data themselves?

ramirezfranciscof:
Two nodes will evaluate to be equal if their uuid coincides.

Fixes #1917

### DIFF
--- a/tests/orm/node/test_node.py
+++ b/tests/orm/node/test_node.py
@@ -54,6 +54,14 @@ class TestNodeAttributesExtras(AiidaTestCase):
         super().setUp()
         self.node = Data()
 
+    def test_equality(self):
+        """Test the `__eq__` method."""
+        node1 = Node()
+        node2 = node1
+        node3 = Node()
+        self.assertTrue(node1 == node2)
+        self.assertTrue(node1 != node3)
+
     def test_attributes(self):
         """Test the `Node.attributes` property."""
         original_attribute = {'nested': {'a': 1}}
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -128,6 +128,18 @@ class Node(Entity, metaclass=AbstractNodeMeta):
         )
         super().__init__(backend_entity)
 
+    def __eq__(self, other):
+        """The uuid should uniquely identify a node.
+
+        Note that although there may be ways in which users may end up manually modifying the
+        uuid of a node, and thus ending up with different nodes with the same uuid, the problem
+        is with them doing these modifications and not with the equality being based on uuid.
+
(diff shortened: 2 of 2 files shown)


ramirezfranciscof:
This is now WIP and paused, see [this comment](https://github.com/aiidateam/aiida-core/issues/1917#issuecomment-686483712) in issue #1917

tests/orm/node/test_node.py
```diff
@@ -54,6 +54,14 @@ def setUp(self):
         super().setUp()
         self.node = Data()
 
+    def test_equality(self):
+        """Test the `__eq__` method."""
+        node1 = Node()
+        node2 = node1
```
greschd, line 60: This test doesn't work quite as intended: Because `node1` and `node2` are the same Python object (`id(node1) == id(node2)`), `node1 == node2` will be `True` even in the current code.

Instead, I think we need to store a `Data` node and load it back, such that `node1` and `node2` are different Python objects.

That can be tested by ``assert node1 is not node2``.

greschd: See the following (current behavior, before the applied fix):

```python
In [1]: from aiida.orm import Data

In [2]: d1 = Data().store()

In [3]: d2 = d1

In [4]: d3 = load_node(d1.uuid)

In [5]: d1 == d2
Out[5]: True

In [6]: d1 is d2
Out[6]: True

In [7]: d1 == d3
Out[7]: False

In [8]: d1 is d3
Out[8]: False
```

ramirezfranciscof: Thanks! I had a slight suspicion of this but I found the problems with the intrinsic python-mock node types before I could get to that. I would anyways have had to ask you guys how to check this for sure, haha.

ramirezfranciscof:
After some discussion in the original issue #1917 I will be closing this down until we reach a final agreement and then I can implement a more thorough and complete solution.

PR #4353, opened 2020-09-04
Bump base docker image version.

### LINKED ISSUE #4339: Newest `aiidateam/aiida-core:stable` docker image contains broken `pymatgen` dependency
### Short description

The [newest `aiidateam/aiida-core:stable` docker image](https://hub.docker.com/layers/aiidateam/aiida-core/stable/images/sha256-74a6e6e0fa060d04dbd72c1409f2d1fc2207bd5a40406ade7bb4f11bbfbd0214?context=explore) contains `pymatgen`, but that is missing its dependency `ruamel.yaml`.

**Update:** The `1.3.0.1` image is also broken, while `1.3.0` works. I suspect it's something about these being built more recently - the latest `pymatgen` release is from 2020-08-13.

### Steps to reproduce

- `docker pull aiidateam/aiida-core:stable`
- `docker run aiidateam/aiida-core:stable`
- `docker exec --tty $DOCKERID wait-for-services`
- `docker exec -it $DOCKERID bash`
- `su aiida`
- `verdi shell`
- `import pymatgen`

Result:
```python
(base) aiida@3e0b99c6409f:~$ verdi shell
Python 3.7.4 (default, Aug 13 2019, 20:35:49) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.17.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import pymatgen
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-a3a39e60e980> in <module>
----> 1 import pymatgen

/opt/conda/lib/python3.7/site-packages/pymatgen/__init__.py in <module>
     12 from fnmatch import fnmatch
     13 
---> 14 import ruamel.yaml as yaml
     15 
     16 from monty.json import MontyEncoder, MontyDecoder, MSONable

ModuleNotFoundError: No module named 'ruamel'
```

### Expected behavior

The last command should import `pymatgen`.

### Your environment

- Docker version 19.03.12, build 48a66213fe
- aiida-core version: 1.3.1
- Python version: 3.7.4

Tagging @csadorf @yakutovicha

yakutovicha:
fixes #4339 

The new image is based on the latest stable Bionic Beaver (18.04)
Ubuntu distribution, which contains fixes for critical vulnerabilities.

Additionally, this image contains a fix for `ruamel.yaml` package
installation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM aiidateam/aiida-prerequisites:0.2.0
+FROM aiidateam/aiida-prerequisites:0.2.1
 
 USER root
 


greschd (approved):

greschd:
Thanks! Since I think this will only affect the `latest` tag, how are we fixing this for `stable`? Backporting this commit to the release branch?

yakutovicha:
> Thanks! Since I think this will only affect the `latest` tag, how are we fixing this for `stable`? Backporting this commit to the release branch?

I can make a new docker image, tagged as `1.3.1.1` (based on AiiDA 1.3.1) or `1.3.0.2` (based on AiiDA 1.3.0).

yakutovicha:
> > Thanks! Since I think this will only affect the `latest` tag, how are we fixing this for `stable`? Backporting this commit to the release branch?
> 
> I can make a new docker image, tagged as `1.3.1.1` (based on AiiDA 1.3.1) or `1.3.0.2` (based on AiiDA 1.3.0).

any preferences, or should I do both?

greschd:
Probably the version based on AiiDA 1.3.1 (`1.3.1.1`) at least should be fixed, right? Not sure if the 1.3.0 - based one makes sense.. that also won't contain the bugfixes between 1.3.0 and 1.3.1.

For me it doesn't make much of a difference, I'm using these just for testing - there's nothing critical I rely on different between these versions.

yakutovicha:
> Probably the version based on AiiDA 1.3.1 (`1.3.1.1`) at least should be fixed, right? Not sure if the 1.3.0 - based one makes sense.. that also won't contain the bugfixes between 1.3.0 and 1.3.1.
> 
> For me it doesn't make much of a difference, I'm using these just for testing - there's nothing critical I rely on different between these versions.

Alright, then. I will build this updated image locally and push it to the DockerHub.

greschd:
Probably a decent policy is always having one maintained image per maintained AiiDA release branch.

yakutovicha:
@greschd both images (1.3.0.2 and 1.3.1.1) are pushed.

greschd:
@yakutovicha I guess the `stable` tag should now point to `1.3.1.1`? Or is that just automatically built from `master` and can't be changed otherwise?

greschd:
I see - probably not worth it to cherry-pick that onto master, right? It would probably lead to a bit of a mess next time we merge develop into master.

yakutovicha:
> I see - probably not worth it to cherry-pick that onto master, right?

I think it is not worth. At least I don't see a clear advantage of doing it vs the mess it would create.

greschd:
Ok - since it's harder to update, I think we should advise (don't know where) people to prefer pinning specific tags (like `1.3.1.1`) instead of the `stable` tag.

PR #4355, opened 2020-09-04
🔧 Add tox configuration

chrisjsewell:
Surprised no one has looked to add this before.
[tox](https://github.com/tox-dev/tox) is a ubiquitous test automation tool,
simply `pip/conda install tox` then run e.g. `tox -- tests/tools/groups/test_paths.py` and it will set up your test environment (or re-use the cached one) and run pytest.
Massively helpful when you are working across multiple repositories with different test environments!

Want to run against a different python version / backend? `tox -e py38-sqla -- tests/tools/groups/test_paths.py`
Want to test the docs? `tox -e docs-clean` or `tox -e docs-update`


greschd:
Tagging @csadorf, guess it's time to pick [this discussion](https://github.com/aiidateam/aiida-core/pull/4204#issuecomment-652281096) back up.

chrisjsewell:
> Edit: found that they are considering it

yeh but hey at least it can go in there 🤷 

>  assumed that the [build-system] is the only key in our pyproject.toml and will regenerate it from scratch

what is this madness lol, maybe it would just be better to just check that it is consistent with setup.json in the pre-commit rather than rebuilding

Alternatively, for now, I could add a `tox.ini` and move the contents of `pytest.ini` into it; keeping the same net number of files 😉

sphuber:
> what is this madness lol, maybe it would just be better to just check that it is consistent with setup.json in the pre-commit rather than rebuilding
> 
> Alternatively, for now, I could add a `tox.ini` and move the contents of `pytest.ini` into it; keeping the same net number of files wink

Nonono, in the long run we definitely want to start relying more heavily on `pyproject.toml` (in my opinion) and so anyway we will have to address the fact that it is plainly overwritten now. No point in pushing this further down the line.

chrisjsewell:
Yeh was just trying to lazy lol.
FYI, I've been working quite closely with the main maintainer of pip recently (pradyunsg, who added all the new resolver stuff).
He's pretty big on pyproject.toml, and also https://flit.readthedocs.io for packaging 
https://flit.readthedocs.io/ (not that I expect aiida-core to use it, but worth a look if you're interested)

csadorf:
> what is this madness lol, maybe it would just be better to just check that it is consistent with setup.json in the pre-commit rather than rebuilding

It is likely possibly to improve the logic, but I also have only so much time to spend on these things. Feel free to create a PR with an improved logic, the validator function is already implemented.

Concerning tox, I moved away from it in the vast majority of my projects, because I had more trouble with it then gaining from it. But if it works well here, I see no reason why not to include it.

chrisjsewell:
>  because I had more trouble with it then gaining from it. But if it works well here, I see no reason why not to include it.

Interesting, I've being using it on about 10 different projects, including aiida-core, and it's worked like a charm

csadorf:
> >  because I had more trouble with it then gaining from it. But if it works well here, I see no reason why not to include it.
> 
> Interesting, I've being using it on about 10 different projects, including aiida-core, and it's worked like a charm 
> (in particular using with the tox-conda plugin)

It's been a few years since I touched it so possible that it improved or maybe I didn't use it right or maybe there was some other reason that it didn't fit and I decided to abandon it. So, just to emphasize: no qualms with applying it here if it works well. :+1:

chrisjsewell:
>It's been a few years

yeh quite possibly, I only "discovered" it recently and am now wondering how I lived without it lol

Ok well I will try to make the pre-commit fix here

chrisjsewell:
> Ok well I will try to make the pre-commit fix here

All done.
As you'll see, I've move from [toml](https://github.com/uiri/toml) to [tomlkit](https://github.com/sdispater/tomlkit) for read/write of pyproject.toml. This is because it preserves formatting and comments, which toml does not (https://github.com/uiri/toml/issues/77).

utils/dependency_management.py
```diff
@@ -161,37 +161,43 @@ def generate_environment_yml():
...
-        'build-system': {
-            'requires': ['setuptools>=40.8.0,<50', 'wheel',
-                         str(reentry_requirement), 'fastentrypoints~=0.12'],
-            'build-backend': 'setuptools.build_meta:__legacy__',
-        }
+    # update the build-system key
+    pyproject['build-system'] = {
+        'requires': ['setuptools>=40.8.0,<50', 'wheel',
+                     str(reentry_requirement), 'fastentrypoints~=0.12'],
+        'build-backend': 'setuptools.build_meta:__legacy__',
```
csadorf, line 188: I understand that we want to implement the "update" a bit more fine-grained instead of just replacing the whole file, so I would suggest to also apply that to the individual keys, e.g., by using `.setdefault()` instead of `__setitem__()`.

utils/dependency_management.py
```diff
@@ -161,37 +161,43 @@ def generate_environment_yml():
...
     }
```
chrisjsewell, line 189:
```suggestion
    pyproject.setdefault('build-system', {})
    pyproject['build-system'].update({
        'requires': ['setuptools>=40.8.0,<50', 'wheel',
                     str(reentry_requirement), 'fastentrypoints~=0.12'],
        'build-backend': 'setuptools.build_meta:__legacy__',
    })
```

is this what you had in mind?

utils/dependency_management.py
chrisjsewell: see below

utils/dependency_management.py
```diff
@@ -161,37 +161,43 @@ def generate_environment_yml():
 
 
 @cli.command()
-def generate_pyproject_toml():
+def update_pyproject_toml():
     """Generate 'pyproject.toml' file."""
```
csadorf, line 165: Please adjust the doc-string as well to reflect the change in function.

utils/dependency_management.py
csadorf: Yes. You may even be able to chain it, but I'm not 100% sure on copy-vs-reference behavior here.

chrisjsewell: Yeh I couldn't remember off hand either, so I thought better safe than sorry lol

utils/dependency_management.py
chrisjsewell, line 165:
```suggestion
    """Generate a 'pyproject.toml' file, or update an existing one.
    
    This function generates/updates the ``build-system`` section,
    to be consistent with the 'setup.json' file.
    """
```

utils/dependency_management.py
greschd: You can chain it, but this is more readable IMO.

csadorf (dismissed):
Thx a lot!

csadorf (approved):

PR #4357, opened 2020-09-04
deprecate `--clear` option of `verdi group delete`

ltalirz:
fixes #4357


ltalirz:
I do have one more question about the purpose of `clear()`ing the group before deletion.

What is the difference between clearing the contents of a group and then deleting it vs directly deleting the group (vs `Group.objects.delete`)?
The nodes remain untouched in either case - it seems like an unnecessary extra step.

The user interface could remain the same in order to accidentally delete groups that contain nodes (although I would suspect that this is actually the more common use of the command than to delete empty groups).

sphuber:
I added the `clear` option in 719d65abd7086d26b028967365dc3a1e12a643b0 . It's functionality used to be combined with the `--force` flag, which I though was potentially dangerous. It is true though that it should just serve as a barrier to people accidentally deleting non empty groups, there is no need to call `group.clear`. Could you update the PR to include this change? I think that would be fine.

ltalirz:
Thanks, I will make the change.

Just wondering, should there also be a flag to actually delete the nodes as well?
It seems to me that this would be quite useful in practice (although it should obviously be off by default).

sphuber:
> Just wondering, should there also be a flag to actually delete the nodes as well?
> It seems to me that this would be quite useful in practice (although it should obviously be off by default).

Think that makes sense, however, it should come with a confirmation prompt, because it will have to follow the consistency rules and it may need to delete a lot more nodes than just those that were in the group.

ltalirz:
> > Just wondering, should there also be a flag to actually delete the nodes as well?
> > It seems to me that this would be quite useful in practice (although it should obviously be off by default).
> 
> Think that makes sense, however, it should come with a confirmation prompt, because it will have to follow the consistency rules and it may need to delete a lot more nodes than just those that were in the group.

Great - I'll open an issue for this, since it is a bit beyond the scope of this PR.
The PR should be ready to go.

sphuber:
@mbercx will you have time to do the implementation of #4358 relatively soon? Because that would make these changes obsolete anyway.

mbercx:
> @mbercx will you have time to do the implementation of #4358 relatively soon? Because that would make these changes obsolete anyway.

Sure, I should be able to work on this early next week! So I guess we can close this PR?

ltalirz:
I'd would still merge it. You anyhow won't be able to remove the `--clear` flag before 2.0, i.e. the docs will still need to mention it.

sphuber:
Fine by me to postpone the new interface until 2.0 since we are changing it, but then it would be good to put official deprecation warning for the clear option in this PR. If you can add that then it's fine for me to merge

ltalirz:
> then it would be good to put official deprecation warning for the clear option in this PR.

I agree. Just to clarify, though: this means I will also remove the necessity to specify the `--clear` option (we don't want to be forcing people to use a deprecated option). 
I think this change in functionality is ok - it should not break existing scripts.

sphuber:
> ust to clarify, though: this means I will also remove the necessity to specify the `--clear` option (we don't want to be forcing people to use a deprecated option).
> I think this change in functionality is ok - it should not break existing scripts.

I agree that it is not optimal, but I am not sure I agree with the assessment. Some people may actually have come to rely on the behavior, calling it without `--clear` because they only want to delete empty groups. If we silently remove that safety, that breaks that assumption. "Scripts will continue to work" but maybe not as intended. So I would actually keep the option and update the deprecation warning to say "the `--clear` option is deprecated and will be removed in 2.0 after which empty and non-empty groups will be treated equally"

mbercx:
@ltalirz should we still add this deprecation in the 1.5.0 release? It's important to add this asap, so we can include #4425 in 2.0.

I do tend to agree with @sphuber that we should not just remove the options functionality, but just add a deprecation warning.

ltalirz:
Right - I certainly get the argument (this change could in principle break user's scripts).
The reason I am a bit hesitant to implement this suggestion is that this is a somewhat special case: the suggestion here is to mark an API as "to be removed" without giving users an alternative. I.e. by deprecating it we force any user who wants to delete a non-empty group to read our deprecation warning until 2.0 is out, and we're forcing them to keep using an API that will break in 2.0 (meaning their scripts will be *guaranteed* to break).

The reason we are doing this is to prevent the potential case of users relying on the fact that leaving out the `--clear` option will prevent them from deleting non-empty groups (I would add that this concern only applies to `--non-interactive` use, since without the `--non-interactive` flag there is still a confirmation prompt).
Granted - it is possible that users use `verdi group delete -n` and rely on the fact that this will fail when a group is not empty but I would bet that I can count the number of users who do this on one hand (if it isn't zero), while the number of users impacted by the situation described in the first paragraph is much larger.

So, in my view it's a balance to strike here and I lean towards changing the behavior so that people have time to adapt their code to remove the `--clear` flag in preparation of the 2.0 release.

sphuber:
> So, in my view it's a balance to strike here and I lean towards changing the behavior so that people have time to adapt their code to remove the `--clear` flag in preparation of the 2.0 release.

That's exactly the point though. How will you get people to stop using it, without having it print a deprecation warning?

By the way, there is no `--non-interacitve` option, I presume you meant `--force`. I think a deprecation warning on the command line is not that big of a deal and at this point we can no longer really argue like "I don't think there is anyone using this". I am also tempted to use this reasoning to be looser with deprecation pathways, but I think we start to have a large group of users now that we don't speak to directly and so have no way of knowing of how they use AiiDA.

ltalirz:
> That's exactly the point though. How will you get people to stop using it, without having it print a deprecation warning?

Sorry if I'm not being clear - I am proposing the current state of the PR, where we of course have a deprecation warning (but we are also changing the behavior of the command without the `--clear` option).

> By the way, there is no `--non-interacitve` option, I presume you meant `--force`

Right, sorry.

> we can no longer really argue like "I don't think there is anyone using this". I am also tempted to use this reasoning to be looser with deprecation pathways, but I think we start to have a large group of users now that we don't speak to directly and so have no way of knowing of how they use AiiDA.

Well, my point was that we are not offering a way out until 2.0 and that there are definitely going to be a significant number of people affected by this (basically anyone who uses the `verdi group delete command`), who will then be forced to adapt immediately with 2.0 rather than gradually starting today.

ltalirz:
@mbercx I believe this is still waiting for your review

aiida/cmdline/commands/cmd_group.py
```diff
@@ -59,33 +61,27 @@ def group_remove_nodes(group, nodes, clear, force):
...
     label = group.label
 
-    if group.count() > 0 and not clear:
-        echo.echo_critical((
-            'Group<{}> contains {} nodes. Pass `--clear` if you want to empty it before deleting the group'.format(
-                label, group.count()
-            )
-        ))
+    if clear:
+        warnings.warn('`--clear` is deprecated and no longer has any effect.', AiidaDeprecationWarning)  # pylint: disable=no-member
```
mbercx, line 81: Perhaps we can still add a warning in case the group is not empty and the `force` flag isn't set?

```suggestion
    if group.count() > 0 and not force:
        echo.echo_warning(f'Group <{label}> contains {group.count()} nodes.')

```

mbercx (changes requested):
Sorry for the delay, @ltalirz! I have just one small suggestion.

aiida/cmdline/commands/cmd_group.py
sphuber: N.B.: `.count()` can potentially be expensive, so it's better to call it once and assign to variable

ltalirz: thanks @mbercx - I added it but then I had second thoughts 

 * Warning people how many nodes there are inside the group can give the impression that we're deleting nodes here (while it is just the group that is being deleted)
 * As @sphuber mentions, it can slow down the command

I.e. unless you feel strongly about this, I prefer to leave the code as it is

mbercx: > Warning people about how many nodes there are inside the group can give the impression that we're deleting nodes here (while it is just the group that is being deleted)

That's a good point, I hadn't thought about that. I suppose we can still rephrase the warning, but it's not critical.

> As @sphuber mentions, it can slow down the command

I actually tested this for one of our larger groups on theossrv5, and it's not _that_ time consuming:

```Python
In [10]: g.count()                                                                                                                     
Out[10]: 61134

In [11]: timeit g.count()                                                                                                              
174 ms ± 1.39 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

But, I do not feel super strongly about it, so feel free to merge!

mbercx (approved):

ltalirz:
Thanks @mbercx !

aiida/cmdline/commands/cmd_group.py
sphuber: this may be on the sqlalchemy backend, where I remember trying to optimize the shit out of it because it was really slow at some point and it shouldn't be. could be different on Django, could be also ok though, not sure. But I am also fine with leaving it as is as @ltalirz suggests. Let's get this thing merged.

PR #4362, opened 2020-09-09
🧪 Add Pytest Benchmark runs and GH Action

chrisjsewell:
Ok lets open this up for review!

(long story, short, check out https://aiidateam.github.io/aiida-core/dev/bench/ubuntu-18.04/django/ 😄)

---

You can look at https://github.com/aiidateam/github-action-benchmark, and its README, for details of the GH Action. But the basic steps of the workflow are:

1. Run pytest to generate JSON data. 

By default, these tests are switched off (see pytest.ini) but to run them locally, simply use `pytest tests/benchmark --benchmark-only` (or `tox -- tests/benchmark --benchmark-only` 😉).
This runs each test, marked as a benchmark, n-times and records the timing statistics (see [pytest-benchmark](https://pytest-benchmark.readthedocs.io)).
When run also with `--benchmark-json benchmark.json`, a JSON file will also be created, with all the details about each test.

2. Extract information from the above JSON, and also data about the system (number of CPUs, etc) and created a "simplified" JSON object.

3. Read the JSON object from the specified `gh-pages` folder (`data.js`), which contains a list of all these JSON objects.

I have split these by OS and backend, see:
- https://github.com/aiidateam/aiida-core/blob/gh-pages/dev/bench/ubuntu-18.04/django/data.js
- https://github.com/aiidateam/aiida-core/blob/gh-pages/dev/bench/ubuntu-18.04/sqlalchemy/data.js

4. If available, compare the new JSON section against the last one to be added `data.js`, and comment in the PR and/or fail the workflow if the timings have sufficiently degraded (depending on GH action configuration).

5. If configured, add the new data to `data.js`, update the other website assets (HTML/CSS/JS) and commit the updates to `gh-pages`. This then updates e.g.: https://aiidateam.github.io/aiida-core/dev/bench/ubuntu-18.04/django.
  Note the on this page you can:
  - hover over individual data points, to see all their associated data/metadata
  - clicking on the data point will take you to the commit diff
  - you can click on a legend label, to hide that line
  - you can drag a box over an area to zoom in the charts

Naturally, these tests are slower than standard unit tests: currently they take ~7/8 minutes, with fairly conservative tests/# of repetitions.
So most likely you would not want to run these on every PR commit. My current solution for this is to have two workflow jobs: 
1. the first runs on every commit to `develop` (if it's not just updating the documentation), updating the `gh-pages` data.
2. the second is triggered by a commit to a branch with an open PR to `develop`, **but only** if you include in the commit message `[run bench]`. This will report back the timing data, but not update `gh-pages`, see https://github.com/aiidateam/aiida-core/pull/4362#commitcomment-42214213.
  I would envisage you would run this on a "final" check of a PR, before merging

---

On to the actual tests.
As you can see on the `gh-pages`, I have split these into three categories that I think could be helpful:

1. basic node storage/deletion (i.e. interactions with the ORM). With my new GH action assets, you'll see that the tests in a group are added to the same chart, which I think is nice to get a sense of comparative timings.
  (Writing this now, I think perhaps a good additional section would be for bulk storage timings)

2. Runs of workchains with internal (looped) calls to workchains/calcjobs.
   These are duplicated using both a local runner and a daemon runner.
   The daemon runner code is a bit tricky, and I imagine may break on the move to asyncio?
   Monitoring rabbitmq (*via* the management plugin on port 15672) I'm 99% sure these tests are "engaging" the messaging system 😬 
   Ideally, I guess you would also want to start multiple workers before running the daemon tests, to gauge multiprocessing performance.

3. Expoting/importing archives. As you can see in the code, I've written a nice little `recursive_provenance` function to generate an initial provenance tree, with a specified breadth/depth 😄 

---

**So, over to you guys! Initial impressions?**

Obviously it is very subjective as to what tests should be added, and we can always add more in the future. But let me know if you think there is any other major category that I have missed.
In terms, of the webpage formatting; just let me know of any suggestions, since its pretty easy to tweak now in https://github.com/aiidateam/github-action-benchmark


chrisjsewell:
See the initial comment for details!

@csadorf, the pip install failure looks to be a `2020-resolver` issue?

chrisjsewell:
Oh and I came across what seems to be a glaring issue with `actions/checkout@v2` https://github.com/actions/checkout/issues/299 (not checking out the correct commit!) 😬

csadorf:
> See the initial comment for details!
> 
> @csadorf, the pip install failure looks to be a `2020-resolver` issue?

It seems so, since our `test-install` workflow also failed [last night](https://github.com/aiidateam/aiida-core/runs/1099766800?check_suite_focus=true).

chrisjsewell:
I think that maybe this 2020-resolver is not quite as ready for release as we were led to believe 😬
Particularly given the issues with #4317, perhaps we should roll it’s use back?

csadorf:
> I think that maybe this 2020-resolver is not quite as ready for release as we were led to believe 😬
> Particularly given the issues with #4317, perhaps we should roll it’s use back?

I don't think we are "using" it as of yet, we are testing it as part of our workflows for this exact reason. To identify issues and report them. Considering the sheer amount of issues with the new resolver, I can't imagine that the pip maintainers will stick to their original plan of rolling it out in October, but we will see.

However, I'm working on updating our test-install workflow to not fail when issues with the 2020-resolver occur, but rather notify us in some other form (still figuring out the right way to do this).

ltalirz (commented):
First of all, thanks a lot @chrisjsewell , this is great! 

Looking through the results page (not the code yet), here is some feedback from my side:
 1. minor: can one turn on a toolbar in the plots that allows to reset them after one has zoomed in (workaround at the moment: just reload the page)
 2. if I understand correctly, the current fluctuations are not actually due to changes in the code. Since the fluctuation are quite substantial (~20%), it will be difficult to make real use of this data unless we are able to get those fluctuations down.
   Does increasing the number of test repetitions decrease these fluctuations or is it caused by the load (or perhaps even varying CPU architecture) on the corresponding runners?
 3. the y axis label is currently "iter/sec" - I guess this means more is faster? I would propose to switch this around, i.e. report the time per iteration. It's more natural to say "storing a node has gone from 20ms to 30ms" than to talk about iterations per second
 4. if at all possible we need a logarithmic y scale. one cannot really see what is going on for the slow operations


P.S. Will this allow us to retire our jenkins server?

chrisjsewell:
> 1. minor: can one turn on a toolbar in the plots that allows to reset the view after one has zoomed in (workaround at the moment: just reload the page)

> 4. if at all possible we need a logarithmic y scale. Looking at the current plots, one cannot really see what is going on for the slow operations and  we do care about relative changes in timings - whether the individual operations are fast or slow.

Done https://github.com/aiidateam/github-action-benchmark/commit/8b0a284373d24691f26a472da8f6b5c59b364e7b and updated on https://aiidateam.github.io/aiida-core/dev/bench/ubuntu-18.04/django/ 👍 

> 3. the y axis label is currently "iter/sec" - I guess this means more is faster? I would propose to switch this around, i.e. report the time per iteration. It's more natural to say "storing a node has gone from 20ms to 30ms" than to talk about iterations per second

Yep it taken from the pytest-benchmark `ops` field: https://github.com/ionelmc/pytest-benchmark/blob/0508a7509355e6912499726f390753b110515cae/src/pytest_benchmark/stats.py#L168, so can just add a switch to use mean = 1/ops

> 2. if I understand correctly, the current fluctuations are not actually due to changes in the code. Since the fluctuation are quite substantial (~20%), it will be difficult to make real use of this data unless we are able to get those fluctuations down.
> Does increasing the number of test repetitions decrease these fluctuations or is it caused by the load (or perhaps even varying CPU architecture) on the corresponding runners?

> P.S. Will this allow us to retire our jenkins server?

Well I guess it depends on what you mean by "real use"? I know we are computational so are used to exact numbers, but laboratory scientists happily deal with fluctuations in their data all the time 😉
That being said:

- One of the main reason I forked the existing `github-action-benchmark` was to plot groups of benchmarks together: even if the absolute results fluctuate, the relative results are quite constant and I think that's very helpful, e.g. looking at the "possible" store time of database data (`store_backend`) compared to the time taken via the aiida/django ORM, which is 4/5 times slower. 

- As discussed in [pytest-benchmark FAQ](https://pytest-benchmark.readthedocs.io/en/latest/faq.html#frequently-asked-questions), it is naturally difficult to achieve full hardware isolation on  a VM (and really on any system). Also with the GitHub VMs obviously you only have access to 2 CPUs per VM.
  With Jenkins we have access to a larger (and possibly more stable) hardware on theossrv6, but its a bit of pain to work with.
  It would be ideal, for these larger tests, if we could move this to using theossrv6 as a GitHub [self-hosted runner](https://docs.github.com/en/actions/hosting-your-own-runners).
  This big sticking point here though is [network security](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories).
  Off the top of my head, some partial solutions would be to move these tests to a separate, private, repository, or try to implement a similar setup to jenkins, whereby everything gets run within a Docker container, making it a bit more isolated.

ltalirz:
Thanks, Chris!

> Well I guess it depends on what you mean by "real use"? 

To clarify: I understood that one of the purposes of these automatic benchmarks is to detect when changes are introducing performance hits (or improvements).
Ideally, these changes would be detectable when they are being made (i.e. in a PR).  Any code changes that result in performance differences that fall within the fluctuations will be difficult to detect.

chrisjsewell:
> Ideally, these changes would be detectable already at the point when they are being proposed (i.e. in a PR). Any code changes that result in performance differences falling within the fluctuations will be difficult to detect in that context.

It just depends how "strict" you want to be on these check; currently they are set to fail/report on >200% changes, which will check for changes that are degrading performance by orders of magnitude.

Outside of that, I'm open to suggestions but don't see any trivial changes I can make to improve the current situation.
(I don't think that increasing the iterations will decrease fluctuations that much, it's just a matter of trial and error really).

ltalirz:
> It just depends how "strict" you want to be on these check; currently they are set to fail/report on >200% changes, which will check for changes that are degrading performance by orders of magnitude.
> 
> Outside of that, I'm open to suggestions but don't see any trivial changes I can make to improve the current situation.
> (I don't think that increasing the iterations will decrease fluctuations that much, it's just a matter of trial and error really).

Ok, my point was just that any improvement we can make in this regard translates into a better ability to detect changes in actual performance.

It's certainly already very useful to be able to spot changes of 100% and more.

chrisjsewell:
> Ok, my point was just that any improvement we can make in this regard translates into a better ability to detect changes in actual performance.

indeed 😄

giovannipizzi:
I think that (unless there are other some quick suggestions) I would start merging this as is.
I agree that fluctuations don't help, but in the end the goal here is to see if a change *significantly* degrades the performance. If we lose a 10% it's acceptable.

Also we need a quick development workflow. Waiting 10 minutes more for every PR commit it's going to kill productivity.

The goal of this (in the way I see it, and in the way I implemented it also in the disk-objectstore package) is that even if there are (relatively big) oscillations, after a few commits you see if the average has gone worse or not (a bit like with COVID data ;-) ).

So, my only suggestion would be to:
- make sure that everything that is merged on develop IS benchmarked even if the user doesn't ask so explicitly. It's important to have this historical data
- do not make the test fail if oscillations are > 200% but just to report it back in the PR/commit (it's already a feature), e.g. mentioning a few maintainers of the teams so they are also alerted. In my experience, when working with the disk especially, you can randomly have one test much worse than the average
- I would therefore suggest to either run the workflow only on merges on develop, or as a scheduled cron job nightly. So, even if it takes 10+ mins, it's not a problem.

If we realise there is a efficiency problem, OR we end up on the benchmark page and we see a clear worsening of the results from a commit on, we will open an issue and investigate more in detail.

As a last comment: while using a log scale is useful to show them all together, we need to be careful about not hiding the oscillations we need to observe. Anyway, for now I think it's ok and we can just change this later if needed.

So, I suggest to merge and do further fine-tuning in additional PRs if there are no major concerns (so @chrisjsewell can then start merging a few important performance improvements and we can check if these would be visible in these plots - this will also be an 'answer' to the doubts on the size of the oscillations and whether they are too big to be useful).

(@chrisjsewell - I didn't check, but make sure you are adding at least one benchmark on something you know it's going to be faster after your future fixes).

chrisjsewell:
> do not make the test fail if oscillations are > 200% but just to report it back in the PR/commit (it's already a feature), e.g. mentioning a few maintainers of the teams so they are also alerted. 

Added in d25edfd 👍 

> Also we need a quick development workflow. Waiting 10 minutes more for every PR commit it's going to kill productivity.
> I would therefore suggest to either run the workflow only on merges on develop

Yep thats already how its setup 👍 As outlined in my initial comment; it is triggered on pushes to develop, or commits to PRs, BUT **only** if they include `[run bench]` in their commit message, which I think could be helpful when working on a performance based PR, and you want to check the benchmarks before merging.

So, for example, d25edfd triggered `run-on-comment` https://github.com/aiidateam/aiida-core/pull/4362/checks?check_run_id=1102778327, which created this comment https://github.com/aiidateam/aiida-core/pull/4362#commitcomment-42235216 (but did not push the data to `gh-pages`)

Then `run-and-upload` runs on `develop` commits, uploads the data to `gh-pages` and alerts us if it did not meet the alert threshold.

> but make sure you are adding at least one benchmark on something you know it's going to be faster after your future fixes

yep I think I've captured all of this.
But if I realise later others are relevant, then I'll look to make an additional pre-PR, with any additional test(s), before the actual fix PR.

chrisjsewell:
> As a last comment: while using a log scale is useful to show them all together, we need to be careful about not hiding the oscillations we need to observe. 

Note, you can change this configuration at any time *via* .github/workflows/benchmark-config.json

chrisjsewell:
> So, I suggest to merge

Happy to merge once someone drops me an approving review 😄

sphuber:
> Yep thats already how its setup +1 As outlined in my initial comment; it is triggered on pushes to develop, or commits to PRs, BUT **only** if they include `[run bench]` in their commit message, which I think could be helpful when working on a performance based PR, and you want to check the benchmarks before merging.

I am not sure I understand the reasoning behind this. In my mind, the whole reason of having the benchmark tests is to be alerted if you _inadvertently_ affect the performance in a negative way. I would therefore expect this to run for each PR such that we can evaluate its effect on performance _before_ merging it. If users are required to put `[run bench]` in the commit message, to me that defeats the purpose. Sure if you _know_ you are working on something that might affect performance, you can include it, but the whole problem is that you might affect the performance without even realizing it. Is there no way to have this always run on a PR? What is the current cost of these benchmark tests? Are these too expensive? So far we have not had any problems with GHA throughput.

chrisjsewell:
> Is there no way to have this always run on a PR? What is the current cost of these benchmark tests? Are these too expensive?

Sure we can turn them on for ever PR commit, they currently take ~6:30 minutes to run, and the unit test suite takes about ~11:30 minutes, so its probably not a massive deal. Up to you guys?

As discussed, we can probably get some benefit from testing up front, at the PR level, although fluctuations will prevent the checks from being too lenient.
The more prominent benefit/goal IMO though, is to track performance over time, which can be reviewed periodically.

(on a side note, these comments that the action can create, https://github.com/aiidateam/aiida-core/pull/4362#commitcomment-42235216, I think should ideally "self-update" each commit like the codecov one does, https://github.com/aiidateam/aiida-core/pull/4362#issuecomment-690781401. But I can mess around with that at a later date)

csadorf:
> > Is there no way to have this always run on a PR? What is the current cost of these benchmark tests? Are these too expensive?
> 
> Sure we can turn them on for ever PR commit, they currently take ~6:30 minutes to run, and the unit test suite takes about ~11:30 minutes, so its probably not a massive deal. Up to you guys?

Before we do that, let's make sure that those tests reliably indicate a performance regression. Otherwise we will start getting desensitized to negative test results.

chrisjsewell:
> Before we do that, let's make sure that those tests reliably indicate a performance regression

So what's the consensus? Shall we merge as is for now, then review whether to turn it on for all PR commits once we "lived with it" for a bit longer?

tests/benchmark/test_engine.py
```diff
@@ -0,0 +1,191 @@
...
+class WorkchainLoop(WorkChain):
+    """A basic Workchain to run a looped step n times."""
+
+    @classmethod
+    def define(cls, spec):
+        super().define(spec)
+        spec.input('iterations', required=True)
+        spec.input('code', required=False)
+        spec.outline(cls.init_loop, while_(cls.terminate_loop)(cls.run_task))
+        # spec.output('result')
```
ltalirz, line 38: leftover comment

tests/benchmark/test_engine.py
```diff
@@ -0,0 +1,191 @@
...
+    'basic-loop': (WorkchainLoop, 4, 0),
+    'serial-wc-loop': (WorkchainLoopWcSerial, 4, 4),
+    'threaded-wc-loop': (WorkchainLoopWcThreaded, 4, 4),
+    'serial-calcjob-loop': (WorkchainLoopCalcSerial, 4, 4),
+    'threaded-calcjob-loop': (WorkchainLoopCalcThreaded, 4, 4),
+}
+
+
+@pytest.mark.parametrize('workchain,iterations,outgoing', WORKCHAINS.values(), ids=WORKCHAINS.keys())
+@pytest.mark.usefixtures('clear_database_before_test')
```
ltalirz, line 120: what is the advantage of adding it this way rather than making it an argument of the test function like all other fixtures?

Note: one might be able to turn on this fixture for benchmark tests automatically by placing a `conftest.py` with content
```python
import pytest

@pytest.fixture(scope='function', autouse=True)
def clear(clear_database_before_test):
    pass
```
but I'm not sure and perhaps it's more confusing than helpful.

ltalirz (commented):
I've had a quick look through the code and it looks fine to me.
Just two minor questions/comments

tests/benchmark/test_engine.py
sphuber: Main argument is that this fixture is not actually supposed to be called by the test function itself. It should just be injected. You could add it as an argument, but then linters, such as pylint, will think you made a mistake and issue a warning. This approach makes it clearer that it is a non-callable fixture

sphuber (commented):
Since the majority seems to be ok with merging this, I won't stand in the way. However, I still want to point out that past experience has shown that having these non-committal CI steps largely tend to get ignored and so ultimately do not accomplish their intended goal. Example was the Jenkins machine: we have been running it to run some extra tests that are more costly, but we do not require them. There has been multiple extended periods where the machine wasn't working properly and since no-one is used to checking the results, it just gets ignored. The same goes for test coverage I would say. We are so used to seeing false positives that also this is still not required and gets ignored. I know the argument is that this way we keep a history that can be checked, but does anyone actually do this? I think we should just be watchful that the benchmarks won't suffer a similar fate and will just be there running in the background but they won't actually be used.

tests/benchmark/test_engine.py
ltalirz: Would be good to decide whether we want to adopt this approach going forward - both to know what to follow in new tests and because it means that it is no longer enough to look at the test function arguments to understand which fixtures were called.

sphuber: I have actually already been using this in all the pytests I have added, so I am definitely in favor. I don't see the problem in having the fixtures in a decorator if they don't get explicitly called in the test body.

setup.json
```diff
@@ -105,6 +105,7 @@
             "pytest-timeout~=1.3",
             "pytest-cov~=2.7",
             "pytest-rerunfailures~=9.0",
+            "pytest-benchmark~=3.2.3",
```
chrisjsewell, line 108:
```suggestion
            "pytest-benchmark~=3.2",
```

tests/benchmark/test_engine.py
chrisjsewell, line 38:
```suggestion
```

csadorf:
> Since the majority seems to be ok with merging this, I won't stand in the way. However, I still want to point out that past experience has shown that having these non-committal CI steps largely tend to get ignored and so ultimately do not accomplish their intended goal. Example was the Jenkins machine: we have been running it to run some extra tests that are more costly, but we do not require them. There has been multiple extended periods where the machine wasn't working properly and since no-one is used to checking the results, it just gets ignored. The same goes for test coverage I would say. We are so used to seeing false positives that also this is still not required and gets ignored. I know the argument is that this way we keep a history that can be checked, but does anyone actually do this? I think we should just be watchful that the benchmarks won't suffer a similar fate and will just be there running in the background but they won't actually be used.

How about we put an item on the AiiDA calendar for when we turn on the benchmarks as mandatory check? Let's say in 30 days? In that way we resolve any problems that occur, but then it won't suffer the fate of being completely ignored at some point.

tests/benchmark/test_engine.py
chrisjsewell: Yep, it's a standard pytest practice, so I don't see any issue with using it.
If you want to understand what fixtures are being called, you shouldn't look at the function anyway, you call `pytest --setup-plan` 😉

chrisjsewell:
> I think we should just be watchful that the benchmarks won't suffer a similar fate and will just be there running in the background but they won't actually be used.

Yep completely valid point, I certainly intend to utilise them over the coming month, at which point hopefully will have a clearer picture on how useful they are.

> How about we put an item on the AiiDA calendar for when we turn on the benchmarks as mandatory check? Let's say in 30 days?

That sounds reasonable to me 👍

sphuber:
Do we know why some steps of the `test-install` workflow are failing? I noticed that I didn't have this fail in other PRs that I opened today, but then again, that workflow wasn't even run. What is the condition for it running?

csadorf:
> Do we know why some steps of the `test-install` workflow are failing? I noticed that I didn't have this fail in other PRs that I opened today, but then again, that workflow wasn't even run. What is the condition for it running?

Those are issues related to the pip 2020-resolver. I've opened a PR to fix the issue which is ready for review: https://github.com/aiidateam/aiida-core/pull/4369

The workflow runs when any changes are made to the dependency specification (including also implicit and potential changes, e.g. modifications of setup.py), the workflow definition itself, as well as nightly.

sphuber:
Thanks @csadorf for the clarification. I will have a look at that PR now. Would prefer to merge that first before merging this PR.

sphuber (approved):
Thanks a lot @chrisjsewell

PR #4363, opened 2020-09-09
ssh transport : adapt for subclassing

adegomme:
follow up of #4199 ( and #4235)

To be able to release aiida-sshonly as a plugin without duplicating code, I had to make some tiny changes to ssh.py, easing its subclassing to better isolate the file transfer parts and keep sftp references wrapped.
- this adds some simple wrappers around some sftp calls, that can be overriden in another class. Most methods don't need changing, as it has to be overriden fully, mainly utilities are changed (stat, lstat, symlink) 
- new method to initialize file transport separately. Also adds error checking for SFTP initialization, with an explicit message if it fails to launch, and a possible solution.
- also limit the output of a log message to 1000 characters, as whole files can be printed in it when using a transport sending large ssh commands.

Merging this would would allow (even better if #3787 is also merged) the plugin to use ssh.py from aiida instead of a copy, which leads to incompatibilities and duplications.


ltalirz:
thanks a lot, @adegomme !
@giovannipizzi would you be available for reviewing this?

aiida/transports/plugins/ssh.py
```diff
@@ -437,14 +437,26 @@ def open(self):
...
+    def open_file_transport(self):
+        """
+        Open the SFTP channel, and handle error by directing customer to try another transport
+        """
+        from aiida.common.exceptions import InvalidOperation
+
+        try:
+            self._sftp = self._client.open_sftp()
+        except Exception:
+            raise InvalidOperation('SFTP seems not functional. Try using sshonly plugin instead?')
```
ramirezfranciscof, line 454: Not sure if this message is super clear given that (1) people may not know that SFTP refers to what the default ssh plugin uses and (2) the process of actually using another plugin means you need to set up another copy of the computer with that different plugin. This is a cluster-configuration problem, right? Its not that the protocol might be temporary down. In that case what do you think of...?

```suggestion
            raise InvalidOperation('Error in ssh transport plugin. This may be due to the remote computer not supporting SFTP. Try setting it up with the sshonly plugin instead.')
```

aiida/transports/plugins/ssh.py
```diff
@@ -1228,7 +1256,7 @@ def _exec_command_internal(self, command, combine_stderr=False, bufsize=-1):  #
         else:
             command_to_execute = command
 
-        self.logger.debug('Command to be executed: {}'.format(command_to_execute))
+        self.logger.debug('Command to be executed: {}'.format(command_to_execute[:1000]))
```
ramirezfranciscof, line 1259: So the rest of the re-organization and wrapping of methods seems ok to me, but this is the only modification that I am unsure about. Can't you be loosing important log info like this? I would also maybe ask for the input of @giovannipizzi and/or @sphuber in this, since they probably know better what to expect inside that command and why is being logged.

ramirezfranciscof (changes requested):
Hi @adegomme , and thanks for this re-structuring! This all looks really good, I just had one modification request for the error message and maybe some discussion regarding the 1000 char cap on the logging.

aiida/transports/plugins/ssh.py
```diff
@@ -1309,6 +1337,15 @@ def gotocomputer_command(self, remotedir):
         )
         return cmd
 
+    def symlink_internal(self, source, dest):
```
sphuber, line 1340: Since this is just an internal method, maybe a better name would be simply `_symlink`?

aiida/transports/plugins/ssh.py
```diff
@@ -437,14 +437,26 @@ def open(self):
...
+
+    def open_file_transport(self):
+        """
+        Open the SFTP channel, and handle error by directing customer to try another transport
+        """
+        from aiida.common.exceptions import InvalidOperation
+
+        try:
+            self._sftp = self._client.open_sftp()
+        except Exception:
```
sphuber, line 453: Can we catch a more specific exception here? Is it predictable what would be raised? What did it raise in your case?

aiida/transports/plugins/ssh.py
sphuber, line 454: I would use the full entry point name to make it more clear what `sshonly` actually is, e.g.:
`Try setting up the computer with the `aiida.schedulers:ssh_only` plugin. Note that I would also call it `ssh_only` which is more readable, I think

aiida/transports/plugins/ssh.py
```diff
@@ -533,6 +545,22 @@ def normalize(self, path='.'):
         """
         return self.sftp.normalize(path)
 
+    def stat(self, path):
+        """
+        Wrap SFTP stat call
```
sphuber, line 550: It would be more useful to know exactly what this method is supposed to do and what it returns. The fact that it wraps `sftp.stat` in this implementation is not that useful. Since you had to implement this in pure ssh I think you know what it should return. Could you please add that in the docstring instead? Same for `lstat`.

sphuber (changes requested):
Thanks a lot @adegomme in addition to @ramirezfranciscof comments, I have some other small comments, but other than that this looks good and almost ready to go.

aiida/transports/plugins/ssh.py
sphuber: Also interested to know why @adegomme thinks this should be cropped.

adegomme: To send a file with the sshonly plugin, the resulting command is equivalent to `ssh remote  "echo '$(cat localfile)' | cat > remotefile" ` , with the '$(cat localfile)' being a python read instead, as I never managed to have the cat version behave correctly.
This logger call means that the whole content of the file is printed in the log :
example when making a 'verdi computer test'
09/17/2020 10:17:12 AM <1120861> aiida.transport.SshOnlyTransport: [DEBUG] Command to be executed: cd '/path/aiida' && echo 'Test from '"'"'verdi computer test'"'"' on 2020-09-17T10:17:05.830707' | cat > /path/aiida/tmpbtteknws

So this adds the whole file content to the log output when debug is activated, which is not great. Every input file, structure file, pseudopotential file gets printed in the stdout or the log when debug is on, and when several MB get dumped on you, this can be troublesome.
Cropping is meant to avoid ruining the log and the terminal. 1000 char is a bit arbitrary, and maybe we could add a comment when it's cropped ( simply "...", or "cropped to 1000 chars to maintain readability") to avoid confusion. 
Another idea would be to have an optional parameter on _exec_command_internal  to disable this log message, but you would not see any file upload log in this case.

aiida/transports/plugins/ssh.py
adegomme: Indeed. Done in 81319e4

aiida/transports/plugins/ssh.py
adegomme: in my case it was   paramiko.ssh_exception.SSHException: Channel closed. I'll catch only this, as it's apparently the one raised in this case. Done in 13fa71a

aiida/transports/plugins/ssh.py
adegomme: I added the ssh_transports:sshonly bit. The name might be changed soon, I have to check a few things before

aiida/transports/plugins/ssh.py
adegomme: I added some comments for these calls in  88b88c9

aiida/transports/plugins/ssh.py
adegomme: I changed the port name in the plugin, published it, it's now ssh_transports:ssh_only. This is reflected in 
9f52676

aiida/transports/plugins/ssh.py
sphuber: I see. Thanks for the explanation. In any case, even if we keep this, we should definitely comment this, because it is not obvious at all, especially given that the actual use case comes from a subclass that lives in another plugin package. @giovannipizzi do you see a way around this? If we make the cropping size an optional argument to the function, at least `SshOnly` can override the copy file method to call `_exec_command_internal` to pass `max_log_size=1000` or something to that effect.

sphuber:
Thanks for all the changes @adegomme . I am happy with the current state, except I would like to think a bit more about the cropping of the logging in the `_exec_internal_command`. Once we have the input of @giovannipizzi I think this should be ready to go.

aiida/transports/plugins/ssh.py
giovannipizzi: After discussing with @sphuber I think a good solution is to have a class attribute, e.g. `_MAX_EXEC_COMMAND_LOG_SIZE=None` in this base class.
Here the logic would not crop if `self._MAX_EXEC_COMMAND_LOG_SIZE` is None, otherwise it would crop to this length (`command_to_execute[:self._MAX_EXEC_COMMAND_LOG_SIZE]`).

In this way the current behaviour is unchanged, and in your subclass @adegomme  you can just set `_MAX_EXEC_COMMAND_LOG_SIZE` to the value you want.

@adegomme what do you think? Are you OK to do this small change? After, it's ready to merge for me

giovannipizzi (changes requested):

adegomme:
The _MAX_EXEC_COMMAND_LOG_SIZE seems to be a great solution, I will do this immediately.

sphuber (approved):

sphuber:
Thanks a lot @adegomme !

PR #4366, opened 2020-09-09
Implement GaussianCubeData data type.

### LINKED ISSUE #4329: New CubeFileData class.
[Gaussian cube file](http://paulbourke.net/dataformats/cube/) is probably the most common format to represent the 3D mesh data in computational molecular science. Many codes are capable of writing/reading the cube files, therefore it makes sense to provide support for the file format natively in AiiDA.

## API
I propose the following code API:

```python
    GaussianCubeData(ArrayData)  # The class should be derived from the ArrayData
    def __init__():
        self.comment  # The contents extracted from the comment lines.
        self.origin  # Position of the origin of the volumeric data (bohr)
        self.voxel  # 3 vectors defining a voxel  (bohr)
        self.atomic_numbers  # Atomic nubers.
        self.atomic_coordinates  # Atomic coordinates.
        self.data  #  3D array containing the data.
        self.data_units  # Can be defined by the parser.
    def read(file)  # Read the cube file.
    def write(file)  # Write the cube file.
    def get_structuredata()  # Returns structure data object extracted from the cube file.
    def crop_data(threshold_min=None, threshold_max=None)  # Crop the mesh according to the threshold value, adap the position of the structure.
    def clip_data(absmin, absmax)  # Set values lower than absmin to zero, set values higher than absmax to absmax
    def import_from_...()  # Import data from another file formats. For instance XSF
```

## Current situation

For the moment there is no standard way to store the mesh data in AiiDA. Possible alternatives are to either use `ArrayData` object or to use a `FileData` object. The problem with those two approaches that they are not standardized and, therefore, do not always come with a clear way of using them.

yakutovicha:
fixes #4329 

The class is meant to deal with Gaussian Cube files. The current
implementation is capable of reading and exporting the cube files.
Additionally, a minimal checker is added to verify the consistency
of data.

### DIFF
--- a/aiida/orm/nodes/data/array/cube.py
+++ b/aiida/orm/nodes/data/array/cube.py
@@ -193,7 +193,7 @@ class GaussianCubeData(ArrayData):
 
         return all_good and super()._validate(), return_msg
 
-    def export(self, path, fileformat='cube', encoding='utf-8', overwrite=False, **kwargs):
+    def export(self, path, fileformat='cube', overwrite=False, **kwargs):
         """Export data to the selected file."""
 
         if not self._validate():
@@ -204,7 +204,7 @@ class GaussianCubeData(ArrayData):
 
         if path:
             if overwrite or not os.path.isfile(path):
-                with open(path, mode='w', encoding=encoding) as fobj:
+                with open(path, mode='w', encoding='utf-8') as fobj:
                     fobj.write(content)
             else:
                 raise FileExistsError(
--- a/aiida/orm/nodes/data/array/cube.py
+++ b/aiida/orm/nodes/data/array/cube.py
@@ -180,21 +180,23 @@ class GaussianCubeData(ArrayData):
         if data_units:
             self.data_units = data_units
 
-    def everything_ok(self):
+    def _validate(self):
         """Check if the object contains all the data to construct a valid cube file."""
         all_good = True
+        return_msg = ''
         for attr in ['comment', 'voxel', 'origin', 'atomic_numbers', 'atomic_charges', 'atomic_coordinates', 'data']:
             try:
                 getattr(self, attr)
             except (AttributeError, KeyError):
                 all_good = False
-                print(f'The attribute `{attr}` is not set.')
-        return all_good
+                return_msg += f'The attribute `{attr}` is not set.\n'
+
+        return all_good and super()._validate(), return_msg
 
     def export(self, path, fileformat='cube', encoding='utf-8', overwrite=False, **kwargs):
         """Export data to the selected file."""
 
-        if not self.everything_ok():
+        if not self._validate():
             raise ValueError("Can't export the cube file.")
 
         if fileformat == 'cube':
--- a/aiida/orm/nodes/data/array/cube.py
+++ b/aiida/orm/nodes/data/array/cube.py
@@ -156,17 +156,13 @@ class GaussianCubeData(ArrayData):
         self.voxel = voxel
 
         # Atomic numbers, charges and positions.
-        numbers = np.empty(natoms, int)
-        charges = np.empty(natoms, int)
-        positions = np.empty((natoms, 3))
+        num_charge_pos = np.empty(natoms, int), np.empty(natoms, int), np.empty((natoms, 3))
         for i in range(natoms):
             line = next(fiter).split()
-            numbers[i] = int(line[0])
-            charges[i] = float(line[1])
-            positions[i] = [float(s) for s in line[2:]]
-        self.atomic_numbers = numbers
-        self.atomic_charges = charges
-        self.atomic_coordinates = positions
+            num_charge_pos[0][i] = int(line[0])
+            num_charge_pos[1][i] = float(line[1])
+            num_charge_pos[2][i] = [float(s) for s in line[2:]]
+        self.atomic_numbers, self.atomic_charges, self.atomic_coordinates = num_charge_pos
 
         # Data mesh.
         data = np.empty(shape[0] * shape[1] * shape[2], dtype=float)
--- a/aiida/orm/nodes/data/array/cube.py
+++ b/aiida/orm/nodes/data/array/cube.py
@@ -1,0 +1,246 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""
+Module of the GaussianCubeData class, defining the AiiDA data type for storing
+Gaussian cube files.
+"""
+import os
+import sys
+import numpy as np
+
+from .array import ArrayData
+
+
+class GaussianCubeData(ArrayData):
+    """Class that allows managing gaussian cube files."""
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        comment=None,
+        voxel=None,
+        origin=None,
+        atomic_numbers=None,
+        atomic_charges=None,
+        atomic_coordinates=None,
+        data=None,
+        data_units=None,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
+        if comment:
+            self.comment = comment
+        if voxel:
+            self.voxel = voxel
+        if origin:
+            self.origin = origin
+        if atomic_numbers:
+            self.atomic_numbers = atomic_numbers
+        if atomic_charges:
+            self.atomic_charges = atomic_charges
+        if atomic_coordinates:
+            self.atomic_coordinates = atomic_coordinates
+        if data:
+            self.data = data
+        if data_units:
+            self.data_units = data_units
+
+    @property
+    def comment(self):
+        """First two lines of the cube file that contain the comment"""
+        return self.get_attribute('comment')
+
+    @comment.setter
+    def comment(self, comment):
+        if isinstance(comment, (list, tuple)) and len(comment) == 2:
+            if isinstance(comment[0], str) and isinstance(comment[1], str):
+                self.set_attribute('comment', comment)
+                return
+        raise ValueError(f'The attibue `comment` must be a list or a tuple containing two strings, got: {comment}')
+
+    @property
+    def voxel(self):
+        """Three vectors defining a voxel  (bohr)"""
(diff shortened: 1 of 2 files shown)


yakutovicha:
@sphuber before I continue with the implementation, I would like to discuss/agree on a few points:

1. In this PR I would like to provide only the minimal implementation, containing `__init__`, `read`, `export`, and `get_structure_data`. The tests will follow, of course. The clipping-cropping functionality I would like to add in a separate PR.
1. I took a slightly different approach with respect to the one [you proposed](https://github.com/aiidateam/aiida-core/issues/4329#issuecomment-682022863). Essentially, as the cube file format is well defined it is quite easy to create a checker function (which I call `everything_ok`, can be renamed of course). This way the user can populate the object either at its creation, or by calling the `read` method, or by populating the attributes one by one. The `everything_ok` method can be called any time to check the consistency of the data.

The second point is more about the API, I am not sure who else has an opinion on that matter. Pinging @giovannipizzi, @chrisjsewell, @csadorf, @greschd just in case.

greschd:
Regarding the second point:
Is there a use case for populating the attributes one by one, or mixing `read` with individually set attributes? Otherwise, I would actually prefer the "complete" constructor + classmethod `from_file` approach. 
The reason for that is that having "incomplete" objects (with some attributes missing) can lead to edge cases when a user naively assumes they will always be there.

If we keep the current approach, `everything_ok` should definitely be called before / in `.store()`.

aiida/orm/nodes/data/array/cube.py
```diff
@@ -0,0 +1,246 @@
...
+    @data_units.setter
+    def data_units(self, data_units):
+        self.set_attribute('data_units', data_units)
+
+    def read(self, file, data_units=None):
+        """Parse cube file."""
+
+        # It should be either a string pointing to a file, or a file object.
+        if isinstance(file, str):
+            fiter = iter(open(file).readlines())
```
greschd, line 135: We need to close this file, i.e. use the context manager

```python
with open(file, mode='r') as f_obj:
    fiter = iter(f_obj.readlines())
```

Not sure if we should also specify `encoding='utf-8'` here, or just go with system default encoding. Maybe adding it as an optional keyword argument to `read` would be good - even though it's used only in the `file: str` case.

aiida/orm/nodes/data/array/cube.py
```diff
@@ -0,0 +1,246 @@
...
+        origin=None,
+        atomic_numbers=None,
+        atomic_charges=None,
+        atomic_coordinates=None,
+        data=None,
+        data_units=None,
+        **kwargs
+    ):
+        super().__init__(**kwargs)
+        if comment:
```
greschd, line 37: These checks should all be `is not None`. I'm not sure if any of the inputs can be valid when they are `False`-y, but better to be safe than sorry.

aiida/orm/nodes/data/array/cube.py
```diff
@@ -0,0 +1,246 @@
...
+
+        if not self.everything_ok():
+            raise ValueError("Can't export the cube file.")
+
+        if fileformat == 'cube':
+            content = self._write_gaussian_cube()
+
+        if path:
+            if overwrite or not os.path.isfile(path):
+                with open(path, 'w') as fobj:
```
greschd, line 204: Yeah, here we should also add the encoding option.

yakutovicha:
> The reason for that is that having "incomplete" objects (with some attributes missing) can lead to edge cases when a user naively assumes they will always be there.

The point here is that even after the creation of a "complete" object, a user might want to modify it and, eventually, either make it incomplete or inconsistent. So I think that one would anyways need to check the object before storing/exporting.

greschd:
I don't think I completely understand this point: After the `__init__` (assuming `None` values are not allowed), all the attributes are set. And then we have setters, but no deleters - how do you get to a state where `everything_ok` would be `False`?

The question is if there is a use case where you _need_ to partially construct the object (pass `None` for some attributes). Since this is "just" a data container (i.e., doesn't give you ways to operate on the partial data) I don't see that - but then again I don't know the full context here.

The advantage of enforcing "full" objects is that it simplifies / reduces the interface: the getters will always return the content, instead of potentially raising. In the using code (if I'm just given such an object) I would otherwise explicitly have to check `everything_ok`, and handle the `False` case.

If there's a convincing use case where that is needed - great, let's keep it that way. Otherwise, it's probably not worth the cost of making the interface more complex.

aiida/orm/nodes/data/array/cube.py
```diff
@@ -0,0 +1,246 @@
...
+        for i, line in enumerate(fiter):
+            l_split = line.split()
+            data[cursor:cursor + len(l_split)] = l_split
+            cursor += len(l_split)
+        self.data = data.reshape(shape)
+
+        if data_units:
+            self.data_units = data_units
+
+    def everything_ok(self):
```
sphuber, line 182: Is there a reason you are only checking whether these fields are set and not the values whatsoever? In the setters, the only explicit check is on the comments, to verify they are strings. For all the others, both attributes and arrays, you do not check the value whatsoever. The `set_array` will check that what is passed is a numpy array, so at least the type will be correct, but you could set an empty array, or an array with incommensurate dimensions. The `set_attribute` is worse, because it happily accepts `None`. I am not sure what the usefulness of this validator method is if its checks are very limited. Even though `None` won't be accepted in the constructor, afterwards it is perfectly fine to set it.

aiida/orm/nodes/data/array/cube.py
```diff
@@ -0,0 +1,246 @@
...
+    @property
+    def data_units(self):
+        """Units of the mesh data."""
+        return self.get_attribute('data_units')
+
+    @data_units.setter
+    def data_units(self, data_units):
+        self.set_attribute('data_units', data_units)
+
+    def read(self, file, data_units=None):
```
sphuber, line 130: What is the argument for making this an instance method as opposed to a class method? Why would you want to have to construct an instance first, if after all what you are doing here _is_ constructing a new instance

aiida/orm/nodes/data/array/cube.py
```diff
@@ -0,0 +1,246 @@
...
+        """Parse cube file."""
+
+        # It should be either a string pointing to a file, or a file object.
+        if isinstance(file, str):
```
sphuber, line 134: I wouldn't really provide the option to have either a filename or file object. Easiest and safest is just to demand a filelike object. For a caller it is trivial to open a file handle if they have the filename and it saves a whole lot of problems by not accepting multiple types

aiida/orm/nodes/data/array/cube.py
```diff
@@ -0,0 +1,246 @@
...
+            at_x, at_y, at_z = self.atomic_coordinates[i]
+            filecontent += '{:5d} {:12.6f} {:12.6f} {:12.6f} {:12.6f}\n'.format(
+                self.atomic_numbers[i], self.atomic_charges[i], at_x, at_y, at_z
+            )
+
+        # Data mesh.
+        fmt = ' {:11.4e}'
+        flat_data = self.data.ravel()
+        for line in range(flat_data.shape[0] // 6):
+            filecontent += (' {:11.4e}' * 6 + '\n').format(*flat_data[line * 6:(line + 1) * 6])
```
sphuber, line 242: You could reuse `fmt` for the `{:11.4e}` formatter here

aiida/orm/nodes/data/array/cube.py
```diff
@@ -0,0 +1,246 @@
...
+
+        # Number of atoms and origin.
+        filecontent += '{:5d} {:12.6f} {:12.6f} {:12.6f}\n'.format(
+            natoms, self.origin[0], self.origin[1], self.origin[2]
+        )
+
+        # Data shape and voxel.
+        for i in range(3):
+            filecontent += '{:5d} {:12.6f} {:12.6f} {:12.6f}\n'.format(
+                self.data.shape[i], self.voxel[i][0], self.voxel[i][1], self.voxel[i][2]
```
sphuber, line 228: This is not very efficient, you are calling `self.data` in each iteration (and also further down in the function) which causes a database hit every time. Best to get the data once and assign to local variable. Same for other data that are stored as attributes.

aiida/orm/nodes/data/array/cube.py
```diff
@@ -0,0 +1,246 @@
...
+            except (AttributeError, KeyError):
+                all_good = False
+                print(f'The attribute `{attr}` is not set.')
+        return all_good
+
+    def export(self, path, fileformat='cube', overwrite=False, **kwargs):
+        """Export data to the selected file."""
+
+        if not self.everything_ok():
+            raise ValueError("Can't export the cube file.")
```
sphuber, line 197: Again, don't really see the point of this. There are a number of ways wherein this test will pass, but the `_write_gaussian_cube` will fail. Why do you want to have a separate check that only catches a small portion of cases?

sphuber (changes requested):
Thanks @yakutovicha I haven't gone into detail in the logic of the `read` and `_write_gaussian_data` but have some suggestions on overall design. Also, I have to echo @greschd 's concern. I don't really see why you do not want to have a `from_file` class method and why you would like to provide setters on instances. Could you please explain the use case where having this functionality is so important? As explained by Dominik, this really complicates the design of the class' interface and implementation (and currently there are actually quite some problems with the validation as indicated in my comments) and it would be good to try and understand the justification for this.

giovannipizzi:
About the `everything_ok`, Node classes have a [`_validate` method](https://github.com/aiidateam/aiida-core/blob/d1bc5134261cba62e3fed26801631c696574f2de/aiida/orm/nodes/node.py#L162), that is [called during store](https://github.com/aiidateam/aiida-core/blob/d1bc5134261cba62e3fed26801631c696574f2de/aiida/orm/nodes/node.py#L1028).

So you just need to call your `everything_ok` from inside the `_validate`, overriding (and extending, not replacing) the super class behaviour.

Also, I'm not sure if I would call it `everything_ok`.
If you really want to keep it as a public method (you might not need to, if you only use it internally), maybe just call it `validate()` for consistency?
Otherwise, just implement the logic inside `_validate`.

sphuber:
@yakutovicha I am going through old PRs and see what can be closed or needs updates before merging. What is the status here?  Can you update this soon or shall we close this for now until the design has been fleshed out?

yakutovicha:
Hey, yes, sorry this one got abandoned. I think we agreed to move the class into another SCI-data repositor. Has it already been created? I remember @mbercx was taking care of that.

sphuber:
Ok, then I will close this for now and we can open a new PR into the other repository

mbercx:
> I remember @mbercx was taking care of that.

image

PR #4369, opened 2020-09-11
Ci/continue on errors with 2020 resolver

csadorf:
Updates the `test-install` workflow to not fail when any of the pip-2020 resolver related steps or jobs fails. This is to reduce disruption to our own processes related to that feature.

See this run for what this looks like in practice: https://github.com/aiidateam/aiida-core/actions/runs/253813527

The compromise here is that in order to not block merges, the associated steps and jobs still get a little green check-mark, but there is a warning displayed if issues with the 2020-resolver are encountered. Unfortunately I did not find a better way to indicate issues, because GitHub actions is not really good about this particular use case. If you know of a better way, I'm open to suggestions, but keep in mind that I already spend some time on researching and experimenting with this.

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -43,6 +43,7 @@ jobs:
 
     continue-on-error: ${{ contains(matrix.pip-feature-flag, '2020-resolver') }}
     strategy:
+      fail-fast: false
       matrix:
         pip-feature-flag: [ '', '--use-feature=2020-resolver' ]
         extras: [ '', '[atomic_tools,docs,notebook,rest,tests]' ]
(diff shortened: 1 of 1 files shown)


.github/workflows/test-install.yml
```diff
@@ -41,7 +41,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
 
+    continue-on-error: ${{ contains(matrix.pip-feature-flag, '2020-resolver') }}
     strategy:
+      fail-fast: false
```
ltalirz, line 46: Just out of curiosity: without this, would the workflow fail even if `continue-on-error` was set to true?

ltalirz (approved):
thanks, fine with me.

.github/workflows/test-install.yml
csadorf: No, but it would fail for those where that is not true, which is also undesired behavior in this particular case.

PR #4370, opened 2020-09-15
`CalcJob`: improve logging in `parse_scheduler_output`

sphuber:
The level of the log that is fired if no detailed job info is available
is changed from `WARNING` to `INFO`. Since not all schedulers implement
the feature of retrieving this detailed job info, such as the often used
`DirectScheduler`, using a warning is not very apt. If the information
is missing, nothing is necessarily wrong, so `INFO` is better suited.

On the contrary, if the `Scheduler.parse_output` excepts, that is grave
and so its level is changed from a warning to an error.

Finally, a new condition is added where the scheduler does implement the
method to retrieve the detailed job info, but the command fails. In this
case, the return value will be non-zero. This value is now checked
explicitly and if the case, a info log is fired and the detailed job
info is set to `None`, which will cause the parsing to be skipped. This
case can for example arise when using the `SlurmScheduler` plugin, which
does implement the detailed job info feature, however, not all SLURM
installations have the job accounting feature enabled, which is required
by the plugin.


sphuber:
@ramirezfranciscof could you review this please? This needs to go in before the next release which we would like to do this week

ramirezfranciscof:
Yes, sure. One general question meanwhile I see the code:

> The level of the log that is fired if no detailed job info is available is changed from `WARNING` to `INFO`. Since not all schedulers implement the feature of retrieving this detailed job info, such as the often used `DirectScheduler`, using a warning is not very apt. If the information is missing, nothing is necessarily wrong, so `INFO` is better suited.

Isn't this more of a design issue? Meaning, that this was set to be a `WARNING` because we expected the schedulers to implement this feature? In which case I would expect some comment regarding why we changed our mind and why we think the detailed job info is not that relevant after all.

sphuber:
> Isn't this more of a design issue? Meaning, that this was set to be a `WARNING` because we expected the schedulers to implement this feature? In which case I would expect some comment regarding why we changed our mind and why we think the detailed job info is not that relevant after all.

I think it was set to `WARNING` without too much thought. The detailed job info, as it is currently defined, is optional and with good reason, as not all scheduler _can_ implement it. Even for SLURM, that does have a way to get detailed job info at the end of a job, this is not enabled for default SLURM configurations and so sometimes will work and sometimes won't. I am therefore convinced that the missing of the detailed job info should not be a warning, because it is expected and correct functionality of AiiDA does not rely on it. It's is just an extra, that when available, it can provide more value.

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -358,11 +362,11 @@ def parse_scheduler_output(self, retrieved):
...
         try:
             exit_code = scheduler.parse_output(detailed_job_info, scheduler_stdout, scheduler_stderr)
-        except exceptions.FeatureNotAvailable as exception:
-            self.logger.warning('could not parse scheduler output: {}'.format(exception))
+        except exceptions.FeatureNotAvailable:
+            self.logger.info('`{}` does not implement scheduler output parsing'.format(scheduler.__class__.__name__))
             return
         except Exception as exception:  # pylint: disable=broad-except
-            self.logger.warning('the `parse_output` method of the scheduler excepted: {}'.format(exception))
+            self.logger.error('the `parse_output` method of the scheduler excepted: {}'.format(exception))
```
ramirezfranciscof, line 369: Would it be easy to add a test that tries to catch this exception? You would have to monkeypatch the parser/calcjob?

tests/engine/test_calc_job.py
```diff
@@ -354,62 +354,106 @@ def test_parse_retrieved_folder(self):
...
 
-        Here we check explicitly that the parsing does not except even if the scheduler does not implement the method.
-        """
-        process = self.instantiate_process()
-        retrieved = orm.FolderData().store()
-        retrieved.add_incoming(process.node, link_label='retrieved', link_type=LinkType.CREATE)
 
-        process.node.set_attribute('detailed_job_info', {})
+@pytest.mark.usefixtures('clear_database_before_test', 'override_logging')
+def test_parse_insufficient_data(process):
```
ramirezfranciscof, line 384: So, this is maybe separate from this PR but just to learn: what is the difference between passing a fixture via `@pytest.mark.usefixtures` and passing it as a parameter of the test?

tests/engine/test_calc_job.py
```diff
@@ -354,62 +354,106 @@ def test_parse_retrieved_folder(self):
...
+    """Test the scheduler output parsing logic in `CalcJob.parse`.
+
+    This is testing the case where the `detailed_job_info` is incomplete because the call failed. This is checked
+    through the return value that is stored within the attribute dictionary.
+    """
+    retrieved = orm.FolderData().store()
+    retrieved.add_incoming(process.node, link_label='retrieved', link_type=LinkType.CREATE)
+
+    process.node.set_attribute('detailed_job_info', {'retval': 1, 'stderr': 'accounting disabled', 'stdout': ''})
+    process.parse()
```
ramirezfranciscof, line 422: So, isn't the `detailed_job_info` overwritten by the `parse` method of the calculation? Also, I am assuming that you are setting the `retrieved` folder in all these tests because it is necessary for this method.

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -337,8 +337,12 @@ def parse_scheduler_output(self, retrieved):
         filename_stdout = self.node.get_option('scheduler_stdout')
 
         detailed_job_info = self.node.get_detailed_job_info()
+
         if detailed_job_info is None:
-            self.logger.warning('could not parse scheduler output: the `detailed_job_info` attribute is missing')
+            self.logger.info('could not parse scheduler output: the `detailed_job_info` attribute is missing')
+        elif detailed_job_info.get('retval', 0) != 0:
+            self.logger.info('could not parse scheduler output: return value of `detailed_job_info` is non-zero')
+            detailed_job_info = None
```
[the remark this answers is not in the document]
sphuber: > Isn't this a breaking change? Are we sure there is no other place in the code that will check the `retval` and have a different behaviour if its `None` or a non zero integer?

This can't be breaking because the code that is changed hasn't been released yet.

> Also, isn't a value other than 0 a parse that actually had a problem, and thus worthy of a `warning` type of log?

As described in the commit message, the command can fail just because the scheduler installation does not support it. This is for example the case for vanilla SLURM schedulers. Since retrieving and parsing the detailed job info is completely optional, I don't think it should be a warning if it fails. Ideally, we could detect the special case of it failing for SLURM just not having accounting enabled, but we can't in the generic way, and there may be comparable cases for other schedulers.

tests/engine/test_calc_job.py
sphuber: Re the `retrieved` folder, yes this is required by `parse` so that is why I add it. As for the `detailed_job_info`, I am not sure what you mean with overwrites? If you mean in the implementation of `parse_scheduler_output` then yes, but it just overwrites the variable in memory, not the attribute. And it only does this to prevent the `scheduler.parse_output` being called, because anyway that is pointless with a detailed job info that does not contain correct information

tests/engine/test_calc_job.py
sphuber: Pass it as function argument if the fixture actually needs to be called in the test body, otherwise use it in `pytest.mark.usefixtures`. The reason is simply that, although the `pytest` magic of automatically injecting fixtures into a test by defining them as an argument, it leads to many unused function arguments which is weird to read, and leads to `pylint` complaining, for the very same reason.

aiida/engine/processes/calcjobs/calcjob.py
sphuber: Yeah, should be easy enough to monkey path the parser, but checking logs can be a bit trickier. But I will give it a shot.

sphuber:
@ramirezfranciscof I added the test

aiida/engine/processes/calcjobs/calcjob.py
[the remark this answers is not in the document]
ramirezfranciscof: Ahh, ok, I missunderstood that last paragraph. I said nothing then.

ramirezfranciscof (approved):
Awesome, thanks!

tests/engine/test_calc_job.py
ramirezfranciscof: Mmmm this I didn't fully understand yet though. What do you mean the variable in memory but not the attribute? I don't get the distinction you are making.

sphuber: I was responding to your question:

> So, isn't the detailed_job_info overwritten by the parse method of the calculation?

which I didn't understand. So I guess now we are going in circles of confusion :)

ramirezfranciscof: Hahaha, I think we can try to clear this out tomorrow. BTW I need to send the email...

PR #4371, opened 2020-09-15
Update citations in `README.md` and documentation landing page

sphuber:
The second AiiDA paper was published in Scientific Data on September 8,
2020. The suggested citations are updated, where the original AiiDA
paper is kept to be cited when people use AiiDA with version before v1.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,13 +51,10 @@ If you are experiencing problems with your AiiDA installation, please refer to t
 
 ## How to cite
 
-If you use AiiDA in your research, please consider citing the AiiDA paper:
+If you use AiiDA in your research, please consider citing the following publications:
 
-> Giovanni Pizzi, Andrea Cepellotti, Riccardo Sabatini, Nicola Marzari,
-> and Boris Kozinsky, *AiiDA: automated interactive infrastructure and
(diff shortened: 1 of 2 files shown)


ltalirz (approved):
Thanks @sphuber 

Note that you used plural in the readme (publications) but singular in the docs (work).
Singular works fine, I think, unless we want to ask people to actually cite both.

sphuber:
> Note that you used plural in the readme (publications) but singular in the docs (work).
> Singular works fine, I think, unless we want to ask people to actually cite both.

We discussed with @giovannipizzi and @muhrin at the time, and we decided that we keep the old publication in the list. If they use `aiida-core<1.0.0` or cite the ADES model, they should cite the 2016 paper. Anything else the 2020 one. Note however, that soon we will have the engine paper that will be published. Currently under review but will most likely be accepted. Also there we will encourage people that use `aiida-core>=1.0.0` to cite it.

@giovannipizzi How do you prefer we frame this for now? Do we use singular or plural?

giovannipizzi (approved):
Work can also be used as a collective term, I think. So I'm OK as is.
If we realise it's not clear, or we want to change, we can change as soon as the engine paper is out.

PR #4373, opened 2020-09-16
`CalcJob`: allow nested target paths for `local_copy_list`

### LINKED ISSUE #4350: Transports can't create directories when copying files
### Describe the bug

Both the `local` and `ssh` transport except when folders need to be created for the full path of a copied file. This manifest itself when paths like `dirname/filename.txt` are added to the copy list of calcjobs and the `dirname` folder does not already exist.

### Steps to reproduce

1. Set up a remote computer with `ssh` or `local` transport and configure it. In my case, I have `daint` for `ssh` and `localhost` for `local` transport.
2. Set up a code on those computers by running the following command: `verdi code setup -L bash -P templatereplacer --on-computer --computer=daint --remote-abs-path=/bin/bash --non-interactive` (the only flag you need to customize is the `--computer=localhost`)
3. Run (`verdi run`) the following script:

```
from aiida.orm import Dict, SinglefileData, load_code
from aiida.engine import run_get_node

import tempfile
with tempfile.NamedTemporaryFile(mode='w+') as handle:
    filepath = handle.name
    handle.write('content of file')
    handle.flush()
    sfile_node = SinglefileData(file=filepath).store()

param_dict = Dict(dict={'x': 3, 'y': 4}).store()
templ_dict = Dict(dict={
    'input_file_name': 'script.sh',
    'input_file_template': "echo $(({x} * {y}))",
    'cmdline_params': ['script.sh'],
    'output_file_name': 'stdout.txt',
    #'files_to_copy': [('inpfile','file.txt')],
    #'files_to_copy': [('inpfile','dir1/file.txt')],
    }).store()

code = load_code('bash@daint')
#code = load_code('bash@localhost')

builder = code.get_builder()
builder.parameters = param_dict
builder.template = templ_dict
builder.files = {'inpfile': sfile_node}
builder.metadata.options = {
    'resources': {'num_machines': 1, 'tot_num_mpiprocs': 1},
    'custom_scheduler_commands': '#SBATCH --constraint=mc\n#SBATCH --cpus-per-task=36',
    'max_wallclock_seconds': 60,
    }

print('param dict:  {}'.format(param_dict.uuid))
print('templ dict:  {}'.format(templ_dict.uuid))
print('sfile node:  {}'.format(sfile_node.uuid))

results, calcjobnode = run_get_node(builder)
print('calculation: {}'.format(calcjobnode.uuid))
```

You can change the two commented sections. The first one lets you check that setting a path that doesn't need to create folders will work. The second one is for testing the `ssh` or `local` transports and see they behave the same.

### Expected behavior

I would expect that all necessary directories would be automatically created if they don't exist. This is because there are some programs that will require a certain file structure, and there is no independent way of just creating folders.

Instead I get this error message for `ssh` transport:

```
09/04/2020 11:17:51 AM <2377> aiida.engine.transports: [ERROR] Exception whilst using transport:
  File "/home/framirez/Workgit/aiida-core/aiida/engine/transports.py", line 103, in request_transport
  File "/home/framirez/Workgit/aiida-core/aiida/engine/processes/calcjobs/tasks.py", line 83, in do_upload
    execmanager.upload_calculation(node, transport, calc_info, folder)
  File "/home/framirez/Workgit/aiida-core/aiida/engine/daemon/execmanager.py", line 205, in upload_calculation
    transport.put(handle.name, target)
  File "/home/framirez/Workgit/aiida-core/aiida/transports/plugins/ssh.py", line 765, in put
    self.putfile(localpath, remotepath, callback, dereference, overwrite)
  File "/home/framirez/Workgit/aiida-core/aiida/transports/plugins/ssh.py", line 792, in putfile
    return self.sftp.put(localpath, remotepath, callback=callback)
  File "/home/framirez/.virtualenvs/aiida-core/lib/python3.6/site-packages/paramiko/sftp_client.py", line 759, in put
    return self.putfo(fl, remotepath, file_size, callback, confirm)
  File "/home/framirez/.virtualenvs/aiida-core/lib/python3.6/site-packages/paramiko/sftp_client.py", line 714, in putfo
    with self.file(remotepath, "wb") as fr:
  File "/home/framirez/.virtualenvs/aiida-core/lib/python3.6/site-packages/paramiko/sftp_client.py", line 372, in open
    t, msg = self._request(CMD_OPEN, filename, imode, attrblock)
  File "/home/framirez/.virtualenvs/aiida-core/lib/python3.6/site-packages/paramiko/sftp_client.py", line 813, in _request
    return self._read_response(num)
  File "/home/framirez/.virtualenvs/aiida-core/lib/python3.6/site-packages/paramiko/sftp_client.py", line 865, in _read_response
    self._convert_status(msg)
  File "/home/framirez/.virtualenvs/aiida-core/lib/python3.6/site-packages/paramiko/sftp_client.py", line 894, in _convert_status
    raise IOError(errno.ENOENT, text)
FileNotFoundError: [Errno 2] No such file
```

And this one for `local` transport:

```
09/04/2020 11:07:39 AM <1447> aiida.engine.transports: [ERROR] Exception whilst using transport:
  File "/home/framirez/Workgit/aiida-core/aiida/engine/transports.py", line 103, in request_transport
  File "/home/framirez/Workgit/aiida-core/aiida/engine/processes/calcjobs/tasks.py", line 83, in do_upload
    execmanager.upload_calculation(node, transport, calc_info, folder)
  File "/home/framirez/Workgit/aiida-core/aiida/engine/daemon/execmanager.py", line 205, in upload_calculation
    transport.put(handle.name, target)
  File "/home/framirez/Workgit/aiida-core/aiida/transports/plugins/local.py", line 296, in put
    self.putfile(localpath, full_destination, overwrite)
  File "/home/framirez/Workgit/aiida-core/aiida/transports/plugins/local.py", line 333, in putfile
    shutil.copyfile(localpath, the_destination)
  File "/home/framirez/.virtualenvs/aiida-core/lib/python3.6/shutil.py", line 121, in copyfile
    with open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: '/scratch/framirez/aiida/49/17/2007-d8be-48f5-8c9b-80c1d7d222ff/dir1/file.txt'
```

### Your environment

- OS: Ubuntu 18.04.5 LTS
- Python 3.6.9
- aiida-core develop branch, commit 7c31bc2d54df8ee22db2fd20f6078a44ba871a89.

sphuber:
Fixes #4350

If a `CalcJob` would specify a `local_copy_list` containing an entry
where the target remote path contains nested subdirectories, the
`upload_calculation` would except unless all subdirectories would
already exist. To solve this, one could have added a transport call that
would create the directories if the target path is nested. However, this
would risk being very inefficient if there are many local copy list
instructions with relative path, where each would incurr a command over
the transport.

Instead, we change the design and simply apply the local copy list
instructions to the sandbox folder on the local file system. This also
at the same time allows us to get rid of the inefficient workaround of
writing the file to a temporary file, because the transport interface
doesn't accept filelike objects and the file repository does not expose
filepaths on the local file system.

The only additional thing to take care of is to make sure the files from
the local copy list do not end up in the repository of the node, which
was the whole point of the `local_copy_list`'s existence in the first
place. But this is solved by simply adding each file, that is added to
the sandbox, also to the `provenance_exclude_list`.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -206,27 +206,6 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
             logger.debug('[submission of calculation {}] copying file/folder {}...'.format(node.pk, filename))
             transport.put(folder.get_abs_path(filename), filename)
 
-    if dry_run:
-        if remote_copy_list:
-            with open(os.path.join(workdir, '_aiida_remote_copy_list.txt'), 'w') as handle:
-                for remote_computer_uuid, remote_abs_path, dest_rel_path in remote_copy_list:
-                    handle.write(
-                        'would have copied {} to {} in working directory on remote {}'.format(
-                            remote_abs_path, dest_rel_path, computer.label
-                        )
-                    )
-
-        if remote_symlink_list:
-            with open(os.path.join(workdir, '_aiida_remote_symlink_list.txt'), 'w') as handle:
-                for remote_computer_uuid, remote_abs_path, dest_rel_path in remote_symlink_list:
-                    handle.write(
-                        'would have created symlinks from {} to {} in working directory on remote {}'.format(
-                            remote_abs_path, dest_rel_path, computer.label
-                        )
-                    )
-
-    else:
-
         for (remote_computer_uuid, remote_abs_path, dest_rel_path) in remote_copy_list:
             if remote_computer_uuid == computer.uuid:
                 logger.debug(
@@ -268,6 +247,25 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
                     'It is not possible to create a symlink between two different machines for '
                     'calculation {}'.format(node.pk)
                 )
+    else:
+
+        if remote_copy_list:
+            with open(os.path.join(workdir, '_aiida_remote_copy_list.txt'), 'w') as handle:
+                for remote_computer_uuid, remote_abs_path, dest_rel_path in remote_copy_list:
+                    handle.write(
+                        'would have copied {} to {} in working directory on remote {}'.format(
+                            remote_abs_path, dest_rel_path, computer.label
+                        )
+                    )
+
+        if remote_symlink_list:
+            with open(os.path.join(workdir, '_aiida_remote_symlink_list.txt'), 'w') as handle:
+                for remote_computer_uuid, remote_abs_path, dest_rel_path in remote_symlink_list:
+                    handle.write(
+                        'would have created symlinks from {} to {} in working directory on remote {}'.format(
+                            remote_abs_path, dest_rel_path, computer.label
+                        )
+                    )
 
     # Loop recursively over content of the sandbox folder copying all that are not in `provenance_exclude_list`. Note
     # that directories are not created explicitly. The `node.put_object_from_filelike` call will create intermediate
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -14,6 +14,7 @@ the routines make reference to the suitable plugins for all
 plugin-specific operations.
 """
 import os
+import shutil
 
 from aiida.common import AIIDA_LOGGER, exceptions
 from aiida.common.folders import SandboxFolder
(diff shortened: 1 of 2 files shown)


aiida/engine/daemon/execmanager.py
```diff
@@ -196,13 +192,16 @@ def find_data_node(inputs, uuid):
...
+            provenance_exclude_list.append(target)
+
+    # In a dry_run, the working directory is the raw input folder, which will already contain these resources
+    if not dry_run:
+        for filename in folder.get_content_list():
+            logger.debug('[submission of calculation {}] copying file/folder {}...'.format(node.pk, filename))
+            transport.put(folder.get_abs_path(filename), filename)
 
     if dry_run:
         if remote_copy_list:
```
ramirezfranciscof, line 207: So, you moved this here and now you have a `dry_run` check right next to another; would it perhaps be better to unify these two conditional branches? (the first is just `if not dry_run`, but the second has an `else` clause in line 225).

ramirezfranciscof (commented):
So, apparently it all worked basically with the modifications we discusses as you had anticipated! I only have 2 comments: one I already did in the code, the second is if it would be possible to add a test that checks this now allows to generate directories by providing full paths in the `local_copy_file`.

ramirezfranciscof:
So, I've also noticed that in the if section I mentioned in the review there is a part where if you are doing a dry run, it would copy the files requested from the remote machine locally to your sandbox folder. It would seem to me that this part would be easy to modify if we wanted to allow a list for copying files from different machines (it would just do this code even if it was not a dry run and the machine uuid was different, and then it would add the files to the provenance exlude list as we are doing in this PR).

https://github.com/aiidateam/aiida-core/blob/b002a9a9a898a05338ff522b46686657fb468f66/aiida/engine/daemon/execmanager.py#L226-L248

aiida/engine/daemon/execmanager.py
sphuber: Done and also added a test for the nested paths

sphuber:
@ramirezfranciscof thanks for the review. I addressed your comments and this should be ready for review

ramirezfranciscof (approved):
Cool stuff; good to go for me! I'll see if I can rebase transport after you merge and start finishing that up.

PR #4376, opened 2020-09-16
ORM: factor out attributes/extras code into mixin classes for front- end backend

mbercx:
Currently, the methods for `attributes` and `extras` of the `Node` backend entity classes are implemented for both the Django and SqlAlchemy backend classes (`DjangoNode` and `SqlaNode`). Since the methods are very similar, it's better to homogenize them and introduce mixin classes which we can add as parents to the `BackendNode` class. This has the added advantage that in case we want to add `attributes` or `extras` to other entities, we only have to add either mixin class as a parent to the backend entity class to achieve this.

We can do the same for the front-end `Node` class, however here we only create the mixin class for the `extras`, as we are not planning to add `attributes` to other entities for now, we only add the `EntityExtrasMixin` class, since the plan is to add `extras` to the `Group` entity (see #4328).

This PR does the following:

* Homogenize the attributes/extras methods for the `DjangoNode` and `SqlaNode` backend classes.
* Move all methods of the `DjangoNode` and `SqlaNode` backend classes related to `attributes` and `extras` to two mixin classes: AttributesBackendEntity and ExtrasBackendEntity. Add these as parent classes to the `BackendNode` class.`
* Move the methods of the `Node` front-end class related to `extras` to the `EntityExtrasMixin` class.


mbercx:
@sphuber thanks for all the help with the reviewing so far! I still have two questions:
* As you can see, I've only made a front-end mixin class for the `extras`, as we don't plan to add `attributes` to other entities for now. It might be a little strange to make a mixin class for the `attributes`, when these methods are currently only used for one front-end entity. However, it would be more consistent to do so. What do you think? 
* I can't seem to fix this pylint issue related to the __all__variable in `aiida.orm.implementation.backends` (`undefined-all-variable`). Can you spot what I'm missing here? :sweat_smile:

sphuber:
I could have sworn I typed what I am about to type yesterday evening already, but apparently it didn't get through. So I will redo it :)

> thanks for all the help with the reviewing so far! I still have two questions:

Right back at you! Great work so far.

> As you can see, I've only made a front-end mixin class for the `extras`, as we don't plan to add `attributes` to other entities for now. It might be a little strange to make a mixin class for the `attributes`, when these methods are currently only used for one front-end entity. However, it would be more consistent to do so. What do you think?

I see your point, but there is an additional advantage of separating it into a separate file: it actually makes the code more readable. As it stands, the `Node` class is huge and contains a lot of implementation. Having the full attribute functionality in a separate class, the mixin, will make the code easier to work with. In addition, this brings it on par with the extras, even though (for now) the attributes mixin will only be used by the `Node`. I think this is fine though and the preferable solution.

>     * I can't seem to fix this pylint issue related to the __all__variable in `aiida.orm.implementation.backends` (`undefined-all-variable`). Can you spot what I'm missing here?

I have no idea. It looks fine. This might actually just be a pylint bug. These are notoriously hard to fix or simply even get rid off. I will try to do some testing locally.

In the meantime, since this PR will need to keep more than 1 commit, and so I won't be able to squash, it will be important to have the commits exactly as they will have to be merged. Your first two commits are already perfect. I would just suggest to add a third on top of that, that simply creates the front end mixins. Then you just have to put the fixes of the additional commits and rebase them in the respective commits where they are supposed to go.

Finally, I would maybe prefix each commit title with "ORM:" such that it is immediately clear what part of the code it touches.
E.g.:

* ORM: homogenize attributes/extras methods of backend node
* ORM: move attributes/extras methods of backend node to mixins
* ORM: move attributes/extras methods of frontend node to mixins

mbercx:
> I see your point, but there is an additional advantage of separating it into a separate file: it actually makes the code more readable. As it stands, the `Node` class is huge and contains a lot of implementation. Having the full attribute functionality in a separate class, the mixin, will make the code easier to work with. In addition, this brings it on par with the extras, even though (for now) the attributes mixin will only be used by the `Node`. I think this is fine though and the preferable solution.

Agreed, the `Node` class is pretty huge. 👍 

> I have no idea. It looks fine. This might actually just be a pylint bug. These are notoriously hard to fix or simply even get rid off. I will try to do some testing locally.

My local pre-commit runs don't actually raise this issue. Might be related to the pylint version we are using?

> In the meantime, since this PR will need to keep more than 1 commit, and so I won't be able to squash, it will be important to have the commits exactly as they will have to be merged. Your first two commits are already perfect. I would just suggest to add a third on top of that, that simply creates the front end mixins. Then you just have to put the fixes of the additional commits and rebase them in the respective commits where they are supposed to go.

Done! Let me know if you have any further comments.

aiida/orm/implementation/entities.py
```diff
@@ -0,0 +1,451 @@
...
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Classes and methods for backend non-specific entities"""
+import abc
+import typing
+
+from aiida.orm.implementation.utils import clean_value, validate_attribute_extra_key
+
+__all__ = ('BackendEntity', 'BackendCollection', 'EntityType', 'BackendEntityAttributesMixin', \
```
sphuber, line 16: Just small reference for the future (don't need to adapt it here): the slash is not necessary when you are in between parentheses.

sphuber:
Thanks a lot @mbercx , this is perfect. All we need to do is figure out this pylint bug and then rebase it on latest develop and we can merge this. I will also see if I can find a solution for the pylint thing, but will have to wait until Tuesday.

sphuber (approved):
Thanks a lot @mbercx ! The final problem with the pylint false positive on the `__all__` in the `aiida.orm.implementation.backends` had to do with the fact that the `aiida.orm.implementation.sql` module _also_ defines a `backends` module and imports the base backends. Some complex process of import order and potential overwriting, caused `pylint` to not understand what was going on. Removing the `__all__` from the `aiida.orm.implementation.sql` module fixed it. It is ok because that module is not used outside the backend and doesn't really need a shorthand import path

PR #4377, opened 2020-09-16
🧪 Skip code CI if docs only PR

### LINKED ISSUE #3055: Speeding up Travis tests
It would be nice if we could skip code tests in Travis for PRs, where _only_ files under `aiida_core.docs.source.` have been changed, if possible. Automatically accepting them as OK.
This would save a lot of time on Travis for docs updates.

Naturally, this cannot be done when changing the code, since some docs are auto-generated from the code.

chrisjsewell:
Split the `ci.yml` into `ci-code.yml` and `ci-style.yml`, so that ci-code jobs can be skipped if the only change in the PR are to files in the docs folder.

closes #3055

### DIFF
--- a/.github/workflows/ci-code.yml
+++ b/.github/workflows/ci-code.yml
@@ -5,37 +5,9 @@ on:
     branches-ignore: [gh-pages]
   pull_request:
     branches-ignore: [gh-pages]
+    paths-ignore: ['docs/**']
 
 jobs:
-
-  pre-commit:
-
-    runs-on: ubuntu-latest
-    timeout-minutes: 30
-
-    steps:
-    - uses: actions/checkout@v2
-
-    - name: Set up Python 3.8
-      uses: actions/setup-python@v2
-      with:
-        python-version: 3.8
-
-    - name: Install system dependencies
-      run: |
-        sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
-        sudo apt update
-        sudo apt install libkrb5-dev ruby ruby-dev
-
-    - name: Install python dependencies
-      run: |
-        pip install numpy==1.17.4
-        pip install -e .[all]
-        pip freeze
-
-    - name: Run pre-commit
-      run:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-code.yml
```diff
@@ -5,38 +5,10 @@ on:
     branches-ignore: [gh-pages]
   pull_request:
     branches-ignore: [gh-pages]
+    paths-ignore: ['docs/**']
```
sphuber, line 8: How does this paths-ignore work? If no files are changed, the workflow does not even run, correct? And so the paths-ignore will ignore changes in those files. So this workflow will only not run, if there are only changes in the `docs` directory. Is that correct?

chrisjsewell: Yep, it will only run if the PR includes at least one file outside the docs directory
See: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths

csadorf (commented):
What happens when there are style issues with the "code PRs"? I'd prefer that the whole workflow is skipped when our pre-commit checks fail.

chrisjsewell:
> What happens when there are style issues with the "code PRs"? I'd prefer that the whole workflow is skipped when our pre-commit checks fail.

I'm not sure if I follow you here?
As I understand, currently the pre-commit job runs for every PR commit and, if it fails, it does not affect any other jobs/workflows (i.e. causing them to fast-fail or inhibiting them from running.)

This behaviour will not change when moving the pre-commit job to ci-style.yml

csadorf:
> > What happens when there are style issues with the "code PRs"? I'd prefer that the whole workflow is skipped when our pre-commit checks fail.
> 
> I'm not sure if I follow you here?
> As I understand, currently the pre-commit job runs for every PR commit and, if it fails, it does not affect any other jobs/workflows (i.e. causing them to fast-fail or inhibiting them from running.)
> 
> This behaviour will not change when moving the pre-commit job to ci-style.yml

You are absolutely correct. I was under the false assumption that we had made the pass of pre-commit checks requirement for the other jobs. I'm withdrawing my comment. 😉

csadorf (approved):
Looks fine to me.

PR #4378, opened 2020-09-17
Unpin numpy version from 1.17.X to allow the newest version (1.19)

CasperWA:
This is an attempt to try and unpin the hard constraint on `numpy`.

This was introduced in f5d6cba, the reason for it can be seen in the commit message.

I have tried following the [wiki page](https://github.com/aiidateam/aiida-core/wiki/AiiDA-Dependency-Management) on how to update dependencies, and am currently trying the semi-automatic update of the requirements files by creating this PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
             - cache-pip
 
       - run: |
-          pip install numpy==1.17.4
+          # pip install numpy==1.17.4
           pip install --user .[docs,tests]
 
       - save_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,9 +13,7 @@ jobs:
           keys:
             - cache-pip
 
-      - run: |
-          # pip install numpy==1.17.4
-          pip install --user .[docs,tests]
+      - run: pip install --user .[docs,tests]
 
       - save_cache:
           key: cache-pip
--- a/aiida/tools/data/array/kpoints/legacy.py
+++ b/aiida/tools/data/array/kpoints/legacy.py
@@ -342,7 +342,7 @@ def get_explicit_kpoints_path(
     explicit_kpoints = [tuple(point_coordinates[path[0][0]])]
     labels = [(0, path[0][0])]
 
-    if not all([_.is_integer() for _ in num_points]):
+    if not all([_.is_integer() for _ in num_points if isinstance(_, (float, numpy.float64))]):
         raise ValueError('Could not determine number of points as a whole number. num_points={}'.format(num_points))
     num_points = [int(_) for _ in num_points]
 
--- a/aiida/tools/data/array/kpoints/legacy.py
+++ b/aiida/tools/data/array/kpoints/legacy.py
@@ -342,6 +342,10 @@ def get_explicit_kpoints_path(
     explicit_kpoints = [tuple(point_coordinates[path[0][0]])]
     labels = [(0, path[0][0])]
 
+    if not all([_.is_integer() for _ in num_points]):
+        raise ValueError('Could not determine number of points as a whole number. num_points={}'.format(num_points))
+    num_points = [int(_) for _ in num_points]
+
     for count_piece, i in enumerate(path):
         ini_label = i[0]
         end_label = i[1]
--- a/aiida/tools/data/array/kpoints/legacy.py
+++ b/aiida/tools/data/array/kpoints/legacy.py
@@ -342,8 +342,8 @@ def get_explicit_kpoints_path(
     explicit_kpoints = [tuple(point_coordinates[path[0][0]])]
     labels = [(0, path[0][0])]
 
-    if not all([_.is_integer() for _ in num_points if isinstance(_, (float, numpy.float64))]):
-        raise ValueError('Could not determine number of points as a whole number. num_points={}'.format(num_points))
+    assert all([_.is_integer() for _ in num_points if isinstance(_, (float, numpy.float64))]
+               ), 'Could not determine number of points as a whole number. num_points={}'.format(num_points)
     num_points = [int(_) for _ in num_points]
 
     for count_piece, i in enumerate(path):
--- a/.github/workflows/ci-code.yml
+++ b/.github/workflows/ci-code.yml
@@ -137,7 +137,7 @@ jobs:
 
     - name: Install python dependencies
       run: |
-        pip install numpy==1.17.4
+        # pip install numpy==1.17.4
         pip install -e .
 
     - name: Run verdi
--- a/.github/workflows/ci-code.yml
+++ b/.github/workflows/ci-code.yml
@@ -136,9 +136,7 @@ jobs:
         python-version: 3.8
 
     - name: Install python dependencies
-      run: |
-        # pip install numpy==1.17.4
-        pip install -e .
+      run: pip install -e .
 
     - name: Run verdi
       run: |
--- a/.github/workflows/ci-style.yml
+++ b/.github/workflows/ci-style.yml
@@ -29,7 +29,6 @@ jobs:
 
     - name: Install python dependencies
       run: |
-        # pip install numpy==1.17.4
         pip install -e .[all]
         pip freeze
 
--- a/.github/workflows/ci-style.yml
+++ b/.github/workflows/ci-style.yml
@@ -29,7 +29,7 @@ jobs:
 
     - name: Install python dependencies
       run: |
-        pip install numpy==1.17.4
+        # pip install numpy==1.17.4
         pip install -e .[all]
         pip freeze
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
 
     - name: Install python dependencies
       run: |
-        pip install numpy==1.17.4
+        # pip install numpy==1.17.4
         pip install transifex-client sphinx-intl
         pip install -e .[docs,tests]
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,6 @@ jobs:
 
     - name: Install python dependencies
       run: |
-        # pip install numpy==1.17.4
         pip install transifex-client sphinx-intl
         pip install -e .[docs,tests]
 
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
 - ipython~=7.0
 - jinja2~=2.10
 - kiwipy[rmq]~=0.5.5
-- numpy~=1.19
+- numpy~=1.17
 - paramiko~=2.7
 - pika~=1.1
 - plumpy~=0.15.0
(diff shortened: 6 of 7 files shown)


csadorf (changes requested):
According to our current [compatibility timeline](http://www.aiida.net/download/) we must support numpy version 1.17 until 26-Jul-2021. The compatiblity timeline is derived from [AEP 003](https://github.com/aiidateam/AEP/tree/master/003_adopt_nep_29#detailed-explanation). 

Please not that there the 26-Jul-2021 is not explicitly mentioned in AEP 003, but required to be compatible with [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

Edit: This means the requirement should be specified as: `numpy~=1.17`.

.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ jobs:
             - cache-pip
 
       - run: |
-          pip install numpy==1.17.4
+          # pip install numpy==1.17.4
```
csadorf, line 17: I recommend to disable these commands in a separate commit (same PR but individual test run). We do not use this work-around in the installation tests (test-install).

csadorf (changes requested):
Please update the requirements/*.txt files according to the following table:
- 3.5: `numpy==1.18.5`
- 3.6: `numpy==1.19.2`
- 3.7: `numpy==1.19.2`
- 3.8: `numpy==1.19.2`

csadorf:
It appears that there is an actual incompatibility:
```
tests/test_dataclasses.py:3203: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
aiida/tools/data/array/kpoints/__init__.py:87: in get_explicit_kpoints_path
    return method(structure, **kwargs)
aiida/tools/data/array/kpoints/__init__.py:216: in _legacy_get_explicit_kpoints_path
    cell=structure.cell, pbc=structure.pbc, **kwargs
aiida/tools/data/array/kpoints/legacy.py:353: in get_explicit_kpoints_path
    numpy.linspace(ini_coord[0], end_coord[0], num_points[count_piece]),
<__array_function__ internals>:6: in linspace
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
```
My assumption is that the third argument in the call to `numpy.linspace` is a float instead of an integer here: https://github.com/aiidateam/aiida-core/blob/b002a9a9a898a05338ff522b46686657fb468f66/aiida/tools/data/array/kpoints/legacy.py#L353

CasperWA:
> It appears that there is an actual incompatibility:
> 
> ```
> tests/test_dataclasses.py:3203: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> aiida/tools/data/array/kpoints/__init__.py:87: in get_explicit_kpoints_path
>     return method(structure, **kwargs)
> aiida/tools/data/array/kpoints/__init__.py:216: in _legacy_get_explicit_kpoints_path
>     cell=structure.cell, pbc=structure.pbc, **kwargs
> aiida/tools/data/array/kpoints/legacy.py:353: in get_explicit_kpoints_path
>     numpy.linspace(ini_coord[0], end_coord[0], num_points[count_piece]),
> <__array_function__ internals>:6: in linspace
>     ???
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> ```
> 
> My assumption is that the third argument in the call to `numpy.linspace` is a float instead of an integer here:
> 
> https://github.com/aiidateam/aiida-core/blob/b002a9a9a898a05338ff522b46686657fb468f66/aiida/tools/data/array/kpoints/legacy.py#L353

Yes, already looked into this and just pushed a possible fix.

CasperWA:
> Please update the requirements/*.txt files according to the following table:
> 
> * 3.5: `numpy==1.18.5`
> * 3.6: `numpy==1.19.2`
> * 3.7: `numpy==1.19.2`
> * 3.8: `numpy==1.19.2`

Just to understand. Did you write this comment manually or is this part of the semi-automatic workflow?

.circleci/config.yml
CasperWA: Sure

csadorf:
> > Please update the requirements/*.txt files according to the following table:
> > 
> > * 3.5: `numpy==1.18.5`
> > * 3.6: `numpy==1.19.2`
> > * 3.7: `numpy==1.19.2`
> > * 3.8: `numpy==1.19.2`
> 
> Just to understand. Did you write this comment manually or is this part of the semi-automatic workflow?

I checked those manually. The automated workflow creates a PR if and only if the tests succeed and also only if the requirements files are outdated, which with the change to `numpy~=1.17` they actually are not. So we actually don't *have* to change them and now that I think about it, maybe we shouldn't.

.circleci/config.yml
CasperWA: Still not sure why this change is has to be in a separate commit. Perhaps for the end PR it should be in a separate PR; that's fine. But if it's important for now, then please clarify why.

csadorf: These commands represent a work-around that is addressing an issue that is independent of our general numpy dependency constraints. That said, it appears that all of the tests pass now even without this work-around so I'm fine with completely removing this work-around (separate commit or not).

aiida/tools/data/array/kpoints/legacy.py
```diff
@@ -342,6 +342,10 @@ def _num_points_from_coordinates(path, point_coordinates, kpoint_distance=None):
     explicit_kpoints = [tuple(point_coordinates[path[0][0]])]
     labels = [(0, path[0][0])]
 
+    if not all([_.is_integer() for _ in num_points if isinstance(_, (float, numpy.float64))]):
+        raise ValueError('Could not determine number of points as a whole number. num_points={}'.format(num_points))
```
csadorf, line 346: Can you just duplicate the related unit test, but provide the faulty input to cover this section?

csadorf (commented):
Looks good to me. Codecov is complaining about the coverage, I'm ok to just ignore that, although we might need to check with someone else whether we can override that check.

But before do that, maybe you can give it a 5min try of providing the faulty input to cover the section?

aiida/tools/data/array/kpoints/legacy.py
```diff
@@ -342,6 +342,10 @@ def _num_points_from_coordinates(path, point_coordinates, kpoint_distance=None):
...
+    num_points = [int(_) for _ in num_points]
```
sphuber, line 347: I don't see this variable `num_points` being defined in the old code, why do we need to define this now?

CasperWA: It is defined in the `if-else` block immediately above.

sphuber: Actually, I think this code will except in the exception, because it references `num_points` in the exception message but it is only defined afterwards. Would highly recommend actually adding a test...

CasperWA: The issue here is that the new `numpy` version apparently makes the contents of the list `numpy.float64` types, where it was previously integers.

CasperWA: > Actually, I think this code will except in the exception, because it references `num_points` in the exception message but it is only defined afterwards. Would highly recommend actually adding a test...

No? It's defined above. Otherwise the creation of the list for the `all()` check would fail ...

CasperWA: ...
So all this code is doing, is simply making sure that the contents of the previously created `num_points` are all integers as is expected from this code, apparently.

sphuber: I see, apologies, didn't think about it being defined above. Last question, is there a reason for the `if isinstance(_, (float, numpy.float64))`? Shouldn't they all be integers for it to be valid, i.e. should the following not be sufficient:
```
if any([not _.is_integer() for _ in num_points]):
    raise ValueError('Could not determine number of points as a whole number. num_points={}'.format(num_points))
```

CasperWA: > I see, apologies, didn't think about it being defined above.

No worries :)

> Last question, is there a reason for the `if isinstance(_, (float, numpy.float64))`? Shouldn't they all be integers for it to be valid, i.e. should the following not be sufficient:
> 
> ```
> if any([not _.is_integer() for _ in num_points]):
>     raise ValueError('Could not determine number of points as a whole number. num_points={}'.format(num_points))
> ```

I did that originally, but for some of the cases in the `if-else` block above, it correctly creates instances of `int`s, which don't have the `is_integer()` method ... so it failed in the CI and I added the `if`-condition.

In the end, there shouldn't be an issue for this check at all, we could just force cast the type to `int` and be done with it. But as I'm not completely sure what is being done in this code, I'd rather be overly careful than simply force type casting.

aiida/tools/data/array/kpoints/legacy.py
CasperWA: Looking into the code and how it works, I am unsure whether this case will ever arise. But I am also not 100 % sure it won't. I simply cannot determine the pathway through the code to reproduce a `num_points` that contains values that would not constitute integers in a float format.

I can either remove the check, or leave it in for "safety"?

In the end, the codecov patch issue is not relevant. Although, it did correctly find that this particular line is not checked, I guess.

CasperWA: I can also remove the block from codecov completely?

csadorf: > Looking into the code and how it works, I am unsure whether this case will ever arise. But I am also not 100 % sure it won't. I simply cannot determine the pathway through the code to reproduce a `num_points` that contains values that would not constitute integers in a float format.

There are just too many paths to determine that reliably. I mean clearly this has been an implicit assertion up until this point. So my suggestion is to make that assertion explicit:

`assert all([_.is_integer() for _ in num_points if isinstance(_, (float, numpy.float64))])`

The advantages of this handling is that the code segment would be covered by tests, it clearly indicates to future developers what assumptions are made here, and would make it very easy to diagnose the problem, should a user unexpectedly trigger this bug.

csadorf (changes requested):

aiida/tools/data/array/kpoints/legacy.py
CasperWA: Ah, you wanted to change the lines here to an `assert`. That makes more sense to me 👍 Can do. I misunderstood your original comment to mean adding a test in our test suite.

CasperWA:
@csadorf and @sphuber This is ready for a re-review + merge. Your old review comments have all been addressed.

sphuber (approved):
All good, thanks @CasperWA

PR #4375, opened 2020-09-16
🧪 DEV: Use -e install for tox runs

chrisjsewell:
Quick one: Improves startup time for test runs

PR #4379, opened 2020-09-17
FIX: Only remove temporary folder if it is present

chrisjsewell:
Surprised no one has had this issue before, but on OSX at least this `shutil.rmtree` fails if `retrieved_temporary_folder` is None:

```python
TypeError: lstat: path should be string, bytes or os.PathLike, not NoneType
```


sphuber:
This line
https://github.com/aiidateam/aiida-core/blob/c6bca066106c8ee92178923ea3a0b6ab0b3657e2/aiida/engine/processes/calcjobs/calcjob.py#L315
might then also bit us in the ass on OS at some time. For now it is never `None` because the temporary directory is always created and passed. Why is Mac OS so shit? ;)

chrisjsewell:
> Why is Mac OS so shit? ;)

You mean because it does the right thing 😉 bet this would fail a type check

chrisjsewell:
BTW, as mentioned in https://github.com/aiidateam/aiida-core/pull/4375#issuecomment-694497919, I was just trying to get all the tests to pass locally;

The tests/cmdline/commands/test_daemon.py::TestVerdiDaemon tests fail, is that expected, if you are just using the temporary profile creation?

chrisjsewell:
Also, after the passing tests finish, I get 100s of the same message:

```python
Exception ignored in: <generator object create_task.<locals>.run_task at 0x7f84deb94e50>
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/futures.py", line 102, in run_task
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/contextlib.py", line 158, in __exit__
AttributeError: 'NoneType' object has no attribute 'exc_info'
```

which doesn't seem healthy 😬

sphuber:
> Also, after the passing tests finish, I get 100s of the same message:

This has been fixed in the latest `plumpy`. Just install the latest patch version and those will go away.

sphuber:
> BTW, as mentioned in [#4375 (comment)](https://github.com/aiidateam/aiida-core/pull/4375#issuecomment-694497919), I was just trying to get all the tests to pass locally;
> 
> The tests/cmdline/commands/test_daemon.py::TestVerdiDaemon tests fail, is that expected, if you are just using the temporary profile creation?

Cannot say without having a closer look. Would it make sense to hold of with merging this and trying to tackle the rest as well and go in one go?

chrisjsewell:
> Would it make sense to hold of with merging this and trying to tackle the rest as well and go in one go?

Up to you

For the sphinx ones, not sure if this will solve it, but it would probably be better to use the "proper" sphinx test fixtures: https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/testing/fixtures.py

For the `TestVerdiDaemon` ones, it would be helpful to know if `pytest tests/cmdline/commands/test_daemon.py` and/or `tox -- tests/cmdline/commands/test_daemon.py` simply passes on Linux, or if you need to use a "pre-created" profile like in `.github/workflows/ci-code.yml:jobs:tests`

sphuber:
> For the `TestVerdiDaemon` ones, it would be helpful to know if `pytest tests/cmdline/commands/test_daemon.py` and/or `tox -- tests/cmdline/commands/test_daemon.py` simply passes on Linux, or if you need to use a "pre-created" profile like in `.github/workflows/ci-code.yml:jobs:tests`

Nope, with on the fly profile this also fails on Ubuntu. However with existing profile this works just fine.

chrisjsewell:
I'm merging this, since its a trivial thing

PR #4380, opened 2020-09-21
Do not run `process_handler` when `exit_codes=[]`.

greschd:
Currently, a `process_handler` that gets passed an empty `exit_codes` list will _always_ run. I think it would be more consistent to _never_ run these, and _always_ run only when `exit_codes=None`. 

That's certainly debatable, but since the change is simple enough I decided to make a PR to decide this over instead of making an issue first.

One could also consider emitting a warning when an empty list is passed (since the handler is then essentially dead code) - but if the list of exit codes is "dynamically determined" it could lead to false warnings.

**Context:** I tried removing `ERROR_COMPUTING_CHOLESKY` from [the following](https://github.com/aiidateam/aiida-quantumespresso/blob/031b92b9c85ed01588de332f039af05738607375/aiida_quantumespresso/workflows/pw/base.py#L367-L370), and was surprised that it then always ran: 
```python
    @process_handler(priority=590, exit_codes=[
        PwCalculation.exit_codes.ERROR_COMPUTING_CHOLESKY,
    ])
    def handle_known_unrecoverable_failure(self, calculation):
```

**Update:** Just noticed that the docstring for the `exit_codes` parameter supports changing the behavior (or otherwise, the docstring would need changing):
```
    :param exit_codes: single or list of `ExitCode` instances. If defined, the handler will return `None` if the exit
        code set on the `node` does not appear in the `exit_codes`. This is useful to have a handler called only when
        the process failed with a specific exit code.
```


sphuber (approved):
Certainly is debatable, but don't feel like it now. You have made the case sufficiently ;) Thanks @greschd

PR #4386, opened 2020-09-23
Drop support for Python 3.5

### LINKED ISSUE #4367: drop python 3.5 support
Following the [AiiDA compatibility timeline](http://www.aiida.net/download/), we will drop python 3.5 support from September 13th 2020 onwards, i.e. with the upcoming 1.4.0 release.

This means python3.6+ only features are now fair game and we can stop testing on python 3.5 on CI
@chrisjsewell @greschd @dev-zero @sphuber @csadorf

sphuber:
Fixes #4367 

Python 3.5 is EOL as of September 13 2020. CI testing will now only be
done against Python 3.6 and 3.8.


utils/dependency_management.py
```diff
@@ -30,7 +30,7 @@
     'graphviz': 'python-graphviz',
 }
 
-CONDA_IGNORE = ['pyblake2', r'.*python_version == \"3\.5\"']
+CONDA_IGNORE = ['pyblake2']
```
ltalirz, line 33: didn't you remove pyblake2 from the dependency list?

ltalirz (commented):
thanks @sphuber !

just one question

utils/dependency_management.py
sphuber: I guess, it is a backport from `hashlib.blake2b` which become part of stdlib from Python 3.6. So this would then also apply to conda, wouldn't it? Do you want me to also remove the `CONDA_IGNORE` variable and associated code, or we just leave it as an empty list.

ltalirz: > Do you want me to also remove the CONDA_IGNORE variable and associated code, or we just leave it as an empty list.

Up to you - it's difficult for me to guess how likely it is that we will need such exceptions again.

If you decide to leave it in for the moment, perhaps add a comment explaining what it does and setting an ultimatum in 6? months after which it can be removed

sphuber: @csadorf I think this is your code. What do you prefer? I leave it in for now?

csadorf: I'd prefer to leave it in.

ltalirz (approved):
thanks!

PR #4388, opened 2020-09-23
Fix to the exception message of the LinkManager

giovannipizzi:
The link manager was always referring to an 'input link' while
it should instead refer on an 'input link label' or 'output link label'
depending on the value of the link direction, determined by the `self._incoming`
attribute.


ramirezfranciscof (approved):
Pretty straightforward change, good to go! Can I merge it? What kind of merge do we use here?

sphuber:
> Pretty straightforward change, good to go! Can I merge it? What kind of merge do we use here?

Since this is a simple single-commit PR, we use a squash-merge. See the wiki for details on merge policy

PR #4393, opened 2020-09-24
verdi migrate --in-place between file systems

### LINKED ISSUE #4392: verdi import --in-place crashes when export file is on different file system
### Describe the bug

`verdi import --in-place` assumes that the temporary directory and the working directory are on the same file system.
If that is not the case, one can run into errors like these:
```
      File "/home/ubuntu/.virtualenvs/aiida/bin/verdi", line 12, in <module>
        sys.exit(verdi())
      File "/home/ubuntu/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 764, in __call__
      File "/home/ubuntu/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 717, in main
        rv = self.invoke(ctx)
      File "/home/ubuntu/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
      File "/home/ubuntu/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
      File "/home/ubuntu/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 956, in invoke
      File "/home/ubuntu/.virtualenvs/aiida/lib/python3.6/site-packages/click/core.py", line 555, in invoke
      File "/home/ubuntu/aiida-core/aiida/cmdline/commands/cmd_export.py", line 235, in migrate
        os.rename(output_file, input_file)
    OSError: [Errno 18] Invalid cross-device link: '/tmp/tmpkoqd1ucq/archive.aiida' -> '/home/ubuntu/aiida_profiles/home/ubuntu/.aiida/repository/sssp/archives/SSSP.1_5.aiida'
```

[`os.rename` should be replaced by `shutil.move`](https://stackoverflow.com/a/42541917/1069467)

### Steps to reproduce
 1. Place an AiiDA export file on a file system that is not the root volume (the one hosting temporary files)
 2. Try importing that file via `verdi import --in-place`

### Your environment

Current aiida-core `develop` branch.

### Additional context

One can use the `TMPDIR` environment variable to work around this issue by setting it e.g. to the current working directory.

ltalirz:
fix #4392 

verdi import --in-place assumed that the temporary directory for the
migrated archive and the working directory are on the same file system.
If that was not the case, it would crash with

  OSError: [Errno 18] Invalid cross-device link

Switching from os.rename to shutil.move should resolve this issue.


greschd:
Just to mention: While `os.rename` is guaranteed to be atomic (on Linux), AFAIK think `shutil.move` makes no such promises. I guess that's why it's able to move between file systems in the first place.

Not sure if that's a guarantee that we need here - I guess it would be nice to be protected against corruption of the in-place-migrated archive.

PR #4395, opened 2020-09-25
Adding entry point group `aiida.tools.data.orbitals` to the mapping

giovannipizzi:
The new entrypoint had been introduced, but it wasn't in the mapping,
so it couldn't be accessed for instance with `verdi plugin list`.


sphuber:
I noticed that there are more missing:

* `aiida.cmdline.computer.configure`
* `aiida.tools.calculations`

Should we add those as well?

ltalirz:
Just a minor nitpick: I believe you're adding a new entry point *group* here (just for the commit message).

ramirezfranciscof (approved):
Pretty straightforward to review. Thanks!

PR #4398, opened 2020-09-25
Dependencies: increase minimum version requirement `plumpy~=0.15.1`

### LINKED ISSUE #4346: Noisy but harmless exceptions reported when interpreter, that has run processes, ends
When a Python interpreter ends that has run processes, the stdout will be flooded with the following stack trace:
```
Exception ignored in: <generator object create_task.<locals>.run_task at 0x7f174d3ce6d0>
  File "/home/sph/code/aiida/env/dev/plumpy/plumpy/futures.py", line 102, in run_task
  File "/usr/lib/python3.7/contextlib.py", line 158, in __exit__
AttributeError: 'NoneType' object has no attribute 'exc_info'
```
This is due to the change of f9f6c9ca4ddd4f216c5a0e3e784976a78e8c5ca6 that now closes the event loop if the `Runner` shutsdown. Closing the loop allows these messages that have been collected to be printed. The messages themselves come from the future that add and remove task and broadcast subscribers, when a `Process` is initialized and destroyed, respectively. 

These exceptions do not not show up in Python 3.8, but do for lower version. Have confirmed it for Py 3.5 and 3.7. The exception seems to come from the following line in the Python built in module `contextlib`:
https://github.com/python/cpython/blob/4e02981de0952f54bf87967f8e10d169d6946b40/Lib/contextlib.py#L158

Somehow, it seems that at that point `sys` is already `None`.
We should try and find a way to prevent this exception from happening, or at the very least suppress the messages, because there wlll be two copies printed for each process run, and so they are really numerous.

sphuber:
Fixes #4346 

The patch release of `plumpy` comes with a simple fix that will prevent
the printing of many warnings when running processes. So although not
critical, it does improve user experience.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -22,7 +22,7 @@ dependencies:
 - numpy~=1.17
 - paramiko~=2.7
 - pika~=1.1
-- plumpy~=0.15.0
+- plumpy~=0.15.1
 - pgsu~=0.1.0
 - psutil~=5.6
 - psycopg2>=2.8.3,~=2.8
--- a/requirements/requirements-py-3.6.txt
+++ b/requirements/requirements-py-3.6.txt
@@ -78,7 +78,7 @@ pgtest==1.3.2
 pickleshare==0.7.5
 pika==1.1.0
 pluggy==0.13.1
-plumpy==0.15.0
+plumpy==0.15.1
 prometheus-client==0.7.1
 prompt-toolkit==3.0.4
 psutil==5.7.0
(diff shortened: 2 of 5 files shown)


ltalirz (approved):
thanks @sphuber !

sphuber:
@ltalirz apparently plumpy==0.15.1 wasn't built on conda yet. Is there a way to trigger the PR? Normally this is done automatically by the bot, isn't it?

PR #4399, opened 2020-09-25
Implement `next` and `iter` for the `Node.open` deprecation wrapper

### LINKED ISSUE #4396: New deprecation wrapper for `Node.open` causes problems with `numpy`
To prepare for the new repository interface in v2.0 the `Node.open` method wrapped the return value in the `WarnWhenNotEntered` class. However, this gives problems when used with `numpy` for example:
```
numpy.genfromtext(node.open('some_file'))
```
will raise a `TypeError`. The reason is because `numpy` will call `iter` on the file descriptor we pass, which is not implemented. We can easily implement this, but the question is what else have we missed? The idea of the wrapper was to prevent code from breaking when we go to v2.0, but is that really worth it if we break current code?

sphuber:
Fixes #4396 

The return value of `Node.open` was wrapped in `WarnWhenNotEntered` in
`aiida-core==1.4.0` in order to warn users that use the method without a
context manager, which will start to raise in v2.0. Unfortunately, the
raising came a little early as the wrapper does not implement the
`__iter__` and `__next__` methods, which can be called by clients.

An example is `numpy.genfromtxt` which will notice the return value of
`Node.open` is filelike and so will wrap it in `iter`. Without the
current fix, this raises a `TypeError`. The proper fix would be to
forward all magic methods to the wrapped filelike object, but it is not
clear how to do this.

### DIFF
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -67,6 +67,12 @@ class WarnWhenNotEntered:
     def __del__(self):
         self._warn_if_not_entered('del')
 
+    def __iter__(self):
+        return self._fileobj.__iter__()
+
+    def __next__(self):
+        return self._fileobj.__next__()
+
     def read(self, *args, **kwargs):
         self._warn_if_not_entered('read')
         return self._fileobj.read(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


greschd (approved):
Do we want a regression test for this?

sphuber:
> Do we want a regression test for this?

Done

sphuber:
@greschd I added a test, this is ready for second review

greschd (approved):
Thanks @sphuber !

PR #4400, opened 2020-09-25
Replace old format string interpolation with f-strings

### LINKED ISSUE #4394: Replace old format string interpolation with f-strings
Now that Python 3.5 support has been dropped we can start using f-strings, which are both more readable and more efficient compared to the older format string interpolation.

sphuber:
Fixes #4394 

Since Python 3.5 is no longer supported, format string interpolations
can now be replaced by f-strings, introduced in Python 3.6, which are
more readable, require less characters and are more efficient.

Note that `pylint` issues a warning when using f-strings for log
messages, just as it does for format interpolated strings. The reasoning
is that this is slightly inefficient as the strings are always
interpolated even if the log is discarded, but also by not passing the
formatting parameters as arguments, the available metadata is reduced.
I feel these inefficiencies are premature optimizations as they are
really minimal and don't weigh up against the improved readability and
maintainability of using f-strings. That is why the `pylint` config is
update to ignore the warning `logging-fstring-interpolation` which
replaces `logging-format-interpolation` that was ignored before.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,6 +8,15 @@ repos:
     -   id: mixed-line-ending
     -   id: trailing-whitespace
 
+-   repo: https://github.com/ikamensh/flynt/
+    rev: '0.55'
+    hooks:
+    -   id: flynt
+        args: [
+            '--line-length=120',
+            '--fail-on-change',
+        ]
+
 -   repo: https://github.com/pre-commit/mirrors-yapf
     rev: v0.30.0
     hooks:
--- a/.ci/polish/cli.py
+++ b/.ci/polish/cli.py
@@ -114,23 +114,23 @@ def launch(expression, code, use_calculations, use_calcfunctions, sleep, timeout
     valid, error = validate(expression)
 
     if not valid:
-        click.echo("the expression '{}' is invalid: {}".format(expression, error))
+        click.echo(f"the expression '{expression}' is invalid: {error}")
         sys.exit(1)
 
-    filename = 'polish_{}.py'.format(str(uuid.uuid4().hex))
+    filename = f'polish_{str(uuid.uuid4().hex)}.py'
     evaluated = evaluate(expression, modulo)
     outlines, stack = generate_outlines(expression)
     outlines_string = format_outlines(outlines, use_calculations, use_calcfunctions)
     write_workchain(outlines_string, filename=filename)
 
-    click.echo('Expression: {}'.format(expression))
+    click.echo(f'Expression: {expression}')
 
     if not dry_run:
         try:
-            workchain_module = 'polish_workchains.{}'.format(filename.replace('.py', ''))
+            workchain_module = f"polish_workchains.{filename.replace('.py', '')}"
             workchains = importlib.import_module(workchain_module)
         except ImportError:
-            click.echo('could not import the {} module'.format(workchain_module))
+            click.echo(f'could not import the {workchain_module} module')
             sys.exit(1)
 
         inputs = {'modulo': Int(modulo), 'operands': Str(' '.join(stack))}
@@ -153,8 +153,7 @@ def launch(expression, code, use_calculations, use_calcfunctions, sleep, timeout
             if timed_out:
                 click.secho('Failed: ', fg='red', bold=True, nl=False)
                 click.secho(
-                    'the workchain<{}> did not finish in time and the operation timed out'.format(workchain.pk),
-                    bold=True
+                    f'the workchain<{workchain.pk}> did not finish in time and the operation timed out', bold=True
                 )
                 sys.exit(1)
 
@@ -162,17 +161,17 @@ def launch(expression, code, use_calculations, use_calcfunctions, sleep, timeout
                 result = workchain.outputs.result
             except AttributeError:
                 click.secho('Failed: ', fg='red', bold=True, nl=False)
-                click.secho('the workchain<{}> did not return a result output node'.format(workchain.pk), bold=True)
+                click.secho(f'the workchain<{workchain.pk}> did not return a result output node', bold=True)
                 sys.exit(1)
 
         else:
             results, workchain = run_get_node(workchains.Polish00WorkChain, **inputs)
             result = results['result']
 
-    click.echo('Evaluated : {}'.format(evaluated))
+    click.echo(f'Evaluated : {evaluated}')
 
     if not dry_run:
-        click.echo('Workchain : {} <{}>'.format(result, workchain.pk))
+        click.echo(f'Workchain : {result} <{workchain.pk}>')
 
         if result != evaluated:
             click.secho('Failed: ', fg='red', bold=True, nl=False)
--- a/.ci/polish/lib/expression.py
+++ b/.ci/polish/lib/expression.py
@@ -73,7 +73,7 @@ def validate(expression):
     try:
         symbols = expression.split()
     except ValueError as exception:
-        return False, 'failed to split the expression into symbols: {}'.format(exception)
+        return False, f'failed to split the expression into symbols: {exception}'
 
     while len(symbols) > 1:
         try:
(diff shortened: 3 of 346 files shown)


sphuber:
Who dares review this PR :smile: @greschd @ltalirz ?

ltalirz:
First of all, thanks @sphuber  for taking on the effort to clean this up!

> Who dares review this PR 😄 @greschd @ltalirz ?

Hehe ;-) To me, the question is more *how* to best review this PR.
Could you perhaps describe how you did this - was this all done manually, did you use some search & replace operations, some refactoring tool, ...?

P.S. This is just out of curiosity: Do you know why test coverage decreased (very mildly), despite f-strings being less verbose?

sphuber:
> Hehe ;-) To me, the question is more _how_ to best review this PR.
> Could you perhaps describe how you did this - was this all done manually, did you use some search & replace operations, some refactoring tool, ...?

Sure, I sure as hell didn't do this manually but used the [`flynt` tool](https://pypi.org/project/flynt/). I first ran with default settings, meaning it will perform the safest transformation. I checked the changes manually and did some minor corrections and pushed to have the tests run. When all tests passed, I ran the tool again, but this time with the options `--transform-concats --agressive -ll 120`. Here also I checked the changes manually and did some minor corrections. Then I manually searched with regex `['"]\.format\(` and fixed some cases manually. There are still quite a few remaining but didn't want to start addressing them all manually so I left them for now.

> I wonder whether going through all of this line by line is the best way to spot potential issues...

Probably not. These kinds of changes are always a bit risky since we are relying purely on the tests but many of these lines may not actually be covered. `pylint` will give us some extra security but probably not a 100%. To be honest I am also not sure how much this can and should be reviewed.

> P.S. This is just out of curiosity: Do you know why test coverage decreased (very mildly), despite f-strings being less verbose?

I have no idea, but honestly this happens all the time and we always just ignore it.

chrisjsewell:
> This is just out of curiosity: Do you know why test coverage decreased, despite f-strings being less verbose?

Simple, because there're fewer lines covered 😄, e.g.

```diff
- "AttributeError: '{}' is not a valid attribute of the object "
- "'{}'".format(attr, self.__class__.__name__)
+ f"AttributeError: '{attr}' is not a valid attribute of the object '{self.__class__.__name__}'"
```

So now `<lines covered>/<total lines>` is a smaller fraction

> we always just ignore it.

You might not have to after https://github.com/aiidateam/aiida-core/pull/4413 😉

chrisjsewell:
For what its worth, I took a quick parse over the diff, and didn't notice any errors (my vision did start to blur a bit after the first hundred diffs though lol)

Note, flynt has a pre-commit hook: https://github.com/ikamensh/flynt#pre-commit-hook, should we consider adding it?

sphuber:
> Note, flynt has a pre-commit hook: https://github.com/ikamensh/flynt#pre-commit-hook, should we consider adding it?

I thought about this yeah, but decided not to. But to be honest, I am not sure anymore why I thought it best not to add it. I have now added it. Let's see how that goes

greschd:
TBH I don't see why _always_ using f-strings should be enforced. Something like the following still makes sense in my book:

```python
some_dict = {'a': 1, 'b': 2}
<...>
string = 'values: a={a}, b={b}'.format(**some_dict)
```

chrisjsewell:
>  Something like the following still makes sense in my book

Yeh but I don't think flynt would convert that?
anyway looks to be a weird problem with its CLI interaction with pre-commit?

sphuber:
> TBH I don't see why _always_ using f-strings should be enforced. Something like the following still makes sense in my book:
> 
> ```python
> some_dict = {'a': 1, 'b': 2}
> <...>
> string = 'values: a={a}, b={b}'.format(**some_dict)
> ```

Flynt won't touch these though. There are indeed some cases where it is better to use implicit conversion with `format`

greschd:
Alright, in that case it's fine. Haven't tried `flynt` myself, so I'm commenting from a place of ignorance 🙃

ltalirz:
> > This is just out of curiosity: Do you know why test coverage decreased, despite f-strings being less verbose?
> 
> Simple, because there're fewer lines covered 😄,

Right, my misunderstanding came from the assumption that what mattered was the number of uncovered lines, but we are actually triggering on the percentage of uncovered lines.
I guess it's a little more complicated than you suggest as well, since there will be less lines both in the uncovered and in the covered code. Assuming that coverage is counted correctly, the question becomes: is the *density* of lines saved higher in covered or uncovered code (a priori, I would probably have bet on the latter, since it's often error messages in `Except` blocks that go uncovered).

Naively, when making a PR what I actually care about is whether/how much I'm increasing the total number of uncovered lines.
The problem is that this will be slightly nonzero in almost all cases, and so it is difficult to decide when to raise a warning flag.
At the moment, we're essentially saying: for new code, cover it at least as well as the average codebase (which is a moving target as our coverage has been slowly increasing over time).

greschd:
Ideally (IMO), the coverage criteria would be:
- for any new _logic_, a fixed coverage (e.g. 95%) needs to be reached
- for slight modifications to existing code (such as here), coverage should remain the same, with some tolerance

Of course the problem is that there's no way of distinguishing between the two cases, so we just fall back on the second (less strict) requirement. But it's a red flag if the number of _missed_ lines increases by a substantial number.

ltalirz:
> But it's a red flag if the number of missed lines increases by a substantial number.

Fully agree, and it would be nice to be able to track this (the [codecov.yml reference](https://docs.codecov.io/docs/codecovyml-reference) does not seem to offer this option (?)).

The first priority, however, is probably still to get the random fluctuations under control.
To stay with the the codecov report above, it actually report an increase in the *total* number of missed lines as well:
image

You may have noticed that, some time ago, codecov started adding comments to missed lines in the "Files" tab, which is helpful. In this case, it make clear that the report of these supposedly "uncovered lines" is really a bug, see e.g.
https://github.com/aiidateam/aiida-core/pull/4400/files#diff-31db4ebe16273a17841c21ee58d0d9baL184

.pre-commit-config.yaml
```diff
@@ -8,6 +8,17 @@ repos:
...
     -   id: trailing-whitespace
 
+-   repo: https://github.com/ikamensh/flynt/
+    rev: '0.55'
+    hooks:
+    -   id: flynt
+        pass_filenames: false
+        args: [
+            '--line-length 120',
+            '--transform-concats',
```
greschd, line 18: I think this argument should be removed: it's available only on py3.8+ interpreters.

.pre-commit-config.yaml
```diff
@@ -8,6 +8,17 @@ repos:
     -   id: mixed-line-ending
     -   id: trailing-whitespace
 
+-   repo: https://github.com/ikamensh/flynt/
+    rev: '0.55'
+    hooks:
+    -   id: flynt
+        pass_filenames: false
+        args: [
+            '--line-length 120',
```
greschd, line 17: @sphuber I think the reason it was behaving weirdly is that this should be `'--line-length', '120'`. At least locally for me, this fixes the issue.

greschd: And we can remove the `pass_filenames: false` again - otherwise it'll either do nothing, or scan the entire code base.

.pre-commit-config.yaml
sphuber: But we are running our pre-commit on 3.8. I think it is an important option to have. It will just not work for local dev environments with an outdated python version. Will it crash or simply ignore that option? If the latter, I vote to keep it in, even though now pre-commit may succeed locally but then fail on GHA..... arrgh

.pre-commit-config.yaml
sphuber: Ah yeah, I think either that or `'--line-length=12'` might also work

greschd: Yep.. pretty unintuitive error message, I must say.

.pre-commit-config.yaml
greschd: It crashes..

greschd: But I think it's not actually _that_ important to have it in - we can transform the bulk of them by manual invocation of `flynt`.

sphuber: Fine, I cave ;) has been removed

greschd: We could add `language_version: 'python3.8'` to the pre-commit config - that just means that `python3.8` is a hard requirement (to be installed on the machine, not necessarily that venv) for running the pre-commit hook.

sphuber: But then it won't run at all for devs without 3.8. Let's just omit it for now.

greschd (approved):

PR #4405, opened 2020-09-26
`verdi setup`: forward broker defaults to interactive mode

### LINKED ISSUE #4404: Broker options do not suggest defaults in interactive mode of `verdi setup`
For most setups, the broker options should just use the defaults, which are defined in the code, but are not automatically suggested in interactive mode of `verdi setup`.

sphuber:
Fixes #4404 

The options for the message broker configuration do define defaults,
however, the interactive clones for `verdi setup`, which are defined in
`aiida.cmdline.params.options.commands.setup` override the default with
the `contextual_default` which sets an empty default, unless it is taken
from an existing profile. The result is that for new profiles, the
broker options do not specify a default, even though for most use cases
the defaults will be required.

### DIFF
--- a/aiida/cmdline/params/options/commands/setup.py
+++ b/aiida/cmdline/params/options/commands/setup.py
@@ -18,6 +18,7 @@ from aiida.backends import BACKEND_DJANGO
 from aiida.cmdline.params import options, types
 from aiida.manage.configuration import get_config, get_config_option, Profile
 from aiida.manage.external.postgres import DEFAULT_DBINFO
+from aiida.manage.external.rmq import BROKER_DEFAULTS
 
 PASSWORD_UNCHANGED = '***'  # noqa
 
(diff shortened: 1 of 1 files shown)


ltalirz (approved):
Thanks @sphuber , I guess this is in response to the recent post to the AiiDA mailing list?

The changes look fine to me and I'm approving this, but you might still consider adding a test to check that new profiles have the right settings (that would have caught the issue we saw on the mailing list).

Also, it took me a bit to follow the slightly convoluted sentence structure in the commit message - here an alternative suggestion:
```
The `verdi setup` cli was overriding default values for the message broker
configuration with `None`, leading to a broken profile configuration.
This did not affect `verdi quicksetup`, where proper defaults were used.
```

sphuber:
> Thanks @sphuber , I guess this is in response to the recent post to the AiiDA mailing list?

Not really. I haven't seen any post with respect to this? Which do you mean?

> The changes look fine to me and I'm approving this, but you might still consider adding a test to check that new profiles have the right settings (i.e. a test that would catch the issue we saw on the mailing list).

It is not actually a problem with the created profiles, they are fine. It is just that when running `verdi setup` in interactive mode, the prompt won't provide a default for the broker settings, so the user will have to type it manually, even though the non-interactive version uses the correct default. So everything works fine, it is just a question of user-friendliness. After this change, 99/100 users can simply press enter for all broker options as the default will be just fine for their use case.

> Also, I found it not so easy to follow the slightly convoluted sentence structure in the commit message - here an alternative suggestion:
> 
> ```
> The `verdi setup` cli was overriding default values for the message broker
> configuration with `None`, leading to a broken profile configuration.
> This did not affect `verdi quicksetup`, where proper defaults were used.
> ```

Maybe this has to do with the previous point, but this message is incorrect. There was no risk of broken profiles, just the prompt did not provide sensible defaults.

sphuber:
I have added an extra sentence to the commit message to make clarify that this just concerns the prompt displaying a proper default.

ltalirz:
The mailing list: https://groups.google.com/g/aiidausers/c/mxW8NfRwVJM

sphuber:
> The mailing list: https://groups.google.com/g/aiidausers/c/mxW8NfRwVJM

Thanks. I don't think this is related. This exception is known to happen before the new broker configuration of v1.4. It is being discussed in [this issue](https://github.com/aiidateam/aiida-core/issues/3973). So far we haven't been able to fully pin point the problem nor find a solution. I will respond to the mailing list with a link to the issue and ask if he can add his example.

PR #4407, opened 2020-09-26
`verdi process show`: order called by ctime and use process label

### LINKED ISSUE #4406: Use process type and sort by creation time for caller and called links in `verdi process show`
Currently the node type is shown, which is often not very informative and the sorting is random. Showing the process type and sorting by creation time shows a more natural overview of, especially, what sub processes were called in what order.
Example, currently:
```
Called       PK  Type            
--------  -----  ----------------
CALL      19012  WorkChainNode    
CALL      19003  WorkChainNode   
CALL      18999  CalcFunctionNode
CALL      18988  WorkChainNode   
CALL      18979  WorkChainNode   
CALL      18969  WorkChainNode    

```
but which could become
```
Called       PK  Type
--------  -----  -----------------------
CALL      18969  PwBaseWorkChain
CALL      18979  PwBaseWorkChain
CALL      18988  HpWorkChain
CALL      18999  structure_relabel_kinds
CALL      19003  PwBaseWorkChain
CALL      19012  HpWorkChain
```

sphuber:
Fixes #4406 

The command was showing the called subprocesses in a random order and
used the node type, which is often uninformative. For example, all
workchains are always shown as `WorkChainNode`. By using the process
label instead, which is more specific, and ordering the called nodes by
creation time, the list gives a more natural overview of the order in
which the subprocesses were called.

### DIFF
--- a/aiida/cmdline/utils/common.py
+++ b/aiida/cmdline/utils/common.py
@@ -147,17 +147,21 @@ def get_node_info(node, include_summary=True):
     nodes_input = node.get_incoming(link_type=(LinkType.INPUT_CALC, LinkType.INPUT_WORK))
     nodes_output = node.get_outgoing(link_type=(LinkType.CREATE, LinkType.RETURN))
 
-    if nodes_caller:
-        result += '\n' + format_nested_links(nodes_caller.nested(), headers=['Called by', 'PK', 'Type'])
-
     if nodes_input:
         result += '\n' + format_nested_links(nodes_input.nested(), headers=['Inputs', 'PK', 'Type'])
 
     if nodes_output:
         result += '\n' + format_nested_links(nodes_output.nested(), headers=['Outputs', 'PK', 'Type'])
 
+    if nodes_caller:
+        result += '\n' + format_flat_links(
+            sorted(nodes_caller.all(), key=lambda x: x.node.ctime), headers=['Caller', 'PK', 'Type']
+        )
+
     if nodes_called:
-        result += '\n' + format_flat_links(nodes_called.all(), headers=['Called', 'PK', 'Type'])
+        result += '\n' + format_flat_links(
+            sorted(nodes_called.all(), key=lambda x: x.node.ctime), headers=['Called', 'PK', 'Type']
+        )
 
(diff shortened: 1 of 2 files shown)


ramirezfranciscof (commented):
This looks great, and it will be much better to see the labels rather than the class name! I only wanted to discuss 1 comment before I approve.

aiida/cmdline/utils/common.py
```diff
@@ -147,17 +147,21 @@ def get_node_info(node, include_summary=True):
...
     if nodes_input:
         result += '\n' + format_nested_links(nodes_input.nested(), headers=['Inputs', 'PK', 'Type'])
 
     if nodes_output:
         result += '\n' + format_nested_links(nodes_output.nested(), headers=['Outputs', 'PK', 'Type'])
 
+    if nodes_caller:
+        result += '\n' + format_flat_links(
+            sorted(nodes_caller.all(), key=lambda x: x.node.ctime), headers=['Caller', 'PK', 'Type']
+        )
```
ramirezfranciscof, line 159: (oops, I think I forgot to add this to the review, I'll just leave it here)

This "if-path" seems not to be covered by the tests; I think the easiest way to include it would be by either adding a cli-invocation on one of the calcjob nodes + respective checks, or by adding a call link between the workchains and just adding a check after the already existing cli-invoke (I woult think the second one is the most "economical one", but I'm not sure if you had a reason for creating those two workchains and leaving one of those disconected).

Also, tangential question: a process will either have one caller or none at all, right? Why do we write this here like there could be many callers?

sphuber: > I woult think the second one is the most "economical one", but I'm not sure if you had a reason for creating those two workchains and leaving one of those disconected

No particular reason to leave it unconnected just that I only needed one of those two to do the test. I will add the second one as the caller of the first one and add the test.

> Also, tangential question: a process will either have one caller or none at all, right? Why do we write this here like there could be many callers?

Absolutely correct. There is no specific reason other than that it is just easier to write and understand compared to the `nodes_called`.

ramirezfranciscof (approved):
Seems ready to go now, thanks!

PR #4408, opened 2020-09-27
`verdi setup`: improve validation and help string of broker virtual host

### LINKED ISSUE #4403: Forward slashes are not allowed in broker virtual host in profile setup
The new broker configuration parameters that have been added in `v1.4.0` have an optional broker virtual host field. Currently, names containing a forward slash fail validation, because the validator is based on a hostname validator, which cannot contain forward slashes. However, virtual hosts typically start with a forward slash, so users may be tempted to pass it but will be presented with a validation error. The help string message is misleading because it says that forward slashes need to be encoded.

The code that forms the URL based on the broker configuration of the profile will take care of adding a leading forward slash in case of a virtual host and it will encode the entire path. So at the very least, we should adapt the help string to say that the leading slash is not necessary and remove the comment about encoding. Ideally, any leading slashes are stripped before passing it to the validation.

sphuber:
Fixes #4403 

The help string of the `--broker-virtual-host` option of `verdi setup`
incorrectly said that forward slashes have to be escaped but this is not
true. The code will escape any characters necessary when constructing
the URL to connect to RabbitMQ. On top of that, slashes would fail the
validation outright, even though these are common in virtual hosts. For
example the virtual host always starts with a leading forward slash, but
our validation would reject it. Also the leading slash will be added by
the code and so does not have to be used in the setup phase. The help
string and the documentation now reflect this.

The exact naming rules for virtual hosts, imposed by RabbitMQ or other
implementations of the AMQP protocol, are not fully clear. But instead
of putting an explicit validation on AiiDA's side and running the risk
that we incorrectly reject valid virtual host names, we simply accept
all strings. In any case, any non-default virtual host will have to be
created through RabbitMQ's control interface, which will perform the
validation itself.

### DIFF
--- a/aiida/cmdline/params/options/__init__.py
+++ b/aiida/cmdline/params/options/__init__.py
@@ -315,10 +315,10 @@ BROKER_PORT = OverridableOption(
 
 BROKER_VIRTUAL_HOST = OverridableOption(
     '--broker-virtual-host',
-    type=types.HostnameType(),
+    type=click.types.StringParamType(),
     default=BROKER_DEFAULTS.virtual_host,
     show_default=True,
-    help='Name of the virtual host for the message broker. Forward slashes need to be encoded'
+    help='Name of the virtual host for the message broker without leading forward slash.'
 )
 
 REPOSITORY_PATH = OverridableOption(
--- a/docs/source/intro/installation.rst
+++ b/docs/source/intro/installation.rst
@@ -687,7 +687,7 @@ The following parameters can be configured:
 +--------------+---------------------------+---------------+-------------------------------------------------------------------------------------------------------------------------+
 | Port         | ``--broker-port``         | ``5672``      | The port to which the server listens.                                                                                   |
 +--------------+---------------------------+---------------+-------------------------------------------------------------------------------------------------------------------------+
-| Virtual host | ``--broker-virtual-host`` | ``''``        | Optional virtual host. If defined, needs to start with a forward slash.                                                 |
+| Virtual host | ``--broker-virtual-host`` | ``''``        | Optional virtual host. Should not contain the leading forward slash, this will be added automatically by AiiDA.         |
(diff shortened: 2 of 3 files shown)


aiida/cmdline/params/options/__init__.py
```diff
@@ -315,10 +315,10 @@ def decorator(command):
 
 BROKER_VIRTUAL_HOST = OverridableOption(
     '--broker-virtual-host',
-    type=types.HostnameType(),
+    type=click.types.StringParamType(),
```
ltalirz, line 318: Just to understand: Didn't the HOSTNAME_REGEX already exclude slashes?
https://github.com/sphuber/aiida-core/blob/dbbea0e3582894f8ef88e0f534db7560a1d8994d/aiida/cmdline/params/types/strings.py#L62
Why not keep the HostnameType?

sphuber: > Just to understand: Didn't the HOSTNAME_REGEX already exclude leading slashes?

Yes, which is exactly the problem. A virtual host with one or more slashes is perfectly valid. See also my commit message:

> On top of that, slashes would fail the validation outright, even though these are common in virtual hosts.

In the final paragraph of the commit message I explain why we resort to a type with as little restrictions as possible, namely a simple string type. I checked the RabbitMQ docs for virtual host naming rules but couldn't find it, so I asked on the mailing list and experimented: https://groups.google.com/u/1/g/rabbitmq-users/c/ZT56iwGbxy0   The last response is what I found by experimenting, almost anything goes. So I think it is best not to restrict anything in our interface.

ltalirz: > Yes, which is exactly the problem. A virtual host with one or more slashes is perfectly valid. 

Well, I was going by the help string, in which you ask people *not* to add the leading slash.

But I see now that these virtual hosts can contain almost any string anywhere, thanks for the info!

ltalirz (approved):

PR #4413, opened 2020-09-30
🔧 Add patch threshold for codecov

chrisjsewell:
The project diff percentage is the change in coverage w.r.t. all lines in the project, whereas the patch diff percentage is the change in coverage w.r.t. only lines touched by the PR
The patch threshold is currently defaulting to 0%, hence it is very easy to fail.

You might even want to increase this threshold, since a smaller amount of uncovered lines will lead to a larger diff percentage, relative to the project diff.


sphuber (commented):
Great if this can get rid of false positives of failing builds due to so-called reduced coverage. However, I definitely think we need to up the threshold. The 0.02% is based on all the lines in the project which is huge. Very few patches will come close. I am thinking 0.1% would already be a lot more reasonable don't you think? With 0.02% we would still see the false positives very often I think

codecov.yml
```diff
@@ -12,3 +12,6 @@ coverage:
     project:
       default:
         threshold: 0.02%
+    patch:
+      default:
+        threshold: 0.02%
```
chrisjsewell, line 17:
```suggestion
        threshold: 0.1%
```

chrisjsewell:
absolutely, just changed

sphuber (approved):
Good to go thanks @chrisjsewell

PR #4415, opened 2020-09-30
`CalcJob`: make sure `local_copy_list` files do not end up in node repo

### LINKED ISSUE #4414: Files in `local_copy_list` permanently stored in the repository
### Describe the bug

When submitting a calculation with a local copy list, files are also stored in the repository permanently.

### Expected behavior

The goal of the local copy list is exactly to avoid that these files are stored, since (at least in current versions of AiiDA, might change in the future with the new disk-objectstore) they occupy space, and e.g. for pseudopotentials in a DFT calculation there is no need to repeat these files for each node.

### Steps to reproduce

I'm using AiiDA 1.4.0.

1. submit and AiiDA-quantumespresso calculation
2. Check the repository content, and you will find the pseudos in the `pseudos` subfolder.

### Additional context

I think this was introduced when implementing the `provenance_exclude_list`.

The `local_copy_list` now leverages it, and the files in the local_copy_list are added here:
https://github.com/aiidateam/aiida-core/blob/4544bc49a50c9aa3abebd2837efb5626958ee2b4/aiida/engine/daemon/execmanager.py#L201

And then here they are not added to the final repository:
https://github.com/aiidateam/aiida-core/blob/4544bc49a50c9aa3abebd2837efb5626958ee2b4/aiida/engine/daemon/execmanager.py#L277-L283

However, this seems not to be working.
I think the reason is that the files in the `local_copy_list` are specified by aiida-quantumespresso as `./pseudo/Co.upf` and then the function above checks instead `pseudo/Co.upf`.
However, I think what aiida-quantumespresso is doing is not invalid, and the execmanager of AiiDA should take care of properly normalising also the relative path provided in the local_copy_list.

sphuber:
Fixes #4414 

The concept of the `local_copy_list` is to provide a possibility to
`CalcJob` plugins to write files to the remote working directory but
that are not also copied to the calculation job's repository folder.
However, due to commit 9dfad2efbe9603957a54d0123a3cec2ee48b54bd this
guarantee is broken.

The relevant commit refactored the handling of the `local_copy_list` in
the `upload_calculation` method to allow the target filepaths in the
list to contain nested paths with subdirectories that might not yet
necessarily exist. The approach was to first write all files to the
sandbox folder, where it is easier to deal with non-existing
directories. To make sure that these files weren't then also copied to
the node's repository folder, the copied files were also added to the
`provenance_exclude_list`. However, the logic in that part of the code
did not normalize filepaths, which caused files to be copied that
shouldm't have. The reason is that the `provenance_exclude_list` could
contain `./path/file_a.txt`, which would be compared to the relative
path `path/file_a.txt` which references the same file, but the strings
are not equal.

The solution is to ensure that all paths are fully normalized before
they are compared. This will turn the relative path `./path/file_a.txt`
into `path/file_a.txt`.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -273,11 +273,15 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
     # folders that would be empty when considering the `provenance_exclude_list` will *not* be copied to the repo. The
     # advantage of this explicit copying instead of deleting the files from `provenance_exclude_list` from the sandbox
     # first before moving the entire remaining content to the node's repository, is that in this way we are guaranteed
-    # not to accidentally move files to the repository that should not go there at all cost.
+    # not to accidentally move files to the repository that should not go there at all cost. Note that all entries in
+    # the provenance exclude list are normalized first, just as the paths that are in the sandbox folder, otherwise the
+    # direct equality test may fail, e.g.: './path/file.txt' != 'path/file.txt' even though they reference the same file
+    provenance_exclude_list = [os.path.normpath(entry) for entry in provenance_exclude_list]
+
     for root, _, filenames in os.walk(folder.abspath):
         for filename in filenames:
             filepath = os.path.join(root, filename)
-            relpath = os.path.relpath(filepath, folder.abspath)
+            relpath = os.path.normpath(os.path.relpath(filepath, folder.abspath))
             if relpath not in provenance_exclude_list:
                 with open(filepath, 'rb') as handle:
                     node._repository.put_object_from_filelike(handle, relpath, 'wb', force=True)  # pylint: disable=protected-access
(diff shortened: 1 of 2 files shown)


giovannipizzi (approved):
Thanks for the quick fix!

giovannipizzi:
I guess we need to remove the required flag from the 3.5 tests?

sphuber:
> I guess we need to remove the required flag from the 3.5 tests?

Not really. This is going to be merged in `release/1.4.2` which still supports Python 3.5. That is why I am merging this straight into the release branch and not into `develop`. Not sure why the tests didn't run though. The `release/1.4.2` branch and this PR branch still define `3.5` in the workflow yml.

espenfl:
Thanks for issuing a fix for this asap. Greatly appreciated. What releases have this issue? We need to block them in the VASP plugin.

sphuber:
@espenfl  v1.4 and v1.4.1 are affected by this bug. I will release a patch release a.s.a.p. Just waiting for one other potential bug that needs to be fixed.

espenfl:
@sphuber Thanks. I will block those. Also, I will test to see if 1.4 triggers our tests for this.

PR #4416, opened 2020-09-30
Add path creation when copying remote files

ramirezfranciscof:
If the `remote_copy_list` or the `symlink_copy_list` were provided with
a target path that required the creation of intermediate folders, the
CalcJob would have fail. This should fix that problem.

* Also added tests for the `upload_calculation` method of the
execmanager (in general and for this specific case)

* Replaced `execlogger` by `EXEC_LOGGER` to comply with pylint naming
conventions for global variables.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -159,6 +159,13 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
     remote_symlink_list = calc_info.remote_symlink_list or []
     provenance_exclude_list = calc_info.provenance_exclude_list or []
 
+    # Creates the directory structure locally and first to prevent unnecessary transport
+    # calls when copying the individual files
+    for _, _, target_relpath in local_copy_list + remote_copy_list + remote_symlink_list:
+        dirname = os.path.dirname(target_relpath)
+        if dirname:
+            os.makedirs(os.path.join(folder.abspath, dirname), exist_ok=True)
+
     for uuid, filename, target in local_copy_list:
         logger.debug(f'[submission of calculation {node.uuid}] copying local file/folder to {target}')
 
@@ -190,9 +197,6 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
         if data_node is None:
             logger.warning(f'failed to load Node<{uuid}> specified in the `local_copy_list`')
         else:
-            dirname = os.path.dirname(target)
-            if dirname:
-                os.makedirs(os.path.join(folder.abspath, dirname), exist_ok=True)
             with folder.open(target, 'wb') as handle:
                 with data_node.open(filename, 'rb') as source:
                     shutil.copyfileobj(source, handle)
@@ -212,10 +216,6 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
                     )
                 )
                 try:
-                    # We have to create the directory first: if it doesn't exist the copy call will fail
-                    dirname = os.path.dirname(dest_rel_path)
-                    if dirname:
-                        transport.makedirs(dirname, ignore_existing=True)
                     transport.copy(remote_abs_path, dest_rel_path)
                 except (IOError, OSError):
                     logger.warning(
@@ -237,10 +237,6 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
                     )
                 )
                 try:
-                    # We have to create the directory first: if it doesn't exist the copy call will fail
-                    dirname = os.path.dirname(dest_rel_path)
-                    if dirname:
-                        transport.makedirs(dirname, ignore_existing=True)
                     transport.symlink(remote_abs_path, dest_rel_path)
                 except (IOError, OSError):
                     logger.warning(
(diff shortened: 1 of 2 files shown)


ramirezfranciscof:
So, codecov is complaining about coverage in the lines I changed regarding the pylint `execlogger` complaint because there are no tests for those logging cases. I think I can get a pass on these, no?

sphuber:
Thanks @ramirezfranciscof I will give this a look later. If you want this to be included in `1.4.2` you should rebase this onto `release/1.4.2`. If it can go in `1.5.0` fine to merge in `develop`.

sphuber:
> So, codecov is complaining about coverage in the lines I changed regarding the pylint `execlogger` complaint because there are no tests for those logging cases. I think I can get a pass on these, no?

Yes I guess so, but on the other hand, is it necessary to make those naming changes to the logger? Without those, you won't have the coverage problem.The old name was there before and therefore accepted by `pylint` right? I also updated this file in my recently opened PR and `pylint` then also didn't complain

ramirezfranciscof:
> Thanks @ramirezfranciscof I will give this a look later. If you want this to be included in 1.4.2 you should rebase this onto release/1.4.2. If it can go in 1.5.0 fine to merge in develop.

Mmm, I am kind of indiferent to this I think. This could be either be seen as a compliment to the `local_copy_list` modification you did or as a separate feature that implements a similar functionality to a different list. GIven that this I mostly need this for the transfer stuff I guess it can wait until 1.5 (also there is more time to check in case we need to fix some similar bug to the one in the `local_copy_list`).

> Yes I guess so, but on the other hand, is it necessary to make those naming changes to the logger? Without those, you won't have the coverage problem.The old name was there before and therefore accepted by pylint right? I also updated this file in my recently opened PR and pylint then also didn't complain

Ah, yes, in my case pylint didn't allow me to commit without changing the logger. It actually did when I did the first fix attempt to the `local_copy_list` and was a bit surprised that you didn't change it in your PR.

sphuber (changes requested):
Thanks @ramirezfranciscof just some minor changes.

sphuber:
@giovannipizzi would be good if you could give this a look. I am not sure if there is a better more efficient way of doing this. We are now making potentially two transport calls per file copy/symlink, but not sure there is a way around it. Besides keeping a reference list of what subdirectories have already been made and not creating the directories more than once, but that opens the code up to more potential bugs and am not sure if the increased efficiency would be worth it. Might be rare to be copying many files in the same directory

ramirezfranciscof:
> We are now making potentially two transport calls per file copy/symlink, but not sure there is a way around it.

I mean, maybe one possible option if we are really worried about this is to have a first pass through the `remote_copy_list` and `symlink_copy_list` that just gets the dirname (`dirname = os.path.dirname(dest_rel_path)`) and creates it in the sandbox folder, so all the structure is ready once this is copied to the cluster.

sphuber:
> I mean, maybe one possible option if we are really worried about this is to have a first pass through the `remote_copy_list` and `symlink_copy_list` that just gets the dirname (`dirname = os.path.dirname(dest_rel_path)`) and creates it in the sandbox folder, so all the "empty" folder structure is ready once this is copied to the cluster.

Good point. Would be good to investigate this to see if there are potential problems with this approach, but on first thought it seems feasible

ramirezfranciscof:
I made the corrections in a first commit and then applied the change in the structure in a separate one (so it would be easier to revert if necessary). Check it out and let me know if its ok @sphuber @giovannipizzi .

aiida/engine/daemon/execmanager.py
```diff
@@ -159,6 +159,13 @@ def upload_calculation(node, transport, calc_info, folder, inputs=None, dry_run=
     remote_symlink_list = calc_info.remote_symlink_list or []
     provenance_exclude_list = calc_info.provenance_exclude_list or []
 
+    # Creates the directory structure locally and first to prevent unnecessary transport
```
giovannipizzi, line 162: Maybe for this comment, more than explaining that this is a solution to a problem that never was in the code (which anyway you could keep, but maybe saying 'if we were to create folders only on the remote, for each file), you could add an explanation of the original problem you wanted to solve (if the copy_list include only the filename of a file in a subfolder, and the subfolder does not exist, they would fail - so we first create possibly empty folders where appropriate.

giovannipizzi (commented):
Apart for a small suggestion for a change in the comment, I think this looks fine to be merged, thanks a lot! I don't seem to find any potential issue

aiida/engine/daemon/execmanager.py
ramirezfranciscof: Changed it for the following text; hopefully this is clearer now, if you don't object I'll merge when the tests pass.

```
    # First creates the directory structure locally before copying the sandbox folder, so that all the intermediate
    # folders for the files in the copy_lists are there before calling the copy methods of the transport (or else
    # these will fail).
    # Alternatively, one would have to call the path creation methods of the transports just before calling the
    # copy methods to make sure each path is there, unnecessarily duplicating the number of connections requested.
```

ramirezfranciscof:
@giovannipizzi @sphuber Well, I actully can't merge without an approving review now that I notice, he he. BTW: codecov is complaining not sure why, it doesn't show me any "lack of converage" warnings in the code and I also can't figure it out for looking at the details of the check...

aiida/engine/daemon/execmanager.py
sphuber: Text is fine, just a minor nitpick: it is not so much requesting new connections each time, it already has an open connection. But each command *is* calling a command over the transport, which has a certain base delay that can be non-negligible.

sphuber (approved):

PR #4418, opened 2020-10-01
A tiny fix into cmdline utils

blokhin:
`value` should be unconditionally defined


sphuber:
Thanks @blokhin . Could you perhaps tell us how you encountered this bug? It would be good if we can add a quick unit test to make sure the function always behaves properly. I can help you with pointers on where and how to add those if you would like to

blokhin:
@sphuber that happened during the fresh installation of the last version. However we have rather non-typical environment, so it is hard for me to provide you all the details, since there are lot of them. For now, please, just note, the variable is defined inside a condition, and if condition fails, the variable stays not properly defined.

blokhin:
Seems to be fixed.

PR #4419, opened 2020-10-01
Add option to delete builder attributes

### LINKED ISSUE #4381: Add possibility to remove inputs from a builder with `delattr`
Currently, once you have a `builder`, you might want to remove an optional output (e.g. if you get a builder from a `get_builder_restart` and want to drop an optional input).

Currently, this is achievable by deleting an item using dictionary syntax:
```python
del builder['input_name']
```
However, the following syntax does not work:
```python
del builder.input_name
```

I suggest to support also this syntax, since it is already possible (and 'suggested') to use builders with attribute access for setting data, thanks to the TAB-completion support (this can just proxy to the existing deletion function for dictionary items).

ramirezfranciscof:
Fixes #4381 

The builder object was able to delete already set inputs as
items (`del builder['input_name']`), but not as attributes
(`del builder.input_name`). This is not consistent with how
these inputs can be set or accessed (both `builder.input_name`
and `builder['input_name']` are possible).

Also added tests for these two ways of accessing inputs.

### DIFF
--- a/aiida/engine/processes/builder.py
+++ b/aiida/engine/processes/builder.py
@@ -7,7 +7,6 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-# pylint: disable=cell-var-from-loop, line-too-long
 """Convenience classes to help building the input dictionaries for Processes."""
 import collections
 
@@ -49,14 +48,14 @@ class ProcessBuilderNamespace(collections.abc.MutableMapping):
                     return self._data.get(name)
             elif port.has_default():
 
-                def fgetter(self, name=name, default=port.default):
+                def fgetter(self, name=name, default=port.default):  # pylint: disable=cell-var-from-loop
                     return self._data.get(name, default)
             else:
 
                 def fgetter(self, name=name):
                     return self._data.get(name, None)
 
-            def fsetter(self, value):
+            def fsetter(self, value, name=name):
                 self._data[name] = value
 
             fgetter.__doc__ = str(port)
--- a/aiida/engine/processes/builder.py
+++ b/aiida/engine/processes/builder.py
@@ -7,10 +7,11 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-# pylint: disable=cell-var-from-loop
+# pylint: disable=cell-var-from-loop, line-too-long
 """Convenience classes to help building the input dictionaries for Processes."""
 import collections
 
+from aiida.orm import Node
 from aiida.engine.processes.ports import PortNamespace
 
 __all__ = ('ProcessBuilder', 'ProcessBuilderNamespace')
@@ -112,6 +113,9 @@ class ProcessBuilderNamespace(collections.abc.MutableMapping):
     def __delitem__(self, item):
         self._data.__delitem__(item)
 
+    def __delattr__(self, item):
+        self._data.__delitem__(item)
+
     def _update(self, *args, **kwds):
         """Update the values of the builder namespace passing a mapping as argument or individual keyword value pairs.
 
@@ -160,8 +164,6 @@ class ProcessBuilderNamespace(collections.abc.MutableMapping):
         :param value: a nested mapping of port values
         :return: the same mapping but without any nested namespace that is completely empty.
         """
-        from aiida.orm import Node
-
         if isinstance(value, collections.abc.Mapping) and not isinstance(value, Node):
             result = {}
             for key, sub_value in value.items():
@@ -174,7 +176,7 @@ class ProcessBuilderNamespace(collections.abc.MutableMapping):
         return value
 
 
-class ProcessBuilder(ProcessBuilderNamespace):
+class ProcessBuilder(ProcessBuilderNamespace):  # pylint: disable=too-many-ancestors
     """A process builder that helps setting up the inputs for creating a new process."""
 
     def __init__(self, process_class):
@@ -188,4 +190,5 @@ class ProcessBuilder(ProcessBuilderNamespace):
 
     @property
     def process_class(self):
+        """Straighforward wrapper for the process_class"""
         return self._process_class
--- a/aiida/engine/processes/builder.py
+++ b/aiida/engine/processes/builder.py
@@ -37,6 +37,10 @@ class ProcessBuilderNamespace(collections.abc.MutableMapping):
         self._valid_fields = []
         self._data = {}
 
+        # The name and port objects have to be passed to the defined functions as defaults for
+        # their arguments, because this way the content at the time of defining the method is
+        # saved. If they are used directly in the body, it will try to capture the value from
+        # its enclosing scope at the time of being called.
         for name, port in port_namespace.items():
 
             self._valid_fields.append(name)
--- a/aiida/engine/processes/builder.py
+++ b/aiida/engine/processes/builder.py
@@ -193,5 +193,5 @@ class ProcessBuilder(ProcessBuilderNamespace):  # pylint: disable=too-many-ances
 
     @property
     def process_class(self):
-        """Straighforward wrapper for the process_class"""
+        """Return the process class for which this builder is constructed."""
         return self._process_class
--- a/tests/engine/processes/test_builder.py
+++ b/tests/engine/processes/test_builder.py
@@ -15,15 +15,18 @@ from aiida.engine.processes.builder import ProcessBuilder
 from aiida import orm
 
 
-#@pytest.mark.usefixtures('clear_database_before_test')
 def test_access_methods():
-    """Test for the access methods (setter, getter, delete)."""
-    builder = ProcessBuilder(ArithmeticAddCalculation)
+    """Test for the access methods (setter, getter, delete).
 
+    The setters are used again after calling the delete in order to check that they still work
+    after the deletion (including the validation process).
+    """
     node_numb = orm.Int(4)
     node_dict = orm.Dict(dict={'value': 4})
 
     # AS ITEMS
+    builder = ProcessBuilder(ArithmeticAddCalculation)
+
     builder['x'] = node_numb
     assert dict(builder) == {'metadata': {'options': {}}, 'x': node_numb}
 
@@ -37,6 +40,9 @@ def test_access_methods():
     assert dict(builder) == {'metadata': {'options': {}}, 'x': node_numb}
 
     # AS ATTRIBUTES
+    del builder
+    builder = ProcessBuilder(ArithmeticAddCalculation)
+
     builder.x = node_numb
     assert dict(builder) == {'metadata': {'options': {}}, 'x': node_numb}
 
--- a/tests/engine/processes/test_builder.py
+++ b/tests/engine/processes/test_builder.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Tests for `aiida.engine.processes.builder.ProcessBuilder`."""
+
+from aiida.calculations.arithmetic.add import ArithmeticAddCalculation
+from aiida.engine.processes.builder import ProcessBuilder
+
+
+def test_access_methods():
+    """Test for the access methods (setter, getter, delete)."""
+    builder = ProcessBuilder(ArithmeticAddCalculation)
+
+    # AS ITEMS
+    builder['metadata'] = {'label': 'test'}
+    assert dict(builder) == {'metadata': {'label': 'test'}}
+    del builder['metadata']
+    assert dict(builder) == {}
+
+    # AS ATTRIBUTES
+    builder.metadata = {'label': 'test'}
+    assert dict(builder) == {'metadata': {'label': 'test'}}
+    del builder.metadata
+    assert dict(builder) == {}
--- a/tests/engine/processes/test_builder.py
+++ b/tests/engine/processes/test_builder.py
@@ -8,23 +8,43 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Tests for `aiida.engine.processes.builder.ProcessBuilder`."""
+import pytest
 
 from aiida.calculations.arithmetic.add import ArithmeticAddCalculation
 from aiida.engine.processes.builder import ProcessBuilder
+from aiida import orm
 
 
+#@pytest.mark.usefixtures('clear_database_before_test')
 def test_access_methods():
     """Test for the access methods (setter, getter, delete)."""
     builder = ProcessBuilder(ArithmeticAddCalculation)
 
+    node_numb = orm.Int(4)
+    node_dict = orm.Dict(dict={'value': 4})
+
     # AS ITEMS
-    builder['metadata'] = {'label': 'test'}
-    assert dict(builder) == {'metadata': {'label': 'test'}}
-    del builder['metadata']
-    assert dict(builder) == {}
+    builder['x'] = node_numb
+    assert dict(builder) == {'metadata': {'options': {}}, 'x': node_numb}
+
+    del builder['x']
+    assert dict(builder) == {'metadata': {'options': {}}}
+
+    with pytest.raises(ValueError):
+        builder['x'] = node_dict
+
+    builder['x'] = node_numb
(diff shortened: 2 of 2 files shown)


aiida/engine/processes/builder.py
```diff
@@ -7,10 +7,11 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-# pylint: disable=cell-var-from-loop
+# pylint: disable=cell-var-from-loop, line-too-long
```
giovannipizzi, line 10: I wouldn't disable this.

giovannipizzi (commented):
All good thanks! Only, I wouldn't disable the long lines check, I think that's one we want to keep. Otherwise it's good to go for me

aiida/engine/processes/builder.py
```diff
@@ -188,4 +190,5 @@ def __init__(self, process_class):
 
     @property
     def process_class(self):
+        """Straighforward wrapper for the process_class"""
```
sphuber, line 193: Not really a useful docstring I think, how about
```suggestion
        """Return the process class for which this builder is constructed."""
```

aiida/engine/processes/builder.py
sphuber: I agree. I have to say the `cell-var-from-loop` is also probably not a good idea to ignore. I know it was already there, but this might point at a bug

aiida/engine/processes/builder.py
```diff
@@ -7,10 +7,11 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-# pylint: disable=cell-var-from-loop
+# pylint: disable=cell-var-from-loop, line-too-long
 """Convenience classes to help building the input dictionaries for Processes."""
 import collections
 
+from aiida.orm import Node
```
sphuber, line 14: Why are you importing this?

tests/engine/processes/test_builder.py
```diff
@@ -0,0 +1,30 @@
...
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Tests for `aiida.engine.processes.builder.ProcessBuilder`."""
+
+from aiida.calculations.arithmetic.add import ArithmeticAddCalculation
+from aiida.engine.processes.builder import ProcessBuilder
+
+
+def test_access_methods():
```
sphuber, line 16: I think we should add more explicit tests that after calling `del` the actual property still works properly. For example directly after deleting it, you can still set it and the validation still works. You are implicitly checking parts of this because you use the same builder, but only for `__setattr__`. Could you use a port  with validation, for example `builder.x`?

sphuber (changes requested):
Thanks @ramirezfranciscof some requests

aiida/engine/processes/builder.py
greschd: When enabling the `cell-var-from-loop` check, it complains about lines 52 (`port`) and 60 (`name`).

https://github.com/aiidateam/aiida-core/blob/9244969fe53b16a89309624b32a50b0c6b3ee768/aiida/engine/processes/builder.py#L41-L65

To me, the first looks like a false positive, while the second could be a bug. Here's some roughly-equivalent test code:
```python
In [4]: funcs = []

In [5]: mapping = dict()

In [6]: for x in range(10):
   ...:     def foo(val=x):
   ...:         mapping[x] = val
   ...:         return val
   ...:     funcs.append(foo)
   ...:
   ...:

In [7]: [f() for f in funcs]
Out[7]: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

In [8]: mapping
Out[8]: {9: 9}
```

What's puzzling to me is that this looks like quite often-used code. Maybe both are actually false positives. But it certainly warrants looking into; `cell-var-from-loop` is quite a useful check in my experience, since it covers easily-overlooked corner cases of the language.

aiida/engine/processes/builder.py
ramirezfranciscof: It was being imported below (see [here](https://github.com/aiidateam/aiida-core/pull/4419/files#diff-72e781fa100001865d9fa09b91d92567L163)) but pylint was complaining and wanted me to move the import to the top of the file.

aiida/engine/processes/builder.py
ramirezfranciscof: 1. Re `cell-var-from-loop`, I'm not sure I fully understand the problem (it looks somewhat simililar to the recursively expanded and simply expanded variables that were such a headache when coding makefiles).

2. Re `line-too-long`, I've managed most of them but there are two that are still there and I don't know how to fix without introducing intermediate variables (all "aesthetical" modifications are reverted by yapf):

```
    def __dir__(self):
        return sorted(set(self._valid_fields + [key for key, _ in self.__dict__.items() if key.startswith('_')]))
(...)
    if not (isinstance(pruned, collections.abc.Mapping) and not pruned and not isinstance(pruned, Node)):
        result[key] = pruned
```

greschd: @sphuber I'm seriously confused by that code. Specifically, the `setattr(self.__class__, name, getter)` sets the property _on the `ProcessBuilderNamespace` class_ - but doesn't that mean all `ProcessBuilderNamespace` instances will share the same properties?

Also, when trying to reproduce the class behavior, I can't get the `fsetter` to work - instead, the `__setattr__` is needed. But why even have the `fsetter`, then?

The following code might illustrate my confusion somewhat:

```python
class Foo:
    def __init__(self, num_vals, obj_name):
        self._data = {}
        for x in range(num_vals):
            name = chr(97 + x)

            def fgetter(self, name=name, obj_name=obj_name):
                print(f'fgetter for {obj_name}')
                return self._data.get(name)

            def fsetter(self, value, obj_name=obj_name):
                print(f'fsetter for {obj_name}')
                self._data[name] = value

            getter = property(fgetter)
            getter.setter(fsetter)

            setattr(self.__class__, name, getter)

    def __setattr__(self, attr, value):
        if attr.startswith('_'):
            object.__setattr__(self, attr, value)
        else:
            self._data[attr] = value


f = Foo(3, 'f')
print(f.a)
f.a = 2
print(f.a)
try:
    print(f.j)
except AttributeError as exc:
    print(str(exc))

g = Foo(10, 'g')
print(g.j)
print(f.j)
```
Output:
```
fgetter for f
None
fgetter for f
2
'Foo' object has no attribute 'j'
fgetter for g
None
fgetter for g
None
```

greschd: > Re cell-var-from-loop, I'm not sure I fully understand the problem (it looks somewhat simililar to the recursively expanded and simply expanded variables that were such a headache when coding makefiles).

@ramirezfranciscof  Right, so this is bit of an obscure part of Python, it touches on how _function closure_ works. When you define

```python
x = 1
def foo():
    print(x)
```
how is `x` "passed" into `foo`? In the above example it doesn't matter, but what about

```python
funcs = []
for x in range(10):
    def foo():
         print(x)
    funcs.append(foo)

for foo in funcs:
    foo()
```

What would you expect that to print?

---

In short, the closure is done _by name_, and if that name is bound to a different value in the enclosing scope, calling `foo` gets the _changed_ `x`.
So yeah, that's quite a gotcha, and it's nice that `pylint` warns us about it.

greschd: Looking at that code, I actually found a (quite serious IMO) bug, see https://github.com/aiidateam/aiida-core/issues/4420. It's not related directly to this change, though.

ramirezfranciscof: @greschd Ah, I think I understand. When you use an externally defined variable inside the definition of a function, when you call the function it will use whatever is stored in that variable at the time it evaluates the function. We are doing a subtle version of this, not by using the external variable inside the function, but using it to set a default.

I am now wondering what is the proper way to "automatically generate parametrized functions". For example, I want to have a list of 100 functions, where the i-function will add i to the number passed. I would have tried the exact thing in your example:

```python
funcs = []
for x in range(100):
    def foo(n):
         return n+x
    funcs.append(foo)
```

But then this doesn't work...how would you do this then?

greschd: > We are doing a subtle version of this, not by using the external variable inside the function, but using it to set a default.

When setting the default, it actually binds to the _value_ of that variable at the time of _defining_ the function. See the following:

```python
funcs = []
for x in range(10):
    def foo(y=x):
         print('x', x)
         print('y', y)
    funcs.append(foo)

for foo in funcs:
    foo()
```

This is the "false positive", which I think is this `pylint` issue: https://github.com/PyCQA/pylint/issues/3107

The "true positive" in our code is this line:
https://github.com/aiidateam/aiida-core/blob/9244969fe53b16a89309624b32a50b0c6b3ee768/aiida/engine/processes/builder.py#L60

ramirezfranciscof: > The "true positive" in our code is this line:

AH, because it is interpreting `name` as the external variable instead of `name` the internal function parameter?

> This is the "false positive", which I think is this pylint issue: PyCQA/pylint#3107. And incidentally, this is also how you can circumvent the problem.

Do you mean, in my example you would have to pass x as the default of an "unused/artificial/forced" function parameter in order to do my list of summing functions?


```python
funcs = []
for x in range(100):
    def foo(n, int_x=x):
         return n + int_x
    funcs.append(foo)
```

Mmmm It feels kind of hacky (the summing term is an internal parameter, I don't want to have to define an input that the user could potentially overwrite).

greschd: > 
> 
> > The "true positive" in our code is this line:
> 
> AH, because it is interpreting `name` as the external variable instead of `name` the internal function input parameter?
> 

Right, because that function doesn't _have_ a `name` parameter, it's just "captured" from the enclosing scope.

> 
> Do you mean, in my example you would have to pass x as the default of an "unused/artificial/forced" function parameter in order to do my list of summing functions?
> 
> ```python
> funcs = []
> for x in range(100):
>     def foo(n, int_x=x):
>          return n + int_x
>     funcs.append(foo)
> ```
> 
> Mmmm It feels kind of hacky (the summing term is an internal parameter, I don't want to have to define an input that the user could potentially overwrite).

Yep, that'd work. If you want to avoid exposing the extra keyword argument but still bind it to the object, you can use `functools.partial`, like so:

```python
In [1]: from functools import partial

In [4]: funcs = []
   ...: for x in range(10):
   ...:     def foo(y):
   ...:         return y
   ...:     funcs.append(partial(foo, y=x))
   ...:
```

Note also how you wouldn't even have to re-define `foo` in every loop iteration here:
```python
In [6]: funcs = []
   ...: def foo(y):
   ...:     return y
   ...: for x in range(10):
   ...:     funcs.append(partial(foo, y=x))
   ...:
```

For the general behavior, there's also an entry in the Python FAQ: https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result

greschd: And here's the somewhat terse description of name binding, which is the background for _why_ it's behaving that way: https://docs.python.org/3.8/reference/executionmodel.html#naming-and-binding

ramirezfranciscof: > Right, because that function doesn't have a name parameter, it's just "captured" from the enclosing scope.

I see. I was mixing up `fsetter` with `fgetter`, but you are right, the error comes from the one with no `name` parameter. Ok, I think the solution is then to copy the def just above and add the `name=name` thing, right? And add the ignore in the specific line for that other function.

Thanks for the thorough explanation, btw! Crystal clear!

greschd: > Ok, I think the solution is then to copy the def just above and add the `name=name` thing, right?

In principle yes, but I think the rest of that `__init__` code doesn't do exactly what it's supposed to - which is why I opened #4420. I'm not 100% sure, but I think the `fsetter` is never actually invoked. 

But I think it would be ok to fix it in this way here (for the false positive, just explicitly ignore the pylint error _on that line_). Resolving #4420 is probably not closely related to this PR - I just happened to stumble upon it while looking at this.

> Thanks for the thorough explanation, btw! Crystal clear!

Happy to help!

tests/engine/processes/test_builder.py
ramirezfranciscof: I'm not sure I understood what you meant here. Can you check my new update? If that was not it, I may need a re-explanation he - he ( ^ - ^ `).

aiida/engine/processes/builder.py
ramirezfranciscof: Yeah, I just did the modifications. This way you will also be able to get "cleaner" base for the test of #4420 .

aiida/engine/processes/builder.py
```diff
@@ -23,11 +23,13 @@ class ProcessBuilderNamespace(collections.abc.MutableMapping):
...
+        """Dynamically construct the get and set properties for the ports of the given port
+        namespace.
 
-        For each port in the given port namespace a get and set property will be constructed dynamically
-        and added to the ProcessBuilderNamespace. The docstring for these properties will be defined
-        by calling str() on the Port, which should return the description of the Port.
+        For each port in the given port namespace a get and set property will be constructed
+        dynamically and added to the ProcessBuilderNamespace. The docstring for these
+        properties will be defined by calling str() on the Port, which should return the
+        description of the Port.
```
sphuber, line 32: Did a linter adjust the length of all docstrings or did you do this manually? Anyway, please revert. We are maintaining a 120 character limit for the entire project, including docstrings. Doesn't make sense to start changing this here.

aiida/engine/processes/builder.py
```diff
@@ -48,14 +50,14 @@ def fgetter(self, name=name):
...
-                def fgetter(self, name=name, default=port.default):
+                def fgetter(self, name=name, default=port.default):  # pylint: disable=cell-var-from-loop
                     return self._data.get(name, default)
             else:
 
                 def fgetter(self, name=name):
                     return self._data.get(name, None)
 
-            def fsetter(self, value):
+            def fsetter(self, value, name=name):
```
sphuber, line 60: Would be good to add a comment to this method on why we are using `name=name`. Look at the amount of investigation we spent in the PR discussion to figure out why this is necessary. Just give a quick explanation that by binding the value as a default to a keyword argument, you are fixing the current value in the loop, because otherwise it would be scoped to the last value in the iteration.

tests/engine/processes/test_builder.py
```diff
@@ -0,0 +1,50 @@
...
+###########################################################################
+"""Tests for `aiida.engine.processes.builder.ProcessBuilder`."""
+import pytest
+
+from aiida.calculations.arithmetic.add import ArithmeticAddCalculation
+from aiida.engine.processes.builder import ProcessBuilder
+from aiida import orm
+
+
+#@pytest.mark.usefixtures('clear_database_before_test')
```
sphuber, line 18:
```suggestion
```

tests/engine/processes/test_builder.py
```diff
@@ -0,0 +1,50 @@
...
+    assert dict(builder) == {'metadata': {'options': {}}, 'x': node_numb}
+
+    del builder['x']
+    assert dict(builder) == {'metadata': {'options': {}}}
+
+    with pytest.raises(ValueError):
+        builder['x'] = node_dict
+
+    builder['x'] = node_numb
+    assert dict(builder) == {'metadata': {'options': {}}, 'x': node_numb}
```
sphuber, line 37: This is indeed what I meant with my comment that you weren't sure you understood. Would be good to add a quick comment on why we are doing this, because it is a literal repeat of the first few lines. Something like:
```
# Set the same node again to check the setter still works despite deleting the item/attribute
```

sphuber (changes requested):
Thanks @ramirezfranciscof . Almost good to go, just revert the formatting changes and add a few comments as indicated in my comments

aiida/engine/processes/builder.py
ramirezfranciscof: Ok, maybe something in my pylint got messed up, but it didn't let me commit with lines longer than a 100 chars (that is why I originally put the `# pylint: disable=line-too-long` that I was requested to remove).

sphuber: If there is a problem with your dev installation, it would be more sensible to try and fix that, instead of adapting the code to the broken environment, wouldn't it? Anyway, you should simply update your installation by running `pip install -e .[pre-commit]` which will install the correct version of `pylint`. The configuration has moved to `pyproject.toml` instead of the `.pylintrc` file, but this is only supported starting from `pylint==2.6`,

aiida/engine/processes/builder.py
```diff
@@ -188,4 +193,5 @@ def __init__(self, process_class):
 
     @property
     def process_class(self):
+        """Straighforward wrapper for the process_class"""
```
sphuber, line 196: Seems this docstring got reverted incorrectly

ramirezfranciscof: Thanks for the catch! I think it should be ready now.

sphuber (approved):

PR #4421, opened 2020-10-04
Release/1.4.2

sphuber:


ltalirz:
thanks @sphuber , I'll leave the review to @ramirezfranciscof if that's ok, since he reviewed the original rewrite of the logic as well.

In general I'm fine with this change, just one question: what if a user explicitly requests one of the items from the `local_copy_list` also via the `retrieve_list` - will this override the `provenance_exclude_list`?

sphuber:
Just to be clear: this is just a PR of the release, meaning we don't have to revisit the actual changes of the commits that are being merged, except for the last one, which bumps the number and updates the change log. The other commits have already passed through review. In this case there is only one actual commit being merged, but that is because this is a critical bug that I don't want to left unfixed too long.

Regarding your question:
>  what if a user explicitly requests one of the items from the `local_copy_list` also via the `retrieve_list` - will this override the `provenance_exclude_list`?

The `local_copy_list` and `retrieve_list` are not related though. The former has to do with what ends up (or actually what doesn't end up) in the repository folder of the `CalcJobNode` itself, whereas the latter instructs what should go in the `retrieved` output node of said node. So I am not sure what you are asking here.

ltalirz (approved):
Ah sorry, somehow I got confused and thought the fixing commit was part of this PR and hadn't been reviewed yet.

And thanks for the clarification on the `retrieve_list`. 

In that case, it looks fine to me.

PR #4424, opened 2020-10-06
`CalcJob`: add the option to stash files after job completion

### LINKED ISSUE #2150: Add optional `TransportTask` for `CalcJob` to stash files on the remote after the job has finished
A common use case, is that after a `JobProcess` has finished, one might want to copy some files from the work directory, which is often on a scratch space, to a more durable location and attach them through a `RemoteFolder` to the `JobCalculation` node. This is different from the files that are already retrieved for parsing, as often these additional files are big, and one wants to keep them on the remote computer. A typical example are big files that are necessary to restart a calculation from, that are too big to retrieve but also too costly to lose and have to recompute. There is a simple and naive solution of just running an inline calculation that will open transport to the remote computer and use `copy_from_remote_to_remote` to copy the files and create a `RemoteFolder` object to attach to the computer, but this can be dangerous. If an exception occurs, the entire process, most likely the `WorkChain` that launched the `JobCalculation` will except as well. Of course one can catch the exceptions but then the files will not be copied. The `JobProcess` infrastructure has a rich architecture to deal with tasks that require transport and can fail. The best option would be to leverage this functionality for this particular purpose by creating an optional `TransportTask` to copy files after the job has finished, but before going over to the retrieval and parse step.

sphuber:
Fixes #2150 

Tests and documentation still to be added once API is agreed upon.


aiida/engine/daemon/execmanager.py
```diff
@@ -394,6 +394,58 @@ def retrieve_calculation(calculation, transport, retrieved_temporary_folder):
...
+        return
+
+    if stash_mode != StashMode.COPY.value:
+        execlogger.warning(f'stashing mode {stash_mode} is not implemented yet.')
+        return
+
+    execlogger.debug(f'stashing files for calculation<{calculation.pk}>: {source_list}', extra=logger_extra)
+
+    uuid = calculation.uuid
+    target_basepath = os.path.join(stash_options['path'], uuid[:2], uuid[2:4], uuid[4:])
```
giovannipizzi, line 427: TBD (please open issue): have `stash_options['path']` have a default in the AuthInfo

aiida/engine/daemon/execmanager.py
```diff
@@ -394,6 +394,58 @@ def retrieve_calculation(calculation, transport, retrieved_temporary_folder):
...
+
+    for source_filename in source_list:
+
+        source_filepath = os.path.join(calculation.get_remote_workdir(), source_filename)
+        target_filepath = os.path.join(target_basepath, source_filename)
+
+        # If the source file is in a (nested) directory, create those directories first in the target directory
+        target_dirname = os.path.dirname(target_filepath)
```
giovannipizzi, line 435: General comment - we should be careful on using os.path.* methods on remote paths, as the FS might be different. I think anyway we're doing it anywhere, so OK to keep it as it is for now.

aiida/engine/daemon/execmanager.py
```diff
@@ -394,6 +394,58 @@ def retrieve_calculation(calculation, transport, retrieved_temporary_folder):
...
+        transport.makedirs(target_dirname, ignore_existing=True)
+
+        try:
+            transport.copy(source_filepath, target_filepath)
+        except (IOError, ValueError) as exception:
+            execlogger.warning(f'failed to stash {source_filepath} to {target_filepath}: {exception}')
+        else:
+            execlogger.debug(f'stashd {source_filepath} to {target_filepath}')
```
giovannipizzi, line 443:
```suggestion
            execlogger.debug(f'stashed {source_filepath} to {target_filepath}')
```

aiida/engine/daemon/execmanager.py
giovannipizzi: is this the correct name: `stash_options['path']`?

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -218,8 +220,8 @@ async def task_retrieve_job(node, transport_queue, retrieved_temporary_folder, c
     :type cancellable: :class:`aiida.engine.utils.InterruptableFuture`
     :raises: TransportTaskException if after the maximum number of retries the transport task still excepted
     """
-    if node.get_state() == CalcJobState.PARSING:
-        logger.warning(f'CalcJob<{node.pk}> already marked as PARSING, skipping task_retrieve_job')
+    if node.get_state() == CalcJobState.STASHING:
+        logger.warning(f'CalcJob<{node.pk}> already marked as STASHING, skipping task_retrieve_job')
```
giovannipizzi, line 224: should say `task_stash_job`?

giovannipizzi (commented):
Some suggestion for changes (I didn't do a complete review)

ramirezfranciscof:
I think I am a bit wary of introducing more explicit file handling inside of the engine. I rememeber that not so far ago we were discussing some limitations that were the result of the engine copying files directly instead of using the methods of `RemoteData`. Was this for extending the copying of files from one remote machine to another? I don't remember now; I do remember the underlying issue was that these methods would open and close transports in an inefficient way and that it might be convenient to have methods that USE already open transports.

Perhaps if we were to deal with that now (since we are anyways modifying the `RemoteData` class) we could minimize the logic contained inside of the engine, who would just orchestrate the methods of the target remote node. Moreover, this would also prevent having the stash and unstash logic separated (so both would live inside the respective remote class instead of being one in the engine and the other inside the transfer calcjob).

So instead of having a `RemoteFolderData` where we run jobs and `RemoteStashData` where we stash them, we would have a `RemoteFolderData` to run jobs but which could also be used as a class for the stashing (if you just copy), or you could also stash in a `RemoteZipData`, or a `RemoteScriptData` (which uses a bashscript for its methods), etc.

Does this idea make any sense?

sphuber:
> I think I am a bit wary of introducing more explicit file handling inside of the engine. I rememeber that not so far ago we were discussing some limitations that were the result of the engine copying files directly instead of using the methods of `RemoteData`. Was this for extending the copying of files from one remote machine to another? I don't remember now; I do remember the underlying issue was that these methods would open and close transports in an inefficient way and that it might be convenient to have methods that USE already open transports.

Yes, the problem with the current implementation of `RemoteData` is that it will open a new connection itself. If this code is then directly used in processes, if they are run in large quantities, you can easily open too many connections and get banned. You _could_ change the implementation to receive an open connection that it then uses, but then you are simply moving the problem. The code that calls the `RemoteData` is then responsible for opening a connection. If this is not done in a clever way, you end up with the same problem. The advantage of having the engine do these things, is that it has all the necessary infrastructure such as the transport queue.

Note that this is not the only complexity that the engine does well that would be difficult to reproduce by custom code. What happens if your call over transport fails, because the connection is interrupted. The engine has again complex functionality in the EBM to automatically retry and pause if the error persists. Recreating this behavior in client scripts is not trivial. I don't think that you can get away from the fact that if you want to perform tasks over transport, you will have to write  asynchronous code that for most users will not be trivial. It is for that reason that I don't think we _can_ make the `RemoteData` have a simple interface to do operations on. At least not when used in automated processes. For manual use in let's say a shell or command line, that would be different. But you wouldn't be able to just use it in your work chains or calculation jobs. I am not sure if all this is clear.

> So instead of having a `RemoteFolderData` where we run jobs and `RemoteStashData` where we stash them, we would have a `RemoteFolderData` to run jobs but which could also be used as a class for the stashing (if you just copy), or you could also stash in a `RemoteZipData`, or a `RemoteScriptData` (which uses a bashscript for its methods), etc.

I think this refers more to issue #4626 does it? As [I responded there](https://github.com/aiidateam/aiida-core/issues/4626#issuecomment-741675678) to one of Leo's comments, I don't think the class should be determined by what type of stashing was used. It should rather distinguish it from a `RemoteFolderData` in the sense that that one can be readily accessed, whereas the `RemoteStashData` might not be, because it may be archived, written to tape, etc.

> Does this idea make any sense?

That depends. If I understand your concern correctly (correct me if I am wrong) you do not like the concept that the interaction with this remote data has to be done indirectly through the engine. Instead, you would prefer to see this functionality directly accessible and usable through the various data classes. I hope I have explained clearly why I think that this approach won't work for cases where these actions have to be done in an automated way, that is to say in workchains. Of course for the single manual operation it might work, but I doubt that this is the majority of the use cases.

aiida/engine/daemon/execmanager.py
sphuber: Nope, that should be `stash_options['target_basepath']`

aiida/engine/daemon/execmanager.py
sphuber: Good point. Surprised this has never caused problems. Wouldn't this break when you run AiiDA on windows and submit to a linux based computer? For example the `upload_calculation` function would be formatting paths to be created on the remote using windows path separators. Guess no one has ever used a Windows client or a Windows-based cluster :sweat_smile:

aiida/engine/processes/calcjobs/tasks.py
sphuber: nope. This code is in the `task_retrieve_job`. Since stashing is done before retrieving, this is a safeguard to ensure it is only executed once.

aiida/engine/daemon/execmanager.py
giovannipizzi: I guess we suggest to use the WSL in windows so it's still using unix-like paths (I think?) and I never saw a Windows supercomputer (where I think anyway many of the actual codes might also break :-) )

giovannipizzi:
To comment on the discussion above: I think this might be right moment to change the interface of the RemoteData concerning auto-opening transports.
I would add a parameter `transport` to all concerned functions, e.g. `is_empty`, `getfile`, `listdir`, `listdir_withattributes`, `_clean`.
For backward compatibility, we make it optional for now - if not specified, we open a transport but also raise a deprecation warning. In 2.0, we make the parameter compulsory.

We instead add a `get_transport` method to do it (I would make it public, I think it's convenient for the users - this is the reason we were doing it "implicitly").

If possible (but I'm not sure it's easy) we should also check that the transport is "correct", i.e. to the correct machine (but I'm not sure it's easy).

Reason: while I agree that this only shifts the problem, it shifts it in the right direction: now, a conscious user cannot avoid to open a lot of connections. After this change, you can at least do something like:

- get a list of remote data, group by computer
- open the transport once
- call all methods, for all remote data on the same computer, within the *same* connection

This will open up the possibility in the future to have e.g. a workflow step get a transport to perform operations (now you have to reimplement those methods instead, if that possibility is implemented).

giovannipizzi:
Regarding multiple (sub)classes vs. same class with an attribute to indicate the type: I guess it's (up to a certain level) a matter of taste.
In the meeting a few days ago, we converged on a single StashData, with attributes distinguishing the type (and different from RemoteFolderData).
I guess having an attribute makes it easy to support more types without having to define more classes.
The only thing to think about is how different the Python API will be.
If we expect them to be essentially the same, independent of the StashData type, then a single class+attribute to distinguish is OK.
If we want to provide API to act on each of them, and this is different, we should then probably instead go the multiple-subclasses approach? I'm not sure however of what we might want to support.
Definitely, however, even if we support multiple subclasses, I would vote for a single (sub)class for the StashData generated by the custom script installed on the supercomputer (to be still implemented), since we have no control on what the script will do except for the API that was used to invoke it.


Finally, regarding conversion to RemoteFolderData:
- in the case of `copy`, we can/should provide a simple calcfunction that converts a `StashData` in mode `copy` to a `RemoteData`
- for any other type, I think converting a StashData to a RemoteData should/could be done with new options of the TransferCalcJob
-

ramirezfranciscof:
> Note that this is not the only complexity that the engine does well that would be difficult to reproduce by custom code. What happens if your call over transport fails, because the connection is interrupted. The engine has again complex functionality in the EBM to automatically retry and pause if the error persists. Recreating this behavior in client scripts is not trivial. I don't think that you can get away from the fact that if you want to perform tasks over transport, you will have to write asynchronous code that for most users will not be trivial. It is for that reason that I don't think we can make the RemoteData have a simple interface to do operations on. At least not when used in automated processes. For manual use in let's say a shell or command line, that would be different. But you wouldn't be able to just use it in your work chains or calculation jobs. I am not sure if all this is clear.

I don't see that what I'm proposing necessarily implies giving up on the engine extra safeguards. For example, I'm thinking about changing:

```python
        try:
            transport.copy(source_filepath, target_filepath)
        except (IOError, ValueError) as exception:
            execlogger.warning(f'failed to stash {source_filepath} to {target_filepath}: {exception}')
```

For:

```python
        try:
            remote_folder.raw_copy(transport, source_filepath)
        except (SomeErrorType) as exception:
            execlogger.warning(f'failed to stash {source_filepath} to {remote_folder.path}: {exception}')
```

The `raw_copy` method does not currently exist but would be like the `copy` one except it uses the provided transport instead of opening one. This might seem like not much but would for example allow to instead of specifying a string for the kind of stashing it could just pass the class (`RemoteFolderData`, `RemoteZipData`, `RemoteScriptData`, etc.) which could significantly reduce upkeeping and interdependence between the different parts of the code.

One thing that might complicate this, however, is that a real generic copy procedure depends both on the transport and class of the source and the transport and class of the target. For example, I am not sure what kind of transport is being used in the example above (or in the execmanager in general) because it allows to copy a source from the remote machine into a target in the same remote machine: is this `ssh`? Because I would expect `ssh` transport copy stuff from the local machine to the remote machine. Now looking at the plugin code, I see that apparently it has methods for both: `copy` for the first and `put` for the second. More confusingly though, the local transport also has both these methods? And now I'm starting to get lost on how transports are supposed to work...can you copy files between machines with `local` transport `put`?

> I think this refers more to issue #4626 does it? As I responded there to one of Leo's comments, I don't think the class should be determined by what type of stashing was used. It should rather distinguish it from a RemoteFolderData in the sense that that one can be readily accessed, whereas the RemoteStashData might not be, because it may be archived, written to tape, etc.

Yeah, sorry; I did see the comment there. I actually started writting my comment there, but since it was a more "bigger picture" and related many aspects, I wasn't sure exactly where to put it and ended up in here (there are a lot of current threads related to the different parts of this project). Also, maybe this part you quote makes more sense in combination with what I'm proposing before, so the class actually contains the logic on how to access the data through its methods.

sphuber:
> To comment on the discussion above: I think this might be right moment to change the interface of the RemoteData concerning auto-opening transports. I would add a parameter `transport` to all concerned functions, e.g. `is_empty`, `getfile`, `listdir`, `listdir_withattributes`, `_clean`.

I agree that this would definitely be a step in the right direction. These methods should ultimately be moved from the `RemoteData` to the `RemoteFolderData`, because the `RemoteData` will no longer necessarily be a "folder". It won't even be instantiable any more. Have to think of how the deprecation pathway would go.

> For backward compatibility, we make it optional for now - if not specified, we open a transport but also raise a deprecation warning. In 2.0, we make the parameter compulsory.

I am afraid that might be too fast. We sort of agreed to no longer deprecate things in 1.6 that will be removed in 2.0.

> We instead add a `get_transport` method to do it (I would make it public, I think it's convenient for the users - this is the reason we were doing it "implicitly"). If possible (but I'm not sure it's easy) we should also check that the transport is "correct", i.e. to the correct machine (but I'm not sure it's easy).

This is indeed a problem and a critical check to perform but I am not sure it is currently possible. The `Transport` only get's the hostname of the `Computer` instance. Now it stands to argue that the hostname should uniquely determine the actual computer, however, in AiiDA it is definitely possible to have multiple `Computer` instances with the same hostname. A `RemoteData` will however be coupled to only one of those instances. It is currently not possible to determine from the `Transport` instance which computer that should be. However, I am also not sure if getting the wrong computer in that case could cause problems. They do represent the same physical machine after all. I guess the easiest would be to pass in the `Computer` instance in the `Transport` constructor, but this then tightly couples that class to an AiiDA ORM class, which is subideal as well.

> Reason: while I agree that this only shifts the problem, it shifts it in the right direction: now, a conscious user cannot avoid to open a lot of connections. This will open up the possibility in the future to have e.g. a workflow step get a transport to perform operations (now you have to reimplement those methods instead, if that possibility is implemented).

While this is true, this does not solve the problem of connection problems. If you add this code into your workflow, even though it may use controlled transports to not overflow the machine, the transport operation may fail, which will except and which will except the entire workchain, toppling everything upstream as well. I think one would _have_ to write this code such that it is wrapped in an EBM-like retry. This is possible, it would require the user to start writing asynchronous code.


@ramirezfranciscof I see now that you are proposing to move the logic for the actual operations as much as possible to within the data types. I am not opposed to this per sé and might be good in certain cases. Would have to look more into this. However, as explained above, this would still not solve the main problems unless they are actually _called_ by the engine in the calcjobs. Users calling this code themselves in their workchains is problematic. Now it would be great if we could fix this in a generic way, but this will certainly be far from trivial.

> The raw_copy method does not currently exist but would be like the copy one except it uses the provided transport instead of opening one. This might seem like not much but would for example allow to instead of specifying a string for the kind of stashing it could just pass the class (RemoteFolderData, RemoteZipData, RemoteScriptData, etc.)

What do you mean with "instead of specifying a string"? You mean the input in the stash metadata options to indicate the type of stashing that needs to be performed? I don't see what passing a class would solve. Since this will have to be stored as an attribute, the class will have to be serialized, which essentially means it is transformed into a string anyway.

> For example, I am not sure what kind of transport is being used in the example above (or in the execmanager in general) because it allows to copy a source from the remote machine into a target in the same remote machine: is this ssh?

It is whatever transport the computer is configured to require, and the computer being the one to which that calculation job was submitted. If it was submitted to a computer with local transport, it will be a local transport instance.

> Because I would expect ssh transport copy stuff from the local machine to the remote machine.

You are confusing the fact here that the "remote" machine is simply the computer to which the calcjob was submitted. This might just as well be the localhost, in which case it would most likely use `local` transport but not necessarily. You can also configure to connect over ssh to your localhost. Again, the "remote" computer in AiiDA is not always "really remote". Important not to forget that.

> More confusingly though, the local transport also has both these methods? And now I'm starting to get lost on how transports are supposed to work...can you copy files between machines with local transport put?

The `Transport` is simply an abstract interface to copy and retrieve files and folders from one computer to the other. AiiDA is always running on the "one" computer, and it is connecting to the "other". The `SshTransport` facilitiates this interaction if the "other" computer is (only) reachable over an SSH connection. The `LocalTransport` implements the operations if the "other" computer _is_ the same as the "one" computer. That's all.

ramirezfranciscof:
> It is whatever transport the computer is configured to require, and the computer being the one to which that calculation job was submitted. If it was submitted to a computer with local transport, it will be a local transport instance.
> 
> You are confusing the fact here that the "remote" machine is simply the computer to which the calcjob was submitted. This might just as well be the localhost, in which case it would most likely use `local` transport but not necessarily. You can also configure to connect over ssh to your localhost. Again, the "remote" computer in AiiDA is not always "really remote". Important not to forget that.
> 
> The `Transport` is simply an abstract interface to copy and retrieve files and folders from one computer to the other. AiiDA is always running on the "one" computer, and it is connecting to the "other". The `SshTransport` facilitiates this interaction if the "other" computer is (only) reachable over an SSH connection. The `LocalTransport` implements the operations if the "other" computer _is_ the same as the "one" computer. That's all.

Ah, I see. Yes, I do know the remote computer can actually be your local one (I have run in localhost many times, haha), but I thought it might use the same `ssh` transport for that (I mean, you can ssh to your own computer), but it is true that it does not have the information to do that is the computer transport is set to local (i.e. passwords, key, etc that goes into the computer config). It makes sense in retrospect.

Also thanks for the explanation on the transports; so basically for `local` transport the `put` and `copy` methods will do the same thing.

> @ramirezfranciscof I see now that you are proposing to move the logic for the actual operations as much as possible to within the data types. I am not opposed to this per sé and might be good in certain cases. Would have to look more into this. However, as explained above, this would still not solve the main problems unless they are actually _called_ by the engine in the calcjobs. Users calling this code themselves in their workchains is problematic. Now it would be great if we could fix this in a generic way, but this will certainly be far from trivial.

I see what you mean; I guess @giovannipizzi was thinking more exclusively on allowing the user the "performant" way of transporting, not necessarily the whole retry on error infrastructure. They would anyways be able to do some manual retry, no? Like, for example:

```python
# very pseudo pseudo-code
tries = 0
while error and tries < 5:
    tries =+ 1
    with RaisedError as error:
       folder_node.copy(transport,file)
```

The "general solution" problem seems interesting though. I clearly don't know enough of how this part work to propose a very specific solution, but would it be possible to automatically convert the steps of the workchain into async routines? And then maybe when you execute them if they catch a `RetryError` type (which would have to be manually raised by the workchain designer) it would mean they have to retry this step?
 
> What do you mean with "instead of specifying a string"? You mean the input in the stash metadata options to indicate the type of stashing that needs to be performed? I don't see what passing a class would solve. Since this will have to be stored as an attribute, the class will have to be serialized, which essentially means it is transformed into a string anyway.

Well, sure you serialize to store but you already have rules to serialize classes, no? What I mean is now you don't need to create and maintain an `extra` equivalence relationship with the enum of possible stashing methods (you have the def of the enum on one part of the code, but then you have to "if ==" through it to decide which code to execute). The classes available are the possible stashing methods (and the code of what to do is in the class and passing the class is giving the instructions, there is only one place where the option is fully specified).

If you will, it would be the same if your enum was instead a dict with `translator = {option_name: code_to_execute}` and you just `execute translator[option_passed]` (if such a thing was possible).

Since now there is only one option possible, perhaps I'm not imagining correctly how you were thinking to implement the selector of what to do.

sphuber:
> I see what you mean; I guess @giovannipizzi was thinking more exclusively on allowing the user the "performant" way of transporting, not necessarily the whole retry on error infrastructure. They would anyways be able to do some manual retry, no? 

Yes, they would be able to do this. Still that doesn't solve the problem where people will put this in workflows, which is where it will break down. You will want a "sleep" in your retry to give a grace period, and if you don't do this async, you block the entire thread. Then of course you still want to "pause" and not except after max retries which again is not trivial.

> The "general solution" problem seems interesting though. I clearly don't know enough of how this part work to propose a very specific solution, but would it be possible to automatically convert the steps of the workchain into async routines? And then maybe when you execute them if they catch a `RetryError` type (which would have to be manually raised by the workchain designer) it would mean they have to retry this step?

As far as I am aware, it is not possible to convert sync code to async code on-the-fly. It is not just slapping `async` in front of the method definition, you will also have to add `await` in parts of the code where you call other coroutines. The only solution is to require users to write async code to begin with. But I am not sure if we can allow to write "mixed" code, i.e. have only _some_ of the outline steps in a workchain be async.

> Since now there is only one option possible, perhaps I'm not imagining correctly how you were thinking to implement the selector of what to do when we extended the functionality.

No, that is more or less how I was imagining implementing it and extending it once more modes become available. I don't really see how this is such a problem though. Sure you can remove the if/else and just rely on passing the class and passing the single responsible method. To me that doesn't change that much, but it does add complexity elsewhere. We do not yet support serialization of arbitrary classes in attributes. Additionally, your approach assumes that the interface of this `stash` method on those `RemoteStashData` subclasses can have exactly the same interface. For normal copy and tar, maybe yes, but also for the custom remote script one? In the end, you may end up doing the same if/else, but on the class instead of the mode enum. Again, I don't see this as the main problem in this whole discussion.

**TLDR: SUMMARY OF THREAD** to summarize this very long discussion in how I see things:
* The question and difference of opinion seems to be whether the `RemoteData` classes should function mostly as a data _representer_ or as an _operator_. That is to say, should they merely store and reflect what data they contain, or should they also actually be functional classes to operate on that data.
* From how classes are typically used in Python, it seems intuitive to make them _operators_ and therefore implement functionality to manipulate their data. This seems to be what @ramirezfranciscof is suggesting.
* On the contrary, I am not convinced that this is a wise design. The `RemoteData` needs a lot more information that is not really part of _it_, to perform these operations on data. It depends on the computer that it is one (which, technically can be retrieved from the `RemoteData`) and a transport to connect to that computer. The `RemoteData` is really just a "symbolic link". Any instance of it, doesn't really have access to its data, like an instance of a class typically has. It just has a reference which allows to address it, but only when it gets external resources that are required. It is therefore that I think we should treat these objects more as simple containers of a _reference_ to data. The operator should be external to it and use it just to get the correct "address" and format of the data they reference.
* The problem that I see with treating `RemoteData` as operator classes, is shown when it is used in automated code. It needs to have the transport passed in (doable, but already moves in the direction of the `RemoteData` no longer being the operator) and it does not have the capability itself to automatically recover from transient problems. Also this has to be done from the outside.
* Now of course, these objections are not as much of a problem when used interactively, but AiiDA is mostly an automated infrastructure and not primarily designed to be an interactive tool.
* That being said, I see the argument that it should be usable as both. My final observation is just that if you give the `RemoteData` class "operating" abilities, then people will use it and they will put it in workflows. We can of course document that they shouldn't do this, but we will then still have to implement the operating code in the engine.

**FINAL QUESTION** Does any of this discussion have any impact on the fundamental question of whether we should have the stashing functionality on a `CalcJob` at all? Or are we merely talking on how the various subclasses of `RemoteData` should be organized and what exact functionality they should possess?

sphuber:
@chrisjsewell my `pre-commit` and `docs` checks started failing here ever since you released `plumpy==0.18.4`. Could you have a look and see whether you know why and if this is something we have to fix in `aiida-core` or in `plumpy`?

chrisjsewell:
> @chrisjsewell my pre-commit and docs checks started failing here ever since you released plumpy==0.18.4. Could you have a look and see whether you know why and if this is something we have to fix in aiida-core or in plumpy?

well that will be motivation for you to have a look at #4669 then 😉

chrisjsewell:
as I mention there, I'm not quite sure why mypy is not recognizing e.g. `plumpy.Wait`, and you need to have the full path `plumpy.process_spec`Wait`

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -361,12 +409,16 @@ def load_instance_state(self, saved_state, load_context):
...
         node = self.process.node
         transport_queue = self.process.runner.transport
-        command = self.data
         result: plumpy.process_states.State = self
 
+        if isinstance(self.data, tuple):
+            command, exit_code = self.data
+        else:
+            command = self.data
+
```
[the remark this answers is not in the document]
mbercx: I see now that on line 526 a state is created with `data=(STASH_COMMAND, exit_code)`. Still, do we not have to replace `self.data` on line 439 and 451 with `command`?

mbercx (commented):
Thanks @sphuber! I gave the code a first pass and left some comments. I'll do a bit of field testing and if I don't run into any issues write a first draft for the documentation.

setup.json
```diff
@@ -163,6 +163,7 @@
             "numeric = aiida.orm.nodes.data.numeric:NumericType",
             "orbital = aiida.orm.nodes.data.orbital:OrbitalData",
             "remote = aiida.orm.nodes.data.remote:RemoteData",
```
mbercx, line 165: While field testing, I got this error:

```
aiida.common.exceptions.StoringNotAllowed: class `aiida.orm.nodes.data.remote.base:RemoteData` does not have registered entry point
```

I'm assuming since this needs needs to be changed to:

```suggestion
            "remote = aiida.orm.nodes.data.remote.base:RemoteData",
```

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -97,6 +97,31 @@ def validate_calc_job(inputs: Any, ctx: PortNamespace) -> Optional[str]:  # pyli
     return None
 
 
+def validate_stash_options(stash_options: Any, _: Any) -> Optional[str]:
+    """Validate the ``stash`` options."""
+    from aiida.common.datastructures import StashMode
+
+    # Normally we would check if `stash_options` is not `None, but due to a bug in `plumpy`, even if not specified, the
```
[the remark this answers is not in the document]
sphuber: yeah, it was fixed in `0.17.0`. I kept it in because there was a reason I couldn't rely on this version of `plumpy` being installed, but I realize now that that is for `aiida-quantumespresso`. For the current state of `aiida-core` we require `plumpy~=0.18.4` so we can indeed remove this guard.

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -216,9 +241,14 @@ def define(cls, spec: CalcJobProcessSpec) -> None:  # type: ignore[override]
         spec.input('metadata.options.additional_retrieve_list', required=False,
             valid_type=(list, tuple), validator=validate_additional_retrieve_list,
             help='List of relative file paths that should be retrieved in addition to what the plugin specifies.')
+        spec.input_namespace('metadata.options.stash', required=False, populate_defaults=False,
+            validator=validate_stash_options,
+            help='Optional directives to stash files after the calculation job has completed.')
```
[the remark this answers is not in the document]
sphuber: No reason. Will add :)

setup.json
sphuber: Hmm, I thought that would have worked since it is also importable from `aiida.orm.nodes.data.remote`. But apparently the entry point needs the direct path where it is defined. Will change it.

aiida/engine/processes/calcjobs/tasks.py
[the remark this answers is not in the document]
sphuber: yes, those should become `command` indeed, will adapt

mbercx:
@sphuber is this ready for another round of review/field testing? 🙃

sphuber:
In fact yes, I hadn't said so yet because the tests are failing. The unit tests are passing but the tests hang in the daemon tests, .co/test_daemon.py . If you could have a look why that'd be great! I can't work on it this week unfortunately
  

  

  

  
  
  
  
  
>

aiida/orm/nodes/data/remote/stash.py
```diff
@@ -0,0 +1,84 @@
...
+    A stashed folder is essentially an instance of ``RemoteData`` that has been archived. Archiving in this context can
+    simply mean copying the content of the folder to another location on the same or another filesystem as long as it is
+    on the same machine. In addition, the folder may have been compressed into a single file for efficiency or even
+    written to tape. The ``stash_mode`` attribute will distinguish how the folder was stashed which will allow the
+    implementation to also `unstash` it and transform it back into a ``RemoteData`` such that it can be used as an input
+    for new ``CalcJobs``.
+    """
+
+    def __init__(
+        self, stash_mode: StashMode = None, target_basepath: str = None, stash_list: typing.List = None, **kwargs
```
mbercx, line 24: Caught another bug during field testing: 🙃

```suggestion
        self, stash_mode: StashMode = None, target_basepath: str = None, source_list: typing.List = None, **kwargs
```

mbercx:
> In fact yes, I hadn't said so yet because the tests are failing. The unit tests are passing but the tests hang in the daemon tests, .co/test_daemon.py . If you could have a look why that'd be great! I can't work on it this week unfortunately

Not entirely sure yet why the tests are hanging, but something funky is going on when I try to run a calculation job without stashing:

```console
$ verdi process list -ap1
  PK  Created    Process label             Process State    Process status
----  ---------  ------------------------  ---------------  ------------------------------------------------------------------------------------
  10  16m ago    ArithmeticAddCalculation  ⏹ Finished [0]   Waiting for transport task: retrieve
```

Probably something went wrong in the changes of the `Waiting` state, am going to look into it now.

aiida/engine/daemon/execmanager.py
```diff
@@ -404,6 +404,63 @@ def retrieve_calculation(calculation: CalcJobNode, transport: Transport, retriev
...
+            transport.copy(source_filepath, target_filepath)
+        except (IOError, ValueError) as exception:
+            execlogger.warning(f'failed to stash {source_filepath} to {target_filepath}: {exception}')
+        else:
+            execlogger.debug(f'stashed {source_filepath} to {target_filepath}')
+
+    remote_stash = RemoteStashData(
+        computer=calculation.computer,
+        target_basepath=target_basepath,
+        stash_mode=stash_mode,
```
mbercx, line 458: I think here we have to convert the `stash_mode` into its `StashMode` enum:

```suggestion
        stash_mode=StashMode(stash_mode),
```

aiida/orm/nodes/data/remote/stash.py
```diff
@@ -0,0 +1,84 @@
...
+        return self.get_attribute('target_basepath')
+
+    @target_basepath.setter
+    def target_basepath(self, value: str):
+        """Set the target basepath.
+
+        :param value: the target basepath.
+        """
+        type_check(value, str)
+        self.set_attribute('target_basepath', value.value)
```
mbercx, line 67:
```suggestion
        self.set_attribute('target_basepath', value)
```

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -467,6 +538,9 @@ def parse(self, retrieved_temporary_folder: str) -> plumpy.process_states.Runnin
             ProcessState.RUNNING, self.process.parse, retrieved_temporary_folder
         )  # type: ignore[return-value]
 
+    def finalize(self, exit_code):
+        return self.create_state(ProcessState.RUNNING, lambda: exit_code)
```
mbercx, line 542: My `plumpy` knowledge is probably lacking, but I don't understand why we pass the `exit_code` as a lambda function here.

mbercx (commented):
@sphuber I still left some final comments, after those I left one by one while I was debugging the stashing feature. Still don't understand why the daemon tests are hanging, though...

aiida/orm/nodes/data/remote/stash.py
sphuber: looks like I have a test to add :)

aiida/engine/processes/calcjobs/tasks.py
sphuber: This is actually a work around to an awful hack and I will change it. Essentially `create_task` expects a callable as the first argument, which is normally the function that should be called once the `Running` state is entered again. Since I didn't really have a function to call anymore, since the process was already parsed and therefore done, I return a lambda. The lambda simply return the exit code, which is what the parsing method would have done.

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -396,12 +452,22 @@ async def execute(self) -> plumpy.process_states.State:  # type: ignore[override
...
                 result = self.retrieve()
 
-            elif self.data == RETRIEVE_COMMAND:
+            elif command == RETRIEVE_COMMAND:
                 node.set_process_status(process_status)
                 # Create a temporary folder that has to be deleted by JobProcess.retrieved after successful parsing
                 temp_folder = tempfile.mkdtemp()
                 await self._launch_task(task_retrieve_job, node, transport_queue, temp_folder)
-                result = self.parse(temp_folder)
+                exit_code = self.process.parse(temp_folder)
```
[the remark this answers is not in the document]
sphuber: This code has been reverted so we still use `self.parse`

sphuber:
@mbercx tests are passing now. You can check the relevant changes in the last commit. Not important to keep separate but just to make it easier to see the changes after review. Now only Jenkins is failing....

mbercx (dismissed):
Great, thanks @sphuber! Final changes look good to me. I've also field tested this both locally using an `ArithmeticAddCalculation` and remotely with an `aiida-quantumespresso` `PwBaseWorkChain`. Everything seems to be working as expected, so for me this is good to go.

sphuber:
Thanks for the review @mbercx . There were some merge conflicts I had to fix, so had to force push which canceled your review. Before you reapprove, just wanted to highlight two more points concerning the design:

1. The `CalcJob` option is called `target_base` whereas the resulting attribute on the `RemoteStashData` is called `target_basepath`. This initially caught me off guard but when thinking about it, it actually makes sense. The `target_base` is currently used to communicate the absolute path on the remote machine, but the `target_basepath` will include additional subfolders based on the sharded calculation node UUID. I was tempted to rename `target_base` _also_ to `target_basepath` but then remembered that once we implement different stashing modes it no longer necessarily corresponds to a filepath. Anyway, in short, I think the current naming is correct, just wanted to highlight it to make sure you agree.
2. The original design _only_ stashed if the parsing returned exit code zero. I changed this such that it _always_ stashes, before parsing even happens. I think this is the better default but in the future we might want to add options to maybe change this. Although if we want to allow complex rules based on the exact exit code returned, rather than just an on/off switch, the interface might become too complicated. But I would leave that for a later worry.

@giovannipizzi your thoughts before merging this?

mbercx (approved):
Thanks @sphuber! I agree with both comments. In case the stashing fails because a file is missing, it should just add a warning message to the log, right?

I'll already re-approve the PR, will ask @giovannipizzi to have a final look during the AiiDA meeting.

giovannipizzi:
Thanks to all of you!

A few questions to double check before I merge:
- the entry point is now `remote.stash`, which means looking for RemoteData will also find these new nodes by default (of course one could do subclassing=False, but then if we add a third 'remote.somethingelse' it becomes difficult to e.g. only filter for `remote` and `remote.stash` but not `remote.somethingelse`). I think (but I'm not sure anymore as we discussed a lot... hence I'm asking) that we eventually decided to just give it a different entry point (e.g. stash) and one can use tuples in the QueryBuilder to get both `(RemoteData, StashData)`.
- on a similar note, are we planning to make the extension of StashData (for e.g. zip etc.) the same class or a different class? In this case I would have defined this e.g. as a `StashFolderData` and give it an entry point `stash.folder`, have a base class StashData (entry point `stash`) and allow for more modes in the future (`stash.script`, `stash.zip`, ...). Again, sorry to raising this again, but I'm not sure anymore what was the final decision and if/why we agreed that the current implementation is better
-testing: there are some lines that are not covered by testing - is it possible to run a test that actually does the stashing, in the daemon tests? Or this is the case but those lines are not included in the test coverage?
- base vs. basepath: I think the general attribute should be called `base` (as you say in the future could be e.g. a zip file) so OK. Base path is indeed different, so ok to keep a different name. *But*: this works only if new extensions (zip, remote script) will be different Data classes - if they are the same, I'm not sure that using `basepath` in the attributes is a good idea (so I need also here a refresher)
- docs are missing, but I made @mbercx swear that he would prepare them soon, so OK :-)

Sorry for the additional questions, I just want to make sure that we are not making a design mistake that then is hard to fix to keep backward-compatibility

sphuber:
Thanks for the questions @giovannipizzi , it is exactly why I asked you to have another look, so I get what I asked for :)

> the entry point is now `remote.stash`, which means looking for RemoteData will also find these new nodes by default (of course one could do subclassing=False, but then if we add a third 'remote.somethingelse' it becomes difficult to e.g. only filter for 
`remote` and `remote.stash` but not `remote.somethingelse`).

This is a generic problem though and not really a result of the choices made here, right? For example, trying to find all `ArrayData` and `KpointsData` but not include `BandsData` nor `TrajectoryData` is similarly also not possible in a single query. In the end though, I feel that semantically  `RemoteStashData` _should_ be in the `remote.` branch of entry points. I would like to have all symbolic "remote" data types to be in the same entry point branch. Ideally the current `RemoteData` that corresponds to the base `remote` entry point should become a non-instantiable class and the current implementation should become `RemoteFolderData` with entry point `remote.folder`. The only reason we didn't do it now is because it would break existing code by changing `RemoteData` to `RemoteFolderData`. But so I would already put the `RemoteStashData` in the `remote.` branch of entry points given that changing entry points in the future is essentially a no-go.

> on a similar note, are we planning to make the extension of StashData (for e.g. zip etc.) the same class or a different class? In this case I would have defined this e.g. as a `StashFolderData` and give it an entry point `stash.folder`, have a base class StashData (entry point `stash`) and allow for more modes in the future (`stash.script`, `stash.zip`, ...). Again, sorry to raising this again, but I'm not sure anymore what was the final decision and if/why we agreed that the current implementation is better

It's a good point. I think originally the plan was to not use various subclasses but store the "mode" as an attribute. However, thinking about it now, I think this approach may be the wrong one. Using a single class for fundamentally different things and switching through some attributes has given problems in the past already (looking at you `KpointsData`). Maybe with the current implementation we are indeed falling into the same trap. So maybe we should indeed create an non instantiable `remote.stash` entry point and then have `remote.stash.folder : RemoteStashFolderData` as the one that is used for stashing using the `copy` mode. 

>  base vs. basepath: I think the general attribute should be called `base` (as you say in the future could be e.g. a zip file) so OK. Base path is indeed different, so ok to keep a different name. _But_: this works only if new extensions (zip, remote script) will be different Data classes - if they are the same, I'm not sure that using `basepath` in the attributes is a good idea (so I need also here a refresher)

This point would also be solved by accepting the suggestion in the previous point. If we make the `RemoteStashFolderData` and `RemoteStashZipData` different subclasses and entry points, we can have `basepath` in the former, and `path` (or something like that) in the latter. So this problem is another motivation to use separate subclasses for the various types of stashing.

>  testing: there are some lines that are not covered by testing - is it possible to run a test that actually does the stashing, in the daemon tests? Or this is the case but those lines are not included in the test coverage?

There actually is a test in the `test_daemon.py` script that performs the stashing. The uncovered lines are the handling of exceptions.
 
>     docs are missing, but I made @mbercx swear that he would prepare them soon, so OK :-)

Love me some docs. Would indeed be great if you could add them @mbercx :+1:


**Summary of action points:** compiling the comments, I think I have to do the following, please comment if you agree.
- Create the new `remote.stash.folder: RemoteStashFolderData` to be used for the `copy` mode.
- Change the current `remote.stash: RemoteStashData` entry point to be a non-instantiable base class
- I would leave the `stash` based entry points under the `remote.` hierarchy. Any downsides beside the query problem marked by Giovanni?

giovannipizzi:
> Create the new remote.stash.folder: RemoteStashFolderData to be used for the copy mode.

OK

> Change the current remote.stash: RemoteStashData entry point to be a non-instantiable base class

OK

> I would leave the stash based entry points under the remote. hierarchy. Any downsides beside the query problem marked by Giovanni?

I see your comment on `array` vs `array.kpoints` (even if in that case it's really just an `ArrayData` node with some additional python methods, so really a subclass also in the python sense, here is a different implementation). On the other hand I agree that it's more intuitive if they live all inside `remote`. I don't see other downsides currently (and I really see limited need to query only for exactly RemoteData + only one subclass of stash) so OK, for me, to go ahead like this, also not to block this forever, and I think this is the easiest approach.
[EDIT: thinking better to it, one issue is people searching for all output `RemoteData` nodes of a ProcessNode and now (by default, with subclassing = True) finding more than one. But maybe it's OK - there is the word 'remote' in the `RemoteStash*Data` class name(s), and if they have some stash data, it means they explicitly asked to stash].

It would be great, though, if @mbercx (and @ramirezfranciscof ?) could confirm that there are no other issues, and (especially for @mbercx who's been keeping the design notes updated) could confirm that we're not missing one of the points we've been discussing, before you @sphuber do the work.

mbercx:
I don't know the history of that _naughty_ `KpointsData` class, but I also don't see any downsides to having multiple classes for different types of stashing modes. It just offers us more flexibility in the future, and allows the user to query for a specific type of stashed data.

I've gone through the notes again, and I don't see any more issues with the currently proposed implementation. Just wanted to highlight again that we should also consider the unstashing of the data, via e.g. the `TransferCalcJob`. I can't think of any problems right now, but maybe you can still think of something that I've missed.

sphuber:
@ramirezfranciscof and @mbercx this is ready for review

mbercx (changes requested):
Thanks @sphuber! Changes look good, but I think you still need to add `_storable = True` to the `RemoteStashFolderData` class?

I'll open a separate PR for the documentation. I think I'll add a subsection to the "How to run external codes" section.

sphuber:
You are absolutely right. I have fixed it and added a unit test for it. I am hoping that this is what caused the daemon tests to hang.

sphuber:
Alright @mbercx seems like the halting daemon tests were indeed just due to the unstorableness. That is fixed now and all tests pass. @ramirezfranciscof this is ready for review should you have the time.

chrisjsewell:
@mbercx, @ramirezfranciscof can you give a timeline on when you can review this?

mbercx:
> @mbercx, @ramirezfranciscof can you give a timeline on when you can review this?

This is good to go for me, the only issue I still saw was the storability one. I think we're just waiting for @ramirezfranciscof's stamp of approval?

mbercx (approved):

ramirezfranciscof:
Yeah, sorry, I'm half way through the thing but I have another big pressing priority I need to finish right now and is consuming a lot of my attention :sweat_smile:. If you are tired of waiting I would understand, I can send the couple of comments I had (very minor, nothing critical so far I think) and just move on. Otherwise I could try to finish this tomorrow maybe (it really does depend if I manage to break through some specific obstacle I'm dealing with).

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -97,18 +97,39 @@ def validate_calc_job(inputs: Any, ctx: PortNamespace) -> Optional[str]:  # pyli
...
+    if not isinstance(target_base, str) or not os.path.isabs(target_base):
+        return f'`metadata.options.stash.target_base` should be an absolute filepath, got: {target_base}'
+
+    if not isinstance(source_list, list) or any(not isinstance(src, str) or os.path.isabs(src) for src in source_list):
+        return f'`metadata.options.stash.source_list` should be a list of relative filepaths, got: {source_list}'
+
+    try:
+        StashMode(stash_mode)
+    except ValueError:
+        return f'`metadata.options.stash.mode` should be a member of StashMode, got: {stash_mode}'
```
ramirezfranciscof, line 117: Since the user needs to provide the enum, would it be useful to specify where to get it from? So `...should be a member of aiida.common.datastructures.StashMode., got:...`, which would also be consistent with the help message (although not sure why we use `aiida.common.datastructures.StashMode.` instead of something like `aiida.common.datastructures:StashMode`).

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -97,18 +97,39 @@ def validate_calc_job(inputs: Any, ctx: PortNamespace) -> Optional[str]:  # pyli
...
+
+    target_base = stash_options.get('target_base', None)
+    source_list = stash_options.get('source_list', None)
+    stash_mode = stash_options.get('mode', StashMode.COPY.value)
+
+    if not isinstance(target_base, str) or not os.path.isabs(target_base):
+        return f'`metadata.options.stash.target_base` should be an absolute filepath, got: {target_base}'
+
+    if not isinstance(source_list, list) or any(not isinstance(src, str) or os.path.isabs(src) for src in source_list):
+        return f'`metadata.options.stash.source_list` should be a list of relative filepaths, got: {source_list}'
```
ramirezfranciscof, line 112: I'm confused with the specification in the calcjob. There it says `source_list` could be tuple or list, but here it seems to only accept lists. Shouldn't it be list or tuple or is there some subclassing relationship I'm not aware of? In any case, wouldn't this already be checked automatically?

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -233,8 +236,8 @@ async def task_retrieve_job(
 
     :raises: TransportTaskException if after the maximum number of retries the transport task still excepted
     """
-    if node.get_state() == CalcJobState.PARSING:
-        logger.warning(f'CalcJob<{node.pk}> already marked as PARSING, skipping task_retrieve_job')
+    if node.get_state() == CalcJobState.DONE:
+        logger.warning(f'CalcJob<{node.pk}> already marked as DONE, skipping task_retrieve_job')
```
ramirezfranciscof, line 240: I'm not entirely sure why of this change. Maybe you already commented on this on one of the meetings we had and I don't remember, but I couldn't find anything in the text of the PR. Why did you add the `DONE` state? And why did you change it here instead of `PARSING`? What is the purpose of these checks here if the states are supposed to be called in order, just a sanity check?

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -270,15 +273,65 @@ async def do_retrieve():
...
+    try:
+        await exponential_backoff_retry(
+            do_stash,
+            initial_interval,
+            max_attempts,
+            logger=node.logger,
+            ignore_exceptions=plumpy.process_states.Interruption
+        )
+    except plumpy.process_states.Interruption:
+        raise
```
ramirezfranciscof, line 325: What kind of exception is this catching? Can't we re-raise something more informative?

ramirezfranciscof (commented):
Apologies for taking a long time to see this, but all great! I just have a couple of comments and questions to understand this better (since this was already approved I'll just leave this as a comment).

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: is catching what it says its catching 😜 an interruption like pausing or killing, so no

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -271,15 +274,65 @@ async def do_retrieve():
...
+    :raises: TransportTaskException if after the maximum number of retries the transport task still excepted
+    """
+    if node.get_state() == CalcJobState.RETRIEVING:
+        logger.warning(f'calculation<{node.pk}> already marked as RETRIEVING, skipping task_stash_job')
+        return
+
+    initial_interval = get_config_option(RETRY_INTERVAL_OPTION)
+    max_attempts = get_config_option(MAX_ATTEMPTS_OPTION)
+
+    authinfo = node.computer.get_authinfo(node.user)
```
chrisjsewell, line 308:
```suggestion
    authinfo = node.get_authinfo()
```

as per 2acffe70836d47a4e6925d4f90baba596b770095, and will fix the pre-commit

tests/orm/data/test_remote_stash.py
```diff
@@ -0,0 +1,82 @@
...
+        'source_list': ('relative/folder', 'relative/file'),
+    }
+
+    with pytest.raises(TypeError):
+        if argument == 'stash_mode':
+            kwargs['stash_mode'] = value
+        elif argument == 'target_basepath':
+            kwargs['target_basepath'] = value
+        elif argument == 'source_list':
+            kwargs['source_list'] = value
```
chrisjsewell, line 80:
```suggestion
        kwargs[argument] = value
```

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: > What is the purpose of these checks here if the states are supposed to be called in order, just a sanity check?

No @ramirezfranciscof this is if e.g. the daemon was stopped/restarted during the transport tasks, we don't want to repeat a task that has already been finished.

but yes @sphuber it appears here that you have simply replaced `PARSING` with `DONE`?
Is this the case and what was the reason?
If it is a swap then surely `PARSING` should be removed from `CalcJobState` and also the places in the tests that use `PARSING` should also be replaced with `DONE`.
Technically this would also warrant a database migration, which brings into question if it is really worth it 😬

chrisjsewell (changes requested):
from a quick skim, generally fine cheers, but still a few changes to make

aiida/engine/processes/calcjobs/tasks.py
ramirezfranciscof: Right, but my underlying question is why would you catch the exception to just raise with nothing? Why not just leave `plumpy.process_states.Interruption` be the raising error?

sphuber: Just in case this what your question was trying to address: if you catch and then simply call `raise` again, it will raise the original exception. Why is this useful? Because this way you can treat certain exceptions different from others. In this case we want exceptions to be treated a certain way, except the `Interruption`, which should just be bubbled up.

tests/orm/data/test_remote_stash.py
```diff
@@ -0,0 +1,82 @@
...
+    target_basepath = '/absolute/path'
+    source_list = ['relative/folder', 'relative/file']
+
+    data = RemoteStashFolderData(stash_mode, target_basepath, source_list)
+    data.store()
+
+    assert data.is_stored
+    assert data.stash_mode == stash_mode
+    assert data.target_basepath == target_basepath
+    assert data.source_list == source_list
```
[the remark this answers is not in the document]
sphuber: Added it with slight modification of checking attributes are still correct after storing as well

aiida/engine/processes/calcjobs/calcjob.py
sphuber: I have added tuple as a valid type. It is true that the `valid_type` in the port declaration should already catch this, but I still include it here to make the validator more robust. If somebody changes the `valid_type` the conditional in the validator will not fail in the second part when it tries to iterate over the value when it is not fact iterable.

aiida/engine/processes/calcjobs/tasks.py
sphuber: This change was necessary for one of the previous designs where I think the stashing was performed after the parsing and so a new state was necessary. However, in the current design, the stashing (if needed) is done before retrieving and the parsing then follows retrieving as it used to. So indeed now the change is merely replacing `PARSING` with `DONE` so I can revert it. Note that this state attribute is merely for internal purposes and is even removed after the process is terminated, so a migration wouldn't have been necessary.

aiida/engine/processes/calcjobs/calcjob.py
chrisjsewell: note, I think these days, rather than using `isinstance(obj, (list, tuple))` its better to use `isinstance(obj, Sequence)`. But not a deal breaker 😉

sphuber: I used to do this, but I have bitten by this before. The `Sequence` does not only match `list` and `tuple` and sometimes the other matches are not anticipated and can cause problems. So I prefer to be explicit now

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: cheers

> Note that this state attribute is merely for internal purposes and is even removed after the process is terminated, so a migration wouldn't have been necessary.

well I meant in the (very rare case) a migration occurred whilst you still had pending calculations in the `PARSING` state, but yeh a technicality to be sure

aiida/engine/processes/calcjobs/calcjob.py
chrisjsewell: fair

sphuber:
Alright gents, thanks for the review. Should have addressed all of them. If it is ok to be approved, please let me merge myself as I need to adapt the commit message a bit.

chrisjsewell (approved):
good job 👌

aiida/engine/processes/calcjobs/tasks.py
ramirezfranciscof: > if you catch and then simply call `raise` again, it will raise the original exception.

Ahh ok. Yeah, sorry, I did not know this, I thought you had to explicitly provide the exception to raise.

ramirezfranciscof (approved):
Great work, thanks!

PR #4425, opened 2020-10-07
Add `--delete-nodes` flag to verdi group delete

### LINKED ISSUE #4356: Docs: docstring of `verdi group delete` command can be confusing
This is the current docstring of `verdi group delete`:
https://github.com/aiidateam/aiida-core/blob/5e02e164b589f9baec2521980ae006a061c76301/aiida/cmdline/commands/cmd_group.py#L66-L71

The differentiation between a node being removed from a group and a node being deleted is unclear.
Furthermore, the docstring is inconsistent in itself & contains unnecessary repetitions.

Will provide a suggestion for improvement.

### LINKED ISSUE #4358: Add option to `verdi group delete` to delete nodes themselves
So far, `verdi group delete` only acts on groups - there is no option to also delete the nodes inside a group.

It would be useful if there was an option `--delete-nodes` (or similar) that allows to delete both the group as well as the nodes it contains.
As [mentioned](https://github.com/aiidateam/aiida-core/pull/4357#issuecomment-687229723) by @sphuber, this should be followed by the usual prompt listing the number of nodes actually to be deleted (when including the provenance rules).

P.S. This is just a minor nitpick but I actually prefer the `verdi node delete` command without the `--clear` option introduced [here](https://github.com/aiidateam/aiida-core/commit/719d65abd7086d26b028967365dc3a1e12a643b0).
In particular, it is not obvious to me why deleting a group that contains nodes should be treated any different than deleting a group that doesn't contain nodes - after all, it is only the group instance being deleted in both cases (happy to be convinced otherwise). 

Here is what I would suggest:

 * By default (no flags), the command prompts when trying to delete a group. The prompt clearly shows how many nodes the group contains. All groups are treated equal.
 * When using the `-f/--force` flag, no prompt is shown and the group is deleted
 * When using the `--delete-nodes` flag, a prompt is shown that describes how many nodes the group contains as well as how many nodes would be deleted (usually more). The prompt includes some sort of warning sign, similar to what appears in `verdi node delete`
 * When using both `--force` and `--delete-nodes`, no prompt is shown.

mbercx:
Fixes #4358
Fixes #4356.

The current implementation of `verdi group delete` uses two flags:

* `--clear` removes all the nodes from the group before deleting it. Deleting a group that contains nodes fails unless the group is empty or the `--clear` flag is passed to the command.
* `--force` means that the user will not be prompted for confirmation.

This PR removes the `--clear` flag, instead warning the user in case the group is not empty. Moreover, it adds the `--delete-nodes` flag, which allows the user to delete all nodes in a group along with the group itself. The current implementation works as follows (Slightly adjusted from #4358):

* By default (no flags), the command prompts when trying to delete a group. If the group is not empty, the prompt clearly shows how many nodes the group contains with a warning.
* When using the `-f/--force` flag, no warning or prompt is shown and *only* the group is deleted, even if it is not empty.
* When using the `--delete-nodes` flag, two warnings are shown that describe how many nodes the group contains as well as how many nodes would be deleted using the default [traversal rules](https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/provenance/consistency.html#traversal-rules) (usually more). The prompt includes a warning that this operation cannot be undone, similar to what appears in `verdi node delete`.
 * When using both `--force` and --delete-nodes, no prompt is shown, and both the group as well as its nodes are deleted, according to the default [traversal rules](https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/provenance/consistency.html#traversal-rules).

### Questions/TODO

- [ ] Fix/Add tests.
- [ ] Fix/add documentation.
- [ ] Currently the implementation only uses the default traversal rules for deleting the nodes in the group. Should the user also be able to add traversal rules options using `graph_traversal_rules`? Of course, these don't make sense unless the `--delete-nodes` flag is passed, so the command should probably fail in case the user passes any traversal rules options without also passing the `--delete-nodes` flag. What do you think?

### DIFF
--- a/aiida/cmdline/commands/cmd_group.py
+++ b/aiida/cmdline/commands/cmd_group.py
@@ -59,35 +59,45 @@ def group_remove_nodes(group, nodes, clear, force):
 
 @verdi_group.command('delete')
 @arguments.GROUP()
-@options.GROUP_CLEAR(help='Remove all nodes before deleting the group itself.')
+@options.GROUP_DELETE_NODES()
 @options.FORCE()
 @with_dbenv()
-def group_delete(group, clear, force):
+def group_delete(group, delete_nodes, force):
     """Delete a group.
 
-    Note that a group that contains nodes cannot be deleted if it contains any nodes. If you still want to delete the
-    group, use the `-c/--clear` flag to remove the contents before deletion. Note that in any case, the nodes themselves
-    will not actually be deleted from the database.
+    Note that by default deleting a group will not delete the nodes in the group. Use the `--delete-nodes` flag to
+    delete the nodes as well, using the standard traversal rules.
     """
     from aiida import orm
+    from aiida.manage.database.delete.nodes import delete_nodes as delete_nodes_list
+    from aiida.tools.graph.graph_traversers import get_nodes_delete
 
     label = group.label
 
-    if group.count() > 0 and not clear:
-        echo.echo_critical((
-            'Group<{}> contains {} nodes. Pass `--clear` if you want to empty it before deleting the group'.format(
-                label, group.count()
+    if group.count() > 0 and not force:
+        echo.echo_warning(f'Group <{label}> contains {group.count()} nodes.')
+
+    if group.count() > 0 and delete_nodes:
+        group_nodes = [node.pk for node in group.nodes]
+        node_pks_to_delete = get_nodes_delete(group_nodes)['nodes']
+
+        if not force:
+            echo.echo_warning(
+                f'Deleting all nodes in group <{label}> will result in the deletion of {len(node_pks_to_delete)} nodes!'
+                ' THIS CANNOT BE UNDONE!'
             )
-        ))
 
     if not force:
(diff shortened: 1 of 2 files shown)


aiida/cmdline/commands/cmd_group.py
```diff
@@ -59,35 +59,45 @@ def group_remove_nodes(group, nodes, clear, force):
...
 
-    if group.count() > 0 and not clear:
-        echo.echo_critical((
-            'Group<{}> contains {} nodes. Pass `--clear` if you want to empty it before deleting the group'.format(
-                label, group.count()
+    if group.count() > 0 and not force:
+        echo.echo_warning(f'Group <{label}> contains {group.count()} nodes.')
+
+    if group.count() > 0 and delete_nodes:
+        group_nodes = [node.pk for node in group.nodes]
```
chrisjsewell, line 81: this is way too slow 😛 . I'll make a `group_delete`function and you can use that

mbercx:
Superseded by #4578.

PR #4427, opened 2020-10-07
✨ NEW: `structure.data.import` plugin entrypoint

dev-zero:
The use case is that I would like to add an option to import structures from a CP2K input file.
While this could be done in an `aiida-cp2k` cli or an extension of `verdi data`, it would make more sense to add it directly here.
On the other hand it might not be a good idea to open up too many entry points since we would have to maintain those.

### DIFF
--- a/aiida/cmdline/commands/cmd_data/cmd_structure.py
+++ b/aiida/cmdline/commands/cmd_data/cmd_structure.py
@@ -17,6 +17,7 @@ from aiida.cmdline.commands.cmd_data.cmd_export import data_export, export_optio
 from aiida.cmdline.commands.cmd_data.cmd_list import data_list, list_options
 from aiida.cmdline.params import arguments, options, types
 from aiida.cmdline.utils import decorators, echo
+from aiida.cmdline.utils.pluginable import Pluginable
 
 LIST_PROJECT_HEADERS = ['Id', 'Label', 'Formula']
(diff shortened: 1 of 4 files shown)


ltalirz (approved):
thanks @dev-zero - I see no issue with this and would support it (this would anyhow move outside aiida-core together with the materials-science specific classes, if that ever happens).

Could you please bring the PR up to date?

@chrisjsewell any comments?

dev-zero:
@ltalirz thanks a lot for the positive feedback, rebased as requested

ltalirz (approved):
@chrisjsewell I approve this but I'd like one more pair of eyes on this.
Feel free to merge if you agree

chrisjsewell (changes requested):
Looks good cheers, but you add a small bit of documentation of the feature here thanks: https://github.com/aiidateam/aiida-core/blob/develop/docs/source/topics/plugins.rst#aiidacmdline

dev-zero:
> Looks good cheers, but could you add a small bit of documentation of the feature around here thanks: https://github.com/aiidateam/aiida-core/blob/develop/docs/source/topics/plugins.rst#aiidacmdline

Done. Based on the already existing one with a spec, example and example usage.

ltalirz (approved):
thanks @dev-zero !

ltalirz:
@chrisjsewell can you approve & merge?

chrisjsewell (approved):
cheers!

Issue #1320, opened 2018-03-22
Group.add_nodes() function should be able to get a list of 'pk' or 'uuid' as the argument.


yakutovicha:
To export a significant amount of nodes it would be good to NOT load them directly, but rather to provide a list of pk or uuid numbers to the add_nodes() function.

I think rather then doing that: 
```
n = QueryBuilder()
n.append(Node)
flat_list = [item for sublist in n.all() for item in sublist]
some_group.add_nodes(flat_list)
```

it would be much nicer to do: 
```
n = QueryBuilder()
n.append(Node, project=['id'])
flat_list = [item for sublist in n.all() for item in sublist]
some_group.add_nodes(flat_list)
```
Calling some_group.add_nodes() for every loaded node is very slow, so I would prefer to avoid this.

ltalirz:
I don't quite get the need for the sublist here (the desired list of nodes should simply come out of the query by applying the appropriate filters), but I completely agree that it should not be necessary to load a node into memory in order to add it to a group.

PR #4428, opened 2020-10-07
✨ NEW: Add `verdi group move-nodes` command

### LINKED ISSUE #4426: Add `verdi group move-nodes` command
When moving nodes from one group to another, the user now has to execute two commands, e.g.:
```
verdi group remove-nodes -G source_group <node_pks>
verdi group add-nodes -G target_group <node_pks>
```
Instead, it might be useful to have a `verdi group move-nodes` command, which executes both steps in one command:
```
verdi group move-nodes -s source_group -t target_group <node_pks>
```

mbercx:
Fixes #4426.

Add an extra subcommand `move-nodes` to `verdi group` that moves nodes
from a source group to a target group, instead of having to use the
`remove-nodes` and `add-nodes` subcommands in order to achieve this.

Similar to the `remove-nodes` and `add-nodes` commands, the move can be
made without prompts by adding the `--force` flag, and the nodes are
presented as a list of arguments.

Add corresponding test method to the `TestVerdiGroup` class of the
test_group.py suite.


aiida/cmdline/commands/cmd_group.py
```diff
@@ -57,6 +57,25 @@ def group_remove_nodes(group, nodes, clear, force):
...
+    """Move nodes from one group to another."""
+    if not force:
+        click.confirm(
+            f'Do you really want to move {len(nodes)} nodes from Group<{source_group.label}> to '
+            f'Group<{target_group.label}>?',
+            abort=True
+        )
+
+    source_group.remove_nodes(nodes)
+    target_group.add_nodes(nodes)
```
ramirezfranciscof, line 76: Maybe I'm being super paranoid, but if `add_nodes` fails after `remove_nodes` then you just lost track of your nodes, right? Would perhaps be slightly better to do it the other way around and/or `Try` to catch the errors and issue a warning saying the groups might have been messed up (or more specifically if you wrap each on a different `Try`)?

tests/cmdline/commands/test_group.py
```diff
@@ -261,6 +261,44 @@ def test_add_remove_nodes(self):
...
+        result = self.cli_runner.invoke(cmd_group.group_show, ['dummygroup1'])
+        self.assertClickResultNoException(result)
+        self.assertIn('CalculationNode', result.output)
+        self.assertIn(str(node_01.pk), result.output)
+        self.assertIn('Int', result.output)
+        self.assertIn(str(node_02.pk), result.output)
+        self.assertIn('Bool', result.output)
+        self.assertIn(str(node_03.pk), result.output)
+
+        # Move all three nodes to the second dummy group
```
ramirezfranciscof, line 286: The test looks pretty neat in general, but would it perhaps be more versatile to only move two of the nodes? This way you are also checking the integrity of the nodes that are left on the source group.

ramirezfranciscof (changes requested):
It all looks pretty good! I just added two ideas that might be worth considering, let me know what you think.

aiida/cmdline/commands/cmd_group.py
mbercx: Hmm, I don't immediately see a case where `add_nodes` would fail if `remove_nodes` already succeeded, but you're probably right to be paranoid. 😄 

Doing it the other way round seems more dangerous, e.g. if the user wants to move nodes that are not in the `source_group`, the command would add these nodes to the `target_group` and then fail to remove them from the `source_group`. In fact, I just noticed that `remove_nodes(nodes)` shows no error or warning when one of the `nodes` is not in the list of nodes, so we should probably fix that first. Let me open an issue for this.

I could use `try...except` to add a warning, but I wouldn't know which error to except, and if an error is raised, wouldn't the user be aware that the groups might have been messed up?

tests/cmdline/commands/test_group.py
mbercx: Good point! I should probably also add a test that checks if the command correctly fails in case we try to move nodes that are **not** in the `source_group`, once we fix this behaviour.

mbercx:
Thanks for the review, @ramirezfranciscof! I've replied to your comments, will make changes once we have resolved the issues.

aiida/cmdline/commands/cmd_group.py
ramirezfranciscof: > Doing it the other way round seems more dangerous, e.g. if the user wants to move nodes that are not in the source_group, the command would add these nodes to the target_group and then fail to remove them from the source_group. In fact, I just noticed that remove_nodes(nodes) shows no error or warning when one of the nodes is not in the list of nodes, so we should probably fix that first. Let me open an issue for this.

This is a very good point. I'm generally a big proponent of "programs should raise and not just ignore you when you ask them to do something they can't do" (unless you `-force`), but I think not all people share this view.

Maybe we need to do some checks before executing this commands? (i.e. that all nodes are in the source_group and none are in the target group? do we just need to check if they are in the source group? I'm not sure which behaviour we expect from this "mv" now =S)

> I could use try...except to add a warning, but I wouldn't know which error to except, and if an error is raised, wouldn't the user be aware that the groups might have been messed up?

Do we have a way to "append" to the error message? That way you can append something like "WARNING: Nodes have already been added to the target group" that will show at the end of the raise at a failed posterior "delete".

ramirezfranciscof: Perhaps we should ask @giovannipizzi what he would expect of this command? Now I noticed his comment in #4426 mentions something related to this.

mbercx: Another question is at which level we want to do this check. Just at the level of the CLI command? Or maybe at the `GroupBackend` level? Should `group.remove_nodes(nodes)` not at least warn the user in case one of the nodes is not in the group?

ramirezfranciscof: Yes, I know what you mean. Again, I would tipically be in favour of warning or even raising when you are asking to delete stuff that is not there. The only problem I see with this change is that it has the potential of being highly disruptive (meaning: backwards incompatible). I would think that for now we just have to accept that `remove_nodes` works this way and in any case start considering if this is a change we would want for 2.0. For now I would just reduce this discussion to how this "move" command behaves.

giovannipizzi: Mmm, I think that in many cases it's good to 'stop/raise' if there is something really wrong (but maybe in a 'atomic' way, without having performed only half of the job?) but there are cases where this is not ideal (e.g. when a code "crashes" because some convergence loop and you don't even get partial information).
Here, for the internal API, I am not sure if we want to change the behaviour as this is indeed backward incompatible.
But for the CLI at least we should echo if something is unexpected.
Also, different tools behave differently - e.g. `mv` warns but continues, but `git mv` is much more strict.

Suggestion: propose the behaviour at the next AiiDA meeting and see if there is consensus.

mbercx:
Usage:

```console
$ verdi group move-nodes -h
Usage: verdi group move-nodes [OPTIONS] [NODES]...

  Move nodes from one group to another.

Options:
  -s, --source-group GROUP  A single group identified by its ID, UUID or
                            label.  [required]

  -t, --target-group GROUP  A single group identified by its ID, UUID or
                            label.  [required]

  -f, --force               Do not ask for confirmation and skip all checks.
  -h, --help                Show this message and exit.
```

I've created two test groups with the following content:

```console
$ verdi group show test1
-----------------  ----------------
Group label        test1
Group type_string  core
Group description  <no description>
-----------------  ----------------
# Nodes:
  PK  Type    Created
----  ------  --------------
 161  Int     7D:10h:42m ago
 157  Int     7D:10h:44m ago
 159  Int     7D:10h:43m ago
```

```console
$ verdi group show test2
-----------------  ----------------
Group label        test2
Group type_string  core
Group description  <no description>
-----------------  ----------------
# Nodes:
  PK  Type    Created
----  ------  --------------
 159  Int     7D:10h:43m ago
```

When I try to move a node that is not in `test1` to `test2`:

```console
$ verdi group move-nodes -s test1 -t test2 151
Critical: None of the specified nodes are in Group<test1>.
```

When I try to move two nodes, one which is in `test1` and one that isn't:

```console
$ verdi group move-nodes -s test1 -t test2 151 161
Warning: 1 nodes with PK {151} are not in Group<test1>.
Are you sure you want to move 1 nodes from Group<test1> to Group<test2>? [y/N]:
```

When I try to move three nodes, two which is in `test1` and one that isn't, but one of the two that is in `test1` is also in `test2`:

```console
$ verdi group move-nodes -s test1 -t test2 161 151 159
Warning: 1 nodes with PK {151} are not in Group<test1>.
Warning: 1 nodes with PK {159} are already in Group<test2>. These will still be removed from Group<test1>.
Are you sure you want to move 2 nodes from Group<test1> to Group<test2>? [y/N]:
```

When I try to move no nodes at all:

```console
$ verdi group move-nodes -s test1 -t test2
Critical: No nodes identifiers were provided.
```

When I try to move the nodes to the same target group as the source group:

```console
$ verdi group move-nodes -s test1 -t test1 161
Critical: Source and target group are the same: Group<test1>
```

When I come to my senses, and try to move two nodes which _are_ in `test1`, but _aren't_ in `test2`:

```console
$ verdi group move-nodes -s test1 -t test2 161 157
Are you sure you want to move 2 nodes from Group<test1> to Group<test2>? [y/N]:
```

mbercx:
@ramirezfranciscof I've updated the PR in line with with the changes that were made for the `verdi group remove-nodes` command in cba7d6d58baeac67d5860234cf296b60f2d23ef6. I've added examples of the usage above, and described the warnings/failures added in the commit message of 7bef6a3. I will update the tests and add documentation when we're agreed on how it should work.

To come back to your original worry about the `source_group.remove_nodes(nodes)` failing: I think this is either not an issue or at least not an issue that should be solved here. The `verdi group remove-nodes` also calls this command. What if it fails there? In case the command fails here, the issue is the same. However, just glancing at the `remove_nodes` method for the `Group` class and its backends, I think it's written pretty robustly, so that it will only remove any of the nodes if there are no issues.

I've also added @yakutovicha as a reviewer, since he actually requested to be a reviewer for the `move-nodes` command, not the changes I made for the `remove-nodes` command. 😅

yakutovicha (commented):
Thanks, @mbercx, I think you did a great job. I also agree with the logic you've come up with to make the moving process as safe as possible: first, decide what to do and only then do it.

I have to admit though, that unrelated to this PR we have to add some functionality of the groups to simplify the code and to make things faster:

1. In principle, we don't need to load nodes when adding them to the group, but we can't use the `pk` or `UUID` to reference the node. I made an issue some years ago about that (#1320). 
2. We should be able to directly get the pks of the nodes that belong to a group. I made an issue some seconds ago about that (#4755)

aiida/cmdline/commands/cmd_group.py
```diff
@@ -90,6 +90,70 @@ def group_remove_nodes(group, nodes, clear, force):
...
+            query.append(Group, filters={'label': source_label}, tag='group')
+            query.append(Node, with_group='group', filters={'id': {'in': node_pks}}, project='id')
+
+            source_group_node_pks = query.all(flat=True)
+
+            if not source_group_node_pks:
+                echo.echo_critical(f'None of the specified nodes are in {source_class}<{source_label}>.')
+
+            if len(node_pks) > len(source_group_node_pks):
+                absent_node_pks = set(node_pks).difference(set(source_group_node_pks))
```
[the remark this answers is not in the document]
mbercx: I learned from the best ;)

mbercx:
Thanks for the review @yakutovicha! I agree with both your points, and would still add that perhaps there should be a click option that just passes the identifiers without loading the nodes first, removing the need for e.g. this line:

```Python
node_pks = [node.pk for node in nodes]
```

Then we can execute these group commands and their checks without ever having to load the nodes. However, somewhere there should be a check that the identifiers do indeed correspond to nodes.

If it's ok for you, I'll keep the slow version for now, and open an issue regarding improving the speed of these commands that refers to your issues, which we will solve for v2.0.0.

yakutovicha:
> If it's ok for you, I'll keep the slow version for now, and open an issue regarding improving the speed of these commands that refers to your issues, which we will solve for v2.0.0.

Definitely. From my side, the PR is good to be merged. Should I approve or do we want @ramirezfranciscof to have a look as well?

yakutovicha:
provided that the pre-commit issue is fixed, of course :)

mbercx:
> provided that the pre-commit issue is fixed, of course :)

For the record, that pre-commit issue is not from files changed in this PR. :) But I still have to add more tests anyways, so I'll look into adding some pylint disables for those `no-member` issues.

chrisjsewell:
> For the record, that pre-commit issue is not from files changed in this PR.

just investigating

chrisjsewell:
see https://github.com/aiidateam/aiida-core/issues/4756

aiida/cmdline/params/options/main.py
```diff
@@ -204,6 +205,22 @@ def set_log_level(_ctx, _param, value):
     '-G', '--group', 'group', type=types.GroupParamType(), help='A single group identified by its ID, UUID or label.'
 )
 
+SOURCE_GROUP = OverridableOption(
```
sphuber, line 208: Do we really need these to be reusable? They are currently only used once. The docstrings are also not very useful. They are generic, but when used for `move-nodes` the help will be not so useful. Would simply use the following in the command directly:
```python
@options.GROUP(
    '-s',
    '--source-group',
    required=True,
    help='The group whose nodes to move.'
)
@options.GROUP(
    '-t',
    '--target-group',
    required=True,
    help='The group to which the move to.
)
```

aiida/cmdline/commands/cmd_group.py
```diff
@@ -84,6 +84,70 @@ def group_remove_nodes(group, nodes, clear, force):
         group.remove_nodes(nodes)
 
 
+@verdi_group.command('move-nodes')
+@options.SOURCE_GROUP(required=True)
+@options.TARGET_GROUP(required=True)
+@arguments.NODES()
```
sphuber, line 90: This needs a help string. Since arguments can't have one, you should add it manually to the docstring of the command.

aiida/cmdline/commands/cmd_group.py
```diff
@@ -84,6 +84,70 @@ def group_remove_nodes(group, nodes, clear, force):
         group.remove_nodes(nodes)
 
 
+@verdi_group.command('move-nodes')
+@options.SOURCE_GROUP(required=True)
+@options.TARGET_GROUP(required=True)
```
sphuber, line 89: Maybe interesting feature (but not necessary to add here) would be that this group can be created if it doesn't yet exist.

aiida/cmdline/commands/cmd_group.py
```diff
@@ -84,6 +84,70 @@ def group_remove_nodes(group, nodes, clear, force):
...
+            node_pks = [node.pk for node in nodes]
+
+            query = QueryBuilder()
+            query.append(Group, filters={'label': source_label}, tag='group')
+            query.append(Node, with_group='group', filters={'id': {'in': node_pks}}, project='id')
+
+            source_group_node_pks = query.all(flat=True)
+
+            if not source_group_node_pks:
+                echo.echo_critical(f'None of the specified nodes are in {source_class}<{source_label}>.')
```
sphuber, line 118: Your formatting is exactly that of `Group.__str__` so you can just plonk it in. Apply also in other places
```suggestion
                echo.echo_critical(f'None of the specified nodes are in {source_group}.')
```

aiida/cmdline/commands/cmd_group.py
```diff
@@ -84,6 +84,70 @@ def group_remove_nodes(group, nodes, clear, force):
...
+                    f'{target_class}<{target_label}>. These will still be removed from {source_class}<{source_label}>.'
+                )
+
+            click.confirm(
+                f'Are you sure you want to move {len(nodes)} nodes from {source_class}<{source_label}> '
+                f'to {target_class}<{target_label}>?',
+                abort=True
+            )
+        else:
+            echo.echo_critical('No nodes identifiers were provided.')
```
sphuber, line 145: Should this be required? Shouldn't it be possible to move _all_ nodes from source to another group?

sphuber: If you want to keep it, I would invert the logical clause and just have `if not nodes: echo_critical`. Then you can dedent all the other logic making it easier to read.

sphuber (changes requested):
Thanks @mbercx , concept is looking good, just have some comments about the implementation. Also, it would be good to have some better test coverage once we agree on the implementation and features.

aiida/cmdline/commands/cmd_group.py
mbercx: True, but then we should probably also raise a warning for this, since we want to avoid group creation because of a typo. :)

aiida/cmdline/commands/cmd_group.py
mbercx: Right, I probably should have reviewed this old PR myself before highlighting it at the meeting. 😅

aiida/cmdline/params/options/main.py
mbercx: Right, I remembered I tried this myself, but got the following error:

```
verdi group move-nodes --help
  File "/home/mbercx/.virtualenvs/aiida-core/bin/verdi", line 5, in <module>
  File "/home/mbercx/envs/aiida-core/code/aiida-core/aiida/cmdline/commands/__init__.py", line 14, in <module>
    from aiida.cmdline.commands import (
  File "/home/mbercx/envs/aiida-core/code/aiida-core/aiida/cmdline/commands/cmd_group.py", line 88, in <module>
    @options.GROUP(
TypeError: __call__() takes 1 positional argument but 3 were given
```

I guess you can only override the `OverridableOption` args using keyword arguments, and I'm not sure what the keywords are for the option flags?

aiida/cmdline/commands/cmd_group.py
sphuber: In order to not complicate this PR, let's just leave the for another time.

aiida/cmdline/commands/cmd_group.py
mbercx: Fair point, moving all nodes to another group is a valid use case that currently isn't covered. I'm wondering if:

1. This should be the behaviour if no nodes are specified.
2. The user has to add e.g. `-a` / `--all` to move all nodes.

Perhaps [2] is better, because if the user accidentally already added `-f` and then presses enter without specifying a node, it'll be annoying to move the nodes back again.

In case neither nodes nor `--all` are specified, the command should perhaps also just do nothing. That's also what happens when you do e.g. `verdi process show` without specifying any nodes.

aiida/cmdline/params/options/main.py
sphuber: Yeah, or course, forgot about that, but that makes sense. The idea of a reusable option is to guarantee it has the same flag everywhere so allowing to change it doesn't make sense. No matter though, just use the suggestion still but simply use  the plain `@click.option` and add `type=GroupParameterType`.

aiida/cmdline/commands/cmd_group.py
sphuber: Yeah you are probably right that making it explicit with an option is the safest option

aiida/cmdline/params/options/main.py
mbercx: Makes sense, thanks! 👍

aiida/cmdline/commands/cmd_group.py
mbercx: I can fully agree with not introducing scope creep. :D

aiida/cmdline/commands/cmd_group.py
```diff
@@ -84,6 +84,67 @@ def group_remove_nodes(group, nodes, clear, force):
...
+)
+@options.FORCE(help='Do not ask for confirmation and skip all checks.')
+@options.ALL(help='Move all nodes from the source to the target group.')
+@with_dbenv()
+def group_move_nodes(source_group, target_group, force, nodes, all_entries):
+    """Move the specified NODES from one group to another."""
+    from aiida.orm import Group, Node, QueryBuilder
+
+    if not force and source_group.pk == target_group.pk:
+        echo.echo_critical(f'Source and target group are the same: {source_group}.')
```
sphuber, line 101: I wonder if skipping this check ever makes sense, even if `force` is used. It would be a no-op essentially wouldn't it? Can't see this ever making sense so best just raise.

aiida/cmdline/commands/cmd_group.py
```diff
@@ -84,6 +84,67 @@ def group_remove_nodes(group, nodes, clear, force):
...
+
+    if not nodes:
+        if not all_entries:
+            return
```
sphuber, line 105: I know you said you are replicating the behavior of `verdi process show` but here I think a warning saying that neither was specified would be a lot more useful. It is not unthinkable that people think just specifying source and target automatically implies to move all, so they would be confused if no output was produced and nothing was done.

aiida/cmdline/commands/cmd_group.py
```diff
@@ -84,6 +84,67 @@ def group_remove_nodes(group, nodes, clear, force):
...
+        nodes = list(source_group.nodes)
```
sphuber, line 106: I would refactor this slightly such that the same warning is printed in case `target` already contains some of the nodes. So essentially take the `target` check out of the next logical block and make it common.

aiida/cmdline/commands/cmd_group.py
```diff
@@ -84,6 +84,67 @@ def group_remove_nodes(group, nodes, clear, force):
...
+            echo.echo_critical(f'None of the specified nodes are in {source_group}.')
+
+        if len(node_pks) > len(source_group_node_pks):
+            absent_node_pks = set(node_pks).difference(set(source_group_node_pks))
+            echo.echo_warning(f'{len(absent_node_pks)} nodes with PK {absent_node_pks} are not in {source_group}.')
+            nodes = [node for node in nodes if node.pk in source_group_node_pks]
+
+        query = QueryBuilder()
+        query.append(Group, filters={'id': target_group.pk}, tag='group')
+        query.append(Node, with_group='group', filters={'id': {'in': node_pks}}, project='id')
```
sphuber, line 129: You should probably update `node_pks` at this point to be just those pks that are actually in `source`. Now you are still using the list specified by the user, but some of which may not actually exist in `source` so the warning at line 134 would be incorrect, wouldn't it?

Edit: I see that you write explicitly in the warning ".... if present". Still wouldn't it just be clearer if you only do this for nodes that are actually in `source`?

sphuber (changes requested):
Thanks @mbercx some final requests to the behavior.

aiida/cmdline/commands/cmd_group.py
mbercx: Yeah, I agree 👍

mbercx:
Thanks for the comments @sphuber! I agree with the revised logic, I think originally I was for some reason very focused on `--force` not returning any output for some reason. 🤷 

Here is an updated example of the usage:

For two groups:

```console
$ verdi group list -C
Report: To show groups of all types, use the `-a/--all` option.
  PK  Label        Type string    User                Node count
----  -----------  -------------  ----------------  ------------
   2  dummygroup1  core           mbercx@gmail.com             2
   3  dummygroup2  core           mbercx@gmail.com             2
```

Which each contain two nodes:

```console
$ verdi group show 2
-----------------  ----------------
Group label        dummygroup1
Group type_string  core
Group description  <no description>
-----------------  ----------------
# Nodes:
  PK  Type             Created
----  ---------------  --------------
 279  CalculationNode  3h:30m:37s ago
 280  Int              3h:30m:37s ago

$ verdi group show 3
-----------------  ----------------
Group label        dummygroup2
Group type_string  core
Group description  <no description>
-----------------  ----------------
# Nodes:
  PK  Type    Created
----  ------  --------------
 280  Int     3h:30m:40s ago
 281  Bool    3h:30m:40s ago
```

Trying to move nodes to the same group:

```console
$ verdi group move-nodes -s dummygroup1 -t dummygroup1
Critical: Source and target group are the same: Group<dummygroup1>.
```

Not specifying any nodes or `--all`:

```console
$ verdi group move-nodes -s dummygroup1 -t dummygroup2
Critical: Neither NODES or the `-a, --all` option was specified.
```

No nodes you want to move aren't in the source group:

```console
$ verdi group move-nodes -s dummygroup1 -t dummygroup2 281
Critical: None of the specified nodes are in Group<dummygroup1>.
```

Some nodes you want to move aren't in the source group:

```console
$ verdi group move-nodes -s dummygroup1 -t dummygroup2 1 2 3 279
Warning: 3 nodes with PK {1, 2, 3} are not in Group<dummygroup1>.
Are you sure you want to move 1 nodes from Group<dummygroup1> to Group<dummygroup2>? [y/N]:
```

The node you want to move is in the source group, but also already in the target group:

```console
$ verdi group move-nodes -s dummygroup1 -t dummygroup2 280
Warning: 1 nodes with PK {280} are already in Group<dummygroup2>. These will still be removed from Group<dummygroup1>.
Are you sure you want to move 1 nodes from Group<dummygroup1> to Group<dummygroup2>? [y/N]:
```

Mixing it up:

```console
$ verdi group move-nodes -s dummygroup1 -t dummygroup2 1 2 3 280 279
Warning: 3 nodes with PK {1, 2, 3} are not in Group<dummygroup1>.
Warning: 1 nodes with PK {280} are already in Group<dummygroup2>. These will still be removed from Group<dummygroup1>.
Are you sure you want to move 2 nodes from Group<dummygroup1> to Group<dummygroup2>? [y/N]:
```

sphuber (approved):

PR #4431, opened 2020-10-08
Non-zero return code upon error in `verdi import`

### LINKED ISSUE #4430: Return error code != 0 for errors during `verdi import`
### Describe the bug

If a migration of an export archive fails, the return code is still `0`.

### Steps to reproduce

Use `verdi export migrate` with an invalid export archive.
Check return code in the terminal afterwards (`$ echo $?`).

### Expected behavior

For any reporting within `verdi`, if `ERROR:` is printed/returned, the return code should _not_ be `0`.

### Your environment

- Operating system: Ubuntu 18.04 LTS in WSL 2.0 on Windows 10.
- Python version: 3.7
- aiida-core version: 1.4.2

### Additional context

This was found by @asle85 during automatic migration of archives for Materials Cloud via Ansible playbooks.

CasperWA:
Fixes #4430

`verdi import` will now exit with return code `1` if an error occurs at
any point during the import/migration.

A `--fail-fast` flag is added to `verdi import`, which will ensure an
immediate exit if an error occurs during the import/migration of a list
of archives (either local files or URLs).

---

Missing:
- [ ] Tests

### DIFF
--- a/aiida/cmdline/commands/cmd_import.py
+++ b/aiida/cmdline/commands/cmd_import.py
@@ -8,8 +8,9 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """`verdi import` command."""
-# pylint: disable=broad-except,too-many-arguments,too-many-locals,too-many-branches
+# pylint: disable=broad-except,too-many-arguments,too-many-locals,too-many-branches,global-statement
 from enum import Enum
+import sys
 import traceback
 import urllib.request
 
@@ -24,6 +25,8 @@ EXTRAS_MODE_EXISTING = ['keep_existing', 'update_existing', 'mirror', 'none', 'a
 EXTRAS_MODE_NEW = ['import', 'none']
 COMMENT_MODE = ['newest', 'overwrite']
 
+RETURN_NON_ZERO = False
+
 
 class ExtrasImportCode(Enum):
     """Exit codes for the verdi command line."""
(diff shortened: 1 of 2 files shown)


CasperWA:
@ltalirz Feel free to go through this PR already, since I consider my "fix" complete. Only tests for it is missing.

aiida/cmdline/commands/cmd_import.py
```diff
@@ -47,33 +50,44 @@ def _echo_error(  # pylint: disable=unused-argument
     :type more_archives: bool
     :param raised_exception: Exception raised during error.
     :type raised_exception: `Exception`
+    :param fail_fast: Exit the import/migration immediately if an error occurs.
+    :type fail_fast: bool
     """
     from aiida.tools.importexport import close_progress_bar, IMPORT_LOGGER
 
+    global RETURN_NON_ZERO
```
giovannipizzi, line 58: Aaahhh a global variable!! :-D

Do you really need it? Essentially you want to check if the _echo_error was ever called?
Then just do it in the caller, no? Any call to _echo_error should return a boolean flag 'return_non_zero' and you pass it up to the relevant function?

CasperWA: Yeah. It was the quickest (and dirtiest) way to do it 😅 

I'll take a look at rewriting it.

CasperWA: Alright, I've rewritten it, but the code is now much more unclear I think.
The issue here is that it should retain the information if it is set to `True`, but still allow for trying to import more archives, if supplied.

greschd: Agree it's now a bit more involved, but the global variable would probably get you into trouble when testing (because it's run more than once from the same interpreter instance).

CasperWA: > Agree it's now a bit more involved, but the global variable would probably get you into trouble when testing (because it's run more than once from the same interpreter instance).

... which is why I shouldn't wait with writing up the tests I guess 😅

CasperWA:
It seems this issue has already been resolved #4532.

PR #4434, opened 2020-10-10
node.open: more useful warnings

ltalirz:
aiida-core started warning when methods like `Node.read` and `Node.del`
are used outside a context manager.
The warning, however, did not include any indication of where the
problematic function call was made (nor did it include instructions on
how to fix it).

Furthermore, this PR fixes an overlooked use of `Node.read` in
aiida-core that did not rely on the context manager.

### DIFF
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -11,6 +11,7 @@
 """Package for node ORM classes."""
 import importlib
 import warnings
+import traceback
 
 from aiida.common import exceptions
 from aiida.common.escaping import sql_string_match
@@ -47,9 +48,18 @@ class WarnWhenNotEntered:
     def _warn_if_not_entered(self, method):
         """Fire a warning if the object wrapper has not yet been entered."""
         if not self._was_entered:
-            msg = '`{}` used without context manager for {}. This will raise starting from `aiida-core==2.0.0`'.format(
-                method, self._name
-            )
+            msg = f'\nThe method `{method}` was called on the return value of `{self._name}.open()`' + \
+                    ' outside of a context manager.\n' + \
+                  'Please wrap this call inside `with <node instance>.open(): ...` to silence this warning. ' + \
+                  'This will raise an exception, starting from `aiida-core==2.0.0`.\n'
+
+            try:
+                caller = traceback.format_stack()[-3]
+            except Exception:  # pylint: disable=broad-except
+                msg += 'Could not determine the line of code responsible for triggering this warning.'
+            else:
+                msg += f'The offending call comes from:\n{caller}'
+
             warnings.warn(msg, AiidaDeprecationWarning)  # pylint: disable=no-member
 
     def __enter__(self):
--- a/aiida/orm/nodes/data/cif.py
+++ b/aiida/orm/nodes/data/cif.py
@@ -417,7 +417,10 @@ class CifData(SinglefileData):
         """
         if not kwargs and self._ase:
             return self.ase
-        return CifData.read_cif(self.open(), **kwargs)
+        with self.open() as handle:
+            cif = CifData.read_cif(handle, **kwargs)
+
+        return cif
 
     def set_ase(self, aseatoms):
         """


ltalirz:
Example warning:
```
In [1]: n = load_node(123)

In [2]: n.open('aiida.out').read()[1]
/Users/leopold/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/orm/nodes/node.py:64: AiidaDeprecationWarning:
`Node.read` used outside context manager for <FolderData: uuid: 2146a20c-193a-4e76-b492-f868a8cf31a0 (pk: 123)>.
Please use `with <node>.open(): ...` instead. This will raise starting from `aiida-core==2.0.0`
File <ipython-input-2-6452f92039da>, line 1 in module <module>  File "<ipython-input-2-6452f92039da>", line 1, in <module>
    n.open('aiida.out').read()[1]

  warnings.warn(msg, AiidaDeprecationWarning)  # pylint: disable=no-member
/Users/leopold/Personal/Postdoc-MARVEL/repos/aiida/aiida_rmq/aiida/orm/nodes/node.py:64: AiidaDeprecationWarning:
`Node.del` used outside context manager for <FolderData: uuid: 2146a20c-193a-4e76-b492-f868a8cf31a0 (pk: 123)>.
Please use `with <node>.open(): ...` instead. This will raise starting from `aiida-core==2.0.0`
File <ipython-input-2-6452f92039da>, line 1 in module <module>  File "<ipython-input-2-6452f92039da>", line 1, in <module>
    n.open('aiida.out').read()[1]

  warnings.warn(msg, AiidaDeprecationWarning)  # pylint: disable=no-member
```

It's a bit ugly that `traceback.StackSummary.from_list(caller_traceback).format()` adds the newline.
I could also use `inspect.getframeinfo` and then essentially recreate 
```
File <ipython-input-2-6452f92039da>, line 1 in module <module>  File "<ipython-input-2-6452f92039da>", line 1, in <module>
    n.open('aiida.out').read()[1]
```
from this but that also seemed like overkill.
Let me know in case you know a better way (maybe I'm overlooking some obvious built-in function here)

aiida/orm/nodes/node.py
```diff
@@ -47,9 +49,16 @@ def __init__(self, fileobj, name):
     def _warn_if_not_entered(self, method):
         """Fire a warning if the object wrapper has not yet been entered."""
         if not self._was_entered:
-            msg = '`{}` used without context manager for {}. This will raise starting from `aiida-core==2.0.0`'.format(
-                method, self._name
-            )
+            msg = f'\n`Node.{method}` used outside context manager for {self._name}.\n' + \
+                  'Please use `with <node>.open(): ...` instead. ' + \
+                  'This will raise starting from `aiida-core==2.0.0`\n'
+            caller_frame = inspect.currentframe().f_back.f_back
```
greschd, line 55: I wonder if there is a possibility that this code fails - maybe we can just wrap it in a try / except such that it fails gracefully?

aiida/orm/nodes/node.py
```diff
@@ -47,9 +49,16 @@ def __init__(self, fileobj, name):
     def _warn_if_not_entered(self, method):
         """Fire a warning if the object wrapper has not yet been entered."""
         if not self._was_entered:
-            msg = '`{}` used without context manager for {}. This will raise starting from `aiida-core==2.0.0`'.format(
-                method, self._name
-            )
+            msg = f'\n`Node.{method}` used outside context manager for {self._name}.\n' + \
```
sphuber, line 52: The `Node.{method}` is not actually correct and may be misleading. The `method` here is what is actually called on the *return value* of `Node.open`. Normally this is a filelike object, but now it is this instance of `WarnWhenNotEntered` that we added. You can see this when this object is garbage collected, because it will then say `Node.del used outside ..`  but `Node` doesn't have a `del` method. Same goes for `read` though.

Consider following example:
```
handle = node.open()
content = handle.read()
```
It is when `read` is called on `handle` that we get the warning, not on the node. Likewise, it is `del handle` that is called when it is garbage collected that incurs the second warning.

This shows another potential problem, with the example above, the line of code that you will include in your warning will be `content = handle.read()` and people may be confused by you to tell them that this should be `with node.open()`. Note that in this case, the fix may be obvious, because they will go to the line mentioned, and see the `node.open()` call straight above it. However, they are not necessarily very close. They could have passed the handle to some function, so then it may not be clear at all where the actual problem is.

aiida/orm/nodes/node.py
```diff
@@ -47,9 +49,16 @@ def __init__(self, fileobj, name):
...
+                  'Please use `with <node>.open(): ...` instead. ' + \
+                  'This will raise starting from `aiida-core==2.0.0`\n'
+            caller_frame = inspect.currentframe().f_back.f_back
+            caller_tb = inspect.getframeinfo(caller_frame)
+            msg += f'File {caller_tb.filename}, line {caller_tb.lineno} in module {caller_tb.function}'
```
sphuber, line 57: Not sure why, but I see the line with `File {caller_tb.filename} ..` twice
```
  /home/sph/code/aiida/env/dev/aiida-core/aiida/orm/nodes/node.py:62: AiidaDeprecationWarning: 
  `Node.read` used outside context manager for <ForceConstantsData: uuid: 6444b4db-4c69-484d-87d1-5ab6d38d8e46 (unstored)>.
  Please use `with <node>.open(): ...` instead. This will raise starting from `aiida-core==2.0.0`
  File /home/sph/code/aiida/env/dev/aiida-quantumespresso/aiida_quantumespresso/data/force_constants.py, line 22 in module set_file  File "/home/sph/code/aiida/env/dev/aiida-quantumespresso/aiida_quantumespresso/data/force_constants.py", line 22, in set_file
      content = handle.read().splitlines()
  
    warnings.warn(msg, AiidaDeprecationWarning)  # pylint: disable=no-member
```
Any idea why?

sphuber:
@ltalirz I think it would be good to try and get this into the v1.5.0 release as well. Do you have the time to still take a look at this?

ltalirz:
Yes

ltalirz:
Could you please check https://github.com/aiidateam/aiida-core/pull/4393 and https://github.com/aiidateam/aiida-core/pull/4511 ?

aiida/orm/nodes/node.py
ltalirz: Done (for the case when `f_back` returns `None`

aiida/orm/nodes/node.py
ltalirz: > The `Node.{method}` is not actually correct and may be misleading. The `method` here is what is actually called on the _return value_ of `Node.open`.

Right, I've now tried to rephrase the message to make this clearer. Further suggestions welcome.

> the line of code that you will include in your warning will be `content = handle.read()` and people may be confused by you to tell them that this should be `with node.open()`

You are very welcome to suggest a better way.
In my view, this warning already improves the situation enormously, since it shows which call caused the warning rather than where the warning is defined.
For me this is good enough.

aiida/orm/nodes/node.py
ltalirz: Sorry, this didn't show up for me since I was working inside ipython, and so I thought I had to recreate this line manually.
I've removed it now; should be fixed.

aiida/orm/nodes/node.py
```diff
@@ -47,9 +49,18 @@ def __init__(self, fileobj, name):
...
+            try:
+                caller_frame = inspect.currentframe().f_back.f_back
+            except AttributeError:
+                warnings.warn(msg, AiidaDeprecationWarning)
+                return
+
+            caller_traceback = traceback.extract_stack(caller_frame, limit=1)
+            for item in traceback.StackSummary.from_list(caller_traceback).format():
+                msg += item
             warnings.warn(msg, AiidaDeprecationWarning)  # pylint: disable=no-member
```
sphuber, line 64: I think putting the one call in try/except helps but is potentially not enough. We really want to avoid exceptions occurring in this handler. I also think the code to extract the relevant line of code can be much simpler. I propose the following for the entire method, which has updated warning message and stack extraction:
```
        if not self._was_entered:
            msg = f'\nThe method `{method}` was called on the return value of `{self._name}.open()` outside of a context manager.\n' + \
                  'Please wrap this call inside `with <node instance>.open(): ...` to silence this warning. ' + \
                  'This will raise an exception, starting from `aiida-core==2.0.0`.\n'

            try:
                caller = traceback.format_stack()[-3]
            except Exception:  # pylint: disable=broad-except
                msg += 'Could not determine the line of code responsible for triggering this warning.'
            else:
                msg += f'The offending call comes from:\n{caller}'

            warnings.warn(msg, AiidaDeprecationWarning)  # pylint: disable=no-member
```

ltalirz: sounds good to me!

sphuber (approved):
Thanks @ltalirz

PR #4436, opened 2020-10-11
Fix `UnboundLocalError` in `aiida.cmdline.utils.edit_multiline_template`

### LINKED ISSUE #4435: `aiida.cmdline.utils.edit_multiline_template` can raise `UnboundLocalError`

sphuber:
Fixes #4435 

If `click.edit` returns a falsy value, the following conditional would
be skipped and the `value` variable would be undefined causing an
`UnboundLocalError` to be raised. This bug was reported by @blokhin but
the exact conditions under which it occurred are not clear.


qiaojunfeng:
I find one occurrence of this error, that is when setup a computer: when verdi asks for `PREPEND_TEXT`, and if I type `:q` for `vi`, the `click.edit` will return `None`, leading to this `UnboundLocalError`.

sphuber:
@chrisjsewell and @ramirezfranciscof this has been open for a while and needs to go in the next release. Would be good if any you two can review it.

chrisjsewell (approved):
Yep seems pretty trivial to me 👍

PR #4446, opened 2020-10-16
👌 Improve archive export  code

chrisjsewell:
This PR is part of a set that aims to decouple the generation of data to export, from the actual writing of the archive; as a prelude to introducing a new archive format (see also #4448).

This particular PR does not introduce any breaking changes, it:

- splits the amazingly complex `export_tree` function into a number of separate, self-contained, functions
- adds mypy compliant typing (and adds the file to the pre-commit mypy list)
- fixes a bug whereby, in python 3.8, `logging.disable(level=` has changed to `logging.disable(lvl=`


chrisjsewell:
Oh and it is also actually compliant with: https://github.com/PyCQA/isort (which wouldn't be a bad idea t add to our pre-commit)

chrisjsewell:
the docker fail is obviously nothing to do with this PR 😒

sphuber:
Just a question about timeline here. I am wondering if we should hold of with this after we merged my work on the new repository. There I also had to touch this code to make it compatible and also added migrations for existing export archives. Maybe we should quickly discuss how we are going to approach this in order to prevent many complicated merge conflicts and double work

chrisjsewell:
yeh we can discuss in the meeting at 2pm 👍

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -9,79 +9,108 @@
...
+)
 
 from aiida import get_version, orm
 from aiida.common import json
 from aiida.common.exceptions import LicensingException
-from aiida.common.folders import RepositoryFolder, SandboxFolder, Folder
+from aiida.common.folders import Folder, RepositoryFolder, SandboxFolder
 from aiida.common.lang import type_check
-from aiida.common.log import override_log_formatter, LOG_LEVEL_REPORT
+from aiida.common.log import LOG_LEVEL_REPORT, override_log_formatter
```
ltalirz, line 36: hehe, wow - things are even ordered alphabetically now?
I wonder at which points the import lines also have to be in alphabetic order ;-)

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -9,79 +9,108 @@
...
     """Export AiiDA data
 
     .. deprecated:: 1.2.1
         Support for the parameters `what` and `outfile` will be removed in `v2.0.0`.
         Please use `entities` and `filename` instead, respectively.
 
-    :param entities: a list of entity instances; they can belong to different models/entities.
-    :type entities: list
+    :param entities: a list of entity instances;
+        they can belong to different models/entities.
```
ltalirz, line 88: Am I wrong or does mypy somehow apply a shorter line length than the one we have configured?

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -9,79 +9,108 @@
...
-    :type silent: bool
 
-    :param use_compression: Whether or not to compress the archive file (only valid for the zip file format).
-    :type use_compression: bool
+    :param use_compression: Whether or not to compress the archive file
+        (only valid for the zip file format).
 
-    :param allowed_licenses: List or function. If a list, then checks whether all licenses of Data nodes are in the
-        list. If a function, then calls function for licenses of Data nodes expecting True if license is allowed, False
+    :param allowed_licenses: List or function.
```
ltalirz, line 105: if this is an allowed kwarg, then this doesn't really match with calling the kwargs `traversal_rules`
should this be made an explicit parameter?

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -9,79 +9,108 @@
...
+        If a list, then checks whether all licenses of Data nodes are in the list. If a function,
+        then calls function for licenses of Data nodes expecting True if license is allowed, False
         otherwise.
     :type allowed_licenses: list
 
-    :param forbidden_licenses: List or function. If a list, then checks whether all licenses of Data nodes are in the
-        list. If a function, then calls function for licenses of Data nodes expecting True if license is allowed, False
+    :param forbidden_licenses: List or function. If a list,
```
ltalirz, line 111: same here

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -93,12 +122,13 @@ def export(
         Default: True, *include* logs in export.
     :type include_logs: bool
 
-    :param kwargs: graph traversal rules. See :const:`aiida.common.links.GraphTraversalRules` what rule names
-        are toggleable and what the defaults are.
+    :param kwargs: graph traversal rules. See :const:`aiida.common.links.GraphTraversalRules`
```
ltalirz, line 125: should this then be named `traversal_rules`?

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -108,28 +138,38 @@ def export(
...
-    entities = deprecated_parameters(
-        old={
-            'name': 'what',
-            'value': kwargs.pop('what', None)
-        },
-        new={
-            'name': 'entities',
-            'value': entities
-        },
+    entities = cast(
```
ltalirz, line 141: just wondering - wouldn't it be easier to annotate the `deprecated_parameters` function instead?

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -167,113 +211,142 @@ def export(
...
+        logging.disable(logging.NOTSET)
 
 
-def export_zip(entities=None, filename=None, use_compression=True, **kwargs):
+def export_zip(
+    entities: Optional[Iterable[Any]] = None,
+    filename: Optional[str] = None,
+    use_compression: bool = True,
+    **traversal_rules: bool,
+) -> Tuple[float, ...]:
```
ltalirz, line 232: Couldn't that be 
```suggestion
) -> Tuple[float, float]:
```

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -167,113 +211,142 @@ def export(
...
 
     return (time_start, time_end)
 
 
-def export_tar(entities=None, filename=None, **kwargs):
+def export_tar(
+    entities: Optional[Iterable[Any]] = None,
+    filename: Optional[str] = None,
+    **traversal_rules: bool,
+) -> Tuple[float, ...]:
```
ltalirz, line 297: At least to me, this function doesn't seem to return anything (?)

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -420,13 +596,23 @@ def export_tree(
...
+    return entities_starting_set, given_node_entry_ids
+
+
+def collect_export_nodes(given_node_entry_ids: Set[int], silent: bool,
+                         **traversal_rules: bool) -> Tuple[Set[int], Dict[int, str], List[dict], Dict[str, bool]]:
+    """Iteratively explore the AiiDA graph to find further nodes that should also be exported
+
+    At the same time, we will create the links_uuid list of dicts to be exported
+
+    :param silent: suppress console prints and progress bar.
```
ltalirz, line 608: given_node_entry_ids and traversal_rules missing

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -420,13 +596,23 @@ def export_tree(
...
+    """
+    from aiida.tools.graph.graph_traversers import get_nodes_export
```
ltalirz, line 610: I guess this should go to the top of the file? wondering why pylint does not complain...

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -448,29 +634,83 @@ def export_tree(
...
+        graph_traversal_rules,
+    )
+
+
+def check_node_licenses(
+    node_ids_to_be_exported: Set[int],
+    allowed_licenses: Optional[Union[list, Callable]],
+    forbidden_licenses: Optional[Union[list, Callable]],
+) -> None:
+    # TODO (Spyros) To see better! Especially for functional licenses
```
ltalirz, line 655: no docstring... I see pylint is very lenient in this file... ;-)

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -647,16 +904,21 @@ def export_tree(
...
+    all_node_uuids: Set[str],
+    export_data: Dict[str, Dict[int, dict]],
+    node_attributes: Dict[str, dict],
+    node_extras: Dict[str, dict],
+    groups_uuid: Dict[str, List[str]],
+    links_uuid: List[dict],
+    silent: bool,
+) -> None:
+    """Store data to the archive."""
     ######################################
```
ltalirz, line 922: delete

ltalirz (changes requested):
thanks @chrisjsewell 

I have a couple of minor comments/questions - the most important one concerns perhaps the change in line length/formatting, which makes reviewing a bit more tedious then necessary.

Feel free to ignore comments on functions that you are removing in #4448

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: yep I run it with this in the pre-commit 😄 

```yaml
- repo: https://github.com/timothycrosley/isort
  rev: 5.5.3
  hooks:
  - id: isort
```

chrisjsewell: does it all for you!

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: I don't think mypy dictates a line length, I think I just changed this when I was writing the code in https://github.com/aiidateam/aiida-archive, which dictates a line length of 88 (via black) which FYI does a lot better job of the formatting, especially for things like multi-line dictionaries

ltalirz: Ok, but please don't do this in your PRs on aiida-core, it is not reviewer friendly.

If you want to propose this as a repo-wide policy, that is another thing, but this needs to be discussed.

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: I'm not sure what you mean by this? the `**` syntax gathers all kwargs that are not already directly specified, so it would not include `allowed_licenses`

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: oh yeh well I wouldn't normally be writing the code in a different repo, its just that I only decided this should be upstreamed after I had written it all

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: typing of variables was only introduced in python 3.6. I can't remember are we allowed to drop 3.5 now?

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: no because `export_tar` returns `Tuple[float, float, float, float]`, so you get an error later on

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: see below...

chrisjsewell: ah I can't comment on it, but its line 357

```python
return (time_export_start, time_export_end, time_compress_start, time_compress_end)
```

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: Well I didn't change the pylint rules that were already there 😉

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: note already fixed in #4448

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: note already fixed in #4448

aiida/tools/importexport/dbexport/__init__.py
ltalirz: Right, but `allowed_licenses` is not listed among the function arguments in the code above ;-)

aiida/tools/importexport/dbexport/__init__.py
ltalirz: yes, see https://github.com/aiidateam/aiida-core/pull/4386
next aiida-core release will be 1.5.0 without python 3.5 support

aiida/tools/importexport/dbexport/__init__.py
ltalirz: ok!

aiida/tools/importexport/dbexport/__init__.py
ltalirz: Ok!

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: actually I don't think pylint checks for docstrings anyway, because there nothing disabling it

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: fixed in #4448

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: fixed in #4448

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: ah actually you still need to cast here, because the function signature says that `entities` can also be `None`, then this cast "updates" that to say it is now only an iterable

chrisjsewell: also `deprecated_functions` can return multiple different types

ltalirz:
Closing, since #4448 will be made backwards-compatible and includes these changes.

PR #4458, opened 2020-10-19
Docs: add "how to share data"

ltalirz:
fixes #4432 
fixes #3998 
partly addresses #4450

The "how to share data" section includes instructions both for dealing with AiiDA archives (e.g. for publishing AiiDA graphs alongside your publication) and for using the AiiDA REST API.

Issue #4120, opened 2020-05-26
Check for missing and add useful cross-links.


csadorf:
After merging all sections, we should add cross-links that were previously hard to identify.

csadorf:
@sphuber I was actually trying to imply that this issue cannot be addressed before the other PRs are merged, not that this is blocking.

sphuber:
Ah ok, but that is not what the `pr/blocked` label is intended for. It is to label PRs that are blocked from being merged, as written in the label description.

sphuber:
Did you still want to do this before the merge into `develop` @csadorf ?

csadorf:
@csadorf No, it will have to wait.

chrisjsewell:
Can this issue now be closed?

PR #4463, opened 2020-10-19
Docs: Add "Sharing data" to "How to work with data"

### LINKED ISSUE #3998: Docs: How to share data

CasperWA:
Closes #3998

Do not move over the technical aspects, i.e., the sections discussing the export archive format and the JSON files.

This is now related to #4458.


ramirezfranciscof:
Thanks for the work @CasperWA !

For the reviewer: here is a checklist of some of the relevant stuff to verify for your convenience (use it as you please).
- [ ] Pull the PR and check that it can locally compile correctly.
- [ ] There are no format-display issues (unrecognized code blocks, broken links, etc.).
- [ ] All links that must remain temporarily unresolved have been reported in issue #4120.

When checking the formatting, make sure that the [guidelines](https://github.com/aiidateam/aiida-core/wiki/Writing-documentation) are being followed. The most relevant ones:
- There is only one sentence per line and otherwise no manual line wrapping.
- All headers are in sentence-case ('Entry points' instead of 'Entry Points')
- Paragraphs are separated by a single line (not more).
- Code blocks: 'bash' and 'python' for scripts, 'console' and 'ipython' for interactive sessions and verdi shell ('$' for normal bash user, '#' for root).

ltalirz:
After discussing with @CasperWA , I will pull in these changes into a new top level "How to share data" section

giovannipizzi:
As a comment, I would be very careful of only moving part of an old page - i.e., if you really don't want to move the internals, then at least make 100% sure it's clear what part of the old page is still not migrated, or we'll just "lose" it.
If it's easy to just copy-paste and adapt the style to the new docs, I'd suggest to do it at the same time, if possible

giovannipizzi:
Sorry - after discussion with @sphuber I realised the old content is still in the source code, unlinked. The important thing to remember is to delete the corresponding text from the old docs when we add the new docs, so at the end we can easily see what has still been left out from the new docs.
Also,  I realised that the parts on the `internals` have been done by Francisco in #4467

Issue #4450, opened 2020-10-16
Docs: remove old content from 'import_export/'


ramirezfranciscof:
This is a structure that was left over from the old documentation and needs to be removed, but first we need to check if all the content has been re-assigned in the new documentation. The files considered are:

- [ ] [main.rst](https://aiida-core.readthedocs.io/en/v1.1.0/import_export/main.html)
- [ ] [dbimporters/cod.rst](https://aiida-core.readthedocs.io/en/v1.1.0/import_export/external_dbs.html)
- [ ] [dbimporters/icsd.rst](https://aiida-core.readthedocs.io/en/v1.1.0/import_export/external_dbs.html)
- [ ] [external_dbs.rst](https://aiida-core.readthedocs.io/en/v1.1.0/import_export/external_dbs.html)

### Proposition:

The first "main" page (about the export format) we already decided to move to "Internals: Data storage (Export archive)" (see [here](https://github.com/aiidateam/aiida-core/issues/4037)). The other three I think would be better suited for a "howto/import_from_online_dbs" section.

mbercx:
> The first "main" page (about the export format) we already decided to move to "Internals: Data storage (Export archive)" (see [here](https://github.com/aiidateam/aiida-core/issues/4037)). 

Perhaps how to export data also belongs in the "How to work with data" section? There is still a section marked as TODO on "How to share data" (see also #3998).

> The other three I think would be better suited for a "howto/import_from_online_dbs" section.

Should this also be a part of "How to work with data" -> ["Importing data"](https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/data.html#importing-data), under a subsection "Importing from external databases"?

sphuber:
Section "Export File Format" should go to the "Internals: Data storage" section, for which there already is issue #4037
The other sections "Export", "Import", "Export Archive Migration" and "Inspect Export Archives" should be moved together to the "How to - share data" section. Note that this section is targeted at end users that just want to use it, so any technical background information that may be present in the existing sections should be removed or reworded.

sphuber:
The page on external importers should probably not go directly in "Importing Data" because this is a lot of information that is quite specific and it will break the flow of the entire page, which is meant to be more generic. However, the content is still important so we propose to create a new page that is dedicated to docs on dbimporters that ship with `aiida-core`, which is linked to from the "Importing data" section, with maybe a short preface of explaining how the `DbImporter` concept works and can be implemented for any external database

[...]

PR #4467, opened 2020-10-19
Docs: add info on export format to internals section

### LINKED ISSUE #4037: Docs: Internals: Data storage (Export archive)

ramirezfranciscof:
Fix  #4037
Partially addresses #4450


giovannipizzi:
Hi Francisco, can you also remove the old content from the previous source (I think it was in `import_export/main.rst`), either only the content you moved, or all the file, so at the end we know what we didn't migrate yet?

ramirezfranciscof:
Ok; my original idea was that I could anyways wait until all associated new content was merged and then delete all files in a given "remove" issue at once (seemed better than have each one decide what part to remove of the files), but from the message in the slack and this comment I assume you prefer that each deletes the part that each copies.

PD: It seems that we are having a BIG bottleneck in the tests. Maybe we should discuss if there is something that can be done about this in the afternoon meeting...

ramirezfranciscof:
I updated all references to "export" into "AiiDA archive". There are some inconsitencies due to the references in the actual code (`export_version` inside the `metadata.json`, or the command being `verdi export migrate`) but I don't think there is much I can do about those.

Now should be ready to review @giovannipizzi @sphuber

docs/source/internals/data_storage.rst
```diff
@@ -1,32 +1,104 @@
...
 
     `#4036`_
 
-    .. _internal_architecture:orm:export:
+.. _internal_architecture:orm:archive:
+
+AiiDA archive format
+====================
+
+An AiiDA archive file is a ``.zip`` or ``.tar.gz`` file with the following content:
```
giovannipizzi, line 26:
```suggestion
An AiiDA archive file is a file usually ending with extension ``.aiida``, typically compressed in ``.zip`` or ``.tar.gz`` format, with the following content:
```

docs/source/internals/data_storage.rst
```diff
@@ -1,32 +1,104 @@
...
+
+* ``metadata.json`` file containing information on the version of AiiDA as well as the database schema.
+* ``data.json`` file containing the nodes and their links.
+* ``nodes/`` directory containing the repository files corresponding to the nodes.
+
+.. _metadata-json:
```
giovannipizzi, line 32: You might want to use a longer reference name using the full prefix of where this is - these are global labels. So e.g. something like `_internal_architecture:orm:archive:metadata-json:`

docs/source/internals/data_storage.rst
```diff
@@ -1,32 +1,104 @@
...
+* a change occurs in what can or cannot be exported for each entity,
+* the database and/or archive schemes are updated or changed,
+* or standardized exported property values are updated in AiiDA.
+
+.. note::
+    For archives of version 0.3 and older it is advisable that you manually try to convince yourself that the migration was completely successful.
+    While all migrations are tested, trying to include reasonable edge-cases, the migrations involed from version 0.3 to 0.4 are intricate and the possibility of a missing edge-case test is quite real.
+    It is worth noting that if you ever have an issue, please report it on `GitHub <https://www.github.com/aiidateam/aiida_core/issues/new>`_, join the `AiiDA mailing list <http://www.aiida.net/mailing-list/>`_, or use the `contact form <http://www.aiida.net/contact-new/>`_.
+
+The entities that are exported are mentioned in *unique_identifiers* with their respective unique identifiers.
```
giovannipizzi, line 60:
```suggestion
The entities that are exported are mentioned in `unique_identifiers` with their respective unique identifiers.
```

docs/source/internals/data_storage.rst
```diff
@@ -1,32 +1,104 @@
...
+This is used to avoid any incompatibilities among different versions of AiiDA.
+It should be noted that the schema described in ``metadata.json`` is related to the data itself - abstracted schema focused on the extracted information - and not how the data is stored in the database (database schema).
+This makes the import/export mechanism transparent to the database system used, backend selected and how the data is organised in the database (database schema).
+
+Let's have a look at the contents of ``metadata.json``:
+
+.. literalinclude :: includes/metadata.json
+   :language: json
+
+At the beginning of the file, we see the version of the archive file (under `export_version`) and the version of the AiiDA code.
```
giovannipizzi, line 48:
```suggestion
At the beginning of the file, we see the version of the archive file (under ``export_version``) and the version of the AiiDA code.
```

giovannipizzi: I think here and below sometimes single backpacks are used, sometimes stars etc - I suggest using double backticks everywhere we refer to a literal string (e.g. the key of one of the JSON files) - I didn't put a comment on each of them

docs/source/internals/data_storage.rst
giovannipizzi: Same for the fields below

docs/source/internals/data_storage.rst
```diff
@@ -1,32 +1,104 @@
...
+.. note::
+
+    If you have migrated an archive file to the newest version, there may be an extra entry in ``metadata.json``.
+    This simply states from which archive version the file was migrated.
+
+.. note::
+
+    If you supply an old archive file that the current AiiDA code does not support, ``verdi import`` will automatically try to migrate the archive by calling ``verdi export migrate``.
+
+.. _data-json:
```
giovannipizzi, line 77: As above

docs/source/internals/data_storage.rst
```diff
@@ -1,32 +1,104 @@
...
+``data.json``
+-------------
+
+A sample of the ``data.json`` file follows:
+
+.. literalinclude :: includes/data.json
+   :language: json
+
+At the start of the JSON file shown above, the links among the various AiiDA nodes are stated (*links_uuid* field).
+For every link the UUIDs (universal unique identifiers) of the connected nodes, as well as the name of the link, are mentioned.
```
giovannipizzi, line 88:
```suggestion
For every link the UUIDs of the connected nodes, as well as the name of the link, are mentioned.
```
(I think people should know by now what the UUID of a node is)

docs/source/internals/data_storage.rst
```diff
@@ -1,32 +1,104 @@
...
+
+Then follows the data "grouped" into entity types.
+It is worth noticing the references between the instances of the various entities.
+For example the DbNode with identifier *5921143* belongs to the user with identifier 2 and was generated by the computer with identifier 1.
+
+The name of the entities is a reference to the base ORM entities.
+This ensuries that the archive files are cross-backend compatible.
```
giovannipizzi, line 95:
```suggestion
This ensures that the archive files are compatible between different actual database backend/schema implementations (e.g. Django, SQLAlchemy, ...).
```

giovannipizzi (changes requested):
Some minor suggestions.

Also, please add a redirect from the old URL describing the data storage to this new URL.

ramirezfranciscof:
@giovannipizzi for now the redirection works for the full page (so `import_export/main.html --> internals/data_storage.html`), I have just asked if it was possible to do section redirection (`import_export/main.html#export-file-format --> internals/data_storage.html#aiida-archive-format`) which is the link that you originally wanted to fix.

ramirezfranciscof:
For which the answer is that apparently this can't be done, so I guess this is the best we can do.

giovannipizzi (approved):
Thanks!

PR #4457, opened 2020-10-19
Docs: Add redirects from old documenation

chrisjsewell:
This adds redirects from old documentation,
e.g. if you try to open https://aiida.readthedocs.io/projects/aiida-core/en/latest/install/quick_installation.html, it will redirect you to https://aiida.readthedocs.io/projects/aiida-core/en/v1.0.0/intro/get_started.html

PR #4468, opened 2020-10-19
Docs: add "How to install plugins" section

### LINKED ISSUE #4459: Docs: How to install and manage plugins
Comes from issue #4453 . 

Instructions from that thread (striketrhough mine, we first do it this way and then change if necessary):

> There should be a dedicated section that explains how to install plugins. Should go a bit like:
> * Link to registry explaining how plugins can be found
> * explain installation: `pip install` and `reentry scan`
> * explain `verdi plugin list`.
> * Say very important, after you install new plugin -> `verdi daemon restart --reset`.
>
> This section ~~could either be in a "Installing plugins" subsection of the "How to manage installation" section, or it could even become~~ ***should be*** its own toplevel How-to "How to install plugins". This is very important information and should maybe not be buried in some other section. Once we have this easy to find section, other locations that explain this process in any amount of detail can simply be updated to link to this new section instead, reducing duplication.
>
> The more detailed information about what `pip install -e` actually does, should not go in this section. If we want to keep it, we should move it probably to the "How to develop a plugin" section for plugin developers.

sphuber:
Fixes #4459

There was no centralized location yet that explains how a plugin package
is installed, yet this is one of the most critical first steps for new
AiiDA users. A new how-to section is created "How to install plugins"
that details this information. Since a file `plugins.rst` already
existed, which contains information on how to develop a plugin package,
it is renamed to `plugins_develop.rst`.


docs/source/howto/plugins_develop.rst
```diff
@@ -128,8 +128,20 @@ Adding a new entry point consists of the following steps:
 
 Your new entry point should now show up in ``verdi plugin list aiida.calculations``.
 
+.. note::
+
+    What does ``pip install`` do?
```
giovannipizzi, line 133:
```suggestion
    For a package called e.g. `aiida_diff`, what does ``pip install aiida_diff`` do?
```

docs/source/howto/plugins_install.rst
```diff
@@ -0,0 +1,65 @@
...
+    $ pip install aiida-plugin-name
+
+A package can also be installed from the source code.
+In this case, instead of passing the plugin package name, pass the name of the directory that contains the ``setup.py`` file.
+For example, if the code is available through a Git repository:
+
+.. code-block:: console
+
+    $ git clone https://github.com/aiidateam/aiida-diff
+    $ pip install aiida-diff
```
giovannipizzi, line 26:
```suggestion
    $ cd aiida-diff
    $ pip install .
```
(to make sure `aiida-diff` here is a folder name and not the name on pypi).

docs/source/howto/plugins_install.rst
```diff
@@ -0,0 +1,65 @@
...
+If you forget to run this command, AiiDA will not be able to find the plugins.
+
+.. note::
+    The reentry cache can also be updated from python when access to the commandline is not available (e.g. in Jupyter notebooks).
+
+    .. code-block:: python
+
+        from reentry import manager
+        manager.scan(group_re='aiida')
+
```
giovannipizzi, line 43: This is not really critical (actually you could just run `!reentry scan`). Anyway, the important thing to emphasise is the one above - so I would put the note above (on the fact `reentry scan` is important)

giovannipizzi (changes requested):

giovannipizzi:
Let's also remember to add a new redirect plugins->plugins_develop in the open PR #4457

giovannipizzi (approved):

Issue #3320, opened 2019-09-13
Increase performance when creating a new node (before storing) and when using a base numeric node with overloading.


giovannipizzi:
Creating a new node is relatively slow now (even if maybe acceptable):
```
  %%timeit
  b = orm.Float(2)

10000 loops, best of 3: 136 µs per loop
```

Things get much worse when using a Numeric ORM data type and doing operations on it (even before storing). If you first create (once) a Float node with `b = orm.Float(2)` and then execute the following, it's *extremely* slow:
```
  %%timeit
  a = np.ones([10,10,10])
  a = a/b

10 loops, best of 3: 147 ms per loop
```

Actually, the reason is not the call to `.value`: this is pretty fast, because if the node is unstored, `.value` will just access the underlying in-memory dictionary:
```
  %%timeit
  a = np.ones([10,10,10])
  a = a/b.value

The slowest run took 7.52 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 9.19 µs per loop
```
Instead, I think that the problem is that when overloading the operators (in this case the division), AiiDA will do this:
https://github.com/aiidateam/aiida-core/blob/56e0100c8098796f8a9f8ed120be06d8f5f29b97/aiida/orm/nodes/data/numeric.py#L20-L43
Which means it will regenerate, for every element, a new AiiDA node, and I think this is slow.

So a few questions:
- why is this slow? 
  - Is it because of generating a new node (even if it doesn't seem to be so slow)?
  - Is it a slowness of the single dispatch functionality?
     https://github.com/aiidateam/aiida-core/blob/56e0100c8098796f8a9f8ed120be06d8f5f29b97/aiida/orm/nodes/data/base.py#L28
  - Do we really want to keep this possibility of overloading the operators? To me it seems easier to drop it (even if we will have to do it in 2.0, and deprecate it before) as this is quite counterintuitive, unless we manage to speed it up to decent levels.

ConradJohnston:
On the final question, assuming there is no nice fix, as it stands, one has to dereference numeric ORM types do anything serious with Numpy arrays. 
Additionally, for anything other than the numeric types, one has to do some house-keeping (.get_list() and the like) to use them.
This might lead you to conclude that dropping the functionality is better, rather than having to think each time time about whether to use the type directly, and use the AiiDA overloaded operators, or to dereference and use the builtins.

giovannipizzi:
I agree with @ConradJohnston 

[...]

PR #4469, opened 2020-10-19
Docs: Move data types to Topics section

### LINKED ISSUE #4449: Docs: move 'datatypes' content to new topic section
This is a structure that was left over from the old documentation and needs to be removed, but first we need to check if all the content has been re-assigned in the new documentation. The files considered are:

- [ ] [bands.rst](https://aiida-core.readthedocs.io/en/v1.1.0/datatypes/bands.html)
- [ ] [functionality.rst](https://aiida-core.readthedocs.io/en/v1.1.0/datatypes/functionality.html)
- [ ] [index.rst](https://aiida-core.readthedocs.io/en/v1.1.0/datatypes/index.html)
- [ ] [kpoints.rst](https://aiida-core.readthedocs.io/en/v1.1.0/datatypes/kpoints.html)

### Proposition:

It is not clear to me what to do with this, maybe we should create a "Topics/data" section for this?

mbercx:
Fixes #4449 

There was still content from the previous documentation on data types that was missing from the revamped docs. Here we move this content into the "Topics - Data types" section.

### Done

- Move all the content from the source/datatypes directory to the 'Data types' topics section, removing the source/datatypes directory afterwards.

- Restructure the content to first discuss the 'core' types, which we intend to keep in aiida-core in the long term, then the 'materials science' types, which may be moved to a plugin at some point. Move the export content to the end of the file.

- Reformat the content to the current style preferences:
  * Use one sentence per line.
  * Use the proper section headers.
  * Add the references in the proper format.
  * Switch to using the `code-block` directive instead of `::`.

### TODO

The section can still do with quite a large overhaul, I think. Some items on the TODO list:

- [x] Write a short text for each data type, which can start from the "Aim" bullet point for each type.
- [x] Move the other bullet points into a table, perhaps one for the "Core" and "Materials design" types each.
- [x] Show some simple use cases (e.g. how to load a dictionary into a `Dict` node) for each of the data types.
- [ ] Test all of the `KpointsData` and `BandsData` examples. Consider removing legacy code.
- [ ] Consider moving "Export data nodes to various formats" content elsewhere:
  * Move the general text on the `export` method to the new "How to share data" section (#3998).
  * Move the data type specific information to the usage examples.
- [x] Add redirects.


docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+
+Core data types
+===============
+
+Here follows a short summary of common data types already provided with AiiDA.
+This list is not complete, see also inside :py:mod:`aiida.orm.nodes.data` for the list of all available plugins.
+
+We also mention, when relevant, what is stored in the database (as attributes, so that it can be easily queried e.g. with the :ref:`QueryBuilder <how-to:data:find>`) and what is stored in the file repository (providing access to the file contents, but not efficiently queryable: this is useful for e.g. big data files that don't need to be queried for).
+
+For all data types, you can follow the link to the data class to read more about the methods provided, how to access them, and so on.
```
giovannipizzi, line 20: If this is in the "reference" section, should we say something like "follow the link to the reference page on the corresponding data class" or similar?

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+.. _topics:data_types:core:base:
+
+Base types
+----------
+
+In the :py:mod:`aiida.orm.nodes.data.base` module there are a number of useful classes that wrap base python datatypes (like :py:class:`~aiida.orm.nodes.data.int.Int`, :py:class:`~aiida.orm.nodes.data.float.Float`, :py:class:`~aiida.orm.nodes.data.str.Str`, ...).
+These are automatically loaded with the verdi shell, and also directly exposed from :py:mod:`aiida.orm`.
+These classes are particularly useful when you need to provide a single parameter to e.g. a :py:class:`~aiida.engine.processes.functions.workfunction`.
+
+Each of these classes can most often be used transparently (e.g. you can sum two :py:class:`~aiida.orm.nodes.data.int.Int` objects, etc.). If you need to access the bare value and not the whole AiiDA class, use the ``.value`` property.
```
giovannipizzi, line 33: I would add a warning below, saying something along the lines:

Warning! While this is convenient if e.g. you need to do simple manipulations, like multiplying two numbers, be very careful of not passing such nodes instead of the corresponding python values to libraries that perform heavy computations with them: in fact, any operation on the value would be replaced with an operation creating new AiiDA nodes, that however can be orders of magnitude slower. In this case, remember to pass the `node.value` to the mathematical function instead. One can reference #3320

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+
+In the same module, there is also a :py:class:`~aiida.orm.nodes.data.list.List` class to store a list of base data types.
```
giovannipizzi, line 35: Since for most classes you can retrieve the underlying value with `.value`, but for lists and dicts it's different, it would be good to add a small paragraph on this

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+
+.. _topics:data_types:core:dict:
+
+Dict
+----
+
+* **Class**: :py:class:`~aiida.orm.nodes.data.dict.Dict`
+* **String to pass to the** :py:func:`~aiida.plugins.factories.DataFactory`: ``dict``
+* **Aim**: store a dictionary of python base types in the database.
+  It can store any dictionary where elements can be a base python type (strings, floats, integers, booleans, None type, datetime objects) and lists or dictionaries of them, at any depth level (e.g. a dictionary where a value is a list of dictionaries of strings and floats).
```
giovannipizzi, line 48: date time objects are now not storable anymore. We should say it can store any JSON-serializable data.
One can use utility functions from the `datetime` module or from eg. this:
https://github.com/aiidateam/aiida-core/blob/0fb6f72333b0e553590326fc348acbee3ef0763b/aiida/common/timezone.py#L122
to convert datetimes to either strings or floats (seconds from epoch) depending on the querability needs.

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+
+.. _topics:data_types:core:array:
+
+ArrayData
+---------
+
+* **Class**: :py:class:`~aiida.orm.nodes.data.array.ArrayData`
+* **String to pass to the** :py:func:`~aiida.plugins.factories.DataFactory`: ``array``
+* **Aim**: store generic numeric arrays
+* **What is stored in the database**: the shape of the arrays and the name of the arrays
```
giovannipizzi, line 60: Probably here an example would be useful, and also in some below (like SinglefileData) to see what the keys are

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+.. _topics:data_types:core:remote:
+
+RemoteData
+----------
+
+* **Class**: :py:class:`~aiida.orm.nodes.data.remote.RemoteData`
+* **String to pass to the** :py:func:`~aiida.plugins.factories.DataFactory`: ``remote``
+* **Aim**: this basically represents a "symbolic link" to a specific folder on a remote computer.
+  Its main use is to allow users to persist the provenance when e.g. a calculation produces data in a raw/scratch folder, and the whole folder needs to be provided to restart/continue.
+* **What is stored in the database**: the path of the folder (and the remote computer as a `.computer` property, not as an attribute).
```
giovannipizzi, line 94:
```suggestion
* **What is stored in the database**: the absolute path of the remote folder (and the remote computer as a `.computer` property, not as an attribute).
```

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+
+.. _topics:data_types:materials:structure:
+
+StructureData
+-------------
+
+* **Class**: :py:class:`~aiida.orm.nodes.data.structure.StructureData`
+* **String to pass to the** :py:func:`~aiida.plugins.factories.DataFactory`: ``structure``
+* **Aim**: store a crystal structure to be used by atomistic codes
+* **What is stored in the database**: all atomic positions, species, kinds,
```
giovannipizzi, line 123:
```suggestion
* **What is stored in the database**: all atomic positions, species, kinds, periodic boundary conditions, unit cell
```

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+.. _topics:data_types:materials:trajectory:
+
+TrajectoryData
+--------------
+
+* **Class**: :py:class:`~aiida.orm.nodes.data.array.trajectory.TrajectoryData`
+* **String to pass to the** :py:func:`~aiida.plugins.factories.DataFactory`: ``array.trajectory``
+* **Aim**: store molecular trajectories (i.e. sequences of StructureData objects, where
+  then number of atomic kinds and sites does not change over time).
+  beside the coordinates, it can also optionally store velocities.
```
giovannipizzi, line 138:
```suggestion
  Beside the coordinates, it can also optionally store velocities.
```

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+
+.. _topics:data_types:materials:trajectory:
+
+TrajectoryData
+--------------
+
+* **Class**: :py:class:`~aiida.orm.nodes.data.array.trajectory.TrajectoryData`
+* **String to pass to the** :py:func:`~aiida.plugins.factories.DataFactory`: ``array.trajectory``
+* **Aim**: store molecular trajectories (i.e. sequences of StructureData objects, where
+  then number of atomic kinds and sites does not change over time).
```
giovannipizzi, line 137:
```suggestion
* **Aim**: store molecular trajectories (i.e. sequences of StructureData objects, where then number of atomic kinds and sites does not change over time).
  Note however that the internal data format is different than StructureData, see below.
```

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+
+Additional parameters are passed as ``kwargs`` to the underlying implementation, that often accepts a different number of parameters.
+
+.. _topics:data_types:materials:kpoints:seekpath:
+
+Seekpath implementation
+^^^^^^^^^^^^^^^^^^^^^^^
+
+When specifying ``method='seekpath'``, the `seekpath <https://github.com/giovannipizzi/seekpath/>`_ library is used to generate the path.
+Note that this requires  ``seekpath`` to be installed (this is not available by default, in order to reduce the dependencies of AiiDA core, but can be easily installed using ``pip install seekpath``).
```
giovannipizzi, line 205: Should we mention as the 'best' way to install this, to use instead the correct extras in `pip install aiida-core`?

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+
+    bands_spins = [bands, bands-0.3] # to distinguish the bands of different spins we substitute 0.3 from the second band structure
+    bs.set_bands(bands_spins, units='eV')
+    bs.show_mpl()
+
+Now the shape of the bands array becomes ``nspins * nkpoints * nstates``
+
+.. _topics:data_types:export:
+
+Export data nodes to various formats
```
giovannipizzi, line 384: As you already mentioned, maybe all this section should be moved to an 'exporting data' section - like in "importing", we have two types (import an AiiDA DB, and import from external sources). Also here, we might want to have two types of export: export to AiiDA export file, and export to other formats (this section for the AiiDA data types).

docs/source/topics/data_types.rst
mbercx: I agree, but just out of curiosity: why is this different? Also, should we perhaps consider changing the `__str__` method of e.g. `List` in this case, as now this can be somewhat confusing:

```Python
In [1]: l = List(list=[1, 2])

In [2]: l
Out[2]: <List: uuid: fc6335b7-69d9-4d65-bc1a-9adb5242a69f (unstored) value: [1, 2]>

In [3]: l.value
<ipython-input-5-aa957b8aeff4> in <module>
----> 1 l.value

AttributeError: 'List' object has no attribute 'value'
```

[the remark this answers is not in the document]
mbercx: 👍 I'm combining `List` and `Dict` into a subsection called "Iterables".

mbercx: Another question I had regarding the `List` and `Dict` classes is the fact that the user cannot simply initialize e.g. a `List` using:

```Python
l = List([1, 2, 4])
```

Instead, he or she has to specify the `list` argument with its key:

```Python
l = List(list=[1, 2, 4])
```

It's also not immediately clear from the constructor signature or docstring that you have to initialize the `List` this way:

```Python
In [1]: List?
Init signature: List(**kwargs)
Docstring:      `Data` sub class to represent a list.
Init docstring:
:param backend_entity: the backend model supporting this entity
:type backend_entity: :class:`aiida.orm.implementation.entities.BackendEntity`
File:           ~/envs/aiida-docs/code/aiida-core/aiida/orm/nodes/data/list.py
Type:           AbstractNodeMeta
Subclasses:    
```

Is there a good reason for this? I.e. why don't we change the current constructor:

```Python
def __init__(self, **kwargs):
    data = kwargs.pop('list', list())
    super().__init__(**kwargs)
    self.set_list(data)
```

to something like:


```Python
def __init__(self, input_list=None, **kwargs):
    input_list = input_list or list()
    super().__init__(**kwargs)
    self.set_list(input_list)
```

If so, we should at least add a docstring to the constructor. 😅

giovannipizzi: On the first point: indeed, adding a .value to both List and Dict would be OK I think - do you want to open an issue, and/or ask this later at the meeting? I don't know if we also need to remove get_list(), get_dict(), maybe it's OK to keep both?

On the second point: 
- we cannot call it 'input_list' as this is backward-incompatible, we need to call it list (and of course this shadows python's list - but then one can write something like
  ```python
  import builtins
  list = list or builtins.list()
  ```
  (with a comment explaining what's going on ;-) )
- I think there was some historical reasons because, for *any* node, if you passed a kwarg `aaa`, it would have been passed to a method `set_aaa`. Probably this does not exist anymore? (pinging @sphuber) Anyway, maybe this is OK and what you suggest is OK, apart from the parameter name?

ramirezfranciscof:
@mbercx does this still require discussion or was this settled yesterday? If it does need discussion, let's remember to bring it up during today's meeting at 11.

docs/source/topics/data_types.rst
mbercx: 👍 I've removed it from the list here.

docs/source/topics/data_types.rst
mbercx: So far I have moved the CLI examples to the corresponding data type section (as they are rather specific) and kept the general description of the `export()` method at the end of the file. I suppose it could be fit into the "How to share data" section?

giovannipizzi: Yes, but I think we also discussed the option to just leave them at the bottom of the page in yesterday's meeting - I'm not sure which solution is better

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,508 @@
...
+* **String to pass to the** :py:func:`~aiida.plugins.factories.DataFactory`: ``array.bands``
+* **Aim**: store electronic structure bands (of phonon bands)
+* **What is stored in the database**: like ``ArrayData``, the shape of the arrays and the name of the arrays
+* **What is stored in the file repository**:  like ``ArrayData``, the array data in numpy format
+* **Additional functionality**:
+
+  * :ref:`Export to a number of formats (xmgrace, gnuplot, png, pdf, ...)<topics:data_types:export:specific:bands>`
+
+
+``BandsData`` object is dedicated to store bands object of different types (electronic bands, phonons).
```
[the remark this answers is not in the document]
mbercx: I agree that some usage examples should be collapsible, but I'm not sure if this should be the case for all usage examples. The ones I've made for the core types are not very long, so these should be ok. I'll make some examples collapsible based on their length and generality, and then we can have another look.

docs/source/topics/data_types.rst
mbercx: You mean to install the `atomic_tools` extras via:

```
 pip install aiida-core[atomic_tools]
```

?

docs/source/topics/data_types.rst
mbercx: Thanks @giovannipizzi! I opened issue #4495 to discuss these suggestions further.

mbercx:
Thanks @giovannipizzi for the review! Besides your comments, I've already made a lot of changes to the previous version. I think the section is already in pretty good shape, but there are still some data types missing (`CifData`, `OrbitalData` and `ProjectionData`). However, since I still need to familiarize myself with these classes and their API before I can write their docs, this can still be quite time consuming. So either we:

* Continue with the content we have now so we can merge the PR asap, and open an issue to add the other data types in a next round of documentation days.
* Keep the PR open until I've added the discussion/examples of the other data types (but I'm not sure if I can manage that this week).

Let me know what you think is best! In case we want to merge it asap, I'll also look for another reviewer since you indicated you'd be busy today.

chrisjsewell:
Before you merge this, please ensure that `datatypes/index.rst` is redirected to a suitable place (in `docs/source/redirects.txt`)

mbercx:
@giovannipizzi this is ready for another round of review! Also fine to merge in its current state, I'll give the whole section another pass when adding the missing data types (#4529).

docs/source/redirects.txt
```diff
@@ -17,6 +17,7 @@ working_with_aiida/index.rst howto/index.rst
 working_with_aiida/python_api.rst reference/api/public.rst
 working_with_aiida/resultmanager.rst howto/exploring.rst
 working_with_aiida/scripting.rst howto/interact.rst
+datatypes/index.rst topics/data_types.rst
```
giovannipizzi, line 20: Let's also redirect the other deleted files

docs/source/topics/data_types.rst
```diff
@@ -4,8 +4,790 @@
...
+
+.. _topics:data_types:core:base:iterable:
+
+AiiDA has also implemented data classes for two basic Python iterables: :py:class:`~aiida.orm.nodes.data.list.List` and :py:class:`~aiida.orm.nodes.data.dict.Dict`. They can store any list or dictionary where elements can be a base python type (strings, floats, integers, booleans, None type):
+
+.. code-block:: ipython
+
+  In [1]: l = List(list=[1, 'a', False])
+
+Note the use of the keyword argument ``list``, this is required for the constructor of the :py:class:`~aiida.orm.nodes.data.list.List` class. You can also store lists or dictionaries within the iterable, at any depth level. For example, you can create a dictionary where a value is a list of dictionaries:
```
giovannipizzi, line 105: I think the idea is to have a newline after each full stop.

docs/source/topics/data_types.rst
```diff
@@ -4,8 +4,790 @@
...
+
+ArrayData
+---------
+
+The :py:class:`~aiida.orm.nodes.data.array.ArrayData` class can be used to represent `numpy <https://numpy.org/>`_ arrays in the provenance.
+Each array is assigned to a name specified by the user using the :py:meth:`~aiida.orm.nodes.data.array.ArrayData.set_array()` method:
+
+.. code-block:: ipython
+
+  In [1]: SinglefileData = DataFactory('singlefile'); import numpy as np
```
giovannipizzi, line 157: Do we need this line?

docs/source/topics/data_types.rst
```diff
@@ -4,8 +4,790 @@
...
+  Out[7]: ['file3.txt', 'module.py']
+
+The content can once again be shown using the :py:meth:`~aiida.orm.nodes.node.Node.get_object_content()` method by passing the correct path:
+
+.. code-block:: ipython
+
+ In [8]: folder.get_object_content('subdir/file3.txt')
+ Out[8]: 'File 3 content\n'
+
+Since the :py:class:`~aiida.orm.nodes.data.folder.FolderData` node is simply a collection of files, it simply stores these files in the repository
```
giovannipizzi, line 294:
```suggestion
Since the :py:class:`~aiida.orm.nodes.data.folder.FolderData` node is simply a collection of files, it simply stores these files in the repository.
```

docs/source/topics/data_types.rst
```diff
@@ -4,8 +4,790 @@
...
+
+.. code-block:: ipython
+
+  In [1]: RemoteData = DataFactory('remote')
+
+  In [2]: computer = load_computer(label='computer_label')
+
+  In [3]: remote = RemoteData(remote_path='/absolute/path/to/remote/directory' computer=local)
+
+You can see the contents of the remote folder by using the :py:meth:`~aiida.orm.nodes.data.remote.RemoteData.listdir()` method:
```
giovannipizzi, line 313: A big warning is due here - this will open a SSH connection, so this should NOT be done in scripts and even less in workflows; it's just a utility for the user.

docs/source/topics/data_types.rst
```diff
@@ -4,8 +4,790 @@
...
+^^^^^^^^^
+
+The following export formats are available for :py:class:`~aiida.orm.nodes.data.structure.StructureData`:
+
+* ``xsf`` (format supported by e.g. XCrySDen and other visualization software;
+  supports periodic cells)
+* ``xyz`` (classical xyz format, does not typically support periodic cells (even if
+  the cell is indicated in the comment line)
+* ``cif`` (export to CIF format, without symmetry reduction, i.e. always storing the
+  structure as P1 symmetry)
```
giovannipizzi, line 431:
```suggestion
* ``cif`` (export to CIF format, without symmetry reduction, i.e. always storing the structure as P1 symmetry)
```

docs/source/topics/data_types.rst
```diff
@@ -4,8 +4,790 @@
...
+
+Where ``<IDENTIFIER>`` is one of the possible identifiers of the node, e.g. its PK or UUID.
+This outputs the structure in ``xsf`` format and writes it to a file.
+
+.. _topics:data_types:materials:trajectory:
+
+TrajectoryData
+--------------
+
+The :py:class:`~aiida.orm.nodes.data.array.trajectory.TrajectoryData` data type represents a sequences of StructureData objects, where the number of atomic kinds and sites does not change over time.
```
giovannipizzi, line 447: We should emphasise here that, at variance with the StructureData, here coordinates are stored in an array and not in the DB.

docs/source/topics/data_types.rst
```diff
@@ -4,8 +4,790 @@
...
+^^^^^^^^^
+
+You can export the py:class:`~aiida.orm.nodes.data.array.trajectory.TrajectoryData` node with ``verdi data trajectory export``, which accepts a number of formats including ``xsf`` and  ``cif``, and additional parameters like ``--step NUM`` (to choose to export only a given trajectory step).
+
+The following export formats are available:
+
+* ``xsf`` (format supported by e.g. XCrySDen and other visualization software;
+  supports periodic cells)
+* ``cif`` (export to CIF format, without symmetry reduction, i.e. always storing the
+  structures as P1 symmetry)
```
giovannipizzi, line 467:
```suggestion
* ``cif`` (export to CIF format, without symmetry reduction, i.e. always storing the structures as P1 symmetry)
```

docs/source/topics/data_types.rst
```diff
@@ -4,8 +4,790 @@
...
+
+.. todo::
+
+  Think of more example use cases.
+
```
giovannipizzi, line 472: Do we want to keep a todo in the docs? Maybe better to open an issue?

docs/source/topics/data_types.rst
giovannipizzi: yes

docs/source/topics/data_types.rst
```diff
@@ -4,8 +4,790 @@
...
+  The intervals between the kpoints on the X axis are proportional to the cartesian distance between them.
+
+.. dropdown:: Dealing with spins
+
+  The ``BandsData`` object can also deal with the results of spin-polarized calculations.
+  Two provide different bands for two different spins you should just merge them in one array and import them again using the ``set_bands()`` method:
+
+  .. code-block:: python
+
+      bands_spins = [bands, bands-0.3] # to distinguish the bands of different spins we substitute 0.3 from the second band structure
```
giovannipizzi, line 708:
```suggestion
      bands_spins = [bands, bands-0.3] # to distinguish the bands of different spins we subtract 0.3 from the second band structure
```

giovannipizzi (changes requested):
Thanks @mbercx ! A few comments to address, then it's ok to merge with me and the rest can be addressed in a future PR

docs/source/topics/data_types.rst
mbercx: Whoops! That's some sloppy copy-paste work, my bad! This line needs to import the `ArrayData` class, of course. Thanks for spotting it. 👍

docs/source/topics/data_types.rst
mbercx: I've opened an issue on adding the missing data types to this section in #4529. This todo is just to make sure I also expand the usage section on `TrajectoryData` a bit when I work on these.

ramirezfranciscof: I think what we were currently doing is both. So maybe add a reference to that PR in the `.. todo::` (see other sections like `internals/engine.rst` for reference) and also mention in the issue the specific change expected here.

mbercx: 👍  Alright, thanks @ramirezfranciscof!

giovannipizzi:
Thanks @mbercx !

PR #4466, opened 2020-10-19
tox: add env for updating the requirements file(s)

chrisjsewell:
Simply run `tox -e py38-update-reqs` or if you have multiple python's installed: `tox -e py{36,37,38}-update-reqs`

PR #4470, opened 2020-10-19
Docs: ⬆️ Update sphinx + extensions versions

### LINKED ISSUE #3689: Sphinx directives are no longer hooked into sphinx.ext.autodoc
The AiiDA sphinx extension previously supported just using ``.. automodule::``, by hooking into ``sphinx.ext.autodoc``. This no longer works, as can be seen by this part of the test / demo documentation:

image

It should also be checked if there is a way to turn it off. For example https://github.com/lsmo-epfl/aiida-lsmo/blob/develop/docs/source/apidoc/aiida_lsmo.rst currently uses ``automodule``, and I'm not sure if the intended behavior is the "regular" class documentation or not.

chrisjsewell:
In, particular sphinx v2 -> v3

Also moved sphinxext testing to the official sphinx testing infrastructure

fixes #3689


chrisjsewell:
@csadorf can you review this, with respect to dependency changes ta. I updated the requirements as mentioned in: https://github.com/aiidateam/aiida-core/pull/4466#issuecomment-712108328

chrisjsewell:
@ltalirz and @greschd do you want to check the sphinx extension side of this PR

I will also have a look at https://github.com/aiidateam/aiida-core/issues/3300 and https://github.com/aiidateam/aiida-core/issues/3689 in a bit and see if they can be closed
(I'm tempted to change the dropdown implementation to https://sphinx-panels.readthedocs.io/en/latest/#dropdown-usage, but I don't know if I'll have time right now)

aiida/sphinxext/calcjob.py
```diff
@@ -27,8 +29,7 @@ class AiidaCalcJobDocumenter(AiidaProcessDocumenter):
 
     @classmethod
     def can_document_member(cls, member, membername, isattr, parent):
-        from aiida.engine import CalcJob
-        return issubclass(cls, CalcJob)
```
chrisjsewell, line 31: Found your error @ltalirz 😄

greschd:
> Also moved sphinxext testing to the official sphinx testing infrastructure

Is there an easy way to manually check the HTML output with the official testing infra? That was the reason for that `Makefile`.

chrisjsewell:
> Is there an easy way to manually check the HTML output with the official testing infra? That was the reason for that `Makefile`.

I generally use beautifulsoup + pytest-regressions: https://github.com/executablebooks/MyST-Parser/blob/750ea18db6e7eccc147e3bf45921ac4eb6715ab5/tests/test_sphinx/conftest.py#L80

aiida/sphinxext/calcjob.py
ltalirz: Great, thanks a lot Chris!

aiida/orm/nodes/data/list.py
```diff
@@ -77,15 +77,18 @@ def remove(self, value):
         del self[value]
 
     def pop(self, **kwargs):  # pylint: disable=arguments-differ
+        """Remove and return item at index (default last)."""
```
ltalirz, line 80: That function does not seem to return anything or am I wrong?

aiida/tools/groups/paths.py
```diff
@@ -257,7 +257,7 @@ def walk_nodes(
 
         :param filters: filters to apply to the node query
         :param node_class: return only nodes of a certain class (or list of classes)
-        :param int batch_size: The size of the batches to ask the backend to batch results in subcollections.
+        :param query_batch: The size of the batches to ask the backend to batch results in subcollections.
```
ltalirz, line 260: `batch_size` might actually be the better name; anyhow it's fine to adapt the docs instead.

docs/source/conf.py
```diff
@@ -279,6 +279,14 @@
 # If false, no module index is generated.
 #latex_domain_indices = True
 
+# this is to avoid the error
+# aiida/orm/implementation/django/querybuilder.py:docstring of
+# aiida.orm.implementation.django.querybuilder.array_length._compiler_dispatch:1: WARNING: Unknown target name: "visit".
```
ltalirz, line 284: weird!
do we know what's so special about this code?

greschd:
Not sure if that answers my question (maybe I'm just being dense): Can I create the HTML output of the "test app" without changing the test code?

tests/sphinxext/sources/workchain/conf.py
```diff
@@ -86,7 +86,7 @@
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'sphinx_rtd_theme'
+html_theme = 'alabaster'
```
ltalirz, line 89: why change this?

tests/sphinxext/sources/workchain_broken/conf.py
```diff
@@ -86,7 +86,7 @@
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'sphinx_rtd_theme'
+html_theme = 'alabaster'
```
ltalirz, line 89: same here

chrisjsewell:
Gah-damn, now that #3689 is fixed, you need to reference the processes with their new object types e.g. `:py:process:` or `:process:`, not `:py:class:` or `:class:`

tests/sphinxext/test_workchain.py
```diff
@@ -8,31 +8,27 @@
...
-def test_workchain_build(build_sphinx, xml_equal, reference_result):
+def test_workchain_build(sphinx_build_factory, xml_equal, reference_result):
     """Test building sphinx documentation for WorkChain.
 
     Builds Sphinx documentation for workchain and compares against expected XML result.
     """
-    out_dir, completed_proc = build_sphinx(WORKCHAIN)
-    assert completed_proc.returncode == 0
-    index_file = join(out_dir, 'index.xml')
+    sphinx_build = sphinx_build_factory('workchain', buildername='xml')
```
chrisjsewell, line 19: @greschd you just change the `buildername` here

tests/sphinxext/reference_results/workchain.xml
```diff
@@ -70,25 +70,37 @@ finalize</literal_block></paragraph>
...
-        <index entries="['single',\ 'DemoWorkChain\ (class\ in\ demo_workchain)',\ 'demo_workchain.DemoWorkChain',\ '',\ None]"></index>
         <desc desctype="class" domain="py" noindex="False" objtype="class">
-            <desc_signature class="" first="False" fullname="DemoWorkChain" ids="demo_workchain.DemoWorkChain" module="demo_workchain" names="demo_workchain.DemoWorkChain"><desc_annotation xml:space="preserve">class </desc_annotation><desc_addname xml:space="preserve">demo_workchain.</desc_addname><desc_name xml:space="preserve">DemoWorkChain</desc_name><desc_parameterlist xml:space="preserve"><desc_parameter xml:space="preserve">inputs=None</desc_parameter><desc_parameter xml:space="preserve">logger=None</desc_parameter><desc_parameter xml:space="preserve">runner=None</desc_parameter><desc_parameter xml:space="preserve">enable_persistence=True</desc_parameter></desc_parameterlist></desc_signature>
+            <desc_signature first="False" fullname="WorkChain"><desc_annotation xml:space="preserve">workchain</desc_annotation><desc_addname xml:space="preserve">demo_workchain.</desc_addname><desc_name xml:space="preserve">DemoWorkChain</desc_name></desc_signature>
```
ltalirz, line 76: Looks like the output changed here (e.g. from `DemoWorkChain` to `WorkChain`), is this intended?

ltalirz (approved):
thanks a lot @chrisjsewell !

Just a couple of comments & minor questions.

@greschd Please also let us know whether you're happy with this solution.

tests/sphinxext/test_workchain.py
greschd: Hmm, I think the reason I had `xml` here initially is because the HTML output would be too flimsy a test (slight changes and such). If your (more recent) experience is that it's not it would be fine to change. 
Otherwise I'd prefer keeping a way to build the "test documentations" outside of the actual tests - the reason being that is that it's a nice way to iterate on the design of the HTML.

ltalirz:
Sorry, that should have been "request changes" ;-)

tests/sphinxext/sources/workchain/conf.py
greschd: Guess this is now the only place where we need `sphinx_rtd_theme`..

ltalirz:
Is there a links to the docs that we can check?

tests/sphinxext/sources/workchain/conf.py
chrisjsewell: because `sphinx_rtd_theme` is a separate theme package that we no longer have as a dependency, whereas alabaster if the built-in theme

tests/sphinxext/sources/workchain_broken/conf.py
chrisjsewell: same as above

chrisjsewell:
you just click on:

image

then artifacts/html/index.html

image

https://1443-77234579-gh.circle-artifacts.com/0/html/index.html

docs/source/conf.py
chrisjsewell: I know, Maybe something to-do with the `_` 🤷 https://github.com/sphinx-doc/sphinx/issues/4350, but that docs are inherited comes from sqlachemy anyway (I think), so we can't really "fix" it

aiida/tools/groups/paths.py
chrisjsewell: yeh it probably should be, but thats outside the scope of this PR

aiida/orm/nodes/data/list.py
chrisjsewell: Hmm, yeh good point it should actually return; thats a bug in the code I would say.

Its a bit out of scope of this PR, so perhaps we should fix in another small PR.
(I only actually made this change to fix a warning by sphinx about the inherited docstring)

tests/sphinxext/test_workchain.py
chrisjsewell: Which is why I mentioned using beautifulsoup; to sanitise the output before comparison.

If you want to build the docs, just run: `sphinx-build -b html -n -W --keep-going tests/sphinxext/sources/workchain tests/sphinxext/sources/workchain/build`, you don't need a Makefile.

Note at present, you would have to add to conf.py something like:

```python
import sys
sys.path.append("/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/tests/sphinxext/workchains")
from aiida.manage.configuration import load_documentation_profile
load_documentation_profile()
```

The latter part we should actually add to the extension (as an optional step governed by a config value), because I imagine you can't actually use it without running that.

alternatively to iterate on the design (this is how I do it), just paste somewhere in the docs, e.g.

```
.. aiida-calcjob:: CalcJob
    :module: aiida.engine.processes.calcjobs
    :expand-namespaces:
```

and run `tox py38-docs-live`, and you get live updates of the docs builds in the browser

aiida/orm/nodes/data/list.py
ltalirz: Up to you - in any case, the docstring should be consistent and if you prefer not to fix it here, we should open an issue

tests/sphinxext/sources/workchain/conf.py
ltalirz: Thanks for the clarification!

chrisjsewell:
@ltalirz @greschd I've rolled back (i.e. commented out) the autodoc writer, because it doesn't work with referencing. From the current implementation, I'm pretty sure that has never worked.

As I mention in https://github.com/aiidateam/aiida-core/issues/3300#issuecomment-712502001, I think the whole extension needs a bit of a re-write to fix this.
But I'm definitely not doing it in this PR, since the point was really only to upgrade to sphinx 3 and now I've already wasted enough time messing around with it 😒

tests/sphinxext/reference_results/workchain.xml
chrisjsewell: removed this change

aiida/orm/nodes/data/list.py
chrisjsewell: ok I've fixed it here then 446d1a3

setup.json
```diff
@@ -70,11 +70,11 @@
...
             "pygments~=2.5",
-            "pydata-sphinx-theme~=0.3.0",
-            "sphinx~=2.2",
+            "pydata-sphinx-theme~=0.4.0",
+            "sphinx~=3.2",
             "sphinxcontrib-details-directive~=0.1.0",
-            "sphinx-panels~=0.4.0",
-            "sphinx-copybutton~=0.2.12",
+            "sphinx-panels~=0.5.0",
+            "sphinx-copybutton~=0.3.0",
```
csadorf, line 77: Is there a reason that we restrict the sphinx extensions to a minor version?

csadorf (approved):
> @csadorf can you review this, with respect to dependency changes ta. I updated the requirements as mentioned in: [#4466 (comment)](https://github.com/aiidateam/aiida-core/pull/4466#issuecomment-712108328)

I have one question, but generally fine w.r.t. the updates to the dependency specification.

tests/sphinxext/test_workchain.py
greschd: > The latter part we should actually add to the extension (as an optional step governed by a config value), because I imagine you can't actually use it without running that.

Actually, you can. The extension itself doesn't rely on the profile being loaded, so if the `.spec()` of the plugin can be created without it it'll work. But yeah, I think 99% of plugins will use it so we might as well add that.

> and run tox py38-docs-live, and you get live updates of the docs builds in the browser

Good enough for me :+1:

greschd:
> @ltalirz @greschd I've rolled back (i.e. commented out) the autodoc writer, because it doesn't work with referencing. From the current implementation, I'm pretty sure that has never worked.

Can you revert this? There are existing plugins using ``automodule`` - and I think referencing being broken isn't as bad as regressing to a "plain" Sphinx class.

setup.json
sphuber: I think probably just because they haven't release a v1.0 yet. So far we have been maintaining the rule of thumb that for packages with at least a v1 we pin to the major version, whereas for v0.* releases we pin to the minor version. This is because typically packages are a bit more liberal with breaking the interface before a v1 release.

chrisjsewell: Indeed, these packages are both from EBP and that version convention is specifically outlined (by me 😄) https://github.com/executablebooks/.github/blob/master/CONTRIBUTING.md#releases-and-change-logs

chrisjsewell:
> Can you revert this? There are existing plugins using `automodule` - and I think referencing being broken isn't as bad as regressing to a "plain" Sphinx class.

@greschd Unfortunately no. When I say revert, I mean revert from earlier in this PR. `automodule` has already regressed to plain sphinx classes (for about a year!) and turning it on breaks 116 references in the aiida-core documentation, with no fix until the whole extension is basically re-wrote.

greschd:
Ah yeah now I remember, that was #3689. I'm not sure referencing ever worked (although I _think_ I was following some guide when writing the directive). Out of curiosity, why does the extension need to be rewritten for referencing to work?

chrisjsewell:
Basically we need to "integrate" it into the "py" domain, a bit like `sphinx/domains/python::PyClasslike` does, such that it also provides an XRefRole and stores an object in the py domain.
The actual creation of the doctree is generally fine though, its just the stuff around it.

setup.json
csadorf: Makes sense. 👍

PR #4475, opened 2020-10-19
Docs: add "How to interact with AiiDA" section

### LINKED ISSUE #4462: Docs: How to interact with AiiDA
Regarding the content of `working_with_aiida/scripting.rst` in issue #4452 

> this information is important and is not yet present. Should probably be a new page in the how-tos, the very first one, called "How to interact with AiidA". Should explain that there are multiple ways of interacting, give itemization: shell, script, REST API. Each one is then explained a bit, with links to other parts of documentation with more detailed information (e.g. in the case of REST API), not sure it applies to the shell and scripts.

sphuber:
Fixes #4462 

This section gives a high-level overview of the various methods with
which can interact with AiiDA. For details it refers to relevant
sections with more information.


sphuber:
As discussed during the meeting today: I think having a page similar to what this PR is adding *somewhere* will be very useful as a starting point to direct to other more detailed pages in the docs that may otherwise be more difficult to find. It may not merit its own how-to top level section and may fit better elsewhere. Open to suggestions and happy to discuss.

Note that this also still needs to update the links in the REST API stub, which is waiting for @ltalirz to finish his section on said REST API. Once that is merged, I can point to the correct sections.

Pinging @csadorf and @giovannipizzi also for discussion.

chrisjsewell:
Definitely add a link to this at the bottom of get_started.rst

csadorf:
I would be fine with introducing this as a top-level how-to section.

docs/source/howto/interact.rst
```diff
@@ -0,0 +1,112 @@
...
+
+    In [2]: load_profile()
+    Out[2]: <aiida.manage.configuration.profile.Profile at 0x7fccfd6c50a0>
+
+One can pass a particular profile name to :meth:`~aiida.load_profile`, otherwise the default profile is loaded.
+Alternatively, the command line interface ``verdi`` provides a handy shortcut:
+
+.. code-block:: console
+
+    $ verdi shell
```
giovannipizzi, line 46: Since you speak about profiles above, maybe this is the right place to remind people that the standard way of switching profile with verdi (not only for `shell`) is with `verdi -p profilename shell`

EDIT: I see this is explained below - as I suggest below, I would swap the order, so how to run a script is explained first, and this includes also `verdi -p XXX run`

docs/source/howto/interact.rst
```diff
@@ -0,0 +1,112 @@
...
+
+This command will open a normal IPython shell but automatically loads the AiiDA profile.
+In addition, certain modules from AiiDA's API that are used very often are automatically imported.
```
giovannipizzi, line 49: There used to be a documentation of this, is this still available, or should we put an incomplete list here at least to give a feeling to users of what is auto-imported? Otherwise, pointing to https://github.com/aiidateam/aiida-core/blob/2174924e40c3c9f00263f2e892e63a07eb90b40f/aiida/cmdline/utils/shell.py#L12 
could be an option.

Also, if no docs for `verdi shell` exists, we should probably add (here or a specific page) on this, at least to  mention the `verdi.shell.auto_import` option of `verdi config`

ramirezfranciscof:
@sphuber the test for the docs seems to have failed, is this the problem that was mentioned yesterday of the limit of 2 in the doc builds or is this something to be fixed in the code?

docs/source/howto/interact.rst
```diff
@@ -0,0 +1,112 @@
...
+
+.. code-block:: ipython
+
+    In [1]: from aiida import load_profile
+
+    In [2]: load_profile()
+    Out[2]: <aiida.manage.configuration.profile.Profile at 0x7fccfd6c50a0>
+
+One can pass a particular profile name to :meth:`~aiida.load_profile`, otherwise the default profile is loaded.
+
```
giovannipizzi, line 68: We should link here to the documentation of the `%aiida` magic. The documentation is here: https://github.com/aiidateam/aiida-core/blob/develop/docs/source/intro/installation.rst#using-aiida-in-jupyter
https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/installation.html#using-aiida-in-jupyter

docs/source/howto/interact.rst
```diff
@@ -0,0 +1,112 @@
...
+
+.. _how-to:interact-scripts:
+
+Scripts
+=======
+
+AiiDA's Python API can be used in Python scripts mixed with any other Python code.
+The only requirement, is that before the API is used, an AiiDA profile is loaded.
```
giovannipizzi, line 76:
```suggestion
The only requirement is that before the API is used, an AiiDA profile is loaded.
```

docs/source/howto/interact.rst
```diff
@@ -0,0 +1,112 @@
...
+This can be accomplished by adding the following lines to the script:
+
+.. code-block:: python
```
giovannipizzi, line 79: This is a bit of repetition.
Can we start with this section, and then later when we discuss `ipython` and `jupyter` we just say that we need to use the same `load_profile` command already explained?

docs/source/howto/interact.rst
```diff
@@ -0,0 +1,112 @@
...
+The downside is that it does not allow to specify a particular profile, but it always loads the default profile.
+
+
+.. _how-to:interact-restapi:
+
+REST API
+========
+
+AiiDA ships with a built in REST API, that allows you to query the data of a particular profile.
+Refer to section TODO_LINK to learn how to start the REST API.
```
giovannipizzi, line 111: You already mentioned this, but I'm adding as a review comment so we don't forget - this needs to be updated as soon as the REST API docs go back in place.

giovannipizzi (changes requested):

docs/source/howto/interact.rst
sphuber: How do you want to reorganize it though? I wanted the separate sections to be independent and not refer to one another to be fully understandable. I thought the repetition was worth it. How would you restructure it?

docs/source/howto/interact.rst
sphuber: I would keep the shell first, but I can move the explanation for the `-p` option to here as well. Would that work?

sphuber:
@giovannipizzi I have restructured the page and addressed all comments. I have also already added the correct links to the anchors of the REST API section that Leo has added in his PR. Would be good to already re-review the rest of the content so that it is ready to be merged once his PR is merged.

giovannipizzi (approved):

giovannipizzi:
Approved it, but before merging there are two warnings to fix so that tests pass. One is just the missing link to the REST API section; the other warning however is for `py:meth reference target not found: aiida.load_profile`

PR #4482, opened 2020-10-20
Docs: move data/plugin from how-to to topics

### LINKED ISSUE #4280: Docs: Move up "querying", move down "adding support for custom datatypes"

Currently, there is the section "Adding support for custom datatypes" as the second entry, before querying
I advocate that this should be essentially the last section of the section (and maybe even be moved somewhere else?) and Querying should be first (or at worst second after importing - logically you might want to import first, but in practice when you search for the docs you will search much more often how to query rather than how to import that is easy to remember).

I think we should change the order, move the "Adding support" much down, also below the sections like organising and deleting data, and probably also after 'serving data to others', and make sure the query part is first or second.

Pinging @csadorf @ltalirz @chrisjsewell for feedback

(This originated from a user that contacted me and couldn't easily find that section, and more generally the second part of this comment):

> Aiida is on version 1.3, which broke most of the Google results as they return Documentation for 1.2 <https://aiida-core.readthedocs.io/en/stable/restapi/index.html>, which can be viewed cached but can be outdated). The documentation is heavily documented for *running* jobs, but much lighter from what I can find on querying the data in ways that work with your dataset.

(For the first part, I think it's OK, but we should be careful when changing the RST file names - we should try to reduce to the minimum now I think to avoid breaking links)

ramirezfranciscof:
fixes #4280


sphuber:
We discussed yesterday to move this to a topic instead. It was decided that the tutorials on writing new plugins (with the exception of `CalcJob` and `Parser` for now) should go to a topic. So I guess this section that you moved, should instead go to the topics. I think it would make sense to move it to the same page that @mbercx is writing on the various data types.

ramirezfranciscof:
@sphuber Tis done

sphuber:
Shall we wait for @mbercx 's section to be moved in?

ramirezfranciscof:
@sphuber I made the scafolding exactly as he has it in his branch and checked with him and told me it was ok to proceed with this before; it should rebase without any problem.

docs/source/topics/data_types.rst
```diff
@@ -0,0 +1,177 @@
+.. _topics:data_types:
+
+**********
+Data types
+**********
+
+.. note:: Here comes datatypes (Marknik)
```
sphuber, line 7: Note sure if this typo is worth updating the code, since "Marknik" will be removing it soon anyway ;)
```suggestion
.. note:: Here comes datatypes (Marnik)
```

sphuber (approved):
Alrighty then!

PR #4488, opened 2020-10-20
Docs: update REST API wsgi scripts

### LINKED ISSUE #4030: Docs: update wsgi example for REST API
- [x] update wsgi example for new `configure_api` interface of the REST API
- [ ] simplify it (just copy-paste the necessary code directly into the respective section for deploying a web server
- [ ] re-read the section on deploying a web server to check it is consistent

ltalirz:
fixes #4030

Open questions:
 - [ ] The old documentation mentioned the possibility of running multiple wsgi scripts using the same daemon process ([one.conf](https://github.com/aiidateam/aiida-core/blob/v1.2.1/docs/wsgi/one.conf)). Does this actually work / has it ever been used in production?


giovannipizzi (approved):
Thanks!

giovannipizzi:
About your question:
The file was added by Fernando here: 
https://github.com/aiidateam/aiida-core/commit/8004ae71dac6f86e75e600e84636de3f5a6f6e28#diff-32fed9844368101395bbc050c45cf94e4ffbe93d96fd4bd648a3f5639e2cadc0

My guess is that he had actually tested it - but it's also true that it was a very old version of AiiDA so not even sure if this still works.
Do you think this might help us in Materials Cloud?
In this case it would be worth to investigate (or ask someone in the team to investigate)

ltalirz:
> Do you think this might help us in Materials Cloud?
> In this case it would be worth to investigate (or ask someone in the team to investigate)

I didn't really know what the difference was, so I wasn't sure.

I've now had a quick look and I think it wouldn't really make a difference - even WSGI applications running inside the same process (group) [will run in separate sub interpreters](https://modwsgi.readthedocs.io/en/develop/user-guides/configuration-guidelines.html#defining-process-groups) and thus lead to (practically the same) memory consumption.

There is, however, also the [application group](https://modwsgi.readthedocs.io/en/develop/user-guides/configuration-guidelines.html#defining-application-groups) which lets wsgi applications run inside the same interpreter.
I don't think this would work with the current state (profile switching needed), but I've opened an issue to keep track of this for the future https://github.com/aiidateam/aiida-core/issues/4494

PR #4491, opened 2020-10-20
Docs: add the "How to explore the provenance graph" section

### LINKED ISSUE #4464: Docs: How to explore the provenance graph
Regarding the content of `working_with_aiida/resultmanager.rst` from issue #4452 

> we couldn't actually find anywhere in the current docs that explains the `inputs`, `outputs` and `res` managers, but this is crucial information. We think there should be a new top-level How-to called "How to explore the provenance graph" that describes all these methods: `get_incoming`, `get_outgoing`, the `inputs` and `outputs` short cuts and the `res` manager for `CalcJobNodes`.

sphuber:
Fixes #4464 

This section adds completely new documentation on the graph exploring
methods `get_incoming` and `get_outgoing` that were not yet documented.
It also documents the `creator`, `caller` and `creator` properties that
also weren't documented yet.

The `working_with_aiida/resultmanager.rst` is removed and its content,
that documents the `inputs` and `outputs` properties of process nodes as
well as the `res` property of the `CalcJobNode` class, is added to the
new section in heavily rewritten form.


docs/source/howto/exploring.rst
```diff
@@ -0,0 +1,137 @@
...
+ * ``link_label_filter``: accepts a string regex (with optional wildcards), only neighboring nodes that are linked with a link label that matches the regex will be returned
+
+As an example:
+
+.. code-block:: python
+
+    node.get_incoming(node_class=Data, link_type=LinkType.INPUT_CALC, link_label_filter='output%node_').all_nodes()
+
+will return only neighboring data nodes that are linked to the ``node`` with a link of type ``LinkType.INPUT_CALC`` and where the link label matches the regex ``'output%node_'``.
+The ``%`` character matches one or more characters and the ``_`` character matches exactly one character.
```
giovannipizzi, line 57:
```suggestion
Reminder on the syntax of SQL `LIKE` patterns: The ``%`` character matches any string of zero or more characters, while the ``_`` character matches exactly one character. These two special characters can be escaped by prepending them with a backslash (note that when putting a backslash in a python string you have to escape the backslash itself, so you'll need two backslashes: e.g., to match exactly a link label `a_b` you need to pass `link_label_filter='a\\b'`).
```

docs/source/howto/exploring.rst
```diff
@@ -0,0 +1,137 @@
...
+ * ``link_type``: accepts a vale of :class:`~aiida.common.links.LinkType`, only neighboring nodes that are linked with this link type will be returned
+ * ``link_label_filter``: accepts a string regex (with optional wildcards), only neighboring nodes that are linked with a link label that matches the regex will be returned
+
+As an example:
+
+.. code-block:: python
+
+    node.get_incoming(node_class=Data, link_type=LinkType.INPUT_CALC, link_label_filter='output%node_').all_nodes()
+
+will return only neighboring data nodes that are linked to the ``node`` with a link of type ``LinkType.INPUT_CALC`` and where the link label matches the regex ``'output%node_'``.
```
giovannipizzi, line 56:
```suggestion
will return only neighboring data nodes that are linked to the ``node`` with a link of type ``LinkType.INPUT_CALC`` and where the link label matches the pattern ``'output%node_'``.
```

docs/source/howto/exploring.rst
```diff
@@ -0,0 +1,137 @@
...
+However, often one doesn't need this expressivity and simply wants to retrieve all neighboring nodes with a syntax that is as succint as possible.
+A prime example is to retrieve the *inputs* or *outputs* of :ref:`a process <topics:processes:concepts>`.
+Instead of using :meth:`~aiida.orm.nodes.node.Node.get_incoming` and :meth:`~aiida.orm.nodes.node.Node.get_outgoing`, to get the inputs and outputs of a ``process_node`` one can do:
+
+.. code-block:: python
+
+    inputs = process_node.inputs
+    outputs = process_node.outputs
+
+These properties do not return the actual inputs and outputs directly, but instead return an instance of :class:`~aiida.orm.utils.managers.NodeLinksManager`
```
giovannipizzi, line 75:
```suggestion
These properties do not return the actual inputs and outputs directly, but instead return an instance of :class:`~aiida.orm.utils.managers.NodeLinksManager`.
```

docs/source/howto/exploring.rst
```diff
@@ -0,0 +1,137 @@
...
+The reason is because through the manager, the inputs or outputs are accessible through their link label and it is tab-completed.
```
giovannipizzi, line 76:
```suggestion
The reason is because through the manager, the inputs or outputs are accessible through their link label (that, for inputs and outputs of processes, is unique) and can be tab-completed.
```

docs/source/howto/exploring.rst
```diff
@@ -0,0 +1,137 @@
...
+.. code-block:: python
+
+    process_node.outputs['result']
+
+If there is no neighboring output with the given link label, a :class:`~aiida.common.exceptions.NotExistentAttributeError` or :class:`~aiida.common.exceptions.NotExistentKeyError` will be raised, respectively.
+
+.. note::
+
+    The ``inputs`` and ``outputs`` properties are only defined for :class:`~aiida.orm.nodes.process.process.ProcessNode`'s.
+    This means that you cannot *chain* these calls, because an input or output of a process node is guaranteed to be a :class:`~aiida.orm.nodes.data.Data` node, which does not have inputs or outputs.
```
giovannipizzi, line 94: I would put this comment below, after explaining creator, to say that inputs+creator allows to go up in the provenance graph as much as you want, chaining them. Instead, going down, you cannot for the reason you mention

docs/source/howto/exploring.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Creator, caller and called
+==========================
+
+Similar to the ``inputs`` and ``outputs`` properties of process nodes, there are some more properties that make exploring the provenance graph easier:
+
+    * :meth:`~aiida.orm.nodes.process.process.ProcessNode.called`: defined for :class:`~aiida.orm.nodes.process.process.ProcessNode`'s and returns the list of process nodes called by this node.
+      If this process node did not call any other processes, this property returns an empty list.
+    * :meth:`~aiida.orm.nodes.process.process.ProcessNode.caller`: defined for :class:`~aiida.orm.nodes.process.process.ProcessNode`'s and returns the process node that called this node.
+      If this ndoe was not called by a process, this property returns ``None``.
```
giovannipizzi, line 107:
```suggestion
      If this node was not called by a process, this property returns ``None``.
```

docs/source/howto/exploring.rst
```diff
@@ -0,0 +1,137 @@
...
+
+
+.. _how-to:exploring:creator-caller-called:
+
+Creator, caller and called
+==========================
+
+Similar to the ``inputs`` and ``outputs`` properties of process nodes, there are some more properties that make exploring the provenance graph easier:
+
+    * :meth:`~aiida.orm.nodes.process.process.ProcessNode.called`: defined for :class:`~aiida.orm.nodes.process.process.ProcessNode`'s and returns the list of process nodes called by this node.
```
giovannipizzi, line 104: Not relevant for this PR, but I realise now that we could have defined this only for WorkflowNodes, actually, since the list will always be empty for CalculationNodes

docs/source/howto/exploring.rst
```diff
@@ -0,0 +1,137 @@
...
+
+will return a list of all the keys in the output node.
+Individual keys can then be accessed through attribute dereferencing:
+
+.. code-block:: python
+
+    node.res.some_key
+
+In an interactive shell, the available keys are also tab-completed.
+If you type ``node.res.`` followed by the tab key twice, a list of the available keys is printed.
```
giovannipizzi, line 137: I would add that, if the default output node link label is `output_parameters`, then `node.res.some_key` is equivalent to `node.outputs.output_parameters.dict.some_key` and stress that we are looking at the keys of a different node (this, together with the fact that `verdi node outputcat/outputls` act on a different node, does not seem to be super intuitive at first for new users, in my experience)

giovannipizzi (changes requested):

docs/source/howto/exploring.rst
sphuber: Did you mean `link_label_filter='a\\_b'` there in the last sentence? I think you may have forgotten the underscore

docs/source/howto/exploring.rst
sphuber: Yeah, and I think the `ProcessNode` should also have a `inputs` and `outputs` implementation. Now I cannot reference it

docs/source/howto/exploring.rst
sphuber: Not sure where exactly you intend this to be placed. It is true that chaining `creator` and `inputs` alternatively works, but that is just one example and not every node has a creator. Instead, I kept this as is and added another note after the explanation of `creator` that gives the example you propose.

sphuber:
@giovannipizzi addressed all your comments

giovannipizzi (approved):
Thanks!

Issue #4254, opened 2020-07-11
Docs: flow of Introduction section unclear


ltalirz:
### Describe the current issue

The 2020 virtual tutorial left a number of participants confused about the flow of the "Introduction" section.

In particular:
 * For some people taking the conda route, it was unclear that they needed to initialize the database and start the postgres server manually (+rabbitmq)
 * For some people taking the "system postgresql/rabbitmq" route it was unclear that they do *not* need to manually create a database and start the postgres server (+rabbitmq)
 * Some people taking the conda route ended up in the "detailed installation" section and trying things, while their installation was already ready to go

### Describe the solution you'd like

Looking back at the Introduction section, I see two possible sources of this confusion:

1) The 4 choices at the beginning of the "Getting started" section have blue buttons that send you directly to parts of the "detailed instructions". Many users probably clicked on them and therefore simply jumped over the "Quick start" section.

My proposed solution:
Move the 4 choices to the end of the "Getting started" section and begin with the "Quick start" right away (if you like, we can leave a link to "alternative ways of installing AiiDA" in the beginning of "Getting started").

2) The current layout suggests a sequential structure, where the "detailed instructions" come after the "getting started". In my opinion, after "Getting started" the "Basic tutorial" should be next.
![image](https://user-images.githubusercontent.com/1053245/87144095-854d8b00-c2a7-11ea-9b53-e50139eb7ff4.png)
  
My proposed solution: Move the detailed installation towards the end of the "Introduction" section

What is AiiDA
Getting started
Basic tutorial
Troubleshooting
Detailed installation

Mentioning @chrisjsewell @mbercx @giovannipizzi 

P.S. I have no strong opinion on the order of Troubleshooting vs Detailed installation; swapping them around would also be fine.

chrisjsewell:
Thanks @ltalirz. Hmm, its difficult, I don't have a clear solution. Me and @csadorf went over this before and I get the impression that everyone is going to have a different opinion on this, and it's always going to confuse at least one person. Although I agree there is room for improvement I would note:

> For some people taking the conda route, it was unclear that they needed to initialize the database and start the postgres server manually (+rabbitmq)

The section headings in the quick start are literally:

- Install Software
- Initialise Data Storage
- Start Computation Services

How much more clear can you be lol?

> For some people taking the "system postgresql/rabbitmq" route it was unclear that they do not need to manually create a database and start the postgres server (+rabbitmq)

Yes I think this needs to be clearer

[...]

PR #4496, opened 2020-10-21
Docs: Revise "Get started" flow

csadorf:
This is a comprehensive revision of the "Get started" and "Advanced installation" page to address the issues identified in #4254 .

In summary, the "Get started" page has been completely revised to provide a more linear flow for all supported setup flows and the "Detailed installation" has been largely pruned and renamed to "Advanced configuration" with most content directly incorporated into the linear setup flows.

These are the changes in detail:

- The "routes overview" panels have been revised to be simpler and more descriptive in their differences. 
- While the "routes overview" panels are still principally distinguished by "level of virtualization", the primary distinction between the first two supported installation methods ("system-wide installation" and "pure conda installation") is the path for installing prerequisite services, **not** whether the AiiDA Python package is installed into a virtual environment (via conda/venv). The primary difference here is that the first method requires administrative privileges, the latter does not. The revised descriptions should make this much clearer.
- Each route panel is clickable and leads directly to a full guide for the specific route on a different page.
- All routes are presented with a linear path, choices (e.g. operating system, Conda/pip) are represented via tabs. 
- Optional steps and potential issues are presented as drop-downs in order to incorporate them into the linear flow without being too distracting to those users where they don't apply.
- No specific route is clearly recommend over another because it really depends on the user environment and use case. However the *system-wide installation* is likely most applicable to novice users who – given the presented options – are unsure about which route is best for them. This is based on the assumption that the alternative routes would likely not be applicable to those users who are unfamiliar with the mentioned technology in the first place. For example, a user should probably not take the docker route if they are unfamiliar with docker, however users who are familiar with docker, will likely be able to evaluate whether that is the right choice for them. The system-wide installation is therefore now explicitly recommended to those users *who are unsure* about which route is best for them.
- The "Detailed Installation" page has been renamed to "Advanced configuration". This is motivated by the fact that all supported setup routes are now fully covered on the "Getting started" page. Caveats and potential issues are by and large directly incorporated into those linear paths. The information found on the "Advanced configuration" page now covers edge cases for advanced users who want to further customize their setup or who have rare environmental conditions. The fact that the content of that chapter will likely not be applicable to most users is made explicit at the top of the page.
- Slight duplication between different routes is deliberately tolerated to be able to represent a linear flow to users.
- The instructions for the Windows Subsystem for Linux have been updated and now support both version 1 and 2.


chrisjsewell:
Ah look at those awesome tabs! I wonder who wrote that code 😝 

https://1535-77234579-gh.circle-artifacts.com/0/html/intro/get_started.html

csadorf:
@greschd and @zhubonan Could you help me fix up the Windows Subsystem installation flow? I don't think it is 100% correct/complete at the moment. Unfortunately I do not have access to a windows machine.
https://1538-77234579-gh.circle-artifacts.com/0/html/intro/get_started.html#intro-get-started-system-wide-install

csadorf:
> Ah look at those awesome tabs! I wonder who wrote that code 😝
> 
> https://1535-77234579-gh.circle-artifacts.com/0/html/intro/get_started.html

The tabs are really great! 👍 👍

chrisjsewell (approved):
initial comments

chrisjsewell (changes requested):
that was meant to be request changes

greschd:
> 
> @greschd and @zhubonan Could you help me fix up the Windows Subsystem installation flow? I don't think it is 100% correct/complete at the moment. Unfortunately I do not have access to a windows machine.
> https://1538-77234579-gh.circle-artifacts.com/0/html/intro/get_started.html#intro-get-started-system-wide-install

I use the Ubuntu rabbitmq now (I think this didn't work in WSL1) - not sure if we should change the instructions accordingly. @zhubonan what's your experience here?
If we switch to Ubuntu rabbitmq, the startup script needs to include `sudo service rabbitmq-server start`.

Two other points:
- With the `visudo`'d file, we should also make sure the instructions are such that the file is _owned by_ and _only editable by_ root. If any user can edit the file, this is a privilege escalation.
- We can add instructions for executing that file on startup with Task Scheduler.

ramirezfranciscof:
Paging @ltalirz and @sphuber who probably also are interested in checking how this turned out. Should @csadorf  maybe give a general overview/demo of how it looks like today in the AiiDA meeting so that @giovannipizzi can take a look at it without having to go through the PR now?

csadorf:
@ramirezfranciscof I've marked this PR as a draft, because the WSL instructions still need to be fixed.

docs/source/intro/get_started.rst
```diff
@@ -10,213 +10,507 @@ An AiiDA installation consists of three core components (plus any external codes
...
+   Does not require the separate installation of prerequisite services.
+   Especially well-suited to get directly started on the **tutorials**.
 
    ---------------
 
-   :fa:`cube,mr-1` **Containers**
+   :fa:`cloud,mr-1` **Run via virtual machine**
 
-   *Use a pre-made image of all the required software.*
+   .. link-button:: intro:get_started:virtual-machine
```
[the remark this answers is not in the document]
csadorf: I think that's a good idea. Do you know if it is possible to open the link in a new tab/window?

docs/source/intro/get_started.rst
```diff
@@ -10,213 +10,507 @@ An AiiDA installation consists of three core components (plus any external codes
...
+   :container: container-lg pb-3
+   :column: col-lg-12 p-2
+
+   **Install prerequisite services**
+
+      AiiDA is designed to run on `Unix <https://en.wikipedia.org/wiki/Unix>`_ operating systems and requires a `bash <https://en.wikipedia.org/wiki/Bash_(Unix_shell)>`_ or `zsh <https://en.wikipedia.org/wiki/Z_shell>`_ shell, and Python >= 3.6.
+
+   .. tabbed:: Ubuntu
+
+      *AiiDA is tested on Ubuntu versions 14.04, 16.04, and 18.04.*
```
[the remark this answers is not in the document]
csadorf: I will just update that now.

csadorf:
> I use the Ubuntu rabbitmq now (I think this didn't work in WSL1) - not sure if we should change the instructions accordingly. @zhubonan what's your experience here?
> If we switch to Ubuntu rabbitmq, the startup script needs to include `sudo service rabbitmq-server start`.
> 
> Two other points:
> 
>  * With the `visudo`'d file, we should also make sure the instructions are such that the file is _owned by_ and _only editable by_ root. If any user can edit the file, this is a privilege escalation.
> 
>  * We can add instructions for executing that file on startup with Task Scheduler.

@greschd Thank you very much, that is already very helpful. Would you potentially be able/willing to directly push an edit to this branch?

docs/source/intro/get_started.rst
```diff
@@ -10,213 +10,507 @@ An AiiDA installation consists of three core components (plus any external codes
...
+   .. tabbed:: pip + venv
+
+      *Install the aiida-core package from PyPI into a virtual environment.*
+
+      Open a terminal and execute:
+
+      .. code-block:: console
+
+          $ python -m pip venv ~/envs/aiida
+          $ source ~/envs/aiida/bin/activate
```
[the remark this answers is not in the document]
csadorf: Done.

zhubonan:
@csadorf @greschd I think on WSL1 the Ubuntu native RabbitMQ does not work. We should definitely differentiate between WSL1 and WSL2. The current instructions seem OK to me for WSL1. 

I have not used WSL2 @greschd. My main concern is that the files system access to the Windows side appears to be much slower in WSL2 than WSL1 and I store my data in the Windows file system. Also, WSL2 seems to block quite a bit RAM? I have WSL2 installed in a different machine but I don't have AiiDA installed there. How did you find the WSL2 experience so far? Maybe there should be some remarks about choosing between the two,

greschd:
Yeah I've had mostly good experiences with WSL2 - but the files I work on with WSL aren't on the Windows drive. Not sure about the RAM issue, I would probably need to look more closely. But this is just my development environment, production is on a separate machine.

The only difference between WSL1/2 in terms of installing would be where RabbitMQ is installed. The "startup via TaskScheduler" would again be the same.

zhubonan:
@greschd I see. For development it should be pretty good, I run both production and development in WSL so disk space is a concern. Also, the growing VHD files for WSL2 will be very big if used for production I would imaging, and that is not easily reclaimable. 
@csadorf I have a feeling that we should warn about potential problems of WSL2 (e.g. disk space) for production?

csadorf:
@zhubonan It seems to me like we should show instructions for WSL 2, but show a disclaimer that there *might be issues* with file system performance in which case we recommend that users should use version 1 (and mention the difference w.r.t. RabbitMQ). Beyond that I think we should be careful about taking on too much responsibility about shortcomings of various systems that are outside of our control.

@greschd Could you help me with the instructions for the scheduler? Do users still need to create the mentioned script and then place it in the TaskScheduler or so?

Both, can you verify that the mentioned issue concerning time zones is still a problem? Reading through the issue comments it might be resolved?

greschd:
Yes, the TaskScheduler would just execute that script at startup; I can write some instructions, but not sure how fool-proof that will be.

I never had the timezone issue AFAICR.

csadorf:
> Yes, the TaskScheduler would just execute that script at startup; I can write some instructions, but not sure how fool-proof that will be.

That would be very much appreciated. I will try to find some additional testers as well.

> I never had the timezone issue AFAICR.

Ok, I'm inclined to remove that.

docs/source/intro/run_docker.rst
```diff
@@ -0,0 +1,61 @@
+.. _intro:get_started:docker:
+.. _intro:install:docker:
+
+****************************
+Run AiiDA via a Docker image
```
chrisjsewell, line 5: This removes a number of helpful commands from the original "Using the Docker image" section, for example the one you need to actual enter the container:

```
docker exec -it --user aiida aiida-container /bin/bash
```

plus the note about using ``/tmp/my_data``, which is the only place writable by the aiida user, and the command to mount it

chrisjsewell (changes requested):

docs/source/intro/run_docker.rst
csadorf: Yeah, I was actually a bit confused about that section. I'll restore the things that got removed, but I also want to avoid making this too much of a general docker tutorial. But I'm slightly less concerned about the length now that the instructions are on their own separate page.

greschd:
@csadorf I added the text but didn't check the formatting - my local install is slightly broken ATM (at least, it can't build the docs). Could you format it to your liking?

csadorf:
> @csadorf I added the text but didn't check the formatting - my local install is slightly broken ATM (at least, it can't build the docs). Could you format it to your liking?
> Feel free to also rephrase it however you see fit - I didn't spend a ton of time on that.

Great, very much appreciated. No worries about the formatting, I will take care of that.

csadorf:
@ramirezfranciscof @chrisjsewell @sphuber This is ready for your review.
@greschd Can you have a last look at the WSL instructions?

sphuber (commented):
Thanks a lot @csadorf . This looks really good, excellent work. I have not been too closely involved in this part of the docs, so I will abstain from a full judgement, but from looking at the built documentation roughly, it looks good. This would be fine for me to merge, but I let the approving up to someone else.

One comment: sometimes a "Hint" box is used where I think "Tip" may be better. For example, on the [system-wide installation page](https://1757-77234579-gh.circle-artifacts.com/0/html/intro/install_system.html#intro-get-started-system-wide-install) the "hint" "Do not start more daemons then there are physical processors on your system." seems more like a tip. Some goes for some of the others. Not sure if there is no "Tip" box or if this decision was deliberate. Anyway, it is a tiny comment and not very important

docs/source/intro/install_system.rst
```diff
@@ -0,0 +1,278 @@
...
+
+          #. Open Task Scheduler.
+
+          #. In the "Actions" menu, click "Create Task".
+
+          #. In "General/Security options", select "Run whether user is logged on or not".
+
+          #. In the "Triggers" tab, click "New...".
+
+             #. In the "Begin the task:" dropdown, select "At startup".
```
greschd, line 112: I've tested this for a bit now, and think it's not 100% bulletproof.. maybe because WSL may not be started when the task is trying to execute. Previously I had used `"At log on"`, which seemed to work fine. One could also set both triggers.

Not sure which is the best one to recommend here 🤔

greschd (approved):
The WSL instructions look good to me, except I'm not 100% sure about the instructions I wrote anymore 😅

docs/source/intro/install_system.rst
csadorf: I say, let's keep it *as is* for now (it's at least not worse than the previous instructions) and we can update them once we know more.

csadorf:
> Thanks a lot @csadorf . This looks really good, excellent work. I have not been too closely involved in this part of the docs, so I will abstain from a full judgement, but from looking at the built documentation roughly, it looks good. This would be fine for me to merge, but I let the approving up to someone else.
> 
> One comment: sometimes a "Hint" box is used where I think "Tip" may be better. For example, on the [system-wide installation page](https://1757-77234579-gh.circle-artifacts.com/0/html/intro/install_system.html#intro-get-started-system-wide-install) the "hint" "Do not start more daemons then there are physical processors on your system." seems more like a tip. Some goes for some of the others. Not sure if there is no "Tip" box or if this decision was deliberate. Anyway, it is a tiny comment and not very important

Thanks a lot! I am actually not 100% clear on the difference between "hint" and "tip". Maybe that's something we should clarify in the wiki? I'd be fine with just using "tip" everywhere instead of "hint".

docs/source/intro/install_system.rst
greschd: Yeah, agreed. If it ends up being a common source of trouble we can always revisit it.

csadorf:
@greschd @chrisjsewell Sorry, your reviews were dismissed when I replaced all 'hints' with 'tips'. This should have been the last commit on this.

sphuber (approved):

csadorf:
@ramirezfranciscof Do you still want to review this or should I merge it?

ramirezfranciscof:
I was going to give it a pass today but got other inconvenients I had to (and still have to) solve. Since I think you already received enough reviews from other people I think it is ok if you want to go ahead and merge it.

csadorf:
> I was going to give it a pass today but got other inconvenients I had to (and still have to) solve. Since I think you already received enough reviews from other people I think it is ok if you want to go ahead and merge it.

Ok, I'll merge it then.

ramirezfranciscof (approved):
Pass given, this looks great!

My only tiny silly comment would be if we could add somewhere in the `System-wide` tile that this is the recommended installation for a work setup. It might be obvious from the text in the tiles that if you don't have admin then you should do the conda-env, but perhaps if you do have it, you may be unsure of wether you should be using it or not (kind of like you shouldn't use "sudo pip install" even if you are sudoer).

Anyway, up to you.

csadorf:
> Pass given, this looks great!
> 
> My only tiny silly comment would be if we could add somewhere in the `System-wide` tile that this is the recommended installation for a work setup. It might be obvious from the text in the tiles that if you don't have admin then you should do the conda-env, but perhaps if you do have it, you may be unsure of wether you should be using it or not (kind of like you shouldn't use "sudo pip install" even if you are sudoer).
> 
> Anyway, up to you.

Like I said in the description of this PR, I don't think it is that easy to recommend one method over the other. It depends on your circumstances, your needs, and your familiarity with certain technologies. That the system-wide method requires admin rights, and the pure-conda method does not should be abundantly clear now. Therefore, in the proposed revision we do *not* recommend any specific method over another, but instead show a slight preference for specific methods via the ordering of the tiles, and furthermore, via this sentence:

> If you are unsure, use the system-wide installation method.

ramirezfranciscof:
Ah, sorry, I totally missed that sentence. Disregard the comment, all GTG.

PR #4501, opened 2020-10-22
Pre-commit: reintroduce pylint rules

ltalirz:
In 65ad067b18cffeb639994efe9a372ec1475e1615 the following pylint rules
were accidentally disabled:

 * missing-class-docstring
 * missing-function-docstring
 * too-many-ancestors
 * too-many-locals

This commit reintroduces all but the "too-many-ancestors" rule, which
was deemed not to be helpful, since it has never prompted refactoring in
the past.

### DIFF
--- a/aiida/manage/manager.py
+++ b/aiida/manage/manager.py
@@ -182,16 +182,7 @@ class Manager:
         if task_prefetch_count is None:
             task_prefetch_count = self.get_config().get_option('daemon.worker_process_slots', profile.name)
 
-        url = profile.get_rmq_url()
         prefix = profile.rmq_prefix
-
-        # This needs to be here, because the verdi commands will call this function and when called in unit tests the
-        # testing_mode cannot be set.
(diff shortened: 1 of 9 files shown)


ltalirz:
@sphuber I started reducing the number of local variables in a few places until I realized that we had increased the acceptable limit to 20 ;-)

aiida/cmdline/commands/cmd_data/cmd_show.py
```diff
@@ -10,6 +10,9 @@
 """
 This allows to manage showfunctionality to all data types.
 """
+import sys
```
greschd, line 13: Could those have been put in the functions to improve startup performance?

ltalirz: done!

you're right, here it makes sense to move imports inside functions (although those system libraries are likely imported somewhere anyhow)

sphuber (approved):
Thanks @ltalirz

Issue #4053, opened 2020-05-06
Callback concept for the progress bar


giovannipizzi:
Rather than wrapping generators returned by internal AiiDA functions in `tqdm` progress bars, I suggest to use a `callback` pattern.
The reason is that, in this way, the internal functions can accept a callback with a pre-decided function signature, that they can call during iterations. This callback parameter would be optional, so if the function is called internally without a callback, there is no printed output.
This is important because the function to export nodes, for instance, is an internal AiiDA function.

As a note, instead, the `aiida.cmdline` module is specific on CLI interaction, so there it would be OK to just wrap tqdm.

To give an idea of what I have in mind, here a possible way to implement a callback pattern:

```python
import random
import time
import tqdm
import contextlib
import functools
import sys

@contextlib.contextmanager
def tqdm_callback():
    def _progress_callback(tqdm_instance, new_processed, total=None):
        if total is not None:
            tqdm_instance.clear()
            tqdm_instance.total = total
            tqdm_instance.refresh()
        if new_processed is not None:
            tqdm_instance.update(new_processed)

    with tqdm.tqdm(total=9e9) as progressbar: # Large number to be later updated, as suggested in tqdm readme
        yielded_function = functools.partial(_progress_callback, progressbar)
        yielded_function.progressbar = progressbar # In this way we make it easy to access it, e.g. for printing
        yield yielded_function

def slow_iteration(num_iter=1000, max_time_per_iter=0.01, callback_every = 1, progress_callback=None):
    def printer(string):
        if progress_callback:
            progress_callback.progressbar.write(string)
        else:
            print(string, file=sys.stderr)

    if progress_callback:
        # Set num_iter
        progress_callback(None, num_iter)

    new_process_count = 0
    for i in range(num_iter):
        time.sleep(random.random()*max_time_per_iter)
        # Call callback with current iteration and total iterations
        if progress_callback:
            if new_process_count % callback_every == 0:
                progress_callback(new_process_count)
                new_process_count = 0
                printer('currently at {}/{}'.format(i, num_iter))
            new_process_count += 1

    if progress_callback and new_process_count:
        progress_callback(new_process_count)

if __name__ == "__main__":
    with tqdm_callback() as callback:
        slow_iteration(num_iter=10, max_time_per_iter=0.1, progress_callback=callback, callback_every=2)
```

[...]

PR #4504, opened 2020-10-23
Refactor archive progress bar

chrisjsewell:
As proposed in https://github.com/aiidateam/aiida-core/issues/4053#issuecomment-714873506, this PR refactors the progress bar implementation in for the archive export to remove the singleton approach and make all calls to the progress bar a context manager.
I also improved some progress bars to give better feedback (in particular the entity field collection).

I've tested this implementation (see below) and find no issues.

As an added bonus, I have also added a verbose option to the CLI (-v), such that this will set the progress bars to be left on screen (i.e. `leave=True`)

### DIFF
--- a/aiida/cmdline/commands/cmd_export.py
+++ b/aiida/cmdline/commands/cmd_export.py
@@ -100,12 +100,10 @@ def create(
     their provenance, according to the rules outlined in the documentation.
     You can modify some of those rules using options of this command.
     """
-    from tqdm import tqdm
     from aiida.common.log import override_log_formatter_context
-    from aiida.common.progress_reporter import set_progress_reporter
+    from aiida.common.progress_reporter import set_progress_bar_tqdm
     from aiida.tools.importexport import export, ExportFileFormat, EXPORT_LOGGER
     from aiida.tools.importexport.common.exceptions import ArchiveExportError
-    from aiida.tools.importexport.common.config import BAR_FORMAT
 
     entities = []
 
@@ -143,7 +141,7 @@ def create(
         export_format = ExportFileFormat.TAR_GZIPPED
 
     if verbosity in ['DEBUG', 'INFO']:
-        set_progress_reporter(tqdm, bar_format=BAR_FORMAT, leave=(verbosity == 'DEBUG'))
+        set_progress_bar_tqdm(leave=(verbosity == 'DEBUG'))
     EXPORT_LOGGER.setLevel(verbosity)
 
     try:
--- a/aiida/cmdline/commands/cmd_export.py
+++ b/aiida/cmdline/commands/cmd_export.py
@@ -67,7 +67,13 @@ def inspect(archive, version, data, meta_data):
 @options.NODES()
 @options.ARCHIVE_FORMAT()
 @options.FORCE(help='overwrite output file if it already exists')
-@options.VERBOSE(help='Do not remove progress bars after a process completes')
+@click.option(
+    '-v',
+    '--verbosity',
+    default='INFO',
+    type=click.Choice(['DEBUG', 'INFO', 'WARNING', 'CRITICAL']),
+    help='Control the verbosity of console logging'
+)
 @options.graph_traversal_rules(GraphTraversalRules.EXPORT.value)
 @click.option(
     '--include-logs/--exclude-logs',
@@ -84,7 +90,7 @@ def inspect(archive, version, data, meta_data):
 @decorators.with_dbenv()
 def create(
     output_file, codes, computers, groups, nodes, archive_format, force, input_calc_forward, input_work_forward,
-    create_backward, return_backward, call_calc_backward, call_work_backward, include_comments, include_logs, verbose
+    create_backward, return_backward, call_calc_backward, call_work_backward, include_comments, include_logs, verbosity
 ):
     """
     Export subsets of the provenance graph to file for sharing.
@@ -97,7 +103,7 @@ def create(
     """
     from tqdm import tqdm
     from aiida.common.progress_reporter import set_progress_reporter
-    from aiida.tools.importexport import export, ExportFileFormat
+    from aiida.tools.importexport import export, ExportFileFormat, EXPORT_LOGGER
     from aiida.tools.importexport.common.exceptions import ArchiveExportError
     from aiida.tools.importexport.common.config import BAR_FORMAT
 
@@ -136,7 +142,9 @@ def create(
     elif archive_format == 'tar.gz':
         export_format = ExportFileFormat.TAR_GZIPPED
 
-    set_progress_reporter(partial(tqdm, bar_format=BAR_FORMAT, leave=verbose))
+    if verbosity in ['DEBUG', 'INFO']:
+        set_progress_reporter(partial(tqdm, bar_format=BAR_FORMAT, leave=(verbosity == 'DEBUG')))
+    EXPORT_LOGGER.setLevel(verbosity)
 
     try:
         export(entities, filename=output_file, file_format=export_format, **kwargs)
--- a/aiida/cmdline/commands/cmd_export.py
+++ b/aiida/cmdline/commands/cmd_export.py
@@ -9,7 +9,6 @@
 ###########################################################################
 # pylint: disable=too-many-arguments,import-error,too-many-locals
 """`verdi export` command."""
-from functools import partial
 import os
 import tempfile
 
@@ -144,7 +143,7 @@ def create(
         export_format = ExportFileFormat.TAR_GZIPPED
 
     if verbosity in ['DEBUG', 'INFO']:
-        set_progress_reporter(partial(tqdm, bar_format=BAR_FORMAT, leave=(verbosity == 'DEBUG')))
+        set_progress_reporter(tqdm, bar_format=BAR_FORMAT, leave=(verbosity == 'DEBUG'))
     EXPORT_LOGGER.setLevel(verbosity)
 
     try:
--- a/aiida/cmdline/commands/cmd_export.py
+++ b/aiida/cmdline/commands/cmd_export.py
@@ -67,7 +67,7 @@ def inspect(archive, version, data, meta_data):
 @options.NODES()
 @options.ARCHIVE_FORMAT()
 @options.FORCE(help='overwrite output file if it already exists')
-@options.VERBOSE()
+@options.VERBOSE(help='Do not remove progress bars after a process completes')
 @options.graph_traversal_rules(GraphTraversalRules.EXPORT.value)
 @click.option(
     '--include-logs/--exclude-logs',
--- a/aiida/cmdline/commands/cmd_export.py
+++ b/aiida/cmdline/commands/cmd_export.py
@@ -9,7 +9,7 @@
 ###########################################################################
 # pylint: disable=too-many-arguments,import-error,too-many-locals
 """`verdi export` command."""
-
+from functools import partial
 import os
 import tempfile
 
@@ -95,8 +95,11 @@ def create(
     their provenance, according to the rules outlined in the documentation.
     You can modify some of those rules using options of this command.
     """
+    from tqdm import tqdm
+    from aiida.common.progress_reporter import set_progress_reporter
     from aiida.tools.importexport import export, ExportFileFormat
     from aiida.tools.importexport.common.exceptions import ArchiveExportError
+    from aiida.tools.importexport.common.config import BAR_FORMAT
 
     entities = []
 
@@ -133,8 +136,10 @@ def create(
     elif archive_format == 'tar.gz':
         export_format = ExportFileFormat.TAR_GZIPPED
 
+    set_progress_reporter(partial(tqdm, bar_format=BAR_FORMAT, leave=verbose))
+
     try:
-        export(entities, filename=output_file, file_format=export_format, verbose=verbose, **kwargs)
+        export(entities, filename=output_file, file_format=export_format, **kwargs)
     except ArchiveExportError as exception:
         echo.echo_critical(f'failed to write the archive file. Exception: {exception}')
     else:
--- a/aiida/cmdline/commands/cmd_export.py
+++ b/aiida/cmdline/commands/cmd_export.py
@@ -102,6 +102,7 @@ def create(
     You can modify some of those rules using options of this command.
     """
     from tqdm import tqdm
+    from aiida.common.log import override_log_formatter_context
     from aiida.common.progress_reporter import set_progress_reporter
     from aiida.tools.importexport import export, ExportFileFormat, EXPORT_LOGGER
     from aiida.tools.importexport.common.exceptions import ArchiveExportError
@@ -147,7 +148,8 @@ def create(
     EXPORT_LOGGER.setLevel(verbosity)
 
     try:
-        export(entities, filename=output_file, file_format=export_format, **kwargs)
+        with override_log_formatter_context('%(message)s'):
+            export(entities, filename=output_file, file_format=export_format, **kwargs)
     except ArchiveExportError as exception:
         echo.echo_critical(f'failed to write the archive file. Exception: {exception}')
     else:
--- a/aiida/cmdline/commands/cmd_export.py
+++ b/aiida/cmdline/commands/cmd_export.py
@@ -67,6 +67,7 @@ def inspect(archive, version, data, meta_data):
 @options.NODES()
 @options.ARCHIVE_FORMAT()
 @options.FORCE(help='overwrite output file if it already exists')
+@options.VERBOSE()
 @options.graph_traversal_rules(GraphTraversalRules.EXPORT.value)
 @click.option(
     '--include-logs/--exclude-logs',
@@ -83,7 +84,7 @@ def inspect(archive, version, data, meta_data):
 @decorators.with_dbenv()
 def create(
     output_file, codes, computers, groups, nodes, archive_format, force, input_calc_forward, input_work_forward,
-    create_backward, return_backward, call_calc_backward, call_work_backward, include_comments, include_logs
+    create_backward, return_backward, call_calc_backward, call_work_backward, include_comments, include_logs, verbose
 ):
     """
     Export subsets of the provenance graph to file for sharing.
@@ -133,7 +134,7 @@ def create(
         export_format = ExportFileFormat.TAR_GZIPPED
 
     try:
-        export(entities, filename=output_file, file_format=export_format, **kwargs)
+        export(entities, filename=output_file, file_format=export_format, verbose=verbose, **kwargs)
     except ArchiveExportError as exception:
         echo.echo_critical(f'failed to write the archive file. Exception: {exception}')
     else:
--- a/aiida/common/log.py
+++ b/aiida/common/log.py
@@ -220,9 +220,7 @@ def override_log_formatter_context(fmt: str):
     Be aware! This may fail if the number of handlers is changed within the decorated function/method.
     """
     temp_formatter = logging.Formatter(fmt=fmt)
-    cached_formatters = []
-    for handler in AIIDA_LOGGER.handlers:
-        cached_formatters.append(handler.formatter)
+    cached_formatters = [handler.formatter for handler in AIIDA_LOGGER.handlers]
 
     for handler in AIIDA_LOGGER.handlers:
         handler.setFormatter(temp_formatter)
--- a/aiida/common/log.py
+++ b/aiida/common/log.py
@@ -212,6 +212,27 @@ def override_log_level(level=logging.CRITICAL):
         logging.disable(level=logging.NOTSET)
 
 
+@contextmanager
+def override_log_formatter_context(fmt: str):
+    """Temporarily use a different formatter for all handlers.
+
+    NOTE: One can _only_ set `fmt` (not `datefmt` or `style`).
+    Be aware! This may fail if the number of handlers is changed within the decorated function/method.
+    """
+    temp_formatter = logging.Formatter(fmt=fmt)
+    cached_formatters = []
+    for handler in AIIDA_LOGGER.handlers:
+        cached_formatters.append(handler.formatter)
+
+    for handler in AIIDA_LOGGER.handlers:
+        handler.setFormatter(temp_formatter)
+
+    yield
+
+    for index, handler in enumerate(AIIDA_LOGGER.handlers):
+        handler.setFormatter(cached_formatters[index])
+
+
 def override_log_formatter(fmt: str):
     """Temporarily use a different formatter for all handlers.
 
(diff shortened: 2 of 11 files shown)


chrisjsewell:
Oct-23-2020 07-41-45

ltalirz:
Hi @chrisjsewell, before going through the code in detail, here is a question:

There is an open PR https://github.com/aiidateam/aiida-core/pull/3896 to homogenize verbosity handling across the whole verdi cli.
As is common with loggers, the approach taken there is to configure a `VERDI_LOGGER` singleton instance, which then has the advantage that one can simply `VERDI_LOGGER.debug(...)` or `VERDI_LOGGER.warn(...)`. 

I am wondering whether a similar approach would not work for the progress bar as well.
Is it really necessary to pass the instance around to every little sub function?
And by making the progress bar a *required* argument we force anyone who would like to use these functions to set up a fake progress bar just to call it.

chrisjsewell:
> I am wondering whether a similar approach would not work for the progress bar as well.
> Is it really necessary to pass the instance around to every little sub function?
> And by having the progress bar as a required argument we force anyone who would like to use these functions to set up a fake progress bar just to call it.

Eurh, I think exactly the opposite; I think singleton's (and using mutable global variables in general) are awful and should only be used when absolutely necessary. I would point you, for example, towards https://stackoverflow.com/a/138012/5033292

chrisjsewell:
if anything, the private functions in this class + the `export` function should be turned into a class, with a shared state

ltalirz:
> I think singleton's (and using mutable global variables in general) are awful and should only be used when absolutely necessary.

I think, as with many questions, there is a balance to be found.
Of course I'm aware of the drawbacks of global variables - at the same time, passing every variable around every time can hurt code readability.

Using loggers is a standard approach suggested in the official python tutorial https://docs.python.org/3/howto/logging.html#configuring-logging
Would you suggest that we start passing loggers around explicitly in aiida-core?

In any case, I agree that the progress bar (which is around only temporarily and specific to this section of hte code) differs from loggers (which are assumed to be there all the time), so perhaps we will need to pass it around in the end.

> if anything, the private functions in this class + the `export` function should be turned into a class, with a shared (local) state.
> I would be happy to make this change

Yes, a more localized way of sharing state would make sense.
Whatever the solution, it would be helpful to have an obvious way of using functions that *can* interact with the progress bar also in scenarios where one doesn't have one at hand.

chrisjsewell:
> Would you suggest that we start passing loggers around explicitly in aiida-core?

Well loggers are one of them rare use cases. Although still it can be a real pain to ensure the behaviour of logging, for example sphinx sets up a global logger which makes debugging failed tests a real annoyance (docutils/sphinx also have some other singletons which are horrible to override, and one of the reasons for my hatred of them lol).

chrisjsewell:
>  it would be helpful to have an obvious way of using functions that can interact with the progress bar also in scenarios where one doesn't have a progress bar at hand

Just to note, none of the public functions require that a progress bar be specifically passed

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -131,23 +134,35 @@ def total_time(self) -> float:
...
         """An archive writer
 
         :param filename: the filename (possibly including the absolute path)
             of the file on which to export.
+        :param progress_context: A context manager for creating and updating a progress bar
+
         """
         # pylint: disable=unused-argument
         self._filename = filename
+        self._progress_context = progress_context or partial(tqdm, bar_format=BAR_FORMAT)
```
chrisjsewell, line 154:
```suggestion
        self._progress_context = progress_context or partial(tqdm, disable=True)
```

perhaps the default should be to not print anything?

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -131,23 +134,35 @@ def total_time(self) -> float:
...
 class ArchiveWriterAbstract(ABC):
     """An abstract interface for AiiDA archive writers."""
 
-    def __init__(self, filename: str, **kwargs: Any):
+    def __init__(self, filename: str, progress_context: Optional[ProgressContext] = None, **kwargs: Any):
         """An archive writer
 
         :param filename: the filename (possibly including the absolute path)
             of the file on which to export.
+        :param progress_context: A context manager for creating and updating a progress bar
```
ltalirz, line 149: maybe add an example of what one could provide here (`tqdm.tqdm`) and what is expected of this progress bar (which methods does it need to have, ...)

aiida/tools/importexport/dbexport/__init__.py
ltalirz: right, that's the question of the dummy logger...
I don't have a clear answer at the moment; perhaps we should look around a bit what others have found to work best in such cases.

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -299,6 +309,18 @@ class WriterJsonTar(ArchiveWriterAbstract):
     The entire containing folder is then compressed as a tar file.
     """
 
+    def __init__(self, filename: str, progress_context: Optional[ProgressContext] = None, **kwargs: Any):
+        """A writer for zipped archives.
+
+        :param filename: the filename (possibly including the absolute path)
+            of the file on which to export.
+        :param progress_context: A context manager for creating and updating a progress bar
```
ltalirz, line 317: also here expanding the description would make semse

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -299,6 +309,18 @@ class WriterJsonTar(ArchiveWriterAbstract):
...
+        :param sandbox_in_repo: Create the temporary uncompressed folder within the aiida repository
```
ltalirz, line 318: should this be made an explicit argument then?

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -272,21 +284,19 @@ def file_format_verbose(self) -> str:
     def export_version(self) -> str:
```
ltalirz, line 284: should this move to the abstract class?

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -725,83 +739,81 @@ def _get_starting_node_ids(entities: List[Any], silent: bool) -> Tuple[DefaultDi
...
+        return entities_starting_set, given_node_entry_ids
+
+    with progress_context(desc='Collecting chosen entities', total=total) as progress_bar:
+        for entry in entities:
+
+            # This returns the class name (as in imports). E.g. for a model node:
+            # aiida.backends.djsite.db.models.DbNode
+            # entry_class_string = get_class_string(entry)
+            # Now a load the backend-independent name into entry_entity_name, e.g. Node!
+            # entry_entity_name = schema_to_entity_names(entry_class_string)
```
ltalirz, line 757: I guess we can actually remove this (?)
At least it does not really seem to make sense any longer in this context

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -862,109 +872,116 @@ def _collect_entity_queries(
...
-        builder.append(
-            orm.Log,
-            filters={'dbnode_id': {
-                'in': node_ids_to_be_exported
-            }},
-            project='uuid',
-        )
-        res = set(builder.all(flat=True))
-        given_log_entry_ids.update(res)
+    total = 1 + ((1 if include_logs else 0) + (1 if include_logs else 0) if node_ids_to_be_exported else 0)
```
ltalirz, line 885: that should probably be
```suggestion
    total = 1 + ((1 if include_logs else 0) + (1 if include_comments else 0) if node_ids_to_be_exported else 0)
```

ltalirz: by the way, if there are no node ids to be exported, what is the `1` for?

ltalirz (changes requested):
thanks @chrisjsewell 

only minor comments

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: The idea is that the abstract class is publicly available, such that (in principle) anyone can make their own archiver by implementing a writer and reader (to come) which would have their own versions.
The `EXPORT_VERSION` in aiida I feel is specific only to the current JSON archivers.
(still have to look into how this relates to migrations though)

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: removed

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: agree, but the problem I had is that mypy then complains about the signature (same for `use_compression` in the zip writer). I'll see if there is a workaround, or just use `type: ignore`.

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: yeh this is a bit of a pain. Ideally this "description" would be provided entirely via the typing (i.e. `PROGRESS_CONTEXT`). In theory this can be done by replacing `tdqm` with a `Protocol` class, but this is only available natively from python 3.8 or by the backport https://pypi.org/project/typing-extensions/. So I'm not sure what the feeling is about adding yet another dependency (albeit that it will naturally phase itself out)

For full completness you would also need https://pypi.org/project/mypy-extensions/, to use https://mypy.readthedocs.io/en/stable/additional_features.html#extended-callable-types.

Something like

```python
class ProgressBar(Protocol):
   def set_description_str(self, text: str, refresh: bool = True):
      pass
   def update(self, n: int = 1):
      pass

ProgressContext = Callable[[DefaultArg(str, 'desc'), DefaultArg(int, 'total')], ContextManager[ProgressBar]]

```

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -131,23 +134,35 @@ def total_time(self) -> float:
         return (self.time_write_stop or self.time_collect_stop) - self.time_collect_start
 
 
+# should make more generic
```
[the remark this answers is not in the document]
chrisjsewell: see below

aiida/tools/importexport/dbexport/__init__.py
ltalirz: From my point of view, a textual description will suffice fully here.
I think adding dependencies just for this one instance would be overkill; we can decide to do this at a later point.

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: oops, yeh I was having an issue without the 1, but I will re-check

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: done

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: changed

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: changed

aiida/cmdline/commands/cmd_export.py
```diff
@@ -67,6 +67,7 @@ def inspect(archive, version, data, meta_data):
 @options.NODES()
 @options.ARCHIVE_FORMAT()
 @options.FORCE(help='overwrite output file if it already exists')
+@options.VERBOSE()
```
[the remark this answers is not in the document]
chrisjsewell: changed

chrisjsewell:
@ltalirz inspired by the discussion in #3896, I have re-re-factored in a54c319!

Now the progress bar is once again a singleton, but now it is fully generalised to just the concept of a progress reporter. As such it is not specific to importexport and so I have added its code as `aiida/common/progress_reporter.py`.

Control of setting the progress reporter is now moved to the CLI (perhaps it could somehow be integrated into #3896?), the exporter functions just retrieve it, with the default being a "null" reporter.

For now aiida/tools/importexport/common/progress_bar.py remains, but in the import refactor, this should be removed

aiida/cmdline/commands/cmd_export.py
```diff
@@ -132,6 +136,8 @@ def create(
     elif archive_format == 'tar.gz':
         export_format = ExportFileFormat.TAR_GZIPPED
 
+    set_progress_reporter(partial(tqdm, bar_format=BAR_FORMAT, leave=verbose))
```
chrisjsewell, line 139: the CLI now sets the progress reporter, not the `export` function

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -395,7 +389,7 @@ def get_writer(file_format: str) -> Type[ArchiveWriterAbstract]:
             f'Can only export in the formats: {tuple(writers.keys())}, please specify one for "file_format".'
         )
 
-    return writers[file_format]
+    return cast(Type[ArchiveWriterAbstract], writers[file_format])
```
chrisjsewell, line 392: this is required due to the differing __init__ signatures of the different implementations (with specific keywords like `use_compression`)

chrisjsewell:
Note there is some code to set/reset the logging level in the `export` function, which could maybe be removed in #3896

chrisjsewell:
In 19da2ad I have also moved control of the logger level to the CLI, so it should now be very easy to add in the #3896 code later

chrisjsewell:
Oct-24-2020 04-00-43

aiida/common/log.py
```diff
@@ -212,6 +212,27 @@ def override_log_level(level=logging.CRITICAL):
...
 
+@contextmanager
+def override_log_formatter_context(fmt: str):
+    """Temporarily use a different formatter for all handlers.
+
+    NOTE: One can _only_ set `fmt` (not `datefmt` or `style`).
+    Be aware! This may fail if the number of handlers is changed within the decorated function/method.
+    """
+    temp_formatter = logging.Formatter(fmt=fmt)
+    cached_formatters = []
```
ltalirz, line 223: could use list comprehension here

aiida/common/progress_reporter.py
```diff
@@ -0,0 +1,111 @@
...
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+# pylint: disable=global-statement,unused-argument
+"""Provide a singleton progress reporter implementation.
```
ltalirz, line 11: Do you have a specific use case in mind that is not a progress bar?

If not, instead of "progress reporter" I would prefer the term "progress bar", which everybody immediately understands rather than trying to be more generic for a use case that we may never have.

A related question is: do you think there is a need to be generic with respect to the progress bar implementation, rather than hardcoding `tqdm`?

I certainly get the point of making the progress bar available not just in the import/export but in `aiida.common` but I'm wondering a bit whether all this boilerplate code is needed.
In the end, it might be a case of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) ;-)

ltalirz (commented):
Hi @chrisjsewell , thanks for putting more thought into this an coming up with this solution.

I have some questions concerning the interface before finishing the review; will get back to the laptop later tonight

aiida/common/progress_reporter.py
chrisjsewell: Of the top of my head, maybe you want to report progress via a web based application (like aiidalab) or in a GUI. These would not use tqdm, and there is also nothing in the interface that requires it to be a bar.

I'm not that bothered about changing to bar, but also I really don't see it as that big an issue. I'm sure people can grasp the idea of a progress reporter.

I am absolutely definitely against hard coding tqdm. For starters, the CLI already requires that it be a partial tqdm.

The major problem with the import/export (and moving to a new format) is that half the stuff is hard-coded. I would much rather have YAGNI (within reason), if it means you don't have to later refactor everything.
Also, its hardly that much boilerplate, the file has less code than the original progress_bar.py one.

aiida/common/log.py
chrisjsewell: yep will do (I just copy/pasted the code from `override_log_formatter`)

aiida/common/progress_reporter.py
chrisjsewell: Also talking about YAGNI in aiida-core, where we keep around two database backends (django/sqla) for really no reason now (since JSONB was implemented in django), is a bit ironic 😜

aiida/common/log.py
```diff
@@ -212,6 +212,27 @@ def override_log_level(level=logging.CRITICAL):
...
+    for handler in AIIDA_LOGGER.handlers:
+        cached_formatters.append(handler.formatter)
```
chrisjsewell, line 225:
```suggestion
    cached_formatters = [handler.formatter for handler in AIIDA_LOGGER.handlers]
```

aiida/common/progress_reporter.py
ltalirz: > Also talking about YAGNI in aiida-core, where we keep around two database backends (django/sqla) for really no reason now (since JSONB was implemented in django), is a bit ironic 😜

Well, you know that I am very much in favor of taking a very hard look at dropping one of the two - to be fair, though, there was a clear use case when sqla was introduced.

As for the progress reporter/bar, I don't want to be the one to impose my opinion and since we want to get this merged, I'm ok with keeping it.
It is true that the code is relatively compact; it's just perhaps not very obvious how to use it (will add another comment in the final review)

aiida/common/progress_reporter.py
```diff
@@ -0,0 +1,111 @@
...
+"""
+from contextlib import contextmanager
+from functools import partial
+from typing import Any, Callable, ContextManager, Iterator, Optional
+
+__all__ = ('get_progress_reporter', 'set_progress_reporter', 'progress_reporter_base', 'ProgressIncrementerBase')
+
+
+class ProgressIncrementerBase:
+    """A base class for incrementing a progress reporter."""
```
ltalirz, line 28: Could you expand in the docstring a bit on the actual usage of this base class?

If I understand correctly, it is a dummy interface class that does nothing and specifically designed to be replaced by tqdm in practice (?).

aiida/common/progress_reporter.py
```diff
@@ -0,0 +1,111 @@
...
+    def update(self, n: int = 1):  # pylint: disable=invalid-name
+        """Update the progress counter.
+
+        :param n: Increment to add to the internal counter of iterations
+
+        """
+
+
+@contextmanager
+def progress_reporter_base(*,
```
ltalirz, line 47: I haven't implemented my own context managers yet, so this might be a stupid question but:
 what is the purpose of these function arguments if they aren't passed to anything?

aiida/common/progress_reporter.py
```diff
@@ -0,0 +1,111 @@
...
+                           total: int,
+                           desc: Optional[str] = None,
+                           **kwargs: Any) -> Iterator[ProgressIncrementerBase]:
+    """A context manager for providing a progress reporter for a process.
```
ltalirz, line 51: Can you add here that this progress reporter is a dummy that does not print anything?

ltalirz (changes requested):
Thanks @chrisjsewell ; mostly requests for documentation + one simplification

aiida/common/progress_reporter.py
chrisjsewell: Oh they are just not being used here because it is a dummy/null implementation, but in an actual implementation e.g. tqdm, they would do something a bit like:

```python
@contextmanager
def progress_reporter_base(total, desc):
   progress_bar = setup_progress_bar(total, desc)
   yield ProgressIterator(progress_bar)
   progress_bar.teardown()
```

aiida/common/progress_reporter.py
```diff
@@ -0,0 +1,111 @@
...
+
+    Example Usage::
+
+        with progress_reporter(total=10, desc="A process:") as progress:
```
chrisjsewell, line 55:
```suggestion
        with progress_reporter_base(total=10, desc="A process:") as progress:
```

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -414,6 +408,21 @@ def export(
...
 
+    Note, the logging level and progress reporter should be set externally, for example::
+
+        from functools import partial
+        from tqdm import tqdm
+        from aiida.common.progress_reporter import set_progress_reporter
+        from aiida.tools.importexport.common.config import BAR_FORMAT
+
+        EXPORT_LOGGER.setLevel('DEBUG')
+        set_progress_reporter(partial(tqdm, bar_format=BAR_FORMAT, leave=True))
```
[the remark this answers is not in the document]
chrisjsewell: added

chrisjsewell:
@ltalirz in 1ab807a I have re-written the progress reporter contextmanager as a single class.
It works exactly the same as the decorated function, but I hope is a bit more easy to understand.

I have also added the `set_progress_bar_tqdm` convenience function

aiida/tools/importexport/common/progress_bar.py
```diff
@@ -15,7 +15,7 @@
 
 from aiida.common.lang import type_check
 
-from aiida.tools.importexport.common.config import BAR_FORMAT
+from aiida.common.progress_reporter import TQDM_BAR_FORMAT as BAR_FORMAT
```
ltalirz, line 18: The progress_bar.py file will be removed when the import refactoring PR is merged, correct?

ltalirz (approved):
Thanks @chrisjsewell , at least to me this version of the progress reporter is much more obvious to interact with.

Let's merge this :-)
Edit: After the tests pass ;-)

aiida/tools/importexport/common/progress_bar.py
chrisjsewell: yep thats the intention 👍

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -512,66 +521,54 @@ def export(
...
-            except (exceptions.ArchiveExportError, LicensingException) as exc:
-                if os.path.exists(filename):
-                    os.remove(filename)
-                raise exc
+    extract_time = report_data['time_collect_stop'] - report_data['time_collect_start']
+    EXPORT_LOGGER.debug(f'Data extracted in {extract_time:6.2g} s.')
 
-            write_time = report_data['time_write_stop'] - report_data['time_write_start']
-            EXPORT_LOGGER.debug(f'Data written in {write_time:6.2g} s.')
+    if export_data is None:
```
chrisjsewell, line 549:
```suggestion
    if export_data is not None:
```

fix accidental change (whilst testing)

ltalirz (approved):

PR #4508, opened 2020-10-24
Docs: Add  `verdi process status` command to tutorial

### LINKED ISSUE #4505: Docs: use `verdi process status` in tutorial
The `verdi process status` command is very useful to debug runs of WorkChains, since it helps pinpointing where the original exception occurred - e.g.

```
$ verdi process status 596175
ZeoppMultistageDdecWorkChain<596175> Excepted [2:run_zeopp_after]
    ├── NetworkCalculation<596176> Finished [0]
    ├── get_structure_from_cif<596184> Finished [0]
    └── Cp2kMultistageDdecWorkChain<596186> Excepted [1:run_cp2kddec]
        └── Cp2kMultistageWorkChain<596187> Excepted [None]
            └── check_resize_unit_cell_legacy<596188> Finished [0]
```
This command is currently mentioned only once in the documentation (in [topics/processes/usage](https://aiida-core.readthedocs.io/en/latest/topics/processes/usage.html#verdi-process-status) ).

One reason why this currently isn't mentioned in the tutorial may be because the tutorial relies on `verdi graph generate` and `verdi process status` is relevant mainly for debugging.
I'm not saying that we need to introduce a bug in the example workchain that they need to fix (that would make things too long), but in my experience, debugging workchain runs is one of the most likely things new users need to do and we should  make sure that they know the `verdi process status` command.

I propose to add this command to the tutorial at the end of https://aiida-core.readthedocs.io/en/latest/intro/tutorial.html#workflows as another quick way to look at the tree of a workchain for debugging purposes.

Thoughts @mbercx @csadorf ?

mbercx:
Fixes #4505 

Currently the `verdi process status` command is only mentioned once in the documentation (in [topics/processes/usage](https://aiida-core.readthedocs.io/en/latest/topics/processes/usage.html#verdi-process-status) ). Since this command is both useful for getting a nice succinct overview of a work chain and debugging, it deserves a mention in the basic tutorial.

This PR adds the command just before generating the final provenance graph.

### DIFF
--- a/docs/source/intro/tutorial.rst
+++ b/docs/source/intro/tutorial.rst
@@ -570,7 +570,7 @@ Depending on which step the workflow is running, you should get something like t
 We can see that the ``MultiplyAddWorkChain`` is currently waiting for its *child process*, the ``ArithmeticAddCalculation``, to finish.
 Check the process list again for *all* processes (You should know how by now!).
 After about half a minute, all the processes should be in the ``Finished`` state.
-You can also see a nice overview of the processes called by the work chain with the ``verdi process status`` command:
+The ``verdi process status`` command prints a *hierarchical* overview of the processes called by the work chain:
 
 .. code-block:: console
 
--- a/docs/source/intro/tutorial.rst
+++ b/docs/source/intro/tutorial.rst
@@ -570,6 +570,17 @@ Depending on which step the workflow is running, you should get something like t
 We can see that the ``MultiplyAddWorkChain`` is currently waiting for its *child process*, the ``ArithmeticAddCalculation``, to finish.
 Check the process list again for *all* processes (You should know how by now!).
 After about half a minute, all the processes should be in the ``Finished`` state.
+You can also see a nice overview of the processes called by the work chain with the ``verdi process status`` command:
+
+.. code-block:: console
+
+    $ verdi process status 19
+    MultiplyAddWorkChain<19> Finished [0] [3:result]
+        ├── multiply<20> Finished [0]
+        └── ArithmeticAddCalculation<22> Finished [0]
+
+The bracket ``[3:result]`` indicates the current step in the outline of the :py:class:`~aiida.workflows.arithmetic.multiply_add.MultiplyAddWorkChain` (step 3, with name ``result``).
+The ``process status`` is particularly useful for debugging complex work chains, since it helps pinpoint where a problem occurred.
 
 We can now generate the full provenance graph for the ``WorkChain`` with:
 


docs/source/intro/tutorial.rst
```diff
@@ -570,6 +570,16 @@ Depending on which step the workflow is running, you should get something like t
 We can see that the ``MultiplyAddWorkChain`` is currently waiting for its *child process*, the ``ArithmeticAddCalculation``, to finish.
 Check the process list again for *all* processes (You should know how by now!).
 After about half a minute, all the processes should be in the ``Finished`` state.
+You can also see a nice overview of the steps the work chain with the ``verdi process status`` command:
```
sphuber, line 573:
```suggestion
You can also see a nice overview of the steps of the work chain with the ``verdi process status`` command:
```
Small correction aside, I am not sure if I would phrase it this way. In the context the word "steps" is often used to refer to the steps of the outline. However, what the command mostly shows is the "call stack", or the hierarchy of process calls. Now this term is probably too technical too use, but we could find a better simpler explanation. Note that the outline steps do enter in the output, but they are in between the square brackets at the end of each line. It shows the current step of each work chain, or the last step, in the case the work chain has finished.

docs/source/intro/tutorial.rst
```diff
@@ -570,6 +570,16 @@ Depending on which step the workflow is running, you should get something like t
...
+
+.. code-block:: console
+
+    $ verdi process status 19
+    MultiplyAddWorkChain<19> Finished [0] [3:result]
+        ├── multiply<20> Finished [0]
+        └── ArithmeticAddCalculation<22> Finished [0]
+
+This command can also be very useful when debugging a work chain.
```
sphuber, line 582: Why? Might be good to give a bit more detail

sphuber (changes requested):
Thanks @mbercx . Have two comments

docs/source/intro/tutorial.rst
```diff
@@ -570,6 +570,16 @@ Depending on which step the workflow is running, you should get something like t
 We can see that the ``MultiplyAddWorkChain`` is currently waiting for its *child process*, the ``ArithmeticAddCalculation``, to finish.
 Check the process list again for *all* processes (You should know how by now!).
 After about half a minute, all the processes should be in the ``Finished`` state.
+You can also see a nice overview of the steps the work chain with the ``verdi process status`` command:
+
+.. code-block:: console
+
+    $ verdi process status 19
+    MultiplyAddWorkChain<19> Finished [0] [3:result]
```
mbercx, line 578: One question here: what does the `[3:result]` represent? I see that it prints the `stepper_state_info` by digging a bit through the code, but it isn't clear to me what the stepper state is. 😅

sphuber: it follows the "steps through the outline" if you check the outline of the `MultiplyAddWorkChain` you will see there are three steps, the third of which is called `result`. The stepper_state_info also correctly takes while loops and if conditions into account

docs/source/intro/tutorial.rst
mbercx: How about: "You can also see a nice overview of the processes called by the work chain with the `verdi process status` command:"? And then explain after the code snippet that the square brackets at the end of each work chain line represents the step in the outline.

docs/source/intro/tutorial.rst
mbercx: How about (also considering your other comment [above](https://github.com/aiidateam/aiida-core/pull/4508#discussion_r511509657)):

> Between the final brackets next to the `MultiplyWorkChain` you can also see the current step in the outline (`[3:result]`). This information combined with the process state can be very useful when debugging a work chain.

ltalirz: Yes, let's simply replace this sentence with

```
The bracket ``[3:result]`` indicates the current step in the outline of the :py:class:`~...MultiplyWorkChain` (step 3, with name `result`).
The ``process status`` is particularly useful for debugging complex WorkChains, since it helps pinpoint where a problem occurred.

File: docs/source/intro/tutorial.rst [hunk 4]
```diff
@@ -570,6 +570,17 @@ Depending on which step the workflow is running, you should get something like t
 We can see that the ``MultiplyAddWorkChain`` is currently waiting for its *child process*, the ``ArithmeticAddCalculation``, to finish.
 Check the process list again for *all* processes (You should know how by now!).
 After about half a minute, all the processes should be in the ``Finished`` state.
+You can also see a nice overview of the processes called by the work chain with the ``verdi process status`` command:
```
ltalirz, line 573:
```suggestion
The ``verdi process status`` command prints a *hierarchical* overview of the processes called by the work chain:
```

I would usually leave it up to the user to decide whether the view is nice, but that's just a matter of taste.

ltalirz (dismissed):
thanks @mbercx  - looks good to me.

don't know whether you want to wait for more input before merging

ltalirz (dismissed):

mbercx:
Thanks @ltalirz! Will still give @sphuber a chance to comment before merging.

csadorf (approved):
LGTM.

sphuber (commented):
All good for me. I was just wondering if we should add that it is recursive, meaning that if one of the called processes itself also calls subprocesses, those will also be shown recursively. This might not be obvious since the example only has a single level

mbercx:
> All good for me. I was just wondering if we should add that it is recursive, meaning that if one of the called processes itself also calls subprocesses, those will also be shown recursively. This might not be obvious since the example only has a single level

I think it's ok to leave it as it is. As you said, the example doesn't demonstrate the recursion, and I think it will be clear if they use this command on a work chain where one of the processes also calls subprocesses.

sphuber (dismissed):
Good to go

mbercx:
Had to rebase to `develop`, which dismissed the reviews. I always do this by rebasing on my work station and force-pushing to my repo branch. Is that the recommended approach?

sphuber:
> Had to rebase to `develop`, which dismissed the reviews. I always do this by rebasing on my work station and force-pushing to my repo branch. Is that the recommended approach?

If the commits of the PR will be squashed anyway, then this is not necessary. You can simply use the "Update branch" button of Github to merge `develop` into the branch. That will not dismiss the reviews, and so is the preferred option. For PRs that contain multiple commits that *have* to be kept, and so either merge-commit or rebase-merge strategy will be used, in that case one has to rebase locally and force push. The PR will have to be reaccepted in this case, but that is actually a good thing, since it will have to check the rebase was done correctly

sphuber (approved):

mbercx:
Right, we weren't running all the checks for the documentation PR's. How do I get GitHub to run all the checks so I can merge the PR?

sphuber:
We cannot, because the checks re explicitly skipped, so we can only merge with admin rights, bypassing the checks. This is not ideal. If you could look, maybe with @chrisjsewell , how the skipping can be done within the action, such that it still runs, but it just shortcuts if the unit tests don't actually have to run. This would be the better solution, because then the PRs will also not show up as "still running" with the yellow circle but will get the green checkmark

PR #4510, opened 2020-10-24
Refactor Import Archive

chrisjsewell:
FYI @ltalirz this is my initial (incomplete) implementation of the import code (that builds on #4504) 

Mainly see `aiida/tools/importexport/dbimport/readers.py` and `tests/tools/importexport/test_reader.py`


tests/tools/importexport/test_reader.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_json_zip():
+    """Test the JSON zip reader."""
+    with ReaderJsonZip(PATH) as reader:
+        assert reader.metadata.export_version == '0.9'
+        assert reader.metadata.aiida_version == '1.4.2'
+        assert sorted(reader.entity_names) == ['Comment', 'Computer', 'Group', 'Log', 'Node', 'User']
+        assert reader.entity_count('Node') == 40065
+        assert sum(1 for _ in reader.iter_entity_fields('Node')) == 40065
```
ltalirz, line 23: just wondering - can one not just do something like `len(list(reader.iter_entity_fields('Node')))`?

aiida/tools/importexport/dbimport/utils.py
```diff
@@ -218,6 +218,8 @@ def deserialize_attributes(attributes_data, conversion_data):
 
 def deserialize_field(key, value, fields_info, import_unique_ids_mappings, foreign_ids_reverse_mappings):
     """Deserialize field using deserialize attributes"""
+    if key in ('attributes', 'extras'):
```
ltalirz, line 221: was this broken before?
is there not test for this?

aiida/tools/importexport/dbimport/readers.py
```diff
@@ -0,0 +1,328 @@
...
+    if zipfile.is_zipfile(in_path):
+        return 'zip'
+    raise ImportValidationError(
+        'Unable to detect the input file format, it is neither a '
+        'tar file, nor a (possibly compressed) zip file.'
+    )
+
+
+@dataclasses.dataclass
+class ExportMetadata:
```
ltalirz, line 46: do we need to replicate this or can it be reused between import and export?

aiida/tools/importexport/dbimport/readers.py
```diff
@@ -0,0 +1,328 @@
...
+    # Entity type -> database ID key
+    unique_identifiers: Dict[str, str] = dataclasses.field(repr=False)
+    # Entity type -> database key -> meta parameters
+    all_fields_info: Dict[str, Dict[str, Dict[str, str]]] = dataclasses.field(repr=False)
+    aiida_version: str
+    export_version: str
+
+
+@decorator
+def ensure_within_context(wrapped, instance, args, kwargs):
```
ltalirz, line 62: this seems like a widely applicable decorator - can it be made available more centrally?

tests/tools/importexport/test_reader.py
chrisjsewell: In general the `sum` method here is better for getting the length of an iterator, because with the `len(list` approach you have to read the entire contents of the iterator into memory

aiida/tools/importexport/dbimport/utils.py
chrisjsewell: So before, the attributes and extras fields were appended to the data after this deserialization. This was really just due to the implementation detail of the current archive format, whereby the node attributes and extras are "special cased" and stored in separate sections of the `data.json`.

In `ReaderJsonZip.iter_entity_fields`, you will see that I re-merge these fields into the fields dict, before yielding it.

I guess rather than having this test here, ideally we would change `aiida/tools/importexport/common/config.py::get_all_fields_info` to include these fields in the returned dict.
This would though necessitate a migration, to update this dict where it is stored in the `0.9`  archive (in `metadata.json`).

One thing I should mention here actually, is that really I want to eventually do a "second round" of refactoring of `import_data_dj`/`import_data_sqla`, to find a way for the process to actually make use of the `iter_entity_fields` iterator, and not have to read all the entities and their fields into memory at the same time.
This would not affect the current archive format, because you have to read the whole of `data.json` into memory anyway. But obviously with a new format this does not necessarily have to be the case.

aiida/tools/importexport/dbimport/readers.py
chrisjsewell: no indeed ideally these could/will be merged. I have just copied it here whilst I am working out the reader interface.

aiida/tools/importexport/dbimport/readers.py
chrisjsewell: yeh possibly, although obviously it is only applicable to classes that set/unset this `_in_context` attribute in their `__enter__`/`__exit__` methods

tests/tools/importexport/test_reader.py
ltalirz: Thanks for pointing this out, makes sense!

aiida/tools/importexport/dbimport/utils.py
ltalirz: > I guess rather than having this test here, ideally we would change `aiida/tools/importexport/common/config.py::get_all_fields_info` to include these fields in the returned dict.
> This would though necessitate a migration, to update this dict where it is stored in the `0.9` archive (in the `all_fields_info` key of `metadata.json`).

Ok. I will anyhow need a migration for the group extras, so before merging this, I could give you write access to my fork to add this

> One thing I should mention here actually, is that really I want to eventually do a "second round" of refactoring of `import_data_dj`/`import_data_sqla`, to find a way for the process to actually make use of the `iter_entity_fields` iterator, and not have to read all the entities and their fields into memory at the same time.
> This would not affect the current archive format, because you have to read the whole of `data.json` into memory anyway. But obviously with a new format this does not necessarily have to be the case.

Right. I agree it's good to hold this off for a later round/PR (unless it just "falls out" naturally).

aiida/tools/importexport/dbimport/readers.py
ltalirz: I see. I believe something similar was already added here in a different context:
https://github.com/aiidateam/aiida-core/blob/02248cf3686a0ab89faf1625e0da24d9e33d8cde/aiida/orm/nodes/node.py#L39

aiida/tools/importexport/dbimport/utils.py
chrisjsewell: > Ok. I will anyhow need a migration for the group extras, so before merging this, I could give you write access to my fork to add this

yep makes sense to do it in a single migration. I guess we want to give these fields a new `converter_type`, that basically tells `deserialize_attributes` to do nothing to the value.
Something like:

```python
def deserialize_attributes(attributes_data, conversion_data):
    """Deserialize attributes"""
    import datetime
    import pytz

   if conversion_data == "null":
       return attributes_data
```

aiida/tools/importexport/dbimport/readers.py
chrisjsewell: yep I recall conversation about that ta 👍 
I guess it is potatas/potatas on whether to add this as deorator or a call to a method (maybe a call to a method is better though, because it does not have any limitations e.g. for `@property`)

I actually copied the decorator from `aiida/tools/importexport/common/archive.py:Archive` (note this class also needs to be deprecated, to replaced by `ReaderJsonZip` etc). There it was used as a method on the class, but this did not work with class inheritance

chrisjsewell: I'm not sure how we want to go about the deprecation of `aiida/tools/importexport/common/archive.py` and `aiida/tools/importexport/common/progress_bar.py`. I guess they need to be left in for aiida v1.5.0. possibly you can put a deprecation warning in the module itself, i.e. to trigger when the module is imported, or you put them in all the contained functions/classes

aiida/tools/importexport/dbimport/utils.py
chrisjsewell: > Right. I agree it's good to hold this off for a later round/PR 

yep indeed 👍

FYI, this will also include probably passing `ArchiveData.node_data` to the writer as a query or generator for node extras/attributes (as opposed to the current iterator), so that it can perform a chunked `query.iterall(batch_size=x)` write into the archive and potentially (with a new format) not have to read all that data into memory at the same time.
(which I reckon is the largest source of data from the database).

aiida/tools/importexport/dbimport/readers.py
```diff
@@ -0,0 +1,328 @@
...
+    def _extract(self, path_prefix: str = ''):
+        """Extract repository data to a temporary folder.
+
+        :param path_prefix: Only extract paths starting with this prefix.
+
+        :raises TypeError: if parameter types are not respected
+        """
+        try:
+            with zipfile.ZipFile(self.filename, 'r', allowZip64=True) as handle:
+                for membername in handle.namelist():
```
chrisjsewell, line 238: @ltalirz note this, together with the `node_repository` method, is a bit of a performance regression for the current archive format, because you have to iterate through all of `handle.namelist()` each time you want to extract/return a node repository, e.g. with

```python
for uuid in reader.iter_node_uuids():
   folder = reader.node_repository(uuid)
```

you are looping through `handle.namelist()` for every uuid (but still only performing the same amount of extractions).

The original implementation unpacked the entire zip folder in one loop.
This is more efficient for a full import, but has the massive drawback for general reads, whereby anytime you want to retrieve a single piece of data or a single node repo from the archive, you have to unpack the whole archive.

Could possibly add a "bulk extract" method, but not quite sure how this would work with the abstract interface and also the progress reporter.

chrisjsewell: maybe have a `prepare_for_import` method, which is described as "Prepare the archive for a full import". This would be optional to call and implementation specific (defaulting to doing nothing), but in this case would perform the full unpacking in one loop

chrisjsewell: I might also add a `count_objects` method that e.g. for zip just returns the length of `handle.namelist()`. Might be nice for introspection.

aiida/tools/importexport/dbimport/readers.py
ltalirz: > I'm not sure how we want to go about the deprecation of `aiida/tools/importexport/common/archive.py` and `aiida/tools/importexport/common/progress_bar.py`.
> I guess they need to be left in for aiida v1.5.0.
> Possibly you can put a deprecation warning in the module itself, i.e. to trigger when the module is imported, or you put them in all the contained functions/classes

In terms of import depth, I think both modules are far enough down not to be considered part of the official public API 
(I was never quite sure of the exact rules; I've just opened a PR to clarify them https://github.com/aiidateam/aiida-core/pull/4511).

In any case, I care more about *actual* uses in the wild, and I highly doubt that the progress bar was used by any plugin (I've never seen/heard of this). I.e. I'd be fine with removing `aiida/tools/importexport/common/progress_bar.py` in 1.5.0

For the `export_tar`,... functions in `archive.py` I know that they actually have been used in the wild, so here we need to provide a deprecation path. A deprecation warning at the module level can work as long as it is possible to provide a warning message that clearly explains what to do for any of the deprecated functionality (otherwise it's better to keep one per function).

For the `Archive` class, I have personally never seen it used and I'm having a tough time imagining anyone using it outside aiida-core, although it is possible. A [quick github search](https://github.com/search?p=2&q=aiida+Archive+unpack&type=Code) did not turn up anything.
Here I would leave it up to you if you want to leave it in.

aiida/tools/importexport/dbimport/readers.py
chrisjsewell: > is a bit of a performance regression

Although actually, if you are re-importing an archive, it's a lot more performant, because you only have to unpack new nodes 🤷 (this is probably the less usual case though?)

ltalirz: > this, together with the `node_repository` method, is a bit of a performance regression for the current archive format

I suggest you first check how much of a performance regression we're talking about here.
If we're talking about 10% that's one thing - if we're talking about 100% or more that's another.

You could try e.g. the archive from here: https://archive.materialscloud.org/record/2020.107

It might make a difference whether this is done on a SSD or a magnetic disk - I can do a test on a magnetic disk later if you like.

> Although actually, if you are re-importing an archive, it's a lot more performant, because you only have to unpack new nodes 🤷 (this is probably the less usual case though?)

It is a less usual case but it is still relevant (not so much re-importing the same archive, but importing an updated archive that contains a lot of data that was already imported).

chrisjsewell: I've actually fixed this now (commit coming soon..), to change `node_repository` to `iter_node_repos` and it first computes the common prefix of all the required node repositories and only extracts that; so performant for both the single and many node use cases 😄

aiida/tools/importexport/dbimport/readers.py
chrisjsewell: I've changed the decorator to be a method

aiida/tools/importexport/dbimport/utils.py
ltalirz: > Ok. I will anyhow need a migration for the group extras,

Just wanted to mention that Casper pointed out that, since the group extras are a simple addition (and we don't offer back-migrations), a migration is actually not needed to support group extras, i.e. we could also defer this migration to a later point.

aiida/tools/importexport/common/archive.py
```diff
@@ -304,36 +61,55 @@ def extract_zip(infile, folder, nodes_export_subfolder=None, **kwargs):
...
     """
     # pylint: disable=fixme
+
     if nodes_export_subfolder:
         if not isinstance(nodes_export_subfolder, str):
             raise TypeError('nodes_export_subfolder must be a string')
     else:
         nodes_export_subfolder = NODES_EXPORT_SUBFOLDER
 
+    if not kwargs.get('silent', False):
```
ltalirz, line 71: In the export, the progress is now being set by the CLI command.
Is there a reason to take a different approach here?

aiida/tools/importexport/common/zip_folder.py
```diff
@@ -0,0 +1,152 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Export a zip-file."""
```
ltalirz, line 10: Could you add some explanation why this is necessary for zip but not for tar?

aiida/tools/importexport/common/zip_folder.py
```diff
@@ -0,0 +1,152 @@
...
+# pylint: disable=missing-docstring,redefined-builtin
+import io
+import os
+import zipfile
+
+
+class MyWritingZipFile:
```
ltalirz, line 17: Not sure this name is self-explanatory. Could you add some docstrings to explain what this does?

aiida/tools/importexport/common/zip_folder.py
```diff
@@ -0,0 +1,152 @@
...
+        self.open()
+        return self
+
+    def __exit__(self, type, value, traceback):
+        self.close()
+
+
+class ZipFolder:
+    """
+    To improve: if zipfile is closed, do something
```
ltalirz, line 47: The current docstring explains what remains to be fixed rather than explaining the purpose/usage of the class

ltalirz: Also this class has many functions without docstrings

aiida/tools/importexport/common/zip_folder.py
```diff
@@ -0,0 +1,152 @@
...
+        return self._pwd
+
+    def open(self, fname, mode='r'):
+        if mode == 'w':
+            return MyWritingZipFile(zip_file=self._zipfile, fname=self._get_internal_path(fname))
+        # else
+        return self._zipfile.open(self._get_internal_path(fname), mode)
+
+    def _get_internal_path(self, filename):
+        return os.path.normpath(os.path.join(self.pwd, filename))
```
ltalirz, line 104: shouldn't we be using pathlib?
(several other uses of os.path in this class)

ltalirz (commented):
hi @chrisjsewell , I've reviewed up to importexport/common/zip_folder.py

before I continue, it would be great if you could rebase this PR on `develop`.
In the meanwhile, I will test importing stuff and check the timings

ltalirz:
Tests:

- archive with 25.8k links, 9.4k nodes
- django backend

New implementation:

- Importing chris.aiida into empty profile: 3m25s
- Importing chris.aiida again into same profile: 14s
- Importing chris.aiida into new empty profile: 2m12s
    - ~60% of the time spent importing links
    - Note: it's not entirely clear in which way the disk has "warmed up" here... I'll just check again
- Importing chris.aiida into another new profile: 2m7s

Old implementation:

- Importing chris.aiida into empty profile: 3m24s
- Importing chris.aiida into same profile: 40s
- Importing chris.aiida into new empty profile: 3m21s

Conclusions

- The new implementation is faster both on first import and, for obvious reasons, for re-import
- In both the old and the new implementation, the import of links is slow.
In the new implementation, importing links now dominates total import time (~60%).
This can probably be optimized at a later point.

ltalirz:
Memory usage during import:

Old implementation
image

New implementation
image


Also looks fine (new implementation uses a tiny bit less).

aiida/tools/importexport/common/zip_folder.py
chrisjsewell: FYI I simply moved this file from its original location `aiida/tools/importexport/dbexport/zip.py`, without change. But no problem, I can address the review comments here and below

Also, I left the original file there with a deprecation warning, but can fully remove if you think this is not necessary for back-compatibility.

ltalirz: I personally don't expect that `aiida/tools/importexport/dbexport/zip.py` was used directly by plugins, so for me it's ok to remove.

aiida/cmdline/commands/cmd_export.py
```diff
@@ -40,21 +40,35 @@ def inspect(archive, version, data, meta_data):
...
+                echo.echo_dictionary(dataclasses.asdict(reader.metadata))
             else:
-                info = archive_object.get_info()
-                data = sorted([(k.capitalize(), v) for k, v in info.items()])
-                data.extend(sorted([(k.capitalize(), v) for k, v in archive_object.get_data_statistics().items()]))
-                echo.echo(tabulate.tabulate(data))
+                statistics = {
+                    'Version aiida': reader.metadata.aiida_version,
+                    'Version format': reader.metadata.export_version,
+                    'Computers': reader.entity_count('Computer'),
```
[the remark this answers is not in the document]
chrisjsewell: Well this would be dependent on the actual implementation reader, but ideally yes. The current JSON one caches only the `data.json` read, then always calls `len(data['export_data']['name'])`

aiida/tools/importexport/archive/common.py
```diff
@@ -0,0 +1,54 @@
...
+import dataclasses
+import os
+from typing import Dict, List, Optional, Set
+
+__all__ = ('ArchiveMetadata', 'detect_archive_type')
+
+
+@dataclasses.dataclass
+class ArchiveMetadata:
+    """Class for storing metadata about this archive."""
```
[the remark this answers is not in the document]
chrisjsewell: the required fields are those that are strictly necessary for recovering the archive, i.e. they are actually used during the import process.
The others are just "nice to have" and describe the parameters used to create the archive. I don't think they should default to the CLI, because this is misleading in implying that these values are contained within the archive. The CLI defaults could change between aiida versions and this also assumes that the archive was created via the CLI which is not strictly necessary.

aiida/tools/importexport/archive/readers.py
```diff
@@ -0,0 +1,509 @@
...
+
+    @abstractmethod
+    def iter_node_repos(self,
+                        uuids: Iterable[str],
+                        progress: bool = True,
+                        description='Iterating node repos') -> Iterator[Folder]:
+        """Yield temporary folders containing the contents of the repository for each node.
+
+        :param uuids: The node UUIDs to yield
+        :param progress: report progress
```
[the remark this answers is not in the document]
chrisjsewell: as you can see, in the current implementation this is a two stage process; first to ensure the folders are unzipped to the temporary folder, then to actually yield the subfolders.
Both require separate progress bars, and so it is not possible to simply wrap this iterator in a (single) external progress reporter

aiida/tools/importexport/archive/readers.py
```diff
@@ -0,0 +1,509 @@
...
+        return EXPORT_VERSION
+
+    def __enter__(self):
+        super().__enter__()
+        self._sandbox = SandboxFolder(self._sandbox_in_repo)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._sandbox.erase()
+        self._sandbox: Optional[SandboxFolder] = None
```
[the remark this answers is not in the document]
chrisjsewell: I think this was perhaps due to an issue with mypy failing. Will check if removing it still passes

aiida/tools/importexport/archive/readers.py
```diff
@@ -0,0 +1,509 @@
...
+            'include_comments': export_parameters.get('include_comments', None),
+            'include_logs': export_parameters.get('include_logs', None),
+            'conversion_info': metadata.get('conversion_info', [])
+        }
+        try:
+            return ArchiveMetadata(**output)
+        except TypeError as error:
+            raise CorruptArchive(f'Metadata invalid: {error}')
+
+    def entity_count(self, name: str) -> int:
```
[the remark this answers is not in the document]
chrisjsewell: As mentioned on slack; It’s because they inherit their docstrings from the abstract base class. I would argue that they should not have their own docstrings, since this would essentially be code duplication and would allow the docstring to become out of sync with the base class

aiida/tools/importexport/archive/readers.py
```diff
@@ -0,0 +1,509 @@
...
+        data = self._get_data().get('export_data', {}).get(name, {})
+        return len(data)
+
+    @property
+    def link_count(self) -> int:
+        return len(self._get_data()['links_uuid'])
+
+    def iter_entity_fields(self,
```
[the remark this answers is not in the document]
chrisjsewell: see above

aiida/tools/importexport/archive/writers.py
```diff
@@ -0,0 +1,301 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
chrisjsewell: yes just moved, nothing to see here 👀

aiida/tools/importexport/archive/writers.py
```diff
@@ -0,0 +1,301 @@
...
+        :param export_data: The data to export
+        :param get_node_repo: A callable: UUID -> to a folder of the node's repository
+        :returns: A dictionary of data about the write process
+
+        """
+
+
+def _write_to_json_archive(
+    folder: Union[Folder, ZipFolder], export_data: ArchiveData, export_version: str, get_node_repo: Callable[[str],
+                                                                                                             Folder]
```
chrisjsewell, line 107: This is why I hate yapf 😬, it does such a poor job at handling long lines, which black is so much better at

aiida/tools/importexport/common/archive.py
chrisjsewell: this was simply for back compatibility. these functions are no longer actually used for read/write, only in the archive migrations where the CLI has not yet been changed.
When I refactor that (I think now in a separate PR) I will probably remove this

aiida/tools/importexport/archive/common.py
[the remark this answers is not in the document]
ltalirz: Makes sense!
Could you add a short version of this to the docstring?

aiida/tools/importexport/dbimport/backends/django.py
```diff
@@ -0,0 +1,689 @@
...
+    # Store new links
+    if links_to_store:
+        IMPORT_LOGGER.debug('   (%d new links...)', len(links_to_store))
+
+        models.DbLink.objects.bulk_create(links_to_store, batch_size=batch_size)
+    else:
+        IMPORT_LOGGER.debug('   (0 new links...)')
+
+
+def _import_groups(
```
ltalirz, line 664:
```suggestion
def _store_groups(
```

aiida/tools/importexport/dbimport/backends/django.py
```diff
@@ -0,0 +1,689 @@
...
+    *, group_count: int, group_uuids: Iterable[Tuple[str, Set[str]]], foreign_ids_reverse_mappings: Dict[str, Dict[str,
+                                                                                                                   int]]
+):
+    """Import groups."""
```
ltalirz, line 668:
```suggestion
    """Store groups in the database."""
```

aiida/tools/importexport/dbimport/backends/django.py
```diff
@@ -0,0 +1,689 @@
...
+    from aiida.backends.djsite.db import models
+
+    if not group_count:
+        return
+
+    pbar_base_str = 'Groups - '
+
+    with get_progress_reporter()(total=group_count, desc=pbar_base_str) as progress:
+        for groupuuid, groupnodes in group_uuids:
+            if not groupnodes:
```
ltalirz, line 678: I saw a similar logic elsewhere - it's not really clear to me why empty groups should be skipped?
In particular now that groups can store extras, one might well want to import an empty group

aiida/tools/importexport/dbimport/backends/sqla.py
```diff
@@ -0,0 +1,681 @@
+# -*- coding: utf-8 -*-
```
ltalirz, line 1: I'm going to mostly skip this file (I guess most comments apply both to the django/sqla version)

tests/tools/importexport/migration/test_migration.py
```diff
@@ -104,20 +106,24 @@ def test_migrate_recursively_specific_version(self):
         """Test the `version` argument of the `migrate_recursively` function."""
         filepath_archive = get_archive_file('export_v0.3_simple.aiida', **self.core_archive)
 
-        with Archive(filepath_archive) as archive:
+        reader_cls = get_reader(detect_archive_type(filepath_archive))
+        with reader_cls(filepath_archive) as archive:
+
+            metadata = deepcopy(archive._get_metadata())  # pylint: disable=protected-access
```
ltalirz, line 112: Can't you just use
```suggestion
            metadata = deepcopy(archive.metadata)
```
?

tests/tools/importexport/migration/test_migration_array.py
```diff
@@ -36,11 +36,12 @@ def migration_data(request):
 
     filepath_archive = get_archive_file(f'export_v{version_old}_simple.aiida', filepath='export/migrate')
 
-    with Archive(filepath_archive) as archive:
-        metadata_old = copy.deepcopy(archive.meta_data)
-        data_old = copy.deepcopy(archive.data)
+    reader_cls = get_reader(detect_archive_type(filepath_archive))
+    with reader_cls(filepath_archive) as archive:
+        metadata_old = copy.deepcopy(archive._get_metadata())  # pylint: disable=protected-access
```
ltalirz, line 41: same here

tests/tools/importexport/test_reader.py
```diff
@@ -0,0 +1,81 @@
...
+
+from aiida.common import InvalidOperation
+from aiida.tools.importexport import EXPORT_VERSION, CorruptArchive
+from aiida.tools.importexport.archive import get_reader
+from tests.utils.archives import get_archive_file
+
+
+@pytest.fixture
+def archive_reader():
+    """Return an initiated archive reader."""
```
ltalirz, line 22: a short documentation of example use would make sense here I think

tests/tools/importexport/test_reader.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def _func(filename=None):
+        archive_path = get_archive_file(filename or f'export_v{EXPORT_VERSION}_simple.aiida', 'export/migrate')
```
ltalirz, line 25: not sure you really want to default to a particular export file here; I think it's easier to understand if one needs to  pass the filename explicitly
(e.g. in the test below you're checking things that are very specific to this export file)

tests/tools/importexport/test_reader.py
```diff
@@ -0,0 +1,81 @@
...
+        return reader_cls(archive_path)
+
+    return _func
+
+
+def test_reader(archive_reader):
+    """Test reader API."""
+
+    with archive_reader() as archive:
+        assert archive.export_version == '0.9'
```
ltalirz, line 36: looking a the filename of the file, this could be replaced by `EXPORT_VERSION` (otherwise one has to always fix it here).

ltalirz (changes requested):
@chrisjsewell I'm basically through now - looks all good; only minor comments.

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -0,0 +1,99 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+# pylint: disable=protected-access,fixme,too-many-arguments,too-many-locals,too-many-statements,too-many-branches,too-many-nested-blocks
```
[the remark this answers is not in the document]
chrisjsewell: removed all

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -0,0 +1,99 @@
...
+        # Get an unique name for the import group, based on the current (local) time
+        basename = timezone.localtime(timezone.now()).strftime('%Y%m%d-%H%M%S')
+        counter = 0
+        group_label = basename
+
+        while Group.objects.find(filters={'label': group_label}):
+            counter += 1
+            group_label = f'{basename}_{counter}'
+
+            if counter == 100:
```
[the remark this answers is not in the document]
chrisjsewell: done

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -0,0 +1,99 @@
...
+                raise exceptions.ImportUniquenessError(
+                    "Overflow of import groups (more than 100 import groups exists with basename '{}')"
```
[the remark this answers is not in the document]
chrisjsewell: lol fixed

tests/tools/importexport/test_reader.py
chrisjsewell: no the idea is to get the latest/current export version and check against that.
The reader should always be compatible with EXPORT_VERSION, then the migrator would be used to "get" to this version

tests/tools/importexport/test_reader.py
chrisjsewell: done

tests/tools/importexport/test_reader.py
chrisjsewell: done

tests/tools/importexport/migration/test_migration_array.py
chrisjsewell: see above

tests/tools/importexport/migration/test_migration.py
chrisjsewell: no because `.metadata` returns a dataclass (validated against the latest version), but here we want the "raw" file contents.
Really we shouldn't be using the reader class here, which intended to be used only with the latest vesion of the archive, rather a separate "migrator" class that can handle all versions.
But that is for another PR 😉

aiida/tools/importexport/dbimport/backends/django.py
chrisjsewell: This function does not actually store the groups (that is done in `_store_entity_data`), it just adds the nodes to the groups. I will change the name/docstring of the function to reflect that

aiida/tools/importexport/dbimport/backends/sqla.py
chrisjsewell: yep 👍

tests/tools/importexport/test_reader.py
ltalirz: well, I'm just saying that your reader is defaulting to a very specific export file from the test data, which is not at all obvious from the name of the fixture `archive_reader`.

all I'm saying is that it makes sense to pass the filename of the export file explicitly

aiida/tools/importexport/dbimport/backends/django.py
```diff
@@ -0,0 +1,689 @@
...
+                if entity_name == NODE_ENTITY_NAME:
+                    # format extras
+                    fields = _sanitize_extras(fields)
+                    if extras_mode_new != 'import':
+                        fields.pop('extras', None)
+                new_entries[entity_name][str(pk)] = fields
+
+
+def _store_entity_data(
+    *, reader: ArchiveReaderAbstract, entity_name: str, comment_mode: str, extras_mode_existing: str,
```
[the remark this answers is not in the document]
chrisjsewell: IMO you should always use a `*` before keyword arguments, i.e. enforcing the use of `func(a=1, b=2)` as opposed to `func(1, 2)` (and this should be implemented more in aiida). It makes the function more resistant to refactoring, because calls to the function will not accidentally retrieve the wrong arguments if they have been re-ordered or removed.

aiida/tools/importexport/dbimport/backends/django.py
```diff
@@ -0,0 +1,689 @@
...
+        )
+
+    # Summarize import
+    result_summary(ret_dict, getattr(group, 'label', None))
+
+    return ret_dict
+
+
+def _generate_entity_data(
+    *, entity_name: str, reader: ArchiveReaderAbstract, new_entries: Dict[str, Dict[str, dict]],
```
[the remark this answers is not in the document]
chrisjsewell: I think you are misunderstanding this `*`; it is not the same as `*args`, it enforces that all calls to the function must use keyword arguments for all the arguments after the `*`, i.e. `func(a=1, b=2)` instead of `func(1, 2)`.
Definitely start using it in your own code 😄 

(see below also)

aiida/tools/importexport/dbimport/backends/django.py
```diff
@@ -0,0 +1,689 @@
...
+            }
+
+        # count total number of entities to import
+        number_of_entities: int = sum(reader.entity_count(entity_name) for entity_name in entity_order)
+        IMPORT_LOGGER.debug('Importing %s entities', number_of_entities)
+
+        ###########################################
+        # IMPORT ALL DATA IN A SINGLE TRANSACTION #
+        ###########################################
+        from django.db import transaction  # pylint: disable=import-error,no-name-in-module
```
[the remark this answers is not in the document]
chrisjsewell: indeed

aiida/tools/importexport/dbimport/backends/django.py
```diff
@@ -0,0 +1,689 @@
...
+    :raises `~aiida.tools.importexport.common.exceptions.CorruptArchive`: if the provided archive at ``in_path`` is
+        corrupted.
+    :raises `~aiida.tools.importexport.common.exceptions.IncompatibleArchiveVersionError`: if the provided archive's
+        export version is not equal to the export version of AiiDA at the moment of import.
+    :raises `~aiida.tools.importexport.common.exceptions.ArchiveImportError`: if there are any internal errors when
+        importing.
+    :raises `~aiida.tools.importexport.common.exceptions.ImportUniquenessError`: if a new unique entity can not be
+        created.
+    """
+    # Initial check(s)
```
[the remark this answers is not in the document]
chrisjsewell: yep 👍 literally the only difference in this top level function, to the sqla one, is that `transaction.atomic()` is replaced by `sql_transaction()` and you have to pass `batch_size` to some funcs in django and `session` to some funcs in sqla

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -0,0 +1,99 @@
...
+        # If not a valid UUID
+        return False
+
+    # Check if there was any kind of conversion of the hex during
+    # the validation
+    return str(parsed_uuid) == given_uuid
+
+
+def _sanitize_extras(fields):
+    """Remove unwanted extra keys."""
```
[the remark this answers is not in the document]
chrisjsewell: done

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -0,0 +1,99 @@
...
+            progress.update()
+            nodes.append(entry[0])
+
+        group.add_nodes(nodes)
+        progress.set_description_str('Done (cleaning up)', refresh=True)
+
+    return group
+
+
+def _validate_uuid(given_uuid):
```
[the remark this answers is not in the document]
chrisjsewell: done

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -0,0 +1,99 @@
...
+    existing = existing_entries.get(NODE_ENTITY_NAME, {})
+    existing_pk = [foreign_ids_reverse_mappings[NODE_ENTITY_NAME][v['uuid']] for v in existing.values()]
+    new = new_entries.get(NODE_ENTITY_NAME, {})
+    new_pk = [foreign_ids_reverse_mappings[NODE_ENTITY_NAME][v['uuid']] for v in new.values()]
+
+    pks_for_group = existing_pk + new_pk
+
+    # So that we do not create empty groups
+    if not pks_for_group:
+        IMPORT_LOGGER.debug('No Nodes to import, so no Group created, if it did not already exist')
```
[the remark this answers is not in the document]
chrisjsewell: done

aiida/tools/importexport/archive/common.py
```diff
@@ -0,0 +1,54 @@
...
+    # Entity type -> UUID list
+    entities_starting_set: Optional[Dict[str, Set[str]]] = dataclasses.field(default=None)
+    include_comments: Optional[bool] = dataclasses.field(default=None)
+    include_logs: Optional[bool] = dataclasses.field(default=None)
+    # list of migration event notifications
+    conversion_info: List[str] = dataclasses.field(default_factory=list, repr=False)
+
+
+def detect_archive_type(in_path: str) -> str:
+    """For back-compatibility, but should be replaced with direct specification of classes."""
```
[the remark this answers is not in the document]
ltalirz: still to do?

aiida/tools/importexport/archive/readers.py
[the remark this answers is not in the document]
ltalirz: What was the result?

aiida/tools/importexport/common/archive.py
ltalirz: I'll keep this open in case you ever look back to this PR

aiida/tools/importexport/archive/readers.py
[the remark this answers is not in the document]
chrisjsewell: oh yeh I removed this 👍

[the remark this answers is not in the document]
chrisjsewell: oops no I didn't just doing now..

aiida/tools/importexport/archive/common.py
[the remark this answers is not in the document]
chrisjsewell: oops yeh just added

tests/tools/importexport/test_reader.py
chrisjsewell: I have anyhow documented that it defaults to the latest archive

aiida/tools/importexport/common/zip_folder.py
chrisjsewell: Err, I'm going to be lazy here and say not now. I have anyhow allowed for the filename in the __init__ to be a Path and thats the main one

ltalirz (dismissed):
Thanks a lot for the dedicated effort @chrisjsewell !

The import/export has definitely been one of the most "smelly" subpackages of AiiDA for a long time; it's really great to see it return to a stage where others will feel comfortable improving it as well.

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -0,0 +1,99 @@
...
+from aiida.tools.importexport.common import exceptions
+from aiida.tools.importexport.common.config import NODE_ENTITY_NAME
+from aiida.tools.importexport.dbimport.utils import IMPORT_LOGGER
+
+
+def _make_import_group(
+    *, group, existing_entries: Dict[str, Dict[str, dict]], new_entries: Dict[str, Dict[str, dict]],
+    foreign_ids_reverse_mappings: Dict[str, Dict[str, int]]
+):
+    """Make an import group containing all imported nodes."""
```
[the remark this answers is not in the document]
chrisjsewell: just added

chrisjsewell:
> Thanks a lot for the dedicated effort @chrisjsewell !

and thanks for the prompt reviews!
Good team work; this is how you implement a PR and don't let it stagnate like some of the currently open ones 😜

ltalirz (approved):

PR #4511, opened 2020-10-25
Docs: clarify public API

### LINKED ISSUE #4558: Docs: clarify public API

The definition of what is the [public API of AiiDA](https://aiida-core.readthedocs.io/en/latest/reference/api/public.html) should be clarified a bit (partly addressed by draft PR https://github.com/aiidateam/aiida-core/pull/4511)

However, while looking into this, we realized that not only was the [list of public resources](https://aiida-core.readthedocs.io/en/latest/reference/api/public.html) wildly out of date; it seems that there's been unintentional "bubbling up" of resources to the second nesting level (and therefore into the public API).

For example, all arguments and options are exposed at the `aiida.cmdline` level, mixing the namespaces of these modules in an uncontrolled way. There is both a `CALCULATION` option and a  `CALCULATION` argument. Which is exposed at `aiida.cmdline` is a matter of luck of in which order someone decided to place the imports in the `__init__.py`:
```
In [4]: from aiida.cmdline.params.arguments import CALCULATION

In [5]: type(CALCULATION)
Out[5]: aiida.cmdline.params.arguments.overridable.OverridableArgument

In [6]: from aiida.cmdline.params.options import CALCULATION

In [7]: type(CALCULATION)
Out[7]: aiida.cmdline.params.options.overridable.OverridableOption

In [8]: from aiida.cmdline import CALCULATION

In [9]: type(CALCULATION)
Out[9]: aiida.cmdline.params.options.overridable.OverridableOption
```

Almost all second-level packages expose way more resources than indicated in the [list](https://aiida-core.readthedocs.io/en/latest/reference/api/public.html)
```
In [10]: import aiida.schedulers

In [11]: dir(aiida.schedulers)
Out[11]:
['JobInfo',
 'JobResource',
 'JobState',
 'JobTemplate',
 'MachineInfo',
 'NodeNumberJobResource',
 'ParEnvJobResource',
 'Scheduler',
 'SchedulerError',
 'SchedulerParsingError',
 '__all__',
 '__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__path__',
 '__spec__',
 'datastructures',
 'scheduler']

In [12]: import aiida.plugins

In [13]: dir(aiida.plugins)
Out[13]:
['BaseFactory',
 'CalculationFactory',
 'DataFactory',
 'DbImporterFactory',
 'GroupFactory',
 'OrbitalFactory',
 'ParserFactory',
 'SchedulerFactory',
 'TransportFactory',
 'WorkflowFactory',
 '__all__',
 '__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__path__',
 '__spec__',
 'entry_point',
 'factories',
 'load_entry_point',
 'load_entry_point_from_string']
In [14]: import aiida.common

In [15]: dir(aiida.common)
Out[15]:
['AIIDA_LOGGER',
 'AiidaException',
 'AttributeDict',
 'CalcInfo',
 'CalcJobState',
 'CodeInfo',
 'CodeRunMode',
 'ConfigurationError',
 'ConfigurationVersionError',
 'ContentNotExistent',
 'DbContentError',
 'DefaultFieldsAttributeDict',
 'EntryPointError',
 'FailedError',
 'FeatureDisabled',
 'FeatureNotAvailable',
 'FixedFieldsAttributeDict',
 'GraphTraversalRule',
 'GraphTraversalRules',
 'IncompatibleDatabaseSchema',
 'InputValidationError',
 'IntegrityError',
 'InternalError',
 'InvalidEntryPointTypeError',
 'InvalidOperation',
 'LicensingException',
 'LinkType',
 'LoadingEntryPointError',
 'MissingConfigurationError',
 'MissingEntryPointError',
 'ModificationNotAllowed',
 'MultipleEntryPointError',
 'MultipleObjectsError',
 'NotExistent',
 'NotExistentAttributeError',
 'NotExistentKeyError',
 'OutputParsingError',
 'ParsingError',
 'PluginInternalError',
 'ProfileConfigurationError',
 'ProgressReporterAbstract',
 'RemoteOperationError',
 'StoringNotAllowed',
 'TQDM_BAR_FORMAT',
 'TestsNotAllowedError',
 'TransportTaskException',
 'UniquenessError',
 'UnsupportedSpeciesError',
 'ValidationError',
 '__all__',
 '__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__path__',
 '__spec__',
 'constants',
 'create_callback',
 'datastructures',
 'escaping',
 'exceptions',
 'extendeddicts',
 'files',
 'folders',
 'get_progress_reporter',
 'hashing',
 'json',
 'lang',
 'links',
 'log',
 'override_log_formatter',
 'override_log_level',
 'progress_reporter',
 'set_progress_bar_tqdm',
 'set_progress_reporter',
 'timezone',
 'utils',
 'validate_link_label',
 'warnings']
```

We need to:

 1. Come up with a list of resources we *want* to be part of the public API (perhaps starting from the [list in the documentation](https://aiida-core.readthedocs.io/en/latest/reference/api/public.html))
 2. Deprecate resources from the second nesting level that don't belong there (removing them in 2.0)
 3. Document the practice, ideally generating the list (semi-)automatically

ltalirz:
Fixes #4558 

Make the definition of the AiiDA public API more explicit and add
examples.


ltalirz:
I have two remaining questions:

 * If the `aiida.cmdline.params.arguments` module should be considered part of the public python API, I suggest we expose this at `aiida.cmdline.arguments` (same for `options`, `types` and the stuff under `utils`)
 * Do we really need this list of public API resources?
   If I understand correctly, then everything in the `__all__` of the second-level packages is fair game.
   If that is true, this manual list (which is bound to date quickly) can either be removed entirely or replaced by an automatically generated one.

docs/source/reference/api/public.rst
```diff
@@ -3,11 +3,28 @@
...
+    .. code-block:: python
+
+        from aiida import load_profile  # OK, top-level import
+        from aiida.orm import QueryBuilder  # OK, second-level import
+        from aiida.tools.importexport import Archive # NOT PUBLIC API
+
+.. warning::
+
+    The implementation of resources that are *not* considered part of the public API can change between minor AiiDA releases.
+    When relying on such resources in scripts or AiiDA plugins, your code can therefore break in between minor AiiDA releases.
```
sphuber, line 26:
```suggestion
    The interface and implementation of resources that are *not* considered part of the public API can change between minor AiiDA releases, and can even be moved or fully removed, without a deprecation period whatsoever.
    Be aware that scripts or AiiDA plugins that rely on such resources, can therefore break unexpectedly in between minor AiiDA releases.
```

sphuber (changes requested):
Thanks @ltalirz apart from a suggestion, all good. Regarding your other questions:

1. I am not sure I fully understand. If we were to expose it under `aiida.cmdline.arguments` it would not be public API as per our own definitions. Currently all its resources are importable from `aiida.cmdline` so I am not sure what you want to change here.

2. I agree that maintaining this list manually is tedious and prone to go out of sync. However, I think it is useful to have in the reference. I don't think we can expect all users to always go into the source code and look at the `__all__` to see what is supposed to be public. If we can automate the creation of this list in a pre-commit hook, that'd be great. Not necessary for this PR I would say. Fine to open separate issue for that if you don't have the time to look into it now

chrisjsewell:
> If we can automate the creation of this list in a pre-commit hook, that'd be great.

What would be even better, is if we could actually offer a pre-commit hook, that fails if you access a non-public API (or if there was a way to get e.g. pylint to respect it). This documentation is great and all, but I'm way too lazy to actually use look at it when I'm writing a plugin 😛

ltalirz:
> 1. I am not sure I fully understand. If we were to expose it under `aiida.cmdline.arguments` it would not be public API as per our own definitions. Currently all its resources are importable from `aiida.cmdline` so I am not sure what you want to change here.

Oh, great that we discussed this!
I (mis)understood the sentence
> If a module is mentioned, then all the resources defined in its `__all__` are included

to mean that if we put a module in this list, then all members of this module are fair game *even though they are not exposed on second-level packages*.

I now realize that putting modules in the list was just a "shorthand" for the resources contained in their `__all__`.

Indeed, all arguments and options are exposed at the `aiida.cmdline` level, which I see as a pretty dangerous practice - for example, there is both a `CALCULATION` option and a  `CALCULATION` argument. Which is exposed at `aiida.cmdline` is a matter of luck of in which order someone decided to place the imports in the `__init__.py`:
```
In [4]: from aiida.cmdline.params.arguments import CALCULATION

In [5]: type(CALCULATION)
Out[5]: aiida.cmdline.params.arguments.overridable.OverridableArgument

In [6]: from aiida.cmdline.params.options import CALCULATION

In [7]: type(CALCULATION)
Out[7]: aiida.cmdline.params.options.overridable.OverridableOption

In [8]: from aiida.cmdline import CALCULATION

In [9]: type(CALCULATION)
Out[9]: aiida.cmdline.params.options.overridable.OverridableOption
```

I found my original interpretation of the sentence quite reasonable - it would give us another nesting level where needed, i.e. we could expose things at `aiida.cmdline.arguments.*`
The only alternative I see is to start including argument/option in the names of the resources (CALCULATION_OPTION), which would be a significant refactoring at this stage.

sphuber:
I see the point of "bubbling up" imports leading to namespace pollution and therefore naming conflicts. The example of cmdline arguments and options is a good example. I think this is also exactly the reason why I added the explicit list, because I realized that we might not always be able to make a public resource importable from a second level package. The fact that `aiida.cmdline.params.arguments` and `aiida.cmdline.params.options` are both exposed upwards is a mistake.

Still, I am not sure how to address this. In any case, if we do what you propose and make the resources of `aiida.cmdline.params.arguments` importable from `aiida.cmdline.arguments` we would have to change the phrasing. You can then no longer say that only what is importable from a second-level package is public, because `aiida.cmdline.arguments` is clearly third-level. The question then also becomes what the benefit of exposing it there is, just so you can skip `.params` when importing? Not quite sure if that is worth it in general.

The refactoring option, I don't really see as an option to be honest. Aside from the work involved, I simply think the bubbling up of imports there should not be done. It can be done, as long as the namespace removal that comes with it, does not cause naming conflicts.

ltalirz:
> The question then also becomes what the benefit of exposing it there is, just so you can skip .params when importing?

Right, this would be the only benefit, and I agree it's not a big one - keeping `.params` is fine as well.

We can simply stay with the 2-level rule as the default and add additional modules to this rule, where exposing at the second level does not make sense.
I think everybody will find this approach reasonable.

As for fixing this in the code, is there a good way of deprecating the old imports from `aiida.cmdline`?
A [quick github search](https://github.com/search?q=%22from+aiida.cmdline+import%22&type=code) didn't turn up any uses of options/arguments/types but I guess we cannot fully exclude it.

sphuber (approved):

PR #4519, opened 2020-10-27
Revert PR #4416

ramirezfranciscof:
This PR reverts the changes made in commit 16bc30548f7f1c686d200935174533535e850fd5. (Titled: "`CalcJob`: support nested directories in target of `remote_copy/symlink_list` (#4416)"). The reasons for the reversal is that apparently part of the behavior of the copy lists may have been inadvertidly changed; this needs to be explored so that we know in more details these behavior and then the feature can re-introduced after we can be more certain it won't break backwards compatibility.


sphuber (approved):
:+1:

PR #4521, opened 2020-10-27
Add group extras to archive

### LINKED ISSUE #4497: Add `Group.extras` to export archives
The `Group` class received `extras` in `v1.4.0` but they are currently not exported.

chrisjsewell:
fixes #4497 

Group extras were introduced recently but not yet exported to AiiDA archives.
This commit adds group extras to the AiiDA archive.

Instead of special-casing deserialization of attributes and extras based on the field name, a `convert_type: "jsonb"` is introduced, which is used to indicate JSON-binary fields.

### DIFF
--- a/aiida/tools/importexport/dbimport/backends/django.py
+++ b/aiida/tools/importexport/dbimport/backends/django.py
@@ -377,6 +377,12 @@ def _store_entity_data(
     fields_info = reader.metadata.all_fields_info.get(entity_name, {})
     unique_identifier = reader.metadata.unique_identifiers.get(entity_name, None)
 
+    if entity_name == NODE_ENTITY_NAME:
+        # in the current export process these fields are not present,
+        # because they are retrieved by a separate query
+        fields_info.setdefault('attributes', {'convert_type': 'jsonb'})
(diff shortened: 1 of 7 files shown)


aiida/tools/importexport/dbimport/backends/django.py
```diff
@@ -377,6 +377,12 @@ def _store_entity_data(
     fields_info = reader.metadata.all_fields_info.get(entity_name, {})
     unique_identifier = reader.metadata.unique_identifiers.get(entity_name, None)
 
+    if entity_name == NODE_ENTITY_NAME:
```
chrisjsewell, line 380: this should eventually be removed, but it requires (a) changes to the export code and (b) a migration

ltalirz (approved):
Tests before `StrictVersion` pass - thanks a lot @chrisjsewell , that's indeed a much more compact and elegant way to add the extras.
Looks like editing import/export starts being more fun again :-)

I've updated the PR description; feel free to use for the commit message

chrisjsewell:
> I've updated the PR description; feel free to use for the commit message

will do thanks!

PR #4345, opened 2020-09-01
Implement the new file repository

sphuber:
Fixes #3445 
Fixes #1663
Fixes #64 
Fixes #1675 
Fixes #3437 
Fixes #2580 
Fixes #3764
Fixes #4020

PR #4522, opened 2020-10-28
Add `reset` method to`ProgressReporterAbstract`

chrisjsewell:
This PR adds the `update` method to the progress reporter.
This in-turn, allows for the implmentation of a callback mechanism in `ArchiveReaderAbstract.iter_node_repos`.
The callback implmentation is taken from the [disk-objectstore](https://github.com/aiidateam/disk-objectstore) package, and so facilitates a later migration to its use.

The PR also moves the (common) repository import code out of the backend specific modules, reducing code duplication.

This PR is intended to facilitate the transition to the new node repository in #4345


chrisjsewell:
I've tested this against my 40,000 node archive, and looks to be all be working as expected:

Oct-28-2020 04-19-22

aiida/common/progress_reporter.py
```diff
@@ -50,9 +50,13 @@ def __init__(self, *, total: int, desc: Optional[str] = None, **kwargs: Any):
...
-        self.total = total
-        self.desc = desc
-        self.increment = 0
+        self._total = total
+        self._desc = desc
+        self._increment = 0
+
+    @property
+    def total(self) -> int:
+        return self._total
```
chrisjsewell, line 59:
```suggestion
    def total(self) -> int:
        """Return the total iterations expected."""
        return self._total
```

aiida/tools/importexport/archive/readers.py
```diff
@@ -357,26 +369,19 @@ def iter_node_repos(self,
...
-            for uuid, path_prefix in zip(uuids, path_prefixes):
-                subfolder = self._sandbox.get_subfolder(path_prefix)
-                if not subfolder.exists():
-                    raise CorruptArchive(
-                        f'Unable to find the repository folder for Node with UUID={uuid} in the exported file'
-                    )
-                yield subfolder
+            self._extract(path_prefix=common_prefix, callback=callback)
+
+        if callback is not None:
```
ltalirz, line 374: Or one could default to a dummy callback, which would make using this feel more like using a logger.

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -22,6 +25,34 @@
...
+    """Copy repositories of new nodes from the archive to the AiiDa profile.
+
+    :param uuids_to_create: the node UUIDs to copy
+    :param reader: the archive reader
+
+    """
+    if not uuids_to_create:
+        return
+    IMPORT_LOGGER.debug('CREATING NEW NODE REPOSITORIES...')
+    with get_progress_reporter()(total=1) as progress:
```
ltalirz, line 38: Ok, so this progress reporter is meant to be overwritten by the callback, right?
Perhaps add a comment to that effect.

ltalirz (commented):
Just some questions before approving (see comments)

aiida/common/progress_reporter.py
```diff
@@ -50,9 +50,14 @@ def __init__(self, *, total: int, desc: Optional[str] = None, **kwargs: Any):
...
 
         """
-        self.total = total
-        self.desc = desc
-        self.increment = 0
+        self._total = total
+        self._desc = desc
+        self._increment = 0
+
+    @property
```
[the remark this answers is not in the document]
chrisjsewell: no reason really I can make all of them public or none, there not used currently anywhere and I just wanted to make them not writable

aiida/tools/importexport/dbimport/backends/common.py
chrisjsewell: yes the callback manipulates the progress reporter.
I thought this would be clear in the `_callback` function below? but if not can add a comment

aiida/tools/importexport/archive/readers.py
chrisjsewell: I've add a `null_callback` function as default. Is this what you had in mind?

aiida/tools/importexport/dbimport/backends/common.py
chrisjsewell: I've added a comment though

aiida/tools/importexport/archive/readers.py
ltalirz: exactly, thanks!

ltalirz (approved):
Thanks @chrisjsewell !

Issue #3678, opened 2019-12-16
Unit test timings - worst offenders


ltalirz:
One rule of thumb for writing unit tests is that they should test one unit of code and run *in the blink of an eye*.

While this is not always possible, we have a large number of tests that violate this rule quite drastically, and should probably be refactored to save everybody's time.

Here are the worst offenders in terms of runtime as of 2019-12-16:
```
21.01s call     aiida/backends/tests/tools/importexport/orm/test_links.py::TestLinks::test_high_level_workflow_links
9.69s call     aiida/backends/tests/tools/importexport/orm/test_links.py::TestLinks::test_complex_workflow_graph_export_sets
7.42s call     aiida/backends/tests/tools/importexport/orm/test_links.py::TestLinks::test_link_flags
7.17s call     aiida/backends/tests/test_nodes.py::TestNodeDeletion::test_delete_cases
5.73s call     aiida/backends/tests/engine/test_rmq.py::TestProcessControl::test_launch_with_inputs
5.46s call     aiida/backends/djsite/db/subtests/migrations/test_migrations_many.py::TestDuplicateNodeUuidMigration::test_deduplicated_uuids
5.43s call     aiida/backends/tests/engine/test_futures.py::TestWf::test_calculation_future_broadcasts
5.42s call     aiida/backends/tests/engine/test_rmq.py::TestProcessControl::test_exception_process
5.40s call     aiida/backends/tests/engine/test_rmq.py::TestProcessControl::test_submit_simple
5.05s call     aiida/backends/djsite/db/subtests/migrations/test_migrations_many.py::TestUuidMigration::test_uuid_untouched
4.11s call     aiida/backends/djsite/db/subtests/migrations/test_migrations_many.py::TestCalcAttributeKeysMigration::test_attribute_key_changes
3.87s call     aiida/backends/djsite/db/subtests/migrations/test_migrations_many.py::TestDbLogMigrationBackward::test_objpk_objname
3.76s call     aiida/cmdline/utils/test_multiline.py::TestMultilineInput::test_edit_pre_bash_comment
3.71s call     aiida/backends/djsite/db/subtests/migrations/test_migrations_many.py::TestGroupRenamingMigration::test_group_string_update
3.70s call     aiida/backends/tests/cmdline/commands/test_code.py::TestVerdiCodeSetup::test_interactive_remote
[...]

Issue #3156, opened 2019-07-10
Streamline and clean files related to export archive migration tests


CasperWA:
The tests can be simplified, if

- the current `remove_fields` utility function is remade into `replace_or_remove_fields`, and make it agnostic to whether metadata or data was supplied or not;
- ~`arithmetic.add_old.aiida` export archive is not needed;~ and
- tests for migrations are split up into smaller tests, similar to the way the migrations are now split.

_Edit_: The test `test_calcjob_inoutputcat_old` under `aiida.backends.tests.cmdline.commands.test_calcjob.TestVerdiCalculation` is the only test that uses the `airthmetic.add_old.aiida` archive, and for a specific purpose at that.

danielhollas:
Closing this since the new archive format in #5145 made this irrelevant

> the current remove_fields utility function is remade into replace_or_remove_fields, and make it agnostic to whether metadata or data was supplied or not;

`remove_fields` function no longer exists.

PR #4532, opened 2020-10-29
Refactor archive migrations

chrisjsewell:
This PR primarily refactors the archive migrations, to provide an `ArchiveMigratorAbstract` interface, which is agnostic to the internal implementation of the archive (i.e. not dependent on the presence of metadata.json and data.json)
This will allow for subsequent changes to the archive format.
To facilitate this:

- `MIGRATE_FUNCTIONS` now includes both the to and from versions of the migration,
- this allows for a change, from a recursive migration approach to pre-computing the migration path, then applying the migrations iteratively
- this also allows for a progress reporter of the migration steps
- the signature of migration step functions has been changed, such that they now only receive the uncompressed archive folder, and not also specifically the data.json and metadata.json dicts.
- instead, the folder is wrapped in a new `CacheFolder` class, which caches file writes in memory, so reading of the files from the file system only happen once and they are written after all the migrations have finished.
- the `--verbose` flag has been added to `verdi export migrate`, unfortunately only the long-format can be used, since `-v` is already reserved for the version.
- consolidated the extracting of tar/zip into `safe_extract_tar`/`safe_extract_zip`.
  These include callbacks, for which I created `aiida.common.progress_reporter::create_callback`, to provide a convenience function.
- moved all migration tests to pytest

This is the output of an example migration:

```
$ verdi export migrate tmp/mount_folder/tests/static/export/migrate/export_v0.4_simple.aiida --verbosity DEBUG -v 0.9 -f -F tar.gz tmp/mount_folder/tests/static/export/migrate/test.tar.gz 
Migration pathway: 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8 -> 0.9
Extracting archive to temporary folder
Performing migrations: 0.8 -> 0.9        100.0%|█████████████████████████████████████████████████████| 5/5
Re-compressing archive as 'tar.gz'
Moving archive to: tmp/mount_folder/tests/static/export/migrate/test.tar.gz
Success: migrated the archive to version 0.9
```


chrisjsewell:
@ltalirz I have not yet updated the tests, but you can start looking at this 😄

ltalirz:
> * `-v` is already reserved for the version.

Hm yeah, I think this should be deprecated.

ltalirz:
Looks very nice!

chrisjsewell:
Oh and I also need to update `cmd_import`, to improve the migration integration (e.g. using the correct verbosity)

ltalirz:
> Oh and I also need to update `cmd_import`

By the way, the `cmd_import` can probably be shortened a lot - there is a lot of duplication for archives coming from urls or from files that is mostly entirely unnecessary

ltalirz:
Came across this https://github.com/aiidateam/aiida-core/issues/3156 - just in case it's easy to fix

ltalirz:
and this https://github.com/aiidateam/aiida-core/issues/3193

ltalirz:
And curious to see how migration tests will be affected in https://github.com/aiidateam/aiida-core/issues/3678

chrisjsewell:
> And curious to see how migration tests will be affected in #3678

some of them should definitely be quicker, since `get_json_files` was extracting the whole archive, just to retrieve data.json and metadata.json, and so I've replaced that with the new `read_file_in_zip` and `read_file_in_tar` functions

chrisjsewell:
ok @ltalirz I have updated all the tests!

I moved all migration ones to pytest, and additionally I have consolidated the extracting of tar/zip into `safe_extract_tar`/`safe_extract_zip`.
These include callbacks, for which I created `aiida.common.progress_reporter::create_callback`, to provide a convenience function.
This means that the CLI now includes a progress bar for this:

```
$ verdi export migrate tmp/mount_folder/tests/static/export/migrate/export_v0.4_simple.aiida --verbosity DEBUG -v 0.9 -f -F tar.gz tmp/mount_folder/tests/static/export/migrate/test.tar.gz 
Migration pathway: 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8 -> 0.9
Extracting archive to temporary folder
Extracting zip files                     100.0%|██████████████████████| 76/76
Performing migrations: 0.8 -> 0.9        100.0%|████████████████████████| 5/5
Re-compressing archive as 'tar.gz'
Moving archive to: tmp/mount_folder/tests/static/export/migrate/test.tar.gz
Success: migrated the archive to version 0.9
```

In terms of test timings, here are the top ones. The long-running ones all belong to `test_links.py`, which I have not touched (and don't intent to in this PR 😬).

```
9.72s call     tests/tools/importexport/orm/test_links.py::TestLinks::test_high_level_workflow_links
5.32s call     tests/tools/importexport/orm/test_links.py::TestLinks::test_complex_workflow_graph_export_sets
3.78s call     tests/tools/importexport/orm/test_links.py::TestLinks::test_link_flags
0.90s call     tests/tools/importexport/test_complex.py::TestComplex::test_reexport
0.73s call     tests/tools/importexport/migration/test_migration.py::TestExportFileMigration::test_migrate_to_newest[0.2]
0.66s call     tests/tools/importexport/orm/test_calculations.py::TestCalculations::test_calcfunction
0.66s call     tests/tools/importexport/migration/test_migration.py::TestExportFileMigration::test_migrate_to_newest[0.6]
0.64s call     tests/tools/importexport/migration/test_migration.py::TestExportFileMigration::test_migrate_to_newest[0.4]
0.63s call     tests/tools/importexport/orm/test_users.py::TestUsers::test_non_default_user_nodes
0.62s call     tests/tools/importexport/orm/test_links.py::TestLinks::test_complex_workflow_graph_links
0.62s call     tests/tools/importexport/migration/test_migration.py::TestExportFileMigration::test_migrate_to_newest[0.7]
0.62s call     tests/tools/importexport/orm/test_comments.py::TestComments::test_reimport_of_comments_for_single_node
0.61s call     tests/tools/importexport/migration/test_migration.py::TestExportFileMigration::test_migrate_to_newest[0.5]
0.60s call     tests/tools/importexport/migration/test_migration.py::TestExportFileMigration::test_migrate_to_newest[0.8]
0.59s call     tests/tools/importexport/orm/test_extras.py::TestExtras::test_extras_import_mode_correct
0.59s call     tests/tools/importexport/test_prov_redesign.py::TestProvenanceRedesign::test_node_process_type
0.56s call     tests/tools/importexport/test_complex.py::TestComplex::test_complex_graph_import_export
0.55s call     tests/tools/importexport/orm/test_logs.py::TestLogs::test_reimport_of_logs_for_single_node
0.52s call     tests/tools/importexport/orm/test_computers.py::TestComputer::test_different_computer_same_name_import
```

aiida/cmdline/commands/cmd_export.py
```diff
@@ -186,86 +186,52 @@ def create(
...
+        version = EXPORT_VERSION
 
-        if zipfile.is_zipfile(input_file):
-            extract_zip(input_file, folder, silent=silent)
-        elif tarfile.is_tarfile(input_file):
-            extract_tar(input_file, folder, silent=silent)
-        else:
-            echo.echo_critical('invalid file format, expected either a zip archive or gzipped tarball')
+    migrator_cls = get_migrator(detect_archive_type(input_file))
+    migrator = migrator_cls(input_file)
```
ltalirz, line 226: hehe, migrator is not a context manager?

I'm shocked :-D

aiida/common/progress_reporter.py
```diff
@@ -174,3 +174,26 @@ def set_progress_bar_tqdm(bar_format: Optional[str] = TQDM_BAR_FORMAT, leave: Op
     """
     from tqdm import tqdm
     set_progress_reporter(tqdm, bar_format=bar_format, leave=leave, **kwargs)
+
+
+def create_callback(progress_reporter: ProgressReporterAbstract) -> Callable[[str, Any], None]:
```
ltalirz, line 179: or make it a method of the progress reporter?

aiida/tools/importexport/archive/common.py
```diff
@@ -61,3 +74,144 @@ def detect_archive_type(in_path: str) -> str:
         'Unable to detect the input file format, it is neither a '
         'folder, tar file, nor a (possibly compressed) zip file.'
     )
+
+
+def read_file_in_zip(filepath: str, path: str) -> str:
```
ltalirz, line 79: this returns the file contents as a string, right (i.e. it shouldn't be used for binary files)?

I think this should be mentioned in the docstring.
Ideally also in the function name but perhaps that would get too long...

aiida/tools/importexport/archive/common.py
```diff
@@ -61,3 +74,144 @@ def detect_archive_type(in_path: str) -> str:
...
+            if isinstance(output, bytes):
+                return output.decode('utf8')
+    except tarfile.ReadError:
+        raise ValueError('The input file format is not valid (not a tar file)')
+    except (KeyError, AttributeError):
+        raise CorruptArchive(f'required file `{path}` is not included')
+
+
+def _get_filter(only_prefix: Iterable[str], ignore_prefix: Iterable[str]) -> Callable[[str], bool]:
+    """Create filter for members to extract."""
```
ltalirz, line 116:
```suggestion
    """Create name filter."""
```

ltalirz: + docs of params please

aiida/tools/importexport/archive/migrations/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+from .v01_to_v02 import migrate_v1_to_v2
+from .v02_to_v03 import migrate_v2_to_v3
+from .v03_to_v04 import migrate_v3_to_v4
+from .v04_to_v05 import migrate_v4_to_v5
+from .v05_to_v06 import migrate_v5_to_v6
+from .v06_to_v07 import migrate_v6_to_v7
+from .v07_to_v08 import migrate_v7_to_v8
+from .v08_to_v09 import migrate_v8_to_v9
+
+# version from -> version to, func(path to (uncompressed), dict of cached data) -> dict of cached data
```
ltalirz, line 25:
```suggestion
# version from -> version to, func(path to uncompressed json, dict of cached data) -> dict of migrated data
```

ltalirz: However, this doesn't really do this function justice, since what matters is the side effect of the migrated data being written to disk, right?

aiida/tools/importexport/archive/migrations/v01_to_v02.py
```diff
@@ -27,41 +29,54 @@ def migrate_v1_to_v2(metadata, data, *args):
...
             if key.startswith(old_start):
-                new_key = get_new_string(key)
+                new_key = get_new_string(key, old_start, new_start)
                 metadata[field][new_key] = metadata[field][key]
                 del metadata[field][key]
 
-    metadata['all_fields_info'] = replace_requires(metadata['all_fields_info'])
+    metadata['all_fields_info'] = replace_requires(metadata['all_fields_info'], old_start, new_start)
+
+    (folder / 'metadata.json').write_text(json.dumps(metadata), encoding='utf8')
```
ltalirz, line 57: Why does this need to be written every time?

It seems to me that the most efficient way to do this would be to:
 * use a context manager
 * let the migration function read data when needed, which then causes read of json file via context manager
 * only update metadata/data in context manager (never write to disk during migration)
 * on exit (after all migrations), let context manager check which of metadata/data were updated; write only those back

ltalirz (commented):
thanks @chrisjsewell , looks great!

I started going through this; will continue tomorrow at some point

aiida/common/progress_reporter.py
chrisjsewell: tqdm does not have a `create_callback` method though. The only way this could work is if you dynamically added the method in `set_progress_bar_tqdm`, which is a bit hacky

aiida/tools/importexport/archive/common.py
chrisjsewell: sure can add it to the function string, but I think the typing makes it quite clear

aiida/cmdline/commands/cmd_export.py
chrisjsewell: lol no I didn't see the need

aiida/tools/importexport/archive/migrations/__init__.py
chrisjsewell: yep

aiida/tools/importexport/archive/migrations/v01_to_v02.py
chrisjsewell: well firstly you are assuming that metadata.json and data.json will always be present. Once the archive format is changed that will certainly not be the case (and is the main point of this PR 😉).
When there is a migration which converts this data.json to the new format, this method will no longer work.
Similarly, once we don't have to, we will not want to hold all the data in memory until all migrations are complete.
Essentially we are sacrificing some performance, for a more general approach, which will eventually allow for a large reduction in memory usage.

aiida/tools/importexport/archive/common.py
chrisjsewell: done

aiida/tools/importexport/archive/migrations/v01_to_v02.py
chrisjsewell: I think passing the migration functions the migrator would be a big hassle, especially for testing and modularity.
But you could pass them something like:

```python
class CacheFolder:
  """Caches writes."""
  def __init__(self, path: Path):
    self._path = path
    self._cache = {}
  @property
  def path(self):
    return self._path
  def flush(self):
     for path, val in self._cache.items():
        (self._path / path).write_text(val, encoding="utf8")
  def write_text(path, content):
    self._cache[path] = content
  def read_text(path):
    if path in self._cache:
       return self._cache[path]
    return (self._path / path).read_text("utf8")
  def __enter__(self):
     return self
  def __exit__(self, *args):
     self.flush()
```

what do you think about that?

chrisjsewell: implemented in cbcaec6

ltalirz: > what do you think about that?

Looks great!

tests/tools/importexport/migration/test_migration.py
```diff
@@ -8,509 +8,186 @@
...
-        new_version = migrate_recursively(metadata, {}, None)
-        self.assertEqual(new_version, newest_version)
-
-    @with_temp_dir
-    def test_v02_to_newest(self, temp_dir):
-        """Test migration of exported files from v0.2 to newest export version"""
-        # Get archive file with export version 0.2
-        input_file = get_archive_file('export_v0.2.aiida', **self.external_archive)
-        output_file = os.path.join(temp_dir, 'output_file.aiida')
+        self.external_archive = {'filepath': 'archives', 'external_module': 'aiida-export-migration-tests'}
```
ltalirz, line 45: constant could be reused here

ltalirz (commented):
thanks @chrisjsewell 

I went through the code except for the individual tests.

As discussed, we may want to think a bit more about the cachefolder approach

aiida/cmdline/commands/cmd_import.py
```diff
@@ -347,9 +189,210 @@ def cmd_import(
...
+    echo.echo(exception)
+
+    if more_archives:
+        # There are more archives to go through
+        if non_interactive:
+            # Continue to next archive
+            pass
+        else:
+            # Ask if one should continue to next archive
+            click.confirm('Do you want to continue?', abort=True)
```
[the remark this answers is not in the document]
chrisjsewell: well yeh I didn't decide on the logic to allow for failures, it was already there and I didn't want to change the behaviour.
Happy to remove it though

aiida/tools/importexport/archive/common.py
```diff
@@ -61,3 +77,253 @@ def detect_archive_type(in_path: str) -> str:
...
+
+        # all path/to/file.txt will now have been written to the file system
+
+    """
+
+    def __init__(self, path: Union[Path, str], *, limit: int = 100, encoding: str = 'utf8'):
+        """Initialise cached folder.
+
+        :param path: folder path to cache
+        :param limit: maximum number of files to cache ()
```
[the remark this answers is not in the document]
chrisjsewell: sure, I can remove it from the public API.
I think its good to have some sort of limit though.

aiida/tools/importexport/archive/migrators.py
```diff
@@ -0,0 +1,240 @@
...
+    def _retrieve_version(self) -> str:
+        """Retrieve the version of the input archive."""
+        raise NotImplementedError()
+
+    def _extract_archive(self, filepath: Path, callback: Callable[[str, Any], None]):
+        """Extract the archive to a filepath."""
+        raise NotImplementedError()
+
+    @staticmethod
+    def _compress_archive_zip(in_path: Path, out_path: Path, compression: int):
```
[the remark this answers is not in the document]
chrisjsewell: Primarily I wanted / needed to support the existing `--archive-format` option of `verdi export migrate`, which does in fact allow you to re-compress in a different format to the original file.

tests/tools/importexport/migration/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+
+import pytest
+
+from aiida.tools.importexport.archive import safe_extract_tar, safe_extract_zip, CacheFolder
+from aiida.tools.importexport.archive.migrations.utils import verify_metadata_version
+from tests.utils.archives import get_archive_file
+
+
+@pytest.fixture()
+def core_archive():
```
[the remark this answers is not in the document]
chrisjsewell: the core archive or ones that point to files in the aiida-core repository, as opoosed to the external ones, which point to the `aiida-export-migration-tests` package.
Although I did point out that having the archives in a separate repo may not really be necessary anymore: https://github.com/aiidateam/aiida-export-migration-tests/issues/13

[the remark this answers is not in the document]
chrisjsewell: it was just the easiest way really, to minimise the changes require during the conversion pytest. I don't think its too bad, if they are tightly scoped (i.e. the conftest only applies to this single folder),
plus its a lot easier now that vs code's intellisense recognises pytest fixtures, so that they exactly the same as if you had imported them 😄

aiida/cmdline/commands/cmd_import.py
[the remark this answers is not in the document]
ltalirz: Up to you

aiida/tools/importexport/archive/common.py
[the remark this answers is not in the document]
ltalirz: > sure, I can remove it from the public API.
> I think its good to have some sort of limit though.

Right. I guess what we typically care about is memory usage so I guess it would be based on the total size of the cache.
In any case; it will be easy to add once we know what we need.

aiida/tools/importexport/archive/migrators.py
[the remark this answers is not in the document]
ltalirz: Ok, thanks I suspected that.

tests/tools/importexport/migration/conftest.py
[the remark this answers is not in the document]
ltalirz: > Although I did point out that having the archives in a separate repo may not really be necessary anymore: [aiidateam/aiida-export-migration-tests#13](https://github.com/aiidateam/aiida-export-migration-tests/issues/13)

Right, you are welcome to change this here (not sure when someone will touch this for the next time.

By the way, there is also the git-lfs for storing larger binary files so it wouldn't even clog up repository size; I think git now automatically suggests it when trying to store larger files.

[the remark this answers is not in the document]
chrisjsewell: interesting thanks I will check that out

aiida/tools/importexport/archive/common.py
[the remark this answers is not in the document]
chrisjsewell: done

chrisjsewell:
@ltalirz in b44e350, I made the alterations to `CacheFolder`, to allow for only one copy of a dict to be maintained. The catch is that obviously any mutations of such a dict will affect the cache, so hopefully I have made that clear in `CacheFolder.load_json` 

Unscientific test of caching performance, with test duration for: `tests/tools/importexport/migration/test_migration.py::TestExportFileMigration::test_migrate_to_newest[0.2]`

`CacheFolder._max_items = 100`

0.64s, 0.65s, 0.63s, 0.67s, 0.66s

`CacheFolder._max_items = 0` (i.e. always read/write)

0.65s, 0.70s, 0.98s, 0.68s, 0.77s

But obviously its a bit hard to tell definitively from such a small archive.

tests/tools/importexport/migration/test_migration.py
chrisjsewell: removed

aiida/cmdline/commands/cmd_import.py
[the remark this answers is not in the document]
chrisjsewell: I've simplified the cmd_import in ffb3699.
As the commit message says, I also removed the click.confirm for migrations, since this is already controlled by --migration/--no-migration

This is the output of a migrated import:

```
$ verdi import -v DEBUG tmp/mount_folder/tests/static/export/migrate/export_v0.2_simple.aiida
Calling import function import_data_dj for the django backend.
Warning: initial import of tmp/mount_folder/tests/static/export/migrate/export_v0.2_simple.aiida failed, trying migration: IncompatibleArchiveVersionError: Archive file version is 0.2, can read only version 0.9
Use 'verdi export migrate' to update this archive file.
Migration pathway: 0.2 -> 0.3 -> 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8 -> 0.9
Extracting archive to temporary folder
Extracting zip files                     100.0%|██████████████████████████████████████████████████████████| 76/76
Performing migrations: 0.8 -> 0.9        100.0%|████████████████████████████████████████████████████████████| 7/7
Re-compressing archive as 'zip'
Moving archive to: /home/aiida/.aiida/repository/other3/sandbox/tmpo644hzqn/migrated_archive.zip
Info: proceeding with import
Calling import function import_data_dj for the django backend.

IMPORT
--------  --------------------
Archive   migrated_archive.zip

Parameters
--------------------------  ------
Comment rules               newest
New Node Extras rules       import
Existing Node Extras rules  kcl
CHECKING IF NODES FROM LINKS ARE IN DB OR ARCHIVE...
CREATING PK-2-UUID/EMAIL MAPPING...
Importing 16 entities
ASSESSING IMPORT DATA...
Finding existing entities - User         100.0%|████████████████████████████████████████████████████████████| 1/1
Reading archived entities - User         100.0%|████████████████████████████████████████████████████████████| 1/1
Finding existing entities - Computer     100.0%|████████████████████████████████████████████████████████████| 1/1
Reading archived entities - Computer     100.0%|████████████████████████████████████████████████████████████| 1/1
Finding existing entities - Node         100.0%|██████████████████████████████████████████████████████████| 10/10
Reading archived entities - Node         100.0%|██████████████████████████████████████████████████████████| 10/10
Finding existing entities - Group        100.0%|████████████████████████████████████████████████████████████| 4/4
Reading archived entities - Group        100.0%|████████████████████████████████████████████████████████████| 4/4
STORING ENTITIES...
Users -  existing entries                100.0%|████████████████████████████████████████████████████████████| 1/1
Computers -  existing entries            100.0%|████████████████████████████████████████████████████████████| 1/1
Nodes -  existing entries                100.0%|██████████████████████████████████████████████████████████| 10/10
Nodes - UUID=f75aaf40 - Extras           100.0%|██████████████████████████████████████████████████████████| 10/10
Groups -  existing entries               100.0%|████████████████████████████████████████████████████████████| 4/4
STORING NODE LINKS...
Links - label=output_array               100.0%|████████████████████████████████████████████████████████████| 8/8
   (0 new links...)
STORING GROUP ELEMENTS...
Groups -                                 100.0%|████████████████████████████████████████████████████████████| 4/4
Done (cleaning up)                       100.0%|██████████████████████████████████████████████████████████| 10/10

Summary
-----------------------  ---------------
Auto-import Group label  20201031-082514
User(s)                  1 existing
Computer(s)              1 existing
Node(s)                  10 existing
Group(s)                 4 existing

Success: imported archive tmp/mount_folder/tests/static/export/migrate/export_v0.2_simple.aiida
```

or with --no-migration

```
$ verdi import -v DEBUG tmp/mount_folder/tests/static/export/migrate/export_v0.2_simple.aiida --no-migration
Calling import function import_data_dj for the django backend.
  File "/tmp/mount_folder/aiida/cmdline/commands/cmd_import.py", line 220, in _import_archive
    import_data(archive_path, **import_kwargs)
  File "/tmp/mount_folder/aiida/tools/importexport/dbimport/__init__.py", line 87, in import_data
    return import_data_dj(in_path, group=group, **kwargs)
  File "/tmp/mount_folder/aiida/tools/importexport/dbimport/backends/django.py", line 128, in import_data_dj
    reader.check_version()
  File "/tmp/mount_folder/aiida/tools/importexport/archive/readers.py", line 140, in check_version
    raise IncompatibleArchiveVersionError(msg)
aiida.tools.importexport.common.exceptions.IncompatibleArchiveVersionError: Archive file version is 0.2, can read only version 0.9
Use 'verdi export migrate' to update this archive file.

Critical: an exception occurred while importing the archive tmp/mount_folder/tests/static/export/migrate/export_v0.2_simple.aiida: IncompatibleArchiveVersionError: Archive file version is 0.2, can read only version 0.9
Use 'verdi export migrate' to update this archive file.
```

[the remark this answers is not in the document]
chrisjsewell: should also check manually with a webpage. What a good one to try?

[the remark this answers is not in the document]
ltalirz: > I've simplified the cmd_import in [ffb3699](https://github.com/aiidateam/aiida-core/commit/ffb36990ec9069b8f078be2eb20dd6aece2193b7).

Thanks!


> I also removed the click.confirm for migrations, since this is already controlled by --migration/--no-migration

For me that change is fine, please also update the documentation (search for "prompt" in share_data.rst


> should also check manually with a webpage. Whats a good one to try?

One of the files from https://archive.materialscloud.org/record/2018.0001/v1 ?

ltalirz:
> But obviously its a bit hard to tell definitively from such a small archive.

I see; i.e. for such small archives we were perhaps already using some OS-level caching mechanism.

I'll have another look through the files that changed.

aiida/cmdline/commands/cmd_import.py
```diff
@@ -252,104 +104,141 @@ def cmd_import(
...
+    message = f'{msg}: {exception.__class__.__name__}: {str(exception)}'
+    if warn_only:
+        echo.echo_warning(message)
+    else:
+        IMPORT_LOGGER.debug('%s', traceback.format_exc())
+        echo.echo_critical(message)
+
+
+def _gather_imports(archives, webpages) -> List[Tuple[str, bool]]:
+    """Gather archives to import and sort into local and web-base.
```
ltalirz, line 157:
```suggestion
    """Gather archives to import and sort into local files and URLs.
```

aiida/cmdline/commands/cmd_import.py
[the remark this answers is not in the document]
chrisjsewell: yep that worked fine 👍 

```
$ verdi import -w https://archive.materialscloud.org/record/2018.0001/v1Info: retrieving archive URLS from https://archive.materialscloud.org/record/2018.0001/v1
Success: 2 archive URLs discovered and added
Info: downloading archive https://archive.materialscloud.org/record/file?record_id=20&file_id=c8c80235-cf81-4660-b0cf-3ca5824c3725&filename=SSSP_efficiency_pseudos.aiida
Success: archive downloaded, proceeding with import
Warning: initial import of https://archive.materialscloud.org/record/file?record_id=20&file_id=c8c80235-cf81-4660-b0cf-3ca5824c3725&filename=SSSP_efficiency_pseudos.aiida failed, trying migration: IncompatibleArchiveVersionError: Archive file version is 0.3, can read only version 0.9
Use 'verdi export migrate' to update this archive file.
Migration pathway: 0.3 -> 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8 -> 0.9
Extracting archive to temporary folder
Re-compressing archive as 'zip'                                                                                  
Moving archive to: /home/aiida/.aiida/repository/other3/sandbox/tmpubonj4bn/migrated_archive.zip
Info: proceeding with import

IMPORT
--------  --------------------
Archive   migrated_archive.zip

Parameters
--------------------------  ------
Comment rules               newest
New Node Extras rules       import
Existing Node Extras rules  kcl
                                                                                                                 
Summary
-----------------------  ---------------
Auto-import Group label  20201031-110237
User(s)                  1 new
Node(s)                  85 new
Group(s)                 1 new

Success: imported archive https://archive.materialscloud.org/record/file?record_id=20&file_id=c8c80235-cf81-4660-b0cf-3ca5824c3725&filename=SSSP_efficiency_pseudos.aiida
Info: downloading archive https://archive.materialscloud.org/record/file?record_id=20&file_id=bb240983-f808-4a0c-aba9-ce4a728820ac&filename=SSSP_accuracy_pseudos.aiida
Success: archive downloaded, proceeding with import
Warning: initial import of https://archive.materialscloud.org/record/file?record_id=20&file_id=bb240983-f808-4a0c-aba9-ce4a728820ac&filename=SSSP_accuracy_pseudos.aiida failed, trying migration: IncompatibleArchiveVersionError: Archive file version is 0.3, can read only version 0.9
Use 'verdi export migrate' to update this archive file.
Migration pathway: 0.3 -> 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8 -> 0.9
Extracting archive to temporary folder
Re-compressing archive as 'zip'                                                                                  
Moving archive to: /home/aiida/.aiida/repository/other3/sandbox/tmpk18emw11/migrated_archive.zip
Info: proceeding with import

IMPORT
--------  --------------------
Archive   migrated_archive.zip

Parameters
--------------------------  ------
Comment rules               newest
New Node Extras rules       import
Existing Node Extras rules  kcl
                                                                                                                 
Summary
-----------------------  ---------------
Auto-import Group label  20201031-110248
User(s)                  1 existing
Node(s)                  11 new
Group(s)                 1 new

Success: imported archive https://archive.materialscloud.org/record/file?record_id=20&file_id=bb240983-f808-4a0c-aba9-ce4a728820ac&filename=SSSP_accuracy_pseudos.aiida
```

aiida/cmdline/commands/cmd_import.py
```diff
@@ -252,104 +104,141 @@ def cmd_import(
...
+            temp_folder.create_file_from_filelike(response, 'downloaded_archive.zip')
+            archive_path = temp_folder.get_abs_path('downloaded_archive.zip')
+            echo.echo_success('archive downloaded, proceeding with import')
 
         try:
-            response = urllib.request.urlopen(archive)
+            import_data(archive_path, **import_kwargs)
+        except IncompatibleArchiveVersionError as exception:
+            if try_migration:
+                _echo_exception(f'initial import of {archive} failed, trying migration', exception, warn_only=True)
```
ltalirz: Maybe you did this to have information on the archive version included in the message?
Perhaps `IncompatibleArchiveVersionError` could have that version set as an attribute (or you can print the message of hte exception anyhow).

I just feel that a "warning" gives the idea that something is wrong.

aiida/cmdline/commands/cmd_import.py
```diff
@@ -252,104 +104,141 @@ def cmd_import(
...
+                # try migration
+                try:
+                    migrator = get_migrator(detect_archive_type(archive_path))(archive_path)
+                    archive_path = temp_folder.get_abs_path('migrated_archive.zip')
+                    migrator.migrate(EXPORT_VERSION, archive_path, out_compression='zip')
+                except Exception as exception:
+                    _echo_exception(f'an exception occurred while migrating the archive {archive}', exception)
+                # re-try import
+                echo.echo_info('proceeding with import')
```
ltalirz, line 233:
```suggestion
                echo.echo_info('proceeding with import of migrated archive')
```

aiida/cmdline/commands/cmd_import.py
```diff
@@ -252,104 +104,141 @@ def cmd_import(
...
+                try:
+                    import_data(archive_path, **import_kwargs)
+                except Exception as exception:
+                    _echo_exception(
+                        f'an exception occurred while re-trying import of the migrated archive {archive}', exception
```
ltalirz, line 238:
```suggestion
                        f'an exception occurred while trying to import the migrated archive {archive}', exception
```

chrisjsewell:
> I see; i.e. for such small archives we were perhaps already using some OS-level caching mechanism.

Oh no I just mean that the fluctuations in timings are almost as big as the time to read/write the relatively small data.json.
(although the average definitely appears to be lower with caching)

aiida/cmdline/commands/cmd_import.py
```diff
@@ -252,104 +104,141 @@ def cmd_import(
...
+                    )
+            else:
+                _echo_exception(f'an exception occurred while importing the archive {archive}', exception)
```
ltalirz, line 241:
```suggestion
                _echo_exception(f'an exception occurred while trying to import the archive {archive}', exception)
```

aiida/cmdline/commands/cmd_import.py
```diff
@@ -252,104 +104,141 @@ def cmd_import(
...
         except Exception as exception:
-            _echo_error(f'downloading archive {archive} failed', raised_exception=exception, **import_opts)
-
-        with SandboxFolder() as temp_folder:
-            temp_file = 'importfile.tar.gz'
+            _echo_exception(f'an exception occurred while importing the archive {archive}', exception)
```
ltalirz, line 243:
```suggestion
            _echo_exception(f'an exception occurred while trying to import the archive {archive}', exception)
```

aiida/tools/importexport/archive/common.py
```diff
@@ -61,3 +78,302 @@ def detect_archive_type(in_path: str) -> str:
...
+
+
+class CacheFolder:
+    """A class to encapsulate a folder path with cached read/writes.
+
+    The class can be used as a context manager, and will flush the cache on exit::
+
+        with CacheFolder(path) as folder:
+            # this stored in memory (no disk write)
+            folder.write_text('path/to/file.txt', 'content')
```
ltalirz, line 236: Could you add a json example here as well?

aiida/tools/importexport/archive/common.py
```diff
@@ -61,3 +78,302 @@ def detect_archive_type(in_path: str) -> str:
...
+        if ctype == 'text':
+            return content
+        if ctype == 'json':
+            return json.dumps(content)
+
+        raise TypeError(f"content of type '{ctype}' could not be converted to text")
+
+    def write_json(self, path: str, data: dict):
+        """write dict to the cache as json.
+
```
ltalirz, line 327: Add something like:

Dictionary is stored in memory until cache is flushed, at which point the dictionary is serialized to json and written to disk.

aiida/tools/importexport/archive/common.py
```diff
@@ -61,3 +78,302 @@ def detect_archive_type(in_path: str) -> str:
...
+        ctype, content = self._cache[path]
+        if ctype == 'text':
+            return content
+        if ctype == 'json':
+            return json.dumps(content)
+
+        raise TypeError(f"content of type '{ctype}' could not be converted to text")
+
+    def write_json(self, path: str, data: dict):
+        """write dict to the cache as json.
```
ltalirz, line 326:
```suggestion
        """Write dict to the folder, to be serialized as json.
```

aiida/tools/importexport/archive/common.py
```diff
@@ -61,3 +78,302 @@ def detect_archive_type(in_path: str) -> str:
...
+
+        :param path: path relative to base folder
+
+        """
+        json.dumps(data)  # make sure that the data can be converted to json
+        self._cache[path] = ('json', data)
+        self._limit_cache()
+
+    def load_json(self, path: str, ensure_copy: bool = False) -> Tuple[bool, dict]:
+        """Load a json file from the cache or base folder.
```
ltalirz, line 336:
```suggestion
        """Load a json file from the cache folder.
```

aiida/cmdline/commands/cmd_import.py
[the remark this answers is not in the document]
chrisjsewell: and I've updated `share_data.rst`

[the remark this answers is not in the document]
chrisjsewell: > I thought there was a test for archives imported from URLs as well... perhaps I remember incorrectly?

there may well be, I haven't looked to change any of them tests

ltalirz:
Thanks @chrisjsewell - I have no further comments besides those above; perhaps moving the export files back to aiida-core could be done in a separate PR.

I will now check timings / memory consumption for a larger migration.

aiida/cmdline/commands/cmd_import.py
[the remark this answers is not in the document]
chrisjsewell: changed

chrisjsewell: yeh fair now changed.
it was mainly just left from the original code, the subsequent migration stdout gives info on the versions, so I don't think it is necessary to include that in the message

> Perhaps IncompatibleArchiveVersionError could have that version set as an attribute

yeh good idea, although I'm not going to do that in this PR, there's enough changes already lol

aiida/tools/importexport/archive/common.py
chrisjsewell: done

aiida/tools/importexport/archive/common.py
chrisjsewell: added

chrisjsewell:
> perhaps moving the export files back to aiida-core could be done in a separate PR.

oh yes certainly 👍

ltalirz:
I originally wanted to use the aiida file from https://archive.materialscloud.org/record/2017.0008/v3 to test the migration timings but it turns out that extracting tar files is extremely slow

E.g. on the openstack VM where I wanted to test, it took me 72 minutes to get to 7.8% extracted (at which point I cancelled this):
```
 $ time verdi export migrate two_dimensional_database.aiida 2d_migrated.aiida
Migration pathway: 0.3 -> 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8 -> 0.9
Extracting archive to temporary folder
Extracting tar files                       7.8%|
```

This is a *lot* slower than the previous extraction.

Can you please check whether we can fix this to get around this performance regression?
https://github.com/aiidateam/aiida-core/blob/4238cb2ec3a391e62b54efa596335025ee5b5f2d/aiida/tools/importexport/archive/common.py#L179
``


https://github.com/aiidateam/aiida-core/blob/4238cb2ec3a391e62b54efa596335025ee5b5f2d/aiida/tools/importexport/archive/common.py#L179

chrisjsewell:
I've got an idea of why this is: in `safe_extract_tar` the progress is updated for every file. This should be changed to only call update every x%.
I'm out at the moment but you could also first try replacing it with https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall as a baseline, presumably this is the pure python fastest implementation  (but unsafe and no progress)

ltalirz:
> I've got an idea of why this is: in `safe_extract_tar` the progress is updated for every file. This should be changed to only call update every x%.

Are you sure? 
tqdm has some internal logic for when to refresh the progress bar and when you don't request a refresh explicitly, it should be doing the right thing out of the box.

Also, I believe the approach in `safe_extract_zip` is the same and we didn't notice any slowdown with the previous archives we looked at...

chrisjsewell:
> Are you sure?
> tqdm has some internal logic for when to refresh the progress bar and when you don't request a
> refresh explicitly, it should be doing the right thing out of the box.

Yep that's not the case: https://github.com/tqdm/tqdm/blob/0f823e79f303b4a93ef1381badb1e65757e5070f/tqdm/std.py#L1226

ltalirz:
> Yep that's not the case: https://github.com/tqdm/tqdm/blob/0f823e79f303b4a93ef1381badb1e65757e5070f/tqdm/std.py#L1226

Sorry, but what is not the case?

It clearly uses the `mininterval`, which defaults to 100ms.

chrisjsewell:
fair, are you sure this is slower than on develop?
I can't see any difference:

It calls for the extraction here:
 
https://github.com/aiidateam/aiida-core/blob/9ff07c166a559f98b5b2be71537814ec00d3f18d/aiida/cmdline/commands/cmd_export.py#L224

which then runs through essentially the same iterative file extraction here:

https://github.com/aiidateam/aiida-core/blob/9ff07c166a559f98b5b2be71537814ec00d3f18d/aiida/tools/importexport/common/archive.py#L112

I'm anyhow going to try/time now some different methods of extracting that file...

ltalirz:
> fair, are you sure this is slower than on develop?
> I can't see any difference:

I tested `verdi export migrate`, where extraction is a lot faster on `develop` (it's not using this `safe_extract_tar`)

ltalirz:
Hmm... you're right, the code looks pretty similar.
I can check the timings again - perhaps it was somehow benefitting from caching when I tested on `develop`?

chrisjsewell:
> it's not using this safe_extract_tar

yeh its using `extract_tar`, which was basically renamed to `safe_extract_tar` and the new progress bar implmentation added.

chrisjsewell:
First test (on my macbook 2 GHz Quad-Core Intel Core i5), double-clicking on file (after changing extension to `tar.gz`): ~ 3 minutes

ltalirz:
On `develop`: Stopped ~38s for 20%, i.e. ~95s for 100%
```
$ verdi export migrate two_dimensional_database.aiida two_dimensional_database_migrated.aiida
EXTRACTING: path                          37.0%|█████▌         | 229040/618360
Aborted!
```

On your branch: takes some minutes to even start extraction; then clearly super slow

E.g. takes ~40s for 0.1%, i.e. 40'000s for 100%
```
$ verdi export migrate two_dimensional_database.aiida two_dimensional_database_migrated.aiida
Migration pathway: 0.3 -> 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8 -> 0.9
Extracting archive to temporary folder
                                           0.0%|                         | 0/1
```

Are you saying you don't see this on your machine?

chrisjsewell:
How long does it take to run the extraction through a standard extraction tool on your computer?

chrisjsewell:
Also do you see the same speed if you run it against the 1.4.2 tag code?

(just checking if it is actually unpacking all the files in develop)

ltalirz:
> How long does it take to run the extraction through a standard extraction tool on your computer?

144s

Before I do any further tests, can you please confirm that you also see what I am seeing on this PR vs `develop`?

chrisjsewell:
> Before I do any further tests, can you please confirm that you also see what I am seeing on this PR vs `develop`?

You'll have to bare with me, I just running:

```python
In [4]: import tarfile

In [5]: %%timeit
   ...: with tarfile.open("two_dimensional_database.tar.gz", 'r:*', format=tarfile.PAX_FORMAT) as handle:
   ...:     handle.extractall("output_tar")
   ...: 
```

which is taking way longer than 3 minutes!

my thinking is that perhaps there was a bug introduced, whereby the current develop is not actually extracting all files.
The progress is updated for all members, but actually it will skip extraction for:

```python
                    elif not member.name.startswith(nodes_export_subfolder + os.sep):
                        continue
```

If you look in the migrated file, do you actually see all the files you would expect?

chrisjsewell:
feels a bit unbelievable that it would take 144s to extract with built-in tools and only 95s to extract via a python loop!

ltalirz:
> perhaps there was a bug introduced

What if you just extract one of the zipped test archives, recompress it using tar, and then re-run the tests on develop (you can just replace one of the zipped ones)?

ltalirz:
> feels a bit unbelievable that it would take 144s to extract with built-in tools and only 95s to extract via a python loop!

true

ltalirz:
> If you look in the migrated file

I've never let it run through; takes too long. Let's test this with small archives

chrisjsewell:
fixed it: 12960b4 🤦 
but before you try again, I'm going to make a few more improvements

ltalirz:
Thanks @chrisjsewell - let me know when I should check it again (if you already tested it, please report your results here; there's no need for us to test twice).

chrisjsewell:
@ltalirz, ok I've moved the compression code to the `common` module and added a progress bar.

Run through below with both zip and tar.
The longest part is the recompression. As such, the last thing I am going to do is figure the best way in `cmd_import` to not have to (needlessly) compress after the migration, allowing the import to read directly on the migrated tempfolder.

```
(base) aiida@162adc8d0bf8:/$ verdi export migrate tmp/mount_folder/two_dimensional_database.tar.gz tmp/mount_folder/output_08.zip --verbosity DEBUG --version 0.8
Reading archive version
Migration pathway: 0.3 -> 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8
Extracting archive to temporary folder
Extracting tar files                     100.0%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 618360/618360
Performing migrations: 0.7 -> 0.8        100.0%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5
Flushing cache
Re-compressing archive as 'zip'
Compressing objects as zip               100.0%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 614731/614731
Moving archive to: tmp/mount_folder/output_08.zip
Cleaning temporary folder
Success: migrated the archive to version 0.8
(base) aiida@162adc8d0bf8:/$ verdi export migrate tmp/mount_folder/output_08.zip tmp/mount_folder/output_09.tar.gz --verbosity DEBUG --version 0.9 -F tar.gz
Reading archive version
Migration pathway: 0.8 -> 0.9
Extracting archive to temporary folder
Extracting zip files                     100.0%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 614731/614731
Performing migrations: 0.8 -> 0.9        100.0%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1
Flushing cache
Re-compressing archive as 'tar.gz'
Compressing objects as tar               100.0%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 614731/614731
Cleaning temporary folder
Success: migrated the archive to version 0.9
```

chrisjsewell:
Ok done. So there is no longer a compression/extraction during an `verdi import` auto-migration (it reads directly from the migrated folder in the temp folder):

```
$ verdi import -w https://archive.materialscloud.org/record/2018.0001/v1
Info: retrieving archive URLS from https://archive.materialscloud.org/record/2018.0001/v1
Success: 2 archive URLs discovered and added
Info: downloading archive https://archive.materialscloud.org/record/file?record_id=20&file_id=c8c80235-cf81-4660-b0cf-3ca5824c3725&filename=SSSP_efficiency_pseudos.aiida
Success: archive downloaded, proceeding with import
Info: incompatible version detected for https://archive.materialscloud.org/record/file?record_id=20&file_id=c8c80235-cf81-4660-b0cf-3ca5824c3725&filename=SSSP_efficiency_pseudos.aiida, trying migration
Reading archive version
Migration pathway: 0.3 -> 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8 -> 0.9
Extracting archive to work directory
Info: proceeding with import of migrated archive                                                                                                                                            

IMPORT
--------  ---------
Archive   extracted

Parameters
--------------------------  ------
Comment rules               newest
New Node Extras rules       import
Existing Node Extras rules  kcl
                                                                                                                                                                                            
Summary
-----------------------  ---------------
Auto-import Group label  20201102-132157
User(s)                  1 existing
Node(s)                  85 existing
Group(s)                 1 existing

Success: imported archive https://archive.materialscloud.org/record/file?record_id=20&file_id=c8c80235-cf81-4660-b0cf-3ca5824c3725&filename=SSSP_efficiency_pseudos.aiida
Info: downloading archive https://archive.materialscloud.org/record/file?record_id=20&file_id=bb240983-f808-4a0c-aba9-ce4a728820ac&filename=SSSP_accuracy_pseudos.aiida
Success: archive downloaded, proceeding with import
Info: incompatible version detected for https://archive.materialscloud.org/record/file?record_id=20&file_id=bb240983-f808-4a0c-aba9-ce4a728820ac&filename=SSSP_accuracy_pseudos.aiida, trying migration
Reading archive version
Migration pathway: 0.3 -> 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8 -> 0.9
Extracting archive to work directory
Info: proceeding with import of migrated archive                                                                                                                                            

IMPORT
--------  ---------
Archive   extracted

Parameters
--------------------------  ------
Comment rules               newest
New Node Extras rules       import
Existing Node Extras rules  kcl
                                                                                                                                                                                            
Summary
-----------------------  ---------------
Auto-import Group label  20201102-132200
User(s)                  1 existing
Node(s)                  85 existing
Group(s)                 1 existing

Success: imported archive https://archive.materialscloud.org/record/file?record_id=20&file_id=bb240983-f808-4a0c-aba9-ce4a728820ac&filename=SSSP_accuracy_pseudos.aiida
```

chrisjsewell:
@ltalirz ready for your review 👍

aiida/cmdline/commands/cmd_import.py
```diff
@@ -224,8 +224,9 @@ def _import_archive(archive: str, web_based: bool, import_kwargs: dict, try_migr
                 echo.echo_info(f'incompatible version detected for {archive}, trying migration')
                 try:
                     migrator = get_migrator(detect_archive_type(archive_path))(archive_path)
-                    archive_path = temp_folder.get_abs_path('migrated_archive.zip')
-                    migrator.migrate(EXPORT_VERSION, archive_path, out_compression='zip')
+                    archive_path = migrator.migrate(
+                        EXPORT_VERSION, None, out_compression='none', work_dir=temp_folder.abspath
```
ltalirz, line 228: I thought we were always going to specify the function arguments ;-)

aiida/tools/importexport/archive/common.py
```diff
@@ -61,3 +78,384 @@ def detect_archive_type(in_path: str) -> str:
         'Unable to detect the input file format, it is neither a '
         'folder, tar file, nor a (possibly compressed) zip file.'
     )
+
+
+def read_file_in_zip(filepath: str, path: str) -> str:
```
ltalirz, line 83: Should this also be called `safe_read_file_in_zip` then?

ltalirz (commented):
thanks @chrisjsewell - just one request: could you check whether there currently are tar files in the test set?
If not, could you please add one (e.g. use one of the early zipped archives and tar it), so that we are safe from accidentally breaking tar in the future.

aiida/tools/importexport/archive/common.py
```diff
@@ -61,3 +78,384 @@ def detect_archive_type(in_path: str) -> str:
...
+
+    :param callback_description: the description to return in the callback
+
+    :raises `~aiida.tools.importexport.common.exceptions.CorruptArchive`: if the file cannot be read
+
+    """
+    _filter = _get_filter(only_prefix, ignore_prefix)
+    try:
+        with zipfile.ZipFile(in_path, 'r', allowZip64=True) as handle:
+            callback('init', {'total': 1, 'description': 'Computing zip objects to extract'})
```
ltalirz, line 170:
```suggestion
            callback('init', {'total': 1, 'description': 'Gathering list of files to extract from zip'})
```

aiida/tools/importexport/archive/common.py
```diff
@@ -61,3 +78,384 @@ def detect_archive_type(in_path: str) -> str:
...
+            members = [name for name in handle.namelist() if _filter(name)]
```
ltalirz, line 171: Just to understand: Is it necessary/more performant to build this list first?
Why not just iterate over`handle.namelist()` and `continue` if the name doesn't fit
This way it's easier for users to follow the progress.

aiida/tools/importexport/archive/migrators.py
```diff
@@ -0,0 +1,267 @@
...
+            self._extract_archive(extracted, callback)
+
+        with CacheFolder(extracted) as folder:
+            with get_progress_reporter()(total=len(pathway), desc='Performing migrations: ') as progress:
+                for from_version in pathway:
+                    to_version = MIGRATE_FUNCTIONS[from_version][0]
+                    progress.set_description_str(
+                        f'Performing migrations: {from_version} -> {to_version}', refresh=False
+                    )
+                    progress.update()
```
ltalirz, line 181: I guess this update meant to update the description of the progress bar (?)

For me it showed 
```
Performing migrations:            0.0%
```
for quite some time - perhaps you want to force a refresh here.

ltalirz:
@chrisjsewell In my test, the run in this PR takes quite a bit more memory than both `develop` and `v1.4.2`. Can we get this down?

It also runs slower than `develop`, but if I understand correctly, we accidentally introduced a bug into the migration (is that correct? if yes, can we add a test that would have flagged this?).


This PR: 16m18s; top memory consumption ~2200 MB
```
$ time verdi export migrate two_dimensional_database.aiida 2d_migrated_develop.aiida
Reading archive version
^[Migration pathway: 0.3 -> 0.4 -> 0.5 -> 0.6 -> 0.7 -> 0.8 -> 0.9
Extracting archive to work directory
Re-compressing archive as 'zip'
Moving archive to: 2d_migrated_develop.aiida
Success: migrated the archive to version 0.9

real    16m18.294s
user    12m1.563s
sys     1m43.576s
```
image


`develop`: 12m28s; top memory consumption ~1400MB
```
$ time verdi export migrate two_dimensional_database.aiida 2d_migrated_develop.aiida
Success: migrated the archive from version 0.3 to 0.9

real    12m28.035s
user    8m39.272s
sys     1m27.630s
```
image

`v1.4.2`: 19m30s; top memory consumption ~1400MB
```
$ time verdi export migrate two_dimensional_database.aiida 2d_migrated_v1.4.2.aiida
Success: migrated the archive from version 0.3 to 0.9

real    19m30.655s
user    10m21.675s
sys     1m48.365s
```
image

aiida/cmdline/commands/cmd_import.py
chrisjsewell: "do as I say, not as I do" 😆
but anyway here the `*` in the `migrate` method arguments is placed after the version and filename, which are variables that will always be required for a migration.
I don't think it's necessary for absolutely all arguments but certainly for most (if not all) optional keyword arguments and/or when there is more than 1 or 2 arguments

aiida/tools/importexport/archive/common.py
chrisjsewell: I'd say no, because there is no specific safety features in the read logic (although I don't think there is anything to make it unsafe)

aiida/tools/importexport/archive/common.py
chrisjsewell: It wouldn't make it easier to follow progress when you are providing a prefix filter (as is done in the archive reader / import code), whereby you may only be extracting a subfolder within the zip folder. Then the progress bar would be very misleading, because it might list 100,000s of objects to extract but actually only a small subset of them are really extracted.

chrisjsewell:
> Can we get this down?

Firstly I looked at procpath and it seems way over-complicated 😬.
See below, I've used https://github.com/jeetsukumaran/Syrupy which is a lot simpler and easier to customise.

In 4dd7fa807c0b1652558be9f1454bfbe63c9eec00 I have reduced the memory usage to the same as develop (and also it looks reduced the time a bit)

This is the run when only making the `_write_object` modification. 

```console
$ cd tmp
$ syrupy.py -i 1 --separator=, --no-align verdi export migrate mount_folder/two_dimensional_database.tar.gz mount_folder/output_09.zip
SYRUPY: Writing process resource usage samples to 'syrupy_20201103014840.ps.log'
SYRUPY: Writing raw process resource usage logs to 'syrupy_20201103014840.ps.raw'
SYRUPY: Executing command 'verdi export migrate mount_folder/two_dimensional_database.tar.gz mount_folder/output_09.zip'
SYRUPY: Redirecting command output stream to 'syrupy_20201103014840.out.log'
SYRUPY: Redirecting command error stream to 'syrupy_20201103014840.err.log'
SYRUPY: Completed running: verdi export migrate mount_folder/two_dimensional_database.tar.gz mount_folder/output_09.zip
SYRUPY: Started at 2020-11-03 01:48:40.621727
SYRUPY: Ended at 2020-11-03 02:12:45.366769
SYRUPY: Total run time: 0 hour(s), 24 minute(s), 04.745042 second(s)
$ python -c 'import pandas as pd; ax = pd.read_csv("syrupy_20201103014840.ps.log").set_index("ELAPSED").plot(y="RSS", grid=True); ax.get_figure().savefig("mount_folder/output.png")'
```

output

This is the run when also removing the full `json.dumps` check in `write_json`

```console
$ syrupy.py -i 1 --separator=, --no-align verdi export migrate mount_folder/two_dimensional_database.tar.gz mount_folder/output_09.zipSYRUPY: Writing process resource usage samples to 'syrupy_20201103023904.ps.log'
SYRUPY: Writing raw process resource usage logs to 'syrupy_20201103023904.ps.raw'
SYRUPY: Executing command 'verdi export migrate mount_folder/two_dimensional_database.tar.gz mount_folder/output_09.zip'
SYRUPY: Redirecting command output stream to 'syrupy_20201103023904.out.log'
SYRUPY: Redirecting command error stream to 'syrupy_20201103023904.err.log'
SYRUPY: Completed running: verdi export migrate mount_folder/two_dimensional_database.tar.gz mount_folder/output_09.zip
SYRUPY: Started at 2020-11-03 02:39:04.003449
SYRUPY: Ended at 2020-11-03 03:01:28.654327
SYRUPY: Total run time: 0 hour(s), 22 minute(s), 24.650878 second(s)
$ python -c 'import pandas as pd; ax = pd.read_csv("syrupy_20201103023904.ps.log").set_index("ELAPSED").plot(y="RSS", grid=True); ax.get_figure().savefig("mount_folder/output.png")'
```

output

chrisjsewell:
> It also runs slower than develop, but if I understand correctly, we accidentally introduced a bug into the migration (is that correct? if yes, can we add a test that would have flagged this?).

No the bug was in this PR, for unpacking tar. You would be able to add a normal test, since it was just writing the same files multiple times, but not actually causing an error (just overwriting it).
Obviously this could be picked up in a benchmark timing test.

aiida/tools/importexport/archive/migrators.py
chrisjsewell: fixed in 026b5cb5d755e249407ab2a463529e369523e3c4

chrisjsewell:
> could you check whether there currently are tar files in the test set?
If not, could you please add one (e.g. use one of the early zipped archives and tar it), so that we are safe from accidentally breaking tar in the future.

Added in 348242f0892e63c109a48a1db2e681396f1eaaba

ltalirz:
> In [4dd7fa8](https://github.com/aiidateam/aiida-core/commit/4dd7fa807c0b1652558be9f1454bfbe63c9eec00) I have reduced the memory usage to the same as develop

Thanks for fixing!

> This is the run when also removing the full `json.dumps` check in `write_json`

Ah, I see... I was wondering where this weird oscillation came from; so it's throwing the JSON string away but only after it reaches some buffer size?


> No the bug was in this PR, for unpacking tar.

Ok.

ltalirz (approved):
thanks @chrisjsewell , this is good to go from my side

chrisjsewell:
Thanks!

> so it's throwing the JSON string away but only after it reaches some buffer size?

Yeh I guess something like that 🤷‍♂️

PR #4534, opened 2020-10-30
Archive export refactor (2)

chrisjsewell:
FYI @ltalirz just a heads up, this is the final PR I alluded to (written on top of #4532) 

Based on the experience from the import refactor, it refactors the archive writer to be somewhat similar; using it as a context manage and "streaming" the data to its methods, as opposed to now where all the data is passed to it in a single dataclass.

The refactor also removes a few spurious database queries; merging them in to the entity queries.

~~A `--batch-size` argument is also added to the CLI, to allow control of query batch sizing.~~

The implementation tentatively works for zip so far (as in, that is what I was running in the meeting demo) but requires a bit more work to clean up and apply to tar as well.


ltalirz:
Ok, I will check this after #4532 is merged.

> A `--batch-size` argument is also added to the CLI, to allow control of query batch sizing.

Is this already impacting memory significantly?

If not, I suggest to keep this parameter at the python API only for the moment and don't expose it on the cli yet - we should limit the number of flags and having the flag there raises expectations that we may not be able to meet (yet).

aiida/tools/importexport/common/config.py
```diff
@@ -178,7 +178,13 @@ def get_all_fields_info():
             'related_name': 'dbnodes'
         },
         'description': {},
-        'process_type': {}
+        'process_type': {},
+        'extras': {
```
ltalirz, line 182: This changes the export format, right?

I.e. one can now either have export files where the node extras are in `all_fields_info` or where the extras are in `node_extras` (same for attributes).

If that is what we want, we should at the very least document it (but it might make sense to mint a new version, since otherwise we have to support both options in the code).

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -107,6 +107,7 @@ def export(
     allowed_licenses: Optional[Union[list, Callable]] = None,
     forbidden_licenses: Optional[Union[list, Callable]] = None,
     writer_init: Optional[Dict[str, Any]] = None,
+    batch_size: int = 100,
```
ltalirz, line 110: Is 100 a reasonable default?
This needs performance testing.

aiida/tools/importexport/archive/writers.py
```diff
@@ -91,81 +82,142 @@ def file_format_verbose(self) -> str:
...
+        }
+
+    def write_link(self, data: Dict[str, str]):
+        # input: str, output: str, label: str, type: str
+        self._link_uuids.append(data)
+
+    def write_group_uuids(self, uuid: str, node_uuids: List[str]):
+        self._group_uuids[uuid] = node_uuids
+
+    def write_entity_data(self, name: str, pk: str, fields: Dict[str, Any]):
```
ltalirz, line 211: pk comes as a str here?

ltalirz (commented):
thanks @chrisjsewell 
I started having a look through and adding comments; I then saw that it's still in draft status

let me know when I should re-review this

chrisjsewell:
> I started having a look through and adding comments; I then saw that it's still in draft status

Yes indeed patience lol.
The general structure of the `export` function I believe is finished, but the `writers.py` I am working on now, since at this point is a bit hacky just to get an initial working implementation.

aiida/tools/importexport/archive/writers.py
chrisjsewell: changed already

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: No idea and indeed would be best to performance test, but this is already the default value for `QueryBuilder.iterall` so it is not actually changing anything from the original code

aiida/tools/importexport/common/config.py
chrisjsewell: no the node extras are still stored in `node_extras`, same for attributes.
Then currently they are merged in the reader method before yielding the data, i.e. from the interface it does not matter that they are stored separately.
The only difference to the archive file will be these extra entries in the metadata,
but this is currently mitigated (so that you don't need a migration for older files) by already existing lines 383-384 in `aiida/tools/importexport/dbimport/backends/django.py`:

```python
    if entity_name == NODE_ENTITY_NAME:
        # in the current export process these fields are not present,
        # because they are retrieved by a separate query
        fields_info.setdefault('attributes', {'convert_type': 'jsonb'})
        fields_info.setdefault('extras', {'convert_type': 'jsonb'})
```

these lines can be removed when we add a migration for the next version.

aiida/tools/importexport/archive/writers.py
```diff
@@ -91,81 +82,142 @@ def file_format_verbose(self) -> str:
...
-        'unique_identifiers': export_data.metadata.unique_identifiers,
-        'export_parameters': {
-            'graph_traversal_rules': export_data.metadata.graph_traversal_rules,
-            'entities_starting_set': export_data.metadata.entities_starting_set,
-            'include_comments': export_data.metadata.include_comments,
-            'include_logs': export_data.metadata.include_logs,
-        },
-        'conversion_info': export_data.metadata.conversion_info
-    }
+    def close(self, excepted: bool):
```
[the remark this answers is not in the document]
chrisjsewell: it allows the writer to decide whether it wants to actually write the output file, given that the export process excepted

aiida/tools/importexport/archive/writers.py
```diff
@@ -91,81 +82,142 @@ def file_format_verbose(self) -> str:
...
+        self._in_context = True
+        # reset the export information
+        self._info = {}
+        self.open()
+        return self
+
+    def __exit__(
+        self, exctype: Optional[Type[BaseException]], excinst: Optional[BaseException], exctb: Optional[TracebackType]
+    ):
+        """Open the contextmanager """
```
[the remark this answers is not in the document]
chrisjsewell: already changed

aiida/tools/importexport/archive/writers.py
```diff
@@ -41,45 +49,28 @@ def get_writer(file_format: str) -> Type['ArchiveWriterAbstract']:
...
     @property
-    def filename(self) -> str:
-        """Return the filename to write to."""
-        return self._filename
+    def filepath(self) -> Path:
+        """Return the filepath to write to."""
+        return self._filepath
+
+    @property
+    def get_export_info(self):
```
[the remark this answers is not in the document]
chrisjsewell: already changed

aiida/tools/importexport/common/config.py
ltalirz: Ah, I see. Perhaps add a comment to that effect also here in the schema

aiida/cmdline/commands/cmd_export.py
```diff
@@ -101,10 +101,18 @@ def inspect(archive, version, data, meta_data):
     show_default=True,
     help='Include or exclude comments for node(s) in export. (Will also export extra users who commented).'
 )
+@click.option(
```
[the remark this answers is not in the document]
chrisjsewell: This allows for performance (cpu/memory) testing from the CLI, as we have done for other PRs (see below). I would just comment it out before merging, because it is definitely something very useful and something that will be beneficial if not now then for the new format.

chrisjsewell:
@ltalirz this is now ready for review
(I have not yet run performance testing)

chrisjsewell:
With ~40,000 node repo

develop:

```console
(base) aiida@088cc428b4c4:/$ cd tmp
(base) aiida@088cc428b4c4:/tmp$ syrupy.py -i 1 --separator=, --no-align verdi export create -G 1 -- mount_folder/export_develop.aiida
SYRUPY: Writing process resource usage samples to 'syrupy_20201105105553.ps.log'
SYRUPY: Writing raw process resource usage logs to 'syrupy_20201105105553.ps.raw'
SYRUPY: Executing command 'verdi export create -G 1 -- mount_folder/export_develop.aiida'
SYRUPY: Redirecting command output stream to 'syrupy_20201105105553.out.log'
SYRUPY: Redirecting command error stream to 'syrupy_20201105105553.err.log'
SYRUPY: Completed running: verdi export create -G 1 -- mount_folder/export_develop.aiida
SYRUPY: Started at 2020-11-05 10:55:53.406105
SYRUPY: Ended at 2020-11-05 10:59:35.599326
SYRUPY: Total run time: 0 hour(s), 03 minute(s), 42.193221 second(s)
(base) aiida@088cc428b4c4:/tmp$ python -c 'import pandas as pd; ax = pd.read_csv("syrupy_20201105105553.ps.log").set_index("ELAPSED").plot(y="RSS", grid=True); ax.get_figure().savefig("mount_folder/output_develop.png")'
```

output_develop

this PR:

```console
(base) aiida@088cc428b4c4:/tmp$ syrupy.py -i 1 --separator=, --no-align verdi export create -G 1 -- mount_folder/export_new.aiida
SYRUPY: Writing process resource usage samples to 'syrupy_20201105110111.ps.log'
SYRUPY: Writing raw process resource usage logs to 'syrupy_20201105110111.ps.raw'
SYRUPY: Executing command 'verdi export create -G 1 -- mount_folder/export_new.aiida'
SYRUPY: Redirecting command output stream to 'syrupy_20201105110111.out.log'
SYRUPY: Redirecting command error stream to 'syrupy_20201105110111.err.log'
SYRUPY: Completed running: verdi export create -G 1 -- mount_folder/export_new.aiida
SYRUPY: Started at 2020-11-05 11:01:11.163114
SYRUPY: Ended at 2020-11-05 11:03:34.289323
SYRUPY: Total run time: 0 hour(s), 02 minute(s), 23.126209 second(s)
(base) aiida@088cc428b4c4:/tmp$ python -c 'import pandas as pd; ax = pd.read_csv("syrupy_20201105110111.ps.log").set_index("ELAPSED").plot(y="RSS", grid=True); ax.get_figure().savefig("mount_folder/output_new.png")'
```

output_new

```console
(base) aiida@088cc428b4c4:/tmp$ verdi export create -G 1 --verbosity DEBUG -- mount_folder/export_new2.aiida
/
EXPORT
--------------  ------------------------------
Archive         mount_folder/export_new2.aiida
Format          JSON Zip (compression=8)
Export version  0.9

Inclusion rules
-----------------  ----
Include Comments   True
Include Logs       True

Traversal rules
---------------------------------  -----
Follow links input calc forwards   False
Follow links input calc backwards  True
Follow links create forwards       True
Follow links create backwards      True
Follow links return forwards       True
Follow links return backwards      False
Follow links input work forwards   False
Follow links input work backwards  True
Follow links call calc forwards    True
Follow links call calc backwards   True
Follow links call work forwards    True
Follow links call work backwards   True

STARTING EXPORT...
Collecting nodes in groups               100.0%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 40065/40065
Traversing provenance via links ...      100.0%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1
WRITING METADATA...
Writing links                            100.0%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 40064/40064
Building entity database queries         100.0%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4
Writing entity data                      100.0%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 40066/40066
Writing group UUID -> [nodes UUIDs]
Exporting node repositories: 40065       100.0%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 40065/40065
FINALIZING EXPORT...
Exported Entities:
  - Node  : 40065
  - User  : 1
  - Group : 1

Success: wrote the export archive file to mount_folder/export_new2.aiida
```

ltalirz:
Thanks for the performance tests!

Do you know why the memory usage increases?

And what are the units here - are 30'000 = 30 GB? That's a lot (of course also before this PR)

aiida/tools/importexport/archive/common.py
```diff
@@ -47,7 +47,7 @@ class ArchiveMetadata:
     # optional data
     graph_traversal_rules: Optional[Dict[str, bool]] = dataclasses.field(default=None)
     # Entity type -> UUID list
-    entities_starting_set: Optional[Dict[str, Set[str]]] = dataclasses.field(default=None)
+    entities_starting_set: Optional[Dict[str, List[str]]] = dataclasses.field(default=None)
```
ltalirz, line 50: I guess there is a good reason for this change; just mentioning that `set` is also in the name of the attribute...

aiida/tools/importexport/archive/zip_path.py
```diff
@@ -0,0 +1,325 @@
...
+
+    For writing zip files, you should use within a context manager, or directly call the ``close`` method::
+
+        with ZipPath('path/to/file.zip', mode='w', compression=zipfile.ZIP_DEFLATED) as path:
+            (path / 'new_file.txt').write_text('hallo world')
+            # there are also some features equivalent to shutil
+            (path / 'other_file.txt').copyfile('path/to/external_file.txt')
+            (path / 'other_folder').copytree('path/to/external_folder', pattern='*')
+
+    Note with zipfile is not possible to overwrite existing files (it will raise a ``FileExistsError``).
```
ltalirz, line 46:
```suggestion
    Note that ``zipfile`` does not allow to overwrite existing files (it will raise a ``FileExistsError``).
```

aiida/tools/importexport/archive/zip_path.py
```diff
@@ -0,0 +1,325 @@
...
+from pathlib import Path
+import posixpath
+from types import TracebackType
+from typing import cast, Iterable, Optional, Set, Type, Union
+import zipfile
+
+__all__ = ('ZipPath',)
+
+
+class ZipPath:
```
ltalirz, line 25: nice!

aiida/tools/importexport/common/config.py
ltalirz: (meaning please add a comment here that these fields are not actually used)

chrisjsewell:
> And what are the units here - are 30'000 = 30 GB? That's a lot (of course also before this PR)

I think your missing a zero there lol, its in bytes so 3Gb

chrisjsewell:
> Do you know why the memory usage increases?

Could not say off hand

ltalirz (commented):
thanks @chrisjsewell ; looks good to me

ltalirz:
> I think your missing a zero there lol, its in bytes so 3Gb

Ah, right. Thanks, that is much more sensible.

> Could not say off hand

Ok. Might be worth a quick look, since memory can be a hard limit, i.e. it may prevent people from exporting.

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -233,193 +226,116 @@ def export(
...
+
+        # check that no nodes are being exported with incorrect licensing
+        _check_node_licenses(node_ids_to_be_exported, allowed_licenses, forbidden_licenses)
+
+        # write the link data
+        if traverse_output['links']:
+            with get_progress_reporter()(total=len(traverse_output['links']), desc='Writing links') as progress:
+                for link in traverse_output['links']:
+                    progress.update()
+                    writer_context.write_link({
```
ltalirz, line 281: so, this interface seems to assume that links are to be written one at a time (not streamed).
should they not be streamed as well eventually?

ltalirz: ok, I see now that also the entities are just written one at a time.
I guess when writing the commit message for this PR, we'll need to be a bit careful with the word "stream" since one could understand something different

chrisjsewell: I'm not sure if I understand this; they are "streamed" to the writer one at a time.
The write can decide how it writes them; caching batches before writing, or otherwise.

Are you suggesting that you should send the links to the writer already in batches?

chrisjsewell: what would you understand by stream?

ltalirz: > what would you understand by stream?

On second thought, I think you are right and your use of "stream" makes sense.

aiida/tools/importexport/archive/writers.py
```diff
@@ -35,176 +43,244 @@ def get_writer(file_format: str) -> Type['ArchiveWriterAbstract']:
...
+                'include_comments': data.include_comments,
+                'include_logs': data.include_logs,
+            },
+            'conversion_info': data.conversion_info
+        }
+        with self._zippath.joinpath('metadata.json').open('wb') as handle:
+            json.dump(metadata, handle)
+
+    def write_link(self, data: Dict[str, str]):
+        self._data['links_uuid'].append(data)
```
ltalirz, line 269: A long shot concerning the memory usage:
What exactly happens under the hood here?

Python lists are variable-length arrays, i.e. over the course of `write_link` being called N times for N links it probably needs to reallocate the array multiple times.

If the implementation is clever, it might realize that it needs to do this super often and therefore get more and more liberal in the size of the pre-allocation.

However, this goes to an implementation issue - at least for the current implementation, we know in advance how many links we will have, so it would make sense to communicate this to the writer in some way.

chrisjsewell:
@ltalirz I added a null writer in b3c0ea2, which I think is handy for the performance analysis, but also can be used for test runs of an export.
Here is the output of using that:

output_null

chrisjsewell:
This is with `self._data['links_uuid'].append(data)` commented out:

output_other

and this is with also the `write_node_repo_folder` code commented out:

output_other2

chrisjsewell:
out of interest, I tried the memory for reading the 2-D archive (614731 objects):

```
syrupy.py -i 1 --separator=, --no-align python -c "import zipfile; z = zipfile.ZipFile('mount_folder/two_dimensional_database.aiida')"
```

zip_read

aiida/tools/importexport/archive/writers.py
chrisjsewell: this does not appear to be an issue (see performance analysis in comments)

aiida/tools/importexport/common/config.py
chrisjsewell: > (meaning please add a comment here that these fields are not actually used)

these fields are used though; the writer pops out `node_extras` and `node_attributes` from the fields, before storing to `data.json` (to keep the existing format), but then the reader re-combines them before returning the fields.
The only reason they have to be added here is that, up until this PR, these fields were not stored in fields_info in the archive `metadata.json`.

aiida/tools/importexport/archive/common.py
chrisjsewell: yeh I went back and forth on this, because it has to be converted to a list before storing as a json and so it was a little easier to convert before passing to the writer.
but your right that the naming is now a little off

aiida/cmdline/commands/cmd_export.py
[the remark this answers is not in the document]
chrisjsewell: commented out

chrisjsewell:
Ooops I got the memory size wrong.
Its in kilobytes, and so the max memory is actually ~0.4GB, so yeh not too big an issue

aiida/tools/importexport/archive/zip_path.py
```diff
@@ -114,10 +117,20 @@ def _all_at_set(self) -> Set[str]:
         """Iterate through all file and directory paths in the zip file.
 
         Note: this is necessary, since the zipfile does not strictly store all directories in the namelist.
```
ltalirz, line 119: Just wondering: so what is the zipfile policy for storing stuff in the namelist?
Is the problem that empty directories aren't stored?

ltalirz:
> Its in kilobytes, and so the max memory is actually ~0.4GB for `two_dimensional_database.aiida`, so yeh not too big an issue

So, just to understand: you recompressed the AiiDA archive file for the 2d materials DB with zip, and these 400MB are just the overhead for building an index of the file in python using `zipfile`, correct?
That is quite substantial considering that this is comparable to the size of the zipfile itself, and 614'731 files and folders is actually not super large, one can easily imagine an archive with an order of magnitude more.

Are we building this index in the export as well?
At least for the export it seems entirely unnecessary to do this (?)

ltalirz:
Another test on a larger export

`develop`: timing - 53m0s
image

0f1cb3c073e1b973361c2d386ec82ca8d9aacce3 : timing 36m45s (to take with a grain of salt, since it an often be faster the second time)
image


Conclusion: It seems we've gotten rid of the early spike in memory consumption from `develop` - memory consumption towards the end is again increased by ~10% compared to the current `develop` but to me that is acceptable.

chrisjsewell:
sorry @ltalirz not quite done with this yet 😬 
As you may have noticed, I have moved the compression code to https://github.com/aiidateam/archive-path (and added `TarPath` there)

One of the big improvements is that both `read_file_in_tar` and `read_file_in_zip` are now optimised to only read the index up to the required file (and not store the index). So now, for example, reading the archive version (from metadata.json) is a lot faster:

old zip:

```console
$ time verdi export migrate mount_folder/two_dimensional_database.zip --in-place
Reading archive version
No migration required
Success: migrated the archive to version 0.9

real    0m10.906s
user    0m9.072s
sys     0m1.315s
```

old zip:

```console
$ time verdi export migrate mount_folder/two_dimensional_database.zip --in-place
Reading archive version
No migration required
Success: migrated the archive to version 0.9

real    0m2.744s
user    0m1.916s
sys     0m0.565s
```

old tar.gz:

```console
$ time verdi export migrate mount_folder/two_dimensional_database.tar.gz --in-place --version 0.3
Reading archive version
No migration required
Success: migrated the archive to version 0.3

real    2m15.946s
user    2m9.304s
sys     0m3.338s
```

new tar.gz:

```console
$ time verdi export migrate mount_folder/two_dimensional_database.tar.gz --in-place --version 0.3
Reading archive version
No migration required
Success: migrated the archive to version 0.3

real    0m8.916s
user    0m6.229s
sys     0m1.013s
```

chrisjsewell:
The other thing is, at this moment for zip, I've added optional caching of the index (i.e. of the `ZipInfo` dict). Naturally, this uses less memory, but is slower, so I'm not sure if its worth having in a default:

un-cached export:

no_cached_export

cached export:

cached_export

ltalirz:
> The other thing is, at this moment for zip, I've added optional caching of the index (i.e. of the `ZipInfo` dict, see `WriterJsonZip.open`). Naturally, this uses less memory, but is slower, so I'm not sure if its worth having as a default:

I see - when you say "caching" you mean writing the index to disk?
What is causing the slowness here - if it is the writing to disk, then that could probably be mitigated by batching those.

Moving certain parts of the code out to a dependency is ok, please just keep in mind that you will also need a conda-forge recipe for those then.

chrisjsewell:
> I see - when you say "caching" you mean writing the index to disk?
> What is causing the slowness here - if it is the writing to disk, then that could probably be mitigated by batching those.

It simply replaces the standard dict with a https://docs.python.org/3/library/shelve.html one. There is no batching available. Obviously you could write your own implementation, to be more efficient, but thats a load more code.

> please just keep in mind that you will also need a conda-forge recipe for those then.

way ahead of you 😉 : https://github.com/conda-forge/staged-recipes/pull/13145 (just fixed the dist to include the LICENSE)

aiida/tools/importexport/archive/zip_path.py
chrisjsewell: basically when you store a file (or directory) it just stores the path to that file in the index, not any of the parent directories. Then when extracting it makes sure the directories are created.
So if you want to check that a directory exists, you can't just look in the index, you have to also check parent directories

chrisjsewell:
Note that now tar files are also written directly (not written to a folder first then compressed), its a lot quicker:

develop:

```console
$ time verdi export create -G 1 -F tar.gz --force -- mount_folder/export_new4.aiida
...
Exported Entities:
  - Node  : 40065
  - User  : 1
  - Group : 1

Success: wrote the export archive file to mount_folder/export_new4.aiida

real    7m40.950s
user    2m47.366s
sys     1m38.453s
```

now:

```console
$ time verdi export create -G 1 -F tar.gz --force -- mount_folder/export_new4.aiida
...
Success: wrote the export archive file to mount_folder/export_new4.aiida

real    4m24.923s
user    2m32.837s
sys     0m42.912s

```

chrisjsewell:
Ok @ltalirz, this is definitely good to go now, just waiting on https://github.com/conda-forge/staged-recipes/pull/13145.

I've now also moved `safe_extract_tar`/`safe_extract_zip` to `archive-path` (as `extract_tree` method).

For the index caching, I've turned it off by default, but added `zip-lowmemory` to the format options. I think that's a good compromise for speed vs memory.
Here are the timings:

```console
(base) aiida@088cc428b4c4:/tmp$ time verdi export create -G 1 -F zip --force -- mount_folder/export_new4.aiida
/
EXPORT
--------------  ------------------------------
Archive         mount_folder/export_new4.aiida
Format          JSON Zip (compression=8)
Export version  0.9

...

Finalizing Export...                                                                        
Exported Entities:
  - Node  : 40065
  - User  : 1
  - Group : 1

Success: wrote the export archive file to mount_folder/export_new4.aiida

real    3m10.904s
user    1m59.881s
sys     0m23.813s
```

```console
$ time verdi export create -G 1 -F zip-lowmemory --force -- mount_folder/export_new4.aiida
-
EXPORT
--------------  ------------------------------
Archive         mount_folder/export_new4.aiida
Format          JSON Zip (compression=8)
Export version  0.9

Finalizing Export...                                                                        
Exported Entities:
  - Node  : 40065
  - User  : 1
  - Group : 1

Success: wrote the export archive file to mount_folder/export_new4.aiida

real    4m29.731s
user    2m20.843s
sys     0m38.281s
```

tests/tools/importexport/migration/conftest.py
```diff
@@ -54,9 +55,9 @@ def _migrate(filename_archive, version_old, version_new, migration_method, archi
         out_path = tmp_path / 'out.aiida'
 
         if zipfile.is_zipfile(archive_path):
-            safe_extract_zip(archive_path, out_path)
+            ZipPath(archive_path).extract_tree(out_path)
```
ltalirz, line 58: `extract_tree` seems to differ in style from `puttree`

tests/utils/archives.py
```diff
@@ -14,8 +14,9 @@
 import zipfile
 from typing import List
 
+from archive_path import read_file_in_tar, read_file_in_zip
```
ltalirz, line 17: just a minor nitpick: I'd probably call it `read_file_from_tar` (but not necessary to change)

aiida/cmdline/commands/cmd_export.py
```diff
@@ -80,7 +80,7 @@ def inspect(archive, version, data, meta_data):
 @options.GROUPS()
 @options.NODES()
 @options.ARCHIVE_FORMAT(
-    type=click.Choice(['zip', 'zip-uncompressed', 'tar.gz', 'null']),
+    type=click.Choice(['zip', 'zip-uncompressed', 'zip-lowmemory', 'tar.gz', 'null']),
```
ltalirz, line 83: Do we want to expose this at the CLI level already?

We just need to keep in mind that this means we should continue supporting this option until 2.0.
(probably ok)

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -104,6 +104,10 @@ def export(
         Support for the parameter `silent` will be removed in `v2.0.0`.
         Please set the log level and progress bar implementation independently.
 
+    .. deprecated:: 1.5.0
+        Support for the parameter `use_compression` will be removed in `v2.0.0`.
+        Please use `writer_init={'use_compression': True}`.
```
ltalirz, line 109: I didn't quite follow the logic of which options go into the `writer_init` and which don't - `writer_init` in my view is not very specific.
E.g. from a user perspective,`file_format`, `overwrite`, etc. also specify aspects of how the file should be written.

I guess you introduced the `writer_init` in order to avoid having yet another parameter for activating the low-memory option.
I think it's better to add it (or - what I would prefer - to stick to the strings for `file_format` and then do the translation inside the `export` function).

ltalirz (changes requested):
thanks @chrisjsewell 

I've reviewed the latest commits - some minor comments

aiida/cmdline/commands/cmd_export.py
```diff
@@ -158,10 +159,13 @@ def create(
...
         export_format = ExportFileFormat.ZIP
-        kwargs.update({'use_compression': True})
+        kwargs.update({'writer_init': {'use_compression': True}})
     elif archive_format == 'zip-uncompressed':
         export_format = ExportFileFormat.ZIP
-        kwargs.update({'use_compression': False})
+        kwargs.update({'writer_init': {'use_compression': False}})
+    elif archive_format == 'zip-lowmemory':
+        export_format = ExportFileFormat.ZIP
+        kwargs.update({'writer_init': {'cache_zipinfo': True}})
```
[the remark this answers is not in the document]
ltalirz: Please see my last comment first

aiida/tools/importexport/dbexport/__init__.py
ltalirz: Ah hm.. I see that 

> or - what I would prefer - to stick to the strings for `file_format` and then do the translation into classes inside the `export` function

Ah hm, I see that `file_format` here is already a translated version of the string passed on the CLI, so that suggestion doesn't work.
One could make it a separate class - or add the additional parameter.

aiida/cmdline/commands/cmd_export.py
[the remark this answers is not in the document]
chrisjsewell: because use_compression=True is anyhow the default. Really I don't see much use for zip-uncompressed, but it was already offered.

aiida/tools/importexport/archive/migrators.py
```diff
@@ -227,39 +227,54 @@ def _extract_archive(self, filepath: Path, callback: Callable[[str, Any], None])
     @staticmethod
     def _compress_archive_zip(in_path: Path, out_path: Path, compression: int):
         """Create a new zip compressed zip from a folder."""
-        with get_progress_reporter()(total=1) as progress:
+        with get_progress_reporter()(total=1, desc='Compressing to zip') as progress:
             _callback = create_callback(progress)
-            compress_folder_zip(in_path, out_path, compression=compression, callback=_callback)
+            with ZipPath(out_path, mode='w', compression=compression, allow_zip64=True) as path:
+                path.puttree(in_path, check_exists=False, callback=_callback, cb_descript='Compressing to zip')
```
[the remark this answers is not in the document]
chrisjsewell: No just what I choose for the `ZipPath` API. You can question that over on archive-path if you want 😉

aiida/tools/importexport/archive/writers.py
```diff
@@ -268,17 +280,20 @@ def open(self):
...
             return
         # write data.json
-        with self._zippath.joinpath('data.json').open('wb') as handle:
+        with self._archivepath.joinpath('data.json').open('wb') as handle:
             json.dump(self._data, handle)
         # close the zipfile to finalise write
-        self._zippath.close()
+        self._archivepath.close()
+        if self._cache_zipinfo:
+            self._zipinfo_cache.close()
```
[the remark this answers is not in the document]
chrisjsewell: Its a private attribute, it won't get changed unless someone access it "illegally"

aiida/tools/importexport/common/archive.py
```diff
@@ -61,7 +61,7 @@ def extract_zip(infile, folder, nodes_export_subfolder=None, check_files=('data.
         incorrect formats
     """
     warnings.warn(
-        'extract_zip function is deprecated and will be removed in AiiDA v2.0.0, use `safe_extract_zip` instead',
+        'extract_zip function is deprecated and will be removed in AiiDA v2.0.0, use the archive-path package instead',
```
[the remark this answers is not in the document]
chrisjsewell: TBH they can do whatever they want, they just can't use this function lol. `export` does not offer the same, safe by default (no paths accessed outside of the archive, no symlinks/devices) or progress callback that this function does.

tests/tools/importexport/migration/conftest.py
chrisjsewell: again you should raise this in archive-path

aiida/cmdline/commands/cmd_export.py
chrisjsewell: yep I'm happy with supporting it

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: > I didn't quite follow the logic of which options go into the writer_init and which don't

simply options that are writer specific. It is misleading to have `use_compression` as a core option, because it does not apply to tar files.

> I guess you introduced the writer_init in order to avoid having yet another parameter for activating the low-memory option.

no `writer_init` was already in the code way before this

> One could make a separate class for the writer with zipinfo caching 

no I disagree, you should not have to make a separate class for every possible option, this is what `writer_init` is for.

aiida/tools/importexport/common/archive.py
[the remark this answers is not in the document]
chrisjsewell: the "function" in question is the `extract_tree` method. But I don't think it would be useful to specify this, in case it changes in the future.
(I imagine it would be very rare anyway that someone is currently using this function)

aiida/tools/importexport/archive/writers.py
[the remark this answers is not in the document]
chrisjsewell: > Is {} the right "null" object for __zipinfo_cache?

yes it definitely is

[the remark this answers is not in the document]
chrisjsewell: Anyhow I have now changed the default to `None` 👍 (see 88d75bb)

chrisjsewell:
> I've reviewed the latest commits - some minor comments

thanks, Ive responded to the comments

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: essentially `file_format` could now maybe be more accurately described as `writer_cls`, which either points to the "entry-point name" or actual class

tests/utils/archives.py
chrisjsewell: again an archive-path matter 😄

aiida/tools/importexport/common/archive.py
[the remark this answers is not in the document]
ltalirz: Thanks - I think my original suggestion was partly a misunderstanding on my part that `safe_extract_zip` was no longer in aiida-core.

I think the main question is whether we want `safe_extract_zip` to be part of the public API (independent of nesting depth) - if yes, then your suggestion is fine, if not, we should suggest to use something else.

tests/tools/importexport/migration/conftest.py
ltalirz: given that `shutil` also uses [inconsistent](https://docs.python.org/3/library/shutil.html#shutil.copyfileobj) [function](https://docs.python.org/3/library/shutil.html#shutil.get_archive_formats) naming, I'm not going to insist

tests/utils/archives.py
ltalirz: > again an archive-path matter 😄

Correct me if I'm wrong but at least the API of the package should remain part of this review (probably the code should as well).

In any case, this particular suggestion is very minor and not important.

aiida/tools/importexport/dbexport/__init__.py
ltalirz: The main issue I have here is that this additional nesting is rather non-obvious  for a user who shouldn't have to care about the internal implementation - while the other parameter names are somewhat self-explanatory,`writer_init` is rather opaque.

If we could find a way to make things more obvious that would be very useful - `export` is *the* function people are going to use to create export files from scripts, so we should try to make it easy to use.

aiida/tools/importexport/common/archive.py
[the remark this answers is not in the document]
chrisjsewell: > I think the main question is whether we want safe_extract_zip to be part of the public API

I bit confused by this, `safe_extract_zip` is now removed from aiida-core so no it won't be part of the API?
Note it was added and removed before a release was made so no deprecation needed

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: >  a user who shouldn't have to care about the internal implementation

these are only to set fine-grained control of the writer though, above the defaults, which should be fine for most users.
Its more for "power" users that want to create their own writer classes.
I would use **kwargs, but that is already taken by the graph traversal rules lol

aiida/tools/importexport/common/archive.py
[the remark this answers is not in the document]
ltalirz: Ok, so I was right after all (sorry, perhaps I checked out the wrong branch locally to check) - well, then you should certainly adapt the warning

chrisjsewell:
archive-path has just been merged into conda-forge 🎉

aiida/tools/importexport/common/archive.py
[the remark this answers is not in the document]
chrisjsewell: >  well, then you should certainly adapt the warning

you mean change what it is now or keep as is?

aiida/tools/importexport/dbexport/__init__.py
chrisjsewell: removed suggestion to use writer init for use_compression (as discussed) in 98f0b9d

aiida/tools/importexport/common/archive.py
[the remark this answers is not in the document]
chrisjsewell: point to extract_tree in archive-path in 98f0b9d

chrisjsewell:
FYI, in case it was not clear, but the test coverage % is reduced because all the code for compression has now been moved to archive-path, so less lines are now tested

aiida/tools/importexport/dbexport/__init__.py
```diff
@@ -222,7 +208,19 @@ def export(
         name: traversal_rules.get(name, rule.default) for name, rule in GraphTraversalRules.EXPORT.value.items()
     }
 
-    writer = get_writer(file_format)(filename=filename, use_compression=use_compression, **(writer_init or {}))
+    # setup the archive writer
+    writer_init = writer_init or {}
+    if use_compression is not None:
+        warnings.warn(
+            'use_compression argument is deprecated and will be removed in AiiDA v2.0.0', AiidaDeprecationWarning
```
ltalirz, line 215:
```suggestion
            'use_compression argument is deprecated and will be removed in AiiDA v2.0.0 (which will always compress)', AiidaDeprecationWarning
```

ltalirz:
> FYI, in case it was not clear, but the test coverage % is reduced because I've actually removed more code than I've added (including all the code for compression being moved to archive-path), so less code lines are now tested

Ok! Just a last suggestion to accept and we can merge

chrisjsewell:
> Ok! Just a last suggestion to accept and we can merge

done cheers

ltalirz (dismissed):

ltalirz:
approved!

ltalirz (approved):

PR #4536, opened 2020-11-02
Fix hill notation for `get_formula()`

CasperWA:
C and H should be put first in the formula, and every other element should be sorted alphabetically.

It used to be so that it would only sort H if C was also present.

This PR fixes this as well as introduce a test to check both `'hill'` and `'hill_compact'` modes follow these rules correctly.

### DIFF
--- a/aiida/orm/nodes/data/structure.py
+++ b/aiida/orm/nodes/data/structure.py
@@ -503,11 +503,13 @@ def get_formula(symbol_list, mode='hill', separator=''):
     if mode in ['hill', 'hill_compact']:
         symbol_set = set(symbol_list)
         first_symbols = []
-        for special_symbol in ['C', 'H']:
-            # remove C and H if present from list and put them at the beginning
-            if special_symbol in symbol_set:
-                symbol_set.remove(special_symbol)
-                first_symbols.append(special_symbol)
+        # remove C (and possibly H) if present from list and put them at the beginning
+        if 'C' in symbol_set:
+            symbol_set.remove('C')
+            first_symbols.append('C')
+            if 'H' in symbol_set:
+                symbol_set.remove('H')
+                first_symbols.append('H')
         ordered_symbol_set = first_symbols + list(sorted(symbol_set))
         the_symbol_list = [[symbol_list.count(elem), elem] for elem in ordered_symbol_set]
 
--- a/aiida/orm/nodes/data/structure.py
+++ b/aiida/orm/nodes/data/structure.py
@@ -503,14 +503,11 @@ def get_formula(symbol_list, mode='hill', separator=''):
     if mode in ['hill', 'hill_compact']:
         symbol_set = set(symbol_list)
         first_symbols = []
-        if 'C' in symbol_set:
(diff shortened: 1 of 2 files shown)


ramirezfranciscof:
> It used to be so that it would only sort H if C was also present.

But...isn't this correct? I'm not an expert on this but [wikipedia](https://github.com/aiidateam/aiida-core/pull/4536) says H is ordered alphabetically when there is no carbon:

> The Hill system (or Hill notation) is a system of writing empirical chemical formulae, molecular chemical formulae and components of a condensed formula such that the number of carbon atoms in a molecule is indicated first, the number of hydrogen atoms next, and then the number of all other chemical elements subsequently, in alphabetical order of the chemical symbols. **When the formula contains no carbon, all the elements, including hydrogen, are listed alphabetically.** 

Is this not so? Where did you get that definition of hill notation?

(BTW: I still think adding a test for this would be valuable regardless of what conclusion we reach re-the definition)

CasperWA:
> > It used to be so that it would only sort H if C was also present.
> 
> But...isn't this correct? (...)

Looking around a bit more, this is indeed the case of what I can find - your wrongly linked wikipedia reference aside ;)

I'll update the PR accordingly.

ramirezfranciscof:
Ah, yes, the redundant linking to the same page, haha. Ok, ping me again when you update the tests!

aiida/orm/nodes/data/structure.py
```diff
@@ -503,14 +503,11 @@ def get_formula(symbol_list, mode='hill', separator=''):
...
-            first_symbols.append('C')
-            if 'H' in symbol_set:
-                symbol_set.remove('H')
-                first_symbols.append('H')
+        for special_symbol in ['C', 'H']:
+            # remove C and H if present from list and put them at the beginning
+            if special_symbol in symbol_set:
+                symbol_set.remove(special_symbol)
+                first_symbols.append(special_symbol)
         ordered_symbol_set = first_symbols + list(sorted(symbol_set))
```
csadorf: Assuming that the interpretation of the Hill notation is correct, I think this code could be simplified:
```suggestion
        ordered_symbol_set = list(sorted(set(symbol_list), key=lambda s: {'C': '0', 'H': '1'}.get(s, s)))
```

csadorf (commented):
I agree that we should add one or two tests for this.

aiida/orm/nodes/data/structure.py
CasperWA: It isn't, unfortunately.
Just pushed the update (reverting to previous implementation).

csadorf: Could still be massively simplified?
```python
if 'C' in symbol_list:
    ordered_symbol_set = list(sorted(set(symbol_list), key=lambda s: {'C': '0', 'H': '1'}.get(s, s))
else:
    ordered_symbol_set = list(sorted(set(symbol_list)))
```
Or am I missing something?

CasperWA: This is a really nice condensation, however!

csadorf (approved):

PR #4539, opened 2020-11-03
Make process functions submittable

unkcpz:
Feature #2965 to make ProcessFunction submittable, which avoid process functions(which may run for a long time) from interrupt to excepted state when the daemon is shutdown. 
However, this means function process(submittable one) should be persisted and can be recreate_from just like `CalcJob` and `WorkChain`, i.e. can be imported from the package. 
So I guess this kind of process function must be declared in the entry_points? Does this violate the basic design of the function process?

pinning you for the comments. @sphuber @muhrin @giovannipizzi


tests/engine/test_launch.py
```diff
@@ -141,6 +141,24 @@ def test_submit_store_provenance_false(self):
...
+        def local_calcfunction(x: orm.Int):
+            return x + 1
+
+        res = launch.run(local_calcfunction, x=orm.Int(1))
+        self.assertEqual(res.value, 2)
+
+        # The local calcfunction can not be create, since class can
+        # not be found when create persistence from local declare function
+        # inputs = {'x': orm.Int(2)}
+        # node = launch.submit(local_calcfunction, **inputs)
```
unkcpz, line 160: The `submit` can not apply to the locally announced calcfunction.

greschd:
I think this is great, thanks @unkcpz 

As you mention, there are restrictions on which process functions can be submitted. In general I don't see a problem with that, it's the same as for other processes. 
However, there are of course many process functions already "in the wild" that may not be submittable. I think it would make sense to give the end user (whoever runs or submits the process function) some way of distinguishing whether a given process function can be submitted. That could be e.g. a `utils.is_submittable(process_function)`, or a property of the process functions themselves.

Can we determine in code whether a process function is submittable, or is it something the developer needs to tell us?

For some context, in `aiida-optimize` I have the following code, that automatically selects whether to `run` or `submit` a sub-process [(link)](https://github.com/greschd/aiida-optimize/blob/ccd0889c6186d434f31efbb048237111a1a2290f/aiida_optimize/wrappers/_run_or_submit.py#L18):
```python
    def run_or_submit(self, proc: Process, **kwargs: ty.Any) -> orm.ProcessNode:
        if utils.is_process_function(proc):
            _, node = run_get_node(proc, **kwargs)
            return node
        return self.submit(proc, **kwargs)
```
I wouldn't know how to re-write that to take advantage of being able to submit _only some_ process functions.

sphuber:
> Can we determine in code whether a process function is submittable, or is it something the developer needs to tell us?

Really, for something to be "submittable", simply means that a daemon worker can properly execute the Python code, whether that be a class in the case of a `WorkChain` or `CalcJob`, or a function in the case of a process function. That means that if the resource can be imported, there should be no problem. The easiest is through entry points, but it can also be done through regular old imports. As long as the function is in the `PYTHONPATH` and the daemon has it correctly defined, it should be able to read it.

> For some context, in aiida-optimize I have the following code, that automatically selects whether to run or submit a sub-process

Note that once process functions are submittable, this will no longer be necessary. Then you can always just call `self.submit` to launch a sub-process. By calling the `submit` of the parent process, which is required (calling `aiida.engine.submit` will raise), it will use the `submit` method of the underlying `Runner`, which will automatically replicate the behavior with which the top process was launched. If the top process was launched with `run`, all subprocesses will also be `run` in the sense that they won't be sent to RabbitMQ but run in the same event loop of the same runner. Likewise, if the top process was submitted, each sub process launched through `self.submit` will also be submitted over RabbitMQ and be picked up by a (possibly different) daemon runner.

greschd:
Right, my concern is what happens with process functions that are _not_ top-level importable and executable - such as the locally-defined example made by @unkcpz in the tests.

I don't really know how common these are - but since that was _perfectly legal_ up until now, I would like my code to continue working with them. Ideally it would also take advantage of being able to `submit` process functions (when possible), so that's where I would like to distinguish them.

unkcpz:
> The easiest is through entry points, but it can also be done through regular old imports. As long as the function is in the PYTHONPATH and the daemon has it correctly defined, it should be able to read it.

@sphuber thanks for pointing it out. I just check and find there is are no hints for users when the process is not importable. Is that necessary to make the exceptional info more readable?

sphuber:
> I don't really know how common these are - but since that was _perfectly legal_ up until now, I would like my code to continue working with them. Ideally it would also take advantage of being able to `submit` process functions (when possible), so that's where I would like to distinguish them.

Existing code should not be affected, because currently they can only be run directly in the local interpreter. Since they are defined there, they are importable by definition. Or am I missing something?

tests/engine/test_launch.py
sphuber: This is to be expected though. The same holds exactly for all other processes, like `WorkChains`. If you define a `WorkChain` locally, either in a file that is not in the `PYTHONPATH` or let's say an interactive shell, when you submit it, it will fail, because the daemon cannot import the corresponding class.

unkcpz:
Then I will just add this feature to the docs. Do you have any suggestions of designs or changes I should add in code?

sphuber:
> thanks for pointing it out. I just check and find there is are no hints for users when the process is not importable. Is that necessary to make the exceptional info more readable?

It is really difficult to do this on the side that is submitting the function. The reason is that it may very well be importable in that code, but the interpreter of the daemon worker that will actually pick it up, may have a different environment where that function is *not* importable. It is practically impossible to dynamically inspect the environment of a Python interpreter on a different system process to check whether a resource is importable. The problem then is that the exception will appear in the daemon worker as soon as it picks up the process and attempts to start running it. Again, the same problem applies to `WorkChain`s and users have faced this often in the past, and we have put quite some effort into the exception handling code to make sure users understand the problem. The exception will end up in the daemon log of course.

tests/engine/test_launch.py
```diff
@@ -141,6 +141,24 @@ def test_submit_store_provenance_false(self):
         with self.assertRaises(exceptions.InvalidOperation):
             launch.submit(AddWorkChain, term_a=self.term_a, term_b=self.term_b, metadata={'store_provenance': False})
 
+    def test_submit_process_function(self):
+        """Test the `submit` of process function."""
+        inputs = {'term_a': orm.Int(2), 'term_b': orm.Int(4)}
+        node = launch.submit(add, **inputs)
+        self.assertIsInstance(node, orm.CalcFunctionNode)
```
sphuber, line 148: This merely tests that the `submit` method does not except and returns `CalcFunctionNode`, which is a good start, but we should also test that the process can actually be executed by a daemon runner. You can try to add this test here, but you will need to fake the daemon runner of course. You can try and take inspiration of the `tests/cmdline/commands/test_process.py` tests, or you can add submitting a process function to the `.ci/test_daemon.py` script. This will also run in the CI but with a normal daemon running, making it easier to test this kind of stuff.  You should test submitting a process function that is defined normally and is importable through the `PYTHONPATH` as well as through an entry point. For the latter you can use `aiida.workflows:arithmetic.add_multiply` which comes with `aiida-core` so you don't have to create one especially for the test.

aiida/engine/launch.py
```diff
@@ -96,7 +96,7 @@ def submit(process, **inputs):
     :return: the calculation node of the process
     :rtype: :class:`aiida.orm.ProcessNode`
     """
-    assert not is_process_function(process), 'Cannot submit a process function'
+    # assert not is_process_function(process), 'Cannot submit a process function'
```
sphuber, line 99:
```suggestion
```

sphuber (changes requested):
Thanks @unkcpz . I think the code is fine, straightforward enough. We should just add tests as I described in a code comment

greschd:
> Existing code should not be affected, because currently they can only be run directly in the local interpreter. Since they are defined there, they are importable by definition. Or am I missing something?

It needs to be defined in the interpreter that _launches it_, not necessarily the local one. That is, it can be defined inside a daemon worker; doesn't mean it is necessarily importable by a different daemon worker.

Totally un-advisable example (you can `submit(SomeWorkChain)` without issue):
```python
from aiida import orm
from aiida.engine import WorkChain, run_get_node, utils, workfunction


class RunOrSubmitWorkChain(WorkChain):
    """
    Adds a 'run_or_submit' method to the WorkChain class, which uses
    'run' for process functions and 'submit' else.
    """
    def run_or_submit(self, proc, **kwargs):
        if utils.is_process_function(proc):
            _, node = run_get_node(proc, **kwargs)
            return node
        return self.submit(proc, **kwargs)


class Sub1(WorkChain):
    @classmethod
    def define(cls, spec):
        super().define(spec)
        spec.outline(cls.do_stuff)

    def do_stuff(self):
        self.report('Sub1 doing stuff.')


class SomeWorkChain(RunOrSubmitWorkChain):
    @classmethod
    def define(cls, spec):
        super().define(spec)
        spec.outline(cls.do_stuff)

    def do_stuff(self):
        @workfunction
        def sub_2():
            # Getting 'self' from the 'SomeWorkChain' for added "fun".
            self.report("sub 2 doing stuff.")

        self.to_context(sub_1=self.run_or_submit(Sub1),
                        sub_2=self.run_or_submit(sub_2))
```

Just to note: If it isn't possible to construct a `utils.is_submittable` this PR is still an improvement (the `run_or_submit` code would continue to work), but it would be _nice_ being able to distinguish them.

sphuber:
I don't think it is possible to create the `utils.is_submittable` function as I described in a comment above. It simply is impossible to reliably determine whether a resource will be importable in a Python interpreter in another system process. Let alone one that might not even be running (you might want to submit a process function without the daemon even running). I don't see the problem of that though, because we have always had the exact same problem with all other process types. So nothing new on the horizon

greschd:
>  It simply is impossible to reliably determine whether a resource will be importable in a Python interpreter in another system process.

Yeah, agreed - whether the current interpreter can import it is a good indicator, but not a _reliable_ test. It would work in all the examples given here, but fail as soon as `sys.path` is different for the daemon. I think I'd probably be willing to put that into my plugin code, but it shouldn't go into `aiida-core` as it'll be a hack at best.

Alright now, I'll stop derailing this discussion with my obscure use cases 😅

tests/engine/test_launch.py
unkcpz: Thanks! I add tests in `.ci/test_daemon.py`. That's true these 'real' tests bring out a subtle issue, that `recreate_from` method should be given to the process function.

unkcpz:
@sphuber I add the tests and remove the 'process function can not be submitted' warning from the docs.
I have a weak impression that there is an instruction in documentation (or maybe in one of the tutorial) says WorkChains needed to be implemented in a package or importable in the PYTHONPATH. If so, the same feature of process function should also be mentioned there.

sphuber:
Yes, the exact same criterion applies to process functions as for `WorkChain` or `CalcJob` classes after this PR. The function needs to be importable, meaning that it either a) has an associated entry point or b) its module path is included in the `PYTHONPATH` that the daemon workers will have. Please also add this to the documentation on process functions, where it now says that they are submittable

greschd:
> a) has an associated entry point or b) its module path is included in the `PYTHONPATH` that the daemon workers will have

Even if it has an entry point, its module still needs to be on the `PYTHONPATH`, no?

sphuber:
> Even if it has an entry point, its module still needs to be on the `PYTHONPATH`, right?

I don't think so. As long as there is an entry point that is properly loaded in the `reentry` cache, then it is loadable by the daemon worker, even if that module is not in the `PYTHONPATH`. Not a 100% sure but we could easily test this.

greschd:
I think it needs to end up on `sys.path`, whether through a "regular" install or via `PYTHONPATH` environment variable; Python cannot import modules that aren't on `sys.path`.

Whether there is an entry point is a somewhat orthogonal question: If there is one, the `aiida.plugins` loaders can be used with the entry point names.

(I did a tiny bit of testing, but feel free to double-check this)

.ci/test_daemon.py
```diff
@@ -72,6 +73,30 @@ def print_report(pk):
...
+        calc = load_node(pk)
+        if not calc.is_finished_ok:
+            print(f'Calc<{pk}> not finished ok: process_state<{calc.process_state}> exit_status<{calc.exit_status}>')
+            print_report(pk)
+            valid = False
+
+        try:
+            actual_result = calc.outputs.result
+        except exceptions.NotExistent:
+            print(f'Could not retrieve `output_parameters` node for Calculation<{pk}>')
```
sphuber, line 89:
```suggestion
            print(f'Could not retrieve `result` output for process<{pk}>')
```

.ci/test_daemon.py
```diff
@@ -356,6 +410,20 @@ def main():
...
+        inputval = counter
+        proc, expected_result = launch_calcfunction(counter=counter, inputval=inputval)
+        expected_results_calcfunctions[proc.pk] = expected_result
+
+    # Submitting the workfunction through the launchers
+    print(f'Submitting {NUMBER_CALCULATIONS} workfunction to the daemon')
+    for counter in range(1, NUMBER_CALCULATIONS + 1):
+        inputval = counter
+        proc, expected_result = launch_workfunction(counter=counter, inputval=inputval)
+        expected_results_workfunctions[proc.pk] = expected_result
```
sphuber, line 425: I think launching just one of each here is fine. Don't think running multiple process functions tests anything of additional value

docs/source/topics/processes/concepts.rst
```diff
@@ -142,6 +142,7 @@ When a process is 'submitted', an instance of the ``Process`` is created, along
 This is called a 'process checkpoint', more information on which :ref:`will follow later<topics:processes:concepts:checkpoints>`.
 Subsequently, the process instance is shut down and a 'continuation task' is sent to the process queue of RabbitMQ.
 This task is simply a small message that just contains an identifier for the process.
+In order to reconstruct the process from 'process checkpoint', the process needs to be importable in the daemon environment by a) giving it an associated entry point or b) including its module path in the ``PYTHONPATH`` that the daemon workers will have.
```
sphuber, line 145:
```suggestion
In order to reconstruct the process from a `checkpoint`, the process needs to be importable in the daemon environment by a) giving it an :ref:`associated entry point<how-to:plugin-codes:entry-points>` or b) :ref:`including its module path<how-to:faq:process-not-importable-daemon>` in the ``PYTHONPATH`` that the daemon workers will have.
```
You will have to add the line
```
.. _how-to:faq:process-not-importable-daemon:`
```
to the file `docs/source/howto/faq.rst` on line 65 for the second link

docs/source/topics/processes/usage.rst
```diff
@@ -330,7 +330,7 @@ For example, when we want to run an instance of the :py:class:`~aiida.calculatio
 The function will submit the calculation to the daemon and immediately return control to the interpreter, returning the node that is used to represent the process in the provenance graph.
 
 .. warning::
-    Process functions, i.e. python functions decorated with the ``calcfunction`` or ``workfunction`` decorators, **cannot be submitted** but can only be run.
+    To submit process, the process needs to be importable, meaning that it either a) has an associated entry point or b) its module path is included in the ``PYTHONPATH`` that the daemon workers will have.
```
sphuber, line 333:
```suggestion
    For a process to be submittable, the class or function needs to be importable in the daemon environment by a) giving it an :ref:`associated entry point<how-to:plugin-codes:entry-points>` or b) :ref:`including its module path<how-to:faq:process-not-importable-daemon>` in the ``PYTHONPATH`` that the daemon workers will have.
```

tests/engine/test_launch.py
```diff
@@ -141,6 +141,24 @@ def test_submit_store_provenance_false(self):
         with self.assertRaises(exceptions.InvalidOperation):
             launch.submit(AddWorkChain, term_a=self.term_a, term_b=self.term_b, metadata={'store_provenance': False})
 
+    def test_submit_process_function(self):
```
sphuber, line 144: Since this was supposed to test submitting a process function, but it not actually does (and this is done isntead in `test_daemon.py`, this entire test can be removed.

tests/engine/test_process_function.py
```diff
@@ -350,8 +350,10 @@ def test_launchers(self):
         self.assertEqual(result, get_true_node())
         self.assertTrue(isinstance(node, orm.CalcFunctionNode))
 
-        with self.assertRaises(AssertionError):
-            submit(self.function_return_true)
+        # process function can be submitted but the process should be importable
+        # in python interpreter
```
sphuber, line 354:
```suggestion
        # Process function can be submitted and will be run by a daemon worker as long as the function is importable
        # Note that the actual running is not tested here but is done so in `.ci/test_daemon.py`.
```

tests/engine/test_process_function.py
```diff
@@ -350,8 +350,10 @@ def test_launchers(self):
...
+        from aiida.workflows.arithmetic.add_multiply import add_multiply
```
sphuber, line 355: Move import to top of file

tests/engine/test_process_function.py
```diff
@@ -350,8 +350,10 @@ def test_launchers(self):
...
+        submit(add_multiply, x=orm.Int(1), y=orm.Int(2), z=orm.Int(3))
```
sphuber, line 356:
```suggestion
        node = submit(add_multiply, x=orm.Int(1), y=orm.Int(2), z=orm.Int(3))
        assert isinstance(node, WorkFunctionNode)
```

sphuber (changes requested):
Thanks @unkcpz a few minor changes requested

docs/source/howto/faq.rst
```diff
@@ -77,3 +77,5 @@ Run the command ``verdi daemon logshow`` in a separate terminal to see the loggi
 If the root cause is indeed due to an import problem, it will probably appear as an ``ImportError`` exception in the daemon log.
 To solve these issues, make sure that all the Python code that is being run is properly importable, which means that it is part of the `PYTHONPATH <https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH>`_.
 Make sure that the PYTHONPATH is correctly defined automatically when starting your shell, so for example if you are using bash, add it to your ``.bashrc``.
+
+.. _how-to:faq:process-not-importable-daemon:
```
sphuber, line 81: This anchor should be on line 65, just before the header "Why would...."

sphuber (changes requested):
I am afraid I still need one more change, you put the anchor for the docs on the wrong line. Once you correct that I will merge this.

docs/source/howto/faq.rst
unkcpz: Thanks, fixed. No much experience about `rst` ;-p

sphuber (approved):

sphuber:
Thanks a lot @unkcpz !

PR #4542, opened 2020-11-04
RabbitMQ: remove validation of `broker_parameters` from profile

### LINKED ISSUE #4540: Remove validation of `broker_parameters` from profile when constructing RabbitMQ URL
This validation was added as an attempt to help users with detecting invalid parameters in the `broker_parameters` dictionary of a profile, but `aiida-core` internally has no real limitations here. It is the libraries underneath that decide what is acceptable and this can differ from library to library plus it is not always clear. For example, currently we use `topika` and `pika` which behave different from `aiormq` which will be replacing them soon once `tornado` is replaced with `asyncio`. It is best to not limit the options on `aiida-core`'s side and just let it fail downstream as to not artificially limit the parameters that might be perfectly acceptable by the libraries downstream.

sphuber:
Fixes #4540 

This validation was added as an attempt to help users with detecting
invalid parameters in the `broker_parameters` dictionary of a profile,
but `aiida-core` internally has no real limitations here. It is the
libraries underneath that decide what is acceptable and this can differ
from library to library plus it is not always clear. For example,
currently we use `topika` and `pika` which behave different from
`aiormq` which will be replacing them soon once `tornado` is replaced
with `asyncio`. It is best to not limit the options on `aiida-core`'s
side and just let it fail downstream as to not artificially limit the
parameters that might be perfectly acceptable by the libraries
downstream.

### DIFF
--- a/aiida/manage/external/rmq.py
+++ b/aiida/manage/external/rmq.py
@@ -40,16 +40,6 @@ BROKER_DEFAULTS = AttributeDict({
     'heartbeat': 600,
 })
 
-BROKER_VALID_PARAMETERS = [
-    'heartbeat',  # heartbeat timeout in seconds
-    'cafile',  # string containing path to ca certificate file
-    'capath',  # string containing path to ca certificates
-    'cadata',  # base64 encoded ca certificate data
-    'keyfile',  # string containing path to key file
-    'certfile',  # string containing path to certificate file
-    'no_verify_ssl',  # boolean disables certificates validation
-]
-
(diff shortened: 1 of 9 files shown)


sphuber:
@muhrin @ramirezfranciscof I had the contact person at LLNL test this branch and they confirmed it still works with their environment, so this should be good to merge. Then I will prepare the 1.4.3 release.

muhrin (approved):
Looks good

Issue #4451, opened 2020-10-16
Docs: remove old content from 'developer_guide/'


ramirezfranciscof:
This is a structure that was left over from the old documentation and needs to be removed, but first we need to check if all the content has been re-assigned in the new documentation. The files considered are:

- [ ] [design/changes.rst](https://aiida-core.readthedocs.io/en/v1.1.0/developer_guide/design/changes.html)
- [ ] [core/modifying_the_schema.rst](https://aiida-core.readthedocs.io/en/v1.1.0/developer_guide/core/modifying_the_schema.html)
- [ ] [core/extend_restapi.rst](https://aiida-core.readthedocs.io/en/v1.1.0/developer_guide/core/extend_restapi.html)
- [ ] core/transport_template.py (downloadable file in core/transport.rst)
- [ ] [core/transport.rst](https://aiida-core.readthedocs.io/en/v1.1.0/developer_guide/core/transport.html)
- [ ] [core/internals.rst](https://aiida-core.readthedocs.io/en/v1.1.0/developer_guide/core/internals.html)
- [ ] [core/caching.rst](https://aiida-core.readthedocs.io/en/v1.1.0/developer_guide/core/caching.html)
- [x] plugins.rst (only index, can be deleted directly)
- [x] core.rst (only index, can be deleted directly)

### Proposition:

This content seems to be appropriate for the internals section (the only thing I am not sure of is the contents assigned to the engine section:

- Global design: `design/changes.rst`
- RestAPI: `core/extend_restapi.rst`
- DB schema: `core/modifying_the_schema.rst`
- Entities: `core/internals.rst`
- Engine: `core/caching.rst`? `core/transport.rst`? (and consequently `core/transport_template.py`)

Since the internals are not a priority right now, I wonder if we should just copy this pages into these new placeholders and leave it there until those sections are tackled so that we can remove the old structure from the repository. Alternatively we could put the files inside of `internals/global_design.src` folders if we don't want to display unfinished content in the documentation.

sphuber:
* `core/caching.rst`: The first section is "Controlling hashing" needs to go to `howto/run_codes` and be appended at the end in the caching section, since this is user facing information, not for core developers. The "Design guidelines" should be merged into this new section.
* `core/caching.rst`: The "Controlling caching" section should go to a section in the `internals/engine.rst`

giovannipizzi:
[...]

PR #4546, opened 2020-11-06
Docs: add content from old documentation on caching/hashing

### LINKED ISSUE #4472: Docs: Move docs on caching
From #4451 

Move the content of "Controlling hashing" and "Design guidelines" inside of `core/caching.rst` to `topics/provenance/caching` (see the [old documentation](https://aiida-core.readthedocs.io/en/v1.1.0/developer_guide/core/caching.html)).

ramirezfranciscof:
close #4472 (now for real)

Moved the content of "Controlling hashing" and "Design guidelines" inside of `developer_guide/core/caching.rst` to `topics/provenance/caching`.


ramirezfranciscof:
@ltalirz I'm tagging you since we discussed this earlier so I think you have this fresh and it would be quick for you to review it, but let me know if you prefer for me to find someone else.

ramirezfranciscof:
@ltalirz @sphuber this should be a fairly short change, mostly moved content around with minor adaptations. Could any of you take a look at it?

docs/source/topics/provenance/caching.rst
```diff
@@ -53,10 +57,26 @@ In order to figure out why a calculation is *not* being reused, the :meth:`~aiid
...
+
+* To ignore specific attributes, a :py:class:`~aiida.orm.nodes.Node` subclass can have a ``_hash_ignored_attributes`` attribute.
+  This is a list of attribute names, which are ignored when creating the hash.
+* For calculations, the ``_hash_ignored_inputs`` attribute lists inputs that should be ignored when creating the hash.
+* To add things which should be considered in the hash, you can override the :meth:`~aiida.orm.nodes.Node._get_objects_to_hash` method.
+  Note that doing so overrides the behavior described above, so you should make sure to use the ``super()`` method.
+* Pass a keyword argument to :meth:`~aiida.orm.nodes.Node.get_hash`.
+  These are passed on to :meth:`~aiida.common.hashing.make_hash`.
+
+
```
sphuber, line 75: This section should be clarified. Part of this is aimed at `aiida-core` developers and some of it can also be directed to plugin developers. The behavior that is described can only be adapted outside of `aiida-core` for `Data` subclasses, because process nodes cannot be subclassed. So I think this content should go in some form to the Internal Architecture section, but it should also stay here, in rewritten form, where it makes clear that this can only be done for `Data` plugins. With that given, it should also give more clear examples making it more practically targeted at plugin developers.

sphuber (changes requested):
Thanks @ramirezfranciscof . If I understand correctly, you are moving in two sections essentially verbatim and adding a "todo" paragraph in the internal architecture. Besides not moving in all existing documentation, why not actually write the documentation instead of writing a small todo?

docs/source/topics/provenance/caching.rst
greschd: In the part for plugin developers, there should also be a mention of the `invalidates_cache` keyword argument on exit codes.

docs/source/internals/engine.rst
```diff
@@ -6,6 +6,15 @@
...
     ******
 
+    .. _internal_architecture:engine:caching:
+
+    Controlling caching
+    -------------------
+
+    Old Docs: This should also contain the contents of "Controlling caching" inside of developer_guide/core/caching.rst.
+    You should include links to the other sections topics:provenance:caching: and how-to:run-codes:caching:, and in those sections as well.
+    Example for topics:provenance:caching: If you want more details of the internal architecture instead, you can check out the :ref:`relevant subsection <internal_architecture:engine:caching>` in the engine documentation.
```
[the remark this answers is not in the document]
ramirezfranciscof: The engine part was being done by @chrisjsewell and a reminder to do this was already added to that issue. I added this here mostly because since we already had the "duplicated" Engine task (meaning, both the issue and the todo in the files), we might as well have all the info in both places. If I copy it myself it might conflict with the other content of the section and be annoying to resolve, not to mention that it might already have been included in that work.

docs/source/developer_guide/core/caching.rst
```diff
@@ -4,20 +4,6 @@ Caching: implementation details
...
-* To ignore specific attributes, a :py:class:`~aiida.orm.nodes.Node` subclass can have a ``_hash_ignored_attributes`` attribute.
-  This is a list of attribute names, which are ignored when creating the hash.
-* For calculations, the ``_hash_ignored_inputs`` attribute lists inputs that should be ignored when creating the hash.
-* To add things which should be considered in the hash, you can override the :meth:`~aiida.orm.nodes.Node._get_objects_to_hash` method. Note that doing so overrides the behavior described above, so you should make sure to use the ``super()`` method.
-* Pass a keyword argument to :meth:`~aiida.orm.nodes.Node.get_hash`.
-  These are passed on to :meth:`~aiida.common.hashing.make_hash`.
-
 .. _devel_controlling_caching:
 
 Controlling caching
```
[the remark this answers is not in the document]
ramirezfranciscof: [I understood](https://github.com/aiidateam/aiida-core/issues/4451#issuecomment-711792600) that the whole "Controlling Hashing" sections was supposed to go to the internal architecture page, not to the topics. Is it only the "Workflownode example" that has to go there then? If the main part of "Controlling Hashing" has user-facing relevant information, then the example also looks to be dealing with the same kind of stuff.

docs/source/internals/engine.rst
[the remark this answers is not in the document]
sphuber: If he is writing and improving the caching documentation, wouldn't it then make sense to have him also move/incorporate these parts, so he can make sure the whole thing is coherent?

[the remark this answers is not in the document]
ramirezfranciscof: Not the caching in particular, he is writing essentially from scratch the docs for the internals of the engine, which were lower priority than trying to get all of the old documentation back somewhere in the new and hence he didn't need to finish this for the release. What we did want was to try to get back as much of the old docs as possible, and since the "topics" section was already present, moving all of the "user-facing" content of caching there seemed like a possibility. If you think is better to wait to emphasize on the consistency over having available these sections on caching/hashing now for the release, then I guess I can block this until the engine section is finished.

docs/source/topics/provenance/caching.rst
ramirezfranciscof: > The behavior that is described can only be adapted outside of aiida-core for Data subclasses, because process nodes cannot be subclassed.

@sphuber These are all for `Data` subclasses? Then how do you use the `_hash_ignored_inputs`? Isn't there a way to control the hashing of processes via the definition of the calcjob plugin?

> In the part for plugin developers, there should also be a mention of the `invalidates_cache` keyword argument on exit codes.

@greschd Could you explain to me how does this work? I am not familiar with it.

sphuber: > sn't there a way to control the hashing of processes via the definition of the calcjob plugin?

There are some ways to control hashing of processes, for example the `invalidates_cache` of the `ExitCode`, but anything that is defined on the `Node` class can not be changed because node classes can not be subclassed and controlled by plugins

greschd: >  Could you explain to me how does this work? I am not familiar with it.

The `spec.exit_code` function takes an `invalidates_cache` keyword argument. If `True`, a calculation that has this exit code will not be used as a cache source.
Plugin developers can use this to e.g. disable caching for calculations that failed with an unknown error (that could well be node failure or similar). Simple example from `aiida-wannier90`: https://github.com/aiidateam/aiida-wannier90/pull/104

docs/source/internals/engine.rst
```diff
@@ -1,11 +1,49 @@
...
+On the level of the generic :class:`orm.Node <aiida.orm.nodes.Node>` class:
+
+* The :meth:`~aiida.orm.nodes.Node.is_valid_cache` property determines whether a particular node can be used as a cache.
+  This is used for example to disable caching from failed calculations.
+* Node classes have a ``_cachable`` attribute, which can be set to ``False`` to completely switch off caching for nodes of that class.
+  This avoids performing queries for the hash altogether.
+
+On the level of the :class:`Process <aiida.engine.processes.process.Process>` and :class:`orm.ProcessNode <aiida.orm.nodes.process.ProcessNode>` classes:
+
+* The :meth:`ProcessNode.is_valid_cache <aiida.orm.nodes.process.ProcessNode.is_valid_cache>` calls :meth:`Process.is_valid_cache <aiida.engine.processes.process.Process.is_valid_cache>`, passing the node itself.
```
greschd, line 30: Since this is a way for plugin developers to override the caching behavior, shouldn't this be mentioned in the section for plugin developers?

docs/source/topics/provenance/caching.rst
```diff
@@ -53,10 +58,34 @@ In order to figure out why a calculation is *not* being reused, the :meth:`~aiid
...
+  Note that doing so also overrides the behavior described above, so make sure to use the ``super()`` method in order to prevent this.
+* You can pass a keyword argument to :meth:`~aiida.orm.nodes.Node.get_hash`.
+  These are, in turn, passed on to :meth:`~aiida.common.hashing.make_hash`.
+
+The process nodes have a fixed behavior that is internal to AiiDA and are not subclassable, so they can't be customized in a direct way.
+To know more about the specifics of these internals you can visit the :ref:`corresponding section <internal_architecture:engine:caching>`.
+There are only two ways in which these can be influenced by plugin designers:
+
+* Indirectly via the hash criteria for the associated data types of their inputs.
+* The :meth:` spec.exit_code <aiida.engine.processes.process_spec.ProcessSpec.exit_code>` has a keyword argument ``invalidates_cache``.
```
greschd, line 81: This isn't controlling _hashing_, it's controlling _caching_. Since the title doesn't apply, maybe it should have a separate short section, together with the `is_valid_cache` mechanism?

greschd (commented):
Thanks @ramirezfranciscof, looks good overall.

Since I don't have a top-level view of the documentation structure I'll leave the main review up to @sphuber.

docs/source/internals/engine.rst
ramirezfranciscof: I interpret that you mean it in the sense that "...shouldn't this ALSO be mentioned...", right? Since here it is explaining the internal mechanisms by which the feature is implemented (in the same way the 3rd point below this explains where the `exit_codes` with `invalidate_cache` are processed). If that is the case, yes, I agree there should be also a more "application oriented" reference to this in the topics.

greschd: Yep, _also_ mentioned in the topics.

ramirezfranciscof:
Thanks for the comments @greschd ! I have now incorporated them into the PR, let me know what you think.

greschd:
Yep, all good from my POV. Still think someone else should give it a pass though, because I'm not familiar with the design of the docs.

ramirezfranciscof:
Maybe @sphuber can give a last pass or suggest some other person to do this?

sphuber (approved):
Thanks @ramirezfranciscof . Let's merge this and then further improvements can be added later on

PR #4548, opened 2020-11-06
docs: include ipython magics from source code

### LINKED ISSUE #4547: %aiida ipython magic broken?
### Describe the bug

It seems the [instructions for enabling the `%aiida` ipython magic](https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/installation.html#using-aiida-in-jupyter) are broken.

### Steps to reproduce

1. `pip install aiida-core[notebook]`
1. Copy the snippet from the documentation into `~/.ipython/profile_default/startup/aiida_magics.py` 
1. Start `ipython` shell
1. Run `%aiida`

```
In [1]: %aiida
UsageError: Line magic function `%aiida` not found.
```

### Expected behavior

The magics are enabled.

Furthermore, it seems the snippet in the docs is outdated wrt to the one in the codebase (but neither work).

The ipython magics is already listed as one of the few remaining offenders of completely untested code: https://github.com/aiidateam/aiida-core/issues/2835

### Your environment

- Python version [e.g. 3.7.1]: 3.7
- aiida-core version [e.g. 1.2.1]: 1.4.2

ltalirz:
fixes #4547 

 * docs: Remove duplication of snippet to register the AiiDA ipython magics from the
aiida-core codebase
 * docs: Also revisit the corresponding section of the documentation, starting with the setup, and removing some generic information about jupyter.
 * test AiiDA ipython magic
 * move some additional lines from the registration snippet to
aiida-core (where we can adapt it if the IPython API ever changes)
 * rename misnomer `load_ipython_extension` to
   `register_ipython_extension` with deprecation far in the future

### DIFF
--- a/.ci/test_ipython_magics.py
+++ b/.ci/test_ipython_magics.py
@@ -21,6 +21,8 @@ def test_ipython_magics():
 %aiida
 qb=QueryBuilder()
 qb.append(Node)
+qb.all()
+Dict().store()
 """
     result = ipy.run_cell(cell)
 
--- a/.ci/test_ipython_magics.py
+++ b/.ci/test_ipython_magics.py
@@ -7,23 +7,21 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
+"""Test the AiiDA iPython magics."""
+from IPython.testing.globalipapp import get_ipython
+from aiida.tools.ipython.ipython_magics import register_ipython_extension
+
+
+def test_ipython_magics():
+    """Test that the %aiida magic can be loaded and adds the QueryBuilder and Node variables."""
+    ipy = get_ipython()
+    register_ipython_extension(ipy)
+
+    cell = """
+%aiida
+qb=QueryBuilder()
+qb.append(Node)
 """
-File to be executed by IPython in order to register the line magic %aiida
+    result = ipy.run_cell(cell)
 
-This file can be put into the startup folder in order to have the line
-magic available at startup.
-The start up folder is usually at ``.ipython/profile_default/startup/``
-"""
-
-# DOCUMENTATION MARKER
-if __name__ == '__main__':
-
-    try:
-        import aiida
-        del aiida
-    except ImportError:
-        # AiiDA is not installed in this Python environment
-        pass
-    else:
-        from aiida.tools.ipython.ipython_magics import register_ipython_extension
-        register_ipython_extension()
+    assert result.success
--- a/.github/workflows/tests.sh
+++ b/.github/workflows/tests.sh
@@ -24,6 +24,7 @@ verdi daemon stop
 
 # tests for the testing infrastructure
 pytest --noconftest .ci/test_test_manager.py
+pytest --noconftest .ci/test_ipython_magics.py
 pytest --noconftest .ci/test_profile_manager.py
 python .ci/test_plugin_testcase.py  # uses custom unittest test runner
 
--- a/aiida/tools/ipython/aiida_magic_register.py
+++ b/aiida/tools/ipython/aiida_magic_register.py
@@ -15,6 +15,7 @@ magic available at startup.
 The start up folder is usually at ``.ipython/profile_default/startup/``
 """
 
+# DOCUMENTATION MARKER
 if __name__ == '__main__':
 
     try:
--- a/aiida/tools/ipython/ipython_magics.py
+++ b/aiida/tools/ipython/ipython_magics.py
@@ -34,10 +34,8 @@ Usage
    In [2]: %aiida
 """
 
-from IPython import version_info  # pylint: disable=no-name-in-module
-from IPython.core import magic  # pylint: disable=no-name-in-module,import-error
-
-from aiida.common import json
+from IPython import version_info, get_ipython
+from IPython.core import magic
 
 
 def add_to_ns(local_ns, name, obj):
@@ -99,6 +97,8 @@ class AiiDALoaderMagics(magic.Magics):
         """
         Output in JSON format.
         """
+        from aiida.common import json
+
         obj = {'current_state': self.current_state}
         if version_info[0] >= 3:
             return obj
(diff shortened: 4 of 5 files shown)


.ci/test_ipython_magics.py
```diff
@@ -0,0 +1,27 @@
...
+
+def test_ipython_magics():
+    """Test that the %aiida magic can be loaded and adds the QueryBuilder and Node variables."""
+    ipy = get_ipython()
+    register_ipython_extension(ipy)
+
+    cell = """
+%aiida
+qb=QueryBuilder()
+qb.append(Node)
```
chrisjsewell, line 23: should you not actually run the query here, to check that you can connect to the aiida database?

```suggestion
qb.append(Node)
qb.all()
```

or e.g.

```python
orm.Dict().store()
```

chrisjsewell (changes requested):
just a minor nitpick

chrisjsewell:
Actually also, why can't .ci/test_ipython_magics.py just be in the normal `tests` folder and use the `aiida_profile` fixture?

.ci/test_ipython_magics.py
ltalirz: Sounds good!

ltalirz:
> Actually also, why can't .ci/test_ipython_magics.py just be in the normal `tests` folder and use the `aiida_profile` fixture?

The test might actually run fine, but I'm not sure whether there will be interference between the dbenv loaded by the test fixtures and the dbenv loading by `%aiida`.
If you can exclude that there is an interaction, it is ok to move - otherwise I prefer to test this separately to make sure that the magic does its work also when run independently.

ltalirz:
@chrisjsewell You see that we again have this pre-commit failure that I'm working around in https://github.com/aiidateam/aiida-core/pull/4550

https://github.com/aiidateam/aiida-core/pull/4548/checks?check_run_id=1384616711#step:6:45

Since this only occurs in this PR, I can only assume it has something to do with me removing the comments to disable pylint rules on two specific imports here (?)
https://github.com/aiidateam/aiida-core/pull/4548/files#diff-3ca2eb4751bee66e85c985bb1c509e482c87bde017fc6238df2f18fd19a90eebL37

I'll check whether that is the case (even though it's totally unclear to me what gives rise to this behavior)

chrisjsewell:
> I'll check whether that is the case (even though it's totally unclear to me what gives rise to this behavior)

Surely you would need to add `.ci/` to the `PYTHONPATH`, otherwise it will not know where to find `lib`

ltalirz:
> Surely you would need to add `.ci/` to the `PYTHONPATH`, otherwise it will not know where to find `lib`

I guess you mean `.ci/polish`?

Can you tell me which of my edits you believe is triggering this issue? As far as I'm aware I'm not doing anything that has something to do with this bit of code (?)

chrisjsewell:
you could do what is already done in `.github/workflows/tests.sh`: `export PYTHONPATH="${PYTHONPATH}:${GITHUB_WORKSPACE}/.ci"`.

Something like:

```yaml
    - name: Run pre-commit
      run:
        pre-commit run --all-files || ( git status --short ; git diff ; exit 1 )
      env:
        PYTHONPATH: "${PYTHONPATH}:${GITHUB_WORKSPACE}/.ci"
```

ltalirz:
Ok

chrisjsewell:
> The test might actually run fine, but I'm not sure whether there will be interference between the dbenv loaded by the test fixtures and the dbenv loading by %aiida.

ok fair, I'm not going to insist, but really I'd rather get rid of the mess that is the .ci folder lol.

In fact, perhaps it would be possible to replace it with molecule testing. Then it would mean that you could actually run all these tests locally, rather than just via Jenkins/GH actions

ltalirz:
> you could do what is already done in 

Unfortunately, this didn't solve the issue.
I will now apply the fix done in https://github.com/aiidateam/aiida-core/pull/4550

ltalirz:
> In fact, perhaps it would be possible to replace it with molecule testing. Then it would mean that you could actually run all these tests locally, rather than just via Jenkins/GH actions

That sounds like a good idea.
Provisioner could be ansible but doesn't have to be.
The dependencies for this should go into a separate extra (`docker-tests` or similar)

chrisjsewell:
> should I open an issue for this

done https://github.com/aiidateam/aiida-core/issues/4557 😄

chrisjsewell:
> @chrisjsewell tests pass

yeh but now you have broken jenkins 😆

ltalirz:
Ah, I thought that was somehow always broken.
Is it supposed to run through?

chrisjsewell:
> Is it supposed to run through?

yeh, it failing on your change from `lib` to `.lib`

sphuber:
> In fact, perhaps it would be possible to replace it with molecule testing. Then it would mean that you could actually run all these tests locally, rather than just via Jenkins/GH actions

Note that this is not the reason we run these tests outside of the unit tests though. The tricky part with some of these tests is that they require a running daemon, which is not always trivial to actually mock in a unit tests. And some tests actually require the loading of the db environment, but that will already have been loaded by the pytest fixtures and so there can be unintentional and problematic interaction between, as @ltalirz already mentioned. I am all for cleaning up the `.ci` folder, but need to sit down and discuss a bit before we start hacking away

sphuber:
And Jenkins is failing because you are using relative imports in a Python script, which is not supported.

chrisjsewell:
> The tricky part with some of these tests is that they require a running daemon, which is not always trivial to actually mock in a unit tests. 

Yeh this is pretty trivial to do with molecule. As mentioned in the issue I opened it creates a temporary docker container and runs the tests in there, which is exactly what you are already doing on Jenkins and GH actions

ltalirz:
@chrisjsewell Tests pass now, could you please re-review?

ltalirz:
Ah hm, I notice one of the  django tests timed out... not sure it's worth paying attention to this; I also can't really tell what is the reason (hanging test or just slow setup, ...)

chrisjsewell:
> Ah hm, I notice one of the django tests timed out... not sure it's worth paying attention to this; I also can't really tell what is the reason (hanging test or just slow setup, ...). The same test passes with sql and also when switching python version

no I think its just because we have so many test jobs now, some intermittently hang

chrisjsewell (approved):

PR #4553, opened 2020-11-10
Improve mypy type checking

chrisjsewell:
Currently, the mypy in the pre-commit is used as a "non-local" import and adds the blanket `--ignore-missing-imports` flag.
This greatly reduces the effectiveness of the type checking, because it does not check any types from classes/functions imported from third-party packages.
For example in #4534 I add the beautifully typed https://github.com/aiidateam/archive-path (note the `py.typed` file in the module signalling it is typed), but the code in aiida-core is not checked against these types at present.

Similarly, adding `check_untyped_defs = True` fixes this issue: https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code.

This PR is blocked until #4534 is merged (which should fix the current errors).

Note that if we did want to type check against e.g. django there is available stub packages: https://github.com/typeddjango/django-stubs


greschd (approved):
Looks good, thanks @chrisjsewell. I wasn't aware the pre-commit/mirrors-mypy adds a blanket `--ignore-missing-imports`, we definitely want to get rid of that.

> This PR is blocked until #4534 is merged (which should fix the current errors).

Why exactly is this blocked?

greschd:
By the way, should we also add a `py.typed` file, or does that only apply once a package is _fully_ typed?

chrisjsewell:
> By the way, should we also add a `py.typed` file, or does that only apply once a package is _fully_ typed?

See https://www.python.org/dev/peps/pep-0561 for guidance. It seems maybe a bit premature to add one yet, given the sparsity of files checked just in .pre-commit-config.yaml (I'll pushing people to add more lol)

chrisjsewell:
>  this doesn't necessarily have to block this PR

well I'm hoping to merge very soon so might as well wait

greschd:
> See https://www.python.org/dev/peps/pep-0561 for guidance. It seems maybe a bit premature to add one yet, given the sparsity of files checked just in .pre-commit-config.yaml (I'll pushing people to add more lol)

Hmm, it's not 100% clear to me, but it seems that the annotations should be at least somewhat complete when adding `py.typed`:

> Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing. This marker applies recursively: if a top-level package includes it, all its sub-packages MUST support type checking as well.

I guess that makes sense, because otherwise the calling code will get a bunch of errors if `mypy` can't automatically deduce the types. 
It seems we could add `py.typed` only to a sub-package, but that would need checking.

chrisjsewell:
ok this should be good to go now

chrisjsewell:
> But technically we're not a "stub package" I think - so does this still apply?

no I don't believe this applies

> It seems we could add py.typed only to a sub-package, but that would need checking.

yeh we can consider this later on, because there is not a whole lot of code that is type checked right now

.pre-commit-config.yaml
```diff
@@ -38,7 +38,7 @@ repos:
         entry: mypy
         language: python
         types: [python]
-        args: [--scripts-are-modules]
+        args: [--scripts-are-modules, --warn-unused-ignores]
```
greschd, line 41: Could these also be set in the `mypy.ini` file instead? I think it'd make sense to keep all configuration in one place.

greschd (approved):

.pre-commit-config.yaml
chrisjsewell: yep good point, done

greschd (approved):

greschd:
Good to go, thanks @chrisjsewell!

Issue #4543, opened 2020-11-05
Why `get_function_source_code` return source file of the function not the source function itself?


unkcpz:
https://github.com/aiidateam/aiida-core/blob/06063e7c38b6e40d3d2fbafe84e4e7f6adc8782a/aiida/orm/utils/mixins.py#L108-L113

Would this more make sense if the source code of function rather than the source of the file is returned? 
Besides, if I understand it correctly, every process function will be stored along with a source file in the repository. Is that too redundant to store the same files every time when the `ProcessNode` call and store?

sphuber:
Not sure, because the rest of the file can also be important for the process function, for example other functions that may be defined or imports. Of course we can add a method that gives just the source code of the function itself, and potentially rename the current one, that would be possible I would say.

I just noticed by the way that the docstring is incorrect, the `:returns:` is incorrect as it says the filepath will be returned.

With respect to the duplication of the source code: this is indeed very inefficient, but fortunately with the new repository implementation that will be merged for `aiida-core==2.0.0`, that will be automatically fixed. That repo implementation will automatically deduplicate content.

unkcpz:
I just find that display the source code of the function rather than the whole file is much concise in Material Cloud explore. I propose to add the method to give the source code.
However, It might not be a good idea to store the source of the function as an attribute.

ltalirz:
> I propose to add the method to give the source code.

That sounds good to me.

> However, It might not be a good idea to store the source of the function as an attribute.

I agree that this can be problematic and would advise against it, see also https://github.com/aiidateam/aiida-core/issues/3714 
The basic principle should be to store in the DB what you want to query for and I imagine that querying the source code of workfunctions is not really a use case people have (?)

Issue #4401, opened 2020-09-25
Missing migration for source_file attribute


ltalirz:
This was discovered together with @asle85 when inspecting migrated databases on Materials Cloud.

### Describe the bug
Early aiida-core versions (pre 0.11.1) used to store the whole source file of an inline function in a `source_file` attribute of the corresponding node.

The `source_file` attribute was removed in #1082, and storing the source file in the repository was added in #1196, 
both were merged between aiida-core versions v0.11.0 and v0.11.1

However, it seems a migration was never added, i.e. after migrating databases with pre-0.11.1 calculations to v1.X, the nodes still have the `source_file` attribute and the source file is not added to the file repository:

```
python
In [1]: n = load_node('da3793cf-5b5b-4b6a-b4aa-2cffc5202b04')

In [2]: type(n)
Out[2]: aiida.orm.nodes.process.calculation.calcfunction.CalcFunctionNode

In [3]: list(n.attributes)
Out[3]:
['sealed',
 'namespace',
 'source_code',
 'source_file',
 'function_name',
 'first_line_source_code']

In [4]: n.list_object_names()
Out[4]: []
```

The REST API v4 expects to find the source file in the file repository and therefore errors when encountering such nodes, e.g. 
https://dev-aiida-dev.materialscloud.org/2dstructures/api/v4/nodes/da3793cf-5b5b-4b6a-b4aa-2cffc5202b04/repo/contents?filename=source_file

### Steps to reproduce

Here is an example export file to try: [source_file.zip](https://github.com/aiidateam/aiida-core/files/5283150/source_file.zip) 

### Expected behavior

Both new and old nodes should store the source file in the same place.

In my opinion, the source code is a very important part of the provenance, and so I believe we need to introduce a migration for this.
Luckily, all the information seems still to be present, i.e. we don't need to ask people to re-migrate old databases.

Mentioning @giovannipizzi @sphuber for input.

giovannipizzi:
Here is how I would proceed: I would make sure that the frontend does not make assumptions on the existence of the source code. If it's there, it should be visualised, but if it's not there, nothing should be displayed. This should be true for any attribute in general, we should only do assumptions on basic things like uuid, links, ...

Regarding the migration: I'm not sure, for a few reasons:
- these, when run, weren't even CalcFunctions but InlineCalculations. I think they should be treated similarly to LegacyCalculationJobs, where some fields might differ
[...]

PR #4554, opened 2020-11-11
function node can return the source of function rather than only the source of file

unkcpz:
As mention in #4543 
When checking the source of the process function over MaterialCloud discover(or when explore the aiida archive of others), the whole source file returns and show makes it hard to quickly find the code if there are many other things in the source file. @ltalirz You may want to comment on this ;)
In this PR, I simply rename the `node.get_function_source_code` to `get_function_source_file` and retrieve the source of the function by `inspect.getsource` and store it as an attribute of the node. 

Not sure about whether to store it as a node attribute is a good idea, since if the code of the function is very long it seems not wise to store a long string in attributes. The other choice is along with storing the 'function_starting_line_number' also storing the number of lines of function, and clip the source code of function from these two value.


sphuber:
Thanks @unkcpz . As you said yourself in your comment, we shouldn't store the function's source code in the database. It should be stored in the repository, but there it already is stored, so storing it again would be inefficient and unnecessary. Your other suggestion may actually be better, to, in addition to the starting line number which is already stored, add the total number of lines. The only problem is backwards compatibility. Whatever function that may use this new total number of lines attribute, will have to deal with it not existing, since all existing process function nodes won't have that attribute and I don't think we can or should write a database migration for this. On that note of backwards compatibility, your current PR also breaks it as you change the behavior of the `get_funciton_source_code`, which we cannot do just like that

unkcpz:
Thanks, @sphuber! I notice as you mentioned there is a backwards compatibility problem. But I have no experience in database or orm stuff, what am I suppose to do? Can you point out which part of code is involved and I should look at?

sphuber:
> Thanks, @sphuber! I notice as you mentioned there is a backwards compatibility problem. But I have no experience in database or orm stuff, what am I suppose to do? Can you point out which part of code is involved and I should look at?

You should revert the change of `get_function_source_code` and keep it as it was, i.e. return the entire file. You can then add a method that returns just the source code of the function, but you will have to find a different method name for this.

Secondly, as we discussed, it is not a good idea to store the source code of the function in the database. The best we can do is store the number of lines as an attribute. This you can then use together with the `function_starting_line_number` you can return the correct line range. However, you need to built a case in the logic that accounts for nodes not having this new `line_count`, because all existing nodes won't have it. In that case you need to raise an exception saying that it cannot be done.

aiida/orm/utils/mixins.py
```diff
@@ -105,13 +107,45 @@ def _set_function_starting_line_number(self, function_starting_line_number):
         """
         self.set_attribute(self.FUNCTION_STARTING_LINE_KEY, function_starting_line_number)
 
+    @property
+    def function_number_of_lines(self):
+        """Return the number of lines of the wrapped function in its source file.
+
+        :returns: the number of lines or None
+        """
+        line_count = self.get_attribute(self.FUNCTION_NUMBER_OF_LINES_KEY)
```
unkcpz, line 116: I assume `get_attribute` with specifying default return value will raise the exception for old nodes, correct?

unkcpz (commented):
hi @sphuber I change the code as your recommend. 

Just out of curiosity, what to do if I want to change the method name?

aiida/orm/utils/mixins.py
sphuber: Yes, you can just pass the default to the `get_attribute` call though
```suggestion
        line_count = self.get_attribute(self.FUNCTION_NUMBER_OF_LINES_KEY, None)
```

aiida/orm/utils/mixins.py
```diff
@@ -105,13 +107,45 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
+
+        return line_count
+
+    def _set_function_number_of_lines(self, function_number_of_lines):
+        """Set the number of lines of the wrapped function in its source file.
+
+        :param function_number_of_lines: the number of lines
+        """
```
sphuber, line 124: Maybe add a check that it is a positive integer?

aiida/orm/utils/mixins.py
```diff
@@ -105,13 +107,45 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
+        self.set_attribute(self.FUNCTION_NUMBER_OF_LINES_KEY, function_number_of_lines)
+
     def get_function_source_code(self):
-        """Return the absolute path to the source file in the repository.
+        """Return the content to the source file where the process function locate in the repository.
 
-        :returns: the absolute path of the source file in the repository, or None if it does not exist
+        :returns: the content of the source file where the process function locate in the repository,
+            or None if it does not exist
```
sphuber, line 131:
```suggestion
        :returns: the source file content.
```

aiida/orm/utils/mixins.py
```diff
@@ -105,13 +107,45 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
+    def _set_function_number_of_lines(self, function_number_of_lines):
+        """Set the number of lines of the wrapped function in its source file.
+
+        :param function_number_of_lines: the number of lines
+        """
+        self.set_attribute(self.FUNCTION_NUMBER_OF_LINES_KEY, function_number_of_lines)
+
     def get_function_source_code(self):
-        """Return the absolute path to the source file in the repository.
+        """Return the content to the source file where the process function locate in the repository.
```
sphuber, line 128:
```suggestion
        """Return the content of the entire source file that contains the process function.
```

aiida/orm/utils/mixins.py
```diff
@@ -105,13 +107,45 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
 
-        :returns: the absolute path of the source file in the repository, or None if it does not exist
+        :returns: the content of the source file where the process function locate in the repository,
+            or None if it does not exist
         """
         return self.get_object_content(self.FUNCTION_SOURCE_FILE_PATH)
 
+    def get_function_source(self):
+        """Return the source code of the function, one and only the function,
```
sphuber, line 136:
```suggestion
        """Return the source code of the function and only of the function.
```

aiida/orm/utils/mixins.py
```diff
@@ -105,13 +107,45 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
+
+        :returns: the content of the function
```
sphuber, line 138:
```suggestion
        :returns: the function source code.
```

sphuber (changes requested):

sphuber:
> Just out of curiosity, what to do if I want to change the method name?

You will have to deprecate the `get_function_source_code` saying that in the future this will return only the source of the function and that in the future you should use `get_function_source_file` to get the entire content of the file, and then you forward the `get_function_source_code` by calling `get_function_source_file`. You can look in source code for examples of deprecation warnings. Look for `warnings.warn('some message', AiidaDeprecationWarning)`.

Now, the problem is, that ideally, in the future when we remove the deprecation and make `get_function_source_code` return just the function source, you can do so, but you will remove `get_function_source` (the function you added now) but that would also be breaking. At that point, what we should do is do the change and deprecate `get_function_source` and only remove that a release cycle later. I hope that is clear, if not let me know

aiida/orm/utils/mixins.py
unkcpz: I set the default value to `-1` so for the old nodes this function will return the content to end of the file.

aiida/orm/utils/mixins.py
unkcpz: I don't think it is needed, since this value is a `len`.

unkcpz:
hi @sphuber, anymore to add of this? I also add the deprecated warning.

aiida/orm/utils/mixins.py
```diff
@@ -105,13 +109,47 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
+        """Return the content of the entire source file that contains the process function.
 
-        :returns: the absolute path of the source file in the repository, or None if it does not exist
+        :returns: the content of the source file
         """
+        warnings.warn(
+            'in the future this will return only the source of the function and'
+            ' please use `get_function_source_file` to get'
+            ' the entire content of the file instead', AiidaDeprecationWarning
+        )
```
ramirezfranciscof, line 136: So, there are a couple of comments regarding this depracation you are doing:

1. You need to already include the `get_function_source_file` method so that users can immediately start fixing their scripts with the new way of doing it.
2. The depracation path for this is a bit complicated. I know you discussed with @sphuber in the general thread what would be the *hypothetical* way to do this, but I'm not sure he was actually agreeing it would be a good idea to go this way (maybe he can confirm in case I misinterpreted?). I personally agree that it is better "namewise", but changing to that might generate a bit of annoyance on the users.
3. The warning itself it is a bit unclear to me, I would maybe add a legthier description if we are going to go this way. Something like:

```suggestion
        warnings.warn(
            'you are currently using `get_function_source_code` to get the entire'
            ' file where the process is defined, but in the future this method will'
            ' be used to return only the part of that file that corresponds to the'
            ' process itself. Please use the method `get_function_source_file`'
            ' instead.', AiidaDeprecationWarning
        )
```

aiida/orm/utils/mixins.py
```diff
@@ -105,13 +109,47 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
         return self.get_object_content(self.FUNCTION_SOURCE_FILE_PATH)
 
+    def get_function_source(self):
+        """Return the source code of the function and only of the function.
```
ramirezfranciscof, line 140: If we are doing the deprecation path, I would add to this docstring that this is a temporary method.

I'm also wondering if we should add a test for this or not, considering that it is going to be temporary. I guess it would be the same that adding a unit test for the warning. @sphuber would it be a good idea to have a unit test file with all the deprecation warnings, considering it would have to be explicitly updated when removing the deprecations? Somehow it sounds both a bit insane but not a bad idea at the same time, as a way to keep track of all deprecations in a single place...

ramirezfranciscof (commented):
Hey @unkcpz , nice work! I have a couple of comments to discuss with you and @sphuber. Also, to justify why I'm suddenly interested in this, it might become necessary for an upcoming PR to solve #4401 (see [this comment](https://github.com/aiidateam/aiida-core/issues/4401#issuecomment-775078931), the row for `source_code`).

aiida/orm/utils/mixins.py
unkcpz: > You need to already include the get_function_source_file method so that users can immediately start fixing their scripts with the new way of doing it.

My bad. add it as the implemented function and forward it to `get_function_source_code`. 

> The depracation path for this is a bit complicated. I know you discussed with @sphuber in the general thread what would be the hypothetical way to do this, but I'm not sure he was actually agreeing it would be a good idea to go this way (maybe he can confirm in case I misinterpreted?). I personally agree that it is better "namewise", but changing to that might generate a bit of annoyance on the users.

However, I can not figure out a better way to do this. I think it can reduce the burden to users only by a clear docstring and comprehensive warning information.

> The warning itself it is a bit unclear to me, I would maybe add a legthier description if we are going to go this way. 

Thanks! changed as you recommend.

aiida/orm/utils/mixins.py
unkcpz: The three relevant functions are closely packed in the code. So I think the source code itself with its docstring makes enough explanation for how to do the future migration. Let's see if @sphuber has any better ideas.

unkcpz (commented):
@ramirezfranciscof My apologise for the delay and thanks for such details reviewing.

aiida/orm/utils/mixins.py
ramirezfranciscof: > However, I can not figure out a better way to do this

So, I think the alternative is to accept that `get_function_source_code` now returns the full file and create a new name for the one that just returns the function (`get_function_source_def`?). This is less consistent with the original information (where `source_file` had the file and `source_code` had the definition) but more consistent with the current status (where `source_code` is the name of the file) and also causes less friction with the users because there is no deprecations needed. This was what I understood from the first comment that @sphuber made:

> You should revert the change of get_function_source_code and keep it as it was, i.e. return the entire file. You can then add a method that returns just the source code of the function, but you will have to find a different method name for this.

There was no mention of any further change there, just to rename the new function and leave the old one as it was. Then the longer path in the latter comment read more like a hipothetical situation than a prescription for this case.

Again, I am personally in favour of doing the longer deprecation path, I just want to make sure that we are all on the same page and there is not misunderstandings or something we are not considering.

unkcpz: @ramirezfranciscof  Thanks for the clarification. So in this simple pr, only need to add a new method for section of source code, then left deprecation and renaming to the future, correct? I rebase the PR and please have a look at your available.

aiida/orm/utils/mixins.py
```diff
@@ -105,13 +107,45 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
+
+        :returns: the number of lines or None
+        """
+        return self.get_attribute(self.FUNCTION_NUMBER_OF_LINES_KEY, None)
+
+    def _set_function_number_of_lines(self, function_number_of_lines):
+        """Set the number of lines of the wrapped function in its source file.
+
+        :param function_number_of_lines: the number of lines
+        """
```
sphuber, line 122:
```suggestion
        """
        type_check(function_number_of_lines, int)
```

aiida/orm/utils/mixins.py
```diff
@@ -105,13 +107,45 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
+        :returns: the function source code.
+        """
+        content_list = self.get_function_source_code().splitlines()
+        start_line = self.function_starting_line_number
+        end_line = start_line + self.function_number_of_lines
+
+        # start_line - 1 to include the decorator
+        function_source = '\n'.join(content_list[start_line - 1:end_line])
+
+        return function_source
```
sphuber, line 147:
```suggestion
    def get_function_source_code(self):
        """Return the absolute path to the source file in the repository.
        :returns: the absolute path of the source file in the repository, or None if it does not exist
        """
        warn_deprecation('This method will be removed, use `get_source_code_file` instead.', version=3)
        return self.get_object_content(self.FUNCTION_SOURCE_FILE_PATH)
        
    def get_source_code_file(self) -> str | None:
        """Return the source code of the file in which the function was defined.

        :returns: The source code of the file or ``None`` if not available.
        """
        return self.get_object_content(self.FUNCTION_SOURCE_FILE_PATH, None)

    def get_source_code_function(self) -> str | None:
        """Return the source code of the function including the decorator.

        :returns: The source code of the function or ``None`` if not available.
        """
        source_code = self.get_source_code_file()
        
        if source_code is None:
            return None

        content_list = source_code.splitlines()
        start_line = self.function_starting_line_number
        end_line = start_line + self.function_number_of_lines

        # Start at ``start_line - 1`` to include the decorator
        return '\n'.join(content_list[start_line - 1:end_line])
```

sphuber (changes requested):
Thanks @unkcpz . I think this is good to go in. I have just suggested some changes to deal with the inconvenient name. I would deprecate `get_function_source_code` and rename it `get_source_code_file` and then add `get_source_code_function` with your new implementation.

tests/engine/test_process_function.py
```diff
@@ -179,6 +181,9 @@ def test_process_function(data):
     assert node.function_name == function_name
     assert isinstance(node.function_starting_line_number, int)
 
+    # Check the source code of the function is stored
+    assert node.get_function_source_def() == inspect.getsource(test_process_function)
```
sphuber, line 185:
```suggestion
    assert node.get_function_source_file() == inspect.getsource(test_process_function)
```

aiida/orm/utils/mixins.py
```diff
@@ -106,20 +109,66 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
+        the source code using ``inspect.getsourcefile``.
+
         :returns: The source code of the function or ``None`` if it could not be determined when storing the node.
         """
         try:
             return self.base.repository.get_object_content(self.FUNCTION_SOURCE_FILE_PATH)
         except FileNotFoundError:
             return None
 
+    def get_function_source_function(self):
```
sphuber, line 155: I personally find `get_function_source_function` a bit weird, with the double `function` in there. I would propose `get_source_code_function` and  `get_source_code_file` for the entire file source code. Think that makes more sense no?

sphuber (commented):
All good, thanks @unkcpz , just a question on the method naming

aiida/orm/utils/mixins.py
unkcpz: Yes, I mess up when resolve confliction on github interface. Will change it now.

unkcpz: Changed, I hope it is all fine.

aiida/orm/utils/mixins.py
```diff
@@ -106,20 +109,66 @@ def _set_function_starting_line_number(self, function_starting_line_number):
...
+        """
+        return self.get_attribute(self.FUNCTION_NUMBER_OF_LINES_KEY, None)
+
+    def _set_function_number_of_lines(self, function_number_of_lines):
+        """Set the number of lines of the wrapped function in its source file.
+
+        :param function_number_of_lines: the number of lines
+        """
+        type_check(function_number_of_lines, int)
+        self.set_attribute(self.FUNCTION_NUMBER_OF_LINES_KEY, function_number_of_lines)
```
sphuber, line 126:
```suggestion
        self.base.attributes.set(self.FUNCTION_NUMBER_OF_LINES_KEY, function_number_of_lines)
```

aiida/orm/utils/mixins.py
```diff
@@ -106,20 +109,66 @@ def _set_function_starting_line_number(self, function_starting_line_number):
         """
         self.base.attributes.set(self.FUNCTION_STARTING_LINE_KEY, function_starting_line_number)
 
+    @property
+    def function_number_of_lines(self):
+        """Return the number of lines of the wrapped function in its source file.
+
+        :returns: the number of lines or None
+        """
+        return self.get_attribute(self.FUNCTION_NUMBER_OF_LINES_KEY, None)
```
sphuber, line 118:
```suggestion
        return self.base.attributes.get(self.FUNCTION_NUMBER_OF_LINES_KEY, None)
```

unkcpz:
> Sorry @unkcpz , two more tiny things I missed. They are suggestions so you can just commit them directly, and I will merge. Thanks!

No worries, I am also supposed to find it 😆. Done!

sphuber:
@unkcpz while typing up the commit message, I realized there still was a bug. The new method `get_source_code_function` relies on the `FUNCTION_NUMBER_OF_LINES_KEY` attribute being set, but this will only exist for nodes stored with this patch. Existing nodes will return `None` and so `get_source_code_function` needs to check for this. The same goes for the `starting_line_number`. Anyway, I added type checking and enabled the file for `mypy` which confirmed the bugs that were hiding.

unkcpz:
You are right. The `mypy` type check only manifests the issue but you did not fix it, correct? Do you have any idea how to process if the attributes do not exist for old nodes?

sphuber:
> You are right. The `mypy` type check only manifests the issue but you did not fix it, correct? Do you have any idea how to process if the attributes do not exist for old nodes?

No it didn't automatically fix it. I fixed it by checking whether those attributes return `None` and if so, just return `None`, just as if the source code file doesn't exist. It now reads:
```python
    def get_source_code_function(self) -> str | None:
        """Return the source code of the function including the decorator.
        :returns: The source code of the function or ``None`` if not available.
        """
        source_code = self.get_source_code_file()

        if source_code is None or self.function_number_of_lines is None or self.function_starting_line_number is None:
            return None

        content_list = source_code.splitlines()
        start_line = self.function_starting_line_number
        end_line = start_line + self.function_number_of_lines

        # Start at ``start_line - 1`` to include the decorator
        return '\n'.join(content_list[start_line - 1:end_line])
```
See the conditional `if source code ....`

unkcpz:
Thanks a lot! I retrigger the failed test (3.10) which I guess it is failed randomly. Only the docs tests failed, but I have no idea why it shows here, I don't think there is any code related to that.

sphuber:
I don't understand either why it is failing all of a sudden. I can reproduce it locally, but the problem seems to be with the tutorial notebook. It creates and loads a temporary profile but doesn't set it as a default. So later when it calls `verdi` it operates on a different default profile (the one of the system). This should have nothing to do with these changes, so I wonder if all builds will now fail.

sphuber:
Ok I found the problem, but I understand literally nothing. In the constructor if we do
```python
        try:
            source_file_path = inspect.getsourcefile(func)
            if source_file_path:
                with open(source_file_path, 'rb') as handle:
                    self.base.repository.put_object_from_filelike(  # type: ignore[attr-defined]
                        handle, self.FUNCTION_SOURCE_FILE_PATH
                    )
        except (IOError, OSError):
            pass
```
it works without any provblems. However if we do
```python
        try:
            source_file_path = inspect.getsourcefile(func)
        except (IOError, OSError):
            pass
        else:
            if source_file_path is not None:
                with open(source_file_path, 'rb') as handle:
                    self.base.repository.put_object_from_filelike(  # type: ignore[attr-defined]
                        handle, self.FUNCTION_SOURCE_FILE_PATH
                    )
```
somehow the docs build fails 😕 what the f?

unkcpz:
It seems the `docs/source/intro/tutorial.md` having process function generated inside the jupyter notebook but can not find it afterwards. I got the following log from running `make -C docs html`.

```
~/Projects/WP-aiida/aiida-core/aiida/orm/utils/mixins.py[0m in [0;36mstore_source_info[0;34m(self, func)[0m
[1;32m     64[0m         [0;32melse[0m[0;34m:[0m[0;34m[0m[0;34m[0m[0m
[1;32m     65[0m             [0;32mif[0m [0msource_file_path[0m [0;32mis[0m [0;32mnot[0m [0;32mNone[0m[0;34m:[0m[0;34m[0m[0;34m[0m[0m
[0;32m---> 66[0;31m                 [0;32mwith[0m [0mopen[0m[0;34m([0m[0msource_file_path[0m[0;34m,[0m [0;34m'rb'[0m[0;34m)[0m [0;32mas[0m [0mhandle[0m[0;34m:[0m[0;34m[0m[0;34m[0m[0m
[0m[1;32m     67[0m                     self.base.repository.put_object_from_filelike(  # type: ignore[attr-defined]
[1;32m     68[0m                         [0mhandle[0m[0;34m,[0m [0mself[0m[0;34m.[0m[0mFUNCTION_SOURCE_FILE_PATH[0m[0;34m[0m[0;34m[0m[0m

[0;31mFileNotFoundError[0m: [Errno 2] No such file or directory: '/home/jyu/Projects/WP-aiida/aiida-core/docs/source/intro/ipykernel_2627531/3198245695.py'
FileNotFoundError: [Errno 2] No such file or directory: '/home/jyu/Projects/WP-aiida/aiida-core/docs/source/intro/ipykernel_2627531/3198245695.py'
```

It has some timestamps in the log, but you can find it point to the code we changed in this PR.

sphuber:
> BTW, is the "f" for "fix" as I did for the commit message? laughing

No, in this case it was something else :sweat_smile:

sphuber (approved):

unkcpz:
I understand more why it was going on for the docs fail. By testing the behavior in the IPython shell, it proves my guess that [It seems in ipython or Jupyter notebook, the classes/functions defined or the __main__ module does not have a __file__ attribute associated with them so inspect wasn't able to retrieve a source file.](https://stackoverflow.com/a/51566918) The issue can be reproduced by:

```python
In [1]: from aiida import orm                                                 
                                                                              
In [2]: from aiida.engine import calcfunction                                 
                                                                                                                                                             
In [3]: @calcfunction                                                         
   ...: def add(a, b):                                                                                                                                       
   ...:     return a + b                                                                                                                                     
   ...:                                                                                                                                                      
                                                                                                                                                             
In [4]: d = add(orm.Int(4), orm.Int(5))                                       
<ipython-input-3-f301de63af89>                                                                                                                               
['@calcfunction\n', 'def add(a, b):\n', '    return a + b\n']   

FileNotFoundError                         Traceback (most recent call last)                                                                                  
<ipython-input-4-677ea2e3e348> in <module>                                                                                                                   
----> 1 d = add(orm.Int(4), orm.Int(5))                   
.....

~/Projects/WP-aiida/aiida-core/aiida/engine/processes/functions.py in _setup_db_record(self)                                                                 
    361         """Set up the database record for the process."""                                                                                            
    362         super()._setup_db_record()                                    
--> 363         self.node.store_source_info(self._func)                       
    364                                                                       
    365     @override                                                         

~/Projects/WP-aiida/aiida-core/aiida/orm/utils/mixins.py in store_source_info(self, func)                                                                    
     66                 print(source_file_path)                               
     67                 print(source_list)                                    
---> 68                 with open(source_file_path, 'rb') as handle:                                                                                         
     69                     self.base.repository.put_object_from_filelike(  # type: ignore[attr-defined]                                                     
     70                         handle, self.FUNCTION_SOURCE_FILE_PATH                                                                                       

FileNotFoundError: [Errno 2] No such file or directory: '<ipython-input-3-f301de63af89>'   
```

So in principle, the try-catch block with `IOError` catch this exception and will not store the source file for the jupyter-notebook or IPython shell. It can be regarded as a bug since now we add the `get_function_source_code` which is supposed to return the code of the function but for the IPython shell it returns nothing. 
I think there is an alternative to `inspect` called [`dill`](https://github.com/uqfoundation/dill) that seems able to pickle the jupyter notebook cells.

sphuber:
ah yeah of course, source code storing for interactive shells and notebooks has never been supported. I would rather call this a feature request then.

PR #4555, opened 2020-11-11
🐛 FIX: slurm resource validation

### LINKED ISSUE #4530: The Slurm scheduler sets --ncpus-per-task to a float
### Describe the bug

When submitting a job defining `num_cores_per_machine`,  the scheduler is unable to submit the job because `--ncpus-per-task` is set to a float, which SLURM doesn't understand. The workchain gets then paused forever because it is unsuccesful at submitting the job.

### Steps to reproduce

Using a slurm scheduler in a computer with `mpiprocs_per_machine` defined:

```python
submit(CalculationClass, options={"resources": {"num_machines": x ,"num_cores_per_machine": y})
```

### Your environment

- Operating system [e.g. Linux]: Linux
- Python version [e.g. 3.7.1]: 3.6.9
- aiida-core version [e.g. 1.2.1]: 1.4.2

### Additional context

I think it's related to this part of the code:
https://github.com/aiidateam/aiida-core/blob/9ff07c166a559f98b5b2be71537814ec00d3f18d/aiida/schedulers/plugins/slurm.py#L133-L139

As I understand, this division should return an integer, but the opposite check is performed. I don't know, maybe I'm wrong.

chrisjsewell:
fixes #4530 

I also added resource validation tests and moved most of the tests to pytest from unittest (would have been all, but I didn't find a good replacement for `AssertLogs`)

### DIFF
--- a/aiida/schedulers/plugins/slurm.py
+++ b/aiida/schedulers/plugins/slurm.py
@@ -137,6 +137,7 @@ class SlurmJobResource(NodeNumberJobResource):
                     '`num_cores_per_machine` must be equal to `num_cores_per_mpiproc * num_mpiprocs_per_machine` and in'
                     ' particular it should be a multiple of `num_cores_per_mpiproc` and/or `num_mpiprocs_per_machine`'
                 )
+            resources.num_cores_per_mpiproc = int(resources.num_cores_per_mpiproc)
 
         return resources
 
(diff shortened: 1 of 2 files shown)


sphuber (approved):
Looks good to me. Thanks @chrisjsewell

PR #4559, opened 2020-11-12
Improve archive import memory usage

chrisjsewell:
This is a small iterative improvement to the archive import logic, added to reduce memory overhead, by reducing the number of variables in memory at any one time


ltalirz (approved):
thanks @chrisjsewell , this looks reasonable

ltalirz:
Please choose a more specific title for the commit when merging

Issue #4551, opened 2020-11-08
indicate development version in AiiDA version number


ltalirz:
### Is your feature request related to a problem? Please describe

As of today, the AiiDA version number in the `develop` branch tends to be the version number of the last release.

This means when running on the `develop` branch and
 - creating AiiDA archives
 - running calculations
 - ...

the version number recorded will be that of the latest AiiDA release, despite the fact that the source code may have changed significantly since then.

The output of 
```
$ verdi --version
AiiDA version 1.4.2
```
also currently gives no indication on whether one is running an official release or a development version.

### Describe the solution you'd like

It would be useful, if the AiiDA version reported by an install based on the `develop` branch would have *some* indication of the fact that it is a development version.

This is not something that would have to be invented from scratch; other packages have already built solutions for this - see .e.g

 - [bokeh](https://github.com/bokeh/bokeh/blob/branch-2.3/bokeh/__init__.py#L67)
 - [panel](https://github.com/holoviz/panel/blob/master/panel/__init__.py#L10)
 - [django](https://github.com/django/django/blob/master/django/__init__.py)

(please feel free to add to the list)

greschd:
I see basically two options:
- use the last-released version + suffix
- use the upcoming version + suffix

Since it's not always known which will be the next version, the first option seems better to me. PEP440 then dictates which kind of suffixes can be used: https://www.python.org/dev/peps/pep-0440

I see two options that would make sense (in order of preference):
- `x.y.z+dev`: https://www.python.org/dev/peps/pep-0440/#local-version-identifiers
- `x.y.z.post0`: https://www.python.org/dev/peps/pep-0440/#post-releases (drawback: this has the semantic of being a post-_release_)

Of course you could also do something where the git commit hash is incorporated in some way like `x.y.z+<hash_part>`. Maybe that'd involve too much extra scaffolding?

ltalirz:
> Since it's not always known which will be the next version, the first option seems better to me.

Fully agree.

`x.y.z+dev` sounds fine to me - I don't have strong feelings about the exact string.

> Of course you could also do something where the git commit hash is incorporated in some way like `x.y.z+<hash_part>`. Maybe that'd involve too much extra scaffolding?

Well, that scaffolding already exists e.g. in panel/bokeh
```
>>> import panel
>>> panel.__version__
'0.10.0a25.post3+g834940c'
```
```
>>> import bokeh
>>> bokeh.__version__
[...]

PR #4503, opened 2020-10-23
🧪 add docker-compose for local testing

chrisjsewell:
As outlined in https://github.com/aiidateam/aiida-core/pull/4448#issuecomment-714812704, the purpose of this is to be able to test e.g the CLI and the functionality in a fresh environment:

Creake the docker container and enter it:
```console
$ docker-compose -f .ci/docker-aiida-core.yml up -d --build
$ docker exec -it --user aiida aiida-core /bin/bash
$ verdi shell
```

Create a large node tree (~60,000 nodes)
```python
root = orm.Dict()
recursive_provenance(root, depth=9, breadth=3, num_objects=1)
```

Try running the export function:
```console
$ verdi export create --input-calc-forward -N 1 -- tmp/mount_folder/out.aiida
```

PR #4564, opened 2020-11-13
Add .dockerignore

chrisjsewell:
Currently, when we build our docker image we have the potential to pull in a lot of unnecessary junk (in the step `COPY . aiida-core`).
The most egregious of these is the `.git/` folder which can get particularly large when you are working locally (mine is 3.2Gb!) I don't see any reason why the docker image needs to contain this folder.
Moreover, it almost doubles the build time:

Without `.dockerignore`:

```console
$ docker build -t aiida-core-big .
[+] Building 469.3s (10/10) FINISHED                                                                                            
 => [internal] load .dockerignore                                                                                          0.1s
 => => transferring context: 45B                                                                                           0.1s
 => [internal] load build definition from Dockerfile                                                                       0.1s
 => => transferring dockerfile: 622B                                                                                       0.1s
 => [internal] load metadata for docker.io/aiidateam/aiida-prerequisites:0.2.1                                             0.0s
 => CACHED [1/5] FROM docker.io/aiidateam/aiida-prerequisites:0.2.1                                                        0.0s
 => [internal] load build context                                                                                        175.4s
 => => transferring context: 3.52GB                                                                                      175.3s
 => [2/5] COPY . aiida-core                                                                                               24.0s
 => [3/5] RUN pip install ./aiida-core[atomic_tools]                                                                     227.7s
 => [4/5] COPY .docker/opt/configure-aiida.sh /opt/configure-aiida.sh                                                      0.1s 
 => [5/5] COPY .docker/my_init.d/configure-aiida.sh /etc/my_init.d/40_configure-aiida.sh                                   0.0s 
 => exporting to image                                                                                                    41.7s 
 => => exporting layers                                                                                                   41.7s 
 => => writing image sha256:a414a5c2bd74ec102d41f77523c421f6af0cfca2f62cbc110142831bbcaeef40                               0.0s 
 => => naming to docker.io/library/aiida-core-big                                                                          0.0s 
```

Final image size: 5,180,584,518 bytes

With `.dockerignore`:

```console
$ docker build -t aiida-core-light .
[+] Building 185.6s (10/10) FINISHED                                                                                
 => [internal] load build definition from Dockerfile                                                           0.2s
 => => transferring dockerfile: 622B                                                                           0.1s
 => [internal] load .dockerignore                                                                              0.1s
 => => transferring context: 35B                                                                               0.0s
 => [internal] load metadata for docker.io/aiidateam/aiida-prerequisites:0.2.1                                 0.0s
 => [1/5] FROM docker.io/aiidateam/aiida-prerequisites:0.2.1                                                   0.0s
 => [internal] load build context                                                                              0.4s
 => => transferring context: 117.84kB                                                                          0.3s
 => CACHED [2/5] COPY . aiida-core                                                                             0.0s
 => [3/5] RUN pip install ./aiida-core[atomic_tools]                                                         168.8s
 => [4/5] COPY .docker/opt/configure-aiida.sh /opt/configure-aiida.sh                                          0.0s 
 => [5/5] COPY .docker/my_init.d/configure-aiida.sh /etc/my_init.d/40_configure-aiida.sh                       0.0s 
 => exporting to image                                                                                        16.1s 
 => => exporting layers                                                                                       16.1s 
 => => writing image sha256:dfccbe94265b71a01a565b3e3b6431abaabca18f6aee19dff364785d9eef3f50                   0.0s 
 => => naming to docker.io/library/aiida-core-light                                                            0.0s 
```

Final image size: 1,684,792,282 bytes


ltalirz (commented):
Thanks @chrisjsewell - I agree that it's good to keep the docker image clean of junk, i.e. I agree with basically all items in the list; just with `.git` I'm not sure: it can be useful to have the git repo in the container (e.g. in case you want to quickly update).
Would there be a way to have a more shallow history?

And how did you manage to get your .git folder to 3GB?

chrisjsewell:
>  in case you want to quickly update

Well this was the point of #4503, if you want to work in the docker container with your local (editable) code, you can just mount it:

```yaml
# A small configuration for use in local CI testing,
# to create a container with the current code installed.

# Simply install docker, then run:
# $ docker-compose -f .ci/docker-aiida-core.yml up -d --build
# $ docker exec aiida-core wait-for-services
# use local aiida-core code
# $ docker exec -i aiida-core /bin/bash -c 'pip install --no-deps -e tmp/mount_folder/'
# $ docker exec -it --user aiida aiida-core /bin/bash

# and to power down, after testing:
# $ docker-compose -f .ci/docker-aiida-core.yml down


version: '3.4'
services:
  aiida:
    build:
      context: ../
      dockerfile: Dockerfile
    image: "aiidateam/aiida-core:latest"
    container_name: aiida-core
    networks:
      - aiida-core
    volumes:
      - "../:/tmp/mount_folder"

networks:
  aiida-core:
```

(this what I've been using for all the archive CLI development)

chrisjsewell:
> And how did you manage to get your .git folder to 3GB?

I know, well I have all yours and sebs branches and most probably when I stash changes on changing branches (can't find any left though 🤷)

chrisjsewell:
anyhow we create an image for literally every commit, so you can also just be using the different images when comparing versions etc

sphuber:
I have been developing a lot of my `aiida-core` repo and it is only 145M. Seriously, how the hell did you get it to 3.2GB? Clean out some stale branches and clean out that stash, my goodness...  The difference in size between the two images that you report  is almost exclusively due to the size of your .git folder, when it doesn't need to be that big at all.

chrisjsewell:
Yeh sure. 
But there is still no need for the git folder in the docker container.

sphuber:
> But there is still no need for the git folder in the docker container.

I personally don't have a strong opinion about this, but @ltalirz seems to argue that it can be useful. I will  leave that up to the two of you. Can the file be placed in the `.docker` folder, to try and keep the top level folder from over cluttering with files?

chrisjsewell:
> Can the file be placed in the .docker folder, to try and keep the top level folder from over cluttering with files?

No I'm afraid not, it's read from the context path of the docker build.

ltalirz:
Removing the .git means that it will be no longer easy to figure out from the code which version one is running (also `verdi --version` won't be able to help, even after https://github.com/aiidateam/aiida-core/issues/4551 is addressed).
I.e. if you're testing, say, your plugin against the `latest` docker container, then it might make debugging harder.

For the official docker images built on CI, it is my understanding that this
https://github.com/aiidateam/aiida-core/blob/develop/.github/workflows/ci-code.yml#L147
just fetches one commit (https://github.com/actions/checkout#checkout-v2) i.e. I guess there is no issue.
I.e. it only affects local builds.

I personally haven't used the aiida containers much, and I do get the point that one can select a specific commit by selecting a docker container.

If @csadorf and @yakutovicha are ok with this change, I won't object.

csadorf:
Since the `Dockerfile` and the repository to be copied share the same directory it is not possible to simply do a shallow clone during the build to my knowledge.

That said, it seems to me that the only image file size that really matters is the one from a freshly cloned repository, i.e., build with CI, and that *can* be a shallow clone in which case the size of the `.git/` directory should not matter much. I would therefore argue against removing that directory and just ensuring that whenever the image is build, it is done from a shallow clone to keep the [build context](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#understand-build-context) as small as possible.
```
git clone --depth=1 file://`realpath aiida-core` aiida-core-build-ctx/
cd aiida-core-build-ctx
docker build .
```
With these steps, the repository size is only 14MB.

For me the biggest culprit for build context bloating is `.tox/` with 3.7 GB.

yakutovicha:
> If @csadorf and @yakutovicha are ok with this change, I won't object.

Maybe having a git repo is sometimes useful if one wants to do a quick test/update directly in the Docker container (I did it one or two times). Also, when reporting problems, one may want to know exactly which commit they are working on (as mentioned already by @ltalirz). So I have a slight preference towards keeping the `.git` folder.

chrisjsewell:
The life of a programmer; being mocked for the size of your `.git` 😆 

Anyhow, I wish I hadn't bought it up as its a bit of a red-herring.
The point of this is really was not for the CI built image, it was a pre-cursor to #4256 
With the dockerignore it just ensures that this can be run a quick as possible, I've been using these dockerized integration tests a lot for e.g. the archive import/export development.
If you really want the git in there (even if no one actually uses it 😛) then so be it, it not the end of the world. 

> For me the biggest culprit for build context bloating is .tox/ with 3.7 GB.

indeed, glad to see your using tox 😄

chrisjsewell:
ok, in the interest of getting this merged I have removed `.git` from the ignore list

ltalirz (approved):
thanks @chrisjsewell , I think that `.dockerignore` is not uncontroversial.

ltalirz:
One of the django tests failed with
```
E       django.db.models.deletion.ProtectedError: ("Cannot delete some instances of model 'DbNode' because they are referenced through a protected foreign key: 'DbLink.input'", <QuerySet [<DbLink: WorkFunctionNode (1677) --> Int (1702)>]>)
```

I haven't come across this, have you?

chrisjsewell:
> I haven't come across this, have you?

no 😬 (I've just re-run the job)

PR #4568, opened 2020-11-16
Docs: update citations with AiiDA workflows paper

sphuber:
Citation for the latest paper on the engine is added to the README and
the documentation index page. The paper in `aiida/__init__.py` is also
updated which was still referencing the original publication of 2016.


aiida/__init__.py
```diff
@@ -34,10 +34,12 @@
...
-    'G. Pizzi, A. Cepellotti, R. Sabatini, N. Marzari, and B. Kozinsky,'
-    '"AiiDA: automated interactive infrastructure and database for computational science", '
-    'Comp. Mat. Sci 111, 218-230 (2016); https://doi.org/10.1016/j.commatsci.2015.09.013 '
-    '- http://www.aiida.net.'
+    'Sebastiaan. P. Huber, Spyros Zoupanos, Martin Uhrin, Leopold Talirz, Leonid Kahle, Rico Häuselmann, '
+    'Dominik Gresch, Tiziano Müller, Aliaksandr V. Yakutovich, Casper W. Andersen, Francisco F. Ramirez, Carl S. Adorf,'
+    ' Fernando Gargiulo, Snehal Kumbhar, Elsa Passaro, Conrad Johnston, Andrius Merkys, Andrea Cepellotti, '
+    'Nicolas Mounet, Nicola Marzari, Boris Kozinsky, and Giovanni Pizzi, *AiiDA 1.0, a scalable computational '
+    'infrastructure for automated reproducible workflows and data provenance*, Scientific Data **7**, 300 (2020); '
+    'DOI: [10.1038/s41597-020-00638-4](https://doi.org/10.1038/s41597-020-00638-4)'
```
giovannipizzi, line 42: For this string I would keep a text-only format, rather than markdown, so just put the DOI URL. Also, I would put shortened names (this will be printed in a few places) just with initials - or even simply S.P.Huber et al.? But at least use the initials, to shorten a bit.

Finally, this string is printed also in headers of files. I hope that in 2020 Unicode is supported, but one cannot be so sure (e.g. this is printed in the header of xmgrace files, that most probably could get confused) - so maybe it's safer use the less-nice form `Häuselmann` -> `Haeuselmann` and `Müller` -> `Mueller`

```suggestion
    'https://doi.org/10.1038/s41597-020-00638-4'
```

giovannipizzi (changes requested):

aiida/__init__.py
```diff
@@ -34,10 +34,8 @@
...
 __authors__ = 'The AiiDA team.'
 __paper__ = (
-    'G. Pizzi, A. Cepellotti, R. Sabatini, N. Marzari, and B. Kozinsky,'
-    '"AiiDA: automated interactive infrastructure and database for computational science", '
-    'Comp. Mat. Sci 111, 218-230 (2016); https://doi.org/10.1016/j.commatsci.2015.09.013 '
-    '- http://www.aiida.net.'
+    'S. P. Huber et al. AiiDA 1.0, a scalable computational infrastructure for automated reproducible workflows and '
+    'data provenance, Scientific Data 7, 300 (2020); https://doi.org/10.1038/s41597-020-00638-4'
 )
 __paper_short__ = 'G. Pizzi et al., Comp. Mat. Sci 111, 218 (2016).'
```
giovannipizzi, line 40: Really sorry, I realised only now that there is a `__paper__` and a `__paper_short__`... 

So  one should update also the short one, and use the short or long one where appropriate. Also, note the comma after 'et al.' and I added the double quotes around the title, to keep the same styling as before and also because we have a lot of punctuation in our new title ;-)
I'm suggesting to keep anyway the `et al.` also for the long version that I think it's a good compromise.

```suggestion
    'S. P. Huber et al., "AiiDA 1.0, a scalable computational infrastructure for automated reproducible workflows and '
    'data provenance", Scientific Data 7, 300 (2020); https://doi.org/10.1038/s41597-020-00638-4'
)
__paper_short__ = 'S. P. Huber et al., Sci. Data 7, 300 (2020).'
```

sphuber: Haha no need to be sorry, cannot believe I completely overlooked that

sphuber:
@giovannipizzi this should be good to go

giovannipizzi (approved):
Thanks!

PR #4573, opened 2020-11-17
Enforce verdi quicksetup --non-interactive

chrisjsewell:
When in non-interactive mode, do not ask whether to use existing user/database

### DIFF
--- a/aiida/manage/external/postgres.py
+++ b/aiida/manage/external/postgres.py
@@ -166,7 +166,7 @@ class Postgres(PGSU):
         :returns: tuple (dbname, created)
         """
         if not self.interactive:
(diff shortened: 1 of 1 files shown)


aiida/manage/external/postgres.py
```diff
@@ -163,6 +165,8 @@ def check_db(self, dbname):
         :param str dbname: Name of the database to be created or reused.
         :returns: tuple (dbname, created)
         """
+        if not self.interactive:
+            return dbname, not self.db_exists(dbuser)
```
chrisjsewell, line 169:
```suggestion
            return dbname, not self.db_exists(dbname)
```

ltalirz (approved):
thanks @chrisjsewell

PR #4575, opened 2020-11-17
Improve `verdi node delete` performance

chrisjsewell:
`verdi node delete` is amazingly inefficient;

1) it loads every node as part of click (from a list of pk/uuids) (in `node_delete`)
2) then converts that back to a list of pks (in `node_delete`)
3) then loads all the nodes again to check if they exist (in `delete_nodes`)
4) then performs a query to check (again!) for non-existent pks (in `traverse_graph`)
5) and finally uses the pks to actually compute the full pk/link set to delete, and deletes them.

This PR essentially removes steps 1-3.

It also adds type checking to the files concerned.

### DIFF
--- a/aiida/manage/database/delete/nodes.py
+++ b/aiida/manage/database/delete/nodes.py
@@ -8,12 +8,15 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Function to delete nodes from the database."""
+from typing import Iterable
 
 import click
 from aiida.cmdline.utils import echo
 
 
-def delete_nodes(pks, verbosity=0, dry_run=False, force=False, **kwargs):
+def delete_nodes(
+    pks: Iterable[int], verbosity: int = 0, dry_run: bool = False, force: bool = False, **traversal_rules: bool
+):
     """Delete nodes by a list of pks.
 
     This command will delete not only the specified nodes, but also the ones that are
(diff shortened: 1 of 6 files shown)


chrisjsewell:
@mbercx this will affect #4425. I can actually just make a PR to re-do that if you want

aiida/tools/graph/graph_traversers.py
```diff
@@ -8,34 +8,59 @@
...
     :param starting_pks: Contains the (valid) pks of the starting nodes.
 
-    :param bool get_links:
+    :param get_links:
         Pass True to also return the links between all nodes (found + initial).
 
-    :param bool create_forward: will traverse CREATE links in the forward direction.
-    :param bool call_calc_forward: will traverse CALL_CALC links in the forward direction.
-    :param bool call_work_forward: will traverse CALL_WORK links in the forward direction.
+    :param missing_callback: A callback to handle missing starting_pks or if None raise NotExistent
```
ltalirz, line 50: might be useful to add your example implementation here

ltalirz (commented):
thanks @chrisjsewell 

looks good, just some minor comments.

by the way, since you mention performance, can you add some numbers on the performance improvement?

aiida/tools/graph/graph_traversers.py
```diff
@@ -204,31 +226,31 @@ def traverse_graph(starting_pks, max_iterations=None, get_links=False, links_for
...
     query_nodes.append(orm.Node, project=['id'], filters={'id': {'in': operational_set}})
     existing_pks = set(query_nodes.all(flat=True))
     missing_pks = operational_set.difference(existing_pks)
-    if missing_pks:
+    if missing_pks and missing_callback is None:
         raise exceptions.NotExistent(
-            f'The following pks are not in the database and must be pruned before this   call: {missing_pks}'
+            f'The following pks are not in the database and must be pruned before this call: {missing_pks}'
         )
+    elif missing_pks and missing_callback is not None:
```
ltalirz, line 249: you really don't like nesting, do you ;-)

chrisjsewell:
Note this does not touch the code for deleting the data and repository, so should not introduce any conflicts for #4345

aiida/tools/graph/graph_traversers.py
chrisjsewell: nope, the flatter the better 😆

aiida/manage/database/delete/nodes.py
```diff
@@ -36,61 +39,57 @@ def delete_nodes(pks, verbosity=0, dry_run=False, force=False, **kwargs):
...
-            load_node(pk)
-        except exceptions.NotExistent:
-            echo.echo_warning(f'warning: node with pk<{pk}> does not exist, skipping')
-        else:
-            starting_pks.append(pk)
+    def _missing_callback(_pks: Iterable[int]):
+        for _pk in _pks:
+            echo.echo_warning(f'warning: node with pk<{_pk}> does not exist, skipping')
+
+    # echo.echo('Finding nodes to delete...')
```
chrisjsewell, line 64:
```suggestion
```

oops yeh, artifact from testing

mbercx:
> @mbercx this will affect #4425. I can actually just make a PR to re-do that if you want?

No, that's ok. I still have to fix the tests and rebase it anyways once we're preparing to release v2.0.0.

aiida/tools/graph/graph_traversers.py
chrisjsewell: something like this?

```suggestion
    :param missing_callback: A callback to handle missing starting_pks or if None raise NotExistent
    	For example to ignore them: ``missing_callback=lambda missing_pks: None``
```

aiida/manage/database/delete/nodes.py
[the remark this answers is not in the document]
chrisjsewell: done

chrisjsewell:
> by the way, since you mention performance, can you add some numbers on the performance improvement?

So if you create some empty nodes:

```python
for _ in range(100):
	print(Data().store())
```

current:

```console
$ time verdi node delete --force {1..100}
\
real    0m31.846s
user    0m5.034s
sys     0m0.787s
```

this PR:

```console
$ time verdi node delete --force {101..200}
|
real    0m4.455s
user    0m3.207s
sys     0m0.596s
```

so yeh pretty big!

ltalirz (approved):
thanks @chrisjsewell !

PR #4576, opened 2020-11-18
Fix `verdi --version` in editable mode

### LINKED ISSUE #4556: `verdi --version` can report incorrect version if package is installed in editable mode
To reproduce:

1) Install `aiida-core==1.4.2` in editable mode
2) `verdi --version` prints `1.4.2`
3) Switch to `aiida-core==1.4.3`
4) `verdi --version` still prints `1.4.2`

This is a problem of the `@click.version_option` decorator which does not dynamically inspect the linked library version but the version that was copied when it was installed.

chrisjsewell:
fixes #4556

### DIFF
--- a/aiida/cmdline/commands/cmd_verdi.py
+++ b/aiida/cmdline/commands/cmd_verdi.py
@@ -85,6 +85,7 @@ class MostSimilarCommandGroup(click.Group):
 
 @click.command(cls=MostSimilarCommandGroup, context_settings={'help_option_names': ['-h', '--help']})
 @options.PROFILE(type=types.ProfileParamType(load_profile=True))
+# Note, __version__ should always be passed explicitly here, because click does not retrieve a dynamic version when installed in editable mode
 @click.version_option(__version__, '-v', '--version', message='AiiDA version %(version)s')
 @click.pass_context
 def verdi(ctx, profile):
--- a/aiida/cmdline/commands/cmd_verdi.py
+++ b/aiida/cmdline/commands/cmd_verdi.py
@@ -82,9 +82,26 @@ class MostSimilarCommandGroup(click.Group):
         return None
 
 
+def version_callback(ctx, param, value):  # pylint: disable=unused-argument
+    """Callback for supplying version information"""
+    if not value or ctx.resilient_parsing:
+        return
+    from aiida import __version__
+    click.echo(f'AiiDA version {__version__}')
+    ctx.exit()
+
+
 @click.command(cls=MostSimilarCommandGroup, context_settings={'help_option_names': ['-h', '--help']})
 @options.PROFILE(type=types.ProfileParamType(load_profile=True))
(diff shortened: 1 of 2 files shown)


aiida/cmdline/commands/cmd_verdi.py
```diff
@@ -82,9 +82,26 @@ def get_command(self, ctx, cmd_name):
...
+    if not value or ctx.resilient_parsing:
+        return
+    from aiida import __version__
+    click.echo(f'AiiDA version {__version__}')
+    ctx.exit()
+
+
 @click.command(cls=MostSimilarCommandGroup, context_settings={'help_option_names': ['-h', '--help']})
 @options.PROFILE(type=types.ProfileParamType(load_profile=True))
-@click.version_option(None, '-v', '--version', message='AiiDA version %(version)s')
```
greschd, line 87: I think there's a simpler way to do this by just replacing `None` with `aiida.__version__` in the `version_option`.

chrisjsewell: This is more customisable

greschd: Guess the question the is whether we need that customizability.

It's ok for me to go either way, but we should maybe add a comment somewhere pointing to the problem with a bare `version_option`. I can see someone a few releases from now thinking "hey, I can just replace this with `click.version_option`.

sphuber: Agree with @greschd , we should at least add a comment why `click's` built in does not work if we go the custom route. That being said, I also prefer the more simple implementation. Don't see why preparing for a potential future need for customizability is necessary right now. If it can be solved as suggested by just replacing `None` with `aiida.__version__` then I would go that route

chrisjsewell: so be it, changed

greschd: I think a short comment regarding the click issue still couldn't hurt, something like "We need to pass the `__version__` explicitly because click's default can be wrong when the project is installed in editable mode, see issue #4556."

chrisjsewell: ok fair, this was all clearer with the initial (explicit) implementation I wrote, rather than calling this "faulty" black box convenience function 😛

sphuber: I think it is fine, there is a reason the `click>version_option` provides an argument to not automatically determine it but specify explicitly the version. I agree that a comment is still useful, because the reason _why_ we are passing in an explicit version may not be directly obvious

chrisjsewell: hows that 4bfac3a

greschd: Beautiful, thanks @chrisjsewell

greschd (approved):

greschd (approved):

PR #4577, opened 2020-11-18
Fix wrong raise from dict nodes

### LINKED ISSUE #4454: `Dict` object returns different exceptions for the missing keys.
### Describe the bug
For the dictionary object created like so `d = Dict(dict={'a':{'b':1}})` different exceptions are raised for `d['x']` and for `d['a']['x']`:
- `d['x']` raises `AttributeError`
-  `d['a']['x']` raises `KeyError`

The final result is that 
```python
try:
    d['x']
except KeyError:
    print("Test")
```
Can't catch the error. 

### Steps to reproduce

From the `verdi shell`:
```python
In [1]: from aiida.orm import Dict

In [2]: d = Dict(dict={'a':{'b':1}})

In [3]: d['x']
KeyError                                  Traceback (most recent call last)
/opt/conda/lib/python3.7/site-packages/aiida/orm/implementation/entities.py in get_attribute(self, key)
    153         try:
--> 154             return self._dbmodel.attributes[key]
    155         except KeyError as exception:

KeyError: 'x'

<ipython-input-3-7657742692bd> in <module>
----> 1 d['x']

/opt/conda/lib/python3.7/site-packages/aiida/orm/nodes/data/dict.py in __getitem__(self, key)
     63
     64     def __getitem__(self, key):
---> 65         return self.get_attribute(key)
     66
     67     def __setitem__(self, key, value):

/opt/conda/lib/python3.7/site-packages/aiida/orm/entities.py in get_attribute(self, key, default)
    343         """
    344         try:
--> 345             attribute = self.backend_entity.get_attribute(key)
    346         except AttributeError:
    347             if default is _NO_DEFAULT:

/opt/conda/lib/python3.7/site-packages/aiida/orm/implementation/entities.py in get_attribute(self, key)
    154             return self._dbmodel.attributes[key]
    155         except KeyError as exception:
--> 156             raise AttributeError('attribute `{}` does not exist'.format(exception)) from exception
    157
    158     def get_attribute_many(self, keys):

AttributeError: attribute `'x'` does not exist
```

### Expected behavior

`d['x']` should except with the `KeyError`
### Your environment

- Operating system Linux:
- Python 3.7.4:
- aiida-core 1.4.2:

ramirezfranciscof:
When accessing an attribute as a dict (`dictnode['key']`), the error raised was an AttributeError instead of a KeyError, which is rather unusual. This has been fixed and a test that checks the correct error raise has been added.

Fixes #4454

### DIFF
--- a/aiida/orm/nodes/data/dict.py
+++ b/aiida/orm/nodes/data/dict.py
@@ -62,7 +62,10 @@ class Dict(Data):
             self.set_dict(dictionary)
 
     def __getitem__(self, key):
-        return self.get_attribute(key)
+        try:
+            return self.get_attribute(key)
+        except AttributeError as exc:
+            raise KeyError from exc
 
     def __setitem__(self, key, value):
         self.set_attribute(key, value)
--- a/tests/orm/data/test_dict.py
+++ b/tests/orm/data/test_dict.py
@@ -50,3 +50,15 @@ class TestDict(AiidaTestCase):
         self.assertEqual(self.node['value'], 2)
         self.node.dict.value = 3
         self.assertEqual(self.node['value'], 3)
+
+    def test_correct_raises(self):
+        """Test that the methods for accessing the item raise the correct error.
+
+        * `dictnode['inexistent']` should raise KeyError
+        * `dictnode.dict.inexistent` should raise AttributeError
+        """
+        with self.assertRaises(KeyError):
+            _ = self.node['inexistent_key']
+
+        with self.assertRaises(AttributeError):
+            _ = self.node.dict.inexistent_key


sphuber:
Changes look fine to me, I am just wondering about backwards compatibility. Some code out there may rely on this (maybe incorrectly) raising an `AttributeError` instead of a `KeyError`. Thoughts @ltalirz and @giovannipizzi ?

ramirezfranciscof:
Also paging @yakutovicha who run into this problem originally and actually had to code around it.

ltalirz:
Hm... that's a tricky one.
There are [quite a lot of search results on github](https://github.com/search?p=1&q=aiida+except+AttributeError&type=Code) - not all of them would break but some of them would, e.g.
https://github.com/electronic-structure/aiida-exciting/blob/c7f4d2d2370dc0b5cb1270883ff24d4880a11dad/aiida_exciting/calculations/exciting.py#L148-L153

In that particular case it wouldn't be too bad since an exception is raised anyhow but it's possible that other plugins catch the exception and do something else.
In principle this is a breaking change.

giovannipizzi:
I fear here we have to define a mixed exception that inherits from both?
The question, however, is how to issue a deprecation warning if people catch the AttributeError?
Maybe we just mention in the docs and will remove in 2.0 and for now we just put in the list of things to mention for backward-incompatibility (and we already inform plugin developers to catch the new one, or even both, from the next 1.x release).

Or we accept that forever in the future the exception raised there is a mixed exception (BTW, the aiida-exciting plugin is still at version 0.x so I wouldn't worry about it, but it's true that others might be catching it.

yakutovicha:
> Changes look fine to me, I am just wondering about backwards compatibility. Some code out there may rely on this (maybe incorrectly) raising an `AttributeError` instead of a `KeyError`. Thoughts @ltalirz and @giovannipizzi ?

I think people, who relied on the "incorrect" behaviour should have put a catch for both: `AttributeError` and `KeyError`exceptions, otherwise, they would experience the same issue as I: https://github.com/aiidateam/aiida-cp2k/pull/116

ramirezfranciscof:
This behaviour was introduced in version `1.4.0` and is present in all `1.4.x` and `1.5.0` versions, how do you want to proceed @sphuber ?

sphuber:
Not sure, `v1.4` has been out for over 2 months and since this bug was introduced there, we should technically fix it in `v1.4` and release `v1.4.4`.

ramirezfranciscof:
I mean, I think this fix should be simple enough that I can do a fix in `1.4.4` and `1.5.1` (you can tell me exactly how best to proceed to do this). But the question is do we need to concern ourselves with backwards compatibility for this?

sphuber:
> I mean, I think this fix should be simple enough that I can do a fix in `1.4.4` and `1.5.1` (you can tell me exactly how best to proceed to do this).

Just create a `support/1.4.x` that branches of `v1.4.3` and change this PR to merge on to that support branch. Since `master` is already at `v1.5` we are not going to merge this support branch in, but we release directly off of that branch. Then you simply create another PR onto `release/1.5.1` (which I already created), cherry-picking then this merged commit that fixes the problem. Once we release `v1.5.0` we also merge that back into `develop` which will then also have the fix.

> But the question is do we need to concern ourselves with backwards compatibility for this?

No. Since we simply added a bug in v1.4 and we are now fixing it in a patch version of the minor branch, backwards compatibility doesn't really apply here.

sphuber:
@ramirezfranciscof can you rebase this on `release/1.5.1` and update the PR so we can merge this? It would be good to release `1.5.1` a.s.a.p.

ramirezfranciscof:
@sphuber hey, I'm not sure what is happening here. The two tests for 3.6 are failing during installation and django 3.8 (but not sqlalchemy 3.8?) is failing due to an error that doesn't seem to be related to dicts.

sphuber:
See [this issue](https://github.com/aiidateam/aiida-core/issues/4614). I have already opened PR #4615 to fix it. Just needs to be approved.

ramirezfranciscof:
@sphuber this one should be ready to merge.

sphuber (approved):
I am getting a déjà vu...

PR #4579, opened 2020-11-18
CI: Add workflow to run tests against various RabbitMQ versions

sphuber:
The main test workflow runs against a single version of RabbitMQ but
experience has shown that the code can break for different versions of
the RabbitMQ server. Here we add a new CI workflow that runs various
unit tests through pytest that simulate the typical interaction with the
RabbitMQ server in normal AiiDA operation. The difference is that these
are tested against the currently available versions of RabbitMQ.

The current setup, still only tests part of the functionality that AiiDA
uses, for example, the default credentials and virtual host are used.
Connections over TLS are also not tested. These options would require
the RabbitMQ service that is running in a docker container to be
configured differently. It is not clear how these various options can be
parametrized in concert with the actual unit tests.

### DIFF
--- a/.github/workflows/rabbitmq.yml
+++ b/.github/workflows/rabbitmq.yml
@@ -1,0 +1,67 @@
+name: rabbitmq
+
+on:
+    push:
+        branches-ignore: [gh-pages]
+    pull_request:
+        branches-ignore: [gh-pages]
+        paths-ignore: ['docs/**']
+
+jobs:
+
+    tests:
+
+        runs-on: ubuntu-latest
+        timeout-minutes: 30
+
+        strategy:
+            fail-fast: false
+            matrix:
+                rabbitmq: [3.5, 3.6, 3.7, 3.8]
+
+        services:
+            postgres:
+                image: postgres:10
+                env:
+                    POSTGRES_DB: test_django
+                    POSTGRES_PASSWORD: ''
+                    POSTGRES_HOST_AUTH_METHOD: trust
+                options: >-
+                    --health-cmd pg_isready
+                    --health-interval 10s
+                    --health-timeout 5s
+                    --health-retries 5
+                ports:
+                -   5432:5432
+            rabbitmq:
+                image: rabbitmq:${{ matrix.rabbitmq }}
+                ports:
+                -   5672:5672
+
+        steps:
+        -   uses: actions/checkout@v2
+
+        -   name: Set up Python 3.8
+            uses: actions/setup-python@v2
+            with:
+                python-version: 3.8
+
+        -   name: Install system dependencies
+            run: |
+                sudo apt update
+                sudo apt install postgresql-10
+
+        -   name: Upgrade pip
+            run: |
+                pip install --upgrade pip
+                pip --version
+
+        -   name: Install aiida-core
+            run: |
+                pip install -r requirements/requirements-py-3.8.txt
+                pip install --no-deps -e .
+                reentry scan
+                pip freeze
+
+        -   name: Run tests
+            run: pytest -sv tests/manage/external/test_rmq.py
--- a/.github/workflows/ci-code.yml
+++ b/.github/workflows/ci-code.yml
@@ -81,7 +81,6 @@ jobs:
 
     - name: Install system dependencies
       run: |
-        sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
         sudo apt update
         sudo apt install postgresql-10 graphviz
 
--- a/.github/workflows/ci-style.yml
+++ b/.github/workflows/ci-style.yml
@@ -23,7 +23,6 @@ jobs:
 
     - name: Install system dependencies
       run: |
-        sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
         sudo apt update
         sudo apt install libkrb5-dev ruby ruby-dev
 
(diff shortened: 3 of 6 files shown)


.github/workflows/rabbitmq.yml
```diff
@@ -0,0 +1,68 @@
...
+        -   uses: actions/checkout@v2
+
+        -   name: Set up Python 3.8
+            uses: actions/setup-python@v2
+            with:
+                python-version: 3.8
+
+        -   name: Install system dependencies
+            run: |
+                sudo rm -f /etc/apt/sources.list.d/dotnetdev.list /etc/apt/sources.list.d/microsoft-prod.list
```
chrisjsewell, line 51: just out of interest, what are these and why do they need to be removed?

sphuber: They are also in the `ci-code.yml` workflow. They were added a few months back because that source list was causing the `apt` commands to fail. I can remove it here to see if the problem has been resolved.

.github/workflows/rabbitmq.yml
```diff
@@ -0,0 +1,68 @@
...
+                    --health-cmd pg_isready
+                    --health-interval 10s
+                    --health-timeout 5s
+                    --health-retries 5
+                ports:
+                -   5432:5432
+            rabbitmq:
+                image: rabbitmq:${{ matrix.rabbitmq }}
+                ports:
+                -   5672:5672
```
chrisjsewell, line 39:
```suggestion
                ports:
                -   5672:5672
                options: >-
                    --health-cmd rabbitmq-diagnostics -q ping
                    --health-interval 30s
                    --health-timeout 30s
                    --health-retries 5
```

this I believe is the recommended health check for rabbitmq (see `.ci/docker-rabbitmq.yml`) but obviously not strictly

sphuber:
@chrisjsewell do you have an idea how we can parametrize these tests over different configurations of the RabbitMQ server? As I understand, this RabbitMQ service in the workflow is simply a direct installation of their official [docker image](https://hub.docker.com/_/rabbitmq/) which actually allows to configure the credentials, virtual host and SSL configuration, but as I understand it, it can only take one value.

The only possibility I see now is to have a manual step (not using the automated service) to install the docker instance, go in and use the management tools to create an additional user and virtual host and then in the unit tests parametrize over the various options to connect with these different parameters.

chrisjsewell:
> different configurations of the RabbitMQ server

maybe something like?

```yaml
services:
	rabbitmq:
	image: rabbitmq:${{ matrix.rabbitmq }}
	ports:
	-   5672:5672
    options: -v ${{ github.workspace }}/path/to/rabbitmq-${{ matrix.config }}.conf:/etc/rabbitmq/rabbitmq.conf
```

chrisjsewell:
> As I understand, this RabbitMQ service in the workflow is simply a direct installation of their official docker image

FYI you can actually just start these services "directly", by running a docker-compose step.

```yaml
    steps:
    - uses: actions/checkout@v1
    - name: Build the docker-compose stack
      run: docker-compose -f .ci/docker-rabbitmq.yml up -d
    - name: Check running containers
      run: docker ps -a
```

this way you could test if its working locally

.github/workflows/rabbitmq.yml
sphuber: it seems like adding these options has caused shit to stop working. Is the syntax different with GHA? It doesn't seem to like the `-q` flag. I think because of quoting probably, it should be interpreted as part of the value for `--health-cmd` but is seen as a separate flag.

chrisjsewell: yep that should do it

chrisjsewell: weird, I literally copy/pasted the command GH generates and it run fine locally: `docker create --name 9489d5a84fa44440b65da8d4368eb107_rabbitmq36_8515e2 --label 179394 --network github_network_bba19cb5634245c991e75611b37d8419 --network-alias rabbitmq -p 5672:5672 --health-cmd 'rabbitmq-diagnostics -q ping' --health-interval 30s --health-timeout 30s --health-retries 5 -e GITHUB_ACTIONS=true -e CI=true rabbitmq:3.6`

chrisjsewell: maybe double-quote or bust

sphuber: > unknown shorthand flag: 'q' in -q

sphuber: double quoting seems to have worked and it now runs for 3.7 and 3.8. However, for 3.5 and 3.6 the health checks fail. Are these options specific for the newer versions?

sphuber:
@chrisjsewell  I removed the health checks because for some unexplicable reason they fail for 3.5 and 3.6. The tests work fine without them. I think this needs to be merged before #4317 which now is fully ready to go, besides this PR. Is there anything else that needs to be changed here?

chrisjsewell (approved):
yep looks fine to me 👍

chrisjsewell:
As I mention above, I definitely think there are ways to set up rabbitmq differently, if we wanted to test SSL etc. But I haven't got the time or inclination to do that right now lol.

sphuber:
Thanks for the review @chrisjsewell 

> As I mention above, I definitely think there are ways to set up rabbitmq differently, if we wanted to test SSL etc. But I haven't got the time or inclination to do that right now lol.

I agree that it can and ideally should be improved. Also fine to do this at a later point. I have opened an issue for that: https://github.com/aiidateam/aiida-core/issues/4591

PR #4583, opened 2020-11-21
Add options for transport tasks

mbercx:
When encountering failures during the execution of transport tasks, a runner
will wait for a time interval between transport task attempts. This time
interval between attemps is increased using an exponential backoff
mechanism, i.e. the time interval is equal to:

`(TRANSPORT_TASK_RETRY_INITIAL_INTERVAL) * 2 ** (N_ATTEMPT - 1)`

where `N_ATTEMPT` is the number of failed attempts. This mechanism is
interrupted once the `TRANSPORT_TASK_MAXIMUM_ATTEMPTS` is reached.

The initial interval and maximum attempts are currently fixed to 20
seconds and 5, respectively. This commit adds two configuration options
that use these defaults, but allow the user to adjust them using `verdi
config`.

### DIFF
--- a/aiida/engine/processes/calcjobs/tasks.py
+++ b/aiida/engine/processes/calcjobs/tasks.py
@@ -32,6 +32,9 @@ UPDATE_COMMAND = 'update'
 RETRIEVE_COMMAND = 'retrieve'
 KILL_COMMAND = 'kill'
 
+RETRY_INTERVAL_OPTION = 'transport.task_retry_initial_interval'
+MAX_ATTEMPTS_OPTION = 'transport.task_maximum_attempts'
+
 logger = logging.getLogger(__name__)
 
 
@@ -61,8 +64,8 @@ def task_upload_job(process, transport_queue, cancellable):
         logger.warning(f'CalcJob<{node.pk}> already marked as SUBMITTING, skipping task_update_job')
         raise Return
 
-    initial_interval = get_config().get_option('transport.task_retry_initial_interval')
-    max_attempts = get_config().get_option('transport.task_maximum_attempts')
+    initial_interval = get_config().get_option(RETRY_INTERVAL_OPTION)
+    max_attempts = get_config().get_option(MAX_ATTEMPTS_OPTION)
 
     authinfo = node.computer.get_authinfo(node.user)
 
@@ -122,8 +125,8 @@ def task_submit_job(node, transport_queue, cancellable):
         logger.warning(f'CalcJob<{node.pk}> already marked as WITHSCHEDULER, skipping task_submit_job')
         raise Return(node.get_job_id())
 
-    initial_interval = get_config().get_option('transport.task_retry_initial_interval')
-    max_attempts = get_config().get_option('transport.task_maximum_attempts')
+    initial_interval = get_config().get_option(RETRY_INTERVAL_OPTION)
+    max_attempts = get_config().get_option(MAX_ATTEMPTS_OPTION)
 
     authinfo = node.computer.get_authinfo(node.user)
 
@@ -170,8 +173,8 @@ def task_update_job(node, job_manager, cancellable):
         logger.warning(f'CalcJob<{node.pk}> already marked as RETRIEVING, skipping task_update_job')
         raise Return(True)
 
-    initial_interval = get_config().get_option('transport.task_retry_initial_interval')
-    max_attempts = get_config().get_option('transport.task_maximum_attempts')
+    initial_interval = get_config().get_option(RETRY_INTERVAL_OPTION)
+    max_attempts = get_config().get_option(MAX_ATTEMPTS_OPTION)
 
     authinfo = node.computer.get_authinfo(node.user)
     job_id = node.get_job_id()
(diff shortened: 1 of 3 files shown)


aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -31,8 +32,8 @@
 RETRIEVE_COMMAND = 'retrieve'
 KILL_COMMAND = 'kill'
 
-TRANSPORT_TASK_RETRY_INITIAL_INTERVAL = 20
-TRANSPORT_TASK_MAXIMUM_ATTEMTPS = 5
+TRANSPORT_TASK_RETRY_INITIAL_INTERVAL = get_config().get_option('transport.task_retry_initial_interval')
+TRANSPORT_TASK_MAXIMUM_ATTEMPTS = get_config().get_option('transport.task_maximum_attempts')
```
chrisjsewell, line 36: I think these should just be called dynamically in the functions, rather than set as global variables

aiida/manage/configuration/options.py
```diff
@@ -185,6 +185,22 @@
         'description': 'Boolean whether to print AiiDA deprecation warnings',
         'global_only': False,
     },
+    'transport.task_retry_initial_interval': {
+        'key': 'task_retry_initial_interval',
+        'valid_type': 'int',
+        'valid_values': None,
```
chrisjsewell, line 191: would be good here if we could validate for `>0`

chrisjsewell (changes requested):

aiida/manage/configuration/options.py
chrisjsewell: I could imagine changing this `valid_type`/`valid_values` into something akin to a [jsonschema](https://pypi.org/project/jsonschema/) or just allowing it to be callable.

e.g.

```json
{
  `validator`: {'type': 'integer', 'minimum': 1}
}
```

or

```json
{
  `validator`: lambda: v: isinstance(v, int) and v > 0
}
```

aiida/engine/processes/calcjobs/tasks.py
mbercx: I also considered this. I left the global variables because this way you only have to get the options once, instead of every time a function is called. However, I'm not very familiar with how the engine works, and the functions aren't called that often, so you're probably right.

aiida/manage/configuration/options.py
mbercx: Extending the validation is definitely a good idea, but perhaps we should open a new issue/PR for this? Note that the `valid_type` is also used to parse the option value, so I'm not sure if we can just merge `valid_type`/`valid_values` into one `validator` function.

aiida/manage/configuration/options.py
```diff
@@ -185,6 +185,22 @@
...
+        'default': 20,
+        'description': 'Initial time interval for the exponential backoff mechanism.',
+        'global_only': False,
+    },
+    'transport.task_maximum_attempts': {
+        'key': 'task_maximum_attempts',
+        'valid_type': 'int',
+        'valid_values': None,
+        'default': 5,
```
mbercx, line 200: Something I also noticed is that some defaults are just set in the option `dict`, and some are set by global variables  at the top of the module. We should probably be consistent in this. I'd be in favour of removing the global variables, I think the defaults are easy enough to find.

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: No because this would mean if you changed the config during your python session, you would still be using the old value. It definitely needs to be changed.

mbercx: Right, good point 👍

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -63,8 +61,8 @@ def task_upload_job(process, transport_queue, cancellable):
         logger.warning(f'CalcJob<{node.pk}> already marked as SUBMITTING, skipping task_update_job')
         raise Return
 
-    initial_interval = TRANSPORT_TASK_RETRY_INITIAL_INTERVAL
-    max_attempts = TRANSPORT_TASK_MAXIMUM_ATTEMTPS
+    initial_interval = get_config().get_option('transport.task_retry_initial_interval')
```
chrisjsewell, line 64: a nitpick would be that you could make the actual option strings global variables, e.g.

```suggestion
    initial_interval = get_config().get_option(RETRY_INTERVAL_KEY)
```

aiida/manage/configuration/options.py
chrisjsewell: yeh makes sense if they are not used anywhere else

aiida/engine/processes/calcjobs/tasks.py
mbercx: 👍 I've named the global variables `RETRY_INTERVAL_OPTION` and `MAX_ATTEMPTS_OPTION`, since each option also has a `'key'` key, and this is not the one you pass to `get_option()`.

chrisjsewell (approved):
Cheers

mbercx:
Thanks for the review, @chrisjsewell!

Issue #4586, opened 2020-11-23
`get_config().get_option()` always returns the default value


mbercx:
### Describe the bug

When trying to use `get_config().get_option()` to obtain the value of a configured option within the current profile, I still obtain the default value.

### Steps to reproduce

```console
$ verdi config runner.poll.interval 123
```

In a `verdi shell`:

```Python
In [1]: from aiida.manage.configuration import get_config, get_config_option, load_config

In [2]: get_config_option('runner.poll.interval')
Out[2]: 123

In [3]: get_config().get_option('runner.poll.interval')
Out[3]: 60
```

### Expected behavior

I'd expect the `get_config` function to obtain the current configuration, so:

```Python
In [3]: get_config().get_option('runner.poll.interval')
Out[3]: 123
```

Is this not the intended behaviour?

### Your environment

- Operating system [e.g. Linux]: Ubuntu
- Python version [e.g. 3.7.1]: 3.7.7
- aiida-core version [e.g. 1.2.1]: develop branch

### Additional context

The recent PR on adding options for transport tasks EBM (#4583) doesn't actually work, since it relies on using `get_config().get_option()` instead of `get_config_option` (which would have definitely been better, in hindsight).

mbercx:
@chrisjsewell this means a2d6c7673952ae48c0d5ae58aff2d2c808e3a982 needs to be fixed, since now the default is always used, even when the option has been configured. Would it be best to open another PR for this, or can we still `--amend` the commit?

mbercx:
@chrisjsewell i've added bdb3060 in the branch of my fork that fixes this issue for the EBM transport options. Now the configured value is picked up:

```console
$ verdi config transport.task_maximum_attempts 
1
$ verdi process list
  PK  Created    Process label    Process State    Process status
----  ---------  ---------------  ---------------  ------------------------------------------------------------------------------------
 593  3m ago     PwBaseWorkChain  ⏵ Waiting        Waiting for child processes: 598
 598  3m ago     PwCalculation    ⏸ Waiting       Pausing after failed transport task: submit_calculation failed 1 times consecutively
```

PR #4587, opened 2020-11-24
Fix command for getting config options

mbercx:
Currently the transport options for the EBM (#4583) are obtained by using 
the get_config function, e.g.:

initial_interval = get_config_option(RETRY_INTERVAL_OPTION)

However, it seems that get_config does not get you the current
configuration, see #4586. 

Replacing get_config().get_option() with get_config_option() fixes this
issue for the EBM options.


chrisjsewell (approved):
yeh easier to just add a new commit

PR #4593, opened 2020-11-27
Dependencies: update minimum version for `notebook>=6.1.5`

sphuber:
Lower versions suffer from vulnerability `GHSA-c7vm-f5p4-8fqh`.

Also update the requirement files to only use explicit pinned versions.
The compatibility operator was erroneously used for the `aio-pika`,
`pamqp` and `pytest-asyncio` dependencies.

For `pamqp` the minimum required version is upped to `2.3` since that
was the version that introduced the `support_deprecated_rabbitmq`
function that is required from that library.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -38,7 +38,7 @@
         "jinja2~=2.10",
         "kiwipy[rmq]~=0.6.1",
         "numpy~=1.17",
-        "pamqp~=2.0",
+        "pamqp~=2.3",
         "paramiko~=2.7",
         "plumpy~=0.17.1",
         "pgsu~=0.1.0",
@@ -91,7 +91,7 @@
         ],
         "notebook": [
             "jupyter~=1.0",
-            "notebook~=6.0"
+            "notebook~=6.0,>=6.1.5"
         ],
         "pre-commit": [
             "mypy==0.790",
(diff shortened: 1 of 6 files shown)


csadorf (changes requested):
The specs in the `requirements/` files should be pinned.

sphuber:
> The specs in the `requirements/` files should be pinned.

There are some other dependencies in those files that use the compatibility operator `~=`. Do you want me to update those in this PR as well?

csadorf:
> > The specs in the `requirements/` files should be pinned.
> 
> There are some other dependencies in those files that use the compatibility operator `~=`. Do you want me to update those in this PR as well?

Yes, please. These files are used to install our test environments, so they should be pinned.

sphuber:
@csadorf I have updated the requirement files. Pinning `pamqp` revealed a problem with the original requirements in the `setup.json` which declared `~-2.0` but in fact we need `~=2.3` at the least. I included this in this PR, even though that wasn't the original intention.

setup.json
```diff
@@ -91,7 +91,7 @@
         ],
         "notebook": [
             "jupyter~=1.0",
-            "notebook~=6.0"
+            "notebook~=6.0,>=6.1.5"
```
csadorf, line 94: Technically valid, but I think this would be better:
```suggestion
            "notebook~=6.1,>=6.1.5"
```

csadorf (changes requested):

csadorf (approved):

PR #3308, opened 2019-09-11
upf to json converter

simonpintarelli:
- `_prepare_json` to upf class
- added export, e.g. `verdi data upf export PK` to cli interface

PR #4597, opened 2020-11-29
WIP: parse more info from upf header

unkcpz:
The `parse_upf` method is very old and only support parsing the element from upf file. I find it is sometimes helpful to also parse the `z_valence`(I want to get this value to set `nbands` before I run the scf calculation, I am curious why we cannot set things like `nbands_factor` in quantum-espresso) and more information from the upf file's `PP_HEADER` such as the type of PP. Just like what @simonpintarelli did in https://github.com/aiidateam/aiida-core/pull/3308 to support converting upf to json. I propose to have a tool to elaborately parsing the upf(or maybe if necessary for other types of pseudopotentials). From the time being, I just add a parser for 'z_valence'. It might useful to parse the following infos from file, [element, z_valence, pp_type, pp_energy, functional, wfc_cutoff, rho_cutoff, relativistic]. 

Here are the things worth to be discussed:

- I am not sure whether to set each of the info as an attribute of `UpfData` or collect them as a dict and set it as one attribute. 
- What's kinds of infos from pp file going to be store?
- Do we need and maintain another package to parse the upf(such as [upf_to_json](https://github.com/simonpintarelli/upf_to_json) by @simonpintarelli , and I think that package can be simplified or replaced by a more versatile one) . Meanwhile, how versatile it would be for other simulation code. Or maybe this tool can be a part of `aiida-pseudo`?

Pining @giovannipizzi @simonpintarelli for inputs before I moving forward.

### DIFF
--- a/aiida/orm/nodes/data/upf.py
+++ b/aiida/orm/nodes/data/upf.py
@@ -36,6 +36,24 @@ REGEX_ELEMENT_V2 = re.compile(
     """, re.VERBOSE
 )
 
+REGEX_HEADER_V1 = re.compile(
+    r"""
+    <PP_HEADER>\s*
+    (?P<header>.*)\s*
+    </PP_HEADER>
+    """, re.DOTALL | re.VERBOSE
+)
+
+REGEX_HEADER_V2 = re.compile(r"""
+    <PP_HEADER\s
+    (?P<header>.*?)\s*
+    /?>
+    """, re.DOTALL | re.VERBOSE)
+
+REGEX_HEADER_PARA_V2 = re.compile(r"""
+    (\w*)\s*=\s*"\s*(.*)\s*"
+    """, re.VERBOSE)
+
 
 def get_pseudos_from_structure(structure, family_name):
     """Return a dictionary mapping each kind name of the structure to corresponding `UpfData` from given family.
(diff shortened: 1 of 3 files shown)


simonpintarelli:
@unkcpz the `upf_to_json` is a copy of the scripts we are using in SIRIUS to create input files. It is needed for running SIRIUS accelerated CP2K via aiida, but the code was too messy to be merged into aiida, thus it became a standalone package :smile:. I'm not aware that any other DFT code needs UPF files converted to something else.

sphuber:
@unkcpz I actually [already implemented the same thing in `aiida-pseudo`](https://github.com/aiidateam/aiida-pseudo/pull/24) that has another implementation of `UpfData` that (I think should) become the successor of the `UpfData` currently shipped with `aiida-core`. We were planning on removing all materials science specific code to plugins for `aiida-core==2.0.0`. I started `aiida-pseudo` for different reasons, but I thought it made a lot of sense to move `UpfData` there and so already implemented a clean version there. The z valence is thus already parsed in `aiida-pseudo==0.3.0`

unkcpz:
@sphuber thanks! I'll have a try on this feature of `aiida-psuedo`.

PR #4602, opened 2020-11-30
Fix bug in SqlAlchemy migrations

### LINKED ISSUE #4590: Potential bug in SqlAlchemy migrations
I have a production SqlAlchemy database that fails to migrate with `aiida-core==1.5.0`. It starts at revision `118349c10896`, migrates successfully to `bf591f31dd12`, but then fails to migrate to revision `0edcdd5a30f0` with the exception:
```
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1278, in _execute_context
    cursor, statement, parameters, context
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 593, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.ObjectInUse: cannot ALTER TABLE "db_dbgroup" because it has pending trigger events


  File "/home/sph/.virtualenvs/aiida_repo/bin/verdi", line 12, in <module>
    sys.exit(verdi())
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/click/core.py", line 829, in __call__
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/click/core.py", line 610, in invoke
  File "/home/sph/code/aiida/env/repo/aiida-core/aiida/cmdline/commands/cmd_database.py", line 43, in database_migrate
    backend.migrate()
  File "/home/sph/code/aiida/env/repo/aiida-core/aiida/orm/implementation/sqlalchemy/backend.py", line 47, in migrate
    self._schema_manager.migrate()
  File "/home/sph/code/aiida/env/repo/aiida-core/aiida/backends/manager.py", line 149, in migrate
    self._migrate_database_version()
  File "/home/sph/code/aiida/env/repo/aiida-core/aiida/backends/sqlalchemy/manager.py", line 151, in _migrate_database_version
    command.upgrade(config, 'head')
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/command.py", line 298, in upgrade
    script.run_env()
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/script/base.py", line 489, in run_env
    util.load_python_file(self.dir, "env.py")
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/util/pyfiles.py", line 98, in load_python_file
    module = load_module_py(module_id, path)
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/util/compat.py", line 184, in load_module_py
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/sph/code/aiida/env/repo/aiida-core/aiida/backends/sqlalchemy/migrations/env.py", line 75, in <module>
    run_migrations_online()
  File "/home/sph/code/aiida/env/repo/aiida-core/aiida/backends/sqlalchemy/migrations/env.py", line 68, in run_migrations_online
    context.run_migrations()
  File "<string>", line 8, in run_migrations
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/runtime/environment.py", line 846, in run_migrations
    self.get_context().run_migrations(**kw)
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/runtime/migration.py", line 520, in run_migrations
    step.migration_fn(**kw)
  File "/home/sph/code/aiida/env/repo/aiida-core/aiida/backends/sqlalchemy/migrations/versions/0edcdd5a30f0_dbgroup_extras.py", line 37, in upgrade
    op.alter_column('db_dbgroup', 'extras', nullable=False)
  File "<string>", line 8, in alter_column
  File "<string>", line 3, in alter_column
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/operations/ops.py", line 1779, in alter_column
    return operations.invoke(alt)
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/operations/base.py", line 374, in invoke
    return fn(self, operation)
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/operations/toimpl.py", line 56, in alter_column
    **operation.kw
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/ddl/postgresql.py", line 147, in alter_column
    **kw
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/ddl/impl.py", line 178, in alter_column
    existing_comment=existing_comment,
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/alembic/ddl/impl.py", line 140, in _exec
    return conn.execute(construct, *multiparams, **params)
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1014, in execute
    return meth(self, multiparams, params)
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/sqlalchemy/sql/ddl.py", line 72, in _execute_on_connection
    return connection._execute_ddl(self, multiparams, params)
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1076, in _execute_ddl
    compiled,
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1318, in _execute_context
    e, statement, parameters, cursor, context
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1512, in _handle_dbapi_exception
    sqlalchemy_exception, with_traceback=exc_info[2], from_=e
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/sqlalchemy/util/compat.py", line 178, in raise_
    raise exception
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1278, in _execute_context
    cursor, statement, parameters, context
  File "/home/sph/.virtualenvs/aiida_repo/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 593, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (psycopg2.errors.ObjectInUse) cannot ALTER TABLE "db_dbgroup" because it has pending trigger events

[SQL: ALTER TABLE db_dbgroup ALTER COLUMN extras SET NOT NULL]
(Background on this error at: http://sqlalche.me/e/13/e3q8)
```

Interestingly, both `bf591f31dd12` and `0edcdd5a30f0` alter the `db_dbgroup` table. The exception suggests that when the second migration comes around, the previous one has not completely finished or at least left unfinished trigger events on the table. This prompted me to try and perform the migrations one-by-one using the command line tool that ships with `aiida-core`:
```
python aiida/backends/sqlalchemy/manage.py upgrade bf591f31dd12
python aiida/backends/sqlalchemy/manage.py upgrade 0edcdd5a30f0
```
and lo and behold, it works.
So I think there is something wrong in the migration, either the way we defined them or a bug in alembic, where the first migration is not properly finished/flushed/whatever before the second one starts. No idea why though.

I tried splitting the revision `0edcdd5a30f0` into two revisions moving the line `op.alter_column('db_dbgroup', 'extras', nullable=False)` to the second revision, however, the exact same exception was thrown. It seems then that the problem has to do with all revisions happening in the same transaction. When we migrate manually one revision at a time, each revision is done in its own separate transaction. However, we set the flag `transaction_per_migration` to `True` when configuring the alembic environment, so that should enforce using a single transaction for each revision. (This actually emulates the behavior of the database migrations which is why we set it.)

sphuber:
Fixes #4590 

### SqlAlchemy: improve the alembic migration code

The code in `aiida.backends.sqlalchemy.manager` is used, among other
things, to control the database migrations. It can be used to get info
like the current revision of the database, current revision of the code
and to actually migrate the database. The migrations are performed
through `alembic`. This library has the somewhat weird approach of
loading the `env.py` file, which then actually calls the migrations.
This file somehow needs to get the configuration of the environment, for
example the connection to the database, and it does so by loading it
from the config that is imported. This config is actually built
dynamically in the `SqlBackendManager`.

The old implementation was correctly building up the configuration
object but was then calling the CLI commands of `alembic` to perform the
various operations, which was unnecessarily complex and inefficient. For
most operations, essentially everything but performing the actual
migrations, one can construct a `MigrationContext` directly, which
allows to inspect the state of the database, for example to get the
current database revision. For anything related to the revisions in the
code, e.g., to get the current head of the code, we can instantiate an
instance of the `ScriptDirectory` which allows one to inspect the
available revisions.

Through these changes, the `env.py` file is no longer loaded everytime,
even when it is not necessary, such as when simply retrieving the
current revision of the database.

### SqlAlchemy: fix bug in `Group` extras migration with revision `0edcddc585917

This migration added the JSONB column `extras` to the `DbGroup` model.
Since this needs to be a non-nullable column, but existing group rows
would have a null value, the migration was implemented by first adding
the column as nullable, after which all rows had their value set to the
default, before making the column non-nullable.

This approach was working when the revision was run by itself in an
isolated way, which is how the unit test is run, and so that passed.
However, as soon as it was run with multiple revisions in a sequence the
migration would fail with the message that the `db_dbgroup` table has
pending event triggers as soon as the final instruction to change the
nullability of the column was executed. The cause for this was never
found.

As an alternative, the migration is adapted to instead create the column
directly as non-nullable, but also provide a server default. This will
also avoid the exception of existing rows with a null value, just as the
original approach, but now when we remove the server default in a second
operation, we no longer hit the problem of the pending trigger events.

### DIFF
--- a/aiida/backends/sqlalchemy/manager.py
+++ b/aiida/backends/sqlalchemy/manager.py
@@ -7,16 +7,9 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-# pylint: disable=import-error,no-name-in-module
 """Utilities and configuration of the SqlAlchemy database schema."""
-
 import os
 import contextlib
-
-from alembic import command
-from alembic.config import Config
-from alembic.runtime.environment import EnvironmentContext
-from alembic.script import ScriptDirectory
 
 from sqlalchemy.orm.exc import NoResultFound
 
@@ -24,7 +17,6 @@ from aiida.backends.sqlalchemy import get_scoped_session
 from aiida.common import NotExistent
 from ..manager import BackendManager, SettingsManager, Setting
 
-ALEMBIC_FILENAME = 'alembic.ini'
 ALEMBIC_REL_PATH = 'migrations'
 
 # The database schema version required to perform schema reset for a given code schema generation
@@ -37,18 +29,44 @@ class SqlaBackendManager(BackendManager):
     @staticmethod
     @contextlib.contextmanager
     def alembic_config():
-        """Context manager to return an instance of an Alembic configuration with the current connection inserted.
+        """Context manager to return an instance of an Alembic configuration.
 
-        :return: instance of :py:class:`alembic.config.Config`
+        The current database connection is added in the `attributes` property, through which it can then also be
+        retrieved, also in the `env.py` file, which is run when the database is migrated.
         """
         from . import ENGINE
+        from alembic.config import Config
 
         with ENGINE.begin() as connection:
             dir_path = os.path.dirname(os.path.realpath(__file__))
-            config = Config(os.path.join(dir_path, ALEMBIC_FILENAME))
+            config = Config()
             config.set_main_option('script_location', os.path.join(dir_path, ALEMBIC_REL_PATH))
             config.attributes['connection'] = connection  # pylint: disable=unsupported-assignment-operation
             yield config
+
+    @contextlib.contextmanager
+    def alembic_script(self):
+        """Context manager to return an instance of an Alembic `ScriptDirectory`."""
+        from alembic.script import ScriptDirectory
+
+        with self.alembic_config() as config:
+            yield ScriptDirectory.from_config(config)
+
+    @contextlib.contextmanager
+    def migration_context(self):
+        """Context manager to return an instance of an Alembic migration context.
+
+        This migration context will have been configured with the current database connection, which allows this context
+        to be used to inspect the contents of the database, such as the current revision.
+        """
+        from alembic.runtime.environment import EnvironmentContext
+        from alembic.script import ScriptDirectory
+
+        with self.alembic_config() as config:
+            script = ScriptDirectory.from_config(config)
+            with EnvironmentContext(config, script) as context:
+                context.configure(context.config.attributes['connection'])
+                yield context.get_context()
 
     def get_settings_manager(self):
         """Return an instance of the `SettingsManager`.
@@ -80,29 +98,13 @@ class SqlaBackendManager(BackendManager):
 
         :return: boolean, True if the database schema version is ahead of the code schema version.
         """
-        from alembic.util import CommandError
-
-        # In the case of SqlAlchemy, if the database revision is ahead of the code, that means the revision stored in
-        # the database is not even present in the code base. Therefore we cannot locate it in the revision graph and
-        # determine whether it is ahead of the current code head. We simply try to get the revision and if it does not
-        # exist it means it is ahead.
-        with self.alembic_config() as config:
-            try:
-                script = ScriptDirectory.from_config(config)
-                script.get_revision(self.get_schema_version_database())
-            except CommandError:
-                # Raised when the revision of the database is not present in the revision graph.
-                return True
-            else:
-                return False
+        with self.alembic_script() as script:
+            return self.get_schema_version_database() not in [entry.revision for entry in script.walk_revisions()]
 
     def get_schema_version_code(self):
         """Return the code schema version."""
-        with self.alembic_config() as config:
-            script = ScriptDirectory.from_config(config)
-            schema_version_code = script.get_current_head()
-
-        return schema_version_code
+        with self.alembic_script() as script:
+            return script.get_current_head()
 
     def get_schema_version_reset(self, schema_generation_code):
         """Return schema version the database should have to be able to automatically reset to code schema generation.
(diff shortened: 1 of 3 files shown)


sphuber:
Thinking about this, I think we need to release this as `v1.5.1` and so not merge into `develop`. @giovannipizzi it would still be good to review and then I can start to prepare the `release/1.5.1` branch

giovannipizzi (approved):
This seems all very good to me - even if of course something might always slip, alembic is a big beast :-D 

Anyway, I think there might be a simple explanation for the other issue we were seeing - I think PSQL has to build a table of constraints when you put them, and to avoid blocking it will do it "in the background" while you continue operation.
Of course, if you have to use that contraint check, it will fail with the error you see.
There are ways to tell PSQL "Please create all pending constraint tables and then continue", and one could put this between the transactions, but probably this is a better approach.

PR #4605, opened 2020-12-01
Config: do not overwrite when loaded and not migrated

### LINKED ISSUE #4606: Config file gets written each time it is read in `Config.from_file` even when not migrated

sphuber:
Fixes #4606 

The `Config.from_file` classmethod used to load the `config.json` file
into memory was always writing the contents from memory to disk if it
was read, even if the content was not migrated. This is unnecessary and
so the write is now only performed if the config was migrated.

### DIFF
--- a/aiida/manage/configuration/config.py
+++ b/aiida/manage/configuration/config.py
@@ -51,14 +51,19 @@ class Config:  # pylint: disable=too-many-public-methods
             config = Config(filepath, check_and_migrate_config({}))
             config.store()
         else:
+            migrated = False
+
             # If the configuration file needs to be migrated first create a specific backup so it can easily be reverted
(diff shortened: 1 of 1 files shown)


ltalirz (approved):
thanks @sphuber !

sphuber:
@ltalirz  Since we are going to release a v1.5.1, I am merging this into the `release/1.5.1` branch instead. Once released, this will also be merged into `develop`, which should be somewhere next week

ltalirz (approved):

ltalirz:
@sphuber By the way, I wonder whether we should add a test to prevent a regression of this?

Not sure it's necessary

sphuber:
> Not sure it's necessary

I think it is fine not to include one now

Issue #4604, opened 2020-12-01
AiiDA configuration file sometimes gets erased.


yakutovicha:
### Describe the bug

The aiida configuration file sometimes gets erased when the machine where AiiDA is running is overloaded. More specifically, if one calls `verdi` and then, while it is running, stops it using the `Ctrl+C` there is a high risk of losing the config file.

Additional note: every time the `verdi` command is called, the config file timestamp changes, which suggests that the file gets overwritten.

### Expected behaviour

The `verdi` command should not change the config file unless it is `verdi config` or a version update.

ltalirz:
The problem is this line, which results in the configuration file being overwritten whenever it is successfully read using  `.from_file`:
https://github.com/aiidateam/aiida-core/blob/02248cf3686a0ab89faf1625e0da24d9e33d8cde/aiida/manage/configuration/config.py#L61


This change was introduced in https://github.com/aiidateam/aiida-core/commit/2f9224a4c8a1637a6de4eed136505024a263b976 (February 2020).

@greschd asked whether the addition of `config.store()` was necessary for tests to pass https://github.com/aiidateam/aiida-core/pull/3797#pullrequestreview-363336212
@sphuber can you comment?

If it was necessary for tests to pass, perhaps the tests need to be amended?

greschd:
Might be related to #4523, insofar as that can lead to "inconsistent" config states - and maybe that leads to storing it over and over again (just a hunch, could be wrong).

chrisjsewell:
It looks like it only needs to be stored if `config_needs_migrating(config) is True`?

sphuber:
> If it was necessary for tests to pass, perhaps the tests need to be amended?
> In any case, I think this line storing the config should be removed again.

No the reason is that the line before it migrates the config if necessary, but only in memory. The file on disk will remain in its old form. The line was added to persist the changes to disk. Otherwise the migration would be performed each time the config file was read. What we should do instead is to have the writing to disk only happen _if_ an actual migration was performed.

I am not convinced however that this is the source of the bug described by @yakutovicha . The problem you describe is merely the explanation for why the config file is written each time it is read. This doesn't necessarily explain how it can be accidentally deleted.

chrisjsewell:
I thinks its quite probably the source of the bug. As already mentioned, it may be something in `shutil.copy` whereby if it is interrupted it leads to the loss of the file.

Additionally, this

[...]

PR #1666, opened 2018-06-19
Fix interactive and conditional options in  click

giovannipizzi:
to comply with the expected behaviour, now also tested.

PR #4607, opened 2020-12-02
Config: make writing to disk as atomic as possible

sphuber:
Potentially addresses #4604 

The original implementation of the `Config.store` method, that writes
the configuration from memory to disk, was using `shutil.copy` which is
not an atomic operation. It is therefore possible that if the copy
operation is interrupted the configuration is not written entirely and
is left either empty or in a corrupted state.

Making this an atomic operation in 100% of the cases is not trivial and
even on the application level there are no guarantees as even on the
file system level operations that should be atomic can be interrupted
under very rare circumstances.

However, we can make a significant improvement by replacing the use of
`shutil.copy` with `os.rename`, which because it is a file move
operation should be atomic and therefore less prone to being
interrupted. The new algorithm for storing the config is as follows:

 1. Create temporary file in same directory as target file
 2. Write content of in memory config to the temporary file and flush
 3. Use `os.rename` to move the temporary file to target destination

This should prevent most common cases that could corrupt the existing
configuration file.

Finally, in `Config.from_method` now only `FileNotFoundError` is caught
when trying to read the config file from disk. Originally both `IOError`
and `OSError` were being caught, which can also be raised if an existing
file could not be read. Simply overwriting in this case with a default
config is not desirable and so in this case we simply let the exception
bubble up.

### DIFF
--- a/aiida/manage/configuration/config.py
+++ b/aiida/manage/configuration/config.py
@@ -8,9 +8,9 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Module that defines the configuration file of an AiiDA instance and functions to create and load it."""
-
 import os
 import shutil
+import tempfile
 
 from aiida.common import json
 
@@ -47,7 +47,7 @@ class Config:  # pylint: disable=too-many-public-methods
         try:
             with open(filepath, 'r', encoding='utf8') as handle:
                 config = json.load(handle)
-        except (IOError, OSError):
+        except FileNotFoundError:
             config = Config(filepath, check_and_migrate_config({}))
             config.store()
         else:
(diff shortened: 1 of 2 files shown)


aiida/manage/configuration/config.py
```diff
@@ -378,38 +378,52 @@ def store(self):
...
-            with open(self.filepath, 'wb') as handle:
-                self._write(handle)
+            self._atomic_write()
             return self
 
         # Otherwise, we write the content to a temporary file and compare its md5 checksum with the current config on
         # disk. When the checksums differ, we first create a backup and only then overwrite the existing file.
         with tempfile.NamedTemporaryFile() as handle:
-            self._write(handle)
+            json.dump(self.dictionary, handle, indent=DEFAULT_CONFIG_INDENT_SIZE)
```
ltalirz, line 392: is there a reason not to use `self._write` here?

tests/manage/configuration/test_config.py
```diff
@@ -185,20 +184,32 @@ def compare_config_in_memory_and_on_disk(self, config, filepath):
...
+        """Test that ``Config.from_file`` does not overwrite if the content was not migrated."""
+        from time import sleep
+
+        # Construct the ``Config`` instance and write it to disk
+        Config(self.config_filepath, self.config_dictionary).store()
+
+        timestamp = os.path.getmtime(self.config_filepath)
+        Config.from_file(self.config_filepath)
+
+        # Sleep a second, because for some operating systems the time resolution is of the order of a second
```
ltalirz, line 208: Wow, which OS has 1s time resolution?
It's interesting that this can work with processors that have clock cycles below ns ;-)

As far as I'm aware, tests should run "in the blink of an eye", i.e. 1s is already too long.. 0.1s would already be better..

ltalirz (commented):
thanks @sphuber , just some small comments

aiida/manage/configuration/config.py
sphuber: `self._write` is no longer there, it has been completely replaced by `_atomic_write`

ltalirz: ah!

tests/manage/configuration/test_config.py
sphuber: [l'histoire se répete](https://github.com/aiidateam/aiida-core/pull/1666#pullrequestreview-130247279)
TLDR: blame MacOS :)

ltalirz: Not quite sure what type of time resolution @giovannipizzi was referring to, but on my MacOS it is quite substantially better

```ipython
In [1]: from os.path import getmtime

In [5]: from time import sleep

In [7]: from pathlib import Path

In [8]: a = Path('a')

In [9]: for i in range(10):
   ...:     sleep(0.1)
   ...:     a.touch()
   ...:     print(getmtime('a'))
   ...:
1606909606.250035
1606909606.350413
1606909606.452756
1606909606.556789
1606909606.658411
1606909606.758832
1606909606.864015
1606909606.968789
1606909607.069461
1606909607.171968

In [10]: for i in range(10):
    ...:     sleep(0.01)
    ...:     a.touch()
    ...:     print(getmtime('a'))
    ...:
1606909614.754383
1606909614.765453
1606909614.775664
1606909614.785863
1606909614.796015
1606909614.806161
1606909614.817056
1606909614.827201
1606909614.837322
1606909614.848603

In [11]: for i in range(10):
    ...:     sleep(0.001)
    ...:     a.touch()
    ...:     print(getmtime('a'))
    ...:
1606909621.740552
1606909621.741794
1606909621.74325
1606909621.744598
1606909621.745944
1606909621.747289
1606909621.748634
1606909621.749978
1606909621.751066
1606909621.752403
In [12]: for i in range(10):
    ...:     sleep(0.0001)
    ...:     a.touch()
    ...:     print(getmtime('a'))
    ...:
1606909728.212414
1606909728.212609
1606909728.212802
1606909728.213136
1606909728.213311
1606909728.213494
1606909728.21371
1606909728.213929
1606909728.21422
1606909728.214399

```

sphuber: If this problem has been addressed in at least some of the MacOS versions, maybe it would be best to remove these sleeps. Note that this is not the only place where I am adding it. It is used for a lot of the `verdi` tests that need to use the interactive writer. I would leave that for a separate PR though. I would be fine to remove it since the only problem it tackles is that certain unit tests will fail on certain MacOS systems. Probably not reason enough to slow down many of these tests all the time.

ltalirz (approved):
ok, in that case good to merge from my side.

let's follow up on the timing issue though

sphuber:
> ok, in that case good to merge from my side.

Do you think this fixes issue #4604 and so I close it? Or do we leave it open?

ltalirz:
Let's close it and reopen in case it reoccurs

tests/manage/configuration/test_config.py
giovannipizzi: So: 
- I have fixed the (testing) issue already upstream in click some time ago - this will [appear in v8.0](https://github.com/pallets/click/commit/97fde8bf50058c19a77d4de46b9c54b1731b1750) (it's now in v8.0.0a1 according to GitHub)
- According to my old comments in that commit, the precision is 2 seconds (!!) on FAT32!
- According to my commit, at the time I was using Mac OS X 10.12. In 10.13 Apple switched to [APFS](https://en.wikipedia.org/wiki/Apple_File_System) that I guess fixed this and that must be why Leo does not see it anymore (BTW, it [introduced a much more serious bug IMO](https://github.com/giovannipizzi/concurrent-delete-race-problem) with atomicity of file replacements, that I unveiled when hard-testing disk-objectstore! Still no reply from the developers, though). Since this is only for testing so it matters only to us developers, and I think nowadays probably everybody is developing on at least 10.13, probably we could do a PR and clean up these sleeps.

giovannipizzi: Just to bring references: in the official [Microsoft docs](https://docs.microsoft.com/en-us/windows/win32/sysinfo/file-times?redirectedfrom=MSDN) they say

> [...] the resolution of create time on FAT is 10 milliseconds, while write time has a resolution of 2 seconds and access time has a resolution of 1 day, so it is really the access date. The NTFS file system delays updates to the last access time for a file by up to 1 hour after the last access.

And then you guys complain about the 1s resolution on older Macs? :-D

ltalirz: Thanks @giovannipizzi for the info, that's good to know!

Does this mean that since we are now always creating a temporary file and renaming it, we actually have a time resolution of 10ms on FAT32 as well?

In general, one could also think of an alternative way of testing this: as long as we know where in our code the config file can be written (e.g. the atomic_write function), we could use `pytest-mock` to mock this function and just check whether it is called. Then we don't need to rely on timing updates at the FS level.

PR #4608, opened 2020-12-02
Daemon: replace deprecated classmethods of `asyncio.Task` in shutdown

### LINKED ISSUE #4594: AttributeError: type object 'Task' has no attribute 'all_tasks' for python3.9

### Describe the bug
OK, here is a issue after I start using asyncio version with PY3.9. The  `asyncio.Task.all_tasks()` and `asyncio.Task.current_task()` is deprecated in PY3.9 but needed for PY3.6. I think before we dropping py3.6 we need to add a python version check (`sys.version_info`) to support both of them. 

OR any other workaround exist?

```
File "/data/CONDA_ENV/aiida-testing/lib/python3.9/asyncio/tasks.py", line 256, in __step                              
    result = coro.send(None)                                                                                              
File "/home/unkcpz/pyProject/aiida_core/aiida/engine/daemon/runner.py", line 39, in shutdown                          
    tasks = [t for t in asyncio.Task.all_tasks() if t is not asyncio.Task.current_task()]
AttributeError: type object 'Task' has no attribute 'all_tasks'
```

### Steps to reproduce

Just stop the daemon.

### Your environment

- Operating system Linux:
- Python version 3.9:
- aiida-core version [develop-branch with asyncio]:

sphuber:
Fixes #4594 

The `shutdown` function, that was attached to the loop of the daemon
runner in `aiida.engine.daemon.runner.start_daemon`, was calling the
classmethods `current_task` and `all_tasks` of `asyncio.Task` which have
been deprecated in Python 3.7 and are removed in Python 3.9. This would
prevent the daemon runners from being shutdown in Python 3.9. The
methods have been replaced with top level functions that can be imported
directl from `asyncio`.

This was not noticed in the tests because in the tests the daemon is
stopped but it is not checked whether this happens successfully. Anyway,
the error would only show up in the daemon log. To test the shutdown
method, it has been made into a standalone coroutine and renamed to
`shutdown_runner`.

Since the `shutdown_runner` is a coroutine, the unit test that calls it
also has to be one and therefore we need `pytest-asyncio` as a
dependency. The `event_loop` fixture, that is provided by this library,
is overrided such that it provides the event loop of the `Manager`,
since in AiiDA only ever this single reentrant loop should be used.

Note that the current CI tests run against Python 3.6 and Python 3.9 and
so will still not catch this problem, however, the `test-install`
workflow _does_ run against Python 3.9. I have opted not to change the
continuous integrations to run against Python 3.9 instead of 3.8, since
they take more than twice the time. Supposedly this is because certain
dependencies have to be built and compiled from scratch when the
testing environment is started.

### DIFF
--- a/requirements/requirements-py-3.9.txt
+++ b/requirements/requirements-py-3.9.txt
@@ -103,6 +103,7 @@ pytest-benchmark==3.2.3
 pytest-cov==2.10.1
 pytest-rerunfailures==9.1.1
 pytest-timeout==1.4.2
+pytest-asyncio==0.12.0
 python-dateutil==2.8.1
 python-editor==1.0.4
 python-memcached==1.59
--- a/setup.json
+++ b/setup.json
@@ -106,6 +106,7 @@
             "pg8000~=1.13",
             "pgtest~=1.3,>=1.3.1",
             "pytest~=6.0",
+            "pytest-asyncio~=0.12",
             "pytest-timeout~=1.3",
             "pytest-cov~=2.7",
             "pytest-rerunfailures~=9.1,>=9.1.1",
--- a/aiida/engine/daemon/runner.py
+++ b/aiida/engine/daemon/runner.py
@@ -19,6 +19,28 @@ from aiida.manage.manager import get_manager
 LOGGER = logging.getLogger(__name__)
 
 
+async def shutdown_runner(runner):
+    """Cleanup tasks tied to the service's shutdown."""
+    LOGGER.info('Received signal to shut down the daemon runner')
+
+    try:
+        from asyncio import all_tasks
+        from asyncio import current_task
+    except ImportError:
+        # Necessary for Python 3.6 as `asyncio.all_tasks` and `asyncio.current_task` were introduced in Python 3.7. The
+        # Standalone functions should be used as the classmethods are removed as of Python 3.9.
+        all_tasks = asyncio.Task.all_tasks
+        current_task = asyncio.Task.current_task
+
+    tasks = [task for task in all_tasks() if task is not current_task()]
+
+    for task in tasks:
+        task.cancel()
+
+    await asyncio.gather(*tasks, return_exceptions=True)
+    runner.close()
+
+
 def start_daemon():
     """Start a daemon runner for the currently configured profile."""
     daemon_client = get_daemon_client()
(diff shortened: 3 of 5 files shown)


sphuber:
As mentioned in the commit, I decided to not run the `continuous-integration` workflow against Python 3.9 instead of 3.8 because it takes super long to run (more than 30 mins). I did run [a build on my fork](https://github.com/sphuber/aiida-core/runs/1487686705) successfully. I reasoned that this is ok since the `test-install` workflow _does_ run against 3.9 @csadorf do you agree that this is ok or do we have to start testing against 3.9 and find a way to speed up the build.  @chrisjsewell do you have an idea why creating the Python environment and installing all packages takes longer?

aiida/engine/runners.py
```diff
@@ -150,7 +150,10 @@ def close(self):
         """Close the runner by stopping the loop."""
         assert not self._closed
         self.stop()
-        reset_event_loop_policy()
+        try:
+            reset_event_loop_policy()
+        except AttributeError:
```
unkcpz, line 155: I see no reason to pass this exception, shouldn't let it raised? And why the `AttributeError` capture here?

unkcpz (changes requested):
@sphuber thanks! looks all good to me. Just a minor question I didn't quite understand.

aiida/engine/runners.py
sphuber: Good point. I had to add it there at some point to get the test to pass but the failure was because I wasn't overriding the `event_loop` fixture of `pytest-asyncio`. Now it is no longer needed.

unkcpz (approved):
Thanks @sphuber. What surprises me is the test coverage increase such a lot by this change. How on earth that happened?

requirements/requirements-py-3.9.txt
```diff
@@ -103,6 +103,7 @@ pytest-benchmark==3.2.3
 pytest-cov==2.10.1
 pytest-rerunfailures==9.1.1
 pytest-timeout==1.4.2
+pytest-asyncio==0.12.0
```
csadorf, line 106: Is this dependency specific to 3.9? If yes, then this should be indicated via markers in setup.json. Otherwise, I'd suggest we add this to all requirements files.

setup.json
```diff
@@ -106,6 +106,7 @@
             "pg8000~=1.13",
             "pgtest~=1.3,>=1.3.1",
             "pytest~=6.0",
+            "pytest-asyncio~=0.12",
```
csadorf, line 109: See comment above.

csadorf (commented):
I think it is perfectly fine to run our CI tests against Python 3.8 for the time being. I expect that we will transition to 3.9 once 3.6 is deprecated. The test-install (nightly) workflow tests against all supported Python version and is expected to run much longer.

However, I do have a question with respect to the dependency specification here. It is very possible that I misunderstood something.

requirements/requirements-py-3.9.txt
sphuber: This _is_ present in all requirement files. They were added I imagine because `pytest-ascynio` is a dependency of one of our dependencies that installs it for testing. Since now we add it as our own dependency it also ends up in the Python 3.9 requirements. Not 100% sure why it wasn't there yet.

csadorf: Ah, sorry, I didn't check the other files.

csadorf (approved):
Approved w.r.t. dependencies.

PR #4613, opened 2020-12-03
CLI: add the `verdi database version` command

sphuber:
Fixes #4613 

This shows the schema generation and version of the database of the
given profile.

### DIFF
--- a/aiida/manage/manager.py
+++ b/aiida/manage/manager.py
@@ -86,8 +86,9 @@ class Manager:
 
         # Do NOT reload the backend environment if already loaded, simply reload the backend instance after
         if configuration.BACKEND_UUID is None:
-            manager = self.get_backend_manager()
-            manager.load_backend_environment(profile, validate_schema=schema_check)
+            from aiida.backends import get_backend_manager
+            backend_manager = get_backend_manager(self.get_profile().database_backend)
+            backend_manager.load_backend_environment(profile, validate_schema=schema_check)
             configuration.BACKEND_UUID = profile.uuid
 
         backend_type = profile.database_backend
@@ -123,8 +124,10 @@ class Manager:
         :return: the database backend manager
         :rtype: :class:`aiida.backend.manager.BackendManager`
         """
+        from aiida.backends import get_backend_manager
+
         if self._backend_manager is None:
-            from aiida.backends import get_backend_manager
+            self._load_backend()
             self._backend_manager = get_backend_manager(self.get_profile().database_backend)
 
         return self._backend_manager
--- a/aiida/cmdline/commands/cmd_database.py
+++ b/aiida/cmdline/commands/cmd_database.py
@@ -23,6 +23,22 @@ def verdi_database():
     """Inspect and manage the database."""
 
 
+@verdi_database.command('version')
+def database_version():
+    """Show the version of the database.
+
+    The database version is defined by the tuple of the schema generation and schema revision.
+    """
+    from aiida.manage.manager import get_manager
+
+    backend_manager = get_manager().get_backend_manager()
+
+    echo.echo('Generation: ', bold=True, nl=False)
+    echo.echo(backend_manager.get_schema_generation_database())
+    echo.echo('Revision:   ', bold=True, nl=False)
+    echo.echo(backend_manager.get_schema_version_database())
+
+
 @verdi_database.command('migrate')
 @options.FORCE()
 def database_migrate(force):
--- a/docs/source/reference/command_line.rst
+++ b/docs/source/reference/command_line.rst
@@ -199,6 +199,7 @@ Below is a list with all available subcommands.
     Commands:
       integrity  Check the integrity of the database and fix potential issues.
       migrate    Migrate the database to the latest schema version.
+      version    Show the version of the database.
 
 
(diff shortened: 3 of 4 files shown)


aiida/manage/manager.py
```diff
@@ -86,8 +86,9 @@ def _load_backend(self, schema_check=True):
 
         # Do NOT reload the backend environment if already loaded, simply reload the backend instance after
         if configuration.BACKEND_UUID is None:
-            manager = self.get_backend_manager()
-            manager.load_backend_environment(profile, validate_schema=schema_check)
+            from aiida.backends import get_backend_manager
+            backend_manager = get_backend_manager(self.get_profile().database_backend)
```
ltalirz, line 90: this is safer?

aiida/manage/manager.py
```diff
@@ -123,8 +124,10 @@ def get_backend_manager(self):
         :return: the database backend manager
         :rtype: :class:`aiida.backend.manager.BackendManager`
         """
+        from aiida.backends import get_backend_manager
```
ltalirz, line 127: not quite following the logic here - in your other change you were placing the import inside the if clause...

aiida/manage/manager.py
```diff
@@ -123,8 +124,10 @@ def get_backend_manager(self):
...
+
         if self._backend_manager is None:
-            from aiida.backends import get_backend_manager
+            self._load_backend()
```
ltalirz, line 130: `get_backend_manager` should load the backend?

isn't that a significant change?
should this be in this PR?

I probably just miss the reasoning for these changes, could you please explain?

ltalirz (commented):
Thanks @sphuber !

besides the comments on the code: since this may be the first time the schema "generation" makes it into user space, could you perhaps add something about this to the documentation?

I.e. to explain why we need to have both a schema generation and a schema version; when do you need one or the other etc.

aiida/manage/manager.py
sphuber: All the changes in this file can be explained as follows, and they should answer all three of your questions. The `get_backend_manager` was so far _not_ loading the backend, although that really doesn't make any sense. It is providing access to data from the database, but to do so the backend should be loaded, otherwise a connection isn't possible. This problem went unnoticed, because the `BackendManager` is only used in `aiida.engine.utils.set_process_state_change_timestamp`. By the time this gets used, the database backend will already have been loaded through another code path.

For the changes in this PR, however, I needed to get _just_ the backend manager and have the backend loaded, which wasn't the case. I couldn't simply call `_load_backend()` in the `get_backend_manager` because this would lead to infinite recursion as `_load_backend()` also calls `get_backend_manager`. Therefore I had to refactor `_load_backend` to not call the former and directly fetch it through `aiida.backends.

aiida/manage/manager.py
sphuber: place of the import wasn't done intentional, could have been in either place and is not really that important.

aiida/manage/manager.py
sphuber: no, was simply necessary to prevent infinite recursion (see other comment)

sphuber:
Thanks for the review @ltalirz 

> besides the comments on the code

I have addressed those in the code. The explanation I give there should really have been in the commit. I will add it there.

> since this may be the first time the schema "generation" makes it into user space, could you perhaps add something about this to > the documentation? I.e. to explain why we need to have both a schema generation and a schema version; when do you need one or the other etc.

Would be fine with adding this, but for now I think this is mostly a command for developers or for helping with debugging. Would it be ok to add this (with a lot more detailed documentation) when we do the database schema reset? We would then anyway have to write quite some documentation, both for users and developers and so it make more sense to write this then,

aiida/manage/manager.py
ltalirz: Thanks a lot for the clarification, very helpful!

One last question: `_load_backend()` now uses a backend manager that is then thrown away.
If these two things indeed have to go together, why not let `_load_backend()` also set `self._backend_manager`?

The current solution seems to duplicate the creation of the backend manager, once in `_load_backend()` and once in `get_backend_manager()`.

sphuber: I left it in because I wasn't sure that there wasn't a code path (unlikely, but still) where the backend manager instance would _not_ be constructed in `_load_backend` because it is in the conditional `if configuration.BACKEND_UUID is None:`. If that somehow is `False` the calling `_load_backend` will not create an instance of the backend manager. So calling `get_backend_manager` will return `None`. I have looked more into the code to see where the design mistake is, but I am not sure yet how to improve the flow. Most of this code that makes it complicated is to deal with the possibility of unloading/reloading different profiles, which is currently anyway not supported. Maybe it is best to leave like this for now, but if you prefer I can assign the backend manager in `_load_backend` to `self._backend_manager` and remove the construction from `get_backend_manager`.

ltalirz: Ok, please add a comment in the corresponding issue so that we don't forget

ltalirz (approved):

aiida/manage/manager.py
sphuber: [Done!](https://github.com/aiidateam/aiida-core/issues/3011)

PR #4615, opened 2020-12-03
CI: manually install `numpy==1.19.4` to prevent release candidate

### LINKED ISSUE #4614: CI breaking because `pymatgen` is installing release candidate of `numpy`
It is that time of the year again and our [CI builds on Python 3.6 are failing](https://github.com/aiidateam/aiida-core/pull/4613/checks?check_run_id=1495195639) because `numpy` released a new candidate version that will remove Python 3.6 support. Since `pymatgen` still hasn't addressed their build problem, it will install the latest release candidate and so we are fucked....

sphuber:
Fixes #4614 

This is the usual problem that occurs when `numpy` releases a candidate
release. Even though we specify limits to not install this, `pymatgen`
installs it as a dependency in their build process where limits are
ignored and so the latest release gets installed. In this case
`numpy==1.20.0rc1` is getting installed and that dropped support for
Python 3.6 and so our builds on that version brick.

The workaround is to manually install the compatible version, in this
case `numpy==1.19.4` before installing `aiida-core`. This way `pymatgen`
will simply use that in their build process.

### DIFF
--- a/.github/workflows/ci-code.yml
+++ b/.github/workflows/ci-code.yml
@@ -92,6 +92,7 @@ jobs:
 
     - name: Install aiida-core
       run: |
+        pip install numpy==1.19.4
         pip install --use-feature=2020-resolver -r requirements/requirements-py-${{ matrix.python-version }}.txt
         pip install --use-feature=2020-resolver --no-deps -e .
         reentry scan
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -155,6 +155,12 @@ jobs:
         sudo apt install postgresql-10 graphviz
 
     - run: pip install --upgrade pip
+    # Work-around issue caused by pymatgen's setup process, which will install the latest
+    # numpy version (including release candidates) regardless of our actual specification:
+    - name: Install numpy [py36]
+      if: matrix.python-version == 3.6
+      run: |
+        pip install `grep 'numpy==' requirements/requirements-py-${{ matrix.python-version }}.txt`
 
     - name: Install aiida-core
       run: |


sphuber:
Thinking about this, I will actually have to base this of `release/1.5.1` because those tests will also be broken and we need to release that tomorrow or early next week.

chrisjsewell:
Does this relate to a particular open issue on pymatgen, so I can go and voice my displeasure lol

Also, should we add a note on this somewhere in the documentation; in the install or troubleshooting section?

sphuber:
> Does this relate to a particular open issue on pymatgen, so I can go and voice my displeasure lol

No current one, but we have dealt with this problem already at least 4 times. Each time `numpy` makes a new minor version release essentially. We have to introduce this explicit limit again to get everything running. At the time, we opened [this issue](https://github.com/materialsproject/pymatgen/issues/1520) which got rejected because it concerned `numpy` removing Python 2.7 support and the `pymatgen` developers didn't want to add fixes for this. It seems that this is not just a Python 2.7 problem but a generic problem with their build requirement definitions. Not sure if they would be amenable to considering addressing this time around. We could maybe open an issue again and ask at least.

> Also, should we add a note on this somewhere in the documentation; in the install or troubleshooting section?

Would be a good idea since it is happening often. At least by know I am able to recognize the problem almost instantly and add the workaround in place without wasting a lot of precious time, but we should document this. Still other developers should then still be aware of this phenomenon. In any case, given that it concerns `aiida-core` developers, it shouldn't go in the docs but the wiki if anything.

chrisjsewell:
On a related note @csadorf we can remove the 2020-resolver now:

> WARNING: --use-feature=2020-resolver no longer has any effect, since it is now the default dependency resolver in pip. This will become an error in pip 21.0.

chrisjsewell:
> In any case, given that it concerns aiida-core developers, it shouldn't go in the docs but the wiki if anything.

Will this not then affect people running `pip install aiida-core[atomic_tools]` then?

I was actually just trying to reproduce this locally, but just saw that its not an issue for Macs, because there is a pymatgen wheel (https://github.com/materialsproject/pymatgen/issues/1520#issue-463761473). Macs are clearly the best 😝

csadorf:
> > In any case, given that it concerns aiida-core developers, it shouldn't go in the docs but the wiki if anything.
> 
> Will this not then affect people running `pip install aiida-core[atomic_tools]` then?
> 

Yes, it does: https://github.com/aiidateam/aiida-core/runs/1496526194?check_suite_focus=true

It is quite frustrating that this is problem is arising yet again. Those are the actions I would recommend:

1. Create an issue on `pymatgen` outlining the problem and its community impact once more.
2. If we know a solution, create a pull request (or at least offer to create one) that resolves or mitigates the issue.
3. Add a warning to the installation documentation that warns about this issue, specifically related to "Python 3.6" and the `atomic_tools` extra.

For clarification, restricting our dependencies to versions of pymatgen that support the same Python versions that we support with markers does not solve the problem, correct?

sphuber:
> For clarification, restricting our dependencies to versions of pymatgen that support the same Python versions that we support with markers does not solve the problem, correct?

Indeed, this does not solve the problem, because the limitations we impose through the `setup.json` are not respected by `setuptools` which is what `pymatgen` runs to build itself. So when we ask to install `pymatgen` it will build taking the latest available version of `numpy` (including release candidates and pre-releases) _unless_ some version already is installed in the environment

.github/workflows/ci-code.yml
```diff
@@ -92,6 +92,7 @@ jobs:
 
     - name: Install aiida-core
       run: |
+        pip install numpy==1.19.4
```
csadorf, line 95: Let's at least ensure that the version from the requirements files is installed and add a comment as to why this work-around is necessary:
```suggestion
        # Work-around issue caused by pymatgen's setup process, which will install the latest 
        # numpy version (including release candidates) regardless of our actual specification:
        pip install `grep 'numpy==' requirements/requirements-py-${{ matrix.python-version }}.txt`
```

.github/workflows/test-install.yml
```diff
@@ -158,6 +158,7 @@ jobs:
 
     - name: Install aiida-core
       run: |
+        pip install numpy==1.19.4
```
csadorf, line 161: I would suggest to make this step explicitly conditional and add a comment as to why it is necessary:
```suggestion
    # Work-around issue caused by pymatgen's setup process, which will install the latest 
    # numpy version (including release candidates) regardless of our actual specification:
    - name: Install numpy [py36]
      if: matrix.python-version == 3.6
      run: |
        pip install `grep 'numpy==' requirements/requirements-py-${{ matrix.python-version }}.txt`

    - name: Install aiida-core
      run: |
```

csadorf (approved):
I've made some suggestions on how to make this work-around fix a bit less intrusive, but overall it is of course fine, at least for the ci-code workflow. We can't have this impede our dev work. I'm a bit more hesitant to accept this for the test-install workflow.

Either way, I want to be pragmatic about this, so the fix is fine, even without my proposed changes, albeit I would strongly recommend to at least accept the in-line comments.

.github/workflows/test-install.yml
sphuber: I see the idea of wanting to limit the scope to the bare minimum but I don't think this is the right thing to do in this case? With this scoped limitation, for all other versions, the release candidate of `numpy` will still get installed with all the potential breaking of code that comes with it. Maybe the install will be fine since the Python version is still compatible but other backwards incompatible changes or bugs may be introduced.

csadorf: I indeed overlooked this, good point.

sphuber: Ok, I will then merge this. Thanks for the review

sphuber:
For reference: the final solution that got merged here was not ideal. It hardcoded `numpy==1.19.4` whereas it should just get the version from the requirements file for the corresponding Python version, because that is guaranteed to be compatible. The same problem needed to be addressed on `support/1.4.x` in order for us to be able to release `v1.4.4`, so we applied the [same fix but in the correct general way](c063609f48c494418164c3e043568628bf66c3a8). This fix was also applied to `release/1.5.1` and replaced the commit merged in this PR. The new commit is 8058a97e483ea7a0b0d347f709a123f2f9b911f0.

mkhorton:
> Indeed, this does not solve the problem, because the limitations we impose through the setup.json are not respected by setuptools which is what pymatgen runs to build itself. So when we ask to install pymatgen it will build taking the latest available version of numpy (including release candidates and pre-releases) unless some version already is installed in the environment

Is this not expected behavior, and the responsibility of setuptools to make that determination?

> If we know a solution, create a pull request (or at least offer to create one) that resolves or mitigates the issue.

This would definitely be appreciated and in the spirit of open source. As mentioned more specifically in the #4614 issue and the [referenced issue in pymatgen](https://github.com/materialsproject/pymatgen/issues/2010), we have been unable to identify or reproduce the issue.

More broadly, for creating re-producible builds, I can recommend [poetry](https://python-poetry.org) since pip alone can often be very brittle. One thing that also comes to mind is that pip 20.3 has also made significant changes in how its dependency resolver works, so maybe that could be causing issues.

sphuber:
@mkhorton thanks a lot for following up on this, really appreciate it. I responded to the issue over at pymatgen with some additional information on how we encounter the problem and where I think the problem comes from, but it is still not entirely clear. I propose we continue the discussion there and see if we can nail down the origin of the problem and hopefully come up with a solution.

PR #4616, opened 2020-12-04
Fix wrong raise from dict nodes (v1.4.x)

ramirezfranciscof:
When accessing an attribute as a dict (`dictnode['key']`), the error
raised was an AttributeError instead of a KeyError, which is rather
unusual. This has been fixed and a test that checks the correct error
raise has been added.

### DIFF
--- a/aiida/orm/nodes/data/dict.py
+++ b/aiida/orm/nodes/data/dict.py
@@ -62,7 +62,10 @@ class Dict(Data):
             self.set_dict(dictionary)
 
     def __getitem__(self, key):
-        return self.get_attribute(key)
+        try:
+            return self.get_attribute(key)
+        except AttributeError as exc:
+            raise KeyError from exc
 
     def __setitem__(self, key, value):
         self.set_attribute(key, value)
--- a/tests/orm/data/test_dict.py
+++ b/tests/orm/data/test_dict.py
@@ -50,3 +50,15 @@ class TestDict(AiidaTestCase):
         self.assertEqual(self.node['value'], 2)
         self.node.dict.value = 3
         self.assertEqual(self.node['value'], 3)
+
+    def test_correct_raises(self):
+        """Test that the methods for accessing the item raise the correct error.
+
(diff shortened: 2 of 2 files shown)


ramirezfranciscof:
@sphuber this can be merged in without the [CI fix](https://github.com/aiidateam/aiida-core/pull/4615) right?

sphuber:
> @sphuber this can be merged in without the [CI fix](https://github.com/aiidateam/aiida-core/pull/4615) right?

Prefer not to. I am about to merge the fix into `release/1.5.1` so then you can rebase it and then I will merge it. That should then also be the last item for `v1.5.1` so we could do the release. Could you prepare and do it?

ramirezfranciscof:
So, I already rebased the one that goes into `release/1.5.1` ([here](https://github.com/aiidateam/aiida-core/pull/4577))but this is the one that goes into `1.4.4`, I don't need to do any rebasing or anything for this right?

sphuber:
> So, I already rebased the one that goes into `release/1.5.1` ([here](https://github.com/aiidateam/aiida-core/pull/4577))but this is the one that goes into `1.4.4`, I don't need to do any rebasing or anything for this right?

Ah damn, didn't realize this is another copy of the same PR. Well I think these tests would have failed as well and will start to fail once we merge it in `support/1.4.x` so we should actually put the same fix on that branch as well. I will do it now.

ramirezfranciscof:
@sphuber Python 3.5 seems to have a problem with numpy

```
ERROR: No matching distribution found for numpy==1.19.4
Error: Process completed with exit code 1.
```

I'm not sure if maybe we need to copy this fix to python 3.5 as well??

```
    # Work-around issue caused by pymatgen's setup process, which will install the latest
    # numpy version (including release candidates) regardless of our actual specification:
    - name: Install numpy [py36]
      if: matrix.python-version == 3.6
      run: |
        pip install `grep 'numpy==' requirements/requirements-py-${{ matrix.python-version }}.txt`
```

sphuber (approved):
Good to go! Thanks

PR #4618, opened 2020-12-04
Release `v1.5.1`

ramirezfranciscof:


ramirezfranciscof:
@sphuber  Now that I think about it maybe I need to point this to master instead, no? I got confused because tipically when you open the release branch that would be local/on your fork, but this one is on the main repo so I wasn't sure if I had to first PR to add the release commit there or directly PR to master.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## v1.4.3
+
+### Bug fixes
+- `Dict`: accessing an inexistent key now raises a `KeyError` (instead of `AttributeError`) [[#4577]](https://github.com/aiidateam/aiida-core/pull/4577)
+- Config: make writing to disk as atomic as possible [[#4607]](https://github.com/aiidateam/aiida-core/pull/4607)
+- Config: do not overwrite when loaded and not migrated [[#4605]](https://github.com/aiidateam/aiida-core/pull/4605)
+- SqlAlchemy: fix bug in `Group` extras migration with revision `0edcdd5a30f0`
+- SqlAlchemy: improve the alembic migration code
```
sphuber, line 10:
```suggestion
## v1.5.1

### Bug fixes
- Config: make writing to disk as atomic as possible [[#4607]](https://github.com/aiidateam/aiida-core/pull/4607)
- Config: do not overwrite when loaded and not migrated [[#4605]](https://github.com/aiidateam/aiida-core/pull/4605)
- SqlAlchemy: fix bug in `Group` extras migration with revision `0edcdd5a30f0` [[#4602]](https://github.com/aiidateam/aiida-core/pull/4602)

### Developers
- SqlAlchemy: improve the alembic migration code [[#4602]](https://github.com/aiidateam/aiida-core/pull/4602)
```

sphuber (changes requested):
Suggested changes for the changelog and also here you forgot to change the version numbers

sphuber:
> Now that I think about it maybe I need to point this to master instead, no? I got confused because tipically when you open the release branch that would be local/on your fork, but this one is on the main repo so I wasn't sure if I had to first PR to add the release commit there or directly PR to master.

There is two options:

* Create PR for release commit on release branch (as you have done now) and once accepted open PR for `release` into `master`. At that point nothing has to be reviewed anymore
* Directly create the release commit on release branch and directly open release into master PR. At that point, the reviewer of course has to still review the last commit to make sure it contains everything.

Both work and it doesn't really matter where the release branch is, with the exception that in the second option you have to commit locally and push, which you may not have the right to do on a branch directly on `aiidateam`, but if you're admin that is no problem.

CHANGELOG.md
ramirezfranciscof: One question though: why did you remove the `Dict` entry? There is no previous `v1.4.4` in this log where this got fix so it should appear here, no?

sphuber: I removed it because it will be released with `1.4.4` and we "forward" it to `1.5.1` to also be included there. We should simply copy the changelog section of `v1.4.4` (once we approved and merged the final version) also in this changelog (since we won't be merging `support/1.4.x` into `develop`.

One thing that I am thinking about now, is that with that changelog, it may seem that the bug is not present in `1.5.0` wherein actuality it is. Hmmmm. Hadn't thought about that yet. Maybe then it _is_ better to keep the same exact fix also in the `1.5.1` changelog. Then again, having the same line twice might not necessarily make obvious that the bug is present in 1.5.0. I am not sure there is a perfect solution, so I would still be tempted to _not_ include it again in the 1.5.1 changelog if it will already be in `1.4.4`.

ramirezfranciscof: But I think the problem arises from including `v1.4.4` info in the changelog of a branch that doesn't have `v1.4.4` in its history. Why do you think we need to include it anyways?

sphuber: Because I think that, even though the explicit changes in the code of explicit support patch versions might not end up in `master`, I think it _is_ important that the corresponding change log entries _are_ available from `master`. This is because anyone usually always inspects the changelog _from_ the `master` branch. If we don't include it, people would even think `v1.4.4` doesn't even exist when looking at the changelog

sphuber (approved):
Thanks @ramirezfranciscof

PR #4625, opened 2020-12-09
NodeLinkManager support dot separated namespace attributes retrieving

### LINKED ISSUE #3978: Add automated namespace support to `inputs` and `outputs` node neighbor manager
The `inputs` and `outputs` node neighbor managers support addressing process node inputs or outputs through attribute dereferencing by the link labels, e.g.:
```
node.outputs.some_output
```
Both the inputs and outputs can have arbitrarily nested namespaces, for example:
```
{
    'relax': {
        'structure': StructureData()
    }
}
```
However, since the links are flat, the resulting link label is `relax__structure`, where the namespace is indicated by the double underscore. To retrieve this output one should use `node.outputs.relax__structure`. However, it would be nice if the neighbor manager understands the double underscore and allows it to be retrieved through the syntax `node.outputs.relax.structure`.

unkcpz:
Fix #3978 

By iterate over the attributes of `inputs` or `outputs` `NodeLink`, create a new `AttrubuteDict` with both 'double underscore' attributes and parse it into a nested dict. 

For example, for attributes keys `['some__name__space__a', 'some__name__b']` with node value 4, 5
Creating an `AttributeDict`:

```
AttrubuteDict({
    'some': {
        'name': {
            'space': {
                'a': 4,
            },
            'b': 5
        }
    }
})
```
Then in `_get_keys` and `_get_node_by_link_label` , interact through the AttributeDict.
However, the code looks a bit of nasty and update the dict time by time seems don't have good performance.

### DIFF
--- a/tests/engine/test_process.py
+++ b/tests/engine/test_process.py
@@ -68,6 +68,8 @@ class TestProcessNamespace(AiidaTestCase):
         # Check that the link of the process node has the correct link name
         self.assertTrue('some__name__space__a' in proc.node.get_incoming().all_link_labels())
         self.assertEqual(proc.node.get_incoming().get_node_by_label('some__name__space__a'), 5)
+        self.assertEqual(proc.node.inputs.some.name.space.a, 5)
+        self.assertEqual(proc.node.inputs['some']['name']['space']['a'], 5)
 
 
 class ProcessStackTest(Process):
--- a/aiida/orm/utils/managers.py
+++ b/aiida/orm/utils/managers.py
@@ -8,11 +8,10 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """
-Contain utility classes for "managers", i.e., classes that act allow
+Contain utility classes for "managers", i.e., classes that allow
 to access members of other classes via TAB-completable attributes
 (e.g. the class underlying `calculation.inputs` to allow to do `calculation.inputs.<label>`).
 """
-
 from aiida.common import AttributeDict
 from aiida.common.links import LinkType
 from aiida.common.exceptions import NotExistent, NotExistentAttributeError, NotExistentKeyError
@@ -47,10 +46,9 @@ class NodeLinksManager:
         self._incoming = incoming
 
     def _construct_attribute_dict(self, incoming):
-        """
-        Construct an attribute dict to support nest namespace
-        :param incoming: if True, inspect incoming links, otherwise inspect
-            outgoing links
+        """Construct an attribute dict from all links of the node, recreating nested namespaces from flat link labels.
+
+        :param incoming: if True, inspect incoming links, otherwise inspect outgoing links.
         """
         if incoming:
             links = self._node.get_incoming(link_type=self._link_type)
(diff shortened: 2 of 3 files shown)


unkcpz:
The problem of this implementation is the new create namespace from double underscore link label can be overridden by the link label of the same name and v.v. For example:
```
a = orm.Data()
a.store()
b = orm.Data()
b.store()

# Create calc with inputs
calc = orm.CalculationNode()
calc.add_incoming(a, link_type=LinkType.INPUT_CALC, link_label='inp__a')
calc.add_incoming(b, link_type=LinkType.INPUT_CALC, link_label='inp')
calc.store()
```
The second incoming labeled as `inp` will override the first `inp` generated from `inp__a` and vice versa.

sphuber:
I guess it is true that for the example you give where you manually link up the nodes in this fashion, there is an overlap and clash of link labels. It is important to note that this could never happen through the "automated" creation in normal usage, because there the `inp` input cannot be a namespace _and_ a regular input at the same time.

unkcpz:
It is easy to add a check and raise the exception to the situation here. Or do you you think can just let it slide?

unkcpz:
new implementation inspired by https://github.com/aiidateam/aiida-core/pull/4863, using `node.nested()`.

tests/engine/test_process.py
```diff
@@ -65,7 +65,7 @@ def test_namespaced_process(self):
 
         # Check that the link of the process node has the correct link name
         self.assertTrue('some__name__space__a' in proc.node.get_incoming().all_link_labels())
-        self.assertEqual(proc.node.get_incoming().get_node_by_label('some__name__space__a'), 5)
+        self.assertEqual(proc.node.inputs.some.name.space.a, 5)
```
sphuber, line 68: I would keep the original test to make sure that also still works and then simply add the one you added. It would be good to also test `node.inputs['some']['name']['space']['a']

Actually, it would make more sense to add these tests to `tests/orm/utils/test_managers.py`. There are the explicit tests for the `LinkManager` class

sphuber (changes requested):
Thanks @unkcpz looks good. Would just add some more tests and in a different location

tests/engine/test_process.py
unkcpz: thanks @sphuber

> I would keep the original test to make sure that also still works and then simply add the one you added. It would be good to also test `node.inputs['some']['name']['space']['a']

added. And should be noticed(and more discuss) that retrieving by `node.inputs.some__name__space__a` not work anymore, which is available previously. This might break the compatibility of some plugins if they use the double underscore to access the node.

> Actually, it would make more sense to add these tests to tests/orm/utils/test_managers.py. There are the explicit tests for the LinkManager class

True, I was trying to add the tests there but it is hard to test in the original testcase. This time I add the test in a new testcase, I hope I did it right and clean.

sphuber: > And should be noticed(and more discuss) that retrieving by node.inputs.some__name__space__a not work anymore, which is available previously. This might break the compatibility of some plugins if they use the double underscore to access the node.

This is a very important point. Is it not possible to keep support for this? Because if so, we can merge this and release it in `v1.6.2` soon. Otherwise this will have to go in v2.0 but even that I am not sure, because we are trying to keep the breaking of plugins in v2.0 minimal. The best option would be to detect if we get a label with double underscores, print a deprecation warning saying to use just normal dots, and then we can remove the support in v2.0

espenfl: > > 
> 
> This is a very important point. Is it not possible to keep support for this? Because if so, we can merge this and release it in `v1.6.2` soon. Otherwise this will have to go in v2.0 but even that I am not sure, because we are trying to keep the breaking of plugins in v2.0 minimal. The best option would be to detect if we get a label with double underscores, print a deprecation warning saying to use just normal dots, and then we can remove the support in v2.0

Thanks a lot @unkcpz and @sphuber for fixing this and the `expose`. This will simplify our work on the plugin side significantly now that the workchain complexity grows. I agree that it would be ideal to keep both. At least until v2.0.

espenfl (dismissed):

sphuber:
@unkcpz I have pushed a commit that keeps support for the old way and just emits a deprecation warning. This way we can ship it with 1.6.2. Hope you don't mind.

@chrisjsewell and @ramirezfranciscof I have requested you as a reviewer now, since I have contributed to this PR now. I think this should now be fine to be merged without it breaking backwards compatibility. Still would be good to have a second pair of eyes.

unkcpz:
> I have pushed a commit that keeps support for the old way and just emits a deprecation warning. This way we can ship it with 1.6.2. Hope you don't mind.

@sphuber surely not mind, and I agree to keep it until 2,0. thanks for the fix!

ramirezfranciscof (approved):
All looks good with the code!

My only worry is that I'm not sure I agree with the general concept behind the change (that this is more intuitive). I mean, it might be more convenient in some cases, but I think "overloading" the notation may also make it less clear. Before the logic was always chains of `node.method`, so it was easier to understand at first sight the following (you already know what type of object comes after each dot):

```
node.outputs.other_node.outputs.yet_another__nested_linked_node.method
```

Now this might be `node.outputs.nested_link.to_node.method`, in which we add an extra possibility to what type of object the middle things could be. Actually, now that I think of it, `node.outputs.nested_link` would return a dict, right? **Should we also test for this (truncated nested links)?**

Anyways, I know I might be a bit late to the planning party so. It is also true that if any other method returns a dict of nodes, we could have a similar situation (although I do think that the `outputs`/`inputs` methods hold a special place in being the main "graph navigation" methods). In any case, I just wanted to leave the comment here to make sure it was a considered aspect (however briefly) but won't delay the progress of this, so feel free to merge.

sphuber:
Thanks @ramirezfranciscof . It's a good point and I agree that it certainly is not necessarily a trivial decision, but I do think that this is almost a necessary solution. I think the following is true in principle:

* The fact that link namespaces are converted to double underscores in the link label in the database is an implementation detail and the user should not have to know this in order to work with them.
* Like you mentioned, the problem you describe may also arise through different circumstances. As soon as `Node.ouputs.link` returns an object that has a property that returns another node, you can get the ambiguity that you describe. It already exists in fact, take for example: `node.outputs.some_output.creator.caller` can actually return `node` again if it represents a workflow that called a calculation which created `some_output` and the workflow also returned that.

So in the end, I don't see the need to limit the chaining of property calls to merely accessing direct inputs and output. And on the flipside, I think this solves a real usability problem, so I would still go ahead with it.

> Actually, now that I think of it, node.outputs.nested_link would return a dict, right?

Yes, it returns an `AttributeDict` instance. I will add a test for this to make sure this keeps working.

sphuber (approved):

PR #4627, opened 2020-12-09
Data: add plugin type `RemoteFolderData` and `RemoteStashData`

### LINKED ISSUE #4626: Add new data types for symbolic remote data
This issue has flown from #2150 that is adding functionality to _stash_ files from the `remote_folder` of a `CalcJob` once it has terminated. This data would be attached as the `stash_folder` output node to the calculation job node. The question arose what `Data` type it should use. To be able to distinguish it from the `remote_folder ` it was suggested to use a different class from `RemoteData`, for example `StashData`.

This raised another question of how the subclassing should be done, because in a sense the `StashData` functions exactly as a `RemoteData`. They are both "symbolic links", in the sense that AiiDA only stores a reference to where the data that they contain is stored. If the remote data is removed AiiDA doesn't necessarily know but it can no longer be used for reproducibility. So far, `RemoteData` is the only data type that is symbolic in this way, but `StashData` would also become one. In that sense, it would make sense that they share a common parent class.

If we would not have to worry about backwards compatibility, we could have created a `SymbolicData` class that would be the parent of `RemoteData` and `StashData`. However, this would require a change in entry point string for `RemoteData`, since the `QueryBuilder` subclassing is based on that. It should be changed from `remote` to `symbolic.remote` if `symbolic` is the entry point name of the `SymbolicData` base class.

Instead, we can bombard `RemoteData` to the parent class, such that the entry point does not have to be changed. We then create two new subclasses `RemoteFolderData` and `RemoteStashData`:
```
"remote = aiida.orm.nodes.data.remote:RemoteData",
"remote.folder = aiida.orm.nodes.data.remote.folder:RemoteFolderData",
"remote.stash = aiida.orm.nodes.data.remote.stash:RemoteStashData",
```
Making it `RemoteFolderData` is actually quite good as it makes for a better parallel with the `FolderData` local analog.

The migration plan would be to already change the `remote_folder` output from `RemoteData` to `RemoteFolderData`. Queries for `RemoteData` will continue to work, since `RemoteFolderData` is a subclass, _unless_ users explicitly say `subclassing=False` in the query, which should be rare. We can then optionally add deprecation warnings when a raw `RemoteData` is stored and tell people to use the new `RemoteFolderData` instead. When sufficient time has passed for transition, we remove the warnings and make `RemoteData` not storable, essentially making it a pure base class (similar to `ProcessNode`). This step is optional, I am not yet sure if this is necessary and whether it is ok to allow `RemoteData` nodes to continue to be created. Actually not even sure if this is being done by users. If we do decide to make `RemoteData` non-storable, we can add a migration changing existing nodes' type strings from `remote` to `remote.folder`.

sphuber:
Fixes #4626 

These new data plugins are a subclass of the existing `RemoteData`. The
`RemoteFolderData` is identical to the original `RemoteData` and is
essentially replacing it. The `RemoteData` type will at some point
become unstorable and will then merely function as a parent class for
querying. The `RemoteStashData` is new and is the second remote type of
data and will be used to "stash" data from the remote working directory
of a completed calculation job for safe keeping. Since this data remains
also on the remote, it is "symbolic" just as the `RemoteFolderData`
because only a reference of its location on the remote is kept. If the
data is removed on the remote, over which AiiDA has no control, any
process that uses it as input is no longer reproducible. The symbolic
data will have to be recreated first.

Since the two types share this "symbolic" character, they need to have a
shared parent class. For backward compatiblity reasons, `RemoteData` is
chosen as that parent class which means the entry point string can
remain unchanged and the subclasses can extend the `remote` namespace.


sphuber:
After discussion, this pull request will be closed. One of the main advantages of having a shared parent class for `RemoteFolderData` and `RemoteStashData` was such that `RemoteData` could be used in queries to target all remote datas. However, this can just as easily be done by passing the tuple `(RemoteFolderData, RemoteStashData)`. In addition, the idea was that the `CalcJob` class, which requires `RemoteData` for the `parent_folder`, would automatically and transparently accept both remote folder data and remote stash data. However, if the engine has to unstash the data, which can be a heavy operation, it can block the engine if many unstash operations have to be performed. This happening transparently may catch the user off guard and so it is argued that it is preferable that unstashing should be an explicit action (through a separate `TransferCalcjob` for example. As soon as the unstashing can be implemented in a non-blocking way, this objection disappears, but since we cannot do that at the moment, for now we do not want to offer this functionality.

So in essence, the changes only provide slights advantages that can be accomplished relatively easy in another way. Conversely, these change are backwards incompatible and requires a database migration, so the cost is quite large.

PR #4633, opened 2020-12-15
Engine: remote `with_persistence=False` from process function runner

### LINKED ISSUE #4632: Running a process function will create `Runner` without persister causing any submission after that to fail
When a process function is run, a runner is obtained through `Manager.get_runner`, but it passes `with_persistence=False`:
https://github.com/aiidateam/aiida-core/blob/073639aeb6d844acece2fda70aeefdca9d9005fc/aiida/engine/processes/functions.py#L116

In principle this would be fine since a process function runs in one go and so does not need to be persisted. The problem arises when the process function is the first thing to run in the interpreter, which causes the `Runner` instance of the `Manager` to be set to the persistence-less one which will cause any subsequent process that is submitted to fail, since that calls `runner.persister` to save the checkpoint after instantiating the process.

sphuber:
Fixes #4632 

In principle the runner for a process function does not need a persister
since it runs in one go and does not have intermediate steps at which
the progress needs to be persisted. However, since the process function
implementation calls `Manager.get_runner`, if a runner has not yet been
created in the interpreter, one will be created and set to be the global
one. This is where the problem occurs because the process function
specifies `with_persistence=False` for the runner. This will cause any
subsequent process submissions to fail since the `submit` function will
call `runner.persister.save_checkpoint` which will fail since the
`persister` of the runner is `None`.

### DIFF
--- a/aiida/engine/processes/functions.py
+++ b/aiida/engine/processes/functions.py
@@ -113,7 +113,7 @@ def process_function(node_class):
             :rtype: (dict, int)
             """
             manager = get_manager()
-            runner = manager.get_runner(with_persistence=False)
+            runner = manager.get_runner()
             inputs = process_class.create_inputs(*args, **kwargs)
 
             # Remove all the known inputs from the kwargs


unkcpz (approved):
Thanks @sphuber

PR #4637, opened 2020-12-16
`RemoteData`: deprecate automatic opening of a transport

### LINKED ISSUE #4636: Make `transport` a required argument of `RemoteData`
Currently, the `RemoteData` methods will open a transport itself, which is not scalable and can lead to too many connections being opened to a machine. Instead, the transport should be passed in such that a client can open a single connection and pass it around.

sphuber:
Fixes #4636 

The methods of the `RemoteData` class that require a transport now, for
convenience, automatically open that transport for the caller. However,
in a high-throughput context this is not scalable as each call opening
its own transport will incur too many opened connections to the target
machine.

Instead, a single transport should be opened that is then passed to the
method calls. All affected methods now have an argument `transport` that
allows to pass a transport in. If it is not passed a deprecation warning
is emitted as this usage will raise in the future as the argument will
be made required.

Note that there is one backwards-incompatible change. The `is_empty`
property had to be changed into a method in order for it to take the
transport as argument. It is not clear whether there is a scheme in
which a property can be transformed into a method in a backwards
compatible way through a deprecation pathway.

### DIFF
--- a/aiida/orm/nodes/data/remote.py
+++ b/aiida/orm/nodes/data/remote.py
@@ -9,7 +9,9 @@
 ###########################################################################
 """Data plugin that models a folder on a remote computer."""
 import os
+import warnings
 
+from aiida.common.warnings import AiidaDeprecationWarning
 from aiida.orm import AuthInfo
 from .data import Data
 
@@ -42,13 +44,17 @@ class RemoteData(Data):
     def set_remote_path(self, val):
         self.set_attribute('remote_path', val)
 
-    @property
-    def is_empty(self):
+    def is_empty(self, transport=None):
         """
         Check if remote folder is empty
+
+        :param transport: optional transport to use, if not specified one will be opened. From `v3.0.0` this argument
+            will be required.
         """
-        authinfo = self.get_authinfo()
-        transport = authinfo.get_transport()
+        if transport is None:
+            warnings.warn('`transport` is `None`, this will start to raise in `v3.0.0`', AiidaDeprecationWarning)  # pylint: disable=no-member
+            authinfo = self.get_authinfo()
+            transport = authinfo.get_transport()
 
         with transport:
             try:
@@ -59,16 +65,21 @@ class RemoteData(Data):
 
             return not transport.listdir()
 
-    def getfile(self, relpath, destpath):
+    def getfile(self, relpath, destpath, transport=None):
         """
         Connects to the remote folder and retrieves the content of a file.
 
         :param relpath:  The relative path of the file on the remote to retrieve.
         :param destpath: The absolute path of where to store the file on the local machine.
+        :param transport: optional transport to use, if not specified one will be opened. From `v3.0.0` this argument
+            will be required.
         """
-        authinfo = self.get_authinfo()
+        if transport is None:
+            warnings.warn('`transport` is `None`, this will start to raise in `v3.0.0`', AiidaDeprecationWarning)  # pylint: disable=no-member
+            authinfo = self.get_authinfo()
+            transport = authinfo.get_transport()
 
-        with authinfo.get_transport() as transport:
+        with transport:
             try:
                 full_path = os.path.join(self.get_remote_path(), relpath)
                 transport.getfile(full_path, destpath)
@@ -79,30 +90,33 @@ class RemoteData(Data):
                             full_path,
                             self.computer.label  # pylint: disable=no-member
                         )
-                    )
+                    ) from exception
                 raise
 
-    def listdir(self, relpath='.'):
+    def listdir(self, relpath='.', transport=None):
         """
         Connects to the remote folder and lists the directory content.
 
         :param relpath: If 'relpath' is specified, lists the content of the given subfolder.
+        :param transport: optional transport to use, if not specified one will be opened. From `v3.0.0` this argument
+            will be required.
         :return: a flat list of file/directory names (as strings).
         """
-        authinfo = self.get_authinfo()
+        if transport is None:
+            warnings.warn('`transport` is `None`, this will start to raise in `v3.0.0`', AiidaDeprecationWarning)  # pylint: disable=no-member
+            authinfo = self.get_authinfo()
+            transport = authinfo.get_transport()
 
-        with authinfo.get_transport() as transport:
+        with transport:
             try:
                 full_path = os.path.join(self.get_remote_path(), relpath)
                 transport.chdir(full_path)
             except IOError as exception:
                 if exception.errno == 2 or exception.errno == 20:  # directory not existing or not a directory
-                    exc = IOError(
-                        'The required remote folder {} on {} does not exist, is not a directory or has been deleted.'.
-                        format(full_path, self.computer.label)  # pylint: disable=no-member
-                    )
-                    exc.errno = exception.errno
-                    raise exc
+                    raise IOError(
+                        f'The required remote folder {full_path} on {self.computer.label} does not exist, is not a '
+                        'directory or has been deleted.'
+                    ) from exception
                 else:
                     raise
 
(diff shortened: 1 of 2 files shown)


greschd:
Do we already have a way for workchain steps to acquire / request the transport? If that isn't possible, code that uses `RemoteData` in workchain steps currently won't have a clear way of how to update to the new behavior.
I seem to remember there were discussions at the last Hackathon about enabling this, but don't know if anything came from that.

> Note that there is one backwards-incompatible change. The `is_empty`
> property had to be changed into a method in order for it to take the
> transport as argument. It is not clear whether there is a scheme in
> which a property can be transformed into a method in a backwards
> compatible way through a deprecation pathway.

I think `is_empty` could be a wrapper object that has a `__call__` to implement the new behavior, and `__bool__` to implement the old behavior (and warn). This sort of assumes that the previous `is_empty` is used in a boolean test (and e.g. an `is False` test could still fail), but it's better than the alternative: If it's just a method, `bool(obj.is_empty)` will always be `True`.

sphuber:
> Do we already have a way for workchain steps to acquire / request the transport? If that isn't possible, code that uses `RemoteData` in workchain steps currently won't have a clear way of how to update to the new behavior.
> I seem to remember there were discussions at the last Hackathon about enabling this, but don't know if anything came from that.
atible way through a deprecation pathway.

I was experimenting with this just now, but I am afraid that this will take significant changes to the engine. The reason being that to allow getting transport in a workchain step (or `CalcJob.prepare_for_submission` for that matter) that does not fully block everything else, it has to be done asynchronously. However, as far as I can tell, this will force us to make all process steps coroutines, meaning all existing workchain implementations will have to be updated. Now, in the grand scheme of things, this may actually be desirable, but it will take a big effort and not something that we can quickly implement. The reasoning for this change is that whatever we will end up doing, this change should be one in the right direction. The other option would be to have some kind of a singleton that acts as a transport queue that can be called from within the `RemoteData` methods but that still would require asynchronous code for it to work properly, and I don't like this design of tight coupling. Dependency injection of the transport seems like a better design. Happy to discuss though.

What I think we _should_ have as an interface is something like the following:
```python
class TransportWorkChain(engine.WorkChain):

    @classmethod
    def define(cls, spec):
        super().define(spec)
        spec.outline(
            cls.use_transport
        )

    @auto_retry(initial_interval=10, max_attempts=5, ignore_exceptions=None)
    async def use_transport(self):
        with self.request_transport() as transport:
            transport.do_something()
```
I think each process step should be made asynchronous which will then make it easy for a user to write a step that can get transport through the queue, respecting the connection limits, and by applying the safety of the EBM we currently have for `CalcJob` transport tasks. If we can get this to work technically, I think this would be a super smooth interface given the complexity that it abstract from the user and the powerful feature that it provides. The downside (as far as I can tell) is that this forces _all_ workchain logic to be async and so users will always have to write this. Maybe this is a small price to pay in the end, given that most of the time they just have to slap `async` in front of method definitions and sometimes add `await`.


> I think `is_empty` could be a wrapper object that has a `__call__` to implement the new behavior, and `__bool__` to implement the old behavior (and warn). This sort of assumes that the previous `is_empty` is used in a boolean test (and e.g. an `is False` test could still fail), but it's better than the alternative: If it's just a method, `bool(obj.is_empty)` will always be `True`.
> 
> The obvious other way would be choosing a different name for the new `is_empty`, but that may be less desirable in terms of consistency with other parts of the code.

I did indeed want to try and avoid this. I will try your trick first and see if I can get that to work

greschd:
Lots to unpack here, I'll have to think about it a bit. For now, just a few quick thoughts:

- What prevents us from _allowing_ workchain steps to be `async`, but only _requiring_ it when an async feature is actually used? It's relatively straightforward to wrap a normal function s.t. it becomes async, we can do that on our side. Example code: https://github.com/FrescolinoGroup/pyasynctools/blob/master/fsc/async_tools/_wrap_to_coroutine.py
- It should probably be `async with self.request_transport() as transport:`, right? That would also make it break if used from a sync function, which is what we want IMO.

Maybe we can already merge this PR while the details of transport in steps is being worked out, but IMO we shouldn't _remove_ the deprecated code before it's solved. Having a clear upgrade path on deprecation would be nicer, but not strictly necessary.

sphuber:
>What prevents us from _allowing_ workchain steps to be `async`, but only _requiring_ it when an async feature is actually used? It's relatively straightforward to wrap a normal function s.t. it becomes async, we can do that on our side. Example code: https://github.com/FrescolinoGroup/pyasynctools/blob/master/fsc/async_tools/_wrap_to_coroutine.py

This is a good point and we actually have utilities already in `aiida-core` for this in `aiida.engine.utils.ensure_coroutine`.

> It should probably be `async with self.request_transport() as transport:`, right? That would also make it break if used from a sync function, which is what we want IMO.

No, there is no need to put `async` in front of the context manager. In fact, I think that is invalid syntax. Or maybe I misunderstand your point.

> Maybe we can already merge this PR while the details of transport in steps is being worked out, but IMO we shouldn't _remove_ the deprecated code before it's solved. Having a clear upgrade path on deprecation would be nicer, but not strictly necessary.

Even if we don't give a fully new robust way of doing things, the code will still be usable with the same problems of the current design. I.e. even when we remove the deprecated code, people can still do
```
with remote_data.computer.get_transport() as transport:
    remote_data.listdir(transport)
```
This can be integrated in a workchain without a problem and suffers from exactly the same problems as the current implementation. It just takes one line extra but already prepares the interface for a future more advanced way of retrieving and using transports

greschd:
> No, there is no need to put async in front of the context manager. In fact, I think that is invalid syntax. Or maybe I misunderstand your point.

Maybe I'm misunderstanding the intent here, but shouldn't `request_transport` yields the transport when it is available? That means we can't block the event loop, otherwise it will never become available. Which means it needs to be a coroutine, specifically an [asynchronous context manager](https://docs.python.org/3/glossary.html#term-asynchronous-context-manager) (implementing `__aenter__` and `__aexit__` coroutines) called with [`async with`](https://docs.python.org/3/reference/compound_stmts.html#the-async-with-statement).

> Even if we don't give a fully new robust way of doing things, the code will still be usable with the same problems of the current design. [...]

Gotcha, didn't realize this was quite so simple. So they would have to add the (normal) context manager now, and then just replace which (possibly async) context manager is used once that's available.

giovannipizzi:
Thanks Sebastiaan! 

1. regarding the design of the steps, I like it - maybe can you copy-paste the relevant bits in the correct issue #3478 so we don't lose track of this?
2. regarding this PR: I know I pushed for it, but there is still a bit that leaves me "confused", that is how to check if the transport is the "correct" one. Since anyway the deprecation is not for 3.0: would it be safer just *not* to deprecate the old behaviour for now, but just add the transport? We can always deprecate it later in 2.0 when we're sure that it's safe, after having tested the new functionality where you can pass the transport (that IMO is instead very important to have, so that if I am in the shell and want to do an operation on many remote datas (e.g. to clean them up) I have a way to do it with a single transport).
   Essentially, this means just keeping the new code but removing the deprecation warnings.

Finally for is_empty - honestly, I don't think it makes sense to keep it; you can just call do `not remote.listdir()`, which is what is being done internally.
I would just keep as it was, with no adaptation, but deprecate the function suggesting to use `not remote.listdir()`, and we remove it in 3.0.

ramirezfranciscof:
>     2\. regarding this PR: I know I pushed for it, but there is still a bit that leaves me "confused", that is how to check if the transport is the "correct" one. Since anyway the deprecation is not for 3.0: would it be safer just _not_ to deprecate the old behaviour for now, but just add the transport? We can always deprecate it later in 2.0 when we're sure that it's safe, after having tested the new functionality where you can pass the transport (that IMO is instead very important to have, so that if I am in the shell and want to do an operation on many remote datas (e.g. to clean them up) I have a way to do it with a single transport).
>         Essentially, this means just keeping the new code but removing the deprecation warnings.

Yeah, this was also something I didn't fully understand. Do we need to deprecate this? Can't we just keep the old behaviour that when no transport is provided it will just open one? It is true that it leaves a path open for inefficiency if users start doing this in automated scripts, but I think it could be useful for "pedagogical" purposes (when learning or showcasing AiiDA we might want to illustrate some points regarding access to remote data without having to go into the details of the transports just yet).

On the other hand, could we maybe take these lines:

```python

        if transport is None:
            warnings.warn('`transport` is `None`, this will start to raise in `v3.0.0`', AiidaDeprecationWarning)  # pylint: disable=no-member
            authinfo = self.get_authinfo()
            transport = authinfo.get_transport()
```

...and replace them with a call (`transport = checkget_transport( transport )`) to an internal method? I know right now its only replacing 3 (2 removing the deprecation?) lines for 1 so there is not THAT MUCH to gain from code de-duplication, but it might already have a nice setup for when we figure out the way to check if the transport is appropriate and more lines need to be added.

sphuber:
After discussion, this PR is going to be closed for now. Even if the transport will be made a required argument, there is still the challenge that it needs to be checked that it is a transport for a computer that matches that of the `RemoteData`. One idea would be to extend the `Transport` class to have a `computer_uuid` attribute, that could be set to the UUID of the `Computer` to which it belongs. Note that this should not be required, as the `Transport` class really should remain independent on AiiDA. This optional class attribute is maybe the least intrusive way of connecting the transport to the computer, even though conceptually it is still not ideal.

PR #4638, opened 2020-12-16
CI: remove the `numpy` install workaround for `pymatgen`

sphuber:
The problem occurred due to an outdated version of `setuptools` which
would be invoked when `pymatgen` gets installed from a tarball, in which
case the wheel has to be built. In this scenario, the build requirements
get installed by `setuptools`, which at outdated versions did not
respect the Python requirements of the dependencies which would cause
incompatible version of `numpy` to be installed, calling the build to
fail. By updating `setuptools` the workaround of manually installing a
compatible `numpy` version beforehand is no longer necessary.

### DIFF
--- a/.github/workflows/ci-code.yml
+++ b/.github/workflows/ci-code.yml
@@ -84,17 +84,11 @@ jobs:
         sudo apt update
         sudo apt install postgresql-10 graphviz
 
-    - name: Upgrade pip
+    - name: Upgrade pip and setuptools
+      # It is crucial to update `setuptools` or the installation of `pymatgen` can break
       run: |
-        pip install --upgrade pip
+        pip install --upgrade pip setuptools
         pip --version
-
-    # Work-around issue caused by pymatgen's setup process, which will install the latest
-    # numpy version (including release candidates) regardless of our actual specification
-    # By installing the version from the requirements file, we should get a compatible version
-    - name: Install numpy
-      run: |
-        pip install `grep 'numpy==' requirements/requirements-py-${{ matrix.python-version }}.txt`
 
     - name: Install aiida-core
       run: |
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -153,14 +153,11 @@ jobs:
         sudo apt update
         sudo apt install postgresql-10 graphviz
 
-    - run: pip install --upgrade pip
-
-    # Work-around issue caused by pymatgen's setup process, which will install the latest
-    # numpy version (including release candidates) regardless of our actual specification
-    # By installing the version from the requirements file, we should get a compatible version
-    - name: Install numpy
+    - name: Upgrade pip and setuptools
+      # It is crucial to update `setuptools` or the installation of `pymatgen` can break
       run: |
-        pip install `grep 'numpy==' requirements/requirements-py-${{ matrix.python-version }}.txt`
+        pip install --upgrade pip setuptools
+        pip --version
 
     - name: Install aiida-core
       run: |
--- a/tests/restapi/test_threaded_restapi.py
+++ b/tests/restapi/test_threaded_restapi.py
@@ -62,8 +62,7 @@ def test_run_threaded_server(restapi_server, server_url, aiida_localhost):
             pytest.fail('Thread did not close/join within 1 min after REST API server was called to shutdown')
 
 
-# Tracked in issue #4281
-@pytest.mark.flaky(reruns=2)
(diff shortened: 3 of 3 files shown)


sphuber:
@csadorf I think I have found the problem and it basically is due to an outdated version of `setuptools`. Updating that in our CI solves the issue. I think this means we can also provide a guaranteed installation pathway for users if we tell in the installation docs to update `setuptools` before calling `pip install aiida-core`.

csadorf (commented):
Looks good. I assume that the failing tests are unrelated?

sphuber:
> Looks good. I assume that the failing tests are unrelated?

Looks like it, but it is really peculiar that all four of them fail. I will relaunch them and see what happens.

sphuber:
@csadorf they failed again, with same problem. That makes it really unlikely to be coincidence. However, I have no idea how the hell the change of this commit can affect the REST API test of the SqlAlchemy session. No clue even where to start looking

csadorf:
> @csadorf they failed again, with same problem. That makes it really unlikely to be coincidence. However, I have no idea how the hell the change of this commit can affect the REST API test of the SqlAlchemy session. No clue even where to start looking

@sphuber Looks like this test has been failing for 5 days now: https://github.com/aiidateam/aiida-core/runs/1544307503?check_suite_focus=true
So this is indeed unrelated, but not clear to me what is causing it.

sphuber:
@CasperWA is the one who wrote it. Do you have an idea as to why it might be failing?

CasperWA:
> @CasperWA is the one who wrote it. Do you have an idea as to why it might be failing?

No idea, sorry.
The point of the test is to check that a bug exists. So we could skip/remove it?
There may have been an update to the used packages, especially SQLAlchemy that fixes the QueuePool stuff? But that's the only thing I can think of.

If it is fixed though, it could suggest that we can update/rethink the session implementation we have done for the special QueryBuilder session? Although it might just suggest that SQLAlchemy has implemented a temporary workaround if the session setup is indeed wrong, but at least it doesn't make it fail for users, making our setup is still the valid design for how to do this.

In summary. No idea 😓

sphuber:
Made a diff for the `pip freeze` of the Python 3.9 SqlAlchemy run for the last build that worked and the one that fails, which yields:
```diff
--- /home/sph/code/aiida/env/dev/aiida-core/success.dat Thu Dec 17 16:11:53 2020
+++ /home/sph/code/aiida/env/dev/aiida-core/fail.dat    Thu Dec 17 16:11:23 2020
@@ -42,7 +42,7 @@
 idna==2.10
 imagesize==1.2.0
 iniconfig==1.1.1
-ipykernel==5.4.0
+ipykernel==5.4.2
 ipython==7.19.0
 ipython-genutils==0.2.0
 ipywidgets==7.5.1
@@ -71,7 +71,7 @@
 networkx==2.5
 notebook==6.1.5
 numpy==1.19.4
-packaging==20.7
+packaging==20.8
 palettable==3.3.0
 pamqp==2.3.0
 pandas==1.1.5
@@ -92,7 +92,7 @@
 psutil==5.7.3
 psycopg2-binary==2.8.6
 ptyprocess==0.6.0
-py==1.9.0
+py==1.10.0
 py-cpuinfo==7.0.0
 PyCifRW==4.4.2
 pycparser==2.20
@@ -103,7 +103,7 @@
 PyNaCl==1.4.0
 pyparsing==2.4.7
 pyrsistent==0.17.3
-pytest==6.1.2
+pytest==6.2.0
 pytest-asyncio==0.14.0
 pytest-benchmark==3.2.3
 pytest-cov==2.10.1
@@ -148,7 +148,7 @@
 sqlalchemy-diff==0.1.3
 SQLAlchemy-Utils==0.34.2
 sqlparse==0.4.1
-sympy==1.7
+sympy==1.7.1
 tabulate==0.8.7
 terminado==0.9.1
 testpath==0.4.4
```
The package `py` which has been deprecated [updated to `1.10` exactly 5 days ago](https://pypi.org/project/py/1.10.0/#history), but from the [changes in the `CHANGELOG`](https://github.com/pytest-dev/py/commit/e5ff378fc3bd3f7c366dec769a718bdb1ceca1f1) it seems unlikely to be the culprit. The other packages I am not sure yet

CasperWA:
Hmm. I suggest to skip this test in this PR, open an issue for it (put this issue number into the skip message) and follow up on this separately, pointing to the investigative [comment above](https://github.com/aiidateam/aiida-core/pull/4638#issuecomment-747502926) by @sphuber.

sphuber:
@CasperWA this test was already known to be problematic and was being rerun with flaky, and as such is already being tracked in #4281 . I have skipped it entirely for now

csadorf (approved):

PR #4641, opened 2020-12-18
CLI: ensure `verdi database version` works even if schema outdated

### LINKED ISSUE #4640: `verdi database version` does not work if schema version is out of sync

sphuber:
Fixes #4640 

The command was failing if the database schema was out of sync because
the backend was loaded, through `get_manager`, with the default schema
check on. Since the database does not actually have to be used, other
than to retrieve the current schema version and generation, we can load
the backend without the check.


mbercx (approved):
All good, thanks @sphuber! Works like a charm.

Issue #4647, opened 2021-01-07
Unpin doc/dev dependencies


chrisjsewell:
@csadorf in 7b03f04 I had to pin sphinx and pylint-django to less than the latest versions:

For sphinx, on CircleCI we are getting: https://app.circleci.com/pipelines/github/aiidateam/aiida-core/2403/workflows/18751484-3474-49c1-849b-4bd307cd43b8/jobs/2395

```
reading sources... [ 60%] reference/apidoc/aiida.restapi

Exception occurred:
  File "/home/circleci/.local/lib/python3.7/site-packages/flask/globals.py", line 38, in _lookup_req_object
    raise RuntimeError(_request_ctx_err_msg)
RuntimeError: Working outside of request context.

This typically means that you attempted to use functionality that needed
an active HTTP request.  Consult the documentation on testing for
information about how to avoid this problem.
The full traceback has been saved in /tmp/sphinx-err-0am0v1k8.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!

Exited with code exit status 2

```

for pylint-django (see : https://github.com/aiidateam/aiida-core/runs/1657433968

```
pylint...................................................................Failed
- hook id: pylint
- exit code: 2

************* Module tests.manage.test_caching_config
tests/manage/test_caching_config.py:1:0: E5110: Django was not configured. For more information runpylint --load-plugins=pylint_django --help-msg=django-not-configured (django-not-configured)

------------------------------------
Your code has been rated at 10.00/10

************* Module aiida.workflows.arithmetic
aiida/workflows/arithmetic/__init__.py:1:0: E5110: Django was not configured. For more information runpylint --load-plugins=pylint_django --help-msg=django-not-configured (django-not-configured)

------------------------------------
Your code has been rated at 10.00/10
```

See https://github.com/PyCQA/pylint-django/blob/master/CHANGELOG.rst#version-240-06-jan-2021

these are not "production" dependencies, only for development/documentation, so I wouldn't say this is high priority

csadorf:
Thank you for letting me know and creating the issue 👍 !

chrisjsewell:
another one to add to the list: remove astroid from pre-commit extras (added in #4757) after https://github.com/PyCQA/astroid/issues/895 fixed

chrisjsewell:
[...]

PR #4657, opened 2021-01-12
docs: clarify difference between jobinfos

### LINKED ISSUE #4656: Docs: clarify last_job_info vs detailed_job_info

The `last_job_info` and `detailed_job_info` are different pieces of information stored on the node, acquired by different means.

The documentation of the functions is currently lacking in terms of how/when this information was acquired and how they differ.

```
    def get_detailed_job_info(self):
        """Return the detailed job info dictionary.
```

```
    def get_last_job_info(self):
        """Return the last information asked to the scheduler about the status of the job.
```

Note: we noticed that the status of the calculation reported in the `last_job_info` can differ from the status of the calculation reported in the `detailed_job_info` - in particular, the status reported in the `last_job_info` always seems to be `RUNNING`, while the `detailed_job_info` is acquired after the job finishes (?).

ltalirz:
fix #4656 

CalcJobNodes contain two different job infos, the detailed_job_info and
the last_job_info.

The distinction between the two was not obvious, and not documented.

### DIFF
--- a/aiida/orm/nodes/process/calculation/calcjob.py
+++ b/aiida/orm/nodes/process/calculation/calcjob.py
@@ -427,6 +427,8 @@ class CalcJobNode(CalculationNode):
     def get_detailed_job_info(self):
         """Return the detailed job info dictionary.
 
+        The scheduler is polled for the detailed job info after the job is completed and ready to be retrieved.
+
         :return: the dictionary with detailed job info if defined or None
         """
         return self.get_attribute(self.SCHEDULER_DETAILED_JOB_INFO_KEY, None)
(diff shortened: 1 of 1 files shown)


ltalirz:
@giovannipizzi 
It may actually be less confusing to update the `last_jobinfo` also at the very last poll (setting it to `None`), even if this means some "loss of information".
Anyhow, perhaps better not to change it now since people may rely on it...

giovannipizzi (dismissed):
Thanks! I would personally keep the current behaviour (second-to-last) exactly because it's a different set of information (comes from `squeue` instead of `sacct`) and good to keep together with the detailed job info for debugging.

giovannipizzi:
Do you know why the circleci docs fail?

ltalirz:
thanks for the review - it seems the reference was not found; trying again

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -179,11 +179,11 @@ async def do_update():
         if job_info is None:
             # If the job is computed or not found assume it's done
             node.set_scheduler_state(JobState.DONE)
-            job_done = True
         else:
             node.set_last_job_info(job_info)
             node.set_scheduler_state(job_info.job_state)
-            job_done = job_info.job_state == JobState.DONE
+
+        job_done = job_info.job_state == JobState.DONE
```
sphuber, line 186: This change will now raise if `job_info` is `None`, right? Don't see why this change is necessary

ltalirz: Oops, thanks for spotting!

I guess this should have been
```
job_done = node.get_scheduler_state() == JonState.DONE
```
but then it's a database access, so I will revert this

chrisjsewell:
> Do you know why the circleci docs fail?

#4647

ltalirz:
I think it should be fine to merge now

sphuber (approved):
Thanks @ltalirz

PR #4662, opened 2021-01-14
docs: simplify proxycommand

ltalirz:
The 'netcat mode' `-W` was added in [OpenSSH version 5.4](https://www.openssh.com/txt/release-5.4), released March 2010.
Given that this simplifies the setup and delegates handling of netcat
to ssh, this is what we should recommend.

For example, MacOS ships with OpenSSH 5.6 since [MacOS 10.7](https://opensource.apple.com/source/OpenSSH/OpenSSH-166/openssh/version.h.auto.html), released October 2010.


giovannipizzi (approved):
Thanks! I didn't know about this - seems to be working fine, so I'm approving it!

giovannipizzi:
How does it work in this case, I guess that tests are skipped since it's just a doc change? Should one force the squash/merge?

chrisjsewell:
Yep

PR #4663, opened 2021-01-14
cli: notify users of runner usage

ltalirz:
The extra query performed in `verdi process list` to get the load on the
daemon runners is slow - and usually results in no output for the user.
This PR at least lets the user know what is being queried (which also may be useful
to monitor runner load).

### DIFF
--- a/aiida/cmdline/commands/cmd_process.py
+++ b/aiida/cmdline/commands/cmd_process.py
@@ -54,6 +54,7 @@ def process_list(
     # pylint: disable=too-many-locals
     from tabulate import tabulate
     from aiida.cmdline.utils.common import print_last_process_state_change, check_worker_load
+    from aiida.engine.daemon.client import get_daemon_client
 
     relationships = {}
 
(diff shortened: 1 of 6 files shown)


ltalirz:
Not sure why the tests failed... didn't change docs + this should not be backend-specific behavior

sphuber (changes requested):
Thanks @ltalirz . Fine by me, would just not add the case where daemon is not running as that is already covered in `verdi process list` itself.

ltalirz:
> Thanks @ltalirz . Fine by me, would just not add the case where daemon is not running as that is already covered in `verdi process list` itself.

Thanks - I've pulled out that check from the previous function and now call the worker query only when needed.

docs/source/reference/command_line.rst
```diff
@@ -323,11 +323,9 @@ Below is a list with all available subcommands.
                                       .aiida extension for these HTTP addresses. Automatically
                                       discovered archive URLs will be downloaded and added to
                                       ARCHIVES for importing
```
ltalirz, line 325: Note: I first removed these changes again but then noticed that they also appear on the [CI pre-commit check](https://github.com/ltalirz/aiida-core/runs/1721347516?check_suite_focus=true) - some updated version perhaps?

ltalirz:
@sphuber Now that the print is gone when no daemon is running, the test checking its output fails.

What do you think is the best approach here - fake a daemon or move this test to the daemon tests?
I'm wondering whether one could fake a running daemon by running things inside a `with` block that creates the pid file, something along the lines of 

```python
with open(get_daemon_client().circus_port_file, 'w', encoding='utf8') as fhandle:
   fhandle.write(123)

   test..
```

Perhaps there are cleaner approaches

sphuber:
> Perhaps there are cleaner approaches

The typical way is to simply monkeypatch the `DaemonClient.is_daemon_running` property to always return `True`. This is made easy using `pytest`, so I have migrated the test to `pytest` and it should now work. Let me know if you also are happy with the changes, then I will approve and merge @ltalirz

ltalirz (commented):
thanks @sphuber !

sphuber (approved):

Issue #4298, opened 2020-08-12
kill the parent workchain does not kill its child process 


unkcpz:
### Steps to reproduce

1. running the following script
```python
from aiida.plugins import CalculationFactory, WorkflowFactory
from aiida.engine import run


MultiplyAddWorkChain = WorkflowFactory('arithmetic.multiply_add')

code = load_code(label='add@localhost')
inputs = {
    'x': Int(1),
    'y': Int(2),
    'z': Int(3),
    'code': code,
}

# Normal inputs should run just fine
results, node = run.get_node(MultiplyAddWorkChain, **inputs)
```
2. Before it finished terminate the process by `Ctrl+c`
3. See error: the child process `ArithmeticAddCalculation` is in waiting and unreachable.

### Your environment

- Operating system [e.g. Linux]:
- Python version [e.g. 3.8.1]:
- aiida-core version [e.g. 1.3.0]:

Other relevant software versions, e.g. Postres & RabbitMQ

sphuber:
This is a duplicate of #3776 but we can keep this around as it gives an actual example to reproduce it.

unkcpz:
More information about this issue. 
1) If use submit to launch the workchain, and kill the parent process, the child process is able to be killed. 
2) Before the parent process kill the child by `controller.kill_process(child.pk, 'msg')` in `aiida.engine.processes.process` the child process is unreachable. `controller.kill_process` return the exception `kiwipy.exceptions.UnroutableError: ('NO_ROUTE', '[rpc].102')`

ltalirz:
> 2\. See error: the child process `ArithmeticAddCalculation` is in waiting and unreachable.

@unkcpz Could you please elaborate a bit on this - which error?


I just tried reproducing this issue on aiida-core v1.5.2 and depending on when I abort, I either get
```
$ verdi run runtest.py
^C
Aborted!
```
or
```
$ verdi run runtest.py
^CTraceback (most recent call last):
  File "/Users/leopold/Applications/miniconda3/envs/aiida_rmq_py3/bin/verdi", line 12, in <module>
    sys.exit(verdi())
  File "/Users/leopold/Applications/miniconda3/envs/aiida_rmq_py3/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/leopold/Applications/miniconda3/envs/aiida_rmq_py3/lib/python3.6/site-packages/click/core.py", line 717, in main
[...]

Issue #3776, opened 2020-02-20
Ensure that children processes are properly killed when parent killed and clean up nodes


sphuber:
There are various scenarios possible where when killing a parent process not all children processes are properly killed as well, or maybe just the nodes are not properly updated. Make sure that the process tasks are properly acknowledged so they don't remain in the queue and make sure to wrap up the nodes.

yakutovicha:
One more scenario, when killing the top work chain doesn't kill its children:

So I am running the following set of processes `EquationOfState` (EOF)-> `Cp2kOpt` -> `Cp2kBase` -> `Cp2kCalc`. I'm NOT submitting the calculation to the daemon, but running it.

When trying to kill the top work chain (EOS) using `verdi process kill`, I got the second one (`Cp2kOpt`) killed as well. At the same time, both `Cp2kBase` and `C2pk` calculation remain unreachable and can't be killed.

If I submit the work chain to the daemon instead - everything can be killed just fine.

ltalirz:
@yakutovicha Could you please check the daemon logs for

```
AttributeError: 'NoneType' object has no attribute 'set_result'
```

If that is found, then https://github.com/aiidateam/aiida-core/pull/4669 may fix this issue

chrisjsewell:
Firstly, to note that now #4669 has been merged, instead of the `AttributeError: 'NoneType' object has no attribute 'set_result'`, there should now be a logger warning `logger.warning(f'killed CalcJob<{node.pk}> but async future was None')` 

Secondly, I have now created https://github.com/aiidateam/aiida-integration-tests.
The README explains what this is in detail, but basically the idea is to create a reproducible environment (with Docker Compose) to investigate these kind of "production" issues.

(see also https://github.com/aiidateam/aiida-core/issues/4603#issuecomment-767646196)

I can reproduce the issue whereby killing a running workchain (not submitted to the daemon), leaves the children unreachable:

Here is the outcome of `CTRL-C` a running workchain with children:

```console
root@2940cef2c10d:~# aiida-sleep workchain -nw 1 -nc 10 -t 120
^C01/26/2021 04:14:18 PM <294> aiida.engine.runners: [CRITICAL] runner received interrupt, killing process 13
Traceback (most recent call last):
  File "/opt/venv/bin/aiida-sleep", line 8, in <module>
    sys.exit(main())
  File "/opt/venv/lib/python3.8/site-packages/click/core.py", line 764, in __call__
[...]

PR #4669, opened 2021-01-22
Type Checking: `aiida/engine` (+bug fixes)

chrisjsewell:
This PR updates plumpy, to a new typed version, and adds type checking to the whole of `aiida/engine`.

Notes:

- mypy seems to not play nice with `__all__`, because it could not find things installed from top-level plumpy, so I had to change e.g. `plumpy.Wait` to `plumpy.process_states.Wait`
(from a Google I couldn't find any obvious fixes)
- I removed some `yapf: disable` statements, because they actually disable yapf for the entire rest of the file, even if added inside a function

I've identified two definite bugs, which perhaps could be related to open issues:

1. Here the actual state should be set, not the key: https://github.com/aiidateam/aiida-core/blob/61e48d7a3dac0bc9956f8ac34b8e3cc19db1fc3e/aiida/engine/processes/process.py#L532

2. Here `self._killing` may be `None` and so could raise the wrong exception (`AttributeError` not `KillInterruption`): https://github.com/aiidateam/aiida-core/blob/61e48d7a3dac0bc9956f8ac34b8e3cc19db1fc3e/aiida/engine/processes/calcjobs/tasks.py#L376


chrisjsewell:
weird that the pylint errors have shown up:

```
 aiida/engine/processes/process.py:277:12: W0201: Attribute '_pid' defined outside __init__ (attribute-defined-outside-init)
aiida/engine/processes/process.py:279:12: W0201: Attribute '_pid' defined outside __init__ (attribute-defined-outside-init)
aiida/engine/processes/process.py:362:8: W0201: Attribute '_pid' defined outside __init__ (attribute-defined-outside-init)
aiida/engine/processes/process.py:553:20: W0201: Attribute '_state' defined outside __init__ (attribute-defined-outside-init)
```

`_pid` is definitely defined in the `super` init

aiida/engine/processes/process.py
```diff
@@ -499,37 +526,33 @@ def build_process_type(cls):
...
+
         """
         self._node = self.get_or_create_db_record()
         self._setup_db_record()
         if self.metadata.store_provenance:
             try:
                 self.node.store_all()
                 if self.node.is_finished_ok:
-                    self._state = ProcessState.FINISHED
+                    self._state = Finished(self, None, True)  # TODO check correct
```
chrisjsewell, line 555: This fix needs checking

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -370,12 +407,13 @@ async def execute(self):  # pylint: disable=invalid-overridden-method
...
 
         except TransportTaskException as exception:
-            raise plumpy.PauseInterruption(f'Pausing after failed transport task: {exception}')
-        except plumpy.KillInterruption:
+            raise plumpy.process_states.PauseInterruption(f'Pausing after failed transport task: {exception}')
+        except plumpy.process_states.KillInterruption:
             await self._launch_task(task_kill_job, node, transport_queue)
-            self._killing.set_result(True)
+            if self._killing is not None:  # TODO possible source of bug
+                self._killing.set_result(True)
```
chrisjsewell, line 414: This fix needs checking

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -52,41 +56,41 @@ def __init__(self, authinfo, transport_queue, last_updated=None):
...
         """
         return self._logger
 
-    def get_minimum_update_interval(self):
+    def get_minimum_update_interval(self) -> float:
         """Get the minimum interval that should be respected between updates of the list.
 
         :return: the minimum interval
-        :rtype: float
+
```
sphuber, line 76: Why keep these lines empty. Could just remove all of these, no? Dead whitespace

aiida/engine/processes/calcjobs/manager.py
```diff
@@ -188,12 +192,12 @@ async def updating():
             )
 
     @staticmethod
-    def _has_job_state_changed(old, new):
+    def _has_job_state_changed(old: Optional['JobInfo'], new: Optional['JobInfo']) -> bool:
         """Return whether the states `old` and `new` are different.
 
         :type old: :class:`aiida.schedulers.JobInfo` or `None`
         :type new: :class:`aiida.schedulers.JobInfo` or `None`
```
sphuber, line 199: can remove these lines

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -11,19 +11,27 @@
...
+from aiida.engine.transports import TransportQueue
+from aiida.engine.utils import exponential_backoff_retry, interruptable_task, InterruptableFuture
 from aiida.schedulers.datastructures import JobState
 from aiida.manage.configuration import get_config_option
 
 from ..process import ProcessState
 
+if TYPE_CHECKING:
+    from aiida.orm.nodes.process.calculation.calcjob import CalcJobNode
+    from .calcjob import CalcJob
```
sphuber, line 33: Is this just for efficiency? I don't think this could actually cause problems like cyclic dependencies, could it? Wouldn't it just be better to import it and use `CalcJobNode` for typing instead of `'CalcJobNode'`. Not sure what the standard is here for `mypy`.

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -396,39 +434,45 @@ async def _launch_task(self, coro, *args, **kwargs):
...
             self._task.interrupt(reason)
 
-        if isinstance(reason, plumpy.KillInterruption):
+        if isinstance(reason, plumpy.process_states.KillInterruption):
             if self._killing is None:
-                self._killing = plumpy.Future()
+                self._killing = plumpy.futures.Future()
             return self._killing
+
+        return None
```
sphuber, line 478: are these explicit `return None`'s actually required?

aiida/engine/processes/ports.py
```diff
@@ -195,7 +197,7 @@ def validate_port_name(port_name):
         if any([len(entry) > PORT_NAME_MAX_CONSECUTIVE_UNDERSCORES for entry in consecutive_underscores]):
             raise ValueError(f'invalid port name `{port_name}`: more than two consecutive underscores')
 
-    def serialize(self, mapping, breadcrumbs=()):
+    def serialize(self, mapping: Optional[Dict[str, Any]], breadcrumbs: Sequence[str] = ()) -> Optional[Dict[str, Any]]:
```
sphuber, line 200: Shouldn't we be more strict here?
```suggestion
    def serialize(self, mapping: Optional[Dict[str, Port]], breadcrumbs: Sequence[str] = ()) -> Optional[Dict[str, Port]]:
```

aiida/engine/processes/process_spec.py
```diff
@@ -76,22 +78,36 @@ def exit_code(self, status, label, message, invalidates_cache=False):
 
         self._exit_codes[label] = ExitCode(status, message, invalidates_cache=invalidates_cache)
 
+    # override return type to aiida's PortNamespace subclass
+
+    @property
+    def ports(self) -> PortNamespace:
```
sphuber, line 84: Hmm, not sure about this. You have to change it here _because_ the port namespace type is changeable through the `PORT_NAMESPACE_TYPE` class attribute. If this were to be changed dynamically, the typing would again be wrong. How is this typically dealt with in `mypy`? You cannot do variable dereferencing I suppose, i.e. `-> 'ProcessSpec.PORT_NAMESPACE_TYPE'`?

aiida/engine/processes/workchains/restart.py
```diff
@@ -101,50 +112,73 @@ def handle_problem(self, node):
...
         """Define the process specification."""
-        # yapf: disable
         super().define(spec)
-        spec.input('max_iterations', valid_type=orm.Int, default=lambda: orm.Int(5),
-            help='Maximum number of iterations the work chain will restart the process to finish successfully.')
-        spec.input('clean_workdir', valid_type=orm.Bool, default=lambda: orm.Bool(False),
-            help='If `True`, work directories of all called calculation jobs will be cleaned at the end of execution.')
-        spec.input('handler_overrides',
-            valid_type=orm.Dict, required=False, validator=functools.partial(validate_handler_overrides, cls),
+        spec.input(
```
sphuber, line 130: It is good that we apply yapf too as much of the file as possible, but I would reinstate the original formatting for the `define` method here and simply add `#yapf: enable` at the end. The goal of the formatting is too make things more readable and frankly I find the formatting by yapf grotesque here. Everything is different and makes it difficult to read I find. Would it be ok to revert the formatting for the just the `define` method here?

aiida/engine/processes/workchains/restart.py
```diff
@@ -202,10 +236,11 @@ def inspect_process(self):  # pylint: disable=too-many-branches
         last_report = None
 
         # Sort the handlers with a priority defined, based on their priority in reverse order
-        for handler in sorted(self.get_process_handlers(), key=lambda handler: handler.priority, reverse=True):
+        get_priority = lambda handler: handler.priority
+        for handler in sorted(self.get_process_handlers(), key=get_priority, reverse=True):
```
sphuber, line 240: why this change?

aiida/sphinxext/process.py
```diff
@@ -117,9 +117,12 @@ def build_content(self):
...
         if hasattr(self.process_spec, 'get_outline'):
-            outline = self.process_spec.get_outline()
-            if outline is not None:
-                content += self.build_outline_doctree(outline=outline)
+            try:
+                outline = self.process_spec.get_outline()
+                if outline is not None:
+                    content += self.build_outline_doctree(outline=outline)
+            except AssertionError:
+                pass
```
sphuber, line 125: Why is this necessary?

aiida/engine/processes/process.py
```diff
@@ -139,56 +169,54 @@ def get_exit_statuses(cls, exit_code_labels):
...
 
         :return: metadata dictionary
-        :rtype: dict
+
         """
         try:
+            assert self.inputs is not None
             return self.inputs.metadata
-        except AttributeError:
+        except (AssertionError, AttributeError):
```
sphuber, line 216: why the `AssertioneError`?

aiida/engine/processes/process.py
sphuber: seems correct that it was the wrong type, but not sure if constructing a state like this is correct. @muhrin should this work as intended?

sphuber (commented):
Thanks @chrisjsewell . I have gone over the changes, but have not checked all types in detail. I spotted some that could probably be made more specific, but don't have the time to check all in detail. Regarding the removal of yapf disabling, that is great, except for in the `define` methods of processes. I feel the new formatting is actually not more legible at all, which is of course its goal. I would vote to restore that the original formatting and just put `# yapf: enable` at end of the method. Also bunch of empty lines in the docstrings where you removed a typed out param or return type.

Regarding the two fixes, I am not sure how these bugs could have manifested themselves and therefore whether they can be linked to any of the open issues.

chrisjsewell:
cheers @sphuber!

>  I feel the new formatting is actually not more legible at all

Oh yeh well thats yapf is generally just shit lol, the black formatting is so much better and more consistent. But anyhow yeh I will revert (probably by running black on the file, then adding the yapf disables, then running yapf)

> Also bunch of empty lines in the docstrings where you removed a typed out param or return type.

yeh I left a line between `:return:` and `"""`  on purpose, I find it more readable

> Regarding the two fixes, I am not sure how these bugs could have manifested themselves and therefore whether they can be linked to any of the open issues.

Indeed, I guess we will just have to retry some of these issues after merging

aiida/engine/processes/calcjobs/manager.py
chrisjsewell: I just find it more readable to have a blank line before `"""`

aiida/engine/processes/calcjobs/manager.py
chrisjsewell, line 199:
```suggestion
```

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: yeh just for cyclic dependencies. I can't remember if I specifically tested this one, but generally erred on the side of caution

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -11,19 +11,27 @@
...
-from aiida.engine.utils import exponential_backoff_retry, interruptable_task
+from aiida.engine.transports import TransportQueue
+from aiida.engine.utils import exponential_backoff_retry, interruptable_task, InterruptableFuture
 from aiida.schedulers.datastructures import JobState
 from aiida.manage.configuration import get_config_option
 
 from ..process import ProcessState
 
+if TYPE_CHECKING:
+    from aiida.orm.nodes.process.calculation.calcjob import CalcJobNode
```
chrisjsewell, line 32:
```suggestion
from aiida.orm.nodes.process.calculation.calcjob import CalcJobNode
from aiida.schedulers.datastructures import JobState
from aiida.manage.configuration import get_config_option

from ..process import ProcessState

if TYPE_CHECKING:
```

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -100,7 +108,7 @@ async def do_upload():
         return skip_submit
 
 
-async def task_submit_job(node, transport_queue, cancellable):
+async def task_submit_job(node: 'CalcJobNode', transport_queue: TransportQueue, cancellable: InterruptableFuture):
```
chrisjsewell, line 111:
```suggestion
async def task_submit_job(node: CalcJobNode, transport_queue: TransportQueue, cancellable: InterruptableFuture):
```

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -145,7 +157,7 @@ async def do_submit():
         return result
 
 
-async def task_update_job(node, job_manager, cancellable):
+async def task_update_job(node: 'CalcJobNode', job_manager, cancellable: InterruptableFuture):
```
chrisjsewell, line 160:
```suggestion
async def task_update_job(node: CalcJobNode, job_manager, cancellable: InterruptableFuture):
```

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -205,7 +221,10 @@ async def do_update():
         return job_done
 
 
-async def task_retrieve_job(node, transport_queue, retrieved_temporary_folder, cancellable):
+async def task_retrieve_job(
+    node: 'CalcJobNode', transport_queue: TransportQueue, retrieved_temporary_folder: str,
```
chrisjsewell, line 225:
```suggestion
    node: CalcJobNode, transport_queue: TransportQueue, retrieved_temporary_folder: str,
```

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -264,7 +288,7 @@ async def do_retrieve():
         return result
 
 
-async def task_kill_job(node, transport_queue, cancellable):
+async def task_kill_job(node: 'CalcJobNode', transport_queue: TransportQueue, cancellable: InterruptableFuture):
```
chrisjsewell, line 291:
```suggestion
async def task_kill_job(node: CalcJobNode, transport_queue: TransportQueue, cancellable: InterruptableFuture):
```

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: yep, they fail mypy

aiida/engine/processes/ports.py
chrisjsewell: this isn't correct though, the values are not ports, they are data nodes, mappings, or well any value

aiida/engine/processes/process.py
chrisjsewell: to pick up the `assert is not None`

aiida/engine/processes/process_spec.py
chrisjsewell: > You cannot do variable dereferencing

no, not to my knowledge

the point is though that `PORT_NAMESPACE_TYPE` can no longer be any subclass of the plumpy `PortNamespace`, because the aiida `Process` requires it to have a `serialize` method. So even if you dynamically change it, it still has to be a subclass of the aiida `PortNamespace`.

aiida/engine/processes/workchains/restart.py
chrisjsewell: because `handler` is a wrapped function with dynamically added attributes (in this case `priority`). There is no way to properly type these (same with decorated function processes) and so mypy was failing

aiida/sphinxext/process.py
chrisjsewell: because `get_outline` can no longer return `None` (required by plumpy typing), instead it raises an assertion error if the outline is not yet loaded

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: @sphuber do you think then that all aiida.orm imports in aiida/engine should be at the top of the file, since at present there is a number within functions/methods.

Note, I checked in `aiida/orm` and there is no `aiida.engine` imports except inside `aiida.orm.querybuilder.get_querybuilder_classifiers_from_cls` and inside `aiida.orm.utils.links.LinkManager.nested`

aiida/engine/processes/process.py
ltalirz: That's not for this PR, but should aiida-core be tampering with this`_state` attribute and the process state classes directly?
This seems like an example where plumpy might want to offer an API that avoids this (it looks like `_state` is used only used in 3 places in aiida-core) and decouple the implementation a bit.

aiida/engine/processes/calcjobs/tasks.py
ltalirz: that's an interesting catch. 

Does this mean that when this is being encountered, somewhere (in the logs?) there should be an error message like
```python
In [4]: class A():
   ...:     def __init__(self):
   ...:         self._killing = None
   ...:
   ...:     def b(self):
   ...:         return self._killing.set_result(True)
   ...:
   ...:

In [5]: a=A()

In [6]: a.b()
<ipython-input-6-4f53a1908077> in <module>
----> 1 a.b()

<ipython-input-4-3b0911a3e651> in b(self)
      4
      5     def b(self):
----> 6         return self._killing.set_result(True)
      7
      8

AttributeError: 'NoneType' object has no attribute 'set_result'
```
`set_result` [is not mentioned in the aiida issues](https://github.com/aiidateam/aiida-core/issues?q=is%3Aissue+is%3Aopen+set_result) but I guess one could still check whether any of the [kill issues](https://github.com/aiidateam/aiida-core/issues?q=is%3Aissue+is%3Aopen+kill) are resolved by this.

ltalirz: P.S. Bingo

```
aiida-dev_aiida.log.4:AttributeError: 'NoneType' object has no attribute 'set_result'
aiida-dev_aiida.log.4:AttributeError: 'NoneType' object has no attribute 'set_result'
aiida-dev_aiida.log.4:AttributeError: 'NoneType' object has no attribute 'set_result'
```

chrisjsewell: how did you get that to happen @ltalirz?

aiida/engine/processes/process.py
chrisjsewell: yeh, its a little misleading that `Process.state` returns the label and not the actual state 😬 but probably "too late" to change that

ltalirz: Well, if I understand correctly, then the actual process state class is only used in three places in aiida-core.
If one would delegate any operation on this class to plumpy, then the confusion about what type of process state one needs to deal with in aiida-core would go away (it would always just be dealing with labels).

aiida/engine/processes/calcjobs/tasks.py
ltalirz: See above. It seems to have happened in a child process being killed by a parent process (probably following a manual `verdi process kill` command).

chrisjsewell: so possibly the cause of #3776 and #4298

aiida/engine/processes/process.py
muhrin: Yes, setting to `ProcessState.FINISHED` was not right.  The general logic to `_state` was that, I really wanted to push clients (even subclasses) to only use `transition_to` for state transitions as this way we could try and guarantee that state transition hooks would be called (e.g. `on_exiting/entering`) which are legtimate places for clients to put code that they always expect to be hit when transitionining. _However_, there need to be exceptions when deserialising a process (e.g in `load_instance_state`) because if it was serialised in that state then the transition already happened at some point and it's OK to just set the `_state` directly.

Sorry for the sloppiness in `this.state` actually returning the state label, this was just me trying to be economical and avoid having to call it `state_label` but I recognise it's confusing when working on the internals of a `Process`.

Anyway, if I'm understanding this code segment it's OK to set the state like this because if the node says it's done, presumably at some point the correct state transition did happen.  That said, I'm not sure how this code state could be reached in the first place (and perhaps it never was if this didn't cause issues before).

@ltalirz , yes, no problem to have a `_set_state()`  method so long as there is a _BIG_ warning to never call it except for the situations I mentioned above.

aiida/engine/processes/calcjobs/tasks.py
muhrin: Yes, nice.  It's _probably_ a bit of a problem if the `future` is `None` so could be logged as a warning but with concurrency it's difficult to think through all the possible paths that could lead to that situation.

aiida/engine/processes/calcjobs/tasks.py
sphuber: I think indeed that `orm` should not depend on `engine`. So it should be possible to move all imports of `aiida.orm` in `aiida.engine` to the top and not have circular import problems. And unless this is a well-accepted standard practice in `mypy` to use the `if TYPE_CHECKING` guard if you only import something for typing, then I would get rid of it and just import it normally. I think the overhead should be minimal and it is less clutter

chrisjsewell: sounds good to me 👍

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: added a log warning (8151a83)

aiida/engine/processes/calcjobs/manager.py
chrisjsewell: done

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: done (498dd02)

aiida/engine/processes/workchains/restart.py
chrisjsewell: done (24140ade25bc5add3fe3f34dadb5d4e609b09e30)

chrisjsewell:
right that's the whole of `aiida.engine` typed. I will be merging this by Wednesday, so reviews should be done before then

ltalirz:
No objections from my side ;-) I'll let Seb give the final ok here; if you need my review, just ping me (I'll unsubscribe for now)

aiida/engine/processes/ports.py
sphuber: of course, I misread the code there. Please ignore

aiida/engine/processes/workchains/restart.py
sphuber: Could have used a comments perhaps, because I am sure I will at some point try to make it a one-liner again and be faced with a failing pre-commit. Fine to leave as is though

aiida/engine/processes/process_spec.py
sphuber: yes, but even if you define another `PORT_NAMESPACE_TYPE` that is a subclass of `PortNamespace`, you also have to override the properties again just to get the type hinting correct? That seems like problematic behavior for a use case that should be quite common. Is this a known limitation of mypy at least? I won't make this stop this PR from being merged but would like to know if these kinds of designs will make type hinting painful

aiida/engine/processes/calcjobs/manager.py
sphuber: That's fine, but that is not really what we have been doing so far in the code base. Ideally we have a pre-commit like `pydocstyle` which I have added for some other repo's, but fixing the errors for `aiida-core` has been to much of an undertaking that I haven't done it so far. Would be really good to add it at some point though. Anyway, I won't make you get rid of all those empty lines for this PR I guess.

aiida/engine/processes/workchains/restart.py
sphuber: Cheers, much appreciated. Would you please do the same for the `CalcJob` spec definition?

sphuber:
@chrisjsewell alright, I have gone through your comments and resolved all of them. I would just appreciate it if you could undo the reformatting of the `CalcJob` spec just as you already did for the `BaseRestartWorkChain`. If you do that and update the branch I will merge this PR. Thanks

aiida/engine/processes/process_spec.py
chrisjsewell: >  Is this a known limitation of mypy at least?

Err well personally I would say it is more highlighting a "problematic" code design, than a limitation of mypy.
Its fine if you are just using `PORT_NAMESPACE_TYPE` to define different concrete implementations of the same abstract class (with the same methods), then you would not need these method overrides.
But when you are using it to define subclasses that do not adhere to the base class (e.g. adding new methods) I am not sure that is an ideal design for dependency injection.

aiida/engine/processes/workchains/restart.py
chrisjsewell: done

aiida/engine/processes/process_spec.py
sphuber: Wait, maybe I am misunderstanding, but we _are_ returning a subclass of the original `PortNamespace`, or is that not the problem? Surely that should be fine as a design. I agree that it should respect the same interface as a minimum, but why would it be bad to _add_ methods? I agree that removing them is a bad idea

aiida/engine/processes/workchains/restart.py
chrisjsewell: fair, can't be bothered now though lol
technically I find these design patterns, where you are dynamically adding a new method a bit problematic, same with how the function processes are implemented. But I see the thinking behind it: being able to use `functools.wrap`

aiida/engine/processes/process_spec.py
chrisjsewell: Because `aiida.engine.Process` has fundamentally changed the required interface of the `PortNamespace` (by requiring the `serialize` method), i.e. the`aiida.engine.Process` will fail if you set its `PORT_NAMESPACE_TYPE = plumpy.PortNamespace`.

chrisjsewell:
I think you've pushed github actions to its limit, adding those rabbitmq tests lol; always at least one job now seems to stall then timeout

sphuber (approved):
Alright, let's get this show on the road

PR #4670, opened 2021-01-22
Feat: Improve error message for unconfigured computers

### LINKED ISSUE #4645: Improve exception message when running on unconfigured computer
When trying to run with a code on an unconfigured computer, the error is quite generic: `NotExistent: No result was found`.

The last part of the traceback follows:
```
[...]
~/git/aiida-core/aiida/engine/utils.py in execute_coroutine()
    107         """Coroutine that wraps the original coroutine and sets it result on the future only if not already set."""
    108         try:
--> 109             result = await coro(future)
    110         except Exception as exception:  # pylint: disable=broad-except
    111             if not future.done():

~/git/aiida-core/aiida/engine/processes/calcjobs/tasks.py in task_upload_job(process, transport_queue, cancellable)
     64     max_attempts = get_config_option(MAX_ATTEMPTS_OPTION)
     65 
---> 66     authinfo = node.computer.get_authinfo(node.user)
     67 
     68     async def do_upload():

~/git/aiida-core/aiida/orm/computers.py in get_authinfo(self, user)
    591         from . import authinfos
    592 
--> 593         return authinfos.AuthInfo.objects(self.backend).get(dbcomputer_id=self.id, aiidauser_id=user.id)
    594 
    595     def is_user_configured(self, user):

~/git/aiida-core/aiida/orm/entities.py in get(self, **filters)
    134         """
    135         res = self.query(filters=filters)
--> 136         return res.one()[0]
    137 
    138     def find(self, filters=None, order_by=None, limit=None):

~/git/aiida-core/aiida/orm/querybuilder.py in one(self)
   2181             raise MultipleObjectsError('More than one result was found')
   2182         elif len(res) == 0:
-> 2183             raise NotExistent('No result was found')
   2184         return res[0]
   2185 

NotExistent: No result was found
```

I think we can catch the `NotExistent` exception in `~/git/aiida-core/aiida/engine/processes/calcjobs/tasks.py` at line: 
```
---> 66     authinfo = node.computer.get_authinfo(node.user)
```
and raise a better AiiDA exception with a more sensible message, like "Computer XXX not configured for user YYY - use `verdi computer configure` first".

ramirezfranciscof:
Fixes #4645

I added a check to the `get_authinfo` method of the computer class that improves on the error raised by the call to `authinfos.AuthInfo.objects(...).get(...)` when a (computer, user) couple is not found for the requested case. Also added a test that checks for the correct error raised, and also that the critical pieces of information are present in the error message.

### DIFF
--- a/aiida/orm/computers.py
+++ b/aiida/orm/computers.py
@@ -600,7 +600,15 @@ class Computer(entities.Entity):
         """
         from . import authinfos
 
-        return authinfos.AuthInfo.objects(self.backend).get(dbcomputer_id=self.id, aiidauser_id=user.id)
+        try:
+            authinfo = authinfos.AuthInfo.objects(self.backend).get(dbcomputer_id=self.id, aiidauser_id=user.id)
+        except exceptions.NotExistent:
+            raise exceptions.NotExistent(
+                f'Computer `{self.label}` (ID={self.id}) not configured for user `{user.get_short_name()}` '
+                f'(ID={user.id}) - use `verdi computer configure ` first'
+            )
+
+        return authinfo
 
     def is_user_configured(self, user):
         """
--- a/aiida/orm/computers.py
+++ b/aiida/orm/computers.py
@@ -602,11 +602,11 @@ class Computer(entities.Entity):
 
         try:
             authinfo = authinfos.AuthInfo.objects(self.backend).get(dbcomputer_id=self.id, aiidauser_id=user.id)
-        except exceptions.NotExistent:
+        except exceptions.NotExistent as exc:
             raise exceptions.NotExistent(
                 f'Computer `{self.label}` (ID={self.id}) not configured for user `{user.get_short_name()}` '
-                f'(ID={user.id}) - use `verdi computer configure ` first'
-            )
+                f'(ID={user.id}) - use `verdi computer configure` first'
+            ) from exc
 
         return authinfo
 
--- a/tests/orm/test_computers.py
+++ b/tests/orm/test_computers.py
@@ -103,3 +103,19 @@ class TestComputerConfigure(AiidaTestCase):
 
         with self.assertRaises(ValueError):
             comp.configure(username='radames', invalid_auth_param='TEST')
+
+    def test_non_configure_error(self):
+        """Configure a computer for local transport and check it is configured."""
+        self.comp_builder.label = 'test_non_configure_error'
+        self.comp_builder.transport = 'local'
+        comp = self.comp_builder.new()
+        comp.store()
+
+        with self.assertRaises(exceptions.NotExistent) as exc:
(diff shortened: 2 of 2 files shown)


ramirezfranciscof:
@sphuber I still didn't add the [check you wanted](https://github.com/aiidateam/aiida-core/issues/4645#issuecomment-754885539) because I have a problem. In principle I would just need to call `computer.get_authinfo( <SOME_USER> )` right after [this line](https://github.com/aiidateam/aiida-core/blob/61e48d7a3dac0bc9956f8ac34b8e3cc19db1fc3e/aiida/engine/processes/calcjobs/calcjob.py#L78).

However, I'm not sure what user to provide or how to access it. In the case of the call inside the engine (the error shown in the issue #4645), the task was provided with the `process` as an input argument so it can get its underlying calcjob node and the user it belongs to. In the validator, however, I have no acces to the node (I only get the inputs and the context, but I couldn't find the class definition of the context to check if it has a reference to the calcjob node). Is there a way to either get that node or get the user from the profile that is currently loaded?

ltalirz:
Also these test failures seem to have been caused by the Open SSH upgrade - please continue with review/approval

aiida/orm/computers.py
```diff
@@ -590,7 +590,15 @@ def get_authinfo(self, user):
         """
         from . import authinfos
 
-        return authinfos.AuthInfo.objects(self.backend).get(dbcomputer_id=self.id, aiidauser_id=user.id)
+        try:
+            authinfo = authinfos.AuthInfo.objects(self.backend).get(dbcomputer_id=self.id, aiidauser_id=user.id)
+        except exceptions.NotExistent:
+            raise exceptions.NotExistent(
+                f'Computer `{self.label}` (ID={self.id}) not configured for user `{user.get_short_name()}` '
+                f'(ID={user.id}) - use `verdi computer configure ` first'
```
giovannipizzi, line 598:
```suggestion
                f'(ID={user.id}) - use `verdi computer configure` first'
```

giovannipizzi: Question I am not sure of, and maybe @sphuber can help - should this `raise` be a `raise from` or it's not needed?

giovannipizzi (commented):
I have just only minor typo a dn one doubt where @sphuber might help, otherwise looks good to me

aiida/orm/computers.py
sphuber: it is not necessary, bu it is recommended, and in fact `pylint` should warn for this. So I wonder why that is not the case.

ramirezfranciscof: Not sure what you mean by needed, but I agree it would be better.

giovannipizzi (approved):
Good to go!

PR #4671, opened 2021-01-22
Fix: Command `play all` only affects active processes.

### LINKED ISSUE #4528: `verdi process play --all` should filter for active processes only
Currently it only filters on the `paused` attribute. It will also find terminated processes, that due to another bug, still have the `paused` attribute. Filtering on active states only should speed up the query anyway.

ramirezfranciscof:
Fixes #4528 

The command `verdi process play -all` now will only affect active (`created`, `waiting`, `running`) nodes.


ramirezfranciscof:
@sphuber I wanted to add some tests for the functionality but...

1. The `test_pause_play_kill` does some extensive and specific setting up that I'm not sure how to adapt to get the kind of terminated process you wanted to avoid querying.
2. I would like to check that neither (a) non-paused active processes nor (b) paused non-active processes are affected by this. For (b) I guess it would trigger the `echo.echo_error(f'Process<{process.pk}> is already terminated')`, although I'm not sure how to assert for that message. But for (a), I'm not even sure I can notice that something was attempted (I can't find the definition for the `play_process` method of the controller, but I would assume if I try to play a process that is already not paused it will just ignore, right? or will it raise?).

The test for `test_pause_play_kill` passes, so at least it is already "unpausing" the processes we care about, but I'm not sure how critical it is to check that it is also not trying to "unpause" other ones.

tests/cmdline/commands/test_process.py
```diff
@@ -123,6 +132,26 @@ def test_pause_play_kill(self):
...
+        self.assertTrue(calc.paused)
+
+        result = self.cli_runner.invoke(cmd_process.process_play, ['--wait'])
+        self.assertIsNone(result.exception, result.output)
+        self.assertFalse(calc.paused)
+
+        result = self.cli_runner.invoke(cmd_process.process_kill, ['--wait', str(calc.pk)])
+        self.assertIsNone(result.exception, result.output)
+        self.assertTrue(calc.is_terminated)
+        self.assertTrue(calc.is_killed)
```
sphuber, line 153: Am I missing something or does this test not really pertain to the changes in the code? You are simply adding this as an additional test. Fine in principle, just to make sure.

On a side note, given the high degree of overlap with `test_pause_play_kill` these two would be a perfect candidate for `pytest.mark.parametrize`.

sphuber (changes requested):
Thanks @ramirezfranciscof . The change itself is fine. Regarding the tests, you could create a terminated process node and set the paused flag and then you check that after the command invocation it is still paused. This is a bit artificial though and I don't think we really need to test this

tests/cmdline/commands/test_process.py
ramirezfranciscof: Ah, sorry. Now I see I forgot to add the `-all` flag.

> On a side note, given the high degree of overlap with `test_pause_play_kill` these two would be a perfect candidate for `pytest.mark.parametrize`.

Mmm yeah, I agree, but this test is using the old testing structure. Can I use pytest fixtures on `AiidaTestCase` methods?

sphuber: no, but you can start the migration to using `pytest`. We have been doing this bit by bit whenever we touch tests that are in the old style. Would be good opportunity to do it for this file

ltalirz:
Looks to me like the test failures were caused by the openssh upgrade - please continue with review / approval

tests/cmdline/commands/test_process.py
ramirezfranciscof: Ah, I see. Haha, that is some sneaky way to suggest to migrate the test :sweat_smile:...

But yeah, I gave it a shot. It was not easy because I have to try a couple of things to know what parts of the `setUp` and `tearDown` where necessary with the daemon being a relevant part of the test, but I think I manage to find a configuration that at least seems to work.

Ironically, in the end the test was not such a good candidate for `pytest.mark.parametrize` because the parameter actually relied on the pk of a node that has to be created during the test. Anyways I did parametrize using a boolean, but it was not as clean as originally expected.

ramirezfranciscof:
@sphuber My only worry with migrating the test is that I'm getting a strange memory error that I'm not sure how to handle. Maybe you would have a better idea if I'm forgetting to close down anything related to the Daemon?

```bash
=================================================== test session starts ===================================================
platform darwin -- Python 3.8.6, pytest-6.1.2, py-1.9.0, pluggy-0.13.1
benchmark: 3.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/workuser/Codes/aiida-core, configfile: pyproject.toml
plugins: rerunfailures-9.1.1, cov-2.10.1, asyncio-0.14.0, benchmark-3.2.3, timeout-1.4.2
collected 10 items                                                                                                        

tests/cmdline/commands/test_process.py ..........                                                                   [100%]

==================================================== warnings summary =====================================================
tests/cmdline/commands/test_process.py::test_pause_play_kill[True]
tests/cmdline/commands/test_process.py::test_pause_play_kill[True]
tests/cmdline/commands/test_process.py::test_pause_play_kill[True]
tests/cmdline/commands/test_process.py::test_pause_play_kill[False]
tests/cmdline/commands/test_process.py::test_pause_play_kill[False]
tests/cmdline/commands/test_process.py::test_pause_play_kill[False]
  /Users/workuser/.virtualenvs/aiida-full/lib/python3.8/site-packages/aio_pika/message.py:714: DeprecationWarning: Use "async with message.process()" instead
    warn('Use "async with message.process()" instead', DeprecationWarning)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
============================================= 10 passed, 6 warnings in 12.46s =============================================
/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py:641: RuntimeWarning: coroutine 'create_task.<locals>.run_task' was never awaited
Coroutine created at (most recent call last)
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 603, in run_until_complete
    self.run_forever()
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 570, in run_forever
    self._run_once()
  File "/Users/workuser/.virtualenvs/aiida-full/lib/python3.8/site-packages/nest_asyncio.py", line 132, in _run_once
    handle._run()
  File "/Users/workuser/.virtualenvs/aiida-full/lib/python3.8/site-packages/nest_asyncio.py", line 201, in run
    ctx.run(self._callback, *self._args)
  File "/Users/workuser/.virtualenvs/aiida-full/lib/python3.8/site-packages/nest_asyncio.py", line 159, in step
    step_orig(task, exc)
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/tasks.py", line 280, in __step
    result = coro.send(None)
  File "/Users/workuser/.virtualenvs/aiida-full/lib/python3.8/site-packages/kiwipy/rmq/communicator.py", line 217, in _on_broadcast
    await receiver(
  File "/Users/workuser/.virtualenvs/aiida-full/lib/python3.8/site-packages/kiwipy/rmq/utils.py", line 118, in wrap
    return coro_or_fn(*args, **kwargs)
  File "/Users/workuser/.virtualenvs/aiida-full/lib/python3.8/site-packages/plumpy/communications.py", line 73, in converted
    task_future = futures.create_task(msg_fn, loop)
  File "/Users/workuser/.virtualenvs/aiida-full/lib/python3.8/site-packages/plumpy/futures.py", line 76, in create_task
    asyncio.run_coroutine_threadsafe(run_task(), loop)
  self._ready.clear()
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

ramirezfranciscof:
So @sphuber , state of the union:

1. I finished moving the test to pytest format.
2. We [recently disabled](https://github.com/aiidateam/aiida-core/pull/4747) this test but I comment out the `pytest.mark.skip` to see how the tests perform here. I'm going to un-comment it once you see this and give me the green light.
3. The reason we skip it is because it has some random fails: it seems to be what it is doing here in the CI (`RuntimeError: Timed out waiting for process to enter waiting state`), but since some of the tests it seems to pass I assume the modifications are ok. The same is true locally for me, it passes half of the time and the other half it times out.
4. Even when it passes, I still get the "Coroutine was never awaited" warnign I describe above.

I think the errors in (3) and (4) are beyond the scope of this PR though. Let me know if you disagree; otherwise this should be ready for review.

tests/cmdline/commands/test_process.py
```diff
@@ -490,3 +395,100 @@ def test_multiple_processes(self):
...
+    # To simulate a process without a corresponding task, we simply create a node and store it. This node will not
+    # have an associated task at RabbitMQ, but it will be a valid `ProcessNode` so it will pass the initial
+    # filtering of the `verdi process` commands
+    orphaned_node = WorkFunctionNode().store()
+    non_existing_process_id = str(orphaned_node.pk)
+    for command in [process_pause, process_play, process_kill]:
+        result = cli_runner.invoke(command, [non_existing_process_id])
+        import traceback
+        assert result.exception is None, ''.join(traceback.format_exception(*result.exc_info))
+        assert 'Error:' in result.output
```
sphuber, line 451: For this to still work, it does need to explicitly set the `process_state` to an active state. Now this node will never even be matched in the query in the command because you added filters on the `process_state`.

tests/cmdline/commands/test_process.py
```diff
@@ -490,3 +395,100 @@ def test_multiple_processes(self):
...
+    result = cli_runner.invoke(process_play, ['--wait', cmd_option])
+    assert result.exception is None, result.output
+    assert not calc.paused
+
+    result = cli_runner.invoke(process_kill, ['--wait', str(calc.pk)])
+    assert result.exception is None, result.output
+    assert calc.is_terminated
+    assert calc.is_killed
+
+    os.kill(daemon.pid, signal.SIGTERM)
```
sphuber, line 494: This cleanup will now sometimes not be called if during the test an error occurs. We should create a fixture, let's say `with_daemon` that looks a bit like:
```python
@pytest.fixture
def with_daemon():
    # start the daemon process
    yield
    os.kill(daemon.pid, signal.SIGTERM)
```
This way the `kill` will be guaranteed to be called

sphuber (changes requested):
Thanks @ramirezfranciscof . Two comments. Would be fine to merge this with the test disabled, after you confirmed that it passes at least sometimes locally and tackle the issue in another PR. Would be good if you could give it a shot though now that you have been working on this code. @chrisjsewell has recently dealt a lot with tests failing to await futures as is apparently the problem here. I think he fixed one of these recently by explicitly awaiting futures at the end of the test. Also the fixes to the daemon runner may have an effect on this. Would be good if you could check together to see if things are related.

ramirezfranciscof:
Ok, I'll try to take a look at this with @chrisjsewell after the release.

Please check that I applied the changes correctly; I'm still a bit unsure of what all the procedures prior to doing the actual testing (that seems [only be this part](https://github.com/aiidateam/aiida-core/pull/4671/files#diff-a1ae4b7a9a47fb5de6130c0c4d940b2aa8bb9cb8a3e15bc88a5753b197ac06a0R458-R476)) are doing or why are they necessary. When transforming this to pytest format I tried to leave that whole part alone, integrating the `setUp` and `tearDown`,  but only further adapted the test section for the checks I was interested in.

tests/cmdline/commands/test_process.py
```diff
@@ -490,3 +393,84 @@ def test_multiple_processes(self):
         self.assertIn('No callers found', get_result_lines(result)[0])
         self.assertIn(str(self.node_root.pk), get_result_lines(result)[1])
         self.assertIn(str(self.node_root.pk), get_result_lines(result)[2])
+
+
+@pytest.mark.skip(reason='fails to complete randomly (see issue #4731)')
+@pytest.mark.requires_rmq
+@pytest.mark.usefixtures('with_daemon')
+@pytest.mark.usefixtures('clear_database_before_test')
```
sphuber, line 401: You can use a single mark statement to inject fixtures
```suggestion
@pytest.mark.usefixtures('with_daemon', 'clear_database_before_test')
```

tests/cmdline/commands/test_process.py
```diff
@@ -490,3 +393,84 @@ def test_multiple_processes(self):
...
+    Test the pause/play/kill commands
+    """
+    # pylint: disable=no-member, too-many-locals
+    from aiida.cmdline.commands.cmd_process import process_pause, process_play, process_kill
+    from aiida.manage.manager import get_manager
+    from aiida.engine import ProcessState
+    from aiida.orm import load_node
+
+    runner = get_manager().create_runner(rmq_submit=True)
+    cli_runner = CliRunner()
```
sphuber, line 414: I made a specific fixture for running click commands. It is in `tests/cmdline/commands/conftest.py` and is called `run_cli_command`. Search in existing tests to see how it is used. It also makes it a lot easier to check the results. Would be good to use this fixture every time you call a command in this test

sphuber (changes requested):
The `with_daemon` fixture looks ok, if you confirmed that at least once the test works properly I think it is working as intended. Unfortunately I realized one more thing we should quickly fix that I missed the first time around. If you take care of that and confirm the test runs locally when not skipped, we can merge this

tests/cmdline/commands/test_process.py
ramirezfranciscof: Haha, funny thing. Originally when I moved the test to pytest I just removed all the daemon setups and was actually using `run_cli_command`; however when I rebased there were a couple of conflicts with the stuff I removed and since in the middle of this was the problem with the test failing randomly, I didn't want to risk it so [I put it all back](https://github.com/aiidateam/aiida-core/pull/4671/commits/c6fc336b96b4194b0fc329e3b2214a54bab25d5d) the way it was before. Then basically with the `with_daemon` fixture and now with this one I basically took all out again xD...

Anyways: ok, I'll change it, but what do you mean it will "make it a lot easier to check the results"? The way I did it before I just replaced all the `cli_runner.invoke` with `run_cli_command` and removed the setup you are pointing in here, but the asserts I kept all the same no?

sphuber: It provides some additional functionality that makes the checking more succinct, for example checking for an exception, you can replace
```
result = cli_runner.invoke(command, [non_existing_process_id])
import traceback
assert result.exception is None, ''.join(traceback.format_exception(*result.exc_info))
assert 'Error:' in result.output
```
with
```
result = run_cli_command(command, [non_existing_process_id], raises=True)
assert 'Error:' in result.output
```
Also there is an attribute on the `result` called `output_lines` which already explodes the output string on the newline characters which is often something one needs to do in checking. Maybe not in this test though

sphuber: We could also think of adding a `match` argument to it like `pytest.raises` to take a regex to match the exception message against

ramirezfranciscof: Ok, from looking at the code of `run_cli_command`, more specifically this part:

```python
        if raises:
            assert result.exception is not None, result.output
            assert result.exit_code != 0
        else:
            assert result.exception is None, ''.join(traceback.format_exception(*result.exc_info))
            assert result.exit_code == 0, result.output
```

I am assuming I can also remove these lines:

```python
    result = run_cli_command(process_pause, ['--wait', str(calc.pk)])
    assert result.exception is None, result.output # <---- This can be removed
    assert calc.paused
```

However, there is something that feels backwards: the `assert result.exception is not None, result.output` was being used when there was no exception, but inside `run_cli_command` it is called when `raises = True` (and viceverse for line `assert result.exception is None, ''.join(traceback.format_exception(*result.exc_info))`). Am I misunderstanding something or should these be switched?

> We could also think of adding a match argument to it like pytest.raises to take a regex to match the exception message against

Do you mean like, now for this PR =P? This test just does a single check of the output (the one you quoted), so maybe we can leave this change for a change where it would be used more. I could open an issue with a low priority tag.

sphuber: > I am assuming I can also remove these lines:

Yes you can remove the explicit check that the exception is `None` because that is done automatically.

> However, there is something that feels backwards: the `assert result.exception is not None, result.output` was being used when there was no exception, but inside `run_cli_command` it is called when `raises = True` (and viceverse for line `assert result.exception is None, ''.join(traceback.format_exception(*result.exc_info))`). Am I misunderstanding something or should these be switched?

To me the logic in `run_cli_command` makes sense. If the caller says `raises=True` they expect the command to raise, therefore we check that the `exception is not None` and if that is not true (i.e. the exception _was_ `None` and so the command didn't actually fail although it should have) then we print the output so the caller can inspect it to debug. Conversely, if `raises=False`, we make sure that the command did not except or else we show the output.

> > We could also think of adding a match argument to it like pytest.raises to take a regex to match the exception message against
> 
> Do you mean like, now for this PR =P? This test just does a single check of the output (the one you quoted), so maybe we can leave this change for a change where it would be used more. I could open an issue with a low priority tag.

No, not necessary for this PR. Was just an idea for a future feature. Indeed low priority issue would be great.

ramirezfranciscof:
@sphuber Sorry for the delay, changes applied and ready (pre-commit seems to be having some issue with `pymatgen` but this looks like a general problem and unrelated to the PR).

sphuber:
> @sphuber Sorry for the delay, changes applied and ready (pre-commit seems to be having some issue with `pymatgen` but this looks like a general problem and unrelated to the PR).

Think that problem was already dealt with by @ltalirz so if you simply make sure the branch is up to date with develop it should be fixed.

tests/cmdline/commands/test_process.py
ramirezfranciscof: > To me the logic in `run_cli_command` makes sense. If the caller says `raises=True` they expect the command to raise, therefore we check that the `exception is not None` and if that is not true (i.e. the exception _was_ `None` and so the command didn't actually fail although it should have) then we print the output so the caller can inspect it to debug. Conversely, if `raises=False`, we make sure that the command did not except or else we show the output.

Mmm ok, but then the second code snippet on your second comment should not have the `raises = True`, correct? The first snippet seems to be expecting no raise. In the current develop code it is slightly different but also seems to be expecting there to be no raise:

```python
result = self.cli_runner.invoke(command, [non_existing_process_id])
self.assertClickResultNoException(result)
self.assertIn('Error:', result.output)

# Defined elsewhere
def assertClickResultNoException(self, cli_result):  # pylint: disable=invalid-name
        self.assertIsNone(cli_result.exception, ''.join(traceback.format_exception(*cli_result.exc_info)))
```

> No, not necessary for this PR. Was just an idea for a future feature. Indeed low priority issue would be great.

:ok_hand:

ramirezfranciscof:
> the pymatgen issue was due to another breaking change made at a later point
> will be fixed by [#4796 (review)](https://github.com/aiidateam/aiida-core/pull/4796#pullrequestreview-607361020)

These people of pymatgen sure give us trouble uh? =P haha

sphuber (approved):
All good, thanks @ramirezfranciscof

PR #4672, opened 2021-01-24
Add `to_aiida_type` to the public API

chrisjsewell:
Given that it is documented for use by users, it should be in the public API: https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/processes/usage.html#automatic-input-serialization


docs/source/topics/processes/include/snippets/serialize/workchain_serialize.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 from aiida.engine import WorkChain
-from aiida.orm.nodes.data import to_aiida_type
+from aiida.orm import to_aiida_type
 # The basic types need to be loaded such that they are registered with
 # the 'to_aiida_type' function.
 from aiida.orm.nodes.data.base import *
```
chrisjsewell, line 6: is this actually correct? won't these basic types be loaded simply by importing from `aiida.orm`. I've personally never had an issue not using this line

greschd: Needs checking, it's quite possible this is outdated.

sphuber (commented):
Thanks @chrisjsewell . If you could check whether the comment you highlighted is indeed outdated and then remove it, I will approve. Also, you might want to add this function then now to the manually curated list of public API in the docs. I know there are plans to autoamte and improve it, but while its there might as well keep it up to date as much as possible

docs/source/topics/processes/include/snippets/serialize/workchain_serialize.py
chrisjsewell: yeh I've checked and you definitely don't, since they are all imported via the `__all__` mechanism

chrisjsewell, line 6:
```suggestion
```

chrisjsewell:
> Also, you might want to add this function then now to the manually curated list of public API in the docs.

@sphuber added 👍 ready for your approval

ltalirz (approved):
thanks @chrisjsewell 

I checked the last two commits, which do address the comments by @sphuber .
Good to go!

PR #4673, opened 2021-01-24
📚 DOCS: tutorial, add `from aiida.orm import load_code`

chrisjsewell:


ramirezfranciscof:
Do you need to import `load_code`? I thought the verdi shell automatically imported all the "load" methods (`load_code`, `load_node`, etc)

chrisjsewell:
>  I thought the verdi shell automatically imported all the "load" methods

oh yeh fair enough. I feel it makes it a bit unclear where to import it, when you are not in a verdi shell, but a not a big deal

PR #4678, opened 2021-01-26
Fix: pre-store hash for -0. and 0. is now the same

### LINKED ISSUE #4610: make_hash: positive and negative zero have different hashes
The `make_hash` function produces different results for positive (``0.``) and negative zero (``-0.``) floats:

```python
In [1]: from aiida.common.hashing import make_hash

In [2]: make_hash(0.)
Out[2]: '20a995d105ee978f320718c436d64a725454f61a3684b4751b34756e04d58365'

In [3]: make_hash(-0.)
Out[3]: '7c8fd8f1e2ffeffd8938d846a8aa1be14780cc8b6d003cd6fe124cb30ccb2cc8'
```

The code responsible for hashing floats is 

https://github.com/aiidateam/aiida-core/blob/e303cc450d3dd1e949de39785cbc141aa340ac70/aiida/common/hashing.py#L194-L200

and 

https://github.com/aiidateam/aiida-core/blob/e303cc450d3dd1e949de39785cbc141aa340ac70/aiida/common/hashing.py#L283-L292

The "truncation" to a specific number of significant digits here is to ensure consistent hashes before and after storing, because floats can be slightly changed when stored to the database. In this case however, this goal isn't met:

```python
In [2]: from aiida import orm

In [3]: f = orm.Float(-0.)

In [4]: f
Out[4]: <Float: uuid: 1723092d-5e30-46d3-b72a-b9957d5479ec (unstored) value: -0.0>

In [5]: f._get_hash()
Out[5]: '446391aa2fa4107ce1a3640eb96d4b3fcc2c4e107925362e9616eb3af0bd1b93'

In [6]: f.store()
Out[6]: <Float: uuid: 1723092d-5e30-46d3-b72a-b9957d5479ec (pk: 252658) value: 0.0>

In [7]: f.get_hash()
Out[7]: '09193d9becebc2ebb09dae8102f02a04aa82478ac7507c6d9b55d253ed7167f6'
```

Now technically `-0.` and `0.` are different objects, and it's valid (but not required) that they hash differently. However, since the storing converts `-0.` to `0`, I feel it would be more practical if they had the same hash. This could be achieved e.g. by adding 
```python
if value == 0:
    value = 0.
```
to `float_to_text`.

I believe this won't need rehashing of the DB, because _when stored_ the zeroes are already converted, and this only affects unstored values. Not 100% sure there isn't some edge case here, though.

ramirezfranciscof:
Fixes #4610

The `-0.` is converted to `0.` in the process of storing the node in the DB, but this was not being considered by the `aiida/common/hashing.py : make_hash` method. This led to `aiida/orm/nodes/node.py : _get_hash` would return a different hash for unstored and stored `Float` nodes with content `-0.`

### DIFF
--- a/aiida/common/hashing.py
+++ b/aiida/common/hashing.py
@@ -288,5 +288,7 @@ def float_to_text(value, sig):
     :param value: the float value to convert
     :param sig: choose how many digits after the comma should be output
     """
+    if value == 0:
+        value = 0.  # Identify value of -0. and overwrite with 0.
     fmt = f'{{:.{sig}g}}'
     return fmt.format(value)
--- a/tests/common/test_hashing.py
+++ b/tests/common/test_hashing.py
@@ -36,6 +36,7 @@ class FloatToTextTest(unittest.TestCase):
     """
 
     def test_subnormal(self):
+        self.assertEqual(float_to_text(-0.00, sig=2), '0')  # 0 is always printed as '0'
         self.assertEqual(float_to_text(3.555, sig=2), '3.6')
         self.assertEqual(float_to_text(3.555, sig=3), '3.56')
         self.assertEqual(float_to_text(3.141592653589793238462643383279502884197, sig=14), '3.1415926535898')


ramirezfranciscof:
It is still not ideal than the logic to convert values into storable values is repeated in `aiida/orm/nodes/node.py` and `aiida/common/hashing.py`. Ideally it shoud be in a single place and both should be using that single point of reference to convert their values, correct?

If you agree, then I can open an issue for this.

greschd:
> 
> It is still not ideal than the logic to convert "raw values" into "storable values" is repeated in `aiida/orm/nodes/node.py` and `aiida/common/hashing.py`. Ideally it shoud be in a single place and both should be using that single point of reference to convert their values, correct?

I'm not sure they should be combined. The hashing method does _more_ than just converting to a "storable" value, for floats it also truncates them to a fixed precision s.t. small numeric differences result in the same hash. That definitely _shouldn't_ be part of the conversion before storing.
So, since the functions are subtly different, maybe it is better to keep them separate.

aiida/common/hashing.py
```diff
@@ -288,5 +288,7 @@ def float_to_text(value, sig):
     :param value: the float value to convert
     :param sig: choose how many digits after the comma should be output
     """
+    if value == 0:
```
greschd, line 291: Maybe we should add a comment saying why this is necessary. Just looking at the code, it seems superfluous.

greschd (approved):

greschd:
Hmm, should we maybe add a test for this?

greschd (approved):

greschd:
Thanks @ramirezfranciscof!

ramirezfranciscof:
@greschd thanks for the review! I'll finish merging when the test pass.

> I'm not sure they should be combined. The hashing method does more than just converting to a "storable" value, for floats it also truncates them to a fixed precision s.t. small numeric differences result in the same hash. That definitely shouldn't be part of the conversion before storing.

Uhm, I see what you mean. However I still believe that the hashing has to consider how the storage of the nodes will change the values, right? So to speak, it may not make sense to hash up to the 10th decimal if we can't ensure the store will respect beyond the 8th decimal. So maybe the hash can add requirements above the generic "clean_value", but still has to rely on that beforehand.

I mean, we can also go the route of having the criteria of value hashing and value storing be independent and just achieve consistent via only ensuring the hash of stored nodes. But then the reason of this PR is actually ensuring coherence of the pre-store and after-store hash, so there is some expectation that the hash will know how the store may change the number, right? :thinking:

greschd:
> But then the reason of this PR is actually ensuring coherence of the pre-store and after-store hash, so there is some expectation that the hash will know how the store may change the number, right?

Yes, the hashing is definitely made with that in mind. The problem is that we can't really do this float transformation in pure Python, that is something that is implemented in PostgreSQL. It's just a consequence of converting from Python types to postgres types ([see their documentation](https://www.postgresql.org/docs/13/datatype-numeric.html)) and back.

The reduction in float precision before hashing is made to trade off false negatives (where the hash changes after a store round-trip) with false positives (the hash should only be equal if the values can reasonably be considered the same).

As far as I understand, the purpose of `clean_value` is different: it should change the data to make it _storable_ in the database, but doesn't (need to) take into account how the values may be changed once they are actually stored.
If `clean_value` changes the _types_ of data (e.g., converting a `tuple` to a `list` or similar) it might make sense to also do that before hashing.

As you noted, the problem with using `clean_value` directly is that it doesn't handle all contents of `_get_objects_to_hash`. I'm not 100% sure about this, but if it's applicable to all _attributes_, we could apply it to `val` in https://github.com/aiidateam/aiida-core/blob/4c9d44af4d8c2550444d9d528dce1b890c7772f6/aiida/orm/nodes/node.py#L1080

Two things to consider then:
- Performance: it seems from the docstring `clean_value` does some copying; we'd need to check how that affects hashing performance, and if there's a way _not_ to do it if the node is already stored.
- Changing hashes: generally, we want to avoid changes that lead to existing nodes getting different hashes, because they require a "rehash" of the entire DB. If it's unavoidable that isn't a blocking requirement, but we should be careful not to do it accidentally.

PR #4688, opened 2021-01-27
Improve namedtuples in aiida/engine

chrisjsewell:
Replace old-style namedtuples with `typing.NamedTuple` sub-classes.
This allows for typing and proper default values.

Note a minor wrinkle is that this requires python>=3.6.1 since, as per https://docs.python.org/3/library/typing.html#typing.NamedTuple;

> Changed in version 3.6.1: Added support for default values, methods, and docstrings.

### DIFF
--- a/aiida/engine/transports.py
+++ b/aiida/engine/transports.py
@@ -8,7 +8,6 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """A transport queue to batch process multiple tasks that require a Transport."""
-from collections import namedtuple
 import contextlib
 import logging
 import traceback
@@ -41,7 +40,6 @@ class TransportQueue:
     up to that point.  This way opening of transports (a costly operation) can
     be minimised.
     """
-    AuthInfoEntry = namedtuple('AuthInfoEntry', ['authinfo', 'transport', 'callbacks', 'callback_handle'])
 
     def __init__(self, loop: Optional[asyncio.AbstractEventLoop] = None):
         """
--- a/aiida/engine/processes/calcjobs/calcjob.py
+++ b/aiida/engine/processes/calcjobs/calcjob.py
@@ -370,7 +370,7 @@ class CalcJob(Process):
         for entry in self.node.get_outgoing():
             self.out(entry.link_label, entry.node)
 
-        return exit_code or ExitCode(0)  # type: ignore[call-arg]
+        return exit_code or ExitCode(0)
 
     def parse_scheduler_output(self, retrieved: orm.Node) -> Optional[ExitCode]:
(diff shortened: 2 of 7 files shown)


chrisjsewell:
How do folks feel about this requirement of python>=3.6.1? (mainly @csadorf)

aiida/engine/transports.py
```diff
@@ -41,7 +40,6 @@ class TransportQueue:
     up to that point.  This way opening of transports (a costly operation) can
     be minimised.
     """
-    AuthInfoEntry = namedtuple('AuthInfoEntry', ['authinfo', 'transport', 'callbacks', 'callback_handle'])
```
chrisjsewell, line 44: this is not used anywhere in the code base. Is it necessary?

ltalirz:
> Replace old-style namedtuples with `typing.NamedTuple`.
> This allows for typing and proper default values.
> 
> Note a minor wrinkle is that this requires python>=3.6.1 since, as per https://docs.python.org/3/library/typing.html#typing.NamedTuple;
> 
> > Changed in version 3.6.1: Added support for default values, methods, and docstrings.

Introducing features in a security-fix release... tztzt

I don't have a strong opinion on it. python 3.6.0 was [released dec 2016](https://www.python.org/downloads/release/python-360/) and python 3.6.1 [march 2017](https://www.python.org/downloads/release/python-361/).
Ubuntu switched from [3.5.1 on 16.04](https://packages.ubuntu.com/xenial/python3) directly to [3.6.5 on 18.04](https://packages.ubuntu.com/bionic/python3), i.e. for ubuntu users there shouldn't be issues.

Would be ok for me.

aiida/engine/transports.py
ltalirz: looks like this was introduced by @muhrin in https://github.com/aiidateam/aiida-core/commit/776d3a359c025f66c65b0d25be8e6cdc97e88239 (Jul 2017)

If it hasn't been used in 3.5 years, it would seem that this can be removed.
I'd like to wait for an OK from either @sphuber or @muhrin on this though

csadorf:
> How do folks feel about this requirement of python>=3.6.1? (mainly @csadorf)

The support for Python 3.6 is already dropped according to our [support table](https://www.aiida.net/download/).

chrisjsewell:
> The support for Python 3.6 is already dropped

perfect 😄

aiida/engine/transports.py
chrisjsewell: @sphuber / @muhrin can you drop a quick approval, then I can merge 😬

muhrin: haha, please drop away.

ltalirz (approved):
good to merge!

PR #4689, opened 2021-01-27
CI: Remove `--use-feature=2020-resolver` pip feature flag tests.

csadorf:
The feature is now on by default in the latest stable release.

PR #4690, opened 2021-01-27
CI: Notify slack on failure of the test-install workflow.

csadorf:


ltalirz (commented):
thanks @csadorf - just two questions

 * does it work?
 * can the notification task fail? if yes, we probably want to ignore such a failure

csadorf:
> thanks @csadorf - just two questions
> 
>     * does it work?

Yes, see #dev-aiida-core channel on the AiiDA slack workspace. 

> 
>     * can the notification task fail? if yes, we probably want to ignore such a failure

Yes, but I'd consider that an issue that I'd prefer not to silently ignore. The notification steps are currently only triggered on the "scheduled" runs. Since this would only be triggered if there is already an issue with the workflow, there is no benefit in silently ignoring this, is there?

.github/workflows/test-install.yml
```diff
@@ -72,6 +72,14 @@ jobs:
       run:
         python -c "import aiida"
 
+    - name: Send Slack notification
+      if: ${{ failure() && github.event_name == 'schedule' }}
```
chrisjsewell, line 76: after #4689 is merged, I think this will never trigger, because the `continue-on-error` has been removed

chrisjsewell (changes requested):

.github/workflows/test-install.yml
csadorf: The `continue-on-error` field was previously added to ensure that the overall workflow continues even if there are issues with the pip-2020-resolver. To the best of my knowledge this should not affect the execution of those steps to run on failure.

chrisjsewell: oh yeh I think you are correct

chrisjsewell (changes requested):
if you resolve the conflicts, I'm happy to approve 👍

chrisjsewell (approved):

PR #4692, opened 2021-01-28
🐛 FIX: Standardise transport task interrupt handling

chrisjsewell:
As per the other transport tasks,
`do_upload` should be interruptable via the `Interruption` exception,
e.g when the process is killed.

Addresses https://github.com/aiidateam/aiida-core/issues/4649#issuecomment-768443928


chrisjsewell:
@sphuber I assume this is what you had in mind?

sphuber:
> @sphuber I assume this is what you had in mind?

Sure did. I just noticed some more inconsistencies however that probably merit fixing in this PR. The upload and submit tasks catch the exception but than simply pass, essentially swallowing it. The other tasks, upload and retrieve, properly reraise the exception. I think that should be the behavior.

Then a second thing. The upload task catches also the `CancelledError` which is raised I think if the future is cancelled. I think the same concept should apply to the other tasks, but none of them add this as an ignored exception and so also no catch and reraise it. It feels like they should do the same as the upload task. Note that `Interruption` and `CancelledError` do not share a common parent so have to be defined separately as to be ignored exceptions. @muhrin should the `CancelledError` apply to all tasks?

muhrin (dismissed):
Looks reasonable to me!

muhrin:
Sorry, missed the query from you @sphuber .  As far as I remember, yes, any `CancelledException` should be raised by these tasks, so as you can see there it's being intercepted just so that it can be raised.  I think the other tasks should do the same (if they catch the general `Exception`)

sphuber:
alright @chrisjsewell , if you could please add the few patches I mentioned, I will merge this

chrisjsewell:
> The other tasks, upload and retrieve, properly reraise the exception. I think that should be the behavior.

>  As far as I remember, yes, any CancelledException should be caught and passed (or whatever the noop would be).

@sphuber @muhrin hmm, you seem to be contradicting each other?
To clarify:

1. Should `plumpy.futures.CancelledError` be passed or reraised
2. Should `plumpy.process_states.Interruption` be passed or reraised

sphuber:
I think they should be reraised. We catch them because we want them to be treated differently from all other exceptions, which should raise a `TransportTaskException`. The `CancelledError` and `Interruption` are caught in the `Waiting.execute` method:
https://github.com/aiidateam/aiida-core/blob/443dc01f0fdaba11060a2d6ddcd9cebf911d22c6/aiida/engine/processes/calcjobs/tasks.py#L418

where it is then handled accordingly. @muhrin ?

chrisjsewell:
> The CancelledError and Interruption are caught in the Waiting.execute method

cheers, makes sense 👍

chrisjsewell:
> where it is then handled accordingly. @muhrin ?

pinging @muhrin for final approval 😉

muhrin:
Hey chaps.  Looking at it more carefully, I agree with @sphuber , raising is the way to go as it's dealt with by the caller.  Merge away.

muhrin (approved):

chrisjsewell:
cheers guys!

PR #4695, opened 2021-01-30
engine: try cleanup transports

ltalirz:
@sphuber : 

~~this gets rid of a leftover transport reference~~
No it doesn't - I had a "bug" in my tests ;-)

### DIFF
--- a/aiida/engine/transports.py
+++ b/aiida/engine/transports.py
@@ -23,10 +23,47 @@ _LOGGER = logging.getLogger(__name__)
 class TransportRequest:
     """ Information kept about request for a transport object """
 
-    def __init__(self):
+    def __init__(self, authinfo: AuthInfo):
         super().__init__()
         self.future: asyncio.Future = asyncio.Future()
         self.count = 0
+
+        self._callback_handle: Optional[asyncio.TimerHandle] = None
+        self._authinfo = authinfo
+        self._transport = authinfo.get_transport()
+
+    def do_open(self):
+        """ Actually open the transport
+
+        Sets self.future to the open transport, if successful, or to an exception.
+        """
+        if self.count > 0:
+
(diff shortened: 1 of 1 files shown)


aiida/engine/transports.py
```diff
@@ -23,11 +23,48 @@
...
+            # The user still wants the transport so open it
+            _LOGGER.debug('Transport request opening transport for %s', self._authinfo)
+            try:
+                self._transport.open()
+            except Exception as exception:  # pylint: disable=broad-except
+                _LOGGER.error('exception occurred while trying to open transport:\n %s', exception)
+                self.future.set_exception(exception)
+
+                # Cleanup of the stale TransportRequest with the excepted transport future
+                # self._transport_requests.pop(authinfo.id, None)
```
ltalirz, line 51: this is what would have happened in the previous codebase.

does this cleanup really need to happen here or is it ok to do it at the end?
I guess all the other context managers waiting for this transport will anyhow get the excepted one.

on the other hand it's true - if this is cleaned up here already, follow-up transport requests can already get a fresh one

it just makes the code more difficult to reason about

chrisjsewell:
Thanks @ltalirz, is there a unit test(s) we can add to check that there are indeed to leftover transport requests (one that would fail with the current code)?

chrisjsewell:
While we are at it, it might also be good to add a brief entry somewhere in docs/source/internals, with an explanation of how the transport works, and/or in the module docstring of aiida/engine/transports.py

ltalirz:
Yes, I'll try to add some tests later... so far this is really just a basis for discussion

Since we're seeming to get closer to the root issue, I'll try to look at that first

ltalirz:
And indeed i had a bug in my checks - the transport and transport request  is still there.
Let's leave this cleanup Pr for later

Issue #4691, opened 2021-01-27
Drop support for Python 3.6.


csadorf:
According to our [support table](https://www.aiida.net/download/), support for Python 3.6 is dropped.

PR #4697, opened 2021-02-01
CI: Increase timeout for test-install test suite.

csadorf:
The current timeout of 35min is regularly exceeded resulting in test failures. This change increases it to 55min.

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -121,7 +121,7 @@ jobs:
 
     needs: [install-with-pip, install-with-conda]
     runs-on: ubuntu-latest
-    timeout-minutes: 35
+    timeout-minutes: 55
 
     strategy:
       fail-fast: false


csadorf:
There appears to be some issue with the pre-commit upstream. I'll just rebase and rerun this later today.

chrisjsewell:
Hopefully this helps, but this is not really the underlying cause of the issue. It is because we have too many test jobs, which ends up causing some to go very slow (or even hang)

csadorf:
> Hopefully this helps, but this is not really the underlying cause of the issue. It is because we have too many test jobs, which ends up causing some to go very slow (or even hang)

This is primarily to address the failure of [scheduled nightly runs ](https://github.com/aiidateam/aiida-core/actions?query=workflow%3Atest-install+event%3Aschedule)which should not really be affected by our general workload.

csadorf:
As a related note, we could host our own runner which might alleviate the situation.

chrisjsewell:
> which should not really be affected by our general workload.

Oh it is definitely to do with the work load; the test suite only took < 12 minutes to run for all other jobs in the matrix, but the failing job got killed at 31 minutes: https://github.com/aiidateam/aiida-core/runs/1803637702?check_suite_focus=true. I have had this happen numerous times in the last few weeks.

No problem to increase to the timeout, but I would also suggest for this workflow, you can now drop the python 3.6 jobs, inline with #4691 

> As a related note, we could host our own runner which might alleviate the situation.

I was looking into that for replacing Jenkins, but it has a number of security implications

csadorf:
> > which should not really be affected by our general workload.
> 
> Oh it is definitely to do with the work load; the pytest suite only took < 12 minutes to run for all other jobs in the matrix, but the failing job got killed at 31 minutes: https://github.com/aiidateam/aiida-core/runs/1803637702?check_suite_focus=true. I have had this happen numerous times in the last few weeks.

What I mean is that the runners reserved for this repository should be exclusively used for the scheduled `test-install` workflow in the middle night and it shouldn't be affected by any other CI work.

> No problem to increase the timeout, but I would also suggest for this workflow, you can now drop the python 3.6 jobs, inline with #4691
> 

I am planning on tackling that as part of the upcoming coding week.

> > As a related note, we could host our own runner which might alleviate the situation.
> 
> I was looking into that for replacing Jenkins, but it has a number of security implications, and in the documentation it basically suggests to only use your own runner on private repositories

Oh wow, that's really bad. I hope that they will work on that.

chrisjsewell (changes requested):
let me know when you have fixed the pre-commit 😄

chrisjsewell:
> aiida/tools/graph/graph_traversers.py:212: error: Incompatible types in assignment (expression has type "float", variable has type "Optional[int]")  [assignment]
Found 1 error in 1 file (checked 55 source files)

Started seeing this in other PRs as well (cc @ltalirz), possible dependency version change?

chrisjsewell:
See #4700

ltalirz:
yes, numpy 1.20 has been released https://numpy.org/doc/1.20/release/1.20.0-notes.html

already broke pymatgen ;-)

chrisjsewell:
> already broke pymatgen ;-)

No surprise there 😆

chrisjsewell (approved):

csadorf:
I believe this problem is actually related to #4731 which I would like to prioritize to resolve.

PR #4700, opened 2021-02-01
🐛 FIX: typing failure

chrisjsewell:
Failing in other open PRs: it appears `numpy.inf` is no longer typed as an integer.

### DIFF
--- a/aiida/tools/graph/graph_traversers.py
+++ b/aiida/tools/graph/graph_traversers.py
@@ -9,7 +9,7 @@
 ###########################################################################
 """Module for functions to traverse AiiDA graphs."""
 import sys
-from typing import Any, Callable, cast, Dict, Iterable, List, Mapping, Optional, Set
+from typing import Any, Callable, cast, Dict, Iterable, List, Mapping, Optional, Set, Union
 
 from numpy import inf
 
@@ -182,7 +182,7 @@ def validate_traversal_rules(
 
 def traverse_graph(
     starting_pks: Iterable[int],
-    max_iterations: Optional[int] = None,
+    max_iterations: Union[None, int, inf] = None,
     get_links: bool = False,
     links_forward: Iterable[LinkType] = (),
     links_backward: Iterable[LinkType] = (),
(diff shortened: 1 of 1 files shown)


greschd:
AFAICT, passing `numpy.inf` would still complain.. if we're ok with adding `typing_extensions` to the dependencies for python<3.8, that could be typed as `Optional[Union[int, Literal[inf]]]`.

ltalirz (approved):
thanks

Just wondering - there is no special treatment needed for `None` anymore either?

ltalirz:
hm.. I also guess this would make any CI run fail that doesn't use numpy 1.20

chrisjsewell:
> AFAICT, passing numpy.inf would still complain.. if we're ok with adding typing_extensions to the dependencies

Yeh it did, I've just cast the `inf` to `int` for now, which should always work.
Would not be adverse to adding `typing_extensions`, but seems a bit drastic for this particular issue

ltalirz (changes requested):
is it possible to support both new and old numpy versions?

it's only for developers, so perhaps it's ok...

 but if you switch to the new numpy, you'll also need to update the numpy dependency as well so that developers won't run into issues

greschd:
> Yeh it did, I've just cast the inf to int for now, which should always work.

Yeah, but that's only for the `None` case.. if someone passes `np.inf` explicitly, it should complain at the call site, no?

chrisjsewell:
> is it possible to support both new and old numpy versions?

not sure what you meany by this? The change in this PR does support both

ltalirz:
> not sure what you meany by this? The change in this PR does support both

Sorry, I was seeing a different change... it also wasn't just the first commit. No idea, I'll leave it to @greschd and you to agree there

chrisjsewell:
> if someone passes `np.inf` explicitly, it would falsely complain at the call site, no?

it would, and that would be correct because the call signature does not allow you to pass `numpy.inf`, that is an internal implementation, and you should use `None` instead

this is only for mypy though obviously, not during actual code execution

greschd:
Yep, sorry, noticed that too late (see my sneaky edit above).

greschd (approved):

chrisjsewell:
> Yep, sorry, noticed that too late (see my sneaky edit above).

very sneaky 😆

PR #4707, opened 2021-02-04
Allow setting `rerunnable` from `metadata['options']`

greschd:
The `rerunnable` option is currently implemented in the scheduler plugins ([LSF](https://github.com/aiidateam/aiida-core/blob/48fa47584c993cdac019c6199bc045a4c19da152/aiida/schedulers/plugins/lsf.py#L306), [PBS](https://github.com/aiidateam/aiida-core/blob/develop/aiida/schedulers/plugins/pbsbaseclasses.py#L197), [SGE](https://github.com/aiidateam/aiida-core/blob/develop/aiida/schedulers/plugins/sge.py#L170), [SLURM](https://github.com/aiidateam/aiida-core/blob/48fa47584c993cdac019c6199bc045a4c19da152/aiida/schedulers/plugins/slurm.py#L272)), but there was no way to activate it for a particular `CalcJob`. This PR adds a `metadata.options.rerunnable` flag that is forwarded to the `JobTemplate`.

This is a follow-up on a [discussion on Slack](https://app.slack.com/client/T470FLBSQ/C466S5PG8/thread/C466S5PG8-1603730677.038100) with @giovannipizzi and @espenfl.

Important points raised there:
- it might depend on scheduler settings of the cluster whether re-queueing gives the same job id
- (depending again on the scheduler config) it's possible that a code would execute more than once, so it should only be applied for idempotent codes

I've now been running this in production for a while, and did not notice any adverse affects. In our case, submitting jobs to SLURM can trigger the allocation of new nodes; if that fails for some reason, SLURM needs to re-queue the jobs such that it can assign different nodes to the job.

Some open questions:
- is there a good place to test this? I looked at [test_calc_job.py](https://github.com/aiidateam/aiida-core/blob/develop/tests/engine/test_calc_job.py), but since the option doesn't affect the direct scheduler there's no way to check it was used. Do we have a calcjob test with different schedulers?
- related to the first point, should the direct scheduler warn when the (unused) `rerunnable` flag is set?
- looking at the [SGE implementation](https://github.com/aiidateam/aiida-core/blob/develop/aiida/schedulers/plugins/sge.py#L170), I'm not sure if that expects a `bool` (as do all the other schedulers), or a `str`.

### DIFF
--- a/aiida/engine/processes/calcjobs/calcjob.py
+++ b/aiida/engine/processes/calcjobs/calcjob.py
@@ -211,6 +211,8 @@ class CalcJob(Process):
                  'inserted before any non-scheduler command')
         spec.input('metadata.options.queue_name', valid_type=str, required=False,
             help='Set the name of the queue on the remote computer')
+        spec.input('metadata.options.rerunnable', valid_type=bool, required=False,
+            help='Determines if the calculation can be requeued / rerun.')
         spec.input('metadata.options.account', valid_type=str, required=False,
             help='Set the account to use in for the queue on the remote computer')
         spec.input('metadata.options.qos', valid_type=str, required=False,
@@ -526,7 +528,7 @@ class CalcJob(Process):
         job_tmpl = JobTemplate()
         job_tmpl.shebang = computer.get_shebang()
         job_tmpl.submit_as_hold = False
-        job_tmpl.rerunnable = False
+        job_tmpl.rerunnable = self.options.get('rerunnable', False)
         job_tmpl.job_environment = {}
         # 'email', 'email_on_started', 'email_on_terminated',
         job_tmpl.job_name = f'aiida-{self.node.pk}'
--- a/aiida/schedulers/plugins/direct.py
+++ b/aiida/schedulers/plugins/direct.py
@@ -166,6 +166,11 @@ class DirectScheduler(aiida.schedulers.Scheduler):
             lines.append('# ENVIRONMENT VARIABLES  END  ###')
             lines.append(empty_line)
 
+        if job_tmpl.rerunnable:
+            self.logger.warning(
+                "The 'rerunnable' option is set to 'True', but has no effect when using the direct scheduler."
+            )
+
         lines.append(empty_line)
 
         ## The following code is not working as there's an empty line
--- a/aiida/schedulers/plugins/sge.py
+++ b/aiida/schedulers/plugins/sge.py
@@ -168,8 +168,9 @@ class SgeScheduler(aiida.schedulers.Scheduler):
             lines.append(f'#$ -h {job_tmpl.submit_as_hold}')
 
         if job_tmpl.rerunnable:
-            # if isinstance(job_tmpl.rerunnable, str):
-            lines.append(f'#$ -r {job_tmpl.rerunnable}')
+            lines.append('#$ -r yes')
+        else:
+            lines.append('#$ -r no')
 
         if job_tmpl.email:
             # If not specified, but email events are set, PBSPro
--- a/docs/source/topics/calculations/usage.rst
+++ b/docs/source/topics/calculations/usage.rst
@@ -555,6 +555,13 @@ The full list of available options are documented below as part of the ``CalcJob
     :expand-namespaces:
 
 
+The ``rerunnable`` option enables the scheduler to re-launch the calculation if it has failed, for example due to node failure or a failure to launch the job. It corresponds to the ``--requeue`` option in SLURM, and the ``-r`` option in SGE, LSF, and PBS. The following two conditions must be met in order for this to work well with AiiDA:
+
+- the scheduler assigns the same job-id to the restarted job
+- the code produces the same results if it has already partially run before (not every scheduler may produce this situation)
+
+Because this depends on the scheduler, its configuration, and the code used, we cannot say conclusively when it will work -- do your own testing! It has been tested on a cluster using SLURM, but that does not guarantee other SLURM clusters behave in the same way.
+
 .. _topics:calculations:usage:calcjobs:launch:
 
 Launch
--- a/tests/engine/test_calc_job.py
+++ b/tests/engine/test_calc_job.py
@@ -9,6 +9,7 @@
 ###########################################################################
 # pylint: disable=too-many-public-methods,redefined-outer-name
 """Test for the `CalcJob` process sub class."""
+import json
 from copy import deepcopy
 from functools import partial
 import io
@@ -387,6 +388,23 @@ class TestCalcJob(AiidaTestCase):
         # Since the repository will only contain files on the top-level due to `Code.set_files` we only check those
         for filename in self.local_code.list_object_names():
             self.assertTrue(filename in uploaded_files)
+
+    @pytest.mark.usefixtures('dry_run_in_tmp')
+    def test_rerunnable(self):
+        """Test that setting `rerunnable` in the options results in it being set in the job template."""
+        inputs = deepcopy(self.inputs)
+        inputs['code'] = self.local_code
+        inputs['metadata']['computer'] = self.computer
+        inputs['metadata']['dry_run'] = True
+        inputs['metadata']['options']['rerunnable'] = True
+
+        _, node = launch.run_get_node(ArithmeticAddCalculation, **inputs)
+        job_tmpl_file = os.path.join(node.dry_run_info['folder'], '.aiida', 'job_tmpl.json')
+
+        with open(job_tmpl_file, mode='r') as in_f:
+            job_tmpl = json.load(in_f)
+
+        assert job_tmpl['rerunnable']
 
     @pytest.mark.usefixtures('dry_run_in_tmp')
     def test_provenance_exclude_list(self):
--- a/tests/schedulers/test_direct.py
+++ b/tests/schedulers/test_direct.py
@@ -9,12 +9,11 @@
 ###########################################################################
 # pylint: disable=invalid-name,protected-access
 """Tests for the `DirectScheduler` plugin."""
-import logging
 import unittest
 
 from aiida.schedulers.plugins.direct import DirectScheduler
 from aiida.schedulers import SchedulerError
-from aiida.common.log import NotInTestingFilter
+from aiida.common.log import AIIDA_LOGGER
 
 # This was executed with ps -o pid,stat,user,time | tail -n +2
 mac_ps_output_str = """21259 S+   broeder   0:00.04
@@ -97,14 +96,14 @@ def test_submit_script_rerunnable(caplog):
     direct = DirectScheduler()
     job_tmpl = JobTemplate()
 
-    # AiiDA by default filters logging when testing
-    direct.logger.removeFilter(NotInTestingFilter)
+    # caplog needs the logs to be propagated
+    original_propagate_value = AIIDA_LOGGER.propagate
+    AIIDA_LOGGER.propagate = True
 
     job_tmpl.rerunnable = True
-    with caplog.at_level(logging.WARNING, logger=direct.logger.name):
-        direct._get_submit_script_header(job_tmpl)
+    direct._get_submit_script_header(job_tmpl)
 
     assert 'rerunnable' in caplog.text
     assert 'has no effect' in caplog.text
 
-    direct.logger.addFilter(NotInTestingFilter)
+    AIIDA_LOGGER.propagate = original_propagate_value
--- a/tests/schedulers/test_direct.py
+++ b/tests/schedulers/test_direct.py
@@ -13,7 +13,6 @@ import unittest
 
 from aiida.schedulers.plugins.direct import DirectScheduler
 from aiida.schedulers import SchedulerError
-from aiida.common.log import AIIDA_LOGGER
 
 # This was executed with ps -o pid,stat,user,time | tail -n +2
 mac_ps_output_str = """21259 S+   broeder   0:00.04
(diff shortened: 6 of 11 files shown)


giovannipizzi:
Thanks @greschd !

A few comments:
- it would be good do add documentation for this, and in particular also an explanation that this was tested with SLURM and it works fine, that it's not implemented in `direct`, and for the others it's untested (in particular, it might generate a new job ID and so AiiDA would not be able to 'pick' the rerunned job), and the other note on idempotent jobs. Maybe even you could open an issue, link to this PR, and in the docs you can point people to the issue, saying to report if they use a different scheduler with rerun so we can have this "tested", in the future
- for testing, you could split the test in two units: one in the place you suggest, checking that the job template attribute is set, just calling `presubmit`; and (more than) one for each scheduler plugin, to check that they write the correct string in the header if the `rerunnable` flag is set to True
- Regarding the direct scheduler - I think it's not crucial, but indeed since by default the flag is set to False, it's ok to issue a warning there if rerunnable is true, and continue
- regarding SGE, [here](http://gridscheduler.sourceforge.net/htmlman/htmlman1/qsub.html) it says:

        -r y[es]|n[o]
          Available for qsub and qalter only.

          Identifies the ability of a job to be rerun or not.  If
          the  value of -r is 'yes', the job will be rerun if the
          job was  aborted  without  leaving  a  consistent  exit
          state.   (This  is  typically  the  case if the node on
          which the job is running crashes).  If -r is 'no',  the
          job will not be rerun under any circumstances.
          Interactive jobs submitted with qsh, qrsh or qlogin are
          not rerunnable.

          Qalter allows changing this option even while  the  job
          executes.

          The value specified with this option or the correspond-
          ing  value  specified  in  qmon  will only be passed to
          defined JSV instances if the value is yes.  The name of
          the  parameter will be r. The value will be y also when
          then long form yes  was  specified  during  submission.
          (see  -jsv  option  above or find more information con-
          cerning JSV in jsv(1))

   So my guess is that the current implementation might even be wrong (I think the combination SGE+rerunnable was never tested) - I suggest you just replace it with
   ```python
        if job_tmpl.rerunnable:
            lines.append(f'#$ -r yes')
   ```

greschd:
Thanks @giovannipizzi, agreed on all points. One quick question: what's the best place to put the documentation? From a quick glance, either in the [options paragraph](https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/calculations/usage.html#options) or a separate paragraph after [dry run](https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/calculations/usage.html#dry-run) makes sense to me, but I'm not very familiar with the organization of the docs.

giovannipizzi:
> What's the best place to put the documentation? From a quick glance, either in the [options paragraph](https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/calculations/usage.html#options) or a separate paragraph after [dry run](https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/calculations/usage.html#dry-run) makes sense to me, but I'm not very familiar with the organization of the docs.

Indeed definitely in the options paragraph; if all the discussion fits in there, great. Otherwise a new paragraph with the more technical details (lined from the options paragraph) is also a good idea.

> If I understood @espenfl's point correctly, we should say that it works fine with _one specific_ SLURM setup. People should do their own testing because SLURM configurations can differ.

I see, good point - great to mention that, then

espenfl:
Yes, SLURM setups are different (as are the other schedulers as well). Also, maybe this is a good time to put some life into my backburner scheduler container work. It was more or less done, just needed to integrate them with the scheduler code in AiiDA and introduce some tests based on those. I will try to do some of that next week.

greschd:
Regarding SGE: Do you know if the default is `no`, or should we add `'#$ -r no'` if `rerunnable` is `False`?

greschd:
> 
> 
> Regarding SGE: Do you know if the default is `no`, or should we add `'#$ -r no'` if `rerunnable` is `False`? From that man page, it wasn't 100% clear to me.

Note that I've now added the `#$ -r no` if `rerunnable` is `False`, but that can easily be reverted if desired. The benefit of adding it is that we the rerunnable option is _definitely_ off when set to `False`, but it may break a setup that inadvertently relied on it being turned off somewhere in the cluster's settings. I think that is ok, because the "correct" fix would then be to turn on `rerunnable` explicitly.

ramirezfranciscof:
Hey @greschd , thanks for the contribution! Sorry it got a bit buried 😅 

I've taken a look and it all seems good; I don't know if you wanted a more specific greenlight for someone in particular, but feel free to tag them if that was the case, otherwise I can approve once the branch is updated. The conflicts in `tests/schedulers/test_lsf.py` are because I recently moved the tests out of the old format into normal pytest functions, I think you can just copy your added test method as it is outside of its current class and it should work (it isn't even using `self` anywhere else). If you don't currently have the time, let me know and I can do the update.

The only thing I'm a bit wary about is this:

> Note that I've now added the `#$ -r no` if `rerunnable` is `False`, but that can easily be reverted if desired. The benefit of adding it is that we the rerunnable option is _definitely_ off when set to `False`, but it may break a setup that inadvertently relied on it being turned off somewhere in the cluster's settings. I think that is ok, because the "correct" fix would then be to turn on `rerunnable` explicitly.

This is true as long as the schedulers didn't have / will not have any sort of modification that will change how this option works or is provided (for example, perhaps previously they only admitted `#$ --rerunable no/yes` or `#$ -r disable/enable`, or maybe the option wasn't even there yet). If this was the case, a "no keyword" approach would still work with old versions but only allowing for explicit enabling/disabling may break without any possible solution for the user. I'm not sure how common this is with these schedulers and if this is a scenario worth "code-considering", but I thought at least it was worth a brief "design-considering" =P.

I guess the super-safe road would be to default these options to `None`, in which case they are ignored, and allow the user to set both `False` or `True` so they can make sure they are being applied in both ways.

greschd:
> This is true as long as the schedulers didn't have / will not have any sort of modification that will change how this option works or is provided (for example, perhaps previously they only admitted #$ --rerunable no/yes or #$ -r disable/enable, or maybe the option wasn't even there yet). If this was the case, a "no keyword" approach would still work with old versions but only allowing for explicit enabling/disabling may break without any possible solution for the user. I'm not sure how common this is with these schedulers and if this is a scenario worth "code-considering", but I thought at least it was worth a brief "design-considering" =P.

Yeah, that is exactly the point where I am not 100% sure. I think it would be good to get buy-in from @giovannipizzi on this.

> I guess the super-safe road would be to default these options to None, in which case they are ignored, and allow the user to set both False or True so they can make sure they are being applied in both ways.

The main difficulty here is that not all schedulers treated `None` the same so far -- PBS didn't emit anything, whereas SLURM and LSF added what is now the `False` option. So it's a trade-off between consistency and strict backwards-compatibility.

ramirezfranciscof:
> The main difficulty here is that not all schedulers treated `None` the same so far -- PBS didn't emit anything, whereas SLURM and LSF added what is now the `False` option. So it's a trade-off between consistency and strict backwards-compatibility.
>
> We _could_ just say that `None` is "legacy behavior", and do some deprecation path. But then you'll end up emitting warnings to users for what may end up being no good reason.

So to make sure I understood this correctly: you mean that some _scheduler plugins_ received the `option=None` and wrote `#$ option = False` in the running script, whereas others didn't write anything with the same input?

If this is the case, I would say there are two cases:

1. **The default behavior of the _scheduler_ is the same when the option is not present or when it is set to `False` (or whatever the _scheduler plugin_ defaults to)**. If this is the case I would argue there is no need to do a deprecation path for the change. I mean, it is true that we change what we write in the script, but if the resulting effect on the scheduler is the same then there is not really any observable change occurring from the point of view of the user. One could say that we need to deprecate because the result we have control over is the writing of the script, not what the scheduler does with it: I would counter that if the scheduler at some point changes what the default behavior is for the missing keyword, then it is the scheduler that is breaking backwards compatibility, not us. Not sure if one argument "objectively" wins over the other, but I like mine better 🤓 .
2. **The default behavior of the _scheduler_ is NOT the same**. Indeed here a warning should be issued and a deprecation path provided. This could generate some friction and is a little extra work, but I think consistency in internal behavior saves you many more headaches in the long run.

In any case, I now understand that perhaps this should be considered a separate discussion and we could open an issue to get other people's input on this. If you think it is a good idea, I may write that later and ping you to check if my description of the situation is correct (or if you think it would be easier to just write it yourself feel free he he 😅 ).

greschd:
> So to make sure I understood this correctly: you mean that some _scheduler plugins_ received the `option=None` and wrote `#$ option = False` in the running script, whereas others didn't write anything with the same `option=None` input?

Exactly. Also agree with your analysis of the two options. Since we're only changing behavior for the PBS plugin, we need to determine the defaults for the PBS scheduler to see if we're in scenario 1. or 2.

greschd:
According to [its manual](http://gridscheduler.sourceforge.net/htmlman/manuals.html), at least the _current_ default for SGE is `-R n` (case 1.):

```
     -R y[es]|n[o]
          Available for qsub, qrsh, qsh, qlogin and qalter.

          Indicates whether a reservation for this job should  be
          done.  Reservation  is  never  done for immediate jobs,
          i.e. jobs submitted using the -now yes option.   Please
          note  that  regardless  of the reservation request, job
          reservation might be disabled using max_reservation  in
          sched_conf(5)  and  might  be limited only to a certain
          number of high priority jobs.

          By default jobs are submitted with the -R n option.

          The value specified with this option or the correspond-
          ing  value  specified  in  qmon  will only be passed to
          defined JSV instances if the value is yes.  The name of
          the  parameter will be R. The value will be y also when
          then long form yes  was  specified  during  submission.
          (see  -jsv  option  above or find more information con-
          cerning JSV in jsv(1))
```

chrisjsewell:
@sphuber mentioned that @atztogo may use the SGE scheduler. Perhaps if he has time, he could quickly comment if he sees any reason that this would break anything for him?
Otherwise, I think this looks good to me thanks 👍

giovannipizzi:
I agree that, according to the docs we could find this looks good to me.
I'd wait a few days in case @atztogo can confirm that this (i.e. explicitly adding `-R n` in the SGE header) does not break anything in his workflows.

Otherwise OK to merge for me

atztogo:
I don't fully understand the discussion on SGE here (a little bit too technical for me), but I think there is no impact for me. The suggestion in the first @giovannipizzi's comment is probably most natural. I leave my SGE setting as below.

I didn't know that this option exists...  Currently I use SGE installed on ubuntu 18.04 (debian package of `8.1.9+dfsg-4+deb9u2`). `man qsub` gives (not `-R` but `-r` if I understand correctly.)
```
       -r y[es]|n[o]
              Available for qsub and qalter only.

              Specifies whether a job can be rerun or not.  If the value of -r
              is 'yes', the job will be rerun if it gets aborted without leav‐
              ing a consistent exit state.  (This is typically the case if the
              node  on  which the job is running crashes).  If -r is 'no' (the
              default), the job will not be rerun  under  such  circumstances.
              It  will  still be rerun if it finishes with exit code 99 unless
              FORBID_RESCHEDULE is set in qmaster_params in sge_conf(5).

              Interactive jobs submitted with qsh, qrsh,  or  qlogin  are  not
              rerunnable.

              Qalter allows changing this option even while the job executes.

              The  value specified with this option or the corresponding value
              specified in qmon will only be passed to defined  JSV  instances
              if  the  value is yes.  The name of the parameter will be r. The
              value will be y also when the long form yes was specified during
              submission.   (See  -jsv  option  above or find more information
              concerning JSV in jsv(1).)
```

The default behaviour (i.e., without writing `#$ -r [yes|no]` on each queue setting (a queue means a set of parameters such as resource, parallelization, etc) seems configurable. Some parameters are overrode by the job-script's header part. I haven't ever touched this option `rerun` or `-r`. So my current setting is FALSE as written in the man page, and could be confirmed by
```
% sudo qconf -sq queue-name
...
rerun                 FALSE
...
```
I tried running with `#$ -r no` and `#$ -r yes` to see how it affects to a job and looked for any log of SGE on this but could not find the information on the status of this option, though there is some way to find it.

atztogo:
> The suggestion in the first @giovannipizzi's comment is probably most natural. I leave my SGE setting as below.

This was not very correct. Rather, the below with `rerunnable=None` as the default for SGE sounds nice to me,

> So to make sure I understood this correctly: you mean that some scheduler plugins received the option=None and wrote #$ option = False in the running script, whereas others didn't write anything with the same option=None input?
```
     if job_tmpl.rerunnable is True:
         lines.append(f'#$ -r yes')
     elif job_tmpl.rerunnable is False:
         lines.append(f'#$ -r no')
     elif job_tmpl.rerunnable is None:
         pass
```

giovannipizzi:
Thanks @atztogo - I realise now that @greschd 's [commment](https://github.com/aiidateam/aiida-core/pull/4707#issuecomment-848025230) was misleading ;-D he reported indeed the `-R` option (reservation) rather than `-r` (rerunnable). So from Togo's comment it seems indeed that there is no SGE default.

So, just so it's clear to me: the "problem" with Togo's last suggestion, i.e. allowing also a `None` value to mean "don't specify => use the scheduler's default", would be that the behaviour is different than the other schedulers (where instead the default is enforced to be `no`?

Maybe it's better to do the opposite then, and do not specify at all the rerunnable option in other plugins, if the value is not specified or `None`? Or there is too much of a risk to create other subtle problems with the other schedulers?

greschd:
> Thanks @atztogo - I realise now that @greschd 's [commment](https://github.com/aiidateam/aiida-core/pull/4707#issuecomment-848025230) was misleading ;-D he reported indeed the `-R` option (reservation) rather than `-r` (rerunnable). So from Togo's comment it seems indeed that there is no SGE default.

Yeah, copied the wrong part :sweat_smile:

sphuber:
Since this has been open for a long time, I have given it a rough look. Please correct me if I'm wrong but I think that adding the new option does not change the current behavior, since the default is to have `rerunnable = False`. Since you also added documentation to explain what the pre-conditions are for this flag to work with AiiDA, I would be tempted to accept and merge this. @giovannipizzi and @ramirezfranciscof any objections?

giovannipizzi (dismissed):
Thanks @greschd , and thanks @sphuber for pinging! After solving the conflicts, it's good to merge for me!

espenfl:
Nice to see this merged. Thanks a lot for the work and discussions.

greschd:
> Thanks @greschd , and thanks @sphuber for pinging! After solving the conflicts, it's good to merge for me!

Rebased and good to merge if checks pass from my side.

sphuber (approved):
Thanks @greschd

Issue #4523, opened 2020-10-28
Dropping of unknown configuration keys can lead to inconsistent config state


greschd:
### Describe the bug

When an unknown key is encountered in the configuration, it is simply dropped instead of going through the config migrations. This can lead to an inconsistent configuration that is not recoverable without manually editing it.

### Steps to reproduce

1. `pip install aiida-core==1.4.1`
2. Set up a new profile. The config will contain
    ```json
    "CONFIG_VERSION": {
        "CURRENT": 4,
        "OLDEST_COMPATIBLE": 3
    }
    ```
    and something like
    ```json
    "tmpfoo": {
        "PROFILE_UUID": "d925405ed3f14935b8fc04dba127ba18",
        "AIIDADB_ENGINE": "postgresql_psycopg2",
        "AIIDADB_BACKEND": "django",
        "AIIDADB_NAME": "tmpfoo",
        "AIIDADB_PORT": 5432,
        "AIIDADB_HOST": "localhost",
        "AIIDADB_USER": "tmpfoo",
        "AIIDADB_PASS": "1234",
        "broker_protocol": "amqp",
        "broker_username": "guest",
        "broker_password": "guest",
        "broker_host": "127.0.0.1",
        "broker_port": 5672,
        "broker_virtual_host": "",
        "AIIDADB_REPOSITORY_URI": "file:///home/a-dogres/.aiida/repository/tmpfoo",
        "options": {},
        "default_user_email": ""
    }
    ```
3. `pip install aiida-core==1.3.1`
4. Run a command, e.g. plain `verdi`. The config will _still_ contain 
    ```json
    "CONFIG_VERSION": {
        "CURRENT": 4,
        "OLDEST_COMPATIBLE": 3
    }
    ```
    but now with 
    ```json"
    tmpfoo": {
        "PROFILE_UUID": "d925405ed3f14935b8fc04dba127ba18",
        "AIIDADB_ENGINE": "postgresql_psycopg2",
        "AIIDADB_BACKEND": "django",
        "AIIDADB_NAME": "tmpfoo",
        "AIIDADB_PORT": 5432,
        "AIIDADB_HOST": "localhost",
        "AIIDADB_USER": "tmpfoo",
        "AIIDADB_PASS": "1234",
        "AIIDADB_REPOSITORY_URI": "file:///home/a-dogres/.aiida/repository/tmpfoo",
        "options": {},
        "default_user_email": ""
    }
    ```
[...]

PR #4712, opened 2021-02-08
♻️ REFACTOR: configuration management API and CLI

### LINKED ISSUE #2701: Merge `config.json` and `caching.yaml` configuration files
When merging the format has to be chosen, where `yaml` may be the best choice.
This would require the configuration migration framework to be rewritten however.

### LINKED ISSUE #4240: Add `verdi` command that parses and prints the caching configuration

### LINKED ISSUE #4584: Refactor `verdi config`
I'm grouping here #4260 and #4480 (also linked to #4240):

`verdi config` currently has limited functionality, since it is a terminal/leaf command, rather than a command group, and also because it relies on tab completion for listing available options.

It should be turned it to a command group with at minimal `list`, `show`, `set`, `unset` commands.

Proposals:

```console
$ verdi config list -h

Usage: verdi config list [OPTIONS]

  List AiiDA options for the current profile.

Options:
  --default/--no-default    Include default options that have not been overriden (prefixed **)
  --global/--no-global      Include global options that have not been overriden (prefixed *)
  -v/--verbose              Include the option help string
  -h, --help  Show this message and exit.

```

```console
$ verdi config list --default --global --verbose
** some.default.option   value   description...
*  some.global.option    value   description...
   some.profile.option   value   description...
```

```console
$ verdi config show -h

Usage: verdi config show [OPTIONS] [CONFIG_NAME]

  Show an AiiDA options for the current profile.

Options:
  -h, --help  Show this message and exit.

```

```console
$ verdi config show some.option

description: This is the variable description ...
valid_type: int
valid_values: None
global_only: False
values:
  default: value
  global:  value
  profile: value
```

```console
$ verdi config set -h

Usage: verdi config set [OPTIONS] CONFIG_NAME CONFIG_VALUE

  Set an AiiDA option.

Options:
  --force       Override existing without confirmation.
  --global      Set the option configuration wide.
  -h, --help  Show this message and exit.

```

```console
$ verdi config unset -h

Usage: verdi config unset [OPTIONS] [CONFIG_NAMES]...

  Unset an AiiDA option.

Options:
  --force       Unset without confirmation.
  --global      Unset the option configuration wide.
  -h, --help  Show this message and exit.

```

thoughts?

chrisjsewell:
- Refactor the verdi configuration to use jsonschema for validation/defaults
- Full `config.json` validate on initiation
- Allow the `config.json` to contain a `$schema` key, that is preserved when storing new data
- Rename option keys (with migration) so they are just the same as the internal representation + rename `user.` fields to `autofill.user.` (https://github.com/aiidateam/aiida-core/issues/2701#issuecomment-654732926)
- Refactor `verdi config` into separate commands: list/get/set/show/unset
- Include deprecation for current `verdi config <KEY>`
- Deprecated `cache_config.yml`: auto-merged into `config.json` (with deprecation warning), then renamed
- CLI to set/append/remove enabled/disabled cache identifiers
- CLI to list all process entry points that are enabled/disabled for caching 
- An `rmq.task_timeout` configuration variable has also been added (with default increased from 5 to 10 seconds), to fix timeout errors at high process loads.

~~Note importlib.resources requires python 3.7~~ python 3.6 dropped

closes #4584, closes #4240 and closes #2701
(also relevant #4243)

Open questions:

- [x] Add more failed validation tests (full config, set option, adding cache identifiers)
- [x] ~~Check with others validating config on read ok (or warning?)~~ checked and people happy with exception
- [x] validate caching identifiers on read? difficult as not part of the jsonschema
- [x] Arrays from CLI split by space (currently) or comma?
- [x] ~~User confirmation of set/unset?~~ (see https://github.com/aiidateam/aiida-core/pull/4712#issuecomment-777424303)
- [x] ~~Do we really need to save a backup for every option change?~~ Leave for now
- [x] Rename cache_config.yml after merge
- [x] How to handle profiles in cache_config.yml that are missing in config.json (they are skipped)?
- [x] check migration compatible version (version 5?)
- [x] ~~Rename other keys, to remove `Profile._map_config_to_internal`?~~ Probably not in this PR
- [x] Add more tests
- [x] auto add `$schema` to config (needs stable URL location)
- [x] documentation update

Also should consider #4523, i.e. not dropping unknown keys 

- [ ] open a new issue to resolve some of the less important things above


chrisjsewell:
examples:

```console
$ verdi config list
name                                   source    value
-------------------------------------  --------  ------------
autofill.user.email                    global    abc@test.com
autofill.user.first_name               global    chris
autofill.user.institution              global    epfl
autofill.user.last_name                global    sewell
daemon.default.workers                 default   1
daemon.timeout                         default   20
daemon.worker_process_slots            default   200
db.batch_size                          default   100000
logging.aiida_loglevel                 default   REPORT
logging.alembic_loglevel               default   WARNING
logging.circus_loglevel                default   INFO
logging.db_loglevel                    default   REPORT
logging.kiwipy_loglevel                default   WARNING
logging.paramiko_loglevel              default   WARNING
logging.plumpy_loglevel                default   WARNING
logging.sqlalchemy_loglevel            default   WARNING
runner.poll.interval                   profile   1
transport.task_maximum_attempts        global    6
transport.task_retry_initial_interval  default   20
verdi.shell.auto_import                default
warnings.showdeprecations              default   True
```

```console
$ verdi config list --description
name                                   source    value         description
-------------------------------------  --------  ------------  ----------------------------------------------------------------------
autofill.user.email                    global    abc@test.com  Default user email to use when creating new profiles.
autofill.user.first_name               global    chris         Default user first name to use when creating new profiles.
autofill.user.institution              global    epfl          Default user institution to use when creating new profiles.
autofill.user.last_name                global    sewell        Default user last name to use when creating new profiles.
daemon.default.workers                 default   1             The default number of workers to be launched by `verdi daemon start`
daemon.timeout                         default   20            The timeout in seconds for calls to the circus client
daemon.worker_process_slots            default   200           The maximum number of concurrent process tasks that each daemon worker
                                                               can handle
db.batch_size                          default   100000        Batch size for bulk CREATE operations in the database. Avoids hitting
                                                               MaxAllocSize of PostgreSQL (1GB) when creating large numbers of
                                                               database records in one go.
logging.aiida_loglevel                 default   REPORT        Minimum level to log to daemon log and the `DbLog` table for the
                                                               `aiida` logger
logging.alembic_loglevel               default   WARNING       Minimum level to log to daemon log and the `DbLog` table for the
                                                               `alembic` logger
logging.circus_loglevel                default   INFO          Minimum level to log to daemon log and the `DbLog` table for the
                                                               `circus` logger
logging.db_loglevel                    default   REPORT        Minimum level to log to the DbLog table
logging.kiwipy_loglevel                default   WARNING       Minimum level to log to daemon log and the `DbLog` table for the
                                                               `kiwipy` logger
logging.paramiko_loglevel              default   WARNING       Minimum level to log to daemon log and the `DbLog` table for the
                                                               `paramiko` logger
logging.plumpy_loglevel                default   WARNING       Minimum level to log to daemon log and the `DbLog` table for the
                                                               `plumpy` logger
logging.sqlalchemy_loglevel            default   WARNING       Minimum level to log to daemon log and the `DbLog` table for the
                                                               `sqlalchemy` logger
runner.poll.interval                   profile   1             The polling interval in seconds to be used by process runners
transport.task_maximum_attempts        global    6             Maximum number of transport task attempts before a Process is Paused.
transport.task_retry_initial_interval  default   20            Initial time interval for the exponential backoff mechanism.
verdi.shell.auto_import                default                 Additional modules/functions/classes to be automatically loaded in
                                                               `verdi shell`
warnings.showdeprecations              default   True          Boolean whether to print AiiDA deprecation warnings
```

chrisjsewell:
Ok, apart from updating the documentation, this is ready for review

chrisjsewell:
also cc'ing @greschd, as I know you worked on the caching

chrisjsewell:
Just noticed #4245, which this supersedes. But might be some stuff to steal from it, like this which might be cleaner than the similar one I created

```python
@pytest.fixture
def create_config_instance(create_empty_config_instance, create_profile) -> Config:
    """Create a temporary configuration instance with one profile.
    This creates a temporary directory with a clean `.aiida` folder and basic configuration file. The currently loaded
    configuration and profile are stored in memory and are automatically restored at the end of this context manager.
    This fixture differs from `create_empty_config_instance` in that here a profile will be created and set as default.
    The defaults of the profile can be overridden in the callable, as well as whether it should be set as default. This
    fixture should probably be used for most tests that assume a basically configured setup.
    :return: a config instance with a configured profile.
    """

    def _create_config_instance(set_as_default=True, **kwargs):
        """Create a temporary configuration instance with one profile.
        :param set_as_default: whether to set the one profile as the default.
        :param kwargs: parameters that are forwarded to the `Profile` constructor.
        """
        profile = create_profile(**kwargs)
        config = create_empty_config_instance
        config.add_profile(profile)

        if set_as_default:
            config.set_default_profile(profile.name, overwrite=True)

        config.store()

        return config

    return _create_config_instance
```

ltalirz:
On minor nitpick - the following is not really self-explanatory (and this is what most users will see)
```
caching.default                        default   False
caching.disabled                       default
caching.enabled                        default
```
It's not clear that the actual on/off switch is `caching.default` and that the other keys ask for lists.

Perhaps something like this
```
caching.enabled_globally        default   False
caching.disabled_for               default
caching.enabled_for                default
```
(open to better suggestions)

ltalirz:
As for the question concerning user confirmation, I would skip it.
Most `set` operations performed by users are likely performed after `list` / `show` , i.e. if they mistakenly delete something they will still have it in the terminal (and in the worst case there's the backup).

tests/manage/test_caching_config.py
```diff
@@ -126,15 +155,17 @@ def test_invalid_configuration_dict(configure_caching, config_dict):
             pass
 
 
-def test_invalid_identifier(use_default_configuration):  # pylint: disable=unused-argument
+def test_invalid_identifier(configure_caching):  # pylint: disable=unused-argument
```
yakutovicha, line 158:
```suggestion
def test_invalid_identifier(configure_caching):
```

tests/manage/test_caching_config.py
```diff
@@ -126,15 +155,17 @@ def test_invalid_configuration_dict(configure_caching, config_dict):
...
     """Test `get_use_cache` raises a `TypeError` if identifier is not a string."""
-    with pytest.raises(TypeError):
-        get_use_cache(identifier=int)
+    with configure_caching({}):
+        with pytest.raises(TypeError):
+            get_use_cache(identifier=int)
 
 
-def test_default(use_default_configuration):  # pylint: disable=unused-argument
+def test_default(configure_caching):  # pylint: disable=unused-argument
```
yakutovicha, line 165:
```suggestion
def test_default(configure_caching):
```

yakutovicha:
@chrisjsewell, just fyi. @ltalirz kindly proposed to take over the review. And I am very happy about it ;).

aiida/manage/configuration/options.py
```diff
@@ -8,252 +8,131 @@
...
+        :raise: ConfigValidationError
+
+        """
+        # pylint: disable=too-many-branches
+        from .config import ConfigValidationError
+        from aiida.manage.caching import _validate_identifier_pattern
+
+        if cast:
+            try:
+                if self.valid_type == 'boolean':
```
ltalirz, line 72: do we really have to do this ourselves or is there some existing library function we can use?

just as an example, I believe this will accept an empty list as a boolean (false) value.

Nothing against the logic implemented here (it seems to work for what I wanted to do) but I'm wondering whether there would be a way for us not to introduce yet another dialect of literals

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                        "INFO",
+                        "DEBUG"
+                    ],
+                    "default": "INFO",
+                    "description": "Minimum level to log to daemon log and the `DbLog` table for the `circus` logger"
+                },
+                "warnings.showdeprecations": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Boolean whether to print AiiDA deprecation warnings"
```
ltalirz, line 152:
```suggestion
                    "description": "Whether to print AiiDA deprecation warnings"
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                    "minimum": 1,
+                    "description": "Initial time interval for the exponential backoff mechanism."
+                },
+                "transport.task_maximum_attempts": {
+                    "type": "integer",
+                    "default": 5,
+                    "minimum": 1,
+                    "description": "Maximum number of transport task attempts before a Process is Paused."
+                },
+                "caching.default": {
```
ltalirz, line 166: this we discussed in a separate comment

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                "AIIDADB_ENGINE",
+                "AIIDADB_HOST",
+                "AIIDADB_NAME",
+                "AIIDADB_PASS",
+                "AIIDADB_PORT",
+                "AIIDADB_USER"
+            ],
+            "properties": {
+                "PROFILE_UUID": {
+                    "description": "The profiles unique key",
```
ltalirz, line 223:
```suggestion
                    "description": "The profile's unique key",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                },
+                "AIIDADB_PASS": {
+                    "type": [
+                        "string",
+                        "null"
+                    ],
+                    "default": null
+                },
+                "broker_protocol": {
+                    "description": "the protocol to use when connecting to the RabbitMQ server",
```
ltalirz, line 269:
```suggestion
                    "description": "Protocol for connecting to the RabbitMQ server",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                    "type": "string",
+                    "enum": [
+                        "amqp",
+                        "amqps"
+                    ],
+                    "default": "amqp"
+                },
+                "broker_username": {
+                    "description": "the username for RabbitMQ authentication",
```
ltalirz, line 278:
```suggestion
                    "description": "Username for RabbitMQ authentication",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                    "type": "string",
+                    "default": "guest"
+                },
+                "broker_password": {
+                    "description": "the password for RabbitMQ authentication",
```
ltalirz, line 283:
```suggestion
                    "description": "Password for RabbitMQ authentication",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                    "type": "string",
+                    "default": "guest"
+                },
+                "broker_host": {
+                    "description": "the hostname of the RabbitMQ server",
```
ltalirz, line 288:
```suggestion
                    "description": "Hostname of the RabbitMQ server",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                    "type": "string",
+                    "default": "127.0.0.1"
+                },
+                "broker_port": {
+                    "description": "the port of the RabbitMQ server",
```
ltalirz, line 293:
```suggestion
                    "description": "Port of the RabbitMQ server",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                    "type": "integer",
+                    "minimum": 1,
+                    "default": 5672
+                },
+                "broker_virtual_host": {
+                    "description": "the RabbitMQ virtual host to connect to",
```
ltalirz, line 299:
```suggestion
                    "description": "RabbitMQ virtual host to connect to",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+        "CONFIG_VERSION": {
+            "description": "The configuration version",
+            "type": "object",
+            "required": [
+                "CURRENT",
+                "OLDEST_COMPATIBLE"
+            ],
+            "properties": {
+                "CURRENT": {
+                    "description": "the current version number",
```
ltalirz, line 328:
```suggestion
                    "description": "Version number of configuration file format",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                    "type": "integer",
+                    "const": 5
+                },
+                "OLDEST_COMPATIBLE": {
+                    "description": "the oldest backwards compatible version number",
```
ltalirz, line 333:
```suggestion
                    "description": "Version number of oldest configuration file format this file is compatible with",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+    "type": "object",
+    "definitions": {
+        "options": {
+            "type": "object",
+            "properties": {
+                "runner.poll.interval": {
+                    "type": "integer",
+                    "default": 60,
+                    "minimum": 0,
+                    "description": "The polling interval in seconds to be used by process runners"
```
ltalirz, line 13:
```suggestion
                    "description": "Polling interval in seconds to be used by process runners"
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                },
+                "daemon.default_workers": {
+                    "type": "integer",
+                    "default": 1,
+                    "minimum": 1,
+                    "description": "The default number of workers to be launched by `verdi daemon start`"
```
ltalirz, line 19:
```suggestion
                    "description": "Default number of workers to be launched by `verdi daemon start`"
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                },
+                "daemon.timeout": {
+                    "type": "integer",
+                    "default": 20,
+                    "minimum": 0,
+                    "description": "The timeout in seconds for calls to the circus client"
```
ltalirz, line 25:
```suggestion
                    "description": "Timeout in seconds for calls to the circus client"
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                },
+                "daemon.worker_process_slots": {
+                    "type": "integer",
+                    "default": 200,
+                    "minimum": 1,
+                    "description": "The maximum number of concurrent process tasks that each daemon worker can handle"
```
ltalirz, line 31:
```suggestion
                    "description": "Maximum number of concurrent process tasks that each daemon worker can handle"
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+                },
+                "OLDEST_COMPATIBLE": {
+                    "description": "the oldest backwards compatible version number",
+                    "type": "integer",
+                    "const": 5
+                }
+            }
+        },
+        "profiles": {
+            "description": "The configured profiles",
```
ltalirz, line 340:
```suggestion
            "description": "Configured profiles",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
...
+            "type": "object",
+            "patternProperties": {
+                ".+": {
+                    "$ref": "#/definitions/profile"
+                }
+            }
+        },
+        "default_profile": {
+            "description": "The default profile to use",
```
ltalirz, line 349:
```suggestion
            "description": "Default profile to use",
```

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -0,0 +1,357 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "description": "Schema for AiiDA configuration version 5",
```
ltalirz, line 3:
```suggestion
    "description": "Schema for AiiDA configuration files, format version 5",
```

aiida/cmdline/commands/cmd_config.py
```diff
@@ -8,46 +8,230 @@
...
-    elif value is None:
-        option_value = config.get_option(option.name, scope=scope, default=False)
-        if option_value:
-            echo.echo(f'{option_value}')
 
-    # Set the specified option
+@verdi_config.command('caching')
+@click.option('-d', '--disabled', is_flag=True, help='List disabled types instead.')
+def verdi_config_caching(disabled):
+    """List caching enabled process types for the current profile."""
```
ltalirz, line 208:
```suggestion
    """List caching-enabled process types for the current profile."""
```

tests/conftest.py
```diff
@@ -220,6 +220,44 @@ def _create_profile(name, **kwargs):
...
 
+@pytest.fixture
+def create_config_instance(create_empty_config_instance, create_profile) -> Config:
+    """Create a temporary configuration instance with one profile.
+    This creates a temporary directory with a clean `.aiida` folder and basic configuration file.
+    The currently loaded configuration and profile are stored in memory,
+    and are automatically restored at the end of this context manager.
+    This fixture differs from `create_empty_config_instance` in that here a profile will be created and set as default.
+    The defaults of the profile can be overridden in the callable, as well as whether it should be set as default.
+    This fixture should probably be used for most tests that assume a basically configured setup.
```
ltalirz, line 231: Can you expand on this a bit?

In particular, I think one should contrast this to the `aiida_profile` fixture that is used automatically by all tests (`autouse=True`) and will already have set up a profile and configuration

tests/manage/test_caching_config.py
```diff
@@ -11,74 +11,103 @@
...
+    current_config_path = current_config.dirpath
+    current_profile_name = configuration.PROFILE.name
+
+    try:
+        reset_profile()
+        configuration.CONFIG = None
+
+        # Create a temporary folder, set it as the current config directory path
+        settings.AIIDA_CONFIG_FOLDER = str(tmp_path)
+        config_dictionary = {
```
ltalirz, line 65: can one reuse the test sample 5.json here?

ltalirz (changes requested):
thanks @chrisjsewell !

Besides the comments I added, you may want to open another issue on the approach to validation in the cli in general.
Many of the values that are provided to the cli (e.g. in setup/quicksetup) end up in the config file and validation could then be done via the JSON schema instead of being duplicated in python code.

aiida/cmdline/commands/cmd_config.py
```diff
@@ -8,46 +8,230 @@
...
 import click
 
 from aiida.cmdline.commands.cmd_verdi import verdi
 from aiida.cmdline.params import arguments
-from aiida.cmdline.utils import echo
+from aiida.cmdline.utils import decorators, echo
 
 
-@verdi.command('config')
+class DeprecatedCommandGroup(click.Group):
```
chrisjsewell, line 20:
```suggestion
class DeprecateConfigCommandsGroup(click.Group):
```

I can't be bothered to make it more general lol, because it is just a transient solution for deprecation, that I doubt will be required more generally and will be removed come v2.0

aiida/cmdline/commands/cmd_config.py
```diff
@@ -8,46 +8,230 @@
...
+            ctx.obj.deprecated_name = cmd_name
+            cmd = click.Group.get_command(self, ctx, '_deprecated')
+            return cmd
+
+        ctx.fail(f"'{cmd_name}' is not a verdi config command.")
+
+        return None
+
+
+@verdi.group('config', cls=DeprecatedCommandGroup)
```
chrisjsewell, line 48:
```suggestion
@verdi.group('config', cls=_DeprecateConfigCommandsGroup)
```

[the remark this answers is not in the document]
chrisjsewell: I think when I tried, it did still give you the typo suggestions. But perhaps you also want to try. Again I didn't want to devote too much time to perfecting a deprecation pattern that will be removed in v2.0

aiida/cmdline/commands/cmd_config.py
```diff
@@ -8,46 +8,230 @@
...
+    from aiida.manage.configuration import Config, Profile
+
+    config: Config = ctx.obj.config
+    profile: Profile = ctx.obj.profile
+
+    option_values = config.get_options(profile.name)
+
+    def _join(val):
+        """split arrays into multiple lines."""
+        if isinstance(val, list):
```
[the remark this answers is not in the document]
chrisjsewell: I would say yes, because these values are read from JSON, arrays will always be lists

aiida/cmdline/commands/cmd_config.py
```diff
@@ -8,46 +8,230 @@
...
+    echo.echo_success(f"'{option.name}' unset {scope_text}")
 
-    # Get the specified option
-    elif value is None:
-        option_value = config.get_option(option.name, scope=scope, default=False)
-        if option_value:
-            echo.echo(f'{option_value}')
 
-    # Set the specified option
+@verdi_config.command('caching')
```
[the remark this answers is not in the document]
chrisjsewell: Yes it was a late addition that I felt would be useful

aiida/cmdline/commands/cmd_config.py
```diff
@@ -8,46 +8,230 @@
...
+        for entry_point in get_entry_point_names(group):
+            identifier = ENTRY_POINT_STRING_SEPARATOR.join([group, entry_point])
+            if get_use_cache(identifier=identifier):
+                if not disabled:
+                    echo.echo(identifier)
+            elif disabled:
+                echo.echo(identifier)
+
+
+@verdi_config.command('_deprecated', hidden=True)
```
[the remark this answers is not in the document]
chrisjsewell: This command is what is called for all the deprecated key commands (from `_DeprecateConfigCommandsGroup`), i.e. if you use `verdi config daemon.default_workers`, and will be removed in v2.0.
It is hidden (so does not show up in the `verdi config --help`) and I think there would be a very small chance that anyone would call it 😅

aiida/manage/caching.py
```diff
@@ -30,91 +26,121 @@
...
+        self._enable_all = True
+        self._disable_all = False
+
+    def disable_all(self):
+        self._enable_all = False
+        self._disable_all = True
+
+    def enable(self, identifier):
+        self._enable.append(identifier)
+        with suppress(ValueError):
```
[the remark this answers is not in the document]
chrisjsewell: yes it is, I only learned of it recently and feel it is a bit "cleaner" 😄

aiida/manage/configuration/config.py
```diff
@@ -8,16 +8,50 @@
...
+    ):
+        super().__init__(message)
+        self._message = message
+        self._keypath = keypath
+        self._filepath = filepath
+        self._schema = schema
+
+    def __str__(self) -> str:
+        prefix = f'{self._filepath}:' if self._filepath else ''
+        path = '/' + '/'.join(str(k) for k in self._keypath) + ': ' if self._keypath else ''
```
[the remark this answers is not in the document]
chrisjsewell: but the key names already have dots in (e.g. `daemon.default_workers`) so I think this would be confusing?

aiida/manage/configuration/migrations/migrations.py
```diff
@@ -98,10 +98,46 @@ def _3_add_message_broker(config):
     return config
 
 
+def _4_simplify_options(config):
+    """Remove unnecessary difference between file/internal representation of options"""
+    conversions = {
```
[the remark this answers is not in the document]
chrisjsewell: we don't use these options as command-line options though, and the dot versions are what is used in all the code base. So either you replace all the options in the code base or you replace them here in the config.json. Its just needless to be maintaining two representations and the translation code to switch between them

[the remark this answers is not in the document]
chrisjsewell: I'd note also that we use entry-points with dot delimitation, so this is very much equivalent

aiida/manage/configuration/options.py
chrisjsewell: I copied this logic from the existing code, but am happy to consider an alternative

aiida/manage/configuration/schema/config-v5.schema.json
chrisjsewell: TODO: see https://github.com/aiidateam/aiida-core/pull/4712#issuecomment-777418357

tests/conftest.py
chrisjsewell: Note I copied this from @sphuber's #4245, so maybe I could defer to him on that 😉

chrisjsewell:
thanks @ltalirz I've now addressed all the "easier" review comments, and will address the remaining in due course.

tests/conftest.py
sphuber: The idea of the fixture is that it should be used for tests that manipulate the config. For those tests we don't want to manipulate the _actual_ config that is being used for the test suite, but we want a dummy one. Otherwise, if the test fails and doesn't undo its changes properly, the actual config is broken and all other tests fail.

aiida/manage/caching.py
```diff
@@ -30,91 +26,121 @@
...
 
+    def __init__(self):
+        self._enable_all = False
+        self._disable_all = False
+        self._enable = []
+        self._disable = []
 
-def _get_config(config_file):
-    """Return the caching configuration.
+    def clear(self):
```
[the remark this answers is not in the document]
chrisjsewell: added in 6b026d3a961c4ad76426376e64d28b0b473f0c8a

aiida/manage/caching.py
```diff
@@ -30,91 +26,121 @@
...
-            config = yaml.safe_load(handle)[profile.name]
-    except (OSError, IOError, KeyError):
-        # No config file, or no config for this profile
-        return DEFAULT_CONFIG
+        # Check validity of enabled and disabled entries
+        try:
+            for identifier in enabled + disabled:
+                _validate_identifier_pattern(identifier=identifier)
+        except ValueError as exc:
+            raise exceptions.ConfigurationError('Invalid identifier pattern in enable or disable list.') from exc
```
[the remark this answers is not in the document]
chrisjsewell: I don't think this is necessary, because it is raised from an exception that specifies the exact identifier and why it is not valid

aiida/manage/caching.py
```diff
@@ -30,91 +26,121 @@
...
-    ConfigKeys.DEFAULT.value: False,
-    ConfigKeys.ENABLED.value: [],
-    ConfigKeys.DISABLED.value: [],
-}
+class _ContextCache:
+    """Cache options, accounting for when in enable_caching or disable_caching contexts."""
 
+    def __init__(self):
+        self._enable_all = False
+        self._disable_all = False
```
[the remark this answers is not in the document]
chrisjsewell: done 093ac73111d7a4aa963e751c30b478be643fb479

tests/conftest.py
chrisjsewell: cheers for the clarification

aiida/manage/configuration/__init__.py
```diff
@@ -101,9 +101,44 @@ def load_config(create=False):
...
+
+    warnings.warn(
+        f'cache_config.yml use is deprecated, merging into config.json and moving to: {cache_path_backup}',
+        AiidaDeprecationWarning
+    )
+    import yaml
+    with open(cache_path, 'r', encoding='utf8') as handle:
+        cache_config = yaml.safe_load(handle)
+    for profile_name, data in cache_config.items():
+        if profile_name not in config.profile_names:
```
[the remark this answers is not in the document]
chrisjsewell: added in 2f38b6259430f9ea1b5db02d333fde5005919056

tests/manage/test_caching_config.py
chrisjsewell: done in d3291674c064165e147765708ae9cad1d363a279

chrisjsewell:
1a3a9586bd92adc0bac3892bbb61138c7f2b4106 has been added to fix #4745

aiida/manage/configuration/options.py
ltalirz: ok, then let's keep it for the moment

tests/conftest.py
```diff
@@ -220,6 +220,44 @@ def _create_profile(name, **kwargs):
     return _create_profile
 
 
+@pytest.fixture
+def create_config_instance(create_empty_config_instance, create_profile) -> Config:
```
ltalirz, line 224: The naming is a bit confusing, since some of these fixtures (`create_config_instance`) are factory-type fixtures (return a function) while others (`create_empty_config_instance`) return the instance.

If we keep the current implementation, I'd suggest something like

 * `create_empty_config_instance` => `empty_config`
 * `create_config_instance` => `config_with_profile_factory`
 * `create_profile` => `profile_factory`

tests/conftest.py
ltalirz, line 231: Ok, here my suggestion:

```suggestion
    
    This fixture builds on the `create_empty_config_instance`, adds one profile and set it as default.
    The defaults of the profile can be overridden in the callable, as well as whether it should be set as default.
    
    Example::
    
       ...
    
    As with `create_empty_config_instance`, the currently loaded configuration and profile are stored in memory,
    and are automatically restored at the end of this context manager.
    This fixture should be used by tests that modify aspects of the AiiDA configuration or profile
    and require a preconfigured profile.
```

ltalirz (changes requested):
Thanks @chrisjsewell - I resolved all but two remaining comments.

As for the open questions:
 * adding more failed validation tests: up to you
 * validate caching identifiers on read: we decided to keep this in python
 * check migration compatible version: not clear to me what this means
 * auto add $schema to config: this can be done after this PR

tests/conftest.py
ltalirz: P.S. It's not very explicit in the code how the db backend of the profile is chosen.

It is important that the backend be the same one of the test profile - otherwise any operations on the DB will fail.

chrisjsewell:
thanks @ltalirz 

> check migration compatible version: not clear to me what this means

this was just to check that

```python
CURRENT_CONFIG_VERSION = 5
OLDEST_COMPATIBLE_CONFIG_VERSION = 5
```

is correct, i.e. I have updated both to the new version which I think is correct?

tests/conftest.py
sphuber: It's intentionally not mentioned because this fixture should not actually be used to "run" the profile. It should merely be used to test functionality that manipulates the content of the config folder, such as the `profile.json`. So I would propose just to change the docsting to make this more prominent. Then it is not a problem that the backend may be incorrect

sphuber:
Can I ask what the reasoning was to stick with JSON instead of migration to YAML? I personally feel that YAML would have been nicer for those that need to manipulate the config manually now and again and since we are now migrating anyway, might as well take the opportunity, no?

tests/conftest.py
ltalirz: Ok, yes then this should be stated clearly in the docstring of the fixture (probably also in the one for the empty config).

I was a bit confused since I saw the `load_profile` - but I guess that does not actually establish a database connection here (?)

chrisjsewell:
> Can I ask what the reasoning was to stick with JSON instead of migration to YAML? I personally feel that YAML would have been nicer for those that need to manipulate the config manually now and again and since we are now migrating anyway, might as well take the opportunity, no?

because with JSON you can add a `$schema` key, which gives you in VS Code (and I imagine in other modern editors) autocompletion/hover descriptions/validation
see https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings

e.g. copy this into a file in VS Code

```json
{
    "$schema": "https://raw.githubusercontent.com/chrisjsewell/aiida_core/verdi-config/aiida/manage/configuration/schema/config-v5.schema.json",
    "CONFIG_VERSION": {
        "CURRENT": 5,
        "OLDEST_COMPATIBLE": 5
    },
    "profiles": {
        "quicksetup": {
            "PROFILE_UUID": "48c21184d3df4bfba4bf3079e3aa05d0",
            "AIIDADB_BACKEND": "django",
            "AIIDADB_NAME": "quicksetup_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9",
            "AIIDADB_PORT": 5432,
            "AIIDADB_HOST": "localhost",
            "AIIDADB_USER": "aiida_qs_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9",
            "AIIDADB_PASS": "1zmaM4PITFA2qESd",
            "broker_protocol": "amqp",
            "broker_username": "guest",
            "broker_password": "guest",
            "broker_host": "127.0.0.1",
            "broker_port": 5672,
            "broker_virtual_host": "",
            "AIIDADB_REPOSITORY_URI": "file:///Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/.aiida/repository/quicksetup",
            "options": {},
            "default_user_email": "abc@test.com"
        }
    },
    "default_profile": "quicksetup",
    "options": {}
}
```

ltalirz:
See also https://github.com/aiidateam/aiida-core/issues/2701#issuecomment-777403558

sphuber:
> > Can I ask what the reasoning was to stick with JSON instead of migration to YAML? I personally feel that YAML would have been nicer for those that need to manipulate the config manually now and again and since we are now migrating anyway, might as well take the opportunity, no?
> 
> because with JSON you can add a `$schema` key, which gives you in VS Code (and I imagine in other modern editors) autocompletion/hover descriptions/validation

Who defines this standard though? I doubt the `$schema` key is a JSON standard? I don't think that VSCode specific functionality should dictate what we choose internally. Have you seen other modern editors that support the exact same convention?

tests/conftest.py
chrisjsewell: > I was a bit confused since I saw the load_profile - but I guess that does not actually establish a database connection here

I actually added that on top of @sphuber's original function, because some tests failed since `config.current_profile` returned `None`, e.g.

```python
_________________________________________________ TestVerdiConfig.test_config_remove_option _________________________________________________

self = <tests.cmdline.commands.test_config.TestVerdiConfig object at 0x7f90129bafd0>
run_cli_command = <function run_cli_command.<locals>._run_cli_command at 0x7f9012b857a0>

    def test_config_remove_option(self, run_cli_command):
        """Test the `verdi config set --remove` command when removing an option value."""
        config = get_config()
    
        option_name = 'caching.disabled_for'
>       config.set_option(option_name, ['x', 'y'], scope=config.current_profile.name)
E       AttributeError: 'NoneType' object has no attribute 'name'

tests/cmdline/commands/test_config.py:125: AttributeError
```

sphuber: correct `load_profile` just sets the global variable with the contents of the JSON settings, but it does not actually load the backend

chrisjsewell:
> Who defines this standard though? I doubt the $schema key is a JSON standard? 

Yes it is: https://json-schema.org/understanding-json-schema/reference/schema.html

ltalirz:
JSON schema is a widely used standard, particularly in the web (for validating responses of APIs).

The VSCode feature of using this to simplify editing the file is a nice gimmick but for me the main argument to be fine with JSON is the widely used validation scheme, combined with making it less and less necessary to edit the configuration file manually.

I still find yaml nicer for files that are intended to be edited by humans.

sphuber:
> > Who defines this standard though? I doubt the $schema key is a JSON standard?
> 
> Yes it is: https://json-schema.org/understanding-json-schema/reference/schema.html

Great, thanks for the link, fine by me then to stick with JSON.

tests/conftest.py
chrisjsewell: changed in ac172c4b8df88df4b8bdd4ba4baa1e7cb4959484

tests/conftest.py
chrisjsewell: ok updated in a9189eccb797affc59399998da17a900595bd1c3

ltalirz (approved):
thanks @chrisjsewell !

aiida/manage/configuration/migrations/migrations.py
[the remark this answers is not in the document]
chrisjsewell: yeh I just think a translation layer over complicates things and should not be used is possible.
TBH, I don't really like the fact that you have `AIIDADB_NAME` in the config.json, `--db-name` at the CLI and `db_name` in the CLI input YAML

sphuber:
Before we merge, I just noticed that you added a migration to homogenize the config keys in the file. Did you move the database keys to a `database` dictionary?

chrisjsewell:
> Did you move the database keys to a database dictionary?

sorry not quite sure what you mean by this?
the database keys, e.g. `AIIDADB_NAME` remain unchanged

chrisjsewell:
TBH If I was to do this from scratch I would rather be more aggressive in the config.json migration; remove the `"option: {}"` key nesting (because really they are all options), allow for all keys to have a global option, and rename all the keys like this.

```json
{
    "CONFIG_VERSION": {
        "CURRENT": 5,
        "OLDEST_COMPATIBLE": 5
    },
    "profiles": {
        "quicksetup": {
            "uuid": "48c21184d3df4bfba4bf3079e3aa05d0",
            "db.engine": "postgresql_psycopg2",
            "db.backend": "django",
            "db.name": "quicksetup_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9",
            "db.port": 5432,
            "db.host": "localhost",
            "db.user": "aiida_qs_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9",
            "db.password": "1zmaM4PITFA2qESd",
			"repo.uri": "file:///Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/.aiida/repository/quicksetup",
            "broker.protocol": "amqp",
            "broker.host": "127.0.0.1",
            "broker.port": 5672,
			"default_user_email": "abc@test.com",
            "runner.poll.interval": 50,
            "daemon.timeout": 20,
            "caching.default_enabled": true,
            "caching.enabled_for": [
                "aiida.calculations:quantumespresso.pw",
                "aiida.calculations:other.*"
            ],
            "caching.disabled_for": [
                "aiida.calculations:templatereplacer"
            ]
        }
    },
    "default_profile": "quicksetup",
    "global": {
        "broker.username": "guest",
        "broker.password": "guest",
        "broker.virtual_host": "",
        "runner.poll.interval": 10,
        "autofill.user.email": "abc@test.com",
        "autofill.user.first_name": "chris",
        "autofill.user.last_name": "sewell",
        "autofill.user.institution": "epfl"
    }
}
```

I guess the drawback of this is (as discussed in https://github.com/aiidateam/aiida-core/issues/4243#issuecomment-656551282) that it would be "too easy" to change and break the database connection configuration

chrisjsewell:
>  has the caching config now been merged into the config.json? Is there an automatic migration for this?

yes and yes, see https://github.com/chrisjsewell/aiida_core/blob/a9189eccb797affc59399998da17a900595bd1c3/aiida/manage/configuration/__init__.py#L109

chrisjsewell:
> Since you are adding a migration to clean up the keys, would have made sense to add this as well.

I could do this. didn't want to go too crazy too fast lol

I'm not sure if it could be a bit of a pain to convert to a nested structure, but maybe ok, e.g.

```json
'profile': {
    'database': {
        'name': ....
        'port': 5432
    },
    'broker': {
        'host': ....
    }
}
``` 

vs

```json
'profile': {
    'database.name': ....
    'database.port': 5432
    'broker.host': ....
}
``` 


@ltalirz what are your thoughts?

ltalirz:
> @ltalirz what are your thoughts?

I agree that switching everything to nesting would be the cleanest way, but I do know of a few cases where templates are used for config.json files e.g. [here](https://github.com/aiidateam/aiida-optimade/blob/develop/profiles/config.j2).
If I see correctly, this template would still work with the current changes in this PR, but it would break with the other changes.

At the same time, we don't really lose anything by pushing the complete rewrite of the config.json out to aiida 2.0.

sphuber:
I am also fine with pushing this to 2.0, given that the migration can be automated and the user shouldn't even notice. Thanks for the additional info, fine for me to go ahead with the merge.

chrisjsewell:
> If I see correctly, this template would still work with the current changes in this PR, but it would break with the other changes.

Wouldn't this just be migrated?

> At the same time, we don't really lose anything by pushing the complete rewrite of the config.json out to aiida 2.0.

but yeh lets put a pin in it for now, and I'll open an issue to note the loose ends of this PR

ltalirz:
> Wouldn't this just be migrated?

Hm, I had not thought of this. You are right - there might then be some warning every time the container starts, but migration should still work.

> but yeh lets put a pin in it for now, and I'll open an issue to note the loose ends of this PR

Thanks.

PR #4713, opened 2021-02-08
Switch matrix order in continuous-integration tests job.

csadorf:
To harmonize with test-install workflow.


ltalirz (changes requested):
Fine! just needs an update

ltalirz (approved):

PR #4714, opened 2021-02-08
bump jenkins dockerimage to 20.04

ltalirz:
Despite python3.7 being installed on the Jenkins dockerimage, pip
install failed after dropping python 3.6 support (likely because pip
from python 3.6 was being used).

We update ubuntu to 20.04, which comes with python 3.8.2 by default.

### DIFF
--- a/.ci/Dockerfile
+++ b/.ci/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 MAINTAINER AiiDA Team <info@aiida.net>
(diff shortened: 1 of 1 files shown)


csadorf (commented):
This looks good to me, however I am not familiar with the jenkins setup, so I'd prefer if @chrisjsewell could also have a look at this.

chrisjsewell:
yeh we've already talked about it; if it runs correctly then its fine

chrisjsewell (approved):

Issue #4668, opened 2021-01-21
`jedi==0.18.0` is not compatible with `ipython==7.19.0`


mbercx:
### Describe the bug

The latest `jedi` release (`0.18.0`) seems to be incompatible with `ipython==7.19.0`. I first noticed this when tab-completion no longer worked in Jupyterlab, but it seems to be causing issues in the `verdi shell` as well:

```python
Python 3.7.9 (default, Aug 18 2020, 06:22:45) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.19.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from aiida.engine import calcfunction
   ...: @calcfunction
   ...: def multiply(x, y):
   ...:     return x * y
   ...: 

In [2]: z = multiply(Int(2), Int(3))
Traceback (most recent call last):
  File "/home/max/.virtualenvs/dev-test/lib/python3.7/site-packages/IPython/terminal/ptutils.py", line 113, in get_completions
    yield from self._get_completions(body, offset, cursor_position, self.ipy_completer)
  File "/home/max/.virtualenvs/dev-test/lib/python3.7/site-packages/IPython/terminal/ptutils.py", line 129, in _get_completions
    for c in completions:
  File "/home/max/.virtualenvs/dev-test/lib/python3.7/site-packages/IPython/core/completer.py", line 438, in _deduplicate_completions
    completions = list(completions)
  File "/home/max/.virtualenvs/dev-test/lib/python3.7/site-packages/IPython/core/completer.py", line 1818, in completions
    for c in self._completions(text, offset, _timeout=self.jedi_compute_type_timeout/1000):
  File "/home/max/.virtualenvs/dev-test/lib/python3.7/site-packages/IPython/core/completer.py", line 1862, in _completions
    full_text=full_text, cursor_line=cursor_line, cursor_pos=cursor_column)
  File "/home/max/.virtualenvs/dev-test/lib/python3.7/site-packages/IPython/core/completer.py", line 2030, in _complete
    cursor_pos, cursor_line, full_text)
  File "/home/max/.virtualenvs/dev-test/lib/python3.7/site-packages/IPython/core/completer.py", line 1374, in _jedi_matches
    text[:offset], namespaces, column=cursor_column, line=cursor_line + 1)
  File "/home/max/.virtualenvs/dev-test/lib/python3.7/site-packages/jedi/api/__init__.py", line 726, in __init__
[...]

PR #4715, opened 2021-02-08
Dependencies: Require `ipython~=7.20`

csadorf:
Package jedi version 0.18 introduces backwards incompatible changes that
break compatibility with ipython<7.20.

Fixes issue #4668.


ltalirz (approved):
fine with me as soon as tests pass

PR #4717, opened 2021-02-08
CI: Update tested Python versions in rabbitmq GA workflow.

csadorf:


csadorf:
Did not realize that this was referring to the RabbitMQ versions.

chrisjsewell:
lol I had the same thought earlier

Issue #4723, opened 2021-02-09
Remove deprecated import from 'collections' instead of 'collections.abc' module


csadorf:
I found at least two more issues with py 3.9 compatibility.

First, I created this: DropD/reentry#57 that needs to be fixed first.

Second: I have (from `pip freeze`): `SQLAlchemy-Utils==0.34.2` that matches our requirements in the setup.json (`sqlalchemy-utils~=0.34.2`). 
However, with this requirement I get:
```
tests/transports/test_all_plugins.py::TestBasicFunctionality::test_is_open
  /Users/pizzi/.virtualenvs/aiida-dev/lib/python3.8/site-packages/sqlalchemy_utils/observer.py:176: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import defaultdict, Iterable, namedtuple
```
I think this has been fixed in [this PR of them](https://github.com/kvesteri/sqlalchemy-utils/pull/397) and, looking [at the corresponding commit](https://github.com/kvesteri/sqlalchemy-utils/commit/45f4e4a3d83f7e45cdf449c7c2c4bfab659d47a7), this should have been released from 0.35.0 (see list of tags in the blue bar at the top of the page from GitHub). So we should update the dependency?

Probably this second issues is not caught in new venvs since the most recent version is installed, but I had an older venv.
[EDIT]: I tested and indeed 0.35.0 does not raise that warning anymore. However, I don't know if the version has been pinned to < 0.35 on purpose or not.

@csadorf I'm reopening this issue, but of course feel free to open another one instead moving my comments there.

_Originally posted by @giovannipizzi in https://github.com/aiidateam/aiida-core/issues/4302#issuecomment-775020859_

csadorf:
Related problem is the deprecated use of `frozendict` in `plumpy`, see: https://github.com/aiidateam/aiida-core/issues/4302#issuecomment-775031129

csadorf:
References:
 - https://docs.python.org/3.9/whatsnew/3.9.html#you-should-check-for-deprecationwarning-in-your-code
 - https://lwn.net/Articles/811369/

ltalirz:
reentry 1.3.2 has been released. updating the dependency in the CI requirements files should fix that warning

Issue #3845, opened 2020-03-11
QueryBuilder: add shortcut to EXPLAIN ANALYZE


ltalirz:
When optimizing SQL queries, using the [EXPLAIN ANALYZE](https://www.postgresql.org/docs/11/sql-explain.html) functionality of postgres can be very useful in understanding how table indices are used and how much time is spent in different aspects of a query.

Currently, it is quite cumbersome for users to provide the output of `EXPLAIN ANALYZE` to developers (they need to print the Querybuilder, log on to psql, paste it after `EXPLAIN ANALYZE`, put a semicolon and run).
It would be useful if the QueryBuilder had a shortcut to simply run `EXPLAIN ANALYZE` on its internal query and return the result as a string.

giovannipizzi:
Here is a possible implementation:

```python
from sqlalchemy.ext.compiler import compiles
from sqlalchemy.sql.expression import Executable, ClauseElement, _literal_as_text

# Adapted from http://www.wmmi.net/documents/SQLAlchemy.pdf
class ExplainAnalyze(Executable, ClauseElement):
    def __init__(self, stmt):
        self.statement = _literal_as_text(stmt)

@compiles(ExplainAnalyze, 'postgresql')
def pg_explain(element, compiler, **kw):
    text = "EXPLAIN ANALYZE "
    text += compiler.process(element.statement, **kw)
    return text

# I already call it self - this is what needs to be put as a new method of the QueryBuilder class
def get_queryplan(self):
    queryplan = self._impl.get_session().execute(ExplainAnalyze(self.get_query())).fetchall()

    queryplan_string = "\n".join(row.values()[0] for row in queryplan)

    return queryplan_string

from aiida.orm import Node, QueryBuilder

qb = QueryBuilder().append(Node).limit(10)
print(get_queryplan(qb))
```

If you like the idea, it's just a matter of copy-pasting the code into `aiida/orm/querybuilder.py` (putting the `get_queryplan()` function as a method of `QueryBuilder`, and then the final call would become simply `print(qb.get_queryplan())`.

@lekah do you see issues with this or it's ok to implement? It's very Postgres specific, but at the moment anyway we only support that...

Another option (maybe better?) is to push this into the `_impl` classes, then have the `get_queryplan` method in the top class call through, and have the default base class implementation return a NotImplementedError?

giovannipizzi:
Note that probably #3844 needs to  be fixed first before this will work, I think - maybe they could be fixed at the same time?

[...]

PR #4724, opened 2021-02-09
Upgrading dependency of sqlalchemy-utils

giovannipizzi:
In sqlalchemy-utils 0.35, imports from collections where correctly
fixed to import from collections.abc (where this is needed).
This removes a few deprecation warnings (claiming that this will not
work in py 3.9, even if in reality this will stop working in py 3.10).
This partially addresses #4723.

We are actually pinning to >=0.36 since in 0.36 a feature was dropped
that we were planning to use (see #3845). In this way, we avoid relying
on a feature that is removed in later versions (risking to implement
something that then we have to remove, or even worse remain "pinned"
to an old version of sqlalchemy-utils because nobody has the time
to fix it with a different implementation [which is tricky, requires
some knowledge of how SqlAlchemy and PosgreSQL work]).


csadorf:
We need to wait for the `test-install` workflow to complete, unfortunately it timed out last time. We might need to increase the timeout.

csadorf:
> We need to wait for the `test-install` workflow to complete, unfortunately it timed out last time. We might need to increase the timeout.

There appears to be an issue with one of the plugin tests. The test suite should complete [within roughly 10 minutes](https://pipelines.actions.githubusercontent.com/lx6f5qJRWQcbiE7vU42yTxrR7aWtThmdNBqldi7jyXdZiiVFpF/_apis/pipelines/1/runs/8583/signedlogcontent/44?urlExpires=2021-02-09T12%3A51%3A32.3505548Z&urlSigningMethod=HMACV1&urlSignature=yFSfQje6ErONTKT74xJ7nF9cfPKpzT8IeAz3aAYWMSA%3D), but the [test times out ](https://pipelines.actions.githubusercontent.com/lx6f5qJRWQcbiE7vU42yTxrR7aWtThmdNBqldi7jyXdZiiVFpF/_apis/pipelines/1/runs/8583/signedlogcontent/45?urlExpires=2021-02-09T12%3A51%3A24.0036996Z&urlSigningMethod=HMACV1&urlSignature=sXPbElk%2FzDGAY0%2FjqEQvFiffGJoqvkeVZPrLrVVYQN0%3D) for the `(py37, django)` combination.

csadorf (approved):

PR #4725, opened 2021-02-09
🔧 MAINTAIN: drop setuptools upper pinning

### LINKED ISSUE #4349: Fix workaround for setuptools version limitation
In PR #4342 we merged a workaround for a bug in the python package ecosystem, involving `pip` and `setuptools`. Currently by not installing `setuptools==0.50.0` fixes the failing builds on Jenkins, but we need to fix this at some point. People have suggested that the actual bug is due to `pip` messing up with its isolated build environment when installing in a virtual env with `--system-site-packages`. Apparently in this case there is cross talk between setuptools installed by `pip` and a version that exists on the system. We don't use virtual envs on the Jenkins machine, but we do install in user space with `pip` using `--user`. The same problem might be at play here. Since on GHA we did not experience problems, where we do in fully isolated virtual envs, simply installing in a virtualenv on Jenkins may fix the problem.

chrisjsewell:
closes #4349

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=40.8.0,<50", "wheel", "reentry~=1.3", "fastentrypoints~=0.12"]
+requires = ["setuptools>=40.8.0", "wheel", "reentry~=1.3", "fastentrypoints~=0.12"]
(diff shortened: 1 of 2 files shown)


csadorf (approved):
Looks like it's all good! 👍

PR #4727, opened 2021-02-09
Allow nested path in symlink, automatically create directory

### LINKED ISSUE #4726: Allow nested remote symlinks
Currently, remote symlinks are only possible for files which are at the top-level of the calculation working directory. In `aiida-quantumespresso`, we have come across a use case (https://github.com/aiidateam/aiida-quantumespresso/issues/618, https://github.com/aiidateam/aiida-quantumespresso/issues/620) for symlinking files that are nested in a subdirectory.

Comments by @giovannipizzi in https://github.com/aiidateam/aiida-quantumespresso/issues/618#issuecomment-731780051_

A few comments and further suggestions:
- if we fix the ssh plugin, we should also fix the local one
- once we go this way, I think the best approach is to create all folders recursively? And then just symlink files? I think we might already have all the logical pieces (e.g. to check if it's a file or a folder). This would solve all problems
- rather than relying on the existence of a globbing character, we should add an optional parameter (that by default keeps the old behaviour) that says "recursively create folders and symlink only files" (with a shorter name ;-))
- Have a way to specify the value of this optional new kwarg here:
  https://github.com/aiidateam/aiida-core/blob/17b77181d87abed9211fdaf1f432e00e276c1c11/aiida/engine/daemon/execmanager.py#L236
- Have a way for the plugin to specify the behaviour for symlinks (but also, for backward-compatibility, to know if the recursive feature is available). For the discoverability of the feature, in the worst case we can just rely on AiiDA's version. But I'm thinking to a few options:
  - have a new flag `remote_recursive_symlink_list` beside `remote_symlink_list` and `remote_copy_list`
  - have a new flag that can be specified at the CalcInfo level that says "use recursiveness for symlinks"

What do people think about this suggestion? Pinging a few people @sphuber @chrisjsewell - if we like the idea we should move this to aiida-core and fix it there first

greschd:
Fixes #4726 

The interface for this feature is still being discussed, hence WIP.


ltalirz:
@giovannipizzi Does that look like a reasonable step
 forward?

sphuber:
I just had a read through the entire stack of issues and the discussion, including on `aiida-quantumespresso` to try and understand the use case and I think I got it, but I may have missed some things (please correct me if I did). I think it would be useful to support the behavior necessary for the presented use case. However, I am not sure if we should do this in the `Transport` level. The reason is that we would be changing the interface and while we can fix the plugins we provide, other plugins may not implement these changes and then functionality provided by `aiida-core` will essentially be broken with other plugins.

Instead, we could simply add this to `aiida.engine.daemon.execmanager.upload_calculation`. This way the new functionality will always work, regardless of the transport plugin used.

Then, whether we need additional interface on the `CalcJob` and `CalcInfo` level: I am not sure we do. Is there a problem with simply _always_ creating the parent directories in the symlink target? If the `remote_symlink_list` supports specifying folders, files and paths with globs, the client can in principle specify anything they would need, wouldn't they? Taking the `aiida-quantumespresso` example, `pw.x` can simply specify the explicit files and not the `out` folder, such that the contents of the parent output folder are not changed. But other plugins that will treat the parent folder as read-only, they can simply specify the folder.

giovannipizzi:
Hi! Indeed, if we can create the folders in the upload_calculation this would be even better.
 
As Sebastiaan says, probably we can just create it when needed.
The client (=the plugin) can create the folder by hand in the `prepare_for_submission`, or specify the folder to symlink, or just a file (and then `upload_calculation` would create the missing parent folders if needed and not already there). I admit however that this was from many months ago so I might be missing something obvious :-D probably trying to implement it will tell us if we are missing something?

sphuber:
Superseded by #5974

PR #4728, opened 2021-02-09
`verdi group remove-nodes`: Add warning when nodes are not in `Group`

### LINKED ISSUE #4438: Removing nodes that aren't in Group raises no warning
Currently, removing nodes that aren't in a `Group` raises no error or warning. E.g. in the `verdi shell`:

```
In [1]: from aiida import orm
In [2]: group = orm.Group('test')
In [3]: group.store()
Out[3]: <Group: "test" [type core], of user mbercx@gmail.com>
In [4]: node = orm.load_node(1)
In [5]: group.remove_nodes(node)
```
The last command raises no error or warning, giving the user the impression that the command has been successfully executed. This behavior is even more confusing when using the CLI, since a prompt is raised that asks if you want to remove from the group:

```
$ verdi group remove-nodes -G test 1
Do you really want to remove 1 nodes from Group<test>? [y/N]: y
$
```

### Expected behavior

I would prefer to see a warning raised for each node that the user tries to remove from the group that is actually not in the group, similar to the linux `rm` command. Nodes that *are* in the group are still removed, however.

The behavior should be the same for the Python command `Group.remove_nodes()` and CLI command `verdi group remove-nodes`. So perhaps this should be implemented at the `Group` or `BackendGroup` level?

As a side note: if the user tries to remove nodes that do not exist in the database, an error is raised and no nodes are removed from the group. Perhaps the behavior here should be similar to the case where the node is not in the group? 

Also see [this discussion](https://github.com/aiidateam/aiida-core/pull/4428#discussion_r502295727) in #4428.

mbercx:
Fixes #4438

Currently the `verdi group remove-nodes` command does not raise any
warning when the nodes that the user wants to remove are not in the
group. It also says it removed the number of requested nodes from the
group, even when none of them are in the group specified.

Here we:

* Raise a warning when any of the nodes requested are not in the
specified group, and list the PK's of the nodes that are missing.
* Have the command fail with a `Critical` message when **neither** the
`--clear` flag nor any nodes are provided.
* Have the command fail with a `Critical` message when **both** the
`--clear` flag nor any nodes are provided.
* Have the command fail with a `Critical` message when none of the
requested nodes are in the group.

Note that the Group.remove_nodes() command still does not raise any
warning when the requested nodes are not in the group.

### DIFF
--- a/aiida/cmdline/commands/cmd_group.py
+++ b/aiida/cmdline/commands/cmd_group.py
@@ -52,6 +52,11 @@ def group_remove_nodes(group, nodes, clear, force):
     label = group.label
     klass = group.__class__.__name__
 
+    if nodes and clear:
+        echo.echo_critical(
+            'Specify either the `--clear` flag to remove all nodes or the identifiers of the nodes you want to remove.'
+        )
+
     if not force:
 
         if nodes:
--- a/aiida/cmdline/commands/cmd_group.py
+++ b/aiida/cmdline/commands/cmd_group.py
@@ -47,12 +47,30 @@ def group_add_nodes(group, force, nodes):
 @with_dbenv()
 def group_remove_nodes(group, nodes, clear, force):
     """Remove nodes from a group."""
-    if clear:
-        message = f'Do you really want to remove ALL the nodes from Group<{group.label}>?'
-    else:
-        message = f'Do you really want to remove {len(nodes)} nodes from Group<{group.label}>?'
+    from aiida.orm import QueryBuilder, Group, Node
 
     if not force:
+        node_pks = [node.pk for node in nodes]
+
+        query = QueryBuilder()
+        query.append(Group, filters={'id': group.pk}, tag='group')
+        query.append(Node, with_group='group', filters={'id': {'in': node_pks}}, project='id')
+
+        group_node_pks = query.all(flat=True)
+
+        if not group_node_pks:
+            echo.echo_critical(f'None of the specified nodes are in Group<{group.label}>.')
+            return
+
+        if len(node_pks) > len(group_node_pks):
+            node_pks = [node_pk for node_pk in node_pks if node_pk not in group_node_pks]
+            echo.echo_warning(f'{len(node_pks)} nodes with PK {node_pks} are not in Group<{group.label}>.')
+
+        if clear:
+            message = f'Are you sure you want to remove ALL the nodes from Group<{group.label}>?'
+        else:
+            message = f'Are you sure you want to remove {len(group_node_pks)} nodes from Group<{group.label}>?'
+
         click.confirm(message, abort=True)
 
     if clear:
--- a/aiida/cmdline/commands/cmd_group.py
+++ b/aiida/cmdline/commands/cmd_group.py
@@ -70,7 +70,6 @@ def group_remove_nodes(group, nodes, clear, force):
 
             if not group_node_pks:
                 echo.echo_critical(f'None of the specified nodes are in {klass}<{label}>.')
-                return
 
             if len(node_pks) > len(group_node_pks):
                 node_pks = set(node_pks).difference(set(group_node_pks))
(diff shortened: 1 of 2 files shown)


mbercx:
Example usage:

```console
 verdi group show test
-----------------  ----------------
Group label        test
Group type_string  core
Group description  <no description>
-----------------  ----------------
# Nodes:
  PK  Type    Created
----  ------  --------------
 157  Int     4h:04m:55s ago
 159  Int     4h:03m:33s ago
 161  Int     4h:02m:54s ago
```

When none of the requested nodes are in the group:

```console
$ verdi group remove-nodes -G test 100
Critical: None of the specified nodes are in Group<test>.
```

When some of the requested nodes are not in the group:

```console
$ verdi group remove-nodes -G test 100 101 159 
Warning: 2 nodes with PK [100, 101] are not in Group<test>.
Are you sure you want to remove 1 nodes from Group<test>? [y/N]: y
```

When all of the requested nodes are in the group:

```console
$ verdi group remove-nodes -G test 157 161
Are you sure you want to remove 2 nodes from Group<test>? [y/N]: y
```

aiida/cmdline/commands/cmd_group.py
```diff
@@ -47,12 +47,30 @@ def group_add_nodes(group, force, nodes):
...
+        query.append(Node, with_group='group', filters={'id': {'in': node_pks}}, project='id')
+
+        group_node_pks = query.all(flat=True)
+
+        if not group_node_pks:
+            echo.echo_critical(f'None of the specified nodes are in Group<{group.label}>.')
+            return
+
+        if len(node_pks) > len(group_node_pks):
+            node_pks = [node_pk for node_pk in node_pks if node_pk not in group_node_pks]
```
yakutovicha, line 66:
```suggestion
            node_pks = set(node_pk).difference(set(group_node_pks))
```
I think manipulating with sets in that context would be much faster comparing to python loops.

aiida/cmdline/commands/cmd_group.py
```diff
@@ -47,12 +47,30 @@ def group_add_nodes(group, force, nodes):
...
+    from aiida.orm import QueryBuilder, Group, Node
 
     if not force:
+        node_pks = [node.pk for node in nodes]
+
+        query = QueryBuilder()
+        query.append(Group, filters={'id': group.pk}, tag='group')
+        query.append(Node, with_group='group', filters={'id': {'in': node_pks}}, project='id')
+
+        group_node_pks = query.all(flat=True)
```
yakutovicha, line 59: Here I am not 100% sure which way is faster, the one you've implemented or something like:

```python
query = QueryBuilder()
query.append(Group, filters={'id': group.pk}, tag='group').append(Node)
group_nodes = set(query.all(flat=True))

group_nodes = group_nodes.intersect(nodes)
```

yakutovicha (commented):
Thanks @mbercx. I think we might consider using sets for that kind of operations. It appears to me that they might be faster when dealing with bigger arrays.

I proposed some possible code modifications below. Let me know what you think.

aiida/cmdline/commands/cmd_group.py
yakutovicha: I discovered an interesting bug while testing this: https://github.com/aiidateam/aiida-core/issues/4736

greschd: AiiDA nodes are not hashable (in the Python sense of having a `__hash__` method), so you can't put them in a `set`.

aiida/cmdline/commands/cmd_group.py
mbercx: Of course, that's quite a bit more sensible. 🤦  Will adjust!

yakutovicha (dismissed):
No complaints from my side =)

Good job @mbercx !

mbercx:
@yakutovicha thanks for the review! I've added the tests, so once you sign off on this it's good to go.

yakutovicha (approved):
Looks good to me, thanks @mbercx !

PR #4730, opened 2021-02-09
👌 IMPROVE: Allow numpy arrays to be serialized on process checkpoints

### LINKED ISSUE #3709: Error when loading checkpoint
After moving from ``1.0.0`` to the latest ``develop``, I'm getting errors where workchains except because their checkpoint could not be loaded (see full traceback below).

These are _new_ processes, i.e. they did not exist before the code update.

At least once, the issue occurred right after the daemon was restarted. This might just be the trigger for actually having to dump / load the checkpoint, though.

I have tried isolating a minimal failing example, but without luck so far. Will update once I've got an example.

@sphuber, could this be related to the switch to safe YAML loading?

```
2020-01-07 15:35:38 [44621 |  ERROR]: Traceback (most recent call last):
  File "/data/aiida/source/aiida-core/aiida/engine/persistence.py", line 124, in load_checkpoint
    bundle = serialize.deserialize(checkpoint)
  File "/data/aiida/source/aiida-core/aiida/orm/utils/serialize.py", line 230, in deserialize
    return yaml.load(serialized, Loader=AiiDALoader)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/__init__.py", line 114, in load
    return loader.get_single_data()
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 43, in get_single_data
    return self.construct_document(node)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 52, in construct_document
    for dummy in generator:
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 596, in construct_python_object
    state = self.construct_mapping(node, deep=deep)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 210, in construct_mapping
    return super().construct_mapping(node, deep=deep)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 135, in construct_mapping
    value = self.construct_object(value_node, deep=deep)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 92, in construct_object 
    data = constructor(self, node)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 420, in construct_undefined
    node.start_mark)
yaml.constructor.ConstructorError: could not determine a constructor for the tag 'tag:yaml.org,2002:python/object/apply: aiida.engine.processes.workchains.awaitable.AwaitableAction'
  in "<unicode string>", line 87, column 11:
      action: !!python/object/apply:aiida.engi ...
              ^

  File "/data/aiida/source/aiida-core/aiida/manage/external/rmq.py", line 187, in _continue
    result = yield super()._continue(communicator, pid, nowait, tag)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/tornado/concurrent.py", line 238, in result
  File "<string>", line 4, in raise_exc_info
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/tornado/gen.py", line 307, in wrapper
    yielded = next(result)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/plumpy/process_comms.py", line 546, in _continue 
    saved_state = self._persister.load_checkpoint(pid, tag)
  File "/data/aiida/source/aiida-core/aiida/engine/persistence.py", line 127, in load_checkpoint
    'Failed to load the checkpoint for process<{}>: {}'.format(pid, traceback.format_exc())
plumpy.exceptions.PersistenceError: Failed to load the checkpoint for process<55902>: Traceback (most recent call last):
  File "/data/aiida/source/aiida-core/aiida/engine/persistence.py", line 124, in load_checkpoint
    bundle = serialize.deserialize(checkpoint)
  File "/data/aiida/source/aiida-core/aiida/orm/utils/serialize.py", line 230, in deserialize
    return yaml.load(serialized, Loader=AiiDALoader)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/__init__.py", line 114, in load
    return loader.get_single_data()
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 43, in get_single_data
    return self.construct_document(node)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 52, in construct_document
    for dummy in generator:
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 596, in construct_python_object
    state = self.construct_mapping(node, deep=deep)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 210, in construct_mapping
    return super().construct_mapping(node, deep=deep)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 135, in construct_mapping
    value = self.construct_object(value_node, deep=deep)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 92, in construct_object    data = constructor(self, node)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/yaml/constructor.py", line 420, in construct_undefined
    node.start_mark)
yaml.constructor.ConstructorError: could not determine a constructor for the tag 'tag:yaml.org,2002:python/object/apply:aiida.engine.processes.workchains.awaitable.AwaitableAction'
  in "<unicode string>", line 87, column 11:
      action: !!python/object/apply:aiida.engi ...
```

greschd:
**NOTE:** This PR requires _extra careful_ review, both because it is security relevant, and because I am not very familiar with the import code.

To allow e.g. numpy arrays to be serialized to a process checkpoint, the `AiiDALoader` is based on `yaml.UnsafeLoader` instead of `yaml.FullLoader`. Since this could pose a security risk when sharing databases with maliciously crafted checkpoints, the checkpoints are removed upon importing an archive.

Fixes #3709.

Questions to consider:
- Is the `AiiDALoader` only used on checkpoints, or also somewhere else?
- Does the `node_type` always start with `process.` for nodes which have checkpoints?
- Anything else you can think of 😄

### DIFF
--- a/aiida/tools/importexport/dbimport/backends/common.py
+++ b/aiida/tools/importexport/dbimport/backends/common.py
@@ -13,7 +13,7 @@ from typing import Dict, List, Optional
 
 from aiida.common import timezone
 from aiida.common.progress_reporter import get_progress_reporter, create_callback
-from aiida.orm import Group, ImportGroup, Node, QueryBuilder
+from aiida.orm import Group, ImportGroup, Node, QueryBuilder, ProcessNode
 from aiida.tools.importexport.archive.readers import ArchiveReaderAbstract
 from aiida.tools.importexport.common import exceptions
 from aiida.tools.importexport.dbimport.utils import IMPORT_LOGGER
@@ -126,4 +126,18 @@ def _sanitize_extras(fields: dict) -> dict:
     fields['extras'] = {key: value for key, value in fields['extras'].items() if not key.startswith('_aiida_')}
     if fields.get('node_type', '').endswith('code.Code.'):
         fields['extras'] = {key: value for key, value in fields['extras'].items() if not key == 'hidden'}
+    return fields
+
+
+def _strip_checkpoints(fields: dict) -> dict:
+    """Remove checkpoint from attributes of process nodes.
+
+    :param fields: the database fields for the entity
+    """
+    if fields.get('node_type', '').startswith('process.'):
+        fields = copy.copy(fields)
+        try:
+            del fields['attributes'][ProcessNode.CHECKPOINT_KEY]
+        except KeyError:
+            pass
     return fields
--- a/aiida/tools/importexport/dbimport/backends/common.py
+++ b/aiida/tools/importexport/dbimport/backends/common.py
@@ -136,8 +136,7 @@ def _strip_checkpoints(fields: dict) -> dict:
     """
     if fields.get('node_type', '').startswith('process.'):
         fields = copy.copy(fields)
-        try:
-            del fields['attributes'][ProcessNode.CHECKPOINT_KEY]
-        except KeyError:
-            pass
+        fields['attributes'] = {
+            key: value for key, value in fields['attributes'].items() if key != ProcessNode.CHECKPOINT_KEY
+        }
     return fields
--- a/aiida/engine/persistence.py
+++ b/aiida/engine/persistence.py
@@ -121,7 +121,7 @@ class AiiDAPersister(plumpy.persistence.Persister):
             raise PersistenceError(f'Calculation<{calculation.pk}> does not have a saved checkpoint')
 
         try:
-            bundle = serialize.deserialize(checkpoint)
+            bundle = serialize.deserialize_unsafe(checkpoint)
         except Exception:
             raise PersistenceError(f'Failed to load the checkpoint for process<{pid}>: {traceback.format_exc()}')
 
--- a/aiida/engine/processes/process.py
+++ b/aiida/engine/processes/process.py
@@ -604,7 +604,7 @@ class Process(plumpy.processes.Process):
         :param encoded: encoded (serialized) inputs
         :return: The decoded input args
         """
-        return serialize.deserialize(encoded)
+        return serialize.deserialize_unsafe(encoded)
 
     def update_node_state(self, state: plumpy.process_states.State) -> None:
         self.update_outputs()
--- a/aiida/manage/manager.py
+++ b/aiida/manage/manager.py
@@ -264,7 +264,7 @@ class Manager:
         if with_orm:
             from aiida.orm.utils import serialize
             encoder = functools.partial(serialize.serialize, encoding='utf-8')
-            decoder = serialize.deserialize
+            decoder = serialize.deserialize_unsafe
         else:
             # used by verdi status to get a communicator without needing to load the dbenv
             from aiida.common import json
--- a/aiida/orm/utils/serialize.py
+++ b/aiida/orm/utils/serialize.py
@@ -176,13 +176,11 @@ class AiiDADumper(yaml.Dumper):
         return super().represent_data(data)
 
 
-class AiiDALoader(yaml.FullLoader):
+class AiiDALoader(yaml.UnsafeLoader):
     """AiiDA specific yaml loader
 
-    .. note:: we subclass the `FullLoader` which is the one that since `pyyaml>=5.1` is the loader that prevents
-        arbitrary code execution. Even though this is in principle only used internally, one could imagine someone
-        sharing a database with a maliciously crafted process instance dump, which when reloaded could execute arbitrary
-        code. This load prevents this: https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation
+    .. note:: The `AiiDALoader` should only be used on trusted input, because it uses the `yaml.UnsafeLoader`. When
+        importing a shared database, we strip all process node checkpoints to avoid this being a security risk.
     """
 
 
--- a/aiida/orm/utils/serialize.py
+++ b/aiida/orm/utils/serialize.py
@@ -217,10 +217,11 @@ def serialize(data, encoding=None):
     return serialized
 
 
-def deserialize(serialized):
+def deserialize_unsafe(serialized):
     """Deserialize a yaml dump that represents a serialized data structure.
 
-    .. note:: no need to use `yaml.safe_load` here because the `Loader` will ensure that loading is safe.
+    .. note:: This function should not be used on untrusted input, because
+        it is built upon `yaml.UnsafeLoader`.
 
     :param serialized: a yaml serialized string representation
     :return: the deserialized data structure
--- a/aiida/tools/importexport/dbimport/backends/django.py
+++ b/aiida/tools/importexport/dbimport/backends/django.py
@@ -32,7 +32,7 @@ from aiida.tools.importexport.archive.common import detect_archive_type
 from aiida.tools.importexport.archive.readers import ArchiveReaderAbstract, get_reader
 
 from aiida.tools.importexport.dbimport.backends.common import (
-    _copy_node_repositories, _make_import_group, _sanitize_extras, MAX_COMPUTERS, MAX_GROUPS
+    _copy_node_repositories, _make_import_group, _sanitize_extras, _strip_checkpoints, MAX_COMPUTERS, MAX_GROUPS
 )
 
 
@@ -355,6 +355,8 @@ def _select_entity_data(
                 if entity_name == NODE_ENTITY_NAME:
                     # format extras
                     fields = _sanitize_extras(fields)
+                    # strip checkpoints
+                    fields = _strip_checkpoints(fields)
                     if extras_mode_new != 'import':
                         fields.pop('extras', None)
                 new_entries[entity_name][str(pk)] = fields
--- a/aiida/tools/importexport/dbimport/backends/sqla.py
+++ b/aiida/tools/importexport/dbimport/backends/sqla.py
@@ -38,7 +38,7 @@ from aiida.tools.importexport.archive.common import detect_archive_type
 from aiida.tools.importexport.archive.readers import ArchiveReaderAbstract, get_reader
 
 from aiida.tools.importexport.dbimport.backends.common import (
-    _copy_node_repositories, _make_import_group, _sanitize_extras, MAX_COMPUTERS, MAX_GROUPS
+    _copy_node_repositories, _make_import_group, _sanitize_extras, _strip_checkpoints, MAX_COMPUTERS, MAX_GROUPS
 )
 
 
@@ -392,6 +392,8 @@ def _select_entity_data(
                 if entity_name == NODE_ENTITY_NAME:
                     # format extras
                     fields = _sanitize_extras(fields)
+                    # strip checkpoints
+                    fields = _strip_checkpoints(fields)
                     if extras_mode_new != 'import':
                         fields.pop('extras', None)
                 new_entries[entity_name][str(pk)] = fields
--- a/docs/source/nitpick-exceptions
+++ b/docs/source/nitpick-exceptions
@@ -138,8 +138,8 @@ py:class yaml.Dumper
 py:class yaml.Loader
 py:class yaml.dumper.Dumper
 py:class yaml.loader.Loader
-py:class yaml.FullLoader
-py:class yaml.loader.FullLoader
+py:class yaml.UnsafeLoader
+py:class yaml.loader.UnsafeLoader
 py:class uuid.UUID
 
 py:class psycopg2.extensions.cursor
--- a/environment.yml
+++ b/environment.yml
@@ -32,7 +32,7 @@ dependencies:
 - psycopg2>=2.8.3,~=2.8
 - python-dateutil~=2.8
 - pytz~=2019.3
-- pyyaml~=5.1.2
+- pyyaml~=5.1
 - reentry~=1.3
 - simplejson~=3.16
 - sqlalchemy-utils~=0.36.0
--- a/setup.json
+++ b/setup.json
@@ -46,7 +46,7 @@
         "psycopg2-binary~=2.8,>=2.8.3",
         "python-dateutil~=2.8",
         "pytz~=2019.3",
-        "pyyaml~=5.1.2",
+        "pyyaml~=5.1",
         "reentry~=1.3",
         "simplejson~=3.16",
         "sqlalchemy-utils~=0.36.0",
--- a/tests/common/test_serialize.py
+++ b/tests/common/test_serialize.py
@@ -9,6 +9,10 @@
 ###########################################################################
 """Serialization tests"""
 
+import types
+
+import numpy as np
+
 from aiida import orm
 from aiida.orm.utils import serialize
 from aiida.backends.testbase import AiidaTestCase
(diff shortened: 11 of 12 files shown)


greschd:
@sphuber do you know if we can remove the version constraint on `pyyaml` with this change?

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -108,3 +108,17 @@ def _sanitize_extras(fields: dict) -> dict:
...
         fields['extras'] = {key: value for key, value in fields['extras'].items() if not key == 'hidden'}
     return fields
+
+
+def _strip_checkpoints(fields: dict) -> dict:
+    """Remove checkpoint from attributes of process nodes.
+
+    :param fields: the database fields for the entity
+    """
+    fields = copy.copy(fields)
```
chrisjsewell, line 118: this should be moved inside the `if`, so we can save time/memory not having to copy the fields for every entity

chrisjsewell (changes requested):

greschd:
Not sure what's up with those missing tests.

chrisjsewell:
> Is the AiiDALoader only used on checkpoints, or also somewhere else?

the `deserializer` method and hence `AiiDALoader` is only used in two places:

```python
aiida/engine/persistence.py:
  117  
  118          checkpoint = calculation.checkpoint
  119  
  120          if checkpoint is None:
  121              raise PersistenceError(f'Calculation<{calculation.pk}> does not have a saved checkpoint')
  122  
  123          try:
  124:             bundle = serialize.deserialize(checkpoint)
  125          except Exception:
  126              raise PersistenceError(f'Failed to load the checkpoint for process<{pid}>: {traceback.format_exc()}')
  127  
  128          return bundle
  129  

aiida/engine/processes/process.py:
  599      def decode_input_args(self, encoded: str) -> Dict[str, Any]:  # pylint: disable=no-self-use
  600          """
  601          Decode saved input arguments as they came from the saved instance state Bundle
  602  
  603          :param encoded: encoded (serialized) inputs
  604          :return: The decoded input args
  605          """
  606:         return serialize.deserialize(encoded)
```

The `deserializer` method docstring should be changed, given at the moment it claims to be safe 😬 

```python
def deserialize(serialized):
    """Deserialize a yaml dump that represents a serialized data structure.

    .. note:: no need to use `yaml.safe_load` here because the `Loader` will ensure that loading is safe.

    :param serialized: a yaml serialized string representation
    :return: the deserialized data structure
    """
    return yaml.load(serialized, Loader=AiiDALoader)
```

Perhaps we should literally change the method name to `deserialize_unsafe`?

greschd:
> Perhaps we should literally change the method name to `deserialize_unsafe`?

I'd be on board with that, @sphuber what's your opinion?

sphuber:
> > Perhaps we should literally change the method name to `deserialize_unsafe`?
> 
> I'd be on board with that, @sphuber what's your opinion?

Would be fine by me. It was so far intended only for internal use and so changing it would be fine, although users may have been using the function as well. Maybe we keep the old `deserialize` with the safe loader and deprecate it, with a warning saying it is unsafe?

> Does the node_type always start with process. for nodes which have checkpoints?

In principle yes, but experience has taught that there are always cases that we didn't think of. Would the performance suffer enormously if we just apply the strip to all nodes? Just always popping the key checkpoints? Then again, there could be a data plugin out there that decided to store an attribute called `checkpoint`. Don't think this is actually the case, but still maybe best not risk it and better assume that the process nodes start with `process.`.

chrisjsewell:
> Maybe we keep the old deserialize with the safe loader and deprecate it, with a warning saying it is unsafe?

meh, thats a lot of code to leave lying around. I'd be inclined to say if people are using a function this deep in the API then that's their fault 😜

greschd:
> Then again, there could be a data plugin out there that decided to store an attribute called `checkpoint`.

Right, that was my reason for not doing it.

> Don't think this is actually the case, but still maybe best not risk it and better assume that the process nodes start with process..

I think the question is this: Can a "reasonable" (i.e., not malicious) plugin create processes with a different `node_type`, that could then later be exploited?

We could also think of renaming the `checkpoints` attribute key, to something like `_aiida_checkpoints`, and then strip it from all nodes. Would that need a migration, or are checkpoints anyway useless between AiiDA versions?

greschd:
I've done the renaming to `deserialize_untrusted` now. Indeed it's a bit of a pain to keep the old version around, because the `yaml.add_constructor` calls etc. need to be duplicated. Doable of course, but maybe not worth it.

sphuber:
> I've done the renaming to `deserialize_untrusted` now. Indeed it's a bit of a pain to keep the old version around, because the `yaml.add_constructor` calls etc. need to be duplicated. Doable of course, but maybe not worth it.

If you guys agree that it can be safely dropped, that is fine by me. I have just become more wary of this since I used to be overly trigger happy with breaking things before v1.0. In this case I agree that it can be argued that it is fine to simply change.

sphuber:
> We could also think of renaming the `checkpoints` attribute key, to something like `_aiida_checkpoints`, and then strip it from all nodes. Would that need a migration, or are checkpoints anyway useless between AiiDA versions?

I guess that technically we state that one should fully finish all active processes before upgrading (as well as make a full backup) but I doubt that many users actually do this, especially between minor or patch versions. Not performing a migration when changing the attribute name will break active processes as the code will look to deserialize the process from an attribute that has the wrong name. The migration would actually be really simple, we have already other examples of simply changing an attribute name, so the effort should be relatively limited, so it would definitely still be an option I think

greschd:
I've made a quick search in the plugins I've got lying around and didn't find a use of `deserialize`; I think @ltalirz has a snazzy way of doing this for all plugins on Github? Of course that's also no guarantee, but at least a good indicator.

Overall I agree on being careful with deprecations; I guess we can just keep it around for one minor version?

Agreed on the migration part -- so do you think it'd be better to rename and drop the key for all nodes?

ltalirz:
> I've made a quick search in the plugins I've got lying around and didn't find a use of `deserialize`; I think @ltalirz has a snazzy way of doing this for all plugins on Github? Of course that's also no guarantee, but at least a good indicator.

[I don't see any use of it](https://github.com/search?o=desc&p=1&q=aiida+%26+%22deserialize%22&s=indexed&type=Code)

chrisjsewell:
> I have just become more wary of this since I used to be overly trigger happy with breaking things before v1.0. In this case I agree that it can be argued that it is fine to simply change.

I agree to be diligent on deprecations (I've had to do enough in #4712 etc 😓) but on the other hand, we do clearly state our public API, and so should not be scared to change things that are not part of it.
Otherwise, we should start using `_` prefixed functions/methods more, so that they are even more clearly highlighted as for internal use only.

sphuber:
Thinking about more about the need of a migration of the `checkpoint` attribute. The reason I didn't prefix it with `_aiida` is because it is an attribute of a `ProcessNode` and since those cannot be subclassed through plugins, _all_ attributes are "internal" AiiDA ones. So I think it is fine to leave as is and just do as you proposed and only target nodes in the archives with a node type starting with `process.`. There is then a risk we miss nodes with the `checkpoint` attribute if somehow they have a faulty `node_type`, but then again, these won't be valid process nodes and won't be imported as such.

greschd:
Great, it seems the re-naming is not needed then.

Should I still add the deprecation for one minor version, or anything else that needs changing?

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -108,3 +108,17 @@ def _sanitize_extras(fields: dict) -> dict:
...
     return fields
+
+
+def _strip_checkpoints(fields: dict) -> dict:
+    """Remove checkpoint from attributes of process nodes.
+
+    :param fields: the database fields for the entity
+    """
+    if fields.get('node_type', '').startswith('process.'):
+        fields = copy.copy(fields)
```
sphuber, line 119: Why is the copy needed here? And if it is needed, shouldn't we use `deepcopy` because we are manipulating a key inside a nested dictionary so the `fields['attribute']` will still have a reference to the original object, wouldn't it?

sphuber (commented):
Just one comment, and a question: shouldn't we also not relax the requirement on `pyyaml`? The whole point was that it is currently pinned to `pyyaml~=5.1.2` and with this change I think we can start to use `5.2` and maybe upwards of that. We should likewise also fix this in `plumpy` because that currently has the same limitation. So maybe we should first fix this in `plumpy`, make a new release and then update the requirements for both `plumpy` and `pyyaml` in this PR before merging it.

Regarding the deprecations. Either we do it properly or we don't, I don't see the point of doing a half-assed deprecation. So I think it is fine to simply rename with deprecation as it currently does.

aiida/tools/importexport/dbimport/backends/common.py
greschd: > Why is the copy needed here?

TBH, I'm not sure why the copy is needed -- that is simply copy/pasted from the `_sanitize_extras` above. I don't have enough context on the whole import procedure to know if it's necessary.

>  the fields['attribute'] will still have a reference to the original object, wouldn't it

Right, good catch. Maybe a complete `deepcopy` is a bit overkill, but we could also `copy.copy` the `fields['attributes']`.

greschd:
> Just one comment, and a question: shouldn't we also not relax the requirement on `pyyaml`? The whole point was that it is currently pinned to `pyyaml~=5.1.2` and with this change I think we can start to use `5.2` and maybe upwards of that. We should likewise also fix this in `plumpy` because that currently has the same limitation. So maybe we should first fix this in `plumpy`, make a new release and then update the requirements for both `plumpy` and `pyyaml` in this PR before merging it.

Even at the 5.1.2 version, this change now allows for loading e.g. numpy arrays. But yeah, I think the order you suggested makes sense. What should the new version specifier be? `~=5.1`, `~=5.2`, `~=5.4`?

sphuber:
> What should the new version specifier be? `~=5.1`, `~=5.2`, `~=5.4`?

I would suggest `~=5.0` as our standard policy is to support all minor versions of a major release. Unless there is a very specific reason why we cannot or should not allow `5.0` and `5.1` I would allow them.

greschd:
Well, `5.0` [doesn't seem to exist](https://pypi.org/project/PyYAML/#history), so maybe `~=5.1`? Although of course there's no functional difference.

greschd:
> So maybe we should first fix this in `plumpy`, make a new release and then update the requirements for both `plumpy` and `pyyaml` in this PR before merging it.

With regards to `plumpy`, is there something in the code that needs changing, or just the requirement?

aiida/tools/importexport/dbimport/backends/common.py
greschd: Since we're now reconstructing the `attributes` dict, this should be resolved - but feel free to double-check @sphuber.

sphuber:
Sorry to only get back to this now @greschd . I guess, for these changes to be useful, this actually _does_ require changes in `plumpy` since that still specifies `pyyaml~=5.1.2` and so the relaxed requirements made here won't actually be usable. @chrisjsewell we should update `plumpy` to also be compatible with `pyyaml~=5.2` and use unsafe loaders and dumpers. Or at the very least, we should make it configurable which ones to use, so in `aiida-core` we can tell it to use the unsafe ones, while keeping the default at the safe ones. The latter would probably be best, but not sure how much work it is going to be. About release schedule, I think ideally we still include this in a `1.*` release. If it is too late for `1.6.0` maybe we can release it in a patch afterwards. What do you guys think?

greschd:
TBH I don't completely understand why this needs changes in `plumpy` - the `UnsafeLoader` is available in `pyyaml~=5.1.2`, so overriding it here changes which loader is used, no?

I've run this (only changing `FullLoader` to `UnsafeLoader`, not the full PR) in production to get around the limitation of not being able to store numpy arrays.

greschd:
AFAICT updating to a newer `pyyaml` version is a related but not absolutely necessary change.

sphuber:
I see what you mean, technically the changes here are compatible with the latest release of `plumpy`, however, you won't actually be able to install `pyyaml~=5.2` since there will be a dependency conflict. And even if you were still able to run with it, I am pretty sure `plumpy` would except, at least the tests will. Maybe we won't hit that code path in AiiDA usage. I realize that upgrading `pyyaml` is not necessary for the changes in this PR, but that was at least part of the goal. Having it pinned at `5.1` is not very good since the package is so generic and probably used by many other parts in the environment.

All that being said, I guess we can decouple them. We could merge this and then adjust plumpy in a separate PR. I was just wondering if that needed to be in conjunction, but I think it is ok to do it decoupled.

greschd:
Ok, since here we have the `pyyaml~=5.1` requirement, I guess we can merge this independently and then work on making `plumpy` compatible? Of course, it would be nice to test that this code, specifically, _is actually_ compatible with `5.2` before we relax the requirement here.

ramirezfranciscof:
Added the blocked tag because this would need an update in PlumPy.

greschd:
@ramirezfranciscof This is only "sort of" blocked by plumpy - we could merge it and fix #3709 without change in plumpy, but it would then probably be better to leave the `pyyaml` dependency at `pyyaml~=5.1.2`.

ltalirz:
@sphuber since you already reviewed this, can you give a final check + the green light for this PR?

Seems to me that the consensus is that this can be merged (while we will still need to wait for fixes in plumpy for actually being able to upgrade pyyaml).

ltalirz:
@greschd would you mind rebasing this on develop?

ltalirz:
Finally, a question for @greschd (sorry if that was already posed):
Are there any precautions one needs to take with an active database here?
I.e. is it possible that there will be some serialized data lying around that I can no longer deserialize after the update?

greschd:
> Are there any precautions one needs to take with active processes here?
> I.e. is it possible that there will be some serialized data lying around that I can no longer deserialize after the update?

Hmm, good question. In principle I think this shouldn't be an issue because we are switching from the more restrictive `FullLoader` to the more liberal `UnsafeLoader`. That means things that `FullLoader` will load should also load with `UnsafeLoader`.
That being said, I'm not sure if this reasoning is watertight, or there might be some edge cases.

greschd:
> @greschd would you mind rebasing this on develop?

done

chrisjsewell (approved):
right lets get this merged cheers! and I'll look at updating plumpy today

PR #4732, opened 2021-02-09
Add CalcJob test over SSH

ltalirz:
As mentioned in https://github.com/aiidateam/aiida-core/issues/4603#issuecomment-769039928 , the original memory leak was observed only when running CalcJobs on a remote computer via an SSH connection.

This PR adds a configuration for a remote computer (slurm container) and uses it to run a CalcJob over SSH.

PR #4735, opened 2021-02-09
Fix hanging direct scheduler+ssh

### LINKED ISSUE #3094: Direct scheduler in combination with `SshTransport` hangs on submit command
The ssh transport command to execute the scheduler submit will not return until the actual job is completed if the `direct` scheduler is used. The command that is executed is:
```
cd '/scratch/aiida/c7/13/9099-ea09-4914-b24f-524e0f4b586c' && bash -e '_aiidasubmit.sh' > /dev/null 2>&1 & echo $!
```
However, the `SshTransport.exec_command_wait` hangs at the line:
```
output_text = stdout.read().decode('utf-8')`
```
It seems that even though the launching of the submit script is backgrounded, the `stdout.read` will not finish until the background process has terminated. As soon as that happens, the command returns and the `CalcJob` continues from the `SUBMIT` task to the `UPDATE` task, at which point it is already finished of course.

giovannipizzi:
The fix is very simple: in the ssh transport, to emulate 'chdir',
we keep the current directory in memory, and we prepend every command
with a `cd FOLDER_NAME && ACTUALCOMMAND`.

One could put `;` instead of `&&`, but then if the folder does not
exist the ACTUALCOMMAND would still be run in the wrong folder, which is
very bad (imagine you are removing files...).

Now, in general this is not a problem. However, the direct scheduler
inserts a complex-syntax bash command to run the command in the background
and immediately get the PID of that process without waiting.
When combined with SSH, this hangs until the whole process is completed, unless
the actual command is wrapped in brackets.

A simple way to check this is running these two commands, that reproduce
the issue with plain ssh, without paramiko:

This hangs for 5 seconds:
```
ssh localhost 'cd tmp && sleep 5 > /dev/null 2>&1 & echo $!'
```

This returns immediately, as we want:
```
ssh localhost 'cd tmp && ( sleep 5 > /dev/null 2>&1 & echo $! )'
```

This fixes #3094


giovannipizzi:
I'm not sure if there is a simple/obvious way to test this... Do you guys feel that testing is crucial in this simple case, where the behaviour is also testable with plain SSH?

giovannipizzi:
Also pinging @ramirezfranciscof - beside reviewing this, if this gets fixed it in time, would be good to have in 1.6; otherwise we'll put it in the next patch release.

ltalirz:
I'm trying to add a test of a calcjob running over ssh (slurm) here https://github.com/aiidateam/aiida-core/pull/4732 (struggling a bit to get authentication to work)
If that works, then adding another one for a direct scheduler should be straightforward

giovannipizzi:
Thanks @ltalirz !

Indeed testing that in general everything works is great.

I guess that the second part that needs to be tested is that indeed the issue is fixed - i.e., that running now does not 'hang'.
Maybe once you're done with your test, one could just put [the logic I put in this comment](https://github.com/aiidateam/aiida-core/issues/3094#issuecomment-776060270) in a test (testing for various combinations of transport and scheduler), run e.g. `sleep 10`, and make sure the next python line is run in, say, < 5 sec? Hopefully using 5 seconds is robust enough even in slow test runners.

giovannipizzi:
Actually, for the direct scheduler, this should be easy and we don't need complex setup, but we could just reuse the same ssh infrastructure of current transport tests?

giovannipizzi:
Ok, I added the tests (and verified that it would have failed before the fix).
For me this is ready to review and merge!

ltalirz:
@giovannipizzi test on ssh container now merged https://github.com/aiidateam/aiida-core/pull/4732

giovannipizzi:
Thanks @ltalirz, this means that now that I merged, my changes to the transport (the brackets) are automatically tested in the new container with SSH+slurm, right?

If this is the case, and tests pass, there is nothing more I need to do - the regression tests for the actual issue at hand have been added in my last commit, and your new tests verify that I didn't break existing functionality (i.e. it's still possible to submit with ssh+slurm - if I made a mistake, those tests would fail).

Therefore, feel free to review and squash-merge!

ltalirz:
> Thanks @ltalirz, this means that now that I merged, my changes to the transport (the brackets) are automatically tested in the new container with SSH+slurm, right?

hehe, no - I just added one test that uses the new container ;-)
if you want to use it as well, you'll need to do a bit of work

giovannipizzi:
Yes, but I mean, when that test now runs, it will use AiiDA's ssh plugin + slurm plugin, no? We at least we're checking that for a simple submission, my change did not break existing functionality (for slurm + ssh). Which, for me, it's OK for this PR to be considered ready to be merged.

ltalirz:
Ah sorry, I misunderstood. Yes - it seems you did not break a basic job submission over SSH :-)

tests/transports/test_all_plugins.py
```diff
@@ -1365,3 +1279,73 @@ def test_exec_with_wrong_stdin(self, custom_transport):
...
+                psutil.pid_exists(job_id), 'The job is not there after a bit more than 1 second! Probably it failed'
+            )
+
+            # Clean up by killing the remote job (if the job succeeds, to avoid
+            # leaving arount it; if it doesn't, it's not a big deal anyway, it's a
+            # sleep that will die in < 10 seconds
+            os.kill(job_id, signal.SIGTERM)
+
+            # Also remove the script
+            os.remove(f'/tmp/{script_fname}')
```
ltalirz, line 1351: Maybe wrap this into a `try` / `finally` (or see suggestion above)

tests/transports/test_all_plugins.py
```diff
@@ -1365,3 +1279,73 @@ def test_exec_with_wrong_stdin(self, custom_transport):
...
+
+        scheduler = SchedulerFactory('direct')()
+        scheduler.set_transport(custom_transport)
+        with custom_transport as transport:
+            with tempfile.NamedTemporaryFile() as tmpf:
+                # Put a submission script that sleeps 10 seconds
+                tmpf.write(b'#!/bin/bash\nsleep 10\n')
+                tmpf.flush()
+
+                transport.chdir('/tmp')
```
ltalirz, line 1313: This is `localhost` here, right?

Instead of the file transfer, it could also just use the temporary file directly.
This would also take care of removing the file in case an exception is encountered.

ltalirz (changes requested):
thanks @giovannipizzi 

just minor comments

tests/transports/test_all_plugins.py
```diff
@@ -1365,3 +1279,73 @@ def test_exec_with_wrong_stdin(self, custom_transport):
...
+    @run_for_all_plugins
+    def test_asynchronous_execution(self, custom_transport):
+        """Test that the execution of a long(ish) command via the direct scheduler does not block.
+
+        This is a regression test for #3094, where running a long job on the direct scheduler
+        (via SSH) would lock the interpreter until the job was done.
+        """
+        # Use a unique name, using the current process PID as well, to avoid concurrent tests (or very rapid
+        # tests that follow each other) to overwrite the same destination
+        script_fname = f'sleep-submit-{os.getpid()}-{custom_transport.__class__.__name__}.sh'
```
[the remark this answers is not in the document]
giovannipizzi: Done

tests/transports/test_all_plugins.py
```diff
@@ -1365,3 +1279,73 @@ def test_exec_with_wrong_stdin(self, custom_transport):
...
+            # So I double check later that the execution was successful.
+            self.assertTrue(
+                elapsed_time < 5,
+                'Getting back control after remote execution took more than 5 seconds! Probably submission is blocking'
+            )
+
+            # Check that the job is still running
+            # Wait 1 more second, so that I don't do a super-quick check that might return True
+            # even if it's not sleeping
+            time.sleep(1)
```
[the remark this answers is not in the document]
giovannipizzi: Done

tests/transports/test_all_plugins.py
giovannipizzi: Done. I've also added additional try/except since the finally block could be called very early, e.g. before the file is created or the process is launched

tests/transports/test_all_plugins.py
giovannipizzi: Yes, but I've preferred to keep it ready to be extended to different machines (I also now remove the file via the transport in the finally block.
I still use `os.kill` for simplicity, though. It should be easy to just change that if we decide to extend testing to further machines (there are other tests that need to be adapted as well) but at least the changes to apply in that case remain confined.

ltalirz (approved):
thanks @giovannipizzi !

giovannipizzi:
They finally enabled "auto-merging"! :-) Let's see how it works

PR #4740, opened 2021-02-10
CI: Increase pytest verbosity in test-install GA workflow.

csadorf:


.github/workflows/test-install.yml
```diff
@@ -183,6 +183,7 @@ jobs:
     - name: Run test suite
       env:
         AIIDA_TEST_BACKEND: ${{ matrix.backend }}
+        PYTEST_ADDOPTS: "-v"
```
chrisjsewell, line 186: perhaps this should be added in `.github/workflows/tests.sh` instead, with the other `PYTEST_ADDOPTS` lines?
I'm not too bothered either way though

chrisjsewell (changes requested):

.github/workflows/test-install.yml
csadorf: I opted to put them here in case we don't want to increase verbosity for all workflows, but it's true, might make sense to increase verbosity in all test runs.

chrisjsewell (approved):
cheers

Issue #4374, opened 2020-09-16
Django backend (possibly) not closing PostgreSQL connections


CasperWA:
In using the [OPTIMADE Client](https://dev-tools.materialscloud.org/optimadeclient/) I found an issue with the Materials Cloud AiiDA-OPTIMADE servers that is noted to be an issue of not closing 100+ PostgreSQL database connections (see [related issue](https://github.com/aiidateam/aiida-optimade/issues/117)).

I was looking into the PR I created to handle the issue of closing SQLAlchemy sessions for the AiiDA REST API (#3974), and came across the following line's asymmetry between the backends, which might be fine, but might not be.

Django:

https://github.com/aiidateam/aiida-core/blob/b002a9a9a898a05338ff522b46686657fb468f66/aiida/backends/djsite/__init__.py#L65

SQLAlchemy:

https://github.com/aiidateam/aiida-core/blob/b002a9a9a898a05338ff522b46686657fb468f66/aiida/backends/sqlalchemy/__init__.py#L57

Could this be the cause of the issue? The database that caused the issue was indeed using a django backend.

Pinging @sphuber and @giovannipizzi

sphuber:
If you are referring to the `expire_on_commit` then most likely the answer is no. That flag merely determines whether the objects that are held in the cache by SqlAlchemy should be expired when the session is committed, meaning they will have to be refreshed from the database when operated on. This should not have anything to do with the connection being closed or not.

ltalirz:
The source of the problem is indeed that idle worker processes (REST APIs, optimade APIs, discover sections... [1]) keep their connection open, see writeup [here](https://github.com/aiidateam/aiida-optimade/issues/117#issuecomment-693723122).

I believe we now need to look into making sure that these web services close the postgresql connection when it is no longer needed.
One could think of offering a context manager replacement of `load_profile()` which takes care of closing the connection on exit (unloading the python modules is not needed). This would, however, require developers of AiiDA web applications to do explicit connection handling.
We should think about whether there is a way to get around this [2] and, if not, we need to pay great attention to make this as simple to use as possible.

P.S. Issue https://github.com/aiidateam/aiida-core/issues/3867 is related but concerns the use case of persistent connections and essentially asks for a way to recover if those connections are interrupted due to external reasons.

[1] The same probably holds in the AiiDA lab, where every jupyter notebook may have its own DB connection. Since these aren't threaded, however, one is unlikely to hit the 100 connection limit.

[...]

PR #4741, opened 2021-02-10
close django connection in REST API

ltalirz:
Note: This is a draft meant for testing e.g. on Materials Cloud.

In regular operation, AiiDA keeps several connections to the postgresql database persistently open.
One set of connections ([up to 5 by default](https://docs.sqlalchemy.org/en/14/core/pooling.html#sqlalchemy.pool.QueuePool)) comes from the SQLA connection pool, which is present on both sqla and django profiles (since both use the QueryBuilder).
For django profiles, there is an additional persistent connection opened by django.
See [here](https://github.com/aiidateam/aiida-core/issues/4374#issuecomment-694303925) for a short summary.

[The default connection limit of postgresql is 100](https://www.postgresql.org/docs/14/runtime-config-connection.html); see [here](https://brandur.org/postgres-connections) for why it is a good idea to keep the number of connections from exploding.
As long as AiiDA users don't use more than 100/6=16 daemon runners (that number would need to include AiiDA shells + REST APIs), they will not hit that limit in regular use.

For Materials Cloud, the situation is different, however.
Currently, Materials Cloud runs one independent AiiDA REST API process per profile, all of which connect to the same postgresql database, and some profiles include further AiiDA-based services that also make use of the REST API.
We've therefore hit the limit of 100 connections (and even more) some time ago.

The Materials Cloud use case is very different from "operational" AiiDA use: on Materials Cloud, database access is needed during user-initiated queries but there are long periods of inactivity between those (there are no running jobs that would access the database with high frequency).
The benefit from keeping postgresql connections open is therefore less relevant (as [shown here](https://github.com/aiidateam/aiida-core/issues/4374#issuecomment-701404620), opening a postgresql connection on materials cloud takes ~6-10ms).

This PR

 * Replaces the default `QueuePool` of sqla by a `NullPool` that will release a connection as soon as it is no longer needed
 * For django profiles: Closes the persistent Django connection at the end of every REST API request

I've tested that even storing ORM entities still works inside the REST
API, i.e. if needed the django connection is simply reestablished on the
fly.

### DIFF
--- a/aiida/backends/utils.py
+++ b/aiida/backends/utils.py
@@ -8,10 +8,21 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Backend-agnostic utility functions"""
+from sqlalchemy.pool import NullPool
 from aiida.backends import BACKEND_SQLA, BACKEND_DJANGO
 from aiida.manage import configuration
+from aiida.common import json
 
 AIIDA_ATTRIBUTE_SEP = '.'
+
+SQLALCHEMY_ENGINE_DEFAULTS = dict(
+    json_serializer=json.dumps,
+    json_deserializer=json.loads,
+    encoding='utf-8',
+    poolclass=NullPool,
+    # pool_size=1,
+    # max_overflow=0,
+)
 
 
 def create_sqlalchemy_engine(profile, **kwargs):
@@ -22,7 +33,6 @@ def create_sqlalchemy_engine(profile, **kwargs):
         more info.
     """
     from sqlalchemy import create_engine
-    from aiida.common import json
 
     # The hostname may be `None`, which is a valid value in the case of peer authentication for example. In this case
     # it should be converted to an empty string, because otherwise the `None` will be converted to string literal "None"
@@ -37,9 +47,9 @@ def create_sqlalchemy_engine(profile, **kwargs):
         port=profile.database_port,
         name=profile.database_name
     )
-    return create_engine(
-        engine_url, json_serializer=json.dumps, json_deserializer=json.loads, encoding='utf-8', **kwargs
-    )
+    kwargs_new = SQLALCHEMY_ENGINE_DEFAULTS.copy()
+    kwargs_new.update(kwargs)
+    return create_engine(engine_url, **kwargs_new)
 
 
 def create_scoped_session_factory(engine, **kwargs):
--- a/aiida/restapi/common/utils.py
+++ b/aiida/restapi/common/utils.py
@@ -858,3 +858,7 @@ def close_session(wrapped, _, args, kwargs):
         return wrapped(*args, **kwargs)
     finally:
         get_manager().get_backend().get_session().close()
+
+        from django.db import connection
+        # Close django connection if open (not needed by REST API)
+        connection.close()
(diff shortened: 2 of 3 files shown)


elsapassaro:
@ltalirz do you think that this PR could be merged? I'm setting up the new aiida servers for Materials Cloud with `aiida-core 1.6.5` and I'm getting again apache errors due to a limited number of connection slots. For now I'll cherry-pick these commits but it would be great to have these changes in the next AiiDA release

ltalirz:
thanks for the note @elsapassaro  - these changes were a quick hack meant for trying things out but if they help for materials cloud perhaps we can figure out a way to get them in.

I've just updated the description of the PR to explain the reasoning here in some detail.

The reason these changes should not be merged as-is is that they can affect regular AiiDA operation:
 1. Using a `NullPool` in sqla can slow down queries (one should check but I am confident it does - otherwise we should indeed use the nullpool)
 2. Similarly, on django profiles if someone makes a request to the REST API, the connection will be closed (it is reestablished automatically when needed, but again this can lead to slowdown)

I propose to add a `verdi config` option that allows users to tell AiiDA to minimize persistent connections (disabled by default).
If it is enabled, the changes made here would go into effect.

What is your perspective @sphuber ?

sphuber:
If this change can be made without affecting normal use of AiiDA, by making it configurable (preferably through the REST API config, instead of the global one, if this is purely for the REST API) then we could add this. It might be worth having a look at #5054 as a potential alternative though. If I understand correctly, the reason of the problems is that you are using a single REST API process per profile and so are running out of connections, but you are only doing that since it is currently not possible to switch profiles. PR #5054 adds that functionality and so would remove your original problem therefore not needing this workaround. @elsapassaro there are already other people testing that branch and if that works we will try to include it with the next release. If you could also test this with the MC that would help speed up the process.

ltalirz:
> making it configurable (preferably through the REST API config, instead of the global one, if this is purely for the REST API)

I guess it would in principle be possible to limit the scope to the REST API but I believe it would require some changes since the explicit creation of the postgresql connections currently does not happen inside the REST API code.
Furthermore, there are Materials Cloud discover sections that use AiiDA for querying the database via the python interface rather than through the REST API (similar to how the AiiDA lab works). 
I think we want these changes to apply globally.

> PR #5054 adds that functionality and so would remove your original problem therefore not needing this workaround.

Fully agree that if #5054 turns out to be performant enough that would be great and would remove the need for this PR.

In this PR (#4741) we just talking about opening/closing a postgresql connection per request (<10ms as mentioned above); in #5054 we are talking about potentially switching profiles per request which is likely orders of magnitude slower.
The only way to figure it out is to test 

@elsapassaro Since testing #5054  requires a migration to the new file repository, I think we should start testing performance locally. 
If I find time next week, I will test it locally (maybe after https://github.com/aiidateam/aiida-core/issues/5116 is addressed) and I suggest you give it a try locally with a small database.

Furthermore, it would be super useful to have a look at the request patterns to the AiiDA REST APIs on Materials Cloud.
In essence what we want to know is: if we combine the logs from all REST API daemons into one time-ordered log:
 * how long do requests typically take?
 * what would be the rate of profile switching?

sphuber:
Since we are getting rid of Django for v2.0, I am closing this for now. If you feel this needs a backport fix on `1.6.x` please reopen and rebase on `support/1.6.x`.

Issue #4722, opened 2021-02-08
Lots of asyncio warnings during test suite run


ltalirz:
After some warnings cleanup, the vast majority of [warnings emitted during runs of the test suite](https://github.com/aiidateam/aiida-core/pull/4719/checks?check_run_id=1858668311#step:9:558)  come from asyncio.

Here are the most prominent ones: 
### [FIXED] DeprecationWarning: Use "async with message.process()"
```
tests/test_dataclasses.py: 27 warnings
tests/calculations/test_transfer.py: 7 warnings
tests/cmdline/commands/test_data.py: 2 warnings
tests/cmdline/commands/test_process.py: 3 warnings
tests/cmdline/commands/test_run.py: 81 warnings
tests/engine/test_calc_job.py: 9 warnings
tests/engine/test_calcfunctions.py: 29 warnings
tests/engine/test_futures.py: 4 warnings
tests/engine/test_launch.py: 42 warnings
tests/engine/test_persistence.py: 2 warnings
tests/engine/test_process.py: 39 warnings
tests/engine/test_process_function.py: 90 warnings
tests/engine/test_rmq.py: 21 warnings
tests/engine/test_run.py: 4 warnings
tests/engine/test_runners.py: 2 warnings
tests/engine/test_work_chain.py: 187 warnings
tests/engine/test_workfunctions.py: 17 warnings
tests/orm/test_querybuilder.py: 4 warnings
tests/orm/data/test_data.py: 2 warnings
tests/parsers/test_parser.py: 8 warnings
tests/tools/importexport/test_prov_redesign.py: 2 warnings
tests/tools/importexport/orm/test_calculations.py: 2 warnings
tests/workflows/arithmetic/test_add_multiply.py: 8 warnings
  /opt/hostedtoolcache/Python/3.8.7/x64/lib/python3.8/site-packages/aio_pika/message.py:714: DeprecationWarning: Use "async with message.process()" instead
    warn('Use "async with message.process()" instead', DeprecationWarning)
```
This is emitted from aiopika [when entering a process context](https://github.com/mosquito/aio-pika/blob/5af608d05b05d65632652f0632d00af4729b265b/aio_pika/message.py#L713-L715)
Can @sphuber or @unkcpz suggest how/where this should be fixed?

### coroutine 'create_task.<locals>.run_task' was never awaited
```
tests/test_dataclasses.py: 1 warning
tests/test_nodes.py: 1 warning
[...]

PR #4742, opened 2021-02-10
🔧 MAINTAIN: Reduce test warnings

chrisjsewell:
This PR is focussed on reducing pytest warnings:

- Replace `collections` with `collections.abc`
- pytest-asyncio does not work with `unittest.TestCase` derived tests (https://github.com/pytest-dev/pytest-asyncio/issues/77).
- `ProcessFuture` already closed via polling should not set a result via a broadcast event.
- Upgrade kiwipy and plumpy to fix:
  - https://github.com/aiidateam/kiwipy/pull/98
  - https://github.com/aiidateam/plumpy/pull/204
  - https://github.com/aiidateam/plumpy/pull/206

partially addresses #4722


chrisjsewell:
719 warnings (https://github.com/aiidateam/aiida-core/runs/1918314856), down to 123 warnings (https://github.com/aiidateam/aiida-core/pull/4742/checks?check_run_id=1918695277)

There is still a lot of:

- ResourceWarning: unclosed event loop <_UnixSelectorEventLoop running=False closed=False debug=True>
- ResourceWarning: unclosed event loop <_UnixSelectorEventLoop running=False closed=False debug=False>
- RuntimeWarning: coroutine 'create_task.<locals>.run_task' was never awaited

but these are more difficult to debug so I will tackle these in another PR.

Also @ltalirz, to note with the new SSH tests, we are getting:

- ResourceWarning: unclosed <socket.socket fd=23, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=0, laddr=('::1', 55576, 0, 0)>

chrisjsewell:
I've released plumpy v0.18.5 https://anaconda.org/conda-forge/plumpy/files, so I don't know why its not finding it

ltalirz:
> Also @ltalirz, to note with the new SSH tests, we are getting:
> 
> * `ResourceWarning: unclosed <socket.socket fd=23, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=0, laddr=('::1', 55576, 0, 0)>`

Exactly, I already opened https://github.com/aiidateam/aiida-core/issues/4744 for this

> I've released plumpy v0.18.5 https://anaconda.org/conda-forge/plumpy/files, so I don't know why its not finding it

In my experience it can take a while.

chrisjsewell:
> In my experience it can take a while to "propagate" to every cache.

yep its installing now 👍

aiida/engine/processes/futures.py
```diff
@@ -59,7 +59,12 @@ def __init__(
 
             # Try setting up a filtered broadcast subscriber
             if self._communicator is not None:
-                broadcast_filter = kiwipy.BroadcastFilter(lambda *args, **kwargs: self.set_result(node), sender=pk)
+
```
ltalirz, line 62: Thanks - I was going to suggest to add a comment here as to why that is necessary but then I realized that this is in the `ProcessFuture` 

https://github.com/aiidateam/aiida-core/issues/4758

ltalirz (commented):
thanks @chrisjsewell , this is great!

I still spotted one overlooked collection import
```
tests/test_dataclasses.py::TestTrajectoryData::test_conversion_from_structurelist
  /home/runner/work/aiida-core/aiida-core/aiida/orm/nodes/data/array/trajectory.py:38: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    if not isinstance(symbols, collections.Iterable):
```

Happy to approve once the tests pass

aiida/engine/processes/futures.py
chrisjsewell: yeh I thought that might be the case

ltalirz (approved):
thanks @chrisjsewell !

Issue #4731, opened 2021-02-09
CI: Non-deterministic timeouts in `test-install` workflow


csadorf:
The `test-install` GitHub actions workflow is sometimes canceled (due to timeout).
The timeouts appear to occur randomly, however they are clustered around two steps:

<details><summary>Canceled on line 29 (<code>
verdi -p test_${AIIDA_TEST_BACKEND} run .ci/test_daemon.py
</code>)</summary>
<ul>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1775661886?check_suite_focus=true">run 1775661886</a></li>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1776053647?check_suite_focus=true">run 1776053647</a></li>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1787650572?check_suite_focus=true">run 1787650572</a></li>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1790065232?check_suite_focus=true">run 1790065232</a></li>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1808225938?check_suite_focus=true">run 1808225938</a></li>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1808280405?check_suite_focus=true">run 1808280405</a></li>
</ul>
</details>

<details>
<summary>Canceled at <code>tests/cmdline/commands/test_plugin.py ......                             [ 28%]</code></summary>
<ul>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1776556930?check_suite_focus=true">run 1776556930</a></li>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1776838336?check_suite_focus=true">run 1776838336</a></li>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1799687152?check_suite_focus=true">run 1799687152</a></li>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1803637702?check_suite_focus=true">run 1803637702</a></li>
<li><a href="https://github.com/aiidateam/aiida-core/runs/1856166142?check_suite_focus=true">run 1856166142</a></li>
</ul>
</details>

Facts:
- For cancellations on line 29, usually all tests fail regardless of tested combination (Python version, backend) except for run 1808225938 and 1808280405.
- For cancellations at `test_plugin.py` there appears to be no correlation with the tested combination, but it is completely random.

Next steps
[...]

PR #4746, opened 2021-02-12
CI: Bump reentry to v1.3.2

### LINKED ISSUE #4723: Remove deprecated import from 'collections' instead of 'collections.abc' module

csadorf:
Closes #4723


ramirezfranciscof:
Hey @csadorf , couple of questions:

(1) Why is this draft?

(2) Do you know/understand why the rmq test failed on push?? It did succeed on PR, so I'm guessing it might be ok to ignore the failed one?

csadorf:
> Hey @csadorf , couple of questions:
> 
> (1) Why is this draft?

I was going to mark it as ready once the tests pass. This is currently affected by #4731 .

> (2) Do you know/understand why the rmq test failed on push?? It did succeed on PR, so I'm guessing it might be ok to ignore the failed one?

No idea.

ltalirz (approved):
thanks.

not sure whether the order of the packages is supposed to be alphabetical (see py3.9) but that doesn't matter

csadorf:
> thanks.
> 
> not sure whether the order of the packages is supposed to be alphabetical (see py3.9) but that doesn't matter

They are in the order as `pip freeze` produces it: https://github.com/aiidateam/aiida-core/blob/e794287218a5d3b85a85ba36eed6d6928a849a26/.github/workflows/test-install.yml#L202

PR #4747, opened 2021-02-12
Skip test 'TestVerdiProcessDaemon::test_pause_play_kill'.

csadorf:
The test randomly fails to complete within a reasonable amount of time
leading to a significant disruption of our CI pipeline.

Investigated in issue #4731.


csadorf:
@sphuber Do you think you will have time to comment on this? I believe the test was originally created by you.

sphuber (approved):
This test is indeed tricky as it needs to mock a daemon process within the unit tests. This has shown problems before so fine to skip for now until it can be made more robust

PR #4750, opened 2021-02-15
Migrate attributes for old inline calculations

### LINKED ISSUE #4401: Missing migration for source_file attribute

ramirezfranciscof:
fixes #4401

Old `InlineCalculation` nodes were migrated to `CalcFunctions`, but their attributes remained the old ones. These migrations updates the attributes as follows to better comply with the current format:

* OLD `sealed` and `function_name` remain the same
* OLD `namespace` becomes `function_namespace`
* OLD `first_line_source_code` becomes `function_starting_line_number`
* OLD `source_file` is moved to repository and dropped from DB
* OLD `source_code` is used to calculate `function_number_of_lines` and dropped from DB
* NEW `exit_status` is initialized at 0
* NEW `process_state` is initialized at "finished"
* NEW `process_label` is set as `Legacy InlineCalculation`
* The attribute `version` is never set for these nodes

Note: `exit_status` and `process_state` are initialized to 0 and "finished" respectively because the
CalcFunctions affected are old InlineCalculation, which were only stored if they succeeded. You can check
the relevant code [here](https://github.com/aiidateam/aiida-core/blob/v0.12.5/aiida/orm/implementation/django/calculation/inline.py#L60) (line 60 has the execution call, it is stored after that).

### Outstanding issues

The initial PR contains the minimal necessary stuff for the migration, but there are a couple of issues that remain:

1. **Reverse Migrations:** Do I need to set these up? I see that some migrations don't have reverse ones, and even the ones that do don't have unit test. I think it would be "easy" to do these reverse migrations, but I don't know if it makes much sense if we are not testing these consistently (unless the idea would be to start doing that from now on).
2. **First SQLalchemy inefficiency [DONE]:** I'm currently setting all the new attributes in separate queries because I couldn't find an easy way to `jsonb_set` more than one attribute at once. This one shouldn't be too difficult (I think), I just need to search and test a bit more (I wanted to have a first version PRed to start looking at the other issues and the django code).
3. **Second SQLalchemy inefficiency [DONE]:** To count the number of lines the source code has, I need to query for the id and source codes first, then do the analysis in python, and finally do one query for each node (filtering for its id and then setting that node's number of lines). There seems to be some ways to execute the analysis functions in SQL (see for example "regexp_split_to_array" in [here](https://www.postgresql.org/docs/9.1/functions-string.html)), but I still can't put them together and it might take more effort than what we actually gain from it?

I'll set it as draft for now and I'll continue working with 2. which seems the most promising and certain, but in the meantime I would appreciate to get some feedback on 1. and 3. and maybe a general overlook at the migrations (@sphuber , @giovannipizzi , should we ask someone else?).

### FInal comment

I think the uuid is not correctly set when creating nodes in the SQLalchemy migration tests. For example, if I insert a `print(node.uuid)` at the end of `TestTrajectoryDataMigration`, between lines 1150 and 1151, and then `pytest -s tests/backends/aiida_sqlalchemy/test_migrations.py::TestTrajectoryDataMigration`, I get a `None` value. All the auxiliary verifications seem to not mind this and work anyways, but perhaps they are actually doing something different and we are not realizing? In any case it might be something we should look into and fix...should I open an issue for this?


ramirezfranciscof:
I found the way to do multiple json modifications at the same time (found two actually, in the migration I do a mix of both because I think it would be useful to keep a reference of how both work).

giovannipizzi:
I won't be able to review this shortly. Maybe @ltalirz could check it? Or @mbercx who's now an expert of migrations? :-) An eye from @sphuber (if he has time) would be also great as he's the most knowledgeable on these attributes

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,102 @@
...
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Migrate attributes from old InlineCalculation that have been transformed into CalcFunctions:
+
+* OLD `sealed` and `function_name` remain the same
```
sphuber, line 12: the `sealed` attribute is not really particular to legacy `InlineCalculation` nodes so not sure we should even mention it.

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,102 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Migrate attributes from old InlineCalculation that have been transformed into CalcFunctions:
```
sphuber, line 10:
```suggestion
"""Migrate attributes from legacy `InlineCalculation` nodes that were migrated to `CalcFunctionNode`s.

The node type migration was performed in `0020_provenance_redesign.py` but the attributes were forgotten.
```

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,102 @@
...
+* OLD `namespace` becomes `function_namespace`
+* OLD `first_line_source_code` becomes `function_starting_line_number`
+* OLD `source_file` is moved to repository and dropped from DB
+* OLD `source_code` is used to calculate `function_number_of_lines` and dropped from DB
+* NEW `exit_status` is initialized at 0
+* NEW `process_state` is initialized at "finished"
+* NEW `process_label` is set as `Legacy InlineCalculation`
+
+Note 1: old InlineCalculation can be identified by the old keywords, and after migration they will be
+identified by having `function_number_of_lines` but no `version`.
```
sphuber, line 22: This is not true I think. Tthere was a time when `CalcFunctionNodes` were created directly but without a `version` attribute.

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,102 @@
...
+#utils.del_object_from_repository(uuid, 'source_file')
+
+#def create_old_attributes(apps, _):
+#"""Re-creates the old attributes"""
+#content_start = modifier.get_value_for_node(pk, 'function_starting_line_number')
+#content_end = content_start + modifier.get_value_for_node(pk, 'function_number_of_lines')
+#content0_text = utils.get_object_from_repository(uuid, 'source_file')
+#content0_list = content0_text.splitlines()
+#content_list = content0_list[content_start:content_end]
+#content_text = "\n".join(content_list)
```
sphuber, line 89:
```suggestion
```

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,102 @@
...
+
+class Migration(migrations.Migration):
+    """Migrate attributes from old InlineCalculation that have been transformed into CalcFunctions."""
+    dependencies = [('db', '0045_dbgroup_extras')]
+    operations = [
+        # 1. We create the new attributes and copy the content
+        migrations.RunPython(create_new_attributes, reverse_code=migrations.RunPython.noop),
+        # 2. We delete the old attributes
+        migrations.RunPython(delete_old_attributes, reverse_code=migrations.RunPython.noop),
+        # 2. Finish migration
```
sphuber, line 100:
```suggestion
        # 3. Finish migration
```

aiida/backends/general/migrations/utils.py
```diff
@@ -50,6 +50,16 @@ def put_object_from_string(uuid, name, content):
         handle.write(content)
 
 
+def del_object_from_repository(uuid, name):
```
sphuber, line 53: don't think you are actually using this, right? So could be removed

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
```diff
@@ -0,0 +1,130 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Migrate attributes from old InlineCalculation that have been transformed into CalcFunctions:
```
sphuber, line 10: Apply same changes as those proposed in Django migration docstring

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
```diff
@@ -0,0 +1,130 @@
...
+
+* OLD `sealed` and `function_name` remain the same
```
sphuber, line 12: idem

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
```diff
@@ -0,0 +1,130 @@
...
+* OLD `namespace` becomes `function_namespace`
+* OLD `first_line_source_code` becomes `function_starting_line_number`
+* OLD `source_file` is moved to repository and dropped from DB
+* OLD `source_code` is used to calculate `function_number_of_lines` and dropped from DB
+* NEW `exit_status` is initialized at 0
+* NEW `process_state` is initialized at "finished"
+* NEW `process_label` is set as `Legacy InlineCalculation`
+
+Note 1: old InlineCalculation can be identified by the old keywords, and after migration they will be
+identified by having `function_number_of_lines` but no `version`.
```
sphuber, line 22: idem

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+        # yapf: enable
+
+        if full_file:
+            text_t = text_i + '\n' + text_t + '\n' + text_f
+
+        return text_t
+
+    def setUpBeforeMigration(self):
+        calcnode = self.DbNode(
```
sphuber, line 47: I would another `CalcFunctionNode` but one that does not corresponds to a legacy node that was migrated. This to test that these nodes are not affected. You could for example set its exit status to a non-zero value and check after migration that it still is non-zero. If the filtering in your migration were to be incorrect, this would have been set to zero.

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Pre existing and renamed attributes
+        self.assertEqual(calcnode.attributes['sealed'], True)
+        self.assertEqual(calcnode.attributes['function_name'], 'sum_inline')
+        self.assertEqual(calcnode.attributes['function_namespace'], '__main__')
+        self.assertEqual(calcnode.attributes['function_starting_line_number'], 4)
+
+        # Newly determined attributes
+        self.assertEqual(calcnode.attributes['exit_status'], 0)
+        self.assertEqual(calcnode.attributes['process_state'], 'finished')  #USE ENUM?
```
sphuber, line 78: If you do decide to use the enum, then you should probably use it also consistently in the migration itself. There you also just use the string. I think it is actually fine to use the string here.
```suggestion
        self.assertEqual(calcnode.attributes['process_state'], 'finished')
```

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,98 @@
...
+        self.assertEqual(calcnode.attributes['process_label'], 'Legacy InlineCalculation')
+        self.assertEqual(calcnode.attributes['function_number_of_lines'], 3)
+
+        # Old attributes that got renamed or removed
+        with self.assertRaises(KeyError):
+            _ = calcnode.attributes['namespace']
```
sphuber, line 84: Would be more natural and shorter to write this as 
```suggestion
        assert 'namespace' not in calcnode.attributes
```

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        UPDATE db_dbnode
+        SET attributes = attributes || '{
+            "exit_status": 0,
+            "process_state": "finished",
+            "process_label": "Legacy InlineCalculation"
+            }'::jsonb ||
+            jsonb_build_object(
+                'function_starting_line_number', attributes->'first_line_source_code',
+                'function_namespace', attributes->'namespace'
```
sphuber, line 65: Does this _need_ to be separate from the first dictionary? Couldn't this be merged? Or do you have to use `jsonb_build_object` here because you need to access existing `attributes`? If so, might be worth a comment

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
```diff
@@ -0,0 +1,130 @@
...
+            WHERE
+                attributes ? 'namespace' AND
+                attributes ? 'first_line_source_code' AND
+                node_type = 'process.calculation.calcfunction.CalcFunctionNode.' AND
+                (attributes -> 'source_code')::TEXT LIKE :symlike;
+            """), replace_dict)
+
+
+    # Saves the source file in the repository
+    sqlquery_results = conn.execute(text("SELECT uuid, attributes -> 'source_file' FROM db_dbnode;"))
```
sphuber, line 102: shouldn't we put another filter to make sure we only target legacy migrated calc nodes to be sure? Here you are fetching all nodes with the attribute `source_file`. There could be others

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
```diff
@@ -0,0 +1,130 @@
...
+                'function_namespace', attributes->'namespace'
+                )
+        WHERE
+            attributes ? 'namespace' AND
+            attributes ? 'first_line_source_code' AND
+            node_type = 'process.calculation.calcfunction.CalcFunctionNode.';
+        """))
+
+
+    # Sets new 'function_number_of_lines' with the number of lines in 'source_code'
```
sphuber, line 74: If I understand correctly, here you are determining the number of lines in the source code by counting the number of new line characters. And to count the new line characters directly in SQL, you are replacing them with empty strings and then compute the character length difference of the new string with the original, which should give the number of lines times the number of characters in the newline character.

It would be good to add this as an explicit comment. Also why you decided to do it this way and not in Python. I could see the point if this would be the only thing you would have to do in Python, but straight after this you are collecting all relevant nodes and writing the source file to the repo. Might as well use the simple Python logic (as in the Django migration) to compute the number of lines. Is a lot less complicated and error prone. Don't think the efficiency would suffer too much.

I am a bit surprised also by the fact that you have to count escaping backslashes in the count. The actual data stored in the database doesn't contain the escape characters does it? So why do you need to count them in the SQL function? Your approach also relies on the fact that you _first_ replace all DOS like endings, then Mac and then Unix. This would also merit a comment because it is not immediately obvious. This is another complexity that would speak in favor of the simply Python approach to be honest

sphuber (changes requested):
Thanks @ramirezfranciscof . I have left some suggestions and comments. Note that certain apply to two files for both backends but sometimes only explicitly mentioned it just in one. Please be sure to check whether it applies to the other as well. There are also quite a few dead code snippets from debugging, please make sure you remove those.

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
ramirezfranciscof: Mmm I'm not sure I understand why this is a factor. I feel that, in this case at least, there is more value in the completeness of being explicit about what I'm doing with all attributes than in saving two words in the docstring, no?

sphuber: Because to me it suggests that legacy `InlineCalculation` already used to have `sealed` nodes and it so happens that doesn't have to be changed. But this is not True, the `sealed` attribute has been added in some other migration.

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
ramirezfranciscof: Yeah, I meant the combination of both, since other `CalcFunctionNodes` don't have `function_number_of_lines`. They could if, when merging the PR that starts keeping track of this new attribute, a migration for older `CalcFunctionNodes` is included, but I think this is not certain since it would involve findig a way to parse the file to get the number of lines and so (I think we discussed this, no?)

Anyways, I added this as a way to identify them before deciding to set the new `process_label` to `Legacy InlineCalculation` instead of just copying the `function_name` there, so I think now that `process_label` should be enough of an identifier and I can just remove the note.

aiida/backends/general/migrations/utils.py
ramirezfranciscof: Ah, yes, I set it for the backwards migrations. If you confirm those are not needed then I will delete it.

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
ramirezfranciscof: Ah, yes, indeed. So this was a question: I was looking if the `process_state` was set as an enum but couldn't find / didn't know how to grep for it. Do you know where it is? I should probably use that, right? (as you say, both in test and in migration)

aiida/backends/general/migrations/utils.py
sphuber: no we don't support backwards migrations.

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
ramirezfranciscof: Mmm yes you are right, but my only question is that `assert 'namespace' not in calcnode.attributes` looks like pytest syntax, can I use that inside an `AiidaTestCase` subclass? (Or even if I could, should I be mixing the two syntaxes?)

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
ramirezfranciscof: > If I understand correctly, here you are determining the number of lines in the source code by counting the number of new line characters. And to count the new line characters directly in SQL, you are replacing them with empty strings and then compute the character length difference of the new string with the original, which should give the number of lines times the number of characters in the newline character.
> 
> It would be good to add this as an explicit comment.
> (...)

But I kind of did, no? See the [two lines below](https://github.com/aiidateam/aiida-core/pull/4750/files/40edeca2e06939f293b5cff2d383e06e6ad73fac#diff-77204f1bd3e424f62e7f81cefb12a090cedaa0188689052fe43957d016144885R75-R76). I mean maybe you mean you want me to explain a bit more?

> Also why you decided to do it this way and not in Python. I could see the point if this would be the only thing you would have to do in Python, but straight after this you are collecting all relevant nodes and writing the source file to the repo. Might as well use the simple Python logic (as in the Django migration) to compute the number of lines. Is a lot less complicated and error prone. Don't think the efficiency would suffer too much.

I had a [first version](https://github.com/aiidateam/aiida-core/pull/4750/commits/ad6efdb1e67a3ad9d9200337a806d05a9d8a4839#diff-77204f1bd3e424f62e7f81cefb12a090cedaa0188689052fe43957d016144885R106-R121) where I did it this way: first I queried for all the `source_code` entries, then I processed those in python and updated the database. The problem with this is that since each update has a specific and different number of lines, I need to do a separate update query for each one of them (thus, the single query to update turns into one query to get the data and n_nodes update queries). Maybe you had a better way in mind that I didn't think of?

Note that for the following section I am doing something similar, yes, but I get all the info for the first query and in the loop I'm just storing the file, there are not queries being made (since I don't need to modify anything that is specific of each node, but just remove the `source_file` attribute once the file is stored, I can just delete all together with the other old attributes in the last query).

I'm honestly not sure what is the performance impact of this: maybe the bottleneck is in the `put_object_from_string` method and then adding a query in each cycle of the same loop does not make a difference. I was under the impression that this is not the case and that each query can take a significant ammount of time, depending on the total size of the DB, but I honestly have made no tests for this. I could try some benchmarking, but if we already have an idea that a single query in a 10MM node DB is comparable to writing a 1k lines file, then I would prefer to accept that and roll with it :sweat_smile:.

> I am a bit surprised also by the fact that you have to count escaping backslashes in the count. The actual data stored in the database doesn't contain the escape characters does it? So why do you need to count them in the SQL function?

Yeah, I didn't understand the why of this either, but I did do some tests to see what was the text stored (because I was getting the wrong count) and indeed the double backlashes were there.

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
ramirezfranciscof: Good point. I modified the `WHERE` clause of all other queries to be even more specific and added it here as well:

```SQL
        SELECT uuid, attributes -> 'source_file'
        FROM db_dbnode
        WHERE
            attributes ? 'namespace' AND
            attributes ? 'source_code' AND
            attributes ? 'source_file' AND
            attributes ? 'first_line_source_code' AND
            node_type = 'process.calculation.calcfunction.CalcFunctionNode.';
```

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
sphuber: it comes from the `plumpy.ProcessState` enum, which you can import from `aiida.engine.processes.process` as it is aliased there

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
sphuber: there is nothing specifically pytest about using `assert`. The `assert` is a normal Python construct. `pytest` just catches the `AssertionError` that get's raised and deals with it. I think you should be able to use it just fine in unittest based test.

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
sphuber: > But I kind of did, no? See the two lines below. I mean maybe you mean you want me to explain a bit more?

Sure, this covers the specifics of the solution but it still leaves the reader to cobble together why all of this is needed in the first place. Simply stating something like "Here we need to compute the number of lines in the `source_code` attribute. To be able to do this directly in SQL, we replace all newline characters with empty strings and then compare the new string length with the original to determine the number of characters removed, which corresponds to the number of lines."

Also, I realize now, this solution does not account for a missing terminating newline character. This would give you a count that is one short. Is it guaranteed that the `source_code` always has a terminating newline?

> I had a first version where I did it this way: first I queried for all the source_code entries, then I processed those in python and updated the database. The problem with this is that since each update has a specific and different number of lines, I need to do a separate update query for each one of them 

You do the same in Django though don't you? You get all the nodes and then in a loop update each node, calling `.save()` on it. The same should be possible in SqlAlchemy. Probably it will update it in memory and then at the end perform single commit to the database.

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
ramirezfranciscof: Indeed, as you say, I need the `jsonb_build_object` since I am extracting the info from the attributes of the each entry. What I could do is use `jsonb_build_object` for all:

```SQL
        SET attributes = attributes || jsonb_build_object(
                'exit_status': 0,
                'process_state': 'finished',
                'process_label': 'Legacy InlineCalculation',
                'function_starting_line_number', attributes->'first_line_source_code',
                'function_namespace', attributes->'namespace'
                )
```

I did it this way because I thought it could be useful to have in the codebase as a reference the two ways of doing so (the one that is sufficient for fixed modifications, and the one you need if you are transforming data for each entry). What would you prefer, that I use the `jsonb_build_object` for all or that I leave both options and include a comment explaining?

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
ramirezfranciscof: Should I also then change all the `self.assertEqual(a,b)` for simpler `assert a == b`?

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
ramirezfranciscof: Ok, thanks! I've actually changed my mind...since the migrations apply to the database, and the database at that point had the `FINISHED` state be `"finished"`, I think the correct way to do this is to target that specific version (i.e. if we later change our `FINISHED` state to `"totally_finished"`, it may break this migration because at the DB where it is applying it the values stored are `"finished"` and not `"totally_finished"`).

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
ramirezfranciscof: > Sure, this covers the specifics of the solution but it still leaves the reader to cobble together why all of this is needed in the first place. Simply stating something like "Here we need to compute the number of lines in the `source_code` attribute. To be able to do this directly in SQL, we replace all newline characters with empty strings and then compare the new string length with the original to determine the number of characters removed, which corresponds to the number of lines."

Ok, gotcha, I will add something along those lines.

> Also, I realize now, this solution does not account for a missing terminating newline character. This would give you a count that is one short. Is it guaranteed that the `source_code` always has a terminating newline?

No, actually I'm assuming it does not have a terminating newline, that is why I always add 1 when storing it:

```SQL
            SET attributes = attributes || jsonb_build_object(
                'function_number_of_lines',
                ( CHAR_LENGTH((attributes -> 'source_code')::TEXT)
                - CHAR_LENGTH( REPLACE((attributes -> 'source_code')::TEXT, :symbol, '' )))
                / :count + 1 -- <<== SEE HERE
                )
```

If it does have a terminating newline then I assume they extra separation line was meant to be considered and add it as well. But I think this starts getting into a terrain that is hard to "homogeneize after the fact". If you really want it, I could add a prior update query that looks for all `source_code` that end in a newline and just remove that, and then apply this, but I don't think I would personally account for that...

> You do the same in Django though don't you? You get all the nodes and then in a loop update each node, calling `.save()` on it. The same should be possible in SqlAlchemy. Probably it will update it in memory and then at the end perform single commit to the database.

This is true, and as you say I did this *with the hope* that the Django ORM is intelligent and knows how to optimize this when building the query since I'm not explicitly saying how to do that. However, I have less confidence that the `conn.execute` method will do this when I'm giving specific SQL queries (I would actually find it a bit concerning if I was explicitly giving queries in SQL and it was deciding to modify them).

As a matter of fact I forgot to do it, but I was actually going to ask if it wouldn't be better to do this part in SQL directly also for django so that we are sure that there is no query bombardment for the ORM not optimizing this request.

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
sphuber: I think it is better to have it consistent and use the one method. I see what you were trying to accomplish but using both just makes the code more difficult to read

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
sphuber: fair enough, you could go for `self.assertFalse('namespace' in calcnode.attributes')` this way you still use the same syntax but it is just a single line per test. Bonus points if you put it in a simple loop for all keys that should be absent ;)

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
sphuber: I am surprised that there is not terminating newline. Maybe the old inline calc code performed a trim on the source code, trimming all leading an trailing whitespace. Would be good if you could check (I think you already looked up the old source to verify that it only got stored on success) and if so then you comment that why it is ok to assume. Then I think it would be fine to keep this SQL solution. I am not 100% sure myself how to do the ORM way through alembic. It is not actually built for data migrations, so maybe it isn't possible. For this item would still like @giovannipizzi to also have a look. Know he is busy, but I don't feel comfortable signing off on migrations that I am not fully sure of.

tests/backends/aiida_django/migrations/test_migrations_0046_inline_calc_attributes.py
ramirezfranciscof: Haha I was actually honestly asking to make the change like a "partial migration to pytest", was not trying to argue against it =P. But ok, I'll do the `self.assertFalse`.

> Bonus points if you put it in a simple loop for all keys that should be absent ;)

Yes brownie points gimme :grimacing:

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
ramirezfranciscof: > I am surprised that there is not terminating newline. 

Ok, I think I may have been unclear on my previous statement, let me try to clarify. My assumption is not on the existence on the ending newline, but on the consistency of how this used to be stored and how we want the soon to be counting to work. Let me explain:

I don't know how AiiDA stored the `source_code` in the attributes, but if we use to store the ending newline I assume we would also want that consider in the new count feature (since it is mostly used internally to return the same content but via a method that takes it from the file instead of keeping it in the DB).

I can try looking in the some of the old DB of the mat cloud for how this was stored, if they had the ending newline or not, and report back (althogh would that be enough to assume it was always the same?). However, independent of confirming its existence, it doesn't really tell me wether I should count or not: I don't know how @unkcpz or you were conceiving this count to work. Again, my assumption was that the new method would return the exact same thing that we use to store in the DB.

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
ramirezfranciscof: So, I already changed it to use only `jsonb_build_object`, but do notice the horrible way of having to provide the dict as a list of `key1, val1, key2, val2, ...` instead of just providing a dict...:anguished:

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
sphuber: > but if we use to store the ending newline I assume we would also want that consider in the new count feature (since it is mostly used internally to return the same content but via a method that takes it from the file instead of keeping it in the DB).

The point is though that in the new version we no longer store the source code lines of the function, but just the number of lines in the source code. Together with the starting line, returning the exact lines of the function is then well-defined. The only question you might have there is whether the string that is returned has the terminating newline or not, but that can be controlled on the fly.

Here, the question I am asking, is whether the calculation you are making to determine the number of lines in the functions source code based on the string in the database is correct. You can assume that it was at least always created in the same way and so if the current method works, it would always work, but that is clearly an assumption and can actually fail. If originally the following was stored:
```
source_code = 'def some_function(arg):\n    return\n
```
but then that was changed to trim the string before storing it such that 
```
source_code = 'def some_function(arg):\n    return
```
was stored instead, your code would fail for certain nodes. For the latter it would correctly return 2 lines, but for the former it would return 3. Now luckily, in this case, the new code would return all the source code _plus an additional line_ from the source file, which is likely to be an empty line, so it won't be a problem. But still, this is the problem I was describing and why I was arguing to instead do this in Python where accounting for both cases is trivial.

chrisjsewell:
@ramirezfranciscof gentle nudge to get a status update 😉

ramirezfranciscof:
> @ramirezfranciscof gentle nudge to get a status update 😉

Oh, sorry, I missed this. I needed to get a DB that had the old types so to check exactly how these are stored, I think I have now found one so I was going to do some tests now and hopefully push the last changes later in the afternoon.

ramirezfranciscof:
I still think that the definition of "number of lines of the function" does not automatically imply if the newline at the end should be counted or not; and now that I checked the [other PR](https://github.com/aiidateam/aiida-core/pull/4554) it is also not trivial if it should include the decorator or not. From how it is being used in [these lines](https://github.com/aiidateam/aiida-core/pull/4554/files#diff-5f56a4ce66bcb5ec7c0aaf89fc7ab18f2d094e0b0f83c42cf40efa5d39c02dd5R140-R145), will assume that we don't want to count the decorator, and from my previous discussions with @sphuber I will assume that the count also include the final newline. So in other words, the following text has 2 lines:

```python
# content of example source_code
(L1) @optional_inline
(L2) def example_sum(num1, num2):
(L3)     return num1 + num2
(L4)
```

I checked in the [2D database](https://archive.materialscloud.org/record/2020.158) and all `source_code` attributes seem to include the decorator and a final newline (some start with indentation and others don't):

```python
In [41]: example.get_attribute('source_code')
Out[41]: '    @optional_inline\n    def build_pw_input_parameters_inline(parameters,pw_parameters):\n        """\n        Build the parameters of a pw calculation.\n(...)            except KeyError:\n                pass\n        \n        return {\'output_pw_parameters\': ParameterData(dict=pw_params_dict)}\n'
```

To consider both cases (ending newline and no ending newline) I have now duplicated the `replace_dicts` so the first three passes will filter any node whose `source_code` specifically finishes in newline (and thus do `count = newlines - 1`) and then search for the rest who don't (and thus `count = newlines`, because the decorator doesn't count).

ramirezfranciscof:
@sphuber I corrected the counting issue, do you want to give it a last pass or we are just waiting for @giovannipizzi ? (in which case I might need to ping him elsewhere)

ramirezfranciscof:
Just for reference, I tried migrating the [2D database](https://archive.materialscloud.org/record/2020.158) and got the following times, which I guess are reasonable for a "small DB" (the sql command `SELECT pg_size_pretty( pg_database_size('aiidadb') );` gives around 300mb).

```bash
# django
real    0m16.078s
user    0m5.428s
sys     0m1.096s

# sqla
real    0m24.617s
user    0m1.960s
sys     0m0.881s
```

sphuber:
I looked at the current code and it now marks the line of the first (it can have multiple of course) decorator as the `function_starting_line_number`. This directly comes from the `inspect` call:
```
_, starting_line_number = inspect.getsourcelines(func)
self._set_function_starting_line_number(starting_line_number)
```
Anyway I agree with your analysis.

> I still think that the definition of "number of lines of the function" does not automatically imply if the newline at the end should be counted or not;

I fully agree. The only thing that is important in this migration is that all lines of the function are included and nothing more. Whether this includes the decorator is not really important. What _is_ important is that we don't accidentally chop off the last non-empty line _or_ adds the next line of the source file that actually is not part of the function. Normally this would be an empty line, so it is not that big of a deal, however, it can simply be the next line of the source file, which we don't really want in there, even though it will be indented differently and could be determined to not actually be part of the file. In short, given:
```python
1. @calcfunction
2. def some_function():
3.     a = 1
4.     return a
5. some_function()
```
We want lines 2, 3 and 4 to be stored (1 is fine but not necessary). If we miss line 4, or we include 5, I would say that is incorrect.

This is why I brought up the question on your method of extracting the number of lines of the function from the `source_code` attribute. You essentially counted the number of newlines and added one. However, if `source_code` contains the following
```python
1. def some_function():
2.     a = 1
3.     return a
```
_with_ an explicit newline character at the end, you would count 3 + 1 = 4 lines. The future method `get_source_code` that uses this number to extract the lines from the source file will then return:
```python
def some_function():
   a = 1
   return a
some_function()
```
which is clearly not what we want.

sphuber:
That all being said, I think that with the current state of the code that sets `function_starting_line_number` as the line of the decorator, we _should_ actually include the decorator in the line count. Otherwise, the future implementation of `get_source_code` that will get the source lines based on starting line number and the number of lines will give inconsistent results.

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,91 @@
...
+
+REVISION = '1.0.46'
+DOWN_REVISION = '1.0.45'
+
+
+def create_new_attributes(apps, _):
+    """Creates the new attributes"""
+    DbNode = apps.get_model('db', 'DbNode')
+
+    with transaction.atomic():
```
giovannipizzi, line 42: Do we really need to wrap this into a transaction? Isn't this already done automatically but the migration system (at least for Django?).

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,91 @@
...
+            # both these string will return a list of 3 elements (but should be counted as 2 lines of code):
+            #   `  @decorator\n  def funct()\n    return None\n`
+            #   `  @decorator\n  def funct()\n    return None`
+            numlines = len(inline_node.attributes['source_code'].splitlines()) - 1
+            inline_node.attributes['function_number_of_lines'] = numlines
+            inline_node.attributes['process_label'] = 'Legacy InlineCalculation'
+            inline_node.attributes['process_state'] = 'finished'
+            inline_node.attributes['exit_status'] = 0
+
+            utils.put_object_from_string(inline_node.uuid, 'source_file', inline_node.attributes['source_file'])
```
giovannipizzi, line 59: Did you check (and maybe with a test) what happens if the object 'source_file' is already there? This can happen if a previous run of the migration partially failed and was rolled back. Ideally, this shouldn't fail in that case.

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,91 @@
...
+
+            inline_node.save()
+
+
+def delete_old_attributes(apps, _):
```
giovannipizzi, line 64: I'm not sure you need to really split this in two, as long as you don't implement reverse migrations - you're not deleting stuff from files, so you could do all in the same loop, and in the worst case the transaction will be rolled back and the old data appear again. This will also have the benefit you loop and save only once, halving the time of the migration. Or am I missing something?

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,91 @@
...
+
+def create_new_attributes(apps, _):
+    """Creates the new attributes"""
+    DbNode = apps.get_model('db', 'DbNode')
+
+    with transaction.atomic():
+        inline_type = 'process.calculation.calcfunction.CalcFunctionNode.'
+        inline_nodes = DbNode.objects.filter(node_type=inline_type, attributes__has_key='namespace')
+
+        for inline_node in inline_nodes:
```
giovannipizzi, line 46: Looping and saving one by one will be slow. Probably, however, since this is only for legacy calculations, we can accept this, hoping that most DBs will have few of these entries... I don't know if the time to speed this up is worth the improvement, in this specific case.

aiida/backends/djsite/db/migrations/0046_inline_calc_attributes.py
```diff
@@ -0,0 +1,91 @@
...
+            #   `  @decorator\n  def funct()\n    return None`
+            numlines = len(inline_node.attributes['source_code'].splitlines()) - 1
+            inline_node.attributes['function_number_of_lines'] = numlines
+            inline_node.attributes['process_label'] = 'Legacy InlineCalculation'
+            inline_node.attributes['process_state'] = 'finished'
+            inline_node.attributes['exit_status'] = 0
+
+            utils.put_object_from_string(inline_node.uuid, 'source_file', inline_node.attributes['source_file'])
+
+            inline_node.save()
```
giovannipizzi, line 61: As a general comment (not only here where we are adding files, but in general for these data migrations, where the node content (DB and/or repo) is changed: should we invalidate and/or regenerate the hash? Or we just don't care? Pinging @sphuber and @greschd

aiida/backends/sqlalchemy/migrations/versions/d102e96ea943_inline_calc_attributes.py
giovannipizzi: I couldn't follow the whole discussion on how to exactly count the number of lines, I have to admit. I trust you did the right counting :-D

I however think that, indeed, it's better to try to keep the same SQL code also in Django, maybe, so we are sure we're doing the very same thing in both backends? Especially in this case, where the logic is *so* different. 
Also in the future, if we have similar issues, we'll need to do it anyway also for SQLA, so I don't think it's a big advantage to keep the Django syntax in the codebase.
On the efficiency, as I mention earlier: every save will trigger a commit to the DB (still within the transaction, if you are in one). So many commits are slower. I cannot say if the time for a commit and the time to save a file are comparable or one is much larger than the other, so I cannot say if moving all in a loop helps or not.
However, the weird substitution is quite complex to understand I have to admit... I cannot say by looking at it if it's correct - I trust that we have enough tests that verify it...

giovannipizzi (commented):
I did my pass (sorry if it took so long).
Most of the comments are non-critical.
I still think using the same logic in both backends is better (i.e. using SQL also for Django).

1. I didn't check in detail the discussion on the number of lines - if there are still doubts, I suggest to add a few more tests on the edge cases. Otherwise, good to go?
2. I didn't check in detail the complex counting of newlines in SQL - again, I hope the tests cover the important use cases...

As a final comment - I guess this is one of those cases where maybe the amount of effort to write a proper migration is larger than the actual gain, considering we're looking at legacy stuff :D At least, it was good for @ramirezfranciscof to learn :-)

sphuber:
As discussed before, we are leaving this for later if necessary. I am closing this for now, since it would probably only require a migration for SqlAlchemy at that point (as Django will have been dropped for v2.0) and it still needs a migration for export archives.

Issue #4100, opened 2020-05-21
Improve parsing of SSH config file for `verdi computer configure ssh`


ConradJohnston:
Related in spirit to issue #3311, and to the "UI Improvements" project.

Currently, when using `verdi computer configure ssh` to configure an SSH connection, the SSH config file, if present, is parsed by Paramiko and `verdi` offers some suggestions for populating the configuration options.

A typical stanza as currently supported might look like this:
```
Host best-hpc.phy.uni.ch
    User Conrad
    IdentityFile ~/.ssh/best-hpc.key
```
One limitation of the current implementation is that one must specify the fully qualified hostname as the `Host` and anything specified using the `Hostname` keyword is only used to substitute `%h` in SSH proxy commands, as I understand it. This is frustrating, as one typically uses the SSH  Config to do something like:
```
Host BestHPC
    Hostname best-hpc.phy.uni.ch
    User Conrad
    IdentityFile ~/.ssh/best-hpc.key
```
 where `Host` is used to define convenience aliases. 

If this more canonical approach was supported it would reduce frustration, but also allow semi-convenient work arounds for locked down supercomputers (see #3929) like:
```
Host BestHPCwith2FA
    Hostname localhost
    User Conrad
    Port 22000
```
where one opens an SSH connection separately, supplies some 2FA, and redirects this connection to some local port, say 22000, and then points AiiDA to this localhost port, using a nice alias.

chrisjsewell:
Sounds reasonable I think. On a semi-related note, I was recently messing around with a bit recently with automating this configuration when building VMs in https://github.com/chrisjsewell/ansible-role-aiida-computer

giovannipizzi:
Indeed, I also had the same experience as you had @ConradJohnston 
Any suggestion on how to allow the parser to do the right job? 
Should we check the hostname? (Doesn't seem enough in the last example) - and if we do it, what should we do if there are multiple matches?

Or should we ask the user to pick from one of the "valid" entries, or maybe just ask to provide a new name, then print out the raw content of the .ssh/config file, and if they are happy, continue with those defaults? (Note that the Host could even just be `Best*` so we should also deal with these wildcard cases)

Or any other suggestions?

ltalirz:
[...]

Issue #3311, opened 2019-09-11
Improve the parsing of the SSH `proxy_command`


giovannipizzi:
When reading the `proxy_command` from YAML file in `verdi computer configure ssh`, and I think also when reading it from command line parameters, the string is not cleaned up from `%h`, `%p` and/or from `>/dev/null` bits.

This logic is present only in the function `_get_proxy_command_suggestion_string(cls, computer) `, that is used to provide the default value by reading from the `~/.ssh/config` file; it relies on paramiko itself changing `%h` and `%p` when reading the config file, and has some logic for removing the redirections.

I think the logic should be extracted to a function, possibly we need to implement the logic for %h and %p if needed, and then call this logic not only to provide a default, but also to validate and change the input of the user.

Otherwise, a second option: we don't change what the user passes (might be a better idea), but we need to document what we expect as a value (i.e. replace `%h` with hostname, don't put redirection, ...) and validate the user input, complaining with an informative message if `%h` or `%p` are found, or a redirection character `>` is found.

PR #4761, opened 2021-02-17
`verdi calcjob gotocomputer`: Add proxy command

mbercx:
Currently the `verdi calcjob gocomputer` command does not work in case
the connection to the remote computer is via a proxy. Here we add the
`ProxyCommand` option to the parameters for the `ssh` command.

### DIFF
--- a/aiida/transports/plugins/ssh.py
+++ b/aiida/transports/plugins/ssh.py
@@ -1340,6 +1340,9 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         if 'key_filename' in self._connect_args and self._connect_args['key_filename']:
             further_params.append(f"-i {escape_for_bash(self._connect_args['key_filename'])}")
 
+        if 'proxy_command' in self._connect_args and self._connect_args['proxy_command']:
+            further_params.append(f"-o ProxyCommand={escape_for_bash(self._connect_args['proxy_command'])}")
+
         further_params_str = ' '.join(further_params)
 
         connect_string = self._gotocomputer_string(remotedir)
--- a/aiida/transports/plugins/ssh.py
+++ b/aiida/transports/plugins/ssh.py
@@ -426,7 +426,7 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
         if self._is_open:
             raise InvalidOperation('Cannot open the transport twice')
         # Open a SSHClient
-        connection_arguments = self._connect_args
+        connection_arguments = self._connect_args.copy()
         if 'key_filename' in connection_arguments and not connection_arguments['key_filename']:
             connection_arguments.pop('key_filename')
         proxystring = connection_arguments.pop('proxy_command', None)
--- a/tests/transports/test_ssh.py
+++ b/tests/transports/test_ssh.py
@@ -59,11 +59,17 @@ class TestBasicConnection(unittest.TestCase):
 
 def test_gotocomputer():
     """Test gotocomputer"""
-    with SshTransport(machine='localhost', timeout=30, use_login_shell=False, key_policy='AutoAddPolicy') as transport:
+    with SshTransport(
+        machine='localhost',
+        timeout=30,
+        use_login_shell=False,
+        key_policy='AutoAddPolicy',
+        proxy_command='ssh -W localhost:22 localhost'
+    ) as transport:
         cmd_str = transport.gotocomputer_command('/remote_dir/')
 
         expected_str = (
-            """ssh -t localhost   "if [ -d '/remote_dir/' ] ;"""
+            """ssh -t localhost -o ProxyCommand="ssh -W localhost:22 localhost" "if [ -d '/remote_dir/' ] ;"""
             """ then cd '/remote_dir/' ; bash  ; else echo '  ** The directory' ; """
             """echo '  ** /remote_dir/' ; echo '  ** seems to have been deleted, I logout...' ; fi" """
         )
--- a/tests/transports/test_ssh.py
+++ b/tests/transports/test_ssh.py
@@ -64,12 +64,12 @@ def test_gotocomputer():
         timeout=30,
         use_login_shell=False,
         key_policy='AutoAddPolicy',
-        proxy_command='ssh -W localhost:22 localhost'
+        proxy_command='ssh -W localhost:22 localhost',
     ) as transport:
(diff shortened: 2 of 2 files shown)


mbercx:
I basically got sick enough of this not working on Piz Daint that I finally got around to fixing it. 😅

ltalirz (commented):
thanks @mbercx , I ran into this in https://github.com/aiidateam/aiida-core/issues/4100#issuecomment-762325711 as well and it's good to fix.

Just to be sure - this change will make sure to read the proxycommand from the profile as well, is that correct?

Happy to do a proper review once the tests pass

ltalirz:
just to mention a related issue that might make sense to fix in one go as your touching this bit of the code https://github.com/aiidateam/aiida-core/issues/3311

(optional, not required of course)

mbercx:
@ltalirz tests are still failing because of #4860, but I think I found the issue that was causing them to fail previously. It was a little subtle. In the `open` method of the `SshTransport`, the `'key_filename'` and `'proxy_command'` keys are `pop`-ed from the `connection_arguments`:

https://github.com/aiidateam/aiida-core/blob/5ab86cca5fc2070937cfa90b1a479d06878e68dd/aiida/transports/plugins/ssh.py#L430-L432

This makes sense, since this variable is later passed to the `connect` command of the `SshClient`, which doesn't accept e.g. `proxy_command` as a keyword argument. However, this also mutated the `SshTransport._connect_args` dictionary, removing the `proxy_command`. So, I decided to `copy` this dictionary instead.

Re #3311: Now that I've had to explore the transport code a bit, I'd be happy to work on this as well, but would fix this in a separate PR since it's not yet clear what the best solution is and although it is _related_ to the transport code, the required changes are still quite different from those made in this PR. :)

mbercx:
> Just to be sure - this change will make sure to read the proxycommand from the profile as well, is that correct?

I'm not sure, actually. You mean the default set in the `.ssh/config` file? I'm still trying to figure out how and where exactly this is read. Frankly, the transport code is a little convoluted. 😅 

The `gotocomputer_command` reads all the connection arguments from the `_connect_args` dictionary, which is populated based on the `kwargs` in the constructor of the `SshTransport` class. However, after doing a bit of digging, it seems that the defaults, parsed from the `.ssh/config` file in the `parse_sshconfig` method, are only obtained when the `transport_option_default` in the `transports/cli.py` module is called. Which I think is done during the creation of the options of the `configure` commands. So I'm assuming that when the computer is first configured, AiiDA does read these config settings from the `.ssh/config` file and add them to the `authinfo` of the computer?

ltalirz (approved):
thanks @mbercx , this looks good to me

ltalirz:
Hi @mbercx - yes, when setting up a computer, AiiDA can parse the `~.ssh/config` file and store this information in the database.

> Just to be sure - this change will make sure to read the proxycommand from the profile as well, is that correct?

What I meant here is:
Different from the regular ssh transport (which uses `paramiko`), the `gotocomputer` command used to directly run `ssh` through bash, which results in it using the `~

> > Just to be sure - this change will make sure to read the proxycommand from the profile as well, is that correct?
> 
> I'm not sure, actually. You mean the default set in the `.ssh/config` file? I'm still trying to figure out how and where exactly this is read. Frankly, the transport code is a little convoluted. 😅

No, I meant that the `gotocomputer` command would now take the configuration from the database (which is indeed what you added).
the reason it *also* reads from the `~.ssh/config` file is that the `gotocomputer` command directly executes `ssh` on the command line (it has to), which then parses the `~/.ssh/config` file. This does not happen when going through `paramiko`.


> The `gotocomputer_command` reads all the connection arguments from the `_connect_args` dictionary, which is populated based on the `kwargs` in the constructor of the `SshTransport` class. However, after doing a bit of digging, it seems that the defaults, parsed from the `.ssh/config` file in the `parse_sshconfig` method, are only obtained when the `transport_option_default` in the `transports/cli.py` module is called. Which I think is done during the creation of the options of the `configure` commands. So I'm assuming that when the computer is first configured, AiiDA does read these config settings from the `.ssh/config` file and add them to the `authinfo` of the computer?

When you set up the computer, you can select whether AiiDA should parse the ssh/config or not.

ltalirz:
I leave the merge to you.
The commit message should mention that this is a slight change in behavior: Instead of the proxycommand stored in `~.ssh/config`,  the `gotocomputer` command will now use the proxycommand stored in the database (which should anyhow work as well since it is used by AiiDA).

mbercx:
> The reason it _also_ reads from the `~.ssh/config` file is that the `gotocomputer` command directly executes `ssh` on the command line (it has to), which then parses the `~/.ssh/config` file. This does not happen in other contexts when going through `paramiko`.

But, this would only work in case the user has set the `Host` as the `HostName`, right? I usually use shorthands for the `host`, e.g.:

```console
Host ela
  HostName ela.cscs.ch
  User mbercx
  IdentityFile ~/.ssh/id_rsa-daint
Host daint
  HostName daint.cscs.ch
  User mbercx
  IdentityFile ~/.ssh/id_rsa-daint
  ProxyJump ela
```

(Which is [in line with our documentation](https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/ssh.html#ssh-configuration)). But AiiDA requests the "fully qualified" host name:

```console
  -H, --hostname HOSTNAME         The fully qualified hostname of the computer
                                  (e.g. daint.cscs.ch). Use "localhost" when
                                  setting up the computer that AiiDA is
                                  running on.  [required]
```

So this is of course why `verdi gotocomputer` was failing in my case, because when you execute `ssh -t daint.cscs.ch ...`, my `.ssh/config` is not being parsed.

> When you set up the computer, you can select whether AiiDA should parse the ssh/config or not.

This faces the same problem. And of course `ProxyJump` is ignored, since paramiko doesn't support it.

PS: Actually, after doing a bit of digging while writing this comment, my whole explanation is raised in [this issue](https://github.com/aiidateam/aiida-core/issues/4100). 😅  I'll continue the discussion there!

mbercx:
> I leave the merge to you.

Thanks for the review, Leo! Will merge once #4860 is merged.

ltalirz:
thanks, will unsubscribe from this pr
feel free to merge without further review (e.g. merging in the postgres updates)

ramirezfranciscof (approved):
Already approved by @ltalirz , just leaving the formal thumbs up.

PR #4762, opened 2021-02-18
👌 IMPROVE: `verdi config list/show`

chrisjsewell:
Ensure they still work before a profile has been configured


chrisjsewell:
@ltalirz just a quick improvement I realised in hindsight

aiida/cmdline/commands/cmd_config.py
```diff
@@ -106,10 +109,14 @@ def verdi_config_show(ctx, option):
         'values': {
             'default': '<NOTSET>' if option.default is NO_DEFAULT else option.default,
             'global': config.options.get(option.name, '<NOTSET>'),
-            'profile': profile.options.get(option.name, '<NOTSET>'),
         }
     }
 
+    if not profile:
+        echo.echo_warning('no profiles configured: run `verdi setup` to create one')
```
ltalirz, line 116: same here

aiida/cmdline/commands/cmd_config.py
```diff
@@ -66,7 +66,10 @@ def verdi_config_list(ctx, prefix, description: bool):
     config: Config = ctx.obj.config
     profile: Profile = ctx.obj.profile
 
-    option_values = config.get_options(profile.name)
+    if not profile:
+        echo.echo_warning('no profiles configured: run `verdi setup` to create one')
```
ltalirz, line 70: is that so?
doesn't this mean that there is no default profile configured?

aiida/manage/configuration/config.py
```diff
@@ -434,7 +434,7 @@ def get_options(self, scope=None) -> Dict[str, Tuple[Option, str, Any]]:
         output = {}
```
ltalirz, line 434: an explanation of what the `scope` is would be useful (optional)

ltalirz (commented):
thanks @chrisjsewell 

two minor questions

aiida/cmdline/commands/cmd_config.py
chrisjsewell: can you have profiles configured without a default profile set?

aiida/manage/configuration/config.py
chrisjsewell: done

aiida/cmdline/commands/cmd_config.py
chrisjsewell: `verdi setup` automatically adds the created profile as the default and there is at least no "user based" method to unset the default (you can just set a different profile) so I think this is accurate. Happy for you to suggest an alternative message though?

ltalirz: hm... I vaguely remember having this situation some time ago but you're right- in the end the vast majority of the cases will be because there are no profiles at all. Let's keep it.

ltalirz (approved):
thanks @chrisjsewell !

aiida/cmdline/commands/cmd_config.py
ltalirz: @chrisjsewell What happens if you delete the default profile?

chrisjsewell: yeh I guess thats an edge case. Lokking at `Config.remove_profile`, it doesn't appear to reset `default_profile`, if the profile being removed is the default.
Perhaps it should do that?
abd perhaps set another profile as the default if there is still existing profiles

ltalirz: Unless someone wants to enforce an explicit specification of the profile, e.g. to avoid accidentally modifying one of the profiles.
In the end it may still the warning message here that is not quite accurate

chrisjsewell: I'll leave it up to you if you want to open a PR to modify it 😉

Issue #4249, opened 2020-07-10
Replace `tests.utils.configuration` with the corresponding pytest fixtures


sphuber:
This will require the tests that use these old utils to first be converted to normal tests, but that is anyway the plan

PR #4764, opened 2021-02-18
🧪 TESTS: add  `config_with_profile` fixture

chrisjsewell:
Using this fixture allows for the removal of `temporary_config_instance` and `with_temporary_config_instance` from `tests/utils/configuration.py`.
Partially addresses #4249


chrisjsewell:
another quick one for you @ltalirz 😬, off the back of #4712

ltalirz (approved):
great, thanks!

ltalirz:
Can `create_mock_profile` also be replaced or are there somehow difficulties?

chrisjsewell:
> Can create_mock_profile also be replaced or are there somehow difficulties?

these require the fixture to be directly passes to the functions (not just using the `usefixture` decorator), which in turn requires all the tests to be rewritten to not subclass `AiidaTestCase` and use regular assets. So just more work that I can be bothered with right now lol.

PR #4766, opened 2021-02-19
📚 DOCS: Add process submit diagram

chrisjsewell:
I've found it helpful:

image


chrisjsewell:
https://2946-77234579-gh.circle-artifacts.com/0/html/topics/processes/concepts.html#process-tasks

ltalirz (commented):
thanks @chrisjsewell - this looks very helpful indeed!

did you draw it yourself?
could you include the source of the figure as well?

I would have a couple of minor suggestions

chrisjsewell:
> did you draw it yourself?
> could you include the source of the figure as well?

yes I did it in powerpoint, so I wasn't sure whether I should include it in the repo?

ltalirz:
Yes, better to include it - I'm sure this figure will need to be updated over time.

chrisjsewell:
ok added

ltalirz (approved):
thanks @chrisjsewell ; 

good to merge for me, we can always iterate on it

Issue #4667, opened 2021-01-21
Tests should check that daemon can be restarted without errors in the log


ramirezfranciscof:
@mbercx and I seem to be having `TIMEOUTS` when restarting or stopping the daemon. It does stop in both cases, and it succeeds when manually `verdi daemon start`ing it:

```
$ verdi daemon stop 
Profile: charge_density
Waiting for the daemon to shut down... TIMEOUT
$ verdi daemon start
Starting the daemon... RUNNING
$ verdi daemon restart 
Restarting the daemon... TIMEOUT
$ verdi daemon status
Profile: charge_density
The daemon is not running
$ verdi daemon start
Starting the daemon... RUNNING
```

We checked our logs and found similar error reports:

```
Exception in callback start_daemon.<locals>.<lambda>() at /home/framirez/Workgit/aiida-core/aiida/engine/daemon/runner.py:59
handle: <Handle start_daemon.<locals>.<lambda>() at /home/framirez/Workgit/aiida-core/aiida/engine/daemon/runner.py:59>
Traceback (most recent call last):
  File "/usr/lib/python3.6/asyncio/events.py", line 145, in _run
    self._callback(*self._args)
  File "/home/framirez/Workgit/aiida-core/aiida/engine/daemon/runner.py", line 59, in <lambda>
    runner.loop.add_signal_handler(s, lambda s=s: asyncio.create_task(shutdown_runner(runner)))
AttributeError: module 'asyncio' has no attribute 'create_task'
```

@mbercx initially reported this behaviour in issue #4648 because it seemed related (see comment [here](https://github.com/aiidateam/aiida-core/issues/4648#issuecomment-762390990)) but apparently the error in the logs is different.

In principle this doesn't seem to come out in a clean environment. We can keep trying to see if we manage to reproduce it there, but hopefully this log gives an idea of what may be happening @sphuber @unkcpz?

ltalirz:
@ramirezfranciscof Thanks for reporting this together with the error log. 

Since the top-level `create_task` function [was introduced only on python 3.7](https://stackoverflow.com/a/53247626/1069467),  this indicates to me you are running python 3.6 - can you confirm?

January 13th 2021 [marks the end of python 3.6 support in AiiDA](https://www.aiida.net/download/); @csadorf is already working on officially dropping  3.6](https://github.com/aiidateam/aiida-core/issues/4691). I believe this will resolve the original reported issue.

[...]

PR #4769, opened 2021-02-22
👌 IMPROVE: verdi daemon status

chrisjsewell:
This is a proof of concept for closing #4667, and feedback would be appreciated.

Currently, when `verdi daemon start x` is run, this starts the circus daemon with settings to start *x* workers via the command `verdi devel run_daemon`. With the default `respawn=True` set, circus will maintain *x* workers running by simply spawning new workers if any die (see https://github.com/circus-tent/circus/blob/4b08197dedbe6416248d23f9983bd42cd97c96ab/circus/watcher.py#L554-L556).

This means that if there are any exceptions raised when starting the daemon it will simply die and be respawned by circus in an endless loop.
`verdi daemon status` does not currently "acknowledge" this issue; you will just see a new PID every time you run it.

Ideally then we want to "ping" the `Runner` inside a daemon worker, to check it is healthy (and return a non-zero exit code if not).
One way to do this is via RabbitMQ, by setting up an additional RPC subscriber, as I have added here.
If I then e,g, add a `raise Exception` to `aiida-core/aiida/cmdline/commands/cmd_devel.py::devel_run_daemon` we now get:

```console
root@ae5122fc408b:~# verdi daemon status
Profile: default
Daemon is running as PID 668593 since 2021-02-22 07:13:26
Active workers [1]:
   PID    MEM %    CPU %  started              status
------  -------  -------  -------------------  --------------
668647        0        0  2021-02-22 07:13:31  Not responding
Use verdi daemon [incr | decr] [num] to increase / decrease the amount of workers
```

This "ping" could also send back additional data and so while I was at it I have added code to record the processes that are currently running on that daemon (started via the `ProcessLauncher`):

```console
root@ae5122fc408b:~# verdi daemon status
Profile: default
Daemon is running as PID 653054 since 2021-02-22 06:45:57
Active workers [4]:
   PID    MEM %    CPU %  started              status
------  -------  -------  -------------------  ----------------
653059   15.893     65.3  2021-02-22 06:45:57  50/200 processes
653060   15.416     64.7  2021-02-22 06:45:57  50/200 processes
653061   14.343     65    2021-02-22 06:45:57  49/200 processes
654850   10.186     55.3  2021-02-22 06:48:34  34/200 processes
Use verdi daemon [incr | decr] [num] to increase / decrease the amount of workers
```

One issue with this is that, particularly when lots of new processes have just been submitted, it may take a while for the worker to reply to the RPC, leading to a laggy `verdi daemon status` and also time outs. So we may not want to ping the workers by default.

```
root@ae5122fc408b:~# verdi daemon status
Profile: default
Daemon is running as PID 653054 since 2021-02-22 06:45:57
Active workers [4]:
   PID    MEM %    CPU %  started              status
------  -------  -------  -------------------  ----------------
653059   13.082      6.7  2021-02-22 06:45:57  Response timeout
653060   12.427      5.2  2021-02-22 06:45:57  Response timeout
653061   11.798      4.6  2021-02-22 06:45:57  Response timeout
654850    9.472      2.6  2021-02-22 06:48:34  Response timeout
Use verdi daemon [incr | decr] [num] to increase / decrease the amount of workers
```

### DIFF
--- a/aiida/cmdline/commands/cmd_daemon.py
+++ b/aiida/cmdline/commands/cmd_daemon.py
@@ -87,7 +87,8 @@ def start(foreground, number):
 
 @verdi_daemon.command()
 @click.option('--all', 'all_profiles', is_flag=True, help='Show status of all daemons.')
-def status(all_profiles):
+@click.option('-w', '--fail-on-worker', is_flag=True, help='Return non-zero exit code on unresponsive worker.')
+def status(all_profiles, fail_on_worker):
     """Print the status of the current daemon or all daemons.
 
     Returns exit code 0 if all requested daemons are running, else exit code 3.
@@ -102,17 +103,22 @@ def status(all_profiles):
         profiles = [config.current_profile]
 
     daemons_running = []
+    workers_responded = []
     for profile in profiles:
         client = get_daemon_client(profile.name)
         delete_stale_pid_file(client)
         click.secho('Profile: ', fg='red', bold=True, nl=False)
         click.secho(f'{profile.name}', bold=True)
-        result = get_daemon_status(client)
+        workers_running, result = get_daemon_status(client)
         echo.echo(result)
+        workers_responded.append(workers_running)
         daemons_running.append(client.is_daemon_running)
 
     if not all(daemons_running):
         sys.exit(3)
+
+    if fail_on_worker and not all(workers_responded):
+        sys.exit(4)
 
 
 @verdi_daemon.command()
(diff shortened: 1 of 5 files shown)


ramirezfranciscof:
I like the idea of getting more information!

But re-closing issue #4667, could you explain how this would help with that in specific? I understand that the problem then was circus constantly respawning workers, how would I see that with this new feature, and how would you use it in a test?

chrisjsewell:
> how would you use it in a test?

because it would return a non-zero exit code if any of the workers do not respond

PR #4770, opened 2021-02-22
🐛 FIX: kill_calculation before job submitted

chrisjsewell:
`job_id` will not yet have been set, so you end up getting:

```
02/22/2021 06:59:32 PM <939726> aiida.scheduler.slurm: [ERROR] Error in _parse_kill_output: retval=1; stdout=; stderr=scancel: error: Invalid job id None

02/22/2021 06:59:32 PM <939726> aiida.engine.transports: [ERROR] Exception whilst using transport:
  File "/root/aiida-core/aiida/engine/transports.py", line 110, in request_transport
  File "/root/aiida-core/aiida/engine/processes/calcjobs/tasks.py", line 308, in do_kill
    return execmanager.kill_calculation(node, transport)
  File "/root/aiida-core/aiida/engine/daemon/execmanager.py", line 426, in kill_calculation
    running_jobs = scheduler.get_jobs(jobs=[job_id], as_dict=True)
  File "/root/aiida-core/aiida/schedulers/scheduler.py", line 341, in get_jobs
    retval, stdout, stderr = self.transport.exec_command_wait(self._get_joblist_command(jobs=jobs, user=user))
  File "/root/aiida-core/aiida/schedulers/plugins/slurm.py", line 235, in _get_joblist_command
    command.append(f"--jobs={','.join(joblist)}")
TypeError: sequence item 0: expected str instance, NoneType found
```


chrisjsewell:
this kind of thing would be caught once the whole code base is typed 😉

ltalirz (approved):
thanks @chrisjsewell !

would it be possible to write a small test for this as well or is this very difficult?

ltalirz (commented):
One more thing: could you document what the return value here is supposed to mean?

I guess before this change, `True` meant that the calculation was successfully killed.
On the other hand, I also don't see it return any other value (it just raises an exception).

I guess the return value is somewhat meaningless?

chrisjsewell:
> I guess the return value is somewhat meaningless?

yeh the result ends up here: https://github.com/aiidateam/aiida-core/blob/da872b2c488c33acf1f5c8628ac1226418cbc4de/aiida/engine/processes/calcjobs/tasks.py#L403
which disregards the return value, so I've changed it to None

chrisjsewell:
> would it be possible to write a small test for this as well or is this very difficult?

yeh I can't think of easy way to write a test for this; you would have to set up the scenario where you have started a CalcJob and somehow paused it before it gets to the submit stage.

As mentioned above, I think the best way to guard against these cases where you have not accounted for a null value is with type checking

PR #4771, opened 2021-02-22
👌 IMPROVE: CTRL-C on running process

chrisjsewell:
Do not call `kill` on a process that is already being killed,
Also log a different message,
so that the user can see that the original CTRL-C was actioned.


aiida/engine/runners.py
```diff
@@ -235,6 +235,9 @@ def _run(self, process: TYPE_RUN_PROCESS, *args: Any, **inputs: Any) -> Tuple[Di
 
             def kill_process(_num, _frame):
                 """Send the kill signal to the process in the current scope."""
+                if process_inited.is_killing:
+                    LOGGER.critical('runner received interrupt, process %s already being killed', process_inited.pid)
```
ltalirz, line 239: does this need to be critical or would a warning suffice?
warning is anyhow the default log level

ltalirz (approved):
thanks @chrisjsewell !

that makes sense

aiida/engine/runners.py
chrisjsewell, line 239:
```suggestion
                    LOGGER.warning('runner received interrupt, process %s already being killed', process_inited.pid)
```

chrisjsewell: yeh sounds good

PR #4772, opened 2021-02-23
👌 IMPROVE: add type checking for aiida/orm/nodes/process

chrisjsewell:
As the title suggests

Two changes to the code that should be reviewed:

1. Change `CalcJobNode.get_description` to return a string
2. In aiida/engine/processes/calcjobs/tasks.py, change `node.computer.get_authinfo(node.user)` to `node.get_authinfo()`, to use `CalcJobNode.get_authinfo` which checks if the computer is set.

One remaining question:

`Node.get_hash` defaults to `ignore_errors=True` which will return `None` on any exception.
This does not seem like a very good practice: is there a particular exception we should ignore and also should we log a warning to alert developers/users about any issues?

### DIFF
--- a/aiida/engine/processes/calcjobs/tasks.py
+++ b/aiida/engine/processes/calcjobs/tasks.py
@@ -130,7 +130,7 @@ async def task_submit_job(node: CalcJobNode, transport_queue: TransportQueue, ca
     initial_interval = get_config_option(RETRY_INTERVAL_OPTION)
     max_attempts = get_config_option(MAX_ATTEMPTS_OPTION)
 
-    authinfo = node.computer.get_authinfo(node.user)
+    authinfo = node.get_authinfo()
 
     async def do_submit():
         with transport_queue.request_transport(authinfo) as request:
@@ -177,7 +177,7 @@ async def task_update_job(node: CalcJobNode, job_manager, cancellable: Interrupt
     initial_interval = get_config_option(RETRY_INTERVAL_OPTION)
     max_attempts = get_config_option(MAX_ATTEMPTS_OPTION)
 
-    authinfo = node.computer.get_authinfo(node.user)
+    authinfo = node.get_authinfo()
     job_id = node.get_job_id()
 
     async def do_update():
@@ -240,7 +240,7 @@ async def task_retrieve_job(
     initial_interval = get_config_option(RETRY_INTERVAL_OPTION)
     max_attempts = get_config_option(MAX_ATTEMPTS_OPTION)
 
-    authinfo = node.computer.get_authinfo(node.user)
+    authinfo = node.get_authinfo()
 
     async def do_retrieve():
         with transport_queue.request_transport(authinfo) as request:
@@ -249,7 +249,7 @@ async def task_retrieve_job(
             # Perform the job accounting and set it on the node if successful. If the scheduler does not implement this
             # still set the attribute but set it to `None`. This way we can distinguish calculation jobs for which the
             # accounting was called but could not be set.
-            scheduler = node.computer.get_scheduler()
+            scheduler = node.computer.get_scheduler()  # type: ignore[union-attr]
             scheduler.set_transport(transport)
 
             try:
@@ -300,7 +300,7 @@ async def task_kill_job(node: CalcJobNode, transport_queue: TransportQueue, canc
         logger.warning(f'CalcJob<{node.pk}> killed, it was in the {node.get_state()} state')
         return True
 
-    authinfo = node.computer.get_authinfo(node.user)
+    authinfo = node.get_authinfo()
 
     async def do_kill():
         with transport_queue.request_transport(authinfo) as request:
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -9,10 +9,13 @@
 ###########################################################################
 # pylint: disable=too-many-lines,too-many-arguments
 """Package for node ORM classes."""
+import datetime
 import importlib
+from logging import Logger
 import warnings
 import traceback
-from typing import List, Optional
+from typing import Any, Dict, IO, Iterator, List, Optional, Sequence, Tuple, Type, Union
+from typing import TYPE_CHECKING
 
 from aiida.common import exceptions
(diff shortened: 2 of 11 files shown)


ltalirz:
> `Node.get_hash` defaults to `ignore_errors=True` which will return `None` on any exception.
> This does not seem like a very good practice: is there a particular exception we should ignore and also should we log a warning to alert developers/users about any issues?

Maybe @greschd has some insight on this?

greschd:
Hmm... I think we were mostly trying to be cautious when introducing the caching; because the hashing happens on all nodes things shouldn't fail catastrophically when that doesn't work.

Looking at the code now, the `_make_hash` dispatch raises `ValueError` (which.. should be `TypeError` 🙃) when no specific implementation exists:
https://github.com/aiidateam/aiida-core/blob/8c079175bbb32a6dc3b4ae663c586c7099cbb9a6/aiida/common/hashing.py#L120-L126

While by now there shouldn't be any node types in `aiida-core` that trigger that, I'm not sure if a data plugin could still do it.

So yeah, I'd probably go with:
- change the `ValueError` to `TypeError` in the `_make_hash` fallback
- catch only `TypeError` in `get_hash`
- emit a warning when the exception is caught

greschd:
Oh, it could also fail with `AttributeError` when the module that defines a node doesn't have a top-level `__version__`:

https://github.com/aiidateam/aiida-core/blob/8c079175bbb32a6dc3b4ae663c586c7099cbb9a6/aiida/orm/nodes/node.py#L1078

greschd:
If we want to turn `ignore_errors` to `False` in the long term, I think we should turn it into a config flag for a while. That way we can surface potential errors, while still allowing to flip a switch and get the old behavior back.

chrisjsewell:
thanks @greschd, I will open a separate PR for that change.

This PR should now pass all tests and be ready for review

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -71,7 +71,7 @@ async def task_upload_job(process: 'CalcJob', transport_queue: TransportQueue, c
     initial_interval = get_config_option(RETRY_INTERVAL_OPTION)
     max_attempts = get_config_option(MAX_ATTEMPTS_OPTION)
 
-    authinfo = node.computer.get_authinfo(node.user)
+    authinfo = node.get_authinfo()
```
ltalirz, line 74: this does somewhat change the code path - in `Node.get_authinfo` it goes via `AuthInfo.from_backend_entity` instead of using `Computer.get_authinfo` directly (which I think it should).

I think this should be fixed

aiida/orm/nodes/node.py
```diff
@@ -63,34 +71,34 @@ def _warn_if_not_entered(self, method):
...
+    def __enter__(self) -> Union[IO[str], IO[bytes]]:
         self._was_entered = True
         return self._fileobj.__enter__()
 
-    def __exit__(self, *args):
+    def __exit__(self, *args: Any) -> None:
         self._fileobj.__exit__(*args)
 
-    def __getattr__(self, key):
+    def __getattr__(self, key: str):
```
ltalirz, line 81: not necessarily, right?
what about `_fileobj`?

ltalirz (commented):
I briefly started but don't have time for the rest. 
if someone else would be up for reviewing this, that would be great

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: Indeed, for reference;

for `Computer.get_authinfo` it is:
`return AuthInfo.objects(self.backend).get(dbcomputer_id=self.id, aiidauser_id=user.id)`
but for `CalcJobNode.get_authinfo` it is:
`return AuthInfo.from_backend_entity(self.backend.authinfos.get(computer=computer, user=self.user))`

@sphuber you added the calcjob one, can you clarify if there is any reason not to replace it with: `return self.computer.get_authinfo(self.user)`

aiida/orm/nodes/node.py
chrisjsewell: not sure what you mean here, `__getattr__` always receives a string

ltalirz: ah, my bad, I was thinking about the return value.
I got confused by the fact that it doesn't need to be specified for the __ methods

chrisjsewell:
I've created #4778 for the node hashing

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: ok I have changed this in 3344edc6cc1fdd98ab8720860af73a1ec1569f4e

chrisjsewell:
I will be merging this by the end of today

ltalirz (approved):
thanks @chrisjsewell 

I've checked the two points mentioned in the PR description - I haven't gone through all type hints one by one; I trust that these are accurate.

docs/source/nitpick-exceptions
```diff
@@ -32,26 +32,36 @@ py:class aiida.engine.runners.ResultAndNode
 py:class aiida.engine.runners.ResultAndPk
 py:class aiida.engine.processes.workchains.workchain.WorkChainSpec
 py:class aiida.manage.manager.Manager
+py:class aiida.orm.nodes.node.WarnWhenNotEntered
 py:class aiida.orm.utils.links.LinkQuadruple
 py:class aiida.tools.importexport.dbexport.ExportReport
 py:class aiida.tools.importexport.dbexport.ArchiveData
 py:class aiida.tools.groups.paths.WalkNodeResult
 
-py:class Node
```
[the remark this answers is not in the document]
chrisjsewell: This is actually something I've raised before: https://github.com/sphinx-doc/sphinx/issues/8498 (well similar). It was closed because they thought it was not an issue and I didn't get round to replying, but yeh from this I'm pretty sure it is still an issue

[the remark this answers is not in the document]
chrisjsewell: so if you have the time, maybe add a comment there and I will re-open it

ltalirz (approved):

Issue #3888, opened 2020-04-02
Exception when killing calculation


greschd:
When killing a calculation, I consistently get the following error in the log:

```
04/02/2020 10:05:18 AM <17123> kiwipy.rmq.tasks: [ERROR] The task excepted
Traceback (most recent call last):
  File "/data/aiida/source/aiida-core/aiida/manage/external/rmq.py", line 187, in _continue
    result = yield super()._continue(communicator, pid, nowait, tag)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 4, in raise_exc_info
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/tornado/gen.py", line 1069, in run
    yielded = self.gen.send(value)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/plumpy/process_comms.py", line 554, in _continue
    raise gen.Return(proc.future().result())
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/plumpy/futures.py", line 34, in result
    return super(Future, self).result(timeout)
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 4, in raise_exc_info
plumpy.exceptions.KilledError: Killed through `verdi process kill`

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/kiwipy/rmq/tasks.py", line 126, in _on_task
    result = yield result
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/home/a-dogres/.virtualenvs/aiida/lib/python3.6/site-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "<string>", line 4, in raise_exc_info
[...]

PR #4773, opened 2021-02-23
🐛 FIX: `ModificationNotAllowed` on workchain kill

chrisjsewell:
copied from https://github.com/aiidateam/aiida-core/issues/3888#issuecomment-784345729

1. in `kill` the parent is killed first, then the children: https://github.com/aiidateam/aiida-core/blob/8c079175bbb32a6dc3b4ae663c586c7099cbb9a6/aiida/engine/processes/process.py#L294
2. when entering the `Wait` state awaitables (e.g. children) are each assigned to `on_process_finished`: https://github.com/aiidateam/aiida-core/blob/8c079175bbb32a6dc3b4ae663c586c7099cbb9a6/aiida/engine/processes/workchains/workchain.py#L263
3. when the child is killed, this calls `resolve_awaitable`, which tries to update the status: https://github.com/aiidateam/aiida-core/blob/8c079175bbb32a6dc3b4ae663c586c7099cbb9a6/aiida/engine/processes/workchains/workchain.py#L145
4. but doh the parent is already terminated and the node sealed -> `ModificationNotAllowed` 🤦 

so here in `resolve_awaitable` we want to check if the parent is already in a terminal state and, if so, skip the ctx and status update.
We still remove the awaitable from `self._awaitables`, to remove the references to the child process (so it can be garbage collected)

### DIFF
--- a/aiida/engine/processes/workchains/workchain.py
+++ b/aiida/engine/processes/workchains/workchain.py
@@ -151,11 +151,6 @@ class WorkChain(Process):
         """
         self._awaitables.remove(awaitable)
 
-        if self.has_terminated:
-            # this can occur, for example, if the process was killed or excepted
-            # then we should not try to update it
-            return
-
         if awaitable.action == AwaitableAction.ASSIGN:
             self.ctx[awaitable.key] = value
         elif awaitable.action == AwaitableAction.APPEND:
@@ -171,6 +166,11 @@ class WorkChain(Process):
             assert f'Unknown awaitable action: {awaitable.action}'
 
         awaitable.resolved = True
+
+        if self.has_terminated:
+            # this can occur, for example, if the process was killed or excepted
+            # then we should not try to update it
+            return
 
         self._update_process_status()
 
--- a/aiida/engine/processes/workchains/workchain.py
+++ b/aiida/engine/processes/workchains/workchain.py
@@ -151,6 +151,11 @@ class WorkChain(Process):
         """
         self._awaitables.remove(awaitable)
 
+        if self.has_terminated:
+            # this can occur, for example, if the process was killed or excepted
+            # then we should not try to update it
+            return
+
         if awaitable.action == AwaitableAction.ASSIGN:
             self.ctx[awaitable.key] = value
         elif awaitable.action == AwaitableAction.APPEND:
--- a/aiida/engine/processes/workchains/workchain.py
+++ b/aiida/engine/processes/workchains/workchain.py
@@ -167,12 +167,10 @@ class WorkChain(Process):
 
         awaitable.resolved = True
 
-        if self.has_terminated:
-            # this can occur, for example, if the process was killed or excepted
+        if not self.has_terminated:
+            # the process may be terminated, for example, if the process was killed or excepted
             # then we should not try to update it
-            return
-
-        self._update_process_status()
+            self._update_process_status()
 
     def to_context(self, **kwargs: Union[Awaitable, ProcessNode]) -> None:
         """Add a dictionary of awaitables to the context.


chrisjsewell:
hmm weird that tests are failing because an already terminated process is not being updated.
trying moving the skip to after the ctx has been updated

aiida/engine/processes/workchains/workchain.py
```diff
@@ -167,6 +167,11 @@ def resolve_awaitable(self, awaitable: Awaitable, value: Any) -> None:
 
         awaitable.resolved = True
 
+        if self.has_terminated:
```
ltalirz, line 170: to make it even more explicit (only update, if process not terminated), you could say

```
if not self.has_terminated:
   self._update_process_status()
```

ltalirz (changes requested):
thanks @chrisjsewell !

this looks correct to me - but would it be possible to add a test here?
killing a calculation is a pretty important use case, so we should really be testing this.
I don't care whether you add it to the daemon tests or to the regular test base

aiida/engine/processes/workchains/workchain.py
chrisjsewell: done

chrisjsewell:
> this looks correct to me - but would it be possible to add a test here?

fair, I will look into it...

chrisjsewell:
> this looks correct to me - but would it be possible to add a test here?

I'm admitting defeat on this and going to punt it to a seperate issue;

with `tests/engine/test_work_chain.py::TestWorkChainAbortChildren::test_simple_kill_through_process` you can get the exception to manifest, with caveats:

1. you need to wait longer than the end of the process, because the exception is raised by later code, e.g.

```python
        async def run_async():
            await run_until_waiting(process)

            result = process.kill()
            if asyncio.isfuture(result):
                await result

            with self.assertRaises(plumpy.KilledError):
                await process.future()

			await asyncio.sleep(1)
```

but what do you wait for to make this consistent?

you could do something like this, but it seems hacky:

```python
        async def no_awaitables(process):
            while process._awaitables:
                await asyncio.sleep(0.1)

        async def run_async():
            await run_until_waiting(process)

            result = process.kill()
            if asyncio.isfuture(result):
                await result

            with self.assertRaises(plumpy.KilledError):
                await process.future()

            await asyncio.wait_for(no_awaitables(process), timeout=1.0)
```

2. pytest still does not actually fail if you do (1), you have to manually except the test, then you will see the excetion in the captured log:

```python
        self.assertEqual(process.node.is_killed, True)
    
>       raise
E       RuntimeError: No active exception to reraise

tests/engine/test_work_chain.py:1043: RuntimeError
------------------------------------------------------------------------------------------- Captured log call -------------------------------------------------------------------------------------------
ERROR    asyncio:base_events.py:1619 Task exception was never retrieved
future: <Task finished coro=<ProcessCallback.run() done, defined at /Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/events.py:99> exception=RuntimeError('Cannot transition from ProcessState.KILLED to ProcessState.EXCEPTED')>
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/events.py", line 103, in run
    await self._callback(*self._args, **self._kwargs)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/processes.py", line 548, in _run_task
    result = await coro(*args, **kwargs)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/utils.py", line 246, in wrap
    return coro_or_fn(*args, **kwargs)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/aiida/engine/processes/workchains/workchain.py", line 308, in on_process_finished
    self.resolve_awaitable(awaitable, value)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/aiida/engine/processes/workchains/workchain.py", line 173, in resolve_awaitable
    self._update_process_status()
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/aiida/engine/processes/workchains/workchain.py", line 192, in _update_process_status
    self.node.set_process_status(None)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/aiida/orm/nodes/process/process.py", line 202, in set_process_status
    self.delete_attribute(self.PROCESS_STATUS_KEY)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/aiida/orm/utils/mixins.py", line 193, in delete_attribute
    raise exceptions.ModificationNotAllowed('attributes of a sealed node are immutable')
aiida.common.exceptions.ModificationNotAllowed: attributes of a sealed node are immutable

  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/asyncio/tasks.py", line 249, in __step
    result = coro.send(None)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/events.py", line 106, in run
    self._process.callback_excepted(self._callback, exc_info[1], exc_info[2])
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/processes.py", line 513, in callback_excepted
    self.fail(exception, trace)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/base/state_machine.py", line 97, in transition
    result = wrapped(self, *a, **kw)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/processes.py", line 1080, in fail
    self.transition_to(process_states.ProcessState.EXCEPTED, exception, trace_back)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/base/state_machine.py", line 335, in transition_to
    self.transition_failed(initial_state_label, label, *sys.exc_info()[1:])
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/base/state_machine.py", line 351, in transition_failed
    raise exception.with_traceback(trace)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/base/state_machine.py", line 317, in transition_to
    self._exit_current_state(new_state)
  File "/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py37-django/lib/python3.7/site-packages/plumpy/base/state_machine.py", line 376, in _exit_current_state
    raise RuntimeError('Cannot transition from {} to {}'.format(self._state.LABEL, next_state.label))
RuntimeError: Cannot transition from ProcessState.KILLED to ProcessState.EXCEPTED
```

how can you get pytest to fail on such (background) asyncio exceptions?

Note, you can intercept the exception by replacing the loops exception handler:

```python
        def _exception_handler(loop, context):
            print(context)

        runner.loop.set_exception_handler(_exception_handler)
        runner.schedule(process)
        runner.loop.run_until_complete(run_async())
```

but I have no idea how you can use this to actually make the test fail

I've tried googling answers, but could not find any obvious answers 😢

chrisjsewell:
opened #4775

PR #4774, opened 2021-02-23
pytest: reset test manager after test run

ltalirz:
fixes part of #4722 

Tests written in pytest did not reset the AiiDA manager, resulting in the
global singleton manager instance remaining after the end of the test.

This was already taken care of in the teardown method of the unit test
class AiidaTestCase.

### DIFF
--- a/.github/workflows/tests.sh
+++ b/.github/workflows/tests.sh
@@ -33,7 +33,7 @@ python ${SYSTEM_TESTS}/test_plugin_testcase.py  # uses custom unittest test runn
 
 # Until the `${SYSTEM_TESTS}/pytest` tests are moved within `tests` we have to run them separately and pass in the path to the
 # `conftest.py` explicitly, because otherwise it won't be able to find the fixtures it provides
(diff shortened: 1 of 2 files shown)


ltalirz:
@chrisjsewell Do you have an idea why this change breaks the memory leak system tests (first commit)?
It looks like the first test passes, and then since the second test will need to recreate the manager (?) the process ends up staying in memory.

When I uncomment the memory leak system tests, the rest of the test suite runs fine (second commit) although the warnings about the unclosed event loop remain.

chrisjsewell:
> Do you have an idea why this change breaks the memory leak system tests (first commit)?

hmm no idea off-hand, will think about it

chrisjsewell:
Perhaps the asycio loop is closed before all background processes are finished and the process has a chance to "clean up" 🤷

ramirezfranciscof:
I understand this is currently WIP, correct? I'll set it to draft, once it is ready to review (or if I misinterpreted the status) you can set the status back and someone will review it.

ltalirz:
Great, feel free to close if you're sure this is fixed

ltalirz:
https://github.com/aiidateam/aiida-core/pull/5330 has been merged

Issue #4758, opened 2021-02-17
Delete unused `ProcessFuture`


ltalirz:
The `ProcessFuture` was intended for a use case that never came to fruition (maybe @muhrin can add 1-2 sentences on that).

`ProcessFuture`s are currently not used in the AiiDA codebase and, when I tested them for investigating the memory leak, they seemed to be broken.
We agreed together with @muhrin to remove them from the code.

chrisjsewell:
yep delete away

ltalirz:
Turns out that `runner.get_process_future()` was used in some *tests* in order to wait until a process completes.
In that case it must (?) work at least to some degree, so let's keep it for the moment.

Issue #4745, opened 2021-02-11
TimeoutError while running many workflows


giovannipizzi:
While running ~1000 workchains (see description [in this comment]), I got some TimeoutErrors.

These are the workflows that a non-zero error status. Some were for convergence issues of QE, but a few were exceptions:
`verdi process status 110696 111576 110019 128494 107304 107365 107405 107439 108306 108636  | grep Except`:
```
    └── PwBaseWorkChain<112063> Excepted [4:validate_resources]
    └── PwBaseWorkChain<112118> Excepted [6:while_(should_run_process)(1:run_process)]
    └── PwBaseWorkChain<110522> Excepted [5:if_(should_run_init)]
        └── PwCalculation<109485> Excepted
        └── PwCalculation<109846> Excepted
```

Here is their report:
`for i in  112063 112118 110522 109485 109846 ; do verdi process report $i ; done`

```
2021-02-10 22:10:17 [35132 | ERROR]: Traceback (most recent call last):
  File "/home/pizzi/git/aiida-core/aiida/manage/external/rmq.py", line 206, in _continue
    result = await super()._continue(communicator, pid, nowait, tag)
  File "/home/pizzi/.virtualenvs/aiida-dev/lib/python3.7/site-packages/plumpy/process_comms.py", line 613, in _continue
    await proc.step_until_terminated()
  File "/home/pizzi/.virtualenvs/aiida-dev/lib/python3.7/site-packages/plumpy/processes.py", line 1213, in step_until_terminated
    await self.step()
  File "/home/pizzi/.virtualenvs/aiida-dev/lib/python3.7/site-packages/plumpy/processes.py", line 1199, in step
    self.transition_to(next_state)
  File "/home/pizzi/.virtualenvs/aiida-dev/lib/python3.7/site-packages/plumpy/base/state_machine.py", line 337, in transition_to
    self.transition_failed(initial_state_label, label, *sys.exc_info()[1:])
  File "/home/pizzi/.virtualenvs/aiida-dev/lib/python3.7/site-packages/plumpy/base/state_machine.py", line 353, in transition_failed
    raise exception.with_traceback(trace)
  File "/home/pizzi/.virtualenvs/aiida-dev/lib/python3.7/site-packages/plumpy/base/state_machine.py", line 322, in transition_to
    self._enter_next_state(new_state)
[...]

PR #4776, opened 2021-02-24
⬆️ UPDATE: kiwipy/plumpy

chrisjsewell:
Update to new patch versions:

kiwipy v0.7.3:

- 👌 IMPROVE: Add debug logging for sending task/rpc/broadcast to RMQ.
- 👌 IMPROVE: Close created asyncio loop on RmqThreadCommunicator.close

plumpy v0.18.6:

- 👌 IMPROVE: Catch state change broadcast timeout

When using an RMQ communicator, the broadcast can timeout on heavy loads to RMQ
(for example see https://github.com/aiidateam/aiida-core/issues/4745).
This broadcast is not critical to the running of the process,
and so a timeout should not except it.
In aiida-core, the broadcast is subscribed to by `verdi process watch` (obviously not critical), in `aiida/engine/processes/futures.py:ProcessFuture` (which is unused https://github.com/aiidateam/aiida-core/issues/4758) and in `aiida/engine/runners.py:Runner.call_on_process_finish` which anyway has a backup polling mechanism on the node.

Note also, as of https://github.com/aiidateam/aiida-core/pull/4712, this timeout is configurable as `rmq.task_timeout`.

Also ensure the process PID is included in all log messages.


chrisjsewell:
just need to get the releases up on Conda

PR #4778, opened 2021-02-24
👌 IMPROVE: capture of node hashing errors

chrisjsewell:
Currently all exceptions are caught and ignored.
This commit adds a specific `HashingError` exception,
for known failure modes.
Only this exception is caught, if `ignore_errors=True`,
and the exception logged.

Also an `aiida_caplog` pytest fixture is added,
to enable logs from `AiiDA_LOGGER` to be captured.


chrisjsewell:
cc @greschd

chrisjsewell:
one other thing to consider https://github.com/aiidateam/aiida-core/pull/4772#issuecomment-784042328

> If we want to turn ignore_errors to False in the long term, I think we should turn it into a config flag for a while. That way we can surface potential errors, while allowing to flip a switch and get the old behavior back.

ltalirz:
I'll leave this to @greschd to review

greschd (approved):
Very nice, thanks @chrisjsewell!

I like the `aiida_caplog`, I can use it in #4707 🚀

PR #4779, opened 2021-02-24
Simplify AiidaTestCase implementation

ltalirz:
This is an attempt at simplifying the AiidaTestCase setup (not yet replacing it with pytest fixtures, but hopefully getting one step closer to doing so eventually)

In particular
 * only truly backend-specific code is left in the backend-specific test classes
 * introduces `refurbish_db()` which includes the combination of cleaning the db and repopulating it with a user (which is a common combination)
 *  move creation of default computer from `setUpClass` to "on demand" (not needed by many tests)
 * merges `reset_database` and `clean_db` function that basically did the same
 * factors out the `get_or_create_default_user` function so that it can be reused outside the AiidaTestCase (`verdi setup`, pytest fixtures, ...)
 * add `orm.Computer.objects.get_or_create` (in analogy to similar methods for user, group, ...)

Note: While this PR cleans up a couple of unnecessary `clean_db` calls and gets rid of unnecessary complexity, it does *not* switch to a mode where the database is cleaned *between every test*.
While some subclasses of `AiidaTestCase` do this, `AiidaTestCase` itself only cleans the database in `setupClass` and some subclasses do significant test setup on the class level, which might slow things down if they have to be done for every test...


ltalirz:
Note: this *does* switch all of the AiidaArchiveTestCase tests to be cleaned in between every test by default (rather than adding this cleaning individually in the many subclasses - not sure whether it even changes anything).

Since timings on github actions ci are not very reliable I've timed `pytest tests/tools/importexport/` locally

this branch:
try 1: real	0m51.025s
try 2: real 0m49.295s

`develop`: 0m49.588s

chrisjsewell:
Thanks @ltalirz this all looks good to me!

> Note: While this PR cleans up a couple of unnecessary clean_db calls and gets rid of unnecessary complexity, it does not switch to a mode where the database is cleaned between every test.
While some subclasses of AiidaTestCase do this, AiidaTestCase itself only cleans the database in setupClass

so in terms of migration to pytest, currently we have:

```python
def clear_database(clear_database_after_test):
    """Alias for 'clear_database_after_test'.

    Clears the database after each test. Use of the explicit
    'clear_database_after_test' is preferred.
    """


def clear_database_after_test(aiida_profile):
    """Clear the database after the test."""
    yield
    aiida_profile.reset_db()


def clear_database_before_test(aiida_profile):
    """Clear the database before the test."""
    aiida_profile.reset_db()
    yield
```

I've already said that I think we should get rid of `clear_database_after_test` because it misses the point of a unit test (you only care about that test, not the one after), and also `clear_database` because it is misleading in that, without reading the docstring, I would assume that it clears the database before the test not after.

but also it might be good to have something like:

```python
@pytest.fixture(scope='class')
def clear_database_before_test_class(aiida_profile):
    """Clear the database before the test."""
    aiida_profile.reset_db()
    yield
```

which would then be a direct replacement:

```
class TestSomething(AiidaTestCase):
```

to

```
@pytest.mark.usefixtures('clear_database_before_test_class')
class TestSomething:
```

ltalirz:
Thanks for the review and the suggestions, @chrisjsewell 

I propose to make adaptations to the pytest framework in a separate PR.
Do I get it right that you approve this PR?

chrisjsewell (approved):

chrisjsewell:
indeed

PR #4781, opened 2021-02-25
Reuse functionality between AiidaTestCase and pytest fixtures

ltalirz:
* add a `User.orm.objects.get_or_create_default` method that can be used anywhere (`verdi setup`, `AiidaTestCase`, pytest fixtures, ...)
* move the only remaining method of the test backend class (`clean_db`) to a private method of the backend (could also be standalone if that is preferred)


chrisjsewell:
note in #4782 I'm trying to remove all `AiidaTestCase` 😄  so this will feed into it

ltalirz:
Thanks, yes I've seen - the changes made here are still relevant I think, and I will then abstain from any work converting the aiidatestcase to pytest.

This reminds me - I'm not sure whether we actually can just remove the aiidatestcase entirely before 2.0.0 since it's being used by a couple of plugins https://github.com/search?q=aiidatestcase&type=code

I will check whether any of those are aiida 1 compatible (and whether any of them used the `create_user` / `create_computer` methods I deleted in https://github.com/aiidateam/aiida-core/commit/c07e3ef0b9e9fff64888115aaf30479bbd76392e

ltalirz:
AiidA 1.0 plugins using aiidatestcase:

https://github.com/dev-zero/aiida-gaussian-datatypes/blob/c069c35f9738541d5aee38a37837e4cb666e3335/tests/test_cli.py
https://github.com/zhubonan/aiida-castep/blob/0f172bff7a2f78b5556c91b94d870ac791f171b3/aiida_castep/tests/dbtests/dbtestcalculation.py

ltalirz:
I.e. we should not delete it (but of course we can still migrate our internal tests).

ltalirz:
fine to close

PR #4782, opened 2021-02-25
🧪 TESTS: convert `AiidaTestCase` -> pytest

chrisjsewell:
Conversion process:

- `unittest2pytest -n -w path/to/file` (see https://github.com/pytest-dev/unittest2pytest)
- add `# pylint: disable=no-self-use`
- replace `AiidaTestCase` with `@pytest.mark.usefixtures('aiida_profile_clean_class')` (I would say this is not recommended unless there are particular performance reasons, since every test should be independent), `@pytest.mark.usefixtures('aiida_profile_clean')`, or using an autouse fixture (see below)
  - replace `def setUp(self):` with `def setup_method(self):` or an autouse fixture
- remove `from aiida.backends.testbase import AiidaTestCase`
- change `@unittest.skip` -> `@pytest.mark.skip`, `@unittest.skipIf` -> `@pytest.mark.skipif`

The were some issues where`assertAlmostEquals` had been used for strange cases, related to this I also added an `__abs__` method to `NumericType`

Annoyingly there are some limitations of using classes with pytest, one being that a class `usefixtures` decorator is called AFTER a `setup_class` method (see
https://stackoverflow.com/questions/31484419/pytest-setup-class-after-fixture-initialization_) so any nodes created in `setup_class` will be wiped from the database by `aiida_profile_clean_class`.
Also class methods do not accept fixtures.
This makes it difficult to do an exact 1-to-1 replacement of any `setUpClass` -> `setup_class` methods used, and also replace the use of `self.computer`

edit: https://docs.pytest.org/en/latest/how-to/unittest.html#using-autouse-fixtures-and-accessing-other-fixtures allows for using fixtures per test, and so can solve the use of `self.computer` etc e.g.

```python
# pylint: disable=attribute-defined-outside-init,no-self-use,unused-argument

class MyTestClass:
    @pytest.fixture(autouse=True)
    def init_db(self, aiida_profile_clean, backend, aiida_localhost):
        self.backend = backend
        self.computer = aiida_localhost
    def test_something(self):
        self.computer
```

or if you don't need any instance attributes:

```python
@pytest.mark.usefixtures('aiida_profile_clean')
class MyTestClass:
   def test_something(self):
```

### DIFF
--- a/aiida/manage/tests/pytest_fixtures.py
+++ b/aiida/manage/tests/pytest_fixtures.py
@@ -70,6 +70,13 @@ def clear_database_before_test(aiida_profile):
     yield
 
 
+@pytest.fixture(scope='class')
+def clear_database_before_test_class(aiida_profile):
+    """Clear the database before a test class."""
+    aiida_profile.reset_db()
+    yield
+
+
 @pytest.fixture(scope='function')
 def temporary_event_loop():
     """Create a temporary loop for independent test case"""
--- a/aiida/orm/nodes/data/numeric.py
+++ b/aiida/orm/nodes/data/numeric.py
@@ -124,3 +124,6 @@ class NumericType(BaseType):
 
     def __int__(self):
         return int(self.value)
+
+    def __abs__(self):
+        return abs(self.value)
--- a/tests/engine/test_work_chain.py
+++ b/tests/engine/test_work_chain.py
@@ -10,14 +10,12 @@
 # pylint: disable=too-many-lines,missing-function-docstring,invalid-name,missing-class-docstring,no-self-use
 """Tests for the `WorkChain` class."""
 import inspect
-import unittest
 import asyncio
 
 import plumpy
 import pytest
 
 from aiida import orm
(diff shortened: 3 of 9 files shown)


chrisjsewell:
cc @ltalirz, this should pass all tests for the files I have changed (plenty more to do), so if you want to have a look do a quick "pre-review" for any issues, before I apply this to any more files?

ltalirz (commented):
thanks @chrisjsewell - looks good to me, no obvious issues come to mind.

when transforming more tests you may run into the fact that the AiidaTestCase is a bit more rigorous in cleaning not only the db but also resetting various caches (see the `clean_db` method).
These may need to be ported over to the `reset_db` method of the ProfileManager

ltalirz:
I guess this can then also be closed?

chrisjsewell:
> I guess this can then also be closed?

Yep 👍, superseded by #5372

Perhaps some of the tips for writing tests, in the initial comment, should be copied somewhere?

Issue #4595, opened 2020-11-28
Issues with new asyncio daemon (stress-test)


giovannipizzi:
I have tried a stress-test of the new AiiDA daemon after the replacement of tornado with asyncio.

I have updated to the most recent develop (716a1d8f6), updated also aiida-qe to develop (commit 1a9713aefbcd235c20ecfb65d0df226b5544bf7d of that repo), pip installed both, run `reentry scan`, and stopped+started the daemon.

I try to roughly describe also what I've been doing.
Then, roughly, I have prepared a script to submit something of the order of ~2000 relax workflows.
While the submission was happening, I quickly reached the number of slots (warning message at the end of verdi process list indicating a % > 100%), so I did `verdi daemon incr 7` to work with 8 workers.
After having submitted more than half of the workflows, I stopped because anyway 8 workers weren't enough and I didn't want to overload the supercomputer with too many connections from too many workers.
I left it run overnight, the next morning I was in a stalled situation all slots were taken, so I increased a bit more the workers, and after a while submitted the rest of the workflows, and let them finish.
Since I realised that most were excepting (see below), I also stopped the daemon (that took a bit, made sure it was stopped, and started again with just one worker to finish the work.

I have seen a number of issues unfortunately ( :-( ) where most calculations had some kind of problem. Pinging @sphuber @unkcpz @muhrin as they have been working on this so they should be able to help debugging/fixing the bugs.

I am going report below as different comments some of the issues that I'm seeing, but I'm not sure how to debug more, so if you need specific logs please let me know what to run (or @sphuber I can give you temporarily access to the machine if it's easier).
While I write I have the last few (~30) jobs finishing, but I can already start reporting the issues I see.

giovannipizzi:
- a few of the relax work chains are left in the `Created` state, as well as some of the internals PwBaseWorkChains, and even some PwCalculations:
```
27034  14h ago    PwRelaxWorkChain  ⏹ Created
33506  7h ago     PwCalculation     ⏹ Created
33986  7h ago     PwCalculation     ⏹ Created
33983  7h ago     PwBaseWorkChain   ⏹ Created
33989  7h ago     PwCalculation     ⏹ Created
33992  6h ago     PwBaseWorkChain   ⏹ Created
34025  6h ago     PwBaseWorkChain   ⏹ Created
34028  6h ago     PwBaseWorkChain   ⏹ Created
34090  6h ago     PwCalculation     ⏹ Created
[...]

PR #4786, opened 2021-02-27
👌 IMPROVE: Add 'exception' to projection mapping

chrisjsewell:
When diagnosing issues with multiple processes excepting (e.g. #4595),
it would be helpful to view the exception type in `verdi process list`.

This PR adds `exception` to the list of allowed projections,
and also standardises the way the exception is set on the node (capturing both the type and message).

For example submitting a calculation after adding `raise AiidaException('oopsie!')` before: https://github.com/aiidateam/aiida-core/blob/c07e3ef0b9e9fff64888115aaf30479bbd76392e/aiida/manage/external/rmq.py#L206

```console
$ verdi process list -S excepted -P ctime pk process_state exception
Created       PK  Process state    Exception
---------  -----  ---------------  -----------------------------------------------
9m ago     10385  Excepted         aiida.common.exceptions.AiidaException: oopsie!
```

### DIFF
--- a/aiida/manage/external/rmq.py
+++ b/aiida/manage/external/rmq.py
@@ -11,6 +11,7 @@
 """Components to communicate tasks to RabbitMQ."""
 from collections.abc import Mapping
 import logging
+import traceback
 
 from kiwipy import communications, Future
 import pamqp.encode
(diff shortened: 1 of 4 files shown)


aiida/manage/external/rmq.py
```diff
@@ -154,7 +155,7 @@ def handle_continue_exception(node, exception, message):
 
         if not node.is_excepted and not node.is_sealed:
             node.logger.exception(message)
```
ltalirz, line 157: Is there a reason thy this goes to the logger rather than to the process report as in the other case?

Could one factor out these two lines into a private node member function that is called in both places?

ltalirz (approved):
thanks @chrisjsewell , this looks good to me

aiida/manage/external/rmq.py
chrisjsewell: The `report` method is on the `Process` class, not `ProcessNode`. `report` then calls the process logger, which in practice should always be the node logger.

But I agree it would be good to streamline more

PR #4787, opened 2021-03-01
docs: reorder/simplify caching howto

ltalirz:
The howto on enabling caching has been reordered to move the concepts to
the beginning and technical details to where they fit better.

The figure has been simplified (complexity introduced by second input
node unnecessary).

Added explicit mention of the fact that hashing is enabled by default
(which may not be obvious).


docs/source/howto/run_codes.rst
```diff
@@ -399,35 +399,37 @@ See :ref:`topics:processes:usage:launching` and :ref:`topics:processes:usage:mon
...
 
-How to save computational resources using caching
-=================================================
+How to save compute time with caching
+=====================================
 
-There are numerous reasons why you might need to re-run calculations you have already run before.
-
-    If you run a great number of complex workflows in high-throughput, that each may repeat the same calculation, or you have to restart an entire workflow that failed somewhere half-way through.
+In practice, one often ends up re-running calculations one has already run before - be it because two workflows include the same calculation or because one needs to restart a workflow that failed due to some infrastructure problem.
```
chrisjsewell, line 405:
```suggestion
Over the course of a project, you may end up re-running the same calculations multiple times - be it because two workflows include the same calculation or because one needs to restart a workflow that failed due to some infrastructure problem.
```

docs/source/howto/run_codes.rst
```diff
@@ -399,35 +399,37 @@ See :ref:`topics:processes:usage:launching` and :ref:`topics:processes:usage:mon
...
 
 Since AiiDA stores the full provenance of each calculation, it can detect whether a calculation has been run before and, instead of running it again, simply reuse its outputs, thereby saving valuable computational resources.
 This is what we mean by **caching** in AiiDA.
 
-.. versionchanged:: 1.6.0
+With caching enabled, AiiDA searches the database for a calculation of the same :ref:`hash<topics:provenance:caching:hashing>`.
```
chrisjsewell, line 410:
```suggestion
With caching enabled, AiiDA searches the database for a (successfully completed) calculation of the same :ref:`hash<topics:provenance:caching:hashing>`.
```

I assume this is the case!

docs/source/howto/run_codes.rst
```diff
@@ -438,39 +440,35 @@ You can view and alter your current caching configuration through the command-li
...
+    Configuring caching via the ``cache_config.yml`` is deprecated as of AiiDA 1.6.0.
+    Existing ``cache_config.yml`` files will be migrated to the central ``config.json`` file automatically.
 
-.. _fig_caching:
-.. figure:: include/images/caching.png
-    :align: center
-    :height: 350px
 
-    When reusing the results of a calculation **C** for a new calculation **C'**, AiiDA simply makes a copy of the result nodes and links them up as usual.
+From this point onwards, when you launch a new calculation, AiiDA will compare its hash (depending both on the type of calculation and its inputs, see :ref:`topics:provenance:caching:hashing`) against other calculations already present in your database.
```
chrisjsewell, line 466:
```suggestion
From this point onwards, when you launch a new calculation, AiiDA will compare its hash (a fixed size string, unique for a calulation's type and inputs, see :ref:`topics:provenance:caching:hashing`) against other calculations already present in your database.
```

chrisjsewell (changes requested):
thanks @ltalirz, just a few suggestions

docs/source/howto/run_codes.rst
ltalirz: that's a good point - is this the case @greschd ?

greschd: yep 👍

greschd: aah well, these days it's a bit more complicated... it depends on the exit code of the calculation

greschd: By default even "unsuccessful" calculations are considered, only when an exit code with `invalidates_cache` attribute is emitted is the calculation _not_ taken into account.

The reason for that is that even a non-zero exit code could be either a "partial success", or an "unconditional failure" (that doesn't depend on machine details). For these, we also don't want to re-run calculations.

So it's up to plugin developers to decide what a "cachable" failure is -- of course with the drawback that it relies on them implementing it.

greschd: The default for non-zero exit codes was switched shortly before 1.0 AFAICR, so if you're porting older documentation that might be wrong.

greschd: But the simple answer for this part of the docs is: No, that's not quite right.

ltalirz: Ok, so if I understand correctly, then I should simply dismiss Chris's suggestion here,
and we should probably somewhere else explain in more detail how calculations with non-zero exit codes are treated.

greschd: Yep. It's described [in the internals section](https://aiida.readthedocs.io/projects/aiida-core/en/latest/internals/engine.html#controlling-caching), but could probably use a more approachable explanation.

greschd: Ah, and there's another part [in topics](https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/provenance/caching.html#controlling-caching).

ltalirz: thanks, added a couple cross-references

chrisjsewell (approved):

Issue #4679, opened 2021-01-26
Computer setup: proposed configure command does not include profile


ltalirz:
### Is your feature request related to a problem? Please describe

When configuring a new computer, the command proposes to configure the computer next

```
(aiida) ubuntu@qmobile:~$ verdi -p failed-mofs-2021-01-14 computer setup --config aiida-code-registry/daint.cscs.ch/multicore/computer-setup.yaml
Success: Computer<4> daint-mc created
Info: Note: before the computer can be used, it has to be configured with the command:
Info:   verdi computer configure ssh daint-mc
```

However, this command is only correct, if the computer is to be configured on the default AiiDA profile.

### Describe the solution you'd like

In order to be able to just copy-paste this command, there should be a check on whether the `setup` command is being run on the default AiiDA profile, and the proposed command should provide the AiiDA profile parameter (if needed).

kenkirito:
can i work on this

lkbhitesh07:
Hello @ltalirz , I would like to work, but while setting up the environment I'm getting this error, will you please help me out.
Thanks
```
Defaulting to user installation because normal site-packages is not writeable
Obtaining file:///home/ht/opensource/aiida-core
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: /usr/bin/python3 /home/ht/.local/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmp3op_m41n
       cwd: /home/ht/opensource/aiida-core
  Complete output (10 lines):
  Traceback (most recent call last):
    File "/home/ht/.local/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 280, in <module>
      main()
    File "/home/ht/.local/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 263, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/home/ht/.local/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 108, in get_requires_for_build_wheel
      backend = _build_backend()
    File "/home/ht/.local/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 99, in _build_backend
      obj = getattr(obj, path_part)
  AttributeError: module 'setuptools.build_meta' has no attribute '__legacy__'
  ----------------------------------------
[...]

PR #4790, opened 2021-03-04
setup: move away from legacy build backend

ltalirz:
AiiDA was still using the `setuptools.build_meta:__legacy__` build
backend.
This can lead to issues when using a *system* version of setuptools
< 40.8.0, see [1].

I've tested that the `reentry_register` hook still works with the new
build backend.

[1] https://github.com/pypa/setuptools/issues/1694#issuecomment-466010982

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = ["setuptools>=40.8.0", "wheel", "reentry~=1.3", "fastentrypoints~=0.12"]
-build-backend = "setuptools.build_meta:__legacy__"
+build-backend = "setuptools.build_meta"
 
 [tool.pylint.master]
 load-plugins = "pylint_django"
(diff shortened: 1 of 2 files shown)


ltalirz:
see also https://github.com/aiidateam/aiida-core/issues/4679#issuecomment-790697089

chrisjsewell:
Why were we originally using legacy?

ltalirz:
I'm not sure there was ever a good reason for it.
We originally did not specify a backend, which implicitly defaults to the legacy one.
It was made explicit in a2bebb422f4a7b75e8ef65fd797f128abf12c6cc by @dev-zero

csadorf (approved):
It seems to me that specifying `__legacy__` explicitly was [never intended use](https://github.com/pypa/setuptools/issues/1689) and it is not clear to me why we specified it in this way. The [setuptools documentation ](https://setuptools.readthedocs.io/en/latest/build_meta.html)also does not give indication for this.

LGTM!

Issue #4648, opened 2021-01-09
Restarting the daemon excepts all jobs (aiida-core 1.6, python 3.7)


giovannipizzi:
Edit: this issue only occurs in python 3.7, so unitl the issue is fuly resolved, a "fix" is to upgrade to python 3.8 

---

### Steps to reproduce

1. Submit a job to the daemon
2. Start the daemon, if not already running
3. Stop or restart the daemon

### Expected behavior

The daemon stops in a reasonably short time, and the job is "frozen" and will safely continue when the daemon restarts.

### Actual problematic behaviour

Instead, when stopping and/or restarting the deamon, I get a `TIMEOUT`. 
Then, I get things like this from `verdi process list`:

```
  PK  Created    Process label    Process State    Process status
----  ---------  ---------------  ---------------  -------------------------------------
 184  20h ago    PwCalculation    ⨯ Excepted       Transport task update was interrupted
 189  1h ago     PwCalculation    ⨯ Excepted       Transport task submit was interrupted
```

and `verdi process report` shows this:
```
$ verdi process report 184
*** 184: None
*** Scheduler output: N/A
*** Scheduler errors: N/A
*** 3 LOG MESSAGES:
+-> ERROR at 2021-01-08 18:53:56.557646+00:00
 | Traceback (most recent call last):
 |   File "/home/pizzi/git/aiida-core/aiida/engine/utils.py", line 171, in exponential_backoff_retry
 |     result = await coro()
 |   File "/home/pizzi/git/aiida-core/aiida/engine/processes/calcjobs/tasks.py", line 177, in do_update
 |     job_info = await cancellable.with_interrupt(update_request)
 |   File "/home/pizzi/git/aiida-core/aiida/engine/utils.py", line 87, in with_interrupt
 |     result = await next(wait_iter)
 |   File "/usr/lib/python3.7/asyncio/tasks.py", line 560, in _wait_for_one
 |     return f.result()  # May raise f.exception().
 | concurrent.futures._base.CancelledError
+-> REPORT at 2021-01-08 18:53:36.744091+00:00
 | [184|PwCalculation|on_except]: Traceback (most recent call last):
 |   File "/home/pizzi/.virtualenvs/aiida-dev/lib/python3.7/site-packages/plumpy/processes.py", line 1072, in step
 |     next_state = await self._run_task(self._state.execute)
[...]

PR #4792, opened 2021-03-04
🐛 FIX: `Task.cancel` should not set state as EXCEPTED

chrisjsewell:
See https://github.com/aiidateam/aiida-core/issues/4648#issuecomment-790870664 for the diagnosis of the issue.

Currently, stopping the daemon in python 3.7 excepts all processes.
This is due to the code in `shutdown_runner`,
which cancels all asyncio tasks running on the loop,
including process continue and transport tasks.

Cancelling a task raises an `asyncio.CancellErrror`.
In python 3.8+ this exception only inherits from `BaseException`,
and so is not caught by any `except Exception` "checkpoints" in plumpy/aiida-core.
In python <= 3.7 however, the exception is equal to `concurrent.futures.CancelledError`,
and so it was caught by one of:
`Process.step`, `Running.execute` or `ProcessLauncher.handle_continue_exception`
and the process was set to an excepted state.

Ideally in the long-term, we will alter `shutdown_runner`,
to not use such a "brute-force" mechanism.
But in the short-term term this commit directly fixes the issue,
by re-raising the `asyncio.CancelledError` exception.


aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -406,7 +407,7 @@ async def execute(self) -> plumpy.process_states.State:  # type: ignore[override
             else:
                 logger.warning(f'killed CalcJob<{node.pk}> but async future was None')
             raise
-        except (plumpy.process_states.Interruption, plumpy.futures.CancelledError):
+        except (plumpy.process_states.Interruption, plumpy.futures.CancelledError, asyncio.CancelledError):
```
ltalirz, line 410: should `plumpy.futures.CancelledError` *be* `asyncio.CancelledError`?

I notice that `concurrent.futures` is used in kiwipy - where are we currently using `concurrent.futures` and where `asyncio` and what's the rationale behind this?
Also mentioning @muhrin

chrisjsewell: > should `plumpy.futures.CancelledError` be `asyncio.CancelledError`

yes I think this should be the case

> where are we currently using concurrent.futures and where asyncio and what's the rationale behind this?

@muhrin can correct me if I am wrong, but we use `concurrent.futures.Future` when we are dealing with threads, and `asyncio.Future` for standard asynchronicity.
Before python 3.8, calling the `cancel()` method of either of these generated `plumpy.futures.CancelledError`, but now they generate different exception so, as I've mentioned (https://github.com/aiidateam/aiida-core/issues/4648#issuecomment-790975477), we should check again at how these are dealt with in kiwipy

ltalirz:
After discussing with @muhrin , the source of the issue may lie in AiiDA directly using `.cancel()` rather than going through the processes built-in interruption mechanism.

Aiida uses `.cancel()` in three places:
https://github.com/aiidateam/aiida-core/blob/285ca45c41db75fbb0ed7eae5f6cdd3afd652da3/aiida/engine/daemon/runner.py#L31-L32

https://github.com/aiidateam/aiida-core/blob/b07841ad5503085913850c96aed85eb89b226d3b/aiida/engine/transports.py#L122-L123

https://github.com/aiidateam/aiida-core/blob/b07841ad5503085913850c96aed85eb89b226d3b/aiida/engine/processes/calcjobs/manager.py#L290-L291

In particular the first one seems problematic

> By looping over all the tasks in the event loop (with no knowledge or care of which process they belong to) it's really ripping the rug out from under the Process and the process should not be expected to deal with such a scenario gracefully
> from an architectural perspective I would think that shutdown_runner would rather be looping over all Processees and interrupting them, and they in turn would do any cleanup of outstanding coroutines, etc.

ltalirz:
Perhaps this should even go inside `Runner.close()`

https://github.com/aiidateam/aiida-core/blob/da872b2c488c33acf1f5c8628ac1226418cbc4de/aiida/engine/runners.py#L161-L170

It seems to me that the runners currently don't keep track of the processes they are running.
@sphuber is that correct?

chrisjsewell:
@ltalirz I think you are one step behind me here lol:

> the source of the issue may lie in AiiDA directly using .cancel() rather than going through the processes' built-in interruption mechanism. In particular the first one seems problematic.

This is what I already said in the original issues description: https://github.com/aiidateam/aiida-core/issues/4648#issuecomment-790870664

> It seems to me that the runners currently don't keep track of the processes they are running, so one would need to add this accounting. @sphuber is that correct? how would one loop over all processes that belong to a runner?

I'm already working on this, and a way to gracefully stop processes: https://github.com/aiidateam/plumpy/pull/213

ltalirz:
I see - sorry, I hadn't seen https://github.com/aiidateam/plumpy/pull/213

Do we still need the changes in plumpy then https://github.com/aiidateam/plumpy/pull/214/files ?

Hm... maybe re-raising makes sense nevertheless...

chrisjsewell:
> Do we still need the changes in plumpy then aiidateam/plumpy#214 (files) ?
> Hm... maybe re-raising makes sense nevertheless...

I think we might as well, because these changes just make sure the same thing is happening for python 3.7 and 3.8+

sphuber:
There was no explicit accounting of what processes a runner was running because once stopped it cannot be started again. Typically then the Python interpreter as a whole would shut down anyway, because the daemon worker or interactive shell was stopping and so the tasks would be requeued automatically at some point by RabbitMQ.

ltalirz:
Thanks for the explanation @sphuber !

I guess we nevertheless all agree that it would be useful if the runners could clean up after themselves, even if just to make it easier to pinpoint leaking tasks (or thinking e.g. of the test suite) and avoiding unexpected side effects of closing a runner.

As for this PR, I guess we can still proceed; the idea being that the fixes will be re-evaluated once the more controlled shutdown is in place.

aiida/manage/external/rmq.py
```diff
@@ -209,6 +210,10 @@ async def _continue(self, communicator, pid, nowait, tag=None):
             message = 'the class of the process could not be imported.'
             self.handle_continue_exception(node, exception, message)
             raise
+        except asyncio.CancelledError:  # pylint: disable=try-except-raise
```
ltalirz, line 213: Also here: while this makes python 3.7 and 3.8 behave the same, can you briefly explain why this (re-raising rather than `handle_continue_exception` is the behavior we want?

chrisjsewell: because `handle_continue_exception` sets the node (and thus process) as excepted, which is why all the processes currently end up as exceted when you stop the daemon in python 3.7

aiida/engine/transports.py
```diff
@@ -108,6 +108,10 @@ def do_open():
         try:
             transport_request.count += 1
             yield transport_request.future
+        except asyncio.CancelledError:  # pylint: disable=try-except-raise
+            # note this is only required in python<=3.7,
+            # where asyncio.CancelledError inherits from Exception
+            raise
```
chrisjsewell, line 114:
```suggestion
            raise
            _LOGGER.debug('Transport task cancelled')
```

aiida/engine/transports.py
```diff
@@ -108,6 +108,10 @@ def do_open():
         try:
             transport_request.count += 1
             yield transport_request.future
+        except asyncio.CancelledError:  # pylint: disable=try-except-raise
```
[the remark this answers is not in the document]
chrisjsewell: sure, can add it as a debug log (see below), but should not be an error otherwise currently users will see a bunch of misleading errors in the daemon log every time the daemon is stopped

aiida/engine/processes/calcjobs/tasks.py
chrisjsewell: yep no problem

chrisjsewell:
Just to note this is the commit that added this extra complexity in the stopping of the daemon worker: https://github.com/aiidateam/aiida-core/commit/281241c6cc86c1e62cabc41ede9cba862df22257
Before this the daemon workers were just immediately killed

aiida/engine/processes/calcjobs/tasks.py
```diff
@@ -406,7 +407,7 @@ async def execute(self) -> plumpy.process_states.State:  # type: ignore[override
...
             node.set_process_status(f'Transport task {command} was interrupted')
             raise
```
chrisjsewell, line 412:
```suggestion
        except (plumpy.futures.CancelledError, asyncio.CancelledError):
            node.set_process_status(f'Transport task {command} was cancelled')
            raise
        except plumpy.process_states.Interruption:
            node.set_process_status(f'Transport task {command} was interrupted')
            raise
```

aiida/manage/external/rmq.py
ltalirz: Hm... I was wondering whether this was the case but then we didn't actually see the 'failed to recreate the process instance in order to continue it' message (or did we)?

In any case, if you're right, then this is an example where this change effectively results in the cancellation of the task being ignored (because we know that only the runner shutdown causes it, and we want to ignore it).

It would be good to mention in the comments the reasons for doing this, and that - eventually - the proper solution might involve handling this exception as well.

chrisjsewell: > but then we didn't actually see the 'failed to recreate the process instance in order to continue it' message (or did we)

because it does not get to this point until you implement the other fixes I outlined

ltalirz: Right. We may still want to expand the comments of this one a bit, anyhow for me it's ok to merge.

ltalirz:
Fine to approve once the tests pass (maybe restarting will be enough here)

aiida/manage/external/rmq.py
chrisjsewell: ok cheers

chrisjsewell:
> Fine to approve once the tests pass (maybe restarting will be enough here)

yeh just the stupid pymatgen issue

Anyhow, first I will now release a new version of plumpy, with the minimal fix https://github.com/aiidateam/plumpy/pull/214 and update to that version in this PR, which should then close #4648, and unblock the 1.6 release

Then I can devote some more time to https://github.com/aiidateam/plumpy/pull/213 and a more complete mechanism for stopping processes (intertwined with this issue of not excepting processes when a daemon worker loses connection with RMQ)

ltalirz:
@chrisjsewell  By the way, do you think it would be possible to test this?
I guess one can manually create a runner, start a task and then shut the runner down.

It would be great if we had some insurance against the bug resurfacing

chrisjsewell:
> By the way, do you think it would be possible to test this?

ermm you mean this 😉: 6f041a5

ltalirz:
Looks great, thanks a lot!

Maybe you want to clean up the PR description now and have it close #4648 ?

ltalirz (approved):

chrisjsewell:
Currently, stopping the daemon in python 3.7 excepts all processes.
This is due to the code in `shutdown_runner`,
which cancels all asyncio tasks running on the loop,
including process continue and transport tasks.

Cancelling a task raises an `asyncio.CancellErrror`.
In python 3.8+ this exception only inherits from `BaseException`,
and so is not caught by any `except Exception` "checkpoints" in plumpy/aiida-core.
In python <= 3.7 however, the exception is equal to `concurrent.futures.CancelledError`,
and so it was caught by one of:
`Process.step`, `Running.execute` or `ProcessLauncher.handle_continue_exception`
and the process was set to an excepted state.

Ideally in the long-term, we will alter `shutdown_runner`,
to not use such a "brute-force" mechanism.
But in the short-term term this commit directly fixes the issue,
by re-raising the `asyncio.CancelledError` exception.

PR #4794, opened 2021-03-05
fix pymatgen imports

### LINKED ISSUE #4793: pymatgen breaks docs build again
https://readthedocs.org/projects/aiida-core/builds/13167658/

https://github.com/materialsproject/pymatgen/commit/8f097cf7ea20b8fb42635452b02f520152218394

https://github.com/materialsproject/pymatgen/blob/master/CHANGES.rst#v202134

My guess is this is going to break a *lot* of projects

All these issues would be solved by them switching from calendar versioning back to semantic versioning.
I really can't wrap my head around why you would like to use calendar versioning for a software that is used as a library in other projects.

ltalirz:
fixes #4793 

pymatgen [made a breaking change in v2021.3.4](https://github.com/materialsproject/pymatgen/blob/master/CHANGES.rst#v202134) that removed many classes
from the top level of the package.
The alternative imports were already available in previous versions,
i.e. we don't need to upgrade the pymatgen dependency.

### DIFF
--- a/aiida/orm/nodes/data/structure.py
+++ b/aiida/orm/nodes/data/structure.py
@@ -1852,7 +1852,7 @@ class StructureData(Data):
         .. note:: Requires the pymatgen module (version >= 3.0.13, usage
             of earlier versions may cause errors)
         """
-        from pymatgen import Structure
+        from pymatgen.core.structure import Structure
 
         if self.pbc != (True, True, True):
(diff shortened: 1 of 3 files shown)


ltalirz:
@chrisjsewell while pymatgen will undo the release, it is nevertheless good to update

chrisjsewell (approved):
Yeh looks good to me ta 👍

PR #4796, opened 2021-03-09
🐛 FIX: `get_pymatgen_version`

chrisjsewell:
In version 2022.0.3 it was moved (https://github.com/materialsproject/pymatgen/issues/2083#issuecomment-793141853)


ltalirz (approved):
thanks!

ltalirz:
in any case, we could now switch to pymatgen<2022 in order to avoid these issues

PR #4800, opened 2021-03-10
Docs: fix the citation links on the index page

sphuber:
The links were still using markdown syntax instead of restructured text.


ltalirz (approved):
thanks!

chrisjsewell (approved):
ta

PR #4802, opened 2021-03-10
maintenance: delete ProcessFuture

### LINKED ISSUE #4758: Delete unused `ProcessFuture`

ltalirz:
fix #4758 

The `ProcessFuture` was intended for a use case that never came to
fruition.

`ProcessFuture`s are currently not used in the AiiDA codebase and, when
tested for investigating the daemon memory leaks, seemed to be broken.


chrisjsewell:
The problem is, it is used in tests

ltalirz:
I see. Then let's keep it for the moment.

PR #4807, opened 2021-03-15
🔧 MAINTAIN: Add PyPI release workflow

chrisjsewell:
This is basically intended to make releases fool-proof and reduce the potential for manual errors and faulty releases 😉 

When you create the release, and hence tag, this workflow is triggered; checks the tag created matches the aiida package version, runs pre-commit and (some) pytests and, if they all pass, deploys to PyPi.

I'm already using a similar workflow for kiwipy (https://github.com/aiidateam/kiwipy#publishing-releases), plumpy and many other projects, although actually this is the first time I've added the tag check, which I have tested here: https://github.com/chrisjsewell/release-test/actions


chrisjsewell:
Uh-oh, some issue with apt installing postgresql-10 😬

.github/workflows/check_release_tag.py
```diff
@@ -0,0 +1,15 @@
+"""Check that the GitHub release tag matches the package version."""
```
chrisjsewell, line 1:
```suggestion
# -*- coding: utf-8 -*-
"""Check that the GitHub release tag matches the package version."""
```

.github/workflows/release.yml
```diff
@@ -1,49 +1,110 @@
...
+# Automate deployment to PyPi when creating a release tag vX.Y.Z
+# will only be published to PyPi if the git tag matches the release version
+# and the pre-commit and tests pass
+
 on:
-  release:
-    types: [published, edited]
+  push:
+    tags:
+      - "v*"
```
sphuber, line 10: Should we make this slightly more specific? what if we create a tag `virtual-machine-tutorial-2024` at some point? Maybe `v[0-9]+.[0-9]+.[0-9]+*` would work?

.github/workflows/check_release_tag.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+"""Check that the GitHub release tag matches the package version."""
+import argparse
+import json
+
+if __name__ == '__main__':
```
sphuber, line 6: I take it you use `argparse` here instead of `click` because the action that calls it doesn't have `click` installed?

.github/workflows/check_release_tag.py
```diff
@@ -0,0 +1,16 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument('GITHUB_REF', help='The GITHUB_REF environmental variable')
+    parser.add_argument('SETUP_PATH', help='Path to the setup.json')
+    args = parser.parse_args()
+    assert args.GITHUB_REF.startswith('refs/tags/v'), f'GITHUB_REF should start with "refs/tags/v": {args.GITHUB_REF}'
+    tag_version = args.GITHUB_REF[11:]
+    with open(args.SETUP_PATH) as handle:
+        data = json.load(handle)
+    pypi_version = data['version']
+    assert tag_version == pypi_version, f'{tag_version} != {pypi_version}'
```
sphuber, line 16:
```suggestion
    assert tag_version == pypi_version, f'The tag version {tag_version} != {pypi_version} specified in `setup.json`'
```

.github/workflows/release.yml
```diff
@@ -1,49 +1,110 @@
...
     steps:
     - uses: actions/checkout@v2
-
-    - name: Set up Python 3.7
+    - name: Set up Python 3.8
       uses: actions/setup-python@v2
       with:
-        python-version: 3.7
+        python-version: 3.8
+    - run: python .github/workflows/check_release_tag.py $GITHUB_REF setup.json
```
sphuber, line 24: will this relative path of `setup.json` always work?

.github/workflows/release.yml
```diff
@@ -1,49 +1,110 @@
...
+        python-version: 3.8
+    - name: Build package
       run: |
-        tx init --no-interactive
-        sphinx-intl update-txconfig-resources --pot-dir locale --transifex-project-name ${PROJECT_NAME}
-        echo $'[https://www.transifex.com]\nhostname = https://www.transifex.com\nusername = '"${USER}"$'\npassword = '"${PASSWD}"$'\n' > ~/.transifexrc
-        tx push -s
+        pip install wheel
+        python setup.py sdist bdist_wheel
+    - name: Publish ot PyPi
```
sphuber, line 106:
```suggestion
    - name: Publish to PyPI
```

.github/workflows/release.yml
```diff
@@ -1,49 +1,110 @@
...
+        pip install --upgrade pip setuptools
+        pip install -r requirements/requirements-py-3.8.txt
+        pip install --no-deps -e .
+        reentry scan
+    - name: Run sub-set of test suite
+      run: pytest -sv -k 'requires_rmq'
+
+  publish:
+
+    name: Publish to PyPi
```
sphuber, line 89:
```suggestion
    name: Publish to PyPI
```

.github/workflows/release.yml
```diff
@@ -1,49 +1,110 @@
 name: release
 
+# Automate deployment to PyPi when creating a release tag vX.Y.Z
+# will only be published to PyPi if the git tag matches the release version
```
sphuber, line 4:
```suggestion
# Automate deployment to PyPI when creating a release tag vX.Y.Z
# will only be published to PyPI if the git tag matches the release version
```

sphuber (changes requested):

.github/workflows/check_release_tag.py
chrisjsewell: yeh exactly, for something so simple I didn't want the extra complication of a click dependency

.github/workflows/release.yml
chrisjsewell: yes the checkout action set the working directory as the repo root. We already assume this with things like `run: .github/workflows/tests.sh`

.github/workflows/release.yml
chrisjsewell: sounds reasonable, I will just give it a try in my test repo...

.github/workflows/release.yml
```diff
@@ -1,49 +1,110 @@
...
+# Automate deployment to PyPI when creating a release tag vX.Y.Z
+# will only be published to PyPI if the git tag matches the release version
+# and the pre-commit and tests pass
+
 on:
-  release:
-    types: [published, edited]
+  push:
+    tags:
+      - "v*"
```
chrisjsewell, line 10:
```suggestion
      - "v[0-9]+.[0-9]+.[0-9]+*"
```

.github/workflows/release.yml
chrisjsewell: yep works fine so changed

sphuber (approved):

PR #4809, opened 2021-03-15
Dependencies: Pin sqlalchemy to minor release

### LINKED ISSUE #4808: SQLAlchemy v1.4 is breaking `verdi setup`
### Describe the bug

It seems the just released v1.4.0 of SQLAlchemy is breaking `verdi setup`.

The error seems to not be able to import `_literal_as_text` from `sqlalchemy.sql.expression`:
https://github.com/Materials-Consortia/optimade-python-tools/pull/740/checks?check_run_id=2115274484#step:10:51

[Here](https://github.com/Materials-Consortia/optimade-python-tools/pull/740/checks?check_run_id=2115274484#step:9:251) is the proof of which `sqlalchemy` version is being installed (1.4.0).

### Steps to reproduce

Try to setup an AiiDA profile using `verdi setup` with a YAML config file.
It might also be reproducible without using a config file.

### Your environment

- Operating system: GH Actions `ubuntu-latest`
- Python versions: 3.6 - 3.9 (the latest of these on GH Actions).
- aiida-core version: v1.5.2

Note, I'm expecting this to fail for `aiida-core`'s own CI as well.

csadorf:
According to https://www.sqlalchemy.org/download.html minor
releases of SqlAlchemy may have breaking changes.

Fixes #4808 .

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -35,7 +35,7 @@ dependencies:
 - reentry~=1.3
 - simplejson~=3.16
 - sqlalchemy-utils~=0.36.0
-- sqlalchemy>=1.3.10,~=1.3
+- sqlalchemy~=1.3.10
 - tabulate~=0.8.5
 - tqdm~=4.45
 - tzlocal~=2.0
--- a/setup.json
+++ b/setup.json
@@ -49,7 +49,7 @@
         "reentry~=1.3",
         "simplejson~=3.16",
         "sqlalchemy-utils~=0.36.0",
-        "sqlalchemy~=1.3,>=1.3.10",
+        "sqlalchemy~=1.3.10",
         "tabulate~=0.8.5",
         "tqdm~=4.45",
         "tzlocal~=2.0",


chrisjsewell (approved):
thanks @csadorf and @CasperWA

sphuber:
Just as some more background information, the change was actually not due to a change in the public interface. `sqlalchemy-utils` fails because it imports `_literal_as_text` which was removed from `sqlalchemy.sql.expression` in [this commit](https://github.com/sqlalchemy/sqlalchemy/commit/f07e050c9ce4afdeb9c0c136dbcc547f7e5ac7b8) of `sqlalchemy`. So really `sqlalchemy-utils` should be fixed and then we can remove the pin for `sqlalchemy`. I will create an issue for this

CasperWA:
> Just as some more background information, the change was actually not due to a change in the public interface. `sqlalchemy-utils` fails because it imports `_literal_as_text` which was removed from `sqlalchemy.sql.expression` in [this commit](https://github.com/sqlalchemy/sqlalchemy/commit/f07e050c9ce4afdeb9c0c136dbcc547f7e5ac7b8) of `sqlalchemy`. So really `sqlalchemy-utils` should be fixed and then we can remove the pin for `sqlalchemy`. I will create an issue for this

Indeed, I wanted to point this out as well. Although it is quite a pedantic/semantic point 😅

PR #4810, opened 2021-03-16
Replacing Input by click.prompt in backup utils

### LINKED ISSUE #3933: backup script contains custom prompting implementation
This custom prompting implementation (built around `input()`) could easily be replaced by `click.prompt`:
https://github.com/aiidateam/aiida-core/blob/0e17ac1d35d3e90b7277fbfe2fa0edabc9732b53/aiida/manage/backup/backup_utils.py#L19-L126

At least it seems well tested, though
https://codecov.io/gh/aiidateam/aiida-core/src/develop/aiida/manage/backup/backup_utils.py

NinadBhat:
Fixes #3933

I have implemented `click.prompt` in functions `query_string `and `query_yes_no`.

For function `ask_question`, `click.prompt` returns a type error when the string is given as an input and type is datetime.datetime. Should I still go ahead with implementing `click.prompt` in `ask_question`?


chrisjsewell:
hey @NinadBhat yes I would still go ahead and try to implement `ask_question` (taking into account that it needs to also parse datetimes)
It would also be good if you could try adding type annotations to the functions, to specify the input/output types 😄

NinadBhat:
@chrisjsewell sorry for multiple updates here. I am going through the errors now. I will update you if I am able to fix them.

sphuber:
@chrisjsewell note that this functionality will be removed entirely in `v2.0.0`. So if we don't include it in 1.6 I am not sure if it will be useful.

ltalirz:
Hey @chrisjsewell - should we nevertheless move forward with this?

sphuber:
Thanks a lot for your contribution @NinadBhat but unfortunately, this code has just been remove entirely, so we no longer need these changes. Apologies for the inconvenience.

PR #4817, opened 2021-03-16
📚 DOCS: Add internal description of the database structure.

ramirezfranciscof:
After working with the migrations I thought having something like this in the documentation might be useful.

The section for "The auxiliary tables" is still incomplete, but I don't really know very well what those tables have or why they are necessary. If someone wants to give me some pointers I could complete them now, or else I can open an issue.

I'm also unsure of what is the purpose of the metadata column in the db_dbauthinfo: for my work database it remains empty as all the properties seem to be in the auth_params column...

Easy access to the circleCI rendering of the added page [here](https://3162-77234579-gh.circle-artifacts.com/0/html/internals/database.html) (I just realized I'll have to update this link every time the PR re-runs the test...)

### DIFF
--- a/docs/source/internals/database.rst
+++ b/docs/source/internals/database.rst
@@ -1,0 +1,275 @@
+.. _internal_architecture:database:
+
+******************
+Database structure
+******************
+
+The database is the main tool that AiiDA uses to keep track of the provenance.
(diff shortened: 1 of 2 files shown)


chrisjsewell:
thanks! I'll review in the not too distant future, although ping me if I forget 😉

Issue #4526, opened 2020-10-28
Scheduler plugins: virtual and physical memory mix-up


eimrek:
Hi AiiDA developers,

In most schedulers the `metadata.options.max_memory_kb` input to a `CalcJob` specifies the physical or "real" memory limit. For example in the Slurm plugin:

https://github.com/aiidateam/aiida-core/blob/55f8706e8ecfb37745efe6bc838719b8dceed1fa/aiida/schedulers/plugins/slurm.py#L382-L395

The `#SBATCH --mem=` directive is used, which set the real (physical) memory limit (https://slurm.schedmd.com/sbatch.html). To set the virtual memory limit, one would use the `VSizeFactor` parameter in `slurm.conf` (https://slurm.schedmd.com/slurm.conf.html). The Slurm plugin implementation is fine, except for the `virtual_memory_kb` variable name.

In the Direct scheduler, however, the virtual memory limit is set with the `ulimit -v` directive:

https://github.com/aiidateam/aiida-core/blob/55f8706e8ecfb37745efe6bc838719b8dceed1fa/aiida/schedulers/plugins/direct.py#L144-L155

In addition to being inequivalent to other schedulers, setting the virtual memory limit is less convenient, as the virtual memory usage is usually much higher than physical memory usage and in the end, much of the used virtual memory might never end up in physical memory (e.g. if the program loads a large library but only uses a small part of it).

A specific example how this problem manifested is in this PR: https://github.com/aiidateam/aiida-common-workflows/pull/57 (i.e. i have to ask a high virtual memory limit for a gaussian calculation, even when the real physical memory usage will be quite low).

chrisjsewell:
thanks @eimrek, do you have any suggestions of what you'd like to see the fix be 😬

eimrek:
hey @chrisjsewell, one option would be to just ignore the `max_memory_kb` for the direct scheduler (or make it unsupported). Another option is to implement something that is proposed e.g. [here](https://unix.stackexchange.com/questions/134414/how-to-limit-the-total-resources-memory-of-a-process-and-its-children/174894). For example the `cgroups`  solution, but that might require installing additional packages and would be quite complex in the end, I think.

PR #4825, opened 2021-03-18
🐛 FIX: `max_memory_kb` usage for direct scheduler

eimrek:
Hi,

I am proposing my solution to the issue I raised a while ago #4526.

The changes are the following: 

* In all scheduler plugins I just renamed `virtual_memory_kb` to `physical_memory_kb` to avoid confusion.

* In the Direct scheduler, the `max_memory_kb` limit is ignored now and just a message is printed to logger. The justification behind this is that the previously used `ulimit -v` did impose a virtual memory limit (as opposed to a physical memory limit that all the other schedulers do) and therefore it's inequivalent (as a result, I always have to treat the direct scheduler differently in my workflows). There doesn't seem to be a straightforward standard way to limit the physical memory usage of a process in linux, so I decided just to ignore this variable.

### DIFF
--- a/aiida/schedulers/plugins/direct.py
+++ b/aiida/schedulers/plugins/direct.py
@@ -142,17 +142,8 @@ class DirectScheduler(aiida.schedulers.Scheduler):
                 lines.append('exec 2>&1')
 
         if job_tmpl.max_memory_kb:
-            try:
-                virtual_memory_kb = int(job_tmpl.max_memory_kb)
-                if virtual_memory_kb <= 0:
-                    raise ValueError
-            except ValueError:
-                raise ValueError(
-                    'max_memory_kb must be '
-                    "a positive integer (in kB)! It is instead '{}'"
-                    ''.format((job_tmpl.max_memory_kb))
-                )
-            lines.append(f'ulimit -v {virtual_memory_kb}')
+            self.logger.info('Physical memory limiting is not supported by the direct scheduler.')
+
         if not job_tmpl.import_sys_environment:
             lines.append('env --ignore-environment \\')
 
--- a/aiida/schedulers/plugins/direct.py
+++ b/aiida/schedulers/plugins/direct.py
@@ -142,7 +142,7 @@ class DirectScheduler(aiida.schedulers.Scheduler):
                 lines.append('exec 2>&1')
 
         if job_tmpl.max_memory_kb:
-            self.logger.info('Physical memory limiting is not supported by the direct scheduler.')
+            self.logger.warning('Physical memory limiting is not supported by the direct scheduler.')
 
         if not job_tmpl.import_sys_environment:
             lines.append('env --ignore-environment \\')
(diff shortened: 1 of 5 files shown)


chrisjsewell:
thanks @eimrek 😄
I haven't had any experience of setting memory limits for the direct scheduler; @sphuber, @ltalirz, @giovannipizzi any use cases you can think of why this should not be ignored?

aiida/schedulers/plugins/direct.py
```diff
@@ -142,17 +142,8 @@ def _get_submit_script_header(self, job_tmpl):
...
-                if virtual_memory_kb <= 0:
-                    raise ValueError
-            except ValueError:
-                raise ValueError(
-                    'max_memory_kb must be '
-                    "a positive integer (in kB)! It is instead '{}'"
-                    ''.format((job_tmpl.max_memory_kb))
-                )
-            lines.append(f'ulimit -v {virtual_memory_kb}')
+            self.logger.info('Physical memory limiting is not supported by the direct scheduler.')
```
ltalirz, line 145: I think this should be a warning - the user is requesting something that is not supported.

ltalirz:
thanks @eimrek 

I agree that having a stricter requirement for the direct scheduler (by limiting virtual memory as limiting physical memory [seems difficult](https://stackoverflow.com/a/6901393/1069467)) than for all other schedulers (that only limit physical memory) is not ideal.

At the same time, it might be useful to be able to limit memory consumption also for the direct scheduler.

If I understand correctly, then your PR is driven by the use case of wanting to transfer a job's memory settings between an hpc scheduler and the direct one unchanged.
Is it really a better solution, if the direct scheduler just ignores the request (rather than trying to enforce it in an approximate way)?

One could also argue for keeping the virtual-memory-based implementation for the direct scheduler; combined with an "info" log message informing the user that this is limiting virtual memory, not physical memory.

However, I haven't been affected by this issue and I can see both sides; I'll let @chrisjsewell and @giovannipizzi decide

eimrek:
Thanks for the comments @ltalirz 

True, it would be useful to be able to limit memory consumption for the direct scheduler. However, limiting virtual memory can be quite tricky and not intuitively match the real memory usage. To illustrate: when I run a small tiny test Gaussian calculation, the physical memory usage can be practically negligible (in the order of 10 MB) but the virtual memory usage is at least 1.5 GB due to large amount of different libraries available to gaussian (and most of them are not used). Additionally, if the code you're running uses a different set of libraries in another version (or perhaps is compiled differently), then the virtual memory usage will vary as well. This makes it quite difficult to limit the real memory usage through the `ulimit -v` directive in a systematic manner.

In principle, currently the implementation is not too restrictive for my use and I could just clear the `max_memory_kb` variable in my workflows for the direct scheduler and the effect would be the same for me. So if you think it's useful to have it there, we can keep it.

Another option we can consider is to define an additional input called something like `metadata.max_virtual_memory_kb` that invokes the `ulimit -v` for the direct scheduler. Could also be implemented for the HPC schedulers but limiting virtual memory in those cases is almost never done or needed afaik, so this probably is not necessary.

ltalirz (dismissed):
I agree that it's difficult to think of a use case for wanting to limit virtual memory (and I've not heard from users who need this).

I support this change.

eimrek:
Fixed another bug that caught my eye near my proposed changes. It's very tiny so I thought I could just include it with this PR. In case we don't merge this PR, I can open another one just for that.

ltalirz (approved):

chrisjsewell:
cheers

PR #4826, opened 2021-03-19
DM: Bump pillow version to 8.1.1.

csadorf:


ramirezfranciscof (approved):
Pretty straightforward it seems, thanks!

Issue #4347, opened 2020-09-02
RecursionError for instances of Orbital class


bosonie:
### Describe the bug

An instance of the `Orbital` class (`aiida.tools.data.orbital.Orbitals`) can not be printed or returned.
A `RecursionError: maximum recursion depth exceeded` is obtained.

### Steps to reproduce
In verdi shell:

      from aiida.tools.data.orbital import Orbital
      a=Orbital(position=[1.3,1,1])
      a

### Solution
I'm not sure what is the deep origin of the problem, but to define a `__str__(self)` for the `Orbital` class should fix the problem. It is in fact `a.__str__()` that fails and in the `__repr__` method of the class, `str(self)` is called.

### Your environment

- Operating system [e.g. Linux]: Linux
- Python version [e.g. 3.7.1]: 3.6.9, 3.7.6
- aiida-core version [e.g. 1.2.1]: 1.3.0

greschd:
> I'm not sure what is the deep origin of the problem, but to define a `__str__(self)` for the `Orbital` class should fix the problem. It is in fact `a.__str__()` that fails and in the `__repr__` method of the class, `str(self)` is called.

That's basically all there is to it. When no `__str__` exists, Python falls back onto `__repr__` for `str(obj)`, which leads to the recursion.

pranjalmish1:
Hey, @greschd, is it still open, can I work on this issue?

greschd:
Yes, this issue is still unresolved.

pranjalmish1:
Hey @greschd, just a basic query, how should I test this bug fix?

greschd:
I think all that's needed here is to call ``str(OrbitalData(position=[1.3,1,1]))`` - while the bug is active, that would raise.

lkbhitesh07:
Hi @greschd @bosonie, I just ran this whole in my local system and got to understand the problem but I'm confused what exactly to return in `__str__` function.
Thanks

NinadBhat:
Hi @greschd I have submitted a PR. The coverage is decreasing but I am not sure whether I should write a test for __str__ method.

greschd:
Given that it was previously failing, I think adding a simple test that calls `str(OrbitalData(...))` makes sense.

Issue #3087, opened 2019-06-26
Define purpose of __repr__, __str__ for AiiDA nodes


ltalirz:
@chrisjsewell 's graphviz PR #2596 needs to know which "short description" should be shown for a given AiiDA node.

It makes sense that this information be provided by the node itself rather than e.g. by the graph visualization tool.

### Status quo 

There are (at least) two built-in python functions used for this:

 * `__str__` and
 * `__repr__` 

Their implementations in AiiDA often differ - see below:
```ipython
In [1]: a= Int(30)

In [19]: a
Out[19]: <Int: uuid: 15f11fcf-02fa-4bbf-b3ef-ee1139f3206c (unstored) value: 30>

In [2]: print(a)
uuid: 15f11fcf-02fa-4bbf-b3ef-ee1139f3206c (unstored) value: 30

In [3]: from aiida.orm import Computer

In [4]: qb=QueryBuilder()

In [5]: qb.append(Computer)
Out[5]: <aiida.orm.querybuilder.QueryBuilder at 0x111300080>

In [6]: l=qb.all()

In [7]: l
Out[7]:
[[<Computer: localhost (localhost), pk: 1>],
 [<Computer: localhost2 (localhost), pk: 2>],
 [<Computer: daint-s888 (daint.cscs.ch), pk: 3>],
 [<Computer: fidis (fidis.epfl.ch), pk: 4>]]

In [9]: print(l[0][0])
localhost (localhost), pk: 1

In [10]: from aiida.orm import StructureData

In [11]: qb=QueryBuilder()

In [12]: qb.append(StructureData)
Out[12]: <aiida.orm.querybuilder.QueryBuilder at 0x111358ba8>

In [14]: s=qb.first()

In [17]: print(s)
uuid: da407675-294f-4e73-8a00-5ea5c37601d0 (pk: 2)

In [20]: qb=QueryBuilder()

In [21]: qb.append(User)
Out[21]: <aiida.orm.querybuilder.QueryBuilder at 0x11138c208>

In [22]: u=qb.first()

In [24]: u[0]
Out[24]: <aiida.orm.users.User at 0x111398f98>

In [25]: print(u[0])
leopold.talirz@epfl.ch
```

### Proposal

The [python docs say on `__repr__`](https://docs.python.org/3/reference/datamodel.html#object.__repr__)

> This is typically used for debugging, so it is important that the representation is information-rich and unambiguous.

[...]

PR #4829, opened 2021-03-20
Adds __str__ method to Orbital Class

NinadBhat:
The absence of __str__ was leading to recursion RecursionError  #4347. So have added a __str__ method.


greschd:
I don't know our convention for what the result of `__str__` should look like, maybe @ltalirz or @chrisjsewell can review this?

sphuber:
There isn't a well defined convention yet, other than the Python standard that `__str__` should be human readable for users and `__repr__` is meant for debugging. For node instances (where there is little difference in the two cases I would say), we have been slowly converging on a format like:
```
NodeClass<IDENTIFIER>: CONTENT
```
so for this case, we could do something like
```
Orbital<1533>: [x, y, z](theta, phi, sigma)
```
For the identifier we typically take the PK. It is less transferable than the UUID, but the UUID is often very long and not always needed. I am not too familiar with the class itself, so the `CONTENT` I proposed is really just a rough idea adding the position vector and the angles and spin orientation in a separate vector. Feel free to adapt this as needed.

aiida/tools/data/orbital/orbital.py
```diff
@@ -112,6 +112,14 @@ def __init__(self, **kwargs):
     def __repr__(self):
         return f'<{self.__class__.__name__}: {str(self)}>'
 
+    def __str__(self) -> str:
+        orb_dict = self.get_orbital_dict()
+
+        position_string = f"{orb_dict['position'][0]:.4f},{orb_dict['position'][1]:.4f},{orb_dict['position'][2]:.4f}"
+        out_string = f' orbital @ {position_string}'
+
+        return out_string
```
sphuber, line 121:
```suggestion
        return f' orbital @ {position_string}'
```

sphuber (changes requested):
Thanks @NinadBhat . I responded in a separate comment on the suggested format of the string returned by `__str__`. Let's wait to see if we can get confirmation from someone else. If they do, please change to that suggested format. Then simply add a test. As @greschd suggested, a simple test that just calls `str(Orbital())` should be sufficient.

ltalirz:
See also https://github.com/aiidateam/aiida-core/issues/3087

NinadBhat:
@ltalirz I have made the changes.

chrisjsewell:
this looks good to me, @ltalirz will merge later today unless you have any objections?

sphuber:
@chrisjsewell in my review I suggested to change the output to `Orbital<1533>: [x, y, z](theta, phi, sigma)`, or at the very least include the identifier of the node. Right now it is impossible to figure out which node is concerned

chrisjsewell:
> in my review I suggested to change the output

Ah fair, I missed that had not been addressed yet

NinadBhat:
@chrisjsewell I am not sure how to access (theta, phi, sigma) and node in the Orbital class.

greschd:
@sphuber note that this is the `Orbital` class, not `OrbitalData` - so these instances won't have a `pk` as far as I understand.

sphuber:
> @sphuber note that this is the `Orbital` class, not `OrbitalData` - so these instances won't have a `pk` as far as I understand.

Ah my bad! I completely missed this, I was indeed thinking of the `OrbitalData`. In that case maybe the current proposal is fine. @greschd as a user of the `Orbital` class, if you think the current `__str__` format is fine, let's merge it.

PR #4831, opened 2021-03-24
🐛 FIX: Add `broker_*` parameters to default test config.

### LINKED ISSUE #4830: Config created by aiida_profile fixture does not pass validation
### Describe the bug

The `aiida_profile` pytest fixture automatically creates a configuration file. However, that file is malformed w.r.t. the schema validation. Specifically, the `broker_*` keys are set to `null`.

### Steps to reproduce

Running the following test produces the error:

```python
from aiida.manage.configuration import get_config
from aiida.manage.configuration.config import Config


def test_profile(aiida_profile):
    config = Config.from_file(get_config().filepath)
```

Error message:
```
E           aiida.manage.configuration.config.ConfigValidationError: Validation Error: /tmp/tmpi8csz00a/.aiida/config.json:/profiles/test_profile/broker_protocol: None is not of type 'string'
E             schema:
E             {'description': 'Protocol for connecting to the RabbitMQ server', 'type': 'string', 'enum': ['amqp', 'amqps'], 'default': 'amqp'}
```

Full configuration file:
```json
{
    "CONFIG_VERSION": {
        "CURRENT": 5,
        "OLDEST_COMPATIBLE": 5
    },
    "profiles": {
        "test_profile": {
            "AIIDADB_BACKEND": "django",
            "AIIDADB_PORT": 50883,
            "AIIDADB_HOST": "localhost",
            "AIIDADB_NAME": "aiida_db",
            "AIIDADB_USER": "aiida",
            "AIIDADB_PASS": "aiida_pw",
            "broker_protocol": null,
            "broker_username": null,
            "broker_password": null,
            "broker_host": null,
            "broker_port": null,
            "broker_virtual_host": null,
            "AIIDADB_REPOSITORY_URI": "file:///tmp/tmp3g0uw68j/test_repo",
            "PROFILE_UUID": "6e41151e0b34449a97315a46e37ec39e",
            "options": {},
            "default_user_email": "tests@aiida.mail"
        }
    },
    "default_profile": "test_profile"
}
```

<details><summary>expand for full pytest log</summary>

```
============================= test session starts ==============================
platform linux -- Python 3.7.10, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /home/a-dogres/programming/test/aiida_pytest_test
collected 1 item

test_bug.py F                                                            [100%]

=================================== FAILURES ===================================
_________________________________ test_profile _________________________________

config = {'CONFIG_VERSION': {'CURRENT': 5, 'OLDEST_COMPATIBLE': 5}, 'default_profile': 'test_profile', 'profiles': {'test_profi...ND': 'django', 'AIIDADB_ENGINE': 'postgresql_psycopg2', 'AIIDADB_HOST': 'localhost', 'AIIDADB_NAME': 'aiida_db', ...}}}
filepath = '/tmp/tmp6s7povlb/.aiida/config.json'

    @staticmethod
    def validate(config: dict, filepath: Optional[str] = None):
        """Validate a configuration dictionary."""
        try:
>           jsonschema.validate(instance=config, schema=config_schema())

../../../.virtualenvs/aiida-pytest-check/lib/python3.7/site-packages/aiida/manage/configuration/config.py:128: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

instance = {'CONFIG_VERSION': {'CURRENT': 5, 'OLDEST_COMPATIBLE': 5}, 'default_profile': 'test_profile', 'profiles': {'test_profi...ND': 'django', 'AIIDADB_ENGINE': 'postgresql_psycopg2', 'AIIDADB_HOST': 'localhost', 'AIIDADB_NAME': 'aiida_db', ...}}}
schema = {'$schema': 'http://json-schema.org/draft-07/schema', 'definitions': {'options': {'properties': {'autofill.user.email'...ption': 'Configured profiles', 'patternProperties': {'.+': {'$ref': '#/definitions/profile'}}, 'type': 'object'}}, ...}
cls = <class 'jsonschema.validators.create.<locals>.Validator'>, args = ()
kwargs = {}
validator = <jsonschema.validators.create.<locals>.Validator object at 0x7f5237ef8950>
error = <ValidationError: "None is not of type 'string'">

    def validate(instance, schema, cls=None, *args, **kwargs):
        """
        Validate an instance under the given schema.
    
            >>> validate([2, 3, 4], {"maxItems": 2})
                ...
            ValidationError: [2, 3, 4] is too long
    
        :func:`validate` will first verify that the provided schema is
        itself valid, since not doing so can lead to less obvious error
        messages and fail in less obvious or consistent ways.
    
        If you know you have a valid schema already, especially if you
        intend to validate multiple instances with the same schema, you
        likely would prefer using the `IValidator.validate` method directly
        on a specific validator (e.g. ``Draft7Validator.validate``).
    
    
        Arguments:
    
            instance:
    
                The instance to validate
    
            schema:
    
                The schema to validate with
    
            cls (IValidator):
    
                The class that will be used to validate the instance.
    
        If the ``cls`` argument is not provided, two things will happen
        in accordance with the specification. First, if the schema has a
        :validator:`$schema` property containing a known meta-schema [#]_
        then the proper validator will be used. The specification recommends
        that all schemas contain :validator:`$schema` properties for this
        reason. If no :validator:`$schema` property is found, the default
        validator class is the latest released draft.
    
        Any other provided positional and keyword arguments will be passed
        on when instantiating the ``cls``.
    
        Raises:
    
            `jsonschema.exceptions.ValidationError` if the instance
                is invalid
    
            `jsonschema.exceptions.SchemaError` if the schema itself
                is invalid
    
        .. rubric:: Footnotes
        .. [#] known by a validator registered with
            `jsonschema.validators.validates`
        """
        if cls is None:
            cls = validator_for(schema)
    
        cls.check_schema(schema)
        validator = cls(schema, *args, **kwargs)
        error = exceptions.best_match(validator.iter_errors(instance))
        if error is not None:
>           raise error
E           jsonschema.exceptions.ValidationError: None is not of type 'string'
E           
E           Failed validating 'type' in schema['properties']['profiles']['patternProperties']['.+']['properties']['broker_protocol']:
E               {'default': 'amqp',
E                'description': 'Protocol for connecting to the RabbitMQ server',
E                'enum': ['amqp', 'amqps'],
E                'type': 'string'}
E           
E           On instance['profiles']['test_profile']['broker_protocol']:
E               None

../../../.virtualenvs/aiida-pytest-check/lib/python3.7/site-packages/jsonschema/validators.py:934: ValidationError

aiida_profile = <aiida.manage.tests.TestManager object at 0x7f5255765f50>

    def test_profile(aiida_profile):
>       config = Config.from_file(get_config().filepath)

test_bug.py:6: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.virtualenvs/aiida-pytest-check/lib/python3.7/site-packages/aiida/manage/configuration/config.py:98: in from_file
    config = Config(filepath, check_and_migrate_config(config))
../../../.virtualenvs/aiida-pytest-check/lib/python3.7/site-packages/aiida/manage/configuration/config.py:146: in __init__
    self.validate(config, filepath)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

config = {'CONFIG_VERSION': {'CURRENT': 5, 'OLDEST_COMPATIBLE': 5}, 'default_profile': 'test_profile', 'profiles': {'test_profi...ND': 'django', 'AIIDADB_ENGINE': 'postgresql_psycopg2', 'AIIDADB_HOST': 'localhost', 'AIIDADB_NAME': 'aiida_db', ...}}}
filepath = '/tmp/tmp6s7povlb/.aiida/config.json'

    @staticmethod
    def validate(config: dict, filepath: Optional[str] = None):
        """Validate a configuration dictionary."""
        try:
            jsonschema.validate(instance=config, schema=config_schema())
        except jsonschema.ValidationError as error:
            raise ConfigValidationError(
>               message=error.message, keypath=error.path, schema=error.schema, filepath=filepath
            )
E           aiida.manage.configuration.config.ConfigValidationError: Validation Error: /tmp/tmp6s7povlb/.aiida/config.json:/profiles/test_profile/broker_protocol: None is not of type 'string'
E             schema:
E             {'description': 'Protocol for connecting to the RabbitMQ server', 'type': 'string', 'enum': ['amqp', 'amqps'], 'default': 'amqp'}

../../../.virtualenvs/aiida-pytest-check/lib/python3.7/site-packages/aiida/manage/configuration/config.py:131: ConfigValidationError
------------------------------ Captured log setup ------------------------------
DEBUG    pgsu:__init__.py:141 Trying to connect via "psycopg2"...
=========================== short test summary info ============================
FAILED test_bug.py::test_profile - aiida.manage.configuration.config.ConfigVa...
============================== 1 failed in 5.34s ===============================
```

</details>

### Expected behavior

The created config file should conform to the schema, or the schema should allow `null` and fall back to the default value.

### Your environment

- Operating system: Ubuntu 18.04 on WSL2
- Python version: 3.7.10
- aiida-core version: 1.6.0

### Additional context

Encountered while running tests with `aiida-pytest`, which executes `verdi` commands while the test profile is active. These failed due to the config validation error.

Tagging @chrisjsewell @ltalirz for input on what the expected behavior should be here.

greschd:
Fixes #4830.

In the `TestProfileManager`, set the `broker_*` parameters only if they are explicitly present in the `profile_info`.

Previously, these were set to `None` by default, which caused the config validation to fail because only string values are allowed.

I think the "root cause" why this wasn't caught is because there is validation in the `Config` constructor, but not the `Profile` constructor. This is the code constructing the profile:

https://github.com/aiidateam/aiida-core/blob/75310e0bfbf8de02302de76f3e452b38e4164f33/aiida/manage/tests/__init__.py#L333-L336

Another option for fixing this issue would be explicitly setting the `broker_*` parameters in `_DEFAULT_PROFILE_INFO`: https://github.com/aiidateam/aiida-core/blob/75310e0bfbf8de02302de76f3e452b38e4164f33/aiida/manage/tests/__init__.py#L28-L42

### DIFF
--- a/aiida/manage/tests/__init__.py
+++ b/aiida/manage/tests/__init__.py
@@ -39,6 +39,12 @@ _DEFAULT_PROFILE_INFO = {
     'repo_dir': 'test_repo',
     'config_dir': '.aiida',
     'root_path': '',
+    'broker_protocol': 'amqp',
+    'broker_username': 'guest',
+    'broker_password': 'guest',
+    'broker_host': '127.0.0.1',
+    'broker_port': 5672,
+    'broker_virtual_host': ''
 }
 
 
--- a/tests/manage/tests/test_pytest_fixtures.py
+++ b/tests/manage/tests/test_pytest_fixtures.py
@@ -1,0 +1,10 @@
+# -*- coding: utf-8 -*-
+"""Tests for the AiiDA pytest fixtures."""
+
+from aiida.manage.configuration import get_config
+from aiida.manage.configuration.config import Config
+
+
+def test_profile_config(aiida_profile):  # pylint: disable=unused-argument
+    """Check that the config file created with the test profile passes validation."""
+    Config.from_file(get_config().filepath)


aiida/manage/tests/__init__.py
```diff
@@ -272,14 +272,11 @@ def profile_dictionary(self):
...
             'database_password': self.profile_info.get('database_password'),
-            'broker_protocol': self.profile_info.get('broker_protocol'),
-            'broker_username': self.profile_info.get('broker_username'),
-            'broker_password': self.profile_info.get('broker_password'),
-            'broker_host': self.profile_info.get('broker_host'),
-            'broker_port': self.profile_info.get('broker_port'),
-            'broker_virtual_host': self.profile_info.get('broker_virtual_host'),
             'repository_uri': f'file://{self.repo}',
         }
+        for key in ['broker_protocol', 'broker_username', 'broker_password', 'broker_host', 'broker_port']:
```
ltalirz, line 277: I guess currently these are the only ones regarded as optional, but one could also imagine e.g. the `database_post`  have a default value.

Why not use this loop for all keys of the dictionary?

ltalirz (commented):
thanks a lot for the fix @greschd 

just one comment

ltalirz:
> I think the "root cause" why this wasn't caught is because there is validation in the Config constructor, but not the Profile constructor. 

This is a bug, right?
At least at the point where the config with the new profiles is written to disk, we should validate.

aiida/manage/tests/__init__.py
chrisjsewell: Just to note, 
for the broker defaults, they currently come from `BROKER_DEFAULTS`: https://github.com/aiidateam/aiida-core/blob/75310e0bfbf8de02302de76f3e452b38e4164f33/aiida/manage/external/rmq.py#L41, and then the database defaults come from somewhere else.

But really I think all the defaults should derive from the schema defaults (in https://github.com/aiidateam/aiida-core/blob/develop/aiida/manage/configuration/schema/config-v5.schema.json), as happens for the options, i.e. the schema would be the "source of truth" for all defaults.

I just didn't have time to do this with the configuration refactor

chrisjsewell:
> This is a bug, right?
> At least at the point where the config with the new profiles is written to disk, we should validate.

I'm sure we discussed this in https://github.com/aiidateam/aiida-core/pull/4712 and agreed not to validate (I can't find exactly where now though)

Note whenever you alter individual configuration options or when you use `verdi setup/quicksetup` these values are already validated then.
So in principle the config in memory will always be valid, because you validate on read then validate on modification. But in this test code you go through the "non-standard" route of generating the configuration

greschd:
> This can be fixed in another PR if you prefer but I think this should be addressed.

I'm not so up-to-date on the configuration code, so yeah it would probably be easier for someone else to address this in a separate PR.

aiida/manage/tests/__init__.py
greschd: > Why not use this loop for all keys of the dictionary?

If `null` is invalid for all keys I agree that makes more sense.

Maybe getting the defaults for both the broker and the other options from the schema can be a separate PR?

ltalirz (dismissed):
thanks @greschd

greschd:
Aah, seems some parts of the test suite require those keys to be there (even if they're `None` I guess).

Maybe the simplest way to fix this then is to add the `broker_*` parameters to `_DEFAULT_PROFILE_INFO`?

greschd:
Alright, next try: Now I simply added the configuration values to `_DEFAULT_PROFILE_INFO`. 

Initially I had them extracted programmatically from the config schema, but then I noticed not all keys are the same in `profile_info` as they are in the schema (e.g., the `AIIDADB_*` keys are renamed to `database_*`). So now I'm not sure if extracting them directly is reliable, or where this key mapping comes from.

chrisjsewell:
cheers

> So now I'm not sure if extracting them directly is reliable, or where this key mapping comes from.

The key mapping comes from:

https://github.com/aiidateam/aiida-core/blob/a8ee41deb873d1bcac9f3438b6ac854773b9f986/aiida/manage/configuration/profile.py#L56

We also discussed removing this mapping in #4712 (but as with https://github.com/aiidateam/aiida-core/pull/4831#discussion_r600346388, that PR was already big enough, so didn't want to add any more complexity)

greschd:
Thanks @chrisjsewell - do you think it'd be better in this PR to hard-code things (and clean up together with further config refactoring), or go through the mapping and get these values from the schema?

I think, since there's a test checking that the test config validates, we should at least _notice_ when things are broken now.

ltalirz (approved):
thanks @greschd - sorry for the back & forth

greschd:
No problem, and thanks for the review @ltalirz!

zhubonan:
Thanks for the fix! The original bug appears to break plugin tests that involve running processes with the daemon. 
Is there any plan to have a bug fix release? This would help with CI tests since we use the last released AiiDA version. Alternatively, we have to pin the tests to a patched AiiDA version.

chrisjsewell:
yes there will be a release

espenfl:
Thanks for fixing this folks. Any ETA on the release?

greschd:
I think it's already in 1.6.1.

chrisjsewell:
Yep 👍

PR #4832, opened 2021-03-25
👌 IMPROVE: Add `account` option to `LsfScheduler`

sphuber:


ramirezfranciscof:
Mmm I know this is a bit of a PITA because most of the these would be more related to the error messages you are fixing than the actual addition of `account`, but could you add the tests for the sections that codecov is complaining about?

BTW: not sure if you requested from @chrisjsewell because you wanted him to check or test for anything specific, the PR looks rather straightforward to me, so I could approve it once the tests are there if you want.

sphuber:
Fine to be reviewed by anyone that feels confident. It isn't a difficult change but some experience with the LSF scheduler would be useful. Don't really have time to be adding tests for these exception cases though.

ramirezfranciscof:
Ah, ok, I see. In that case yes, we can wait for @chrisjsewell review. In the meantime, if you don't mind, I could try adding up the tests for these exception raises.

sphuber:
> if you don't mind, I could try adding up the tests for these exception raises.

That'd be great. Please feel free to go ahead

ramirezfranciscof:
@sphuber did you allow for modifications in your PR-branch? I can't seem to push changes to it. I'm trying the following:

```bash
(aiida-full) workuser@apolos-Air aiida-core % git status

On branch pr4832_lsf
nothing to commit, working tree clean

(aiida-full) workuser@apolos-Air aiida-core % git remote -vv

fork	https://ramirezfranciscof@github.com/ramirezfranciscof/aiida-core.git (fetch)
fork	https://ramirezfranciscof@github.com/ramirezfranciscof/aiida-core.git (push)
sphuber	https://ramirezfranciscof@github.com/sphuber/aiida-core (fetch)
sphuber	https://ramirezfranciscof@github.com/sphuber/aiida-core (push)

(aiida-full) workuser@apolos-Air aiida-core % git push --set-upstream sphuber:fix/lsf-scheduler-add-account-option

ssh: Could not resolve hostname sphuber: nodename nor servname provided, or not known
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

aiida/schedulers/plugins/lsf.py
```diff
@@ -119,13 +119,13 @@ def __init__(self, **kwargs):
 
         try:
             self.parallel_env = str(kwargs.pop('parallel_env', ''))
-        except (TypeError, ValueError):
-            raise TypeError("When specified, 'parallel_env' must be a string")
+        except (TypeError, ValueError) as exc:
+            raise TypeError("When specified, 'parallel_env' must be a string") from exc
```
ramirezfranciscof, line 123: Are we catching anything here? I can't find anything that can't be casted into `str` (see [this](https://stackoverflow.com/questions/38380335/potential-exceptions-using-builtin-str-type-in-python) for example) and the `pop` has a default so...?

aiida/schedulers/plugins/lsf.py
```diff
@@ -698,7 +701,7 @@ def _parse_time_string(self, string, fmt='%b %d %H:%M'):
                 thetime = datetime.datetime.strptime(actual_string, f'{actual_fmt} L')
         except Exception as exc:
             self.logger.debug(f'Unable to parse time string {string}, the message was {exc}')
-            raise ValueError('Problem parsing the time string.')
+            raise ValueError(f'Problem parsing the time string: `{string}`') from exc
```
ramirezfranciscof, line 704: I took the liberty of adding a bit of information for this error message (although the information is being logged, I guess there is no harm in also including it here). I enclosed in inverted commas so that it would be more clear what exactly is the string (also added this for the other error on lines [678/679](https://github.com/aiidateam/aiida-core/pull/4832/files#diff-f3bdb5ef823a005d31025d9d0b851a50587737040d0208810bc3d6eaf2b806b5R678-R679)).

Let me know if you don't like it @sphuber, I can take it out.

ramirezfranciscof (commented):
Sorry, I thought that since I was adding tests and the testsuite was not very complex I would migrate the format to pytest, but apparently this complicated a bit github's understanding of the changes.

If it helps, I think you can just ignore `test_parse_common_joblist_output`, or just verify that the only thing I did was taking it out of the class (idem for `test_submit_output` and `test_kill_output`).

aiida/schedulers/plugins/lsf.py
sphuber: I think you might be right. The default in `get` should guarantee a value. If your reference is correct that `str()` cannot ever raise, then we would not need to catch any exception. Still I am not a 100% sure and if we missed something and introduce a bug that be kind of annoying

tests/schedulers/test_lsf.py
```diff
@@ -32,162 +32,206 @@
...
+
+        def __str__(self):
+            raise TypeError('This type cannot be casted to str')
+
+    with pytest.raises(TypeError):
+        job_tmpl.job_resource = scheduler.create_job_resource(
+            tot_num_mpiprocs=2,
+            parallel_env=StrCastRaiser(),
+        )
+
```
ramirezfranciscof, line 170: I think this is my [hackest hack](https://www.youtube.com/watch?v=uOa-ObWPAKg) so far...

aiida/schedulers/plugins/lsf.py
ramirezfranciscof: So I added a hacky test for this, but thinking a bit more about it...

Should we really be doing implicit validation through this try-catch? Or should do a more explicit input validation for `parallel_env` that raises if anything other than a string is provided? Does it make sense to pass anything else that would get casted, what is this variable for?

sphuber: given the error message, it would make sense to just have something like:
```
if not isinstance(kwargs.pop('parallel_env', ''), str):
    raise TypeError(...)
```
As it stands, the try-catch indeed makes very little sense. I think it would be fine to make that change and then you don't need such a convoluted test setup

ramirezfranciscof: Ok, change done and test streamlined.

But I'll have you know that test was pure coding art =P

sphuber:
Thanks a lot @ramirezfranciscof . For me this would be good to merge

chrisjsewell (approved):
Yep all good in the hood 👍

chrisjsewell:
grr @aiidateam/dependency-manager, pip has suddenly decided that it can't resolve dependencies for `pip install .[all]`: https://github.com/aiidateam/aiida-core/pull/4832/checks?check_run_id=2286189589

csadorf:
> grr @aiidateam/dependency-manager, pip has suddenly decided that it can't resolve dependencies for `pip install .[all]`: https://github.com/aiidateam/aiida-core/pull/4832/checks?check_run_id=2286189589

Looks like it might just be timing out? 😕

chrisjsewell:
> Looks like it might just be timing out? 😕

No thats because it can't work out compatible dependencies, it just goes on forever downloading exponential numbers of package versions. If you can workout what package has changed since it was working, and pin that, then this behaviour will disappear

csadorf:
> > Looks like it might just be timing out? 😕
> 
> Yeh thats because it can't work out compatible dependencies, it just goes on forever downloading exponential numbers of package versions. If you can workout what package has changed since it was working, and pin that, then this behaviour will disappear

I just ran an analysis and found that the following packages had releases since yesterday:

 - django
 - sqlalchemy
 - tqdm
 - pre-commit

The following packages had releases in the past 7 days:
 - django
 - sqlalchemy
 - tqdm
 - docutils
 - pre-commit
 - pg8000
 - pytest

chrisjsewell:
thanks @csadorf, at a guess I reckon it could be this: https://github.com/tlocke/pg8000/commit/16b08f32611c87c51ff1f2955a8057650dcce7bc

chrisjsewell:
nope no joy

csadorf:
> nope no joy

I tried to constrain tqdm, pre-commit, pg8000, and pytest, but no luck so far.

chrisjsewell:
> I've ran pipenv against the requirements with all extras and found the following conflict

nice cheers, out of interest how long did this take to identify the conflict?

csadorf:
> > I've ran pipenv against the requirements with all extras and found the following conflict
> 
> nice cheers, out of interest how long did this take to identify the conflict?

Just a few minutes. Let's continue discussion relating to this particular problem in the associated issue: https://github.com/aiidateam/aiida-core/issues/4850

chrisjsewell (approved):

Issue #4067, opened 2020-05-12
Realising visualization in Grafana using data from AiiDA


espenfl:
Grafana is very popular and it would be great if we could support it. They have the generic JSON plugin, which I think might be the way to go:
https://grafana.com/grafana/plugins/simpod-json-datasource

Notice that:

```
To work with this datasource the backend needs to implement 4 urls:

    / should return 200 ok. Used for "Test connection" on the datasource config page.
    /search should return available metrics when invoked.
    /query should return metrics based on input.
    /annotations should return annotations.

Those two urls are optional:

    /tag-keys should return tag keys for ad hoc filters.
    /tag-values should return tag values for ad hoc filters.
```

Have someone looked at this before? Does it make sense to extend the REST API that is already there, or make a dedicated one?

The additional code we put into AiiDA should also unpack any AiiDA objects so that Grafana does not need to have any knowledge of it.

espenfl:
As a start I think it would be great to do the `ArrayData`.

espenfl:
A general use case would be to use Grafana as a plotting tool. Grafana typically relies on database entries. Instead of making a script which fetched AiiDA objects on the postgresql database, decodes it and puts it back in in some other database so that it can be easily fetched by Grafana it would be nice to be able to do this directly. Two options exists as far as I know; write a custom Grafana plugin (would need to be AiiDA aware to decode) and using the general json plugin which returns fixed data structures according to the specifications in the link above.

giovannipizzi:
Isn't grafana mostly for monitoring rather that looking at the data?
In any case, if you have a minimal working proof of concept that we can look at, I'm happy to discuss with you so we see how to integrate it in AiiDA!

espenfl:
Well, Grafana is great for live data, but is also useful as a visualizer, especially if the dataset evolve over time, which it typically is doing for high throughput. A working proof of concept would be how we can make https://grafana.com/grafana/plugins/simpod-json-datasource work with a request to get the content of a `ArrayData` through the AiiDA API.

PR #4835, opened 2021-03-29
Proof of Concept: Grafana REST-API

ramirezfranciscof:
This is in principle just a proof of concept of how we could communicate with [Grafana](https://grafana.com/) through the REST-API, as suggested in issue #4067. This first toy demo just shows the number of parent workflows currently running, and the number of mpi processors that are being requested and used by the currently running calcjobs. This is a display screenshot:

test

# Steps to reproduce

1. **Sign up:** go to Grafana and [open a free account](https://grafana.com/signup/cloud/connect-account?pg=login). You will have to create an organization when you do this*. Once in the Grafana Cloud Portal, click in the Grafana tile.
2. **Install the plugin:** go to [this page](https://grafana.com/grafana/plugins/simpod-json-datasource/) and you should see the option to install the plugin in blue on the left.
3. **Local setup:** Checkout this branch and then run `verdi restapi --posting`.
4. **Configure source:** Go back to the Grafana Home and look in the left side options for the "Configuration". Click on "Data Sources" and then click on "add data source". Look and "select" the JSON tile. The only thing you need to set is the `URL` to `http://127.0.0.1:5000/api/v4/grafana` and the `Access` to `Browser`. Then just choose a name and save it.
5. **Setup dashboard:** In the Grafana Home look in the left side options for the "Dashboards" and click on "Manage". Then click on "new dashboard", it should already had a panel deployed in which you can click on "Add an empty panel". In the "Query" tab on the bottom region, there is a selection box that currently shows `--Grafana--`: click there and choose the name of your recently selected source. There should be now another couple of boxes below, one of which says `metric`: select `current_resource_usage`. Finally, on the "Visualization" section on the right, select the "Gauge". Click on "Apply".

Now you can run a couple of your workchains and see if the gauges track correctly your resources. NOTE: Please let me know if you are having trouble with any of the steps or if some instructions are unclear, since I'm reconstructing this procedure after doing it myself through trial and error I'm not entirely sure these will be identical for a clean slate account...

*I am not really sure how the accounts and organizations work and depend on one another. If someone wants to check, I could try adding them to my organization to see how that works. Let me know via PM.

# Follow up

I'll leave this draft open for a little while; in the meantime I'll try thinking of a more interesting application of the displays and collect some feedback (ideas from all are welcome). I'll make a couple of specific pings:

- @giovannipizzi if you think this could be an interesting direction for REST-API development in AiiDA.
- @espenfl this was originally your idea, what do you think of this? what kind of information would you like to see displayed?
- @ltalirz and @CasperWA : so you are more than welcome to check this out too, but in principle I'm just pinging you so you are aware of this possible development since it also requires a POST method in the REST-API just as your recent querybuilder interface. This is currently a rather dirty implementation, so if we go forward I'll probably need some pointers on how to properly extend the REST-API from you.

### DIFF
--- a/aiida/restapi/api.py
+++ b/aiida/restapi/api.py
@@ -92,7 +92,7 @@ class AiidaApi(Api):
 
         from aiida.restapi.common.config import CLI_DEFAULTS
         from aiida.restapi.resources import (
-            ProcessNode, CalcJobNode, Computer, User, Group, Node, ServerInfo, QueryBuilder
+            ProcessNode, CalcJobNode, Computer, User, Group, Node, ServerInfo, QueryBuilder, Grafana
         )
 
         self.app = app
@@ -116,6 +116,17 @@ class AiidaApi(Api):
                 QueryBuilder,
                 '/querybuilder/',
                 endpoint='querybuilder',
+                strict_slashes=False,
+                resource_class_kwargs=kwargs,
+            )
+
+            self.add_resource(
+                Grafana,
+                '/grafana/',
+                '/grafana/search/',
+                '/grafana/query/',
+                '/grafana/annotations/',
+                endpoint='grafana',
                 strict_slashes=False,
                 resource_class_kwargs=kwargs,
             )
--- a/aiida/restapi/common/utils.py
+++ b/aiida/restapi/common/utils.py
@@ -420,8 +420,8 @@ class Utils:
 
         ## Type checks
         # mandatory parameters
-        if not isinstance(data, dict):
-            raise InputValidationError('data must be a dictionary')
+        if not isinstance(data, (dict, list)):
+            raise InputValidationError('data must be a dictionary or a list')
 
         # non-mandatory parameters
         if status is not None:
--- a/aiida/restapi/common/utils.py
+++ b/aiida/restapi/common/utils.py
@@ -209,7 +209,7 @@ class Utils:
 
         if path[0] in [
             'projectable_properties', 'statistics', 'full_types', 'full_types_count', 'download', 'download_formats',
-            'report', 'status', 'input_files', 'output_files'
+            'report', 'status', 'input_files', 'output_files', 'search', 'query', 'annotations'
         ]:
             query_type = path.pop(0)
             if path:
--- a/aiida/restapi/grafana.py
+++ b/aiida/restapi/grafana.py
@@ -1,0 +1,161 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+""" Resources for using grafana with the REST API """
+from datetime import datetime
+
+
+def deduce_mpiproc(data_dict):
+    """Description"""
+    data_resources = data_dict['resources']
+    manual_calculation = ('num_mpiprocs_per_machine' in data_resources) and ('num_machines' in data_resources)
+    ready_calculation = 'tot_num_mpiprocs' in data_resources
+
+    mpiproc = 0
+
+    if manual_calculation:
+        np_per_node = data_resources['num_mpiprocs_per_machine']
+        total_nodes = data_resources['num_machines']
+        mpiproc = total_nodes * np_per_node
+
+    if ready_calculation:
+        mpiproc = data_resources['tot_num_mpiprocs']
+
+    if 'last_job_info' in data_dict:
+        mpiproc = data_dict['last_job_info']['num_mpiprocs']
+
+    return mpiproc
+
+
+def process_grafana_request(entry_point, request_dict):
+    """Description"""
+    # pylint: disable=too-many-locals,too-many-branches,too-many-statements
+    from aiida import orm
+
+    if entry_point == 'search':
+        #data = ['resource_usage', 'current_resource_usage']
+        data = ['current_resource_usage']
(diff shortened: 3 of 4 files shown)


espenfl:
@ramirezfranciscof That is one nice Easter surprise. Thanks a lot for approaching this. I will try this out as documented and also take a look at how you did it code wise. What I want is also to be able to fetch data that sits in `ArrayData` and be able to plot that. That requires some way to fetch it and return it in a sensible manner to Grafana. If we could also couple the pk's somehow, so that Grafana can fetch some kind of id's of `ArrayData` and then load incoming ones, that would be super sweet.

ramirezfranciscof:
@espenfl Haha, I'm glad you like it and are still interested in this. From what I understood grafana only works with time series, right? So are we talking about an `ArrayData` of dimensions (2,n)?

giovannipizzi:
Looks very interesting! Rather than showing the instantaneous number, can one show the value as a function of time?
Indeed, the main goal of Grafana (as fas as I understand) is monitoring of resources, so indeed showing things like the following (vs. time) might be useful:
- statistics on workflows and more general calculations and processes not in a final state
- % of "free slots" still available (it's computed to show a warning at the end of `verdi process list`, and printed only if close to the limit, so you can check there)
- statistics on status of calcjobs on the remote computer (maybe more complex)
- statistics on CPU and memory usage of the workers (this is printed by `verdi daemon status` so one should be able to get it programatically)
- number of nodes (maybe of various types: data, calcs, workflows, ...) in the DB

For showing data (as Espen suggests) I would indeed first try to understand which data we want to show, and how much this is something general that can be added to AiiDA, and how much it should be implemented as a "plugin" (e.g. if it's only valid for very specific types of ArrayData). Let's see the usecase of Espen!

Once we have a few relevant examples where this is useful, it would be great to clean up the code.
Finally: does one need to create an account, or can grafana also run locally?

ramirezfranciscof:
> Looks very interesting! Rather than showing the instantaneous number, can one show the value as a function of time?

Thanks! I initially tried doing something like this but it was not obvious how to get information from resource usage from the data stored inside the nodes (you can still see the remnants of that in the code, I was actually using the wrong parameters at that point, like the creation time of the node). Then I found that the place were we typically have the data for this is the `last_job_info` attribute, which I seem to remember was scheduler specific, and probably the format has changed at least a couple of times during the `v0.x` migrations, no? I could try to at least do a demo for the format that I get from daint in my DB.

Also there might be some details to polish with all kinds of resource reporting, such as for example: this request for 36 cpus ends up taking 72 in daint (which we still don't know if is just daint way of reporting that the full node is occupied, considering multithreading even if it is not using it, or what exactly is going on...)

```
In [2]: calcjob_node.attributes
Out[2]: 
{(...)
 'version': {'core': '1.6.0', 'plugin': '3.3.0'},
 'withmpi': True,
 'resources': {'num_machines': 1, 'tot_num_mpiprocs': 1},
 (...)
 'last_job_info': {
  (...)
  'num_machines': 1,
  'num_mpiprocs': 72,
  'dispatch_time': {'date': '2021-03-30T20:03:11.000000', 'timezone': None},
  'submission_time': {'date': '2021-03-30T19:59:37.000000', 'timezone': None},
  'wallclock_time_seconds': 181,
  'requested_wallclock_time_seconds': 14400},
 (...)
 'mpirun_extra_params': ['--hint=nomultithread'],
 (...)
 'custom_scheduler_commands': '#SBATCH --constraint=mc\n#SBATCH --cpus-per-task=36'}
``` 

> Finally: does one need to create an account, or can grafana also run locally?

This is a good point, I initially went with Grafana Cloud because that is what seems to be the most direct advertised thing in their website, but there also seems to be [a local installation](https://grafana.com/docs/grafana/latest/installation/), and apparently the code does not only have a commercial ("Enterprise") version, but also [an open source one](https://github.com/grafana/)! I will probably still keep using the cloud version for my test, but we could dig more into this later.

greschd:
I've used Grafana before (although didn't set it up myself), so here are some observations:

> Rather than showing the instantaneous number, can one show the value as a function of time?

I think there are existing tools that convert instantaneous data into time series. This seems like the right documentation page: https://grafana.com/docs/grafana/latest/getting-started/timeseries/
From memory, I think we were using `collectd` (for more traditional CPU / memory / etc. resource tracking) - I could look up the details if needed.

One really neat thing you can do is add "alerts" (also configured directly on Grafana) that can go to Slack / e-mail / whatever when some resource exceeds a given criterion.

> the code does not only have a commercial ("Enterprise") version, but also an open source one

Can confirm, we used the self-hosted open-source version.

giovannipizzi:
Thanks!
@ramirezfranciscof - what I would investigate is indeed (as @greschd says) how to extract 'instantaneous' data from AiiDA, and let Grafana (or some other tool) collect the time series.
So I wouldn't start by looking into last_job_info, but looking at the instantaneous use of processes, resources etc., and see if either grafana itself or some other tool (see @greschd link) can take care of storing the series

chrisjsewell:
> what I would investigate is indeed (as @greschd says) how to extract 'instantaneous' data from AiiDA, and let Grafana (or some other tool) collect the time series.
So I wouldn't start by looking into last_job_info, but looking at the instantaneous use of processes, resources etc., and see if either grafana itself or some other tool (see @greschd link) can take care of storing the series

this should all be an AEP

sphuber:
@ramirezfranciscof would it be ok to close this PR and to proceed in an AEP further working out the proposed interface and use cases?

ramirezfranciscof:
Yeah ok, anyway I think the changes proposed would mean doing something quite different so ok to close this.

giovannipizzi:
Just as a comment, indeed this could go in an AEP (for the design) or be implemented in a plugin as a first step. I suggest to discuss with @Crivella who recently used Grafana+prometheus, to brainstorm on other possible ideas

PR #4839, opened 2021-03-30
docs: update ssh proxycommand instructions

ltalirz:
It was not obvious how to handle cases where the SSH key for the proxy
server needs to be specified.

Also, remove the note regarding redirection.
This was relevant for the old approach of using netcat directly and is
more confusing than helpful without it.


ramirezfranciscof (approved):
Seems good, thanks!

PR #4855, opened 2021-04-13
Dependencies: update `sqlalchemy` dependency requirements

### LINKED ISSUE #4813: Support SQLAlchemy v1.4
The sqlalchemy dependency is constrained to v1.3 (#4809 ) as v1.4 is incompatible (see #4808 ). The sqlalchemy project follows a [modified semver approach](https://www.sqlalchemy.org/download.html), where minor releases can have major breaking changes.

sphuber:
Fixes #4813 

In `767f57a02b1ba40421ab9b17a67b2532aff556c6` the `sqlalchemy`
dependency was pinned to the `v1.3` minor version, because the recent
`v1.4.0` release would break `sqlalchemy-utils`. The latter has now
released `v0.37.0` which is compatible with `sqlalchemy==1.4`. In theory
this would allows us to support `v1.4` as well, except our own code
breaks with this version since we access quite a few protected
attributes in the query builder implementation.

Since `sqlalchemy-utils` does not follow semver, we have to pin it to
the `0.37.x` series for now. Since version `0.37.0` and `0.37.1` were
yanked as they did not specify the `python_requires` keyword, we set the
lower boundary to `0.37.2`.

The constructor of the `DbLog` class was simplified. Its only purpose is
to make sure that the `_metadata` column is properly initialized. The
column has an underscore prepended because otherwise it would clash with
an attribute of the SqlAlchemy base class, however, the constructor
takes just `metadata` so this has to be transferred. In addition, it
should be initialized to an empty dict if it is `None`. All other
initialization code that was in the constructor will be taken care of by
simply calling the parent class' constructor.

Finally, the `force_instant_defaults` listener of `sqlalchemy-utils` was
updated in v0.37.5 which caused certain tests to fail. The fix ensured
that the `kwargs` passed into the constructor of a model instance would
be updated with the defaults of the column definitions that are set on
the instance by the listener. The problem is that some of our models,
e.g. `DbComment`, rely no distinguishing between the column value being
explicitly passed by the caller or through default. Most notably this is
for the `mtime` which our implementation wants to have undefined before
storing, unless explicitly passed in the constructor. Due to the change
in `sqlalchemy-utils` this distinction could no longer be made and so
the only solution was to copy the original implementation of the
listener utility.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -34,7 +34,7 @@ dependencies:
 - pyyaml~=5.1
 - reentry~=1.3
 - simplejson~=3.16
-- sqlalchemy-utils~=0.36.0
+- sqlalchemy-utils~=0.37.2
 - sqlalchemy~=1.3.10
 - tabulate~=0.8.5
 - tqdm~=4.45
--- a/setup.json
+++ b/setup.json
@@ -48,7 +48,7 @@
         "pyyaml~=5.1",
         "reentry~=1.3",
         "simplejson~=3.16",
-        "sqlalchemy-utils~=0.36.0",
+        "sqlalchemy-utils~=0.37.2",
         "sqlalchemy~=1.3.10",
         "tabulate~=0.8.5",
         "tqdm~=4.45",
--- a/aiida/backends/sqlalchemy/models/__init__.py
+++ b/aiida/backends/sqlalchemy/models/__init__.py
@@ -8,13 +8,30 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Module to define the database models for the SqlAlchemy backend."""
-
-from sqlalchemy_utils import force_instant_defaults
+import sqlalchemy as sa
 
 # SqlAlchemy does not set default values for table columns upon construction of a new instance, but will only do so
 # when storing the instance. Any attributes that do not have a value but have a defined default, will be populated with
 # this default. This does mean however, that before the instance is stored, these attributes are undefined, for example
 # the UUID of a new instance. In Django this behavior is the opposite and more in intuitive because when one creates for
 # example a `Node` instance in memory, it will already have a UUID. The following function call will force SqlAlchemy to
-# behave the same as Django and set model attribute defaults upon instantiation.
-force_instant_defaults()
+# behave the same as Django and set model attribute defaults upon instantiation. Note that this functionality used to be
+# provided by the ``sqlalchemy_utils.force_instant_defaults`` utility function. However, this function's behavior was
+# changed in v0.37.5, where the ``sqlalchemy_utils.listeners.instant_defaults_listener`` was changed to update the
+# original ``kwargs`` passed to the constructor, with the default values from the column definitions. This broke the
+# constructor of certain of our database models, e.g. `DbComment`, which needs to distinguish between the value of the
+# ``mtime`` column being defined by the caller as opposed to the default. This is why we revert this change by copying
+# the old implementation of the listener.
+
+
+def instant_defaults_listener(target, _, __):
+    """Loop over the columns of the target model instance and populate defaults."""
+    for key, column in sa.inspect(target.__class__).columns.items():
+        if hasattr(column, 'default') and column.default is not None:
+            if callable(column.default.arg):
+                setattr(target, key, column.default.arg(target))
+            else:
+                setattr(target, key, column.default.arg)
+
+
+sa.event.listen(sa.orm.mapper, 'init', instant_defaults_listener)
--- a/aiida/backends/sqlalchemy/models/comment.py
+++ b/aiida/backends/sqlalchemy/models/comment.py
@@ -49,7 +49,7 @@ class DbComment(Base):
         """Adding mtime attribute if not present."""
         super().__init__(*args, **kwargs)
         # The behavior of an unstored Comment instance should be that all its attributes should be initialized in
-        # accordance with the defaults specified on the collums, i.e. if a default is specified for the `uuid` column,
+        # accordance with the defaults specified on the columns, i.e. if a default is specified for the `uuid` column,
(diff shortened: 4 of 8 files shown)


environment.yml
```diff
@@ -34,8 +34,8 @@ dependencies:
 - pyyaml~=5.1.2
 - reentry~=1.3
 - simplejson~=3.16
-- sqlalchemy-utils~=0.36.0
-- sqlalchemy~=1.3.10
+- sqlalchemy-utils~=0.37.0
```
csadorf, line 37: It appears that version 0.3.7.0 has not been released for conda-forge yet: https://anaconda.org/conda-forge/sqlalchemy-utils

https://github.com/conda-forge/sqlalchemy-utils-feedstock

csadorf, line 37: I'm afraid we probably have to pin sqlalchemy-utils since the project does not follow any specified versioning scheme (https://github.com/kvesteri/sqlalchemy-utils/issues/438).
```suggestion
- sqlalchemy-utils==0.37.0
```

setup.json
```diff
@@ -48,8 +48,8 @@
         "pyyaml~=5.1.2",
         "reentry~=1.3",
         "simplejson~=3.16",
-        "sqlalchemy-utils~=0.36.0",
-        "sqlalchemy~=1.3.10",
+        "sqlalchemy-utils~=0.37.0",
```
csadorf, line 51:
```suggestion
        "sqlalchemy-utils==0.37.0",
```
See previous comment.

environment.yml
```diff
@@ -34,8 +34,8 @@ dependencies:
...
+- sqlalchemy~=1.3
```
csadorf, line 38:
```suggestion
- sqlalchemy~=1.3.0|~=1.4.0
```
We have to constrain sqlachemy to specific minor versions since the project follows a [modified semantic versioning scheme](https://www.sqlalchemy.org/download.html#versions) where breaking changes can be introduced with new minor versions.

setup.json
```diff
@@ -48,8 +48,8 @@
...
+        "sqlalchemy~=1.3",
```
csadorf, line 52:
```suggestion
        "sqlalchemy~=1.3.0,~=1.4.0",
```
See previous comment.

csadorf (changes requested):
Thanks for the update! I've made some change requests, but this PR is also blocked by a missing release of sqlalchemy-utils version 0.37.0 on [conda-forge](https://anaconda.org/conda-forge/sqlalchemy-utils).

sphuber:
Adding support for `sqlalchemy>=1.4` will also require work on our side. The `QueryBuilder` implementation accesses protected attributes that have been removed and so the tests fail. I had a quick look already but it doesn't seem link a simple fix. In particular it is failing in `QueryBuilder._build` where we are accessing `sqlalchemy.Query._entities` which has been removed. It seems that it is somewhat replaced by `sqlalchemy.Query._raw_columns` but simply replacing that doesn't work seemingly because `_raw_colums` seems to convert the entities being added. Anyway, this will take more work then I can dedicate to this now, so this will be on hold until someone can fix the incompatibilities and use the public API of `sqlalchemy`

csadorf:
> Adding support for `sqlalchemy>=1.4` will also require work on our side. The `QueryBuilder` implementation accesses protected attributes that have been removed and so the tests fail. I had a quick look already but it doesn't seem link a simple fix. In particular it is failing in `QueryBuilder._build` where we are accessing `sqlalchemy.Query._entities` which has been removed. It seems that it is somewhat replaced by `sqlalchemy.Query._raw_columns` but simply replacing that doesn't work seemingly because `_raw_colums` seems to convert the entities being added. Anyway, this will take more work then I can dedicate to this now, so this will be on hold until someone can fix the incompatibilities and use the public API of `sqlalchemy`

Then let's keep the dependency to `sqlalchemy~=1.3.10` here and loose the constraint in a separate PR. Then we can merge this PR once the sqlalchemy-utils conda package is released.

csadorf:
Blocked by: https://github.com/conda-forge/sqlalchemy-utils-feedstock/pull/46

csadorf (approved):
Changes approved, merge pending the publication of the sqlalchemy-utils conda package.

sphuber:
As it turns out, the tests are failing even with just changing `sqlalchemy-utils` so this will require more work.

csadorf:
> As it turns out, the tests are failing even with just changing `sqlalchemy-utils` so this will require more work.

@sphuber Could you comment on the status of this?

sphuber:
@csadorf I have updated the branch. If the tests pass we can merge this and like you say do the update to `sqlalchemy>=1.4` in a separate PR. That will definitely be more involved and I don't have the time for that.

environment.yml
csadorf: The version has been released on conda-forge in the meantime.

csadorf (dismissed):

csadorf:
For future reference, the support of SQLAlchemy v1.4 is tracked in: https://github.com/aiidateam/aiida-core/issues/4813 .

sphuber:
Please don't merge this yet, #5027 needs to be merged first.

sphuber:
Seems that even with just the update of `sqlalchemy-utils` the tests fail...

sphuber:
@csadorf no point in updating the branch. The code needs fixing before this can be merged.

sphuber:
Ok @csadorf , it took quite some work, but I think things are ready now. That'll teach me to think that simply updating a dependency requirement will be quick and painless... As you can see there were quite some additional changes necessary, to keep the old behavior. I searched through the code base and the actual remaining uses of the `sqlalchemy-utils` library are very few. Maybe it would be time to consider dropping it altogether.

csadorf (approved):

csadorf:
@sphuber Please go-ahead with the merge.

PR #4858, opened 2021-04-16
👌 IMPROVE: Check for recycled circus PID

dev-zero:


ramirezfranciscof:
Hey @dev-zero , thanks for the contribution! It seems quite straightforward code-wise, but could you explain a bit the exact objective of this and what kind of problem is addressing in general (maybe you already did this in some issue, in which case, could you link it?)

I guess most of the functionality can be deduced from the change itself but just to be sure intent aligns with execution. 😅

dev-zero:
@ramirezfranciscof in that part of the code we are looking up whether there's a process for the recorded PID. If there is not we can assume the daemon crashed and we have to restart. If there is we have to be a bit more careful since the PID can be recycled (maybe after a restart of the machine, but also on long-running machines). If a process is found we have to see whether the process is in fact a circus process, and if it is a circus process whether it is started by the same user requesting the start. The last one is a real corner case because it would mean that another user on the same machine is running an aiida daemon and by chance got the same PID as this users aiida daemon before.
Besides, one case is still not caught: a user running multiple aiida daemons from different virtualenvs/conda environments.

aiida/cmdline/utils/daemon.py
```diff
@@ -126,8 +126,13 @@ class StartCircusNotFound(Exception):
     if pid is not None:
         try:
             process = psutil.Process(pid)
+            # the PID got recycled, but a different process
             if _START_CIRCUS_COMMAND not in process.cmdline():
                 raise StartCircusNotFound()  # Also this is a case in which the process is not there anymore
+
+            # the PID got recycled, but for a daemon of someone else
+            if process.username() != psutil.Process().username():  # compare against the username of this interpreter
+                raise StartCircusNotFound()
```
unkcpz, line 135: a small piece of advice,  since the exception includes a new case the name should be more generalized(maybe `IncorrectCircusProcess`?). As well as the class docstring of Exception in the few lines above (121). @dev-zero

unkcpz (changes requested):

aiida/cmdline/utils/daemon.py
dev-zero: Not really: the `StartCircusNotFound` is raised when the previous circus process could not be found. This is the same thing in the newly added check: a process with the same PID and `cmdline` was found, but for a different user, hence it's not the previously remembered circus process, hence the conclusion is still "not found in the ps command". In other terms: just the conditions under which a process is considered to be the previously remembered one are tightened a bit.

unkcpz: I see, thanks @dev-zero. Anyway, this is a useful fix, I remember I encountered the case where I have to delete staled circus pid file manually.

unkcpz (dismissed):
@dev-zero thanks, it looks good to me. 

I have another thought about this `delete_stale_pid_file` check. Since it always sticks right after the client creation (except in cmd_daemon::stop show following, but I guess it's fine to move the line backwards),  maybe it is safe to put this check inside `get_daemon_client`? (discussion required, not required for this PR though)

:https://github.com/aiidateam/aiida-core/blob/53c5564529eb636cc376e4b9535fd59c756d5b0f/aiida/cmdline/commands/cmd_daemon.py#L191-L195

dev-zero:
@chrisjsewell can this be merged despite the coverage fail?

PR #4860, opened 2021-04-16
CI: install default `postgresql` instead of specific version

mbercx:
The CI configuration uses `ubuntu-latest` which recently became Ubuntu
Focal Fossa (20.04) which no longer has `postgresql-10` available, but
instead provides `postgres-12`. Here we switch to using the default
`postgresql`, which is line with our installation guide and works with both
versions of Ubuntu.

### DIFF
--- a/.github/workflows/ci-code.yml
+++ b/.github/workflows/ci-code.yml
@@ -86,7 +86,7 @@ jobs:
     - name: Install system dependencies
       run: |
         sudo apt update
-        sudo apt install postgresql-12 graphviz
+        sudo apt install postgresql graphviz
 
     - name: Upgrade pip and setuptools
       # It is crucial to update `setuptools` or the installation of `pymatgen` can break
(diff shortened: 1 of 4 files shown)


mbercx:
Just noticed that my tests were failing in #4761:

https://github.com/aiidateam/aiida-core/pull/4761/checks?check_run_id=2362353306#step:5:63

Had the same issue for `aiida-quantumespresso`, see:

https://github.com/aiidateam/aiida-quantumespresso/pull/674

greschd:
Why don't we just remove the version and install ``postgresql`` instead?

csadorf:
@mbercx I like @greschd ' suggestion, can you try that?

csadorf (changes requested):
Generally, I'd suggest that we fix all dependencies in our CI workflows, but in this case I'm inclined to follow @greschd ' suggestion to install the default version. At the very least we should do that for the `test-install` workflow which should as closely as possible resemble our installation instructions on the docs.

mbercx:
Thanks for the comments @greschd and @csadorf! Have made the suggested adjusment 👍

csadorf (approved):

Issue #4623, opened 2020-12-08
`BaseRestartWorkChain` does not return `output_namespaces` of `_process_class`


bosonie:
I'm not sure whether to consider this as a bug or a feature request.

The `BaseRestartWorkChain`  contains a step `results` that is meant to return the outputs specified in the output specification using the outputs the of last completed process.
The line that does the job is:
https://github.com/aiidateam/aiida-core/blob/e073972ba6b0b83d7d7964750a762da26603fa74/aiida/engine/processes/workchains/restart.py#L277-L280
This mechanism works well for normal outputs, but it ignores `output_namespaces`. In fact the line above checks that the "name" of the output port is equal to any of the return outputs of the last completed process. Supposing the name of the output namespaces is "name", the return outputs of the last completed process will be "name__key" and therefore not detected.

JPchico:
@bosonie do you have a small code where one can reproduce this error?

I have an idea on how to fix it but would be nice to have a small example on this as I do not currently have any `BaseRestartWorkChain` to test this.

bosonie:
No I don't. I can try to build it later. I'll update you tonight

bosonie:
```
from aiida.engine import BaseRestartWorkChain
from aiida.engine import while_, run
from aiida.workflows.arithmetic.multiply_add import MultiplyAddWorkChain

class MyMul(MultiplyAddWorkChain): 
    @classmethod 
    def define(cls, spec): 
             super().define(spec) 
             spec.output_namespace('inpp', valid_type=Int, dynamic=True, required=False) 
         def result(self): 
             inpp={} 
             inpp["test"] = self.ctx.addition.outputs.sum 
             self.out('inpp', inpp) 
             self.out('result', self.ctx.addition.outputs.sum)

class Test(BaseRestartWorkChain): 

         _process_class = MyMul 
         _proc_exit_cod = _process_class.exit_codes 
         
         @classmethod 
         def define(cls, spec): 
             super().define(spec) 
             spec.expose_inputs(MyMul) 
             spec.outline( 
                 cls.setup, 
                 cls.prepare_inputs, 
                 while_(cls.should_run_process)( 
                     cls.run_process, 
                     cls.inspect_process, 
                 ), 
                 cls.results 
             ) 
             spec.expose_outputs(MyMul) 
              
         def prepare_inputs(self): 
[...]

PR #4863, opened 2021-04-19
FIX: respect nested output namespaces in `Process.exposed_outputs`

### LINKED ISSUE #3533: `Process.exposed_outputs` does not work with nested namespaces
Simple test case to demonstrate the problem:
```
class ChildProcess(Process):

    _node_class = orm.WorkflowNode

    @classmethod
    def define(cls, spec):
        super(ChildProcess, cls).define(spec)
        spec.input('input', valid_type=orm.Int)
        spec.output('output', valid_type=orm.Int)
        spec.output('name.space', valid_type=orm.Int)

    @override
    def run(self):
        self.out('output', self.inputs.input)
        self.out('name.space', self.inputs.input)


class ParentProcess(Process):

    _node_class = orm.WorkflowNode

    @classmethod
    def define(cls, spec):
        super(ParentProcess, cls).define(spec)
        spec.input('input', valid_type=orm.Int)
        spec.expose_outputs(ChildProcess)

    @override
    def run(self):
        runner = get_manager().create_runner()
        results, node = runner.run_get_node(ChildProcess, **self.inputs)
        self.out_many(self.exposed_outputs(node, ChildProcess))


class TestProcessNamespace(AiidaTestCase):

    def test_namespaced_process(self):
        results, node = run_get_node(ParentProcess, input=orm.Int(5))
        print(results)
```
The results will only contain `output` and not the `name.space` output.

sphuber:
Fixes #3533 

The `exposed_outputs` method was detecting nested namespaces in the
dictionary of outputs of a node by using the namespace separator
character of the port namespace class. However, this character, which is
the `.`, gets converted to `__` when the link is stored in the database.
This transformation is necessary since the `.` is a reserved character
to enable attribute based dereferencing, e.g.

    node.outputs.some_output

Since link labels can contain underscores, we use a double underscore
which, together with the guarantee that link labels don't have leading
or terminating underscores, can uniquely determine the namespaces in any
given link label.

The solution is to not manually reconstruct the namespaces in the output
dictionary but simply use the `get_outgoing().nested()` method which
does it for us.

### DIFF
--- a/aiida/engine/processes/process.py
+++ b/aiida/engine/processes/process.py
@@ -883,7 +883,7 @@ class Process(plumpy.processes.Process):
         # maps the exposed name to all outputs that belong to it
         top_namespace_map = collections.defaultdict(list)
         link_types = (LinkType.CREATE, LinkType.RETURN)
-        process_outputs_dict = {entry.link_label: entry.node for entry in node.get_outgoing(link_type=link_types)}
+        process_outputs_dict = node.get_outgoing(link_type=link_types).nested()
 
         for port_name in process_outputs_dict:
             top_namespace = port_name.split(namespace_separator)[0]
--- a/tests/engine/test_process.py
+++ b/tests/engine/test_process.py
@@ -7,7 +7,7 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-# pylint: disable=no-member
+# pylint: disable=no-member,too-many-public-methods
 """Module to test AiiDA processes."""
 import threading
 
@@ -343,3 +343,52 @@ class TestProcess(AiidaTestCase):
         self.assertTrue(node.is_finished)
         self.assertTrue(node.is_finished_ok)
         self.assertEqual(node.exit_status, 0)
+
+    def test_exposed_outputs(self):
+        """Test the ``Process.exposed_outputs`` method."""
+        from aiida.common import AttributeDict
+        from aiida.common.links import LinkType
+        from aiida.engine.utils import instantiate_process
+        from aiida.manage.manager import get_manager
+
+        runner = get_manager().get_runner()
+
+        class ChildProcess(Process):
+            """Dummy process with normal output and output namespace."""
+
+            _node_class = orm.WorkflowNode
+
+            @classmethod
+            def define(cls, spec):
+                super(ChildProcess, cls).define(spec)
+                spec.input('input', valid_type=orm.Int)
+                spec.output('output', valid_type=orm.Int)
+                spec.output('name.space', valid_type=orm.Int)
+
+        class ParentProcess(Process):
+            """Dummy process that exposes the outputs of ``ChildProcess``."""
+
+            _node_class = orm.WorkflowNode
+
+            @classmethod
+            def define(cls, spec):
+                super(ParentProcess, cls).define(spec)
+                spec.input('input', valid_type=orm.Int)
+                spec.expose_outputs(ChildProcess)
+
+        node_child = orm.WorkflowNode().store()
+        node_output = orm.Int(1).store()
+        node_output.add_incoming(node_child, link_label='output', link_type=LinkType.RETURN)
+        node_name_space = orm.Int(1).store()
+        node_name_space.add_incoming(node_child, link_label='name__space', link_type=LinkType.RETURN)
+
+        process = instantiate_process(runner, ParentProcess, input=orm.Int(1))
+        exposed_outputs = process.exposed_outputs(node_child, ChildProcess)
+
+        expected = AttributeDict({
+            'name': {
+                'space': node_name_space,
+            },
+            'output': node_output,
+        })
+        self.assertEqual(exposed_outputs, expected)
--- a/tests/engine/test_work_chain.py
+++ b/tests/engine/test_work_chain.py
@@ -9,9 +9,8 @@
 ###########################################################################
 # pylint: disable=too-many-lines,missing-function-docstring,invalid-name,missing-class-docstring,no-self-use
 """Tests for the `WorkChain` class."""
-import inspect
-import unittest
 import asyncio
+import inspect
 
 import plumpy
 import pytest
@@ -1303,7 +1302,6 @@ class TestWorkChainExpose(AiidaTestCase):
             }
         )
 
-    @unittest.skip('Functionality of `Process.exposed_outputs` is broken for nested namespaces, see issue #3533.')
     def test_nested_expose(self):
         res = launch.run(
             GrandParentExposeWorkChain,


JPchico:
> Does this also address #4623 ? I would think it should, or at least be intimately related. It was mentioned by @bosonie in issue #3533 but I've seen no comment regarding that.

I do not think it does, since I think that `node.get_outgoing(link_label_filter=name)` would still fail to get the proper `name` as it would pick just the top level label instead of having something like `namespace__port` which is the kind of label that one would have to pass to `link_label_filter`.

bosonie:
> > Does this also address #4623 ? I would think it should, or at least be intimately related. It was mentioned by @bosonie in issue #3533 but I've seen no comment regarding that.
> 
> I do not think it does, since I think that `node.get_outgoing(link_label_filter=name)` would still fail to get the proper `name` as it would pick just the top level label instead of having something like `namespace__port` which is the kind of label that one would have to pass to `link_label_filter`.

Correct

sphuber:
@ramirezfranciscof could you review this today. This needs to be released in 1.6.2 a.s.a.p as it is blocking some users.

sphuber:
Thanks for the quick review @ramirezfranciscof 

> BTW: I'm a bit curious about having tests for this in test_process.py and test_work_chain.py. It would seem that the main difference is that the first calls the methods directly (tests that the methods work), whereas the second goes through the engine running system (i.e. tests that these methods are actually called during run).

Well, the `exposed_outputs` is defined on the `Process` class and so I added the tests in the corresponding unit test. That being said, exposing outputs really only makes sense for workflow type processes because typically you want to expose the ports of a subprocess that you want to *call*, which is only something that workflows can do. Calculation processes therefore don't really have a use. But that is a different question really. I think that unit tests for a particular method should be done in the file of the respective class. If we move the method to another (sub)class, then the unit tests move as well.

ramirezfranciscof:
> Well, the `exposed_outputs` is defined on the `Process` class and so I added the tests in the corresponding unit test. That being said, exposing outputs really only makes sense for workflow type processes because typically you want to expose the ports of a subprocess that you want to _call_, which is only something that workflows can do. Calculation processes therefore don't really have a use. But that is a different question really. I think that unit tests for a particular method should be done in the file of the respective class. If we move the method to another (sub)class, then the unit tests move as well.

Mmm ok, I think we are saying a similar thing. Let me reformulate maybe: I agree with the following points...
- Since `exposed_outputs` is defined on the `Process` class then the tests for the methods should be on `test_process.py` (I'm not sure why it has to be defined there instead of in the workchain subclass, but anyways I'm ok with that, it was not the issue).
- It is a good idea to test that the engine is calling this method and setting the right thing when we `run` or `submit`. And since it only makes sense to do this with a workchain, this should be what gets built.

My point was more questioning if "integration" test should be on `test_work_chain.py`. Now that I take a second look, I see that all (or at least most) of the tests in `test_work_chain.py` are done through `run` calls to the engine, which is the same way we seem to be testing all specific processes (`test_calc_job.py`, `test_calcfunction.py`, `test_workfunction.py`). 

Now, I think I still have the question of why don't we test more directly (without going through the engine) the methods of specific subclasses of `Process`. And also perhaps if we should make it more clear where should the test for direct methods go, and where should the "integration" ones that go through the engine go.

sphuber:
I don't think that there should be a distinction whether integration tests go into `test_process` or `test_work_chain`. They both should contain unit tests. Now, writing pure unit tests for these complex classes are tricky and often would require a lot of mocking. We are simply using the full execution to do the mocking for us. You can argue that this makes it an integration test, but I don't think it matters too much. I think the idea should be that whenever reasonably possible we should attempt to isolate the direct interface that we are testing and just mock the required state, but in the case of processes, this can be so difficult that simply running them may be a viable solution as well.

mbercx:
While trying to add a note on tab-completion for the `outputs` for the tutorial, I ran into the following deprecation warning:

```python
In [1]: cj_node = load_node(311)
/opt/conda/lib/python3.7/site-packages/aiida/orm/utils/managers.py:98: AiidaDeprecationWarning: dereferencing nodes with links containing double underscores is deprecated, simply replace the double underscores with a single dot instead. For example:
`self.inputs.some__label` can be written as `self.inputs.some.label` instead.
Support for double underscores will be removed in the future.
  'Support for double underscores will be removed in the future.', AiidaDeprecationWarning
In [2]: cj_node.outputs. <TAB>
```

The deprecation warning prints when I try to tab-complete the outputs. @ramirezfranciscof guided me to this PR. I didn't look into much detail what is responsible, but if I revert back to `aiida-core==1.6.1`, before this change was released, I no longer have this problem.

sphuber:
This is not the corresponding PR, but #4625 is. I am bit surprised that the deprecation warning triggers already at tab completion though. It *should* only trigger when actually retrieving a node through a label containing double underscores. So I am confused how this is happening. Still, does the tab-completion still work, despite the warning?

mbercx:
> This is not the corresponding PR, but #4625 is. I am bit surprised that the deprecation warning triggers already at tab completion though. It _should_ only trigger when actually retrieving a node through a label containing double underscores. So I am confused how this is happening. Still, does the tab-completion still work, despite the warning?

Yes, but it prints every time the tab-completion is attempted. 😅

unkcpz:
If I understand it correctly, ipython tab-autocompletion read attributes from `__dir__` method, which in our `NodeLinksManader` it is created from `_get_keys`. @mbercx While in your `cj_node.outputs`, the keys include double-underscore attributes cause `__getattr__` go through it and raise the warning. 

Could you show the  `cj_node.outputs.__dir__()`, I believe there will be at least one underscore key correspond to the attributes. If it is the case, the keys with underscore need to be excluded explicitly from `_get_keys` method.

unkcpz:
By setting a traceback right before the deprecate waring I got the following information:

```
  File "/home/unkcpz/miniconda3/envs/aiida-core-dev/lib/python3.8/site-packages/jedi/cache.py", line 110, in wrapper
    return dct[key]
KeyError: ((), frozenset())  

  File "/home/unkcpz/Projects/python-code/aiida_core/aiida/orm/utils/managers.py", line 88, in _get_node_by_link_label
    node = attribute_dict[label]                                                                                        
KeyError: '__wrapped__'
```

I have no idea where this `__wrapped__` key come from, but escape it solve the issue.

sphuber:
The `__wrapped__` dunder method is added when a function is wrapped by a decorator and this returns the original wrapped function (see [Python docs](https://docs.python.org/3/library/functools.html)). The wrapping seems to come from this `jedi` cache that is present. I have no idea where this is coming from though and who is asking to cache this stuff. The problem is though that in our code, we warn whenever the label contains a double underscore, which is the case for dunder methods, but we should ignore those. I  split this issue off to #5010 and am submitting a PR to fix it.

PR #4865, opened 2021-04-21
Fix `aiida.cmdline.utils.decorators.with_dbenv` always loading the db

### LINKED ISSUE #4864: The `aiida.cmdline.utils.decorators.with_dbenv` loads database environment even if it is already loaded
It calls the function `load_backend_if_not_loaded` but the name is a lie as it will always load the backend, regardless of its state.

sphuber:
Fixes #4864 

The premise of the method `with_dbenv` is that it will load the profile
and corresponding database environment if not already loaded. However,
due to a bug in `load_backend_if_not_loaded` it was actually *always*
loading the environment for every call, even if the database env was
already loaded, leading to very expensive calls for nothing.

### DIFF
--- a/aiida/cmdline/utils/decorators.py
+++ b/aiida/cmdline/utils/decorators.py
@@ -34,11 +34,13 @@ __all__ = ('with_dbenv', 'dbenv', 'only_if_daemon_running')
 
 def load_backend_if_not_loaded():
     """Load the current profile if necessary while running the spinner to show command hasn't crashed."""
-    from aiida.manage.configuration import load_profile
+    from aiida.manage.configuration import load_profile, PROFILE
     from aiida.manage.manager import get_manager
-    with spinner():
-        load_profile()
-        get_manager().get_backend()
+
+    if PROFILE is None:
+        with spinner():
+            load_profile()
+            get_manager().get_backend()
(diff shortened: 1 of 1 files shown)


sphuber:
@ramirezfranciscof also this would be good to quickly merge and release with 1.6.2

ramirezfranciscof:
@sphuber is there a way to add a quick test that uses this method twice and somehow checks that there was a double load? It would be nice to have, but I do recognize that it is not clear to me how it could check for that...

sphuber:
> @sphuber is there a way to add a quick test that uses this method twice and somehow checks that there was not a double load? It would be nice to have, but I do recognize that it is not clear to me how it could check for that...

It could be done: we would have to mock the database loading env by a function that counts the number of times it is called with a global and raises the second time. But I didn't have the time to implement it and don't think it was necessary for this change. Doubt that this will accidentally be reverted at some point.

ramirezfranciscof:
Ok, I'll just open an issue then and we can just merge it as is (maybe if its quick and we don't have other problems I can even take care of it during the coding days after merging the new repository).

ramirezfranciscof (approved):
Good to go!

PR #4871, opened 2021-04-26
engine: implement sub-dicts for context in workchains

### LINKED ISSUE #4972: Add support for implicit namespacing in the keys for `WorkChain.to_context`

dev-zero:
Fixes #4972

### DIFF
--- a/aiida/engine/processes/workchains/workchain.py
+++ b/aiida/engine/processes/workchains/workchain.py
@@ -11,7 +11,7 @@
 import collections.abc
 import functools
 import logging
-from typing import Any, List, Optional, Sequence, Union, TYPE_CHECKING
+from typing import Any, List, Optional, Sequence, Union, Tuple, TYPE_CHECKING
 
 from plumpy.persistence import auto_persist
 from plumpy.process_states import Wait, Continue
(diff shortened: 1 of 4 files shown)


dev-zero:
@sphuber as discussed on Slack. Btw, there is still something fishy in the exception handling: I tried to add a test which triggers one of the `AssertionError`, but then the test always hang.

dev-zero:
also, I checked the codebase for other erroneous uses of `assert`, but didn't find any so far.

aiida/engine/processes/workchains/workchain.py
```diff
@@ -121,24 +121,53 @@ def on_run(self):
...
+        :param key: A key into the context, where words before a dot are interpreted as a key for a sub-dictionary
+        """
+        ctx = self.ctx
+        ctx_path = key.split('.')
+
+        for path in ctx_path[:-1]:
+            try:
+                ctx = ctx[path]
+            except TypeError:
+                raise AssertionError(
```
sphuber, line 138: Shouldn't this really be a `ValueError`? This is raised when the `key` is inconsistent with requested action and `key` is a user supplied value from within the `WorkChain` definition.

aiida/engine/processes/workchains/workchain.py
```diff
@@ -121,24 +121,53 @@ def on_run(self):
...
+                    f'The sub-dictionary part `{path}` of the given nested key path `{key}` is already set to a list'
```
sphuber, line 139: The part `path` here will contain just the immediate part of the full path that causes problems. It would be good if we could provide the full sub path. This will prevent ambiguity if you have a path with sub paths that have the same name, for example `some.name.name.space`.

sphuber, line 139: `is already set to a list` might be true if you are trying to assign a value to a key that is currently a list, but this would also trigger if the sub path already has a value assigned to it, wouldn't it? Let's say the path `some.path` has a value assigned to it, then trying to assign a value to `some.path.namespace` will also raise this `TypeError` and have a misleading error message. Would probably be useful to distinguish these cases individually.

aiida/engine/processes/workchains/workchain.py
```diff
@@ -121,24 +121,53 @@ def on_run(self):
...
+            ctx, (Awaitable, list)
+        ), f'The sub-dictionary path `{".".join(ctx_path[:-1])}` points to an Awaitable or a Sequence'
+
+        return ctx, ctx_path[-1]
+
     def insert_awaitable(self, awaitable: Awaitable) -> None:
         """Insert an awaitable that should be terminated before before continuing to the next step.
 
         :param awaitable: the thing to await
         :type awaitable: :class:`aiida.engine.processes.workchains.awaitable.Awaitable`
```
sphuber, line 156:
```suggestion
```

aiida/engine/processes/workchains/workchain.py
```diff
@@ -121,24 +121,53 @@ def on_run(self):
...
+                    f'The sub-dictionary part `{path}` of the given nested key path `{key}` is already set to a list'
+                )
+            except KeyError:
+                ctx[path] = AttributeDict()  # create the sub-dict and update the context
+                ctx = ctx[path]
+
+        # Awaitables are also AttributeDicts, but we must NEVER attach other Awaitables to them
+        assert not isinstance(
+            ctx, (Awaitable, list)
+        ), f'The sub-dictionary path `{".".join(ctx_path[:-1])}` points to an Awaitable or a Sequence'
```
sphuber, line 148: Just like the other `AssertionError`, I think this should be a `ValueError` because it would be thrown in case of an inconsistent user provided error. A minor-nitpick on the side: you check for `list` but say `Sequence` in the help string, probably good to align these two.

tests/engine/test_work_chain.py
```diff
@@ -780,6 +780,182 @@ def result(self):
...
+        class Workchain(WorkChain):
+
+            @classmethod
+            def define(cls, spec):
+                super().define(spec)
+                spec.outline(cls.begin, cls.result)
+
+            def begin(self):
+                self.to_context(**{'sub1.workchains': append_(self.submit(SimpleWc1))})
+                return ToContext(**{'sub1.workchains': append_(self.submit(SimpleWc2))})
```
sphuber, line 853: Any particular reason you are using two different subworkchain classes that do the same? Using the same `SimpleWc` work chain should achieve the same test coverage, would it not?

sphuber (changes requested):
Thanks a lot @dev-zero . Overall I fully support the feature and implementation, just some minor comments. Also, it would be important to document this feature. I would suggest adding this to this section: https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/workflows/usage.html?highlight=append_#to-context

aiida/engine/processes/workchains/workchain.py
dev-zero: Agreed, I was looking at the surround/previous code where `assert ...` were used so I followed that guideline. A `KeyError` would be the other logical choice.

aiida/engine/processes/workchains/workchain.py
dev-zero: Agreed, will fix.

dev-zero: Actually there are not too many options here if I understood it correctly.
But you're probably right that we have to move the `isinstance(Awaitable, ...)` check here, otherwise we might create a subdict inside an `Awaitable` (what I prevented below for the the leave-node, but may happen inside the loop already).
But I'll take another look.

aiida/engine/processes/workchains/workchain.py
dev-zero: Sure, will do. I'll go toward `Sequence` I guess since that's the more general one.

tests/engine/test_work_chain.py
dev-zero: I wanted to be able to distinguish by value that the order was preserved.
Don't remember why anymore, though.

sphuber: This behavior of `append_` is already tested elsewhere though and so wouldn't really be necessary here, where you are just testing the `ToContext` and `to_context` interface. Still, maybe doesn't hurt, just adds extra lines and makes it a bit more cumbersome to read, but we can keep it this way.

aiida/engine/processes/workchains/workchain.py
dev-zero: Wait, `{key}` is that info, but I will reformat the message to follow "problem: reason" to make it more clear, including `{key}` in the "problem" part.

sphuber: Sure `key` will show that `some.name.name.space` is a problematic key but it will say that it is because

> The sub-dictionary part `name` ... is already set to a list

The point is that then you don't know which `name` part is the problematic one, the first or the second. If you give the full path, i.e., `some.name` or `some.name.name`, the error is unambiguous

dev-zero: ugh, okay, got it: once an Awaitable is resolved it can be many different values.

aiida/engine/processes/workchains/workchain.py
dev-zero: ok, the complete thing needed to change, done.

aiida/engine/processes/workchains/workchain.py
dev-zero: done.

tests/engine/test_work_chain.py
dev-zero: ok :)

dev-zero:
> Thanks a lot @dev-zero . Overall I fully support the feature and implementation, just some minor comments. Also, it would be important to document this feature. I would suggest adding this to this section: https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/workflows/usage.html?highlight=append_#to-context

done

sphuber (approved):

sphuber:
Thanks a lot @dev-zero

PR #4882, opened 2021-04-28
Rename `name` column of `Computer` database table to `label`

### LINKED ISSUE #2926: Rename `name` column of `Computer` to `label`

sphuber:
Fixes #2926 

All other database models already use `label` so using the same for the
`Computer` model is more consistent. The `name` attribute was already
deprecated in the ORM in `aiida-core==1.0.0` and replaced by `label`.
Here we apply the actual database migration to also perform the change
in the database models.


sphuber:
I started the deprecation path for this change in `v1.0.0` by adding `label` in our front-end ORM and raising deprecation warnings when `name` was used. However, we could not provide a deprecation path for all uses. Most notably, when using the `QueryBuilder` one had to continue to use `name` when specifying it in the filters or projections. The same goes for the REST API, where `name` is still being used. With this PR, the database tables are updated to use `label` and the final remnants of `name` are removed. This does mean, however, that this will break code that manually uses the `QueryBuilder` to query for computers and filter or project on `name`, so this could be an argument to not merge these changes.

However, on the other hand, currently there is an inconsistency where in the ORM one has to use `label` whereas in the REST API and QB one should use `name`. By merging these changes, we get rid of this inconsistency.

sphuber:
@ramirezfranciscof @chrisjsewell this is now ready for review

aiida/restapi/translator/computer.py
```diff
@@ -75,9 +75,9 @@ def get_projectable_properties(self):
                 'type': 'int',
                 'is_display': False
             },
-            'name': {
-                'display_name': 'Name',
-                'help_text': 'Name of the object',
+            'label': {
+                'display_name': 'Label',
```
ltalirz, line 79: to be fully consistent, one might argue that `display_name` should become `display_label`.
but I'm ok with keeping this inconsistency

ltalirz (approved):
thanks @sphuber !

I went through the files and the changes all look fine.
It's a pity that we we could not do a deprecation for the querybuilder and the REST API, so some people might run into issues, but this change needs to happen and it should happen in 2.0.

I notice that the file `tests/static/graphs/graph1.aiida` was deleted - was this just no longer needed?

aiida/restapi/translator/computer.py
sphuber: > I notice that the file `tests/static/graphs/graph1.aiida` was deleted - was this just no longer needed?

It wasn't deleted, it was just migrated since it needed to also perform the `name` -> `label` change in the archive. Due to the new archive format that is already there, the size was just reduced.

ltalirz (approved):

PR #4883, opened 2021-04-28
`verdi status`: add version and repository UUID to the output

### LINKED ISSUE #2522: `verdi status`: file repo + version information
* When the file repository does not exist, AiiDA currently creates a new folder automatically and without any warning. This makes the current "file repository status" rather meaningless, since it only checks whether the folder exists. I would suggest to provide a warning, when AiiDA has to create it from scratch and to add _some_ info to the status, e.g. whether the folder is empty or not (open to suggestions)
* could consider adding version information that helps as reports on mailing lists (e.g. aiida version, postgres version, rmq version)

sphuber:
Fixes #2522 

The new disk objectstore-based repository provides a UUID which is also
written in the database to which it is coupled. We add it here to the
output of `verdi status` as additional information.

In addition, we add the version of the package to the output. This way,
in troubleshooting, users do not have to run `verdi --version`
separately to get this information.

### DIFF
--- a/aiida/cmdline/commands/cmd_status.py
+++ b/aiida/cmdline/commands/cmd_status.py
@@ -55,6 +55,7 @@ STATUS_SYMBOLS = {
 def verdi_status(print_traceback, no_rmq):
     """Print status of AiiDA services."""
     # pylint: disable=broad-except,too-many-statements,too-many-branches
+    from aiida import __version__
     from aiida.cmdline.utils.daemon import get_daemon_status, delete_stale_pid_file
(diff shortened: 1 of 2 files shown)


ltalirz (commented):
thanks for the improvement - fully agree with this change!

should we check in the tests that the information shows up as expected?

sphuber:
> should we check in the tests that the information shows up as expected?

Added

ltalirz (approved):
thanks @sphuber !

PR #4888, opened 2021-04-29
ORM: replace `InputValidationError` with `ValueError` and `TypeError`

### LINKED ISSUE #3812: ORM: replace `InputValidationError` with `ValueError` or `TypeError`
The `InputValidationError` was intended for use in `CalcJob.prepare_for_submission` but is abused in the ORM code to throw when arguments have an invalid value. This should just be communicated through a `ValueError`.

This will be backwards incompatible and so needs to go in `v2.0.0`

sphuber:
Fixes #3812 

The front-end ORM code threw an `InputValidationError` in a number a
places. However, this exception was intended for calculation plugins to
throw in the case of invalid input nodes. Where it was being used in the
ORM plain `ValueError` or `TypeError` exceptions should have been used.

Since the `InputValidationError` does not subclass either of the base
exception types, code that was catching this particular AiiDA-specific
exception will break after these changes. However, a scan of the most
used plugins revealed that this exception is actually not being caught
so this change should have little to no actual consequences.

### DIFF
--- a/aiida/cmdline/commands/cmd_code.py
+++ b/aiida/cmdline/commands/cmd_code.py
@@ -19,7 +19,6 @@ from aiida.cmdline.params import options, arguments
 from aiida.cmdline.params.options.commands import code as options_code
 from aiida.cmdline.utils import echo
 from aiida.cmdline.utils.decorators import with_dbenv
-from aiida.common.exceptions import InputValidationError
 
 
 @verdi.group('code')
@@ -76,7 +75,6 @@ def set_code_builder(ctx, param, value):
 @with_dbenv()
 def setup_code(non_interactive, **kwargs):
     """Setup a new code."""
-    from aiida.common.exceptions import ValidationError
     from aiida.orm.utils.builders.code import CodeBuilder
 
     if kwargs.pop('on_computer'):
@@ -88,15 +86,15 @@ def setup_code(non_interactive, **kwargs):
 
     try:
         code = code_builder.new()
-    except InputValidationError as exception:
+    except ValueError as exception:
         echo.echo_critical(f'invalid inputs: {exception}')
 
     try:
         code.store()
-        code.reveal()
-    except ValidationError as exception:
+    except Exception as exception:  # pylint: disable=broad-except
         echo.echo_critical(f'Unable to store the Code: {exception}')
 
+    code.reveal()
     echo.echo_success(f'Code<{code.pk}> {code.full_label} created')
 
 
@@ -244,7 +242,7 @@ def relabel(code, label):
 
     try:
         code.relabel(label)
-    except InputValidationError as exception:
+    except (TypeError, ValueError) as exception:
         echo.echo_critical(f'invalid code label: {exception}')
     else:
         echo.echo_success(f'Code<{code.pk}> relabeled from {old_label} to {code.full_label}')
--- a/aiida/orm/implementation/querybuilder.py
+++ b/aiida/orm/implementation/querybuilder.py
@@ -18,14 +18,12 @@ likely be moved to a `SqlAlchemyBasedQueryBuilder` class and restore this abstra
 import abc
 import uuid
 
-# pylint: disable=no-name-in-module, import-error
+# pylint: disable=no-name-in-module,import-error
 from sqlalchemy_utils.types.choice import Choice
 from sqlalchemy.types import Integer, Float, Boolean, DateTime
 from sqlalchemy.dialects.postgresql import JSONB
 
-from aiida.common import exceptions
 from aiida.common.lang import type_check
-from aiida.common.exceptions import InputValidationError
 
 __all__ = ('BackendQueryBuilder',)
 
@@ -220,7 +218,7 @@ class BackendQueryBuilder:
         elif operator == 'in':
             expr = database_entity.in_(value)
         else:
-            raise InputValidationError(f'Unknown operator {operator} for filters on columns')
+            raise ValueError(f'Unknown operator {operator} for filters on columns')
         return expr
 
     def get_projectable_attribute(self, alias, column_name, attrpath, cast=None, **kwargs):
@@ -244,7 +242,7 @@ class BackendQueryBuilder:
         elif cast == 'd':
             entity = entity.astext.cast(DateTime)
         else:
-            raise InputValidationError(f'Unkown casting key {cast}')
+            raise ValueError(f'Unkown casting key {cast}')
         return entity
 
     def get_aiida_res(self, res):
@@ -399,13 +397,9 @@ class BackendQueryBuilder:
         """
         try:
             return getattr(alias, colname)
-        except AttributeError:
-            raise exceptions.InputValidationError(
+        except AttributeError as exc:
+            raise ValueError(
                 '{} is not a column of {}\n'
                 'Valid columns are:\n'
-                '{}'.format(
-                    colname,
-                    alias,
-                    '\n'.join(alias._sa_class_manager.mapper.c.keys())  # pylint: disable=protected-access
-                )
-            )
+                '{}'.format(colname, alias, '\n'.join(alias._sa_class_manager.mapper.c.keys()))  # pylint: disable=protected-access
+            ) from exc
--- a/aiida/orm/nodes/data/array/xy.py
+++ b/aiida/orm/nodes/data/array/xy.py
@@ -13,7 +13,7 @@ collections of y-arrays bound to a single x-array, and the methods to operate
 on them.
 """
 import numpy as np
-from aiida.common.exceptions import InputValidationError, NotExistent
+from aiida.common.exceptions import NotExistent
 from .array import ArrayData
 
 
@@ -43,19 +43,19 @@ class XyData(ArrayData):
     def _arrayandname_validator(array, name, units):
         """
         Validates that the array is an numpy.ndarray and that the name is
-        of type str. Raises InputValidationError if this not the case.
+        of type str. Raises TypeError or ValueError if this not the case.
         """
         if not isinstance(name, str):
-            raise InputValidationError('The name must always be a str.')
+            raise TypeError('The name must always be a str.')
 
         if not isinstance(array, np.ndarray):
-            raise InputValidationError('The input array must always be a numpy array')
+            raise TypeError('The input array must always be a numpy array')
         try:
             array.astype(float)
-        except ValueError:
-            raise InputValidationError('The input array must only contain floats')
+        except ValueError as exc:
+            raise TypeError('The input array must only contain floats') from exc
         if not isinstance(units, str):
-            raise InputValidationError('The units must always be a str.')
+            raise TypeError('The units must always be a str.')
 
     def set_x(self, x_array, x_name, x_units):
         """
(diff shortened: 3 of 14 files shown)


CasperWA (commented):
If you intend to completely remove the use of these exceptions (`InputValidationError` and `ValidationError`), maybe they should also be removed from `aiida.common.exceptions`?

sphuber:
> If you intend to completely remove the use of these exceptions (`InputValidationError` and `ValidationError`), maybe they should also be removed from `aiida.common.exceptions`?

No, as discussed during the meeting, the exceptions still have their use (they are still used in the `CalcJob.presubmit` and that is what it was intended for, validating calculation job input nodes) and are also still used by plugins to raise. Here we are just making sure our ORM doesn't abuse it.

CasperWA:
> (...) the exceptions still have their use (they are still used in the `CalcJob.presubmit` and that is what it was intended for, validating calculation job input nodes) and are also still used by plugins to raise. Here we are just making sure our ORM doesn't abuse it.

Just read through the related issue. Indeed, you're right.

Looking through the doc strings of the exceptions as well, I think `ValidationError` could get a refresh, but the `InputValidationError` seems clear enough as is. Although it's unclear to me why this cannot be used for workflows as well?

sphuber:
> Although it's unclear to me why this cannot be used for workflows as well?

I mean you could, I just don't see the point. Who would be catching these exceptions? And when you want them to be caught, unless there is a specific reason they need to be very distinguishable, I think custom exceptions often make things more complicated than necessary.

CasperWA:
> (...) And when you want them to be caught, unless there is a specific reason they need to be very distinguishable, I think custom exceptions often make things more complicated than necessary.

I quite disagree. As long as you keep a proper hierarchy it's quite useful to have explanatory exception names in the code. For me it's not so much about the "catchability" (for this, a proper hierarchy and a parent exception can be used), but for more readable code, quickly making it clear _why_ someone thinks it's necessary to raise an exception in the code. But that's just me :)

But for the first part, I would update the doc strings of the exceptions to make it clear where it should be used then, to avoid similar things happening in the future. Although, it can be argued in this case that even though the doc string was already quite explicit for `InputValidationError`, it was still ignored... 😅

sphuber:
@ltalirz could you please follow up on the discussion in the issue of this PR. And maybe some one else still wants to review this? @chrisjsewell @ramirezfranciscof ?

aiida/cmdline/commands/cmd_code.py
```diff
@@ -88,15 +86,15 @@ def setup_code(non_interactive, **kwargs):
...
     try:
         code = code_builder.new()
-    except InputValidationError as exception:
+    except ValueError as exception:
         echo.echo_critical(f'invalid inputs: {exception}')
 
     try:
         code.store()
-        code.reveal()
-    except ValidationError as exception:
```
ramirezfranciscof, line 97: Just as a comment I would maybe add a small note to the commit message saying that you also fixed this bug (the reveal being inside the with construct) and made a more open ended exception catching.

aiida/orm/nodes/data/array/xy.py
```diff
@@ -43,19 +43,19 @@ class XyData(ArrayData):
...
+            raise TypeError('The name must always be a str.')
 
         if not isinstance(array, np.ndarray):
-            raise InputValidationError('The input array must always be a numpy array')
+            raise TypeError('The input array must always be a numpy array')
         try:
             array.astype(float)
         except ValueError:
-            raise InputValidationError('The input array must only contain floats')
+            raise ValueError('The input array must only contain floats')
```
ramirezfranciscof, line 56: Nitpick, but wouldn't this be a `TypeError`?

aiida/orm/querybuilder.py
```diff
@@ -986,22 +985,22 @@ def order_by(self, order_by):
                         elif isinstance(orderspec, dict):
                             this_order_spec = orderspec
                         else:
-                            raise InputValidationError(
+                            raise ValueError(
```
ramirezfranciscof, line 988: Nitpick, but shouldn't this be a `TypeError`?

aiida/orm/utils/builders/code.py
```diff
@@ -195,7 +194,7 @@ def validate_installed(self):
         if messages:
             raise self.CodeValidationError(f'{messages}')
 
-    class CodeValidationError(Exception):
+    class CodeValidationError(ValueError):
```
ramirezfranciscof, line 197: I would maybe add a small comment on the commit message mentioning this change (as it doesn't seem immediately related to the others).

aiida/tools/graph/age_rules.py
```diff
@@ -208,7 +207,7 @@ def _init_run(self, operational_set):
             for proj_tag, projectionlist in projections.items():
                 try:
                     self._querybuilder.add_projection(proj_tag, projectionlist)
-                except InputValidationError:
+                except ValueError:
```
ramirezfranciscof, line 210: `add_projection` can also raise `TypeError`, right? (line 1180 of `aiida.orm.querybuilder` or whereabouts) Are you not catching that one because there is no point in adding the extra info?

Btw, shouldn't we `except ... as ...` and `raise ... from ...` here?

ramirezfranciscof (commented):
Thanks @sphuber ! I have a couple of small things (some may be nitpicks) to mention. One of the comments though might be worth generalizing: since you are adapting all of these raises, would it make sense to turn most (if not all) of them into re-raises? (so, `except ... as ...` and `raise ... from ...`)

aiida/cmdline/commands/cmd_code.py
sphuber: I don't think this is necessarily bug is it?

aiida/orm/nodes/data/array/xy.py
sphuber: Yeah, I also was wondering about this, but since the actual top level type (list) was correct, I turned it into a `ValueError`. But thinking more about it I think you are right. They type should be `List[float]` and so the type of the list elements is part of the entire type, which is not respected. Will change it.

aiida/tools/graph/age_rules.py
sphuber: No reason, you are right, should also catch the `TypeError`.

sphuber: Btw, since I think you wrote this code: why do you raise a `KeyError` here and why has the string an explicit newline?

aiida/orm/utils/builders/code.py
sphuber: Updated

aiida/orm/implementation/querybuilder.py
```diff
@@ -400,7 +398,7 @@ def get_column(self, colname, alias):  # pylint: disable=no-self-use
         try:
             return getattr(alias, colname)
         except AttributeError:
-            raise exceptions.InputValidationError(
+            raise ValueError(
```
ramirezfranciscof, line 401: Should this also `raise ... from ...` ?

aiida/orm/nodes/data/array/xy.py
```diff
@@ -86,20 +86,20 @@ def set_y(self, y_arrays, y_names, y_units):
...
         if len(y_units) != len(y_names):
-            raise InputValidationError('Length of units does not match!')
+            raise ValueError('Length of units does not match!')
 
         # Try to get the x_array
         try:
             x_array = self.get_x()[1]
         except NotExistent:
-            raise InputValidationError('X array has not been set yet')
+            raise ValueError('X array has not been set yet')
```
ramirezfranciscof, line 97: Should this also be `raise ... from ...`?

ramirezfranciscof (approved):
I noticed a couple more of `raise ... from ...` that I think could be added; you can check them out but other than that this is ready!

The only other thing I noticed is that there seem to be a lot of validation pathways that are not being unit-tested. I was going to propose to open an issue because I thought this was mostly for `Querybuilder` error handling, but I see this is also in the data plugins so I'm not sure. Adding one issue for each or one issue for all seem both not ideal, although if I had to choose I would probably go for the latter, no?

aiida/tools/graph/age_rules.py
ramirezfranciscof: Yes, now I don't remember if I added this or it was originally like that on the base AGE branch; in any case I can't think of any good reason so I was probably just blundering 😅 

If you want to change it though, is there a way to re-raise the same kind of exception? (since you are catching both `ValueException` and `TypeException`, you don't know a-priori which one to re-raise...)

aiida/orm/implementation/querybuilder.py
sphuber: I don't understand why `pylint` is not complaining about these.

aiida/orm/utils/builders/code.py
ramirezfranciscof: Ok, great! I meant it more like to take it into consideration for the final marge, but good to already have the original commit as a reference =)

aiida/orm/nodes/data/array/xy.py
ramirezfranciscof: Yes, I also hesitated a bit on this one, t'was a tricky one.

sphuber:
> The only other thing I noticed is that there seem to be a lot of validation pathways that are not being unit-tested. I was going to propose to open an issue because I thought this was mostly for `Querybuilder` error handling, but I see this is also in the data plugins so I'm not sure. Adding one issue for each or one issue for all seem both not ideal, although if I had to choose I would probably go for the latter, no?

I think one for the `QueryBuilder` would be useful maybe since there is a lot of testing missing I think. But if I remember correctly, there may already be an issue related to this, or at least that the code should be cleaned/refactored. Maybe add it just to that one that the test coverage should also be improved.

aiida/cmdline/commands/cmd_code.py
ramirezfranciscof: Yeah, I was thinking that if it was raising because of `reveal` it would actually store the thing and then say it didn't. But now I notice that since it is catching specifically the `ValidationError ` you are probably right (as `reveal` was not raising this kind of exception).

ramirezfranciscof:
I have found nothing with the tags [refactoring](https://github.com/aiidateam/aiida-core/issues?q=is%3Aopen+is%3Aissue+label%3Atype%2Frefactoring), [testing](https://github.com/aiidateam/aiida-core/issues?q=is%3Aopen+is%3Aissue+label%3Atopic%2Ftesting) or [querybuilder](https://github.com/aiidateam/aiida-core/issues?q=is%3Aopen+is%3Aissue+label%3Atopic%2Fquery-builder). I even went through the whole list of a [simple "test" search](https://github.com/aiidateam/aiida-core/issues?page=6&q=is%3Aopen+is%3Aissue+test) and nothing. I also had the idea that there was one, but if it is there is very well camouflaged =S.

Should I create two issues then, one for general coverage and one for the querybuilder in particular?

sphuber:
Sure, if you can't find anything, go ahead and open an issue for it. How it should be split depends really. We had one single issue before about test coverage that @ltalirz opened and then closed once the biggest offenders were fixed. Maybe we need to reinstate it or create a new one where you add these new offenders

PR #4889, opened 2021-04-29
`Manager`: check if repository container has been initialised

### LINKED ISSUE #4885: Repository database compatibility check shows when creating new profile
Just before the migration is performed the warning shows telling that the database has repo UUID `None` which makes sense, but the warning should not be shown here.

mbercx:
Fixes #4885

When loading the backend of the profile with the `Manager._load_backend`
method, the `get_repository_container` method of the `Profile` is
called, initialising a container for the profile in case is has not yet
been initialised. This causes issues during the migration of the
database/repository, since here the presence of the container directory
results in a critical failure of the migration command.

Here we add a check in the `_load_backend` method to see if the
container of the profile has already been initialised. If it has, the
repository uuid obtained from the container, if not it is set to `None`.
This also removes the warning that was printed during the migration
or the creation of a new profile, since the repository uuid obtained
from the database backend is also `None` before the migration.

### DIFF
--- a/aiida/manage/configuration/profile.py
+++ b/aiida/manage/configuration/profile.py
@@ -134,13 +134,26 @@ class Profile:  # pylint: disable=too-many-public-methods
         """
         from disk_objectstore import Container
 
-        filepath = os.path.join(self.repository_path, 'container')
+        filepath = self._container_path
         container = Container(filepath)
 
-        if not container.is_initialised:
+        if not self.container_is_initialised:
             container.init_container(clear=True, **self.defaults['repository'])  # pylint: disable=unsubscriptable-object
 
         return container
+
+    @property
+    def container_is_initialised(self):
+        """Check if the container of the profile file repository has already been initialised."""
+        from disk_objectstore import Container
+        filepath = self._container_path
+        container = Container(filepath)
+        return container.is_initialised
+
+    @property
+    def _container_path(self):
+        """Return the path to the container of the profile file repository."""
+        return os.path.join(self.repository_path, 'container')
 
     @property
     def uuid(self):
--- a/aiida/manage/manager.py
+++ b/aiida/manage/manager.py
@@ -136,7 +136,10 @@ class Manager:
         # yet known, we issue a warning in the case the repo and database are incompatible. In the future this might
         # then become an exception once we have verified that it is working reliably.
         if repository_check and not profile.is_test_profile:
-            repository_uuid_config = profile.get_repository_container().container_id
+            if profile.container_is_initialised:
+                repository_uuid_config = profile.get_repository_container().container_id
+            else:
+                repository_uuid_config = None
             repository_uuid_database = backend_manager.get_repository_uuid()
 
             from aiida.cmdline.utils import echo


mbercx:
I've tried this for my development database and it seems to work fine for the migration. I did get the following warning though:

```console
Warning:                                                                                                                                                                
Detected repository folders that were missing the required subfolder `path` or `raw_input`. The paths of those nodes repository folders have been written to a log file: /home/mbercx/envs/aiida-dev/setup/profile/migration-repository-missing-subfolder-4i32qdwj.json
```

I do tend to mess around with my development database/repo quite a bit, so I may have messed it up somewhere. Will clone a different production database and try again with this branch.

aiida/manage/configuration/profile.py
```diff
@@ -134,14 +134,27 @@ def get_repository_container(self) -> 'Container':
...
+        """Check if the container of the profile file repository has already been initialised."""
+        from disk_objectstore import Container
+        filepath = self._container_path
+        container = Container(filepath)
+        return container.is_initialised
+
+    @property
+    def _container_path(self):
+        """Return the path to the container of the profile file repository."""
+        return os.path.join(self.repository_path, 'container')
```
CasperWA, line 156: `self.repository_path` should be a subclass of `pathlib.Path`.
```suggestion
        return self.repository_path / 'container'
```

aiida/manage/configuration/profile.py
```diff
@@ -134,14 +134,27 @@ def get_repository_container(self) -> 'Container':
...
+    @property
+    def container_is_initialised(self):
+        """Check if the container of the profile file repository has already been initialised."""
+        from disk_objectstore import Container
+        filepath = self._container_path
+        container = Container(filepath)
+        return container.is_initialised
+
+    @property
+    def _container_path(self):
```
CasperWA, line 154:
```suggestion
    def _container_path(self) -> pathlib.Path:
```

aiida/manage/configuration/profile.py
```diff
@@ -134,14 +134,27 @@ def get_repository_container(self) -> 'Container':
...
+        if not self.container_is_initialised:
             container.init_container(clear=True, **self.defaults['repository'])  # pylint: disable=unsubscriptable-object
 
         return container
 
+    @property
+    def container_is_initialised(self):
+        """Check if the container of the profile file repository has already been initialised."""
+        from disk_objectstore import Container
+        filepath = self._container_path
```
CasperWA, line 149: Why didn't `pre-commit` catch this?
```suggestion

        filepath = self._container_path
```

aiida/manage/configuration/profile.py
```diff
@@ -134,14 +134,27 @@ def get_repository_container(self) -> 'Container':
...
         container = Container(filepath)
 
-        if not container.is_initialised:
+        if not self.container_is_initialised:
             container.init_container(clear=True, **self.defaults['repository'])  # pylint: disable=unsubscriptable-object
 
         return container
 
+    @property
+    def container_is_initialised(self):
```
CasperWA, line 146:
```suggestion
    def container_is_initialised(self) -> bool:
```

aiida/manage/configuration/profile.py
```diff
@@ -134,14 +134,27 @@ def get_repository_container(self) -> 'Container':
...
+        """Check if the container of the profile file repository has already been initialised."""
```
CasperWA, line 147:
```suggestion
        """Return whether the container of the profile file repository has already been initialised."""
```

aiida/manage/configuration/profile.py
```diff
@@ -134,14 +134,27 @@ def get_repository_container(self) -> 'Container':
...
+        from disk_objectstore import Container
+        filepath = self._container_path
+        container = Container(filepath)
+        return container.is_initialised
```
CasperWA, line 151:
```suggestion
        return Container(self._container_path).is_initialised
```

aiida/manage/configuration/profile.py
sphuber, line 146: I am wondering if we should remove `container` from these methods. I know that I called the method `get_repository_container` but the `container` part really comes from this particular implementation with the `disk-objectstore`. If we were to ever change it (don't think we will but in principle) the container concept may disappear. Maybe it would be better to have the following method and property names:
```
def get_repository() -> `Repository`
def is_repository_initialised
def initialise_repository
```
Looking at this more, currently `get_repository_container` returns the `Container` instance, but this is also a leaky abstraction. Some of the calls to this method currently just use it to get the UUID of the repo. Since then I added `Profile.get_repository_uuid` which is better and those places should use that instead.

Ok, this was a bit a train of thought and I am now thinking that we should just merge this and then I will do another PR with some proposed interface and name changes. What do you think?

aiida/manage/configuration/profile.py
CasperWA: Also, if this is a more or less "constant" property, you could store it as a private variable and only create the `Container` instance once per call to this property, returning the "cached" private variable all the other times instead. I.e., you store the first result here in the private variable.

CasperWA (changes requested):

aiida/manage/configuration/profile.py
CasperWA: Sounds good to me. Then we can move on and you can disregard my comments since it should be updated anyway in the upcoming PR and they are not functionally critical.

sphuber: OK, then I will merge this now and include your comments in the next PR. Then we can start testing the migration again. Thanks guys

sphuber (approved):
Fanfuckingtastic

PR #4891, opened 2021-04-29
`Profile`: do not leak repository implementation in interface

### LINKED ISSUE #4890: Do not leak disk-objectstore specificities in the public interface of the `Profile`

sphuber:
Fixes #4890 

The `Profile` interface was leaking implementation details of the
repository. For example, the method `get_repository_container` returned
the `Container` class of the `disk-objectstore` library, but this is an
implementation detail. If this ever were to change all client code would
break.

Instead, we add the `get_repository` method which returns an instance of
the generic `Repository` class that will be common to all repository
implementations. The backend implementation can still be obtained
through this object but this should only be done in exceptional cases.

The `Repository` class now has three additional properties and methods:

 * `def uuid(self) -> typing.Optional[str]:`
 * `def initialise(self, **kwargs) -> None:`
 * `def is_initialised(self) -> bool:`

These simply call through to the exact same method/property on the
backend instance. The `AbstractRepositoryBackend` also now has the exact
same attributes and they are implemented for the two currently existing
implementations `DiskObjectStoreRepositoryBackend` and the
`SandboxRepositoryBackend`.

### DIFF
--- a/aiida/cmdline/commands/cmd_status.py
+++ b/aiida/cmdline/commands/cmd_status.py
@@ -84,13 +84,13 @@ def verdi_status(print_traceback, no_rmq):
 
     # Getting the repository
     try:
-        container = profile.get_repository_container()
+        repository = profile.get_repository()
     except Exception as exc:
         message = 'Error with repository folder'
         print_status(ServiceStatus.ERROR, 'repository', message, exception=exc, print_traceback=print_traceback)
         exit_code = ExitCode.CRITICAL
     else:
-        repository_status = f'Connected to {container.get_folder()} [UUID={container.container_id}]'
+        repository_status = f'Connected to {repository}'
         print_status(ServiceStatus.UP, 'repository', repository_status)
 
     # Getting the postgres status by trying to get a database cursor
--- a/aiida/manage/configuration/profile.py
+++ b/aiida/manage/configuration/profile.py
@@ -10,6 +10,7 @@
 """AiiDA profile related code"""
 import collections
 import os
+import pathlib
 from typing import TYPE_CHECKING
 
 from aiida.common import exceptions
@@ -19,7 +20,7 @@ from .options import parse_option
 from .settings import DAEMON_DIR, DAEMON_LOG_DIR
 
 if TYPE_CHECKING:
-    from disk_objectstore import Container
+    from aiida.repository import Repository  # pylint: disable=ungrouped-imports
 
 __all__ = ('Profile',)
 
@@ -127,33 +128,23 @@ class Profile:  # pylint: disable=too-many-public-methods
         # Currently, whether a profile is a test profile is solely determined by its name starting with 'test_'
         self._test_profile = bool(self.name.startswith('test_'))
 
-    def get_repository_container(self) -> 'Container':
-        """Return the container of the profile's file repository.
+    def get_repository(self) -> 'Repository':
+        """Return the repository configured for this profile.
 
-        :return: the profile's file repository container.
+        .. note:: The repository will automatically be initialised if it wasn't yet already.
         """
         from disk_objectstore import Container
+        from aiida.repository import Repository
+        from aiida.repository.backend import DiskObjectStoreRepositoryBackend
 
-        filepath = self._container_path
-        container = Container(filepath)
+        container = Container(self.repository_path / 'container')
+        backend = DiskObjectStoreRepositoryBackend(container=container)
+        repository = Repository(backend=backend)
 
-        if not self.container_is_initialised:
-            container.init_container(clear=True, **self.defaults['repository'])  # pylint: disable=unsubscriptable-object
+        if not repository.is_initialised:
+            repository.initialise(clear=True, **self.defaults['repository'])  # pylint: disable=unsubscriptable-object
 
-        return container
-
-    @property
-    def container_is_initialised(self):
-        """Check if the container of the profile file repository has already been initialised."""
-        from disk_objectstore import Container
-        filepath = self._container_path
-        container = Container(filepath)
-        return container.is_initialised
-
-    @property
-    def _container_path(self):
-        """Return the path to the container of the profile file repository."""
-        return os.path.join(self.repository_path, 'container')
+        return repository
 
     @property
     def uuid(self):
@@ -357,12 +348,12 @@ class Profile:  # pylint: disable=too-many-public-methods
         return self._test_profile
 
     @property
-    def repository_path(self):
+    def repository_path(self) -> pathlib.Path:
         """Return the absolute path of the repository configured for this profile.
 
         :return: absolute filepath of the profile's file repository
         """
-        return self._parse_repository_uri()[1]
+        return pathlib.Path(self._parse_repository_uri()[1])
 
     def _parse_repository_uri(self):
         """
--- a/aiida/manage/manager.py
+++ b/aiida/manage/manager.py
@@ -136,10 +136,7 @@ class Manager:
         # yet known, we issue a warning in the case the repo and database are incompatible. In the future this might
         # then become an exception once we have verified that it is working reliably.
         if repository_check and not profile.is_test_profile:
-            if profile.container_is_initialised:
-                repository_uuid_config = profile.get_repository_container().container_id
-            else:
-                repository_uuid_config = None
+            repository_uuid_config = profile.get_repository().uuid
             repository_uuid_database = backend_manager.get_repository_uuid()
 
             from aiida.cmdline.utils import echo
--- a/aiida/repository/backend/sandbox.py
+++ b/aiida/repository/backend/sandbox.py
@@ -4,6 +4,7 @@ import contextlib
 import io
 import os
 import shutil
+import typing
 import uuid
 
 from .abstract import AbstractRepositoryBackend
@@ -17,6 +18,10 @@ class SandboxRepositoryBackend(AbstractRepositoryBackend):
     def __init__(self):
         self._sandbox = None
 
+    def __str__(self) -> str:
+        """Return the string representation of this repository."""
+        return f'SandboxRepository: {self._sandbox.abspath}'
+
     def __del__(self):
         """Delete the entire sandbox folder if it was instantiated and still exists."""
         if getattr(self, '_sandbox', None) is not None:
@@ -24,6 +29,28 @@ class SandboxRepositoryBackend(AbstractRepositoryBackend):
                 shutil.rmtree(self.sandbox.abspath)
             except FileNotFoundError:
                 pass
+
+    @property
+    def uuid(self) -> typing.Optional[str]:
+        """Return the unique identifier of the repository.
+
+        .. note:: A sandbox folder does not have the concept of a unique identifier and so always returns ``None``.
+        """
+        return None
+
+    def initialise(self, **kwargs) -> None:
+        """Initialise the repository if it hasn't already been initialised.
+
+        :param kwargs: parameters for the initialisation.
+        """
+        # Merely calling the property will cause the sandbox folder to be initialised.
+        self.sandbox  # pylint: disable=pointless-statement
+
+    @property
+    def is_initialised(self) -> bool:
+        """Return whether the repository has been initialised."""
+        from aiida.common.folders import SandboxFolder
+        return isinstance(self._sandbox, SandboxFolder)
 
     @property
     def sandbox(self):
--- a/aiida/repository/repository.py
+++ b/aiida/repository/repository.py
@@ -47,6 +47,27 @@ class Repository:
         self.set_backend(backend)
         self.reset()
 
+    def __str__(self) -> str:
+        """Return the string representation of this repository."""
+        return f'Repository<{str(self.backend)}>'
+
+    @property
+    def uuid(self) -> typing.Optional[str]:
+        """Return the unique identifier of the repository or ``None`` if it doesn't have one."""
+        return self.backend.uuid
+
+    def initialise(self, **kwargs) -> None:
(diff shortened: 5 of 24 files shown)


sphuber:
Haven't gone over all the comments of @CasperWA of the other PR yet, but will do later. I propose @mbercx already goes over the suggested interface changes here. Will have time to get back to this later in the afternoon.

chrisjsewell (approved):

chrisjsewell (changes requested):
well perhaps I should say, approved after it doesn't break all the tests lol

CasperWA:
> Haven't gone over all the comments of @CasperWA of the other PR yet, but will do later (...)

They're all syntax and style, essentially. So no worries.

aiida/manage/configuration/profile.py
```diff
@@ -127,33 +128,43 @@ def __init__(self, name, attributes, from_config=False):
         # Currently, whether a profile is a test profile is solely determined by its name starting with 'test_'
         self._test_profile = bool(self.name.startswith('test_'))
 
-    def get_repository_container(self) -> 'Container':
-        """Return the container of the profile's file repository.
+    @property
+    def _repository_container_path(self) -> str:
```
chrisjsewell, line 132: paths should be return as `pathlib.Path`
if not already the case, it would be good if `Container` accepted a `Path` for its initialisation

CasperWA: One of my comments as well 😅

mbercx: In my (very poor) defence, I only copy-pasted @giovannipizzi's code snippets from slack and tested if it worked. 😅

CasperWA: Those are not meant as the final version to go into the codebase I'm suspecting ;) So that's a "no" on your defence 😂

CasperWA: Also, @mbercx if you haven't already, you should install `pre-commit` in the repo by running `pre-commit install`. If it says you don't have `pre-commit` install with the `pre-commit` extra, i.e., `pip install -U -e .[pre-commit]` (or similar).

aiida/manage/configuration/profile.py
```diff
@@ -127,33 +128,43 @@ def __init__(self, name, attributes, from_config=False):
...
-        """
+    def _get_repository_container(self) -> 'Container':
+        """Return the disk object store ``Container``."""
         from disk_objectstore import Container
+        return Container(self._repository_container_path)
 
-        filepath = self._container_path
-        container = Container(filepath)
+    def get_repository(self) -> 'Repository':
+        """Return the UUID that uniquely identifies the repository configured for this profile."""
```
mbercx, line 142:
```suggestion
        """Return the repository configured for this profile.
        
        .. note:: will initialise the repository if not already done so.
        """
```

aiida/manage/manager.py
```diff
@@ -136,8 +136,8 @@ def _load_backend(self, schema_check: bool = True, repository_check: bool = True
         # yet known, we issue a warning in the case the repo and database are incompatible. In the future this might
         # then become an exception once we have verified that it is working reliably.
         if repository_check and not profile.is_test_profile:
-            if profile.container_is_initialised:
-                repository_uuid_config = profile.get_repository_container().container_id
+            if profile.is_repository_initialised:
```
mbercx, line 139: Is this check still necessary in case `get_repository_uuid()` returns `None` for non-initialised repositories?

aiida/manage/configuration/profile.py
```diff
@@ -127,33 +128,43 @@ def __init__(self, name, attributes, from_config=False):
...
+        self.initialise_repository()
+        backend = DiskObjectStoreRepositoryBackend(container=container)
+        return Repository(backend=backend)
 
-        if not self.container_is_initialised:
-            container.init_container(clear=True, **self.defaults['repository'])  # pylint: disable=unsubscriptable-object
+    def get_repository_uuid(self) -> Optional[str]:
+        """Return the UUID that uniquely identifies the repository configured for this profile.
+
+        .. note:: will return ``None`` if the repository has not yet been initialized.
```
mbercx, line 153:
```suggestion
        .. note:: will return ``None`` if the repository has not yet been initialised.
```
Else @chrisjsewell will be angry. 😁

aiida/manage/manager.py
mbercx: If not, perhaps we should make this a hidden method, or even remove it entirely? I feel we may be overdoing it with the number of repository methods/properties. 😅

mbercx (changes requested):
Thanks @sphuber. Gave the PR a first pass.

aiida/manage/configuration/profile.py
mbercx: I have the pre-commit hooks installed, and they all passed for my commit. 🤷‍♂️  I haven't run `pre-commit install` in a while though, maybe the hooks have changed since then? I didn't know there should be an empty line after an import in a method (I'm assuming you're referring to [this comment](https://github.com/aiidateam/aiida-core/pull/4889#discussion_r622917362)).

CasperWA: Indeed. I thought we were using black now. i guess not? But black always puts an empty new line after imports.

sphuber: the `Container` calls `os.path.realpath` on the path argument. Does that support `Path` objects already in Python 3.6? If so, then we are good.

aiida/manage/manager.py
sphuber: Maybe what would be better is that, now `Profile.get_repository` returns a `Repository` instance, we implement those methods there. That should be the correct way of doing things.

aiida/manage/configuration/profile.py
chrisjsewell: > Edit: Ah, no. We're indeed not yet using black. Still on yapf.... pff

Ah you are preaching to the choir my friend 😆: https://github.com/aiidateam/aiida-core/pull/4345#issuecomment-826762603

> you should install pre-commit in the repo by running pre-commit install

you could really do that with aiida-core, because pylint and mypy use system installs: use tox (its the way forward): `tox -e py38-pre-commit`

chrisjsewell: >  Does that support Path objects already in Python 3.6?

I think everything in `os.path` supports `Path` for >=3.6, but can check

chrisjsewell: yep:

```python
In [5]: realpath??
Signature: realpath(filename)
Source:   
def realpath(filename):
    """Return the canonical path of the specified filename, eliminating any
symbolic links encountered in the path."""
    filename = os.fspath(filename)
    path, ok = _joinrealpath(filename[:0], filename, {})
    return abspath(path)
File:      ~/opt/miniconda3/envs/python-3.6/lib/python3.6/posixpath.py
Type:      function

In [6]: realpath(Path("."))
Out[6]: '/Users/chrisjsewell/Documents/GitHub/aiida-core-origin'
```

aiida/manage/manager.py
mbercx: Indeed, having `Repository.is_initialised` and `Repository.initialise` sounds like a good idea. 👍

sphuber:
Alright, this should be much better. If you guys are happy with the design, I will add the necessary tests for the new `Repository` and `RepositoryBackend` properties and methods.

mbercx:
> Alright, this should be much better. If you guys are happy with the design, I will add the necessary tests for the new `Repository` and `RepositoryBackend` properties and methods.

Looking very good indeed! Much cleaner to have these methods on the `Repository` class. 👌  Will do a proper review and some testing after the wrap-up meeting.

chrisjsewell:
> Looking very good indeed!

not that it means much to have @mbercx 's seal of a approval

CasperWA:
> > Looking very good indeed!
> 
> not that it means much to have @mbercx 's seal of a approval

Indeed. I quickly jumped in to review when I saw I blanketed this PR ... tsk tsk tsk... 😂

mbercx:
> > Looking very good indeed!
> 
> not that it means much to have @mbercx 's seal of a approval

Does that mean I don't have to review PR's anymore? 🥳

aiida/repository/backend/sandbox.py
```diff
@@ -17,6 +18,10 @@ class SandboxRepositoryBackend(AbstractRepositoryBackend):
     def __init__(self):
         self._sandbox = None
 
+    def __str__(self) -> str:
+        """Return the string representation of this repository."""
+        return f'SandboxRepository: {self._sandbox.abspath}'
```
CasperWA, line 23: The UUID is not being returned here, might be nice? Also whether or not it's initialised?

aiida/repository/repository.py
```diff
@@ -47,6 +47,27 @@ def __init__(self, backend: AbstractRepositoryBackend = None):
         self.set_backend(backend)
         self.reset()
 
+    def __str__(self) -> str:
+        """Return the string representation of this repository."""
+        return f'Repository<{str(self.backend)}>'
```
CasperWA, line 52: See comment concerning the string representation of `SandboxRepositoryBackend`.

tests/cmdline/commands/test_setup.py
```diff
@@ -79,8 +79,8 @@ def test_quicksetup(self):
         # Check that the repository UUID was stored in the database
         manager = get_manager()
         backend_manager = manager.get_backend_manager()
-        repository = profile.get_repository_container()
-        self.assertEqual(backend_manager.get_repository_uuid(), repository.container_id)
+        repository_uuid = profile.get_repository().uuid
+        self.assertEqual(backend_manager.get_repository_uuid(), repository_uuid)
```
CasperWA, line 83: It reads a bit redundant now I think. But up to you.
```suggestion
        self.assertEqual(backend_manager.get_repository_uuid(), profile.get_repository().uuid)
```

tests/cmdline/commands/test_setup.py
```diff
@@ -167,5 +167,5 @@ def test_setup(self):
...
```
CasperWA, line 171:
```suggestion
        self.assertEqual(backend_manager.get_repository_uuid(), profile.get_repository().uuid)
```

tests/cmdline/commands/test_setup.py
CasperWA: If it's used later as a proper variable fine to keep (can't see that it is). The same goes for the suggested change below.

CasperWA:
> > > Looking very good indeed!
> > 
> > not that it means much to have @mbercx 's seal of a approval
> 
> Does that mean I don't have to review PR's anymore? 🥳

Nah. Just means more work for the rest of us 😂

aiida/cmdline/commands/cmd_status.py
```diff
@@ -84,13 +84,13 @@ def verdi_status(print_traceback, no_rmq):
...
     try:
-        container = profile.get_repository_container()
+        repository = profile.get_repository()
     except Exception as exc:
         message = 'Error with repository folder'
         print_status(ServiceStatus.ERROR, 'repository', message, exception=exc, print_traceback=print_traceback)
         exit_code = ExitCode.CRITICAL
     else:
-        repository_status = f'Connected to {container.get_folder()} [UUID={container.container_id}]'
+        repository_status = f'Connected to {repository}'
```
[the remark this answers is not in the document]
sphuber: It does now :) I implemented a `__str__` on the repository implementations and the `DiskObjectStore` one includes its UUID as well as absolute filepath

aiida/repository/backend/sandbox.py
sphuber: Sandbox doesn't have a UUID and always returns `None`. It doesn't make sense for a sandbox tbh

aiida/repository/repository.py
sphuber: Since this calls `str(self.backend)` this will include whatever the `__str__` of the backend class implements. So for the DOS one, it will include the UUID.

aiida/repository/backend/sandbox.py
CasperWA: Ah right. Yeah. There were so many layers here I kind of lost track of which string is represented where 😅

sphuber:
Ok, the docs are failing with the following:
```
/home/circleci/.local/lib/python3.7/site-packages/aiida/manage/configuration/profile.py:docstring of aiida.manage.configuration.Profile.get_repository:: WARNING: py:class reference target not found: Repository
/home/circleci/.local/lib/python3.7/site-packages/aiida/manage/configuration/profile.py:docstring of aiida.manage.configuration.profile.Profile.get_repository:: WARNING: py:class reference target not found: Repository
```
This is most likely becuase of the return type, which I put as a string. The reason being that I cannot import `Repository` at the top of the module, because this leads to a circular import. For the typing, I put a guard in `if typing.TYPE_CHECKING` which makes sure `mypy` can properly import it, but for the docs this doesn't work. Not sure how to get around this. Any ideas? Other than maybe trying to break the circular import in the `aiida.repository.Repository` path, but that might not be trivial.

chrisjsewell:
> This is most likely becuase of the return type, which I put as a string. 

yeh its an annoying sphinx thing (https://github.com/sphinx-doc/sphinx/issues/8498), just add it to the `docs/source/nitpick-exceptions`

chrisjsewell (approved):

sphuber:
@mbercx would you still like a gander before we merge?

mbercx:
> @mbercx would you still like a gander before we merge?

LGMT! 🚀

PR #4896, opened 2021-04-30
Config: warn when the configuration directory is created

### LINKED ISSUE #4884: Warn when creating .aiida folder

AiiDA currently creates the `.aiida` folder on the fly if it does not yet exist.
It would be useful if the user is warned (once) when this happens - even if just to inform new users about where on the hard disk AiiDA stores data and in order to help detect issues with the `AIIDA_PATH` (e.g. forgot to set the environment variable) early on.

### Additional context

Transferred from https://github.com/aiidateam/aiida-core/issues/2522

sphuber:
Fixes #4884 

The configuration directory is automatically created when it is loaded
and it doesn't exist yet. This is useful for new installations, however,
it can lead to problems when users accidentally make a mistake when
setting an explicit path with the `AIIDA_PATH` environment variable. If
a type is present in the path, instead of the intended existing config
being used, a new empty one is created.

We add a warning to when the base configuration directory is created, if
and only if the `AIIDA_PATH` variable is set. Otherwise the warning
would always show up, also for new profiles which is most likely not
desirable.

The `aiida.manage.configuration.settings` module is also migrated to use
the `pathlib` module for file system path manipulations.

### DIFF
--- a/aiida/manage/configuration/settings.py
+++ b/aiida/manage/configuration/settings.py
@@ -8,9 +8,10 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Base settings required for the configuration of an AiiDA instance."""
-
-import errno
 import os
+import pathlib
+import typing
+import warnings
 
 USE_TZ = True
 
@@ -24,9 +25,9 @@ DEFAULT_CONFIG_INDENT_SIZE = 4
 DEFAULT_DAEMON_DIR_NAME = 'daemon'
 DEFAULT_DAEMON_LOG_DIR_NAME = 'log'
 
-AIIDA_CONFIG_FOLDER = None
-DAEMON_DIR = None
-DAEMON_LOG_DIR = None
+AIIDA_CONFIG_FOLDER: typing.Optional[pathlib.Path] = None
+DAEMON_DIR: typing.Optional[pathlib.Path] = None
+DAEMON_LOG_DIR: typing.Optional[pathlib.Path] = None
 
 
 def create_instance_directories():
@@ -35,32 +36,26 @@ def create_instance_directories():
     This will create the base AiiDA directory defined by the AIIDA_CONFIG_FOLDER variable, unless it already exists.
     Subsequently, it will create the daemon directory within it and the daemon log directory.
     """
-    directory_base = os.path.expanduser(AIIDA_CONFIG_FOLDER)
-    directory_daemon = os.path.join(directory_base, DAEMON_DIR)
-    directory_daemon_log = os.path.join(directory_base, DAEMON_LOG_DIR)
+    from aiida.common import ConfigurationError
+
+    directory_base = pathlib.Path(AIIDA_CONFIG_FOLDER).expanduser()
+    directory_daemon = directory_base / DAEMON_DIR
+    directory_daemon_log = directory_base / DAEMON_LOG_DIR
 
     umask = os.umask(DEFAULT_UMASK)
 
     try:
-        create_directory(directory_base)
-        create_directory(directory_daemon)
-        create_directory(directory_daemon_log)
+        for path in [directory_base, directory_daemon, directory_daemon_log]:
+
+            if path is directory_base and not path.exists():
+                warnings.warn(f'Creating AiiDA configuration folder `{path}`.')
+
+            try:
+                path.mkdir(parents=True, exist_ok=True)
+            except OSError as exc:
+                raise ConfigurationError(f'could not create the `{path}` configuration directory: {exc}') from exc
     finally:
         os.umask(umask)
-
-
-def create_directory(path):
-    """Attempt to create the configuration folder at the given path skipping if it already exists
-
-    :param path: an absolute path to create a directory at
-    """
-    from aiida.common import ConfigurationError
-
-    try:
-        os.makedirs(path)
-    except OSError as exception:
-        if exception.errno != errno.EEXIST:
-            raise ConfigurationError(f"could not create the '{path}' configuration directory")
 
 
 def set_configuration_directory():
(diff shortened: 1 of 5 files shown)


aiida/manage/configuration/settings.py
```diff
@@ -24,9 +25,9 @@
 DEFAULT_DAEMON_DIR_NAME = 'daemon'
 DEFAULT_DAEMON_LOG_DIR_NAME = 'log'
 
-AIIDA_CONFIG_FOLDER = None
-DAEMON_DIR = None
-DAEMON_LOG_DIR = None
+AIIDA_CONFIG_FOLDER: typing.Optional[pathlib.Path] = None
```
ltalirz, line 28: we are changing the type of some rather important configuration variable here.
a [quick search on github](https://github.com/search?p=2&q=AIIDA_CONFIG_FOLDER&type=Code)  shows that AIIDA_CONFIG_FOLDER is being used e.g. in aiida-vasp or aiida-pytest.

In the two examples I saw it is anyhow converted to a `Path` so I think those would not break, but I wonder whether this change is worth the risk (one could keep it a string and still use pathlib to manipulate the path inside the functions below).

If we are going to make the switch to pathlib for this variable, I would prefer if this was part of an effort to comb through aiida-core and convert all module-level variables that contain paths to `Path`s (`path.join` still gives 100+ hits on the codebase).

ltalirz (commented):
thanks a lot @sphuber !

> We add a warning to when the base configuration directory is created, if
and only if the AIIDA_PATH variable is set. Otherwise the warning
would always show up, also for new profiles which is most likely not
desirable.

Actually, I would have suggested to show this message also when the `AIIDA_PATH` is not set.
I think it is good to show users *once* where this directory is located.
It won't happen for every new profile - only for the very first one.

aiida/manage/configuration/settings.py
```diff
@@ -35,32 +36,27 @@ def create_instance_directories():
...
-        create_directory(directory_daemon_log)
-    finally:
-        os.umask(umask)
-
+        for path in [directory_base, directory_daemon, directory_daemon_log]:
 
-def create_directory(path):
-    """Attempt to create the configuration folder at the given path skipping if it already exists
+            if path is directory_base and environment_variable and not path.exists():
+                warnings.warn(f'config directory `{path}` does not exist, creating it now.')
```
ltalirz, line 52:
```suggestion
                warnings.warn(f'Creating AiiDA configuration folder `{path}`.')
```
[in the docs](https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/installation.html?highlight=backup#backing-up-your-installation) it is just called the `.aiida/` folder

aiida/manage/configuration/settings.py
sphuber: I thought that since most stdlib modules support `pathlib.Path` and `str` interchangeably, this would be ok. At least for Python 3.6 but Python 3.5 is no longer supported and Python 3.6 is even on its way out. In addition, the `AIIDA_CONFIG_FOLDER` variable is not part of the public API. So all these things combined, I didn't think it was a problem. I would love to see a general replacement of all path operations within the source code to use `pathlib` but this would be a lot of work and don't think anyone will do this in one go. Instead I think we should do this progressively as code is being touched, like I do here.

aiida/manage/configuration/settings.py
sphuber: Rather than updating the warning, I would update the documentation to name if the "AiiDA configuration folder", which currently just happens to be named `.aiida`. Would that be ok?

sphuber:
> Actually, I would have suggested to show this message also when the `AIIDA_PATH` is not set.
> I think it is good to show users _once_ where this directory is located (and `AIIDA_PATH` not being set is actually also a very likely reason for using the wrong configuration folder).
> It won't happen for every new profile - only for the very first one.

Happy to change this.

aiida/manage/configuration/settings.py
ltalirz: Ok - if that change is indeed very unlikely to break code in the plugins I'm fine with it.
As I mentioned the two examples I saw are not affected.

aiida/manage/configuration/settings.py
ltalirz: Sure!

sphuber: Done

ltalirz (commented):
thanks @sphuber , this now looks all good to me.

One last thing: this warning now [shows up in the tests](https://github.com/aiidateam/aiida-core/pull/4896/checks?check_run_id=2482965382#step:9:113) (as it should).

Since this is expected, could we ignore this warning in the tests?
E.g. using the `UserWarning` [pytest example](https://docs.pytest.org/en/stable/warnings.html) (or choose a specific name just for this one), and just add the warning to the [ignore list](https://github.com/aiidateam/aiida-core/blob/4a347b6001fc4f4dcc1c0710489a899b6589ee57/pyproject.toml#L66)

sphuber:
> One last thing: this warning now [shows up in the tests](https://github.com/aiidateam/aiida-core/pull/4896/checks?check_run_id=2482965382#step:9:113) (as it should).
> 
> Since this is expected, could we ignore this warning in the tests?
> E.g. using the `UserWarning` [pytest example](https://docs.pytest.org/en/stable/warnings.html) (or choose a specific name just for this one), and just add the warning to the [ignore list](https://github.com/aiidateam/aiida-core/blob/4a347b6001fc4f4dcc1c0710489a899b6589ee57/pyproject.toml#L66)

Done and done!

ltalirz:
> Done and done!

Thanks! 
For some reason I don't see it in the changed files - I tried reloading the page with cmd+shift+R.
You did push the changes that update e.g. the pyproject.toml correct?

sphuber:
> > Done and done!
> 
> Thanks!
> For some reason I don't see it in the changed files - I tried reloading the page with cmd+shift+R.
> You did push the changes that update e.g. the pyproject.toml correct?

I ignored them directly on the test, by decorating them with `pytest.mark.filterwarnings`

ltalirz:
> I ignored them directly on the test, by decorating them with `pytest.mark.filterwarnings`

Ah I see - but that doesn't accomplish what I had in mind.
The warning will be shown when the test profile is created (whichever test is being run).

Even better than ignoring it in the pyproject.toml would be to catch it in the setup of the test profile of the profile manager

sphuber:
> > I ignored them directly on the test, by decorating them with `pytest.mark.filterwarnings`
> 
> Ah I see - but that doesn't accomplish what I had in mind.
> The warning will be shown when the test profile is created (whichever test is being run).

Given that that is exactly the behavior that a normal installation has, I think we should keep that. In addition, this warning will not be captured by `pytest` as a warning because the creation of the test profile is not part of the `pytest` setup. So adding it to the pytest warnings won't get rid of it anyway. What I suppressed are the warnings that are raised in some of the tests that create a new config directory on the fly. For those tests, we know that the warning will be used (which will show up at the end of the test suite and mark the tests "yellow") but it is not the purpose of the test and is a known side-effect, so we want to ignore it.

ltalirz:
Ok, fair enough!

The warning still occurs in [two](https://github.com/aiidateam/aiida-core/pull/4896/checks?check_run_id=2485608260#step:9:159) [places](https://github.com/aiidateam/aiida-core/pull/4896/checks?check_run_id=2485608260#step:9:112) - could you please add the pytest ignore there as well?

ltalirz (approved):
thanks a lot!

ltalirz:
> In addition, this warning will not be captured by `pytest` as a warning because the creation of the test profile is not part of the `pytest` setup.

I now realize that I may have slightly misunderstood your comment - it is true that for our specific test setup the profile is not set up via pytest, but most plugins use the automatic setup mechanism we provide and in that case the warning *does* show up in pytest

https://github.com/aiidateam/aiida-cp2k/pull/140/checks?check_run_id=2490884833#step:7:508

I.e. we will either need to tell plugin developers to add an ignore line for this or find a way to ignore it in tests automatically.

Issue #4897, opened 2021-04-30
Cannot migrate due to `container` folder being present


CasperWA:
### Describe the bug

The issue of not being able to migrate an AiiDA database that should've been fixed in #4889 might've been undone by #4891.

I keep getting the following error (even after deleting the `container` folder):

```console
$ rm -rf /home/candersen/.aiida/repository/li-ion-conductors/container
$ verdi -p li-ion-conductors database migrate -f
...
Operations to perform:
  Apply all migrations: auth, contenttypes, db
Running migrations:
Critical: the container /home/candersen/.aiida/repository/li-ion-conductors/container already exists. If you ran this migration before and it failed simply delete this directory and restart the migration.
  Applying db.0047_migrate_repository...
$
```

### Steps to reproduce

Migrate a database using the latest `develop` commit (fa8d05f)

### Expected behavior

Successful migration.

### Your environment

- Operating system: Ubuntu 20.04
- Python version: 3.8.5
- aiida-core version: fa8d05f

CasperWA:
I think the issue is the removal of the `else` clause [here](https://github.com/aiidateam/aiida-core/pull/4891/files#diff-414702fc1e9c1f26c4adeeb124de237e15be7c7d3824326cd778078bf8515bc2R139).

sphuber:
Hmm, that is pretty bad if this fails again, and especially that we are not catching this in our tests. Just to make sure, if you delete the folder `/home/candersen/.aiida/repository/li-ion-conductors/container` and run the database migration command again, you get the same error every time?

> I think the issue is the removal of the `else`-clause [here](https://github.com/aiidateam/aiida-core/pull/4891/files#diff-414702fc1e9c1f26c4adeeb124de237e15be7c7d3824326cd778078bf8515bc2R139), although I was originally of the impression that #4891 would fix and remove the need for having this `else`-clause.

I removed this intentionally, because now the `uuid` property should return `None` if the repo is not initialised, and this is properly tested on the `Repository` class level.

mbercx:
I think perhaps the `Profile.get_repository()` method should not initialise the repository?

CasperWA:
> Hmm, that is pretty bad if this fails again, and especially that we are not catching this in our tests. Just to make sure, if you delete the folder `/home/candersen/.aiida/repository/li-ion-conductors/container` and run the database migration command again, you get the same error every time?

Yes. I also updated the OP to reflect this.

[...]

PR #4898, opened 2021-04-30
Database schema: set `{}` as default for `DbNode.repository_metadata`

### LINKED ISSUE #4893: Use an empty dictionary as the default for the `repository_metadata` column of `DbNode`
Currently it is `null`. The front-end ORM `Node.repository_metadata` returns an empty dictionary if the database value is `None` so one always gets the same type. However, if code goes around the front-end ORM, such as by using the `QueryBuilder` or what the import export functionality currently does, it needs to start checking the type.

fixes #4892

sphuber:
Fixes #4893 

Initially, when the column `repository_metadata` was added to the
`DbNode` table, it was set to be nullable since a significant number of
nodes will not actually have any files and therefore this value would be
empty. However, to prevent clients having to deal with a null-value, the
front-end ORM `Node.repository_metadata` property would return an empty
dictionary in this case, such that the return type is always a
dictionary. The main argument was that this would prevent unnecessary
bytes from being stored in the database.

However, a bug surfaced where some code expected a dictionary for the
`repository_metadata` but got `None`. This particular instance was in
the import code, which circumvents the ORM and went straight to the
database. This is of course undesirable, but it also happens through the
`QueryBuilder` that doesn't transform the returned attributes of
entities through the ORM interface. Given that there are a number of
layers from the ORM to the database, making sure that the typing across
all layers is consistent would be tricky and prone to more bugs. The
most secure solution is to simply set an empty dict as the default on
the database level. The added cost to the database size should still be
minimal and so is an acceptable downside to the increased stability of
the code.

Note that the column in the model is declared both with a server default
as well as a default on the ORM level. The reason is that the server
default is required for the migration. If the column were to be added
without the default, existing rows would violate the non-nullable
clause. For consistency, the server default is also added to the table
column declaration. The ORM default is necessary to guarantee that an
empty dictionary is set on a new `DbNode` instance when it is created.
SqlAlchemy cannot execute the server default and so would leave it as
`None`, but we require that even for unstored instances, the value
defaults to an empty dictionary.


mbercx (approved):
Thanks @sphuber! It all looks good, but I cannot field test the migration because of #4897. Will do some proper tests once that is fixed, but let's merge this for now.

sphuber:
I am almost ready with a fix for #4897 but it took quite some more rewriting than expected. I found another fundamentally problematic assumption in the code that went by unnoticed until the recent changes.

Issue #4892, opened 2021-04-29
Repository: `verdi archive create` fails due to empty `repository_metadata`


mbercx:
### Describe the bug

After migrating to the new file repository, I am unable to export an archive:

```console
$ verdi archive create -N 269 -- test.zip
|
EXPORT
--------------  ------------------------
Archive         test.zip
Format          JSON Zip (compression=8)
Export version  0.11

Inclusion rules
-----------------  ----
Include Comments   True
Include Logs       True

Traversal rules
---------------------------------  -----
Follow links input calc forwards   False
Follow links input calc backwards  True
Follow links create forwards       True
Follow links create backwards      True
Follow links return forwards       True
Follow links return backwards      False
Follow links input work forwards   False
Follow links input work backwards  True
Follow links call calc forwards    True
Follow links call calc backwards   True
Follow links call work forwards    True
Follow links call work backwards   True

Traceback (most recent call last):                                                                                                                                                                                                            
  File "/home/mbercx/.virtualenvs/aiida-sirius/bin/verdi", line 12, in <module>
    sys.exit(verdi())
  File "/home/mbercx/.virtualenvs/aiida-sirius/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/mbercx/.virtualenvs/aiida-sirius/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/mbercx/.virtualenvs/aiida-sirius/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mbercx/.virtualenvs/aiida-sirius/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mbercx/.virtualenvs/aiida-sirius/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mbercx/.virtualenvs/aiida-sirius/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/mbercx/envs/aiida-sirius/code/aiida-core/aiida/cmdline/utils/decorators.py", line 74, in wrapper
    return wrapped(*args, **kwargs)
[...]

PR #4900, opened 2021-04-30
Profile: only initialise the repository during the migration

### LINKED ISSUE #4897: Cannot migrate due to `container` folder being present

sphuber:
Fixes #4897 

The new disk object store migration that will ship with `v2.0` requires
to be initialised once and only once, and it will generate the necessary
folders and configuration file. This process was being done in the
method `Profile.get_repository` guarded by a check in case it was
already initialised.

The problem with this was that the container could be initialised too
early during the early stages of a profile setup. As soon as the
repository was fetched, it would be initialised generating, among other
things, the UUID. This would then trigger the check that the database
contained the same UUID, which would of course fail, since the database
was empty. This could have been fixed by ignoring the check at this
point, but the real problem is that the repository should not be
initialised at this point. The only point at which the repo should be
initialised is during the corresponding database migration that
introduced the disk object store repository. Both for existing as well
as for new profiles, they will go through this migration and so it and
it alone should be responsible for initialising the repository.

This approach did create problems for the unittests though, as they
would sometimes clean the repository. It would this not by just removing
the contents, but it would delete the entire container. This meant it
had to be recreated, but since in normal operations this only happens
during the migration (which also during tests only happens once, unless
maybe during the migration tests themselves) and so an error would be
raised that the repository is not initialised. The solution is to
reinitialise a new repo as soon as the old one was destroyed. Currently
this is done by simply deleting the folder on disk and reinitialising an
entire new instance. In the future, it would be better if the existing
container could be kept and its contents could simply be dropped, but
this would require a feature in the `disk-objectstore` library.

### DIFF
--- a/aiida/backends/general/migrations/utils.py
+++ b/aiida/backends/general/migrations/utils.py
@@ -153,19 +153,9 @@ def migrate_legacy_repository(node_count, shard=None):
     basepath = pathlib.Path(profile.repository_path) / 'repository' / 'node'
     container = Container(filepath)
 
-    if not basepath.is_dir():
-        # If the database is empty, this is a new profile and so it is normal the repo folder doesn't exist. We simply
-        # return as there is nothing to migrate
-        if profile.is_test_profile or node_count == 0:
-            return None, None
-
-        raise exceptions.DatabaseMigrationError(
-            f'the file repository `{basepath}` does not exist but the database is not empty, it contains {node_count} '
-            'nodes. Aborting the migration.'
-        )
-
-    # When calling this function multiple times, once for each shard, we should only check whether the container has
-    # already been initialized for the first shard.
+    # The new container needs to be initialised. Since this function can be called multiple times, it shouldn't
+    # initialise the repository each time since it will delete all content. By checking the value of the shard, we
+    # should just initialise it for the first iteration.
     if shard is None or shard == '00':
         if container.is_initialised and not profile.is_test_profile:
             raise exceptions.DatabaseMigrationError(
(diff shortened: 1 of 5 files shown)


sphuber:
I haven't added a test yet that should prevent the same regression that we have twice now. But this should fix the bug and improves the code/logic in general. Will think of a test soon, but this way the code can already be reviewed

CasperWA:
I'll test this now.

sphuber:
> I'll test this now (for a small db only).
> 
> Edit: Done. Tested for a DB I had. It represents a part of Thibault's old data.
> I got an error after migration, but not related to the (non-)existence of the `/container` folder.
> Please see the Google Doc for a summary of the whole error message.

Thanks a lot for the testing @CasperWA . As responded in that document, the problem is a new one and unrelated to the issue that this PR fixes. Since you were able to migrate at all, #4897 can be considered fixed by this. We should investigate more how the failure of your test is possible at all. For reference, the migration that changes the `name` column to `label` for the `DbComputer` table fails (for the SqlAlchemy backend) because apparently there are `null` rows. The column is and was defined as:
```
label = Column(String(255), unique=True, nullable=False)
```
So it should not have been able to insert `null` values anyway. We should look if there ever was a time when it was `nullable=True` and people changed it without a migration, but then again, it should have failed at that time I think.

sphuber:
I think this can be merged, since it seems to fix the specific issue it aims to address. The other failures mentioned by @CasperWA are edge-cases of certain databases that are failures of other specific migrations, which should be addressed in a separate issue. @mbercx would you be able to review this PR?

mbercx (approved):
Changes look good, thanks @sphuber! I've also tested the migration for a ~30k node database, and it seems to have gone smoothly. Also tested:

* Querying ✅
* Group operations ✅
* Running computations ✅
* `verdi archive create` - single `StructureData` (to see if #4892 is truly fixed) ✅
* `verdi archive create` - `Group` with `WorkChainNodes` ✅

PR #4901, opened 2021-05-01
`Config`: add the `delete_profile` method

### LINKED ISSUE #695: Use general function for deleting AiiDA file repository
One of the things that the `profile_delete` function in [aiida/cmdline/commands/profile.py](https://github.com/aiidateam/aiida_core/blob/develop/aiida/cmdline/commands/profile.py) needs to remove is the aiida file repository.

Once we support different modes of file storage (file system/object store/...) we should call the general-purpose function here.

sphuber:
Fixes #695

The class already has `remove_profile` that merely removes the profile
from the index and doesn't actually delete its associated resources. The
new `delete_profile` method will by default also delete the repository
and database that are configured for the profile, although this behavior
can be controlled with the `include_database` and `include_repository`
arguments.

The implementation of `verdi profile delete` is update to use this new
functionality instead of the loose functions that were defined in the
`aiida.manage.configuration.setup` module. The latter, since it was not
part of the public API and was only used by this CLI command, has been
removed from the codebase.

Note that the implementation for the deletion of the database is
slightly different. Not only does `Config.delete_profile` not contain
any echoing of information, as this should be done in the CLI command
and not in the generic API, it also does not attempt to delete the
database user. The old function `delete_database` did delete the
database user configured for the profile, as long as no other profiles
in the config used the same. This is dangerous, however, because the
same database user can be used by a program outside of AiiDA and we
don't own those users necessarily.


sphuber:
Need to still add tests for the new `RepositoryBackend.erase` methods, but the rest of the code can already be reviewed

aiida/cmdline/commands/cmd_profile.py
```diff
@@ -92,22 +95,36 @@ def profile_setdefault(profile):
...
-    Delete one or more profiles.
+@click.pass_context
+def profile_delete(ctx, force, include_config, include_database, include_repository, profiles):
+    """Delete one or more profiles.
 
-    You can specify more profile names (separated by spaces).
-    These will be removed from the aiida config file,
-    and the associated databases and file repositories will also be removed.
+    The PROFILES argument takes one or multiple profile names that will be deleted. Deletion here means that the profile
+    will be removed including its file repository and database and all their content. The various options can be used
```
ltalirz, line 103: (minor) I would say that removing the file repository and the database already makes it sufficiently clear that this includes their content. not important though

aiida/cmdline/commands/cmd_profile.py
```diff
@@ -92,22 +95,36 @@ def profile_setdefault(profile):
...
-        echo.echo_info(f"Deleting profile '{profile.name}'")
-        delete_profile(
-            profile,
-            non_interactive=force,
-            include_db=include_db,
-            include_repository=include_repository,
-            include_config=include_config
+
+        if not include_database and not include_repository:
+            deleting = '.'
```
ltalirz, line 112: Since deleting the db and the repo is the default behavior (and rightfully so), I think you actually want to inform the user about things that are *not* being deleted.

aiida/repository/repository.py
```diff
@@ -417,6 +417,14 @@ def delete_object(self, path: FilePath, hard_delete: bool = False):
         directory = self.get_directory(path.parent)
         directory.objects.pop(path.name)
 
+    def delete(self):
+        """Delete the entire repository.
```
ltalirz, line 421:
```suggestion
        """Delete the repository.
```

aiida/repository/repository.py
```diff
@@ -417,6 +417,14 @@ def delete_object(self, path: FilePath, hard_delete: bool = False):
...
+
+        .. important:: this will not just delete the contents of the repository but also the repository itself. For
```
ltalirz, line 423:
```suggestion
        .. important:: This will not just empty the contents of the repository but remove any trace of the repository from the file system. For
```

aiida/repository/repository.py
```diff
@@ -417,6 +417,14 @@ def delete_object(self, path: FilePath, hard_delete: bool = False):
...
+            example, if the repository is a folder on disk, the folder itself will potentially be deleted entirely.
```
ltalirz, line 424:
```suggestion
            example, if the repository is stored inside a folder on disk, the folder may be deleted.
```

ltalirz (commented):
thanks @sphuber !

> The old function `delete_database` did delete the
> database user configured for the profile, as long as no other profiles
> in the config used the same. This is dangerous, however, because the
> same database user can be used by a program outside of AiiDA and we
> don't own those users necessarily.

While this is theoretically possible, I have not heard of a single instance of this occurring.
On the other hand, not removing the user means that AiiDA is not able to properly clean up after itself - e.g. if the password of that user is leaked, then people could still access the postgres instance through it many years later.

I'm not saying that it is a black & white situation, but I would lean towards removing the db user as before.

aiida/repository/repository.py
sphuber: Cannot say file system because this `Repository` class can service any repo, also an object store for example.

aiida/cmdline/commands/cmd_profile.py
sphuber: Done

sphuber:
> While this is theoretically possible, I have not heard of a single instance of this occurring.
> On the other hand, not removing the user means that AiiDA is not able to properly clean up after itself - e.g. if the password of that user is leaked, then people could still access the postgres instance through it many years later.
> I'm not saying that it is a black & white situation, but I would lean towards removing the db user as before.

Any profile that does not get installed with `quicksetup` will be an example of this, and even for `quicksetup` you can specify an existing database. Configuring a profile with an existing database is not an edge-case and is prominently featured in the documentation. We simply cannot assume that the database user is only used by the current AiiDA installation. Even if you have multiple AiiDA installations on the same machine, they could all use the same user and if you delete all profiles of one instance, based on the old logic it would try to delete the user, even though it is still in use.

ltalirz:
What I meant is that I haven't heard of a single complaint about AiiDA deleting a database user that was still supposed to be used (and the code was in there for a long time).

I'd be ok to keep the functionality and have it be turned off by default... I would personally use it every time.

sphuber:
@ltalirz I have added the `--include-database-user/--skip-database-user` option

ltalirz (approved):
thanks @sphuber 

I see that the command will now try to delete the db user, irrespective of whether the database user is used by other profiles.
For `quicksetup` users that makes using this option a bit less convenient... but given that it needs to be turned on manually I guess it's ok.

I'd just ask you to update the commit message to reflect this.

aiida/cmdline/commands/cmd_profile.py
```diff
@@ -92,22 +102,45 @@ def profile_setdefault(profile):
...
+            'database user': include_database_user,
+            'file repository': include_repository
+        }
+
+        if not all(includes.values()):
+            message_suffix = ', excluding: '
+            excludes = []
+            for include_label, include in includes.items():
+                if not include:
+                    excludes.append(include_label)
```
ltalirz, line 128: Or just
```suggestion
            excludes = [ label for label in includes if not includes[label] ]
```

sphuber: Hah, of course. I am definitely including this, it's way better. Thanks :+1:

sphuber:
> I'd just ask you to update the commit message to reflect this.

Done, thanks for the review and comments.

ltalirz (approved):

PR #4902, opened 2021-05-03
docs: add hints on daemon worker slots

ltalirz:
* Add documentation of how to increase the slots per daemon worker.
 * Add section on how to use the rabbitmq dashboard
 * Reorder/group performance tuning tips


docs/source/howto/installation.rst
```diff
@@ -314,7 +314,45 @@ Tuning performance
...
 However, optimal performance at that scale might require some tweaks to the AiiDA configuration to balance the CPU and disk load.
-Here are a few general tips that might improve the AiiDA performance:
+
+Here are a few tips for tuning AiiDA performance:
+
+
+    .. dropdown:: Increase the number of daemon workers
+
+        By default, the AiiDA daemon only uses a single worker, i.e. a single operating system process.
+        If ``verdi daemon status`` shows the daemon worker constantly at high CPU usage, you can use ``verdi daemon incr X`` to add ``X`` parallel daemon workers.
```
sphuber, line 324:
```suggestion
        If ``verdi daemon status`` shows the daemon worker constantly at high CPU usage, you can use ``verdi daemon incr X`` to add ``X`` parallel daemon workers, where ``X`` is a positive integer.
```

docs/source/howto/installation.rst
```diff
@@ -314,7 +314,45 @@ Tuning performance
...
+
+        Keep in mind that other processes need to run on your computer (e.g. rabbitmq, the PostgreSQL database, ...), i.e. you likely want to stop increasing the number of workers *before* you reach the number of cores of your CPU.
+
+        To make the change permanent, set
+        ::
+
+            verdi config set daemon.default_workers 5
```
sphuber, line 331:
```suggestion
            verdi config set daemon.default_workers 5
```

This setting ensures that the next time you (re)start your daemon, it will start with 5 workers instead of 1.

docs/source/howto/installation.rst
```diff
@@ -314,7 +314,45 @@ Tuning performance
...
+
+    .. dropdown:: Increase the number of daemon worker slots
+
+
+        By default, each daemon worker accepts only 200 tasks at a time.
+        If ``verdi daemon status`` constantly warns about a high percentage of the available daemon worker slots being used, you can increase the number of tasks handled by each daemon worker (thus increasing the workload per worker).
```
sphuber, line 337: I think this is in `verdi process list` and not `verdi daemon status`.
```suggestion
        If ``verdi process list`` warns about a high percentage of the available daemon worker slots being used, you can increase the number of tasks handled by each daemon worker (thus increasing the workload per worker).
```

sphuber: Maybe we should say this is an alternative to running more daemon workers if you don't have enough CPUs, but that of  course each machine will have a max at some point and we cannot increase infinitely. It is difficult if not impossible to give settings that work for everyone, because some people might try to run on a small 2 CPU laptop and others have a dedicated 64 CPU workstation or server. I think this is important to convey somehow

docs/source/howto/installation.rst
```diff
@@ -314,7 +314,45 @@ Tuning performance
...
+
+Here are a few tips for tuning AiiDA performance:
+
+
+    .. dropdown:: Increase the number of daemon workers
+
+        By default, the AiiDA daemon only uses a single worker, i.e. a single operating system process.
+        If ``verdi daemon status`` shows the daemon worker constantly at high CPU usage, you can use ``verdi daemon incr X`` to add ``X`` parallel daemon workers.
+
+        Keep in mind that other processes need to run on your computer (e.g. rabbitmq, the PostgreSQL database, ...), i.e. you likely want to stop increasing the number of workers *before* you reach the number of cores of your CPU.
```
sphuber, line 326:
```suggestion
        Keep in mind that other processes also need to run on your computer (e.g. rabbitmq, the PostgreSQL database and any other programs you have running), therefore you probably do not want to have the same amount or even more workers than your machine has CPUs.
```

docs/source/howto/installation.rst
```diff
@@ -314,7 +314,45 @@ Tuning performance
...
+
+    .. dropdown:: Analyze engine performance
+
+        If you're observing slow performance of the AiiDA engine, the rabbitmq message rate is a useful indicator to check.
+
+        rabbitmq comes with a built-in `management plugin <https://www.rabbitmq.com/management.html>`_ that can be enabled via something like::
+
+            sudo rabbitmq-plugins enable rabbitmq_management
+
+        Then, navigate to http://localhost:15672/ and log in with ``guest``/``guest``.
```
sphuber, line 354: I think it might be good to have this info somewhere on how to get the panel up and running, but as is, I don't think it will be super helpful to users. In any case, I don't think we should say that this is per se a useful indicator since we ourselves are not super sure on how to interpret the results. Take as example your recent issues on the "bursts" of messages in RabbitMQ which has not been reproduced yet.

docs/source/howto/installation.rst
```diff
@@ -314,7 +314,45 @@ Tuning performance
 
 AiiDA supports running hundreds of thousands of calculations and graphs with millions of nodes.
 However, optimal performance at that scale might require some tweaks to the AiiDA configuration to balance the CPU and disk load.
-Here are a few general tips that might improve the AiiDA performance:
+
+Here are a few tips for tuning AiiDA performance:
```
sphuber, line 318:
```suggestion
Here are a few tips for tuning AiiDA's performance:
```

sphuber (changes requested):
Thanks @ltalirz , have added some suggestions and minor corrections

docs/source/howto/installation.rst
ltalirz, line 331:
```suggestion
        To make sure the daemon will use 5 workers instead of 1 next time you (re)start it, do
        ::

            verdi config set daemon.default_workers 5
```

docs/source/howto/installation.rst
ltalirz: thanks, happy to discuss but here I prefer my version

docs/source/howto/installation.rst
```diff
@@ -314,7 +314,45 @@ Tuning performance
...
+
+        To make the change permanent, set
+        ::
+
+            verdi config set daemon.default_workers 5
+
+    .. dropdown:: Increase the number of daemon worker slots
+
+
+        By default, each daemon worker accepts only 200 tasks at a time.
```
ltalirz, line 336:
```suggestion
        By default, each daemon worker accepts only 200 tasks at a time, which is a somewhat conservative choice.
```

docs/source/howto/installation.rst
ltalirz, line 337:
```suggestion
        If ``verdi process list`` warns about a high percentage of the daemon worker slots being used, but the CPU usage of the individual workers in ``verdi daemon status`` is not already close to 100% percent, you can increase the number of tasks handled by each daemon worker (thus increasing the workload per worker).
```

docs/source/howto/installation.rst
ltalirz: I agree that it is a bit less practical than the other pieces of advice in the list (it doesn't tell you which know to turn).

Still, I think it's an important piece of information for performance tuning and I didn't find a better place to put it where it would be seen.
Happy to move it somewhere else if you have an idea.

> Take as example your recent issues on the "bursts" of messages in RabbitMQ which has not been reproduced yet.

It's true that we don't yet know what the underlying issue is and how to fix it, but I think the observation is quite clear.

docs/source/howto/installation.rst
```diff
@@ -314,7 +314,45 @@ Tuning performance
...
+        If ``verdi process list`` warns about a high percentage of the daemon worker slots being used, but the CPU usage of the individual workers in ``verdi daemon status`` is not already close to 100%, you can increase the number of tasks handled by each daemon worker (thus increasing the workload per worker).
+        Increasing it to 1000 should typically work.
```
ltalirz, line 338:
```suggestion
        Increasing it to 1000 may work, depending on your hardware and the type of workload in the Processes you are running.
```

docs/source/howto/installation.rst
ltalirz: Yes, it also depends on the Processes you are running (e.g. if they involve significant computation or queries to the DB, ...).
I've tried to incorporate this now

ltalirz:
thanks for the review!

docs/source/howto/installation.rst
sphuber: It just feels really off in a list with concrete action points to _change_ the performance. How about we add a section "Inspecting your installation" on the same level as and just after "Tuning performance", where you put this information instead. In this case I also wouldn't frame it in the context of analysing the performance, but keep it in general. So far we have also needed this to debug problems with missing tasks, so I would just introduce as a useful tool to inspect what is going on with an isntallation in the case of problems, bugs/performance or otherwise.

ramirezfranciscof (approved):
Hey @ltalirz , thanks for the additions! I'll let you decide what to do with the [last comment](https://github.com/aiidateam/aiida-core/pull/4902/files#r625911853) that @sphuber left (which sounded more or less optional) and then update/merge.

ltalirz:
Thanks @ramirezfranciscof - I've now moved the rabbitmq suggestion into a separate list.
I still kept it under the same heading since in my opinion it's very related to performance tuning (the alternative place would have been troubleshooting).

ramirezfranciscof (dismissed):
Thanks @ltalirz and @sphuber , I'll merge it after the build test completes!

PR #4903, opened 2021-05-03
docs: make query docs easier to find

### LINKED ISSUE #4879: Docs: unclear path to querybuilder documentation

I was looking for a place to link to that would explain how the querybuilder works.

The basic tutorial does not contain instructions on how to use the querybuilder. When browsing through the "howto" section, one sees "how to explore the provenance graph" but that actually already assumes one knows the basics of how to get nodes from the DB.

The place I would actually have liked to find would have been "How to work with data => Finding and querying for data" but I didn't find it looking & searching through the documentation web site. I only found it via `git grep` on the repository.

At the very least, the "how to explore the provenance graph" needs to start with a pointer to "Finding and querying for data".

Given that querying is such a big topic, I would also propose to make "How to find and query for data" a top-level howto section, and rename "How to work with data" to something like "How to manipulate and organize data" (or keep the name unchanged if that is preferred).

mentioning @mbercx and @csadorf for input

ltalirz:
fix #4879 

With the Materials Cloud Archive adding an option to explore AiiDA
archive files on renkulab.io, finding and querying for data will be the
*first*  thing that some new AiiDA users will want to do.

Currently, the relevant documentation was hidden inside "How to work
with data", which made it difficult to find by browsing through the web
page (the page was also ranked low when searching for "query" in the
search field).

This PR makes "How to find and query for data" a top-level how-to
section.
The "How to explore the provenance graph" section becomes a subsection
of "How to find and query for data", entitled "Shortcuts" - indeed, this
section deals with shortcuts in the AiiDA python API that save users
from writing QueryBuilder queries for very common generic queries.


docs/source/redirects.txt
```diff
@@ -4,6 +4,7 @@ get_started/index.rst intro/get_started.rst
 get_started/computers.rst howto/run_codes.rst
 get_started/codes.rst howto/run_codes.rst
 howto/plugins.rst howto/plugins_develop.rst
+howto/exploring.rst howto/query.rst
```
csadorf, line 7: Nice! I was about to write a comment about this and then noticed you had already added the redirects. 😄

docs/source/howto/query.rst
```diff
@@ -0,0 +1,415 @@
+.. _how-to:query:
+
+*******************************
+How to find and query for data
+*******************************
```
csadorf, line 5: I did not review this section in detail assuming that it is largely similar to the original section in "exploring". But please let me know if this is not the case and I should read this more thoroughly.

docs/source/topics/data_types.rst
```diff
@@ -26,7 +26,7 @@ The different data types can be accessed through the :py:func:`~aiida.plugins.fa
 
 A list of all the data entry points can be obtain running the command ``verdi plugin list aiida.data``.
 
-For all data types, you can follow the link to the corresponding data class in the API reference to read more about the class and its methods. We also detail what is stored in the database (mostly as attributes, so the information can be easily queried e.g. with the :ref:`QueryBuilder <how-to:data:find>`) and what is stored as a raw file in the AiiDA file repository (providing access to the file contents, but not efficiently queryable: this is useful for e.g. big data files that don't need to be queried for).
+For all data types, you can follow the link to the corresponding data class in the API reference to read more about the class and its methods. We also detail what is stored in the database (mostly as attributes, so the information can be easily queried e.g. with the :ref:`QueryBuilder <how-to:query>`) and what is stored as a raw file in the AiiDA file repository (providing access to the file contents, but not efficiently queryable: this is useful for e.g. big data files that don't need to be queried for).
```
csadorf, line 29: I am honestly a bit puzzled why all of these sections have multiple sentences per line. We may need to add a linter for this.

csadorf (commented):
Looks great, I have one question and them I am happy to approve this!

docs/source/topics/data_types.rst
ltalirz: > I am honestly a bit puzzled why all of these sections have multiple sentences per line. 

I don't know... perhaps copied from the old documentation?

> We may need to add a linter for this.

Feel free to add such a linter if that exists - might not be super straightforward to detect (dot/exclamation mark/... followed by upper case?) but probably you'll catch most cases.
I would say that should probably go into a separate PR.

csadorf: Oh yeah, this would need to go into a separate PR for sure. Just something I noticed while reviewing.

csadorf:
@ltalirz I assume this is ready for merge / final approval? I'm asking, because you neither responded to my question as to whether there was significant change to the text nor re-requested review.

ltalirz:
Yes, sorry - I understood that I should let you know in case you needed to give it a more careful read (and by resolving the point I tried to indicate that this isn't necessary).
Please re-review!

csadorf (approved):

PR #4908, opened 2021-05-04
CLI: set `localhost` as default for database hostname in `verdi setup`

### LINKED ISSUE #4907: Use `localhost` as default for `--database-hostname` in `verdi setup`
Currently it sets no default, but often one wants to configure `localhost`. The problem is that a default *is* being set, but it is taken from `pgsu.DEFAULT_DSN` which defines `None` for the hostname.

sphuber:
Fixes #4907 

The default was actually being defined on the option, however, it was
taken from the `pgsu.DEFAULT_DSN` dictionary, which defines the database
hostname to be `None`. Still, 9 out of 10 times the database is on the
localhost so not having this as a default is kind of annoying and
unnecessary.

### DIFF
--- a/aiida/cmdline/params/options/__init__.py
+++ b/aiida/cmdline/params/options/__init__.py
@@ -249,7 +249,7 @@ DB_HOST = OverridableOption(
     '--db-host',
     type=types.HostnameType(),
     help='Database server host. Leave empty for "peer" authentication.',
-    default=DEFAULT_DBINFO['host']
+    default='localhost'
 )
 
 DB_PORT = OverridableOption(
--- a/aiida/cmdline/params/options/commands/setup.py
+++ b/aiida/cmdline/params/options/commands/setup.py
@@ -259,7 +259,7 @@ SETUP_DATABASE_BACKEND = QUICKSETUP_DATABASE_BACKEND.clone(
 
 SETUP_DATABASE_HOSTNAME = QUICKSETUP_DATABASE_HOSTNAME.clone(
     prompt='Database host',
-    contextual_default=functools.partial(get_profile_attribute_default, ('database_hostname', DEFAULT_DBINFO['host'])),
+    contextual_default=functools.partial(get_profile_attribute_default, ('database_hostname', 'localhost')),
     cls=options.interactive.InteractiveOption
 )
 


ltalirz:
thanks @sphuber - to my knowledge, there actually is a difference between `None` and `localhost`, see
https://github.com/aiidateam/pgsu/blob/43d8d305053e9f6a1352295bf9a3f6955881d5ba/pgsu/__init__.py#L17

In particular, I believe that method `localhost` will not work with the default postgres setup on ubuntu that goes via the `local` connection. 

What is annoying about the default?

sphuber:
> In particular, I believe that `localhost` will not work with the default postgres setup on ubuntu that goes via the `local` connection.

Maybe circumstantial evidence but I have been using `localhost` on Ubuntu 18.04 and 20.04 since forever without any issues.

> What is annoying about the default?

I always have to specify `localhost` when creating a new profile. Since most cases would actually use this, it seems to me like there is not really a reason to _not_ have that as a default.

ltalirz:
> > In particular, I believe that `localhost` will not work with the default postgres setup on ubuntu that goes via the `local` connection.
> 
> Maybe circumstantial evidence but I have been using `localhost` on Ubuntu 18.04 and 20.04 since forever without any issues.

I might be wrong (I'll check). Can you please share the uncommented lines of your `/etc/postgresql/10/main/pg_hba.conf` ?

ltalirz (approved):
On second thought, I think you are right.

`'localhost'` is the correct default there, while `None` is a good thing to try for pgsu (because it is the only way you can access the psql shell without a password).

You may want to include this information in the commit message.

sphuber:
```
local   all             postgres                                peer

# TYPE  DATABASE        USER            ADDRESS                 METHOD

# "local" is for Unix domain socket connections only
local   all             all                                     peer
# IPv4 local connections:
host    all             all             127.0.0.1/32            md5
# IPv6 local connections:
host    all             all             ::1/128                 md5
# Allow replication connections from localhost, by a user with the
# replication privilege.
local   replication     all                                     peer
host    replication     all             127.0.0.1/32            md5
host    replication     all             ::1/128                 md5
```
This is the default config of postgres-12 on Ubuntu 20.04

sphuber:
> You may want to include this information in the commit message.

Cheers, will do.

PR #4913, opened 2021-05-05
Remove remaining deprecated resources for v2.0

sphuber:
A previous commit already removed the majority of deprecated resources,
but these were missed. A few things of notice:

 * The `aiida.common.folders.RepositoryFolder` was never officially
   deprecated, but it was also not part of the official API. A check of
   plugins online shows that it is not in use, so should be fine to be
   removed regardless.

 * The `hook` argument of the `aiida.restapi.run_api.run_api` function
   had been properly deprecated, however, the `verdi restapi` command
   still exposed the `--no-hookup/--hookup` flag which was not
   officially deprecated. However, the deprecation warning would have
   been shown indirectly when `run_api` was called and `--no-hookup`
   really doesn't make sense anyway since it returns a Python object
   that is lost immediately once the command exist.


ltalirz (approved):
thanks @sphuber 

went through everything and all seems fine.

also checked that `RepositoryFolder` is not used by plugins on github

sphuber:
> went through everything and all seems fine.

As I expected, the tests are probably hanging, most likely because of the removal of the `--no-hookup` flag. This was being used by the `verdi restapi` unit test to not actually start the server. Since the command now _always_ starts the server, probably it is hanging there. I think I will have to mock the `run_api` method to be a no-op.

sphuber:
Thanks for the review @ltalirz . I fixed the tests, should be good now. Maybe just want to recheck the `test_restapi.py` file since I changed the tests there since the last review.

ltalirz (approved):
thanks for the fix - indeed, I should have caught that.

great that you managed to retain the test.

PR #4881, opened 2021-04-28
Remove all deprecated resources for `v2.0`

sphuber:
Fixes #2761

PR #4915, opened 2021-05-05
Fix the benchmark tests

sphuber:
They were still using the `silent` keyword for the import and export
functions, which have recently been removed causing the benchmark tests
to except.


sphuber:
@chrisjsewell what was the way to trigger the benchmark tests on a PR again?

chrisjsewell:
> what was the way to trigger the benchmark tests on a PR again?

I had to remove this, because GH actions removed the functionality I'd used 😒 

Isn't it also saying about missing `use_compression`? I hadn't got round to figuring out why that was the case

sphuber:
> Isn't it also saying about missing `use_compression`? I hadn't got round to figuring out why that was the case

It is still part of the constructor of the `WriterJsonZip` class, but I am not sure if it is properly redirected there.

sphuber:
Ah, but I see now it complains in the failing tests about both. Think we can safely remove that one as well.

chrisjsewell:
Ok yeh I think thats ok: https://github.com/aiidateam/aiida-core/pull/4881/files#r626747473

chrisjsewell (approved):

Issue #4321, opened 2020-08-26
`verdi`: add repository maintenance commands to operate on the repository


sphuber:
Once the new repository implementation will be merged, the disk object storage (DOS) will need some maintenance now and again to improve performance, that needs to be controllable from `verdi`. Operations include but are potentially not limited to:

* Packing all loose objects
* Deleting packed loose objects
* Repacking
* Deleting unreferenced objects
* Vacuuming the index database

Some of these operations can be executed during operation of AiiDA, whereas some can only be executed as a maintenance operation where the DOS cannot be used. There is also a difference in operation efficiency, where some can be very quick and efficient, whereas others (such as repacking) can potentially be very expensive. Currently the idea is that there should be one command that will be used for the majority of users that just needs to be called periodically to optimize the repository. However, it would be good to expose additional commands for the sub operations listed above such that it provides more control and for example can be called in CRON jobs.

Current proposal of subcommands for the new command `verdi repository`:

 * `backup`: perform an `rsync` of the DOS.
    - ? Should this perform `rsync` of the state as is, or should it automatically perform some maintenance operations first, or alternatively propose the user with these operations and ask for confirmation
 * `maintenance`: single-stop-shop command for most users. Will automatically or through prompting optimize the DOS in the most effective way possible.
 * `pack`: Pack all loose objects
 * `clean`: Delete packed loose objects, delete unreferenced objects and vacuum the database
    - `--delete-loose/--no-delete-loose`:
    - `--delete-unreferenced/--no-delete-unreferenced`:
    - `--vacuum-index/--vacuum-index`:
    - `--repack`: will fully repack the DOS, which is not always necessary and potentially expensive, but _can_ greatly improve performance

This is an incomplete proposal and names are in no way definitive. Let's discuss below and I will keep updating the design here once certain things become agreed upon.

sphuber:
Pinging @ramirezfranciscof and @giovannipizzi

giovannipizzi:
Do you prefer to add comments here, or that we modify directly the initial comment?
One comment is that I would move `--repack` as an option to `clean` - it's the most aggressive operation and needed to recover space after deleting/cleaning, so logically for me it's the 'last' of clean. I would keep pack instead only to pack loose and to things that can be executed while AiiDA is in use.

sphuber:
I would add comments here as new posts and once we agree I will update the OP to reflect the current design proposal

ramirezfranciscof:
[...]

Issue #4919, opened 2021-05-06
Proposal to add `DbFile` table


chrisjsewell:
In-line with my comments in https://github.com/aiidateam/aiida-core/issues/4321#issuecomment-833167997,
I feel that with the current implementation of `DbNode.repository_metadata` it is really quite difficult to perform any task on repository files, in an efficient manner.

If we had a separate table, something like:

```python
class DbFile(m.Model):
    node = m.ForeignKey('DbNode', on_delete=m.CASCADE)
	path = m.CharField(max_length=255, db_index=True)
    key = m.CharField(max_length=255)
    # also type if we need to store directories?
    class Meta:
        unique_together = (('node', 'path'),)
```

This would make it incredibly more performant and facile for file queries like:

- Generating the list of unique keys (i.e. hashes) in the database (required to clean the repository)
- Performing file system "like" operations that were previously possible when the files were stored on disk, e.g. searching for all files named `data-file.xml` 

@sphuber, @giovannipizzi, @ltalirz?

giovannipizzi:
Hi @chrisjsewell - indeed the table was one of our options for the implementation. 
Here I list the reasons (I remember of) for the current choice. Personally, I'm happy to re-evaluate it, but I want to note other operations that we should benchmark before deciding to switch.

- consistency at the DB level: with the current nested dictionary, it's very clear what are directories and what are files, and there is no possibility to store inconsistent data, e.g. things like: `a//b` (double slash), or twice the same entry with similar syntax `a//b` and `a/b`, or e.g. it would be possible to have a file without the parent folder (`a/b/c` without `a/b`) and other similar things. [This is something I would accept to compromise if really needed]
- Your proposal will of course the table would make queries more efficient (but note that you need to add a `db_index` also to hash, since often you will want to search if a given hash exists, e.g. before deciding to delete an object from the container). However, it's not clear (meaning: preliminary tests showed it to be inefficient, but I don't have the numbers so one would need to double check) if while storing objects, the additional cost of storing many rows (one per file at least) would be instead a big hit for performance (and similarly to get the list of all files in the repo, one would need to fetch many rows, that is often inefficient - at least if not done properly). Before deciding to switch, I'd like to see a comparison of performance also on these kind of operations that are quite common in daily operation in AiiDA.
[...]

PR #4922, opened 2021-05-06
✨ NEW: Add `Node.objects.iter_object_keys`

chrisjsewell:
This PR implements `Node.Collection.iter_object_keys` and `Node.Collection.iter_object_names`,
to retrieve all the repository object keys/names for the given `Node` subclass.
For example:

```python
from aiida.orm import Data
len(set(Data.objects.iter_object_keys()))
```

It is primarily intended to form part of the solution for #4321 (retrieving all object keys on the AiiDA DB, to decide what needs to be deleted from the object store).

I think also this is a good location in the API to expose this, in a generic way from which we can always change the implementation at a later date.

thoughts @sphuber, @giovannipizzi?

(If you agree I will add some tests)

### DIFF
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -84,6 +84,44 @@ class Node(Entity, NodeRepositoryMixin, EntityAttributesMixin, EntityExtrasMixin
 
             self._backend.nodes.delete(node_id)
 
+        def iter_object_keys(self, subclassing: bool = True) -> Iterator[str]:
+            """Iterate over all repository object keys for this ``Node`` class
+
+            .. note::
+                keys will not be deduplicated (wrap in a ``set`` to achieve this)
+            
+            :param bool subclassing: Whether to include subclasses of the given class
+            """
+            from aiida.repository import Repository
+            profile = get_manager().get_profile()
+            assert profile, "profile not loaded"
+            backend = profile.get_repository().backend
+            qb = QueryBuilder()
+            qb.append(self.entity_type, subclassing=subclassing, project=["repository_metadata"])
+            for metadata, in qb.iterall():
+                repo = Repository.from_serialized(backend=backend, serialized=metadata)
+                for hash_key in repo.get_hash_keys():
+                    yield hash_key
+
+        def iter_object_names(self, subclassing: bool = True) -> Iterator[str]:
+            """Iterate over all repository object names for this ``Node`` class
+
+            .. note::
+                names will not be deduplicated (wrap in a ``set`` to achieve this)
+
+            :param bool subclassing: Whether to include subclasses of the given class
+            """
+            from aiida.repository import Repository
+            profile = get_manager().get_profile()
+            assert profile, "profile not loaded"
+            backend = profile.get_repository().backend
+            qb = QueryBuilder()
+            qb.append(self.entity_type, subclassing=subclassing, project=["repository_metadata"])
+            for metadata, in qb.iterall():
+                repo = Repository.from_serialized(backend=backend, serialized=metadata)
+                for hash_key in repo.list_object_names():
+                    yield hash_key
+
     # This will be set by the metaclass call
     _logger: Optional[Logger] = None
 


giovannipizzi:
Thanks @chrisjsewell! Location seems possibly OK. However I'm wondering if we actually need the functionality to iterate only on a subset of nodes types. If this is not a requirement, in the end we just need some iterator over all hash keys; maybe there's a different place to put it (but I don't know where, so in case we can keep it here).

As a second comment, on my mid-size DB (~320k nodes, ~340k different hash keys, ~510k objects listed in the node repository_metadata), I get this timing:
```python
import time

def iter_object_keys():
    from aiida.repository import Repository

    profile = get_manager().get_profile()
    backend = profile.get_repository().backend

    query = QueryBuilder()
    query.append(Node, project=['repository_metadata'])
    for metadata, in query.iterall():
        repo = Repository.from_serialized(backend=backend, serialized=metadata)
        for hash_key in repo.get_hash_keys():
            yield hash_key

t = time.time()
b = set(iter_object_keys())
print(time.time() - t)
```
This runs in ~8.2s.

The following "hardcoded" logic would take only 2.1s:
```python
import aiida.backends.djsite.db.models as djmodels
import collections.abc
import time


def flatten_hashonly(dictionary):
    items = set()
    if not dictionary:
        return items
    for value in dictionary['o'].values():
        try:
            items.add(value['k'])
        except KeyError:
            items.update(flatten_hashonly(value))
    return items

t = time.time()
results = djmodels.DbNode.objects.values_list('repository_metadata', flat=True)
hashes = set()
for data in results:
    if data is not None:
        if data == {}:
            data = {'o': {}}
        hashes.update(flatten_hashonly(data))
print(len(hashes), time.time() - t)
```
(this might also give a sense on how long it takes to list all of them, I imagine this time just scales linearly with the size of the DB, see also discussions in #4321 and #4919).

I don't know why the first implementation (from this PR) is 4x slower (maybe the class creation?); I know the second snippet is quite hard-coded and with no checks; but maybe there is a way to implement a version that is a compromise between the two?
(If it's not obvious how to speed it up, I think we can go ahead and optimise later, anyway, as long as we're happy with the interface).

giovannipizzi:
A partially unrelated question for @sphuber: should the "correct" default be `{'o': {}}` rather than `{}`, unless I'm misunderstanding the syntax? (I know you've been already moving from `None` to `{}`)
Anyway I thought in my first version I had to patch `{}` to  `{'o': {}}`, but this does not seem to be the case, maybe I'm just misunderstanding the syntax.

chrisjsewell:
> I know the second snippet is quite hard-coded and with no checks; but maybe there is a way to implement a version that is a compromise between the two?

I can have a look, but yeh the point was I went for the "abstractually best" solution

giovannipizzi:
yes yes, I know. As I said, it's OK performance wise and we can improve later.
I'd still try to hear from @sphuber if he's better suggestions for the location (do we need filtering by type?)

chrisjsewell:
Note, if you just replace `Repository.from_serialized(backend=backend, serialized=metadata).get_hash_keys()` in my function, with your `flatten_hashonly(metadata)`, you get the same speed up

chrisjsewell:
> do we need filtering by type?

well its noce for summary statistics; do you gain anything by not having it?

giovannipizzi:
> Note, if you just replace `Repository.from_serialized(backend=backend, serialized=metadata).get_hash_keys()` in my function, with your `flatten_hashonly(metadata)`, you get the same speed up

You mean this?
```python
import time

def iter_object_keys():
    from aiida.repository import Repository

    profile = get_manager().get_profile()
    backend = profile.get_repository().backend

    query = QueryBuilder()
    query.append(Node, project=['repository_metadata'])
    for metadata, in query.iterall():
        for item in flatten_hashonly(metadata):
            yield item

t = time.time()
b = set(iter_object_keys())
print(time.time() - t)
```
For reference, I get 5.1s, so in between. But again I wouldn't block this on performance, it's still acceptable (and uses existing code), we can improve later.

> well its nice for summary statistics; do you gain anything by not having it?

Statistics is a good point. I'm just thinking whether by lifting that requirement we can put the logic somewhere else, in a more 'global' place (like a class method of `Repository` just to give an example - but I'm not convinced, maybe the place you suggest is the best place

chrisjsewell:
> like a class method of Repository just to give an example

err, I feel that would be kind of misleading, because obviously the output does not actually come from the repository

giovannipizzi:
I realise I've been unfair in my timings - due to some copy-paste from earlier code, in the faster one I'm using directly django rather than the QueryBuilder - this gives a closer timing of ~5.1s, much closer to the implementation in this PR:
```python
import aiida.backends.djsite.db.models as djmodels
import collections.abc
import time


def flatten_hashonly(dictionary):
    items = set()
    if not dictionary:
        return items
    for value in dictionary['o'].values():
        try:
            items.add(value['k'])
        except KeyError:
            items.update(flatten_hashonly(value))
    return items

t = time.time()
query = QueryBuilder()
query.append(Node, project=['repository_metadata'])
hashes = set()
for data, in query.iterall():
    hashes.update(flatten_hashonly(data))
print(len(hashes), time.time() - t)
```
Sorry about this.

Also, I see your point @chrisjsewell about not putting it in `Repository`

sphuber:
A few points:

* I agree that this shouldn't go in the `Repository` class as it explicitly does not have a connection to the database. It provides an interface to a repo on disk, but those repos are not expected to have a recording of the virtual hierarchy. This needs to be inserted into the `Repository` (through `from_serialized) and it will keep it in memory for the duration of its lifetime. But in AiiDA the actual hierarchy is stored in our database and so I wouldn't add methods to "retrieve" that on the `Repository` class.

* I am not sure why this needs to be defined centrally on the `Node` collection. Would this really be used anywhere else than the maintenance operation for the repository? Anyway, I won't stop this if you think it makes sense here. I would just recommend a slight name change because I was really confused by the name `iter_object_keys`. The fact that this is called as `Node.objects.iter_object_keys` you would think the `objects` is some hashmap and you would iterate over the keys that map the nodes in that collection. Maybe `iter_repository_object_keys` would be better? Then, I can see why we need the keys, in order to know which keys are no longer referenced, but why do we need the files? What would be the point of iterating over all filenames in the entire repository?

* The implementation of `iter_object_names` is incorrect. The `list_object_names` method is not recursive and merely lists the names of the files in the root directory of that node. You can pass a relative path as argument to get the names in that subdirectory. Note also that `list_object_names` returns the name of both files and directories.

chrisjsewell:
> The implementation of iter_object_names is incorrect. The list_object_names method is not recursive and merely lists the names of the files in the root directory of that node.

Oh yeh cheers, I forgot that, ... but this is certainly not evident in the names/docstrings

```python
    def list_objects(self, path: str = None) -> typing.List[File]:
        """Return a list of the objects contained in this repository sorted by name, optionally in given sub directory.

        :param path: the relative path where to store the object in the repository.
        """

    def list_object_names(self, path: str = None) -> typing.List[str]:
        """Return a sorted list of the object names contained in this repository, optionally in the given sub directory.

        :param path: the relative path where to store the object in the repository.
        """
```

which mirrors my comments in https://github.com/aiidateam/aiida-core/issues/4321#issuecomment-834220354, that I feel it would be better to have a `pathlib.Path` like API

dev-zero:
> ```python
> def flatten_hashonly(dictionary):
>     items = set()
>     if not dictionary:
>         return items
>     for value in dictionary['o'].values():
>         try:
>             items.add(value['k'])
>         except KeyError:
>             items.update(flatten_hashonly(value))
>     return items
> ```

As a general notion about recursive functions in Python: Python does not do Tail Call Optimization hence I've come to avoid recursive functions in performance critical code (or code where the depth is not predetermined to avoid hitting the `RuntimeError: maximum recursion depth exceeded`). There are several ways to do it, my favourite is usually an explicit stack which exploits the fact that Python uses references for anything but simple types. Untested, but the following is roughly what I mean:

```python
def flatten_hashonly(dictionary):
    items = set()

    if not dictionary:
        return items

    stack = [dictionary]

    while stack:
        value = stack.pop()

        try:
            items.add(value['k'])
        except KeyError:
            stack += list(dictionary['o'].values())

    return items
```

Maybe this could be of use here as well.

giovannipizzi:
Thanks @dev-zero 
The correct code should be this:
```python
def flatten_hashonly_nonrec(dictionary):
    items = set()

    if not dictionary:
        return items

    stack = [dictionary]

    while stack:
        values = stack.pop()['o'].values()

        for value in values:
            try:
                items.add(value['k'])
            except KeyError:
                stack += [value]

    return items
```
(One important note where it took a moment for me to realise: if `d` is a dictionary, `[d]` is a list with d as an entry, while `list(d)` is a list of its keys).

I checked and the timing seems the same as the recursive version, but I agree that a non-recursive version is better to avoid the recursion error exception

chrisjsewell:
Yep fair 👍

chrisjsewell:
I'm merging this soon

sphuber:
I'll leave the check on the recursive implementation to @giovannipizzi or @dev-zero since they commented on this. I just have to note that I think you haven't addressed my comment on the naming for `iter_object_keys`. When defined on the collection, it leads to `Node.objects.iter_object_keys()` and that strongly suggests you are somehow iterating over the objects (i.e. nodes) in the collection. At no point is it clear that it is dealing with objects from the repository of the node. I really think we should name it `iter_repository_object_keys`.

sphuber (changes requested):
I'll leave the check on the recursive implementation to @giovannipizzi or @dev-zero since they commented on this. I just have to note that I think you haven't addressed my comment on the naming for iter_object_keys. When defined on the collection, it leads to Node.objects.iter_object_keys() and that strongly suggests you are somehow iterating over the objects (i.e. nodes) in the collection. At no point is it clear that it is dealing with objects from the repository of the node. I really think we should name it iter_repository_object_keys.

aiida/orm/nodes/node.py
```diff
@@ -82,6 +82,26 @@ def delete(self, pk: int) -> None:
 
         self._backend.nodes.delete(pk)
 
+    def iter_object_keys(self,
```
chrisjsewell, line 85:
```suggestion
    def iter_repo_keys(self,
```

tests/orm/node/test_node.py
```diff
@@ -966,3 +967,22 @@ def test_uuid_equality_fallback():
     assert hash(node_a) == hash(node_b)
     assert hash(node_a) != hash(node_0)
     assert hash(node_b) != hash(node_0)
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_iter_object_keys():
```
chrisjsewell, line 973:
```suggestion
def test_iter_repo_keys():
```

tests/orm/node/test_node.py
```diff
@@ -966,3 +967,22 @@ def test_uuid_equality_fallback():
     assert hash(node_a) == hash(node_b)
     assert hash(node_a) != hash(node_0)
     assert hash(node_b) != hash(node_0)
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_iter_repo_keys():
+    """Test the ``iter_object_keys`` method."""
```
chrisjsewell, line 974:
```suggestion
    """Test the ``iter_repo_keys`` method."""
```

tests/orm/node/test_node.py
```diff
@@ -966,3 +967,22 @@ def test_uuid_equality_fallback():
...
+    data1 = Data()
+    data1.put_object_from_filelike(BytesIO(b'value1'), 'key1')
+    data1.put_object_from_filelike(BytesIO(b'value1'), 'key2')
+    data1.put_object_from_filelike(BytesIO(b'value3'), 'folder/key3')
+    data1.store()
+    data2 = Data()
+    data2.put_object_from_filelike(BytesIO(b'value1'), 'key1')
+    data2.put_object_from_filelike(BytesIO(b'value4'), 'key2')
+    data2.store()
+    assert set(Data.objects.iter_object_keys()) == {
```
chrisjsewell, line 984:
```suggestion
    assert set(Data.objects.iter_repo_keys()) == {
```

chrisjsewell:
> I think you haven't addressed my comment on the naming for iter_object_keys

cheers, changed the name to `iter_repo_keys`

PR #4923, opened 2021-05-06
👌 IMPROVE: `Repository.hash()` speed

### LINKED ISSUE #4918: `Repository.hash` seems very inefficient
https://github.com/aiidateam/aiida-core/blob/5f87661e9099b1e8875254df8b133dd43b58c5bc/aiida/repository/repository.py#L96-L110

@giovannipizzi @sphuber Why do we need to read every file and regenerate their hash?
why not just re-use the hashes of the files that are already stored e.g.:

```python
    def hash(self) -> str:
        """Generate a hash of the repository's contents.

        :return: the hash representing the contents of the repository.
        """
        objects = {}
        for root, dirnames, filenames in self.walk():
            objects['__dirnames__'] = dirnames
            for filename in filenames:
                objects[str(root / filename)] = self.get_object(root / filename).key

        return make_hash(objects)

```

chrisjsewell:
closes #4918

Uses `hashlib.sha256` by default then, to be super safe, only uses the key for the object-store if `self.container.hash_type == 'sha256'`

before:

```
--------------------------------------------- benchmark 'node': 1 tests ---------------------------------------------
Name (time in ms)              Min       Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_store_with_object     10.4047  105.4098  14.0552  9.3613  13.0171  1.9786       1;4  71.1483     100           1
---------------------------------------------------------------------------------------------------------------------
```

after:

```
------------------------------------------- benchmark 'node': 1 tests -------------------------------------------
Name (time in ms)             Min     Max    Mean  StdDev  Median     IQR  Outliers       OPS  Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------
test_store_with_object     5.8858  9.1815  6.8171  0.7261  6.5448  0.9458      22;3  146.6907     100           1
-----------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/aiida/repository/backend/abstract.py
+++ b/aiida/repository/backend/abstract.py
@@ -7,6 +7,7 @@ This key should then be able to be used to retrieve the bytes of the correspondi
 """
 import abc
 import contextlib
+import hashlib
 import io
 import pathlib
 import typing
@@ -104,6 +105,21 @@ class AbstractRepositoryBackend(metaclass=abc.ABCMeta):
         """
         with self.open(key) as handle:
             return handle.read()
+
+    def get_object_hash(self, key: str) -> str:
+        """Return the hash of a object identified by key.
+
+        .. note::
+            This hash value is dependant on the repository implementation
+            and so may change for different repositories.
+            By default the algorithm is SHA-256
+
+        :param key: fully qualified identifier for the object within the repository.
+        :raise FileNotFoundError: if the file does not exist.
+        :raise OSError: if the file could not be opened.
+        """
+        with self.open(key) as handle:
+            return hashlib.sha256(handle.read()).hexdigest()
 
     def delete_object(self, key: str):
         """Delete the object from the repository.
--- a/aiida/repository/backend/abstract.py
+++ b/aiida/repository/backend/abstract.py
@@ -12,6 +12,8 @@ import io
 import pathlib
 import typing
 
+from aiida.common.hashing import chunked_file_hash
+
 __all__ = ('AbstractRepositoryBackend',)
 
 
@@ -107,19 +109,18 @@ class AbstractRepositoryBackend(metaclass=abc.ABCMeta):
             return handle.read()
 
     def get_object_hash(self, key: str) -> str:
-        """Return the hash of a object identified by key.
+        """Return the SHA-256 hash of an object stored under the given key.
 
-        .. note::
-            This hash value is dependant on the repository implementation
-            and so may change for different repositories.
-            By default the algorithm is SHA-256
+        .. important::
+            A SHA-256 hash should always be returned,
+            to ensure consistency across different repository implementations.
 
         :param key: fully qualified identifier for the object within the repository.
         :raise FileNotFoundError: if the file does not exist.
         :raise OSError: if the file could not be opened.
         """
         with self.open(key) as handle:
-            return hashlib.sha256(handle.read()).hexdigest()
+            return chunked_file_hash(handle, hashlib.sha256)
 
     def delete_object(self, key: str):
         """Delete the object from the repository.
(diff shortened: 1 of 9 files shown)


aiida/repository/backend/abstract.py
```diff
@@ -105,6 +106,21 @@ def get_object_content(self, key: str) -> bytes:
...
+            This hash value is dependant on the repository implementation
+            and so may change for different repositories.
+            By default the algorithm is SHA-256
+
+        :param key: fully qualified identifier for the object within the repository.
+        :raise FileNotFoundError: if the file does not exist.
+        :raise OSError: if the file could not be opened.
+        """
+        with self.open(key) as handle:
+            return hashlib.sha256(handle.read()).hexdigest()
```
giovannipizzi, line 122: This might exhaust RAM if the object is very big (possible).

You should compute the hash by reading the file in chunks.
You can copy-paste and adapt from here as an example:
https://github.com/aiidateam/disk-objectstore/blob/a2561a40cf2e9a4d58325387f913ccf08111f5fd/disk_objectstore/utils.py#L815-L839

(It's possible we've the same issue elsewhere in AiiDA, but I would start improving those one by one)

giovannipizzi (changes requested):
Seems a very good and safe approach.

Just a comment to avoid memory issues with big objects (feel free to dismiss my review once fixed, is @sphuber approves)

aiida/repository/backend/abstract.py
chrisjsewell: cheers will do

aiida/repository/backend/disk_object_store.py
```diff
@@ -99,3 +99,19 @@ def delete_object(self, key: str):
...
+            This hash value is dependant on the repository implementation
+            and so may change for different repositories.
+            By default the algorithm is SHA-256
+
+        :param key: fully qualified identifier for the object within the repository.
+        :raise FileNotFoundError: if the file does not exist.
+
+        """
+        if self.container.hash_type != 'sha256':
+            return super().get_object_hash(key)
```
sphuber, line 116: I was struggling for a bit to understand why this conditional was necessary because the container always uses the hash of the content as the key, so it shouldn't matter what hash type is used. However, then I realized that this would be a problem higher up that expects the same hash of a file regardless of what type of repository it is stored in. I think this is not directly obvious and think this should probably be written in the `Repository.get_hash` method docstring.

This may cause problems in the future when another repo is introduced that uses a different hashing type and all of a sudden caching and other things that rely on hashes may start failing, no?

giovannipizzi: Yes, indeed might not be obvious that this function *must* return a `sha256` hash (as this is what AiiDA expects, and should not change between different backends). Would be good to clarify.

aiida/repository/backend/abstract.py
chrisjsewell: Implemented `aiida.common.hashing.chunked_file_hash`

aiida/repository/backend/disk_object_store.py
chrisjsewell: Changed docstring to assert that SHA-256 hashing must be returned.

> This may cause problems in the future when another repo is introduced that uses a different hashing type

This would only now happen if the repo class specifically overrode `AbstractRepositoryBackend.get_object_hash` and did not "respect" the method's docstring, which would then be a bug in that classes implementation

giovannipizzi (approved):
Thanks!

sphuber (approved):
All good thanks @chrisjsewell

PR #4926, opened 2021-05-06
Migrations: fix the rename of `name` to `label` for `DbComputer`

### LINKED ISSUE #4925: Migration for `Computer` column `name` rename to `label` is broken for SqlAlchemy
it should use `alter_column` in order to prevent violations of the non-nullable new column. Rookie mistake and even worse that it was not caught because yours truly thought this was such a simple migration it didn't need one.

sphuber:
Fixes #4925 

The migration was broken for SqlAlchemy and wasn't detected because
there were no tests. The migration of new profiles still worked because
it only failed for databases that have at least one computer.

The new migration properly uses `alter_column` to change the column
name, instead of creating a new one and dropping the old one. Not only
does this incur the exception that the new column will have null values,
which is not allowed since it is not nullable, but it would also have
lost all the data.

Finally, the `label` column has a uniqueness constraint and the name
should technically also be changed, so the constraint is dropped and
added with the new name based on `label`.

### DIFF
--- a/aiida/backends/sqlalchemy/migrations/versions/535039300e4a_computer_name_to_label.py
+++ b/aiida/backends/sqlalchemy/migrations/versions/535039300e4a_computer_name_to_label.py
@@ -8,7 +8,6 @@ Create Date: 2021-04-28 14:11:40.728240
 
 """
 from alembic import op
-import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = '535039300e4a'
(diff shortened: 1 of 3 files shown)


aiida/backends/sqlalchemy/migrations/versions/535039300e4a_computer_name_to_label.py
```diff
@@ -19,15 +18,13 @@
 
 def upgrade():
     """Migrations for the upgrade."""
-    op.add_column('db_dbcomputer', sa.Column('label', sa.String(length=255), nullable=False))
-    op.drop_constraint('db_dbcomputer_name_key', 'db_dbcomputer', type_='unique')
+    op.alter_column('db_dbcomputer', 'name', new_column_name='label')  # pylint: disable=no-member
     op.create_unique_constraint('db_dbcomputer_label_key', 'db_dbcomputer', ['label'])
-    op.drop_column('db_dbcomputer', 'name')
+    op.drop_constraint('db_dbcomputer_name_key', 'db_dbcomputer')
```
giovannipizzi, line 23: What happens to this constraint when you alter the column? Maybe it's better to drop it before altering (also for performance, so the index does not exist and does not need to be updated while altering)? Same for downgrade (drop before altering)

sphuber: Not sure to be honest. Since this should be a small index (shouldn't be a huge number of computers) I think it won't have a huge impact, but happy to swap the order.

sphuber: swapped them around

giovannipizzi (approved):
Thanks! I guess that it would also just work to rename the constraint - I don't see any other thing the DB can do other than making `db_dbcomputer_label_key` actually be an index/constraing on the new column. But I think that in this way we're safe

PR #4929, opened 2021-05-07
♻️ REFACTOR: `Repository.get_hash_keys` -> `Repository.get_file_keys`

chrisjsewell:
I feel that this make it clearer what this method is returning.
When we refer to a hash, in the context of the repository, we are refering to the (hexidecimal) hash of a file's contents.
By contrast, the key to the file's contents does not have to be a hash of its contents, as is the case for `SandboxRepositoryBackend`, which uses a UUID.


sphuber (approved):
All good, thanks @chrisjsewell

PR #4930, opened 2021-05-07
👌 IMPROVE: Use `PurePosixPath` instead of `Path` in repository

chrisjsewell:
This makes it explicit that accessing paths in the repository is done via posix-like paths (i.e. rather than Windows ones), and that we are not dealing with the actual OS file-system.

Note that (on unix systems) `Path` is a subclass of `PurePosixPath`, so you can still pass those to the methods if you really wanted to.
For the class hierarchy see: https://docs.python.org/3/library/pathlib.html#module-pathlib

(thankfully we don't actually require Windows cross-support, because I have to deal with this all the time for jupyter-book)

### DIFF
--- a/aiida/repository/repository.py
+++ b/aiida/repository/repository.py
@@ -12,7 +12,7 @@ from .common import File, FileType
 
 __all__ = ('Repository',)
 
-FilePath = Union[str, pathlib.Path]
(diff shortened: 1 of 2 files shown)


sphuber (approved):
lgtm, thanks @chrisjsewell

PR #4934, opened 2021-05-07
`Config`: write config to disk at end of `delete_profile`

### LINKED ISSUE #4921: `Config.delete_profile` does not write changes to file
This means that the profile is not removed from the `config.json`. The method should call `self.store()` before returning and we should add a test. Tricky part is that in the test we cannot use the `Config` method to check if the profile still exists, because it won't in memory but we need to check that the file on disk is properly updated as well.

It should also check if the deleted profile was the default, and if so reset it.

sphuber:
Fixes #4921 

The `delete_profile` would successfully remove the profile from the
`Config` instance in memory, however, the changes would not be persisted
to disk. After reloading the config from disk, the deleted profile would
be back. A regression test is added for this behavior and the behavior
is fixed by calling `store` explictly at the end of the method, to
persist the new state to disk.

This PR also adds a separate commit that first refactors the tests
for the `Config` class to use `pytest` straight up. For clarity, these
commits should be merged separately.

### DIFF
--- a/aiida/manage/configuration/config.py
+++ b/aiida/manage/configuration/config.py
@@ -372,6 +372,7 @@ class Config:  # pylint: disable=too-many-public-methods
             postgres.drop_dbuser(profile.database_username)
 
         self.remove_profile(name)
+        self.store()
 
     def set_default_profile(self, name, overwrite=False):
         """Set the given profile as the new default.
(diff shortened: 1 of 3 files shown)


ltalirz (approved):
thanks @sphuber - this looks good to me!

>  After reloading the config from disk, the deleted profile would
be back. 

Do you happen to know when this regression was introduced?
I believe this worked correctly at some point.
Might want to mention this in the commit message.

sphuber:
@ltalirz fixed the commit message of the last commit that addresses the regression to include the commit that introduced it

PR #4935, opened 2021-05-07
`Node`: add the `walk` method to iterate over repository content

### LINKED ISSUE #4927: Expose `Repository.walk` on the `NodeRepositoryMixin` level
This will make the `walk` method available directly from the `Node` interface

sphuber:
Fixes #4927 

The `walk` method is added that simply forwards to the same method of
the underlying `Repository` instance that is assigned to the node. This
allows to recursively iterate over the directories and files that are
stored in the node's repository folder in the style of `os.walk`.

### DIFF
--- a/aiida/orm/nodes/repository.py
+++ b/aiida/orm/nodes/repository.py
@@ -2,14 +2,17 @@
 """Interface to the file repository of a node instance."""
 import contextlib
 import io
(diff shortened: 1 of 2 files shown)


ltalirz (approved):
thanks a lot @sphuber !

PR #4941, opened 2021-05-14
🔧 MAINTAIN: Remove psycopg2 Conda mapping

chrisjsewell:
See: https://github.com/conda-forge/aiida-core-feedstock/pull/54, critically it means that conda environments now pass `pip check` (they didn't before)


ltalirz (approved):
thanks @chrisjsewell , for the fix here and also on the conda package!

csadorf (approved):
Thx!

PR #4949, opened 2021-05-18
Add Schema validation for QueryBuilder

### LINKED ISSUE #4660: Define (JSON) schema for QueryBuilder queryhelp
so far, this actually only checks that the json is a dictionary, right?

I guess we currently don't have a JSON schema for the querybuilder dict yet, but perhaps it would make sense to have  some basic schema validation here, at least to the degree of the assumptions you make in the code below?

e.g. something like (untested)

```python
from voluptuous import Schema

QUERYHELP_SCHEMA = Schema({
  'path' : list,
  'project': dict,
  }
  required=True)

QUERYHELP_SCHEMA(self.trans._query_help)  # raises, if not valid
```

it may result in better error messages when posting invalid files, and it also makes the code easier to read by laying out the expectations about what is being posted.

_Originally posted by @ltalirz in https://github.com/aiidateam/aiida-core/pull/4337#discussion_r554552585_

This issue comes from PR #4337 where a validation of the POSTed queryhelp JSON would be useful for clients and users to understand how the POSTed data might be wrong.

Essentially a data class with validation is needed, which should be added to the `QueryBuilder` class.

NinadBhat:
Fixes #4660


chrisjsewell:
Err, I'm not sure about this voluptuous approach of specifying the schema as a python object, as opposed to having a standalone JSON file containing the schema.

The reason why a JSON file would be nice, is that then you can take it and use it on the client side, to construct a query. The client may not be written in Python.

This is my initial reaction though; I've never heard of voluptuous before so maybe I need to understand it better first (happy to discuss)

For an existing use of JSON schema in aiida, see the config validation that I added

NinadBhat:
@chrisjsewell I used voluptuous because it was mentioned in #4660. I do not mind using something else. Let's discuss what we finally want to move forward with.

PR #4950, opened 2021-05-18
suggest running reentry scan for unstorable Node

ltalirz:
When encountering a Node that isn't storable because of an entry point
that isn't registered, suggest to run `reentry scan`.
We got a user question regarding this, apparently this was not obvious.

### DIFF
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -197,11 +197,11 @@ class Node(Entity, NodeRepositoryMixin, EntityAttributesMixin, EntityExtrasMixin
             raise exceptions.StoringNotAllowed(self._unstorable_message)
 
         if not is_registered_entry_point(self.__module__, self.__class__.__name__, groups=('aiida.node', 'aiida.data')):
-            msg = f'class `{self.__module__}:{self.__class__.__name__}` does not have a registered entry point.' \
-             + ' Consider running `reentry scan`.' \
-             + ' If the issue persists, check that the corresponding plugin is installed ' \
-             + 'and that the entry point shows up in `verdi plugin list`.'
-            raise exceptions.StoringNotAllowed(msg)
+            raise exceptions.StoringNotAllowed(
+                f'class `{self.__module__}:{self.__class__.__name__}` does not have a registered entry point. '
+                'Consider running `reentry scan`. If the issue persists, check that the corresponding plugin is '
+                'installed and that the entry point shows up in `verdi plugin list`.'
+            )
 
     @classproperty
     def class_node_type(cls) -> str:
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -197,7 +197,8 @@ class Node(Entity, NodeRepositoryMixin, EntityAttributesMixin, EntityExtrasMixin
             raise exceptions.StoringNotAllowed(self._unstorable_message)
 
         if not is_registered_entry_point(self.__module__, self.__class__.__name__, groups=('aiida.node', 'aiida.data')):
-            msg = f'class `{self.__module__}:{self.__class__.__name__}` does not have registered entry point'
+            msg = f'class `{self.__module__}:{self.__class__.__name__}` does not have a registered entry point.' \
+             + ' Consider running `reentry scan`.'
             raise exceptions.StoringNotAllowed(msg)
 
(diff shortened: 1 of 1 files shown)


ramirezfranciscof (approved):
Not sure why you wanted @sphuber in particular to see this, seems pretty straightforward and for me its ready to go (I would maybe say "Consider trying again after running `reentry scan`", but it shouldn't make much of a difference).

I'll leave it to you @ltalirz to merge it or wait for the other review as you prefer.

sphuber:
I didn't add this sentence originally, because it is not guaranteed to be the problem. If the plugin is not even installed, running `reentry scan` will have zero effect. For example, if someone clones the repo of a plugin and then imports one of its data plugins directly through the path and tries to store it, they will be faced with this error and `reentry scan` won't help. So maybe adding 

> Please install the corresponding plugin first and run `reentry scan`

would be more accurate?

ramirezfranciscof:
Mmm a bit verbose but all encompassing option?:

> Please check the installed plugins using `verdi plugin list` and `verdi plugin list <PLUGIN_TYPE>`. If it is not listed there, you may need to install the corresponding plugin and/or run `reentry scan` to update the registry.

aiida/orm/nodes/node.py
```diff
@@ -197,7 +197,8 @@ def validate_storability(self) -> None:
             raise exceptions.StoringNotAllowed(self._unstorable_message)
 
         if not is_registered_entry_point(self.__module__, self.__class__.__name__, groups=('aiida.node', 'aiida.data')):
-            msg = f'class `{self.__module__}:{self.__class__.__name__}` does not have registered entry point'
+            msg = f'class `{self.__module__}:{self.__class__.__name__}` does not have a registered entry point.' \
+             + ' Consider running `reentry scan`.'
```
ltalirz, line 201:
```suggestion
             + ' Consider running `reentry scan`.' \
             + ' If the issue persists, check that the corresponding plugin is installed and that the entry point shows up in `verdi plugin list`'
```

ltalirz:
Thanks to both for your suggestions.

I think in most cases the problem is likely to be the reentry cache (as in the user question I got), so running `reentry scan`  should be the first suggestion.

I've incorporated your suggestions in a follow-up sentence.

ramirezfranciscof:
It works for me, but I was also ok with the previous one so this time I'll let @sphuber have the approving review if he agrees.

P.S.: I think the CI apocalypse is due to an extra space after the line continuation `\` on the third line added (Python is moody, hehe). Is this something we should add to our pre-commit hooks to catch, or do they already have this check?

ltalirz:
> P.S.: I think the CI apocalypse is due to an extra space after the line continuation `\` on the third line added (Python is moody, hehe). Is this something we should add to our pre-commit hooks to catch, or do they already have this check?

Sorry, I thought I might be able to do it from the github interface but it's always easy to make mistakes...

aiida/orm/nodes/node.py
```diff
@@ -197,7 +197,10 @@ def validate_storability(self) -> None:
             raise exceptions.StoringNotAllowed(self._unstorable_message)
 
         if not is_registered_entry_point(self.__module__, self.__class__.__name__, groups=('aiida.node', 'aiida.data')):
-            msg = f'class `{self.__module__}:{self.__class__.__name__}` does not have registered entry point'
+            msg = f'class `{self.__module__}:{self.__class__.__name__}` does not have a registered entry point.' \
+             + ' Consider running `reentry scan`.' \
+             + ' If the issue persists, check that the corresponding plugin is installed ' \
+             + 'and that the entry point shows up in `verdi plugin list`.'
             raise exceptions.StoringNotAllowed(msg)
```
sphuber, line 204:
```suggestion
            raise exceptions.StoringNotAllowed(
                f'class `{self.__module__}:{self.__class__.__name__}` does not have a registered entry point. '
                'Consider running `reentry scan`. If the issue persists, check that the corresponding plugin is '
                'installed and that the entry point shows up in `verdi plugin list`.'
            )
```

sphuber (commented):
Fine with the new message. Would just change the code to put the message inside the constructor of the exception. That makes all the `\` and `+` characters unnecessary and so there is less clutter.

sphuber (approved):
Thanks @ltalirz

Issue #4944, opened 2021-05-14
Spurious `paramiko.ssh_exception.SSHException: Error reading SSH protocol banner`


dev-zero:
### Describe the bug

I occasionally get the following error when running calculations on a fresh installations of AiiDA despite having run `verdi computer test` and using a password-less key file:

```
05/14/2021 09:46:11 AM <24898> aiida.transport.SshTransport: [ERROR] Error connecting to 'eiger.cscs.ch' through SSH: [SshTransport] Error reading SSH protocol banner, connect_args were: {'username': 'timuel', 'port': 22, 'look_for_keys': True, 'key_filename': '/users/tiziano/.ssh/id_ed25519.cscs', 'timeout': 60, 'allow_agent': True, 'proxy_command': 'ssh -W eiger.cscs.ch:22 timuel@ela.cscs.ch', 'compress': True, 'gss_auth': False, 'gss_kex': False, 'gss_deleg_creds': False, 'gss_host': 'eiger.cscs.ch'}
05/14/2021 09:46:11 AM <24898> aiida.engine.transports: [ERROR] exception occurred while trying to open transport:
 Error reading SSH protocol banner
05/14/2021 09:46:11 AM <24898> aiida.engine.transports: [ERROR] Exception whilst using transport:
Traceback (most recent call last):
  File "/scratch/tiziano/virtualenvs/aiida/lib/python3.9/site-packages/paramiko/transport.py", line 2211, in _check_banner
    buf = self.packetizer.readline(timeout)
  File "/scratch/tiziano/virtualenvs/aiida/lib/python3.9/site-packages/paramiko/packet.py", line 380, in readline
    buf += self._read_timeout(timeout)
  File "/scratch/tiziano/virtualenvs/aiida/lib/python3.9/site-packages/paramiko/packet.py", line 622, in _read_timeout
    raise socket.timeout()
socket.timeout
[...]
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/scratch/tiziano/work/aiida/aiida-core/aiida/engine/utils.py", line 188, in exponential_backoff_retry
    result = await coro()
  File "/scratch/tiziano/work/aiida/aiida-core/aiida/engine/processes/calcjobs/tasks.py", line 190, in do_update
    job_info = await cancellable.with_interrupt(update_request)
  File "/scratch/tiziano/work/aiida/aiida-core/aiida/engine/utils.py", line 95, in with_interrupt
    result = await next(wait_iter)
[...]

Issue #3929, opened 2020-04-14
More options for using AiiDA with locked-down supercomputers


ltalirz:
In cases where HPC centers don't offer access via SSH keys (e.g. requiring 2-factor authentication), the only way to use AiiDA currently is to install it on the cluster (which is possible and has become a lot easier since the introduction of the `aiida-core` and `aiida-core.services` conda packages).

However, there are also alternative routes we could explore, which I list below such that they don't get lost:

  1. opening an SSH connection once and keep reusing it (e.g. using `ControlPersists yes`,  and `ControlPath ~/.ssh/cm_socket/%r@%h:%p` in the `~/.ssh/config` file). This is currently [not supported by paramiko](https://github.com/paramiko/paramiko/issues/852) but there are alternative python bindings like [ssh2-python](https://github.com/ParallelSSH/ssh2-python) and [parallel-ssh](https://github.com/ParallelSSH/parallel-ssh) we could look into.

  2. the guys at nersc have developed a small set of scripts called [sshproxy](https://www.nersc.gov/assets/Uploads/Using-sshproxy.pdf) that grants temporary access to SSH keys on the cluster side. I.e. you authenticate once, then something on the cluster "enables" your key for a period of time (say 24h), and after that time the key is disabled again. It seems they haven't put it on their [github](https://github.com/NERSC) yet, but if they were asked perhaps they would be fine with open sourcing it. Of course, this route would always require action from the cluster administrator.

P.S. We might anyhow want to look into ssh2-python for [performance reasons](https://parallel-ssh.org/post/ssh2-python/). See also [here](https://github.com/ParallelSSH/parallel-ssh#comparison-with-alternatives) and [here](https://parallel-ssh.readthedocs.io/en/latest/ssh2.html) for a comparison with paramiko.

Mentioning @sphuber for info

ConradJohnston:
Just to comment - I think this could become an increasingly important issue, if current trends continue. There was a security incident affecting Tier 1 and 2 HPC machines in the UK this week and it looks possible that SSH-alone access may no longer be possible. If these sorts of highly-targeted (and possibly state-sponsored?) attacks continue, it seems highly likely that many more HPC machines will become locked down with 2FA and similar measures. This is quite a significant threat to usefulness and impact of AiiDA, if not addressed.

sphuber:
[...]

PR #4951, opened 2021-05-18
transports/ssh: support proxy_jump

dev-zero:


ltalirz:
thanks for the addition @dev-zero !

from my side this is very welcome, we should just think about how to best avoid confusing new users with two ways of configuration and guide them to the new proxyjump command, both in the docs and through the cli.

I could even imagine deprecating (and eventually removing) the proxycommand option from the cli, i.e. just keeping the support in the python API for existing profiles.

dev-zero:
The  proxycommand may be needed for some more exotic cases where people are tunneling through `nc`.
Wrt docs and guiding the user: this is why I kept it as an `RFC` for now, I just need a band-aid for now and the `proxycommand` seems atm to be one of the culprits of hanging workers.

dev-zero:
Oh, and how beautifully it works :-) just submitted 100+ calculations, not a single wait, no workers hitting 100% CPU

dev-zero:
@ltalirz ok, updated the docs and help strings

ltalirz:
> @ltalirz ok, updated the docs and help strings

Thanks ! With "docs" are you referring to the aiida-core documentation?
I guess the place I had in mind was [this](https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/ssh.html?highlight=ssh#ssh-configuration)

By the way, do I understand correctly that for you this fix got rid of the symptom https://github.com/aiidateam/aiida-core/issues/4876#issuecomment-843329087 ?

Do you know whether it is due to the use of the `ProxyJump` or whether it is due to not using a separate process group?

It seems to me that the separate process group was introduced by @greschd in 00ed7f66511678d8ef153e423f94c5bada7b3bc3 , perhaps he can comment on the reasoning behind it.

dev-zero:
> > @ltalirz ok, updated the docs and help strings
> 
> Thanks ! With "docs" are you referring to the aiida-core documentation?
> I guess the place I had in mind was [this](https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/ssh.html?highlight=ssh#ssh-configuration)

Ah, right, I thought more about the documentation strings for the CLI.


> By the way, do I understand correctly that for you this fix got rid of the symptom [#4876 (comment)](https://github.com/aiidateam/aiida-core/issues/4876#issuecomment-843329087) 

That, and all other SSH-related issues I've been seeing.
 
> Do you know whether it is due to the use of the `ProxyJump` or whether it is due to not using a separate process group?

Guessing from https://github.com/paramiko/paramiko/pull/1180 I'd say it is the implementation of `ProxyCommand` in general (and not necessarily ours) and the switch to `ProxyJump` avoids it.

ltalirz:
Wow, so using `ProxyCommand` with `paramiko` can let you end up in an infinite loop, and this was known since 2017 and never fixed https://github.com/paramiko/paramiko/pull/963 . That's a little worrying and perhaps suggests to look at alternatives at some point as suggested [here](https://github.com/aiidateam/aiida-core/issues/3929#issue-599405342)...

In any case, in the short term it makes the proxyjump support even more pressing (if this does indeed get rid of the problem).

@sphuber @giovannipizzi Do you have any comments on the implementation basics here?
@greschd Should one use the separate process group?

sphuber:
Interesting, glad you found the source of your problems. I have to say, I wouldn't have thought to look at the SSH connections in that infinite recursion error. The weird thing is, I have been using the current setup to submit calculations over SSH with proxy _very_ heavily, as in tens of thousands of jobs, without problem. This was for the high-throughput project on Daint which we connect to through Ela. The configuration of the computer authentication in AiiDA that we have been using looks like

    ssh -i ~/.ssh/id_rsa username@ela.cscs.ch /usr/bin/netcat daint.cscs.ch 22`.

Would you have expected this to also run into the issue you have been observing. Or is that the kind of setup that you referred to as
> The proxycommand may be needed for some more exotic cases where people are tunneling through nc

and so here `nc` is the `netcat` that we are using?

dev-zero:
Yes, that's what I was referring to. In principal you wouldn't need the the `netcat` in there, but use the `-W` argument for `ssh` instead (which was my setup and I believe also the one of some others). Maybe the behaviour changes depending on that specific configuration, for example the reaction on reception of a `SIGTERM` or the buffering (in your case ssh forwards the `SIGTERM` to netcat which probably terminates immediately while SSH by itself alone when using `-W` tries to play it a bit safer). The `.~/ssh/config` may also play a role: for example if you have a `MasterSocket` then you may  get a corresponding log message from SSH on stderr.

What puzzled me was the fact that my workers were stalling, meaning 100% CPU but not processing any jobs. Stopping the demon generated a timeout and when I was checking the processes the workers were still running with one SSH zombie process each. Zombie processes usually mean that the caller didn't follow "good protocol" (like not doing a `.communicate()`, but only a `.wait()` which is bad if the kernel still waits for the process to drain a buffer). Following how Paramiko does communication a bit further (namely using only `.recv()`) led me to believe that they are driving the `Popen` (in disguise as a socket) in exactly the way the Python docs warn about that can lead to deadlocks, even though they seemed to have been careful.
Doing a `setsid()` on the SSH `Popen` might also not be the best idea (essentially demonizing the process), but if I remember correctly that was needed to solve some other problem.

I don't see how this patch solves the recursion problem per se, but think that the "setup" of deadlocked workers generated by this problem triggers other issues. While the improber termination of the workers is then probably responsible for the `Process unreachable` problems I've been seeing.

Stopping the demon, reconfiguring the host to use `proxy_jump` and then restarting the demon solved all the issues I have reported or seen reported wrt to SSH, CPU usage, Recursion Errors **so far**. I can say more once CSCS is up again, but yesterday the remaining ~200 jobs which hanged in `Created` or were awaiting upload and ~100 more steps were handled within <1min while it took >5min to submit one single job before.

Sorry that there's still a lot of speculation in some parts, but maybe we can fill those gaps together.

giovannipizzi (commented):
Thanks @dev-zero !

Since this adds support for a new feature, and in addition solves issues, I'm in favour of this!
I didn't check the implementation in detail but the idea seems good.

If we believe that the proxy jump is the 'better' way of doing things, I agree on documenting it as the "default suggestion" in our docs (and just mention proxy_command for advanced users). I wouldn't deprecate the other one for now as there might be use cases for it (already the fact that proxy_command comes later, and the comment on proxy_command to use the proxy jump instead, are enough).

As a comment, I've also been using a lot the proxy command with daint also recently. However, in the past (I think AiiDA 0.x) I remember having (different? or maybe similar) issues with it, where (when using a different nc executable on daint, *and* probably a different proxy_command string) the processes were remaining as zombie (but this was happening on daint - maybe in the end they were happening also on my computer, but I noticed because I was contacted by the admins that saw hundreds of processes on the jump host ela :-D

Final comment - good to check the various YML we have lying around, to make sure that those 1. continue working or 2. are updated to the new proxy jump. In particular for AiiDAlab (pinging @yakutovicha and @csadorf).

On this PR: I would add changes to the docs as discussed, as well as tests on the CLI to test the new CLI option.

dev-zero:
for completeness the `proxy_command` I've been using (taken from a `yml` for Daint):

```
proxy_command          ssh -W eiger.cscs.ch:22 timuel@ela.cscs.ch
```

dev-zero:
There's one caveat to the solution: since this relies on more Paramiko ssh connections (one per proxy and one for the machine) I seem to spawn too many threads. From `top` (with the `nTH` (number of threads) column enable):

```
PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM nTH     TIME+ COMMAND                                                                                                                                                                                    
18450 tiziano   20   0 7858248 237060  37784 S 14.62 1.450 321  76:55.02 verdi                                                                                                                                                                                      
18449 tiziano   20   0 7805784 229932  37328 S 6.312 1.407 319  73:36.11 verdi                                                                                                                                                                                      
18447 tiziano   20   0 7859568 221544  37356 S 5.980 1.356 325  74:18.38 verdi                                                                                                                                                                                      
18448 tiziano   20   0 7851892 239564  38556 S 5.980 1.466 320  75:17.40 verdi                                                                                                                                                                                      
18451 tiziano   20   0 7832944 231792  37440 S 5.980 1.418 322  69:26.28 verdi                                                                                                                                                                                      
18452 tiziano   20   0 7795188 240476  37084 S 5.980 1.471 318  73:59.64 verdi      
```

which causes:

```
-bash: fork: retry: Resource temporarily unavailable
```

My guess is a combination of:

* https://github.com/paramiko/paramiko/issues/949 ... but searching for `thread` on the Paramiko issue tracker gives also other unresolved candidates
* https://bugs.python.org/issue37788

dev-zero:
Ok, all done. The following warnings are as far as I can tell reported from bogus locations since the CI run from the commit before showed the same messages, but emitted simply from a different location. And ran the reported testcase via a script which monkey-patched file and socket open/close and printed tracebacks based on `tracemalloc` and couldn't reproduce it.

NEW:
```
2021-05-19T12:43:04.5716799Z tests/transports/test_ssh.py::test_gotocomputer_proxyjump
2021-05-19T12:43:04.5718005Z   /opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/abc.py:86: ResourceWarning: unclosed file <_io.FileIO name=74 mode='wb' closefd=True>
2021-05-19T12:43:04.5718810Z     _abc_init(cls)
2021-05-19T12:43:04.5719033Z 
2021-05-19T12:43:04.5719535Z tests/transports/test_ssh.py::test_gotocomputer_proxyjump
2021-05-19T12:43:04.5720710Z   /opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/abc.py:86: ResourceWarning: unclosed file <_io.FileIO name=75 mode='rb' closefd=True>
2021-05-19T12:43:04.5721480Z     _abc_init(cls)
2021-05-19T12:43:04.5721714Z 
2021-05-19T12:43:04.5722203Z tests/transports/test_ssh.py::test_gotocomputer_proxyjump
2021-05-19T12:43:04.5723389Z   /opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/abc.py:86: ResourceWarning: unclosed file <_io.FileIO name=77 mode='rb' closefd=True>
2021-05-19T12:43:04.5724162Z     _abc_init(cls)
2021-05-19T12:43:04.5724383Z 
```
https://pipelines.actions.githubusercontent.com/lx6f5qJRWQcbiE7vU42yTxrR7aWtThmdNBqldi7jyXdZiiVFpF/_apis/pipelines/1/runs/11842/signedlogcontent/24?urlExpires=2021-05-19T12%3A46%3A18.9245123Z&urlSigningMethod=HMACV1&urlSignature=tv%2BRl9us0wUkgaVtg%2BtoZNvWoqOs2UxxNoyZcT4ulwg%3D

PREVIOUS:
```
2021-05-16T16:22:23.8667487Z tests/workflows/arithmetic/test_add_multiply.py::test_run
2021-05-16T16:22:23.8668712Z   /opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/traceback.py:359: ResourceWarning: unclosed file <_io.FileIO name=59 mode='wb' closefd=True>
2021-05-16T16:22:23.8669617Z     result.append(FrameSummary(
2021-05-16T16:22:23.8669953Z 
2021-05-16T16:22:23.8670445Z tests/workflows/arithmetic/test_add_multiply.py::test_run
2021-05-16T16:22:23.8671647Z   /opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/traceback.py:359: ResourceWarning: unclosed file <_io.FileIO name=60 mode='rb' closefd=True>
2021-05-16T16:22:23.8672559Z     result.append(FrameSummary(
2021-05-16T16:22:23.8672976Z 
2021-05-16T16:22:23.8673450Z tests/workflows/arithmetic/test_add_multiply.py::test_run
2021-05-16T16:22:23.8674671Z   /opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/traceback.py:359: ResourceWarning: unclosed file <_io.FileIO name=63 mode='rb' closefd=True>
2021-05-16T16:22:23.8675574Z     result.append(FrameSummary(
```
https://pipelines.actions.githubusercontent.com/lx6f5qJRWQcbiE7vU42yTxrR7aWtThmdNBqldi7jyXdZiiVFpF/_apis/pipelines/1/runs/11774/signedlogcontent/24?urlExpires=2021-05-19T13%3A25%3A52.6201442Z&urlSigningMethod=HMACV1&urlSignature=eivYPNcjaiguS7TPL%2FQl29J2j%2Bp5CiGxEwGJtxeMU3k%3D

Also, those warnings from the previous commit don't show up in this PRs CI run and there was nothing in between which could have fixed them.

dev-zero:
Since I was leaking threads/filedescriptors/whatnot in an earlier version of this patch I stumbled over this blog article via HN which is about file descriptors (and how a lot of other things like sockets, timers are filedescriptors nowadays): http://0pointer.net/blog/file-descriptor-limits.html
For AiiDA the limit of open files per process is therefore 1024 since Paramiko (and maybe others) uses `select` and should you ever have the hunch that you might be hitting that limit without the code properly handling the errors from it:

```console
$ for pid in $(pgrep verdi) ; do echo "PID = $pid with $(ls /proc/$pid/fd/ | wc -l) file descriptors" ; done
```

Might be something to extend `verdi daemon status` with it? (together with "number of threads", virt/res memory) And maybe also add a check that the user didn't set the soft limit for number of open files beyond 1024.

giovannipizzi:
Interesting... I read briefly the link. So we should make sure we don't use a limit over 1024, at least as long as we use paramiko? Ok for me to add to verdi test if it's important to check - @sphuber @ramirezfranciscof @chrisjsewell what do you think?

dev-zero:
> Interesting... I read briefly the link. So we should make sure we don't use a limit over 1024, at least as long as we use paramiko?

Correct. There might be other plugins which still use `select` (psycopg mabe?), but paramiko is definitely using it. Normally I wouldn't care about it too much because the softlimit is `1024`, but I see something like `ulimit -a unlimited` when people are doing calculations and there may be a good chance that `verdi` is run with that setting too.

docs/source/howto/ssh.rst
```diff
@@ -196,7 +196,25 @@ This section explains how to use the ``proxy_command`` feature of ``ssh`` in ord
...
+      User USER_TARGET
+      IdentityFile ~/.ssh/aiida
+      ProxyJump USER_PROXY@FULLHOSTNAME_PROXY
+
+  Host FULLHOSTNAME_PROXY
+      IdentityFile ~/.ssh/aiida
+
+replacing the ``..._TARGET`` and ``..._PROXY`` variables with the host/user names of the respective servers.
+
+To setup the *PROXY* configuration with ``ProxyCommand` **instead**, edit the ``~/.ssh/config`` file on the computer on which you installed AiiDA (or create it if missing)
```
ltalirz, line 216:
```suggestion
To setup the proxy configuration with ``ProxyCommand` **instead**, edit the ``~/.ssh/config`` file on the computer on which you installed AiiDA (or create it if missing)
```

docs/source/howto/ssh.rst
```diff
@@ -206,24 +224,30 @@ Edit the ``~/.ssh/config`` file on the computer on which you installed AiiDA (or
 
 replacing the ``..._TARGET`` and ``..._PROXY`` variables with the host/user names of the respective servers.
 
-This should allow you to directly connect to the *TARGET* server using
+In both cases, shis should allow you to directly connect to the *TARGET* server using
```
ltalirz, line 227:
```suggestion
In both cases, this should allow you to directly connect to the *TARGET* server using
```

docs/source/howto/ssh.rst
```diff
@@ -206,24 +224,30 @@ Edit the ``~/.ssh/config`` file on the computer on which you installed AiiDA (or
...
+
 AiiDA configuration
 ^^^^^^^^^^^^^^^^^^^
 
-When :ref:`configuring the computer in AiiDA <how-to:run-codes:computer:configuration>`, AiiDA will automatically parse the required information from your ``~/.ssh/config`` file.
+When :ref:`configuring the computer in AiiDA <how-to:run-codes:computer:configuration>`, AiiDA will automatically parse most of required information from your ``~/.ssh/config`` file.
+
+.. dropdown:: Specifying the proxy_jump
+
+    The ``proxy_jump`` *must** be specified manually when configuring the SSH connection and is not automatically taken from your ``~/.ssh/config``.
```
ltalirz, line 249: will it figure out to use the correct key for the proxy?

docs/source/howto/ssh.rst
```diff
@@ -196,7 +196,25 @@ This section explains how to use the ``proxy_command`` feature of ``ssh`` in ord
...
+To setup the proxy configuration with ``ProxyJump`, edit the ``~/.ssh/config`` file on the computer on which you installed AiiDA (or create it if missing)
+and add the following lines::
+
+  Host SHORTNAME_TARGET
+      Hostname FULLHOSTNAME_TARGET
+      User USER_TARGET
+      IdentityFile ~/.ssh/aiida
+      ProxyJump USER_PROXY@FULLHOSTNAME_PROXY
+
+  Host FULLHOSTNAME_PROXY
```
ltalirz, line 211: does one need that also for the proxycommand version below?

ltalirz (commented):
thanks a lot @dev-zero 

I was hoping @greschd could perhaps give a quick comment on whether this approach without separate process groups seems fine.

in the meanwhile I already gave a quick look

aiida/transports/plugins/ssh.py
```diff
@@ -420,6 +445,8 @@ def open(self):
 
         :raise aiida.common.InvalidOperation: if the channel is already open
         """
+        import paramiko
```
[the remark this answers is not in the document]
dev-zero: done

aiida/transports/plugins/ssh.py
```diff
@@ -118,12 +119,28 @@ class SshTransport(Transport):  # pylint: disable=too-many-public-methods
...
             }
         ),
+        (
+            'proxy_jump', {
+                'prompt':
+                'SSH proxy jump',
+                'help':
+                'SSH proxy jump for tunneling through other SSH hosts.'
+                ' Use a comma-separated list of hosts of the form [user@]host[:port].'
+                ' If user or port are not specified for a host, the user & port values from the target host are used.',
```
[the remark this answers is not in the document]
dev-zero: Ok, done.

aiida/transports/plugins/ssh.py
```diff
@@ -429,9 +456,65 @@ def open(self):
...
-            self._proxy = _DetachedProxyCommand(proxystring)
+
+        proxyjumpstring = connection_arguments.pop('proxy_jump', None)
+        proxycmdstring = connection_arguments.pop('proxy_command', None)
+
+        if proxyjumpstring and proxycmdstring:
+            raise ValueError('The SSH proxy jump and SSH proxy command options can not be used together')
+
+        if proxyjumpstring:
+            matcher = re.compile(r'^(?:(?P<username>[^@]+)@)?(?P<host>[^@:]+)(?::(?P<port>\d+))?\s*$')
```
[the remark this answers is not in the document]
dev-zero: I take that as a compliment. It is based on the spec of `proxy_jump` for `ssh_config` and written by me.

docs/source/howto/ssh.rst
dev-zero: done

docs/source/howto/ssh.rst
dev-zero: 🤣 thanks for the catch

docs/source/howto/ssh.rst
dev-zero: for the proxycommand it would even be crucial since that configuration is not controlled by paramiko but only ssh (since the `ssh` process is spawned). Added.

docs/source/howto/ssh.rst
dev-zero: I added a comment: with the current design the same key is used (if any) as for the main/target host.
Theoretically you could specify a complete set of configurations for each proxy jump (`ssh_config` allows that).
But that would make the interface rather complicated, so I went with 80/20 rule and simply copy the complete config from the main host.

dev-zero:
> thanks a lot @dev-zero
> 
> I was hoping @greschd could perhaps give a quick comment on whether this approach without separate process groups seems fine.

Ok, now I understood what you meant by the previous comment and I can answer: using a separate process group for `proxy_jump` does not make sense. For `proxy_command` you are actually spawning an `ssh` process (you see that in `ps`), for `proxy_jump` you do not. Instead we open a Paramiko connection for each proxy host, create a communication channel on that connection which the remote ssh process directs to the next hop and is locally just another socket which then gets used by the next Paramiko connection. There are no additional processes involved, only communication threads which Paramiko handles internally.

sphuber:
By the way @dev-zero , does this feature address issue #4944 ?

dev-zero:
@sphuber indirectly, yes. Since it doesn't rely on `ssh` anymore which could block when waiting for an agent response I think it is more manageable. But one would have to test with a key coming from an agent again, but even that should then produce a more meaningful error message since a failure in obtaining a key will then directly handled by Paramiko rather than silently swallowed in the spawned subprocess.

docs/source/howto/ssh.rst
```diff
@@ -196,34 +196,67 @@ This section explains how to use the ``proxy_command`` feature of ``ssh`` in ord
...
+      IdentityFile ~/.ssh/aiida
+
+replacing the ``..._TARGET`` and ``..._PROXY`` variables with the host/user names of the respective servers.
```
ltalirz, line 214:
```suggestion
replacing the ``..._TARGET`` and ``..._PROXY`` variables with the host/user names of the respective servers.

.. note :: 

    Please note that when using `ProxyJump` with AiiDA, the proxy and the target host must use the *same* `IdentityFile`.
    If the user directory of the proxy is shared with the target host (often the case in supercomputing facilities), this is ensured automatically.
```

docs/source/howto/ssh.rst
```diff
@@ -196,34 +196,67 @@ This section explains how to use the ``proxy_command`` feature of ``ssh`` in ord
...
+
+To setup the proxy configuration with ``ProxyCommand` **instead**, edit the ``~/.ssh/config`` file on the computer on which you installed AiiDA (or create it if missing)
```
ltalirz, line 216: In order not to overload the user with options, I would suggest to hide this inside a

```
.. dropdown:: :fa:`plus-circle` Alternative setup with ``ProxyCommand``
```

(not mandatory)

docs/source/howto/ssh.rst
```diff
@@ -196,34 +196,67 @@ This section explains how to use the ``proxy_command`` feature of ``ssh`` in ord
...
 AiiDA configuration
 ^^^^^^^^^^^^^^^^^^^
 
-When :ref:`configuring the computer in AiiDA <how-to:run-codes:computer:configuration>`, AiiDA will automatically parse the required information from your ``~/.ssh/config`` file.
+When :ref:`configuring the computer in AiiDA <how-to:run-codes:computer:configuration>`, AiiDA will automatically parse most of required information from your ``~/.ssh/config`` file.
+
+.. dropdown:: Specifying the proxy_jump
+
+    The ``proxy_jump`` *must** be specified manually when configuring the SSH connection and is not automatically taken from your ``~/.ssh/config``.
+    Simply copy & paste the same instructions as you have used for ``ProxyJump`` in your ``~/.ssh/config`` to the input for ``proxy_jump``.
```
ltalirz, line 253:
```suggestion
    Simply copy & paste the instructions you passed to the ``ProxyJump`` directive in your ``~/.ssh/config`` to the input of ``proxy_jump``.
```

docs/source/howto/ssh.rst
```diff
@@ -196,34 +196,67 @@ This section explains how to use the ``proxy_command`` feature of ``ssh`` in ord
...
+
+    The same key (if any) as specified for the *TARGET* host will be used to login to the proxy, hence you have to make sure that the public part
+    of your local key is registered in the ``.ssh/authorized_keys`` file for both the *TARGET* host and all of the proxies/jump hosts.
+    The user directory is usually shared between all nodes in a supercomputing facility, hence this is usually ensured automatically.
+
```
ltalirz, line 258:
```suggestion

```

ltalirz (commented):
thanks @dev-zero - just some final suggestions on the docs; after this I will approve and merge.

anyone wanting to give this another look, please do so now.

P.S. I haven't tested the RST directives in the suggestions - minor fixes may be needed

docs/source/howto/ssh.rst
dev-zero: To be honest I think the formulation of proxy/target "using an `IdentityFile`" is a bit dangerous, people might start copying their private keys. Maybe something like: "the proxy and target must accept the same public key (matching the `IdentityFile`)?

ltalirz: I see. Updated suggestion!

dev-zero: the `IdentityFile` IS the private key

dev-zero:
@ltalirz I wasn't happy with the documentation and redid it. I hope it is clearer now.

dev-zero:
please squash commits on merge

mbercx:
Sorry for not having looked that this earlier, but I'm still having a quick look now, so don't merge yet! 😅

docs/source/howto/ssh.rst
```diff
@@ -185,47 +185,79 @@ Connecting to a remote computer *via* a proxy server
...
+   $ verdi computer configure ssh SHORTNAME_TARGET
+   ...
+   Allow ssh agent [True]:
+   SSH proxy jump []: USER_PROXY@FULLHOSTNAME_PROXY
+
+.. note:: A chain of proxies can be specified as a comma-separated list. If you need to specify a different username, you can so with ``USER_PROXY@...``. If no username is specified for the proxy the same username as for the *TARGET* is used.
 
-.. dropdown:: Specifying the proxy_command manually
+.. dropdown:: Specifying the ``proxy_command`` manually
```
mbercx, line 260: Hmm, now that I see this, I'm not sure if I would hide it in a dropdown. Although using `ProxyJump` is obviously preferable, users that _do_ use `ProxyCommand` should not miss this message. Maybe use `important` instead?

ltalirz (dismissed):
> @ltalirz I wasn't happy with the documentation and redid it. I hope it is clearer now.

All good, an of course your point about the `IdentityFile` was well taken.

@mbercx I'm unsubscribing from this thread. Please take care of merging this after your review

mbercx (changes requested):
Thanks a lot for this, @dev-zero! We've been having issues with using `ProxyCommand` when connecting to CSCS clusters, so having support for `ProxyJump` will be a big plus.

I haven't reviewed the code in much detail, I trust @ltalirz has that covered. I just wanted to try this out by following the documentation to see if everything is clear. However, I haven't been able to make this work for my connection to daint. Maybe I messed up somewhere (it's a Sunday after all)? Here's the relevant piece in the `~/.ssh/config`:

```console
$ grep -A 8 daint-test ~/.ssh/config 
Host daint-test
  HostName daint.cscs.ch
  User mbercx
  IdentityFile ~/.ssh/id_rsa-daint
  ProxyJump mbercx@ela.cscs.ch

Host ela.cscs.ch
  IdentityFile ~/.ssh/id_rsa-daint

```

`ssh daint-test` works fine. I first set up the computer with:

```console
$ verdi computer setup --config computer/daint-test.yaml 
Success: Computer<5> daint-test created
Info: Note: before the computer can be used, it has to be configured with the command:
Info:   verdi computer configure ssh daint-test
```
where

```console
$ more computer/daint-test.yaml 
---
label: daint-test
description: Piz Daint with XC50 Haswell GPU (12 CPUs, 1 GPU) - PRACE project
hostname: daint.cscs.ch
transport: ssh
scheduler: slurm
shebang: '#!/bin/bash'
work_dir: /scratch/snx3000/{username}/aiida
mpirun_command: srun -u -n {tot_num_mpiprocs} --hint=nomultithread --unbuffered
mpiprocs_per_machine: 12
prepend_text: |
   #SBATCH -A pr110
   #SBATCH -C gpu
append_text: ' '
```

I then configured the `ssh` interactively:

```console
$ verdi computer configure ssh daint-test
Info: enter "?" for help
User name [mbercx]: 
Port number [22]: 
Look for keys [True]: 
SSH key file []: 
Connection timeout in s [60]: 
Allow ssh agent [True]: 
SSH proxy jump []: mbercx@ela.cscs.ch
SSH proxy command []: 
Compress file transfers [True]: 
GSS auth [False]: 
GSS kex [False]: 
GSS deleg_creds [False]: 
GSS host [daint.cscs.ch]: 
Load system host keys [True]: 
Key policy (RejectPolicy, WarningPolicy, AutoAddPolicy) [RejectPolicy]: 
Use login shell when executing command [True]: 
Connection cooldown time (s) [30.0]: 
Info: Configuring computer daint-test for user mbercx@gmail.com.
Success: daint-test successfully configured for mbercx@gmail.com
```

When trying to run a job, it gets stuck in the `upload` step:

```console
$ verdi process list
  PK  Created    Process label    Process State    Process status
----  ---------  ---------------  ---------------  ----------------------------------
 559  5m ago     PwBaseWorkChain  ⏵ Waiting        Waiting for child processes: 564
 564  5m ago     PwCalculation    ⏵ Waiting        Waiting for transport task: upload

Total results: 2

Info: last time an entry changed state: 5m ago (at 17:25:28 on 2021-06-06)
```

And the report has the following error message:

```console
$ verdi process report 564
*** 564: None
*** Scheduler output: N/A
*** Scheduler errors: N/A
*** 1 LOG MESSAGES:
+-> ERROR at 2021-06-06 17:25:59.102930+00:00
 | Traceback (most recent call last):
 |   File "/home/mbercx/envs/aiida-dev/code/aiida-core/aiida/engine/utils.py", line 188, in exponential_backoff_retry
 |     result = await coro()
 |   File "/home/mbercx/envs/aiida-dev/code/aiida-core/aiida/engine/processes/calcjobs/tasks.py", line 80, in do_upload
 |     transport = await cancellable.with_interrupt(request)
 |   File "/home/mbercx/envs/aiida-dev/code/aiida-core/aiida/engine/utils.py", line 95, in with_interrupt
 |     result = await next(wait_iter)
 |   File "/usr/lib/python3.8/asyncio/tasks.py", line 608, in _wait_for_one
 |     return f.result()  # May raise f.exception().
 |   File "/usr/lib/python3.8/asyncio/futures.py", line 175, in result
 |     raise self._exception
 |   File "/home/mbercx/envs/aiida-dev/code/aiida-core/aiida/engine/transports.py", line 89, in do_open
 |     transport.open()
 |   File "/home/mbercx/envs/aiida-dev/code/aiida-core/aiida/transports/plugins/ssh.py", line 502, in open
 |     proxy_client.connect(proxy['host'], **proxy_connargs)
 |   File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/paramiko/client.py", line 435, in connect
 |     self._auth(
 |   File "/home/mbercx/.virtualenvs/aiida-dev/lib/python3.8/site-packages/paramiko/client.py", line 765, in _auth
 |     raise SSHException("No authentication methods available")
 | paramiko.ssh_exception.SSHException: No authentication methods available
```

My OpenSSH version is 7.6:

```console
$ ssh -V
OpenSSH_7.6p1 Ubuntu-4ubuntu0.3, OpenSSL 1.0.2n  7 Dec 2017
```

And I'm using Ubuntu 18.04.5 LTS:

```console
$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="18.04.5 LTS (Bionic Beaver)"
```

Let me know if I messed up somewhere! In this case, sorry for delaying the merge. 😅 
I've also left some minor comments re the documentation.

docs/source/howto/ssh.rst
```diff
@@ -185,47 +185,79 @@ Connecting to a remote computer *via* a proxy server
...
       Hostname FULLHOSTNAME_TARGET
       User USER_TARGET
       IdentityFile ~/.ssh/aiida
-      ProxyCommand ssh -W %h:%p USER_PROXY@FULLHOSTNAME_PROXY
+      ProxyJump USER_PROXY@FULLHOSTNAME_PROXY
+
+  Host FULLHOSTNAME_PROXY
+      IdentityFile ~/.ssh/aiida
+
+Replace the ``..._TARGET`` and ``..._PROXY`` variables with the host/user names of the respective servers.
```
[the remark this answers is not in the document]
dev-zero: There are many resources, rather than providing links to a commercial service I'd rather suggest that people use their favorite search engine.

docs/source/howto/ssh.rst
dev-zero, line 260:
```suggestion
.. important:: Specifying the ``proxy_command`` manually
```

dev-zero: Done, thanks.

dev-zero:
@mbercx difficult to say what went wrong since the error from the log is missing which would contain more information (especially the final login information used to login to the proxy).

ltalirz:
thanks @dev-zero 

@mbercx can you please check the changes and provide additional input on the issue you observed?
e.g. the configuration of the ela and daint hosts from your SSH config + the output of `verdi computer test daint`

mbercx (approved):
Hi @ltalirz and @dev-zero! So, it turns out the issue was related to the fact that I was relying on `look_for_keys: True`. I.e. when configuring the test computer with the following YAML file:

```yaml
username: mbercx
port: 22
look_for_keys: True
key_filename: ""
timeout: 60
allow_agent: True
proxy_jump: mbercx@ela.cscs.ch
proxy_command: ""
compress: True
gss_auth: False
gss_kex: False
gss_deleg_creds: False
gss_host: daint.cscs.ch
load_system_host_keys: True
key_policy: RejectPolicy
use_login_shell: True
safe_interval: 30.0
```
> Sidenote: I'm using double quotes to avoid prompts when configuring the computer with the `--config` option. Can you see any issues with that? 🤔 

This resulted in the error I also find when trying to run calculations when I do:

```console
$ verdi computer test daint-test
Info: Testing computer<daint-test> for user<mbercx@gmail.com>...
* Opening connection... 07/14/2021 06:24:16 AM <31330> aiida.transport.SshTransport: [ERROR] Error connecting to proxy 'ela.cscs.ch' through SSH: 
[SshTransport] No authentication methods available, connect_args were: {'username': 'mbercx', 'port': 22, 'look_for_keys': True, 'timeout': 60, 'allow_agent': True, 'compress': True, 'gss_auth': False, 'gss_kex': False, 'gss_deleg_creds': False, 'gss_host': 'daint.cscs.ch'}
[FAILED]: Error while trying to connect to the computer
  Use the `--print-traceback` option to see the full traceback.
Warning: 1 out of 0 tests failed
```

(I added another carriage return to make sure the the `No authentication methods available` is clearly visible.)

This may be related to the fact that I have a rather large number of keys in my `~/.ssh/` folder. If I specify the correct key in the filename:

```yaml
key_filename: /home/mbercx/.ssh/id_rsa-daint
```

and reconfigure the computer, all is well on the western front:

```console
$ verdi computer test daint-test
Info: Testing computer<daint-test> for user<mbercx@gmail.com>...
* Opening connection... [OK]
* Checking for spurious output... [OK]
* Getting number of jobs from scheduler... 07/14/2021 06:33:20 AM <32573> aiida.scheduler.slurm: [WARNING] Unrecognized job_state 'RD' for job id 31333865
[OK]: 1887 jobs found in the queue
* Determining remote user name... [OK]: mbercx
* Creating and deleting temporary file... [OK]
Success: all 5 tests succeeded
```

Also tried submitting some calculations, and didn't encounter anymore issues!

dev-zero:
Wrt large number of keys: AFAIK ssh considers every login attempt with a key it doesn't know as an unsuccessful login attempt and will likely disconnect after 3 attempts, probably resulting in the error message you see. If you have been using `ProxyCommand` before instead: this one was using your `~/.ssh/config` for connecting to ela (since an `ssh` command was run), hence it was using the settings you have in there for ela (which probably included a hard-coded key).

mbercx:
> Wrt large number of keys: AFAIK ssh considers every login attempt with a key it doesn't know as an unsuccessful login attempt and will likely disconnect after 3 attempts, probably resulting in the error message you see. If you have been using `ProxyCommand` before instead: this one was using your `~/.ssh/config` for connecting to ela (since an `ssh` command was run), hence it was using the settings you have in there for ela (which probably included a hard-coded key).

Right, makes sense, thanks @dev-zero! This all looks good to me. Just updated the branch, can you (squash and) merge? I typically prefer to write my own commit messages, so I'll leave that to you. 🙃

dev-zero:
@mbercx I'm not authorized to merge

mbercx:
Alright, happy to write a short commit message, but since it'll be associated with your account, maybe you'd like to write it yourself so I can add it when I merge?

ltalirz:
@dev-zero just gave you maintainer rights for the aiida-core repo; let me know if you're still unable to merge

dev-zero:
@ltalirz thanks, looks good

PR #4954, opened 2021-05-19
Disregard old test repository quirks for migration

### LINKED ISSUE #4910: Object store migration: Node folders `raw_input` vs. `path`
### Describe the bug

There is a known edge case that can happen during the migration to the object store repository, where a `Node` can have both a `path` and a `raw_input` folder in the current repository.
If this is encountered, the migration will fail with a `CRITICAL` and report the `Node` in question, describe the issue and instruct the user to delete one or the other folder.

At the AiiDA team developers meeting (on 05.05.2021) @giovannipizzi mentioned [his comment](https://github.com/aiidateam/aiida-core/pull/4345#issuecomment-749699623) regarding this issue.
His suggested solution is to: "ignore if both are there AND `path` is empty (then just pick `raw_input`)".

### Expected behavior

It would be nice if there is a solvable solution, where the migration doesn't fail, but simply reasons slightly, choosing the correct folder (if possible) and then continues. If this is not possible, then the migration should still fail with a `CRITICAL`, of course.

CasperWA:
An old test created a `.gitignore` file in the Node repository folder
`path`, while also creating files under a `raw_input` folder.
This adds a logic test for this specific edge-case, migrating the
`raw_input` folder and ignoring the `path` folder for the specific Node.

Fixes #4910.

### DIFF
--- a/aiida/backends/general/migrations/utils.py
+++ b/aiida/backends/general/migrations/utils.py
@@ -252,12 +252,16 @@ def get_node_repository_dirpaths(basepath, shard=None):
                 path = None
 
                 if 'path' in subdirs and 'raw_input' in subdirs:
-                    # If the `path` is empty, we simply ignore and set `raw_input` to be migrated, otherwise we add
-                    # the entry to `contains_both` which will cause the migration to fail.
-                    if os.listdir(dirpath / 'path'):
-                        contains_both.append(str(dirpath))
-                    else:
+                    # If the `path` folder is empty OR it contains *only* a `.gitignore`, we simply ignore and set
+                    # `raw_input` to be migrated, otherwise we add the entry to `contains_both` which will cause the
+                    # migration to fail.
+                    # See issue #4910 (https://github.com/aiidateam/aiida-core/issues/4910) for more information on the
+                    # `.gitignore` case.
+                    path_contents = os.listdir(dirpath / 'path')
+                    if not path_contents or path_contents == ['.gitignore']:
                         path = dirpath / 'raw_input'
+                    else:
+                        contains_both.append(str(dirpath))
                 elif 'path' in subdirs:
                     path = dirpath / 'path'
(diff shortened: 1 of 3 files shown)


chrisjsewell:
cheers @CasperWA, just to confirm there is not an analogous change required in the sqlalchemy migrations?

chrisjsewell:
well actually I guess I meant test for the sqlalchemy backend, since the actual change is in general utils, so I assume that applies to both

CasperWA:
> well actually I guess I meant test for the sqlalchemy backend, since the actual change is in general utils, so I assume that applies to both

Yeah, it seems the checks were done backend-agnostically here, which makes sense as it has nothing to do with the DB. But I'm unsure exactly what you mean otherwise here? Do you want a unittest for something or?

chrisjsewell:
> Ah, I think you do want another unit test, but for the SQLAlchemy backend. Hmm, I don't think it's necessary, but I can set it up if you want, just to be absolutely sure?

Yeh that was the point; I just initially noted there was a django named file, but not an sqla one lol.
If its possible to do easily/quickly, I guess it would be nice? (although I'm sure we have better things to do with our lives)

CasperWA:
> If its possible to do easily/quickly, I guess it would be nice? (although I'm sure we have better things to do with our lives)

Mnjeh - you'd say so wouldn't you? 😏 😅 

Anyway, I've made a test, but now I'm not remembering how to test the SQLAlchemy backend, what's the env var?

CasperWA:
> Anyway, I've made a test, but now I'm not remembering how to test the SQLAlchemy backend, what's the env var?

Nevermind, I figurereded it out. I'm totally sticky brain.

(But for real: `AIIDA_TEST_BACKEND=sqlalchemy pytest`...) 👍

chrisjsewell (approved):
Merge that bad boy!

PR #4956, opened 2021-05-20
remove ete3 dependency

ltalirz:
It seems the last use of ete3 was removed in commit
287d1385884c1c77a942d4945506c406ee98acde but the dependency was not
removed.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -17,7 +17,6 @@ dependencies:
 - click~=7.1
 - disk-objectstore~=0.5
 - django~=2.2
-- ete3~=3.1
 - python-graphviz~=0.13
 - ipython~=7.20
 - jinja2~=2.10
--- a/requirements/requirements-py-3.7.txt
+++ b/requirements/requirements-py-3.7.txt
@@ -33,7 +33,6 @@ disk-objectstore~=0.5
 Django==2.2.20
 docutils==0.15.2
 entrypoints==0.3
-ete3==3.1.2
 Flask==1.1.2
 Flask-Cors==3.0.10
 Flask-RESTful==0.3.8
--- a/requirements/requirements-py-3.8.txt
+++ b/requirements/requirements-py-3.8.txt
@@ -33,7 +33,6 @@ disk-objectstore~=0.5
 Django==2.2.20
 docutils==0.15.2
 entrypoints==0.3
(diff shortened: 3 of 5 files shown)


sphuber:
How did you discover this? If you haven't done so already, would it make sense to do a quick scan for other unused dependencies?

ltalirz:
Sure, we can have a look:

### pipreqs
```
$ pip install pipreqs
...
$ pipreqs --print --no-pin aiida | tr '[:upper:]' '[:lower:]' | sort
INFO: Successfully output requirements
aio_pika
aldjemy
alembic
archive_path
ase
beautifulsoup4
bpython
circus
click
click_completion
click_config_file
click_spinner
contextvars
dataclasses
disk_objectstore
docutils
flask
flask_cors
flask_restful
graphviz
ipython
jinja2
jsonschema
kiwipy
matplotlib
mayavi
numpy
pamqp
paramiko
pgsu
pgtest
phonopy
plumpy
psutil
psycopg2
psycopg2_binary
pymatgen
pymysql
pyparsing
pytest
python_dateutil
pytz
pyyaml
pyzmq
reentry
requests
scipy
simplejson
singledispatch
spglib
sphinx
sphinxcontrib_programoutput
sqlalchemy_utils
tabulate
tqdm
tzlocal
upf_to_json
werkzeug
wrapt
```

This misses the `django` and `sqlalchemy` imports; likely because we have folders named `django` and `sqlalchemy` in `aiida.orm.implementation`.

I've checked the rest from the `setup.json` against this list and could not find any other unused package.

### conda_forge_tick

You may have noticed that the conda-forge bot does exactly this analysis (see e.g. [here](https://github.com/conda-forge/aiida-core-feedstock/pull/52) ), which is why I'm quite confident that this was not a problem with any recent AiiDA releases, and I think the commit where deprecated functionality was removed is the most likely place for overlooking something like this.

The bot is using [this](https://github.com/regro/cf-scripts/blob/6773d125e3b70e9866682861e6710bd97b7147f3/conda_forge_tick/audit.py#L83-L90) bit of code.
I haven't installed & run it myself - you're welcome to do so if you like. Otherwise we'll know after the next release ;-)

sphuber (approved):
Thanks @ltalirz

PR #4959, opened 2021-05-21
🐛 FIX: REST API date-time query

### LINKED ISSUE #4957: REST API: Querying for datetime excepts
e.g. `https://15.188.110.176:5000/api/v4/nodes?mtime%3E=2019-04-23` leads to:

```
May 21 00:56:29 ip-172-31-44-253 verdi[13037]:   File "/home/ubuntu/venv/lib/python3.8/site-packages/aiida/common/hashing.py", line 172, in _
May 21 00:56:29 ip-172-31-44-253 verdi[13037]:     return [_single_digest('dict(')] + list(
May 21 00:56:29 ip-172-31-44-253 verdi[13037]:   File "/home/ubuntu/venv/lib/python3.8/site-packages/aiida/common/hashing.py", line 174, in <genexpr>
May 21 00:56:29 ip-172-31-44-253 verdi[13037]:     (k_digest + _make_hash(val, **kwargs)) for k_digest, val in sorted(hashed_key_mapping(), key=itemgetter(0))
May 21 00:56:29 ip-172-31-44-253 verdi[13037]:   File "/usr/lib/python3.8/functools.py", line 875, in wrapper
May 21 00:56:29 ip-172-31-44-253 verdi[13037]:     return dispatch(args[0].__class__)(*args, **kw)
May 21 00:56:29 ip-172-31-44-253 verdi[13037]:   File "/home/ubuntu/venv/lib/python3.8/site-packages/aiida/common/hashing.py", line 127, in _make_hash
May 21 00:56:29 ip-172-31-44-253 verdi[13037]:     raise HashingError(f'Value of type {type(object_to_hash)} cannot be hashed')
May 21 00:56:29 ip-172-31-44-253 verdi[13037]: aiida.common.exceptions.HashingError: Value of type <class 'aiida.restapi.common.utils.DatetimePrecision'> cannot be hashed
```

NinadBhat:
Fixes #4957 

Added hashing function for DatetimePrecision object.

@chrisjsewell Not sure if this is entirely right, but this method solved the error for me.


chrisjsewell (changes requested):
Thanks @NinadBhat! 
The fix looks fine, I think though it would be good to also add a test in `tests/restapi/test_routes.py:RESTApiTestSuite`, with an `mtime`/`ctime` filter on nodes

unkcpz (changes requested):
Here are some questions from my view:
1. We already know the `DatetimePrecision` has a `datatime` and an `int` as its objects, then is there any particular reason that you dispatch it to return like a `abc.Sequence` instead of dispatch it like `numbers.Complex` which also defined by two auguments.
2. I would suggest moving the `DatatimePrecision` definition from `aiida.restapi.common.utils` to `aiida.common`. Since `aiida.common` is a more 'internal' module then `aiida.restapi`. or maybe can put new code in `aiida.restapi.common.utils` and leave `aiida.common` untouched?
3. I believe would be also good to add test in `tests/common/test_hashing.py`.

unkcpz:
just look into the `tests/common/test_hashing.py`, since there are lots of tests assertion which compare expected with hash strings,  it would be suitable to use `pytest-regression` to do this.  But this will be in other PR. Also looking for your opinion @chrisjsewell .

NinadBhat:
@chrisjsewell, the one test failing is because docker does not install flask. Not sure where I should add flask dependency for it to be installed with docker.

unkcpz:
> the one test failing is because docker does not install flask. Not sure where I should add flask dependency for it to be installed with docker.

From my point of view, I think this is a side effect that I mentioned, where you import `restapi.common` in `aiida.common`. In fact, `aiida.common` would be better to know nothing about `aiida.restapi` I guess.

NinadBhat:
@unkcpz  makes sense. Thanks for the quick reply. @chrisjsewell  should I move `DatatimePrecision` to `aiida.common` ?

chrisjsewell:
>  should I move DatatimePrecision to aiida.common

Yeh I don't think that is a problem; it is a very simple class with no dependencies 

> We already know the DatetimePrecision has a datatime and an int as its objects, then is there any particular reason that you dispatch it to return like a abc.Sequence instead of dispatch it like numbers.Complex which also defined by two auguments.

meh, I'm not too bothered either (particularly as we are likely going to scrap all the current restapi code soon 😆). It would actually be kind of nice to know where/why it is actually being hashed in the first place; from a quick look in the code, this was not immediately apparent (but don't worry about spending too long on this)

> it would be suitable to use pytest-regression to do this. But this will be in other PR. Also looking for your opinion @chrisjsewell .

yeh always happy to use pytest-regression 👍 , although I don't think its a test dependency yet so yeh wouldn't add it in this PR

PR #4960, opened 2021-05-21
replace reentry by importlib.metadata

ltalirz:
Note: this is a quick port of an earlier fix intended for testing;
pre-commit still fails because of one missing feature + we may want to
clean up the codebase a bit.

### DIFF
--- a/aiida/plugins/entry_point.py
+++ b/aiida/plugins/entry_point.py
@@ -11,16 +11,35 @@
 import enum
 import traceback
 import functools
-
-try:
-    from reentry.default_manager import PluginManager
-    # I don't use the default manager as it has scan_for_not_found=True
-    # by default, which re-runs scan if no entrypoints are found
-    ENTRYPOINT_MANAGER = PluginManager(scan_for_not_found=False)
-except ImportError:
-    import pkg_resources as ENTRYPOINT_MANAGER
+import importlib
 
 from aiida.common.exceptions import MissingEntryPointError, MultipleEntryPointError, LoadingEntryPointError
+# try:
+#     from reentry.default_manager import PluginManager
+#     # I don't use the default manager as it has scan_for_not_found=True
+#     # by default, which re-runs scan if no entrypoints are found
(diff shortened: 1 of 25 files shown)


aiida/plugins/entry_point.py
```diff
@@ -11,18 +11,31 @@
...
+    """
+
+    def __init__(self):
+        self._map = None
+
+    def get_entry_point_map(self):
+        if self._map is None:
+            self._map = importlib.metadata.entry_points()
+
+        return self._map
```
dev-zero, line 35: since you're handing out r/w permission on `self._map` with it and only need the function for lazy initialization, I'd use:

```suggestion
    @property
    def map(self):
        if self._map is None:
            self._map = importlib.metadata.entry_points()

        return self._map
```

aiida/plugins/entry_point.py
```diff
@@ -238,6 +251,14 @@ def get_entry_point_names(group, sort=True):
     return entry_point_names
 
 
+def get_entry_point_map():
+    """
+    Return a dictionary of all entry point groups (=key) and entry points (=values).
+    :return: dictionary of all entry points
+    """
+    return ENTRYPOINT_MANAGER.get_entry_point_map()
```
dev-zero, line 259:
```suggestion
    return ENTRYPOINT_MANAGER.map
```

ltalirz:
thanks!

chrisjsewell:
Hey @ltalirz, just to note here, as of python 3.10 (https://docs.python.org/3.10/library/importlib.metadata.html#entry-points) and already implemented in the `importlib_metadata` port (https://importlib-metadata.readthedocs.io/en/latest/using.html#entry-points), they are moving away from the dictionary interface for retrieving entry points, to use the more explicit `select` method.

As an example, I have implemented this in another of my packages: https://github.com/executablebooks/MyST-NB/commit/a4c21b16be56725381fd3d829fde6d212d948906

ltalirz:
superseded by https://github.com/aiidateam/aiida-core/pull/5058

PR #4962, opened 2021-05-26
fix verdi computer show tab completion

unkcpz:
Got the following error when doing `verdi computer show <tab>`. This is a quick fix but just thinking when this change made in `DbComputer`, maybe other places also need to be take care of?

```
  File "/home/unkcpz/Projects/python-code/aiida_core/aiida/orm/implementation/querybuilder.py", line 399, in get_column 
    return getattr(alias, colname)                                                                                      
  File "/home/unkcpz/miniconda3/envs/aiida-core-dev/lib/python3.8/site-packages/sqlalchemy/orm/util.py", line 506, in __
getattr__                                                                                                               
    attr = getattr(target, key)                                                                                         
AttributeError: type object 'DbComputer' has no attribute 'name'
...
...
  File "/home/unkcpz/Projects/python-code/aiida_core/aiida/orm/querybuilder.py", line 1225, in _add_to_projections
    entity_to_project = self._get_projectable_entity(alias, column_name, attr_key, cast=cast)
  File "/home/unkcpz/Projects/python-code/aiida_core/aiida/orm/querybuilder.py", line 1199, in _get_projectable_entity
    entity = self._impl.get_column(column_name, alias)
  File "/home/unkcpz/Projects/python-code/aiida_core/aiida/orm/implementation/querybuilder.py", line 401, in get_column
    raise ValueError(
ValueError: name is not a column of aliased(DbComputer)
Valid columns are:
id
uuid
label
hostname
description
scheduler_type
transport_type
metadata
```

### DIFF
--- a/aiida/cmdline/params/types/computer.py
+++ b/aiida/cmdline/params/types/computer.py
@@ -41,7 +41,7 @@ class ComputerParamType(IdentifierParamType):
 
         :returns: list of tuples of valid entry points (matching incomplete) and a description
         """
-        return [(option, '') for option, in self.orm_class_loader.get_options(incomplete, project='name')]
+        return [(option, '') for option, in self.orm_class_loader.get_options(incomplete, project='label')]
 
 
 class ShebangParamType(StringParamType):
--- a/tests/cmdline/params/types/test_computer.py
+++ b/tests/cmdline/params/types/test_computer.py
@@ -8,88 +8,112 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Tests for the `ComputerParamType`."""
-from aiida.backends.testbase import AiidaTestCase
+import pytest
+
 from aiida.cmdline.params.types import ComputerParamType
 from aiida.orm.utils.loaders import OrmEntityLoader
 
 from aiida import orm
 
+# pylint: disable=redefined-outer-name
 
-class TestComputerParamType(AiidaTestCase):
-    """Tests for the `ComputerParamType`."""
 
-    @classmethod
-    def setUpClass(cls, *args, **kwargs):
-        """
-        Create some computers to test the ComputerParamType parameter type for the command line infrastructure
-        We create an initial computer with a random name and then on purpose create two computers with a name
-        that matches exactly the ID and UUID, respectively, of the first one. This allows us to test
-        the rules implemented to solve ambiguities that arise when determing the identifier type
-        """
-        super().setUpClass(*args, **kwargs)
+@pytest.fixture
+def parameter_type():
+    """Return an instance of the `ComputerParamType`."""
+    return ComputerParamType()
 
-        kwargs = {
-            'hostname': 'localhost',
-            'transport_type': 'local',
-            'scheduler_type': 'direct',
-            'workdir': '/tmp/aiida'
-        }
 
-        cls.param = ComputerParamType()
-        cls.entity_01 = orm.Computer(label='computer_01', **kwargs).store()
-        cls.entity_02 = orm.Computer(label=str(cls.entity_01.pk), **kwargs).store()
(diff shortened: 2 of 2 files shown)


aiida/cmdline/params/types/computer.py
```diff
@@ -41,7 +41,7 @@ def complete(self, ctx, incomplete):  # pylint: disable=unused-argument
 
         :returns: list of tuples of valid entry points (matching incomplete) and a description
         """
-        return [(option, '') for option, in self.orm_class_loader.get_options(incomplete, project='name')]
+        return [(option, '') for option, in self.orm_class_loader.get_options(incomplete, project='hostname')]
```
sphuber, line 44:
```suggestion
        return [(option, '') for option, in self.orm_class_loader.get_options(incomplete, project='label')]
```

sphuber (changes requested):
Thanks @unkcpz . The change of `DbComputer.name` to `DbComputer.label` was done in #4882 . It should have updated all other places in the code that still used `name` but this one was clearly missed and it is not tested. It would be great if you could add a quick test for it. Note that I think therefore we should also project on `label` instead of the hostname, because that is essentially what was being projected originally.

unkcpz:
Thanks @sphuber. The test added and also refactoring the unit test to use pytest.

unkcpz:
Well, seems run into pre-commit pylint warnings. Do we need to exclude `tests/` in pre-commit-pylint?

tests/cmdline/params/types/test_computer.py
```diff
@@ -8,88 +8,110 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Tests for the `ComputerParamType`."""
```
sphuber, line 10: # pylint: disable=redefined-outer-name

tests/cmdline/params/types/test_computer.py
```diff
@@ -8,88 +8,110 @@
...
-        result = self.param.convert(identifier, None, None)
-        self.assertEqual(result.uuid, self.entity_03.uuid)
+@pytest.fixture
+def parameter_type():
+    """Return an instance of the `ComputerParamType`."""
+    return ComputerParamType()
+
+
+@pytest.fixture
+def setup_computers(clear_database_before_test):
```
sphuber, line 26:
```suggestion
@pytest.mark.usefixtures('clear_database_before_test')
def setup_computers():
```

tests/cmdline/params/types/test_computer.py
```diff
@@ -8,88 +8,110 @@
...
+
+    options = [item[0] for item in parameter_type.complete(None, 'xa')]
+    assert sorted(options) == sorted([entity_04.label])
+
+
+def test_get_by_id(setup_computers, parameter_type):
+    """
+    Verify that using the ID will retrieve the correct entity
+    """
+    entity_01, entity_02, entity_03 = setup_computers
```
sphuber, line 60:
```suggestion
    entity_01, _, _ = setup_computers
```

sphuber (changes requested):
Thanks a lot for the test and refactoring @unkcpz  . The tests should also be linted so it should not be added to the exclude list. The errors are easy to address though. I made three suggestions that show how to do it. Just the one of unused variables you should apply to the other locations in the code that pylint mentioned

unkcpz:
@sphuber thanks!  It should be all right this time.

tests/cmdline/params/types/test_computer.py
sphuber: Uhm, we do it in plenty of places with no problems. So I doubt that it is not possible. What makes you think that besides that PR you linked?

sphuber (approved):

PR #4963, opened 2021-05-26
RFC: permit class member functions as calcfunctions

dev-zero:
It is already possible to specify a class member function as a "calcfunction"
and use it when not running via the daemon (and not with caching enabled):

    class MyWorkChain(WorkChain):
        @calcfunction
        def mysum(a, b):
            return a+b

        def workfunc(self):
            self.ctx.sum = MyWorkChain.mysum(self.inputs.x, self.inputs.y)

Here I simply extend the loading to unbreak caching and running via the daemon,
while brushing up error handling a bit (catch one exception more and cleanup the traceback).

### DIFF
--- a/aiida/engine/processes/functions.py
+++ b/aiida/engine/processes/functions.py
@@ -280,9 +280,10 @@ class FunctionProcess(Process):
             spec.outputs.valid_type = (Data, dict)
 
         return type(
-            func.__name__, (FunctionProcess,), {
+            func.__qualname__, (FunctionProcess,), {
                 '__module__': func.__module__,
                 '__name__': func.__name__,
+                '__qualname__': func.__qualname__,
                 '_func': staticmethod(func),
                 Process.define.__name__: classmethod(_define),
                 '_func_args': args,
--- a/aiida/orm/nodes/process/process.py
+++ b/aiida/orm/nodes/process/process.py
@@ -219,14 +219,25 @@ class ProcessNode(Sealable, Node):
         except exceptions.EntryPointError as exception:
             raise ValueError(
                 f'could not load process class for entry point `{self.process_type}` for Node<{self.pk}>: {exception}'
-            )
+            ) from exception
         except ValueError:
-            try:
-                import importlib
-                module_name, class_name = self.process_type.rsplit('.', 1)
-                module = importlib.import_module(module_name)
-                process_class = getattr(module, class_name)
-            except (AttributeError, ValueError, ImportError) as exception:
+            import importlib
+
+            def str_rsplit_iter(string, sep='.'):
+                components = string.split(sep)
+                for idx in range(1, len(components)):
+                    yield sep.join(components[:-idx]), components[-idx:]
+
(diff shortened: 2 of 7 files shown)


greschd:
As far as I can see, this doesn't bind either the class or the instance in any way - should there maybe be a `@staticmethod` somewhere to indicate that?

dev-zero:
@greschd at first I thought I'd have to do that too. Indeed my first working attempt was this (since an unbound staticmethod is not a callable you have to call it via the descriptor protocol):

```
class MyWorkChain(WorkChain):
    @staticmethod
    def mysum(a, b):
        return a + b

MyWorkChain.mysum = calcfunction(MyWorkChain.mysum)
```

But the `@calcfunction` decorator wraps/replaces the function with a new type anyway, hence it doesn't seem necessary.

greschd:
Right, I understand the `@staticmethod` isn't actually necessary. What I'm a bit concerned about is that the decorated function _looks_ very similar to a plain method (that would bind `self`), but doesn't act that way. I'm wondering if there's some way we could make that distinction more clear.

dev-zero:
Good point, yes.

Maybe some more things:

* while doing testing related to this I encountered some weird database errors when a process could not be loaded (e.g. when loading from caching). My guess is that the loading failed but the in the error path the exception would have been added to the incompletely loaded process node.
* I started messing with this since I wanted to see whether it would be possible to have class member calcfunctions which pull their arguments from the context or input automatically, such that

```
    @calcfunction
    def mysum(a, b):
        return a+b

    def workfunc(self):
        self.ctx.sum = MyWorkChain.mysum(self.ctx.x, self.ctx.y)
```

could be written as

```
    @calcfunction
    @args_from_context  # automatically passes members x, y from self.ctx
    def sum(x, y):
        return x+y
```

In the sense that such functions could be used in the outline, but also directly act as calcfunctions. So instead of having explicit delayed/async function calls as calcfunctions those class member calcfunctions in the outline would then more or less push/pop from the ctx "stack".

greschd:
>     * while doing testing related to this I encountered some weird database errors when a process could not be loaded (e.g. when loading from caching). My guess is that the loading failed but the in the error path the exception would have been added to the incompletely loaded process node.

Happy to take a look at the caching case if you can create a reproducing example.

I like the idea of a calcfunction that operates directly on the ctx.

dev-zero:
My guess would be that the `self` here is serializable, while a workchain class instance is not.

sphuber:
@dev-zero as you may have noticed, I have been going through old PRs to clean house a bit. This one I think can still be merged, so I took the opportunity to rebase your branch, hope you don't mind. I have also added tests that test the proposed functionality. I am testing two ways of declaring the calcfunction; as a proper staticmethod of the workchain and as a class attribute. The tests validate that it can be run as well as submitted and also that caching works. The question is now whether we should officially recommend these approaches in the documentation. Maybe we should only recommend the staticmethod approach as it seems to make explicit why `self` is not getting bound and passed the the calcfunction when invoked. Thoughts?

sphuber (approved):

PR #4965, opened 2021-05-31
ADD: New repository CLI

ramirezfranciscof:
This PR incorporates the new command line tool to control the maintenance tasks of the repository (fix 4321).

The main challenge of this task is reconciling (1) the need for specific control over the backend and the processes happening underneath and (2) a general frontend interface that is simple for the user and versatile to control many possibly different backends. This is something that is a common issue of many software designs, but this is a particular case where the specifics of the underlying processes are quite important and hiding them behind a generic interface could render the whole feature unusable.

### Current Implementation

There is a single command `verdi repository maintain`, that will perform a full maintenance procedure on the repository backend. This procedure might be time intensive and it requires the user to stop using the database to prevent any data corruption, so a proper warning is displayed asking for confirmation (there is also the possibility of guaranteeing the safety with profile locking, see below).

The command also has the flag `--live` to indicate that only maintenance tasks that can be done while still using AiiDA should be executed. Again, a warning lets the user know that this is not the full procedure and that they should run the full command once they have the time.

I think that this characteristic is essentially the critical minimal information the user needs to handle: can I do this while still using AiiDA or do I need to do a "downtime"? Even considerations of performance control (doing a quick maintenance vs and an in-depth) are secondary, at least in the sense that they are more relevant for "downtime" maintenance but not so much for the "live" option.

Finally I added a `--pass_down` option that accepts a string that gets send to the backend and can be used for testing, performance analysis, or power-using specific backends. You can see how this is currently being used to have a finer grain control of the different stages of maintenance in the objectstore repository backend.

### The case of deletion

One last thing that would be common to all repository backends is the "propagation of the deletion of files". The issue is that the deletion of data only takes direct effect in the AiiDA database (removing the reference from there) without affecting the content of the backend. Therefore it needs to be specifically propagated to the repository backend, a process which is currently performed every time the maintenance is run (with or without `--live`). This is not strictly linked to the underlying maintenance operations though, and in principle we could separate this command so that users can propagate the deletion to the backend independently.

I chose to initially have this be part of the maintenance so as to present a simpler interface, since it is not even guaranteed that this will have any beneficial effect on its own (for example, if the backend also does soft delete and keeps the unreferenced files around until a full maintenance is executed). This means however that it can't be controlled by users externally since this is performed on the AiiDA side and not part of the backend (and thus should not be influenced by things in `pass_down`). In principle for the objectstore they can do this propagation only if they `pass_down` the options to cancel all backend operations, but (1) this is very backend specific and (2) they can't currently choose to skip it in this way. I'm considering adding a `--skip-propagation` flag for this purpose, but I think since it is an addition it is not critical to have it now (unless there is some specific performance issue with this part).

### Tasks:

- [ ] Incorporate daemon locking mechanism (see [this PR](https://github.com/aiidateam/aiida-core/pull/5135))
- [ ] Add documentation section for new command.


ramirezfranciscof:
**This draft does not yet implement the simplified interface that I have described in the OP**. Instead, the interface is left "completely" transparent, in the sense that it allows to easily enable/disable each of the following operations:

* "Pack" files: the [disk-objectstore](https://github.com/aiidateam/disk-objectstore) will initially keep all individual files "loose", and through this command you can tell it to store them in packs (this duplicates the data).
* "Clean" loose files: Remove the loose files that have already been packed (necessary to free the space of the packed files).
* "Transmit" deletion of files: when the user deletes files, they are just making AiiDA remove its internal reference to them in the objectstore. In order to actually delete them from there, one needs to run this to take stock of all objects aiida still keeps track of and ask the repository to specifically remove all the ones that are not being kept track of.
* "Repack" files: when deleting objects in the [disk-objectstore](https://github.com/aiidateam/disk-objectstore), the loose files are removed but the pack files will just lose their reference in the internal database of the objectstore. This re-organizes the packs and removes all data that is not being referenced there.
* "Vacuum" the repository DB: this will re-order the rows of the database so that the tables are optimized for the current status of the packs (which may have changed due to deletion or repacking).

This is so to be able to determine some basic heuristics of the different times that it takes for the [disk-objectstore](https://github.com/aiidateam/disk-objectstore) to perform each process and better figure out which can be associated together without much penalty. Specially within the subgroups that can be performed with the daemon running (pack/transmit) and the ones that can't (clean/repack/vacuum). For this purpose I have also created this [companion script](https://github.com/ramirezfranciscof/aiida-toolbox-ffr/blob/main/active/issue4321/test00_timing.py) (you can download and pip-install the full repository in an AiiDA environment, as the script makes use of other tools in there). If you want to test it yourself you can run it by using the following:

```
(aiida) $  ./test00_timing.py --mid-files
```

The test performs the following procedures:
1. Creates the test repository (takes around 30 min for me)
2. Does a first deletion of 50 nodes and then runs the whole series of "transmit" => "clean" => "vacuum" (twice, to get a baseline of how expensive it is to run these commands when they have nothing to do and also check that effectively there are no cross-related effects).
3. Pack the files of the remaining 150 nodes (twice, to get baseline for "unnecessary" runs) and run the series of "transmit/clean/vacuum" twice.
4. Delete another set of 49/50 nodes (now packed) and run the series of "transmit/clean/vacuum" twice.
5. Repack the files of the remaining 100/101 nodes (twice too) and run the series of "transmit/clean/vacuum" twice.

You can see my full output below, but my summary of it would be:
* Deleting around 250 loose files (in 50 nodes) totalling 2gb takes less than 20 seconds. The baseline (the query process of finding unreferenced files) seems to take only 1-2 seconds (although this DB of 200 nodes / 1000 files is rather small, I want to try in some larger one).
* Packing 6gb of files takes half a minute. Extrapolating a bit, it would mean that packing a 1tb repository could take 1 hour and a half, which seems pretty decent. It is a risky extrapolation and having the disk writing stuff for over an hour is not good for the computer, so periodic packaging should be advised. Baseline if there is nothing to do is around one second.
* Cleaning packed files seems to be faster than deleting a smaller number of loose files? (20 secs vs 2.5 secs?) Maybe this is because the backend only exposes the option to delete a single object so I have to call that for all single files (see line 39 of `aiida.repository.control.py`), while it would be more efficient to give the object store a list of files to delete and have it take care of it (i.e. expose `self.container.delete_objects(...)` in the methods of the abstract repository backend; for example, see line 101 in the `aiida.repository.backend.disk_object_store.py`).
* Deleting packed nodes is significantly more expensive than deleting loose nodes (90sec vs 20sec, although again I would like to check what happens if I unify the deletion in a single call).
* Repacking seems to be "expensive" even if there is not "much to do" (doing two consecutive calls takes 15 secs both, with no reduction in the second one). Again, as with all these results, I would like to see how it changes with many small files.
* Vacuuming seems to always take around 1 second no matter what is there to vacuum.

<details>
 <summary>
Full output
 </summary>

 ```
Running test for "--mid-files"...

 > Now setting up the repository...
 >>> Deleting all nodes...
 >>> Cleaning up database and repository...
 >>> Populating the database...
 > Elapsed time: 1636.7547591612674
 > The database currently holds 200 nodes and the repo occupies 8.1G

 > Now deleting unpacked nodes...
 > Elapsed time: 0.10452547576278448
 > The database currently holds 150 nodes and the repo occupies 8.1G
 > Now cleaning deleted unpacked nodes...
 >>> Elapsed time (transmit): 17.87814964680001
 >>> Elapsed time (clean): 1.1770451520569623
 >>> Elapsed time (vacuum): 1.1977200941182673
 > The database currently holds 150 nodes and the repo occupies 5.8G
 >>> Elapsed time (transmit): 1.6738419053144753
 >>> Elapsed time (clean): 1.1854597837664187
 >>> Elapsed time (vacuum): 1.1461118678562343
 > The database currently holds 150 nodes and the repo occupies 5.8G

 > Now packing the nodes...
 > Elapsed time (pack): 37.37593990098685
 > The database currently holds 150 nodes and the repo occupies 12G

 > Now packing the nodes...
 > Elapsed time (pack): 1.2292702598497272
 > The database currently holds 150 nodes and the repo occupies 12G
 > Now cleaning repacked nodes...
 >>> Elapsed time (transmit): 1.7674363791011274
 >>> Elapsed time (clean): 2.378125532064587
 >>> Elapsed time (vacuum): 1.111530366819352
 > The database currently holds 150 nodes and the repo occupies 5.7G
 >>> Elapsed time (transmit): 1.69200792722404
 >>> Elapsed time (clean): 0.9920809953473508
 >>> Elapsed time (vacuum): 1.1101957960054278
 > The database currently holds 150 nodes and the repo occupies 5.7G

 > Now deleting packed nodes...
 > Elapsed time: 0.09688124433159828
 > The database currently holds 101 nodes and the repo occupies 5.7G
 > Now cleaning deleted packed nodes...
 >>> Elapsed time (transmit): 92.88936862908304
 >>> Elapsed time (clean): 1.0394411101005971
 >>> Elapsed time (vacuum): 1.0659767519682646
 > The database currently holds 101 nodes and the repo occupies 5.8G
 >>> Elapsed time (transmit): 1.5534060336649418
 >>> Elapsed time (clean): 1.0056658750399947
 >>> Elapsed time (vacuum): 1.0443645529448986
 > The database currently holds 101 nodes and the repo occupies 5.8G

 > Now repacking nodes...
 > Elapsed time: 16.907301522791386
 > The database currently holds 101 nodes and the repo occupies 3.7G

 > Now repacking nodes...
 > Elapsed time: 17.614450520370156
 > The database currently holds 101 nodes and the repo occupies 3.7G
 > Now cleaning deleted packed nodes...
 >>> Elapsed time (transmit): 1.517629874870181
 >>> Elapsed time (clean): 1.0210560159757733
 >>> Elapsed time (vacuum): 1.0760509828105569
 > The database currently holds 101 nodes and the repo occupies 3.7G
 >>> Elapsed time (transmit): 1.5449293879792094
 >>> Elapsed time (clean): 1.0809445828199387
 >>> Elapsed time (vacuum): 1.1022412879392505
 > The database currently holds 101 nodes and the repo occupies 3.7G
 ```

 </details>

I would like to try this with a lot of small files (see `--small-files` flag) to confirm the timings, but so far I haven't been able to run it to completion (several problems with the connection to my remote machine closing down before finishing creating the DB, this one seems to take a lot more time). I would like to wait until I have this before making any generalized statement supporting or supplementing the implementation proposed in the OP, but wanted to show the current state in case somebody had some feedback or questions.

aiida/backends/general/migrations/utils.py
```diff
@@ -86,43 +87,36 @@ class NoopRepositoryBackend(AbstractRepositoryBackend):
...
+    def has_objects(self, keys: List[str]) -> List[bool]:
+        raise NotImplementedError()
 
-        :param key: fully qualified identifier for the object within the repository.
-        :return: True if the object exists, False otherwise.
-        """
+    def delete_objects(self, keys: List[str]) -> None:
+        raise NotImplementedError()
+
+    def list_objects(self) -> List[str]:
```
chrisjsewell, line 116:
```suggestion
    def list_objects(self) -> Iterable[str]:
```

aiida/repository/backend/abstract.py
```diff
@@ -84,15 +85,43 @@ def put_object_from_file(self, filepath: typing.Union[str, pathlib.Path]) -> str
...
     def has_object(self, key: str) -> bool:
         """Return whether the repository has an object with the given key.
 
         :param key: fully qualified identifier for the object within the repository.
         :return: True if the object exists, False otherwise.
         """
+        return self.has_objects([key])[0]
+
+    @abc.abstractmethod
+    def list_objects(self) -> Iterator[str]:
```
chrisjsewell, line 107:
```suggestion
    def list_objects(self) -> Iterable[str]:
```

To clarify: Every iterator is also an iterable, but not every iterable is an iterator.
For example, a list is iterable but a list is not an iterator.
(an iterator can be created from an iterable by using the function `iter()`)

aiida/repository/backend/abstract.py
```diff
@@ -84,15 +85,43 @@ def put_object_from_file(self, filepath: typing.Union[str, pathlib.Path]) -> str
...
+        """Lists all available objects by key.
```
chrisjsewell, line 108:
```suggestion
        """Yields all available objects by key.
```

aiida/repository/backend/abstract.py
```diff
@@ -84,15 +85,43 @@ def put_object_from_file(self, filepath: typing.Union[str, pathlib.Path]) -> str
...
+
+        :return: A list of all the available object keys.
```
chrisjsewell, line 110:
```suggestion
        :return: An iterable of all the available object keys.
```

aiida/repository/backend/disk_object_store.py
```diff
@@ -90,15 +86,12 @@ def open(self, key: str) -> typing.Iterator[typing.BinaryIO]:
...
+        super().delete_objects(keys)
+        self.container.delete_objects(keys)
 
-        :param key: fully qualified identifier for the object within the repository.
-        :raise FileNotFoundError: if the file does not exist.
-        :raise OSError: if the file could not be deleted.
-        """
-        super().delete_object(key)
-        self.container.delete_objects([key])
+    def list_objects(self) -> Iterator[str]:
```
chrisjsewell, line 93:
```suggestion
    def list_objects(self) -> Iterable[str]:
```

aiida/repository/backend/sandbox.py
```diff
@@ -110,12 +110,13 @@ def open(self, key: str) -> typing.Iterator[typing.BinaryIO]:
...
+        for key in keys:
+            os.remove(os.path.join(self.sandbox.abspath, key))
 
-        :param key: fully qualified identifier for the object within the repository.
-        :raise FileNotFoundError: if the file does not exist.
-        :raise OSError: if the file could not be deleted.
-        """
-        super().delete_object(key)
-        os.remove(os.path.join(self.sandbox.abspath, key))
+    def list_objects(self) -> Iterator[str]:
```
chrisjsewell, line 118:
```suggestion
    def list_objects(self) -> Iterable[str]:
```

tests/repository/backend/test_abstract.py
```diff
@@ -30,7 +31,19 @@ def erase(self):
...
+    def _put_object_from_filelike(self, handle: BinaryIO) -> str:
+        pass
+
+    def delete_objects(self, keys: List[str]) -> None:
+        pass
+
+    def has_objects(self, keys: List[str]) -> List[bool]:
+        pass
+
+    def list_objects(self) -> List[str]:
```
chrisjsewell, line 43:
```suggestion
    def list_objects(self) -> Iterable[str]:
```

chrisjsewell (changes requested):
Heya, also need `list_objects` for the archive, so just checking we are on the same page

chrisjsewell:
> Also need `list_objects` for the archive, so just checking we are on the same page

might end up extracting this in to a separate PR/commit

aiida/repository/control.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # Propagate the deleted nodes to the repository backend
+    keyset_existing = set(repository.list_objects())
+
+    keyset_recorded = set()
+    query = orm.QueryBuilder()
+    query.append(orm.Node, project=['repository_metadata'])
+    for metadata, in query.iterall():
+        node_repo = Repository.from_serialized(backend=repository, serialized=metadata)
+        keyset_recorded.update(node_repo.get_file_keys())
```
chrisjsewell, line 34: this can all be replaced by:

```suggestion
   keyset_recorded = set(orm.Node.objects(backend).iter_repo_keys(batch_size=batch_size))
```

aiida/repository/control.py
```diff
@@ -0,0 +1,43 @@
...
+    """Performs maintenance tasks on the repository."""
+    from aiida import orm
+    from aiida.manage.manager import get_manager
+    from aiida.repository import Repository
+
+    if not live:
+        print('Here I should be trying to lock down the profile...')
+
+    manager = get_manager()
+    repository = manager.get_backend().get_repository()
```
chrisjsewell, line 24: best to allow the backend to be optionally passed

ramirezfranciscof:
I will now be working in adding some tests before setting this back to a PR but I think almost all the functionality is there. There are only two things missing that I think are modular enough that it would be best to add in separate PRs, so as not to make this one bigger:

1. Getting the size of unreferenced files besides the count in `verdi storage maintain --dry-run` (currently there is no backend method to ask for size of objects)
2. The `verdi storage integrity` checks.

On the other hand, mypy is driving me a bit insane. I don't understand very well what it is complaining about and I tried googling for these problems without much success. The only thing I managed to found / suspect is the it gets cranky when dealing with `dict`s that don't have a clearly pre-defined schemas (and even when they do, if they are minimally complex then you need to define this schema in type classes or something [like this](https://stackoverflow.com/a/63659664)?). Anyways, do you understand what it is happening here and how to fix it @sphuber ?


```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

aiida/repository/backend/disk_object_store.py:131: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "bool")  [assignment]
aiida/repository/backend/disk_object_store.py:134: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "bool")  [assignment]
aiida/repository/backend/disk_object_store.py:137: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "bool")  [assignment]
aiida/repository/backend/disk_object_store.py:140: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "bool")  [assignment]
aiida/repository/backend/disk_object_store.py:169: error: Incompatible types in assignment (expression has type "int", target has type "str")  [assignment]
aiida/repository/backend/disk_object_store.py:171: error: Incompatible types in assignment (expression has type "Dict[str, int]", target has type "str")  [assignment]
aiida/repository/backend/disk_object_store.py:175: error: Incompatible types in assignment (expression has type "Dict[str, float]", target has type "str")  [assignment]
aiida/repository/control.py:29: error: Unsupported target for indexed assignment ("Collection[str]")  [index]
aiida/repository/control.py:30: error: Unsupported target for indexed assignment ("Collection[str]")  [index]
aiida/repository/control.py:35: error: Unsupported target for indexed assignment ("Collection[str]")  [index]
Found 10 errors in 2 files (checked 104 source files)
```

ramirezfranciscof:
Tests are also almost ready. There is an error I can't reproduce locally (perhaps because I am using a test profile instead of having one be created for the test?), which I'm not sure how to solve. Apparently the database created for the test does not have a default user?

```
        backend = get_manager().get_backend()
>       assert orm.User.objects(backend).get_default() is not None
E       AssertionError: assert None is not None
```

I don't quite understand why it is problematic here but not in other tests that create nodes (I added the previous line to check but before this was raising on the following command `datanode = orm.FolderData()`, because the `__init__` of the node requires a `User`).

I'll keep trying to see if I can find out what the problem is, but if you have any idea it will be welcome @sphuber (specially since I can't reproduce this locally).

sphuber:
> Tests are also almost ready. There is an error I can't reproduce locally (perhaps because I am using a test profile instead of having one be created for the test?), which I'm not sure how to solve. Apparently the database created for the test does not have a default user?

You have commented out the `clear_database_before_storage` fixture. This is probably necessary as it will ensure that the database is properly configured. We are in a weird halfway state of using pytest and still the old-style tests, and for now it is necessary to use this fixture for all pytest based tests, even if it doesn't really require the database. This can be tricky to debug because the problem only shows up if you run the test together with some old-style tests. If you run the pytest test in isolation it will work.

ramirezfranciscof:
> You have commented out the `clear_database_before_storage` fixture. This is probably necessary as it will ensure that the database is properly configured. We are in a weird halfway state of using pytest and still the old-style tests, and for now it is necessary to use this fixture for all pytest based tests, even if it doesn't really require the database. This can be tricky to debug because the problem only shows up if you run the test together with some old-style tests. If you run the pytest test in isolation it will work.

Sorry, you were probably looking at one of my tests (since I can't reproduce locally I have been changing and pushing to see the effects here). The original test was using the `'clear_storage_before_test'` fixture, which should in turn be the one using `'clear_database_before_test'`:

```
def clear_storage_before_test():
    """Clears the storage before a test."""
    ...
```

For some reason when I use `@pytest.mark.usefixtures('clear_database_before_test')` on actual test `test_get_unreferenced_keyset` then it works but if I use `@pytest.mark.usefixtures('clear_storage_before_test')` then it doesn't. May be that the `function` scope doesn't get along concatenated fixtures? Not sure. Now I'm checking what happens when I use both directly on the test...

tests/repository/test_control.py
```diff
@@ -0,0 +1,146 @@
...
+
+    repository = get_manager().get_backend().get_repository()
+    object_keys = list(repository.list_objects())
+    repository.delete_objects(object_keys)
+    repository.maintain(full=True)
+
+
+@pytest.mark.usefixtures('clear_database_before_test', 'clear_storage_before_test')
+def test_get_unreferenced_keyset():
+    """Test the ``get_unreferenced_keyset`` method."""
```
ramirezfranciscof, line 60: This test fails if I don't include the `'clear_database_before_test'` fixture for the `test_get_unreferenced_keyset`. I don't understand why it doesn't suffice with `'clear_storage_before_test'`, since it is a fixture that itself uses the other one. If I don't do it this way I get an error when trying to use the orm complaining about the user not being set...

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +118,63 @@ def get_object_hash(self, key: str) -> str:
...
+            pack_loose = kwargs.get('override_pack_loose')  # type: ignore
+
+        if 'override_do_repack' in kwargs:
+            do_repack = kwargs.get('override_do_repack')  # type: ignore
+
+        if 'override_clean_storage' in kwargs:
+            clean_storage = kwargs.get('override_clean_storage')  # type: ignore
+
+        if 'override_do_vacuum' in kwargs:
+            do_vacuum = kwargs.get('override_do_vacuum')  # type: ignore
```
ramirezfranciscof, line 140: I had to disable mypy here because I am getting errors I don't understand:

```
aiida/repository/backend/disk_object_store.py:131: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "bool") [assignment] 
aiida/repository/backend/disk_object_store.py:134: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "bool") [assignment] 
aiida/repository/backend/disk_object_store.py:137: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "bool") [assignment] 
aiida/repository/backend/disk_object_store.py:140: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "bool") [assignment] 
```

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +118,63 @@ def get_object_hash(self, key: str) -> str:
...
+
+        output_info['Objects'] = {  # type: ignore
+            'unpacked': files_data['loose'],
+            'packed': files_data['packed'],
+        }
+        output_info['Size (MB)'] = {  # type: ignore
+            'unpacked': size_data['total_size_loose'] * bytes_to_mb,
+            'packed': size_data['total_size_packfiles_on_disk'] * bytes_to_mb,
+            'other': size_data['total_size_packindexes_on_disk'] * bytes_to_mb,
+        }
```
ramirezfranciscof, line 179: I had to disable mypy here because I am getting errors I don't understand:

```
aiida/repository/backend/disk_object_store.py:169: error: Incompatible types in assignment (expression has type "int", target has type "str") [assignment] 
aiida/repository/backend/disk_object_store.py:171: error: Incompatible types in assignment (expression has type "Dict[str, int]", target has type "str") [assignment] 
aiida/repository/backend/disk_object_store.py:175: error: Incompatible types in assignment (expression has type "Dict[str, float]", target has type "str") [assignment] 
```

aiida/repository/control.py
```diff
@@ -0,0 +1,81 @@
...
+
+    unreferenced_objects = get_unreferenced_keyset(aiida_backend=backend)
+
+    maintainance_report['repository']['unreferenced'] = len(unreferenced_objects)  # type: ignore
+    maintainance_report['repository']['info'] = repository_backend.get_info()  # type: ignore
+
+    repository_backend.delete_objects(list(unreferenced_objects))
+
+    # Perform the maintainance operations in the repository
+    maintainance_report['repository']['maintain'] = repository_backend.maintain(full=full, **kwargs)  # type: ignore
```
ramirezfranciscof, line 35: I had to disable mypy here because I am getting errors I don't understand:

```
aiida/repository/control.py:29: error: Unsupported target for indexed assignment ("Collection[str]") [index] 
aiida/repository/control.py:30: error: Unsupported target for indexed assignment ("Collection[str]") [index] 
aiida/repository/control.py:35: error: Unsupported target for indexed assignment ("Collection[str]") [index] 
```

ramirezfranciscof (commented):
I still can't figure out the two previous problems, but I'm now going to stop trying things out so that this can be reviewed. I set the typing to ignore and I am using the fixture in what would seem to me to be a "redundant" way, but at least all tests pass now.

I would like to not have to use these kind of patches here but the problems seem obscure and/or hard to debug, so in order to keep progress with the feature I think it would be better to just go ahead and I can open separate issues for dealing with them in the future.

aiida/repository/backend/disk_object_store.py
sphuber: Problem here is that those variables (`pack_loose` etc.) are declared as having type `bool`. But the `get` method of a dictionary will return `Optional[Any]`, i.e. any type or `None`. This means you can assign `None` or any other value to what is supposed to be a bool.

Why do you use `**kwargs` in the method signature? If you just explicitly define them, you don't have this problem. Plus, you can get rid of all these initialization rules and just define proper defaults. In any case, I would have used the following syntax:
```python
pack_loose = kwargs.get('override_pack_loose', True)
```
But really the question is why you chose this design anyway? Why have the caller use the argument prefixed with `override_` if you are going to change it internally. Doesn't it make way more sense to just have `pack_loose` as the method argument?

ramirezfranciscof: > Problem here is that those variables (`pack_loose` etc.) are declared as having type `bool`.

I mean, where? Those are internal variables, I never explicitly tell mypy they are anything. Does it interpret this from me doing `pack_loose = True` or `do_repack = full`? Does that mean I can no longer re-assign internal variables to "different" types?

> Plus, you can get rid of all these initialization rules and just define proper defaults. In any case, I would have used the following syntax:
> ```python
> pack_loose = kwargs.get('override_pack_loose', True)
> ```

Mmm this is true. This was a leftover of a previous way of setting the initial values, which I defaulted to the bools for a fast maintenance and then changed them if the user was doing `full`. But now that I use the `full` flag to assign it as the default value itself I guess I can use that with the overrides as well.

> Why do you use **kwargs in the method signature? If you just explicitly define them, you don't have this problem.
> ...
> Why have the caller use the argument prefixed with override_ if you are going to change it internally. Doesn't it make way more sense to just have pack_loose as the method argument?

I think the answer to both of these is the same: the method first needs to decide if it will set these values for a `full` or a `partial` maintenance, and afterwards the overrides can change any of those values independently.

If I was setting defaults in the signature then I would have to choose one (`full` or a `partial`) and then in the body detect if the other is being performed and adapt the values. In that case, I would then be overriding the more specific "overrides" requested by the user. Although you could say that technically the same resulting procedures are available either case, I think it makes more sense that the more specific indications gets respected instead of it being stomp over the "general" setup.

sphuber: > I mean, where? Those are internal variables, I never explicitly tell mypy they are anything. Does it interpret this from me doing pack_loose = True or do_repack = full?

Yes, by assigning a bool to those variables, `mypy` will interpret that this variable is supposed to represents a boolean. So when you then assign `None` to it, it will raise a warning. Note that it _is_ just a warning and you may want to allow this behavior, in which case ignoring the warning is reasonable. However, note that in this case, it probably _is_ a bug. You don't type check and so you could assign anything to those variables, for example `None`, and this could have unintended consequences downstream.

> I think the answer to both of these is the same: the method first needs to decide if it will set these values for a full or a partial maintenance, and afterwards the overrides can change any of those values independently.

Fair enough, but I think this really is just a problem of the design of the signature to begin with. I don't think the concept of `full` and `partial` should be baked into maintain. That seems to be a high-level concept that the CLI will have, but the lower-level interface doesn't need this. I would simplify this to the following signature:
```python
def maintain(self, pack_loose = True, repack = False, clean_storage = False, vacuum = False) -> dict:
```
i.e. with defaults that correspond to your `partial` and then the CLI can call this with all `True` if the user requests a `full` maintenance.

ramirezfranciscof: > Yes, by assigning a bool to those variables, `mypy` will interpret that this variable is supposed to represents a boolean. So when you then assign `None` to it, it will raise a warning. Note that it _is_ just a warning and you may want to allow this behavior, in which case ignoring the warning is reasonable. However, note that in this case, it probably _is_ a bug. You don't type check and so you could assign anything to those variables, for example `None`, and this could have unintended consequences downstream.

I think I get the distinction you are trying to make with it being a superfluous warning or an actual bug catcher. But these are all runtime user end inputs, so the typing is not enforced and adding it wouldn't really catch anything. Would mypy recognize if I do some manual typechecking with `isinstance` to guard for this or what is the option?


> Fair enough, but I think this really is just a problem of the design of the signature to begin with. I don't think the concept of `full` and `partial` should be baked into maintain. That seems to be a high-level concept that the CLI will have, but the lower-level interface doesn't need this. I would simplify this to the following signature:
> ```python
> def maintain(self, pack_loose = True, repack = False, clean_storage = False, vacuum = False) -> dict:
> ```
> i.e. with defaults that correspond to your `partial` and then the CLI can call this with all `True` if the user requests a `full` maintenance.

Normally I agree, but here the idea to call `repository_maintain` with `full=True` or `full=False` is that you don't need to know the specifics of the backend, only the implemented backend knows what `full` means to them. The `repository_maintain` doesn't know what the underlying backend is and thus what its specific switches are: the `full` keyword is the only option we require the repository backend to implement, the rest are optional switches that would be passed from the user to the backend transparently (i.e. `repository_maintain` just passes them, doesn't know what they do), with the understanding that is the user the one who know what specific backend is using and thus the keywords that are available.

P.S. I have not yet applied the modification suggested above in case you are in the middle of reviewing so that it doesn't change stuff. If you haven't started I add it now and push it.

ramirezfranciscof: I'll mark this as resolved just because I modified this as you suggested in the first comment and now pylint doesn't complain anymore (thanks!). If you still disagree about the general design feel free to keep replying though.

chrisjsewell (changes requested):
Please re-write without the use of kwargs

ramirezfranciscof:
> Please re-write without the use of kwargs

Hey, thanks for the review. Would you please care to elaborate a bit more about this? Because I can remove the superfluous `**kwargs` in methods such as `aiida/repository/backend/abstract::get_info` which were mostly added pre-emptively for future usage and for homogeneity with the other methods, sure. But I can't remove all of the usages of kwargs here, most notably:

1. I can't take them out of `aiida/repository/control::maintain` operation because, as discussed before, the whole idea is that it would be passing them transparently from what the user provided to the backend which knows the specifics requested, and can be different from backend to backend.

2. I guess theoretically I could make the `kwargs` from the `DiskObjectStoreRepositoryBackend.maintain` method explicit because there it already knows what are the details of that backend. But when I add them there, then I get a complaint from  `pylint` because the number of arguments does not coincide with the method from the parent class:

```
aiida/repository/backend/disk_object_store.py:122:4: W0221: Number of parameters was 2 in 'AbstractRepositoryBackend.maintain' and is now 6 in overridden 'DiskObjectStoreRepositoryBackend.maintain' method (arguments-differ)
```

(making the method abstract doesn't seem to help btw)

aiida/repository/control.py
```diff
@@ -0,0 +1,81 @@
...
+    This should be all the soft-deleted files.
+    """
+    from aiida import orm
+
+    if aiida_backend is None:
+        aiida_backend = get_manager().get_backend()
+
+    repository_backend = aiida_backend.get_repository()
+
+    keyset_backend = set(repository_backend.list_objects())
```
sphuber, line 60: Think it is a bit confusing to name this variable `_backend` when it is the repository. In current terminology, the backend refers at most to just the PostgreSQL db, but really to DB + repo.

aiida/repository/control.py
```diff
@@ -0,0 +1,81 @@
...
+    keyset_aiidadb = set(orm.Node.objects(aiida_backend).iter_repo_keys())
+
+    if check_consistency:
+        if len(keyset_aiidadb - keyset_backend) > 0:
+            raise RuntimeError('Database seems corrupted (some tracked objects are not in the repository). Aborting!')
```
sphuber, line 65: Should this be phrased differently? We don't know if it is the database that is corrupt or the repository. Should probably say something like "Backend is in inconsistent state: database contains dangling file objects."

aiida/repository/control.py
```diff
@@ -0,0 +1,81 @@
...
+def get_repository_report(backend: Backend = None, **kwargs) -> dict:
+    """Performs a report on the status of the repository."""
+    report = {'user_info': ''}
+
+    if backend is None:
+        backend = get_manager().get_backend()
+    unreferenced_objects = get_unreferenced_keyset(aiida_backend=backend)
+
+    num_unref = len(unreferenced_objects)
+    report['user_info'] += f' > Unreferenced files to delete: {num_unref}'
```
sphuber, line 79: Think in these kinds of report methods that it is better to keep the formatting to a minimum. That should be the task of the caller who is going to print this to screen somehow.
```suggestion
    report['user_info'] += f'Unreferenced files to delete: {num_unref}'
```

aiida/repository/control.py
```diff
@@ -0,0 +1,81 @@
...
+    keyset_aiidadb = set(orm.Node.objects(aiida_backend).iter_repo_keys())
+
+    if check_consistency:
+        if len(keyset_aiidadb - keyset_backend) > 0:
+            raise RuntimeError('Database seems corrupted (some tracked objects are not in the repository). Aborting!')
+
+    return keyset_backend - keyset_aiidadb
+
+
+def get_repository_report(backend: Backend = None, **kwargs) -> dict:
```
sphuber, line 70: Is this function really necessary? You are just calling this in the CLI command and are essentially just calling one other function `get_unreferenced_keyset`. The formatting you do would be much better placed in the CLI command than in this function.

aiida/repository/control.py
```diff
@@ -0,0 +1,81 @@
...
+
+    repository_backend.delete_objects(list(unreferenced_objects))
+
+    # Perform the maintainance operations in the repository
+    maintainance_report['repository']['maintain'] = repository_backend.maintain(full=full, **kwargs)  # type: ignore
+
+    return maintainance_report
+
+
+def repository_info(statistics: bool = False, backend: Backend = None, **kwargs) -> dict:
```
sphuber, line 40: Same here. Does this really merit a function? or can the CLI command just call `repository.get_info(statistics)`? That is a single call. The only thing the command should do is retrieve the `repository` but this seems like a common operation and so would be fine to have a function for.

sphuber (changes requested):
Thanks @ramirezfranciscof . I see your point of having `maintain` be a backend specific method and therefore having to have a flexible signature. I am not sure what @chrisjsewell meant with his comment to "do it without kwargs", but I think he simply means to make the kwargs that the Disk objectstore implementation accepts, explicit in the signature. `pylint` will then complain that the signature differs, but that is justified in this case, so we can put an ignore statement there. Other than that, for me there are just some minor things remaining that shouldn't be too controversial I think.

aiida/cmdline/commands/cmd_storage.py
```diff
@@ -85,41 +85,100 @@ def storage_integrity():
...
+            '\nThis command will perform all maintenance operations on the internal storage that can be safely '
+            'executed while still running AiiDA. '
+            'However, not all operations that are required to fully optimize disk usage and future performance '
+            'can be done in this way. '
+            'Whenever you find the time or opportunity, please consider running `verdi repository maintenance '
+            '--full` for a more complete optimization.\n'
+        )
+
+    if not click.confirm('Are you sure you want continue in this mode?'):
+        return
```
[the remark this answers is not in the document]
ramirezfranciscof: Since the difference between maintenance types is just the impossibility to use the profile, the partial maintenance can still be a complex lengthy task that the user might not want to start by error because he forgot to add a keyword. Also if at some point we add some "verbosity" option that outputs a lot of information while processing, it gives time for the user to actually read the message before it gets buried. I don't know, I figured it wouldn't hurt since it is not a command you are calling frequently anyways and when you do call it you want to make sure you are using the right one. If you are still against it I can take it out, I don't think it is critical.

aiida/repository/backend/abstract.py
```diff
@@ -119,6 +119,23 @@ def list_objects(self) -> Iterable[str]:
         :return: An iterable for all the available object keys.
         """
 
+    def get_info(self, statistics=False, **kwargs) -> dict:  # pylint: disable=unused-argument,no-self-use
```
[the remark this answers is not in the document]
ramirezfranciscof: Well, that is a good question.

I have not yet found a fully satisfactory criteria for when a method should be fully abstract or implement a default behavior. Right now I am more inclined to implement a default whenever possible, since it simplifies testing and prototyping. And technically, the maintain operation doesn't need to do anything (for example, if we implemented here our previous file repository, there is no maintenance to be done). This is opposed as something like the `put`/`get` methods, which actually have requirements that can't be given a default implementation.

I'm open to hearing different criterias though.

aiida/repository/control.py
ramirezfranciscof: Currently this is true, but I will add more features related to the repository in an upcoming release (I didn't want to overcrowd this one). From the design document (I highlighted what I would be missing):

> Verdi storage maintain --dry-run
> - Get some heuristics on whether to vacuum PSQL and/or SQLite -> future, open issue
> - Get a count of objects **(and size in MB)** of unreferenced objects (to know you can delete)

aiida/repository/control.py
ramirezfranciscof: Currently this is true, but I will add more features related to the repository in an upcoming release (I didn't want to overcrowd this one). From the design document (I highlighted what I would be missing):

> Verdi storage info --statistics: more details
> - What is currently in db summary -v info
> - More verbose statistics per type (calcs, data)
> - Count # objects (packed/loose)
> - **Get size on disk, separated by type (packed/loose/recoverable/PSQL DB)**

aiida/repository/control.py
ramirezfranciscof: Well, there is an issue here again related to the inherent versatility of the repository. How can a generic caller like the CLI know how to format the info from the repository if this can be vastly different things? Having the underlying method return a pre-formated text that then the CLI can decide to print (or the ORM can ignore) was the simplest solution, as any other option would seem to require the design of some sort of "communication protocol" that is expected from the repository backend.

PS: Again, here you also need to take into account that you are looking at the line added by the `unreferenced_objects` count that we do on our end, but this `user_info` would also be passed to the repository backend to be populated with further info (like the space in MB these items occupy).

aiida/repository/control.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Module for overall repository control commands."""
```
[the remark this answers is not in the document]
ramirezfranciscof: Ok, I can move it there. However now I am not sure what is the hierarchy of concepts here, and (combined with the other comment on the `repository_backend` jargon) what exactly is what we call the backend. The way I understood it before, I thought the idea was that it would be something like this:

 - Storage Backend
   - Database Backend
   - Repository Backend

But now from these comments I understand it is more like:

 - Storage
   - Database Backend ("**the** backend")
     - Repository

Is this so? Is there any reason why the database and repository can't be "independent" from one another?

aiida/cmdline/commands/cmd_storage.py
```diff
@@ -85,41 +85,100 @@ def storage_integrity():
...
+    repodata = repository_info(statistics=statistics)
+
+    echo.echo_dictionary({'database': dbdata, 'repository': repodata}, sort_keys=False, fmt='yaml')
+
+
+@verdi_storage.command('maintain')
+@click.option(
+    '--full',
+    is_flag=True,
+    help='Performs maintenance tasks that are safe to run while other AiiDA instances are working with the profile.'
```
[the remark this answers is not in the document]
ramirezfranciscof: Fixed

aiida/cmdline/commands/cmd_storage.py
```diff
@@ -85,41 +85,100 @@ def storage_integrity():
...
+)
+@click.option(
+    '--dry-run',
+    is_flag=True,
+    help='Performs maintenance tasks that are safe to run while other AiiDA instances are working with the profile.'
```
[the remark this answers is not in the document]
ramirezfranciscof: Fixed

aiida/cmdline/commands/cmd_storage.py
```diff
@@ -85,41 +85,100 @@ def storage_integrity():
...
+    """Performs maintenance tasks on the repository."""
+    from aiida.repository.control import get_repository_report, repository_maintain
+
+    if dry_run and full:
+        echo.echo_critical('You cannot request both `--dry-run` and `--full` at the same time.')
+
+    if dry_run:
+        maintainance_report = get_repository_report()['user_info']
+        click.echo('Repository:')
+        click.echo(maintainance_report)
```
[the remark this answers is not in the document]
ramirezfranciscof: Fixed

aiida/cmdline/commands/cmd_storage.py
```diff
@@ -85,41 +85,100 @@ def storage_integrity():
...
+        return
+
+    if full:
+
+        click.echo(
```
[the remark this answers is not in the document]
ramirezfranciscof: Fixed

aiida/cmdline/commands/cmd_storage.py
```diff
@@ -85,41 +85,100 @@ def storage_integrity():
...
+        )
+
+    else:
+
+        click.echo(
+            '\nThis command will perform all maintenance operations on the internal storage that can be safely '
+            'executed while still running AiiDA. '
+            'However, not all operations that are required to fully optimize disk usage and future performance '
+            'can be done in this way. '
+            'Whenever you find the time or opportunity, please consider running `verdi repository maintenance '
```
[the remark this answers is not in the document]
ramirezfranciscof: If I use `echo.echo_info` it seems it is not shown by default (or at least not in the testings). Is this so? If it is, maybe this is related to the discussion / other comment below of how important is for the user to see this and confirm.

aiida/cmdline/commands/cmd_storage.py
```diff
@@ -85,41 +85,100 @@ def storage_integrity():
...
+            '--full` for a more complete optimization.\n'
+        )
+
+    if not click.confirm('Are you sure you want continue in this mode?'):
+        return
 
-    echo.echo_dictionary(data, sort_keys=False, fmt='yaml')
+    maintainance_report = repository_maintain(full=full)
+    click.echo('\nMaintainance procedures finished:\n')
+    if 'user_info' in maintainance_report:
```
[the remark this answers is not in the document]
ramirezfranciscof: Fixed

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +118,53 @@ def get_object_hash(self, key: str) -> str:
         if self.container.hash_type != 'sha256':
             return super().get_object_hash(key)
         return key
+
+    def maintain(self, full: bool = False, **kwargs) -> dict:
```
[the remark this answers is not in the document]
ramirezfranciscof: Yes, I agree it would be a good opportunity to start applying these. Fixed!

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +118,53 @@ def get_object_hash(self, key: str) -> str:
...
+
+        operation_results = {}
+
+        if pack_loose:
+            self.container.pack_all_loose()
+            operation_results['pack_loose'] = {'decription': 'Packed all loose files.'}
+
+        if do_repack:
+            self.container.repack()
+            operation_results['repack'] = {'decription': 'Repacked all packages for optimal access.'}
```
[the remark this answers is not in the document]
ramirezfranciscof: Fixed

aiida/repository/control.py
ramirezfranciscof: Fixed (but see other comment referencing this)

ramirezfranciscof:
@sphuber thanks for the review! I think I have addressed all comments (either fixing or responding) so let me know what do you think.

Currently the pre-commit is failing because mipy doesn't like the overriding of the methods:

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

aiida/repository/backend/disk_object_store.py:122: error: Signature of "maintain" incompatible with supertype "AbstractRepositoryBackend"  [override]
aiida/repository/backend/disk_object_store.py:180: error: Signature of "get_info" incompatible with supertype "AbstractRepositoryBackend"  [override]
Found 2 errors in 1 file (checked 104 source files)
```

However, when I try to change this locally, apparently MY mypy doesn't like the ignores (this is my line 180 for example):

```python
    def get_info(  # type: ignore[override] # pylint: disable=arguments-differ
        self,
        statistics=False,
        ) -> dict:
```

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

aiida/repository/backend/disk_object_store.py:122: error: unused "type: ignore" comment
aiida/repository/backend/disk_object_store.py:180: error: unused "type: ignore" comment
Found 2 errors in 1 file (checked 1 source file)

pylint...................................................................Passed
```

I have `mypy==0.910`, which is the same of the automated testing, so I'm not sure what is happening. I guess I could just force commit without running the pre-commit and see if it fixes it here, but that seems less than ideal.

aiida/cmdline/commands/cmd_storage.py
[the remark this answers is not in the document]
sphuber: By using `echo.echo_info` the `--verbosity` option can control what messages are shown. The default logging level is `INFO` and so indeed messages sent with `echo_info` won't show up. If you want it to show up by default, but not be a warning, you can use `echo_report`, which works just as in workchains.

aiida/cmdline/commands/cmd_storage.py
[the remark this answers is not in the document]
sphuber: > I don't know, I figured it wouldn't hurt since it is not a command you are calling frequently anyways and when you do call it you want to make sure you are using the right one.

I was under the impression that the `verdi storage maintain` should be expected to be run often, exactly because it can be done while running AiiDA. So if that is true, it would be a bit annoying if there is a prompt everytime, especially if there is no force option. But also for me it is not critical. If someone else has an opinion either way, that would be fine for me.

aiida/repository/backend/abstract.py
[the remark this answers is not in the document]
sphuber: > since it simplifies testing and prototyping.

That is only an argument in very rare cases, and here I definitely don't think it applies. Changing design or implementation just because it makes testing easier since you do not have to mock the methods for mock implementations, is not really a good approach I would say. Since maintain is expected to do something, I actually think this should be abstract, or at the very least the base class raises `NotImplementedError`. This prevents that you forget to implement it and an implementation actually doesn't do anything by mistake.

aiida/repository/control.py
sphuber: Historically, the database was called the "backend", but this was really a misnomer. It was actually just *a* backend to our database. We then started using "backend" to mean the backend interface that stores all of AiiDA's data (for a given profile). We then also realized, rightfully so, that the repository also belongs to this. The database and repository are both part of the "backend" and they are even no longer fully independent.

So I think we should use the following terminology:

AiiDA uses a data store (or storage). The current backend for this consists of a database (Postgres) and a file repository (disk object store).

In the implementation we refer to it just as the "Backend" (as of @chrisjsewell 's recent changes). I think this is still a bit of a shortcut, because a backend is too general, but for internal purposes, I think it is fine.

aiida/repository/control.py
sphuber: I see the problem, but I still don't think that having these methods that spit out preformatted strings that are then just to be printed as a flexible solution. I think it is better to have a conditinal in the CLI to check the repo type and have it retrieve the information it wants to display and format it however it wants to, so that it is consistent. But if you really think this method is necessary, then I think the formatting should be as minimal as possible. This ` > ` formatting seems really unnecessary.

aiida/repository/control.py
sphuber: Also here, I would just have the repo interface expose methods to retrieve this data, and then have the CLI call this. I realize that this means the CLI has to check the type of the repo, but I think that is its purpose. That logic _should_ go there.

aiida/repository/control.py
sphuber: Same as above. Also, it is not at all clear what the difference should be between `get_repository_report` and `repository_info`. If they really need to stay, why not merge them? Or at the very least make them consistent and rename latter to `get_repository_info`.

aiida/backends/control.py
```diff
@@ -0,0 +1,84 @@
+# -*- coding: utf-8 -*-
```
chrisjsewell, line 1: Add this file to `.pre-commit-config.yaml` mypy files

aiida/backends/control.py
```diff
@@ -0,0 +1,84 @@
...
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Module for overall repository control commands."""
+# Note: these functions are not methods of `AbstractRepositoryBackend` because they need access to the orm.
+# This is because they have to go through all the nodes to gather the list of keys that AiiDA is keeping
+# track of (since they are descentralized in each node entry).
+# See the get_unreferenced_keyset function
+from aiida.manage.manager import get_manager
```
chrisjsewell, line 15:
```suggestion
from typing import Optional, Set

from aiida.manage.manager import get_manager
```

aiida/backends/control.py
```diff
@@ -0,0 +1,84 @@
...
+from aiida.orm.implementation import Backend
+
+
+def repository_maintain(full: bool = False, backend: Backend = None, **kwargs) -> dict:
```
chrisjsewell, line 19:
```suggestion
def repository_maintain(full: bool = False, backend: Optional[Backend] = None, **kwargs) -> dict:
```

aiida/backends/control.py
```diff
@@ -0,0 +1,84 @@
...
+
+    repository.delete_objects(list(unreferenced_objects))
+
+    # Perform the maintainance operations in the repository
+    maintainance_report['repository']['maintain'] = repository.maintain(full=full, **kwargs)  # type: ignore
+
+    return maintainance_report
+
+
+def repository_info(statistics: bool = False, backend: Backend = None, **kwargs) -> dict:
```
chrisjsewell, line 40:
```suggestion
def repository_info(statistics: bool = False, backend: Optional[Backend] = None) -> dict:
```

aiida/backends/control.py
```diff
@@ -0,0 +1,84 @@
...
+    """Returns relevant information on the repository."""
+    if backend is None:
+        backend = get_manager().get_backend()
+    repository = backend.get_repository()
+    return repository.get_info(statistics)
+
+
+def get_unreferenced_keyset(check_consistency: bool = True, aiida_backend: Backend = None, **kwargs) -> set:
```
chrisjsewell, line 48:
```suggestion
def get_unreferenced_keyset(check_consistency: bool = True, aiida_backend: Optional[Backend] = None) -> Set[str]:
```

aiida/backends/control.py
```diff
@@ -0,0 +1,84 @@
...
+        keyset_missing = keyset_aiidadb - keyset_backend
+        if len(keyset_missing) > 0:
+            raise RuntimeError(
+                'There are objects referenced in the database that are not present in the repository. Aborting!'
+            )
+
+    return keyset_backend - keyset_aiidadb
+
+
+def get_repository_report(backend: Backend = None, **kwargs) -> dict:
```
chrisjsewell, line 73:
```suggestion
def get_repository_report(backend: Optional[Backend] = None) -> dict:
```

aiida/cmdline/commands/cmd_storage.py
```diff
@@ -84,42 +84,100 @@ def storage_integrity():
 @click.option('--statistics', is_flag=True, help='Provides more in-detail statistically relevant data.')
 def storage_info(statistics):
```
chrisjsewell, line 85: This now needs to use/adapt `from aiida.cmdline.utils.common import get_database_summary`

tests/backends/test_control.py
```diff
@@ -0,0 +1,148 @@
...
+        return text
+
+
+@pytest.fixture(scope='function')
+@pytest.mark.usefixtures('clear_database_before_test')
+def clear_storage_before_test():
+    """Clears the storage before a test."""
+    #-----------------------------------------------------------------#
+    # For some reason `clear_database_before_test` is not cleaning out
+    # the nodes so I have to do it manually.
```
chrisjsewell, line 42: Have you tried just this

```suggestion
def clear_storage_before_test(clear_database_before_test):
    """Clears the storage before a test."""
```

tests/backends/test_control.py
```diff
@@ -0,0 +1,148 @@
...
+    #-----------------------------------------------------------------#
+
+    repository = get_manager().get_backend().get_repository()
+    object_keys = list(repository.list_objects())
+    repository.delete_objects(object_keys)
+    repository.maintain(full=True)
+
+
+@pytest.mark.usefixtures('clear_database_before_test', 'clear_storage_before_test')
+def test_get_unreferenced_keyset():
```
chrisjsewell, line 59: you are already using `clear_database_before_test` via `clear_storage_before_test`

```suggestion
@pytest.mark.usefixtures('clear_storage_before_test')
def test_get_unreferenced_keyset():
```

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +118,90 @@ def get_object_hash(self, key: str) -> str:
...
+            clean_storage = override_clean_storage
+
+        if override_do_vacuum is not None:
+            do_vacuum = override_do_vacuum
+
+        operation_results = {}
+
+        if pack_loose:
+            self.container.pack_all_loose()
+            operation_results['pack_loose'] = {'decription': 'Packed all loose files.'}
```
chrisjsewell, line 166: Maybe it better to have the as logging reports, as you go, rather than only reporting when everything is finished?

```suggestion
            MAINTAIN_LOGGER.report('Packing all loose files.')
            self.container.pack_all_loose()
```

chrisjsewell (dismissed):

aiida/repository/backend/abstract.py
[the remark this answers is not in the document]
ramirezfranciscof: > Changing design or implementation just because it makes testing easier since you do not have to mock the methods for mock implementations, is not really a good approach I would say.

Well, I agree, but this is not what I was saying:

> I have not yet found a fully satisfactory criteria for when a method should be fully abstract or implement a default behavior.

I am not using the argument to change an otherwise established convenient design. I am saying these two options seem otherwise equivalent to me in all other practical aspects, so that was the only distinction I found to decide on one vs the other.

If it was clear to me that the `maintain` method had anything that would be expected of it, I would have indeed chosen to do this an abstract method. But although it may be true that you typically call it because you expect it to do "something" (although this is also true of a lot of functions who can have "empty" results), since that can be extremely varied and different from backend to backend, I think the most robust thing is that clients of the feature operate under the assumption that it could actually do nothing (in which case setting that as the default is much more natural).

Anyways, having it raise a `NotImplementedError` is ok for me. It would mean clients have to add more safeguards when using it, making their code a bit clunkier, but they also get a more explicit treatment of the doing nothing case. I can't really say a priori if this will be a practically convenient tradeoff but we can try it out.

aiida/repository/control.py
ramirezfranciscof: > I think it is better to have a conditinal in the CLI to check the repo type and have it retrieve the information it wants to display and format it however it wants to, so that it is consistent.

But this basically breaks the modularity of the abstraction, no? The CLI needs to know about what information gets provided by all possible repository backends.

> But if you really think this method is necessary, then I think the formatting should be as minimal as possible. This `>` formatting seems really unnecessary.

Yeah, ok, that is a fair point, I can remove those.

aiida/repository/control.py
ramirezfranciscof: > AiiDA uses a data store (or storage). The current backend for this consists of a database (Postgres) and a file repository (disk object store).

Should we maybe then reflect this in the code structure? i.e.: re-arrange the stuff from `aiida.backend` and `aiida.repository` into `aiida.storage.database` and `aiida.storage.repository`? (or even `aiida.backend.database` and `aiida.backend.repository`)

tests/backends/test_control.py
ramirezfranciscof: No I didn't try that before. Thanks, this seemed to be the problem, I was not aware that you couldn't use marks on fixtures.

aiida/repository/control.py
ramirezfranciscof: Mmm, I'm not sure I agree. The  `get_repository_report` method returns information that would help you decide if you need to run maintenance or not (for example, number of unreferenced files that were soft deleted and you could hard delete). This is displayed by the CLI for human users but could be part of a different UI (maybe even for an automated maintenance cron workflow) that would need to do the exact same "backend customization" logic. This is why I separated that from the click function.

sphuber: That's my point though. If you intend this information to be used by other consumers, like scripts, you wouldn't want this information returned in human-readable format, where the important information (the number of unreferenced files) is embedded in a string. The interface _should_ have an easy method to retrieve this data, but it should not be bothered how it is presented to a user; that is the CLI's job.

ramirezfranciscof:
@chrisjsewell @sphuber I don't understand why exactly this is happening or how to prevent it, but the auto importer is overwritting the `aiida/backends/__init__.py` file and wreaking havoc on everything. This was also happening with the local hooks, but tried skipping it but the autofixer here is still insisting in doing this.

chrisjsewell:
Yeh you need to put any code in `__init__.py` under a `# END AUTO-GENERATED` comment, otherwise it will be overwritten

ramirezfranciscof:
**Problem 1:** mypy pre-commit in github is complaining about the "incompatible" signatures of the `maintain` and `get_info` methods that are overwritten with the extra explicit parameters.

``` bash
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

aiida/repository/backend/disk_object_store.py:122: error: Signature of "maintain" incompatible with supertype "AbstractRepositoryBackend"  [override]
aiida/repository/backend/disk_object_store.py:184: error: Signature of "get_info" incompatible with supertype "AbstractRepositoryBackend"  [override]
Found 2 errors in 1 file (checked 108 source files)
```

I don't get this locally, and my pre-commits complain if I tell them to ignore these:

``` python
    def maintain(  # type: ignore # pylint: disable=arguments-differ
   (...)
    def get_info(  # type: ignore # pylint: disable=arguments-differ
```

``` bash
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

aiida/repository/backend/disk_object_store.py:122: error: unused "type: ignore" comment
aiida/repository/backend/disk_object_store.py:184: error: unused "type: ignore" comment
Found 2 errors in 1 file (checked 1 source file)
```

ramirezfranciscof:
**Problem 2:** pylint pre-commit in github is complaining about the the `maintain` and `get_info` methods being abstract but not implemented in `SandboxRepositoryBackend` and `ZipfileBackendRepository`. This is not so however, those methods are not abstract anymore (nor for [maintain](https://github.com/aiidateam/aiida-core/pull/4965/files#diff-885f125a2d837715bfc0462b3e116bc62eaec359e6f91a75fc1b60baa4e01a91R128-R129) or [get_info](https://github.com/aiidateam/aiida-core/pull/4965/files#diff-885f125a2d837715bfc0462b3e116bc62eaec359e6f91a75fc1b60baa4e01a91R121-R122)).

ramirezfranciscof:
**Problem 3:** The verdi tests are detecting a "potential speed problem". From what I understand this is because it is detecting that the orm is getting imported somewhere it shouldn't in the cmdline, but (1) I have not added any global import it `cmd_storage`, (2) I have not modified any other cmdline here, (3) I haven't modified anything in the modules that are globally imported in `cmd_storage` and (4) I haven't added a global `import orm` or `from orm` anywhere. So I am a bit at a loss here, I'm not sure what else this error could mean.

``` bash
  warnings.warn(f'Creating AiiDA configuration folder `{path}`.')
Critical: potential `verdi` speed problem: `aiida.orm` module is imported which is not in: ('aiida.backends', 'aiida.cmdline', 'aiida.common', 'aiida.manage', 'aiida.plugins', 'aiida.restapi')
Error: Process completed with exit code 1.
```

ramirezfranciscof:
**Outstanding Discussion:** (writing...)

ramirezfranciscof:
Applied the changes discussed earlier today @sphuber @chrisjsewell so this is ready for review. Of the previously mentioned problems I am still having this one:

> **Problem 3:** The verdi tests are detecting a "potential speed problem". From what I understand this is because it is detecting that the orm is getting imported somewhere it shouldn't in the cmdline, but (1) I have not added any global import it `cmd_storage`, (2) I have not modified any other cmdline here, (3) I haven't modified anything in the modules that are globally imported in `cmd_storage` and (4) I haven't added a global `import orm` or `from orm` anywhere. So I am a bit at a loss here, I'm not sure what else this error could mean.
> 
> ```shell
>   warnings.warn(f'Creating AiiDA configuration folder `{path}`.')
> Critical: potential `verdi` speed problem: `aiida.orm` module is imported which is not in: ('aiida.backends', 'aiida.cmdline', 'aiida.common', 'aiida.manage', 'aiida.plugins', 'aiida.restapi')
> Error: Process completed with exit code 1.
> ```

Any ideas?

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +120,116 @@ def get_object_hash(self, key: str) -> str:
...
+        if live:
+            if override_do_repack or override_clean_storage or override_do_vacuum:
+                errmsg = 'a specifically resquest keyword cannot be applied while the profile is in use:\n'
+                if override_do_repack is not None:
+                    errmsg = ' > override_do_repack = {override_do_repack}\n'
+                if override_clean_storage is not None:
+                    errmsg = ' > override_clean_storage = {override_clean_storage}\n'
+                if override_do_vacuum is not None:
+                    errmsg = ' > override_do_vacuum = {override_do_vacuum}\n'
+                raise ValueError(errmsg)
```
sphuber, line 165:
```suggestion
                overrides = {'do_repack': do_repack, 'clean_storage': clean_storage, 'do_vacuum': do_vacuum}
                keys = ', '.join([key for key, override in overrides if override is True])
                raise ValueError('The following overrides were enabled but can't be if `live=True`: {keys}')
```

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +120,116 @@ def get_object_hash(self, key: str) -> str:
...
+        override_pack_loose: bool = None,
+        override_do_repack: bool = None,
+        override_clean_storage: bool = None,
+        override_do_vacuum: bool = None,
+    ) -> dict:
+        """Performs maintenance operations.
+
+        :param full:
+            a flag to perform operations that require to stop using the maintained profile.
+        :param override_pack_loose:
```
sphuber, line 137: Is the `override_` prefix really necessary in the keyword itself? Seems more natural to me to just have it without. The docstring of `live` should then simply explain that it will automatically set values for these and the specific flags can be used to override, noting that for `live=True` not all of them may be overridden

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +120,116 @@ def get_object_hash(self, key: str) -> str:
...
+            files_size = self.container.get_total_size()['total_size_loose'] * BYTES_TO_MB
+            if dry_run:
+                DOSTORE_LOGGER.report(f'Would pack all loose files ({files_numb} files occupying {files_size} MB) ...')
+            else:
+                DOSTORE_LOGGER.report(f'Packing all loose files ({files_numb} files occupying {files_size} MB) ...')
+                self.container.pack_all_loose()
+
+        if do_repack:
+            files_numb = self.container.count_objects()['packed']
+            files_size = self.container.get_total_size()['total_size_packfiles_on_disk'] * BYTES_TO_MB
```
sphuber, line 190: This information is not really adding anything specific to the maintenance operation is it? It just gives the current size, but that doesn't tell what it will be nor what will be saved. Only the latter would be really interesting IMO

tests/cmdline/commands/test_storage.py
```diff
@@ -103,3 +103,52 @@ def mocked_migrate(self):  # pylint: disable=no-self-use
...
+    def mock_maintain(**kwargs):
+        logmsg = 'Provided kwargs:\n'
+        for key, val in kwargs.items():
+            logmsg += f' > {key}: {val}\n'
+        logging.info(logmsg)
+
+    monkeypatch.setattr(control, 'repository_maintain', mock_maintain)
+
+    with caplog.at_level(logging.INFO):
+        _ = run_cli_command(cmd_storage.storage_maintain, user_input='Y')
```
sphuber, line 130: Might as well not assign the result if you are not going to use it
```suggestion
        run_cli_command(cmd_storage.storage_maintain, user_input='Y')
```

sphuber (changes requested):
Thanks @ramirezfranciscof . I think the interface is a lot better now. Just some comments on bits of the new implementation

aiida/backends/control.py
```diff
@@ -23,26 +23,27 @@
...
-    # Perform the maintainance operations in the repository
-    MAINTAIN_LOGGER.info('Performing repository-specific maintenance ...')
-    try:
-        maintainance_report['repository']['maintain'] = repository.maintain(full=full, **kwargs)
-    except NotImplementedError:
-        maintainance_report['repository']['maintain'] = {}
+    if dry_run:
+        MAINTAIN_LOGGER.info(f'Would delete {len(unreferenced_objects)} unreferenced objects ...')
+    else:
+        MAINTAIN_LOGGER.info(f'Deleting {len(unreferenced_objects)} unreferenced objects ...')
```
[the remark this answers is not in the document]
ramirezfranciscof: Ah, sorry, I misunderstood and thought you explicitly said to do it like this. Ok, I can change it.

aiida/backends/control.py
```diff
@@ -0,0 +1,80 @@
...
+# This is because they have to go through all the nodes to gather the list of keys that AiiDA is keeping
+# track of (since they are descentralized in each node entry).
+# See the get_unreferenced_keyset function
+from typing import Optional, Set
+
+from aiida.common.log import AIIDA_LOGGER
+from aiida.manage.manager import get_manager
+from aiida.orm.implementation import Backend
+
+__all__ = ('MAINTAIN_LOGGER',)
```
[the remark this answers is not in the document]
ramirezfranciscof: Well, the `disk_object_store` uses it [here](https://github.com/aiidateam/aiida-core/pull/4965/files#diff-691a383302df0c55699d6ed0ef3b004ac423ed2fc44179296229d6ee7ecb9097R148-R149) and all backends would use it too, no?

```python
        from aiida.backends.control import MAINTAIN_LOGGER
        DOSTORE_LOGGER = MAINTAIN_LOGGER.getChild('disk_object_store')  # pylint: disable=invalid-name
```

aiida/backends/control.py
```diff
@@ -0,0 +1,80 @@
...
+    from aiida import orm
+    MAINTAIN_LOGGER.info('Obtaining unreferenced object keys ...')
+
+    if aiida_backend is None:
+        aiida_backend = get_manager().get_backend()
+
+    repository = aiida_backend.get_repository()
+
+    keyset_backend = set(repository.list_objects())
+    keyset_aiidadb = set(orm.Node.objects(aiida_backend).iter_repo_keys())
```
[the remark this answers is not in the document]
ramirezfranciscof: No nothing, I think this just pre-dated that discussions. I changed it now.

aiida/repository/backend/abstract.py
```diff
@@ -119,6 +119,31 @@ def list_objects(self) -> Iterable[str]:
...
+
+        :return: a dictionary with the information.
+        """
+
+    @abc.abstractmethod
+    def maintain(self, dry_run: bool = False, live: bool = True, **kwargs) -> None:
+        """Performs maintenance operations.
+
+        :param dry_run:
+            flag to only run
```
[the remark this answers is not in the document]
ramirezfranciscof: Well, I think this is clearly trying to express that

aiida/repository/backend/disk_object_store.py
ramirezfranciscof: Well, it can help give you an idea of how long it might take to do the repacking. But ok, I can take it out if you prefer.

aiida/backends/control.py
```diff
@@ -0,0 +1,80 @@
...
+# See the get_unreferenced_keyset function
+from typing import Optional, Set
+
+from aiida.common.log import AIIDA_LOGGER
+from aiida.manage.manager import get_manager
+from aiida.orm.implementation import Backend
+
+__all__ = ('MAINTAIN_LOGGER',)
+
+MAINTAIN_LOGGER = AIIDA_LOGGER.getChild('maintain')
```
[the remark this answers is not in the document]
ramirezfranciscof: Mmm but then the question is how do I manage the inheritance of loggers? Because here I just inherit from the `AIIDA_LOGGER` so I can just do:

```python
MAINTAIN_LOGGER = AIIDA_LOGGER.getChild('backend.control')
```

But then any other logger that inherits from this one will have `aiida.backend.control` prefix even though these are in `aiida.repository.backend.xxxxxx`...

ramirezfranciscof:
Hey @sphuber I think we might have counted differently but this is ready for other review.

BTW related to the error, the only place where I added something related to the orm is [here]():

```python
from aiida.orm.implementation import Backend

```

aiida/backends/control.py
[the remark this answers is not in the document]
sphuber: Ok, let's leave this for now.

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +120,96 @@ def get_object_hash(self, key: str) -> str:
...
+        live: bool = True,
+        pack_loose: bool = None,
+        do_repack: bool = None,
+        clean_storage: bool = None,
+        do_vacuum: bool = None,
+    ) -> dict:
+        """Performs maintenance operations.
+
+        :param full:
+            flag to perform operations that require to stop using the maintained profile.
```
sphuber, line 136:
```suggestion
        :param live:
            if True, will only perform operations that are safe while the repository is (potentially) in use.
```

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +120,96 @@ def get_object_hash(self, key: str) -> str:
...
+        :param override_pack_loose:
+            override flag for forcing the packing of loose files.
```
sphuber, line 138: You forgot to update the docstring after the name change

aiida/repository/backend/disk_object_store.py
```diff
@@ -118,3 +120,96 @@ def get_object_hash(self, key: str) -> str:
...
+        """
+        from aiida.backends.control import MAINTAIN_LOGGER
+        DOSTORE_LOGGER = MAINTAIN_LOGGER.getChild('disk_object_store')  # pylint: disable=invalid-name
+
+        if live and (do_repack or clean_storage or do_vacuum):
+            overrides = {'do_repack': do_repack, 'clean_storage': clean_storage, 'do_vacuum': do_vacuum}
+            keys = ', '.join([key for key, override in overrides if override is True])  # type: ignore
+            raise ValueError(f'The following overrides were enabled but cannot be if `live=True`: {keys}')
+
+        pack_loose = True if pack_loose is None else pack_loose
```
sphuber, line 156: For this type of conditional there is this standard idiom
```suggestion
        pack_loose = pack_loose or True
```

sphuber (changes requested):
Thanks @ramirezfranciscof . Looks good to be merged now; just spotted some docstrings you forgot to update after last changes. If you correct those, I will approve and merge.

aiida/repository/backend/disk_object_store.py
ramirezfranciscof: Mmm but if the user passes `False` I want it to respect that and what you propose would overwrite it no?

sphuber: Nope, this construction only hits the `or` if the variable is `None`, not just something falsy

ramirezfranciscof: Mmmm are you sure? I mean, I don't mean to be skeptic of you but I just tried it and it giving me the behavior I am describing:

```python
In [1]: false_var = False

In [2]: true_var = True

In [3]: none_var = None

In [4]: destination = none_var or 'GOOD'

In [5]: destination
Out[5]: 'GOOD'

In [6]: destination = true_var or 'BAD'

In [7]: destination
Out[7]: True

In [8]: destination = false_var or 'BAD'

In [9]: destination
Out[9]: 'BAD'

In [10]: destination = false_var or True  # to check the exact case in case strings have a different behaviour

In [11]: destination
Out[11]: True
```

Would it make a difference if the variables are arguments of the encompassing function? I can't imagine it would...

sphuber: Yep, you are absolutely right. So forget about this comment

ramirezfranciscof:
All tests seem to be passing. I am now having again the problem I describe [here](https://github.com/aiidateam/aiida-core/pull/4965#issuecomment-987794222), but I just commited ignoring the hooks. Still would like to know why this is happening but I want this merged more.

@sphuber all good now? (notice that there is an [outstanding comment](https://github.com/aiidateam/aiida-core/pull/4965#discussion_r765215418) in your previous review where I was waiting for confirmation if my comment convinced you or you still want the info taken out)

sphuber (approved):
Thanks @ramirezfranciscof . The ignore statement for `mypy` because of the different signatures is not a big deal, it is the same as that of `pylint`. Let's keep that for now and merge this.

PR #4966, opened 2021-06-01
Additional info printing

louisponet:
Following my show & tell and the small discussion we had after, I thought I'd make a quick example case of the additional information that could be printed to help users find their way.

The main goals would be to:
1. immediately show some helpful information
2. guide users to where they might find additional information/what they are looking for
3. Keep things uniform, i.e. ordering of certain things etc (not a huge deal)

As shown by the example, very little needs to be done to supply already a lot of information that's useful, I think this would be a very easy way to increase interactivity of the basic AiiDA use.

Would be great if we can start adding these small QOL improvements asap, and to also have some feedback.
@giovannipizzi @ramirezfranciscof  @mbercx @chrisjsewell @ltalirz  

PS: There's the crayons package that would also be nice to use, i.e. make all the PK numbers green, node types yellow, if errors are present color them red, or something along those lines.

Cheers!

### DIFF
--- a/aiida/orm/nodes/node.py
+++ b/aiida/orm/nodes/node.py
@@ -153,7 +153,7 @@ class Node(Entity, NodeRepositoryMixin, EntityAttributesMixin, EntityExtrasMixin
         if not self.is_stored:
             return f'uuid: {self.uuid} (unstored)'
 
-        return f'uuid: {self.uuid} (pk: {self.pk})'
+        return f'uuid: {self.uuid}\n\tpk: {self.pk}'
 
     def __copy__(self):
         """Copying a Node is not supported in general, but only for the Data sub class."""
--- a/aiida/orm/nodes/process/process.py
+++ b/aiida/orm/nodes/process/process.py
@@ -55,11 +55,21 @@ class ProcessNode(Sealable, Node):
     _unstorable_message = 'only Data, WorkflowNode, CalculationNode or their subclasses can be stored'
 
     def __str__(self) -> str:
-        base = super().__str__()
-        if self.process_type:
-            return f'{base} ({self.process_type})'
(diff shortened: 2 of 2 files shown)


chrisjsewell:
Thanks @louisponet, although I would note that all nodes also have a `get_description` method. We should decide what the balance is between this method and `__str__`. There will be instances, for example, where you may want to print a long list of nodes, for this use case you would prefer a more concise string. I would suggest `get_description` should contain the more verbose string

louisponet:
That is fair, I tend to prefer to print a lot in the "default" setting that will be seen by most users, and provide an option to minimize the output for more specific use cases. I don't know of course how often people print many nodes.

chrisjsewell:
> and provide an option to minimize the output for more specific use cases.

Unfortunately you can't do this for `__str__` though, because it does not allow for any arguments. You could though have a verbosity argument for `get_description`

louisponet:
I meant more an option in the sense of `node.get_concise_repr()` or something like that. It basically depends on what occurs most often, if it occurs most often that you'd want the concise, that should be what's shown in the repl, if more often you'd like to see the info then the longer one would make more sense to be the default.

chrisjsewell:
Another interesting option to note is with IPython you can use rich repr methods: https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

sphuber:
Indeed, I don't think we should `__str__` for this. This is usually called in places where we need a short one line description of the node, such as in CLI outputs or log messages. Having a multiline string each time is going to be messing up the readability and all the database hits are going to slow it down. I think having a dedicated method, like `get_description` makes sense.

chrisjsewell:
I think this is essentially on hold, until #3087 is "answered"

sphuber:
Since this has been on hold for a long time, I am closing this for now. Once #3087 has been discussed and answered it can be reopened or a new PR opened.

PR #4970, opened 2021-06-02
✨ NEW: _repr_pretty_ representation for `ProcessBuilder`

mbercx:
The representation of the `ProcessBuilder` (i.e. `__repr__`) is
difficult to read since it just prints the heavily nested mapping on one
line and doesn't show the contents of e.g. `Dict` nodes.

Here we add a `_repr_pretty_` method to improve the readability
when using the `ProcessBuilder` in Jupyter notebooks and IPython.


mbercx:
@sphuber this is adapted from a method I stole from the `aiida-quantumespresso-epfl` repo. :grin:
@louisponet I couldn't add you as a reviewed, but have a look and let me know what you think.

Some usage examples:

* After using the builder for the `PwCalculation` for a bit:

```python
builder = code.get_builder()
builder.structure = structure
builder.parameters = orm.Dict(dict={
  'CONTROL': {
    'calculation': 'scf',  # self-consistent field
  },
  'SYSTEM': {
    'ecutwfc': 30.,  # wave function cutoff in Ry
    'ecutrho': 240.,  # density cutoff in Ry
  }
})
builder
```

Gives

<p style="margin-bottom: 4px"><b>Process class:</b> PwCalculation <br />  <b>Inputs:</b> </p>

```
"code": Remote code 'qe-v6.7-pw' on localhost, pk: 1, uuid: 47a1c120-47af-4b9e-80c2-5a647c99817e
"metadata": {
    "options": {
        "stash": {}
    },
},
"parameters": {
    "CONTROL": {
        "calculation": "scf"
    },
    "SYSTEM": {
        "ecutwfc": 30.0,
        "ecutrho": 240.0
    }
},
"pseudos": {}
"structure": Si2 (pk: 3)
```

* `PwBaseWorkChain` obtained from `get_builder_from_protocol()` method:

```python
code = orm.load_code(1)
structure = orm.load_node(3)

builder = PwBaseWorkChain.get_builder_from_protocol(
    code=code,
    structure=structure,
)
builder
```

Gives:

<p style="margin-bottom: 4px"><b>Process class:</b> PwBaseWorkChain <br />  <b>Inputs:</b> </p>

```
"clean_workdir": True
"kpoints_distance": 0.15
"kpoints_force_parity": False
"metadata": {}
"pw": {
    "code": Remote code 'qe-v6.7-pw' on localhost, pk: 1, uuid: 47a1c120-47af-4b9e-80c2-5a647c99817e
    "metadata": {
        "options": {
            "max_wallclock_seconds": 43200
            "resources": {
                "num_machines": 1
            },
            "withmpi": True
        },
    },
    "parameters": {
        "CONTROL": {
            "calculation": "scf",
            "forc_conv_thr": 0.0001,
            "tprnfor": true,
            "tstress": true,
            "etot_conv_thr": 2e-05
        },
        "SYSTEM": {
            "nosym": false,
            "occupations": "smearing",
            "smearing": "cold",
            "degauss": 0.01,
            "ecutwfc": 30.0,
            "ecutrho": 240.0
        },
        "ELECTRONS": {
            "electron_maxstep": 80,
            "mixing_beta": 0.4,
            "conv_thr": 4e-10
        }
    },
    "pseudos": {
        "Si": uuid: 07a2d8db-ef3f-43ce-b200-bd68a7b4c6d2 (pk: 933)
    },
    "structure": Si2 (pk: 3)
},
```

mbercx:
@chrisjsewell should I add something to ignore these `mypy` warnings in this file?

```
aiida/engine/processes/builder.py:97: error: Item "Port" of "Union[Port, PortNamespace]" has no attribute "serialize"  [union-attr]
aiida/engine/processes/builder.py:97: error: Item "PortNamespace" of "Union[Port, PortNamespace]" has no attribute "serialize"  [union-attr]
```

mbercx:
Just for reference, this is what you would get for the `PwBaseWorkChain.get_builder_from_protocol()` method one _without_ this representation:

> ```{'metadata': {}, 'pw': {'metadata': {'options': {'resources': {'num_machines': 1}, 'max_wallclock_seconds': 43200, 'withmpi': True}}, 'pseudos': {'Si': <UpfData: uuid: 07a2d8db-ef3f-43ce-b200-bd68a7b4c6d2 (pk: 933)>}, 'code': <Code: Remote code 'qe-v6.7-pw' on localhost, pk: 1, uuid: 47a1c120-47af-4b9e-80c2-5a647c99817e>, 'structure': <StructureData: uuid: e5367a26-ebcd-479d-83db-73cf56ba1539 (pk: 3)>, 'parameters': <Dict: uuid: e6b64e56-2fec-4355-9b16-725914de9d58 (unstored)>}, 'clean_workdir': <Bool: uuid: 57f303c9-8a3e-4f0e-844f-910d4a437c08 (unstored) value: True>, 'kpoints_distance': <Float: uuid: 6055b4e0-5142-4163-9861-63971be7a0a6 (unstored) value: 0.15>, 'kpoints_force_parity': <Bool: uuid: d00c33b3-b27c-41de-b4c7-8fc47352e889 (unstored) value: False>}```

(quoted here to wrap the super long line)

louisponet:
So that's what it would look like in ipython and jupyter notebooks? I love it!

mbercx:
> So that's what it would look like in ipython and jupyter notebooks? I love it!

Unfortunately, _not_ in IPython. 😞  The console doesn't support HTML, notebooks do. See [here](https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display).

louisponet:
`_repr_pretty_` I guess is the way to go for that then, probably not very hard to get a similar repr there.

mbercx:
> `_repr_pretty_` I guess is the way to go for that then, probably not very hard to get a similar repr there.

👀  Indeed! Let me get right on that. 😁

mbercx:
@louisponet there you go! I dumped the `_repr_html_` version. It wasn't really adding anything so far. Now it works in both Jupyter notebooks as well as the IPython console.

mbercx:
Note: the `format_inputs` method should probably be moved elsewhere. I think we can also use it for e.g. formatting `overrides` inputs, or the inputs of a process that has already run.

chrisjsewell:
changing this to a draft while the tests are failing @mbercx. Get the tests passed then we can finalise this I guess 😄

mbercx:
Hmm, so I can't seem to fix the pre-commit because if I actually try to ignore these issues:

```python
aiida/engine/processes/builder.py:96: error: Incompatible types in assignment (expression has type "None", variable has type "Union[Port, PortNamespace]")  [assignment]
aiida/engine/processes/builder.py:98: error: Item "Port" of "Union[Port, PortNamespace]" has no attribute "serialize"  [union-attr]
aiida/engine/processes/builder.py:98: error: Item "PortNamespace" of "Union[Port, PortNamespace]" has no attribute "serialize"  [union-attr]
Found 3 errors in 1 file (checked 74 source files)
```

By adding `# type: ignore[assignment]` and `# type: ignore[union-attr]`, `mypy` complains that these are not used:

```python
aiida/engine/processes/builder.py:96: error: unused 'type: ignore' comment
aiida/engine/processes/builder.py:98: error: unused 'type: ignore' comment
Found 2 errors in 1 file (checked 1 source file)
```

Don't know how to fix this due to my `mypy` ignorance. Will look into it, but @chrisjsewell: you prob can tell me how to fix this off the top of your head? :)

sphuber:
> y adding `# type: ignore[assignment]` and `# type: ignore[union-attr]`, `mypy` complains that these are not used:

Might be related to this issue: https://github.com/aiidateam/aiida-core/issues/5026  I have also suffered from this a few times and have no idea where it comes from nor how to fix it.

chrisjsewell:
> Will look into it, but @chrisjsewell: you prob can tell me how to fix this off the top of your head? :)

How do I never seem to have these issues 😆
I would suggest you make sure your version of mypy is up-to-date, in fact I would suggest you use `tox -r -e py38-pre-commit -- --all mypy`

mbercx:
> How do I never seem to have these issues 😆
> I would suggest you make sure your version of mypy is up-to-date, in fact I would suggest you use `tox -r -e py38-pre-commit -- --all mypy`

_sigh_, alright! I'll finally start looking into using `tox`. 😝

mbercx:
@chrisjsewell finally got around to having a look at `tox`! Good stuff 👌 

With all tests passing and the codecov gods appeased, this one is finally ready for review.

chrisjsewell:
Better late than never 😄

aiida/engine/processes/builder.py
```diff
@@ -238,6 +244,43 @@ def __init__(self, process_class: Type['Process']):
         self._process_spec = self._process_class.spec()
         super().__init__(self._process_spec.inputs)
 
+    def _repr_pretty_(self, p, _):  # pylint: disable=invalid-name
+        """Pretty representation for in the IPython console and notebooks."""
+
+        def format_inputs(inputs, level=0):
```
chrisjsewell, line 250: I would just extract this as a separate staticmethod on the class:

```python
@staticmethod
def _format_inputs(inputs: dict, level: int=0) -> str:
```

aiida/engine/processes/builder.py
```diff
@@ -238,6 +244,43 @@ def __init__(self, process_class: Type['Process']):
...
+                    result.append(fmt_indent('},', level))
+                elif isinstance(value, Dict):
+                    result.append(fmt_indent(f'"{key}": {{', level))
+                    for line in json.dumps(value.get_dict(), indent=4).split('\n'):
+                        if line not in ['{', '}', '{}']:
+                            result.append(fmt_indent(f'{line}', level))
+                    result.append(fmt_indent('},', level))
+                elif isinstance(value, BaseType):
+                    result.append(fmt_indent(f'"{key}": {value.value}', level))
+                elif isinstance(value, Structure):
```
chrisjsewell, line 275: We probably should not start doing "special cases" like this, and ideally use `Node.get_description`, which e.g. for `StructureData.get_description` returns `self.get_formula(mode='hill_compact')`

To be hypocritical, there is similar special-casing I added years back in: https://github.com/aiidateam/aiida-core/blob/2c5cc6fac2a1cea651d06089e44441a7e4e238fc/aiida/tools/visualization/graph.py#L204

but yeh both of these should really be standardised (I guess this ties in a bit with https://github.com/aiidateam/aiida-core/issues/3087)

chrisjsewell (changes requested):
Few comments to address ta

aiida/engine/processes/builder.py
mbercx: I agree that adding special cases is probably a bad idea long term. A bit confusing that there is both a `description` property and `get_description()` method, which return different strings.

I'm not sure if `get_description` should be the default fallback, or the `_repr_`. For a `Code`, you now get a good amount of information in one line:

```
Remote code 'qe-v6.7-pw' on localhost, pk: 1, uuid: 47a1c120-47af-4b9e-80c2-5a647c99817e
```

But if we switch to `get_description`, you get the description that was provided during code setup. That doesn't contain the PK, for example (similar issue for the `StructureData`, but that can be fixed).

aiida/engine/processes/builder.py
mbercx: Had to make it a `classmethod` since it's recursive.

aiida/engine/processes/builder.py
sphuber: The fact that `Code.get_description` returns the value of the `description` property is just a fault in the `Code.get_description` implementation. That method is intended to give a rich description of the object. It is true that the name is not ideal and the implementation for the node types that ship with `aiida-core` are very inconsistent. But I agree with @chrisjsewell , we shouldn't compensate those errors by hard-coding special cases here. We should simply improve the `get_description` implementation of those node types.

aiida/engine/processes/builder.py
```diff
@@ -8,20 +8,25 @@
...
 from aiida.orm import Node
+from aiida.orm.nodes.data.base import BaseType
+from aiida.plugins import DataFactory
 
 if TYPE_CHECKING:
     from aiida.engine.processes.process import Process
 
 __all__ = ('ProcessBuilder', 'ProcessBuilderNamespace')
 
+Dict = DataFactory('dict')
```
chrisjsewell, line 26: just import this from `aiida.orm`

aiida/engine/processes/builder.py
```diff
@@ -242,3 +247,41 @@ def __init__(self, process_class: Type['Process']):
...
+        for key, value in sorted(inputs.items()):
+            if isinstance(value, (dict, ProcessBuilder, ProcessBuilderNamespace)) and len(value) > 0:
+                result.append(fmt_indent(f'"{key}": {{', level))
+                result.append(cls._format_inputs(value, level=level + 1))
+                result.append(fmt_indent('},', level))
+            elif isinstance(value, Dict):
+                result.append(fmt_indent(f'"{key}": {{', level))
+                for line in json.dumps(value.get_dict(), indent=4).split('\n'):
+                    if line not in ['{', '}', '{}']:
+                        result.append(fmt_indent(f'{line}', level))
```
chrisjsewell, line 278: again feels here you should just be using `textwrap.indent`

chrisjsewell (changes requested):

aiida/engine/processes/builder.py
chrisjsewell: (I would basically advocate for never using these plugin loaders, because they are not statically analysable)

chrisjsewell: Not that I recommend lol (I'm sure I've already talked about this in an aiida meeting); we also "guarantee" what is publicly available to import from `aiida.orm`. Ok maybe if we move e.g. `StructureData` out of core you would have to change the import location, but I think that whole change would be breaking anyway and just a simple find/replace

But anyhow, certainly in core we always import rather than load

mbercx:
@chrisjsewell this one is in principle ready for another round of review. However, I remember we once discussed writing a JSON encoder instead and relying on that instead of the current rudimentary `_format_inputs` method. Let me know if you still want to discuss this, or if the current approach is fine.

sphuber:
Thanks @mbercx . This looks all fine to me, just have one question: will `_repr_pretty_` be automatically called when calling `print(builder)` in an interactive shell, or is the user expected to call something specific manually?

tests/engine/processes/test_builder.py
```diff
@@ -54,3 +58,62 @@ def test_access_methods():
...
+
+def test_merge():
+    """Test the ``ProcessBuilder`` merge methods."""
+    builder = ProcessBuilder(ArithmeticAddCalculation)
+
+    builder._update({'metadata': {'options': {'resources': {'num_machines': 1}}}})
+    assert builder.metadata.options.resources['num_machines'] == 1
+    builder._merge({'metadata': {'options': {'resources': {'num_cores_per_machine': 2}}}})
+    assert builder.metadata.options.resources['num_machines'] == 1
+    assert builder.metadata.options.resources['num_cores_per_machine'] == 2
```
sphuber, line 91: Are these being added due to a rebase-fu?

aiida/engine/processes/builder.py
```diff
@@ -242,3 +259,10 @@ def __init__(self, process_class: Type['Process']):
     def process_class(self) -> Type['Process']:
         """Return the process class for which this builder is constructed."""
         return self._process_class
+
+    def _repr_pretty_(self, p, _) -> str:  # pylint: disable=invalid-name
+        """Pretty representation for in the IPython console and notebooks."""
+
```
sphuber, line 265:
```suggestion
```

aiida/engine/processes/builder.py
```diff
@@ -242,3 +259,10 @@ def __init__(self, process_class: Type['Process']):
...
+        return p.text(
+            f'Process class: {self._process_class.get_name()}\nInputs:\n{json.dumps(self, cls=PrettyEncoder, indent=4)}'
```
sphuber, line 267: `get_name()` wait what now? What is this method?

sphuber (changes requested):

tests/engine/processes/test_builder.py
mbercx: Nono, I simply wanted to appease the codecov gods here. :)

aiida/engine/processes/builder.py
mbercx: Umm, I found it when I was trying to get a short version of the process class:

```python
In [1]: from aiida.calculations.arithmetic.add import ArithmeticAddCalculation

In [2]: builder = ArithmeticAddCalculation.get_builder()

In [3]: builder._process_class
Out[3]: aiida.calculations.arithmetic.add.ArithmeticAddCalculation

In [4]: builder._process_class.get_name()
Out[4]: 'ArithmeticAddCalculation'
```

sphuber: Ah, it comes from `plumpy.Process`:
```python
Signature: builder._process_class.get_name() -> str
Source:   
    @classmethod
    def get_name(cls) -> str:
        """Return the process class name."""
        return cls.__name__
File:      ~/code/aiida/env/dev/plumpy/plumpy/processes.py
Type:      method
```
To be honest, I could simply use the builtin `__name__`. See no reason not to

tests/engine/processes/test_builder.py
sphuber: I see. Well I was sure this was already being tested somewhere. It looks like there are two files to test the `ProcessBuilder` and these tests are already in `tests/engine/test_process_builder`.  Think the file you were updating should be the one to keep, so would be good to move the tests from the other there and then you can skip adding these tests as they should be duplicates (largely).

mbercx: This is why the tests folder structure should be the same as the package. 😅

mbercx: So I would suggest to move the `TestProcessBuilder` tests in `tests/engine/test_process_builder` to `tests/engine/process/test_builder` instead.

sphuber: With the risk of being the scope-creeper: the tests of the former are in the old form compared to the pytest style of the latter. Would be might magnificent if you would be willing to update them :innocent:

mbercx: Haha, ok, but only because you _didn't_ ask me to do this in https://github.com/aiidateam/aiida-core/pull/4428/files. 😁

mbercx: alldone

mbercx:
Note: Although I don't want to introduce this massive scope creep in this PR: I think long term this should be part of the discussion in #3087, i.e. do we implement a `_repr__pretty_` method for _all_ nodes, and then have the `PrettyEncoder` use this instead?

mbercx:
> Note: Although I don't want to introduce this massive scope creep in this PR: I think long term this should be part of the discussion in #3087, i.e. do we implement a `_repr__pretty_` method for _all_ nodes, and then have the `PrettyEncoder` use this instead?

mbercx:
(oops, slippery fingers! 😅 )

mbercx:
Actually, maybe it makes _a lot_ more sense to output this in a YAML format... It's meant to be human readable after all. For the `PwBandsWorkChain` example:

<details>
<summary>See pretty representation of builder</summary>

```python
Process class: PwBandsWorkChain
Inputs:
bands:
  metadata: {}
  pw:
    code: Plane wave Quantum ESPRESSO v6.6 - local compilation
    metadata:
      options:
        max_wallclock_seconds: 43200
        resources:
          num_machines: 1
        stash: {}
        withmpi: true
    parameters:
      CONTROL:
        calculation: scf
        etot_conv_thr: 2.0e-05
        forc_conv_thr: 0.0001
        tprnfor: true
        tstress: true
      ELECTRONS:
        conv_thr: 4.0e-10
        electron_maxstep: 80
        mixing_beta: 0.4
      SYSTEM:
        degauss: 0.01
        ecutrho: 240.0
        ecutwfc: 30.0
        nosym: false
        occupations: smearing
        smearing: cold
    pseudos:
      Si: ''
bands_kpoints_distance: 0.025
clean_workdir: true
metadata: {}
nbands_factor: 3.0
relax:
  base:
    kpoints_distance: 0.15
    kpoints_force_parity: false
    metadata: {}
    pw:
      code: Plane wave Quantum ESPRESSO v6.6 - local compilation
      metadata:
        options:
          max_wallclock_seconds: 43200
          resources:
            num_machines: 1
          stash: {}
          withmpi: true
      parameters:
        CELL:
          cell_dofree: all
          press_conv_thr: 0.5
        CONTROL:
          calculation: vc-relax
          etot_conv_thr: 2.0e-05
          forc_conv_thr: 0.0001
          tprnfor: true
          tstress: true
        ELECTRONS:
          conv_thr: 4.0e-10
          electron_maxstep: 80
          mixing_beta: 0.4
        SYSTEM:
          degauss: 0.01
          ecutrho: 240.0
          ecutwfc: 30.0
          nosym: false
          occupations: smearing
          smearing: cold
      pseudos:
        Si: ''
  base_final_scf:
    metadata: {}
    pw:
      metadata:
        options:
          stash: {}
      pseudos: {}
  max_meta_convergence_iterations: 5
  meta_convergence: true
  metadata: {}
  volume_convergence: 0.02
scf:
  kpoints_distance: 0.15
  kpoints_force_parity: false
  metadata: {}
  pw:
    code: Plane wave Quantum ESPRESSO v6.6 - local compilation
    metadata:
      options:
        max_wallclock_seconds: 43200
        resources:
          num_machines: 1
        stash: {}
        withmpi: true
    parameters:
      CONTROL:
        calculation: scf
        etot_conv_thr: 2.0e-05
        forc_conv_thr: 0.0001
        tprnfor: true
        tstress: true
      ELECTRONS:
        conv_thr: 4.0e-10
        electron_maxstep: 80
        mixing_beta: 0.4
      SYSTEM:
        degauss: 0.01
        ecutrho: 240.0
        ecutwfc: 30.0
        nosym: false
        occupations: smearing
        smearing: cold
    pseudos:
      Si: ''
structure: Si
```
</details>

tests/engine/processes/test_builder.py
```diff
@@ -8,49 +8,442 @@
...
+        "name_spaced": "underscored",
+        "dict": {
+            "a": 1,
+            "b": {
+                "c": 2
+            }
+        },
+        "boolean": true
+    }"""
+    pretty_repr = re.sub(r'(?m)^\s{4}', '', pretty_repr)
```
sphuber, line 447: Instead of this, I think it is better to use the `dedent` method from the `textwrap` package. This is clearer I think to what it is doing and we already use it elsewhere in the tests. See comment above how to do it
```suggestion
    pretty_repr = textwrap.dedent(
    """Process class: ExampleWorkChain
    Inputs:
    {
        "metadata": {},
        "dynamic": {
            "namespace": {
                "alp": 1
            }
        },
        "name": {
            "spaced": 1
        },
        "name_spaced": "underscored",
        "dict": {
            "a": 1,
            "b": {
                "c": 2
            }
        },
        "boolean": true
    }""")
```

sphuber (changes requested):
Thanks @mbercx just two minor comments remaining. And then if you are still up for it, I would take you up on your suggestion to use YAML instead of JSON.

aiida/engine/processes/builder.py
```diff
@@ -242,3 +260,10 @@ def __init__(self, process_class: Type['Process']):
     def process_class(self) -> Type['Process']:
         """Return the process class for which this builder is constructed."""
         return self._process_class
+
+    def _repr_pretty_(self, p, _) -> str:  # pylint: disable=invalid-name
+        """Pretty representation for in the IPython console and notebooks."""
+        return p.text(
+            f'Process class: {self._process_class.__name__}\n'
+            f'Inputs:\n{yaml.safe_dump(json.JSONDecoder().decode(PrettyEncoder().encode(self)))}'
```
sphuber, line 268: Doing a full encode/decode cycle in JSON before encoding in YAML is a bit inefficient. Then again, writing serializers in YAML is a bit of a mess (like in `aiida.orm.utils.serialize`, and unfortunately we cannot even reuse that) and this should not be horribly costly, so let's leave this as is.

sphuber (approved):

sphuber:
Thanks a lot @mbercx !

aiida/engine/processes/builder.py
mbercx: Yeah, I know it's a bit of a messy workaround, but didn't know how to write the serializer in YAML to be honest. I timed the line with `%%timeit` and it only took a couple dozen nanoseconds, so should be ok. ^^

PR #4971, opened 2021-06-03
add sphinx dependency to test extra

### LINKED ISSUE #4958: pytest is not able to run without `sphinx` installed.
[version]: develop branch 2f1c1f82c5085bd311303dfc82c0a3bf67629377

I believe some newly added test are depended on sphinx, so it should also be installed in 'tests' extra.

```
  File "/home/unkcpz/miniconda3/envs/aiida-core-dev/lib/python3.8/site-packages/_pytest/config/__init__.py", line 703, i
n import_plugin                                                                                                         
    __import__(importspec)                                                                                              
ModuleNotFoundError: No module named 'sphinx.testing.fixtures' 
```
Installing sphinx manually resolve this.

unkcpz:
fix #4958 

hi @csadorf , here is the fix, I am not sure is this good to have same package `sphinx` in both `tests` and `docs`.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -114,7 +114,8 @@
             "pytest-benchmark~=3.2",
             "pympler~=0.9",
             "coverage<5.0",
-            "sqlalchemy-diff~=0.1.3"
+            "sqlalchemy-diff~=0.1.3",
+            "sphinx~=3.2.1"
         ],
         "bpython": [
             "bpython~=0.18.0"


csadorf:
> hi @csadorf , here is the fix, I am not sure is this good to have same package `sphinx` in both `tests` and `docs`.

Thank you very much! Yes, that is fine. They don't even have to rely on the exact same versions as long as the constraints are compatible. In case that both extras are installed, the extra dependencies are unioned and thus automatically resolved. Our installation should catch any eventual conflict in the future.

csadorf (approved):

PR #4977, opened 2021-06-07
🐛 FIX: Initialising a Node with a User

chrisjsewell:
So it appears that no one has ever tried to specifically set a user for a node (e.g. `Data(user=another_user)`) because, as you see from the diff, it would then try to call `.backend_entity` on the backend entity and fail.
Disappointing that mypy didn't pick this up in the type checking (perhaps there is a stricter setting that I haven't added)

cc @ltalirz as this came up with respect to the REST API (i.e. if we want to create a node for a specific user)

### DIFF
--- a/aiida/manage/tests/pytest_fixtures.py
+++ b/aiida/manage/tests/pytest_fixtures.py
@@ -70,6 +70,13 @@ def clear_database_before_test(aiida_profile):
     yield
 
 
+@pytest.fixture(scope='class')
+def clear_database_before_test_class(aiida_profile):
+    """Clear the database before a test class."""
+    aiida_profile.reset_db()
+    yield
+
+
 @pytest.fixture(scope='function')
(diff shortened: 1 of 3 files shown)


sphuber (changes requested):
Thanks @chrisjsewell . Straight forward enough, but since this wasn't caught by tests and is so basic, would be good to add a simple test for this.

chrisjsewell:
> but since this wasn't caught by tests and is so basic, would be good to add a simple test for this.

yeh yeh I knew you were going to say this 😛 Just wanted to quickly add the code change via github first, so I didn't forget

chrisjsewell:
@sphuber I converted all of `tests/orm/node/test_node.py` to pytest and added some extra tests

sphuber (approved):
Thanks @chrisjsewell

PR #4984, opened 2021-06-11
`ProcessBuilder`: ensure instances do not interfere 

### LINKED ISSUE #4420: Unrelated builders interfere with each other
### Describe the bug

The `ProcessBuilderNamespace` of unrelated processes can interfere with each other, because their properties are set on the `ProcessBuilderNamespace` class instead of the specific instance.

### Steps to reproduce

WorkChain code:

```python
class First(WorkChain):
    @classmethod
    def define(cls, spec):
        super().define(spec)
        spec.input('a', default=lambda: orm.Int(1))
        spec.input('b')


class Second(WorkChain):
    @classmethod
    def define(cls, spec):
        super().define(spec)
        spec.input('a', default=lambda: orm.Int(2))
        spec.input('c')
```

Inspect builder behavior:
```python
In [1]: from test_proc import First, Second                                                                                                                                                                                           

In [2]: builder = First.get_builder() 

In [3]: Second.get_builder() # We simply construct this builder, and do nothing with it.                                                                                                           

In [5]: builder.a() # The default for 'a' has changed to the lambda defined in `Second`.
Out[5]: <Int: uuid: fb6a1556-9f34-4d32-b7ea-cc2800ee00ef (unstored) value: 2> 

In [6]: builder.c # This should raise, since `First` doesn't have a 'c' input.
```

**EDIT:** When actually _running_ the workchain, the default is populated correctly, and validation is also correct. I guess this slightly lowers the severity, but depending on how the user manipulates the builder it could still lead to incorrect inputs.

### Your environment

- Operating system: Ubuntu-18.04 on WSL2
- Python version: 3.7.3
- aiida-core version: `develop` branch

### Additional context

This is a follow-up from a discussion from https://github.com/aiidateam/aiida-core/pull/4419

sphuber:
Fixes #4420 

There was a bug in the `ProcessBuilderNamespace` constructor where it
converts the ports of the port namespace into dynamic properties and
assigns them to the class. The problem here is that the property is
added to the class instead of the instance. This meant that if you have
two classes that happen to have ports with overlapping names, when
constructing the builder for the second, the properties of the first
builder that have port names.

The simple solution of simply add the dynamic properties to the instance
instead of the class won't work, because properties are descriptors and
can only be assigned to a class. The workaround is to create a subclass
on the fly and attach the properties to that new class.

### DIFF
--- a/aiida/engine/processes/builder.py
+++ b/aiida/engine/processes/builder.py
@@ -10,6 +10,7 @@
 """Convenience classes to help building the input dictionaries for Processes."""
 import collections
 from typing import Any, Type, TYPE_CHECKING
+from uuid import uuid4
 
 from aiida.orm import Node
 from aiida.engine.processes.ports import PortNamespace
@@ -41,6 +42,8 @@ class ProcessBuilderNamespace(collections.abc.MutableMapping):
         self._valid_fields = []
         self._data = {}
 
+        dynamic_properties = {}
+
         # The name and port objects have to be passed to the defined functions as defaults for
         # their arguments, because this way the content at the time of defining the method is
         # saved. If they are used directly in the body, it will try to capture the value from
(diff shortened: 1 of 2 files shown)


chrisjsewell:
I take it this is built on (and requires first merging) #4983?

sphuber:
> I take it this is built on (and requires first merging) #4983?

It does indeed. The changes in this PR are minimal and can already be reviewed, but it is not the easiest to distill them from the other changes. Thats why I marked it blocked so we can merge the other first and then this should be straightforward

chrisjsewell:
ugghh, this does look kosher lol. Are sure we cannot use `__getattr__` here?

sphuber:
> ugghh, this does look kosher lol. 

I think you mean _doesn't_ ?

> Are sure we cannot use `__getattr__` here? (as already mentioned in [#4420 (comment)](https://github.com/aiidateam/aiida-core/issues/4420#issuecomment-702752422))

Feel free to try. But with the properties we can get add the help string of the port to the `__doc__` of the property. Don't think we can emulate this behavior using `__getattr__`.

sphuber:
@chrisjsewell I am merging this tomorrow. So if you still want to have a look at another solution, please do so a.s.a.p.

chrisjsewell:
> So if you still want to have a look at another solution, please do so a.s.a.p.

ok gimme til midday tomorrow

chrisjsewell (approved):
Hmm, ok I'll approve this.

I get that the dynamic attribute access help in the context of ipython shells, but urgh, it does feel these builders are getting overly complex.
Then again, I don't have a better solution right now 🤷

PR #4990, opened 2021-06-18
Exception not raise properly when number of codes>2 in calc_info 

unkcpz:
If `calc_info.codes_run_mode` is not set when using more than one code in a calc_info, the `PluginInternalError` is not properly raised. `calc_info.codes_run_mode` will be a `None`, and the exception never raised here.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,9 +4,15 @@ repos:
     hooks:
     -   id: double-quote-string-fixer
     -   id: end-of-file-fixer
+        exclude: &exclude_pre_commit_hooks >
+            (?x)^(
+                tests/.*(?<!\.py)$
+            )$
     -   id: fix-encoding-pragma
     -   id: mixed-line-ending
     -   id: trailing-whitespace
+        exclude: *exclude_pre_commit_hooks
+
 
 -   repo: https://github.com/ikamensh/flynt/
     rev: '0.55'
--- a/aiida/common/datastructures.py
+++ b/aiida/common/datastructures.py
@@ -80,7 +80,8 @@ class CalcInfo(DefaultFieldsAttributeDict):
         either, for example, because they contain proprietary information or because they are big and their content is
         already indirectly present in the repository through one of the data nodes passed as input to the calculation.
     * codes_info: a list of dictionaries used to pass the info of the execution of a code
-    * codes_run_mode: a string used to specify the order in which multi codes can be executed
+    * codes_run_mode: the mode of execution in which the codes will be run (`CodeRunMode.SERIAL` by default,
+        but can also be `CodeRunMode.PARALLEL`)
     * skip_submit: a flag that, when set to True, orders the engine to skip the submit/update steps (so no code will
         run, it will only upload the files and then retrieve/parse).
     """
--- a/aiida/engine/processes/calcjobs/calcjob.py
+++ b/aiida/engine/processes/calcjobs/calcjob.py
@@ -593,12 +593,17 @@ class CalcJob(Process):
                 raise PluginInternalError('CalcInfo should have the information of the code to be launched')
             this_code = load_node(code_info.code_uuid, sub_classes=(Code,))
 
-            this_withmpi = code_info.withmpi  # to decide better how to set the default
-            if this_withmpi is None:
-                if len(calc_info.codes_info) > 1:
-                    raise PluginInternalError('For more than one code, it is necessary to set withmpi in codes_info')
-                else:
-                    this_withmpi = self.node.get_option('withmpi')
+            # To determine whether this code should be run with MPI enabled, we get the value that was set in the inputs
+            # of the entire process, which can then be overwritten by the value from the `CodeInfo`. This allows plugins
+            # to force certain codes to run without MPI, even if the user wants to run all codes with MPI whenever
+            # possible. This use case is typically useful for `CalcJob`s that consist of multiple codes where one or
+            # multiple codes always have to be executed without MPI.
+
+            this_withmpi = self.node.get_option('withmpi')
+
+            # Override the value of `withmpi` with that of the `CodeInfo` if and only if it is set
+            if code_info.withmpi is not None:
+                this_withmpi = code_info.withmpi
 
             if this_withmpi:
                 this_argv = (
@@ -615,15 +620,12 @@ class CalcJob(Process):
             codes_info.append(code_info)
         job_tmpl.codes_info = codes_info
 
-        # set the codes execution mode
+        # set the codes execution mode, default set to `SERIAL`
+        codes_run_mode = CodeRunMode.SERIAL
+        if calc_info.codes_run_mode:
+            codes_run_mode = calc_info.codes_run_mode
 
-        if len(codes) > 1:
-            try:
-                job_tmpl.codes_run_mode = calc_info.codes_run_mode
-            except KeyError as exc:
-                raise PluginInternalError('Need to set the order of the code execution (parallel or serial?)') from exc
-        else:
-            job_tmpl.codes_run_mode = CodeRunMode.SERIAL
+        job_tmpl.codes_run_mode = codes_run_mode
         ########################################################################
 
         custom_sched_commands = self.node.get_option('custom_scheduler_commands')
--- a/tests/engine/test_calc_job.py
+++ b/tests/engine/test_calc_job.py
@@ -75,6 +75,99 @@ class FileCalcJob(CalcJob):
 
 
 @pytest.mark.requires_rmq
+class MultiCodesCalcJob(CalcJob):
+    """`MultiCodesCalcJob` implementation to test the calcinfo with multiple codes set.
+
+    The codes are run in parallel. The codeinfo1 is set to run without mpi, and codeinfo2
+    is set to run with mpi.
+    """
+
+    @classmethod
+    def define(cls, spec):
+        super().define(spec)
+        spec.input('parallel_run', valid_type=orm.Bool)
+
+    def prepare_for_submission(self, folder):
+        from aiida.common.datastructures import CalcInfo, CodeInfo, CodeRunMode
+
+        codeinfo1 = CodeInfo()
+        codeinfo1.code_uuid = self.inputs.code.uuid
+        codeinfo1.withmpi = False
+
+        codeinfo2 = CodeInfo()
+        codeinfo2.code_uuid = self.inputs.code.uuid
+
+        calcinfo = CalcInfo()
+        calcinfo.codes_info = [codeinfo1, codeinfo2]
+        if self.inputs.parallel_run:
+            calcinfo.codes_run_mode = CodeRunMode.PARALLEL
+        else:
+            calcinfo.codes_run_mode = CodeRunMode.SERIAL
+        return calcinfo
+
+
+@pytest.mark.requires_rmq
+@pytest.mark.usefixtures('dry_run_in_tmp')
+@pytest.mark.usefixtures('clear_database_before_test')
+@pytest.mark.parametrize('parallel_run', [True, False])
+def test_multi_codes_run_parallel(aiida_local_code_factory, file_regression, parallel_run):
+    """test codes_run_mode set in CalcJob"""
+
+    inputs = {
+        'code': aiida_local_code_factory('arithmetic.add', '/bin/bash'),
+        'parallel_run': orm.Bool(parallel_run),
+        'metadata': {
+            'dry_run': True,
+            'options': {
+                'resources': {
+                    'num_machines': 1,
+                    'num_mpiprocs_per_machine': 1
+                }
+            }
+        }
+    }
+
+    _, node = launch.run_get_node(MultiCodesCalcJob, **inputs)
+    folder_name = node.dry_run_info['folder']
+    submit_script_filename = node.get_option('submit_script_filename')
+    with open(os.path.join(folder_name, submit_script_filename)) as handle:
+        content = handle.read()
+
+    file_regression.check(content, extension='.sh')
+
+
+@pytest.mark.requires_rmq
+@pytest.mark.usefixtures('dry_run_in_tmp')
+@pytest.mark.usefixtures('clear_database_before_test')
+@pytest.mark.parametrize('calcjob_withmpi', [True, False])
+def test_multi_codes_run_withmpi(aiida_local_code_factory, file_regression, calcjob_withmpi):
+    """test withmpi set in CalcJob only take effect for codes which have codeinfo.withmpi not set"""
+
+    inputs = {
+        'code': aiida_local_code_factory('arithmetic.add', '/bin/bash'),
+        'parallel_run': orm.Bool(False),
+        'metadata': {
+            'dry_run': True,
+            'options': {
+                'resources': {
+                    'num_machines': 1,
+                    'num_mpiprocs_per_machine': 1
+                },
+                'withmpi': calcjob_withmpi,
+            }
+        }
+    }
+
+    _, node = launch.run_get_node(MultiCodesCalcJob, **inputs)
+    folder_name = node.dry_run_info['folder']
+    submit_script_filename = node.get_option('submit_script_filename')
+    with open(os.path.join(folder_name, submit_script_filename)) as handle:
+        content = handle.read()
+
+    file_regression.check(content, extension='.sh')
+
+
+@pytest.mark.requires_rmq
 class TestCalcJob(AiidaTestCase):
     """Test for the `CalcJob` process sub class."""
 
@@ -279,6 +372,7 @@ class TestCalcJob(AiidaTestCase):
 
         self.assertIn('exception occurred in presubmit call', str(context.exception))
 
+    @pytest.mark.usefixtures('dry_run_in_tmp')
     def test_run_local_code(self):
         """Run a dry-run with local code."""
         inputs = deepcopy(self.inputs)
@@ -294,6 +388,7 @@ class TestCalcJob(AiidaTestCase):
         for filename in self.local_code.list_object_names():
             self.assertTrue(filename in uploaded_files)
 
+    @pytest.mark.usefixtures('dry_run_in_tmp')
     def test_provenance_exclude_list(self):
         """Test the functionality of the `CalcInfo.provenance_exclude_list` attribute."""
         import tempfile
--- a/tests/engine/test_calc_job/test_multi_codes_run_parallel_False_.sh
+++ b/tests/engine/test_calc_job/test_multi_codes_run_parallel_False_.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+exec > _scheduler-stdout.txt
+exec 2> _scheduler-stderr.txt
+
+
+'/bin/bash'   
+
+'/bin/bash'   
--- a/docs/source/howto/plugin_codes.rst
+++ b/docs/source/howto/plugin_codes.rst
@@ -183,6 +183,12 @@ In our example, the two input files are already stored in the AiiDA file reposit
 The ``retrieve_list`` on the other hand tells the engine which files to retrieve from the directory where the job ran after it has finished.
 All files listed here will be store in a |FolderData| node that is attached as an output node to the calculation with the label ``retrieved``.
 
+Finally, we pass the |CodeInfo| to a |CalcInfo| object.
+One calculation job can involve more than one executable, so ``codes_info`` is a list.
+If you have more than one executable in your ``codes_info``, you can set ``codes_run_mode`` to specify the mode with which these will be executed (`CodeRunMode.SERIAL` by default).
+We define the ``retrieve_list`` of filenames that the engine should retrieve from the directory where the job ran after it has finished.
+The engine will store these files in a |FolderData| node that will be attached as an output node to the calculation with the label ``retrieved``.
+
 .. admonition:: Further reading
 
     There are :ref:`other file lists available<topics:calculations:usage:calcjobs:file_lists>` that allow you to easily customize how to move files to and from the remote working directory in order to prevent the creation of unnecessary copies.
--- a/requirements/requirements-py-3.7.txt
+++ b/requirements/requirements-py-3.7.txt
@@ -112,6 +112,7 @@ pytest-benchmark==3.2.3
 pytest-cov==2.10.1
 pytest-rerunfailures==9.1.1
 pytest-timeout==1.4.2
+pytest-regressions==2.2.0
 python-dateutil==2.8.1
 python-editor==1.0.4
 python-memcached==1.59
--- a/requirements/requirements-py-3.8.txt
+++ b/requirements/requirements-py-3.8.txt
@@ -111,6 +111,7 @@ pytest-benchmark==3.2.3
 pytest-cov==2.10.1
 pytest-rerunfailures==9.1.1
 pytest-timeout==1.4.2
+pytest-regressions==2.2.0
 python-dateutil==2.8.1
 python-editor==1.0.4
 python-memcached==1.59
--- a/requirements/requirements-py-3.9.txt
+++ b/requirements/requirements-py-3.9.txt
@@ -111,6 +111,7 @@ pytest-benchmark==3.2.3
 pytest-cov==2.10.1
 pytest-rerunfailures==9.1.1
 pytest-timeout==1.4.2
+pytest-regressions==2.2.0
 python-dateutil==2.8.1
 python-editor==1.0.4
 python-memcached==1.59
--- a/setup.json
+++ b/setup.json
@@ -112,6 +112,7 @@
             "pytest-cov~=2.7,<2.11",
             "pytest-rerunfailures~=9.1,>=9.1.1",
             "pytest-benchmark~=3.2",
+            "pytest-regressions~=2.2",
             "pympler~=0.9",
             "coverage<5.0",
             "sqlalchemy-diff~=0.1.3",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -365,3 +365,11 @@ def with_daemon():
 
     # Note this will always be executed after the yield no matter what happened in the test that used this fixture.
     os.kill(daemon.pid, signal.SIGTERM)
+
+
+@pytest.fixture(scope='function')
+def dry_run_in_tmp(request, tmpdir):
+    """change to the tmp directory to do the dry run, then change back to the calling directory to avoid side-effects"""
+    os.chdir(tmpdir)
+    yield
+    os.chdir(request.config.invocation_dir)
--- a/tests/engine/test_calc_job/test_multi_codes_run_parallel_True_.sh
+++ b/tests/engine/test_calc_job/test_multi_codes_run_parallel_True_.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+exec > _scheduler-stdout.txt
+exec 2> _scheduler-stderr.txt
+
+
+'/bin/bash'    &
+
+'/bin/bash'    &
+
+wait
+
--- a/tests/engine/test_calc_job/test_multi_codes_run_withmpi_False_.sh
+++ b/tests/engine/test_calc_job/test_multi_codes_run_withmpi_False_.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+exec > _scheduler-stdout.txt
+exec 2> _scheduler-stderr.txt
+
+
+'/bin/bash'   
+
+'/bin/bash'   
--- a/tests/engine/test_calc_job/test_multi_codes_run_withmpi_True_.sh
+++ b/tests/engine/test_calc_job/test_multi_codes_run_withmpi_True_.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+exec > _scheduler-stdout.txt
+exec 2> _scheduler-stderr.txt
+
+
+'/bin/bash'   
+
+'mpirun' '-np' '1' '/bin/bash'   


sphuber (commented):
Thanks @unkcpz . Shouldn't we just define a default maybe instead of raising? Maybe setting `SERIAL` as a default makes sense and people can switch to `PARALLEL` if they need. Also we really should add a test for this. Should be straightforward enough, just mock the process and call `presubmit`

unkcpz:
Hi @sphuber , I have added a test and also set the default for `codeinfo.withmpi`. Instead of directly mocking a process, I create and run a `MultiCodesCalcJob`, and then check the content of job script generated.

aiida/common/datastructures.py
```diff
@@ -80,7 +80,8 @@ class CalcInfo(DefaultFieldsAttributeDict):
         either, for example, because they contain proprietary information or because they are big and their content is
         already indirectly present in the repository through one of the data nodes passed as input to the calculation.
     * codes_info: a list of dictionaries used to pass the info of the execution of a code
-    * codes_run_mode: a string used to specify the order in which multi codes can be executed
+    * codes_run_mode: `CodeRunMode.SERIAL` or `CodeRunMode.PARALLEL` to specify the order
+        in which multi codes can be executed
```
ramirezfranciscof, line 84: So if I understand correctly this keyword controls if the codes provided will be run one after the other or all at the same time, right? I don't think I would understand this as "order" in which the codes can be executed. I think the noun "mode" you use for keyword is also a better descriptor for its behavior. So something like:

```suggestion
    * codes_run_mode: the mode of execution in which the codes will be run (`CodeRunMode.SERIAL` by default, but can also be `CodeRunMode.PARALLEL`)
```
---
**SIDENOTE** (meaning, this is not super relevant, if it even makes sense we can open a separate issue for this)

I am now wondering why do we need the `CodeRunMode` construct. Are we anticipating there might be other customizable options we might pass through this keyword? Because in principle there is nothing conceptually besides serial and parallel, so this could simply be a flag no? @sphuber ?

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -593,12 +593,9 @@ def presubmit(self, folder: Folder) -> CalcInfo:
...
 
-            this_withmpi = code_info.withmpi  # to decide better how to set the default
-            if this_withmpi is None:
-                if len(calc_info.codes_info) > 1:
-                    raise PluginInternalError('For more than one code, it is necessary to set withmpi in codes_info')
-                else:
-                    this_withmpi = self.node.get_option('withmpi')
+            this_withmpi = self.node.get_option('withmpi')  # set code mpi, default set to as calcjob one
+            if code_info.withmpi:
+                this_withmpi = code_info.withmpi
```
ramirezfranciscof, line 598: So, this change actually makes more sense to me (why was previously the `withmpi` flag related to how many codes where provided before or even in what mode these where run? I mean, `withmpi` only refers to paralellization **within** each run of a code, not if these codes runs are then performed in parallel or serial). But just to mention that it would be good to also add some comment on this change on the commit log.

tests/engine/test_calc_job.py
```diff
@@ -74,6 +74,62 @@ def prepare_for_submission(self, folder):
         return calcinfo
 
 
+@pytest.mark.requires_rmq
+class MultiCodesCalcJob(CalcJob):
+    """`MultiCodesCalcJob` implementation to test the calcinfo with multiple codes set.
+
+    The codes are run in parallel. The codeinfo1 is set to run without mpi, and codeinfo2
+    is set to run with mpi.
```
ramirezfranciscof, line 82: Mmm so I know technically all other tests with the `FileCalcJob` run """serially""", but shouldn't we also check that it works correctly when providing two codes like you do below?

You could just define an input flag in the calcjob and have that activate whether it should set  `calcinfo.codes_run_mode = CodeRunMode.PARALLEL` or not in the `prepare_for_submission`. Then you can just parametrize the test you already made.

ramirezfranciscof (changes requested):
Hey @unkcpz thanks for the contribution! I have some minor comments on it, could you take a look? Let me know if there is anything that is not clear or you don't agree with.

aiida/engine/processes/calcjobs/calcjob.py
unkcpz: > why was previously the withmpi flag related to how many codes where provided before or even in what mode these where run? I mean, withmpi only refers to paralellization within each run of a code, not if these codes runs are then performed in parallel or serial

Not exactly, for example if I have two codes, where the first is a pre-processing helper code must be run without mpi and the second one is the real simulation code better run with mpi. The user may only set `withmpi=True` as a flag in calcjob in intuitively without setting `withmpi=False` paticuliarly for the first hepler code. This will fail the running. 
I guess this is the reason why this default setting is not being properly set previously.

tests/engine/test_calc_job.py
unkcpz: Sorry @ramirezfranciscof, I am not sure I fully understand your suggestion. Do you mean there is no need to have new `MultiCodeCalcJob` but I can test it in `FileCalcJob` with two codes? 

If so. Hmmm... That makes sense, I can try it.

unkcpz (commented):
Hi @ramirezfranciscof, thanks for the review. I accept most of your advice and leave some notes for your comment. Please have another look at your convenience. Cheers!

aiida/common/datastructures.py
sphuber: In a sense yes it is a "flag" as it only has two values, but the "flag" concept only really applies to a command-line interface where either you specify it or not. Even there, if you don't specify it, there is still some default value, namely the feature to which the flag corresponds being turned off. Here, you _could_ think of a "parallel" flag, which when enabled would turn it on, otherwise it would be serial. However, how would you implement this? You would still need a variable with a default (`SERIAL`) value and a value to change it. So in the end, you end up exactly with what we have now. I am not sure how you think a "flag" should be implemented here.

.gitignore
```diff
@@ -17,6 +17,7 @@
 .eggs
 .vscode
 .tox
+submit_test/
```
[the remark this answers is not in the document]
ramirezfranciscof: Ahh I see your point. Yes, I think your solution sounds reasonable. I'm surprised that this hasn't come up before; @sphuber is this the only test which is executing a `dry_run` and comparing the resulting script?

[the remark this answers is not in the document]
sphuber: There are other tests that run a `dry_run`, and so create this folder, but they don't compare the input script.

tests/engine/test_calc_job.py
ramirezfranciscof: Mmm no, sorry I may not have been very clear.

I was just pointing out that you were testing `CodeRunMode.PARALLEL` but not `CodeRunMode.SERIAL` for multiple codes. I only mentioned `FileCalcJob` as a recognition that in principle we are implicitly using `CodeRunMode.SERIAL` in the other tests that were already there and used that prior class, but they were always testing it when there was only one code. The suggestions I made were modifications on `MultiCodeCalcJob` so that you could also use that to test `CodeRunMode.SERIAL` with multiple codes too:

```
class MultiCodeCalcJob(CalcJob):
    (...)
    spec.input('parallel_run', valid_type=Bool, default=False)
    (...)
    if parallel_run:
        calcinfo.codes_run_mode = CodeRunMode.PARALLEL
    else:
        calcinfo.codes_run_mode = CodeRunMode.SERIAL
   (...)


@pytest.mark.requires_rmq
@pytest.mark.parametrize('parallel_run', [True, False])
def test_multi_codes_run_parallel(aiida_local_code_factory, file_regression):
    (...)
    inputs = {
        parallel_run: parallel_run,
    (...)
```

Is this clearer now? Sorry for the misunderstanding; I think I personally preferred when you had the dedicated `MultiCodeCalcJob` for your multiple code test but it is ok if you don't want to change that back. The important thing for me was to try both options.

.pre-commit-config.yaml
```diff
@@ -4,6 +4,10 @@ repos:
     hooks:
     -   id: double-quote-string-fixer
     -   id: end-of-file-fixer
+        exclude: |
+            (?x)^(
+                tests/engine/test_calc_job/.*|
+            )$
```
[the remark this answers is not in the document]
ramirezfranciscof: Uhm, I see the issue. I think I would adapt the test which is more particular rather than start polluting global config files with exceptions, but I'll admit this is a somewhat personal preference. I am thinking of something like this (I took it from [here](https://stackoverflow.com/a/275025), didn't check if it actually works, just to give you an idea):

``` python
    with open(os.path.join(folder_name, submit_script_filename)) as handle:
        content = handle.read()

    content = content.rstrip() + `\n` # linter compliance
    file_regression.check(content, extension='.sh')
```

But yeah, I will agree that is also not ideal. What do you think? @sphuber do you have any opinions on this, is there a better alternative we are missing?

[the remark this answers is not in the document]
sphuber: In a sense yes, it would be good to adapt the test to _not_ be sensitive to trailing whitespace, because we don't really care. Maybe you can actually use `file_regression.check` itself to ignore the whitespace. See [the documentation](https://pytest-regressions.readthedocs.io/en/latest/api.html#file-regression) which shows there is a `newline` argument. Not 100% how it works but it could be useful

.gitignore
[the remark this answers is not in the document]
ramirezfranciscof: Ok, I see, but the question is what do you do with all these newly created test folder/files so that they are not shown on git? I actually don't remember these appearing for me after running tests, are they deleted automatically somehow?

aiida/engine/processes/calcjobs/calcjob.py
ramirezfranciscof: Ahhhhh I understand now what the old code was doing. It was getting the `withmpi` flag from each code and if it didn't find it then it checked if there were one code, in which case it used the `withmpi` flag from the calcjob, or if there were many codes it just raised. It is a bit of a convoluted implementation though xD.

So, let me make sure I understand this: the `calcjob-withmpi` can be set **by the user** and is what is used by default on the codes, unless **the calcjob designer** specifically sets a `code-withmpi` which will override any behavior requested by the user. Is that correct?

Maybe it would be worth to make some info box explaining this in the [How-to sections for writing calcjobs](https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/plugin_codes.html#preparing-for-submission) (not saying to do it now necessarily, we can add an issue for this, but do feel free if you are inspired to do so).

aiida/common/datastructures.py
ramirezfranciscof: Yeah with flag I didn't mean in exclusively in the "user interface" way but as in a boolean. So it would work the same way basically no? I guess the relevant change to understand what I mean is in `aiida/schedulers/scheduler.py`, going from this:

``` python
        if codes_run_mode == CodeRunMode.PARALLEL:
            list_of_runlines.append('wait\n')
            return ' &\n\n'.join(list_of_runlines)

        if codes_run_mode == CodeRunMode.SERIAL:
            return '\n\n'.join(list_of_runlines)
```

To this:

``` python
        if do_parallel_execution:
            list_of_runlines.append('wait\n')
            return ' &\n\n'.join(list_of_runlines)
        else:
            return '\n\n'.join(list_of_runlines)
```

It is a super minor change, but I was just wondering what was the advantage of having an extra construct (however simple it may be) to add a layer of abstraction what seems to be a simple boolean switch.

---

Actually, nevermind: just by looking into this I changed my mind. By grepping `CodeRunMode` I was able to quickly identify in which parts the `run_mode` was being set, modified, and/or applied. That alone is a pretty good advantage in my books, so now I like this better haha. Sorry for having gone on a "circular tangent" 😅  (however contradictory that might be).

unkcpz:
Thanks @ramirezfranciscof ! I open the issue https://github.com/aiidateam/aiida-core/issues/5034 for documentation of the `withmpi` of codeinfo. All other advice is accepted. I'll rebase and make a clear commit once it is approved.

tests/engine/test_calc_job/test_multi_codes_run_parallel_False_.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+exec > _scheduler-stdout.txt
+exec 2> _scheduler-stderr.txt
+
+
+'/bin/bash'
+
+'mpirun' '-np' '1' '/bin/bash'
```
ramirezfranciscof, line 8: Mmm this is strange no? I expected `content = content.rstrip() + '\n'` to add a final newline in the file but apparently it didn't, it just ended the current line. More surprisingly so: the linter didn't complain that there was no newline end?

I mean, if it is working we can leave it like this, it is just...unexpected.

aiida/schedulers/scheduler.py
```diff
@@ -216,7 +216,7 @@ def _get_run_line(self, codes_info, codes_run_mode):
 
             output_string = f'{command_to_exec} {stdin_str} {stdout_str} {stderr_str}'
 
-            list_of_runlines.append(output_string)
+            list_of_runlines.append(output_string.rstrip())
```
ramirezfranciscof, line 219: Hey @unkcpz why did you change this here? The `rstrip()` on the test when you are doing the `file_regression` should suffice for the test to work, no? I wouldn't change this in the actual code just to cater for a test or comply with some linter criteria in a script that is produced during usage and not part of the codebase. I mean, this is changing some intermediate file produce by the code that could be checked or parsed by some users, and its modification may create some problems (however unlikely it might be in this case).

I mean, unless you have a strong reason to propose we change to having our scripts be trimmed at the end, I would say to revert this change.

ramirezfranciscof (changes requested):
Hey @unkcpz its looking great! Thanks for the work, there are just two final details. One is the `.gitignore` modification I'm still trying to [figure out](https://github.com/aiidateam/aiida-core/pull/4990/files#r679203846) and the other is a modif on `aiida/schedulers/scheduler.py` that I don't quite understand (maybe it was done for testing and left unintentionally?). After this is dealt with I would consider this ready to merge.

.gitignore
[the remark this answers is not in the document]
ramirezfranciscof: Hey @sphuber , sorry for the insistence, I just want to make sure there is no already standardized way in which we are dealing with this issue before I propose my own.

To recap the problem: the test needs to run a `dry_run`; this generates a folder with files; this files are left polluting the environment (in this case, the modification aims at removing it from files detected by git). What do other tests that use `dry_run` do? They delete the files manually? Is there a `fixture` that takes care of deleting this folder at the end of the tests?

tests/engine/test_calc_job/test_multi_codes_run_parallel_False_.sh
sphuber: It is expected though. The `rstrip` strips _all_ whitespace, so also the newline character, which you then re-append. However, having a terminating newline character, does not show up visually as an empty line. It just means the line is terminated and the carriage return goes to the _start_ of the next line. If you want to see a visual empty line, you would need another new line character. Note that Github visualizes this difference (in diffs where this changes) with a symbol at the end of the file, saying "no newline character" or something to that effect.

ramirezfranciscof: Ah, yeah, for the `rstrip()` I did some tests in the shell and printing so I didn't know exactly how it would behave with write, there my "I expected..." meant "I intended that...". The unexpected part is more about the linter not complaining.

.gitignore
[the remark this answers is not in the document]
sphuber: There are two test files that run `CalcJob` dry runs: `tests/engine/test_launch.py` and `tests/engine/test_calc_job.py`. The former cleans up after itself, which is why you never saw the directories before. The `TestLaunchersDryRun` class deletes the `submit_test` folder in its `tearDownClass` method:

https://github.com/aiidateam/aiida-core/blob/a46851972d3bcdb9143801f8199e8c3d17ec645a/tests/engine/test_launch.py#L156-L169

For now, @unkcpz can maybe use this in the `test_calc_job.py` file as well. At some point we can find a cleaner fixture to do this that makes it easy to reuse, but since these test files are still using old-style, maybe we leave that for another time.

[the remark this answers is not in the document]
ramirezfranciscof: Thanks @sphuber !

Is it ok then to use some of these at templates to add a cleanup part to your test and remove this from `.gitignore` @unkcpz ?

aiida/schedulers/scheduler.py
unkcpz: This is intended to pass the linter. As you can see from the above line, that the `output_string` will end will spaces if `stdin_str`, `stdout_str` or `stderr_str` is not set. In that case, the command line of submit script will end with spaces wiped out by the linter and fails the pre-commit. 

> I mean, this is changing some intermediate file produce by the code that could be checked or parsed by some users, and its modification may create some problems (however unlikely it might be in this case

I agree, but for this part of the code, I see no downside to strip the line. In other words, I would much more like the bash script ironed without extra spaces in each line ;-)

.gitignore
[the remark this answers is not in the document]
unkcpz: Ah! I see your point! I can do add the fixture here and remove the line from `.gitignore`.

unkcpz:
Hi @ramirezfranciscof , rather than delete the folder, I add the following `dry_run_in_tmp` fixture to change the working dir to `/tmp` and running the test their.

```python
def dry_run_in_tmp(request):
    """change to the tmp directory to do the dry run, then change back to the calling directory to avoid side-effects"""
    os.chdir('/tmp')
    yield
    os.chdir(request.config.invocation_dir)
```

sphuber:
> Hi @ramirezfranciscof , rather than delete the folder, I add the following `dry_run_in_tmp` fixture to change the working dir to `/tmp` and running the test their.
> 
> ```python
> @pytest.fixture(scope='function')
> def dry_run_in_tmp(request):
>     """change to the tmp directory to do the dry run, then change back to the calling directory to avoid side-effects"""
>     os.chdir('/tmp')
>     yield
>     os.chdir(request.config.invocation_dir)
> ```

This looks fine, but I wouldn't hardcode the `'/tmp'` directory. You can use the `tmpdir` fixture to get a temporary directory that is cross-platform compatible and will automatically be cleaned up by `pytest`.

aiida/schedulers/scheduler.py
ramirezfranciscof: > This is intended to pass the linter. As you can see from the above line, that the `output_string` will end will spaces if `stdin_str`, `stdout_str` or `stderr_str` is not set. In that case, the command line of submit script will end with spaces wiped out by the linter and fails the pre-commit.

Right, but I mean, you are not committing any script that is being generated by this, right? The reference file is being created by `file_regression.check(content, extension='.sh')` and you are already using `rstrip` to clean up `content` before putting it there. That you originally take `content` from the printed script does not matter if you clean it between reading it and writing to reference with the `file_regression`. Am I missing something here? Which script directly generated by this is being rejected by the linter?

> I agree, but for this part of the code, I see no downside to strip the line. In other words, I would much more like the bash script ironed without extra spaces in each line ;-)

I also agree that it is very unlikely that this will break someone's code, but it is still possible. If we have had a team discussion on how to do a cleaner script and agreed in some guidelines that we were going to respect going forward, I would totally bite the bullet on the chance of causing some friction. But we are not even sure that we won't have to change this further (maybe to prevent contiguous empty lines due to non existing append/prepend texts) or even go back on it (maybe it turns out there are good reasons to have one final empty line at the end of the file). So yeah, the risk may be minimal but on the other hand there is no objective gain from this change (other than you and me liking it more aesthetically, but that is not very objective =P).

unkcpz: Sorry for not explain it more clear. 

I have to add this to pass the linter. Because the `content.rstrip()` in test function, only removes the final lines of the submit script content. `output_string` is the executable command line not necessary the final line. For clarity, please see the following content generated by in test and after linter pre-commit clean up (I use '␣' to represent space character):

The script generated:
```bash
#!/bin/bash
exec > _scheduler-stdout.txt
exec 2> _scheduler-stderr.txt


'/bin/bash'␣␣␣

'mpirun' '-np' '1' '/bin/bash'
```

After pre-commit:
```bash
#!/bin/bash
exec > _scheduler-stdout.txt
exec 2> _scheduler-stderr.txt


'/bin/bash'

'mpirun' '-np' '1' '/bin/bash'
```

If we agree not to add exclude files in the linter configuration, I can not figure out a simple way to get rid of this and pass the pre-commit check.

unkcpz:
> You can use the tmpdir fixture to get a temporary directory that is cross-platform compatible and will automatically be cleaned up by pytest.

@sphuber thanks! Changed.

aiida/engine/processes/calcjobs/calcjob.py
```diff
@@ -593,12 +593,9 @@ def presubmit(self, folder: Folder) -> CalcInfo:
...
             this_code = load_node(code_info.code_uuid, sub_classes=(Code,))
 
-            this_withmpi = code_info.withmpi  # to decide better how to set the default
-            if this_withmpi is None:
-                if len(calc_info.codes_info) > 1:
-                    raise PluginInternalError('For more than one code, it is necessary to set withmpi in codes_info')
-                else:
-                    this_withmpi = self.node.get_option('withmpi')
+            this_withmpi = self.node.get_option('withmpi')  # set code mpi, default set to as calcjob one
+            if code_info.withmpi:
```
sphuber, line 597: If I understand correctly, there are two ways to set `withmpi`: on the process itself through the `metadata.options.withmpi` input and through the `CodeInfo` that is created by the plugin itself, where the latter should be leading if both are defined.

If that analysis is correct, than I think this code is not correct. By checking just `if code_info.withmpi` the actual value will not be overridden even if it is set to `False`. However, this seems like a good use-case from a plugin perspective. If the plugin needs to run two codes, for which one _always_ needs to be run without mpi, i.e. `code_info.withmpi == False`, then that should override the value taken from `self.node.get_option('withmpi')`. That is to say, `False` is also an actual value that should not be ignored. So should the conditional not be `if code_info.withmpi is not None`?

To make this code a lot clearer, I would propose writing it as follows, with comment:
```suggestion

            # To determine whether this code should be run with MPI enabled, we get the value that was set in the inputs
            # of the entire process, which can then be overwritten by the value from the `CodeInfo`. This allows plugins
            # to force certain codes to run without MPI, even if the user wants to run all codes with MPI whenever
            # possible. This use case is typically useful for `CalcJob`s that consist of multiple codes where one or
            # multiple codes always have to be executed without MPI.

            withmpi = self.node.get_option('withmpi')

            # Override the value of `withmpi` with that of the `CodeInfo` if and only if it is set
            if code_info.withmpi is not None:
                withmpi = code_info.withmpi
```

aiida/schedulers/scheduler.py
sphuber: You are absolutely right @unkcpz the formatter strips all whitespace, also terminating spaces in lines in the middle of the file. Apologies for not having thought of this, I was focusing on the terminating new line characters.

OK, so here is my thought process. We have added the whitespace trimming pre-commit action such that trailing whitespace is automatically cleaned up in source code, as it can make things look messy and more difficult to read than necessary. When I added it, I did not specify what files it should run for, and so let it run for everything, even comparison files that are used in the unit tests, even though those are not the primary target. We don't need to read those often, so trailing newspace is not a real problem. When you consider the motivation for the whitespace trimming, it seems clear that the correct solution for this problem is to exclude the test comparison file from the pre-commit. So I now agree that @unkcpz original solution of updating the exclude rules in the `.pre-commit-config.yaml` file is actually the correct one.

I propose therefore that we revert the changes we requested (really sorry for the back and forth @unkcpz ), removing the manual explicit whitespace stripping in the code and the test and simply update the include rules in the config. Note that instead of excluding explicitly this file, maybe it would be better to simply _include_ all `.py` files, because that is in the end what we are really interested in. We want to clean the source code files, not any other arbitrary text files.

What we could do is use the following regex I think
```
        exclude: >- >
            (?x)^(
                tests/.*(?<!\.py)$
            )$
```

i.e. exclude all files in the `tests` folder that _do not_ finish with the `.py` extension. We still want to clean the test source files of course. Would you agree with this suggestion @ramirezfranciscof ?

sphuber (changes requested):
Thanks a lot @unkcpz and especially for your patience. I realize we have been going back and forth quite a bit on various things. I have now fully reviewed the code (which I hadn't done yet, just replied to question and comments) and I have two comments remaining. Once those are addressed this is good to be merged for me.

aiida/schedulers/scheduler.py
ramirezfranciscof: @unkcpz Ahhhh I see now! Thanks for the explanation, haha, indeed I can see the excess with the newlines but it is more difficult to notice the trailing spaces 😅. Nice idea using the strange underscore symbol to illustrate it.

@sphuber mmm I'm trying to think if this could backfire. I mean, the test section does have other kinds of file (the `test/static` folders have some `.cif`, `.UPF`, `.aiida` exports, `.json`) but perhaps we don't want to be lintering those anyways? I also had the idea that there were some notebooks being used to do some tests, but I couldn't find anything there now. I'm wondering if at some point it won't be convenient to use some bash scripts as aux to some tests and we may want to linter those too. It is also worth mentioning that the `.sh` extension was set explicitly by `file_regression.check(content, extension='.sh')`, so we could also decide on a specific extension for reference files and exclude those (`.pyreg`? `.pytest_regression_uniquely_long_aiida_extension`?). I mean, I agree it would be nice to set some configuration to deal now and for all with all pytest regression files, I'm just not sure just applying the linter to `.py` is the clear way to do this.

There is also one last possible chance that we might be able to deal with this in a simple way inside the test (which is, actually, more self-documenting that the initial one I proposed...):

``` python
    content = '# pylint: skip-file\n' + content
```

In any case, as the proposed alternative would be a modification internal to the tests so we are more free to set it back if we change our minds, I'm ok with you going the way you prefer. It still pains me to have to add the exception to the `.pre-commit-config.yaml`, but since you already tried a suggestion I made that underestimated the problem I won't insist on this if you think the pylint config is the best way to go.

sphuber: > There is also one last possible chance that we might be able to deal with this in a simple way inside the test (which is, actually, more self-documenting that the initial one I proposed...):
> 
> ```python
>     content = '# pylint: skip-file\n' + content
> ```

It is not `pylint` that does the trimming though, so that won't work.

ramirezfranciscof: Mmm I see. I can't find anywhere how to signal inside the file that no hooks should be applied to it, so perhaps this is not possible in a general way. Weird. Anyways, after two unsuccessful suggestions I concede, feel free to go ahead with @sphuber suggested solution then @unkcpz =)

aiida/engine/processes/calcjobs/calcjob.py
ramirezfranciscof: Uhm, I totally missed this on my pass, my bad 😕.

Follow up question: we want `code_info.withmpi` to **always** override the calcjob setting? Or if the user has explicitly set `withmpi` to be `False` then all codes executed should actually respect that? (in which case setting `code_info.withmpi = True` or leaving `code_info.withmpi = None` would have the same effect) Or the idea is that if a plugin sets `code_info.withmpi = True` it means that the code can only be ran with mpi?

Maybe we can make a table of expected results and intent and check these in a test?

|  | [Calcjob] True | [Calcjob] False |
| :---  | :---:  | :---:  |
| [Code] True (the plugin/code can only be run with mpi?) | Run with MPI | Incompatible? Raise? |
| [Code] False (this code does not use parallelism at all)| Run without MPI | Run without MPI |
| [Code] None (this code can run with or without MPI) | Run with MPI | Run without MPI |

sphuber: > Or if the user has explicitly set withmpi to be False then all codes executed should actually respect that?

In the example that I sketched, this would be problematic. Imagine a `CalcJob` that executes two codes: a small pre-processing script that _has_ to be executed serially or it will crash, and then a normal executable (like `pw.x`) that can be executed serially or with MPI depending on how it is compiled. In this case, the `metadata.options.withmpi` is the tool for the user to toggle the behavior for the latter, but it should _not_ override the setting for the pre-processing script because it will fail.

I am not sure if this solution covers all use cases or whether there are exceptions where the user definitely needs to be able to override. However, if the plugin explicitly sets `CodeInfo.withmpi` than there must be a good reason. If it can be run with either setting, maybe the plugin should not set one value either way. Or we should change `metadata.options.withmpi` to be able to take a list, but that may become even more problematic.

unkcpz: @ramirezfranciscof @sphuber thanks for making this more clear. This is an important catch! It is the right code that I intended to implement. 

Only if plugin developer sets `CodeInfo.withmpi` explicitly the `metadata.option.withmpi` will not override it through running `CalcJob`. The ultimate behaviour will be the case shown in the table above. Indeed that does not include the scenario that the plugin user what to override what was set in CodeInfo. I have no strong opinion on whether we should take this into account. 
But I want to mention that maybe we can consider this problem together with [setting default calcjob when configuring code #4991](https://github.com/aiidateam/aiida-core/issues/4991). Plugin user can only set one default CalcJob plugin for the code and this is the code that is implicitly called with `self.inputs.code` in `CalcJob`.  I would assume the plugin user only what to control the MPI behaviour of this particular code. And the other codes need to be explicitly passed as an InputPort in defining the CalcJob, which give plugin user an option to use the correct compiled code(MPI one or serial one) if the `withmpi` of this codeinfo is fixed by the plugin developer.  
Therefore, I prefer to have a detailed comment as shown by @sphuber and add information in documentation to warn the plugin developer to carefully set this `codeinfo.withmpi` in their plugin when there is more than one code.

unkcpz:
@sphuber no worries about back and forth, I was only trying to get rid of the pre-commit error in the previous commit. Now we know what causes the pre-commit failure.  
There is an issue here, I can either exclude files in a specific hook or in top-level for all repos. There is no option to exclude files 'repo' range https://pre-commit.com/#pre-commit-configyaml---repos .

I think we need to add it in top-level, correct?

sphuber:
> @sphuber no worries about back and forth, I was only trying to get rid of the pre-commit error in the previous commit. Now we know what causes the pre-commit failure.
> There is an issue here, I can either exclude files in a specific hook or in top-level for all repos. There is no option to exclude files 'repo' range https://pre-commit.com/#pre-commit-configyaml---repos .
> 
> I think we need to add it in top-level, correct? Otherwise exclude files in two hooks seems nasty.
> There is an issue here, I can either exclude files in a specific hook or in top-level for all repos.

I am pretty sure you can apply the exclude for all pre-commit hooks and not apply them to all repos, like so:

```yaml
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v2.5.0
    hooks:
    -   id: double-quote-string-fixer
    -   id: end-of-file-fixer
    -   id: fix-encoding-pragma
    -   id: mixed-line-ending
    -   id: trailing-whitespace
        exclude: >-
            (?x)^(
                tests/.*(?<!\.py)$
            )$
```

unkcpz:
Hmmm.... no, it is not true. The exclude key only take effect for the hook. With your configuration above, I got the following files modification:

```
(aiida-core-dev) ➜  aiida_core git:(code-run-mode) ✗ pre-commit run --files tests/engine/test_calc_job/test_multi_codes_run_parallel_True_.sh
Fix double quoted strings............................(no files to check)Skipped
Fix End of Files.........................................................Failed
- hook id: end-of-file-fixer
- exit code: 1
- files were modified by this hook

Fixing tests/engine/test_calc_job/test_multi_codes_run_parallel_True_.sh
...
```

Have to configure pre-commit like:
```yaml
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v2.5.0
    hooks:
    -   id: double-quote-string-fixer
    -   id: end-of-file-fixer
        exclude: >-
            (?x)^(
                tests/.*(?<!\.py)$
            )$
    -   id: fix-encoding-pragma
    -   id: mixed-line-ending
    -   id: trailing-whitespace
        exclude: >-
            (?x)^(
                tests/.*(?<!\.py)$
            )$
```

aiida/engine/processes/calcjobs/calcjob.py
ramirezfranciscof: > In the example that I sketched, this would be problematic. Imagine a `CalcJob` that executes two codes: a small pre-processing script that _has_ to be executed serially or it will crash, and then a normal executable (like `pw.x`) that can be executed serially or with MPI depending on how it is compiled. In this case, the `metadata.options.withmpi` is the tool for the user to toggle the behavior for the latter, but it should _not_ override the setting for the pre-processing script because it will fail.

Yes, so I think you are describing what should happen in the table I included when `[Calcjob] True` and `[Code] False`. Here it makes sense that the `CodeInfo.withmpi` overrides the `metadata.options.withmpi` because of the reasons you already explained: even if the user is indicating it has MPI available for the main code called by the calculation, other auxiliary calls may not benefit from or not even be able to run in parallel.

My question was more what should happen in the reciprocal case, when `[Calcjob] False` and `[Code] True`. Here the user is specifically saying to NOT use MPI at all, is it still justified to override this and anyways use MPI on the `CodeInfo.withmpi = True` code? Or does this suggest a more fundamental incompatibility and thus should raise?

sphuber:
> Have to configure pre-commit like:
> 

Ah sure, we need it to apply to multiple hooks, but you can define the list once and use a reference like so

```yaml
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v2.5.0
    hooks:
    -   id: double-quote-string-fixer
    -   id: end-of-file-fixer
        exclude: &exclude_files >
            (?x)^(
                tests/.*(?<!\.py)$
            )$
    -   id: fix-encoding-pragma
    -   id: mixed-line-ending
    -   id: trailing-whitespace
        exclude: *exclude_files
```

aiida/engine/processes/calcjobs/calcjob.py
unkcpz: My apologies @ramirezfranciscof, I gave it a glimpse and take for granted with my own mind. 
I reconsider your advice to raise here and start to like it. Maybe it can be a raising warning rather than an exception? But I still assume the developer of the plugin should be careful and know what they do when setting `codeinfo.withmpi`. 
Do you have a strong opinion on raising an exception or maybe a warning?

unkcpz:
thanks! That makes sense.

aiida/engine/processes/calcjobs/calcjob.py
ramirezfranciscof: My preference is typically on the versatility on the end user and clarity when we are overriding his intentions. So here I would say:

- I would issue a warning if then we let `metadata.option.withmpi` override `codeinfo.withmpi` (we let the user know the plugin designer set this to only be run with MPI for some reason but we allow them to proceed as they see fit)
- In the other case, i.e. if after the warning `codeinfo.withmpi` would override `metadata.option.withmpi`, then I would say it is preferable to raise. This setting is not being used for anything else as far as I know, if the user is setting it, it is specifically for this and we are overriding it. They wouldn't be able to do what they want anyways, so just raise saying "you are not allowed to set this".

Now as to which of those two I prefer, I'm not sure. I'm still a bit puzzled as to in what situation the plugin designer would set something to ONLY be run with MPI, it is a bit bizarre to me. I'll let you decide what is best for this (or @sphuber if he has another opinion).

---

NOTE:  I am now re-reading the previous logic to check all the cases of the table. The previous behavior was that it would first check the individual `codeinfo.withmpi` and only if it was `None` *and* if it was a single code, then it would check `metadata.option.withmpi`. So if there were multiple codes being called, then the plugin designer had to set whether they were run with MPI or not, and the user had no say in this. So before there was actually no way for the plugin designer to leave how to run some of the codes to "the criteria" of the user.

This gets more bizarre the more we look into it...🤔

sphuber:
> thanks @sphuber ! That makes sense. However, the `exclude_files` is already be used in the following. What do you think I name it `exclude_nonpy_tests_files` for the case here?

Maybe just name it after the pre-commit step, i.e., `exclude_pre_commit_hooks`

aiida/engine/processes/calcjobs/calcjob.py
unkcpz: All the bizarre things come from there is only one binary flag that can be set from `metadata.option.withmpi` by plugin user when actually a list of flags need for multiple codes case, which is hard to settle at this PR.

I would like to suggest that we stop here and leave all complicity to the documentation. For plugins that involve more than one code, if the user wants to get full control in MPI mode of how to run codes, they are suggested to split the calcjob into parts with each one running with one code only.

unkcpz:
Hi @ramirezfranciscof, I add a test to check the `withmpi` behaviours and rebase the commits. Maybe we can get it merged at it is now, and leave the discussion when the practical issue happened and report?

sphuber (approved):
Thanks a lot @unkcpz . I am approving this to be merged. I agree that any further complications are indeed due to mismatch in the process inputs and the `CodeInfo` definition. I think we should simply open an issue that aims to improve the documentation on `CodeInfo` which includes a discussion on the `withmpi` option.

ramirezfranciscof (approved):

ramirezfranciscof:
Yes, I'm ok with merging, we have had substantial discussion as it is and I don't want to drag the PR much longer (I hope it wasn't too tedious for you =P).

However I do want to make a final comment that I don't think this is just a matter of documentation: I feel like the current behavior may be limiting in the sense that it may be putting control on how to execute code on the wrong actor (i.e. if MPI is an option with which codes can or not be compiled, it is the user who knows how the code is compiled and I'm not sure the plugin designer can make the decision to force that at code-design time).

PR #4994, opened 2021-06-19
Use sanitised variable name in SGE scheduler job title

mjclarke94:
The  SGE scheduler creates a sanitised `job_title` variable, but doesn't actually use it.

### DIFF
--- a/aiida/engine/processes/builder.py
+++ b/aiida/engine/processes/builder.py
@@ -41,7 +41,7 @@ class ProcessBuilderNamespace(collections.abc.MutableMapping):
         self._valid_fields = []
         self._data = {}
 
-        # The name and port objects have to be passed to the defined functions as defaults for
+        # The name and port objects have to be passed to the defined functions as defaults for
         # their arguments, because this way the content at the time of defining the method is
         # saved. If they are used directly in the body, it will try to capture the value from
         # its enclosing scope at the time of being called.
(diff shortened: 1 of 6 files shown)


sphuber (commented):
Thanks @mjclarke94 . That indeed seems unintended, however, I also wonder whether this entire cleaning block is even needed? The block seems to have been copied verbatim from the `PbsBaseClass` scheduler implementation for PBS based schedulers (as indicated by the reference to `qsub` in the comment). Is this cleaning even required for SGE? Does it have the same limitations on the job title as PBS has?

mjclarke94:
Yeah, seems to be the same restrictions which explains the code reuse.
From the man page of `qsub` on an HPC using SGE:
```
 -N name
              Available for qsub, qsh, qrsh, qlogin and qalter only.

              The name of the job. The name should follow the "name" definition in sge_types(5).  Invalid job names will be denied at submit time.
```
`man sge_types`
```
name
       A name is an arbitrary string of ASCII printing characters, but may not contain  "/", ":", "@", "\", "*",  or "?".
```

sphuber:
Thanks for the update @mjclarke94 . Then it seems all good to merge. It would just be good to add a test. I quickly wrote one that you can add at the end of the file `tests/schedulers/test_sge.py`
```python
def test_job_name_cleaning(self):
    """Test that invalid characters are cleaned from job name."""
    from aiida.schedulers.datastructures import JobTemplate

    scheduler = SgeScheduler()

    job_tmpl = JobTemplate()
    job_tmpl.job_resource = scheduler.create_job_resource(parallel_env='mpi8', tot_num_mpiprocs=16)
    job_tmpl.job_name = 'Some/job:name@with*invalid-characters.'

    header = scheduler._get_submit_script_header(job_tmpl)
    self.assertTrue('#$ -N Somejobnamewithinvalid-characters.' in header, header)
```
Feel free to modify it of course if you think you can improve it. If you add this, we can run the tests and merge

mjclarke94:
Looks good to me! Test added 🙂

mjclarke94:
Oop. Caught by the pre-commit. Struggling to get it to install locally for some reason but I think that change should sort it.

sphuber (approved):
Thanks @mjclarke94

PR #4996, opened 2021-06-25
Pre-commit: move `yapf` and `mypy` config to `pyproject.toml`

sphuber:
The libraries `yapf` and `mypy` support the `pyproject.toml` file for
their configuration starting from v0.31 and v0.9, respectively.


sphuber:
@chrisjsewell to have support for `pyproject.toml` required updating `mypy` which incurred a bunch of new errors in the pre-commit. Don't have time to fix this now, but thought to already create the PR

unkcpz:
Hi @sphuber, since I am here to look at mypy errors, I can try to assistant to complete this PR. Do you already start working on it, or can I directly add more commits into your branch?

sphuber:
@unkcpz I have updated the branch with all that I have currently. It would be great if you could address the `mypy` errors. Feel free to add commits straight to my branch. Let me know if you don't have rights to do so. Thanks!

unkcpz:
Hi @sphuber, I make some commits to settle the mypy errors.

Except for the following types of mypy complaints are not resolved:

###   `error: Module has no attribute "Node"  [attr-defined]` 

And lots of similar no attribute errors which I guess are related to the mypy bug as post at https://github.com/python/mypy/issues/10826. mypy was not able to deduce the correct module from `__all__`.

###  incompatible with supertype errors

Similar errors of `repository_metadata`, `_updatable_attributes` and `_hash_ignored_attributes` from many files.The signatures are the attributes of parent class and are being overridden as properties in child class. Referring to  https://github.com/python/mypy/issues/4125 for more details.

aiida/manage/configuration/profile.py
```diff
@@ -130,12 +131,18 @@ def __init__(self, name, attributes, from_config=False):
 
     def get_repository(self) -> 'Repository':
         """Return the repository configured for this profile."""
-        from disk_objectstore import Container
         from aiida.repository import Repository
+
+        backend = self.get_backend()
+        return Repository(backend=backend)
+
+    def get_backend(self) -> 'DiskObjectStoreRepositoryBackend':
```
sphuber, line 139: I wouldn't call this `get_backend`. It would be too confusing with the database backend. If we really need this method, a better name would be `get_repository_backend`. However, why can't we just call `get_repository().backend`? Is it necessary to have this intermediate method?

aiida/manage/manager.py
```diff
@@ -163,6 +163,9 @@ def _load_backend(self, schema_check: bool = True, repository_check: bool = True
         # are taken into account and the `DbLogHandler` is configured.
         configure_logging(with_orm=True)
 
+        if self._backend is None:
+            raise ValueError('Could not load the backend.')
```
sphuber, line 167: Would be better to put this check directly after the logic that is supposed to set it, i.e. before the `configure_logging` call. Second, I don't think a `ValueError` is the correct exception for this. That is reserved for validation arguments of a method or function. What seems to be happening in this case, the only reason `self._backend` can be `None` is because the `backend_type` is neither `BACKEND_DJANGO` nor `BACKEND_SQLA`. So really we shouldn't have this check but add the `else` clause and raise an `AssertionError` because our internal logic is supposed to set these when loading a profile and that is what we are checking here: internal consistency of program logic.

aiida/manage/manager.py
```diff
@@ -206,7 +209,7 @@ def get_backend(self) -> 'Backend':
 
         """
         if self._backend is None:
-            self._load_backend()
+            return self._load_backend()
```
sphuber, line 212: Why is this change necessary?

aiida/tools/importexport/dbimport/backends/django.py
```diff
@@ -38,7 +38,7 @@
 
 def import_data_dj(
     in_path: str,
-    group: Optional[Group] = None,
+    group: Optional[ImportGroup] = None,
```
sphuber: I reverted this change as it is backwards incompatible. We are currently allowing users to use a plain `Group` to store nodes in upon importing. We cannot narrow this restriction down.

aiida/tools/importexport/dbimport/backends/common.py
```diff
@@ -66,7 +66,7 @@ def collect_hashkeys(objects, hashkeys):
         container_export.export(set(hashkeys), container_profile, compress=True, callback=callback)
 
 
-def _make_import_group(*, group: Optional[ImportGroup], node_pks: List[int]) -> ImportGroup:
+def _make_import_group(*, group: Optional[ImportGroup], node_pks: List[int]) -> Optional[ImportGroup]:
```
sphuber: To fix the mypy error related to inconsistent types, we should rather change this to `Optional[Group]`. I have already done this and reverted your other change. Reason why is explained in another comment

sphuber (commented):
Thanks @unkcpz . Reverted some changes to the `Group` logic and have some other questions.

aiida/manage/manager.py
unkcpz: Good point!

unkcpz (commented):
@sphuber thanks for review!

> I reverted this change as it is backwards incompatible. We are currently allowing users to use a plain Group to store nodes in upon importing. We cannot narrow this restriction down.

Yes. After changing that I find it failed some migrate tests.

aiida/manage/manager.py
unkcpz: Okay, I guess this is the place where the type assertion properly takes effect. We have:
```python
return cast('Backend', self._backend)
```

In that case, I propose `_load_backend` method should have no return value.

aiida/manage/configuration/profile.py
```diff
@@ -130,12 +131,18 @@ def __init__(self, name, attributes, from_config=False):
 
     def get_repository(self) -> 'Repository':
         """Return the repository configured for this profile."""
-        from disk_objectstore import Container
         from aiida.repository import Repository
+
+        backend = self.get_backend()
```
unkcpz, line 136: Looking at failed CI, found here is a mistake. In case of disturbing you with too many push to your branch, I will fix this later with more valid commits.

```suggestion
        backend = self.get_repository_backend()
```

chrisjsewell:
> And lots of similar no attribute errors which I guess are related to the mypy bug as post at python/mypy#10826. mypy was not able to deduce the correct module from `__all__`.

Well in fairness to mypy, these `__all__` + `import *` constructs are basically a bit of a nightmare for static analysers.

We have to use `pylint: disable=wildcard-import` in all these files, which would suggest it is maybe not the best practice, and it also makes it quite difficult to see what APIs are being exposed at each level (#4511)

Additionally, it currently does not work with new default VS Code Python Language Server (Pylance) and so you get no auto-completions, etc.
Although here, following the issues: https://github.com/microsoft/pylance-release/issues/289 -> https://github.com/microsoft/pyright/pull/1877 -> https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface, there is a work-around; to restructure the `__all__` construction like:

```python
__all__ = ()
__all__ += calculation.__all__
__all__ += process.__all__ 
__all__ += workflow.__all__
```

this does not help with mypy though

sphuber:
Ok, well if this is a known problem of `mypy`, is there not some way to ignore these errors in a simple manner, i.e., without having to add inline disable comments on each respective line?

chrisjsewell:
#5061 updates mypy to v0.901 😄

chrisjsewell:
right lets see where we are at with this

chrisjsewell:
Ok well the `Module has no attribute` errors are now gone, but yeh for the rest; I would suggest the code changes should go in a separate PR and this one should just be about moving the configuration

sphuber:
> I would suggest the code changes should go in a separate PR

What code changes? As far as I know, any code changes applied were simply to address the new errors that were now thrown by the updated version of `mypy`, so they would be necessary. And there are still a lot of errors left over.

aiida/manage/configuration/profile.py
```diff
@@ -130,12 +131,16 @@ def __init__(self, name, attributes, from_config=False):
 
     def get_repository(self) -> 'Repository':
         """Return the repository configured for this profile."""
-        from disk_objectstore import Container
         from aiida.repository import Repository
+        backend = self.get_repository_backend()
+        return Repository(backend=backend)
+
+    def get_repository_backend(self) -> 'DiskObjectStoreRepositoryBackend':
```
chrisjsewell, line 138: > what code changes

well you are introducing completely new methods

chrisjsewell:
> moving the configuration should result in zero changes to the mypy outcome

😅 #5066

PR #4999, opened 2021-06-28
👌 IMPROVE: Make `verdi group` messages consistent

### LINKED ISSUE #4998: Docs/verdi: Messages from handling groups are inconsistent in `verdi`

When going through the tutorial section on "Organising your data", it seemed to me that the "Success", "Warning", etc. messages from doing `verdi group` commands were inconsistent in naming and capitalization.

Be consistent.
When referring to a node type, capitalize.
When referring to `label`, don't use `name`.

CasperWA:
Closes #4998.

- Use `label` instead of `name`.
- Use `group.__class__.__name__` always.
- Minor syntax fixes.

**Edit**: This PR has been updated according to the discussion below.

### DIFF
--- a/aiida/cmdline/commands/cmd_group.py
+++ b/aiida/cmdline/commands/cmd_group.py
@@ -32,9 +32,7 @@ def verdi_group():
 def group_add_nodes(group, force, nodes):
     """Add nodes to a group."""
     if not force:
-        click.confirm(
-            f'Do you really want to add {len(nodes)} nodes to {group.__class__.__name__}<{group.label}>?', abort=True
-        )
+        click.confirm(f'Do you really want to add {len(nodes)} nodes to {group}?', abort=True)
 
     group.add_nodes(nodes)
 
@@ -48,9 +46,6 @@ def group_add_nodes(group, force, nodes):
 def group_remove_nodes(group, nodes, clear, force):
     """Remove nodes from a group."""
     from aiida.orm import QueryBuilder, Group, Node
-
-    label = group.label
-    klass = group.__class__.__name__
 
     if nodes and clear:
         echo.echo_critical(
@@ -69,18 +64,18 @@ def group_remove_nodes(group, nodes, clear, force):
             group_node_pks = query.all(flat=True)
 
             if not group_node_pks:
-                echo.echo_critical(f'None of the specified nodes are in {klass}<{label}>.')
+                echo.echo_critical(f'None of the specified nodes are in {group}.')
 
             if len(node_pks) > len(group_node_pks):
                 node_pks = set(node_pks).difference(set(group_node_pks))
-                echo.echo_warning(f'{len(node_pks)} nodes with PK {node_pks} are not in {klass}<{label}>.')
+                echo.echo_warning(f'{len(node_pks)} nodes with PK {node_pks} are not in {group}.')
 
-            message = f'Are you sure you want to remove {len(group_node_pks)} nodes from {klass}<{label}>?'
+            message = f'Are you sure you want to remove {len(group_node_pks)} nodes from {group}?'
 
         elif clear:
-            message = f'Are you sure you want to remove ALL the nodes from {klass}<{label}>?'
+            message = f'Are you sure you want to remove ALL the nodes from {group}?'
         else:
-            echo.echo_critical(f'No nodes were provided for removal from {klass}<{label}>.')
+            echo.echo_critical(f'No nodes were provided for removal from {group}.')
 
         click.confirm(message, abort=True)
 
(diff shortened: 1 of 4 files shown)


sphuber (commented):
I am always a fan of consistency and I also like:
```
{group.__class__.__name__}<{group.label}>
```
as the standard representation of a `Group` instance. But to prevent us having to repeat this lengthy snippet each time, what about we make this the implementation of `Group.__str__`? Then we just put `{group}` in the string and presto. The current `__str__` implementation includes the user name but I think that is used so little that we can remove it.

CasperWA:
> (...) But to prevent us having to repeat this lengthy snippet each time, what about we make this the implementation of `Group.__str__`? Then we just put `{group}` in the string and presto. The current `__str__` implementation includes the user name but I think that is used so little that we can remove it.

Fine by me - if there are no objections to this? @giovannipizzi @ltalirz @chrisjsewell @mbercx ? (I will add a commit with this change as it's quick, which can then be reverted if there's a huge uprising of sorts.)

sphuber:
> **Edit**: Looking at the code I have just realized you are mixing `__repr__` and `__str__`.

Wasn't really mixing them up, but was just describing how I think the `__str__` should be implemented. The `__repr__` should also be improved. The general guideline for these two methods (which we aren't yet widely respecting but we started changing to) is that `__str__` should return a user readable string and `__repr__` should provide the necessary information for a developer to reliably identify the object. So I would actually propose the following for both methods:
```python
def __repr__(self):
    return f'self.__class__.__name__<self.uuid>'

def __str__(self):
    return f'self.__class__.__name__<self.label>'
```
I forgot where, but we started using this definition in other places since some time ago. We just never got around to apply this to all important ORM classes.

CasperWA:
> (...) The general guideline for these two methods (which we aren't yet widely respecting but we started changing to) is that `__str__` should return a user readable string and `__repr__` should provide the necessary information for a developer to reliably identify the object. (...)

I don't completely agree with this, at least not the `__repr__` part.
I believe the true return value for `__repr__` should always be passable to the built-in `eval()` function to instantiate an object with similar attributes and make up as the instance that is represented.

> (...) So I would actually propose the following for both methods:
> 
> ```python
> def __repr__(self):
>     return f'self.__class__.__name__<self.uuid>'
> 
> def __str__(self):
>     return f'self.__class__.__name__<self.label>'
> ```

This would then instead become:

```python
def __repr__(self):
    return f'self.__class__.__name__(label=self.label, user=self.user, description=self.description, type_string=self.type_string, backend=self.backend)'

def __str__(self):
    return f'self.__class__.__name__<self.label>'
```

I'm not sure whether all the attributes here match up exactly with the initialization parameters --- but, yeah.

> I forgot where, but we started using this definition in other places since some time ago. We just never got around to apply this to all important ORM classes.

Might be good to make a collective issue for this then? Low priority but still important.

sphuber:
> I believe the true return value for `__repr__` should always be passable to the built-in `eval()` function to instantiate an object with similar attributes and make up as the instance that is represented.

I know that this is what is often written on this topic, but I don't think this is useful or necessary in practice. Note that even if you get the full signature correct (and even you yourself already admit here that you are not sure your current implementation is correct) there is no guarantee that you know where the class was imported from and so you cannot necessarily reconstruct the object using `eval`. This whole guideline was once conceived to express the goal that the `repr` should be as unambiguous as possible as to what object it represents. It was never intended that it should be possible to reconstruct the object. This is why I think in our use case (for ORM objects at least) we should simply use the UUID as these objects always have one and it will always be guaranteed to be unambiguous.

CasperWA:
> > I believe the true return value for `__repr__` should always be passable to the built-in `eval()` function to instantiate an object with similar attributes and make up as the instance that is represented.
> 
> I know that this is what is often written on this topic, but I don't think this is useful or necessary in practice. Note that even if you get the full signature correct (and even you yourself already admit here that you are not sure your current implementation is correct) there is no guarantee that you know where the class was imported from and so you cannot necessarily reconstruct the object using `eval`. This whole guideline was once conceived to express the goal that the `repr` should be as unambiguous as possible as to what object it represents. It was never intended that it should be possible to reconstruct the object. This is why I think in our use case (for ORM objects at least) we should simply use the UUID as these objects always have one and it will always be guaranteed to be unambiguous.

Okay. I think this is something that should be decided and implemented for all ORM classes in one instead of changing it here. So I'll open an issue for this (if one doesn't already exist) and revert changes to `__repr__`, keeping the current change to `__str__`.

chrisjsewell (approved):
Discussion of the actua group __str__ contruct has moved to https://github.com/aiidateam/aiida-core/issues/3087, so this is good to go thanks 👍

PR #5001, opened 2021-06-30
Docs: fix all links of plugin_develop

unkcpz:

PR #5003, opened 2021-06-30
Docs: fixing links and linkcheck on Github action

unkcpz:
hi @chrisjsewell, the modifications of https://github.com/aiidateam/aiida-core/pull/5001 are included here with linkcheck.

### DIFF
--- a/aiida/tools/data/orbital/realhydrogen.py
+++ b/aiida/tools/data/orbital/realhydrogen.py
@@ -258,7 +258,7 @@ class RealhydrogenOrbital(Orbital):
     OrbitalData class.
 
     Following the notation of table 3.1, 3.2 of Wannier90 user guide
-    http://www.wannier.org/doc/user_guide.pdf
+    (which can be downloaded from http://www.wannier.org/support/)
     A brief description of what is meant by each of these labels:
 
     :param radial_nodes: the number of radial nodes (or inflections) if no
--- a/docs/source/howto/plugins_develop.rst
+++ b/docs/source/howto/plugins_develop.rst
@@ -18,22 +18,22 @@ Creating a plugin package
 =========================
 
 
-AiiDA plugins can be bundled and distributed in a `Python package <packages>`_ that provides a set of extensions to AiiDA.
+AiiDA plugins can be bundled and distributed in a `Python package <packages_>`_ that provides a set of extensions to AiiDA.
 
 .. note::
 
   The Python community uses the term 'package' rather loosely.
-  Depending on context, it may refer simply to a folder containing individual Python modules or it may include the files necessary for building and installing a package to be distributed via the `Python Package Index (PyPI) <pypi>`_.
+  Depending on context, it may refer simply to a folder containing individual Python modules or it may include the files necessary for building and installing a package to be distributed via the `Python Package Index (PyPI) <pypi_>`_.
 
-.. _packages: https://docs.python.org/2/tutorial/modules.html?highlight=package#packages
+.. _packages: https://docs.python.org/3/tutorial/modules.html#packages
 
 
 Quickstart
 ----------
 
-The fastest way to jumpstart an AiiDA plugin package is to use the `AiiDA plugin cutter <plugin-cutter>`_ in order to template the basic folder structure, already customized according to the desired name of your plugin, following AiiDA conventions.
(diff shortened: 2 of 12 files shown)


unkcpz:
Here is a problem with this github action, when the action failed, it doesn't give verbose information at last, so it is hard to find from the action page which link is broken. Have to either view the file from the GitHub page or build locally and search for the keyword 'broken'. 
I currently put the action in test ci for every push and pull request, I believe it might be better to only test this in the new release phase.

ltalirz:
Thanks a lot for this @unkcpz !

Since this turned up quite a number of broken links in the howto section (notifying also @sudarshanv01 ), I think we should get at least the fixes merged today.

> it doesn't give verbose information at last, so it is hard to find from the action page which link is broken. Have to either view the file from the GitHub page or build locally and search for the keyword 'broken'.

This can be done on github actions I believe, i.e. store the output of the action and in the next step fail if `broken` was detected.


> I currently put the action in test ci for every push and pull request, I believe it might be better to only test this in the new release phase.

I guess it would only be needed if a PR changes the docs?
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths
If we limit it to that I think we can run it always

I'll have a quick go at this.

ltalirz:
Or @unkcpz if you have the time to have a look that would be great as well - I will come back to this in 2h (need to work on something else first)

chrisjsewell:
> I guess it would only be needed if a PR changes the docs

For that it would need to go in a separate workflow file, e.g. for .github/workflows/ci-code.yml it is skipped if there are only document changes.

Note obviously this would not pick up any broken links added in docstrings, for a non-doc PR

unkcpz:
Thanks! @ltalirz @chrisjsewell 
The action is put in a separate file now and the outputs are grepped to give the results. 

It's true in this case the changes in docstrings will not trigger the action in the pull request, but I believe this is rare and can be left on `push` phase.

aiida/tools/data/orbital/realhydrogen.py
```diff
@@ -258,7 +258,7 @@ class RealhydrogenOrbital(Orbital):
     OrbitalData class.
 
     Following the notation of table 3.1, 3.2 of Wannier90 user guide
-    http://www.wannier.org/doc/user_guide.pdf
+    (which can be download from http://www.wannier.org/support/)
```
ltalirz, line 261:
```suggestion
    (which can be downloaded from http://www.wannier.org/support/)
```

docs/source/howto/plugins_develop.rst
```diff
@@ -18,22 +18,22 @@ Creating a plugin package
...
+AiiDA plugins can be bundled and distributed in a `Python package <packages_>`_ that provides a set of extensions to AiiDA.
 
 .. note::
 
   The Python community uses the term 'package' rather loosely.
-  Depending on context, it may refer simply to a folder containing individual Python modules or it may include the files necessary for building and installing a package to be distributed via the `Python Package Index (PyPI) <pypi>`_.
+  Depending on context, it may refer simply to a folder containing individual Python modules or it may include the files necessary for building and installing a package to be distributed via the `Python Package Index (PyPI) <pypi_>`_.
 
-.. _packages: https://docs.python.org/2/tutorial/modules.html?highlight=package#packages
+.. _packages: https://docs.python.org/3/tutorial/modules.html?highlight=package#packages
```
ltalirz, line 28:
```suggestion
.. _packages: https://docs.python.org/3/tutorial/modules.html#packages
```

ltalirz (changes requested):
thanks a lot @unkcpz for this, this come just at the right time :-)

I have just two minor change requests, then I suggest we get this merged.

Please let me know in the next 15 minutes whether you're going to work on finishing this - otherwise I can take over (I know it' s quite late where you are by now ;-) )

ltalirz:
@unkcpz about the broken link to aiida-diff - don' get crazy.
You can either add it to the ignore list or remove it for being too specific

I suspect that github is doing a special processing of the "anchor"  value that is not what the sphinx checker expects

ltalirz:
As mentioned above, since it's late where you are, let me know if you're going to finish this

unkcpz:
Hi @ltalirz, I have to go now, please make changes to finish this, if necessary.

ltalirz (approved):

ltalirz:
mentioning @csadorf for info (this is not about dependencies but about the github workflow setup)

I've merged this now since it contains important fixes for the tutorial.
If any modifications should be made to setup of the github workflow that can be done in a follow-up PR

PR #5011, opened 2021-07-06
ORM: fix deprecation warning always being shown in link managers

### LINKED ISSUE #5010: Tab-completion for the `outputs` attribute of a `Node` issues a warning
While trying to add a note on tab-completion for the `outputs` for the tutorial, I ran into the following deprecation warning:

```python
In [1]: cj_node = load_node(311)
/opt/conda/lib/python3.7/site-packages/aiida/orm/utils/managers.py:98: AiidaDeprecationWarning: dereferencing nodes with links containing double underscores is deprecated, simply replace the double underscores with a single dot instead. For example:
`self.inputs.some__label` can be written as `self.inputs.some.label` instead.
Support for double underscores will be removed in the future.
  'Support for double underscores will be removed in the future.', AiidaDeprecationWarning
In [2]: cj_node.outputs. <TAB>
```

The deprecation warning prints when I try to tab-complete the outputs. @ramirezfranciscof guided me to this PR. I didn't look into much detail what is responsible, but if I revert back to `aiida-core==1.6.1`, before this change was released, I no longer have this problem.

_Originally posted by @mbercx in https://github.com/aiidateam/aiida-core/issues/4863#issuecomment-873825647_

sphuber:
Fixes #5010 

The link managers for the `Node` class which are used for the `inputs`
and `outputs` attributes and facilitate the tab-completion of incoming
and outgoing links, was recently changed to deprecate the direct use of
double underscores in link labels in favor of treating them as normal
nested dictionaries. The deprecation warning was thrown whenever the
label contained a double underscore, but this would therefore also
trigger on dunder methods, which is not desirable behaviour.

This inaccuracy manifested itself in the deprecation method being
printed even when just activating the tab-completion on `node.outputs`
or `node.inputs` without even specifying a label with a double
underscore. It is not fully understood how `_get_node_by_link_label` is
called in doing this, but it seems some caching mechanism is calling the
`__wrapped__` attribute on the link manager, which in turn triggers the
deprecation warning. An additional clause in the condition to exclude
dunder methods fixes the behaviour.

### DIFF
--- a/tests/orm/utils/test_managers.py
+++ b/tests/orm/utils/test_managers.py
@@ -152,6 +152,10 @@ def test_link_manager_with_nested_namespaces(clear_database_before_test):
     out1.add_incoming(calc, link_type=LinkType.CREATE, link_label='nested__sub__namespace')
     out1.store()
 
+    out2 = orm.Data()
+    out2.add_incoming(calc, link_type=LinkType.CREATE, link_label='remote_folder')
+    out2.store()
+
     # Check that the recommended way of dereferencing works
     assert calc.inputs.nested.sub.namespace.uuid == inp1.uuid
     assert calc.outputs.nested.sub.namespace.uuid == out1.uuid
@@ -164,10 +168,27 @@ def test_link_manager_with_nested_namespaces(clear_database_before_test):
         assert calc.inputs.nested__sub__namespace.uuid == inp1.uuid
         assert calc.outputs.nested__sub__namespace.uuid == out1.uuid
 
+    # Dunder methods should not invoke the deprecation warning
+    with pytest.warns(None) as record:
+        try:
+            calc.inputs.__name__
+        except AttributeError:
+            pass
+        assert not record
+
     # Must raise a AttributeError, otherwise tab competion will not work
-    with pytest.raises(AttributeError):
-        _ = calc.outputs.nested.not_existent
+    for attribute in ['not_existent', 'not__existent__nested']:
+        with pytest.raises(AttributeError):
+            _ = getattr(calc.outputs.nested, attribute)
 
     # Must raise a KeyError
+    for key in ['not_existent', 'not__existent__nested']:
+        with pytest.raises(KeyError):
+            _ = calc.outputs.nested[key]
+
+    # Note that `remote_folder` corresponds to an actual leaf node, but it is treated like an intermediate namespace
+    with pytest.raises(AttributeError):
+        _ = calc.outputs.remote_folder__namespace
+
     with pytest.raises(KeyError):
-        _ = calc.outputs.nested['not_existent']
+        _ = calc.outputs['remote_folder__namespace']
--- a/aiida/orm/utils/managers.py
+++ b/aiida/orm/utils/managers.py
@@ -87,7 +87,10 @@ class NodeLinksManager:
         try:
             node = attribute_dict[label]
         except KeyError as exception:
-            if '__' in label:
+            # Check whether the label contains a double underscore, in which case we want to warn the user that this is
+            # deprecated. However, we need to exclude labels that corresponds to dunder methods, i.e., those that start
+            # and end with a double underscore.
+            if '__' in label and not (label.startswith('__') and label.endswith('__')):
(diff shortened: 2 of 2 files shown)


sphuber:
I think this affects `v1.6.*` so we might want to consider a patch release with this fix cherry-picked. I also think there have been some other bug fixes on `develop` that also affect the last release branch that we might want to include.

chrisjsewell:
cheers, I will have a look at these later today (once I have done my tutorial stint)

tests/orm/utils/test_managers.py
```diff
@@ -164,10 +168,20 @@ def test_link_manager_with_nested_namespaces(clear_database_before_test):
...
+        try:
+            calc.inputs.__name__
+        except AttributeError:
+            pass
+        assert not record
+
     # Must raise a AttributeError, otherwise tab competion will not work
-    with pytest.raises(AttributeError):
-        _ = calc.outputs.nested.not_existent
+    for attribute in ['not_existent', 'not__existent__nested', 'remote_folder__namespace']:
```
unkcpz, line 180: I guess you should separate `'not__existent__nested'` and `'remote_folder__namespace'` from `'not_existent'`. They are not nested in `nested` namespace.

unkcpz (changes requested):
Thanks @sphuber . I assume the added test cases are in order to test behaviour of return catch, the original `not_existent` is from `nested__not_existent`, so you may need to change either `link_label='remote_folder'` or separate them from the previous one.

tests/orm/utils/test_managers.py
```diff
@@ -164,10 +168,20 @@ def test_link_manager_with_nested_namespaces(clear_database_before_test):
...
+        with pytest.raises(AttributeError):
+            _ = getattr(calc.outputs.nested, attribute)
```
unkcpz, line 182:
```suggestion
    for attribute in ['not__existent__nested', 'remote_folder__namespace']:
        with pytest.raises(AttributeError):
            _ = getattr(calc.outputs, attribute)
```

tests/orm/utils/test_managers.py
```diff
@@ -164,10 +168,20 @@ def test_link_manager_with_nested_namespaces(clear_database_before_test):
...
 
     # Must raise a KeyError
-    with pytest.raises(KeyError):
-        _ = calc.outputs.nested['not_existent']
+    for key in ['not_existent', 'not__existent__nested', 'remote_folder__namespace']:
+        with pytest.raises(KeyError):
+            _ = calc.outputs.nested[key]
```
unkcpz, line 187:
```suggestion
    for key in ['not__existent__nested', 'remote_folder__namespace']:
        with pytest.raises(KeyError):
            _ = calc.outputs[key]
```

tests/orm/utils/test_managers.py
sphuber: Good catch. This was also the reason that the `TypeError` catch block was not being hit in the tests. It should be fixed now.

tests/orm/utils/test_managers.py
```diff
@@ -164,10 +168,27 @@ def test_link_manager_with_nested_namespaces(clear_database_before_test):
...
+
+    # Note that `remote_folder` corresponds to an actual leaf node, but it is treated like an intermediate namespace
+    with pytest.raises(AttributeError):
+        _ = calc.outputs.remote_folder__namespace
+
     with pytest.raises(KeyError):
-        _ = calc.outputs.nested['not_existent']
+        _ = calc.outputs.nested['remote_folder__namespace']
```
unkcpz, line 194:
```suggestion
        _ = calc.outputs['remote_folder__namespace']
```

here also need to be changed, correct?

unkcpz (approved):
Only a minor request, all looks good to me. 

Hoping it get merged soon, when looking at this I just find because of too much of this depracated complaint I turn off the ipython warning.

tests/orm/utils/test_managers.py
sphuber: Absolutely, apologies for the sloppy mistake

PR #5012, opened 2021-07-06
CLI: do not provide command hints during tab-completion

### LINKED ISSUE #3815: TAB completion crashes on unknown commands if followed by parameters
If I write `verdi ru count_query.py` and I press TAB I get on screen
```Traceback (most recent call last):
  File "/Users/pizzi/.virtualenvs/aiida_core_dev3/bin/verdi", line 10, in <module>
    sys.exit(verdi())
  File "/Users/pizzi/.virtualenvs/aiida_core_dev3/lib/python3.6/site-packages/click/core.py", line 764, in __call__
  File "/Users/pizzi/.virtualenvs/aiida_core_dev3/lib/python3.6/site-packages/click/core.py", line 712, in main
    _bashcomplete(self, prog_name, complete_var)
  File "/Users/pizzi/.virtualenvs/aiida_core_dev3/lib/python3.6/site-packages/click_completion/patch.py", line 111, in _shellcomplete
    do_bash_complete(cli, prog_name)
  File "/Users/pizzi/.virtualenvs/aiida_core_dev3/lib/python3.6/site-packages/click_completion/core.py", line 168, in do_bash_complete
    echo('\t'.join(re.sub(r"""([\s\\"'()])""", r'\\\1', opt) for opt, _ in choices), nl=False)
  File "/Users/pizzi/.virtualenvs/aiida_core_dev3/lib/python3.6/site-packages/click_completion/core.py", line 168, in <genexpr>
    echo('\t'.join(re.sub(r"""([\s\\"'()])""", r'\\\1', opt) for opt, _ in choices), nl=False)
  File "/Users/pizzi/.virtualenvs/aiida_core_dev3/lib/python3.6/site-packages/click_completion/core.py", line 86, in get_choices
    ctx = resolve_ctx(cli, prog_name, args)
  File "/Users/pizzi/.virtualenvs/aiida_core_dev3/lib/python3.6/site-packages/click_completion/lib.py", line 82, in resolve_ctx
    cmd = ctx.command.get_command(ctx, a[0])
  File "/Users/pizzi/git/aiida-core/aiida/cmdline/commands/cmd_verdi.py", line 77, in get_command
    '{matches}'.format(cmd=cmd_name, matches='\n'.join('\t{}'.format(m) for m in sorted(matches)))
  File "/Users/pizzi/.virtualenvs/aiida_core_dev3/lib/python3.6/site-packages/click/core.py", line 496, in fail
    raise UsageError(message, self)
click.exceptions.UsageError: 'ru' is not a verdi command.

The most similar commands are: 
        group
        run
```

I think this is triggering some buggy path and this is because there is a parameter after the wrong/incomplete command `ru` (instead tab completion works if I press tab directly after `verdi ru`)

sphuber:
Fixes #3815 

The `verdi` command uses a custom `Group` subclass that overrides the
`get_command` method to provide some additional information in case the
provided command name does not exist. These hints should help the user
spot potential typos by giving a list of existing commands with similar
names.

However, this feature was also being triggered during tab-completion.
For example, typing `verdi comput list` followed by activating
tab-completion would result in the error being displayed since `comput`
is not a valid command. In this case, one does not want to display the
error message with hint at all.

The tricky part is that there is no canonical way to distinguish between
a normal command execution and a tab-completion event. The best bet is
to use the `resilient_parsing` attribute on the `Context` which is set
to `True` during tab-completion. Although this attribute was introduced
into `click` directly to support auto-completion functionality, the
problem is that this is not the only use case for which this flag can be
set. It is therefore possible that there is some code path where this
flag is set to `True` but it does not actually correspond to a
tab-completion event. For now there doesn't seem to be a better solution
though and in most cases this approach should work.

### DIFF
--- a/aiida/cmdline/commands/cmd_verdi.py
+++ b/aiida/cmdline/commands/cmd_verdi.py
@@ -8,8 +8,8 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """The main `verdi` click group."""
-
 import difflib
+
 import click
 
 from aiida import __version__
@@ -52,9 +52,20 @@ class MostSimilarCommandGroup(click.Group):
         """
         cmd = click.Group.get_command(self, ctx, cmd_name)
 
-        # return the exact match
+        # If we match an actual command, simply return the match
         if cmd is not None:
             return cmd
+
+        # If this command is called during tab-completion, we do not want to print an error message if the command can't
+        # be found, but instead we want to simply return here. However, in a normal command execution, we do want to
+        # execute the rest of this method to try and match commands that are similar in order to provide the user with
+        # some hints. The problem is that there is no one canonical way to determine whether the invocation is due to a
+        # normal command execution or a tab-complete operation. The `resilient_parsing` attribute of the `Context` is
+        # designed to allow things like tab-completion, however, it is not the only purpose. For now this is our best
+        # bet though to detect a tab-complete event. When `resilient_parsing` is switched on, we assume a tab-complete
+        # and do nothing in case the command name does not match an actual command.
+        if ctx.resilient_parsing:
+            return
 
         if int(cmd_name.lower().encode('utf-8').hex(), 16) == 0x6769757365707065:
             import base64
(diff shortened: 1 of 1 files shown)


giovannipizzi (approved):
Thanks! I think that's OK - as a note (for myself mostly), I tried and just returning will trigger the 'default' bash completion (files in the current folder). I think that's OK and in line with other codes.

Also, I think using `resilient_parsing` is OK - maybe there are other places where this will be needed, but I think we're being more resilient than before with this change, so this should be good.

PR #5015, opened 2021-07-08
Corrected `except KeyError` to `except AttributeError` in TrajectoryData node methods.

Crivella:
In the TrajectoryData node there are several try checks on the get_attibute method, which except a KeyError in order then pick a default value/action.
`KeyError` should be changed to `AttributeError` as this is what the get_attribute method raises when the attribute is not found.


ramirezfranciscof:
Hey @Crivella ! Thanks for the contribution and sorry for the long wait!

I see that Codecov is complaining that there are not test for this, probably this contributed to us not noticing when this broke down. Would you be up for adding those? I would add a file in `test/orm/data/test_array.py` and just add three functions that trigger each of the raises and checks that the requested outputs are consistent.

Crivella:
HI @ramirezfranciscof , I have added the test that checks if the same try_except statement in the show_mpl_* functions behave properly when the attribute are not found.
I don't think i can check the variable directly as it is internally set andused by the function

ramirezfranciscof (approved):
Thanks @Crivella ! Sorry the delay but I was looking if there was a way to monkeypatch `plot_positions_XYZ` to write to a file or something like that and test that instead of the direct raising of `get_attribute`, which is not optimal but I guess is the best we have so far.

Anyways, I think this can go in now and we can later improve moving to pytest or being more precise with the testing, but great base material!

PR #5017, opened 2021-07-09
Add the `MsonableData` data plugin

### LINKED ISSUE #4975: ✨ NEW: `MSONableNode` data type

One of the challenges of developing for AiiDA is that in case you want to use a certain type of data, you need to be able to store it in the provenance as a node. This means that you can't just use a class representing a type of data in AiiDA without first writing a data type plugin. This can be a barrier to using AiiDA, since it requires extra work.

Below I describe an idea that I've been thinking about for a while to extend the number of data types in AiiDA with all `MSONable`s, and explore some of the consequences for provenance etc. There may still be flaws and in the end we might abandon the idea, but I wanted to write this down so others could comment.

Let's take `pymatgen`'s [`Composition`](https://pymatgen.org/pymatgen.core.composition.html#pymatgen.core.composition.Composition) class as an example. 

Most of the classes in `pymatgen` are `MSONable`, or Monty-JSON-serialisable. You can check the [`monty.json` ](https://guide.materialsvirtuallab.org/monty/monty.json.html) module for more details, but essentially every class that is `MSONable` needs to implement an `as_dict()` method, so that via the `MontyEncoder` and `MontyDecoder`, these classes can be JSON-serialized and stored in a file or a JSON-like database such as MongoDB (this is heavily used by [Fireworks](https://materialsproject.github.io/fireworks/)).

For the `Composition` example, this looks something like this:

```python
In [1]: from pymatgen.core import Composition

In [2]: comp = Composition('LiCoO2')

In [3]: comp.to_json()
Out[3]: '{"Li": 1.0, "Co": 1.0, "O": 2.0, "@module": "pymatgen.core.composition", "@class": "Composition", "@version": "2020.12.31"}'
```

My suggestion is to consider if we can't leverage the JSON-serialisabilty of `MSONable` classes to extend the types of data we can store in the database with _all_ `MSONable` ones. Since we can store JSON in the `postgres` database, this should be fairly straightforward. And since the `module` and `class` are stored in the dictionary returned by the `to_json()` method, we can reconstruct the original `MSONable` instance quite easily.

#### Methods, provenance and immutability

Obtaining the original class from the node should be fairly straightforward for the user:

```python
In [1]: comp_node = load_node(<PK>)

In [2]: comp = comp_node.get_msonable()  # Or get_instance()?
```

The user would then obtain the original instance that was stored and be able to use it in whatever way they want. This I think is already a fine first implementation. They can adapt the instance and then store a node again afterwards.

Of course, these intermediate steps will not be stored in provenance. So maybe we can go one step further, and split up the methods of the class:

* `get` methods that simply return data without modifying the `Node` instance. These _should_ be easily transferable to the `MSONableNode` instance from the `MSONable` class. Continuing from the (hypothetical) example above:
  ```python
  In [3]: comp.get_atomic_fraction('Li')
  Out[3]: 0.25

  In [4]: comp_node.get_atomic_fraction('Li')
  Out[4]: 0.25
  ```

* `set` methods that change something about the instance. These obviously cannot just adapt the `MSONableNode`, since the nodes are immutable by design (and for good reason). However, we could have them return a _new_ `MSONableNode` instance. Moreover, we could store the method used as a `calcfunction`, and add this operation to the provenance. Again, from the example:
  ```python
  In [5] comp_node.add_charges_from_oxi_state_guesses()
  Out[5]: <MSONableNode[Composition]: uuid: b1bc0507-9e0b-464d-b8e7-27ec14f45a8a (pk: 5)>
  ```
  This call would have been added to the provenance as a `calcfunction`, so:
  ```console
  $ verdi process list -a
    PK  Created    Process label                       Process State     Process status
  ----  ---------  ----------------------------------  ----------------  ----------------
   656  5s ago     add_charges_from_oxi_state_guesses  ⏹ Finished [0]
  ```

I'm not sure how difficult it is to check if a method is a `get` or `set` one. Just off the top of my head we'd probably load the origin `MSONable` instance, apply the method and then see if the `as_dict()` methods of the instance before and after are different. I think it should be feasible, but maybe I'm missing something.

### Describe alternatives you've considered

Another thing I've thought about, but I'm not sure if it would be possible, is to write a `MSONable` -> AiiDA data type converter. I'd have to explore this some more to see if it's feasible (even partially, to get a template that can then be adapted). However, this might be an inferior idea, since that means that any change in e.g. the `Composition` class would have to be updated in the corresponding data type class. On the other hand, that may be desirable to avoid things breaking with a breaking change in `pymatgen`.

### EDIT: Additional idea

An additional idea would be to extend the `MSONable` concept to AiiDA nodes, creating a e.g. a `NodeAble` class that can be added to any class to make instances of that class storable in an AiiDA database. I guess it would be a subclass of `MSONable`, where we rely on the `as_dict()` method for the JSON stored in the database and perhaps an `to_repository()` method that details how/what needs to be stored in the repository. Would be very cool if you could make any Python class storeable in an AiiDA database by simply:

* Adding `NodeAble` as a parent class.
* Implementing the `as_dict()` method.
* Implementing the `to_repository()` method.

I think this would be easier and more maintainable than writing separate AiiDA data types for everything.

---

Maybe I'm missing some very important detail, and this is just a pipe dream. 🙃 Comments are most welcome!

sphuber:
Fixes #4975 

The `MSONable` class, provided by the `monty` library, is designed to
make arbitrary classes easily JSON-serializable such that they can be
serialized and, for example, stored in a database. The class is used
extensively in `pymatgen` and many of their classes are MSONable. Since
the `pymatgen` classes are used often in the current AiiDA community, it
would be nice if these objects can be easily stored in the provenance
graph.

The new `MsonableData` data plugin, wraps any instance of a `MSONable`
class. When constructed, it calls `as_dict` which by definition returns
a JSON-serialized version of the object, which we can therefore store in
the nodes attributes and allow it to be stored. The `obj` property will
return the original MSONable instance as it deserializes it from the
serialized version stored in the attributes.

### DIFF
--- a/aiida/orm/nodes/data/msonable.py
+++ b/aiida/orm/nodes/data/msonable.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+"""Data plugin for classes that implement the ``MSONable`` class of the ``monty`` library."""
+import importlib
+
+from monty.json import MSONable
+
+from aiida.orm import Data
+
+
+class MsonableData(Data):
+    """Data plugin that allows to easily wrap objects that are MSONable.
+
+    To use this class, simply construct it passing an isntance of any ``MSONable`` class and store it, for example:
+
+        from pymatgen.core import Molecule
+
+        molecule = Molecule(['H']. [0, 0, 0])
+        node = MsonableData(molecule)
+        node.store()
+
+    After storing, the node can be loaded like any other node and the original MSONable instance can be retrieved:
+
+        loaded = load_node(node.pk)
+        molecule = loaded.obj
+
+    .. note:: As the ``MSONable`` mixin class requires, the wrapped object needs to implement the methods ``as_dict``
+        and ``from_dict``. A default implementation should be present on the ``MSONable`` base class itself, but it
+        might need to be overridden in a specific implementation.
+
+    """
+
+    def __init__(self, obj, *args, **kwargs):
+        """Construct the node from the pymatgen object."""
+        if obj is None:
+            raise TypeError('the `obj` argument cannot be `None`.')
+
+        if not isinstance(obj, MSONable):
+            raise TypeError('the `obj` argument needs to implement the ``MSONable`` class.')
+
+        for method in ['as_dict', 'from_dict']:
+            if not hasattr(obj, method) or not callable(getattr(obj, method)):
+                raise TypeError(f'the `obj` argument does not have the required `{method}` method.')
+
+        super().__init__(*args, **kwargs)
+
+        self._obj = obj
+        self.set_attribute_many(obj.as_dict())
+
+    def _get_object(self):
+        """Return the cached wrapped MSONable object.
+
+        .. note:: If the object is not yet present in memory, for example if the node was loaded from the database,
+            the object will first be reconstructed from the state stored in the node attributes.
+
+        """
+        try:
+            return self._obj
+        except AttributeError:
+            attributes = self.attributes
+            class_name = attributes['@class']
+            module_name = attributes['@module']
+
+            try:
+                module = importlib.import_module(module_name)
+            except ImportError as exc:
+                raise ImportError(f'the objects module `{module_name}` can not be imported.') from exc
+
+            try:
+                cls = getattr(module, class_name)
+            except AttributeError as exc:
+                raise ImportError(
+                    f'the objects module `{module_name}` does not contain the class `{class_name}`.'
+                ) from exc
+
+            self._obj = cls.from_dict(attributes)
+            return self._obj
+
+    @property
+    def obj(self):
+        """Return the wrapped MSONable object."""
+        return self._get_object()
--- a/setup.json
+++ b/setup.json
@@ -165,6 +165,7 @@
             "folder = aiida.orm.nodes.data.folder:FolderData",
             "int = aiida.orm.nodes.data.int:Int",
             "list = aiida.orm.nodes.data.list:List",
+            "msonable = aiida.orm.nodes.data.msonable:MsonableData",
             "numeric = aiida.orm.nodes.data.numeric:NumericType",
             "orbital = aiida.orm.nodes.data.orbital:OrbitalData",
             "remote = aiida.orm.nodes.data.remote.base:RemoteData",
--- a/tests/orm/data/test_msonable.py
+++ b/tests/orm/data/test_msonable.py
@@ -1,0 +1,149 @@
+# -*- coding: utf-8 -*-
+"""Tests for the :class:`aiida.orm.nodes.data.msonable.MsonableData` data type."""
+from monty.json import MSONable
+import pymatgen
+import pytest
+
+from aiida.orm import load_node
+from aiida.orm.nodes.data.msonable import MsonableData
+
+
+class MsonableClass(MSONable):
+    """Dummy class that implements the ``MSONable interface``."""
+
+    def __init__(self, data):
+        """Construct a new object."""
+        self._data = data
+
+    @property
+    def data(self):
+        """Return the data of this instance."""
+        return self._data
+
+    def as_dict(self):
+        """Represent the object as a JSON-serializable dictionary."""
+        return {
+            '@module': self.__class__.__module__,
+            '@class': self.__class__.__name__,
+            'data': self._data,
+        }
+
+    @classmethod
+    def from_dict(cls, d):
+        """Reconstruct an instance from a serialized version."""
+        return cls(d['data'])
+
+
+def test_construct():
+    """Test the ``MsonableData`` constructor."""
+    data = {'a': 1}
+    obj = MsonableClass(data)
+    node = MsonableData(obj)
+
+    assert isinstance(node, MsonableData)
+    assert not node.is_stored
+
+
+def test_constructor_object_none():
+    """Test the ``MsonableData`` constructor raises if object is ``None``."""
+    with pytest.raises(TypeError, match=r'the `obj` argument cannot be `None`.'):
+        MsonableData(None)
+
+
+def test_invalid_class_not_msonable():
+    """Test the ``MsonableData`` constructor raises if object does not sublass ``MSONable``."""
+
+    class InvalidClass:
+        pass
+
+    with pytest.raises(TypeError, match=r'the `obj` argument needs to implement the ``MSONable`` class.'):
+        MsonableData(InvalidClass())
+
+
+def test_invalid_class_no_as_dict():
+    """Test the ``MsonableData`` constructor raises if object does not sublass ``MSONable``."""
+
+    class InvalidClass(MSONable):
+
+        @classmethod
+        def from_dict(cls, d):
+            pass
+
+    # Remove the ``as_dict`` method from the ``MSONable`` base class because that is currently implemented by default.
+    del MSONable.as_dict
+
+    with pytest.raises(TypeError, match=r'the `obj` argument does not have the required `as_dict` method.'):
+        MsonableData(InvalidClass())
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_store():
+    """Test storing a ``MsonableData`` instance."""
+    data = {'a': 1}
+    obj = MsonableClass(data)
+    node = MsonableData(obj)
+    assert not node.is_stored
+
+    node.store()
+    assert node.is_stored
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_load():
+    """Test loading a ``MsonableData`` instance."""
+    data = {'a': 1}
+    obj = MsonableClass(data)
+    node = MsonableData(obj)
+    node.store()
+
+    loaded = load_node(node.pk)
+    assert isinstance(node, MsonableData)
+    assert loaded == node
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_obj():
+    """Test the ``MsonableData.obj`` property."""
+    data = {'a': 1}
+    obj = MsonableClass(data)
+    node = MsonableData(obj)
+    node.store()
+
+    assert isinstance(node.obj, MsonableClass)
+    assert node.obj.data == data
+
+    loaded = load_node(node.pk)
+    assert isinstance(node.obj, MsonableClass)
+    assert loaded.obj.data == data
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_unimportable_module():
+    """Test the ``MsonableData.obj`` property if the associated module cannot be loaded."""
+    obj = pymatgen.core.Molecule(['H'], [[0, 0, 0]])
+    node = MsonableData(obj)
+
+    # Artificially change the ``@module`` in the attributes so it becomes unloadable
+    node.set_attribute('@module', 'not.existing')
+    node.store()
+
+    loaded = load_node(node.pk)
+
+    with pytest.raises(ImportError, match='the objects module `not.existing` can not be imported.'):
+        _ = loaded.obj
+
+
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_unimportable_class():
+    """Test the ``MsonableData.obj`` property if the associated class cannot be loaded."""
+    obj = pymatgen.core.Molecule(['H'], [[0, 0, 0]])
+    node = MsonableData(obj)
+
+    # Artificially change the ``@class`` in the attributes so it becomes unloadable
+    node.set_attribute('@class', 'NonExistingClass')
+    node.store()
+
+    loaded = load_node(node.pk)
+
+    with pytest.raises(ImportError, match=r'the objects module `.*` does not contain the class `NonExistingClass`.'):
+        _ = loaded.obj


aiida/orm/nodes/data/msonable.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
+"""Data plugin for classes that implement the ``MSONable`` class of the ``monty`` library."""
+import importlib
+
+from monty.json import MSONable
```
chrisjsewell, line 5: This isn’t used no?

aiida/orm/nodes/data/msonable.py
```diff
@@ -0,0 +1,81 @@
...
+            raise TypeError('the `obj` argument needs to implement the ``MSONable`` class.')
+
+        for method in ['as_dict', 'from_dict']:
+            if not hasattr(obj, method) or not callable(getattr(obj, method)):
+                raise TypeError(f'the `obj` argument does not have the required `{method}` method.')
+
+        super().__init__(*args, **kwargs)
+
+        self._obj = obj
+        self.set_attribute_many(obj.as_dict())
```
chrisjsewell, line 47: Should we also check here that the @class and @module keys are available

aiida/orm/nodes/data/msonable.py
chrisjsewell: Oh no actually it is, but monty is not a direct dependency of aiida is it?

aiida/orm/nodes/data/msonable.py
```diff
@@ -0,0 +1,81 @@
...
+        loaded = load_node(node.pk)
+        molecule = loaded.obj
+
+    .. note:: As the ``MSONable`` mixin class requires, the wrapped object needs to implement the methods ``as_dict``
+        and ``from_dict``. A default implementation should be present on the ``MSONable`` base class itself, but it
+        might need to be overridden in a specific implementation.
+
+    """
+
+    def __init__(self, obj, *args, **kwargs):
```
chrisjsewell, line 32: Add type checking

aiida/orm/nodes/data/msonable.py
```diff
@@ -0,0 +1,81 @@
...
+        """Construct the node from the pymatgen object."""
+        if obj is None:
+            raise TypeError('the `obj` argument cannot be `None`.')
+
+        if not isinstance(obj, MSONable):
```
chrisjsewell, line 37: Do we actually need to restrict it to this, or can we just say that any class that has these properties can be used, i.e. an as_dict/to_dict that consume/create a json dict with at least the @class and @method keys

chrisjsewell (changes requested):

sphuber:
@zhubonan it would be great if you could give this a spin and is useful for your use case.

zhubonan:
Thanks for this. I will have a go!

zhubonan:
I just had a try and it works great for me. It can be very useful for storing the provenance of, say a phase diagram construction performed with pymatgen. At the moment I do this by manually converting AiiDA's records into `ComputedEntry` on-the-fly and make the phase diagram.
 
With this new capability, the pymatgen's entries (which has no provenance at all, so I have to manually add in the UUID....) can be stored and linked to AiiDA's workflows with the rovenance), and each phase diagram will also automatically have the full provenance record! 

One thing that I realised is that `fireworks` actually uses a different protocol rather than the standard MSONable class, in which case the objects are loaded by `FwSerializer`.  This is needed to allow code refactoring - so the class can be dynamically loaded:

https://github.com/materialsproject/fireworks/blob/a69911f9b99d025c322baa3dd003370b7a21dafb/fireworks/utilities/fw_serializers.py#L183

However, I think it is too code specific and not really a standard. I am happy to just having the `MsonableData` class in `aiida-core`.

zhubonan:
I have found some problems with further testing. One issue I have found is that the not all MSONable objects can be serialised with standard JSON. For example, some pymatgen objects contain NumPy array and NaN values, the latter cannot be stored in PostgreSQL directly because JSON dost not support NaN.

I can see that they have implemented special decoders/encoders for these cases to get around it:

https://github.com/materialsvirtuallab/monty/blob/b6204db0ddebb51441b397cd8c79418022ea8830/monty/json.py#L248

Example content with `nan`:

```
{'@module': 'pymatgen.entries.computed_entries',
 '@class': 'ComputedEntry',
 'energy': -8.93381717,
 'composition': defaultdict(float, {'Li': 1.0, 'O': 1.0}),
 'energy_adjustments': [{'@module': 'pymatgen.entries.computed_entries',
   '@class': 'ConstantEnergyAdjustment',
   '@version': '2020.10.20',
   'value': -0.70229,
   'uncertainty': nan,
   'name': 'MP Anion Correction',
   'cls': {'@module': 'pymatgen.entries.compatibility',
    '@class': 'MaterialsProjectCompatibility',
    '@version': '2020.10.20'},
   'description': 'Constant energy adjustment'}],
 'parameters': {'run_type': 'GGA',
  'is_hubbard': False,
  'pseudo_potential': {'functional': 'PBE',
   'labels': ['Li_sv', 'O'],
   'pot_type': 'paw'},
  'hubbards': {},
  'potcar_symbols': ['PBE Li_sv', 'PBE O'],
  'oxide_type': 'oxide'},
 'data': {'oxide_type': 'oxide'},
 'entry_id': 'mp-1097030',
 'correction': -0.70229}
```

zhubonan:
Oh actually it is the python's JSON encoder/decoder that does the work of sorting out NaN and infinity:

https://docs.python.org/3/library/json.html#json.JSONDecoder

sphuber:
> I have found a problem with further testing. Not all MSONable objects can be serialised with standard JSON. For example, some pymatgen objects contain NumPy array and NaN values, the latter cannot be stored in PostgreSQL directly because JSON dost not support NaN.

Hmm, for numpy arrays, what we could think of doing is after calling `as_dict` on the object, before storing it on the attributes of the node, we first loop over the serialized content and remove all numpy arrays and store them in the repository of the node instead. We leave a placeholder in the attributes such that upon deserialization, we replace those placeholders with the numpy array read from the repository. This would obviously be less performant but should work I think. Do you have an example of a MSONable class that contains a numpy array, which is easy to mock, so I can add this in tests for the implementation?

> Oh actually it is the python's JSON encoder/decoder that does the work of sorting out NaN and infinity:

Regarding the NaN values, we indeed don't support this because PostgreSQL doesn't support it in the JSONB fields: https://www.postgresql.org/docs/current/datatype-json.html#JSON-TYPE-MAPPING-TABLE
But then it is probably because NaN is not a valid JSON token. So it is not so much a problem of the Python encoder/decoders but merely that we can store it in the attributes. Only solution I see is to serialize this value  (and others, such as `infinity` etc.) to some kind of string token that than also has to be deserialized. But not sure if this is going to work.

zhubonan:
> Hmm, for numpy arrays, what we could think of doing is after calling as_dict on the object,
I think the `monty` package solves this by making the NumPy arrays part of the JSON representation: 

https://guide.materialsvirtuallab.org/monty/_modules/monty/json.html#MontyEncoder

We should able to add this to a pair of encoding and decoding functions that process the `dict` before serialization and after deserialization. The ability to store infinite values can also be added that way - they just saves to strings like `NaN` and `-Infinite`, but perhaps it should add it to the parent class `Dict`.  This is what python does anyway when encoding/decodnig JSON.

https://github.com/python/cpython/blob/3.9/Lib/json/encoder.py#L204

zhubonan:
Sorry, I think I have mixed some things up. 

Just to clarify, the support of NumPy array is not something we need to do on our side. Individual classes should implement `as_dict` to be able to support it. In pymatgen, I have seen classes get around this by doing around trip using the MontyEncoder to get a string and then load it as plain python:

https://github.com/materialsproject/pymatgen/blob/5d600ca610d1a4d3e1532293e5f87062332b9a26/pymatgen/entries/computed_entries.py#L549

If we really want this for any class, we could add this functionality by doing the same round trip thing or calling the `default` method of the `MontyEncoder`. But I think it is not necessary in most cases. 

So the only thing we need to add here is for processing the nan and inf values. I have raised a [PR](https://github.com/sphuber/aiida-core/pull/13) to the branch of this PR @sphuber.

sphuber:
> Just to clarify, the support of NumPy array is not something we need to do on our side. Individual classes should implement `as_dict` to be able to support it.

So the `pymatgen` class that you failed to wrap in the `MsonableData` class, was failing because it did not serialize the numpy arrays it contains? I guess that if we specify that `as_dict` _should_ return something that is JSON serializable, than that makes sense and technically we would be _correct_ in raising for classes that don't respect this. However, if many `pymatgen` classes do not actually properly return JSON serializable dictionaries, then that defeats the purpose of our class a bit doesn't it? If the goal is to allow to wrap "any" MSONable class without any changes, then it would be a bit pointless if it fails for most. Do you have an idea what percentage of classes roughly don't work with the current implementation? Did you just happen to find one of few corner cases? Or are there quite a number?

zhubonan:
No, what I meant is that in my case they just fail because of things like NaN or -inf being not JSON serializable. The numpy issue is a similar but different problem. 

Like you said the `as_dict` method of the class *should* take care of converting NumPy objects. But very often it does not but the end result can still be serialized if their `MontyEncoder` when saving it to a `json` file. 

For example:

```
class MsonableClass(MSONable):
    """Dummy class that implements the ``MSONable interface``."""

    def __init__(self, data):
        """Construct a new object."""
        self._data = data

    @property
    def data(self):
        """Return the data of this instance."""
        return self._data

    def as_dict(self):
        """Represent the object as a JSON-serializable dictionary."""
        return {
            '@module': self.__class__.__module__,
            '@class': self.__class__.__name__,
            'data': self._data,
        }

    @classmethod
    def from_dict(cls, d):
        """Reconstruct an instance from a serialized version."""
        return cls(d['data'])

obj = MsonableClass({'a': np.arange(10}))
```


- calling `as_dict` does not necessary give a JSON serializable dictionary - the numpy/datetime objects are still there . The returned dictionary cannot be dump into a file using the standard encoder. 
- calling `to_json` would return the JSON string representation,  encoded using `MontyEncoder`, where the numpy object is serialized, and the NaN, inf objects are converted to string.
- calling `from_dict` would automatically go through the `MontyDecoder` to reconstruct the numpy, datetime, objects etc.

Because we don't store the content in the database as string, but rather passing the JSON directly to PostgreSQL, hence we don't go through the encoder/decoder stages, but they are needed for many of the classes to work. 

As you can see, there is an asymmetry - `as_dict` does not always give a JSON serializable object as it does not go through the encoder, but `from_dict` does process and recover things like numpy arrays and datetime. 

To have the same level of support, we need to make sure when serializing we also go through the encoder. This can be done by processing objects using the `default` method that convert each object to into the JSON serializable format (I will update the code). 

In addition, similar pre-procssing and post-processing steps are needed for the `inf`, `NaN` values. They are implicitly handled by the `JSONEncoder` in the stdlib when dumping to /loading from JSON strings, but as we don't actually dump the object into strings, we have to process them explicitly here.

chrisjsewell:
Two notes that came to mind whilst discussing with @ramirezfranciscof:

1. Personally, I don't think we should need pymatgen/monty as a dependency here, nor even probably call this `MSONable`. It can be more general than that: just wrapping any class with an `as_dict` (instance) method, that creates a dict (which includes the special `@module` and `@class` methods), and a `from_dict` (class) method.
2. What happens if you try to load this with e.g. a Pymatgen version that has changed in a breaking manner, whereby the class is no longer in the same module, or the `from_dict` input dict structure has changed. I think this is going to be an issue for "long term" storage no? As in, you try to load this stuff from an archive, but it fails and you don't know why. I guess this is a similar issue to having data plugins, whereby there is no current functionality for migrations. There though, we do at least record the version of the plugin, and so here I think we should do the same thing (trying to capture the module's version).

zhubonan:
> What happens if you try to load this with e.g. a Pymatgen version that has changed in a breaking manner, whereby the class is no longer in the same module, or the from_dict input dict structure has changed.

I think that the MSONable does try to record the module version (not enforced) and when deserialising it is not checked. 

> Personally, I don't think we should need pymatgen/monty as a dependency here

I agree. We and do not need to check if an object is actually a subclass of `MSONable`. As long as it is `MSONable`-like it should work. However, we would still want to pre-process the return of the `as_dict` so that it can be stored by PostgreSQL as JSON, e.g. work around for the NaN, -inf values.

sphuber:
Note that the whole idea of this plugin was to make `pymatgen` classes that implement the `MSONable` interface easy to wrap and store in AiiDA. If we remove the requirement for accepted objects to be instances of `MSONable` and any class that implements `as_dict` and `from_dict`, I think we have to discuss the interface in more detail. A big problem in the current implementation was that what we really need is a method that returns a serialized version of the object that can be stored as JSON. The `as_dict` method of the `MSONable` class does not do this and so we still need to do a lot of work, including multiple roundtrip JSON (de)serialization. So technically we need `as_json` and `from_json` methods. This would make it very general and well define what objects can successfully be wrapped, but this means none of the `pymatgen` objects can be automatically wrapped, which was the original motivation for this.

zhubonan:
> The as_dict method of the MSONable class does not do this and so we still need to do a lot of work, including multiple roundtrip JSON (de)serialization. 

I probably got you confused... from the specification of `MSONable` class ([link](https://guide.materialsvirtuallab.org/monty/_modules/monty/json.html)) the returned value is supposed to be JSON serialisable (using the Python's encoder for the NaN and inf). However, the problem is that it is that a `MSONable` object may have attributes being dictionaries/lists of `MSONable` objects or numpy arrays, in this case returning a simple, unprocessed dictionary for calling the constructor can break the JSON serialisation requirement. If that is the case it is down to the subclasses to implement the `as_dict` correctly. And one solution in some pymatgen classes is to override the `MSONable.as_dict` with additional steps of make a JSON string using the `MontyEncoder`, and reload it using the plain `JSONDecoder`. Doing such round trips makes what `as_dict` returns being JSON serialisable (using `JSONDecoder`). However, if one's class is simple enough, the `as_dict` method does not need to implement this. 

What I meant is that we do not have to make `monty` a dependency and there is no need to check if an object is a sub-class of `MSONable`. We just implement the interface for serialisation and deserialisation for storing the data (store `as_dict`) in the database and a convenient method for recreating the object from the stored data (identify the class, and call `from_dict`). 

We can remove the step using the `MontyEncoder` when serialisation - this just meant that the classes not implementing the `as_dict` correctly would not work, which is fine. For deserialisation, the default `from_dict` method of `MSONable` does call the `MontyDecoder`, so we can also remove the additional step to call the `MontyDecoder` here. Again, this only makes the classes not implementing the original `MSONable` interface correctly not work, which I think is fine.

sphuber:
> I probably got you confused... 

I am afraid I am still confused :sweat_smile: 

> from the specification of `MSONable` class ([link](https://guide.materialsvirtuallab.org/monty/_modules/monty/json.html)) the returned value is supposed to be JSON serialisable (using the Python's encoder for the NaN and inf).

Exactly, but the problem is that some (if not many or all) implementations do not respect this requirement at all.

> However, the problem is that it is that a `MSONable` object may have attributes being dictionaries/lists of `MSONable` objects or numpy arrays, in this case returning a simple, unprocessed dictionary for calling the constructor can break the JSON serialisation requirement. If that is the case it is down to the subclasses to implement the `as_dict` correctly. And one solution in some pymatgen classes is to override the `MSONable.as_dict` with additional steps of make a JSON string using the `MontyEncoder`, and reload it using the plain `JSONDecoder`. Doing such round trips makes what `as_dict` returns being JSON serialisable (using `JSONDecoder`). However, if one's class is simple enough, the `as_dict` method does not need to implement  this.

I agree, but apparently some classes don't do this. The whole reason we went in this direction is because you reported that you found that certain `pymatgen` classes don't respect this requirement properly.

I am fine with adding a more generic class, that is not necessarily linked to `MSONAble` but accepts anything that respects the spec of `as_dict` and `from_dict`. However, that would mean that many `pymatgen` classes cannot be wrapped as-is as their implementations are incorrect.

zhubonan:
> Exactly, but the problem is that some (if not many or all) implementations do not respect this requirement at all.

>I agree, but apparently some classes don't do this. The whole reason we went in this direction is because you reported that you found that certain pymatgen classes don't respect this requirement properly.

I was wrong about this - most of the `pymatgen` classes do have the overrides for `as_dict` and provision for potential problems. Initially, I looked at the `MSONable.as_dict` and thought surely that would not work for all `MSONable` subclasses. The fact is that most of the latter had overrides. So the get around we need here is a just to deal with the `NaN`, `inf` and `-inf`.

zhubonan:
@sphuber I have made some suggestions on the changes here: https://github.com/sphuber/aiida-core/pull/13. 
I think that we just have to assume that the target class have implemented `as_dict` and `from_dict` properly. It should be the case for most (intended) use cases. 

Here is a ad-hoc test script I use:

```
MSONable = DataFactory('msonable')
from pymatgen.ext.matproj import MPRester
from pymatgen.analysis.phase_diagram import PhaseDiagram
rester = MPRester()
data = rester.get_entries_in_chemsys('Li-O')
pd = PhaseDiagram(data)
pentry = MSONable(pd)
pentry.store()
node = load_node(pentry.uuid)
output = node.obj
```

which is to pull some data entry from the materials project, create a phase diagram and store it into AiiDA. The `PhaseDiagram` class's `as_dict` assemble the dictionary by calling the `as_dict` of each objects (see [here](https://github.com/materialsproject/pymatgen/blob/937eff5cc83febdea57258c0488a5b81cb550506/pymatgen/analysis/phase_diagram.py#L1149-L1158)). The same `as_dict` can be achieved by doing `loads(dumps(pd, cls=MontyEncoder))` though, but that the former is the proper implementation.

zhubonan (dismissed):
Looks good to me. Decoupling from the `MSONable` class is sensible as the `to_dict`, `from_dict` interface is a general one - `MSONable` simply includes convenient default implementations. The issue with common but non-standard JSON objects such as  `-inf`, `inf` and `nan` are also addressed.

aiida/orm/nodes/data/msonable.py
sphuber: This is no longer applicable: dependency has been removed.

aiida/orm/nodes/data/msonable.py
sphuber: We now check explicitly and add it ourselves if it isn't already in the dict returned by the class.

aiida/orm/nodes/data/msonable.py
sphuber: I added type checking, but since we are very flexible, all we can say is `typing.Any` pretty much. Or is there some better typing we can have still?

aiida/orm/nodes/data/msonable.py
sphuber: Unrestricted the interface. Anything that has `as_dict` is accepted, and we add the keys ourselves if need be

sphuber:
I think we have converged on the best design. I have updated the PR to reflect this and have renamed all classes to properly reflect the fact that we support anything that is JSON-serializable. In that sense it is no longer directly tied to `monty`'s `MSONable`, which also makes it less of a problem that certain `pymatgen` classes won't be compatible, since their `as_dict` method's do not return a JSON-serializable dictionary.

@chrisjsewell I also addressed the remarks from your initial PR so this PR is now ready for final review.

aiida/orm/nodes/data/msonable.py
chrisjsewell: The best way is to define a Protocol: https://docs.python.org/3/library/typing.html#typing.Protocol, although you need the typing_extensions backport package for python < 3.8

chrisjsewell: (See e.g. https://github.com/executablebooks/markdown-it-py/blob/1d27af1da41fd3b00da8307040e6e274925437b2/markdown_it/renderer.py#L21)

sphuber:
@chrisjsewell typing is updated. Ready for final review

sphuber:
@chrisjsewell I will be merging this tomorrow. Let me know if you still want to review it.

aiida/orm/nodes/data/jsonable.py
```diff
@@ -0,0 +1,144 @@
...
+
+try:
+    from typing import Protocol
+except ImportError:  # Python <3.8 doesn't have `Protocol` in the stdlib
+    from typing_extensions import Protocol  # type: ignore[misc]
+
+
+class JsonSerializableProtocol(Protocol):
+
+    def as_dict(self) -> typing.MutableMapping[typing.Any, typing.Any]:
```
chrisjsewell, line 17: nitpick: I think to be jsonable, keys need to strings no?

```suggestion
    def as_dict(self) -> typing.MutableMapping[str, typing.Any]:
```

chrisjsewell (changes requested):
Cheers, just a minor nitpick

aiida/orm/nodes/data/jsonable.py
chrisjsewell: meh, I guess it would fail with monty's less defined type: https://github.com/materialsvirtuallab/monty/blob/6149bcd9c3b30fa35ad6dbad608a130b7860fc72/monty/json.py#L115, so don't worry about it

chrisjsewell (approved):

aiida/orm/nodes/data/jsonable.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def _get_object(self) -> JsonSerializableProtocol:
+        """Return the cached wrapped object.
+
+        .. note:: If the object is not yet present in memory, for example if the node was loaded from the database,
+            the object will first be reconstructed from the state stored in the node attributes.
+
+        """
+        try:
+            return self._obj
```
chrisjsewell, line 118: actually, wait when would this raise an `AttributeError`, do we not want to always reload the object, in case the attributes have changes?

chrisjsewell (changes requested):

aiida/orm/nodes/data/jsonable.py
chrisjsewell: Ah yeh the attribute error would be for if its loaded I guess, but yeh what if you did:

```
data = JsonableData(obj)
data.set_attrbiute('x', 1)
data.obj
```

would the user expect the obj to be updated?

sphuber: Good point. This is only really a problem though for unstored nodes and I doubt that users will be manually changing the attributes of the serialized object. Since the whole point of this class is to allow simple wrapping of serializable instances, there seems little point to be manually changing the attributes after construction but before storing. If we do want to support this as a use case, then we should indeed always fully reconstruct the obj from the database. This will introduce a significant performance hit since it needs to hit the db and deserialize the entire object. I would argue that this is not worth it, since these nodes will used most often to retrieve the `obj` from a stored node, rather than mutating unstored versions of them.

TL;DR: I would keep the current implementation and at most just add a warning in the `obj` property docstring that the thing is cached and so messing with attributes manually can break things

chrisjsewell: Yeh that sounds fine to me 👍

PR #5024, opened 2021-07-21
🐛 FIX: `BandsData` matplotlib with NaN values

PhilippRue:
The `BandsData` supports using `numpy.NaN` values in the bands array which can be used to remove some parts of the bands or to fill up parts which might be out of a certain energy window.

The matplotlib plotting functionality (`BandsData.show_mpl()`) however does not work with NaN values when the limits for the plot are found.  

With `nanmin` and `nanmax` instead of the simple `min/max` ignores NaN values in the bands array and makes the matplotlib plotter work for incomplete bands.


unkcpz (approved):
@PhilippRue thanks! This looks good to me. See if others have more to comment.

chrisjsewell:
cheers!

PR #5029, opened 2021-07-30
CI: Test direct installation via conda.

csadorf:
Instead of just testing whether we can install into the environment
defined by environment.yaml.

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -90,15 +90,17 @@ jobs:
     - run: python --version
     - run: which python
 
-    - name: Create conda environment
+    - name: Test direct installation
+      run: |
+        conda create -n test-install aiida-core
+        source activate test-install
+        python -c "import aiida"
+
+    - name: Test conda environment
       run: |
         conda env create -f environment.yml -n test-environment
         source activate test-environment
         python -m pip install --no-deps -e .
-
-    - name: Test importing aiida
-      run: |
-        source activate test-environment
         python -c "import aiida"
 
   tests:


csadorf:
This shows that aiida-core cannot be installed into a Python 3.9 environment, see the [test-install log](https://github.com/aiidateam/aiida-core/pull/5029/checks?check_run_id=3201508343):
```
2021-07-30T11:43:17.0115553Z python-3.8.10        | 26.2 MB   |            |   0% 
2021-07-30T11:43:17.1180013Z python-3.8.10        | 26.2 MB   | #9         |  19% 
2021-07-30T11:43:17.2667417Z python-3.8.10        | 26.2 MB   | ####4      |  44% 
2021-07-30T11:43:17.3697297Z python-3.8.10        | 26.2 MB   | ######5    |  65% 
2021-07-30T11:43:21.3715764Z python-3.8.10        | 26.2 MB   | #########7 |  97% 
2021-07-30T11:43:21.3717265Z python-3.8.10        | 26.2 MB   | ########## | 100% `
```

This is caused by aio-pika package which is constrained to [`~6.6`](https://github.com/aiidateam/aiida-core/blob/58c6d9f2679369c40b2c28cb32e6e1d88cc259ec/setup.json#L26):
```
$ conda create -n aiida-py39 python=3.9 aiida-core
Collecting package metadata (current_repodata.json): \done
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: /
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed

UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions

Package python conflicts for:
aiida-core -> aio-pika~=6.6 -> python[version='2.7.*|3.5.*|3.6.*|>=3.6|>=3.9,<3.10.0a0|>=3|>=3.5,<3.6.0a0|=2.7|>=3.5|>=2.7|3.4.*|3.8.*|3.7.*']
python=3.9
aiida-core -> python[version='>=2.7,<2.8.0a0|>=3.5|~=3.7|>=3.6,<3.7.0a0|>=3.8,<3.9.0a0|>=3.7,<3.8.0a0']
```
This is odd, because [the package is no-arch as of 6.7.1](https://anaconda.org/conda-forge/aio-pika/files) which would be compatible with our constraint. Installing aio-pika in isolation works flawlessly:
```
conda create -n aio-pika python=3.9 aio-pika~=6.6
```
and will install version 6.8.

csadorf:
I'm tracking the current failure to install with conda in combination with `python=3.9` in #5037 .

.github/workflows/test-install.yml
```diff
@@ -90,15 +90,17 @@ jobs:
     - run: python --version
     - run: which python
 
-    - name: Create conda environment
+    - name: Test direct installation
       run: |
-        conda env create -f environment.yml -n test-environment
-        source activate test-environment
-        python -m pip install --no-deps -e .
+        conda env create -n test-install aiida-core
```
ltalirz, line 95: Is that supposed to install aiida-core into the new environment?

When I run it locally, I get
```
$ conda env create -n test-install  aiida-core

SpecNotFound: Invalid name, try the format: user/package
```

Also, I guess it's good to test this but if there are issues with the conda-forge feedstock for aiida-core they will need to be fixed in the respective repository (cannot be fixed here).

ltalirz:
could this be related to https://github.com/conda-forge/aiida-core-feedstock/issues/41 ?

.github/workflows/test-install.yml
csadorf: Hmm.. odd, it appears to have worked even with the `env` command on the runner, however I updated it to use the more standard (?) `create` command.

ltalirz:
Thanks, for me it would be ok to test this here but as mentioned if this breaks it will need fixes in a different repo.

Also, if I understand correctly, the test would fail now - do you still want to merge this?

ltalirz (approved):
ah ok, I see it just installs python3.8

ok for me

PR #5033, opened 2021-08-01
Add the `ElectronicBandsData` data plugin

### LINKED ISSUE #5032: Add support for fermi energy to the `BandsData`
This feature has cropped up in various discussions, see issue #2847 and https://github.com/aiidateam/aiida-common-workflows/issues/210

sphuber:
Fixes #5032 

This data plugin is a subclass of the `BandsData` plugin. It adds
support to define a fermi energy, which is set as an attribute. This is
a typical use case for electronic band structures, which so far were
using the `BandsData` type, but the latter was also being used for other
band structures, such as phonons, which do not share this property.

By making it a subclass, existing processes that accept `BandsData`
instances will automatically also accept `ElectronicBandsData` instances
and queries for `BandsData` will also match the subclasses, unless the
user turns of subclassing explicitly.

### DIFF
--- a/aiida/orm/nodes/data/array/bands/electronic.py
+++ b/aiida/orm/nodes/data/array/bands/electronic.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+"""Data plugin to represent an electronic band structure."""
+from typing import Optional
+
+from numpy.typing import ArrayLike
+
+from aiida.common.lang import type_check
+from ..kpoints import KpointsData
+from .bands import BandsData
+
+__all__ = ('ElectronicBandsData',)
+
+
+class ElectronicBandsData(BandsData):
+    """Data plugin to represent an electronic band structure.
+
+    It subclasses the ``BandsData`` plugin and adds the ``fermi_level`` attribute.
+    """
+
+    KEY_FERMI_LEVEL = 'fermi_level'
+
+    def __init__(self, kpoints: KpointsData, bands: ArrayLike, fermi_level: Optional[float] = None, **kwargs):
+        """Construct a new instance.
+
+        The arguments ``kpoints``, ``bands`` and ``fermi_level`` can be set through the constructor but have to be
+        provided as positional arguments.
+
+        .. note:: Since we are currently still supporting Python 3.7, we cannot yet use the ``/`` positional argument
+            which was added to Python 3.8. Once support for Python 3.7 is dropped we can enforce the correct usage of
+            the arguments through that operator.
+
+        """
+        super().__init__(**kwargs)
+
+        self.set_kpointsdata(kpoints)
+        self.set_bands(bands)
+
+        if fermi_level is not None:
+            self.fermi_level = fermi_level
+
+    @property
+    def fermi_level(self) -> Optional[float]:
+        """Return the fermi level or ``None`` if one is not defined."""
+        return self.get_attribute(self.KEY_FERMI_LEVEL, None)
+
+    @fermi_level.setter
+    def fermi_level(self, value: float):
+        """Set the fermi level."""
+        type_check(value, float)
+        self.set_attribute(self.KEY_FERMI_LEVEL, value)
--- a/aiida/cmdline/commands/cmd_data/cmd_show.py
+++ b/aiida/cmdline/commands/cmd_data/cmd_show.py
@@ -208,7 +208,7 @@ def _show_xmgrace(exec_name, list_bands):
     import sys
     import subprocess
     import tempfile
-    from aiida.orm.nodes.data.array.bands import MAX_NUM_AGR_COLORS
+    from aiida.orm.nodes.data.array.bands.bands import MAX_NUM_AGR_COLORS
 
     list_files = []
     current_band_number = 0
--- a/aiida/orm/nodes/data/array/bands/__init__.py
+++ b/aiida/orm/nodes/data/array/bands/__init__.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+###########################################################################
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Module with `Node` sub classes for band structure data types."""
+from .bands import BandsData, find_bandgap, prepare_header_comment
+from .electronic import ElectronicBandsData
+
+__all__ = ('BandsData', 'find_bandgap', 'prepare_header_comment', 'ElectronicBandsData')
--- a/aiida/orm/nodes/data/array/bands/bands.py
+++ b/aiida/orm/nodes/data/array/bands/bands.py
@@ -18,7 +18,9 @@ import numpy
 
 from aiida.common.exceptions import ValidationError
 from aiida.common.utils import prettify_labels, join_labels
-from .kpoints import KpointsData
+from ..kpoints import KpointsData
+
+__all__ = ('BandsData',)
 
 
 def prepare_header_comment(uuid, plot_info, comment_char='#'):
--- a/docs/source/nitpick-exceptions
+++ b/docs/source/nitpick-exceptions
@@ -93,6 +93,8 @@ py:class sphinx.util.docutils.SphinxDirective
 py:class frozenset
 
 py:class numpy.bool_
+py:class numpy.dtype
+py:class numpy.typing._array_like._SupportsArray
 
 py:class paramiko.proxy.ProxyCommand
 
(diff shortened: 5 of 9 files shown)


sphuber:
Probably need to add a `fermi_energy_unit` property as well, but first wanted to gauge if this solution might cover all use cases that we know off.

aiida/orm/nodes/data/array/bands/electronic.py
```diff
@@ -0,0 +1,49 @@
...
+__all__ = ('ElectronicBandsData',)
+
+
+class ElectronicBandsData(BandsData):
+    """Data plugin to represent an electronic band structure.
+
+    It subclasses the ``BandsData`` plugin and adds the ``fermi_energy`` attribute.
+    """
+
+    KEY_FERMI_ENERGY = 'fermi_energy'
```
mbercx, line 18: Why have this more complicated `KEY_FERMI_ENERGY` class attribute and then set/get the instance attributes using `set_attribute` and `get_attribute`? Also, on line 35 you then still use `self.fermi_energy`, so why not simply use this everywhere?

aiida/orm/nodes/data/array/bands/electronic.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, **kwargs):
```
mbercx, line 20: I must admit I do not like this constructor using `**kwargs` from a UX point of view. ^^ Basically it's impossible to tell for a user that checks the docstring how to use it. What's the advantage of this, and does it outweigh the benefit of the user simply seeing the input arguments and having them documented in the docstring?

aiida/orm/nodes/data/array/bands/electronic.py
sphuber: Not sure what you mean here. The reason for the class attribute is that we need to store the value as an attribute with a certain key. I could have chosen anything, but the same name as the property only seems logical. The attribute is there so as to not have to repeat a string literal. Are you asking why the property `fermi_energy` is there as opposed to letting people use `get_attribute` and `set_attribute` directly?

sphuber:
Thanks for the review @mbercx . Fine for me to change to `fermi_level`. Maybe it would be good to raise this point during a CWF meeting and see that others agree. Would also be a good moment to see if people think we should add an attribute for the energy unit or whether we enforce `eV` as has been the norm for data types in `aiida-core` so far. I think being explicit and storing the unit is the better option even though that means we'd be breaking with current "standard" practice.

aiida/orm/nodes/data/array/bands/electronic.py
mbercx: Hmm, I _was_ a little quick in going through the code, so the comment on line 35 didn't really make sense. 😅 I guess my question is: Is it worth adding this extra complexity versus using e.g. just `_fermi_energy` to store the attribute just to avoid a bit of repetition. I.e. why not remove `KEY_FERMI_ENERGY` and set/get the Fermi level by just assigning it to a hidden attribute:

```python
    @property
    def fermi_energy(self) -> Optional[float]:
        """Return the fermi energy or ``None`` if one is not defined."""
        return self._fermi_energy

    @fermi_energy.setter
    def fermi_energy(self, value: float):
        """Set the fermi energy."""
        if self.is_stored:
            raise ModificationNotAllowed('The node is stored and cannot be modified.')

        type_check(value, float)
        self._fermi_energy = value
```

sphuber: How is the value in your example getting persisted to the database though? Note that an attribute of the Python class `ElectronicBandsData` does not automatically get stored in the attributes _column_ of the underlying node in the database. We need to manually do this mapping. The property `fermi_energy` is there just to make life easier for the user. Instead of having to know or figure out which key we use to store the value in the attributes, they can discover it through tab-completion.

mbercx: I see, there was my lack of understanding. I missed the requirement to use these methods to actually store the information in the columns of the database. Thanks for the clarification! 👍

aiida/orm/nodes/data/array/bands/electronic.py
sphuber: I have changed the interface. However, it does bring other potential use problems with it. See the commit message for an explanation. Using the `/` marker to make them positional only "solves" the problem somewhat.

sphuber: Hmmm, the `/` positional argument only marker is only supported for Python 3.8 and up. Not sure if there is a backport solution for this, but I doubt it since it concerns the basic syntax. @chrisjsewell do you happen to know?

giovannipizzi:
If the units in BandsData are always eV (I think this should be the case now that we are defining a Electronic type), I would vote to also force the fermi_level to be in eV and document it. I think it just makes our life too complex if we need to then define a new layer to convert units between various ElectronicBandsData. If we really want to have multiple units, I still think the units of the Fermi level should be the same as for the bands, so we would define only one unit for bands and the Fermi level.

Also, if we want to go down this route of defining an ElectronicBandsData type, we should move in the new type all functionality that is specific to electronic bands (two bands channels = spin up/down, the find_bandgap function, ...)?
I'm still wondering if this is really useful, in the sense that for me it's a bit different than the case of Kpoints, where we really were storing two different types of kpoints (grid vs. explicit points), and some methods just except if the other type has been set. Here there's just the definition of a "zero", and one could give a more general name (energy_zero for instance). So I think that having a subclass is less critical. But anyway, I'm not against it if we properly move electron-specific methods to this subclass and the users won't notice when the plugin change type.

mbercx:
> Also, if we want to go down this route of defining an ElectronicBandsData type, we should move in the new type all functionality that is specific to electronic bands (two bands channels = spin up/down, the find_bandgap function, ...)?

I'm definitely in favor of implementing the `ElectronicBandsData` type, and adding the methods you mention. The `find_bandgap` method would really be useful, I think. How would I now obtain a band gap from a `BandsData` node? This should be _very_ straightforward for users.

giovannipizzi:
> How would I now obtain a band gap from a BandsData node? 

Ok, my fault - I was under the false impression that there was a `band.find_bandgap()` method, instead there is only a function. I would suggesting to move this to `aiida.tools.data`. To make it easy, I think we should add also the `datanode.tools.*` extensions similar to what I think we have for calcjobs already, but I don't think we have yet for data nodes? So one can just add things to `aiida.tools` rather than to the class itself, that probably should just focus son how to store the data internally, but still you will be able to do `bandnode.tools.find_bandbap` rather than
```python
from aiida.tools.data.array.bands import find_bandgap
find_bandgap(bandnode)
````
(with the additional benefit that one can do `node.tools.<TAB>` to discover which tools exist for that node type.

I can't find the issue on `data.tools`, maybe @sphuber remembers? I think we discussed on the train back from the 2019 coding week.

sphuber:
> I can't find the issue on `data.tools`, maybe @sphuber remembers? I think we discussed on the train back from the 2019 coding week.

The reason that there is no `data.tools` analogue to the `calculation.tools` is because they do not have the same problem. The reason that we implemented the calculation tools concept was because the process node classes in AiiDA are not subclassable. This means that people cannot add their own custom functionality such as utility methods that operate on the various process nodes. The same does not go for `Data` nodes. These _are_ subclassable and so any utility functions can be added directly on those classes. If the `ElectronicBandsData` needs a `get_band_gap` method, then we should just implement on that class. I don't see the added value of implementing this elsewhere through some complicated entry point system.

mbercx:
Regarding placing these in e.g. `BandsData.tools.<method_name>`: Although I do understand that adding these here will make them easier to find considering the vast number of methods available for all node classes, we do need to make sure that this is documented _very_ well, and also introduced at the tutorial.

However, I would be more in favor of not having this kind of workaround, which again requires us to teach the users something to just find the utility methods they need. Instead, it would be better if we could hide methods that users typically do not need, so they can more easily find the methods that we would place in `.tools`. I think I already discussed this with @chrisjsewell at some point, and he mentioned a way to do this elegantly which I now forgot. (this was in the context of https://github.com/aiidateam/aiida-core/issues/4975)

giovannipizzi:
OK to keep it in the main class, but indeed for a user it's overwhelming to find all node methods (that they might not need) + the methods acting on the nodes. Also this makes data classes quite complex, again with a design principle that I would like to push forward that we keep data classes as stable as possible over time, just focusing on storing data in the DB in a consistent format, and move code that processes it outside.

As a historical note we had tried in AiiDA 0.x to hide methods (prefixing with underscore) that might not be needed by a end user (e.g. `set_attribute`), but then we agreed that this was not the right approach, as it was actually hiding methods that, from a python point of view, were not really private, as e.g. a plugin or some other AiiDA-core class might want to call them to set an attribute.
If there is some other way to do the "right" thing I'm all ears (maybe it's overriding the `__dir__`? But I don't know if I would do it, the user would be confused of not finding a method in dir but still be able to use it. Namespacing feels better to me.

But anyway this is also a matter of taste, and it's not crucial for this issue.

sphuber:
Will close this PR for now since it has been inactive for a long time and I don't anticipate having time soon to address the changes. Will reopen when there is a new design incorporating the suggested changes.

PR #5039, opened 2021-08-04
Remove capitalize of Exit codes spec of process show by `verdi plugin list`

unkcpz:
This is a quick fix for the exit code spec message of process print by `verdi plugin list <process>`. As you can see by running `verdi plugin list aiida.workflows quantumespresso.pw.relax` you'll get:
```
Exit codes:
                              ...
                              402:  The final scf pwbaseworkchain sub process failed
```

The expected output shall be:
```
Exit codes:
                              ...
                              402:  the final scf PwBaseWorkChain sub process failed
```

I believe it is better to let plugin developers be responsible for Title the message.


sphuber (dismissed):

sphuber (approved):
Thanks @unkcpz

PR #5046, opened 2021-08-04
Dependencies: Bump jinja2 to version 3.0

csadorf:


csadorf:
The docs failure appears to be related to: https://github.com/mkdocs/mkdocs/issues/864

csadorf:
Fixed in https://github.com/pydata/pydata-sphinx-theme/commit/3455c23ad887401eae585c37921178faf99ef4e5 (v0.6.3).

ltalirz (approved):
thanks @csadorf !

ltalirz:
@csadorf , the man who merges faster than his shadow...

csadorf:
> @csadorf , the man who merges faster than his shadow...

Haha, auto- merge was on. ;)

adegomme:
I just bumped into it, there is an issue with the pyzmq version in the 3.9 requirements file, it lists 22.1.1 (which does not exist) instead of 22.2.1 as i the other ones.

ltalirz:
Thanks for the report @adegomme 

@csadorf do you know how that got in there?
Also, I would have thought that the requirements files are being tested whenever they are touched?

csadorf:
> Thanks for the report @adegomme
> 
> @csadorf do you know how that got in there?
> Also, I would have thought that the requirements files are being tested whenever they are touched?

That was probably caused by a typo while resolving a merge conflict. However, we would probably have noticed if we would actually be testing on Python 3.9. [Right now we are only testing Python 3.7 and 3.8](https://github.com/aiidateam/aiida-core/blob/77aa5a4388d4a7fb6f0dab3723557803de631b4f/.github/workflows/ci-code.yml#L52). I assume to save computational resources?

Should we start testing on Python 3.9?

ltalirz:
Hm... I guess what I'm saying is: why was that not caught by the test-install workflow, which *should* run for every python version?

https://github.com/aiidateam/aiida-core/blob/77aa5a4388d4a7fb6f0dab3723557803de631b4f/.github/workflows/test-install.yml#L1-L14

csadorf:
> Hm... I guess what I'm saying is: why was that not caught by the test-install workflow, which _should_ run for every python version?

That workflow tests the installation of AiiDA and _checks_ whether the requirements file are compliant with our dependency specification. It does not try to create an environment from the requirements files. The non-existent pyzmq version 22.1.1 is compliant with our specification.

ltalirz:
I see, so it installs AiiDA on all versions but not from the requirements files.

Does this mean that the requirements-py-3.9.txt file was completely unused so far (except for being checked to follow the spec)?
It seems to me that this issue indicates that this check is not enough.
Perhaps `test-install` should also test installing from the requirements files?

I would not want to add this to the default CI that runs on every commit, but I think that would be ok to check whenever the requirements files are updated.

csadorf:
Agreed, it makes sense to try to pip-install the environments defined in the requirements files as part of the `test-install` workflow. Since that workflow is not run on every commit, I'd actually prefer to test all supported Python versions instead of swapping py38 with p39.

- [ ] Add tests to create a Python environment from all requirements files.
- [ ] Add Python 3.9 to `tests` job.

If you agree, I can propose that change.

ltalirz:
sounds good!

csadorf:
> sounds good!

Implemented in https://github.com/aiidateam/aiida-core/pull/5082 .

PR #5048, opened 2021-08-04
📚 DOCS: Make clear you can also stash folders

### LINKED ISSUE #4814: Improve stashing documentation
_Originally posted by @sphuber in https://github.com/aiidateam/aiida-core/issues/4812#issuecomment-800111916_

> @sphuber you wrote a lovely commit message for [9359b07](https://github.com/aiidateam/aiida-core/commit/9359b079bc49cec6d3f1e871ec856b71928c0696), but didn't add any of this to the documentation wink
> 
> (also some minor improvements to the new  documentation)

I take pride in my commit messages ;) the reason I didn't add it to the docs was because @mbercx was going to do this. Thanks for opening this PR. I will already give some quick comments, but then I let @mbercx say whether he wants to add more to this PR or we merge this and then he can extend the docs in a later separate PR. Note also that the commit message contained quite some details on design choices, for example on the choice of which  plugin to use for the stash folder. This would be ideal to add to the [internal architecture docs](https://aiida-core.readthedocs.io/en/latest/internals/index.html).

mbercx:
Fixes #4814 

The current documentation on stashing data after a calculation only discusses
the stashing of files with the copy mode. However, it is perfectly possible
to stash folders as well. Here we update the language of the documentation on
stashing to reflect this.

### DIFF
--- a/docs/source/topics/calculations/usage.rst
+++ b/docs/source/topics/calculations/usage.rst
@@ -504,8 +504,8 @@ The ``stash`` option namespace allows a user to specify certain files and/or fol
 This can be useful if these need to be stored for a longer time on a machine where the scratch space is cleaned regularly, but they need to be kept on the remote machine and not retrieved.
 Examples are files that are necessary to restart a calculation but are too big to be retrieved and stored permanently in the local file repository.
 
-The files/folders that need to be stashed are specified through their relative filepaths within the working directory in the ``stash.source_list`` option.
-Using the ``COPY`` mode, the target path defines another location (on the same filesystem as the calculation) to copy the files/folders to, and is set through the ``stash.target_base`` option, for example:
+The files and that need to be stashed are specified through their relative filepaths within the working directory in the ``stash.source_list`` option.
+Using the ``COPY`` mode, the target path defines another location (on the same filesystem as the calculation) to copy the files to, and is set through the ``stash.target_base`` option, for example:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


mbercx:
@ramirezfranciscof not sure on the usage of the "files/folders", but having "files and/or folders" everywhere seemed a bit much, and using e.g. "data" is not quite as explicit.

ramirezfranciscof:
Hey @mbercx thanks for the quick reaction to this! Indeed, I would say that even "files/folders" seems a bit overwhelming haha. Perhaps you can remove some of these, specially on the admonitions? I think as long as it is mentioned once or twice that one can also copy the folder it should be enough, I don't think we need to keep this "explicit precision" every time we refer to copying the data.

mbercx:
I probably went overboard a bit. ^^ Will adjust!

docs/source/topics/calculations/usage.rst
```diff
@@ -495,17 +495,17 @@ The contract of the 'retrieve temporary list' is essentially that the files will
...
+The ``stash`` option namespace allows a user to specify certain files and/or folders that are created by the calculation job to be stashed somewhere on the remote where the job is run.
+This can be useful if these need to be stored for a longer time on a machine where the scratch space is cleaned regularly, but they need to be kept on the remote machine and not retrieved.
 Examples are files that are necessary to restart a calculation but are too big to be retrieved and stored permanently in the local file repository.
 
-The files that are to be stashed are specified through their relative filepaths within the working directory in the ``stash.source_list`` option.
-Using the ``COPY`` mode, the target path defines another location (on the same filesystem as the calculation) to copy these files to, and is set through the ``stash.target_base`` option, for example:
+The files and that need to be stashed are specified through their relative filepaths within the working directory in the ``stash.source_list`` option.
```
ramirezfranciscof, line 507:
```suggestion
The files/folder that need to be stashed are specified through their relative filepaths within the working directory in the ``stash.source_list`` option.
```

I would maybe just use it here too because it is specifically saying how to specify them in the `stash.source_list`

ramirezfranciscof (approved):
Were you done with the modifications or did I jump the gun here? I feel like this is good. I added a last correction but good to merge afterwards!

ramirezfranciscof (approved):
GTG thanks! Btw I'm surprised you managed to find the ñ, I always struggle with that in non-spanish keyboards.

mbercx:
@ramirezfranciscof seems the checks aren't running, prob because this only a documentation update, but it still requires them?

PR #5050, opened 2021-08-04
bugfix/data bands list ungrouped

### LINKED ISSUE #5049: `verdi data bands list` lists only grouped band structures
### Describe the bug

I am running some test calculations with the aiida-cp2k plugin which generate band structures, but after the band structure was stored, `verdi data bands list` remains empty. Only after I add the band structures to a group they appear in `band structure list`.

### Steps to reproduce

1. Let a calculation create a `BandsData`
2. Make sure it is now in a group
3. List bands data with `verdi data bands list`
4. Observe only grouped band structures being listed

### Expected behavior

Have all band structures listed.

### Your environment

- Operating system [e.g. Linux]: Linux
- Python version [e.g. 3.7.1]: 3.9
- aiida-core version [e.g. 1.2.1]: 2.0.0a1

dev-zero:
Fixes #5049 

- abstractqueries/get_bands_and_parents_structure: list ungrouped BandsData

### DIFF
--- a/aiida/orm/nodes/data/array/bands.py
+++ b/aiida/orm/nodes/data/array/bands.py
@@ -1808,7 +1808,7 @@ def get_bands_and_parents_structure(args):
         A list of sublists, each latter containing (in order):
             pk as string, formula as string, creation date, bandsdata-label
     """
-    # pylint: disable=too-many-locals
+    # pylint: disable=too-many-locals,too-many-branches
 
     import datetime
     from aiida.common import timezone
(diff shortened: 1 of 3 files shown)


sphuber:
So that answers that long standing question of "who the hell uses these old abstract queries" :smile: To be honest though, I always felt that these shouldn't really be in `aiida-core`. It is one thing to have domain specific data types in there, but to have these very use case specific queries baked in feels off. They also go way back when there wasn't a single query builder frontend for both database backends. Since we are not removing the data types for v2.0, I guess we can keep this as well, so ok to fix this. But since it concerns a bug, it would be good to add a test for it. I think these abstract queries are notoriously under tested.

dev-zero:
rebased and test added for the first half of the `get_bands_and_parents_structure` (which this fix touches)

sphuber (approved):
Thanks @dev-zero . I have updated the tests to use `pytest` style instead. The `AiidaTestCase` method is being phased out.

PR #5054, opened 2021-08-08
REST API: make the profile configurable as request parameter

### LINKED ISSUE #5052: REST API: make the target profile a query parameter
Currently, the profile on which the REST API operates is fixed at runtime of the REST API and cannot be changed or configured per request. For a machine serving multiple profiles, this necessitates running a separate REST API instance per profile, which is not very performant. It would be great if the profile can be specified per request and the instance switches the loaded profile if needed. Some of the code necessary for this is already present, but it has not been marked as safe for use in production. Issue #2759 is tracking robustifying this and making switching profiles within a Python interpreter safe.

sphuber:
Fixes #5052 

To make this possible, after parsing the query string but before
performing the request, the desired profile needs to be loaded. If any
other profile is already loaded, that first needs to be properly
unloaded, properly unloading the database backend and the manager with
all it resources.

### DIFF
--- a/aiida/restapi/common/utils.py
+++ b/aiida/restapi/common/utils.py
@@ -486,6 +486,7 @@ class Utils:
         extras = None
         extras_filter = None
         full_type = None
+        profile = None
 
         # io tree limit parameters
         tree_in_limit = None
@@ -539,10 +540,17 @@ class Utils:
             raise RestInputValidationError('You cannot specify extras_filter more than once')
         if 'full_type' in field_counts and field_counts['full_type'] > 1:
             raise RestInputValidationError('You cannot specify full_type more than once')
+        if 'profile' in field_counts and field_counts['profile'] > 1:
+            raise RestInputValidationError('You cannot specify profile more than once')
 
         ## Extract results
         for field in field_list:
-            if field[0] == 'limit':
+            if field[0] == 'profile':
+                if field[1] == '=':
+                    profile = field[2]
+                else:
+                    raise RestInputValidationError("only assignment operator '=' is permitted after 'profile'")
+            elif field[0] == 'limit':
                 if field[1] == '=':
                     limit = field[2]
                 else:
@@ -658,7 +666,7 @@ class Utils:
 
         return (
             limit, offset, perpage, orderby, filters, download_format, download, filename, tree_in_limit,
-            tree_out_limit, attributes, attributes_filter, extras, extras_filter, full_type
+            tree_out_limit, attributes, attributes_filter, extras, extras_filter, full_type, profile
         )
 
     def parse_query_string(self, query_string):
@@ -680,7 +688,7 @@ class Utils:
 
         ## Define grammar
         # key types
-        key = Word(f'{alphas}_', f'{alphanums}_')
+        key = Word(f'{alphas}_', f'{alphanums}_-')
         # operators
         operator = (
             Literal('=like=') | Literal('=ilike=') | Literal('=in=') | Literal('=notin=') | Literal('=') |
--- a/aiida/cmdline/commands/cmd_restapi.py
+++ b/aiida/cmdline/commands/cmd_restapi.py
@@ -45,7 +45,8 @@ from aiida.restapi.common import config
     help='Enable POST endpoints (currently only /querybuilder).',
     hidden=True,
 )
-def restapi(hostname, port, config_dir, debug, wsgi_profile, posting):
+@click.pass_context
+def restapi(ctx, hostname, port, config_dir, debug, wsgi_profile, posting):
     """
     Run the AiiDA REST API server.
 
@@ -58,6 +59,7 @@ def restapi(hostname, port, config_dir, debug, wsgi_profile, posting):
     # Invoke the runner
     try:
         run_api(
+            profile=ctx.obj['profile'].name,
             hostname=hostname,
             port=port,
             config=config_dir,
--- a/aiida/manage/configuration/schema/config-v9.schema.json
+++ b/aiida/manage/configuration/schema/config-v9.schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema",
-  "description": "Schema for AiiDA configuration files, format version 8",
+  "description": "Schema for AiiDA configuration files, format version 9",
   "type": "object",
   "definitions": {
     "options": {
@@ -123,6 +123,12 @@
           "default": 5,
           "minimum": 1,
           "description": "Maximum number of transport task attempts before a Process is Paused."
+        },
+        "rest_api.profile_switching": {
+          "type": "boolean",
+          "default": false,
+          "description": "Toggle whether the profile can be specified in requests submitted to the REST API",
+          "global_only": true
         },
         "rmq.task_timeout": {
           "type": "integer",
--- a/aiida/restapi/resources.py
+++ b/aiida/restapi/resources.py
@@ -14,7 +14,7 @@ from flask import make_response, request
 from flask_restful import Resource
 
 from aiida.common.lang import classproperty
-from aiida.manage import load_profile
+from aiida.manage import get_config_option, load_profile
 from aiida.restapi.common.exceptions import RestInputValidationError
 from aiida.restapi.common.utils import Utils, close_thread_connection
 from aiida.restapi.translator.nodes.node import NodeTranslator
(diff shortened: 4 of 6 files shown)


sphuber:
@giovannipizzi @ltalirz this is a quick mockup of have we could make the REST API capable of dynamically loading and changing profiles. I have tested this manually (a bit) and it seems to work. I tried switching from django to django, sqla to sqla, and from django to sqla (and vice versa). Before trying to write unit tests (and perhaps cleaning up the implementation) I would be curious to see if this would work for some of the materials cloud deployments. Would it be possible to try and change a MC dev enviroment that now is using one REST API instance per profile, to simply have a single one serve all profiles and provide the `profile` query parameter in requests?

giovannipizzi:
Thanks @sphuber, seems super useful. Pinging @elsapassaro - is it easy to try this on some dev server of Materials Cloud, with 2 profiles?

@sphuber - if I understand correctly, you now specify profile=... in the URL, right? I'm wondering if this is how we would use it in Materials Cloud (maybe yes? @elsapassaro , @ltalirz , feedback welcome). 
In any case, it would be good to have a way (for Materials Cloud but also in general) to limit which profiles are actually exposed on the REST API. E.g. either with some command line options when you start the REST API (e.g. you start only for the current one by default, and you can specify more profiles (or `--all`) in the command line `verdi` command). Or by having some flags in the `config.json` file (`exposed_via_rest_api=true`) or some combination of them (and we should pick good defaults, where user don't get unexpected behaviour - e.g. running the API and not seeing any profile is bad, but also running the API and exposing more than what they thought is equally not good IMO).
This might be useful both on Materials Cloud (e.g. to expose a subset of DBs we control) but also for users (who might want to give access to others to only one or a few of their profiles, but not to some of the others that might contain confidential data).

aiida/restapi/common/utils.py
```diff
@@ -483,6 +483,7 @@ def build_translator_parameters(self, field_list):
         extras = None
         extras_filter = None
         full_type = None
+        profile = None
```
ltalirz, line 486: `load_profile(None)` defaults to loading the default profile.

I think the correct default here would be the profile that was used in
```verdi -p profile restapi```

**Comment by sphuber** [comment]
@giovannipizzi Indeed, the profile can now be defined as a query parameter in the URL. I don't really see another way how to do it, if the goal is for clients to specify a particular profile. I also think that having an option to restrict the profiles that can be exposed when launching the REST API is a good thing. I am not too familiar with it, but probably the best would be to do this through the config, which can then potentially be exposed through the CLI to be dynamically changed. However, before implementing those additional details I wanted to first check if the current approach satisfies the use case and whether it works robustly. I don't have an idea yet of the overhead of switching profiles under load and if we need (if we can even) optimize things here.

**Comment by ltalirz** [comment]
Thanks a lot @sphuber for this!

>  I tried switching from django to django, sqla to sqla, and from django to sqla (and vice versa).

Somehow I was still under the impression that switching from django to sqla was not safe... but in this case it is?

On materials cloud, we currently run the wsgi daemon processes with several threads - what happens in that case?
Is there a specific thing @elsapassaro should test?

I agree that access to profiles would be configured via `verdi config`.

While I agree that this is the right way forward and that this would strongly reduce e.g. RAM use on Materials Cloud, I imagine that  profile switching can incur a significant wait time on the request level.
We should benchmark this and to see whether we can live with this overhead.

**Comment by sphuber** [comment]
It could still make sense to merge this, although I will refactor it first to use `allow_switch` of `load_profile` that was added after the recent storage loading refactoring.

File: aiida/restapi/common/utils.py (hunk 1 above)
Comment (sphuber) [reply aiida/restapi/common/utils.py:483]: I have kept this as is, because this is just the parsing of the query parameters and I don't think that should be bothered with determining the default profile. This is now instead handled in `BaseResource.load_profile` which knows what default to use if not specified in the query parameters.

**Comment by sphuber** [comment]
I have rebased the PR and now use the `allow_switch` argument of `load_profile` that was recently added. I think it would be good to add this for v2.0 since it is a fully backwards-compatible change and could solve a concrete problem. @ltalirz do you have the time to give this a quick look or should I ask someone else?

**Comment by chrisjsewell** [comment]
One question: have you considered what happens when multi-threading, and thread safety?

**Comment by chrisjsewell** [comment]
> I take it the allow_switch is not thread-safe (yet)?

Indeed, I would say it is not.
Currently, the restapi threads are "abusing" the sqlalchemy scoped session, for thread local connections: https://github.com/aiidateam/aiida-core/blob/ffedc8bfe1832292db6f7ddd4e332961c91eb5ac/aiida/restapi/common/utils.py#L822-L835

This is basically the only part of AiiDA still "hard-coded" to a backend.

Note, sqlalchemy scoped sessions, simply use a `threading.local` to create the session: https://github.com/sqlalchemy/sqlalchemy/blob/03989d1dce80999bb9ea1a7d36df3285e5ce4c3b/lib/sqlalchemy/util/_collections.py#L658

since after #5330, there is now only `Manager` and `Config` are global variables, I was thinking to potentially support threading via making thread local: https://github.com/aiidateam/aiida-core/blob/ffedc8bfe1832292db6f7ddd4e332961c91eb5ac/aiida/manage/manager.py#L33-L38

Although this would be slightly different from having a session as thread local, since you would also have an `sqlalchemy.Engine` per thread

**Comment by sphuber** [comment]
But for the current REST API it probably is not a problem though since none of its end-points allow to mutate state. They are `get` operations only, so problems related to threading should be minimal if not existing, right?

**Comment by eimrek** [comment]
Hi, I tested this a little bit on materials cloud.

Two parts:

**1. general usage**

Currently, we set up a separate WSGI for each aiida profile with a script containing `load_profile`, e.g. this one:

```
# wsgi script for profile mc3d
from aiida.manage.configuration import load_profile
from aiida.restapi.run_api import configure_api
from aiida.restapi.api import App

load_profile('mc3d')

api = configure_api(flask_app=App)
application = api.app
```

which are then served at corresponding URL path of apache (e.g. `/mc3d/`)

With this PR, this `load_profile` ceases to work and always reverts to the default aiida profile. I would propose that it would be good to keep this working (i.e. by setting up the wsgi file this way, the corresponding profile is used).

**2. benchmarking**

I used a simple script using `requests` to benchmark this PR with the old implementation. (here's the notebook: [mc-rest-benchmark.zip](https://github.com/aiidateam/aiida-core/files/9093564/mc-rest-benchmark.zip))

I tested on profiles `pyrene-mofs` (total of 56'832 nodes), `mc3d` (total of 535'005 nodes) and `li-ion-conductors` (total of 4'505'808 nodes).

I accessed 50 uuids for each of these profiles (e.g. `/mc3d/api/v4/nodes/<uuid>`) for both the current implementation and this PR. 

In all cases, the average single access time is the same for all the profiles.

with the current (old) implemention, the average access time for one uuid is around **0.30 seconds**.

with the implementation of this PR, the average access time for one uuid 

* if accessed sequentially the same profile: **0.34 seconds**
* if alternating between two profiles (doesn't matter which two): **0.51 seconds**.

I guess this is something that needs consideration, as the worst case performance is 1.7 times slower.

Regarding memory usage, currently the wsgi daemons are taking about 40 MB each (6th column), which maybe is not too bad:

```
(aiida) ubuntu@dev-aiida-2:~/aiida-core$ ps aux | grep daemon-
ubuntu    221483  0.1  0.2 430120 44408 ?        Sl   15:51   0:00 daemon-mc3d       -k start
ubuntu    221484  0.1  0.2 429416 43504 ?        Sl   15:51   0:00 daemon-mc2d       -k start
ubuntu    221485  0.1  0.2 429416 43584 ?        Sl   15:51   0:00 daemon-2dtopo     -k start
ubuntu    221486  0.1  0.2 429416 43584 ?        Sl   15:51   0:00 daemon-tc-applica -k start
ubuntu    221487  0.1  0.2 429416 43588 ?        Sl   15:51   0:00 daemon-sssp       -k start
ubuntu    221488  0.1  0.2 430120 44404 ?        Sl   15:51   0:00 daemon-pyrene-mof -k start
ubuntu    221489  0.1  0.2 429416 43584 ?        Sl   15:51   0:00 daemon-curated-co -k start
ubuntu    221490  0.1  0.2 429416 43584 ?        Sl   15:51   0:00 daemon-hcofs-co2  -k start
ubuntu    221491  0.1  0.2 363880 43604 ?        Sl   15:51   0:00 daemon-stoceriait -k start
ubuntu    221492  0.1  0.2 363880 43596 ?        Sl   15:51   0:00 daemon-scdm       -k start
ubuntu    221493  0.1  0.2 429416 43600 ?        Sl   15:51   0:00 daemon-tin-antimo -k start
ubuntu    221494  0.1  0.2 429416 43596 ?        Sl   15:51   0:00 daemon-optimade-s -k start
ubuntu    221495  0.1  0.2 429416 43600 ?        Sl   15:51   0:00 daemon-li-ion-con -k start
```

Although maybe i'm missing something about this.

Happy to run any other further tests, if needed.

eimrek (approved):

sphuber:
@eimrek I am pretty sure I think I have found the source of the performance regression. The parsing of the query parameters was happening twice. I have refactored it and tested locally and at least for me the performance is similar to `main` now. Would you like to perform one final benchmark?

eimrek:
hi @sphuber, ran the scripts again, and indeed the performance seems the same now! (Although, for some reason, both benchmarks now give roughly 0.34-0.35 sec per request.)

Good to merge from my side.

PR #5056, opened 2021-08-09
Database: minimize PostgreSQL logs for Django backend

### LINKED ISSUE #5055: Avoid PSQL errors being logged at every start of AiiDA
At every `load_profile`, AiiDA checks if it is an old 0.x version of a new 1.x version by looking at an (old) `tval` field. If the query fails, it knows it's 1.x.
However, this creates a lot of log messages in the PSQL log file that becomes very big after some months (and also can appear on the user screen if they didn't configure PSQL logging properly, as reported by an user (see screenshot below).

If we could avoid that those checks generate SQL errors (in 1.x, accepting that they might generate errors in 0.x instead if needed), by keeping the same logic, that would be great.

verdi_status

(From the error messages, I think the lines are https://github.com/aiidateam/aiida-core/blob/16b326b64fb70bc9b6eb1543e6343d8313556d55/aiida/backends/djsite/manager.py#L110 and https://github.com/aiidateam/aiida-core/blob/16b326b64fb70bc9b6eb1543e6343d8313556d55/aiida/backends/djsite/manager.py#L90). Maybe one could just invert the code in the try and except blocks? Mentioning @sphuber who I think originally wrote that code.

sphuber:
Fixes #5055 

For the Django backend, we are recording the current version of the
database schema manually in our own `DbSetting` table. We use this to
determine whether the database schema version is compatible with that of
the code that is installed.

Since the way the version and generation of the schema are stored change
over the years, we had to rely on a try-except method to determine which
way was the case. The original try was for the original way of v0.x of
`aiida-core`, and if that failed we would fall back to the modern method
of storing the info introduced in `aiida-core==1.0`. However, each time
the except was hit, meaning a programming error was encountered in the
PostgreSQL statement, a number of warning are logged to PostgreSQL. Not
only can they appear in stdout if no log file has been configed for the
PostgreSQL server (can be the case for conda installations) the log file
will also accumulate these necessary logs, which are produced everytime
the database version is checked, i.e., when a profile is loaded, which
is often.

It might not be trivial to get completely rid of these errors being
logged but we can invert the try-except logic. Assuming that the
majority of users are now on v1.0 or above, they would always be hitting
the except. By first checking for the schema-style of v1.0 instead, only
users of v0.x will be getting the logs, which should be negligible.

### DIFF
--- a/aiida/backends/djsite/manager.py
+++ b/aiida/backends/djsite/manager.py
@@ -87,14 +87,16 @@ class DjangoBackendManager(BackendManager):
         backend = get_manager()._load_backend(schema_check=False, repository_check=False)  # pylint: disable=protected-access
 
         try:
-            result = backend.execute_raw(r"""SELECT tval FROM db_dbsetting WHERE key = 'schema_generation';""")
-        except ProgrammingError:
             result = backend.execute_raw(r"""SELECT val FROM db_dbsetting WHERE key = 'schema_generation';""")
-
-        try:
-            return str(int(result[0][0]))
-        except (IndexError, TypeError, ValueError):
+        except ProgrammingError:
+            # If this value does not exist, the schema has to correspond to the first generation which didn't actually
+            # record its value explicitly in the database until ``aiida-core>=1.0.0``.
             return '1'
+        else:
+            try:
+                return str(int(result[0][0]))
+            except (IndexError, ValueError, TypeError):
+                return '1'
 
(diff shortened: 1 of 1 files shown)


sphuber:
Would be good to try and merge this soon so I can also include this in the v1.6.5 backport release that is planned for tomorrow.

giovannipizzi (approved):
Thanks for the quick fix!

Issue #3458, opened 2019-10-23
Plugins: Investigate official importlib.metadata for entrypoint discovery


dev-zero:
Python 3.8 introduced a built-in module for entrypoint discovery https://docs.python.org/3.8/library/importlib.metadata.html with a backport for <3.8 and 2.7 here: https://importlib-metadata.readthedocs.io/en/latest/

While there's nothing wrong with `pkg_resources` other than we need `reentry` to make it scale, we should do a test with `importlib.metadata` to see whether it would be fast enough for us without `reentry`.

Release notes from `pluggy` (`pytest`) and `tox` suggest that it performs better than `setuptool`s `pkg_resources`.

ltalirz:
Sounds great, I'll have a look at this.

ltalirz:
as part of this, one should also check whether the dependency on `fastentrypoints.py` can be removed (see `utils/fastentrypoints.py`)

ltalirz:
I started looking into this here:
https://github.com/ltalirz/aiida_core/tree/issue_3458_importlib_metadata

I'm using a basic in-memory cache to make sure we're loading the entry points just once, but still the cli is a bit less responsive.
not terribly, but noticeably.

ltalirz:
@dev-zero Do you happen to have an AiiDA installation on a magnetic hard disk, where you could give the two tests above a try?

ltalirz:
@giovannipizzi @sphuber 
So, it seems in my environment, the difference between using `reentry` and using the built-in entry-point mechanism is down from roughly an order of magnitude to roughly a factor of 4.

Would you mind giving the PR a try and let me know how the autocompletion speed feels to you?
https://github.com/ltalirz/aiida_core/tree/issue_3458_importlib_metadata

It's certainly a bit slower, but at some point we need to think whether this still warrants the hassle of maintaining reentry, and asking people to `reentry scan`.

sphuber:
> So, it seems in my environment, the difference between using `reentry` and using the built-in entry-point mechanism is down from roughly an order of magnitude to roughly a factor of 4.

For me personally the responsiveness of the CLI is very important. I use it all the time and the slowness of auto-completion really annoys me. A factor of 4 is huge in that context I feel and for me would definitely warrant the continued development of reentry and the annoyance of `reentry scan`.

[...]

PR #5058, opened 2021-08-10
♻️ REFACTOR: Remove `reentry` requirement

chrisjsewell:
Utilise the python 3.10 `select` API (https://docs.python.org/3.10/library/importlib.metadata.html), via [importlib-metadata](https://github.com/python/importlib_metadata).

~The code currently includes a fall back to the old API, since I have previously encountered an issue with old versions of importlib_metadata being installed on Conda: https://github.com/python/importlib_metadata/issues/308~

Also added typing to module and mypy type checking


chrisjsewell:
@ltalirz not sure what was happening with #4960, but yeh wanted to push this through

chrisjsewell:
http://www.quantum-espresso.org/ is currently down 😬

ltalirz:
Just to understand: are you building on top of #4960 or not?

Have all references to reentry been removed from the codebase?

chrisjsewell:
> Just to understand: are you building on top of #4960 or not?

no

> Have all references to reentry been removed from the codebase?

yes

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=40.8.0", "wheel", "reentry~=1.3", "fastentrypoints~=0.12"]
```
ltalirz, line 2: please mention in the eventual commit message that `fastentrypoints` is also being removed

ltalirz (commented):
I notice that you keep updating the PR.

I'll stop going through it now and come back later - please let me know once it's ready for review.

As for using the `importlib_metadata`: I haven't yet looked at your implementation, but I would just like to mention that in python 3.9 (that has the importlib.metadata in the standard library) using the `importlib_metadata` package was significantly slower, see https://github.com/aiidateam/aiida-core/issues/3458#issuecomment-846099826

I.e. if we are going to rely on this, we should make sure to use the fast implementation whenever it is available (maybe you already do this; I will check later).

pyproject.toml
chrisjsewell: just to check, is there a reason you know of why we would still need this? I believe the issue it solved is no longer an issue?

chrisjsewell:
> I notice that you keep updating the PR.

yeh sorry just minor fixes to the tests

chrisjsewell:
>  but I would just like to mention that in python 3.9 (that has the importlib.metadata in the standard library) using the importlib_metadata package as a drop-in replacement was significantly slower

the complication here though comes in that `importlib_metadata` now allows the new `select` API, which is only available in python 3.10. So is it still slower to use that API vs the old API with `importlib.metadata`?

chrisjsewell:
I would also note that from v4.3.0, there is a claimed performance improvement: https://github.com/python/importlib_metadata/pull/317

pyproject.toml
chrisjsewell: This is without fastentrypoints, so I guess it is now ok to remove:

```
#!/Users/chrisjsewell/Documents/GitHub/aiida_core_develop/.tox/py38-pre-commit/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'aiida-core','console_scripts','verdi'
import re
import sys

# for compatibility with easy_install; see #2198
__requires__ = 'aiida-core'

try:
    from importlib.metadata import distribution
except ImportError:
    try:
        from importlib_metadata import distribution
    except ImportError:
        from pkg_resources import load_entry_point


def importlib_load_entry_point(spec, group, name):
    dist_name, _, _ = spec.partition('==')
    matches = (
        entry_point
        for entry_point in distribution(dist_name).entry_points
        if entry_point.group == group and entry_point.name == name
    )
    return next(matches).load()


globals().setdefault('load_entry_point', importlib_load_entry_point)


if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(load_entry_point('aiida-core', 'console_scripts', 'verdi')())

```

setup.json
```diff
@@ -36,6 +36,7 @@
         "jinja2~=2.10",
         "jsonschema~=3.0",
         "kiwipy[rmq]~=0.7.4",
+        "importlib-metadata~=4.0",
```
ltalirz, line 39: we probably want to up that to force people to upgrade to more performant versions

chrisjsewell:
Just playing around with autocompletion on my computer, and the performance seems good

sphuber:
> Just playing around with autocompletion on my computer, and the performance seems good

You upped the time limit for the load time of verdi to 0.5. Is that still necessary after the update to a more recent `importlib.metadata`? In other words, is with the latest version the load time still slower compared to `reentry` or do we have comparable or better performance?

chrisjsewell:
> You upped the time limit for the load time of verdi to 0.5.

Haha you noticed my cheat, yeh a few of them were sneaking over (I added a run for all python versions)

- python 3.7: SUCCESS: loading time 0.37 at iteration 1 below 0.5
- python 3.8: SUCCESS: loading time 0.33 at iteration 1 below 0.5
- python 3.9: SUCCESS: loading time 0.42 at iteration 1 below 0.5

Comparing that to the latest run on develop (python 3.8): SUCCESS: loading time 0.34 at iteration 1 below

so not much in it

setup.json
chrisjsewell: Upped the pin to ~=4.3

pyproject.toml
chrisjsewell: Yeh so actually the current develop `verdi` script is:

```
#!/Users/chrisjsewell/Documents/GitHub/aiida-core-origin/.tox/py39-pre-commit/bin/python
# -*- coding: utf-8 -*-
# EASY-INSTALL-ENTRY-SCRIPT: 'aiida-core','console_scripts','verdi'
__requires__ = 'aiida-core'
import re
import sys

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(verdi())
```

I guess you need importlib_metadata installed to produce the new script?

chrisjsewell: So I re-added fastentrypoints: I guess that the second script, with the direct `verdi` import, is still sightly quicker

chrisjsewell:
> The code currently includes a fall back to the old API, since I have previously encountered an issue with old versions of importlib_metadata being installed on Conda: python/importlib_metadata#308

actually since https://github.com/conda-forge/importlib_metadata-feedstock/commit/c204ab79d1b488c05eaf4ac4e01e51b3f7540f7f, I think this is not a problem, so I will remove the try/except...

chrisjsewell:
ready for re-review @ltalirz cheers

.github/workflows/ci-code.yml
```diff
@@ -127,13 +126,18 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
```
ltalirz, line 127: this is a short run that you're multiplying here, right?

perhaps reduce the max runtime then

aiida/plugins/entry_point.py
```diff
@@ -69,7 +68,13 @@ class EntryPointFormat(enum.Enum):
 }
 
 
-def validate_registered_entry_points():  # pylint: disable=invalid-name
+def parse_entry_point(group: str, spec: str) -> EntryPoint:
```
ltalirz, line 71: should that perhaps be called `create_entry_point`?
how is this method used?

Calling it `parse` may be a bit confusing together with the `parse_entry_point_string` methods

aiida/plugins/entry_point.py
```diff
@@ -213,92 +217,75 @@ def load_entry_point(group, name):
     return loaded_entry_point
 
 
-def get_entry_point_groups():
+def get_entry_point_groups() -> Set[str]:
```
ltalirz, line 220: hm... should call perhaps also be cached?

I'm wondering a bit now what the policy here is for which function calls to cache and which not - I guess any expensive call to `eps()` should be cached.
It would be good to document this in the module docstring

ltalirz (changes requested):
thanks @chrisjsewell , just some minor requests

ltalirz:
I'll also test autocompletion locally now, just to be safe

ltalirz:
Hm... I'm seeing a quite significant slowdown:

`develop`
```
$ COMP_CWORD="2" COMP_WORDS="verdi data g" _VERDI_COMPLETE=complete-bash hyperfine --warmup 3 verdi
Benchmark #1: verdi
  Time (mean ± σ):     250.9 ms ±   3.1 ms    [User: 188.1 ms, System: 40.9 ms]
  Range (min … max):   247.7 ms … 256.8 ms    11 runs
```

`remove-reentry`
```
$ COMP_CWORD="2" COMP_WORDS="verdi data g" _VERDI_COMPLETE=complete-bash hyperfine --warmup 3 verdi
Benchmark #1: verdi
  Time (mean ± σ):     396.7 ms ±   4.7 ms    [User: 307.2 ms, System: 67.6 ms]
  Range (min … max):   391.7 ms … 404.6 ms    10 runs
```
I.e. autocompletion is about 150ms slower, which is noticeable (and these numbers seem to be quite reproducible).

This is unexpected, since the timings of the bare library calls were basically on par with reentry
```
00:17 $ python -mtimeit -n1 -r1 'import importlib_metadata'
1 loop, best of 1: 30.8 msec per loop
00:16 $ python -mtimeit -s 'import importlib_metadata as im' 'im.entry_points()'
50 loops, best of 5: 5.91 msec per loop
```

@chrisjsewell do you happen to have an idea what is going on here?

chrisjsewell:
> autocompletion is about 150ms slower, which is noticeable 

do you find it noticeable when actually using the tab completion though?

I'm a little sceptical of the timing method, since as I was reproducing it, I noticed that it was not feeding back that actually the cli was failing (since on mac you have to first do `autoload -Uz compinit` and `compinit`): so it gave the same times whethere or not there was an exception.
Also I was a little confused that you used `verdi data g`, given there are no data plugins starting with `g`?

aiida/plugins/entry_point.py
chrisjsewell: yeh I was just using the ones that were already there. I would think though that maybe `impotlib_metadata` would already do its own caching, although I guess it can't hurt 🤷

.github/workflows/ci-code.yml
chrisjsewell: sure ting

aiida/plugins/entry_point.py
chrisjsewell: well it was just a replacement of `EntryPoint.parse` in ` aiida/manage/database/integrity/plugins.py`, that is the only place it is used. I thought it would be better to centralise the logic into this module.

chrisjsewell:
But yeh, when I manually tried out both `develop` and this PR, I honestly could not notice any difference

.github/workflows/ci-code.yml
chrisjsewell, line 127:
```suggestion
    timeout-minutes: 15
```

aiida/plugins/entry_point.py
ltalirz: > I would think though that maybe `impotlib_metadata` would already do its own caching

I see, can you check?

ltalirz:
As for the test, I just copied the line by tiziano in #3458 - how do you suggest I change it?
We can also change the `g` but I guess for the timings it should not really matter whether we're looking for a command that exists or not.

I did, however, notice a difference before I started testing. Note that the timings may depend not only on the hardware but also on how many packages are installed in your environment.

chrisjsewell:
So yeh final times from me; reentry is faster by these numbers, but I think it's acceptable since I don't notice the difference when actually using the CLI

reentry
```
$ hyperfine --warmup 3 "verdi --help"                
Benchmark #1: verdi --help
  Time (mean ± σ):     388.4 ms ±  17.8 ms    [User: 303.9 ms, System: 78.2 ms]
  Range (min … max):   366.9 ms … 428.5 ms    10 runs

$ hyperfine --warmup 3 "verdi plugin list aiida.data"
Benchmark #1: verdi plugin list aiida.data
  Time (mean ± σ):      1.291 s ±  0.128 s    [User: 1.400 s, System: 0.311 s]
  Range (min … max):    1.153 s …  1.582 s    10 runs

$ hyperfine --warmup 3 "verdi data"
Benchmark #1: verdi data
  Time (mean ± σ):     440.4 ms ±  56.2 ms    [User: 342.1 ms, System: 88.3 ms]
  Range (min … max):   386.5 ms … 524.8 ms    10 runs

$ hyperfine --warmup 3 "verdi database summary"
Benchmark #1: verdi database summary
  Time (mean ± σ):      1.176 s ±  0.037 s    [User: 1.398 s, System: 0.271 s]
  Range (min … max):    1.145 s …  1.255 s    10 runs
```

importlib-metadata

```
$ hyperfine --warmup 3 "verdi --help"
Benchmark #1: verdi --help
  Time (mean ± σ):     578.3 ms ±  18.2 ms    [User: 433.9 ms, System: 136.3 ms]
  Range (min … max):   551.5 ms … 622.8 ms    10 runs

$ hyperfine --warmup 3 "verdi plugin list aiida.data"
Benchmark #1: verdi plugin list aiida.data
  Time (mean ± σ):      1.533 s ±  0.032 s    [User: 1.668 s, System: 0.375 s]
  Range (min … max):    1.484 s …  1.588 s    10 runs

$ hyperfine --warmup 3 "verdi data"
Benchmark #1: verdi data
  Time (mean ± σ):     728.6 ms ±   6.2 ms    [User: 565.5 ms, System: 156.0 ms]
  Range (min … max):   720.9 ms … 739.3 ms    10 runs

$ hyperfine --warmup 3 "verdi database summary"
Benchmark #1: verdi database summary
  Time (mean ± σ):      1.577 s ±  0.027 s    [User: 1.705 s, System: 0.371 s]
  Range (min … max):    1.534 s …  1.628 s    10 runs
```

Aug-11-2021 13-03-26

chrisjsewell:
As discussed in the aiida dev meeting: @sphuber, @giovannipizzi, @mbercx, @ramirezfranciscof, etc.
It would be good if you could find a little time to clone this PR locally and play around with CLI, to see if you notice any performance degradation. Perhaps by the next dev meeting, so we can discuss then if we are happy to merge this

chrisjsewell:
Thanks for testing @sphuber, yeh that’s my feeling; that it’s the lesser of two evils

ltalirz:
@chrisjsewell as mentioned, I was surprised to find this slowdown since it is not

Turns out the inefficient aspect of your implementation are the repeated calls to `eps()`.
The following patch recovers almost the performance of `reentry`, at least for me:

```diff
diff --git a/aiida/plugins/entry_point.py b/aiida/plugins/entry_point.py
index 5b7c22abc..8a5acbada 100644
--- a/aiida/plugins/entry_point.py
+++ b/aiida/plugins/entry_point.py
@@ -19,6 +19,8 @@ from typing import Any, Optional, List, Sequence, Set, Tuple
 from importlib_metadata import EntryPoint, EntryPoints
 from importlib_metadata import entry_points as eps

+EPS = eps()
+
 from aiida.common.exceptions import MissingEntryPointError, MultipleEntryPointError, LoadingEntryPointError

 __all__ = ('load_entry_point', 'load_entry_point_from_string', 'parse_entry_point')
@@ -194,7 +196,6 @@ def load_entry_point_from_string(entry_point_string: str) -> Any:
     return load_entry_point(group, name)


 def load_entry_point(group: str, name: str) -> Any:
     """
     Load the class registered under the entry point for a given name and group
@@ -218,19 +219,18 @@ def load_entry_point(group: str, name: str) -> Any:
     return loaded_entry_point


 def get_entry_point_groups() -> Set[str]:
     """
     Return a list of all the recognized entry point groups

     :return: a list of valid entry point groups
     """
-    return eps().groups
+    return EPS.groups


 def get_entry_point_names(group: str, sort: bool = True) -> List[str]:
     """Return the entry points within a group."""
-    all_eps = eps()
+    all_eps = EPS
     group_names = list(all_eps.select(group=group).names)
     if sort:
         return sorted(group_names)
@@ -245,7 +245,7 @@ def get_entry_points(group: str) -> EntryPoints:
     :param group: the entry point group
     :return: a list of entry points
     """
-    return eps().select(group=group)
+    return EPS.select(group=group)


 @functools.lru_cache(maxsize=None)
@@ -259,7 +259,7 @@ def get_entry_point(group: str, name: str) -> EntryPoint:
     :raises aiida.common.MissingEntryPointError: entry point was not registered

     """
-    found = eps().select(group=group, name=name)
+    found = EPS.select(group=group, name=name)
     if name not in found.names:
         raise MissingEntryPointError(f"Entry point '{name}' not found in group '{group}'")
     if len(found.names) > 1:
```

chrisjsewell:
cheers @ltalirz, how about 6689fd2 then?

aiida/plugins/entry_point.py
```diff
@@ -27,6 +27,11 @@
 ENTRY_POINT_STRING_SEPARATOR = ':'
 
 
+@functools.lru_cache(maxsize=1)
```
chrisjsewell, line 30: I prefer not to have mutable global variables, i.e. this instead of `EPS = eps()`

ltalirz: is this the recommended way of doing non-mutable global variables?

Anyhow, it's ok with me

ltalirz:
Thanks, I might perhaps suggest to re-add the caching at least to the `load_entry_point` function but perhaps that is fast.
In any case, I'm getting 280-300ms with this, so if there still is something to shave off, it isn't much.

Happy to approve once you revert the increase of the time limits for the performance tests

chrisjsewell:
ah thats a pain:

```
The conflict is caused by:
    The user requested importlib-metadata==4.6.3
    ipykernel 6.0.3 depends on importlib-metadata<4; python_version < "3.8.0"
```

are we ok to use ipykernel v5 for python 3.7

ltalirz:
ipykernel v5 would be ok with importlib-metadata v4 while ipykernel v6 is not?

chrisjsewell:
> ipykernel v5 would be ok with importlib-metadata v4 while ipykernel v6 is not?

yep: 72c99d3

ltalirz:
this requirement in ipykernel actually comes from argcomplete
https://github.com/ipython/ipykernel/issues/645

.github/system_tests/test_verdi_load_time.sh
```diff
@@ -10,7 +10,7 @@ VERDI=`which verdi`
 # tends to go towards ~0.8 seconds. Since these timings are obviously machine and environment dependent, typically these
 # types of tests are fragile. But with a load limit of more than twice the ideal loading time, if exceeded, should give
 # a reasonably sure indication that the loading of `verdi` is unacceptably slowed down.
-LOAD_LIMIT=0.4
+LOAD_LIMIT=0.5
```
chrisjsewell, line 13:
```suggestion
LOAD_LIMIT=0.4
```

chrisjsewell:
> Happy to approve once you revert the increase of the time limits for the performance tests

reverted

ltalirz (dismissed):

chrisjsewell:
> do we have until the next dev meeting or you are going to merge this today?

I would say that, now we have made this final improvement and removed any significant timing difference to reentry, there is no need to wait. Unless obviously you feel strongly we should?

ltalirz:
From my side I agree that there is no reason to wait anymore

chrisjsewell:
@ltalirz, if you have time, perhaps you could quickly pop a new comparison of the new vs reentry times on here? (I'm working on another branch at the mo, so can't)

ltalirz:
this branch
```
$ COMP_CWORD="2" COMP_WORDS="verdi data g" _VERDI_COMPLETE=complete-bash hyperfine --warmup 3 verdi
Benchmark #1: verdi
  Time (mean ± σ):     272.7 ms ±   2.3 ms    [User: 207.6 ms, System: 42.7 ms]
  Range (min … max):   269.8 ms … 277.9 ms    10 runs
```

`develop`
```
$ COMP_CWORD="2" COMP_WORDS="verdi data g" _VERDI_COMPLETE=complete-bash hyperfine --warmup 3 verdi
Benchmark #1: verdi
  Time (mean ± σ):     261.0 ms ±   9.9 ms    [User: 194.2 ms, System: 43.4 ms]
  Range (min … max):   249.0 ms … 281.0 ms    11 runs
```

This is really negligible and won't be noticed by users

giovannipizzi:
Thanks for the clarification, it was just since you asked for feedback and I didn't understand that the time difference has disappeared. If the time difference is negligible I'm ok with the merge.
Just to report, I also run on my computer and I get:

This branch:
```
Benchmark #1: verdi
  Time (mean ± σ):     387.7 ms ±  18.9 ms    [User: 315.7 ms, System: 65.9 ms]
  Range (min … max):   371.0 ms … 428.8 ms    10 runs
```
develop:
```
Benchmark #1: verdi
  Time (mean ± σ):     338.6 ms ±   5.1 ms    [User: 275.4 ms, System: 57.6 ms]
  Range (min … max):   331.4 ms … 346.4 ms    10 runs
```
So I'm OK to merge.

However, while looking into it with cProfile+snakeviz, I realised that a call to cmd_computer was taking a lot of time (~80ms) even when the command was not involved (e.g. when expanding "verdi data bands sho").

In this branch, if you comment out the whole content of the `cmd_computer.py` file, indeed the time goes down significantly from ~370ms to ~305ms.
If you just uncomment this single line:
```python
from aiida.transports import cli as transport_cli
```
Then the time goes up again.

If you want to give a look into this already now, great, otherwise I'm ok with merging this as is since this would be a further improvement, but in this case could you please open an issue mentioning this? (I also have the feeling that there might be at least one or two similar things one could improve, e.g. also cmd_archive seems to be taking some time, but the cProfile+snakeviz is quite complex to "parse" and I still don't know if it's in addition to the time of cmd_computer or not).

chrisjsewell:
thanks @giovannipizzi! I will indeed have a quick check

chrisjsewell:
> the cProfile+snakeviz is quite complex to "parse"

FYI maybe check out https://github.com/benfred/py-spy, which is a lot easier to use

sphuber (dismissed):
let's get this show on the road :rocket:

chrisjsewell:
commit message:

This commit replaces the use of `reentry`, for entry point loading,
with `importlib_metadata` and, in turn,
removes the requirement for users to run `reentry scan` after installations.

aiida-core makes heavy use of entry-points to define plugins.
The `reentry` package was introduced to load these plugins since,
at the time, the de facto `pkg_resources` method for using entry points
was too slow, in particular for responsive CLI usage.
This, however, came with the drawback that users must perform an extra step
to register the plugins before aiida-core can be used, or when new plugins are installed.

In recent years `importlib.metadata` and its backport `importlib_metadata`
has replaced `pkg_resources`, and as of https://github.com/python/importlib_metadata/pull/317
is now on a par with `reentry` for performance.

For now, we use `importlib_metadata` for all python versions,
so that we can use the new python 3.10 API and performance boosts.

chrisjsewell:
are you gonna be merging anymore?

sphuber:
> ah @sphuber you ** lol, merging commits when I'm almost finished running the branch update

Can't help it you're not quick enough, I updated the other branch before you updated this one.

But yeah, not merging anything in the near future.

PR #5060, opened 2021-08-10
Dependencies: update requirement `pyyaml~=5.4`

### LINKED ISSUE #5059: Update the `pyyaml` dependency requirement

sphuber:
Fixes #5059 

Earlier versions have critical security flaws that have been fixed in
`pyyaml==5.4`. Note that `plumpy` also needs to be upgraded to `0.20.0`
which adds support for this version of `pyyaml`.

The `UnsafeLoader` is replaced by the `Loader` which are identical, but
the former is only being kept as an alias for backwards compatibility
but it might be removed in future releases.


sphuber:
Tests are failing because of conda install which is waiting for this release: https://github.com/conda-forge/plumpy-feedstock/pull/54

setup.json
```diff
@@ -39,13 +39,13 @@
...
         "paramiko~=2.7,>=2.7.2",
-        "plumpy~=0.19.0",
+        "plumpy~=0.20.0",
         "pgsu~=0.2.0",
         "psutil~=5.6",
         "psycopg2-binary~=2.8.3",
         "python-dateutil~=2.8",
         "pytz~=2021.1",
-        "pyyaml~=5.1",
+        "pyyaml~=5.4",
```
csadorf, line 48: Are we losing compatibility with the older versions? Otherwise it might be fine to leave this *as is*.

sphuber: This update was not so much for the compatibility, as it was to resolve critical security issues. The whole point of upgrading to `5.4` was that it fixes some critical safety problems and we shouldn't really use the older versions.

csadorf: I don't consider it our responsibility to force users to update 3rd party software as long as we don't prevent them from doing so due to our constraints.

That said, I'm fine with going ahead with this.

sphuber: Then what is the point of upgrading most of the requirements? Not trying to be snarky, really curious and trying to figure out if we should change our dependency management policy. We do this regularly, upping the minimum requirement, but I think that for many of the minor dependencies, we can just keep the lower level requirement and support a higher major version at the same time. Still we do this, but than surely we should _definitely_ do this for upgrading minimum requirements for packages that pose potential security problems.

csadorf: There is in fact no point in updating our dependencies in this way in my opinion and if it is done in this way it is either not with my blessing or it was a mistake on my part.

We should make sure to update our *requirements* (the ones we test against and are specified in the requirements/*.txt files) and make sure that we are not incompatible with the latest versions, especially if those are fixing security fixes. Being compatible with PyYAML>5.1 was the primary point of this effort for me, not updating our dependency specification.

This policicy I've outlined would be in compliance with [AEP 002](https://github.com/aiidateam/AEP/tree/master/002_dependency_management) in my opinion.

sphuber: Just to come back to this, the reason why I thought what I described was the policy, is because that's what you did in these recent two commits: 6558c71de6d4e37d6b9f7ad2a7a792ffb4947a31 and 78ef6336423f0484e27c361aad4f2d9eaeb72946   In both you updated the minimum requirement in the `setup.json`. Was this because adding simultaneous support for the recent versions would have been to difficult?

csadorf: I think I had some reasoning when I made those changes (pytz uses calver and aldjemy was a major version bump), but the simple answer is that I simply didn't think it through when I made those changes.

PR #5061, opened 2021-08-10
♻️ REFACTOR: Make `__all__` imports explicit (add pre-commit)

chrisjsewell:
This PR adds a pre-commit hook which:

1. builds a tree of non-init modules and their `__all__` values
2. walks through `__init__.py` files and auto-generates the `*` imports of submodules and `__all__` list for all (recursive) children in these files
3. Returns non-zero if there are unreadable `__all__` and/or conflicting imports

This intends to address two related issues:

1. As discussed in https://github.com/aiidateam/aiida-core/pull/4996#issuecomment-887984000, the current `*` imports are incompatible with mypy, pylance and a number of other tools which utilise static analysis of the code.
2. The `*` imports make it extremely difficult to understand what is being exposed at each level of the API (#4558)

Currently, this method works best when there is no actual code present in the __init__ files, although I added `# END AUTO-GENERATED` to be able to leave `aiida/manage/configuration/__init__.py`, as the tests were failing when moving them to another module.

For a few other modules I moved the `__init__.py` content to a `main.py`

see if the the tests pass 🤞


utils/make_all.py
```diff
@@ -0,0 +1,166 @@
...
+            continue
+
+        path.write_text('\n'.join(new_content).rstrip() + '\n', encoding='utf8')
+
+    return non_unique
+
+
+if __name__ == '__main__':
+    _folder = Path(__file__).parent.parent.joinpath('aiida')
+    _skip = {
```
chrisjsewell, line 148: note also, here we specify children imports that should always be skipped

chrisjsewell:
all tests passing 🎉

utils/make_all.py
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# pylint: disable=simplifiable-if-statement,too-many-branches
+"""Pre-commit hook to add ``__all__`` installs to ``__init__`` files."""
```
ltalirz, line 4:
```suggestion
"""Pre-commit hook to add ``__all__`` imports to ``__init__`` files."""
```

utils/make_all.py
ltalirz: please document in a comment why these imports are being skipped (individually, if there are different reasons, or once if they are all skipped for the same reason - e.g. import clashes that should be fixed).

ltalirz (changes requested):
thanks @chrisjsewell 

As mentioned in the AiiDa meeting I think this will help us clean up the public API and be more mindful of what we expose going forward.

utils/make_all.py
```diff
@@ -0,0 +1,170 @@
...
+
+if __name__ == '__main__':
+    _folder = Path(__file__).parent.parent.joinpath('aiida')
+    _skip = {
+        'cmdline/params': ['arguments', 'options'],
+        'cmdline/utils': ['echo'],
+        'manage': ['tests'],
+        'orm': ['implementation'],
+        'orm/implementation': ['django', 'sqlalchemy', 'sql'],
+        'restapi': ['*'],
```
chrisjsewell, line 158:
```suggestion
        # skipped since some arguments and options share the same name
        'cmdline/params': ['arguments', 'options'],
        # skipped since the module and its method share the same name
        'cmdline/utils': ['echo'],
        # skipped since this is for testing only not general use
        'manage': ['tests'],
        # skipped since we don't want to expose the implmentation
        'orm': ['implementation'],
        # skipped since both implementations share class/function names
        'orm/implementation': ['django', 'sqlalchemy', 'sql'],
        # skip all since the module requires extra requirements
        'restapi': ['*'],
```

utils/make_all.py
chrisjsewell: done 👍

ltalirz: thanks! 

for the 'implementation' case - just to make sure I understand:
this is just a precaution, correct?
I.e. it would only take effect if we were to `from implementation import *`?

chrisjsewell: yep thats correct: I tried to keep it as close to the existing imports as possible for now, but obviously we can iterate on exactly what is exposed where

ltalirz: Ah, I didn't read the PR description carefully enough - I thought there might be some "priming" by the current imports, but what this PR does is to *impose* `from ... import *` for all submodules, correct?

I.e. there is no way to export a functionality only on aiida.x.y.z - it will always bubble up to aiida.x (unless it's in the skip list).
Is that correct?

ltalirz: In a way, it means that we are using `__all__` (anywhere in the source tree) to define AiiiDA's public API. 

That is is ok, but I think we should document this and get input from some others to check they agree

chrisjsewell: > Is that correct?

yep thats correct, and is basically what was already happening, just more manually 😬 (there was definitely places I think where people had just forgot to add certain imports)

> there is no way to export a functionality only on aiida.x.y.z - it will always bubble up to aiida.x (unless it's in the skip list)

currently yeh, although obviously now the basic code is in place, I could always make it fancier lol, for more fine-grained control

ltalirz (approved):
thanks @chrisjsewell , for me this can proceed.
please wait for at least one more review e.g. from @sphuber or @greschd before proceeding.


I understand that the idea is to have a follow-up PR to update `reference/api/public`, both in terms of creating the list programmatically, and in documenting that the public API is simply anything people put into `__all__` inside the AiiDA codebase (with exceptions defined at ...)

chrisjsewell:
Yeh so @sphuber @giovannipizzi @greschd, whoever lol, it would be good to get the review so I can move on to the next PRs (#5063, ##4996, ...)

aiida/cmdline/utils/echo.py
```diff
@@ -18,7 +18,7 @@
 
 __all__ = (
     'echo', 'echo_info', 'echo_success', 'echo_warning', 'echo_error', 'echo_critical', 'echo_highlight',
-    'echo_dictionary'
+    'echo_dictionary', 'VALID_DICT_FORMATS_MAPPING'
```
sphuber, line 21: Why does this need to be exported? I think this should remain internal.

aiida/common/log.py
```diff
@@ -155,9 +156,11 @@ def configure_logging(with_orm=False, daemon=False, daemon_log_file=None):
     """
     from logging.config import dictConfig
 
+    from aiida.manage.configuration import get_config_option
+
     # Evaluate the `LOGGING` configuration to resolve the lambdas that will retrieve the correct values based on the
     # currently configured profile. Pass a deep copy of `LOGGING` to ensure that the original remains unaltered.
-    config = evaluate_logging_configuration(copy.deepcopy(LOGGING))
+    config = evaluate_logging_configuration(get_logging_config())
```
sphuber, line 163: Should update the comment above since there no longer is a deep copy being made explicitly.

aiida/cmdline/params/options/__init__.py
```diff
@@ -8,590 +8,111 @@
...
-    '-N',
-    '--nodes',
-    'nodes',
-    type=types.NodeParamType(),
-    cls=MultipleValueOption,
-    help='One or multiple nodes identified by their ID or UUID.'
-)
-
-FORCE = OverridableOption('-f', '--force', is_flag=True, default=False, help='Do not ask for confirmation.')
+# yapf: disable
```
sphuber, line 14: Since we are removing all other code from the init's and just have the `__all__`, why do we need to disable and reenable `yapf`? I know that `yapf` would change the one resource per line to just try and fit as much on one line, but why is this not desirable?

aiida/orm/utils/load_funcs.py
```diff
@@ -0,0 +1,205 @@
+# -*- coding: utf-8 -*-
```
sphuber, line 1: Would rename this file `load_funcs.py` to `loaders.py`.

sphuber (changes requested):
Thanks @chrisjsewell . Just some minor changes and questions and this would be good to go for now. There is one bigger question, but that would be for another PR. I think we are exposing too much right now, at least there are quite a few things that were never intended to be public facing API, and since we are now declaring that anything that is present in the init's of the second level modules, we will have to revise this. But as I said, I think we can address this in a separate PR, but this should be done before the 2.0 release. In that PR, we should also not forget to update the docs, which contains a hand-written page with public API resources. this should be removed.

aiida/cmdline/utils/echo.py
chrisjsewell: Its used here: https://github.com/chrisjsewell/aiida_core/blob/e3b386196eb9d5b4be67b4bf979210269eca1900/aiida/cmdline/params/options/main.py#L581

aiida/cmdline/params/options/__init__.py
chrisjsewell: Well firstly IMO, it makes it less readable, but mainly I think it would make the pre-commit hooks clash, since each would keep changing them back and forth

aiida/orm/utils/load_funcs.py
chrisjsewell: Ha, so there is already a loader.py in that module. Open to other suggestions though

aiida/common/log.py
```diff
@@ -155,9 +156,11 @@ def configure_logging(with_orm=False, daemon=False, daemon_log_file=None):
     """
     from logging.config import dictConfig
 
+    from aiida.manage.configuration import get_config_option
+
     # Evaluate the `LOGGING` configuration to resolve the lambdas that will retrieve the correct values based on the
     # currently configured profile. Pass a deep copy of `LOGGING` to ensure that the original remains unaltered.
```
chrisjsewell, line 162:
```suggestion
    # currently configured profile.
```

aiida/common/log.py
chrisjsewell: done

chrisjsewell:
> I think we are exposing too much right now

Yep agree; in this PR I was just trying to keep ~parity with the existing installs, but we can certainly do that

chrisjsewell:
cheers @sphuber, addressed your comments

chrisjsewell:
I've removed  #4558 from the linked PRs, to keep it open for the follow PR

aiida/orm/utils/load_funcs.py
sphuber: Ah yeah of course, my bad. But perfect, that is where they should go. I would just put them at the top of that file.

aiida/cmdline/utils/echo.py
sphuber: But the fact that it is used internally by the module does not mean it should be in the `__all__`. Even if it is not there, it can still be imported without any issue. So I would really just revert this change as I don't think this should necessarily be exposed at the top level.

aiida/common/log.py
```diff
@@ -51,74 +49,77 @@ def filter(self, record):
...
-            'formatter': 'halfverbose',
-            'filters': ['testing']
-        },
-    },
-    'loggers': {
-        'aiida': {
-            'handlers': ['console'],
-            'level': lambda: get_config_option('logging.aiida_loglevel'),
-            'propagate': False,
+def get_logging_config():
```
ramirezfranciscof, line 52: Are the changes in this file necessary for implementing the automation of `__all__`?

setup.json
```diff
@@ -95,7 +95,7 @@
         ],
         "pre-commit": [
             "astroid<2.5",
-            "mypy==0.790",
+            "mypy==0.910",
```
ramirezfranciscof, line 98: Is this dependency bump necessary for implementing the automation of `__all__`?

ramirezfranciscof (approved):
Hey @chrisjsewell , amazing refactor! This makes the code feel much more organized, besides enabling the linting features you already mentioned. For me this would be good to merge once @sphuber remaining comments are addressed.

Perhaps there is only one thing I would point out: all this PR is almost purely the inclusion of the script, the moving around of content of the `__init__` files to `main` files or the like (I understand you didn't modify any of this content), and the automated `__all__` writting. There are however two modifications of content that I'll point out below. Are these **strictly** necessary? If they are just convenient, I would try to put them in a separate PR that we can review very quickly after this one. I hope it doesn't come out as too pedantic, but I just feel that mixing content modification on a PR that is heavily moving content may "hide" it if we need to revisit the changes at some point.

ramirezfranciscof:
> > I think we are exposing too much right now
> 
> Yep agree; in this PR I was just trying to keep ~parity with the existing installs, but we can certainly do that

What would be the best way to do this later? We would need to use this kind of "exceptions" list in your automated code ([this](https://github.com/aiidateam/aiida-core/pull/5061/files#diff-ec40fb414dedf454784c1883820e475b462ac23c50c6b38dcf77c99a373f89a0R152)) or maybe if we are going to increase significantly the number of "unexposed" methods we need to find a cleaner way to do it (which is perhaps "positive" rather than "negative", i.e. we specify what we want to include instead of what we want to skip)?

Another question: does this now automatically clean up the autogenerated docs? Or if not, is there a way in which we could adapt it so that it does (so there is a specific section of the docs with only the exposed public API that is automatically generated)?

aiida/cmdline/utils/echo.py
chrisjsewell: done 👍

aiida/orm/utils/load_funcs.py
chrisjsewell: done 👍

aiida/cmdline/params/options/__init__.py
chrisjsewell: In the spirit of this PR, I'm happy with the explicit nature of using `# yapf: disable`, rather than relying on some obscure interpretation of commas by yapf. So we'll agree to disagree on this one 😉

setup.json
chrisjsewell: Well, one of the primary purposes of this PR was to get the new version of mypy working, so I wanted to make sure that was actually the case

aiida/common/log.py
chrisjsewell: Yeh otherwise there were messy circular dependency errors.
Two key points here are (a) top-level imports should always flow from `aiida.common` to other modules (since its meant to be a common module) and (b) you should always try to avoid mutable global variables

chrisjsewell:
> If it is not in the __all__ it won't be bubbled u

Yeh so I would say we are indeed being "positive" rather than "negative", since you specify positively what you want exposed by adding it to its files `__all__`

PR #5064, opened 2021-08-11
FIX: ``InteractiveOption`` Remove double call of normal ``callback``

### LINKED ISSUE #5023: Callback always called twice when providing ``prompt_fn`` and ``callback`` to ``InteractiveOption``
### Describe the bug

Using both the ``prompt_fn`` and ``callback`` arguments in the ``InteractiveOption`` leads to unexpected additional calls of the ``callback`` function.

### Steps to reproduce

```python
from aiida.cmdline.params.options.interactive import InteractiveOption

def prompt_callback(ctx):
    print('prompt callback')
    return True

def callback(ctx, param, value):
    print('callback')

@click.command('test')
@click.option('--label',
              cls=InteractiveOption,
              prompt='Label',
              prompt_fn=prompt_callback,
              callback=callback)
def cmd_test(label):
    print(label)
```
Output:

```
prompt callback
Info: enter "?" for help
Label: test
callback
callback
test
```

### Expected behavior

The callback should be called only once after the for each prompt

```
prompt callback
Info: enter "?" for help
Label: test
callback
test
```

### Your environment

- Operating system [e.g. Linux]: MacOs
- Python version [e.g. 3.7.1]: 3.7.7
- aiida-core version [e.g. 1.2.1]: develop branch

click version 7.1.2

janssenhenning:
Fixes #5023.

In the ``InteractiveOption`` providing a normal callback is wrapped
in the method ``after_callback`` while the actual callback is replaced
by ``prompt_callback``. Now the following situation occurs when the
parameter should be prompted for:

   1. ``prompt_loop`` is entered via ``prompt_callback`` and the value
      is converted in ``safely_convert``
   2. ``safely_convert`` also includes a call to
      ``prompt_callback``, where a call to ``after_callback`` is
      performed
   3. After a value succeeds the code goes back into the place, where
      the ``prompt_loop`` was entered in ``prompt_callback``
      and continues with an additional call to ``after_callback``
      which already executed before

The fix here is to directly return after the prompt_loop finishes and
handle the case, where the value should not be prompted separately


tests/cmdline/params/options/test_interactive.py
```diff
@@ -95,6 +110,20 @@ def test_callback_prompt_twice(self):
...
+        behaviour: the callback should be called at most once per prompt
+        """
+        cmd = self.simple_command(type=float, callback=validate_positive_number_with_echo)
+        runner = CliRunner()
+        result = runner.invoke(cmd, [], input='string\n-1\n-1\n1\n')
+        self.assertIsNone(result.exception)
+        lines = result.output.split('\n')
+        #The callback should be called once per prompt
+        #where type conversion was successful
+        self.assertEqual(lines.count('Validating Number'), 3)
```
sphuber, line 125: With the change of including the value being passed in the echo string, I think we can make the test more robust and make the changes on false positives smaller. Since the echo should be the only output (right?) we should also be able to check the entire output verbatim. So I would suggest something like the following:

```suggestion
        self.assertEqual(result.output, 'Validating string\nValidating -1\nValidating 1\n')
```

tests/cmdline/params/options/test_interactive.py
```diff
@@ -95,6 +110,20 @@ def test_callback_prompt_twice(self):
...
         self.assertIn(expected_4, lines[12])
 
+    def test_callback_prompt_only_once(self):
+        """
+        scenario: using InteractiveOption with type=float and callback that echos an additional message
+        behaviour: the callback should be called at most once per prompt
+        """
+        cmd = self.simple_command(type=float, callback=validate_positive_number_with_echo)
+        runner = CliRunner()
+        result = runner.invoke(cmd, [], input='string\n-1\n-1\n1\n')
```
sphuber, line 120: If you accept my other suggestions, the input should also be updated.
```suggestion
        result = runner.invoke(cmd, [], input='string\n-1\n1\n')
```

sphuber (commented):
Thanks for your contribution @janssenhenning . The solution looks fine at first look. Just have a small suggestion for the test that I think makes the chances for false positives smaller and is slightly easier to follow as to what is happening. Let me know what you think.

tests/cmdline/params/options/test_interactive.py
janssenhenning: I agree that the test can be made more robust by including the value in the echo, but the change will not work as is, since for example ``'Validating string'`` will not appear in the output. This value will fail when it is converted to a float so the callback will never be called.

sphuber: Ah I see, because it would already fail the validation at the type which is expecting a float. Well, by accident, I think that is actually good to keep the `string` value in, as it tests that that _doesn't_ hit the callback as it shouldn't. So maybe all you have to change in my suggestion is just to remove `Validate string\n` from the expected output

janssenhenning: I will probably need to check specific lines of the output, since there is more stuff in ``result.output`` besides the ``'Validating {value}'`` lines like the lines for entering the values for example but in principle yes

sphuber (approved):
Thanks a lot @janssenhenning

PR #5067, opened 2021-08-11
ORM: deprecate double underscores in `LinkManager` contains

### LINKED ISSUE #4953: Back incompatible changes introduced in #4625
With #4625, the use of double underscore to access nested namespaces has been deprecated. While the direct call using the double underscore (for instance `node.inputs.nested__subspace`) has been maintained (with correct deprecation), the functionality of interrogating is presence in the inputs/outputs is not maintained.
This means that `"nested__subspace" in node.inputs` returns `True` before the introduction of this commit and `False` after.
On the other hand, we now have `"nested" in node.inputs` returning `True` even if there is no incoming link with this name (for instance `node.get_incoming(link_label_filter="pseudos").all()` returns nothing).

sphuber:
Fixes #4953 

The use of double underscores in the interface of the `LinkManager` was
recently deprecated for v2.0, however, it unintentionally broke the
`__contains__` operator. Legacy code that was using code like:

    if 'some__nested__namespace' in node.inputs

which used to work, will now return false, breaking existing code. The
solution is to override the `__contains__` operator and check for the
presence of a double underscore in the key. If that is the case, now a
deprecation warning is emitted, but the key is split on the double
underscores and the namespaces are used to fetch the intended nested
dictionary before applying the actual contains check on the leaf node.

### DIFF
--- a/aiida/orm/utils/managers.py
+++ b/aiida/orm/utils/managers.py
@@ -12,9 +12,12 @@ Contain utility classes for "managers", i.e., classes that allow
 to access members of other classes via TAB-completable attributes
 (e.g. the class underlying `calculation.inputs` to allow to do `calculation.inputs.<label>`).
 """
+import warnings
+
 from aiida.common import AttributeDict
 from aiida.common.links import LinkType
 from aiida.common.exceptions import NotExistent, NotExistentAttributeError, NotExistentKeyError
+from aiida.common.warnings import AiidaDeprecationWarning
 
 __all__ = ('NodeLinksManager', 'AttributeManager')
 
@@ -24,6 +27,8 @@ class NodeLinksManager:
     A manager that allows to inspect, with tab-completion, nodes linked to a given one.
     See an example of its use in `CalculationNode.inputs`.
     """
+
+    _namespace_separator = '__'
 
     def __init__(self, node, link_type, incoming):
         """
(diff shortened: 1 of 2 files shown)


ramirezfranciscof:
Just a small comment: the breaking change was released with [v1.6.2](https://github.com/aiidateam/aiida-core/commits/v1.6.2) (see "Commits on Apr 26, 2021") so perhaps it would be nice if we could add this to the [v1.6.5 release branch](https://github.com/aiidateam/aiida-core/commits/release/1.6.5) you were preparing. Maybe this was your original idea, but since you only mention 2.0 in the OP I just wanted to mention this explicitly.

aiida/orm/utils/managers.py
```diff
@@ -142,6 +148,25 @@ def __getattr__(self, name):
                 f"Node<{self._node.pk}> does not have an {prefix} with link label '{name}'"
             ) from exception
 
+    def __contains__(self, key):
+        """Override the operator of the base class to emit deprecation warning if double underscore is used in key."""
+        if self._namespace_separator in key:
+            warnings.warn(
+                'The use of double underscores in keys is deprecated. Please expand the namespaces manually.',
```
ramirezfranciscof, line 155: "Please expand the name manually" is correct but I'm not sure it will be clear enough for people who don't have the context of this change? I would maybe add an example like in the warning above.

```suggestion
                'The use of double underscores in keys is deprecated. Please expand the namespaces manually: '
                'instead of `nested__name in node.inputs`, use `nested in node.inputs` first and `name in node.inputs.nested` afterwards',
```

ramirezfranciscof (approved):
Great fix! With all paths being tested and all! Only have one comment regarding the clarity of the warning message, but it LGTM.

aiida/orm/utils/managers.py
sphuber: Thanks I have included a slightly different variant of your suggestion

sphuber:
Yes, this should also go in the backport release.

PR #5073, opened 2021-08-12
Entry points: add the `core.` prefix to all entry points

### LINKED ISSUE #4914: add 'core' prefix to entry points defined by aiida-core
In order to be in line with rules for plugins and to avoid polluting the global namespace, the entry points defined by aiida-core should be prefixed by 'core.', e.g. `core.structure`.

See also discussion [here](https://github.com/aiidateam/aiida-registry/issues/60#issuecomment-532757533)

Planned for v2.0.0

sphuber:
Fixes #4914

This PR updates all* entry points that ship with `aiida-core` and prepends the `core.` entry point namespace, as to conform with our own convention. The benefit is not just to be consistent but it will also make it easy to determine exactly where all entry points come from, which may not always have been trivial without the prefix.

The PR includes full database migrations and archive migrations to update all the relevant versions of the entry point names as they are stored on `Computer` and `Node` instances. This makes sure that existing data is fully backwards compatible. As for backwards compatibility of code: all factories include a check for the old entry points and will emit a deprecation warning and will then load the corresponding entry point from the new entry point string. This means that code using the factories is fully backwards compatible. There are a few places in the CLI that are not backwards compatible unfortunately. For example when using `-S slurm` when setting up a computer, one now has to update this to use the new entry point, i.e., `-S core.slurm`. The same goes for REST API queries where query parameters that used `transport_type=local` now should use `transport_type=core.local`. Ultimately, I think this is worth it in the long run.


giovannipizzi:
Thanks @sphuber!
I let the others do a review, just a couple of quick comments, if not done already
- the tutorials need to be updated. Please open an issue in that repository, and add a card/issue for the 2.0 mileston for this (we should fix the tutorials for 2.0 roughly at the same time of the 2.0 release, pinging @mbercx for this)
- We should add a note to https://github.com/aiidateam/aiida-core/wiki/AiiDA-2.0-plugin-migration-guide (this should, I believe, eventually become a guide not only for developers but also for users moving to 2.0?)

sphuber:
>     * We should add a note to https://github.com/aiidateam/aiida-core/wiki/AiiDA-2.0-plugin-migration-guide (this should, I believe, eventually become a guide not only for developers but also for users moving to 2.0?)

Indeed, this is up to date so far with the state of `develop`. If this gets merged I will make sure to add an entry in the wiki.

ltalirz:
Thanks @sphuber !

I have two main questions:

 1. Do we need to deprecate the old strings at all? 
    My gut feeling tells me that most users would intuitively be against the new strings, simply because they are longer (but I'd be happy to be convinced otherwise). 
    I guess we want the new strings for reasons of internal consistency but we might be able to do that without the user having to care about it (at least so far we haven't had an actual case of namespace overlap to my knowledge).

 2. Is now the right time for deprecating the strings?
     I guess we all agree that we won't remove support with v2.0. However, as a diligent plugin developer you don't want your plugin to raise deprecation warnings, and so you would have to update all entry-point strings in their code now.
     If there is ever going to be a move of certain data types outside aiida-core (structuredata), perhaps that would be a better time to deprecate entry point strings so to only cause this disruption once.

sphuber:
>  Do we need to deprecate the old strings at all?
>  My gut feeling tells me that most users would intuitively be against the new strings, simply because they are longer (but I'd be happy to be convinced otherwise).

We kind of have to deprecate them because we cannot keep both at all levels. We can keep the old strings working in the factories (which is already implemented) but for some of the entry points (the calcjobs, workflows, parsers, schedulers and transports) the entry point name is stored on the relevant `Node` or `Computer` in the database. This is used to reload the corresponding class. The migration that I added takes care to migrate existing attributes. So even though some parts of the interface may support both temporarily, the data will have to adopt a single one anyway.

>  I guess we want the new strings for reasons of internal consistency (just to remind me, what is the main problem we're solving here?) but we might be able to do that without the user having to care about it (at least so far we haven't had an actual case of namespace overlap to my knowledge).  Of course, having two sets of strings side-by-side is more complicated than just having one, so there is a downside.

I described the benefits in the OP of the PR. This is to be consistent, conformant with our own guidelines and it will make it more obvious where certain plugins come from. The ecosystem might be small enough now where that is not (yet) difficult to say, but in the future, things might not be so obvious. The explicitness of the `core` prefix is very valuable I think.

> Is now the right time for deprecating the strings? I guess we all agree that we won't remove support with v2.0. However, as a diligent plugin developer you don't want your plugin to raise deprecation warnings, and so you would have to update all entry-point strings in your code now.

I don't agree. The whole point of deprecation warnings is to allow a grace period. The point is not to _never_ have them show up. If that were the case, we wouldn't even need a deprecation period and just work with the developers to adapt their code straight away to the changes before they release the next version. The changes necessary would also be minimal. The entry points are not used that often, but where ever they are used, they can be updated with relatively simple search-and-replace.

> If there is ever going to be a move of certain data types outside aiida-core (structuredata), perhaps that would be a better time to deprecate entry point strings without `core.` prefix as well so to only cause this disruption once.

The discussion of moving out the domain-specific data plugins has been brought up many times, even before 1.0 and we keep postponing it. At this point, we don't know when that is going to happen, so I'd prefer to do this now, so things are consistent and prepared.

sphuber:
Here follows a summary of the discussion and discussion taken at the meeting of Aug 30 2021.

Main argument against this change:
- It adds additional work for plugin developers because they will have to adapt their code to get rid of the deprecation warning.

The counter arguments to this point:
- This change is being added to a major revision, so deprecation warnings can be expected
- Plugins anyway have typically an upper limit on their `aiida-core` requirement to only support a single major version. When v2.0 is released, they will anyway have to make a new release in which case adding a small commit that quickly updates the entry point names with the correct prefix would be trivial.
- Plugins that are not actively maintained anymore and are unlikely to make a new release to deal with the warnings, will anyway not work for the new major version since there are some actual (however few) breaking changes, given that it is a major release. This is therefore not a sufficient reason not to go ahead with this change.

It was decided that with these counter arguments, the main downside of this change of adding some work for plugin developers does not outweigh the long term benefit of having correct and explicit entry point names. Therefore we have decided to go ahead with this change.

sphuber:
@chrisjsewell if we can get an official approval that would be great

ltalirz (approved):

chrisjsewell:
> @chrisjsewell if we can get an official approval that would be great

well better late than never

PR #5080, opened 2021-08-14
👌 IMPROVE: CLI responsiveness

### LINKED ISSUE #5070: Improvements for CLI load time
Thanks for the clarification, it was just since you asked for feedback and I didn't understand that the time difference has disappeared. If the time difference is negligible I'm ok with the merge.
Just to report, I also run on my computer and I get:

This branch:
```
Benchmark #1: verdi
  Time (mean ± σ):     387.7 ms ±  18.9 ms    [User: 315.7 ms, System: 65.9 ms]
  Range (min … max):   371.0 ms … 428.8 ms    10 runs
```
develop:
```
Benchmark #1: verdi
  Time (mean ± σ):     338.6 ms ±   5.1 ms    [User: 275.4 ms, System: 57.6 ms]
  Range (min … max):   331.4 ms … 346.4 ms    10 runs
```
So I'm OK to merge.

However, while looking into it with cProfile+snakeviz, I realised that a call to cmd_computer was taking a lot of time (~80ms) even when the command was not involved (e.g. when expanding "verdi data bands sho").

In this branch, if you comment out the whole content of the `cmd_computer.py` file, indeed the time goes down significantly from ~370ms to ~305ms.
If you just uncomment this single line:
```python
from aiida.transports import cli as transport_cli
```
Then the time goes up again.

If you want to give a look into this already now, great, otherwise I'm ok with merging this as is since this would be a further improvement, but in this case could you please open an issue mentioning this? (I also have the feeling that there might be at least one or two similar things one could improve, e.g. also cmd_archive seems to be taking some time, but the cProfile+snakeviz is quite complex to "parse" and I still don't know if it's in addition to the time of cmd_computer or not).

_Originally posted by @giovannipizzi in https://github.com/aiidateam/aiida-core/issues/5058#issuecomment-897515108_

chrisjsewell:
This commit improves the CLI loading speed,
by loading the `verdi computer configure` commands
(which in-turn load Transport plugins)
only when necessary.

The strictness of the `verdi devel check-load-time`
is also increased, to only allow loading of specific aiida modules.

closes #5070

### DIFF
--- a/aiida/cmdline/commands/cmd_computer.py
+++ b/aiida/cmdline/commands/cmd_computer.py
@@ -541,9 +541,6 @@ def computer_delete(computer):
 class LazyConfigureGroup(click.Group):
     """A click group that will lazily load the subcommands for each transport plugin."""
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-
     def list_commands(self, ctx):
         return list(get_entry_point_names('aiida.transports'))
 
--- a/aiida/cmdline/commands/cmd_computer.py
+++ b/aiida/cmdline/commands/cmd_computer.py
@@ -540,6 +540,7 @@ def computer_delete(computer):
 
 class LazyConfigureGroup(click.Group):
     """A click group that will lazily load the subcommands for each transport plugin."""
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
--- a/aiida/cmdline/commands/cmd_computer.py
+++ b/aiida/cmdline/commands/cmd_computer.py
@@ -21,7 +21,6 @@ from aiida.cmdline.utils import echo
 from aiida.cmdline.utils.decorators import with_dbenv
 from aiida.common.exceptions import ValidationError
 from aiida.plugins.entry_point import get_entry_point_names
-from aiida.transports import cli as transport_cli
 
 
 @verdi.group('computer')
@@ -539,7 +538,20 @@ def computer_delete(computer):
     echo.echo_success(f"Computer '{label}' deleted.")
 
 
-@verdi_computer.group('configure')
+class LazyConfigureGroup(click.Group):
+    """A click group that will lazily load the subcommands for each transport plugin."""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def list_commands(self, ctx):
+        return list(get_entry_point_names('aiida.transports'))
+
+    def get_command(self, ctx, name):
+        from aiida.transports import cli as transport_cli
+        return transport_cli.create_configure_cmd(name)
+
+
+@verdi_computer.group('configure', cls=LazyConfigureGroup)
 def computer_configure():
     """Configure the Authinfo details for a computer (and user)."""
 
@@ -595,7 +607,3 @@ def computer_config_show(computer, user, defaults, as_option_string):
             else:
                 table.append((f'* {name}', '-'))
         echo.echo(tabulate.tabulate(table, tablefmt='plain'))
-
-
-for ep_name in get_entry_point_names('aiida.transports'):
-    computer_configure.add_command(transport_cli.create_configure_cmd(ep_name))
--- a/aiida/cmdline/commands/cmd_computer.py
+++ b/aiida/cmdline/commands/cmd_computer.py
@@ -547,7 +547,7 @@ class LazyConfigureGroup(click.Group):
     def list_commands(self, ctx):
         return list(get_entry_point_names('aiida.transports'))
 
-    def get_command(self, ctx, name):
+    def get_command(self, ctx, name):  # pylint: disable=arguments-differ
         from aiida.transports import cli as transport_cli
         return transport_cli.create_configure_cmd(name)
 
@@ -569,6 +569,7 @@ def computer_configure():
 def computer_config_show(computer, user, defaults, as_option_string):
     """Show the current configuration for a computer."""
     from aiida.common.escaping import escape_for_bash
+    from aiida.transports import cli as transport_cli
 
     transport_cls = computer.get_transport_class()
     option_list = [
--- a/aiida/cmdline/commands/cmd_computer.py
+++ b/aiida/cmdline/commands/cmd_computer.py
@@ -19,7 +19,7 @@ from aiida.cmdline.params import options, arguments
 from aiida.cmdline.params.options.commands import computer as options_computer
 from aiida.cmdline.utils import echo
 from aiida.cmdline.utils.decorators import with_dbenv
-from aiida.common.exceptions import ValidationError
+from aiida.common.exceptions import ValidationError, EntryPointError
 from aiida.plugins.entry_point import get_entry_point_names
 
 
@@ -542,11 +542,17 @@ class LazyConfigureGroup(click.Group):
     """A click group that will lazily load the subcommands for each transport plugin."""
 
     def list_commands(self, ctx):
-        return list(get_entry_point_names('aiida.transports'))
+        subcommands = super().list_commands(ctx)
+        subcommands.extend(get_entry_point_names('aiida.transports'))
+        return subcommands
 
     def get_command(self, ctx, name):  # pylint: disable=arguments-differ
         from aiida.transports import cli as transport_cli
-        return transport_cli.create_configure_cmd(name)
+        try:
+            command = transport_cli.create_configure_cmd(name)
+        except EntryPointError:
+            command = super().get_command(ctx, name)
+        return command
 
 
 @verdi_computer.group('configure', cls=LazyConfigureGroup)
--- a/aiida/cmdline/commands/cmd_devel.py
+++ b/aiida/cmdline/commands/cmd_devel.py
@@ -33,13 +33,21 @@ def devel_check_load_time():
     """
     from aiida.manage.manager import get_manager
 
+    aiida_modules = [key for key in sys.modules.keys() if key.startswith("aiida.")]
+    aiida_modules_str = "\n- ".join(sorted(aiida_modules))
+    echo.echo(f"aiida modules loaded:\n- {aiida_modules_str}")
+
     manager = get_manager()
 
     if manager.backend_loaded:
         echo.echo_critical('potential `verdi` speed problem: database backend is loaded.')
 
-    if 'aiida.orm' in sys.modules:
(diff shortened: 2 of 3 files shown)


chrisjsewell:
Before and after PR:

```console
$ hyperfine --warmup 3 verdi
Benchmark #1: verdi
  Time (mean ± σ):     586.2 ms ±  69.7 ms    [User: 444.6 ms, System: 115.4 ms]
  Range (min … max):   509.7 ms … 693.6 ms    10 runs
 
$ hyperfine --warmup 3 verdi
Benchmark #1: verdi
  Time (mean ± σ):     450.0 ms ±  89.4 ms    [User: 330.5 ms, System: 88.8 ms]
  Range (min … max):   362.2 ms … 671.8 ms    10 runs

```

chrisjsewell:
and https://github.com/aiidateam/aiida-core/runs/3328880012 is down to 0.23s ⚡️

chrisjsewell:
Ok I'm gonna merge this soonish, because I think it's a no-brainer, and will stop the verdi tests failing on other PRs

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -20,26 +22,36 @@ def verdi_devel():
...
+        echo.echo(f'aiida modules loaded:\n- {aiida_modules_str}')
+
     manager = get_manager()
 
     if manager.backend_loaded:
         echo.echo_critical('potential `verdi` speed problem: database backend is loaded.')
 
-    if 'aiida.orm' in sys.modules:
-        echo.echo_critical('potential `verdi` speed problem: `aiida.orm` module is imported.')
+    allowed = ('aiida.backends', 'aiida.cmdline', 'aiida.common', 'aiida.manage', 'aiida.plugins', 'aiida.restapi')
```
ltalirz, line 49: I guess this is a constant, perhaps name it accordingly

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -20,26 +22,36 @@ def verdi_devel():
...
 
         * the database environment is loaded when it doesn't need to be
-        * the `aiida.orm` module is imported when it doesn't need to be
+        * Only expected `aiida.*` modules are imported
 
     If either of these conditions are true, the command will raise a critical error
     """
     from aiida.manage.manager import get_manager
 
+    aiida_modules = [key for key in sys.modules if key.startswith('aiida.')]
```
ltalirz, line 39: just to be more clear
```suggestion
    loaded_aiida_modules = [key for key in sys.modules if key.startswith('aiida.')]
```

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -20,26 +22,36 @@ def verdi_devel():
 
 
 @verdi_devel.command('check-load-time')
-def devel_check_load_time():
+@click.option('-v', '--verbose', is_flag=True, help='Increase verbosity')
```
ltalirz, line 25: I think there is a verbosity option one can simply import... anyhow we'll fix this soon hopefully

aiida/cmdline/commands/cmd_computer.py
```diff
@@ -556,6 +572,7 @@ def computer_configure():
 def computer_config_show(computer, user, defaults, as_option_string):
     """Show the current configuration for a computer."""
     from aiida.common.escaping import escape_for_bash
+    from aiida.transports import cli as transport_cli
```
ltalirz, line 575: just from the PR diff it looks like this is being introduced without being used (weird - shouldn't pylint catch this?)

aiida/cmdline/commands/cmd_computer.py
```diff
@@ -539,7 +538,24 @@ def computer_delete(computer):
     echo.echo_success(f"Computer '{label}' deleted.")
 
 
-@verdi_computer.group('configure')
+class LazyConfigureGroup(click.Group):
+    """A click group that will lazily load the subcommands for each transport plugin."""
+
+    def list_commands(self, ctx):
+        subcommands = super().list_commands(ctx)
+        subcommands.extend(get_entry_point_names('aiida.transports'))
```
ltalirz, line 546: Am I right in assuming that this will result in the list of subcommands being recreated every time `list_commands` is called?

Why not trigger`computer_configure.add_command` the first time either `list_commands` or `get_command` are called?

ltalirz (changes requested):
great, thanks a lot @chrisjsewell !

some comments/questions

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -20,26 +22,36 @@ def verdi_devel():
...
+def devel_check_load_time(verbose):
     """Check for common indicators that slowdown `verdi`.
 
     Check for environment properties that negatively affect the responsiveness of the `verdi` command line interface.
     Known pathways that increase load time:
 
         * the database environment is loaded when it doesn't need to be
-        * the `aiida.orm` module is imported when it doesn't need to be
+        * Only expected `aiida.*` modules are imported
```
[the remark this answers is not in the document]
ltalirz: Edit: re-reading, I think it should just say

Certain "aiida.*" modules are imported when they don't need to me

aiida/cmdline/commands/cmd_computer.py
chrisjsewell: no its just been moved from the top of the file to inside the function, so that its lazy loaded. Its still used inside that function

aiida/cmdline/commands/cmd_computer.py
chrisjsewell: Because I don't want to have to load all the transport plugins, just to list the commands; `add_command` should only be called when you actually need that command.
When you say "every time list_commands is called", why would you expect list_commands to be called multiple times? 
Also note `get_entry_point_names` is cached and so fast on re-calls

aiida/cmdline/commands/cmd_devel.py
chrisjsewell, line 33:
```suggestion
        * Unexpected `aiida.*` modules are imported
```

chrisjsewell: this?

aiida/cmdline/commands/cmd_devel.py
chrisjsewell: what name did you have in mind? I'm not fussed with whatever

aiida/cmdline/commands/cmd_devel.py
chrisjsewell: Sure, although this is a dev command, so I'm not really to bothered about it being that formal

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -10,6 +10,8 @@
 """`verdi devel` commands."""
 import sys
 
+import click
+
 from aiida.cmdline.commands.cmd_verdi import verdi
```
chrisjsewell, line 15:
```suggestion
from aiida.cmdline.commands.cmd_verdi import verdi
from aiida.cmdline.params import options
```

aiida/cmdline/commands/cmd_devel.py
chrisjsewell, line 25:
```suggestion
@options.VERBOSE()
```

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -20,26 +22,36 @@ def verdi_devel():
...
+
     manager = get_manager()
 
     if manager.backend_loaded:
         echo.echo_critical('potential `verdi` speed problem: database backend is loaded.')
 
-    if 'aiida.orm' in sys.modules:
-        echo.echo_critical('potential `verdi` speed problem: `aiida.orm` module is imported.')
+    allowed = ('aiida.backends', 'aiida.cmdline', 'aiida.common', 'aiida.manage', 'aiida.plugins', 'aiida.restapi')
+    for loaded in aiida_modules:
```
chrisjsewell, line 50:
```suggestion
    for loaded in loaded_aiida_modules:
```

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -20,26 +22,36 @@ def verdi_devel():
...
         * the database environment is loaded when it doesn't need to be
-        * the `aiida.orm` module is imported when it doesn't need to be
+        * Only expected `aiida.*` modules are imported
 
     If either of these conditions are true, the command will raise a critical error
     """
     from aiida.manage.manager import get_manager
 
+    aiida_modules = [key for key in sys.modules if key.startswith('aiida.')]
+    aiida_modules_str = '\n- '.join(sorted(aiida_modules))
```
chrisjsewell, line 40:
```suggestion
    aiida_modules_str = '\n- '.join(sorted(loaded_aiida_modules))
```

chrisjsewell:
cheers @ltalirz, addressed all comments

aiida/cmdline/commands/cmd_devel.py
```diff
@@ -10,7 +10,10 @@
 """`verdi devel` commands."""
 import sys
 
+import click
+
```
chrisjsewell, line 14:
```suggestion
```

aiida/cmdline/commands/cmd_computer.py
ltalirz: > Because I don't want to have to load all the transport plugins, just to list the commands; `add_command` should only be called when you actually need that command.

Sure, that part I get.

> When you say "every time list_commands is called", why would you expect list_commands to be called multiple times?

Hm... for example in the test suite?
I don't know whether it does happen in regular use, perhaps not.

aiida/cmdline/commands/cmd_devel.py
ltalirz: capital letters?

chrisjsewell: I’d say capital is only for global variables. I can make it a global variable, but don’t think it’s that necessary

aiida/cmdline/commands/cmd_computer.py
chrisjsewell: Yeh I’d say every time you call the CLI obviously you are reloading everything, so there would not really be much possibility of multiple calls

ltalirz: Ok, I wouldn't call this implementation ideal but it fixes the speed issue and it works, so I guess it's ok.

Feel free to merge or to further improve

ltalirz (approved):

PR #5069, opened 2021-08-12
♻️ REFACTOR: Move to sqlachemy 1.4 (and v2 API)

chrisjsewell:
This PR is intended to update our use of sqlalchemy to the 2.0 API, which is supported by v1.4: https://docs.sqlalchemy.org/en/14/glossary.html#term-1

(note this currently builds on #5063)

PR #5081, opened 2021-08-15
👌 IMPROVE: `QueryBuilder` reprs (incl/ queryhelp -> as_dict)

### LINKED ISSUE #3844: QueryBuilder __str__ does not replace all templates
The query string synthesized by the QueryBuilder sometimes contains template variables that are replaced somewhere down the line before the query is handed over to postgresql.

It seems that there are cases, where calling `print(qb)` on a QueryBuilder instance results in a string, where not all of the variables have been replaced -- meaning, when copy-pasting this string into a psql shell, the query will fail.

A very simple example is this (see `%(extras_1)s`)
```python

In [12]: qb=QueryBuilder()

In [13]: qb.append(Node, filters={"extras.tag4": "appl_pecoal"})
Out[13]: <aiida.orm.querybuilder.QueryBuilder at 0x11d9a9a58>

In [14]: print(qb)
SELECT db_dbnode_1.id, db_dbnode_1.uuid, db_dbnode_1.node_type, db_dbnode_1.process_type, db_dbnode_1.label, db_dbnode_1.description, db_dbnode_1.ctime, db_dbnode_1.mtime, db_dbnode_1.attributes, db_dbnode_1.extras, db_dbnode_1.user_id, db_dbnode_1.dbcomputer_id
FROM db_dbnode AS db_dbnode_1
WHERE CAST(db_dbnode_1.node_type AS VARCHAR) LIKE '%%' AND CASE WHEN (jsonb_typeof(db_dbnode_1.extras #> %(extras_1)s) = 'string') THEN (db_dbnode_1.extras #>> '{tag4}') = 'appl_pecoal' ELSE false END
```

### LINKED ISSUE #3845: QueryBuilder: add shortcut to EXPLAIN ANALYZE

### LINKED ISSUE #4502: Add `as_dict()` and `from_dict()` methods to `QueryBuilder`
During a discussion in #4337, @CasperWA made the excellent point that that the `queryhelp` property name doesn't reflect very well what it is. I also remember that at some point I was actually looking for a way to export a `QueryBuilder` instance, but I glossed over the `queryhelp` when browsing the properties/methods of the `QueryBuilder`.

Most intuitive to me would be two methods:

* `QueryBuilder.as_dict()`: This has the same functionality as the current `queryhelp` property: it returns the `QueryBuilder` instance as a dictionary, which you can use to recreate the same instance with:
* `QueryBuilder.from_dict(d)`: This class method returns the `QueryBuilder` instance from the dictionary created by the `as_dict()` method. This basically just returns `QueryBuilder(**d)`.

I think this API is very clear, but that may be because I was searching for it due to my experience with the [`monty`](https://guide.materialsvirtuallab.org/monty/monty.json.html) package (heavily used in the `pymatgen`-branch of materials science packages).

It's fine to keep the `queryhelp` for now; would it be a good idea to deprecate this at some point?

### LINKED ISSUE #4659: Printing advanced querybuilder objects does not work
If I instantiate this query builder object:
```python
qb = QueryBuilder().append(DataFactory('structure'), project='id', filters={'extras.optimade.elements': {'contains': ['Si']}})
```
and then I print it (`print(qb)`), I get an exception:
```
NotImplementedError: Don't know how to literal-quote value ['Si']
```

By taking inspiration from [this page](https://copdips.com/2020/06/compiling-sqlalchemy-query-to-nearly-real-raw-sql-query.html), I adapted a bit (for JSONB rather than array, and with a fix for strings:
```python
from datetime import date, datetime, timedelta
from sqlalchemy.orm import Query

def render_query(statement, db_session):
    """
    Generate an SQL expression string with bound parameters rendered inline
    for the given SQLAlchemy statement.
    WARNING: This method of escaping is insecure, incomplete, and for debugging
    purposes only. Executing SQL statements with inline-rendered user values is
    extremely insecure.
    Based on http://stackoverflow.com/questions/5631078/sqlalchemy-print-the-actual-query
    """
    if isinstance(statement, Query):
        statement = statement.statement
    dialect = db_session.bind.dialect

    class LiteralCompiler(dialect.statement_compiler):
        def visit_bindparam(
            self, bindparam, within_columns_clause=False, literal_binds=False, **kwargs
        ):
            return self.render_literal_value(bindparam.value, bindparam.type)

        def render_jsonb_value(self, val, item_type):
            if isinstance(val, list):
                return "{}".format(
                    ",".join([self.render_jsonb_value(x, item_type) for x in val])
                )
            if isinstance(val, str):
                return '"{}"'.format(val)
            return self.render_literal_value(val, item_type)

        def render_literal_value(self, value, type_):
            if isinstance(value, int):
                return str(value)
            elif isinstance(value, (str, date, datetime, timedelta)):
                return "'{}'".format(str(value).replace("'", "''"))
            elif isinstance(value, list):
                return "'[{}]'".format(
                    ",".join(
                        [self.render_jsonb_value(x, type_) for x in value]
                    )
                )
            return super(LiteralCompiler, self).render_literal_value(value, type_)

    return LiteralCompiler(dialect, statement).process(statement)
```

This now works:
```python
q = qb.get_query()
print(render_query(q, q.session))
```
returns
```
SELECT db_dbnode_1.id 
FROM db_dbnode AS db_dbnode_1 
WHERE CAST(db_dbnode_1.node_type AS VARCHAR) LIKE 'data.structure.%' AND CAST((db_dbnode_1.extras #> '{optimade, elements}') AS JSONB) @> '["Si"]'
```

This is still probably not complete (we should check the other base types as well as dictionaries) but is a good start.
Should we implement this (or an extension of it) in the method returning the `__str__` representation of a QueryBuilder?

Pinging @ramirezfranciscof (who maybe could implement this?) and @CasperWA as this is related to his OPTIMADE-related queries

chrisjsewell:
This PR principally does:

1. Deprecates `queryhelp` for `as_dict()` and also add a `from_dict` classmethod

As I believe @mbercx and @CasperWA already agree, `queryhelp` is a very unintuitive name, and a stumbling block for many new users.

2. Replace the `__str__` outputs with a representation of `as_dict()` rather than the compiled SQL query (and also use this for `__repr__`).

Users should not need to be exposed to the SQL, this is a "dev-user" feature 

3. Add an `as_sql` method, for retrieving the SQL query string

This has an option to allow for both strings with the parameters inlined or external.
It also fixes  #3844 and #4659 (since the SQL string is no longer generated on print, this is also less of an issue)

4. Add `analyze_query` method, to run postgresql's `EXPLAIN` command.

5. Some more type hinting and tests

closes #4502
closes #3844
closes #3845
closes #4659


chrisjsewell:
Ok this is done, I would like to merge this asap, to feed it up into #5069. So looking to merge in next day or two

aiida/orm/querybuilder.py
```diff
@@ -491,25 +506,82 @@ def __init__(
...
+            'project': self._projections,
+            'order_by': self._order_by,
+            'limit': self._limit,
+            'offset': self._offset,
+        })
+
+    @property
+    def queryhelp(self) -> QueryDict:
+        """"Legacy name for ``as_dict`` method."""
+        # deprecate?
```
chrisjsewell, line 523: The only outstanding question, for me, is whether we actually add a deprecation warning here, or "schedule" it for a later date. I guess essentially it shouldn't be removed until v3?

aiida/orm/querybuilder.py
```diff
@@ -491,25 +506,82 @@ def __init__(
...
+            since normally sqlalchemy will handle this process internally.
+
+        :params inline: Inline bound parameters (this is normally handled by the Python DBAPI).
+        """
+        compiled = self._compile_query(self.get_query(), literal_binds=inline)
+        if inline:
+            return compiled.string + '\n'
+        return f'{compiled.string!r} % {compiled.params!r}\n'
+
+    def __str__(self) -> str:
```
CasperWA, line 577: Wouldn't this be more the `__repr__` functionality? Although I can't think of the essentials to show for the string representation?

aiida/orm/querybuilder.py
```diff
@@ -79,8 +82,20 @@ class PathItemType(TypedDict):
...
         outerjoin: bool
         edge_tag: str
+
+    class QueryDict(TypedDict, total=False):
+        """A JSON serialisable representation of a ``QueryBuilder`` instance"""
+
+        path: List[PathItemType]
+        filters: Dict[str, FilterType]
+        project: Dict[str, ProjectType]
+        order_by: List[str]
```
CasperWA, line 92: Almost sure this is a list of dicts... maybe re-check all of these to be 100 % sure?

aiida/orm/querybuilder.py
```diff
@@ -491,25 +506,82 @@ def __init__(
...
+                    raise
+
+        return _Compiler(dialect, query.statement, compile_kwargs=dict(literal_binds=literal_binds))
+
+    def as_sql(self, inline: bool = False) -> str:
+        """Convert the query to an SQL string representation.
+
+        .. warning::
+
+            This method should be used for debugging puposes only,
```
CasperWA, line 567:
```suggestion
            This method should be used for debugging purposes only,
```

aiida/orm/querybuilder.py
```diff
@@ -2006,46 +2051,33 @@ def get_used_tags(self, vertices: bool = True, edges: bool = True) -> List[str]:
...
-        # It describes whether the current query
-        # which is an attribute _query of this instance is still valid
-        # The queryhelp_hash is used to determine
-        # whether the query is still valid
-
-        queryhelp_hash = make_hash(self.queryhelp)
-        # if self._hash (which is None if this function has not been invoked
-        # and is a string (hash) if it has) is the same as the queryhelp
-        # I can use the query again:
-        # If the query was injected I never build:
```
CasperWA, line 2029: Bad joke incoming ...


As we forget faster and faster and -- what?
Oh right! Maybe put this in the doc string of the method instead of removing completely, to keep some of the humanly laid out logic behind the code.

aiida/orm/querybuilder.py
```diff
@@ -2095,16 +2127,27 @@ def distinct(self) -> 'QueryBuilder':
...
+        compiled = self._compile_query(query, literal_binds=True)
+        options = ', '.join((['ANALYZE'] if execute else []) + (['VERBOSE'] if verbose else []))
+        options = f' ({options})' if options else ''
+        rows = self._impl.get_session().execute(f'EXPLAIN{options} {compiled.string}').fetchall()
+        return '\n'.join(row.values()[0] for row in rows)
 
-            qb = QueryBuilder(**queryhelp)
-            qb.first()
+    def first(self) -> Optional[List[RowType]]:
+        """Executes the query, asking for the first row of results.
```
CasperWA, line 2148:
```suggestion
        """Executes the query, asking for the first row of results.

        Note, this may change if several rows are valid for the query,
        as persistent ordering is not guaranteed unless explicitly specified.
```

aiida/restapi/resources.py
```diff
@@ -209,9 +209,9 @@ def get(self, id=None, page=None):  # pylint: disable=redefined-builtin,invalid-
 
 class QueryBuilder(BaseResource):
     """
-    Representation of a QueryBuilder REST API resource (instantiated with a queryhelp JSON).
+    Representation of a QueryBuilder REST API resource (instantiated with a serialised QueryBuilder instance).
 
-    It supports POST requests taking in JSON :py:func:`~aiida.orm.querybuilder.QueryBuilder.queryhelp`
+    It supports POST requests taking in JSON :py:func:`~aiida.orm.querybuilder.QueryBuilder.as_dict`
```
CasperWA, line 214: Maybe this should reference the new `QueryDict` type instead then? While also referencing the `as_dict()` method as the "driver"?

docs/source/topics/database.rst
```diff
@@ -222,12 +222,18 @@ List of all relationships:
 +------------------+---------------+--------------------+-------------------------------------------------+
 
 .. _topics:database:advancedquery:queryhelp:
+.. _topics:database:advancedquery:querydict:
 
-The queryhelp
--------------
+Converting the QueryBuilder to/from a dictionary
```
CasperWA, line 227: This should be inserted as the new reference for the link in the REST API help...

docs/source/topics/database.rst
```diff
@@ -305,12 +311,12 @@ To do this, you have to specify:
...
         The absolute value of the direction defines the table to join to with respect to your own position in the list.
         An absolute value of 1 joins one table above, a value of 2 to the table defined 2 indices above.
-        The two following queryhelps yield the same query::
+        The two following dictionaries yield the same query::
 
             from aiida.orm import TrajectoryData
             from aiida_quantumespresso.calculations.pw import PwCalculation
             from aiida.orm import Dict
-            qh1 = {
+            qd1 = {
```
CasperWA, line 319:
```suggestion
            query_dict_1 = {
```

docs/source/topics/database.rst
```diff
@@ -327,7 +333,7 @@ To do this, you have to specify:
 
             # returns same query as:
 
-            qh2 = {
+            qd2 = {
```
CasperWA, line 336:
```suggestion
            query_dict_2 = {
```

docs/source/topics/database.rst
```diff
@@ -344,17 +350,17 @@ To do this, you have to specify:
 
             # Shorter version:
 
-            qh3 = {
+            qd3 = {
```
CasperWA, line 353:
```suggestion
            query_dict_3 = {
```

CasperWA (changes requested):
Cool, thanks @chrisjsewell !

Some comments along with the actual review comments:
Maybe rename the awkward `_kw` to the more standard `kwargs`?
Are the *.txt and test_as_dict.yml files in the test folder for the `QueryBuilder` necessary? They seem to maybe be relics from some tests that should be removed or are they intentional?

chrisjsewell:
cheers for the quick review @CasperWA 😄

chrisjsewell:
> Are the *.txt and test_as_dict.yml files in the test folder for the QueryBuilder necessary? They seem to maybe be relics from some tests that should be removed or are they intentional?

No they are intentional and generated in `tests/orm/test_querybuilder.py::TestRepresentations` by pytest-regressions

chrisjsewell:
> Maybe rename the awkward _kw to the more standard kwargs?

done

aiida/orm/querybuilder.py
chrisjsewell: I felt these comments were just describing what you can already easily see from the code and not adding any actual extra insight.

aiida/orm/querybuilder.py
chrisjsewell: so with this PR:

```python
        qb = orm.QueryBuilder().append(orm.Data, project=['id', 'uuid']).order_by({orm.Data: 'id'})
        print(repr(qb))
        print(qb)
```

goes to

```python
<aiida.orm.querybuilder.QueryBuilder object at 0x7fbd0aa64410>

QueryBuilder(path=[{'entity_type': 'data.Data.', 'tag': 'Data_1', 'joining_keyword': None, 'joining_value': None, 'edge_tag': None, 'outerjoin': False}], filters={'Data_1': {'node_type': {'like': 'data.%'}}}, project={'Data_1': [{'id': {}}, {'uuid': {}}]}, order_by=[{'Data_1': [{'id': {'order': 'asc'}}]}], limit=None, offset=None)
```

Personally I think this makes sense, since the `__repr__` uniquely identifies the object, and the string `__str__` provides a more verbose definition of it (and the bonus is you can literally copy/paste it to recreate the instance)

aiida/orm/querybuilder.py
chrisjsewell, line 92:
```suggestion
        order_by: List[dict]
```

chrisjsewell: oops yeh, the others are definitely correct

aiida/restapi/resources.py
chrisjsewell: probably might as well stretch to closing #4660 here as well

chrisjsewell: and then you can reference that

aiida/orm/querybuilder.py
CasperWA: I wouldn't say the `__repr__` defines the instance in a usable unique way, other than referencing the point in memory. Again, I'm more for the `__repr__` string being used for this exact purpose (recreating an instance) - this is the way it is used in the built-in Python libs as well - but that said this is already discussed extensively in #3087.

aiida/orm/querybuilder.py
CasperWA: I agree to an extent. I think the purpose of having the hash is not self-explanatory from the code though, unless you are _very_ into the code and always use hashes this way - and even then, I think it's fair to put in a human-readable comment either by the `make_hash` line or in the doc-string to quickly mention the purpose (i.e., reusing existing `QueryBuilder` instances versus making new ones, I guess?) and that by default the value in the following lines is `None`, something else that might not be clear from the code, since it only relays what to do _if_ the value is `None`.

CasperWA:
> No they are intentional and generated in `tests/orm/test_querybuilder.py::TestRepresentations` by pytest-regressions

Why? Is there a reason this is necessary to keep in the repo? Couldn't it just be deleted after use or kept in memory if it's to be used elsewhere?

chrisjsewell:
> > No they are intentional and generated in `tests/orm/test_querybuilder.py::TestRepresentations` by pytest-regressions
> 
> Why? Is there a reason this is necessary to keep in the repo? Couldn't it just be deleted after use or kept in memory if it's to be used elsewhere?

No this is just how https://pytest-regressions.readthedocs.io works; it’s regressing against the stored files. I’d definitely recommend it, we use it loads, for example, in aiida-quantumespresso

aiida/orm/querybuilder.py
chrisjsewell: yeh lets leave this discussion to that issue lol; IMO this is exactly what the repr and str should provide, otherwise they would just be identical.
Anyhow, I think its certainly better than the current situation, of showing the raw SQL to users (which can also except)

CasperWA: Sure - but again you're introducing another way of string representing AiiDA entities in the code base now. Maybe do it, but make an explicit note of it the general issue - although a future PR for #3087 should already take *all* AiiDA entities into account anyhow ... so yeah. Fine, I guess? 😅

chrisjsewell: The QueryBuilder is not an aiida entity though, i.e. it does not relate to an ORM object, and its string representation was already different to anything else in aiida; so its not introducing another way, its just changing the current way

CasperWA: Hmm, I'm fuzzy on that definition exactly myself. But if you consider everything that is stored as a row in a PostgreSQL table an entity, I'll agree. But I'm not entirely sure that's the definition. But maybe it should be.

chrisjsewell: In fact might just make this both the repr and str

aiida/orm/querybuilder.py
```diff
@@ -2006,46 +2051,33 @@ def get_used_tags(self, vertices: bool = True, edges: bool = True) -> List[str]:
...
-
-        queryhelp_hash = make_hash(self.queryhelp)
-        # if self._hash (which is None if this function has not been invoked
-        # and is a string (hash) if it has) is the same as the queryhelp
-        # I can use the query again:
-        # If the query was injected I never build:
+        need_to_build = True
+        query_hash = make_hash(self.as_dict())
         if self._hash is None:
             need_to_build = True
```
chrisjsewell, line 2065:
```suggestion
            # this is the first time the query has been built
            need_to_build = True
```

aiida/orm/querybuilder.py
chrisjsewell: It is already explained in the docstring, reagrding the use of a hash to avoid rebuilds, I've added an extra comment below though about if the hash is None

aiida/orm/querybuilder.py
chrisjsewell: yeh I've just made the repr and str the same, the basically what happens for all the python built-in stuff

docs/source/topics/database.rst
chrisjsewell: done

chrisjsewell:
@CasperWA I've addressed all your comments cheers

aiida/orm/querybuilder.py
chrisjsewell: I don't think it will be good for users to have deorecation warnings showing up when they've just switched to v2, so I would be tempted to wait

chrisjsewell:
merging this today

CasperWA:
> merging this today (the link check fail is just that https://shapeshed.com/unix-exit-codes/ is down at present)

Sure - but maybe always wait for an approving review, or?

docs/source/topics/database.rst
```diff
@@ -222,12 +222,18 @@ List of all relationships:
...
+------------------------------------------------
 
-The ``queryhelp`` dictionary is a property of the :class:`~aiida.orm.querybuilder.QueryBuilder` class.
-Once you have built your query using the appender method explained in the :ref:`finding and querying for data how-to<how-to:query>` and the advanced sections above, you can easily store your query by saving the ``QueryBuilder.queryhelp`` dictionary as a JSON file for later use:
+.. important::
+
+    In aiida-core version 1, this dictionary was accessed with ``QueryBuilder.queryhelp``, which is now deprecated.
+
+The :class:`~aiida.orm.querybuilder.QueryBuilder` class can be converted to a dictionary and also loaded from a dictionary,
+for easy serialisation and storage.
```
CasperWA, line 235:
```suggestion
The :class:`~aiida.orm.querybuilder.QueryBuilder` class can be converted to a dictionary and also loaded from a dictionary, for easy serialisation and storage.
```

CasperWA (dismissed):
Thanks a bunch @chrisjsewell !
Approved on my end. I found a single "one sentence per one line" thing in the documentation and that's about it.

chrisjsewell:
> but maybe always wait for an approving review,

I had full faith that you would sign it off today 😄

chrisjsewell:
> Approved on my end

Thanks!

PR #5082, opened 2021-08-16
CI: Update/maintain test-install GA workflow

csadorf:
Related to: https://github.com/aiidateam/aiida-core/pull/5046#issuecomment-898386064

- MAINTAIN: Add 'check-yaml' to pre-commit hooks.
  Should catch issues with the definition of workflow files in the future.
- Dependencies: Fix the pyzmq requirement specification for Python 3.9.
  I have verified that the `resolve-pip-dependencies` job (see below) fails without this change.
- CI: Implement _resolve-pip-dependencies_ job.
  Checks whether environments defined in the requirements/ files are resolvable.
- CI: Separate the direct conda installation test into separate job.
  The test whether we can actually install aiida with `conda install aiida-core ...` is run as a dedicated job only after we checked that we can create a conda environment from the `environment.yml` file.
- CI: Run _install-with-conda_ job for all supported Python versions.
  To test whether `conda install aiida-core ...` actually works for all supported Python versions.

Checklist before merge:
- [ ] Restore deleted GA workflows.
- [ ] Re-enable test suite

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -73,6 +73,32 @@ jobs:
         pip install -r requirements/requirements-py-${{ matrix.python-version }}.txt
         pip freeze
 
+  create-conda-environment:
+    # Verify that we can create a valid conda environment from the environment.yml file.
+
+    needs: [validate-dependency-specification]
+    if: github.repository == 'aiidateam/aiida-core'
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setup Conda
+      uses: s-weigand/setup-conda@v1
+      with:
+        python-version: 3.9
+        update-conda: false
+        conda-channels: conda-forge
+
+    - run: conda --version
+    - run: python --version
+    - run: which python
+
+    - name: Test conda environment
+      run: |
+        conda env create --dry-run -f environment.yml -n test-environment
+
   install-with-pip:
 
     if: github.repository == 'aiidateam/aiida-core'
@@ -107,12 +133,18 @@ jobs:
         python -c "import aiida"
 
   install-with-conda:
+    # Verify that we can install AiiDA with conda.
 
     if: github.repository == 'aiidateam/aiida-core'
-    runs-on: ubuntu-latest
-    name: install-with-conda
+    # Currently the installation for Python 3.7 fails in combination with
+    # ubuntu-20.04 due to an incompatibility with glibc.
+    runs-on: ubuntu-18.04
+    timeout-minutes: 10
 
-    timeout-minutes: 5
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


csadorf:
@ltalirz The PR is currently in draft mode, because I have deactivated all unrelated GA runs to save on time and resources. I would appreciate if you could review the PR in draft mode so that all necessary changes and fixes can be implemented in an efficient manner without accidentally merging it. Once the PR is accepted, I will do one final run with all workflows restored before merge.

_I recommend to not squash this PR on merge._

csadorf:
@ltalirz Let me know if I can help with the review.

.github/workflows/test-install.yml
```diff
@@ -169,12 +229,6 @@ jobs:
       run:
         .github/workflows/setup.sh
 
-    - name: Run test suite
```
ltalirz, line 172: Just to check: you were planning to bring this back?

ltalirz (commented):
thanks @csadorf , looks good to me.

My only comment regards the job dependencies: while I get that this makes sense from a perspective of saving resources, does it lead to a slowdown in practice?

I guess the jobs don't actually depend on each other to run successfully but rather if one jobs fail, then the others will as well?

I'm ok with both approaches; just curious to know whether the total runtime differs significantly (maybe the effect is anyhow small).

.github/workflows/test-install.yml
csadorf: Absolutely! 😀

csadorf: > thanks @csadorf , looks good to me.
> 
> My only comment regards the job dependencies: while I get that this makes sense from a perspective of saving resources, does it lead to a slowdown in practice?
> 
> I guess the jobs don't actually depend on each other to run successfully but rather if one job fails, then we know that the others will fail as well?
> 
> I'm ok with both approaches; just curious to know whether the total runtime differs significantly (maybe the effect is anyhow small).

Yes, the idea was to save some resources, but it is true that some of those jobs do not actually depend on each other in the way that it is configured, but rather belong to a "topical unit". I guess it would be better practice to remove those dependencies and run everything maximal parallel. I will rewire the dependencies and request another review.

csadorf:
> thanks @csadorf , looks good to me.
> 
> My only comment regards the job dependencies: while I get that this makes sense from a perspective of saving resources, does it lead to a slowdown in practice?
> 
> I guess the jobs don't actually depend on each other to run successfully but rather if one job fails, then we know that the others will fail as well?
> 
> I'm ok with both approaches; just curious to know whether the total runtime differs significantly (maybe the effect is anyhow small).

Yes, the idea was to save some resources, but it is true that some of those jobs do not actually depend on each other in the way that it is configured, but rather belong to a "topical unit". I guess it would be better practice to remove those dependencies and run everything maximal parallel. I will rewire the dependencies and request another review.

csadorf:
@ltalirz Job dependencies are rewired. The runtime (without actually running the test suite) has been slightly reduced, but I guess this depends on how many workers GitHub makes available to us at a given time.

https://github.com/aiidateam/aiida-core/actions/runs/1142730871

ltalirz:
thanks @csadorf - up to you how to handle the dependency situation, both is fine with me.

Feel free to bring back the rest and merge without further review from my side

PR #5083, opened 2021-08-16
✨ NEW: Add querybuilder jsonschema

chrisjsewell:
This is a draft PR for adding a validation schema for input to the `QueryBuilder`.

Currently the schema is based on the output of `QueryBuilder.as_dict`, which is more restrictive than the actual input to `QueryBuilder`. This is because when init'ing or adding variables to the query, the QueryBuilder coerces certain input types to the final format.
If we use this e.g. for validation in the REST API, do we want to allow all possible formats?

Note also, this PR is currently based on top of #5081


CasperWA:
> If we use this e.g. for validation in the REST API, do we want to allow all possible formats?

Maybe not publicly? However, it also seems unintuitive that a passed dictionary that normally works with `QueryBuilder.from_dict()` doesn't work with the REST API.

Thinking more about it, I'd consider making the `/querybuilder` endpoint POSTed JSON data equal what can be done with the `QueryBuilder.from_dict()` method. It's more intuitive and it makes it easy to handle internally as well. The documentation should be able to be based on the current documentation for what may be passed to the `from_dict()` method and then that's that?
Unless it poses a security vulnerability, of course.
I'm guessing the output of `to_dict()` represents a subset of what can be passed to `from_dict()`? In which case this also works out.

chrisjsewell:
> I'm guessing the output of to_dict() represents a subset of what can be passed to from_dict()? In which case this also works out.

yep

> It's more intuitive and it makes it easy to handle internally as well.

I'd kinda disagree with this: the pro is that you can have smaller / less verbose JSONs, e.g.

```json
{
   "project": {
      "tag1": [{"property1": {}}]
   } 
}
```

rather than:

```json
{
   "project": {
      "tag1": ["property"]
   } 
}
```

But the con is that it goes against the Python zen; "There should be one-- and preferably only one --obvious way to do it".
Users now have to understand multiple possible formats, and the JSON schema has to be more complex to, incorporate the multiple formats, which will probably mean the validation errors it returns will be less "specific" and also validation will be slower (probably not noticeably though)

chrisjsewell:
> It's more intuitive and it makes it easy to handle internally as well.

Actually, another reason that come to mind for wanting to only allow the "final" format in the schema, is that then you could potentially just bypass all the `QueryBuilder` code that does the validation and type coercion, and set the instance attributes directly.
This would intrinsically speed up REST API calls (not sure by how much though)

PR #5085, opened 2021-08-18
CLI: add verbosity control

### LINKED ISSUE #4330: Control verbosity of click output

sphuber:
Supersedes #3896 
Fixes #4330 

For those reviewing this, I recommend looking at the individual commits.

1. Simple change to "formally" add the custom `REPORT` level and the `report` method to the `logging` built-in module. This will allow it to be used just like for example `logging.DEBUG` and `logging.debug()` and the other standard log levels.
2. Define a `-v/--verbosity` option that is automatically added to all `verdi` commands. When specified on the CLI, this option changes the log level for the `AIIDA_LOGGER` by setting the corresponding option value on the `Profile` instance that is loaded in memory
3. Refactor the echo functions from the `aiida.cmdline.utils.echo` module to log their messages to the `CMDLINE_LOGGER`, a direct child of the `AIIDA_LOGGER`, instead of directly echo'ing the messages to `stdout` and `stderr`.
4. Change majority of `echo_info` invocations for `echo_report`. Since `REPORT` is the default log level, after the previous changes the content of `echo_info` will no longer show up, so we change them to `echo_report`.


ltalirz:
thanks a lot @sphuber  - I guess you're looking for a review from one of us three.
I'm currently a bit short of time; @ramirezfranciscof would you perhaps be available for this one?

sphuber:
Thanks @ltalirz . I pinged you since you have wrote the original PR and were involved in all the discussions, but I guess if the other's review it, that would be fine as well. You can always just give feedback on the higher level design and naming decisions and leave the review of the implementation to the rest.

To see what it would look like and how it would work for the user, you can maybe have a quick look at the section I wrote for the docs:
```
Verbosity
=========
All ``verdi`` commands have the ``-v/--verbosity`` option, which allows to control the verbosity of the output that is printed by the command.
The option takes a value that is known as the log level and all messages that are emitted with an inferior log level will be suppressed.
The valid values in order of increasing log level are: `NOTSET`, `DEBUG`, `INFO`, `REPORT`, `WARNING`, `ERROR` and `CRITICAL`.
For example, if the log level is set to ``ERROR``, only messages with the ``ERROR`` and ``CRITICAL`` level will be shown.
The choice for these log level values comes directly from `Python's built-in logging module <https://docs.python.org/3/library/logging.html>`_.
The ``REPORT`` level is a log level that is defined and added by AiiDA that sits between the ``INFO`` and ``WARNING`` level, and is the default log level.

The verbosity option is case-insensitive, i.e., ``--verbosity debug`` and ``--verbosity DEBUG`` are identical.
The option can be passed at any subcommand level, for example:

.. code:: console

    verdi process list --verbosity debug

is identical to

.. code:: console

    verdi --verbosity debug process list

When the option is specified multiple times, only the last value will be considered.
The default value for the logging level can be changed permanently through the configuration system.
For example, to set the default log level to ``WARNING``,

.. code:: console

    verdi config set logging.aiida_loglevel WARNING

This is identical to passing ``--verbosity WARNING`` manually to each invoked command.

.. warning:: Setting the configuration option does not just affect the verbosity of the command line, but the logging for all of AiiDA.
    For example, it also affects the verbosity of process reports, such as those of work chains.
    If the log level is set above ``REPORT``, the reports of work chains will no longer be logged.

```

chrisjsewell:
Thanks @sphuber.
Initial feedback, I note that you have not removed/replaced the `VERBOSE`, which I imagine may mean weird behaviour for these commands:

```
aiida/backends/sqlalchemy/manage.py:
  54: @options.VERBOSE()
  62: @options.VERBOSE()

aiida/cmdline/commands/cmd_code.py:
  147: @options.VERBOSE()
  185: @options.VERBOSE()

aiida/cmdline/commands/cmd_database.py:
  200: @options.VERBOSE()

aiida/cmdline/commands/cmd_devel.py:
  24: @options.VERBOSE()

aiida/cmdline/commands/cmd_group.py:
  96: @options.VERBOSE()

aiida/cmdline/commands/cmd_node.py:
  283: @options.VERBOSE()
```

one of these is `verdi node delete`, which I know directly sets the level of the `DELETE_LOGGER`, which is used in the `delete_nodes` function. I guess this would need to be changed and also all uses of `DELETE_LOGGER.info` -> `DELETE_LOGGER.report`

then there are three instances of directly creating an option with `--verbosity`, all in the archive: 

```
aiida/cmdline/commands/cmd_archive.py:
   87:     '--verbosity',
  206:     '--verbosity',
  247:             'failed to migrate the archive file (use `--verbosity DEBUG` to see traceback): '
  318:     '--verbosity',
```

again they directly set certain logger levels, which are then used in the calling functions, plus there is the relationship between verbosity levels and progress bars, which would need to be adapted a bit

chrisjsewell:
There are 66 instances of `.info(` and we should probably decide if these are now better as `.report(`:

```
66 results - 17 files

aiida/engine/runners.py:
  328:         LOGGER.info('adding subscriber for broadcasts of %d', pk)
  349:             LOGGER.info('%s<%d> confirmed to be terminated by backup polling mechanism', *args)

aiida/engine/daemon/runner.py:
  28:     LOGGER.info('Received signal to shut down the daemon runner')
  37:     LOGGER.info('Daemon runner stopped')
  58:         LOGGER.info('Starting a daemon runner')
  61:         LOGGER.info('Received a SystemError: %s', exception)
  64:     LOGGER.info('Daemon runner started')

aiida/engine/processes/process.py:
  292:         self.node.logger.info(f'Loaded process<{self.node.pk}> from saved state')
  300:         self.node.logger.info(f'Request to kill Process<{self.node.pk}>')
  311:                     self.logger.info('no controller available to kill child<%s>', child.pk)
  319:                     self.logger.info('no connection available to kill child<%s>', child.pk)
  321:                     self.logger.info('kill signal was unable to reach child<%s>', child.pk)

aiida/engine/processes/calcjobs/calcjob.py:
  421:             self.logger.info('could not parse scheduler output: the `detailed_job_info` attribute is missing')
  423:             self.logger.info('could not parse scheduler output: return value of `detailed_job_info` is non-zero')
  445:             self.logger.info(f'`{scheduler.__class__.__name__}` does not implement scheduler output parsing')

aiida/engine/processes/calcjobs/manager.py:
   97:             self.logger.info('waiting for transport')
  114:             self.logger.info(f'AuthInfo<{self._authinfo.pk}>: successfully retrieved status of active jobs')

aiida/engine/processes/calcjobs/tasks.py:
   95:         logger.info(f'scheduled request to upload CalcJob<{node.pk}>')
  108:         logger.info(f'uploading CalcJob<{node.pk}> successful')
  143:         logger.info(f'scheduled request to submit CalcJob<{node.pk}>')
  154:         logger.info(f'submitting CalcJob<{node.pk}> successful')
  204:         logger.info(f'scheduled request to update CalcJob<{node.pk}>')
  215:         logger.info(f'updating CalcJob<{node.pk}> successful')
  266:                 logger.info(f'detailed job info not available for scheduler of CalcJob<{node.pk}>')
  274:         logger.info(f'scheduled request to retrieve CalcJob<{node.pk}>')
  286:         logger.info(f'retrieving CalcJob<{node.pk}> successful')
  318:             logger.info(f'stashing calculation<{node.pk}>')
  336:         logger.info(f'stashing calculation<{node.pk}> successful')
  369:         logger.info(f'scheduled request to kill CalcJob<{node.pk}>')
  377:         logger.info(f'killing CalcJob<{node.pk}> successful')

aiida/engine/processes/workchains/workchain.py:
  332:         self.logger.info('received callback that awaitable %d has terminated', awaitable.pk)

aiida/manage/external/rmq.py:
  194:             LOGGER.info('not continuing process<%d> which is already terminated with state %s', pid, node.process_state)

aiida/parsers/plugins/diff_tutorial/parsers.py:
  40:         self.logger.info(f"Parsing '{output_filename}'")
  61:         self.logger.info(f"Parsing '{output_filename}'")

aiida/schedulers/plugins/direct.py:
  136:                 self.logger.info('sched_join_files is True, but sched_error_path is set; ignoring sched_error_path')
  205:         self.logger.info(f'submitting with: {submit_command}')
  388:         self.logger.info(f'killing job {jobid}')

aiida/schedulers/plugins/lsf.py:
  500:         self.logger.info(f'submitting with: {submit_command}')
  727:         self.logger.info(f'killing job {jobid}')

aiida/schedulers/plugins/pbsbaseclasses.py:
  215:                 _LOGGER.info(
  262:                 _LOGGER.info(
  334:         _LOGGER.info(f'submitting with: {submit_command}')
  718:         _LOGGER.info(f'killing job {jobid}')

aiida/schedulers/plugins/sge.py:
  188:                 self.logger.info(
  225:                 self.logger.info(
  300:         self.logger.info(f'submitting with: {submit_command}')
  499:         self.logger.info(f'killing job {jobid}')
  521:             self.logger.info(

aiida/schedulers/plugins/slurm.py:
  323:                 self.logger.info(
  432:         self.logger.info(f'submitting with: {submit_command}')
  706:         self.logger.info(f'killing job {jobid}')

aiida/tools/graph/deletions.py:
   73:     DELETE_LOGGER.info('%s Node(s) marked for deletion', len(pks_set_to_delete))
   90:         DELETE_LOGGER.info('This was a dry run, exiting without deleting anything')
   95:         DELETE_LOGGER.info('This was a dry run, exiting without deleting anything')
  101:     DELETE_LOGGER.info('Starting node deletion...')
  103:     DELETE_LOGGER.info('Deletion of nodes completed.')

aiida/tools/importexport/archive/migrators.py:
  127:         MIGRATE_LOGGER.info('Reading archive version')
  144:             MIGRATE_LOGGER.info('No migration required')
  147:         MIGRATE_LOGGER.info('Migration pathway: %s', ' -> '.join(pathway + [version]))
  165:         MIGRATE_LOGGER.info('Extracting archive to work directory')
  188:             MIGRATE_LOGGER.info(f"Re-compressing archive as '{out_compression}'")
  202:             MIGRATE_LOGGER.info('Moving archive to: %s', out_path)

aiida/tools/importexport/dbexport/main.py:
  253:             EXPORT_LOGGER.info('No entities were found to export')
  273:         EXPORT_LOGGER.info('Finalizing Export...')
  278:         EXPORT_LOGGER.info('Exported Entities:\n  - ' + export_summary + '\n')
  280:     # EXPORT_LOGGER.info('Writer Information:\n %s', writer.export_info)
```

sphuber:
> Initial feedback, I note that you have not removed/replaced the `VERBOSE`, which I imagine may mean weird behaviour for these commands:

Good point. I will address this, removing the `VERBOSE` option and replacing it with the `VERBOSITY` one.

> There are 66 instances of .info( and we should probably decide if these are now better as .report(:

I think we should not change these to report, because that would change the current behavior. Also, under our definition of `report` and `info` I think it is correct that these are info messages.

chrisjsewell:
> I think we should not change these to report, because that would change the current behavior. Also, under our definition of report and info I think it is correct that these are info messages.

The behaviour will already being changing though won't it? i.e. If the `verdi archive` and `verdi node delete` commands are changed to use this new VERBOSITY option, then the default log level will change from `info` -> `report`, and so the `DELETE_LOGGER`, `MIGRATE_LOGGER` and `EXPORT_LOGGER` qill need to change their `info` messages to `report`, to be seen in the CLI?

sphuber:
Ah yeah, I forgot that Casper and you programmed the delete and export functionality such that info messages would be shown by default. If that is the case, then they should indeed be changed to report. Looking at those messages, I feel that they are still quite verbose and it actually makes sense to not show them by default and the user should just pass `-v info` to see them. But if you prefer to show those messages by default, then we change them to report ones. It is true that that would also keep current behavior so maybe that is best anyway.

aiida/cmdline/commands/cmd_code.py
```diff
@@ -174,28 +173,24 @@ def show(code, verbose):
...
 @options.DRY_RUN()
 @options.FORCE()
 @with_dbenv()
-def delete(codes, verbose, dry_run, force):
+def delete(codes, dry_run, force):
     """Delete a code.
 
     Note that codes are part of the data provenance, and deleting a code will delete all calculations using it.
     """
     from aiida.common.log import override_log_formatter_context
```
chrisjsewell, line 192: Do we still need to use this here?

sphuber: I don't think so no, since the logging now goes through the echo functions which essentially do this formatting. Shall I also remove the utility itself, since its only use is in the cmdline commands?

chrisjsewell: Quite proably yeh, again it would be good to run some of these commands once you are done and print the terminal output here, so we can check they are still looking pretty 😄

sphuber: Uhm, there is still a problem. Even though the logging level of the child loggers is properly set on the `AIIDA_LOGGER` and they inherit it, they still have the default formatting. Removing these overrides will therefore return the "incorrectly" formatted log messages. I think we should still remove them, but like the log level, also change the formatter for the `AIIDA_LOGGER` and all its children, globally.

aiida/cmdline/commands/cmd_archive.py
```diff
@@ -170,11 +165,10 @@ def create(
     elif archive_format == 'null':
         export_format = 'null'
 
-    if verbosity in ['DEBUG', 'INFO']:
-        set_progress_bar_tqdm(leave=(verbosity == 'DEBUG'))
+    if AIIDA_LOGGER.level <= logging.INFO:
```
chrisjsewell, line 168: should this not be `<= logging.REPORT` now

chrisjsewell: (and same for other two instances below)

sphuber: fixed

chrisjsewell: don't see the change yet?

sphuber: Should be there now. Were a bunch of other fixes that had to go in.

aiida/cmdline/commands/cmd_code.py
sphuber: OK, commit 6340311 implements a solution for this problem by configuring a custom formatter for all loggers when the CLI is invoked.

chrisjsewell (approved):
Ok I think this is all good cheers

PR #5086, opened 2021-08-18
Engine: implement functionality to import completed `CalcJobs`

### LINKED ISSUE #1892: Add ImmigrantJobProcess

sphuber:
Fixes #1892
Implementation of [this AEP](https://github.com/aiidateam/AEP/pull/12).

Example of usage:
```python

importer = ArithmeticAddCalculation.get_importer()
remote_data = RemoteData('/some/absolute/path', computer=load_computer('computer'))
inputs = importer.parse_remote_data(remote_data)
results, node = run.get_node(ArithmeticAddCalculation, **inputs)
assert node.is_imported
```

This is as standardized as it can be as the `parse_remote_data` for each importer implementation will probably need custom keyword arguments. For the Python API, we could provide a simple wrapping function, like e.g.:
```python
def import_calcjob(cls: CalcJob, remote_data: RemoteData, **kwargs: Any) -> CalcJobNode:
    """"Import the contents of a completed calcjob.

    :param cls: the `CalcJob` class to which the completed calcjob corresponds, e.g. `ArithmeticAddCalculation`.
    :param remote_data: a ``RemoteData`` node that contains the input and output files of the completed job.
    :param kwargs: any additional inputs that should be passed to the importer plugin.
    :return: the node representing the imported calcjob when successful.
    """
    importer = cls.get_importer()
    inputs = importer.parse_remote_data(remote_data, **kwargs)
    inputs['remote_folder'] = remote_data
    results, node = run.get_node(cls, **inputs)
    return node
```
However, due to the arbitrary keyword arguments that need to be supported for the importer, it will be tricky to turn this into a CLI command. Unless we implement some plugin system for that to dynamically define the options based on the spec of the importer, such as we do for the transport CLI.

I think it would be great if we can try to release this with v2.0 as it is very useful feature, that some users having been waiting a long time for. Of course, this would require that plugin developers have some time to implement the importer, so the sooner we finish this implementation, the sooner we can help them prepare their plugins.

~~P.S.: this should probably be addressed in the AEP (and I will) but we should maybe discuss the naming of "Immigrator" and "immigrating". I have merely kept this because the original concept was implemented for `PwCalculation` and was called the `PwImmigrant`. I am not sure what the reasoning behind this naming was since I wasn't there (@giovannipizzi maybe you know) but maybe "importer/importing" would be better. It sounds more neutral and technically it is also more correct. Immigrating, in my understanding at least, is more something moving as seen from the point of view of the place that it is leaving, not so much the place it is moving towards. That being said, I am not sure if naming this also importing may cause confusion with importing of archives. As said, I will probably add this to the discussion on the AEP, but thought I would mention it here as well.~~


giovannipizzi:
Thanks @sphuber!
I didn't look at the implementation yet, but I think the current interface looks fine, and I like the wrapper (I understand the need of kwargs, I think your suggestion works fine).

I'm not worried about the command line - I think each plugin can decide to create a custom CLI command that explicitly declares which kwargs are needed, and then calls `immigrate_calcjob`. In any case it's a small wrapper, and it's good to let plugin developers define it: with your implementation, the wrapper will be minimal, and can allow the developer to define custom parameters and perform some validation or logic on the kwargs before passing them to `immigrate_calcjob`.

Finally: the name was proposed by Eric Hontz, who first designed it for Quantum ESPRESSO and contributed it (for QE) to AiiDA (AiiDA core at the time, then moved into aiida-quantumespresso), back in 2014-2015. 
I admit I never liked it, so I would also vote for changing it (ideally to something different from "importing", indeed, to avoid confusion with AiiDA archives, but I'm not sure I have a great idea at the moment. Some random thoughts for the verb that could denote the action: inject, reconstruct, replicate, mirror, mock.
So e.g. get_replicator? get_reconstructor?

sphuber:
Note that I have updated the AEP to change the naming officially from "immigrator" to "importer". I am waiting with adapting this implementation until we have discussed and approved this decision in order to prevent unnecessary work should it be rejected.

PR #5087, opened 2021-08-18
🔧 MAINTAIN: Update pylint version (& fix new issues)

### LINKED ISSUE #4647: Unpin doc/dev dependencies

chrisjsewell:
Update pylint pinning to v2.9 and remove restrictions on astroid and pylint-django.

- `django-not-available` -> `django-not-configured`
- Fix new categories: `consider-using-dict-items`, `consider-using-enumerate`, `too-many-function-args`, `consider-using-from-import`, `consider-using-max-builtin`, `arguments-renamed`, `consider-using-with`, `use-maxsplit-arg`, `use-a-generator`, `deprecated-decorator`
- extra disables required for `no-member` (pylint seems to be getting worse at identifying members, or at least more strict)
- disable `unsubscriptable-object` for `cif.values`

closes #4647

### DIFF
--- a/aiida/cmdline/commands/cmd_archive.py
+++ b/aiida/cmdline/commands/cmd_archive.py
@@ -436,10 +436,11 @@ def _import_archive(archive: str, web_based: bool, import_kwargs: dict, try_migr
         if web_based:
             echo.echo_info(f'downloading archive: {archive}')
             try:
-                response = urllib.request.urlopen(archive)
+                with urllib.request.urlopen(archive) as response:
+                    temp_folder.create_file_from_filelike(response, 'downloaded_archive.zip')
             except Exception as exception:
                 _echo_exception(f'downloading archive {archive} failed', exception)
-            temp_folder.create_file_from_filelike(response, 'downloaded_archive.zip')
+
             archive_path = temp_folder.get_abs_path('downloaded_archive.zip')
             echo.echo_success('archive downloaded, proceeding with import')
 
--- a/aiida/cmdline/commands/cmd_archive.py
+++ b/aiida/cmdline/commands/cmd_archive.py
@@ -254,6 +254,7 @@ def migrate(input_file, output_file, force, in_place, archive_format, version, v
 
 class ExtrasImportCode(Enum):
     """Exit codes for the verdi command line."""
+    # pylint: disable=invalid-name
     keep_existing = 'kcl'
     update_existing = 'kcu'
     mirror = 'ncu'
--- a/aiida/cmdline/commands/cmd_daemon.py
+++ b/aiida/cmdline/commands/cmd_daemon.py
@@ -158,7 +158,7 @@ def logshow():
 
     try:
         currenv = get_env_with_venv_bin()
-        process = subprocess.Popen(['tail', '-f', client.daemon_log_file], env=currenv)
+        process = subprocess.Popen(['tail', '-f', client.daemon_log_file], env=currenv)  # pylint: disable=consider-using-with
         process.wait()
     except KeyboardInterrupt:
         process.kill()
--- a/aiida/cmdline/commands/cmd_data/cmd_show.py
+++ b/aiida/cmdline/commands/cmd_data/cmd_show.py
@@ -220,7 +220,7 @@ def _show_xmgrace(exec_name, list_bands):
             'agr', setnumber_offset=current_band_number, color_number=(iband + 1 % MAX_NUM_AGR_COLORS)
         )
         # write a tempfile
-        tempf = tempfile.NamedTemporaryFile('w+b', suffix='.agr')
+        tempf = tempfile.NamedTemporaryFile('w+b', suffix='.agr')  # pylint: disable=consider-using-with
         tempf.write(text)
         tempf.flush()
         list_files.append(tempf)
--- a/aiida/cmdline/commands/cmd_run.py
+++ b/aiida/cmdline/commands/cmd_run.py
@@ -114,7 +114,7 @@ def run(scriptname, varargs, auto_group, auto_group_label_prefix, exclude, inclu
 
     try:
         # Here we use a standard open and not open, as exec will later fail if passed a unicode type string.
-        handle = open(scriptname, 'r')
+        handle = open(scriptname, 'r')  # pylint: disable=consider-using-with
     except IOError:
         echo.echo_critical(f"Unable to load file '{scriptname}'")
     else:
--- a/aiida/cmdline/commands/cmd_status.py
+++ b/aiida/cmdline/commands/cmd_status.py
@@ -130,7 +130,7 @@ def verdi_status(print_traceback, no_rmq):
         delete_stale_pid_file(client)
         daemon_status = get_daemon_status(client)
 
-        daemon_status = daemon_status.split('\n')[0]  # take only the first line
+        daemon_status = daemon_status.split('\n', maxsplit=1)[0]  # take only the first line
         if client.is_daemon_running:
             print_status(ServiceStatus.UP, 'daemon', daemon_status)
         else:
--- a/aiida/cmdline/params/types/identifier.py
+++ b/aiida/cmdline/params/types/identifier.py
@@ -10,7 +10,7 @@
 """
 Module for custom click param type identifier
 """
-from abc import ABC, abstractproperty
+from abc import ABC, abstractmethod
 
 import click
 
@@ -65,7 +65,8 @@ class IdentifierParamType(click.ParamType, ABC):
                 else:
                     self._entry_points.append(entry_point)
 
-    @abstractproperty
+    @property
+    @abstractmethod
     @with_dbenv()
     def orm_class_loader(self):
         """
--- a/aiida/cmdline/params/types/path.py
+++ b/aiida/cmdline/params/types/path.py
@@ -99,7 +99,8 @@ class PathOrUrl(click.Path):
     def checks_url(self, url, param, ctx):
         """Check whether URL is reachable within timeout."""
         try:
-            urllib.request.urlopen(url, timeout=self.timeout_seconds)  # pylint: disable=consider-using-with
+            with urllib.request.urlopen(url, timeout=self.timeout_seconds):
+                pass
         except (urllib.error.URLError, urllib.error.HTTPError, timeout):
             self.fail(
                 '{0} "{1}" could not be reached within {2} s.\n'
--- a/aiida/cmdline/params/types/path.py
+++ b/aiida/cmdline/params/types/path.py
@@ -99,7 +99,7 @@ class PathOrUrl(click.Path):
     def checks_url(self, url, param, ctx):
         """Check whether URL is reachable within timeout."""
         try:
-            urllib.request.urlopen(url, timeout=self.timeout_seconds)
+            urllib.request.urlopen(url, timeout=self.timeout_seconds)  # pylint: disable=consider-using-with
         except (urllib.error.URLError, urllib.error.HTTPError, timeout):
             self.fail(
                 '{0} "{1}" could not be reached within {2} s.\n'
@@ -143,7 +143,7 @@ class FileOrUrl(click.File):
     def get_url(self, url, param, ctx):
         """Retrieve file from URL."""
         try:
-            return urllib.request.urlopen(url, timeout=self.timeout_seconds)
+            return urllib.request.urlopen(url, timeout=self.timeout_seconds)  # pylint: disable=consider-using-with
         except (urllib.error.URLError, urllib.error.HTTPError, timeout):
             self.fail(
                 '"{0}" could not be reached within {1} s.\n'
--- a/aiida/orm/nodes/data/data.py
+++ b/aiida/orm/nodes/data/data.py
@@ -59,6 +59,7 @@ class Data(Node):
 
         :returns: an unstored clone of this Data node
         """
+        # pylint: disable=no-member
         import copy
 
         backend_clone = self.backend_entity.clone()
--- a/aiida/orm/nodes/data/data.py
+++ b/aiida/orm/nodes/data/data.py
@@ -59,13 +59,12 @@ class Data(Node):
 
         :returns: an unstored clone of this Data node
         """
-        # pylint: disable=no-member
         import copy
 
         backend_clone = self.backend_entity.clone()
         clone = self.__class__.from_backend_entity(backend_clone)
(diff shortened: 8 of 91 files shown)


chrisjsewell:
Also I removed `useless-suppressions` from the disabled list (and even tried putting it in an enable list) but pylint still does not seem to identify them. I imagine there is quite a few "dead" ones in the code base now

chrisjsewell:
there 100s of `raise-missing-from` errors, so I'm leaving that as disabled for now

aiida/cmdline/commands/cmd_archive.py
```diff
@@ -254,6 +254,7 @@ def migrate(input_file, output_file, force, in_place, archive_format, version, v
 
 class ExtrasImportCode(Enum):
     """Exit codes for the verdi command line."""
+    # pylint: disable=invalid-name
```
sphuber, line 257: Won't this now disable this for the entire file? Or is it properly scoped to the `enum`?

aiida/cmdline/commands/cmd_daemon.py
```diff
@@ -158,7 +158,7 @@ def logshow():
 
     try:
         currenv = get_env_with_venv_bin()
-        process = subprocess.Popen(['tail', '-f', client.daemon_log_file], env=currenv)
+        process = subprocess.Popen(['tail', '-f', client.daemon_log_file], env=currenv)  # pylint: disable=consider-using-with
```
sphuber, line 161: Instead of disabling, couldn't we actually not better use a with statement here?

aiida/cmdline/commands/cmd_data/cmd_show.py
```diff
@@ -220,7 +220,7 @@ def _show_xmgrace(exec_name, list_bands):
             'agr', setnumber_offset=current_band_number, color_number=(iband + 1 % MAX_NUM_AGR_COLORS)
         )
         # write a tempfile
-        tempf = tempfile.NamedTemporaryFile('w+b', suffix='.agr')
+        tempf = tempfile.NamedTemporaryFile('w+b', suffix='.agr')  # pylint: disable=consider-using-with
```
sphuber, line 223: Idem: Instead of disabling, couldn't we actually not better use a with statement here?

aiida/cmdline/commands/cmd_run.py
```diff
@@ -114,7 +114,7 @@ def run(scriptname, varargs, auto_group, auto_group_label_prefix, exclude, inclu
 
     try:
         # Here we use a standard open and not open, as exec will later fail if passed a unicode type string.
-        handle = open(scriptname, 'r')
+        handle = open(scriptname, 'r')  # pylint: disable=consider-using-with
```
sphuber, line 117: Idem: instead of disabling, couldn't we actually not better use a with statement here?

aiida/cmdline/commands/cmd_status.py
```diff
@@ -130,7 +130,7 @@ def verdi_status(print_traceback, no_rmq):
         delete_stale_pid_file(client)
         daemon_status = get_daemon_status(client)
 
-        daemon_status = daemon_status.split('\n')[0]  # take only the first line
+        daemon_status = daemon_status.split('\n', maxsplit=1)[0]  # take only the first line
```
sphuber, line 133: what does this accomplish?

aiida/cmdline/params/types/identifier.py
```diff
@@ -10,7 +10,7 @@
 """
 Module for custom click param type identifier
 """
-from abc import ABC, abstractproperty
+from abc import ABC, abstractmethod
```
sphuber, line 13: what is wrong with abstractproperty?

aiida/cmdline/params/types/path.py
```diff
@@ -99,7 +99,7 @@ def convert(self, value, param, ctx):
     def checks_url(self, url, param, ctx):
         """Check whether URL is reachable within timeout."""
         try:
-            urllib.request.urlopen(url, timeout=self.timeout_seconds)
+            urllib.request.urlopen(url, timeout=self.timeout_seconds)  # pylint: disable=consider-using-with
```
sphuber, line 102: Also here, even though not crucial, might as well just use a context manager

aiida/orm/nodes/data/data.py
```diff
@@ -59,6 +59,7 @@ def clone(self):
 
         :returns: an unstored clone of this Data node
         """
+        # pylint: disable=no-member
```
sphuber, line 62: should maybe be more specific, on the exact line?

chrisjsewell:
Intend to merge this in the next day, as it will affect all other PRs

aiida/orm/utils/loaders.py
```diff
@@ -8,7 +8,7 @@
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
 """Module with `OrmEntityLoader` and its sub classes that simplify loading entities through their identifiers."""
-from abc import abstractclassmethod
+from abc import abstractmethod
```
sphuber, line 11: same as `abstractproperty` what is wrong with this?

aiida/tools/data/cif.py
```diff
@@ -162,6 +162,7 @@ def refine_inline(node):
 
     .. note:: can be used as inline calculation.
     """
+    # pylint: disable=unsubscriptable-object
```
sphuber, line 165: which line does this come from? I know there is a bug in `pylint` in combination with Python 3.9 that throws these as false positives

sphuber (commented):
Have a couple of questions about certain disables. Initially repeated them at each line that I though they applied, but stopped after a while because there are so many instances. If you agree that in certain places we should actually fix instead of disabling, please have a look at other instances as well

aiida/cmdline/params/types/identifier.py
chrisjsewell: It's been deprecated since python 3.3: https://docs.python.org/3/library/abc.html?highlight=abstractproperty#abc.abstractproperty

aiida/orm/utils/loaders.py
chrisjsewell: again deprecated: https://docs.python.org/3/library/abc.html?highlight=abstractproperty#abc.abstractclassmethod

aiida/cmdline/commands/cmd_status.py
chrisjsewell: Instead of splitting `"a\nb\nc" in to `["a", "b", "c"]`, it will split only on the first occurrence then return, e.g. `["a", "b\nc"]`, which is more performant if you only need the first item. Same with `rsplit`, if you only need the last item

aiida/cmdline/commands/cmd_archive.py
chrisjsewell: yeh, I was hopping it would be scoped to the `enum`, or do you think not?

aiida/cmdline/commands/cmd_daemon.py
chrisjsewell: It wasn't quickly apparent to me how to do this and keep the same behaviour. i.e. is the same?

```
currenv = get_env_with_venv_bin()
with subprocess.Popen(['tail', '-f', client.daemon_log_file], env=currenv) as process:
   try:
     process.wait()
   except KeyboardInterrupt:
     process.kill()
```

probably not 🤷 


same with the other `disable=consider-using-with`

aiida/cmdline/commands/cmd_data/cmd_show.py
chrisjsewell: as above, what code would you propose?

aiida/cmdline/commands/cmd_run.py
chrisjsewell: as above; basically I would maybe open a separate issue to look at removing all the `disable=consider-using-with`, because some of them feel like they need a bit of scrutiny to double check they won't break the current behaviour

aiida/cmdline/params/types/path.py
chrisjsewell, line 102:
```suggestion
            with urllib.request.urlopen(url, timeout=self.timeout_seconds):
                pass
```

aiida/tools/data/cif.py
chrisjsewell: done

aiida/cmdline/params/types/path.py
chrisjsewell: done

aiida/orm/nodes/data/data.py
chrisjsewell: done

chrisjsewell:
addressed your comments thanks @sphuber

aiida/schedulers/plugins/lsf.py
```diff
@@ -151,6 +151,15 @@ def accepts_default_mpiprocs_per_machine(cls):
...
 
+    @classmethod
+    def validate_resources(cls, **kwargs):
+        """Validate the resources against the job resource class of this scheduler.
+
+        :param kwargs: dictionary of values to define the job resources
+        :return: attribute dictionary with the parsed parameters populated
+        :raises ValueError: if the resources are invalid or incomplete
+        """
+
```
chrisjsewell, line 162: Guessing `@abstractclassmethod` didn't actually work, since this was not picked up before? @sphuber I guess this should probably have some content

aiida/cmdline/commands/cmd_data/cmd_show.py
sphuber: I see now. Didn't notice it was creating a list of handles and then processing them. What the hell :man_facepalming:  OK, fine to leave this as is for now

aiida/cmdline/commands/cmd_daemon.py
sphuber: Fair enough. I would be tempted to do:
```
currenv = get_env_with_venv_bin()
try:
    with subprocess.Popen(['tail', '-f', client.daemon_log_file], env=currenv) as process:
        process.wait()
except KeyboardInterrupt:
    process.kill()
```
But I am not sure the try-except is even necessary when using the context manager. Keyboard interrupt should just kill it and the context manager should clean it up properly. Anyway, indeed fine to keep this as is as it is not a trivial change

aiida/cmdline/commands/cmd_run.py
sphuber: Good idea to open an issue and leave all of these with disable statements for now

aiida/cmdline/params/types/identifier.py
sphuber: Huh, you don't say. I wonder why that warning never showed up? I remember the deprecation warnings from the `collections` module showing up, so I don't think that by default deprecation messages from built-in modules don't show. Anyway, good to keep these changes then :+1:

aiida/cmdline/commands/cmd_archive.py
sphuber: Not sure to be honest. You could give it a quick try by adding a global constant after the enum declaration which should then incur a warning from pylint if it is properly scoped.

aiida/cmdline/commands/cmd_status.py
sphuber: I see, probably a nano-optimization, but I will allow it ;)

aiida/schedulers/plugins/lsf.py
sphuber: Yeah that is weird indeed. I guess the validation is already done in the constructor. We can just move that logic to the `validate_resources`. To be safe as to not break anything, you can always still call `self.validate_resources(**kwargs)` from the constructor, which is what the `NodeNumberJobResource` also does.

sphuber (approved):
Thanks @chrisjsewell . Would be good if you can still check the comment on the Enum and the scheduler job resource validation, but am already approving this

aiida/cmdline/commands/cmd_status.py
chrisjsewell: Well that lucky, because pylint won't 😜

aiida/cmdline/commands/cmd_archive.py
chrisjsewell: Yep looks to work as intended, FYI scoping is detailed here: http://pylint.pycqa.org/en/latest/user_guide/message-control.html#block-disables

aiida/schedulers/plugins/lsf.py
chrisjsewell: done

chrisjsewell:
all done cheers

Issue #2577, opened 2019-03-06
Allow usage of `pk` in `QueryBuilder` project and filter statements


sphuber:
The database models use `id` as the column name for the primary key, but unfortunately, `id` is a reserved keyword in python. This is also the reason that Django uses `pk` to refer to the primary key column. In AiiDA we are using a mixture. Some of the ORM classes define both the `id` and the `pk` property, where the `pk` property just proxies through to the `id` one.

Ideally we would change the database models to use only `pk` and then we would only use `pk` throughout the code. Even just restricting the usage of `id` to the database models and using `pk` exclusively everywhere else would already be an improvement.

But the worst is that currently sometimes a user can use either name, but sometimes one *has* to use `id`. For example, attempting to filter on `pk` in the `QueryBuilder` yields the cryptic message:
```
In [3]: QueryBuilder().append(ProcessNode, filters={'pk': 6820483}, tag='parent', project=['id']).append(ProcessNode, with_ancestors='parent', project=['
   ...: id']).all()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-1269b974abcb> in <module>()
----> 1 QueryBuilder().append(ProcessNode, filters={'pk': 6820483}, tag='parent', project=['id']).append(ProcessNode, with_ancestors='parent', project=['id']).all()

/home/aiida/code/aiida/env/sdb2/aiida-core/aiida/orm/querybuilder.pyc in all(self, batch_size)
   2217         :returns: a list of lists of all projected entities.
   2218         """
-> 2219         return list(self.iterall(batch_size=batch_size))
   2220 
   2221     def dict(self, batch_size=None):

/home/aiida/code/aiida/env/sdb2/aiida-core/aiida/orm/querybuilder.pyc in iterall(self, batch_size)
   2163         :returns: a generator of lists
   2164         """
-> 2165         query = self.get_query()
   2166 
   2167         for item in self._impl.iterall(query, batch_size, self._attrkeys_as_in_sql_result):

/home/aiida/code/aiida/env/sdb2/aiida-core/aiida/orm/querybuilder.pyc in get_query(self)
   2049 
   2050         if need_to_build:
-> 2051             query = self._build()
   2052             self._hash = queryhelp_hash
   2053         else:

/home/aiida/code/aiida/env/sdb2/aiida-core/aiida/orm/querybuilder.pyc in _build(self)
[...]

Issue #2196, opened 2018-11-13
Extend QueryBuilder to support mappings of entity attribute name to a differently named column


muhrin:
At the moment the way that querybuilder works is that it takes filters and assumes that the name maps directly to a column of the same name.  This is not the case in, notably, `Group` where the `type_string` attribute is actually stored under a column called `type`.

The most trivial solution would be just to rename the column `type_string` in a migration.  However a better solution would be to have a backend implementation dependent `get_column` method that supports mapping the filter key to the correct column in the model.  This would, however, require that the concrete `QueryBuilder` gives the `get_column` more information, specifically at least the orm class Entity type.

For now, I'm just changing the relevant parts of the code (for `Group`) to use `type` instead of `type_string`.

PR #5088, opened 2021-08-19
 ✨ NEW: Add `orm.Entity.fields` interface for `QueryBuilder`

chrisjsewell:
The `QueryBuilder` currently has a conceptual problem, in that it essentially exposes the backend database to the user. A classic example of this is that, on the front end, we use `Node.pk`, but when using the `QueryBuilder`, one has to use the "backend" name `id` (always confusing for new users) (#2577, #2196).
Additionally, there is currently no "programmatic" way to infer what can be retrieved from the database for each ORM entity, which is something very desirable for the REST API (https://github.com/aiidateam/aiida-restapi/issues/1).

In this PR, I propose a new API (mainly in `aiida/orm/fields.py`), for:

1. ~~"decorating" ORM properties as database fields,~~ Adding a `__qb_fields__` class attribute containing `QbField` instances (and the `QbAttrField` subclass)
2. "gathering" these instances into a new `Entity.fields` class attribute, then
3. allowing these fields to be used when constructing `QueryBuilder` instances.

The API is backward compatible and looks like this (also see `tests/orm/test_fields.py`):

On any ORM entity, it will inherit parent fields, then if you want to add extra fields, you can add a `__fields__` class attribute or directly decorate methods:

```python
from typing import Optional
from aiida.orm import Data, QbAttrField

class NewData(Data):
   
     __fields__ = (
          QbAttrField("new_key", dtype=str, doc="A new key"),
     )

```

This class will now have a `fields` class attribute, which gives you the mapping of ORM properties to database fields:

```python
In [1]: NewData.fields
Out[1]: 
{'ctime': 'QbField(ctime) -> datetime',
 'description': 'QbField(description) -> str',
 'extras': 'QbField(extras.*) -> Dict[str, Any]',
 'label': 'QbField(label) -> str',
 'mtime': 'QbField(mtime) -> datetime',
 'new_key': 'QbAttrField(attributes.new_key) -> str',
 'node_type': 'QbField(node_type) -> str',
 'pk': 'QbField(id) -> int',
 'repository_metadata': 'QbField(repository_metadata) -> Dict[str, Any]',
 'source': 'QbAttrField(attributes.source.*) -> Union[dict, NoneType]',
 'user_pk': 'QbField(user_id) -> int',
 'uuid': 'QbField(uuid) -> str'}
In [2]: NewData.fields.new_key
Out[2]: QbAttrField('new_key', dtype=str)
```

(you can see the fields for all aiida nodes in `tests/orm/test_fields`)

Within the `QueryBuilder` you can then directly use these fields for the `project`, `filters` and `order_by` values e.g.

```python
In [4]: qb = QueryBuilder().append(Node, project=[Node.fields.pk, Node.fields.mtime], filters={Node.fields.pk: 1})
In [5]: qb.order_by({'node_1': Node.fields.mtime})

Out[5]: QueryBuilder(path=[{'entity_type': '', 'orm_base': 'node', 'tag': 'node_1', 'joining_keyword': None, 'joining_value': None, 'edge_tag': None, 'outerjoin': False}], filters={'node_1': {'node_type': {'like': '%'}, 'id': 1}}, project={'node_1': [{'id': {}}, {'mtime': {}}]}, project_map={'node_1': {'id': 'pk'}}, order_by=[{'node_1': [{'mtime': {'order': 'asc'}}]}], limit=None, offset=None, distinct=False)

# to get all fields
In [6]: QueryBuilder().append(Node, project=Node.fields)
Out[6]: QueryBuilder(path=[{'entity_type': '', 'orm_base': 'node', 'tag': 'node_1', 'joining_keyword': None, 'joining_value': None, 'edge_tag': None, 'outerjoin': False}], filters={'node_1': {'node_type': {'like': '%'}}}, project={'node_1': [{'ctime': {}, 'description': {}, 'extras': {}, 'label': {}, 'mtime': {}, 'node_type': {}, 'id': {}, 'repository_metadata': {}, 'user_id': {}, 'uuid': {}}]}, project_map={'node_1': {'id': 'pk', 'user_id': 'user_pk'}}, order_by=[], limit=None, offset=None, distinct=False)

# you can also use field comparitors to construct filters
In [7]: QueryBuilder().append(Node, project=Node.fields.pk, filters=(Node.fields.pk <= 2) & (Node.fields.pk.in_([2, 3])))
Out[7]: QueryBuilder(path=[{'entity_type': '', 'orm_base': 'node', 'tag': 'node_1', 'joining_keyword': None, 'joining_value': None, 'edge_tag': None, 'outerjoin': False}], filters={'node_1': {'node_type': {'like': '%'}, 'id': {'and': [{'<=': 2}, {'in': {2, 3}}]}}}, project={'node_1': [{'id': {}}]}, project_map={'node_1': {'id': 'pk'}}, order_by=[], limit=None, offset=None, distinct=False)

# fields can be designated as subscribtable to index into them
In [8]: QueryBuilder().append(Dict, project=Dict.fields.dict["subkey"])
Out[8]: QueryBuilder(path=[{'entity_type': 'data.core.dict.Dict.', 'orm_base': 'node', 'tag': 'Dict_1', 'joining_keyword': None, 'joining_value': None, 'edge_tag': None, 'outerjoin': False}], filters={'Dict_1': {'node_type': {'like': 'data.core.dict.%'}}}, project={'Dict_1': [{'attributes.subkey': {}}]}, project_map={'Dict_1': {'attributes.subkey': 'dict.subkey'}}, order_by=[], limit=None, offset=None, distinct=False)
```

then run the query as normal.

If you run `qb.dict()`/`qb.iterdict()`, it will correctly give you the front end projection keys, e.g.

```python
In [1]: QueryBuilder().append(Dict, project=[Dict.fields.pk, Dict.fields.dict["a"]]).dict()
Out[1]: 
[{'Dict_1': {'pk': 15, 'dict.a': 1}},
 {'Dict_1': {'pk': 7, 'dict.a': 1}},
 {'Dict_1': {'pk': 3, 'dict.a': None}}]
```

~Probably the main limitation at present, is that the output by `qb.dict()`/`qb.iterdict()` will still return the backend keys, e.g. `id` not `pk`~ (a4ee061)

Some other things that come to mind:

- ~For e.g. `Dict.fields.dict` we would ideally like to be able to do e.g. `Dict.fields.dict.key1`~ (use `subscriptable=True` when defining field)
- ~Some fields are missing for joined ids e.g. `Node.user_id`~ (added `foreign_key` field kwarg)
- ~Composite fields, e.g. in `StructureData` you have `pbc` -> `[attributes.pbc1, attributes.pbc2, attributes.pbc3]` in the database~ (can now use `__qb_fields__`)
- ~Field post-conversions (after having been returned from the database ORM), e.g. list to numpy array~ (possible, but not adding to this PR)
- ~~You have to use some `tyep: ignore[misc]` on the decorators currently, due to https://github.com/python/mypy/issues/1362~~
- ~Obviously, also need to add tests~ (added `tests/orm/test_fields.py`)

Definitely love to have some feedback on this cc @sphuber, @giovannipizzi, @ltalirz, @mbercx, @ramirezfranciscof


chrisjsewell:
another though would be in some way to tie the fields to the backend, e.g. `Node.fields.name` would return a dynamic field name to use for querying based on the backend. This is certainly possible, and could be added at a later date without changing the user API

ltalirz:
thanks @chrisjsewell! 
I definitely agree that a route for accessing database properties of any given node programmatically through the frontend API is badly needed.

One question that comes to mind is how much tangible benefit the separation between frontend and backend will really still bring once we've dropped one of the two backends. But I guess that's still a bit further down the line and more controversial, so by adding this mapping explicitly and making it available for programmatic use I think we are making a step in the right direction.

Thinking about how to use this in e.g. the REST API, and your open question:

> * For e.g. `Dict.fields.dict` we would ideally like to be able to do e.g. `Dict.fields.dict.key1`

If I see correctly, you are using python's type hints to write the "schema" here, which I guess might be limiting when you want to represent more elaborate constraints than data types, such as specific field names in a dictionary that are optional; allowed value ranges (which could also be very interesting to reuse in the verdi CLI) as well as documentation of the fields.

In the aiida-restapi package we use pydantic for this - do you think something like this would make sense here as well?

chrisjsewell:
Another, another thought would be, perhaps you could also extend this to specifying repository attributes, e.g.

```python
@has_fields
class NewData(orm.Data):
   @property
   @field_method("path/to/file", in_repo=True)
   def y(self) -> str:
      return self.get_object_content("path/to/file")
```

chrisjsewell:
thanks @ltalirz 

> you are using python's type hints to write the "schema" here, which I guess might be limiting when you want to represent more elaborate constraints than data types, such as specific field names in a dictionary that are optional

Well I guess here you could just add additional key-words to `field_method`, e.g.

```python
   @field_method("attributes.x", min=0, max=100)
   def x(self) -> Optional[int]:
      return self.get_attribute("x", None)
```

> In the aiida-restapi package we use pydantic for this - do you think something like this would make sense here as well?

I feel the drawback of pydantic is that it does not work well with sub-classing.
Note my method takes some inspiration from https://github.com/python-attrs/attrs and https://docs.python.org/3/library/dataclasses.html

ltalirz:
thanks, I also recently came across `attrs` in a different context - looks pretty cool.

I guess what I'm saying is that when making this change it would be great if you could include with it one small example code of how e.g. the REST API (or the verdi cli) could take advantage of this interface, removing the duplicated schema information that currently resides there (explicitly in the case of the aiida-restapi; implicitly in the verdi cli).

Just to make sure that we're achieving this goal - or if we are not to be aware of what's missing and how we can get there.

chrisjsewell:
> Just to make sure that we're achieving this goal - or if we are not, to be aware of what's still missing and how we can get there.

Yep fair. I don't really want to start rewriting the old REST API though lol, do you have any CLI commands in mind where this could be useful?

chrisjsewell:
(in 74f39f2 I have removed the need for using `has_fields`, and introduced `__fields__`, will update initial comment)

aiida/orm/nodes/process/process.py
```diff
@@ -134,6 +144,23 @@ def process_class(self) -> Type['Process']:
 
         return process_class
 
+    @property  # type: ignore
+    @field_method()
+    def process_type(self) -> Optional[str]:
```
chrisjsewell, line 149: I think `process_type` should be removed from the base `Node` class, since obviously it is just a process specific field and an implementation detail that it is also contained in the database table for all nodes

chrisjsewell: (same for `computer`, that is only relevant to code amd process nodes)

chrisjsewell:
Ok added just about all the fields for all the node types.
It does make it so much easier, in e.g. `verdi shell`, to find what you can query for:

image

image

aiida/orm/nodes/process/process.py
giovannipizzi: No, computer should stay for all nodes I think, some other nodes use it (specifically: RemoteData). Default = None.

giovannipizzi:
Thanks @chrisjsewell, in general I think this was a very important missing feature.

Just a few quick questions:
- to double check: the QueryBuilder interface is backward compatible, meaning you can still use strings to project etc., or now one has to use the new syntax only?
- Once this is merged, would the plan to deprecate the old syntax at some point, or not? Or at least to prioritise one over the other in 1) the documentation, and 2) the usage in the code itself (e.g. all queries in the cmdline)
- Is the limitation you mention ("qb.dict()/qb.iterdict() will still return the backend keys, e.g. id not pk") hard to overcome? I imagine that, in order to go to a fully machine-usable querybuilder, knowing in a reliable way which keys will be returned when one uses .dict() is important?

chrisjsewell:
> to double check: the QueryBuilder interface is backward compatible, meaning you can still use strings to project etc., or now one has to use the new syntax only?

Yes, fully backward compatible

>  nce this is merged, would the plan to deprecate the old syntax at some point, or not? Or at least to prioritise one over the other in 1) the documentation, and 2) the usage in the code itself (e.g. all queries in the cmdline)

certainly to prioritise this syntax, then, once people have had a good chance to play around with it, I would indeed consider deprecating

> Is the limitation you mention ("qb.dict()/qb.iterdict() will still return the backend keys, e.g. id not pk") hard to overcome? I imagine that, in order to go to a fully machine-usable querybuilder, knowing in a reliable way which keys will be returned when one uses .dict() is important?

I think with #5093 there is definitely a way forward with this

aiida/orm/nodes/process/process.py
sphuber: It should definitely stay on the `Node` level.

aiida/orm/fields.py
```diff
@@ -0,0 +1,331 @@
...
+        return deepcopy(self._fields)
+
+
+def field_method(
+    db_field: Optional[str] = None,
+    *,
+    key: Optional[str] = None,
+    dtype: Optional[Type[Any]] = None,
+    foreign_key: Optional[str] = None,
+    subscribtable: bool = False
```
sphuber, line 267: Think this should be `subscriptable` with a 'p' shouldn't it?

aiida/orm/fields.py
```diff
@@ -0,0 +1,331 @@
...
+):  # pylint: disable=unused-argument
+    """A method decorator for database fields of an AiiDA entity.
+
+    :param db_field: The name of the database field. If not given, the name of the method is used.
```
sphuber, line 271:
```suggestion
    :param db_field: The name of the database field. If not given, the name of the method is used, unless `key` is specified.
```

sphuber (commented):
Thanks @chrisjsewell . I think the concept proposed here is very useful and badly needed indeed. I went through the changes and most of it looks fine, with some minor comments here and there, but I also still have some more generic questions about the design and the implications.

With the current design, we are effectively coupling the front-end directly to implementation details in the backend. By having to specify the database column in the front-end, the separation that we had will be gone. I am not saying that this would necessarily be a deal-breaker, since anyway we might be getting rid of the need of a separation if we're getting rid of Django, but I wonder if it is good and necessary in principle. What we really need in the front-end for the user is what you summarized very succinctly elsewhere:

>  keys that relate to "queryable" fields, that are read-only once the node is stored
    keys that relate to "queryable" fields, that are read/write always (i.e. extras)
    keys that relate to "non-queryable" fields

I would maybe add "queryable" but "private" fields that AiiDA stores in the database (hash etc.). This is an effective implementation-free description of what is stored in the repository and what in the database. Could we not move the concept of `OrmFields` to the backend layer and have the front-end `orm.Entity` automatically detect the defined fields on the `BackendEntity` that it contains. This would allow us to later extend the concept of fields in the front-end `Entity` with non-queryable fields, which could be data stored in the repository.

A second question is about consistency. There are two methods of defining a field to an ORM class: by explicitly declaring it in `__fields__` or by decorating a class method or property. This is mostly a problem of historicity and having to keep backwards compatibility, but if we were to design the ORM from scratch, wouldn't we ideally just define the fields explicitly through `__fields__` and have those generate properties dynamically. This would save a lot of code typing and would guarantee a uniform interface. My question is now that if you agree with this, and if so, if we should maybe already try to take this approach and drop the decorator. We can then also deprecate the methods that happen to not follow the naming convention of the automatically and dynamically named getters. There can of course still be custom methods if they need to perform additional logic to just retrieving a database value and returning it.

I think this approach would also prevent inconsistencies in to what fields _are_ marked as fields, and which aren't. Reading through the code, I think there are quite a few entities that do not have certain fields defined, even though they probably should. `Computer.user` is one of them for example. Of course we could always thoroughly go through the code before merging to make sure everything is there, but I was wondering if the other approach may prevent the potential for missing fields.

chrisjsewell:
Thanks for the feedback @sphuber 
some quick mobile points:
(now in a McDonald’s waiting for a train from Zagreb to Munich lol)

> With the current design, we are effectively coupling the front-end directly to implementation details in the backend.
> Could we not move the concept of OrmFields to the backend layer and have the front-end orm.Entity

I certainly agree we want to push good abstractions 👍. Obviously the query builder was actually abstracted in the first place, in the sense you request/receive directly backend columns. I’ll have a think if there is a better abstraction, but a key thing here is that I would like this API to be available to be available to plugin developers, to define their own fields on added data nodes, which I think would not be possible if moved to the backend

> but if we were to design the ORM from scratch, wouldn't we ideally just define the fields explicitly through __fields__ 

yeh interesting, possibly. With the current implementation, I was sort of thinking along the lines of how sqlalchemy works; having field attributes, which auto-generate the __table__ attribute. What you suggest would be, I guess an inversion of this. The only issue I foresee is with static analysis and instance attribute auto completion, since then all these would be dynamically generated

aiida/orm/nodes/data/structure.py
```diff
@@ -716,6 +718,12 @@ class StructureData(Data):
     _dimensionality_label = {0: '', 1: 'length', 2: 'surface', 3: 'volume'}
     _internal_kind_tags = None
 
+    __fields__ = (
+        OrmField('pbc1', 'attributes.pbc1',
+                 dtype=bool), OrmField('pbc2', 'attributes.pbc2',
+                                       dtype=bool), OrmField('pbc3', 'attributes.pbc3', dtype=bool)
```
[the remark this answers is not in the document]
chrisjsewell: One word: black 😉

aiida/orm/nodes/process/calculation/calcjob.py
```diff
@@ -347,6 +352,7 @@ def set_scheduler_state(self, state: 'JobState') -> None:
         self.set_attribute(self.SCHEDULER_STATE_KEY, state.value)
         self.set_attribute(self.SCHEDULER_LAST_CHECK_TIME_KEY, timezone.datetime_to_isoformat(timezone.now()))
 
+    @field_method(f'attributes.{SCHEDULER_STATE_KEY}', key='scheduler_state', dtype=str)
     def get_scheduler_state(self) -> Optional['JobState']:
```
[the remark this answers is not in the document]
chrisjsewell: Yeh, I like the idea of having the field “close” to the method, as it feels like it will make things like refactoring easier, but indeed the special case discrepancies are not ideal. I guess you will still have this problem if, as you mention, we want to look to auto generate the methods from the fields

chrisjsewell:
Ok, in be43aec I have removed the `field_method` decorator method, instead leaving only the single `__qb_fields__` class attribute mechanism for specifying fields.
As @sphuber, it's over-complex to have two mechanisms for specifying these fields.

I have also renamed `__fields__` -> `__qb_fields__` and `OrmField` -> `QbField`, to make it clear that these are fields that can be returned from the `QueryBuilder`.
Lastly, I have added the `QbAttrField` subclass, as a shorthand for specifying keys in the attributes field. This is what data subclasses use to denote data they store in the attributes.

aiida/orm/nodes/process/calculation/calcjob.py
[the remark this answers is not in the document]
chrisjsewell: See https://github.com/aiidateam/aiida-core/pull/5088#issuecomment-994422493

aiida/orm/nodes/data/structure.py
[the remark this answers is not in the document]
chrisjsewell: fixed

aiida/orm/nodes/process/calculation/calcjob.py
```diff
@@ -31,19 +32,22 @@
 
 __all__ = ('CalcJobNode',)
 
+SCHEDULER_STATE_KEY = 'scheduler_state'
+CALC_JOB_STATE_KEY = 'state'
```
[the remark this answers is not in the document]
chrisjsewell: Actually, I realised this change was not necessary, so have reverted it

aiida/orm/nodes/process/process.py
chrisjsewell: reverted this change as no longer necessary

edan-bainglass:
Reviewing this PR. Tests passing where it is currently. I'm now incrementally moving (rebasing) it towards v2.5. 

I've made it as far as #6141. A few minor adjustments along the way. Will discuss these at a later time. 

I'm now running into an issue with #6134. Specifically, the change made to `test_querybuilder.py` raises the following error post-test (test does pass):

```
--- Logging error ---
  File "/home/edanb/PSI/AiiDA/aiida-core/.tox/py39/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 763, in _finalize_fairy
    fairy._reset(pool, transaction_was_reset)
  File "/home/edanb/PSI/AiiDA/aiida-core/.tox/py39/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 1038, in _reset
    pool._dialect.do_rollback(self)
  File "/home/edanb/PSI/AiiDA/aiida-core/.tox/py39/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 683, in do_rollback
    dbapi_connection.rollback()
psycopg2.OperationalError: server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.


  File "/home/edanb/miniconda3/envs/pr-qb/lib/python3.9/logging/__init__.py", line 1086, in emit
    stream.write(msg + self.terminator)
ValueError: I/O operation on closed file.
Call stack:
  File "/home/edanb/PSI/AiiDA/aiida-core/.tox/py39/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 509, in <lambda>
    and _finalize_fairy(
  File "/home/edanb/PSI/AiiDA/aiida-core/.tox/py39/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 791, in _finalize_fairy
    pool.logger.error(
Message: 'Exception during reset or similar'
Arguments: ()
```

Error not raised when test is commented out, so it's this specific test. I checked out the commit (excluding this PR) and ran the test. Same error. Good if someone else can checkout the commit and test, see if it's a local issue. But also, @sphuber comments?

edan-bainglass:
Okay. Caught up to `main` (`e1255ce3`). All good other than the #6134 issue that needs review.

aiida/orm/implementation/querybuilder.py
```diff
@@ -77,6 +77,8 @@ class QueryDictType(TypedDict):
     # mapping: tag -> [] -> field -> 'func' -> 'max' | 'min' | 'count'
     #                                'cast' -> 'b' | 'd' | 'f' | 'i' | 'j' | 't'
     project: Dict[str, List[Dict[str, Dict[str, Any]]]]
+    # mapping: tag -> field -> return key for iterdict method
+    project_map: Dict[str, Dict[str, str]]
```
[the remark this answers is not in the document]
edan-bainglass: @chrisjsewell what's the plan here? I see a thumbs up but not an implementation.

sphuber:
Thanks @edan-bainglass . You are not yet up-to-date with `main`. Note that a few weeks ago (after the release) the `aiida` directory was moved to `src/aiida`. Maybe it is just that, that you haven't moved the changes in yet. If you need help fixing it, let me know

edan-bainglass:
Huh 🤔 I was sure I included the move to src (between v2.5.0 and main). Maybe I didn't push? I'll check in a bit. Out replenishing stock 🛒

sphuber:
> So, with that out of the way, what is the policy? Should I PR to Chris's fork, or just push my rebasing directly to his repo? Also, there are still comments on the PR that I am awaiting response. @chrisjsewell, if you can please address these, that'd be super 🙏

There is no real existing policy. Depends on what @chrisjsewell prefers. But if he is no longer interested or doesn't have the time, which I can understand, maybe we can continue with your fork. We close this PR and you open a new one and continue there. That might be easiest

edan-bainglass:
Sure. If @chrisjsewell has no objections, we can proceed as you suggest.

PR #5091, opened 2021-08-20
CI: run certain workflows only on main repo, not on forks

### LINKED ISSUE #5068: Push to DockerHub action should not run (and fail) on forks
On forks, the [push_image_to_dockerhub](https://github.com/aiidateam/aiida-core/blob/develop/.github/workflows/push_image_to_dockerhub.yml) GitHub actions is run (and fails since it does not have the tokens to login in docker hub), see e.g.:
https://github.com/giovannipizzi/aiida_core/runs/3310106205?check_suite_focus=true

We should probably add a conditional check on `on` that says to run only for the main repo?

sphuber:
Fixes #5068

The following workflows should only be run on the main repo:

 * `build_and_test_docker_on_pr.yml`
 * `post-release.yml`
 * `push_image_to_dockerhub.yml`
 * `release.yml`

### DIFF
--- a/.github/workflows/build_and_test_docker_on_pr.yml
+++ b/.github/workflows/build_and_test_docker_on_pr.yml
@@ -16,6 +16,9 @@ jobs:
 
   build-and-test:
 
+    # Only run this job on the main repository and not on forks
+    if: github.repository == 'aiidateam/aiida-core'
+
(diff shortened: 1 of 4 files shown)


giovannipizzi (approved):
Thanks!

PR #5092, opened 2021-08-21
🔧 MAINTAIN: Make `DjangoQueryBuilder` a subclass of `SqlaQueryBuilder`

chrisjsewell:
Rather than duplicating all the code across the backends,
since DjangoQueryBuilder is just SqlaQueryBuilder+aldjemy,
subclass and only override the differences.

Also `get_filter_expr` was missing from the base `BackendQueryBuilder`


sphuber (approved):
Good to go, thanks @chrisjsewell

PR #5093, opened 2021-08-24
♻️ REFACTOR: Fully abstract QueryBuilder

### LINKED ISSUE #4327: QueryBuilder 'order_by' has no effect if the same type is used twice
### Describe the bug

When the same type is `append`ed to the QueryBuilder twice (with difference filter clauses etc.), the `order_by` statement appears to have no effect.

### Minimal example

Running the following code, the result does not respect the `ctime` ordering:
```python
from aiida import orm
from aiida.engine import calcfunction


@calcfunction
def add(x, y):
    return x + y


x = orm.Int(1).store() # explicitly storing so that I know the order of creation
y = orm.Int(3).store()
z = orm.Int(5).store()

c = add(add(x, y), z)

qb = orm.QueryBuilder()
qb.append(orm.Int, filters={'uuid': c.uuid}, tag='c')
qb.append(orm.Int, with_descendants='c')
qb.order_by({orm.Int: {'ctime': 'desc'}})
print(qb.distinct().all())
```

The printed output is (formatted for readability)
```
[
  [<Int: uuid: 7c1a439c-9c04-419e-81f9-d4b6cb2de073 (pk: 59793) value: 1>], 
  [<Int: uuid: 3fb15a95-81a9-48e3-9ca7-359738de82d8 (pk: 59794) value: 3>], 
  [<Int: uuid: 8459de2a-eeb2-45aa-9d8a-6cd86d27a5ac (pk: 59795) value: 5>], 
  [<Int: uuid: 41f522fa-a473-46e6-be95-b5774e46b074 (pk: 59797) value: 4>]
]
```
which is in _ascending_ `ctime` order. In fact, switching from `desc` to `asc` has no effect.

### Counterexample: using a different type

Using a different type for the two `append` statements, the ordering works as expected:

```python
from aiida import orm
from aiida.engine import calcfunction


@calcfunction
def add(x, y):
    return orm.Float((x + y).value)


x = orm.Int(1).store()
y = orm.Int(3).store()
z = orm.Int(5).store()

c = add(add(x, y), z)

qb = orm.QueryBuilder()
qb.append(orm.Float, filters={'uuid': c.uuid}, tag='c')
qb.append(orm.Int, with_descendants='c')
qb.order_by({orm.Int: {'ctime': 'desc'}})
print(qb.distinct().all())
```

Result:

```
[
  [<Int: uuid: 59350c15-1421-4ca1-ba8a-7da60815030f (pk: 59802) value: 5>], 
  [<Int: uuid: d782fbec-85cf-4c7a-8b0d-a529d17625f1 (pk: 59801) value: 3>], 
  [<Int: uuid: f04c16eb-91f5-4f76-89b7-a268ef747de1 (pk: 59800) value: 1>]
]
```

### Environment

- Operating system: Ubuntu 18.04, via WSL2
- Python version: 3.7.3
- aiida-core version: Latest `develop` (477fe305ede8ee51cc9a8152ef3edc90ca8c4972), Django backend

### Workaround

A workaround is using the `tag` and referring to the entity that should be ordered with that, instead of passing the type. If the above is expected behavior, I think the `order_by` documentation should strongly encourage the use of tags.

chrisjsewell:
Fixes #4327 

There are four primary modules involved in the QueryBuilder abstraction: 

- `aiida/orm/implementation/django/querybuilder.py::DjagoQueryBuilder`
- `aiida/orm/implementation/sqlalchemy/querybuilder.py::SqlaQueryBuilder`
- `aiida/orm/implementation/querybuilder.py::BackendQueryBuilder`
- `aiida/orm/querybuilder.py::QueryBuilder`

Prior to #5092, all of these modules imported objects from SQLAlchemy, i.e. there was actually no abstraction, since everything was tightly coupled to SQLAlchemy.
As well as this lack of abstraction, the spread of logic and state across these modules (as well as the lack of typing and non-uniform method naming) made it very difficult to decipher the workings of the code.

This PR continues from #5092, to fully abstract the QueryBuilder; moving all sqlalchemy code to `SqlaQueryBuilder`, making the `BackendQueryBuilder` a proper abstract class, and reducing `QueryBuilder` backend interaction to these 6 calls:

```python
aiida/orm/querybuilder.py:
   131:         self._impl: BackendQueryBuilder = backend.query()
   952:         return self._impl.as_sql(data=self.as_dict(), inline=inline)
   962:         return self._impl.analyze_query(data=self.as_dict(), execute=execute, verbose=verbose)
   986:         result = self._impl.first(self.as_dict())
   999:         return self._impl.count(self.as_dict())
  1014:         for item in self._impl.iterall(self.as_dict(), batch_size):
  1034:         for item in self._impl.iterdict(self.as_dict(), batch_size):
```

As you can see, whenever the backend is called it is passed the (JSONable) query dict, which it uses to build the query. As was previously the case on the frontend, `SqlaQueryBuilder` hashes this dict, to ensure the query is only rebuilt on changes.

Some other changes of note:

- `aiida/orm/implementation/sqlalchemy/querybuilder.py` is split into a few modules, to improve modularity
- the `QueryBuilder.get_query` and `QueryBuilder.inject_query` method has been removed, since naturally these break abstraction, since they rely on an `sqlalchemy.Query` object (I don't know of any use-cases why these would be used)
- Made the `QueryBuilder.distinct` method save its state to the query dict, rather than directly calling the `sqlachemy.Query.distinct`, e.g.

```python
In [7]: qb = QueryBuilder()

In [8]: qb
Out[8]: QueryBuilder(path=[], filters={}, project={}, order_by=[], limit=None, offset=None, distinct=False)

In [9]: qb.distinct()
Out[9]: QueryBuilder(path=[], filters={}, project={}, order_by=[], limit=None, offset=None, distinct=True)

In [10]: qb.distinct(False)
Out[10]: QueryBuilder(path=[], filters={}, project={}, order_by=[], limit=None, offset=None, distinct=False)
```

- correctly reset `limit` after calling `QueryBuilder.one()`
- Improve validation of `joining_keyword`; before it would only validate against all possible keywords, now it validates against only the keywords for the entity type:

```python
In [2]: QueryBuilder().append(orm.Log, with_incoming='a')
ValueError: 'with_incoming' is not a valid keyword for 'log' joining specification
Valid keywords are: {'with_node'}
```

- Improve the defaults for joinin_keywords, i.e. rather than always setting `with_incoming` for all entities, set `with_node` for non-node entities:

```python
In [5]: QueryBuilder().append(orm.Node).append(orm.Group)
Out[5]: QueryBuilder(path=[{'entity_type': '', 'orm_base': 'node', 'tag': 'node_1', 'joining_keyword': None, 'joining_value': None, 'edge_tag': None, 'outerjoin': False}, {'entity_type': 'group.core', 'orm_base': 'group', 'tag': 'core_1', 'joining_keyword': 'with_node', 'joining_value': 'node_1', 'edge_tag': 'node_1--core_1', 'outerjoin': False}], filters={'node_1': {'node_type': {'like': '%'}}, 'core_1': {'type_string': {'like': '%'}}, 'node_1--core_1': {}}, project={'node_1': [], 'core_1': [], 'node_1--core_1': []}, order_by=[], limit=None, offset=None, distinct=False)
```

- `sqlalchemy_utils` is now only used in `tests/backends/aiida_sqlalchemy/test_utils.py`, so I have moved it to the `tests` extras, rather than `install_requires`
- lots of typing!


chrisjsewell:
This is good to go FYI

aiida/orm/querybuilder.py
```diff
@@ -409,168 +124,90 @@ def __init__(
...
-        })
+            'distinct': self._distinct,
+        }
+        if copy:
+            return deepcopy(data)
+        return data
 
     @property
-    def queryhelp(self) -> QueryDict:
+    def queryhelp(self) -> 'QueryDictType':
```
sphuber, line 199: Shouldn't we just deprecate this so we can remove it at some point? No reason to keep this around if there is a preferred alternative

sphuber (approved):
Thanks @chrisjsewell . I have given the changes a global look, but have to admit that due to the amount of changes have not been able to review everything in detail, so I am relying mostly on the tests here that seem to run without a problem. I left one comment about actually deprecating `queryhelp` but since that was not actually changed in this PR, ok to merge this as is. Would be good to include the PR message in the commit, since that contains a useful summary of the changes

aiida/orm/querybuilder.py
chrisjsewell: Yeh I mentioned this here: https://github.com/aiidateam/aiida-core/pull/5081#discussion_r689362848, lets open an issue for it to track

chrisjsewell: #5096

chrisjsewell:
> I have given the changes a global look, but have to admit that due to the amount of changes have not been able to review everything in detail, so I am relying mostly on the tests here that seem to run without a problem.

Thanks 😄 yeh I promise I was very thorough! and with the rigorous use of type checking, it really does help to make sure all the function/method inputs/outputs are correct

aiida/orm/querybuilder.py
sphuber: Ah yeah, I remember seeing this now. I am a bit surprised by you being against having deprecation warnings for a new major release? Why would that be objectionable? And why wouldn't that be a problem on a minor release?

Issue #4985, opened 2021-06-11
Drop Django Backend


chrisjsewell:
Ok I am going to make a very unnuanced statement to gauge opinion, and would appreciate if you simply respond with: "strongly agree", "weakly agree", "no opinion", "weakly disagree" or "strong disagree"; no discussion, no reasoning.
If the majority of responses are "strongly disagree", we will close this issue and never speak of it again, if not we can discuss further:

There were legacy reasons for maintaining both django and sqlalchemy (e.g. JSONB), but the reasons have all now gone, and so the only thing it serves to do is effectively doubles the development, testing and maintenance costs in many key areas of aiida-core. We can't drop sqlalchemy, since it is the only backend for which the QueryBuilder is/can be implemented (i.e. even when you select django as the backend, you are still calling sqlalchemy for queries), so it has to be django. It has to be now or never, whilst we are moving from v1 to v2, and so breaking changes are allowed. It would be relatively straight-foward to do, and would have minimal impact for most developers/users.

cc @giovannipizzi @sphuber @ltalirz @ramirezfranciscof @CasperWA @mbercx @csadorf

mbercx:
No discussion, no reasoning? I like this game 😁 

"strongly agree"

csadorf:
strongly agree

ltalirz:
weakly agree.

To be clear, I'm assuming the question to be: do you support dropping the django backend before the v2 release if it can be done without much effort and with minimal impact for most users?

"Minimal impact for most users" to me means an automatic migration from django to sqla (that would need to be  demonstrated on a large django database).

chrisjsewell:
To go slightly against my rule of no discussion 😬 (but a point I want to highlight):

> an automatic migration from django to sqla 

To a first approximation there is literally no difference with the actual postgres DB, and no migration needed (except maybe the schema_version), i.e. I tried

1. Create a new profile (with the default "django" backend)
2. `verdi archive import "https://archive.materialscloud.org/record/file?filename=two_dimensional_database.aiida&file_id=98f1789b-d68e-46db-9c67-4ef8dd8c1ebf&record_id=648"`
3. `verdi database summary -v`
4. change the backend to "sqlalchemy" in the config.json
5. I had to override `SettingsManager.validate_schema_version` (because it returned the DB schema as `None`):
   ```python
   def validate_schema_version(self, profile):
      return True
   ```
   but after that...
6. `verdi database summary -v` (works fine)
[...]

PR #3582, opened 2019-11-27
Refactor the code dealing with database into `BackendManager`

sphuber:
Fixes #3530 and fixes #3584 

The code that acts directly with the database, outside of the ORM, for
example to check the database schema generation and version, or to get
data from the `DbSettings` table has been refactored. The ORM requires
the database environment to be loaded, but to do that the database
schema needs to be verified which also requires accessing the database.
This functionality was implemented mostly in free-standing functions and
separately for both backends. This is reorganized into a `BackendManager`
abstract class and implemented for Django and SqlAlchemy. This
centralizes the functionality required to verify database schema and
perform the migrations thereof.

PR #5097, opened 2021-08-25
👌 IMPROVE: Add SQLA migration for parity with Django schema

### LINKED ISSUE #2303: Schema optimisations and consistency of indexes between Django and SqlAlchemy
There are (still) important differences between the schema generated by SQLA and Django.
To make sure they are the same, we SHOULD NOT check only the python code, but actually the information in the DB schema, asking psql.
For instance, while Django correctly adds two indices for type:
```
=> select * from pg_indexes where tablename='db_dbnode' and indexdef LIKE '%type%';
 schemaname | tablename |              indexname               | tablespace |                                                   indexdef                                                   
------------+-----------+--------------------------------------+------------+--------------------------------------------------------------------------------------------------------------
 public     | db_dbnode | db_dbnode_599dcce2                   |            | CREATE INDEX db_dbnode_599dcce2 ON public.db_dbnode USING btree (type)
 public     | db_dbnode | db_dbnode_type_4cda33f938ccd765_like |            | CREATE INDEX db_dbnode_type_4cda33f938ccd765_like ON public.db_dbnode USING btree (type varchar_pattern_ops)
(2 rows)
```
SQLA does not generate the second index (`varchar_pattern_ops`) that is essential for making queries for "startwith" in a string (`LIKE 'xxx%'`) that are used a lot in queries on the type. See also [Postgres docs](https://www.postgresql.org/docs/9.5/indexes-opclass.html).

Tasks:
- [ ] Make sure all relevant char fields have also a `varchar_pattern_ops` 
- [ ] Make a test that checks the schema autogenerated by Django as SQLA and compares it with some reference data to check if indexes and similar are the same (same uniqueness, same indexes and index types, ...)

chrisjsewell:
This PR adds migrations to the SQLAlchemy schema, to make it identical to the Django schema, and adds tests to ensure this parity.

There are three migrations:

1. Ensures all columns that are to be set with the `NOT NULL` restriction, are indeed non-null
2. Sets relevant columns to `nullable=False`
3. Changes in a few column types:
    - `db_dblog.levelname`: `String(255) -> String(50)`
    - `db_dbsetting.key`: `String(255) -> String(1024)`
    - `db_dbsetting.description`: `String(255) -> Text()`
4. Adds/renames all indexes and constraints to be identical to those in the django schema

`tests/backends/test_schema_parity.py` then tests the parity of the schema columns (name, type, nullable), primary key constraints, unique constraints and indexes.

This also required a regression fix for #3582, which had actually broken the sqlalchemy migration mechanism: instead of carrying out each migration in a separate transaction, this had caused all migrations to be carried out in the same transaction. This caused `cannot ALTER TABLE "db_xxx" because it has pending trigger events` errors, where changes in one migration were not "flushed" to the database on a transaction close.

In renaming indexes/constraints the sqlalchemy models now adopt the auto-generated names from django (see https://github.com/django/django/blob/2f33217ea2cad688040dd6044cdda946c62e5b65/django/db/backends/base/schema.py#L989).

In order to control the naming, all indexes and constraints are now specified specifically in the sqlalchemy models (in `__table_args__`), as opposed to being auto-generated via the `index` and `unique` options on the relevant `Column`.

It is also of note, that due to not taking into account https://code.djangoproject.com/ticket/23577, some of the django migrations were not properly written to rename index/constraint names, and so some names do not actually mirror their column names, for example `db_dbgroup_name_...` instead of `db_dbgroup_label_...`.

Finally, the database default for `repository_metadata` was removed, since this is not possible to set in django (see https://code.djangoproject.com/ticket/470), and the `DbNode.reposotry_metadata` properties have been modified, to ensure they return an empty dict, rather than `None`.

closes #2303, and a step towards addressing #4985


chrisjsewell:
Ok this is coming along now.

One thing to note is that django auto-generates names for indexes that have these partial-uuids as part of the name, e.g. `db_dbcomment_user_id_8ed5e360`, with a few also having "wrong" names, e.g. `db_dbgroup_name_66c75272` is an index for the `label` column
(see https://github.com/aiidateam/aiida-core/issues/4985#issuecomment-905205112)
I don't think in django there is any way to directly specify the index names (as sqlalchemy allows)

For parity, we want these names to be exactly the same so, assuming there is no possible django migration to change the names, I guess the migration here should rename all indexes.

Note, alembic does not allow for an actual rename operation: https://github.com/sqlalchemy/alembic/issues/246, so we would need to drop the indexes then recreate them

chrisjsewell:
Analysis of previous unique constraints:

```console
$ verdi -p a-import-django devel run-sql "SELECT tbl.relname,c.conname,c.contype,c.conkey FROM pg_constraint AS c INNER JOIN pg_class AS tbl ON tbl.oid = c.conrelid WHERE c.contype='u'  ORDER BY tbl.relname,c.conname;"
('db_dbauthinfo', 'db_dbauthinfo_aiidauser_id_dbcomputer_id_777cdaa8_uniq', 'u', [5, 6])
('db_dbcomment', 'db_dbcomment_uuid_49bac08c_uniq', 'u', [2])
('db_dbcomputer', 'db_dbcomputer_name_key', 'u', [3])
('db_dbcomputer', 'db_dbcomputer_uuid_f35defa6_uniq', 'u', [2])
('db_dbgroup', 'db_dbgroup_name_type_12656f33_uniq', 'u', [3, 4])
('db_dbgroup', 'db_dbgroup_uuid_af896177_uniq', 'u', [2])
('db_dbgroup_dbnodes', 'db_dbgroup_dbnodes_dbgroup_id_dbnode_id_eee23cce_uniq', 'u', [2, 3])
('db_dblog', 'db_dblog_uuid_9cf77df3_uniq', 'u', [10])
('db_dbnode', 'db_dbnode_uuid_62e0bf98_uniq', 'u', [2])
('db_dbsetting', 'db_dbsetting_key_1b84beb4_uniq', 'u', [2])
('db_dbuser', 'db_dbuser_email_30150b7e_uniq', 'u', [5])

$ verdi -p a-import-sqla devel run-sql "SELECT tbl.relname,c.conname,c.contype,c.conkey FROM pg_constraint AS c INNER JOIN pg_class AS tbl ON tbl.oid = c.conrelid WHERE c.contype='u'  ORDER BY tbl.relname,c.conname;"
('db_dbauthinfo', 'db_dbauthinfo_aiidauser_id_dbcomputer_id_key', 'u', [2, 3])
('db_dbcomment', 'db_dbcomment_uuid_key', 'u', [2])
('db_dbcomputer', 'db_dbcomputer_label_key', 'u', [3])
('db_dbcomputer', 'db_dbcomputer_uuid_key', 'u', [2])
('db_dbgroup', 'db_dbgroup_label_type_string_key', 'u', [3, 4])
('db_dbgroup', 'db_dbgroup_uuid_key', 'u', [2])
('db_dbgroup_dbnodes', 'db_dbgroup_dbnodes_dbgroup_id_dbnode_id_key', 'u', [3, 2])
('db_dblog', 'db_dblog_uuid_key', 'u', [10])
('db_dbnode', 'db_dbnode_uuid_key', 'u', [2])
('db_dbsetting', 'db_dbsetting_key_key', 'u', [2])
```

chrisjsewell:
Note it is non possible to set database level defaults in django (https://code.djangoproject.com/ticket/470), so I am removing the `server_default` on `repository_metadata` in sqlalchemy.

chrisjsewell:
another issue with django, it didn't rename the index & constraint when applying `aiida/backends/djsite/db/migrations/0048_computer_name_to_label.py` (see https://code.djangoproject.com/ticket/23577)

chrisjsewell:
For prosperity, its of note that you can affect the sqlalchemy auto-naming of indexes/constraints: https://docs.sqlalchemy.org/en/14/core/constraints.html#creating-custom-tokens-for-naming-conventions, but I think anyway it better to have them specifically in `__table_args__`

aiida/backends/sqlalchemy/migrations/versions/1feaea71bd5a_migrate_repository.py
```diff
@@ -111,8 +111,8 @@ def upgrade():
     container_id = backend.get_repository().uuid
     statement = text(
         f"""
-        INSERT INTO db_dbsetting (key, val, description)
-        VALUES ('repository|uuid', to_json('{container_id}'::text), 'Repository UUID')
+        INSERT INTO db_dbsetting (key, val, description, time)
+        VALUES ('repository|uuid', to_json('{container_id}'::text), 'Repository UUID', NOW())
```
sphuber, line 115: Will this leave existing databases, who have already applied this migration, in a broken state? They will not have this column set. I think this might require a separate migration that sets an arbitrary value for the `time` column on existing rows.

aiida/backends/sqlalchemy/models/comment.py
```diff
@@ -28,18 +29,30 @@ class DbComment(Base):
...
+        Integer, ForeignKey('db_dbuser.id', ondelete='CASCADE', deferrable=True, initially='DEFERRED'), nullable=False
+    )
+    content = Column(Text, default='', nullable=False)
 
     dbnode = relationship('DbNode', backref='dbcomments')
     user = relationship('DbUser')
 
+    __table_args__ = (
+        # index/constraint names mirror django's auto-generated ones
+        UniqueConstraint(uuid, name='db_dbcomment_uuid_49bac08c_uniq'),
```
sphuber, line 50: If we are going to keep the SqlAlchemy implementation, why are we migrating the SqlAlchemy databases to match the Django one? Wouldn't it be way more logical to migrate Django to match SqlAlchemy?

aiida/backends/sqlalchemy/models/node.py
```diff
@@ -43,18 +44,16 @@ class DbNode(Base):
...
+    node_type = Column(String(255), default='', nullable=False)
+    process_type = Column(String(255))
+    label = Column(String(255), nullable=False, default='')
+    description = Column(Text(), nullable=False, default='')
+    ctime = Column(DateTime(timezone=True), default=timezone.now, nullable=False)
+    mtime = Column(DateTime(timezone=True), default=timezone.now, onupdate=timezone.now, nullable=False)
     attributes = Column(JSONB)
     extras = Column(JSONB)
-    repository_metadata = Column(JSONB, nullable=False, default=dict, server_default='{}')
+    repository_metadata = Column(JSONB, nullable=True, default=dict)
```
sphuber, line 56: I remember adding this explicitly. The commit message (4a347b6001fc4f4dcc1c0710489a899b6589ee57) explains why. I think it makes sense to enforce that this column always has an empty dictionary. It prevented a bug. Why can we now revert this?

sphuber (commented):
Thanks @chrisjsewell . Before we go into detail, I have added a few high-level questions/comments.

aiida/backends/sqlalchemy/models/comment.py
chrisjsewell: Because you cannot control the name of indexes/constraints on the django ORM; they are auto-generated in private methods

aiida/backends/sqlalchemy/models/node.py
chrisjsewell: Because there is not a matching server default in the django schema (it is not possible: https://code.djangoproject.com/ticket/470). The `server_default` did not fix the bug in 4a347b6, it was the addition of `default=dict` on the python side.

aiida/backends/sqlalchemy/models/comment.py
chrisjsewell: Note, you could change the names of indexes/constraints in django migrations, but then the schema you create via the migration route would be different to the one you create via the models.
Although this parity is not actually tested on the django side, as opposed to on the sqlalchemy side it is specifically tested in `tests/backends/aiida_sqlalchemy/migrations/test_schemas.py::TestMigrationSchemaVsModelsSchema::test_model_and_migration_schemas_are_the_same`

chrisjsewell: (As noted in https://github.com/aiidateam/aiida-core/pull/5097#issuecomment-993200033, this parity is actually already violated in a number of django migrations, so perhaps there should be an additional django migration to fix this 🤷)

chrisjsewell:
>  Before we go into detail, I have added a few high-level questions/comments.

Cheers, answered

aiida/backends/sqlalchemy/models/node.py
chrisjsewell: Indeed though, ideally we would make more use of server-side defaults, which will be possible once we don't have to deal with the shortcomings of django 😉

aiida/backends/sqlalchemy/migrations/versions/1feaea71bd5a_migrate_repository.py
sphuber: True, so this change is fine.

aiida/backends/sqlalchemy/models/comment.py
sphuber: Ok, regarding the last comment, I think you have now fixed this, correct? That's the change in the `0048_computer_name_to_label.py ` migration?

Once we have dropped Django, would it make sense to revert some of these changes and no longer define the name explicitly? It really feels counterintuitive to have manually define an identifier with a random string.

aiida/backends/sqlalchemy/models/comment.py
```diff
@@ -28,18 +29,30 @@ class DbComment(Base):
...
+        # Index('db_dbcomment_uuid_49bac08c_uniq', uuid, unique=True),
```
sphuber, line 51:
```suggestion
```

aiida/backends/sqlalchemy/models/computer.py
```diff
@@ -31,13 +32,25 @@ class DbComputer(Base):
...
     id = Column(Integer, primary_key=True)  # pylint: disable=invalid-name
-    uuid = Column(UUID(as_uuid=True), default=get_new_uuid, unique=True)
-    label = Column(String(255), unique=True, nullable=False)
-    hostname = Column(String(255))
-    description = Column(Text, nullable=True)
-    scheduler_type = Column(String(255))
-    transport_type = Column(String(255))
-    _metadata = Column('metadata', JSONB)
+    uuid = Column(UUID(as_uuid=True), default=get_new_uuid, nullable=False)
+    label = Column(String(255), nullable=False)
```
sphuber, line 36: does this not need a default?

aiida/backends/sqlalchemy/models/group.py
```diff
@@ -50,25 +53,45 @@ class DbGroup(Base):
 
     id = Column(Integer, primary_key=True)  # pylint: disable=invalid-name
 
-    uuid = Column(UUID(as_uuid=True), default=get_new_uuid, unique=True)
-    label = Column(String(255), index=True)
+    uuid = Column(UUID(as_uuid=True), default=get_new_uuid, nullable=False)
+    label = Column(String(255), nullable=False)
```
sphuber, line 57: default?

aiida/backends/sqlalchemy/models/log.py
```diff
@@ -26,18 +27,38 @@ class DbLog(Base):
...
 
     id = Column(Integer, primary_key=True)  # pylint: disable=invalid-name
-    uuid = Column(UUID(as_uuid=True), default=get_new_uuid, unique=True)
-    time = Column(DateTime(timezone=True), default=timezone.now)
-    loggername = Column(String(255), index=True, doc='What process recorded the message')
-    levelname = Column(String(255), index=True, doc='How critical the message is')
+    uuid = Column(UUID(as_uuid=True), default=get_new_uuid, nullable=False)
+    time = Column(DateTime(timezone=True), default=timezone.now, nullable=False)
+    loggername = Column(String(255), nullable=False, doc='What process recorded the message')
+    levelname = Column(String(50), nullable=False, doc='How critical the message is')
```
sphuber, line 33: defaults?

aiida/backends/sqlalchemy/models/node.py
```diff
@@ -43,18 +44,16 @@ class DbNode(Base):
...
-    process_type = Column(String(255), index=True)
-    label = Column(
-        String(255), index=True, nullable=True, default=''
-    )  # Does it make sense to be nullable and have a default?
-    description = Column(Text(), nullable=True, default='')
-    ctime = Column(DateTime(timezone=True), default=timezone.now)
-    mtime = Column(DateTime(timezone=True), default=timezone.now, onupdate=timezone.now)
+    uuid = Column(UUID(as_uuid=True), default=get_new_uuid, nullable=False)
+    node_type = Column(String(255), default='', nullable=False)
+    process_type = Column(String(255))
```
sphuber, line 49: default?

tests/backends/test_schema_parity/test_columns.yml
```diff
@@ -0,0 +1,300 @@
...
+    is_nullable: false
+    max_length: 255
+  uuid:
+    data_type: uuid
+    default: null
+    is_nullable: false
+db_dbgroup:
+  description:
+    data_type: text
+    default: null
```
sphuber, line 95: Are these values correct? This says that `DbGroup.description`'s default is `null` but in the models it actually has `default=''`. I was wondering what the hell was going on with the defaults where for many columns you are now defining `default = ''` but not all of them. I had already added quite a few review comments at columns that don't specify a default, asking whether they should. What was the general thinking as to which fields need a default and which don't?

In any case, if the tests pass with this reference file, that seems a problem no? All defaults seem to be null here even though most models actually define one.

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
```diff
@@ -0,0 +1,173 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():  # pylint: disable=too-many-statements
+    """Convert null values to default values"""
+    db_dbauthinfo = sa.sql.table(
+        'db_dbauthinfo',
+        # sa.Column('aiidauser_id', sa.Integer),
+        # sa.Column('dbcomputer_id', sa.Integer),
```
sphuber, line 38:
```suggestion
```

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
```diff
@@ -0,0 +1,173 @@
...
+    op.execute(db_dbauthinfo.update().where(db_dbauthinfo.c.enabled.is_(None)).values(enabled=True))
+    op.execute(db_dbauthinfo.update().where(db_dbauthinfo.c.auth_params.is_(None)).values(auth_params={}))
+    op.execute(db_dbauthinfo.update().where(db_dbauthinfo.c.metadata.is_(None)).values(metadata={}))
+    op.execute(db_dbauthinfo.update().where(db_dbauthinfo.c.auth_params == JSONB.NULL).values(auth_params={}))
+    op.execute(db_dbauthinfo.update().where(db_dbauthinfo.c.metadata == JSONB.NULL).values(metadata={}))
+
+    db_dbcomment = sa.sql.table(
+        'db_dbcomment',
+        # sa.Column('dbnode_id', sa.Integer),
+        # sa.Column('user_id', sa.Integer),
```
sphuber, line 53:
```suggestion
```

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
```diff
@@ -0,0 +1,173 @@
...
+        sa.Column('content', sa.Text),
+        sa.Column('ctime', sa.DateTime(timezone=True)),
+        sa.Column('mtime', sa.DateTime(timezone=True)),
+        sa.Column('uuid', UUID(as_uuid=True)),
+    )
+
+    op.execute(db_dbcomment.update().where(db_dbcomment.c.content.is_(None)).values(content=''))
+    op.execute(db_dbcomment.update().where(db_dbcomment.c.mtime.is_(None)).values(mtime=timezone.now()))
+    op.execute(db_dbcomment.update().where(db_dbcomment.c.ctime.is_(None)).values(ctime=timezone.now()))
```
sphuber, line 62: This would of course really set the wrong time, but I guess that is as opposed to not having a time at all. Also don't see another solution, but it does raise the question: are there really databases out there that have nulls for these fields? Seems like that would have been a source of bugs. Since `Comment` is not used a lot, it probably never surfaced, but for the other tables maybe

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
```diff
@@ -0,0 +1,173 @@
...
+    op.execute(db_dbcomputer.update().where(db_dbcomputer.c.hostname.is_(None)).values(hostname=''))
+    op.execute(db_dbcomputer.update().where(db_dbcomputer.c.metadata.is_(None)).values(metadata={}))
+    op.execute(db_dbcomputer.update().where(db_dbcomputer.c.metadata == JSONB.NULL).values(metadata={}))
+    op.execute(db_dbcomputer.update().where(db_dbcomputer.c.scheduler_type.is_(None)).values(scheduler_type=''))
+    op.execute(db_dbcomputer.update().where(db_dbcomputer.c.transport_type.is_(None)).values(transport_type=''))
+    op.execute(db_dbcomputer.update().where(db_dbcomputer.c.uuid.is_(None)).values(uuid=get_new_uuid()))
+
+    db_dbgroup = sa.sql.table(
+        'db_dbgroup',
+        # sa.Column('user_id', sa.Integer),
```
sphuber, line 85:
```suggestion
```

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
```diff
@@ -0,0 +1,173 @@
...
+    op.execute(db_dbgroup.update().where(db_dbgroup.c.time.is_(None)).values(time=timezone.now()))
+    op.execute(db_dbgroup.update().where(db_dbgroup.c.type_string.is_(None)).values(type_string=''))
+    op.execute(db_dbgroup.update().where(db_dbgroup.c.uuid.is_(None)).values(uuid=get_new_uuid()))
+
+    # db_dbgroup_dbnode = sa.sql.table(
+    #     "db_dbgroup_dbnode",
+    #     # sa.Column('dbgroup_id', sa.Integer),
+    #     # sa.Column('dbnode_id', sa.Integer),
+    # )
+
```
sphuber, line 104:
```suggestion
```

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
```diff
@@ -0,0 +1,173 @@
...
+    db_dblink = sa.sql.table(
+        'db_dblink',
+        # sa.Column('input_id', sa.Integer),
+        # sa.Column('output_id', sa.Integer),
```
sphuber, line 108:
```suggestion
```

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
```diff
@@ -0,0 +1,173 @@
...
+        sa.Column('type', sa.String(255)),
+    )
+
+    op.execute(db_dblink.update().where(db_dblink.c.type.is_(None)).values(type=''))
```
sphuber, line 112: Already commented on the default setting for the `Comment.ctime`, but this thing seems way more serious. I wonder if we should collect all default settings that would be problematic if there are really rows out there that have null values. We should then have a "integrity-check" that finds these. I fear that if we just silently set these missing values to a default, the database will be broken at other points. Having a `Link` without a type is bad...

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
```diff
@@ -0,0 +1,173 @@
...
+        sa.Column('mtime', sa.DateTime(timezone=True)),
+        sa.Column('node_type', sa.String(255)),
+        sa.Column('uuid', UUID(as_uuid=True)),
+    )
+
+    op.execute(db_dbnode.update().where(db_dbnode.c.ctime.is_(None)).values(ctime=timezone.now()))
+    op.execute(db_dbnode.update().where(db_dbnode.c.description.is_(None)).values(description=''))
+    op.execute(db_dbnode.update().where(db_dbnode.c.label.is_(None)).values(label=''))
+    op.execute(db_dbnode.update().where(db_dbnode.c.mtime.is_(None)).values(mtime=timezone.now()))
+    op.execute(db_dbnode.update().where(db_dbnode.c.node_type.is_(None)).values(node_type=''))
```
sphuber, line 147: another problematic one

aiida/backends/sqlalchemy/migrations/versions/1de112340b18_django_parity_3.py
```diff
@@ -0,0 +1,222 @@
...
+    ('db_dbnode', ('ctime',), False, 'db_dbnode_ctime_71626ef5'),
+    ('db_dbnode', ('mtime',), False, 'db_dbnode_mtime_0554ea3d'),
+    ('db_dbnode', ('dbcomputer_id',), False, 'db_dbnode_dbcomputer_id_315372a3'),
+    ('db_dbnode', ('user_id',), False, 'db_dbnode_user_id_12e7aeaf'),
+)
+
+# table name, column name, index name
+MISSING_VARCHAR_INDEXES = (
+    ('db_dbcomputer', 'label', 'db_dbcomputer_label_bc480bab_like'),
+    ('db_dbgroup', 'label', 'db_dbgroup_name_66c75272_like'),
```
sphuber, line 44: Should this not be label after the Django migration 48 is performed correcting the migration name?
```suggestion
    ('db_dbgroup', 'label', 'db_dbgroup_label_66c75272_like'),
```

Why was this not caught by the tests?

aiida/backends/sqlalchemy/migrations/versions/1de112340b18_django_parity_3.py
```diff
@@ -0,0 +1,222 @@
...
+    ('db_dbsetting', ('key',), 'db_dbsetting_key_key'),
+)
+
+# table name, column names, constraint name
+ADD_UNIQUE_CONSTRAINTS = (
+    ('db_dbauthinfo', ('aiidauser_id', 'dbcomputer_id'), 'db_dbauthinfo_aiidauser_id_dbcomputer_id_777cdaa8_uniq'),
+    ('db_dbcomment', ('uuid',), 'db_dbcomment_uuid_49bac08c_uniq'),
+    ('db_dbcomputer', ('label',), 'db_dbcomputer_label_bc480bab_uniq'),
+    ('db_dbcomputer', ('uuid',), 'db_dbcomputer_uuid_f35defa6_uniq'),
+    ('db_dbgroup', ('label', 'type_string'), 'db_dbgroup_name_type_12656f33_uniq'),
```
sphuber, line 77: Maybe this name change is also missing in Django 48 migration?
```suggestion
    ('db_dbgroup', ('label', 'type_string'), 'db_dbgroup_label_type_12656f33_uniq'),
```

sphuber (changes requested):
Have given the code a look now. Will now to try to run the migration locally and test things a bit.

aiida/backends/sqlalchemy/models/node.py
sphuber: Ok not ideal that we have to adapt to Django if that is what we are going to drop, but if that's the way it is. Maybe we can optimize this once Django is dropped and add a migration for SqlAlchemy to do things as they should be ideally

aiida/backends/sqlalchemy/migrations/versions/1de112340b18_django_parity_3.py
chrisjsewell: No this is from `aiida/backends/djsite/db/migrations/0021_dbgroup_name_to_label_type_to_type_string.py` (and there is also `aiida/backends/djsite/db/migrations/0030_dbnode_type_to_dbnode_node_type.py`), which obviously cannot be retroactively fixed.
I tempted then to add a django migration for these.

aiida/backends/sqlalchemy/migrations/versions/1de112340b18_django_parity_3.py
chrisjsewell: See above

aiida/backends/sqlalchemy/migrations/versions/1de112340b18_django_parity_3.py
sphuber: Hang on. If its correct, maybe just leave it. Especially if we think that after Django is fully dropped, we might add a final SQLA migration to undo the things we had to do specifically for django, such as weirdly named indices and no server defaults. If we are planning on doing this, no need to change it for django now

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
chrisjsewell: Well I imagine having them set to `None` would already be bad, and so going from `None -> ''` is not going to break the database any more than it was already.
I think in reality the python code should have already stopped this from occurring, but yeh having an integrity-check could be done

tests/backends/test_schema_parity/test_columns.yml
chrisjsewell: Note, everything in these regression files are server side data; `null` here simply means that no default has been set. By contrast, `default=''` on the model is a client side default

aiida/backends/sqlalchemy/migrations/versions/91b573400be5_prepare_schema_reset.py
```diff
@@ -40,8 +40,8 @@ def upgrade():
     # isolated way, we do not suffer from those problems and can safely use it.
     statement = text(
         """
-        INSERT INTO db_dbsetting (key, val, description)
-        SELECT 'schema_generation', '"1"', 'Database schema generation'
+        INSERT INTO db_dbsetting (key, val, description, time)
+        SELECT 'schema_generation', '"1"', 'Database schema generation', NOW()
```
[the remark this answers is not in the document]
chrisjsewell: Well it already is done, in `aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py`, that sets `null` to the current time.
Off-hand I can't remember why I also added it here as well, I guess it was breaking a test but will check

aiida/backends/sqlalchemy/models/computer.py
chrisjsewell: There is no default on the django side. I imagine we should not be "silently" setting these labels to a default, because not specifying a label is likely to be a bug of the upstream code

aiida/backends/sqlalchemy/models/group.py
chrisjsewell: see above

aiida/backends/sqlalchemy/models/log.py
chrisjsewell: see above

aiida/backends/sqlalchemy/models/node.py
chrisjsewell: the default here, since `process_type` is nullable, will automatically be `None`

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
chrisjsewell: >  are there really databases out there that have nulls for these fields? 

I very much doubt it, this is really just an extra pre-caution, to ensure these migrations can actually be run

aiida/backends/sqlalchemy/models/comment.py
chrisjsewell: > It really feels counterintuitive to have manually define an identifier with a random string.

I would note here that these strings are not random, they are deterministically generated from the table name, column name(s) and suffix (see https://github.com/django/django/blob/43289707809c814a70f0db38ca4f82f35f43dbfd/django/db/backends/base/schema.py#L989)

If there was a django migration, to fix naming from `aiida/backends/djsite/db/migrations/0021_dbgroup_name_to_label_type_to_type_string.py` and `aiida/backends/djsite/db/migrations/0030_dbnode_type_to_dbnode_node_type.py`,
then it may be possible, with https://docs.sqlalchemy.org/en/14/core/constraints.html#creating-custom-tokens-for-naming-conventions, to actually setup up sqlalchemy to auto-generate exactly the same names.

Although, yeh don't know if it is worth the effort

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
chrisjsewell: I would just note here for completeness that, since all the foreign field columns were nullable, in theory there could exist a database where these are null and then the change to `nullable=False` would fail.
But obviously (a) in practice I can't see this ever being the case, since it would already break aiida, and (b) there is no possible default to change these to

aiida/backends/sqlalchemy/migrations/versions/91b573400be5_prepare_schema_reset.py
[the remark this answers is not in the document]
sphuber: I am not sure if this is the cause, but when I run the migrations of your branch on an existing SQLA profile, the first one gets applied, but then it seems to get stuck at the second one. I have to kill the postgres process and then get the following stacktrace:
```
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1802, in _execute_context
    self.dialect.do_execute(
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 732, in do_execute
    cursor.execute(statement, parameters)
psycopg2.OperationalError: SSL SYSCALL error: EOF detected


  File "/home/sph/.virtualenvs/aiida_dev/bin/verdi", line 12, in <module>
    sys.exit(verdi())
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/click/core.py", line 1128, in __call__
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/click/core.py", line 1659, in invoke
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/click/core.py", line 1659, in invoke
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/click/core.py", line 1395, in invoke
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/cmdline/commands/cmd_storage.py", line 42, in storage_migrate
    backend.migrate()
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/orm/implementation/sqlalchemy/backend.py", line 44, in migrate
    self._schema_manager.migrate()
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/backends/manager.py", line 157, in migrate
    self._migrate_database_version()
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/backends/sqlalchemy/manager.py", line 160, in _migrate_database_version
    self.migrate_up('head')
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/backends/sqlalchemy/manager.py", line 147, in migrate_up
    upgrade(config, version)
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/command.py", line 320, in upgrade
    script.run_env()
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/script/base.py", line 563, in run_env
    util.load_python_file(self.dir, "env.py")
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/util/pyfiles.py", line 92, in load_python_file
    module = load_module_py(module_id, path)
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/util/pyfiles.py", line 108, in load_module_py
    spec.loader.exec_module(module)  # type: ignore
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/backends/sqlalchemy/migrations/env.py", line 52, in <module>
    run_migrations_online()
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/backends/sqlalchemy/migrations/env.py", line 45, in run_migrations_online
    context.run_migrations()  # pylint: disable=no-member
  File "<string>", line 8, in run_migrations
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/runtime/environment.py", line 851, in run_migrations
    self.get_context().run_migrations(**kw)
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/runtime/migration.py", line 612, in run_migrations
    step.migration_fn(**kw)
  File "/home/sph/code/aiida/env/dev/aiida-core/aiida/backends/sqlalchemy/migrations/versions/1de112340b17_django_parity_2.py", line 80, in upgrade
    op.alter_column('db_dbsetting', 'time', existing_type=sa.DateTime(timezone=True), nullable=False)
  File "<string>", line 8, in alter_column
  File "<string>", line 3, in alter_column
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/operations/ops.py", line 1880, in alter_column
    return operations.invoke(alt)
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/operations/base.py", line 387, in invoke
    return fn(self, operation)
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/operations/toimpl.py", line 50, in alter_column
    operations.impl.alter_column(
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/ddl/postgresql.py", line 173, in alter_column
    super(PostgresqlImpl, self).alter_column(
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/ddl/impl.py", line 231, in alter_column
    self._exec(
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/alembic/ddl/impl.py", line 197, in _exec
    return conn.execute(construct, multiparams)
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/future/engine.py", line 280, in execute
    return self._execute_20(
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1614, in _execute_20
    return meth(self, args_10style, kwargs_10style, execution_options)
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/sql/ddl.py", line 80, in _execute_on_connection
    return connection._execute_ddl(
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1381, in _execute_ddl
    ret = self._execute_context(
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1845, in _execute_context
    self._handle_dbapi_exception(
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 2026, in _handle_dbapi_exception
    util.raise_(
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/util/compat.py", line 207, in raise_
    raise exception
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1802, in _execute_context
    self.dialect.do_execute(
  File "/home/sph/.virtualenvs/aiida_dev/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 732, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) SSL SYSCALL error: EOF detected

[SQL: ALTER TABLE db_dbsetting ALTER COLUMN time SET NOT NULL]
(Background on this error at: https://sqlalche.me/e/14/e3q8)
```
The final error message seems interesting. It may be coincidence, but it is exactly concerning the time column of `db_dbsetting`.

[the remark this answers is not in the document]
sphuber: Just tried this on an empty database and it also hangs completely. Surely it cannot take more than 5 minutes on an empty database, so it must be a problem.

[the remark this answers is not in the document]
chrisjsewell: What if you undo this change, or comment everything out in the aiida/backends/sqlalchemy/migrations/versions/1de112340b17_django_parity_2.py?
As discussed I had issues with "locked" databases both here, before fixing the issue that separate migrations were not being run in separate transactions, and in #5192, before fixing a separate sqlalchemy session not being closed.

chrisjsewell:
O.M.G.
After an inordinate amount of debugging I found the issue and a tentative fix:

So after here: https://github.com/aiidateam/aiida-core/blob/c743a335480f8bb3a5e4ebd2463a31f9f3b9f9b2/aiida/backends/sqlalchemy/manager.py#L172
and here: https://github.com/aiidateam/aiida-core/blob/c743a335480f8bb3a5e4ebd2463a31f9f3b9f9b2/aiida/backends/sqlalchemy/manager.py#L187
(at least)

you end up within an "idle" transaction:

```console
$ ps x | grep postgres
...
12539   ??  Ss     0:00.01 postgres: aiida_qs_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9 test-sqla9_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9 ::1(51834) idle in transaction 
```

when alembic then tries to alter the table during the migration:

```
01/12/2022 09:29:17 AM <13504> alembic.runtime.migration: [INFO] Running upgrade 1de112340b16 -> 1de112340b17, Parity with Django backend (rev: 0049),
part 2: Alter columns to be non-nullable and change type of some columns.
```

it hangs waiting for this idle transaction to close

```console
 7434   ??  Ss     0:00.01 postgres: aiida_qs_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9 test-sqla9_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9 ::1(50629) idle in transaction  
 7435   ??  Ss     0:00.01 postgres: aiida_qs_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9 test-sqla9_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9 ::1(50630) idle  
 7436   ??  Ss     0:00.04 postgres: aiida_qs_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9 test-sqla9_chrisjsewell_3abef2c96e126c712c1d3e256a3590b9 ::1(50631) ALTER TABLE waiting 
```

I don't know why this idle transaction opens, and cannot think how to stop it from opening.
But, in d522b44, if I add:

```python
        transaction = get_scoped_session().get_transaction()
        if transaction:
            transaction.close()
```

before starting the migration, this closes any current transaction, allowing for the migration to proceed properly

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
sphuber: But so are you suggesting to keeping these commented out lines to remember that these _may_ cause problems? If so, then I would add a comment explaining. Otherwise, I would just remove these comments out lines everywhere. Actually, I think the best solution is to summarize this problem once in the docstring of the migration module saying there are fields that could be null because of the missing explicit `nullable=False`, but that it is unlikely and anyway we couldn't provide a default. And then just delete the commented out lines

sphuber (changes requested):
Thanks @chrisjsewell  I think we can almost merge this. I closed all resolved comments, but there are still a couple outstanding

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
sphuber: Well maybe it would actually be better to have it fail in this case. Assuming that these cases are rare or not existing, we could analyze the problem and provide a proper fix. Tecnhically we would be able to deduce the correct type based on input and output nodes. I think this column is too delicate to be changing silently

aiida/backends/sqlalchemy/manager.py
```diff
@@ -41,7 +41,12 @@ def alembic_config():
 
         from . import ENGINE
 
-        with ENGINE.begin() as connection:
+        transaction = get_scoped_session().get_transaction()
+        if transaction:
+            transaction.close()
```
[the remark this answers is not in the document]
chrisjsewell: It is indeed, and I've added a comment 👍

chrisjsewell:
> Can you please still try to remove this change and so we only rely on the parity migration to set the proper default? Not ideal to be changing a migration that has already been released.

Ok trying reverting in [`be0c2bf` (#5097)](https://github.com/aiidateam/aiida-core/pull/5097/commits/be0c2bf698c4c5de3871448f344459302b5d5b5b)
Hopefully now, this isn't a problem, with the fix regarding idle transactions

chrisjsewell:
Ok I just tested a "manual" migration, and all seems to be running as expected now 🎉

aiida/backends/sqlalchemy/migrations/versions/91b573400be5_prepare_schema_reset.py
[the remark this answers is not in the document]
chrisjsewell: this has been removed

aiida/backends/sqlalchemy/migrations/versions/1de112340b16_django_parity_1.py
chrisjsewell: Ok, so if I just remove this line then.
Is there any other you think should be definitely removed?

I guess we can think to provide a "fix" under `verdi storage` for such issues,
but I imagine we can defer that to another PR?

sphuber: Yes, essentially there are these two comments remaining: (1) remove the setting of defaults that should really be considered as grave bugs (such as this one. Please also check if there are others), and (2) if you keep the commented out lines for operations that should have been there in principle but can't due to problems in original schema, add a generic comment in top docstring.

sphuber (changes requested):
Ok, just found a few more things, some minor, but some very important. I made them all as suggestions, so I think you can just all batch them in one go and we can merge. Think you will agree with all of them

sphuber (approved):
Thanks @chrisjsewell let's get this show on the road

chrisjsewell:
cheers for the review!

PR #5104, opened 2021-08-30
⬆️ UPGRADE: psycopg2 allow v2.9

### LINKED ISSUE #5002: Unpin psycopg2-binary from 2.8
As discussed in https://github.com/aiidateam/aiida-core/pull/4989#issue-672551353, database creation cannot run inside a transaction block.
This was probably the case before (as in it was possible, but not a recommended practice), but from psycopg2 2.9 this now raises an exception.

chrisjsewell:
Fixed in https://github.com/aiidateam/pgsu/pull/27

fixes #5002


chrisjsewell:
will need to wait for the conda release of pgsu v0.2.1

csadorf (approved):
Approved pending that all tests pass. Please let me know if there appears to be an issue with the test-install workflow.

chrisjsewell:
eurhhh; ran into another problem; https://github.com/psycopg/psycopg2/issues/1293, which is not being fixed for django v2: https://github.com/django/django/pull/14530

oh well, guess I'll have to fox the issue by dropping django 😉

chrisjsewell:
should just need to implement the try/except, to close the connection, in `tests/backends/aiida_sqlalchemy/migrations/conftest.py::uninitialised_profile`

sphuber:
> should just need to implement the try/except, to close the connection, in `tests/backends/aiida_sqlalchemy/migrations/conftest.py::uninitialised_profile`

Do you mean `try/finally`?

It is interesting that it is the migration tests that fail in the first PR after merging them, while I was working on creating a nightly workflow to move them to since we said "the migration tests shouldn't break very often" :sweat_smile:

chrisjsewell:
> It is interesting that it is the migration tests that fail in the first PR after merging them

Well, creating an "empty" profile (configured to an empty database/repository), for the duration of a test, is not really tied to migrations.
Its just a standard action that django did not allow, because we could only create a single, global database connection.

sphuber:
> For sure, the polish tests should be at least moved to a separate job though

Well if you think 2:30 is not a problem, then the polish also don't have to be moved. [Running the two of them takes ~4:20](https://github.com/sphuber/aiida-core/runs/5220141109?check_suite_focus=true). It is not a huge difference, but still I think it might make sense not to require those additional minutes on every commit.

sphuber (approved):
Good to go as far as I am concerned

PR #5106, opened 2021-08-30
try adding `consumer_timeout` default to `broker_parameters` config

### LINKED ISSUE #5105: AiiDA will no longer work with rabbitmq>3.7 by default
In https://github.com/rabbitmq/rabbitmq-server/pull/2990 a `consumer_timeout` has been introduced and set to 15 minutes, meaning that any process task that takes longer than 15 minutes will be cancelled 😬
(there is people in that PR none too happy that this was introduced in a minor version)

The quick fix for this for users is either (a) use rabbitmq 3.7 or lower, or (b) configure `consumer_timeout` to false. (see also https://www.rabbitmq.com/consumers.html#acknowledgement-timeout)

As is literally the last comment in that PR, at the time of writing, it is unclear to me off-hand if this can be done using the API (i.e. something aiida-core can handle automatically)?

chrisjsewell:
looking to address #5105; will it work for all rabbitmq versions?

### DIFF
--- a/aiida/manage/configuration/schema/config-v5.schema.json
+++ b/aiida/manage/configuration/schema/config-v5.schema.json
@@ -323,13 +323,20 @@
                     "description": "RabbitMQ arguments that will be encoded as query parameters",
                     "type": "object",
                     "default": {
-                        "heartbeat": 600
+                        "heartbeat": 600,
+                        "consumer_timeout": false
                     },
                     "properties": {
                         "heartbeat": {
                             "description": "After how many seconds the peer TCP connection should be considered unreachable",
                             "type": "integer",
                             "default": 600,
+                            "minimum": 0
+                        },
+                         "consumer_timeout": {
+                            "description": "A timeout for a daemon to complete a task",
+                            "type": ["boolean", "integer"],
+                            "default": false,
                             "minimum": 0
                         }
                     }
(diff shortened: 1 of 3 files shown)


aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -323,14 +323,21 @@
                     "description": "RabbitMQ arguments that will be encoded as query parameters",
                     "type": "object",
                     "default": {
-                        "heartbeat": 600
+                        "heartbeat": 600,
+                        "consumer_timeout": false
```
chrisjsewell, line 327:
```suggestion
                        "consumer_timeout": 1
```

lets try setting to really low, if that breaks rabbitmq 3.8

aiida/manage/configuration/schema/config-v5.schema.json
```diff
@@ -323,14 +323,21 @@
...
                         "heartbeat": {
                             "description": "After how many seconds the peer TCP connection should be considered unreachable",
                             "type": "integer",
                             "default": 600,
                             "minimum": 0
+                        },
+                         "consumer_timeout": {
+                            "description": "A timeout for a daemon to complete a task",
+                            "type": ["boolean", "integer"],
+                            "default": false,
```
chrisjsewell, line 339:
```suggestion
                            "default": 1,
```

PR #5108, opened 2021-08-30
CI: do not run the benchmark on forks

sphuber:
If done, it would fail anyway because it cannot find the `gh-pages`
branch to push the results to.

For example, see: https://github.com/sphuber/aiida-core/runs/3421686169

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -11,7 +11,8 @@ jobs:
 
   run-and-upload:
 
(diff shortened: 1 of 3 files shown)


chrisjsewell (approved):

sphuber:
@chrisjsewell I added a small commit on top of this PR to address the `importlib-metadata` problem. Opened #5109 to make sure we revert it once a fix is released

PR #5110, opened 2021-08-31
🔧 MAINTAIN: Remove `_DeprecateConfigCommandsGroup`

chrisjsewell:
Heya @sphuber, I guess you missed, when you removed the deprecations


sphuber (approved):
Just checked and this was added in v1.6, which was released in March 30 2021. Not the longest of time, but I guess it's fair game by now to remove it.

chrisjsewell:
Yep, it wouldn’t work now anyway though, because you removed the actual command that it called

PR #5111, opened 2021-09-01
Dependencies: update to `click==8.0` and remove `click-completion`

### LINKED ISSUE #5098: Drop the `click-completion` dependency in favor of the native implementation of `click==8.0`
As of `click v8.0`, tab-completion is now supported natively, as [described here in the docs](https://click.palletsprojects.com/en/8.0.x/shell-completion/?highlight=completion). So far we have been using the [`click-completion`](https://github.com/click-contrib/click-completion) extension which may now be no longer necessary.

There has been an [open issue](https://github.com/click-contrib/click-completion/issues/41) for a while asking whether `click-completion` can be considered deprecated or whether it still provides more functionality that isn't yet in `click`. From a first glance, it seems everything we need is there. The only downside is that the completion activation command is now shell dependent and Power shell is not natively supported, but I doubt that this should be a deal breaker.

sphuber:
Fixes #5098 

This has two other commits in it atm, that are supposed to be merged first in #5107 .


sphuber:
@chrisjsewell already assigned you before, but @ramirezfranciscof feel free to review this as well if you'd like

csadorf (commented):
Thanks for requesting my review. I am a bit confused by the size of this PR. Is it because the branch is not up-to-date?

sphuber:
> Thanks for requesting my review. I am a bit confused by the size of this PR. Is it because the branch is not up-to-date?

Not really. This is just because their were quite a bit of chances in the new click version that required significant changes of our code. I would refer you to the commit messages for more information. I have rebased the branch so it is now up to date.

tests/cmdline/commands/test_archive_import.py
```diff
@@ -225,7 +225,7 @@ def test_raise_malformed_url(self):
         self.assertIsNotNone(result.exception, result.output)
         self.assertNotEqual(result.exit_code, 0, result.output)
 
-        error_message = 'Is it a valid path or URL?'
+        error_message = 'could not be reached within'
```
csadorf, line 228: This was one of the changes I stumbled over. It appears like an unrelated change.

csadorf (commented):
All good from my side. I had only a cursory look over the code, so I'm not going to explicitly approve it, but thanks a lot for making sure that we can move forward with click 8.x. That was quite an effort.

PR #4938, opened 2021-05-08
♻️ REFACTOR: Proposal for `RepoPath` API

chrisjsewell:
This a proposal for an implementation of the file system path protocol (PEP 519), and mimics the ``pathlib.Path`` API.
This is exposed as `Node.repo_path`

In addition, the Repository is sub-classed into `MutableRepoFileSystem` (for use before storing) and `FrozenRepoFileSystem` (for use after storing)

```python
In [1]: node = Data()

In [2]: node.repo_path.joinpath('a', 'b', 'c').write_text("hallo")

In [3]: node.repo_content_string()
a: 
a/b: 
a/b/c: d0a02a6a-ba87-4db6-b4ef-833d89ecd2c1

In [4]: node.repo_path.joinpath('a', 'b', 'c')
Out[4]: RepoPath('a/b/c', MutableRepoFileSystem())

In [5]: node.put_object_from_tree('/Users/chrisjsewell/Documents/GitHub/aiida-core-origin/.molecule')

In [6]: node.repo_content_string()
README.md: 90f76af5-f271-4a7e-8483-31591750fd21
a: 
a/b: 
a/b/c: d0a02a6a-ba87-4db6-b4ef-833d89ecd2c1
default: 
default/Dockerfile: 0acbdc57-43dc-4f6f-97e7-d40e0bfd9ad7
default/config_jenkins.yml: d96a8823-2650-4150-aa1c-305e98fed76d
default/config_local.yml: 95e695a7-4044-477b-9d3c-e1d7c38021ab
default/create_docker.yml: 56285e18-c760-420d-b9ac-87acfab1bc9c
default/files: 
default/files/polish: 
default/files/polish/__init__.py: 966bbc4d-042b-4e89-8525-d823efd8bed0
default/files/polish/cli.py: 606f9226-199a-41f7-98e7-17eaa03c41be
default/files/polish/lib: 
default/files/polish/lib/__init__.py: 237d6def-eb89-48c1-b40d-df653a2920e2
default/files/polish/lib/expression.py: ca11e38a-ea14-4894-93d0-8306e6762ee9
default/files/polish/lib/template: 
default/files/polish/lib/template/base.tpl: 027eb8b3-c916-425a-988f-58f3ebb2d5b1
default/files/polish/lib/template/workchain.tpl: 296f0ecb-1b89-4405-adda-c74b81b18da9
default/files/polish/lib/workchain.py: 72e9a6ec-647c-4209-8365-ee46c2f45a25
[...]

PR #5114, opened 2021-09-02
`Node`: add the `copy_tree` method 

### LINKED ISSUE #4928: Add a method to `NodeRepositoryMixin` to recursively copy all content (or a subfolder) of a node's repository to another location
Currently, one has to walk manually over the contents using `Repository.walk`.

Note that the target location might be some path on the local file system, but it would also be great if possibly this can be directly copied into another target node. Example use case is where a subfolder of one `FolderData` needs to be copied to another `FolderData`. This is implementable by using `walk` and copying over all files manually, but a utility method that does this for the caller might be very useful.

sphuber:
Fixes #4928 

This method allows to easily copy the entire repository tree, or a
subdirectory, of a `Node` to a place on the local file system. The
method simply forwards to the `aiida.repository.Repository` class, where
the functionality is actually implemented using the `Repository.walk`
method, recursively looping over all the files contained within the
node's repository.

### DIFF
--- a/aiida/orm/nodes/repository.py
+++ b/aiida/orm/nodes/repository.py
@@ -216,6 +216,14 @@ class NodeRepositoryMixin:
         """
         yield from self._repository.walk(path)
 
+    def copy_tree(self, target: Union[str, pathlib.Path], path: FilePath = None) -> None:
+        """Copy the contents of the entire node repository to another location on the local file system.
+
+        :param target: absolute path of the directory where to copy the contents to.
+        :param path: optional relative path whose contents to copy.
+        """
+        self._repository.copy_tree(target, path)
+
     def delete_object(self, path: str):
         """Delete the object from the repository.
 
--- a/tests/repository/test_repository.py
+++ b/tests/repository/test_repository.py
@@ -4,6 +4,7 @@
 import contextlib
 import io
 import pathlib
+import typing as t
 
 import pytest
 
@@ -55,6 +56,20 @@ def repository_uninitialised(request, tmp_path_factory) -> Repository:
     with request.param(tmp_path_factory.mktemp('container')) as backend:
         repository = Repository(backend=backend)
         yield repository
+
+
+@pytest.fixture(scope='function', params=[True, False])
+def tmp_path_parametrized(request, tmp_path_factory) -> t.Union[str, pathlib.Path]:
+    """Indirect parametrized fixture that returns temporary path both as ``str`` and as ``pathlib.Path``.
+
+    This is a useful fixture to automatically parametrize a test for a method that accepts both types.
+    """
+    tmp_path = tmp_path_factory.mktemp('target')
+
+    if request.param:
+        tmp_path = str(tmp_path)
+
+    yield tmp_path
 
 
 def test_uuid(repository_uninitialised):
@@ -564,6 +579,42 @@ def test_walk(repository, generate_directory):
         (pathlib.Path('relative'), ['sub'], ['file_b']),
         (pathlib.Path('relative/sub'), [], ['file_c']),
     ]
+
+
+@pytest.mark.parametrize('path', ('.', 'relative'))
+def test_copy_tree(repository, generate_directory, tmp_path_parametrized, path):
+    """Test the ``Repository.copy_tree`` method."""
+    directory = generate_directory({'file_a': None, 'relative': {'file_b': None, 'sub': {'file_c': None}}})
+    repository.put_object_from_tree(str(directory))
+
+    repository.copy_tree(tmp_path_parametrized, path=path)
+    for root, dirnames, filenames in repository.walk(path):
+        for dirname in dirnames:
+            assert pathlib.Path(tmp_path_parametrized / root / dirname).is_dir()
+        for filename in filenames:
+            filepath = pathlib.Path(tmp_path_parametrized / root / filename)
+            assert filepath.is_file()
+            with repository.open(root / filename) as handle:
+                assert filepath.read_bytes() == handle.read()
+
+
+@pytest.mark.parametrize(('argument', 'value', 'exception', 'match'), (
+    ('target', None, TypeError, r'path .* is not of type `str` nor `pathlib.Path`.'),
+    ('target', 'relative/path', TypeError, r'provided target `.*` is not an absolute path.'),
+    ('target', pathlib.Path('.'), TypeError, r'provided target `.*` is not an absolute path.'),
+    ('path', pathlib.Path('file_a'), NotADirectoryError, r'object with path `.*` is not a directory.'),
+))
+def test_copy_tree_invalid(tmp_path, repository, generate_directory, argument, value, exception, match):
+    """Test the ``Repository.copy_tree`` method for invalid input."""
+    directory = generate_directory({'file_a': None})
+    repository.put_object_from_tree(str(directory))
+
+    if argument == 'target':
+        with pytest.raises(exception, match=match):
+            repository.copy_tree(target=value)
+    else:
+        with pytest.raises(exception, match=match):
+            repository.copy_tree(target=tmp_path, path=value)
 
 
 def test_clone(repository, generate_directory):
--- a/aiida/repository/repository.py
+++ b/aiida/repository/repository.py
@@ -473,3 +473,40 @@ class Repository:
                 yield from self.walk(path / dirname)
 
         yield path, dirnames, filenames
+
+    def copy_tree(self, target: Union[str, pathlib.Path], path: FilePath = None) -> None:
+        """Copy the contents of the entire node repository to another location on the local file system.
+
+        :param target: absolute path of the directory where to copy the contents to.
+        :param path: optional relative path whose contents to copy.
+        :raises TypeError: if ``target`` is of incorrect type or not absolute.
+        :raises NotADirectoryError: if ``path`` does not reference a directory.
+        """
+        path = self._pre_process_path(path)
+        file_object = self.get_object(path)
+
+        if file_object.file_type != FileType.DIRECTORY:
+            raise NotADirectoryError(f'object with path `{path}` is not a directory.')
+
+        if isinstance(target, str):
+            target = pathlib.Path(target)
+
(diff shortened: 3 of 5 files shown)


sphuber:
@ltalirz @chrisjsewell @ramirezfranciscof any takers?

On a side note, this PR made me think whether we need to add the possibility to create an empty directory in the virtual hierarchy of nodes. Currently this is not possible.

tests/repository/test_repository.py
```diff
@@ -566,6 +581,37 @@ def test_walk(repository, generate_directory):
...
+))
+def test_copy_tree_invalid(tmp_path, repository, argument, value, exception):
+    """Test the ``Repository.copy_tree`` method for invalid input."""
+    if argument == 'target':
+        with pytest.raises(TypeError, match=exception):
+            repository.copy_tree(target=value)
+    else:
+        with pytest.raises(TypeError, match=exception):
+            repository.copy_tree(target=tmp_path, path=value)
+
```
ramirezfranciscof, line 613: Why do you have this `if`, all your parameters have `argument == 'target'`, right? Was this supposed to support also testing the errors on `path`? (although I see no raises related to that variable...should we perhaps raise if the `path` doesn't exist in the node or would the raise inside `Repository.open` give enough info to understand the problem?)

tests/repository/test_repository.py
```diff
@@ -566,6 +581,37 @@ def test_walk(repository, generate_directory):
     ]
 
 
+def test_copy_tree(repository, generate_directory, tmp_path_parametrized):
+    """Test the ``Repository.copy_tree`` method."""
+    directory = generate_directory({'file_a': None, 'relative': {'file_b': None, 'sub': {'file_c': None}}})
+    repository.put_object_from_tree(str(directory))
+
+    repository.copy_tree(tmp_path_parametrized)
```
ramirezfranciscof, line 589: You are not testing when you provide a source path inside the node repo, no? Perhaps I would add a test for that. I wonder if it makes sense to have a cool parametrized fixture like you made for `Union[str, pathlib.Path]` but for `FilePath = Union[str, pathlib.PurePosixPath]` 🤔

ramirezfranciscof (commented):
All looks good! I just have 3 small comments/questions before approving.

chrisjsewell:
Eurgh, yeh it needs to be exposed in some fashion, but you are adding way too many methods on to `Node`, and there's absolutely no naming consistency, to identify what methods are associated with the repository (or anything else). It's just very user unfriendly 😬

This is why I've already looked into refactoring all this as e.g. `Node().repo.method` (#4938), which I may well circle back round to at some point

chrisjsewell:
Basically I would point you towards: https://python-patterns.guide/gang-of-four/composition-over-inheritance/#dodge-mixins: Using ["Composition over inheritance"](https://en.wikipedia.org/wiki/Composition_over_inheritance), is something we do ok in some areas of aiida, but poor in others

aiida/orm/nodes/repository.py
```diff
@@ -216,6 +216,14 @@ def walk(self, path: FilePath = None) -> Iterable[Tuple[pathlib.PurePosixPath, L
         """
         yield from self._repository.walk(path)
 
+    def copy_tree(self, target: Union[str, pathlib.Path], path: FilePath = None) -> None:
```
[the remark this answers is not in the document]
sphuber: The choice for the name of the argument `path` was taken because that is the name used in all other repository related methods where it concerns a relative path in the virtual hierarchy. That's why I decided to keep it the same for consistency.

tests/repository/test_repository.py
sphuber: Yeah, somehow one of the test cases where `path` is specified got lost. I will add it.

tests/repository/test_repository.py
sphuber: I will see how I can add the test.

sphuber:
> Eurgh, yeh it needs to be exposed in some fashion, but you are adding way too many methods on to `Node`, and there's absolutely no naming consistency, to identify what methods are associated with the repository (or anything else). It's just very user unfriendly grimacing
> 
> This is why I've already looked into refactoring all this as e.g. `Node().repo.method` (#4938), which I may well circle back round to at some point

I am not against improving the interface and I see the point for exposing all repository related methods under the `repo` namespace. The main problem is that we cannot change this now, since we introduced this naming in `v1.0` and don't want to break it for `v2.0`. Fine to start thinking about a better solution and introduce it at some point, deprecation the current interface.

Also very happy to have suggestions for the name of this method. I am not at all married to it. I simply took it as that is the name that is used in `shutil`.

Finally, I agree that too many methods at some point makes things difficult to oversee and in that respect it would be good to cleanup the `Node` namespace a bit. However, I don't think we will ever get to a really minimal interface. The reason is that within AiiDA the `Node` takes such a central role, that all reasoning, with respect to the user interface, happens from its perspective. Changing the design to having functions or other classes that consume a node don't solve the problem, because then it is definitely not "discoverable". To me it seems that it is therefore almost unavoidable that it has many methods. Wherever possible it is fine to group them and expose them in some subnamespace, as with the repository methods. But this can have downsides in and of itself for "user-friendliness". I am not convinced that just by tab-completing and looking at the methods you can always successfully intuit how a tool functions, especially not one that is as complex as AiiDA, for better or worse.

sphuber:
Thanks for the review @ramirezfranciscof . I have added the additional tests and included the changes that had snuck into the other PR.

aiida/orm/nodes/repository.py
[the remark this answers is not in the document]
ramirezfranciscof: mmm ok, I still think it is a bit confusing when you have a target path and a source path that one of them is just arbitrarily named `path`, but I agree that if we already got into that problem with other methods we might as well stay consistent now. Maybe if/when we do the migration discussed below we could take this into account for the new interface.

tests/repository/test_repository.py
ramirezfranciscof: I see it now, thanks! However there is no added check for the path existing in the node repo no? 😅 

I mean perhaps it is not necessary and the error that `self.walk(path)` would raise should be enough for the user to identify where he is making the error (if it already mentions the path and the node id, for example), but I just wanted explicit confirmation that it is ok to rely on that here.

tests/repository/test_repository.py
ramirezfranciscof: So you tagged me for re-review but I don't see this test added, did you decided not to do it? I mean, it is ok, I guess it is not super critical, I'm just checking it was intentional and not that you forgot or some commit didn't get pushed =).

ramirezfranciscof:
> Basically I would point you towards: https://python-patterns.guide/gang-of-four/composition-over-inheritance/#dodge-mixins: Using ["Composition over inheritance"](https://en.wikipedia.org/wiki/Composition_over_inheritance), is something we do ok in some areas of aiida, but poor in others

This is actually a super interesting concept I came across somewhat recently while reading [Head First Design Patterns](https://www.goodreads.com/book/show/58128.Head_First_Design_Patterns) (highly recommended if you can get past the silly tone 😅). I would find it really exciting to have these kind of design concepts come up more when discussing and deploying implementations and refactors in the code!

More practically, it would also be very helpful when doing improvements like #4938 if there was a bit more explanation on how these kind of principles affect the decisions on making the specific changes, and how these influence the technical aspects of the implementation. I mean, I understand these might seem obvious being the author, but it is sometimes very hard to just look at the changes and deduce what one is trying to achieve, specially since in implementations these concepts can't be usually applied in "pure" form. For example, while it is true that the `NodeRepositoryMixin` is technically being applied through inheritance, I don't think it is purely that, since the repository is being composed into the mixin (kept in the `_repository_instance` variable). So in a way is like using composition, but further abstracting and separating all methods related to this composition from the class. Of course one would then argue that the point of the composition is not adding extra methods to the class, to which the retort would be that the extra methods serve as "interface shortcuts" for the users (which I think is essentially the summary of the short exchange above). But I guess my point is that (only?) having that discussion in the abstract can soon become either circular, repetitive, or aimless, and can make it hard to turn into something actionable.

In any case, it is true what @sphuber says that this is more a long term concern and not something we can resolve before 2.0, so maybe we can continue this discussion in other more specific issues and accept for now that this PR will have to construct on what is already there.

tests/repository/test_repository.py
sphuber: > However there is no added check for the path existing in the node repo no?

There is though. Granted, the check is led from the perspective of the source `repository`, but it uses `walk(path=path)` to check that all the files that should have been copied, actually exist in `tmp_path_parametrized`. Of course, if `walk` is bugged, we could have a false positive in some really weird obscure case, potentially. But that method itself is also tested so I think it is fine.

sphuber: Rereading your last comment, I think you may have put this under the wrong initial comment which concerns the test for invalid arguments. That does not include the use of `walk` whatsoever. Probably you meant to say this with the other comment below on the `test_copy_tree` test, which does include the use of `walk`. My answer applies the same though.

tests/repository/test_repository.py
sphuber: I have added the test though. It now parametrizes over `path` as follows:
```python
@pytest.mark.parametrize('path', ('.', 'relative'))
```
and `path` is passed both to `copy_tree` and then used in `walk` for the check. I didn't also parametrize each of these values over `str` and `pathlib.Path`.

ramirezfranciscof: Totally right, my bad I missed that!

tests/repository/test_repository.py
ramirezfranciscof: > Granted, the check is led from the perspective of the source `repository`, but it uses `walk(path=path)` to check that all the files that should have been copied

Right, ok, so this is what I meant with "relying on `self.walk(path)` to do the check and raise". It is just that sometimes lower level functions give less informative errors and you need to `try / except` and re-raise with more info, but I guess here it should be fine.

ramirezfranciscof (approved):
This is good to go for me! (it is out-of-date though, not sure if that means you need approval once again after updating)

Issue #4566, opened 2020-11-16
`CalcJob`: characterize the behavior of the copy-lists through unit tests


ramirezfranciscof:
More specifically, I think it is not clear which is the expected behavior when copying recursively. For example, when using as source `./dir/` or `./dir/*`, when using as targets `./dir/` or `./dir/.`. Although this might be partially due to the fact that the underlying tools we use may have different behaviors for these (inside the transport plugins), maybe we should establish our own set of rules and aim for obtaining the same result irrespective of the tool used.

ramirezfranciscof:
This is a continuation from #4499; I'll be closing that issue since the cause of that particular error has been reverted so the example codes no longer really apply.

ramirezfranciscof:
![](https://github.trello.services/images/mini-trello-icon.png) [Testing powerup](https://trello.com/c/AH9eetUA/52-testing-powerup)

greschd:
@ramirezfranciscof not sure if these links are supposed to be publicly accessible, but all I'm seeing is 

![image](https://user-images.githubusercontent.com/5692808/105477768-9b9ac180-5ca2-11eb-9931-483a299447be.png)

ramirezfranciscof:
@greschd Haha, no, sorry, thanks for the heads up. I was trying the github powerup in trello as a way of keeping a private list of issues/prs I want to track, but apparently it is automatically commenting on the github page...not cool.

greschd:
Ha yeah, not ideal 😄

PR #5115, opened 2021-09-03
`CalcJob`: allow directories in `local_copy_list`

### LINKED ISSUE #3869: Allow to pass a folder in `local_copy_list`
Currently the `local_copy_list` only supports the copying of individual files from the repository of any data node. However, sometimes one might want to copy over the entire contents of a node's repository, or at least of a sub directory, to the remote working directory. Implementation should be relatively simple: just check whether the second element of any entry in `local_copy_list` corresponds to a directory instead of a file. If it is a directory copy its contents recursively.

sphuber:
Fixes #3869 

Up till now, the `local_copy_list` could only be used to specify
individual files that should be copied. However, very often, one may
have an input node from whose repository an entire directory, or sub
directory should be copied. The only way to do this was to manually
iterate over the contents of that repository directory and add the files
to the `local_copy_list` one by one.

Here we extend the syntax of the `local_copy_list` and the second
argument can now also point to a directory in the repository of the
source node. Its entire contents will be copied to the relative path
defined by the third element in the tuple. Note that the directory
itself won't be copied, just its contents.

### DIFF
--- a/aiida/engine/daemon/execmanager.py
+++ b/aiida/engine/daemon/execmanager.py
@@ -16,6 +16,7 @@ plugin-specific operations.
 from collections.abc import Mapping
 from logging import LoggerAdapter
 import os
+import pathlib
 import shutil
 from tempfile import NamedTemporaryFile
 from typing import Any, List, Optional, Mapping as MappingType, Tuple, Union
@@ -26,6 +27,7 @@ from aiida.common.folders import SandboxFolder
 from aiida.common.links import LinkType
 from aiida.orm import load_node, CalcJobNode, Code, FolderData, Node, RemoteData
 from aiida.orm.utils.log import get_dblogger_extra
+from aiida.repository.common import FileType
 from aiida.schedulers.datastructures import JobState
 from aiida.transports import Transport
 
@@ -181,7 +183,7 @@ def upload_calculation(
                     transport.put(handle.name, filename)
             transport.chmod(code.get_local_executable(), 0o755)  # rwxr-xr-x
 
-    # local_copy_list is a list of tuples, each with (uuid, dest_rel_path)
+    # local_copy_list is a list of tuples, each with (uuid, dest_path, rel_path)
     # NOTE: validation of these lists are done inside calculation.presubmit()
     local_copy_list = calc_info.local_copy_list or []
     remote_copy_list = calc_info.remote_copy_list or []
@@ -199,13 +201,26 @@ def upload_calculation(
         if data_node is None:
             logger.warning(f'failed to load Node<{uuid}> specified in the `local_copy_list`')
         else:
-            dirname = os.path.dirname(target)
-            if dirname:
-                os.makedirs(os.path.join(folder.abspath, dirname), exist_ok=True)
-            with folder.open(target, 'wb') as handle:
-                with data_node.open(filename, 'rb') as source:
-                    shutil.copyfileobj(source, handle)
-            provenance_exclude_list.append(target)
+
+            # If no explicit source filename is defined, we assume the top-level directory
+            filename_source = filename or '.'
+            filename_target = target or ''
+
+            # Make the target filepath absolute and create any intermediate directories if they don't yet exist
+            filepath_target = pathlib.Path(folder.abspath) / filename_target
+            filepath_target.parent.mkdir(parents=True, exist_ok=True)
+
+            if data_node.get_object(filename_source).file_type == FileType.DIRECTORY:
+                # If the source object is a directory, we copy its entire contents
+                data_node.copy_tree(filepath_target, filename_source)
+                provenance_exclude_list.extend(data_node.list_object_names(filename_source))
+            else:
+                # Otherwise, simply copy the file
+                with folder.open(target, 'wb') as handle:
+                    with data_node.open(filename, 'rb') as source:
+                        shutil.copyfileobj(source, handle)
+
+                provenance_exclude_list.append(target)
 
     # In a dry_run, the working directory is the raw input folder, which will already contain these resources
     if not dry_run:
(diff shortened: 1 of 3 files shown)


sphuber:
@ramirezfranciscof @mbercx want to give this a look? Note that this includes another commit that is being added separately in PR #5114 but the second commit depends on it.

aiida/engine/daemon/execmanager.py
```diff
@@ -199,13 +201,26 @@ def upload_calculation(
...
+                # If the source object is a directory, we copy its entire contents
+                data_node.copy_tree(filepath_target, filename_source)
+                provenance_exclude_list.extend(data_node.list_object_names(filename_source))
+            else:
+                # Otherwise, simply copy the file
+                with folder.open(target, 'wb') as handle:
+                    with data_node.open(filename, 'rb') as source:
+                        shutil.copyfileobj(source, handle)
+
+                provenance_exclude_list.append(target)
```
ramirezfranciscof, line 223: Do we need this conditional? I think `data_node.copy_tree` should work when it is provided with the path to a single file, no? (actually it might be good to add a test for this specifically in #5114 )

Same question for `data_node.list_object_names`; even if it currently doesn't, maybe it would be easier to make that work with paths to files instead of adding complexity here.

aiida/repository/repository.py
```diff
@@ -486,10 +486,10 @@ def copy_tree(self, target: Union[str, pathlib.Path], path: FilePath = None) ->
             target = pathlib.Path(target)
 
         if not isinstance(target, pathlib.Path):
-            raise TypeError('path is not of type `str` nor `pathlib.Path`.')
+            raise TypeError(f'path `{path}` is not of type `str` nor `pathlib.Path`.')
 
         if not target.is_absolute():
-            raise TypeError('provided target is not an absolute path.')
+            raise TypeError(f'provided target `{target}` is not an absolute path.')
```
ramirezfranciscof: Why are you fixing this here instead of modifying the PR #5114 ? I mean, it is still up, wouldn't it be cleaner to have that there?

ramirezfranciscof (commented):
Also good for me! Just two comments to discuss related to the interaction of this PR with the one it is based on.

aiida/repository/repository.py
sphuber: Moved to 5114

aiida/engine/daemon/execmanager.py
sphuber: Because I don't `copy_tree` should accept a file for a `path`. I will add an explicit check in the other PR and add a test.

ramirezfranciscof: Mmm ok, I see. I guess I understand how "listing" is technically different if interpreted as "show the content of" and that doesn't work with files, but "copy" is more general and works with both. Is it the "tree" qualifier that for you implies that what is copied should be a directory and files should be identified and rejected? Because I see no practical complications for copying files and folders indistinctively...

Also, related to this: what would be the problem of doing `provenance_exclude_list.append(target)` when `target` is a folder? I mean, when you do `data_node.list_object_names(filename_source))`, the list you get back also includes subfolders, right? So you are already adding folders to the `exclude_list`.

On the other hand, I guess one problem is that if you add a subfolder like that and then create some input file that should be included in the provenance inside of that subfolder, then it will be ignored because the folder is on the exclude list, right? But we are anyways adding subfolders because of previous paragraph, so this problem is more complex and not solved by this `if`.

sphuber: > Mmm ok, I see. I guess I understand how "listing" is technically different if interpreted as "show the content of" and that doesn't work with files, but "copy" is more general and works with both. Is it the "tree" qualifier that for you implies that what is copied should be a directory and files should be identified and rejected? Because I see no practical complications for copying files and folders indistinctively...

With the name `copy_tree`, I think it should definitely not accept files. The `tree` indeed signifies that it operates on a directory, like `put_object_from_tree`. For files there is `put_object_from_file(like)`. It can be argued that there should be a generic method that copies both, but I don't think it necessarily makes things easier to use or implement.

> Also, related to this: what would be the problem of doing `provenance_exclude_list.append(target)` when `target` is a folder? I mean, when you do `data_node.list_object_names(filename_source))`, the list you get back also includes subfolders, right? So you are already adding folders to the `exclude_list`.

You are correct. Since I expand the filenames in their separate subpaths when they are copied over to the node's repository, this will also include the subpath that equals `target`. So there is indeed no need to expand the target path already here in its subfolders. I will change this.

> On the other hand, I guess one problem is that if you add a subfolder like that and then create some input file that should be included in the provenance inside of that subfolder, then it will be ignored because the folder is on the exclude list, right? But we are anyways adding subfolders because of previous paragraph, so this problem is more complex and not solved by this `if`.

This is a good point. I guess this might indeed be problematic, but I don't see a clean solution right now. If the file really needs to be stored in the node's repo, then they should simply write the entire folder directly to the sandbox and not use the `local_copy_list`. Anyway, this entire system of the `local_copy_list` is not an ideal solution but it is there for historical reasons. Really what we should have done from the beginning is just have the plugin write everything to the sandbox and then specify just the `provenance_from_exclude` list. This gives full flexibility in terms of what gets stored and what not and it is a lot simpler to understand. This is definitely a thing we should consider simplifying for v3.0 if that ever comes out.

sphuber: Actually, thinking about it, we cannot simplify this by just adding `target` to the `provenance_exclude_list`. The reason is that if the target is `.`, i.e., add all contents to the root, we don't want to exclude everything that is in there. Just those resources that were written by that operation. So what I had originally, expanding the source path into its contents, is the correct thing to do.

ramirezfranciscof: Uhm, I see. I still feel a bit uneasy about this, but I think this is a general problem of implementing file copy interfaces. It kind of reminds me of when I tried to add directory creation to the copy lists and we had to roll it back because it changed the behavior of how some copy commands were indicated, prompting the publication of issue #4566 out of my frustration. What a headache =/...

Anyways, I still think it would be good to do a more thorough design of this and define a specific interface, but that is beyond the scope of this PR, so I'm just going to approve this now.

ramirezfranciscof (approved):
All good to go, thanks!

aiida/engine/daemon/execmanager.py
sphuber: I agree that the interface is shit, but the real solution is to get rid of `local_copy_list` entirely. The `remote_copy_list` is slightly different, but ideally that could also be replaced with an interface that allows a user to _directly_ manipulate the remote file system instead of giving abstract indirect instructions.

ramirezfranciscof: I don't know, I get the feeling that you will get this kind of annoying headaches with almost any sort of general "file copying" interface you end up implementing. I mean, just from writing bash scripts with `cp` or `rsync` I remember having problems with suddenly unexpected behavior related to the exact syntax used in the command and the specifics of the situation. I guess that is the reason why you or shutil choose to do `copy_TREE` kind of methods that are more restrictive (instead of what I wanted to just copy whatever you provide) but at least the expected result is more clear.

> The `remote_copy_list` is slightly different, but ideally that could also be replaced with an interface that allows a user to _directly_ manipulate the remote file system instead of giving abstract indirect instructions.

I don't understand, what do you mean by _directly_ manipulating the remote file system?

sphuber: > I don't understand, what do you mean by directly manipulating the remote file system?

By letting the user call direct methods to write/read files instead of using indirect instructions through `remote_copy_list`. It would automatically get a transport, which can then be used to perform the operations directly. This would ideally have an interface that is identical to writing files to the local file system, making it transparent whether you are reading/writing from/to a local or remote file system.

PR #5119, opened 2021-09-07
CLI: rework the logging verbosity implementation

### LINKED ISSUE #5116: Empty profile list when upgrading from v1.6.5 to `develop`
### Describe the bug

This is the output of `verdi profile list` on my machine for aiida-core v1.6.5
```
$ verdi profile list
Info: configuration folder: /.../aiida_rmq/.aiida
  discover-hcofs
  fold-nodes
  mcloud_paper
* nanoporous_pipeline
  pyrene-mofs
  radames_django
```

When I update to `develop`, I get this
```
$ verdi profile list
$ verdi profile show
```

No error message, no warning, just an empty list.

In case it is relevant, I am using the `AIIDA_PATH` variable to specify the location of the AiiDA folder.

This is an excerpt from the config file

```
{
    "CONFIG_VERSION": {
        "CURRENT": 5,
        "OLDEST_COMPATIBLE": 5
    },
    "profiles": {
        "radames_django": {
            "AIIDADB_BACKEND": "django",
            "default_user_email": "radames.verdi@ope.ra",
            "AIIDADB_HOST": "localhost",
            "AIIDADB_PORT": 65256,
            "AIIDADB_NAME": "aiida_test_setup_django",
            "AIIDADB_USER": "aiida_SetupTestCase",
            "AIIDADB_PASS": "...",
            "AIIDADB_REPOSITORY_URI": "file:///.../aiida_rmq/aiida_radames_django",
            "PROFILE_UUID": "0646945891af44a1b0f91529dca7e714"
        },
```

### Expected behavior

If the config file (or repo, ...) need to be migrated I would expect a warning and instructions on what to do.

### Your environment

- Python version [e.g. 3.7.1]: 3.9
- aiida-core version [e.g. 1.2.1]: 1.6.5 + develop

any thoughts @sphuber ?

P.S. When running a command that needs the db, I get

```
$ verdi process list
Critical:
Database schema version `1.0.48` is incompatible with the required schema version `1.0.49`.
To migrate the database schema version to the current one, run the following command:

    verdi -p nanoporous_pipeline database migrate
```

Is it possible that some changes in `develop` are causing AiiDA to need access to the database just in order to do a `verdi profile list`?

sphuber:
Fixes #5116 

The intended behavior of logging of the system when invoked through the
command line interface `verdi` is as follows:

 * The verbosity of output produced by invoked CLI commands can be
   controlled through a single `-v/--verbosity` option that is accepted
   by the CLI at every command level.
 * It takes the values of the log levels of Python's logging system,
   including AiiDA's custom `REPORT` level.
 * The verbosity level should be applied to all logging messages, so not
   just those messages that are logged directly by the command itself,
   but also any messages logged by module code that is indirectly called
   in the stack.

Given that the implementation is built on top of the logging system, the
log level set by the verbosity option has to be added to the logging
configuration somehow. The logging is configured by invoking the method
`aiida.common.log.configure_logging` which consumes a dictionary with
the logging configuration that has certain fields that are dynamically
determined from the loaded profile. Unfortunately, this method can be
called by module code as well, therefore, the verbosity option can not
rely on using this method to configure the logging. Because once the
callback is finished, any other code invoked by the command that calls
this same method, will undo the changes of the verbosity option.

The original implementation attempted to solve this problem by storing
the selected log level through the verbosity option, in the profile
instance loaded in memory. Since the log level would be read dynamically
from the profile in the `configure_logging` method, even when it would
be recalled, the correct log level would be configured.

This approach had various disadvantages, however. It tied the option
closely to the concept of a profile. Even though all `verdi` commands
have the profile option added by default and therefore this does not
really pose a problem, it made the verbosity option useless in any other
context. The tight coupling also made the logic more difficult to
understand. Finally, even though the change of the profile setting was
only made in memory and so in principle the change should be temporary
within the scope of the CLI command, if the command decided to store the
profile, which can for example happen in commands that are designed to
change the profile, the change in log level option would also be written
to disk, causing an undesirable side effect of this implementation.

Here, we change the implementation by fully decoupling the verbosity
option from the profile and simply having the callback write the
selected log level to a global constant in the `aiida.common.log` module
called `CLI_LOG_LEVEL`. The `configure_logging` module will check this
constant and when set, will configure all loggers with the `cli` handler
with the correspondig log level. This design still has a downside in
that it hard-couples the generic `aiida.common.log` module to the more
specific `aiida.cmdline` module, but it is not obvious to get the CLI
dependent log level injected into this module some other way.

Finally, the original implementation allowed a default value for the
verbosity option to be configured through the config by setting the
`logging.aiida_loglevel` option. However, since that also determines the
log level for AiiDA when not run through the CLI, i.e. during normal
operations, this could have undesirable side-effects. For example, if
the option was set to `WARNING` to temporarily limit the number of logs
from processes when being run, certain CLI commands would also no longer
report any output.


sphuber:
@ltalirz given that you requested the change in behavior, would be good if you could have a look to see if this matches your expectation. Note that there is one test failing, but that is the one that I added and it is a peculiarity of the testing itself in combination with the logging. I don't think the underlying functionality is actually broken. So you can already have a look at the implementation while I try to find a way to fix the test.

aiida/cmdline/utils/decorators.py
```diff
@@ -47,10 +47,7 @@ def load_backend_if_not_loaded():
         with spinner():
             load_profile()  # This will load the default profile if no profile has already been loaded
             manager.get_backend()  # This will load the backend of the loaded profile, if not already loaded
-            # Loading the backend will have reconfigured the logging (see ``Manager._load_backend`` which calls the
-            # ``aiida.common.log.configure_logging`` function but with the default ``cli=False``), so here we have to
-            # call it again to ensure that the correct CLI handlers and formatters are configured.
-            configure_logging(with_orm=True, cli=True)
+            configure_logging(with_orm=True)
```
ltalirz, line 50: if the `get_backend` call already configured the logging, is it still necessary to do it again?

docs/source/topics/cli.rst
```diff
@@ -127,18 +127,6 @@ is identical to
     verdi --verbosity debug process list
 
 When the option is specified multiple times, only the last value will be considered.
-The default value for the logging level can be changed permanently through the configuration system.
```
ltalirz, line 130: for the daemon logs this is still mentioned somewhere else, right?

tests/cmdline/params/options/test_verbosity.py
```diff
@@ -0,0 +1,106 @@
...
+        else:
+            assert f'{log_level_name.capitalize()}: verdi' not in result.output
+            assert f'{log_level_name.capitalize()}: aiida' not in result.output
+
+
+@pytest.mark.usefixtures('reset_log_level')
+def test_config_aiida_loglevel(run_cli_command, isolated_config, caplog):
+    """Test the behavior of the ``--verbosity`` option when the ``logging.aiida_loglevel`` config option is set.
+
+    Even though the ``CMDLINE_LOGGER`` is technically a child of the ``AIIDA_LOGLEVEL`` and so normally the former
```
ltalirz, line 76: Since, as you mention, this is a bit counterintuitive behavior, is there something that speaks against reversing the inheritance and making the `CMDLINE_LOGGER` the parent of the AiiDA logger?

In principle one could then even decide not to care about profile-specific reconfiguration of the AiiDA logger (but I guess it's probably still wise to enforce the log level handed down from the command line also there).

aiida/cmdline/params/options/main.py
```diff
@@ -90,18 +90,19 @@ def decorator(command):
     return decorator
 
 
-def set_log_level(ctx, __, value):
-    """Set the log level for the global ``AIIDA_LOGGER``.
+def set_log_level(_, __, value):
```
ltalirz, line 93: just to understand: the first to arguments were arguments that became obsolete over time?

Perhaps it would be worth mentioning this in the docstring.

aiida/cmdline/params/options/main.py
```diff
@@ -90,18 +90,19 @@ def decorator(command):
...
+    """Set the log level for all loggers.
```
ltalirz, line 94:
```suggestion
    """Fix the log level for all loggers from the cli.
```

ltalirz (commented):
thanks a lot @sphuber !

looks good to me, I have mainly one question on the hierarchy of the loggers

aiida/cmdline/params/options/main.py
sphuber: This is the standard interface for a `click` callback. The first is the `ctx` context and the second `param` is the parameter. We used to use the `ctx` to get the loaded profile, but now it is fully decoupled so we no longer need it.

docs/source/topics/cli.rst
sphuber: Yep: https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/troubleshooting.html?highlight=logging.aiida_loglevel#increasing-the-logging-verbosity

aiida/cmdline/utils/decorators.py
sphuber: Good point, not sure where this got added first, but it seems to be doubling up indeed. Will check to make sure and then remove it if really not needed

tests/cmdline/params/options/test_verbosity.py
sphuber: We could try, but I think logger inheritance in the `logging` module works by the module path. So the `aiida` module has `aiida` as the logger name and the `aiida.cmdline` module has `aiida.cmdline` and therefore is a child. I am not even sure if we can even make a logger the parent of `aiida` since that is a top-level namespace.

aiida/cmdline/params/options/main.py
ltalirz: > This is the standard interface for a `click` callback.

I see, perhaps good to mention this in the docstring?
I would have tended to use `_ctx` and `_param` as the argument names to make it easier for the next guy who has to start using them again.

tests/cmdline/params/options/test_verbosity.py
ltalirz: > We could try, but I think logger inheritance in the `logging` module works by the module path.

I think that is what happens if you do `logging.getLogger(__name__)` because `logging.getLogger('aiida.submodule')` means "create logger `submodule` as child of logger `aiida`)
See the [logging docs](https://docs.python.org/3/library/logging.html#logger-objects).

So one could in principle do something like
```
CMDLINE_LOGGER = logging.getLogger('aiida_cli')  # use some unique name
AIIDA_LOGGER = CMDLINE_LOGGER.get('aiida')
```
but that would change the name of the aiida logger to 'aiida_cli.aiida' which does not follow standard practices.

We can do something else though:
```
CMDLINE_LOGGER = logging.getLogger('aiida')
AIIDA_LOGGER = CMDLINE_LOGGER.getChild('api')
```
and let `configure_logging` configure the `aiida.api` logger.

Anyhow, it may be somewhat confusing either way, so I leave the decision up to you.

aiida/cmdline/params/options/main.py
sphuber: Done

aiida/cmdline/params/options/main.py
sphuber: Done

aiida/cmdline/utils/decorators.py
sphuber: Checked and removed it

tests/cmdline/params/options/test_verbosity.py
sphuber: That's also what I surmised. Given that it anyway would be not ideal and there would always be something that would be an bit backwards, I would keep what I have currently.

ltalirz (approved):
thanks a lot @sphuber , looks good to me!

PR #5120, opened 2021-09-07
🔧 MAINTAIN: Pin setuptools<58.0.2

chrisjsewell:
The installation of the coverage package fails
(a dependency of pytest-cov),
due to this breaking change: https://github.com/pypa/setuptools/issues/2775


chrisjsewell:
naturally, we should look to upgrade coverage, but that will take some work to update the rest of the pytest infrastructure

chrisjsewell:
should actually be fixed with 58.0.3 ...

PR #5051, opened 2021-08-04
schedulers: use num_cores_per_mpiproc in direct

dev-zero:

PR #5126, opened 2021-09-09
`DirectScheduler`: use `num_cores_per_mpiproc` if defined in resources

sphuber:
If `num_cores_per_mpiproc` is specified in the job resources, the value
will now be exported as the `OMP_NUM_THREADS` variable.

Supersedes #5051


sphuber:
Merging this myself without approval. It was already approved by @giovannipizzi in #5051

PR #5127, opened 2021-09-09
🧪 TESTS: Add `CREATEDB` priviliges for DB user

chrisjsewell:
Allows for `tests/backends/aiida_sqlalchemy/test_migrations.py`
to run with auto-generated `aiida` users in tests

The tests run on Github Actions are specifially set up with a profile that uses the `postgres` super user: https://github.com/aiidateam/aiida-core/blob/1519278ab57635e1ba26ec95b6747393bd939e51/.github/config/profile.yaml#L12

whereas the default test database user a standard `aiida` user: https://github.com/aiidateam/aiida-core/blob/1519278ab57635e1ba26ec95b6747393bd939e51/aiida/manage/tests/main.py#L45

The `postrges` super user already has all privileges, by default, but the `aiida` user requires `CREATEDB` to carry out the migration tests.

### DIFF
--- a/aiida/manage/external/postgres.py
+++ b/aiida/manage/external/postgres.py
@@ -91,7 +91,7 @@ class Postgres(PGSU):
         """
         return bool(self.execute(_USER_EXISTS_COMMAND.format(dbuser)))
 
-    def create_dbuser(self, dbuser, dbpass, priviliges=""):
+    def create_dbuser(self, dbuser, dbpass, priviliges=''):
         """
         Create a database user in postgres
 
--- a/aiida/manage/external/postgres.py
+++ b/aiida/manage/external/postgres.py
@@ -22,6 +22,7 @@ from aiida.cmdline.utils import echo
 
 __all__ = ('Postgres', 'PostgresConnectionMode', 'DEFAULT_DBINFO')
 
+# The last placeholder is for adding privileges of the user
 _CREATE_USER_COMMAND = 'CREATE USER "{}" WITH PASSWORD \'{}\' {}'
 _DROP_USER_COMMAND = 'DROP USER "{}"'
 _CREATE_DB_COMMAND = (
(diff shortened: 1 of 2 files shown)


aiida/manage/external/postgres.py
```diff
@@ -22,7 +22,7 @@
 
 __all__ = ('Postgres', 'PostgresConnectionMode', 'DEFAULT_DBINFO')
 
-_CREATE_USER_COMMAND = 'CREATE USER "{}" WITH PASSWORD \'{}\''
+_CREATE_USER_COMMAND = 'CREATE USER "{}" WITH PASSWORD \'{}\' {}'
```
ltalirz, line 25: perhaps add a comment here what the last field is for (it's less obvious than for the other fields)

aiida/manage/tests/main.py
```diff
@@ -319,7 +319,9 @@ def create_aiida_db(self):
             self.create_db_cluster()
         self.postgres = Postgres(interactive=False, quiet=True, dbinfo=self.dbinfo)
         # note: not using postgres.create_dbuser_db_safe here since we don't want prompts
-        self.postgres.create_dbuser(self.profile_info['database_username'], self.profile_info['database_password'])
+        self.postgres.create_dbuser(
```
ltalirz, line 322: please add a comment here explaining why we need to have a user with createdb privileges here
(while migration typically should not require createdb privileges)

ltalirz (changes requested):
thanks @chrisjsewell 

just two minor clarifications

aiida/manage/tests/main.py
chrisjsewell: done

aiida/manage/external/postgres.py
chrisjsewell: done

chrisjsewell:
cheers @ltalirz addressed comments

ltalirz (approved):
thanks a lot!

PR #5133, opened 2021-09-13
fix: remove branch restriction for docker push

### LINKED ISSUE #5132: aiida-core docker image outdated
```
$ docker pull aiidateam/aiida-core:1.6.5
Error response from daemon: manifest for aiidateam/aiida-core:1.6.5 not found: manifest unknown: manifest unknown
```
In particular, this means the commands from the [docs](https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/run_docker.html) is currently do not work.

This is a result of docker silently [discontinuing autobuilds for free accounts](https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/run_docker.html).

It seems, however that as an open-source project one can [apply for open source status](https://www.docker.com/blog/expanded-support-for-open-source-software-projects/) and still get this feature.

I will take care of the application, could someone else please upload the latest docker container (perhaps @yakutovicha or @mbercx ?)

ltalirz:
fixes #5132 

The push of docker images to dockerhub was limited to the `master` and
`develop` branches, thus excluding maintenance releases on support
branches.

Here we remove the branch limitation, mirroring the corresponding
selector in the `release` workflow.

### DIFF
--- a/.github/workflows/push_image_to_dockerhub.yml
+++ b/.github/workflows/push_image_to_dockerhub.yml
@@ -5,9 +5,6 @@ name: build-and-push-to-dockerhub
 
 on:
   push:
-    branches:
-      - master
-      - develop
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+*"
 


yakutovicha (commented):
@ltalirz, I think I now understand why version 1.6.5 of the container was not built. It was simply because there is no `build-and-push-to-dockerhub` action there. The action was added recently to the `develop` branch and has never been merged into `master`. Therefore, I think we should close the PR.

ltalirz:
Thanks for looking into this @yakutovicha - makes sense!

I will close this PR but could you please do the following: 
 * check out `support/1.6.x` locally
 * cherry-pick the commit that adds the github action into the branch
 * `git commit --amend`  and add `Cherry-Pick: $hash` at the bottom of the commit message, where $hash is the hash of the original commit
 * push to `support/1.6.x`

There may be further support releases in the 1.6 line, which are not going to be merged into `master` after 2.0 is out, so it would be good to make sure that docker images are being built for them.

PR #5134, opened 2021-09-13
bump aiida-prerequisites

ltalirz:
This includes an upgrade from Ubuntu 18.04 to 20.04.

Also remove installation of custom circus dependency (issue was fixed in
circus 17.0).

### DIFF
--- a/.jenkins/Jenkinsfile
+++ b/.jenkins/Jenkinsfile
@@ -48,10 +48,10 @@ pipeline {
                     // this folder is mounted from a volume so will have wrong permissions
                     sh 'sudo chown root:root /home/jenkins/.cache'
                     // prepare environment (install python dependencies etc)
-                    sh 'pip install -r requirements/requirements-py-3.7.txt --cache-dir /home/jenkins/.cache/pip'
+                    sh 'pip install -r requirements/requirements-py-3.8.txt --cache-dir /home/jenkins/.cache/pip'
                     sh 'pip install --no-deps .'
                     // for some reason if we don't change permissions here then python can't import the modules
-                    sh 'sudo chmod -R a+rwX /opt/conda/lib/python3.7/site-packages/'
+                    sh 'sudo chmod -R a+rwX /opt/conda/lib/python3.8/site-packages/'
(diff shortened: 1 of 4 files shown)


.jenkins/Jenkinsfile
```diff
@@ -48,10 +48,10 @@ pipeline {
                     // this folder is mounted from a volume so will have wrong permissions
                     sh 'sudo chown root:root /home/jenkins/.cache'
                     // prepare environment (install python dependencies etc)
-                    sh 'pip install -r requirements/requirements-py-3.7.txt --cache-dir /home/jenkins/.cache/pip'
+                    sh 'pip install -r requirements/requirements-py-3.8.txt --cache-dir /home/jenkins/.cache/pip'
```
csadorf, line 51: Why 3.8 and not 3.9?

csadorf (changes requested):
There are a lot of checks failing and it looks like the change of the Python version requires an update of the Jenkins CI job.

.jenkins/Jenkinsfile
ltalirz: I think it has to be the native python version (which is 3.8 for ubuntu 20.04)

ltalirz: > There are a lot of checks failing and it looks like the change of the Python version requires an update of the Jenkins CI job.

Ehm... are you looking at the first or the second commit?

csadorf: Indeed, I looked at the first commit for the jenkins failure, but there are still these "failed check" annotations, do you know what this is about?
Screen Shot 2021-09-14 at 09 28 28

ltalirz: I don't know - when I click on the icons in the right top corners it brings to jobs that succeeded just fine.

Also, these tests are unrelated to the dockerfile.

I think it's somehow a leftover from the first commit

csadorf: Ok, I assume it is fine then.

csadorf (approved):

Issue #4991, opened 2021-06-18
The default plugin not mandatory to be given in setting code 


unkcpz:
### Is your feature request related to a problem? Please describe

I have two `CalcJob` classes `ACalculation` and `BCalculation`. `ACalculation` is run with `calcinfo.codes_info = [helpercodeinfo, codeinfoA]`; while `BCalculation` is run with `calcinfo.codes_info = [helpercodeinfo, codeinfoB]`. They share a code corresponding to `helpercodeinfo` in order to pre-excuted to give an extra input file to run the main code. 

However, during setting the code the user is asked to set the default plugin for the code. In my case, the code is not bound to any calculation plugin and I can set any default plugin to the code. 
All the necessary information about the code is the path and the computer, I believe the default plugin is not a mandatory setting for the code. 

### Describe the solution you'd like

I think users are allowed to set the code without setting the default plugin. The downside effect I can see is that code without default plugin prevents getting process builder from code directly. 

### Describe alternatives you've considered

Or a default plugin(maybe called `core.dummy`) can be given with an entry point along with aiida-core to meet the requirement of the case.

PR #5140, opened 2021-09-17
Command line: allow to not set input plugin for code

unkcpz:
For feature #4991. 

An extra flag option `--bind-to-input-plugin` with default flag to `True` is add to `verdi code setup` to allow not to set the input plugin for the code.
This may increase the configuration complex a bit of code setup. Especially when setting code through config file, if user do not want to bind input plugin to the code they need to set this flag to `false` explicitly like: 

```
---
label: {label}                    
bind_to_input_plugin: no
computer: {computer}
remote_abs_path: /remote/abs/path
```

### DIFF
--- a/aiida/cmdline/commands/cmd_code.py
+++ b/aiida/cmdline/commands/cmd_code.py
@@ -82,6 +82,8 @@ def setup_code(non_interactive, **kwargs):
     else:
         kwargs['code_type'] = CodeBuilder.CodeType.STORE_AND_UPLOAD
 
+    kwargs.pop('bind_to_input_plugin')
+
     code_builder = CodeBuilder(**kwargs)
 
     try:
--- a/aiida/cmdline/commands/cmd_code.py
+++ b/aiida/cmdline/commands/cmd_code.py
@@ -61,6 +61,7 @@ def set_code_builder(ctx, param, value):
 @verdi_code.command('setup')
 @options_code.LABEL()
 @options_code.DESCRIPTION()
+@options_code.BIND_TO_INPUT_PLUGIN()
 @options_code.INPUT_PLUGIN()
 @options_code.ON_COMPUTER()
 @options_code.COMPUTER()
--- a/aiida/cmdline/params/options/commands/code.py
+++ b/aiida/cmdline/params/options/commands/code.py
@@ -16,6 +16,7 @@ from aiida.cmdline.params.options.overridable import OverridableOption
 
 
 def is_on_computer(ctx):
+    import ipdb; ipdb.set_trace()
     return bool(ctx.params.get('on_computer'))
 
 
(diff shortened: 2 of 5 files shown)


sphuber:
Thanks @unkcpz . I agree that we should make this input optional, since it is a default value after all. But why don't we just make the option truly optional? In that case, a user simply does not provide a value for the `default plugin` prompt. Does this approach not work somehow?

unkcpz:
It already a optional option since the default value of `required` is `required=False`, and it works quite well with setting code in non-interactive way. The problem comes with set code interactively, what you expect users to input? Since it's not possible to just click the `ENTER` and continous for setting the following options.

There are following chooses(but I think with their own downside) to do it without the extra flag I set it here:
1. Allow user to input `None` to not set this when the option prompt out.
2. The default value is `None` and user can type `Enter` and go ahead.
3. use CHARACTER_IGNORE_DEFAULT `!` as  the input, but I guess the `!` only for TEXT option such as `append_text` and `prepend_text`?

Moreover, input_plugin option read a `PluginParamType` which is a subclass of `NonEmptyStringParamType`, so it is not easy to convert `None` or empty string for it.

sphuber:
I think this is rather just a bug in the behavior. The option is required, so a `None` value is perfectly acceptable. Which means that if the user is prompted but presses ENTER, then that should simply be acceptable as setting `None`. I would rather fix that behavior than complicate the interface with another dependent option.

unkcpz:
Thanks!

> I would rather fix that behavior than complicate the interface with another dependent option.

I agree that it is a bad idea to complicate the interface. Thanks for tell me this is a bug, I thought it was designed to being like that, to prevent user from type ENTER without thinking too much etc.
I'll rollback and fix it in the way you recommended.

sphuber:
I am not sure yet why the prompt keeps prompting even if `None` should be accepted, however, it might have to do with our `InteractiveOption`. It hacks `click`'s prompt cycle quite a bit and may have made a mistake. Note that I have an [open PR to upgrade to `click==8.0`](https://github.com/aiidateam/aiida-core/pull/5111) which gets rid of most of the custom logic, since the prompt logic in `click` was largely reworked. Please have a look there as well, since fixing this issue may have to do with the changes there as well.

unkcpz:
Hi @sphuber, I believe the mentioned repeat prompt behavior is by design. I find a test case to check this: https://github.com/aiidateam/aiida-core/blob/b31e5f506a733d3225391b4ef075fc17f7a4492d/tests/cmdline/params/options/test_interactive.py#L148-L158

sphuber:
Good point. Looking into it more closely, I think the behavior is just the default prompting behavior of `click`. Take the following MWE:
```python
@click.command()
@click.option('--label', required=False, type=str, prompt='Label')
def cmd(label):
    click.echo(f'Label: {label}')
```
This will keep prompting for the value each time an empty value is specified, even though it is not required. I think this is a fundamental behavior of `click` but I have the feeling this doesn't make any sense. If an option is not required and `None` would be perfectly fine, why would we keep prompting? I would say it makes a lot more sense to just accept that. The only downside I can see is if someone hits enter on accident, but that is the same as typing something incorrect and hitting enter, in that case you also cannot correct the "mistake" and simply have to reexecute the command. I really feel that this should be possible. What do you think? Do you think empty value for an optional prompted value should just be accepted?

unkcpz:
I have no strong opinion on this. For a new user who may not understand the exact meaning of the opinion, I sometimes want to type ENTER to see what will happened. I kindly understand why click keep on repeat prompt. If we want do it in the way you mentioned, it would be better to add a instruction next to `?` `!` to let the user know that Enter will leave the option not set. 

The other problem is, for example if we set `is_on_computer=True`, can we still set `computer` to None by enter?

unkcpz:
@sphuber Let's close this for now. I made tests for prompt behavior with rebase during the coding week but indeed still got the issue from `click` side.

Issue #4495, opened 2020-10-21
Small changes to `List` and `Dict` data types API


mbercx:
I'd like to discuss two small changes to the API of the `List` and `Dict` classes:

#### Is your feature request related to a problem? Please describe

When initializing a `List` or `Dict` node, you have to pass `dict` or `list` as a keyword argument:

```Python
l = List(list=[1, 2, 3])
```

If you try to instead just pass the list as a positional argument, you are faced with an error that doesn't clarify proper usage:

```Python
l = List([1, 2, 3])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-47-398676db57d5> in <module>
----> 1 List([1, 2, 3])

TypeError: __init__() takes 1 positional argument but 2 were given
```

and the docstring doesn't explain that you have to use the `list` keyword argument either:

```Python
In [46]: List?
Init signature: List(**kwargs)
Docstring:      `Data` sub class to represent a list.
Init docstring:
:param backend_entity: the backend model supporting this entity
:type backend_entity: :class:`aiida.orm.implementation.entities.BackendEntity`
File:           ~/envs/aiida-docs/code/aiida-core/aiida/orm/nodes/data/list.py
Type:           AbstractNodeMeta
Subclasses:    
```

#### Describe the solution you'd like

I think it would be more intuitive if the user could just pass the list/dict as the first positional argument. For example, we could change the current constructor:

```Python
def __init__(self, **kwargs):
    data = kwargs.pop('list', list())
    super().__init__(**kwargs)
    self.set_list(data)
```

to something like (thanks to @giovannipizzi for the suggestion to import the `builtins` module to avoid conflicts between the `list` argument and builtin, see [this discussion](https://github.com/aiidateam/aiida-core/pull/4469#discussion_r508263946)):

```Python
def __init__(self, list=None, **kwargs):
    import builtins
    list = list or builtins.list()
    super().__init__(**kwargs)
    self.set_list(input_list)
```

Similarly for `Dict`. If for some reason we cannot make this change, we should at least add a docstring that clearly explains the API to the user.

#### Is your feature request related to a problem? Please describe

For mose data types in AiiDA derived from builtin Python types, you can use the `value` property  to get the corresponding Python builtin. For `List` and `Dict` nodes, however, you have to use the `get_list()` and `get_dict()` methods. Adding the 
`value` property to these classes would make the API between the "builtin" AiiDA data types more consistent. 

[...]

PR #5142, opened 2021-09-19
`List`: register the class with the `to_aiida_type` serializer

### LINKED ISSUE #5141: Register `list` for the `to_aiida_type` dispatch
This will allow users to pass normal base type lists to an input port accepting a `List` node and defining the serializer `to_aiida_type`.

sphuber:
Fixes #5141 

This will allow users to pass a plain list as input for a port that
accepts `List` nodes and specified the `to_aiida_type` dispatch as
serializer. The value will be automatically converted to a `List` node.
This brings the functionality for `List` on par with the other base type
data node classes.

To make the newly added test work, the `Dict` class needed a change to allow comparing to normal dictionaries. This is added in a separate commit.

### DIFF
--- a/aiida/orm/nodes/data/dict.py
+++ b/aiida/orm/nodes/data/dict.py
@@ -70,6 +70,12 @@ class Dict(Data):
     def __setitem__(self, key, value):
         self.set_attribute(key, value)
 
+    def __eq__(self, other):
+        if isinstance(other, dict):
+            return self.get_dict() == other
+
+        return self is other
+
     def set_dict(self, dictionary):
         """ Replace the current dictionary with another one.
 
--- a/aiida/orm/nodes/data/list.py
+++ b/aiida/orm/nodes/data/list.py
@@ -10,6 +10,7 @@
 """`Data` sub class to represent a list."""
 from collections.abc import MutableSequence
 
+from .base import to_aiida_type
 from .data import Data
 
 __all__ = ('List',)
@@ -139,3 +140,8 @@ class List(Data, MutableSequence):
         :rtype: bool
         """
         return not self.is_stored
+
+
+@to_aiida_type.register(list)
+def _(value):
+    return List(list=value)
--- a/tests/orm/data/test_dict.py
+++ b/tests/orm/data/test_dict.py
@@ -7,58 +7,91 @@
 # For further information on the license, see the LICENSE.txt file        #
 # For further information please visit http://www.aiida.net               #
 ###########################################################################
-"""Tests for the `Dict` class."""
+# pylint: disable=redefined-outer-name
+"""Tests for :class:`aiida.orm.nodes.data.dict.Dict` class."""
+import pytest
 
-from aiida.backends.testbase import AiidaTestCase
 from aiida.orm import Dict
 
 
-class TestDict(AiidaTestCase):
-    """Test for the `Dict` class."""
+@pytest.fixture
+def dictionary():
+    return {'value': 1, 'nested': {'dict': 'ionary'}}
 
-    @classmethod
-    def setUpClass(cls, *args, **kwargs):
-        super().setUpClass(*args, **kwargs)
-        cls.dictionary = {'value': 1, 'nested': {'dict': 'ionary'}}
-        cls.node = Dict(dict=cls.dictionary)
 
-    def test_keys(self):
-        """Test the `keys` method."""
-        self.assertEqual(sorted(self.node.keys()), sorted(self.dictionary.keys()))
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_keys(dictionary):
+    """Test the ``keys`` method."""
+    node = Dict(dict=dictionary)
+    assert sorted(node.keys()) == sorted(dictionary.keys())
 
-    def test_get_dict(self):
-        """Test the `get_dict` method."""
-        self.assertEqual(self.node.get_dict(), self.dictionary)
 
-    def test_dict_property(self):
-        """Test the `dict` property."""
-        self.assertEqual(self.node.dict.value, self.dictionary['value'])
-        self.assertEqual(self.node.dict.nested, self.dictionary['nested'])
+@pytest.mark.usefixtures('clear_database_before_test')
+def test_get_dict(dictionary):
+    """Test the ``get_dict`` method."""
+    node = Dict(dict=dictionary)
+    assert node.get_dict() == dictionary
 
-    def test_get_item(self):
-        """Test the `__getitem__` method."""
-        self.assertEqual(self.node['value'], self.dictionary['value'])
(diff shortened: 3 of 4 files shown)


mbercx:
Nice! Love the fact that you don't have to worry about passing in a `List` anymore. However, won't this change the way that `Dict`s compare? I.e. won't:

```Python
Dict(dict={'a': 1}) == Dict(dict={'a': 1})
```

Now return `True`? I'm all for it, but I do think this behaviour is different from before (also see the discussion in https://github.com/aiidateam/aiida-core/issues/4495).

chrisjsewell (changes requested):
Yeh cheers, but just stick to the list addition for this PR

sphuber:
Forgot about that long open discussion about node equality. Agree that I shouldn't change that here. Still I think that the `assert dict_node == dict_node.value` should be non-controversial, right? I have updated the implementation to allow for this case, but to leave the original behavior when comparing node instances.

mbercx:
> Forgot about that long open discussion about node equality. Agree that I shouldn't change that here. Still I think that the `assert dict_node == dict_node.value` should be non-controversial, right? I have updated the implementation to allow for this case, but to leave the original behavior when comparing node instances.

I'm still in favour of actually making `Dict` nodes the same as `List` nodes (and other `BaseType`s) with regard to comparisons. ^^ Maybe not change that here, but we should try and hash this out before v2.0, perhaps? If we can get everyone to pitch in at the next AiiDA team meeting, I'm happy to open a PR to fix #4495 after, based on whatever we decide.

sphuber:
> I'm still in favour of actually making `Dict` nodes the same as `List` nodes (and other `BaseType`s) with regard to comparisons. ^^ 

I would have to revise the discussion once again and make a summary of the pro's and con's. I am tempted to agree with you but I remember there being strong con's that I cannot remember right now.

> Maybe not change that here, but we should try and hash this out before v2.0, perhaps?

Whether we change it or not, I think it would be good either way to come to a decision. So indeed good to try and make an executive summary and get a poll going with a final decision.

mbercx:
> Whether we change it or not, I think it would be good either way to come to a decision. So indeed good to try and make an executive summary and get a poll going with a final decision.

Alright, I'll try to go through the whole discussion again and write a summary by Wednesday. Then we can schedule a small meeting after the group meeting with everyone interested in discussing this (probably won't have enough time during the AiiDA meeting).

chrisjsewell:
Yeh I’m for whatever makes the user experience best, which is most likely comparing base types as there value. Just a pain, because there will probably be no way to deprecate in a nice way, and it will likely break someone’s code/plugin who has been relying on previous behaviour. I think perhaps a good way would be like sqlalchemy has done, you can add an environmental variable to turn on certain warnings. We could have this turn on a warning for any code that calls these `__eq__` methods, so people can quickly identify what code they need to change, then turn it off again

chrisjsewell:
(and this warning could be backported to v1

sphuber:
> Alright, I'll try to go through the whole discussion again and write a summary by Wednesday. Then we can schedule a small meeting after the group meeting with everyone interested in discussing this (probably won't have enough time during the AiiDA meeting).

Would be good, if you manage in time, to send the summary by email, such that those who are interested can have a look before. It is not really possible to do this during the meeting. Not that we should have the discussion during the meeting, but it can be helpful for people to have the context beforehand.

sphuber:
@chrisjsewell are there any objections left after I changed the `__eq__` implementation to not change comparison of nodes but only with plain dicts?

@mbercx would also be good to make sure #4495 gets in soon, since we likely also want to add this in 2.0 now that we are on a campaign to make all base types equal. Did you want to do this yourself, or can I also go ahead and implement it?

chrisjsewell:
Will have a double check soon, bare with me, I’m just in the middle of Croatia waiting for a bus lol

mbercx:
> @mbercx would also be good to make sure #4495 gets in soon, since we likely also want to add this in 2.0 now that we are on a campaign to make all base types equal. Did you want to do this yourself, or can I also go ahead and implement it?

I think I should find time to do this on Friday, or next week at the latest. Note that we should probably first agree on what the equality comparison between base types should be though, because making `List` and `Dict` `BaseType`s changes the equality comparison behavior for `Dict`.

sphuber:
@mbercx would addressing some of the feature requests really change the equality behavior? For example, adding the `value` property and changing the signature of the constructor wouldn't affect it, would it? Worst come to worst, we can always add these changes and leave the equality question for what it is if we cannot come to an agreement. In any case, we should probably change these things in separate commits if not also separate PRs.

sphuber:
> Will have a double check soon, bare with me, I’m just in the middle of Croatia waiting for a bus lol

pinging @chrisjsewell

chrisjsewell:
yep will do after lunch

chrisjsewell (approved):
Errr, I still don't feel the equality change needs to be present in this PR 😬
But anyhow, it's very likely it will be changed again before v2 releases, so I'll leave it up to you 🤷

sphuber:
> Errr, I still don't feel the equality change needs to be present in this PR grimacing

Well I guess it is not strictly necessary, but I don't think it hurts. And it allows the tests in the follow-up commit to use the same assert statement `assert converted == value` in the parametrized test. Without this addition, I would have to resort to making an exceptional case for the dictionary test.

chrisjsewell:
Oh I see, in `BaseType` there is already:

```python
    def __eq__(self, other):
        if isinstance(other, BaseType):
            return self.value == other.value
        return self.value == other
```

thats why `Dict` was previously different @sphuber 

This then raises the question, should `Dict` also inherit from `BaseType`?
Or, at a minimum, we should also implement the `__ne__` method on `Dict` as well

sphuber:
The reason I added the `__eq__` is indeed exactly because `Dict` doesn't inherit from `BaseType`. I didn't want to change the base class for `Dict` as that could have had even further consequences to the behavior and I explicitly wanted the change to be minimal and just have the node content compare equal to the base type of its value. I added the `self is other` as the fall-through comparison as that is the base implementation of `object`. However, I realize now, that I probably should have called `super().__eq__(self, other)` since `Node` defines a fall-through based on UUID. We should correct this therefore I think.

Then on whether `__ne__` should be specified: I don't think this is necessary. The [docs specify that](https://docs.python.org/3/reference/datamodel.html#object.__ne__) `__ne__` is automatically defined as the inverse of `__eq__`.

mbercx:
> @mbercx would addressing some of the feature requests really change the equality behavior? For example, adding the `value` property and changing the signature of the constructor wouldn't affect it, would it? Worst come to worst, we can always add these changes and leave the equality question for what it is if we cannot come to an agreement. In any case, we should probably change these things in separate commits if not also separate PRs.

Sorry, missed this reply! Will work on adding the `value` property and changing the constructor today.

PR #5148, opened 2021-09-23
IMPROVE: Add database name to `verdi status`

ramirezfranciscof:
/title

I think this is useful information that was missing from the report printed by the `verdi status` command.


chrisjsewell (dismissed):
Looks good to me thanks

ramirezfranciscof:
Hey @chrisjsewell , thanks for the review! Sorry but I needed to add a test to appease codecov 😅  could I ask you for another one?

(for reference the test is almost identical to the previous one just above it, `test_database_incompatible`)

chrisjsewell (approved):
👍

PR #5149, opened 2021-09-23
BaseRestortWorkChain: clearify the log report when output port not compatible

unkcpz:
The output ports required are reported to missing in two conditions. 
First, when the nested `CalcJob` do not correctly giving the output. Second, when the base workchain does not correctly expose the output port, such as what happened in https://github.com/aiidateam/aiida-quantumespresso/issues/723. 

These should be two different scenario but fine to be report in same conditional branch. In this PR I add the second condition which is not considered previously to give a more comprehensible report logs.


ramirezfranciscof:
> Hi @ramirezfranciscof , I rebase commits to the develop branch. And it dismiss your review. Did I do anything wrong?

Ah, yeah this happens when you re-base, it automatically dismisses all reviews. No worries, update it (however you prefer) and ping me so I can approve it again.

PR #5150, opened 2021-09-23
👌 IMPROVE: Add `Dict.__ne__`

chrisjsewell:
See https://github.com/aiidateam/aiida-core/pull/5142#issuecomment-925724779

@sphuber still not sure what the "default" comparison should be; I you put `self is other` for `__eq__`, was there a reason for this?


sphuber:
Closing this in favor of #5159

PR #5151, opened 2021-09-23
🔧 MAINTAIN: Add isort pre-commit hook

chrisjsewell:
To nicely sort imports: https://pycqa.github.io/isort/

~~I've set it as "opt-in" (via file regex) for now, so we can introduce it gradually, without introducing too many merge conflicts in existing PRs~~

Annoyingly, there is no specific integration with yapf (as there is for black), but from googling the configuration I added should be fine.
I added `force_sort_within_sections=true` to not split `import a` and `from a import` types, which I feel is nicer

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,10 +24,6 @@ repos:
     rev: 5.9.3
     hooks:
     - id: isort
-      files: >-
-         (?x)^(
-            aiida/common/.*py|
-         )$
 
 -   repo: https://github.com/ikamensh/flynt/
     rev: '0.66'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
          (?x)^(
             aiida/tools/groups/paths.py
          )$
-            
+
 -   repo: https://github.com/ikamensh/flynt/
     rev: '0.66'
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,10 +24,10 @@ repos:
     rev: 5.9.3
     hooks:
     - id: isort
-       files: >-
-          (?x)^(
+      files: >-
+         (?x)^(
             aiida/tools/groups/paths.py
-          )$
+         )$
             
 -   repo: https://github.com/ikamensh/flynt/
     rev: '0.66'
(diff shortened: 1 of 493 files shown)


pyproject.toml
```diff
@@ -75,6 +75,13 @@ markers = [
     "sphinx: set parameters for the sphinx `app` fixture"
 ]
 
+[tool.isort]
+line_length = 100
```
sphuber, line 79: Why not 120 as the rest of the code?

.pre-commit-config.yaml
```diff
@@ -20,6 +20,15 @@ repos:
...
 
 
+-   repo: https://github.com/pycqa/isort
+    rev: 5.9.3
+    hooks:
+    - id: isort
+      files: >-
+         (?x)^(
+            aiida/common/.*py|
+         )$
```
sphuber, line 30: I can appreciate that doing everything now would cause merge conflicts with open PRs, but I fear that in this way it is never going to get done. If we are going to add this, I prefer we bite the bullet. Since it only concerns imports, the merge conflicts should not be complicated.

pyproject.toml
chrisjsewell: Fine by me

.pre-commit-config.yaml
chrisjsewell: Sure ting, more lines of code attributed to me lol

pyproject.toml
chrisjsewell, line 79:
```suggestion
line_length = 120
```

.pre-commit-config.yaml
chrisjsewell, line 30:
```suggestion
```

chrisjsewell:
right, removed the restriction on files; lets see if pre-commit.ci does its thing!

aiida/backends/djsite/db/migrations/__init__.py
```diff
@@ -64,6 +68,7 @@ def _upgrade_schema_generation(version, apps, _):
 
 def upgrade_schema_version(up_revision, down_revision):
     from functools import partial
```
chrisjsewell, line 70: weird, only because of the extra new line, does pylint start failing it for not having a docstring 🤷 (https://github.com/aiidateam/aiida-core/runs/3690148576)

```suggestion
def upgrade_schema_version(up_revision, down_revision):
    """Run migrations, to translate the database schema."""
    from functools import partial
```

chrisjsewell:
Ok @sphuber I think this is good to go

sphuber (approved):
Looking good, thanks @chrisjsewell

PR #5153, opened 2021-09-23
Updated LSF scheduler to accept number of nodes

nkeilbart:
I have the scheduler on Lassen at LLNL which is not able to accept the current settings. I have updated the scheduler so that if the option:

use_num_machines : True

is set then you can specify both num_machines and tot_num_mpiprocs as well.

### DIFF
--- a/aiida/schedulers/plugins/lsf.py
+++ b/aiida/schedulers/plugins/lsf.py
@@ -99,11 +99,7 @@ class LsfJobResource(JobResource):
     See https://www-01.ibm.com/support/knowledgecenter/SSETD4_9.1.2/lsf_command_ref/bsub.1.dita?lang=en
     for more details about the parallel environment definition (the -m option of bsub).
     """
-    _default_fields = (
-        'parallel_env',
-        'tot_num_mpiprocs',
-        'default_mpiprocs_per_machine',
-    )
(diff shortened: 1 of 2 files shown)


chrisjsewell:
Thanks @nkeilbart!
Obviously, will have to look at why this is now failing the tests, but off-hand I think this seems reasonable

nkeilbart:
No problem at all. Let me know if there is some documentation somewhere that might need updated as well. Otherwise people might not know that option exists without looking at the source code. I'm still a bit of a novice with all the git repository setup but if you point me in the right direction I'll work on that as well.

sphuber:
Thanks @nkeilbart . The first question that sprang to my mind is: why did the original implementer say that `num_machines` is not supported? Do all LSF schedulers support defining the number of nodes, or might this be something specific to the installation/configuration at LLNL? If it is supported by default, then it would indeed make sense to add here. Once we answer that question, I have some small comments on the implementation, but I wanted to first get this question out of the way. Because if it is not general to all LSF schedulers, then creating a LLNL specific plugin might be better. This would be very easy to do through the plugin system and would prevent breaking the `LsfScheduler` for other users.

sphuber:
@nkeilbart I took the liberty to fix the tests and add additional ones for the added functionality. Also cleaned the code up a bit for readability, hope you don't mind. Let me know if you are still happy with this as is, and then I will accept and merge it.

sphuber (approved):
Thanks for the contribution @nkeilbart

PR #5156, opened 2021-09-29
ADD: Repository methods for repo CLI and other features

ramirezfranciscof:
Methods added and modifications:

* Added has_objects as new abstract and adapted has_object
* Added delete_objects as new abstract and adapted delete_object
* Added list_objects method
* While modifying the inherited classes I removed some of the docstrings
  since these should be anyways inherited from the parent class and were
  just adding visual noise and maintenance costs.

### DIFF
--- a/aiida/repository/backend/abstract.py
+++ b/aiida/repository/backend/abstract.py
@@ -10,8 +10,7 @@ import contextlib
 import hashlib
 import io
 import pathlib
-
-from typing import Union, Optional, Iterator, BinaryIO, List, Iterable
+from typing import BinaryIO, Iterable, Iterator, List, Optional, Union
 
 from aiida.common.hashing import chunked_file_hash
 
--- a/aiida/repository/backend/abstract.py
+++ b/aiida/repository/backend/abstract.py
@@ -33,6 +33,16 @@ class AbstractRepositoryBackend(metaclass=abc.ABCMeta):
     def uuid(self) -> Optional[str]:
         """Return the unique identifier of the repository."""
 
+    @property
+    @abc.abstractmethod
+    def key_format(self) -> Optional[str]:
+        """Return the format for the keys of the repository.
+
+        Important for when migrating between backends (e.g. archive -> main), as if they are not equal then it is
+        necessary to re-compute all the `Node.repository_metadata` before importing (otherwise they will not match
+        with the repository).
+        """
+
     @abc.abstractmethod
     def initialise(self, **kwargs) -> None:
         """Initialise the repository if it hasn't already been initialised.
@@ -104,7 +114,7 @@ class AbstractRepositoryBackend(metaclass=abc.ABCMeta):
 
     @abc.abstractmethod
     def list_objects(self) -> Iterable[str]:
-        """Return iterable that yeilds all available objects by key.
+        """Return iterable that yields all available objects by key.
 
         :return: An iterable for all the available object keys.
         """
--- a/aiida/repository/backend/abstract.py
+++ b/aiida/repository/backend/abstract.py
@@ -10,7 +10,8 @@ import contextlib
 import hashlib
 import io
 import pathlib
-import typing
+
+from typing import Union, Optional, Iterator, BinaryIO, List, Iterable
 
 from aiida.common.hashing import chunked_file_hash
 
@@ -30,7 +31,7 @@ class AbstractRepositoryBackend(metaclass=abc.ABCMeta):
 
     @property
     @abc.abstractmethod
-    def uuid(self) -> typing.Optional[str]:
+    def uuid(self) -> Optional[str]:
         """Return the unique identifier of the repository."""
 
     @abc.abstractmethod
@@ -58,7 +59,7 @@ class AbstractRepositoryBackend(metaclass=abc.ABCMeta):
     def is_readable_byte_stream(handle) -> bool:
         return hasattr(handle, 'read') and hasattr(handle, 'mode') and 'b' in handle.mode
 
-    def put_object_from_filelike(self, handle: typing.BinaryIO) -> str:
+    def put_object_from_filelike(self, handle: BinaryIO) -> str:
         """Store the byte contents of a file in the repository.
 
         :param handle: filelike object with the byte content to be stored.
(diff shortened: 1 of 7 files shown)


ramirezfranciscof:
Created from #4965, @chrisjsewell I modified the type returned so let me know if this now works for you too or there is something else that would need adapting. In the meantime I'll try to add a couple of tests for these new features.

aiida/repository/backend/disk_object_store.py
```diff
@@ -65,16 +65,11 @@ def _put_object_from_filelike(self, handle: typing.BinaryIO) -> str:
         """
         return self.container.add_object(handle.read())
```
chrisjsewell, line 66:
```suggestion
        return self.container.add_streamed_object(handle)
```

Files should be read/written in chunks, so that very large files never fill the RAM

chrisjsewell (changes requested):

chrisjsewell:
Yep cheers, another thing that from #5145 is that I think there should be a `key_format` property, e.g.

for the `DiskObjectStoreRepositoryBackend` it would be:

```python
@property
def key_format(self):
	return self.container.hash_type
```

and for `SandboxRepositoryBackend` it would be:

```python
@property
def key_format(self):
	return 'uuid4'
```

The reason this is necessary is that, when migrating between backends (e.g. archive -> postgresql), one needs to know if the `key_format` are equal. If not, then you would need to re-compute all the `Node.repository_metadata` before importing them, otherwise they will not match with the repository.

aiida/repository/backend/abstract.py
```diff
@@ -84,15 +84,33 @@ def put_object_from_file(self, filepath: typing.Union[str, pathlib.Path]) -> str
...
         """Return whether the repository has an object with the given key.
 
         :param key: fully qualified identifier for the object within the repository.
         :return: True if the object exists, False otherwise.
         """
+        return self.has_objects([key])[0]
+
+    @abc.abstractmethod
+    def list_objects(self) -> Iterable[str]:
+        """Return iterable that yeilds all available objects by key.
```
chrisjsewell, line 107:
```suggestion
        """Return iterable that yields all available objects by key.
```

chrisjsewell: If you use VS Code, I certainly recommend https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker 😉

aiida/repository/backend/abstract.py
```diff
@@ -84,15 +84,33 @@ def put_object_from_file(self, filepath: typing.Union[str, pathlib.Path]) -> str
...
+
+        :return: An iterable for all the available object keys.
+        """
 
     @contextlib.contextmanager
-    def open(self, key: str) -> typing.Iterator[typing.BinaryIO]:
+    def open(self, key: str) -> Iterator[BinaryIO]:
```
chrisjsewell, line 113:
```suggestion
    def open(self, key: str) -> ContextManager[BinaryIO]:
```

A bit more specific: https://docs.python.org/3/library/typing.html#typing.ContextManager (I only noticed this was available myself recently)

chrisjsewell: actually scratch that, I think this is only for without the contextmanager decorator (at least mypy/pylint are raising a lot of errors

aiida/repository/backend/abstract.py
ramirezfranciscof: If I'm developing inside a container, will this extension automatically install inside of it when I attach through the remote explorer or will I have to manually re-install it every time I shutdown and re-start the container?

chrisjsewell: I know with the new vs code devcontainers, you can specify which extensions to load, see e.g. https://github.com/chrisjsewell/aiida-install/blob/56fe6d7398594c8813890652af99c521a77f249c/.devcontainer/devcontainer.json#L41

ramirezfranciscof:
I added the tests I could think of so this is now ready for another round of review. I also added the `key_format` property: if I understood correctly you only need the property defined here, not any feature using it, correct @chrisjsewell ?

chrisjsewell (approved):
All good thanks @ramirezfranciscof 👌

aiida/repository/backend/abstract.py
ramirezfranciscof: When you are using these you need to create the container through the vscode right? I'm not sure this would work for me because I need to initialize mine using docker-compose (I have 3 containers running for this: rabbitmq, postgres, and the work container).

chrisjsewell: I was actually playing around with that a bit but found, at least on my macbook, it was a bit slow a prone to the container dying:

devcontainer.json

```json
// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
// https://github.com/microsoft/vscode-dev-containers/tree/v0.187.0/containers/docker-existing-dockerfile
// https://github.com/microsoft/vscode-dev-containers/tree/main/containers/python-3-postgres
{
    "name": "AiiDA development environment",
    "dockerComposeFile": "docker-compose.yml",
    "service": "main",
    "workspaceFolder": "/workspace",

	// Set *default* container specific settings.json values on container create.
	"settings": {
        "editor.rulers": [100],
        "python.pythonPath": "/usr/local/bin/python",
		"python.defaultInterpreterPath": "/usr/local/bin/python",
        "python.languageServer": "Pylance",
		"python.linting.enabled": true,
		"python.linting.pylintEnabled": true,
        "python.linting.pylintPath": "/usr/local/py-utils/bin/pylint",
        "python.formatting.provider": "yapf",
        "python.formatting.yapfPath": "/usr/local/py-utils/bin/yapf",
    },
	
	// Add the IDs of extensions you want installed when the container is created.
	"extensions": [
		"ms-python.python",
		"ms-python.vscode-pylance",
		"ms-azuretools.vscode-docker",
		"streetsidesoftware.code-spell-checker",
		"chrisjsewell.aiida-explore-vscode"
	],
}
```

```yaml
version: '3.4'

services:

  main:
    image: 'mcr.microsoft.com/vscode/devcontainers/python:3.8'
    volumes:
      - ..:/workspace:cached
    # Overrides default command so things don't shut down after the process ends.
    command: sleep infinity

  postgres:
    image: postgres:12.3
    environment:
      POSTGRES_USER: pguser
      POSTGRES_PASSWORD: password
    restart: unless-stopped

  rmq:
    image: rabbitmq:3.8.3-management
    environment:
      RABBITMQ_DEFAULT_USER: guest
      RABBITMQ_DEFAULT_PASS: guest
    restart: unless-stopped
```

ramirezfranciscof: Uhm, very interesting! Thanks, I will give this a try =D

PR #5157, opened 2021-10-03
📚 DOCS: Add example for `retrieve_temporary_list`

### LINKED ISSUE #5143: Docs: complete section on `retrieve_temporary_list`
It explains [how to define what files to retrieve temporarily](https://aiida.readthedocs.io/projects/aiida-core/en/latest/topics/calculations/usage.html#retrieve-temporary-list) but then doesn't say how to operate on those files in the `Parser.parse` method.

mbercx:
Fixes #5143 

The current section on how the user can retrieve files temporarily to use
during parsing doesn't contain any examples on how to use this feature. Here
we add an example on how to specify the list of files to be retrieved in a
temporary folder, clearly showing the `retrieve_temporary_list` attribute of
the `CalcInfo`. We also add an example of how to obtain the
`retrieved_temporary_folder` from the `kwargs` of the `Parser.parse()` method.

### DIFF
--- a/docs/source/topics/calculations/usage.rst
+++ b/docs/source/topics/calculations/usage.rst
@@ -518,17 +518,32 @@ The target relative path is also compatible with glob patterns in the source rel
 
 Retrieve temporary list
 ~~~~~~~~~~~~~~~~~~~~~~~
+
 Recall that, as explained in the :ref:`'prepare' section<topics:calculations:usage:calcjobs:prepare>`, all the files that are retrieved by the engine following the 'retrieve list', are stored in the ``retrieved`` folder data node.
 This means that any file you retrieve for a completed calculation job will be stored in your repository.
 If you are retrieving big files, this can cause your repository to grow significantly.
 Often, however, you might only need a part of the information contained in these retrieved files.
 To solve this common issue, there is the concept of the 'retrieve temporary list'.
-The specification of the retrieve temporary list is identical to that of the normal :ref:`retrieve list<topics:calculations:usage:calcjobs:file_lists_retrieve>`.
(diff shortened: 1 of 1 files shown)


docs/source/topics/calculations/usage.rst
```diff
@@ -518,17 +518,32 @@ The target relative path is also compatible with glob patterns in the source rel
...
 The only difference is that, unlike the files of the retrieve list which will be permanently stored in the retrieved :py:class:`~aiida.orm.nodes.data.folder.FolderData` node, the files of the retrieve temporary list will be stored in a temporary sandbox folder.
-This folder is then passed to the :ref:`parser<topics:calculations:usage:calcjobs:parsers>`, if one was specified for the calculation job.
+This folder is then passed under the ``retrieved_temporary_folder`` keyword argument to the ``parse`` method of the :ref:`parser<topics:calculations:usage:calcjobs:parsers>`, if one was specified for the calculation job:
+
+.. code-block:: python
+
+    def parse(self, **kwargs):
+        """Parse the retrieved files of the calculation job."""
+
+        retrieved_temporary_folder = kwargs['retrieved_temporary_folder']
```
sphuber, line 542: The only thing we should probably add here is what the type of `kwargs['retrieved_temporary_folder']` is. It might not be immediately obvious that this is just a `str` (should we make it `pathlib.Path`) that represents the absolute filepath to the directory and so can be used with the standard library module `os` to inspect its contents etc.

sphuber (changes requested):
Thanks @mbercx just a small suggestion to add and then this can be merged

mbercx:
Thanks for the review, @sphuber! I've added an "important" note on the type of the `kwargs['retrieved_temporary_folder']`. Re the suggestion of making this a `pathlib.Path` instead, I agree 👍 .

mbercx:
Link to the corresponding section:

https://4639-77234579-gh.circle-artifacts.com/0/html/topics/calculations/usage.html#retrieve-temporary-list

sphuber (approved):

PR #5159, opened 2021-10-04
`Dict`: fix the `__eq__` implementation to call super

sphuber:
Fixes the mistake merged in #5142 and this supersedes #5150 

The `__eq__` method was recently implemented for the `Dict` data type
but it erroneously did not call through to the super implementation in
case the dictionary content does not match that of `other`. This caused
the fallback equality implemented on the `Node` base class, which
ensures that two nodes with the same UUID compare equal, to be missed.


ramirezfranciscof:
> I think you may have misread.

Indeed, haha. Sorry, I was trying to get quick review before something else I had to do I should have read more carefully!

PR #5160, opened 2021-10-05
Fix issue with time in lsf and gaussian plugins

frssp:
This fixes an error that occurs when aiida.gaussian passes the output from '_parse_log_cclib' 
It contains a datetime.timedelta instance which is not json-serializable

Set `LSB_DISPLAY_YEAR=Y` before `bjobs` to display years. 
No workaround (`actual_fmt` and `actual_string`) required.

### DIFF
--- a/aiida/orm/implementation/utils.py
+++ b/aiida/orm/implementation/utils.py
@@ -9,10 +9,10 @@
 ###########################################################################
 """Utility methods for backend non-specific implementations."""
 from collections.abc import Iterable, Mapping
+import datetime
 from decimal import Decimal
 import math
 import numbers
-import datetime
 
 from aiida.common import exceptions
 from aiida.common.constants import AIIDA_FLOAT_PRECISION
@@ -76,7 +76,7 @@ def clean_value(value):
             # see https://www.postgresql.org/docs/current/static/datatype-json.html#JSON-TYPE-MAPPING-TABLE
             raise exceptions.ValidationError('nan and inf/-inf can not be serialized to the database')
 
-        # This fixes an error that occurs when aiida.gaussian passes the output from '_parse_log_cclib' 
+        # This fixes an error that occurs when aiida.gaussian passes the output from '_parse_log_cclib'
         # It contains a datetime.timedelta instance which is not json-serializable
         if isinstance(val, (datetime.timedelta)):
             return int(val.total_seconds())
--- a/aiida/orm/implementation/utils.py
+++ b/aiida/orm/implementation/utils.py
@@ -12,6 +12,7 @@ from collections.abc import Iterable, Mapping
 from decimal import Decimal
 import math
 import numbers
+import datetime
 
 from aiida.common import exceptions
 from aiida.common.constants import AIIDA_FLOAT_PRECISION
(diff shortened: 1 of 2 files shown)


aiida/orm/implementation/utils.py
```diff
@@ -75,6 +76,11 @@ def clean_builtin(val):
             # see https://www.postgresql.org/docs/current/static/datatype-json.html#JSON-TYPE-MAPPING-TABLE
             raise exceptions.ValidationError('nan and inf/-inf can not be serialized to the database')
 
+        # This fixes an error that occurs when aiida.gaussian passes the output from '_parse_log_cclib'
+        # It contains a datetime.timedelta instance which is not json-serializable
+        if isinstance(val, (datetime.timedelta)):
+            return int(val.total_seconds())
```
sphuber, line 82: This is a significant change. I am wondering if the `aiida-gaussian` plugin should simply make sure to remove the datetime, or do the conversion themselves. The reason is that if we do it here, it is a but hidden that the type is changing. If you retrieve the attribute after it has been stored, you won't get the timedelta object that you passed but get an int back.

aiida/schedulers/plugins/lsf.py
```diff
@@ -264,9 +264,11 @@ def _get_joblist_command(self, jobs=None, user=None):
         """
         from aiida.common.exceptions import FeatureNotAvailable
 
-        # I add the environment variable SLURM_TIME_FORMAT in front to be
-        # sure to get the times in 'standard' format
-        command = ['bjobs', '-noheader', f"-o '{' '.join(self._joblist_fields)} delimiter=\"{_FIELD_SEPARATOR}\"'"]
+        # Force to display year
+        command = [
+            "LSB_DISPLAY_YEAR='Y'", 'bjobs', '-noheader',
```
sphuber, line 269: Is this `LSB_DISPLAY_YEAR` setting available for all LSF scheduler versions and installations? If the answer is yes, than we can add it. If not, then I am not sure, because then calculations of people with a scheduler that doesn't support it will start failing.

aiida/schedulers/plugins/lsf.py
```diff
@@ -599,9 +601,9 @@ def _parse_joblist_output(self, retval, stdout, stderr):
 
             this_job.queue_name = partition
 
-            psd_finish_time = self._parse_time_string(finish_time, fmt='%b %d %H:%M')
-            psd_start_time = self._parse_time_string(start_time, fmt='%b %d %H:%M')
-            psd_submission_time = self._parse_time_string(submission_time, fmt='%b %d %H:%M')
+            psd_finish_time = self._parse_time_string(finish_time, fmt='%b %d %H:%M:%S %Y')
+            psd_start_time = self._parse_time_string(start_time, fmt='%b %d %H:%M:%S %Y')
+            psd_submission_time = self._parse_time_string(submission_time, fmt='%b %d %H:%M:%S %Y')
```
sphuber, line 606: Why are you also adding the seconds in the format here? You are only expecting the year to be added no?

sphuber (commented):
Thanks for the PR @frssp . I have some questions about the general applicability of the proposed changes to all LSF schedulers. Those questions also relate directly to why the tests are failing. The tests were built to make sure that datetime strings without the year also get parsed, which after your changes is no longer necessarily, the case.

sphuber:
I am closing due to inactivity. If you want to address the review comments, please feel free to reopen

PR #5161, opened 2021-10-05
Docker: Default to number of physical cores for localhost code.

### LINKED ISSUE #5117: The localhost computer is hard-coded to have only one MPI core
https://github.com/aiidateam/aiida-core/blob/75e7024ecbbb24b2d4b04d744989057d0b925930/.docker/opt/configure-aiida.sh#L45

This line leads any hosts that runs this container to assume that it has only one MPI proc available which is not necessarily true. I recently had to monkey-patch the localhost computer metadata to work-around that constraint in the context of running AiiDAlab with multiple cores.

Is there a specific reason that this parameter is set like this? Pinging @yakutovicha and @sphuber for input.

csadorf:
Adjust how the "localhost" code is configured on the aiida-core Docker
image. The default value is based on the actual number of physical cores
and is further configurable via an environment variable.

Closes #5117.

### DIFF
--- a/.docker/opt/configure-aiida.sh
+++ b/.docker/opt/configure-aiida.sh
@@ -33,7 +33,23 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
 
     # Setup and configure local computer.
     computer_name=localhost
-    cpu_count=`python -c 'import psutil; print(psutil.cpu_count(logical=False))'`
+
+    # Determine the number of physical cores as a default for the number of
+    # available MPI ranks on the localhost. We do not count "logical" cores,
+    # since MPI parallelization over hyper-threaded cores is typically
+    # associated with a significant performance penalty. We use the
+    # `psutil.cpu_count(logical=False)` function as opposed to simply
+    # `os.cpu_count()` since the latter would include hyperthreaded (logical
+    # cores).
+    NUM_PHYSICAL_CORES=$(python -c 'import psutil; print(int(psutil.cpu_count(logical=False)))' 2>/dev/null)
+    LOCALHOST_MPI_PROCS_PER_MACHINE=${MPI_PROCS_PER_MACHINE:-${NUM_PHYSICAL_CORES}}
+
+    if [ -z $LOCALHOST_MPI_PROCS_PER_MACHINE ]; then
+      echo "Unable to automatically determine the number of logical CPUs on this "
+      echo "machine. Please set the LOCALHOST_MPI_PROCS_PER_MACHINE variable to "
+      echo "explicitly set the number of available MPI ranks."
+      exit 1
+    fi
 
     verdi computer show ${computer_name} || verdi computer setup   \
         --non-interactive                                          \
@@ -44,7 +60,7 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
         --scheduler core.direct                                    \
         --work-dir /home/aiida/aiida_run/                          \
         --mpirun-command "mpirun -np {tot_num_mpiprocs}"           \
-        --mpiprocs-per-machine ${MPI_PROCS_PER_MACHINE:-${cpu_count}} && \
+        --mpiprocs-per-machine ${LOCALHOST_MPI_PROCS_PER_MACHINE} && \
     verdi computer configure core.local "${computer_name}"              \
         --non-interactive                                          \
         --safe-interval 0.0
--- a/.docker/opt/configure-aiida.sh
+++ b/.docker/opt/configure-aiida.sh
@@ -42,7 +42,7 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
     # `os.cpu_count()` since the latter would include hyperthreaded (logical
     # cores).
     NUM_PHYSICAL_CORES=$(python -c 'import psutil; print(int(psutil.cpu_count(logical=False)))' 2>/dev/null)
-    LOCALHOST_MPI_PROCS_PER_MACHINE=${MPI_PROCS_PER_MACHINE:-${NUM_PHYSICAL_CORES}}
+    LOCALHOST_MPI_PROCS_PER_MACHINE=${LOCALHOST_MPI_PROCS_PER_MACHINE:-${NUM_PHYSICAL_CORES}}
 
     if [ -z $LOCALHOST_MPI_PROCS_PER_MACHINE ]; then
       echo "Unable to automatically determine the number of logical CPUs on this "
--- a/.docker/opt/configure-aiida.sh
+++ b/.docker/opt/configure-aiida.sh
@@ -33,6 +33,8 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
 
     # Setup and configure local computer.
     computer_name=localhost
+    cpu_count=`python -c 'import psutil; print(psutil.cpu_count(logical=False))'`
+
     verdi computer show ${computer_name} || verdi computer setup   \
         --non-interactive                                          \
         --label "${computer_name}"                                 \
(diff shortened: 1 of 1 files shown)


.docker/opt/configure-aiida.sh
```diff
@@ -33,6 +33,8 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
 
     # Setup and configure local computer.
     computer_name=localhost
+    cpu_count=`python -c 'import psutil; print(psutil.cpu_count(logical=False))'`
```
sphuber, line 36: should we add a check here that the returned value is a valid integer and otherwise unset the variable?

.docker/opt/configure-aiida.sh
```diff
@@ -42,7 +44,7 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
         --scheduler core.direct                                    \
         --work-dir /home/aiida/aiida_run/                          \
         --mpirun-command "mpirun -np {tot_num_mpiprocs}"           \
-        --mpiprocs-per-machine 1 &&                                \
+        --mpiprocs-per-machine ${MPI_PROCS_PER_MACHINE:-${cpu_count}} && \
```
sphuber, line 47: Maybe rename `MPI_PROCS_PER_MACHINE` to `LOCALHOST_MPI_PROCS_PER_MACHINE` to make it more specific, just in case there ever is another machine to be configured

.docker/opt/configure-aiida.sh
sphuber: Is `psutil` also guaranteed to be installed at this point? We currently have it in the `install_requires` of `aiida-core`, which I think answers that question in the positive, given that to run this script, AiiDA has to be installed. Might just pose a problem if we ever remove the last use of psutil in the actual module code of `aiida-core` and then it gets removed, because this use here is easily overlooked. Is there a specific reason to use `psutil` over `os.cpu_count()`?

sphuber (commented):
Thanks @csadorf , some comments below

.docker/opt/configure-aiida.sh
csadorf: > should we add a check here that the returned value is a valid integer and otherwise unset the variable?

I believe the easiest way to protect against undefined behavior here would be to simply wrap the output like this:
```suggestion
    cpu_count=`python -c 'import psutil; print(int(psutil.cpu_count(logical=False)))'`
```
In this way the script will fail if the count cannot be determined.

> Is `psutil` also guaranteed to be installed at this point? We currently have it in the `install_requires` of `aiida-core`, which I think answers that question in the positive, given that to run this script, AiiDA has to be installed. Might just pose a problem if we ever remove the last use of psutil in the actual module code of `aiida-core` and then it gets removed, because this use here is easily overlooked. Is there a specific reason to use `psutil` over `os.cpu_count()`?

If `psutil` is ever removed as an aiida-core dependency, then the build of this image will simply fail, which would directly alert us to the problem. We could add a `pip install psutil~=5.6` command directly in this script if you prefer. We just need to make sure that this would not in any way create conflicts with the dependency specification of aiida-core.

We use `psutil` due to the reasons outlined here: https://github.com/aiidateam/aiida-core/issues/5117#issuecomment-913580903 .

.docker/opt/configure-aiida.sh
csadorf: That seems like a premature generalization to me.

.docker/opt/configure-aiida.sh
sphuber: > In this way the script will fail if the count cannot be determined.

Are you sure? I think the exception will simply be written to `stderr`, the `cpu_count` variable will be empty, and the script will continue. Take this MWE:
```bash
#!/usr/bin/env bash
set -x
cpu_count=`python -c 'import psutil; print(non_existing.cpu_count(logical=False))'`
echo 'CPU_COUNT=' $cpu_count
```
Note that I replaced `psutil` with `non_existing` and so it will except but still get to the next line:
```
++ python -c 'import psutil; print(pstil.cpu_count(logical=False))'
  File "<string>", line 1, in <module>
NameError: name 'pstil' is not defined
+ cpu_count=
+ echo CPU_COUNT=
CPU_COUNT=
```
You could set the `set -e` bash flag to automatically exit the script when a non-zero exit status is returned from any command.

> We could add a pip install psutil~=5.6 command directly in this script if you prefer.

Don't think that is a good idea. I think the script should simply fail if it cannot properly determine the value and not continue silently. Even better, if the env variable is defined (which would override it  anyway) maybe it is best to simply log the error as a warning and then continue anyway.

On a side-note: when is this script being run? When the image is built or when it is being run? That is to say, will this get the number of CPUs of the localhost that is building the image or the one that is running. I hope that it is the latter, because if not it would most likely still get the wrong value configured, would it not?

> We use psutil due to the reasons outlined here

Gotcha. Would have been useful to include this info in a comment, and/or at the very least in the commit message stating that you need `psutil`'s version because `os.cpu_count` does not give merely the physical CPUs, which is what is required here.

.docker/opt/configure-aiida.sh
sphuber: It is not so much just generalization, but mostly for clarity by being more explicit.

.docker/opt/configure-aiida.sh
csadorf: > > In this way the script will fail if the count cannot be determined.
> 
> Are you sure? [...]

No, you are correct. I did not realize the `set -e` flag was not set.

> You could set the `set -e` bash flag to automatically exit the script when a non-zero exit status is returned from any command.

I opted for explicitly handling the failure of determination (which would only matter if the user did not override the variable) in https://github.com/aiidateam/aiida-core/pull/5161/commits/47f8a97b651fd2fd8e616ca356d46633185e2315 .
> 
> > We could add a pip install psutil~=5.6 command directly in this script if you prefer.
> 
> Don't think that is a good idea. I think the script should simply fail if it cannot properly determine the value and not continue silently. Even better, if the env variable is defined (which would override it anyway) maybe it is best to simply log the error as a warning and then continue anyway.
> 

This failure case would be handled by the error handling implemented in https://github.com/aiidateam/aiida-core/pull/5161/commits/47f8a97b651fd2fd8e616ca356d46633185e2315 .

> On a side-note: when is this script being run? When the image is built or when it is being run? That is to say, will this get the number of CPUs of the localhost that is building the image or the one that is running. I hope that it is the latter, because if not it would most likely still get the wrong value configured, would it not?
> 

Good catch, this script only makes sense to run on the host machine which is indeed the case.

> > We use psutil due to the reasons outlined here
> 
> Gotcha. Would have been useful to include this info in a comment, and/or at the very least in the commit message stating that you need `psutil`'s version because `os.cpu_count` does not give merely the physical CPUs, which is what is required here.

I will add a comment.

.docker/opt/configure-aiida.sh
csadorf: DONE.

.docker/opt/configure-aiida.sh
```diff
@@ -33,6 +33,24 @@ if [[ ${NEED_SETUP_PROFILE} == true ]]; then
...
+
+    # Determine the number of physical cores as a default for the number of
+    # available MPI ranks on the localhost. We do not count "logical" cores,
+    # since MPI parallelization over hyper-threaded cores is typically
+    # associated with a significant performance penalty. We use the
+    # `psutil.cpu_count(logical=False)` function as opposed to simply
+    # `os.cpu_count()` since the latter would include hyperthreaded (logical
+    # cores).
+    NUM_PHYSICAL_CORES=$(python -c 'import psutil; print(int(psutil.cpu_count(logical=False)))' 2>/dev/null)
+    LOCALHOST_MPI_PROCS_PER_MACHINE=${MPI_PROCS_PER_MACHINE:-${NUM_PHYSICAL_CORES}}
```
sphuber, line 45:
```suggestion
    LOCALHOST_MPI_PROCS_PER_MACHINE=${LOCALHOST_MPI_PROCS_PER_MACHINE:-${NUM_PHYSICAL_CORES}}
```

sphuber (changes requested):
Thanks @csadorf . All good, I just think you forgot to replace one occurrence of `MPI_PROCS_PER_MACHINE`. Once corrected this can be merged.

csadorf:
> Thanks @csadorf . All good, I just think you forgot to replace one occurrence of `MPI_PROCS_PER_MACHINE`. Once corrected this can be merged.

Looks like a git conflict resolution went wrong...

csadorf:
No idea why the tests are not passing anymore.

sphuber:
> No idea why the tests are not passing anymore.

They were not even running because Github Actions is in degraded mode: https://www.githubstatus.com/

sphuber (approved):

PR #5164, opened 2021-10-05
Backport the Github action workflow, to build and push the Docker image, to the `support/1.6.x` branch

sphuber:
@ltalirz please just check that this adds what you want but _do not merge_ the PR. Will have to be done manually from CLI.

### DIFF
--- a/.github/workflows/push_image_to_dockerhub.yml
+++ b/.github/workflows/push_image_to_dockerhub.yml
@@ -15,6 +15,9 @@ jobs:
 
   build-and-push:
 
+    # Only run this job on the main repository and not on forks
+    if: github.repository == 'aiidateam/aiida-core'
+
     runs-on: ubuntu-latest
     timeout-minutes: 30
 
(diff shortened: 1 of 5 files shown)


.github/workflows/push_image_to_dockerhub.yml
```diff
@@ -0,0 +1,54 @@
+# Build the new Docker image on every commit to master/develop branch and on every new tag.
+# No caching is involved for the image build. The new image is then pushed to the Docker Hub.
+
+name: build-and-push-to-dockerhub
+
+on:
+  push:
+    branches:
```
ltalirz, line 8: Listing these two branches here should not be necessary, but I think it will also not hurt.

If we also want the head of the `support/*` branches to be pushed to dockerhub, we would need to specify that here (but this is not necessary).
The tags should be enough.

ltalirz (approved):
thanks @sphuber , this looks good to me

ltalirz:
closing the PR since the docker-related actions have already completed

PR #5165, opened 2021-10-05
👌 IMPROVE: constructor of base data types

mbercx:
Adapt the constructor of the `Dict` and `List` data types so the
value no longer needs to be provided as a keyword argument, but can
simply be passed as the first positional argument.

Remove the `*args` from the `BaseType` constructor, instead specifying
the `value` input argument. Otherwise users could just pass multiple
positional arguments and it would only use the first without raising an
error.

Also fixes two issues with the `List` class:

* The `remove` method was not working as prescribed. Instead of
removing the first element in the list with the specified value, it
simply deleted the element with index `value`.
* The `set_list()` method didn't make a copy of the `list` input, so any
modification done to the `List` instance would also affect the original
`list`. If the same list is used to initialise several `List` instances,
adapting one would affect the other.

Finally, add tests for the `List` class.


mbercx:
Seems there _were_ already some tests for `List` in `tests/test_base_dataclasses.py`. Will check why this one fails and move them to `tests/orm/data/test_list.py`. 

I also noticed that these tests use test classes based on `AiidaTestCase`. Is there a preference between using this setup and separate functions decorated with `@pytest.mark.usefixtures('clear_database_before_test')`?

chrisjsewell:
> I also noticed that these tests use test classes based on AiidaTestCase. Is there a preference between using this setup and separate functions decorated with @pytest.mark.usefixtures('clear_database_before_test')?

yep `AiidaTestCase` is "deprecated", see https://github.com/aiidateam/aiida-core/pull/4782 for converting

mbercx:
Alright, refactored the basetype tests. I also moved the `data` dir in the tests to `nodes` (renamed from `node`, to match the package directory tree.

Some things I found out while working on this:

1. Although adding together a `str` and `bool` is perfectly possible:
  ```Python
  In [1]: 'a' + 'b'
  Out[1]: 'ab'
  
  In [2]: True + False
  Out[2]: 1
  ```
  Doing this for `Str` or `Bool` raises a `TypeError: unsupported operand type(s)`.

2. `abs` is not implemented for the `BaseType` classes:
```Python
In [3]: abs(Float(-2.0))
<ipython-input-5-324814f1c59f> in <module>
----> 1 abs(Float(-2.0))

TypeError: bad operand type for abs(): 'Float'
```

3. Some interesting inplace addition behaviour:
```Python
In [4]: f = Float(2)

In [4]: f += 2.0

In [5]: f
Out[5]: <Float: uuid: 8ee7db38-6a28-442f-8bfb-827de8f9d897 (unstored) value: 4.0>

In [6]: f.store()
Out[6]: <Float: uuid: 8ee7db38-6a28-442f-8bfb-827de8f9d897 (pk: 1) value: 4.0>

In [7]: f += 0.1

In [8]: f
Out[8]: <Float: uuid: cbf4b5ee-1474-46d6-aa48-c11a0639d12f (unstored) value: 4.1>
```

unkcpz (dismissed):
Thanks, @mbercx! Looks good to me.

As for the tests files move to `node/data/`, I like this new folder refactoring, it has more parallel structure with the code they test.

mbercx:
Thanks for the review @unkcpz! Just rebased, good to merge once the test complete.

unkcpz (approved):
Thanks @mbercx!

PR #5166, opened 2021-10-06
Add capability to download orm.ArrayData from REST-API

### LINKED ISSUE #2809: Rest api: add endpoint to download ArrayData
Currently the endpoint is provided to get general information of ArrayData but there is no endpoint available to download the raw data (numpy array) stored in ArrayData.

Add endpoint for ArrayData to get the size of the raw data and also to download it.

JPchico:
Right now it is not possible to access the arrays stored in the db via the REST-API.

To try to solve this a simple option to download the data in a JSON format has been added.
The function creates a dictionary where the keys are the `arraynames` and the values are lists obtained from the numpy arrays via the `.tolist()` method.

This dictionary is then converted into JSON format so that the user can obtain the contents of the arrays via a REST request in the form
`http://127.0.0.1:5000/api/v4//nodes/UUID_OF_ARRAY/download?download_format=json`.

closes #2809

### DIFF
--- a/aiida/orm/nodes/data/array/array.py
+++ b/aiida/orm/nodes/data/array/array.py
@@ -222,4 +222,4 @@ class ArrayData(Data):
         if comments:
             json_dict['comments'] = get_file_header(comment_char='')
 
-        return json.dumps(json_dict).encode('utf-8'), {}
+        return json.dumps(json_dict, ignore_nan=True).encode('utf-8'), {}
--- a/aiida/orm/nodes/data/array/array.py
+++ b/aiida/orm/nodes/data/array/array.py
@@ -193,3 +193,28 @@ class ArrayData(Data):
                 f'Mismatch of files and properties for ArrayData node (pk= {self.pk}): {files} vs. {properties}'
             )
         super()._validate()
+
+    def _get_array_entries(self):
+
+        array_dict = {}
+        for key, val in self.get_iterarrays():
+            array_dict[key] = val.tolist()
+        return array_dict
+
+    def _prepare_json(self, main_file_name='', comments=True):  # pylint: disable=unused-argument
+        """
+        Prepare a json file in a format compatible with the AiiDA band visualizer
+
+        :param comments: if True, print comments (if it makes sense for the given
+            format)
+        """
+        from aiida import get_file_header
+        from aiida.common import json
+
+        json_dict = self._get_array_entries()
+        json_dict['original_uuid'] = self.uuid
+
+        if comments:
+            json_dict['comments'] = get_file_header(comment_char='')
+
+        return json.dumps(json_dict).encode('utf-8'), {}
--- a/aiida/orm/nodes/data/array/array.py
+++ b/aiida/orm/nodes/data/array/array.py
@@ -195,7 +195,12 @@ class ArrayData(Data):
         super()._validate()
 
     def _get_array_entries(self):
+        """Return a dictionary with the different array entries.
 
+        The idea is that this dictionary contains the array name as a key and
+        the value is the numpy array transformed into a list. This is so that
+        it can be transformed into a json object.
+        """
         array_dict = {}
         for key, val in self.get_iterarrays():
             array_dict[key] = val.tolist()
@@ -203,7 +208,7 @@ class ArrayData(Data):
 
     def _prepare_json(self, main_file_name='', comments=True):  # pylint: disable=unused-argument
         """
-        Prepare a json file in a format compatible with the AiiDA band visualizer
+        Prepare a json file to extract the arrays stored in this node
 
         :param comments: if True, print comments (if it makes sense for the given
             format)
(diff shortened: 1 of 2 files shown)


JPchico:
Hi @chrisjsewell ! I was looking at the serialization of the numpy arrays to see if one can make it more in the spirit of what is done the `JsonableData`. As numpy arrays do not have the `as_dict` method, the dictionary would need to be something like this
```
json_dict = {
    'array_name': {
        '@class': obj.__class__.__name__,
        '@module': obj.__class__.__module__,
        'data': obj.tolist()
    }
}
```
However, the `@class` and `@module` entries would always be `ndarray` and `numpy` respectively. So I'm not too sure about whether this would be useful. One could add an entry for the `'dtype':  obj.dtype` which might be useful.

Regarding the sanitation of the `np.nan` to make the array json compatible, one could use something like what is used in 
https://github.com/aiidateam/aiida-core/blob/0772f3323f0817c1c63cc87d6841d630578f9ece/aiida/orm/nodes/data/jsonable.py#L90-L108

instead of transforming the json variables to float one would do it the other way around transforming the float/numpy variables to strings (e.g. `np.nan -> 'NaN'`). However, a lot of this is already taken care of by the `json.dumps` `ignore_nan=True` option, at least with regards to the `np.nan`. 

So in my opinion the current approach is quite good, one can make the modifications I raise here, but I'm unsure on whether it is something that it would be necessary.

chrisjsewell:
> However, the @class and @module entries would always be ndarray and numpy respectively. So I'm not too sure about whether this would be useful.

Heya, the other reason I would say this is probably not the way to go, is that the REST API should really provide language agnostic data, e.g. if the client was written in JavaScript, then there is not really anything it can do with a Python class & module

JPchico:
Hi @chrisjsewell ! I was looking at the serialization of the numpy arrays to see if one can make it more in the spirit of what is done the `JsonableData`. As numpy arrays do not have the `as_dict` method, the dictionary would need to be something like this
```
json_dict = {
    'array_name': {
        '@class': obj.__class__.__name__,
        '@module': obj.__class__.__module__,
        'data': obj.tolist()
    }
}
```
However, the `@class` and `@module` entries would always be `ndarray` and `numpy` respectively. So I'm not too sure about whether this would be useful. One could add an entry for the `'dtype':  obj.dtype` which might be useful.

Regarding the sanitation of the `np.nan` to make the array json compatible, one could use something like what is used in 
https://github.com/aiidateam/aiida-core/blob/0772f3323f0817c1c63cc87d6841d630578f9ece/aiida/orm/nodes/data/jsonable.py#L90-L108

instead of transforming the json variables to float one would do it the other way around transforming the float/numpy variables to strings (e.g. `np.nan -> 'NaN'`). However, a lot of this is already taken care of by the `json.dumps` `ignore_nan=True` option, at least with regards to the `np.nan`. 

So in my opinion the current approach is quite good, one can make the modifications I raise here, but I'm unsure on whether it is something that it would be necessary.

> > However, the `@class` and `@module` entries would always be ndarray and numpy respectively. So I'm not too sure about whether this would be useful.
> 
> Heya, the other reason I would say this is probably not the way to go, is that the REST API should really provide language agnostic data, e.g. if the client was written in JavaScript, then there is not really anything it can do with a Python class & module

Hi @chrisjsewell I agree, in my opinion the current implementation gives you basically what you need, the arrays are written as python lists, which transform to JSON arrays. The only thing I can think to make it more flexible is the explicit writing of `NaN` and `Infinity`.  Otherwise I think this is pretty much done. Do you agree?

JPchico:
Hi @chrisjsewell ! Just wanted to double check if you thought that the last suggestion was needed to be able to merge the PR, or if there is something else that need to be done. It would be nice to have this as part of the main branch (even if the REST-API might change to GraphQL later)

sphuber (approved):
Thanks @JPchico . I hadn't reviewed this because I typically am not involved much with the REST API and its developments, but I think we should decide on this PR. For me this would be fine to merge. @chrisjsewell or @ltalirz if you still want to have a look, let me know by tomorrow otherwise I will merge this.

sphuber:
@chrisjsewell any objections to merging this today?

chrisjsewell:
> any objections to merging this today?

eurgh, yeh sure. I don't foresee this being part of the future REST-API, but don't want to block on it.

chrisjsewell:
the merge conflict need to be fixed first though obviously

chrisjsewell:
> I don't foresee this being part of the future REST-API

Not that there is anything specifically wrong with this PR FYI. Just the whole of `aiida/restapi` lol

sphuber:
Since the interface doesn't allow to resolve the conflicts online, I had to do it locally and had to make a new PR #5425

sphuber:
Thanks for your contribution @JPchico ! And apologies for the long wait.

PR #5169, opened 2021-10-08
♻️ REFACTOR: Delegate `RepositoryBackend` control to the `Backend`

chrisjsewell:
This PR is a result of observations in #5154 and #5145:

By the current placement of `get_repository` on the `Profile` class, this caused two principal issues:

1. It tied all possible `Backend` implementations to the `DiskObjectStoreRepositoryBackend`
2. It made it impossible to implement/instantiate a `Backend` that was not directly dependent on the currently loaded `Profile`.

These issues, for example, preclude the possibility of implementing a "full" archive backend and instantiating it, whilst a "main" `Backend` is also loaded.

This PR moves the `get_repository` method to the `Backend` class, addressing both these issues.
It also removes the `instatiate` and `delete` methods from the (node) `Repository` class (which were just piping through to the backend), since a single node should not have "direct" access to deleting the whole profile's object store 😬.

Additionally, it improves the docstring of the `Backend` and fixes a minor issue, that `verdi profile delete` left an empty repository folder for that profile.

---

Note, this is likely the first in the number of PRs to address #5154:

No offence to the people involved (hindsight is always 20/20) but the whole abstraction and flow of control/information between `Config`, `Profile`, `Manager`, `BackendManager`, and `Backend` is IMO a bit of a mess really (plus some unnecessary use of global variables, which should always be avoided wherever possible). 
This greatly hampers implementing desirable features like profile switching.

Firtly, the flow from `Profile` -> `Backend` could be improved by having the `Backend` instantiate with a `Profile` (and having this also set the `Backend` on the `Profile`, to ensure only one `Backend` is instantiated per profile).
With this PR, but already with `Backend.get_session`, we are pulling information from the profile via a global variable, rather than via a `self.profile` attribute.
This would then also allow for the `ENGINE` and `SESSION_FACTORY` global variables, in `aiida/backends/djsite/__init__.py` and `aiida/backends/sqlalchemy/__init__.py`, to be removed; delegating responsibility for database session management to the relevant backend.

Similarly, `Config.delete_profile` and `Manager._load_backend` (currently hard coded to sqla/django backends), should delegate some of their responsibilities to the `Backend` implementation.

### DIFF
--- a/aiida/repository/repository.py
+++ b/aiida/repository/repository.py
@@ -493,3 +493,21 @@ class Repository:
 
                 with self.open(root / filename) as handle:
                     filepath.write_bytes(handle.read())
+
+    # these methods are not actually used in aiida-core, but are here for completeness
+
+    def initialise(self, **kwargs: Any) -> None:
+        """Initialise the repository if it hasn't already been initialised.
+
+        :param kwargs: keyword argument that will be passed to the ``initialise`` call of the backend.
+        """
+        self.backend.initialise(**kwargs)
+
+    def delete(self) -> None:
+        """Delete the repository.
+
+        .. important:: This will not just delete the contents of the repository but also the repository itself and all
+            of its assets. For example, if the repository is stored inside a folder on disk, the folder may be deleted.
+        """
+        self.backend.erase()
+        self.reset()


giovannipizzi:
Hi!
I agree that the current flow between the various classes you mention has space for improvement (also for me is hard to remember what is where) and the changes you suggest look good to me.
I'd just wait if possible to have feedback from @sphuber (I think he's OOO this week, so we might need to wait next week).

In the meantime, just to know where we're heading towards (so that it's clear and, if we all agree, will make it faster to review further PRs), @chrisjsewell did you already write down somewhere (e.g. in an issue) a minimal description of how you envision the classes to behave eventually? (In addition, having this written down will make it easy to move it to the design docs).

chrisjsewell:
Bear with me @giovannipizzi, I am literally just writing some UML diagrams for this (before and after)!

Issue #3382, opened 2019-10-04
Ability to "bulk store" Nodes


CasperWA:
In order to properly update and redo the export/import of AiiDA archives, a "bulk store" method is needed through the front-end ORM. Optimally, it should utilize the back-end-specific equivalent function/method (i.e., `session.add_all()` for SQLA and `Model.objects.bulk_create()` for Django).

This may be a method in the Collections, making it similar to invoke as the Django equivalent.

This is needed to optimize speed and code for especially import, as it will also make it more back-end-agnostic.

Issue #2496, opened 2019-02-16
Formalize the usage of transactions between the front end and backend layers


sphuber:
The usage of transactions in the ORM should be formalized in the sense of where they are defined, used and who bears the responsibility. Probably, the backend ORM layer should never explicitly open/control a transaction, because it will very often be a front end instance who will have to ask multiple independent backend instances to store themselves and the entire store should happen in a single transaction because the individual parts would be unintelligible.

For example, consider a `Node` instance. In the near future, the repository will be refactored to store a reference of the files the node contains in a separate database table. When storing a new node instance then, the front end class will have to instruct the backend node to store itself, after which it asks the repository backend instance to store a reference of its contents in the table. Only when both operations succeed should the changes be committed. To avoid a direct dependency between the backend node and backend repository interfaces, it is the front end node class that will have to take responsibility for creating a transaction and committing the changes.

giovannipizzi:
A few comments here.

1. In places like we use rollback: https://github.com/aiidateam/aiida_core/blob/b8eaffca3c448d1242eded845aa118b4bc1ae1a9/aiida/orm/implementation/sqlalchemy/nodes.py#L185-L197
   I think this pattern is incorrect because even if another exception happens that has nothing to do with the DB, we roll back.
   The suggested approach [here](https://docs.sqlalchemy.org/en/13/orm/session_transaction.html) has an additional bit, that is having a `.begin()` before the block, so that you roll back only what you were doing in that bit of code (and actually uses `except` instead of `except Exception`, that in this specific case I think it's better since we anyway re-raise, and we would also properly catch things like CTRL+C. This is the [link for django](https://docs.djangoproject.com/en/1.11/topics/db/transactions/#transaction-rollback)

1. I have the feeling that the approach to follow is:
   - in the backend, provide functionality to manage transactions
   - provide a decorator that the user can use if they want to wrap (we should already have it I think)
   - in the backend, do not wrap any operation in transactions
[...]

PR #5171, opened 2021-10-08
✨ NEW: Add `Backend` bulk methods

chrisjsewell:
Partially extracted from (and required by) #5145 and also work towards #5154

Adds `Backend.bulk_insert` and `Backend.bulk_update` methods, and moves the `delete_nodes_and_connections` function to a method on the `Backend`.
This removes the need for using backend specific code outside of the backend, for these use cases (still plenty more to fix)

`delete_nodes_and_connections` performs multiple operations and so should only be called within a transaction, thus it raises an `AssertionError` if this is not the case.

(this addresses, but maybe not fully, #3382 and #2496)

### DIFF
--- a/aiida/orm/implementation/django/backend.py
+++ b/aiida/orm/implementation/django/backend.py
@@ -168,7 +168,7 @@ class DjangoBackend(SqlBackend[models.Model]):
 
     def delete_nodes_and_connections(self, pks_to_delete: Sequence[int]) -> None:
         if not self.in_transaction:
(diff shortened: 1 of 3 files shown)


chrisjsewell:
> Also, one more general question, now many PRs have at the bottom a number of "errors" under the title "Unchanged files with check annotations". What are these? Should they be ignored?

Yeh I don't know either, its showing errors when it encounters an exception raise, which seems nonsensical for a static analyser. May be related to https://github.com/actions/toolkit/issues/457

Issue #2216, opened 2018-11-19
Add caching of object collections


muhrin:
Currently ORM entity collections are recreated each time `.objects` is called on an entity type e.g. `User.objects`.  This should be cached as the cost of construction is likely to be significant and the number of times this is called could be very lage.  What's more the underlying collection has no reason to be re-instantiated, it's the same collection each time.

A start point would be to add a lazy-caching to this function:

https://github.com/aiidateam/aiida_core/blob/aa7907260906f01d8d7bb04b1b1421b24ab00d3f/aiida/orm/entities.py#L116

ltalirz:
@muhrin I see everything is already pretty much prepared for this, so I'm happy to add this myself.

The question remains what the logic should be to check whether the cache can be used?

Let me know whether you agree with this logic
```python
    @classproperty
    def objects(cls, backend=None):  # pylint: disable=no-self-use, no-self-argument
        """
        Get an collection for objects of this type.

        :param backend: the optional backend to use (otherwise use default)
        :return: an object that can be used to access entites of this type
        """
        from . import backends

        # new_backend always holds an actual backend object
        new_backend = backend or cls._BACKEND or backends.construct_backend()

        if type(cls._BACKEND) != type(new_backend):
            # if the backend has changed, update cache
            cls._BACKEND = new_backend
            cls._OBJECTS = cls.Collection(cls._BACKEND, cls)
        else:
            # if not, use cache (if already populated)
            cls._OBJECTS = cls._OBJECTS or cls.Collection(cls._BACKEND, cls)

        return cls._OBJECTS
```

muhrin:
So I created a working version in my branch `collection_and_user_caching`: https://github.com/muhrin/aiida_core/blob/collection_and_user_caching/aiida/orm/entities.py

Sorry to not update you.  It works, however it exposes one problem I haven't figured out how to deal with which is that the default User gets deleted at the end of a test, however, it is still cached and so points to a non-existent user.

To answer your question, IMHO the criterion should be the backend instance reference, this is what ties everything together.  It's perfectly legitemate to load, e.g., two Django backends that point to different databases.

I would suggest that we proceed as follows:
1. Merge my branch into yours.
2. Figure out/decide what to do for the user deletion issues (and indeed deletion in general)

ltalirz:
@muhrin 
I've merged your branch and merged `provenance_redesign` back into it.
You have write access to my fork, so you can go ahead and continue directly there if you like
[...]

PR #5183, opened 2021-10-19
👌 IMPROVE: Entity Collection typing

chrisjsewell:
This PR makes static types available for `Entity.ojects`.
As an example, in `pyproject.toml` comment out `follow_imports`:

```toml
[[tool.mypy.overrides]]
module = 'aiida.*'
# can only follow these imports when more of the code is typed
# follow_imports = "skip"
```

and run `example.py` against mypy (with the pre-commit env loaded):

```python
from aiida.orm import Computer, Data, Dict, Group, User

reveal_type(Computer.objects)
reveal_type(Group.objects.all())
reveal_type(User.objects.get_default())
reveal_type(Data.objects)
reveal_type(Data.objects(get_manager().get_backend()))
reveal_type(Data.objects.get())
reveal_type(Dict.objects.get())

data = Data.objects.all()[0]
reveal_type(data)
```

```console
(.../aiida_core_develop/.tox/py38-pre-commit) aiida_core_develop/$ mypy example.py
example.py:4: note: Revealed type is "aiida.orm.computers.ComputerCollection"
example.py:5: note: Revealed type is "builtins.list[aiida.orm.groups.Group*]"
example.py:6: note: Revealed type is "aiida.orm.users.User"
example.py:7: note: Revealed type is "aiida.orm.nodes.node.NodeCollection[aiida.orm.nodes.data.data.Data*]"
example.py:8: note: Revealed type is "aiida.orm.nodes.node.NodeCollection[aiida.orm.nodes.data.data.Data]"
example.py:9: note: Revealed type is "aiida.orm.nodes.data.data.Data*"
example.py:10: note: Revealed type is "aiida.orm.nodes.data.dict.Dict*"
example.py:13: note: Revealed type is "aiida.orm.nodes.data.data.Data*"
```

~~One thing to note above, is that entity subclass methods give base-class entity types, `Data.objects.get() -> Node`.
Ideally this would give `Data`, but I haven't been able to work this out (it is already tricky enough, with the cyclic nature of `Entity -> Collection -> Entity ...` 😬)
To stress, this is only an issue with the static typing, not the actual runtime return types.~~ (fixed, see below)

A few points:

- The `Collection` sub-classes are now defined outside of the `Entity` sub-classes. There is no reason why they cannot be used independent of the `Entity` itself.
- I left the `Entity.Collection` attribute, for back-compatibility, but actually it is no longer needed.
- The ` aiida/common/datastructures::LazyStore` has been removed and replaced with https://docs.python.org/3/library/functools.html#functools.lru_cache. Caching was added due to #2216; I'm interested if it really speeds things up that much, but also there's no downside to keeping it.
  You can view it working with:
  ```python
  In [1]: Data.objects; Data.objects; User.objects; Group.objects.get_cached.cache_info()
  Out[1]: CacheInfo(hits=1, misses=3, maxsize=100, currsize=3)
  ```
- There was a number of uses of e.g. `Comment.objects` that were fixed to ensure the `Collection` uses the same backend as the `Entity`, e.g. `Comment.objects(self.backend)`
- #5182 ensures e.g. `Node.objects.get` uses don't raise (incorrect) `no-member` issues in pylint
- I noted that `User.get_or_create` returns the opposite way to `Computer` or `Group`, i.e. `Tuple[bool, 'User']`, not `Tuple['User', bool]`. But I don't see away to deprecate that 🤷

### DIFF
--- a/aiida/orm/computers.py
+++ b/aiida/orm/computers.py
@@ -36,7 +36,6 @@ class ComputerCollection(entities.Collection['Computer']):
         create (and store) a new Computer if such a Computer was not present yet.
 
         :param label: computer label
-        :type label: str
 
         :return: (computer, created) where computer is the computer (new or existing,
             in any case already stored) and created is a boolean saying
@@ -53,7 +52,7 @@ class ComputerCollection(entities.Collection['Computer']):
         """Return a list with all the labels of the computers in the DB."""
         return self._backend.computers.list_names()
 
-    def delete(self, pk: int):
+    def delete(self, pk: int) -> None:
         """Delete the computer with the given id"""
         return self._backend.computers.delete(pk)
 


sphuber:
Thanks @chrisjsewell merge away

PR #5190, opened 2021-10-20
🔧 MAINTAIN: Move pylint plugin to external package

chrisjsewell:
This will allow pylint/pre-commit to be called from anywhere,
and also allows for plugins to use it if necessary.


chrisjsewell:
I guess this seems a good a place as any to house it @sphuber?

As mentioned in #5182, there is probably some improvements that can be made with this plugin, but I think it's already better than having to add a tonne of pylint disables, for uses of e.g. `Entity.objects`.
Naturally, if there are any replies to https://github.com/PyCQA/pylint/issues/1694#issuecomment-946350213, then I will update it.

chrisjsewell:
The other approach would be to turn it into a standalone package. I can do that but probably not today

PR #5192, opened 2021-10-22
🧪 TESTS: SQLA Migrations -> pytest

chrisjsewell:
This PR primarily converts the SQLA Migrations tests from using `AiidaTestCase` to using pytest fixtures, and splits them out into multiple files (to improve readability).
The tests are functionally equivalent, just restructured and with a reduction in duplication when using `sqlalchemy.Session` contexts (hence the reduction in lines).
One more test has been added (`tests/backends/aiida_sqlalchemy/migrations/test_all_basic.py`), to cycle though all migration versions and try migrating them down&up from&to the latest revision (against an empty database). The list of versions is dynamically generated, to ensure none are missed.

To ensure a fully empty database, an option has been added to the test manager `aiida_profile.reset_db(with_user=False)` (defaults to `True`), whereby the default User will not be auto-generated. This was required for some migrations to succeed when migrating `down` then `up`, and was how the current unittest implementation worked.

Note this was actually done in a seperate PR in the end

```
Additionally, methods have been renamed:

- `get_schema_version_database()` -> ~~`get_schema_version_current()`~~ `get_schema_version_backend()`
- `get_schema_version_code()` -> ~~`get_schema_version_latest()`~~ `get_schema_version_head()`

I feel this is clearer what they return
```

Also fixes sqlalchemy v2 deprecation: https://docs.sqlalchemy.org/en/14/orm/extensions/automap.html?highlight=prepare#sqlalchemy.ext.automap.AutomapBase.prepare.params.autoload_with

This is a precursor to #5097


chrisjsewell:
> @ltalirz you also happy with get_schema_version_head and get_schema_version_backend?

Pinging @ltalirz, would like to get this merged ASAP 😄

tests/backends/aiida_sqlalchemy/migrations/test_all_basic.py
```diff
@@ -0,0 +1,35 @@
...
+
+    So we only run for all versions later than this.
+    """
+    from aiida.backends.sqlalchemy.manager import SqlaBackendManager
+    migrator = SqlaBackendManager()
+    all_versions = migrator.list_schema_versions()
+    first_index = all_versions.index('a514d673c163') + 1
+    # ideally we would pytest parametrize this, but then we would need to call list_schema_versions on module load
+    for version in all_versions[first_index:]:
+        print(version)
```
sphuber, line 31:
```suggestion
```

tests/backends/aiida_sqlalchemy/migrations/test_calc_attributes_keys.py
```diff
@@ -0,0 +1,96 @@
...
+# Copyright (c), The AiiDA team. All rights reserved.                     #
+# This file is part of the AiiDA code.                                    #
+#                                                                         #
+# The code is hosted on GitHub at https://github.com/aiidateam/aiida-core #
+# For further information on the license, see the LICENSE.txt file        #
+# For further information please visit http://www.aiida.net               #
+###########################################################################
+"""Tests migration of the keys of certain attribute for ProcessNodes and CalcJobNodes."""
+from .conftest import Migrator
+
```
sphuber, line 12: Why separate out some of these migration tests to separate files, but not all of them? Feels a bit arbitrary to take a few out but still leave a lot of them in `test_other.py`...

chrisjsewell: Yeh fair: I felt it was too much for one file (as before), but overkill to split every migration into a separate file.
I've now split them up a bit better; trying to group them in related/sequential migrations, and added numbering to the file names, to make it easier to see the order.

chrisjsewell:
this is ready for re-review cheers @sphuber

chrisjsewell:
Ah what a pain in the backside: `tox -e py37-sqla tests/backends/aiida_sqlalchemy/migrations` runs fine, but there is something about the prior `tests/backends/aiida_sqlalchemy/test_session.py`, that is leaving the test database in a bad state that is making the migrations stall (only if you call the tests in a certain order)

chrisjsewell:
ok should be fixed now

aiida/backends/manager.py
```diff
@@ -179,8 +179,8 @@ def is_database_schema_ahead(self):
         """
 
     @abc.abstractmethod
-    def get_schema_version_code(self):
-        """Return the code schema version."""
+    def get_schema_version_head(self):
+        """Return the latest schema version."""
```
sphuber, line 183: The docstring could do with some extra clarification, also using the word "head" in it like the method name.

aiida/backends/manager.py
```diff
@@ -191,11 +191,8 @@ def get_schema_version_reset(self, schema_generation_code):
         """
 
     @abc.abstractmethod
-    def get_schema_version_database(self):
-        """Return the database schema version.
-
-        :return: `distutils.version.LooseVersion` with schema version of the database
-        """
+    def get_schema_version_backend(self):
+        """Return the current schema version."""
```
sphuber, line 195: Same thing here, would be good to give a bit more detail explaining that this is the version that is stored in the currently configured backend.

sphuber (approved):
Thanks @chrisjsewell . I have left two suggestions for the docstrings of the renamed methods, which I think would really help. Still, not the most critical thing ever, so I am going to approve.

aiida/backends/manager.py
chrisjsewell: done

aiida/backends/manager.py
chrisjsewell: done

sphuber (approved):

chrisjsewell:
Ah this so ***ing annoying, it runs fine locally, but here it hangs on a migration for some reason

chrisjsewell:
> Ah this so ***ing annoying, it runs fine locally, but here it hangs on a migration for some reason

Yeh err there is just something causing locking of the database in the migration tests, and it seems different whether you are using the test database automatically created by pgtest (what I do locally) or created upfront then setting `AIIDA_TEST_PROFILE` (whats done on GH Actions).

Note this only affects the tests, not running the migrations manually 

It is the same thing I encountered in https://github.com/aiidateam/aiida-core/pull/5097#issue-978831706

@sphuber / @giovannipizzi if you want to give this some "fresh eyes" (namely the `perform_migration` fixture in `tests/backends/aiida_sqlalchemy/migrations/conftest.py`), otherwise I'll revisit when I get back from holiday in a week

PR #5194, opened 2021-10-24
Verdi: Warn for already existing ``code@computer label``

janssenhenning:
See #3303 

Implement reprompting mechanism in ``verdi code setup`` and ``verdi code duplicate`` for code label/computer combinations that are already in the database. The Computer option gains callback checking the DB for the set code label/computer combination

If it is found the user is asked if he would like to change the label. If that is the case the label is set to `None`, which triggers a second label argument, which will perform the same check as the computer option but it will raise an error and reprompt instead. This second label argument is ``hidden=True, expose_value=False`` to avoid duplicate help messages and name clashes with the other label option

Works both for remote and local codes. For non-interactive commands an error is thrown


ramirezfranciscof:
Hey @janssenhenning , sorry but it seems that apparently there were some other details that needed to be sorted out together with this issue 😅 (see for example #5204) so we ended up having to take more personal control of this feature, I hope that is ok. We anyways appreciate a lot the contribution and the initiative to participate in this project! Feel free to check the merged PR (#5205) in case there is any aspect you find still missing from that (and if not, then I'll let you close this PR).

sphuber:
@janssenhenning I had written a message here before merging #5205 but apparently that never made it to the server, not sure what happened.

Anyway, I wrote to thank you for your contribution and that the reason I opened an alternative PR was not my original intention. When @ramirezfranciscof asked me whether the suggested approach was feasible, I remembered having looked into this and found issues due to how `click` works, which I wrote in the post above. Since I thought that solution was a dead end, I didn't want to give the go ahead for you to start changing your PR, because you would end up doing a lot of work for nothing. Therefore I decided to create a quick MWE to demonstrate the problem I thought there to have been, but in doing so, figured out that actually it could be made to work. With some additional kinks to sort out, I ended up with the implementation done. At this point, it seemed more logical to just open a PR as opposed to have you change this one entirely by copy pasting the full implementation. Again, this was not to discourage you or be disrespectful. Your contributions would be very much appreciated and valuable.

janssenhenning:
@ramirezfranciscof @sphuber Absolutely no problem. I had this lying around for quite some time. I used it to get some practice with ``click`` to use it in ``aiida-fleur`` and related packages.

Looking at #5205 what happens if you store the code in the database so ``--store-in-db``? Then the ``computer`` option is ``None`` and the label can be duplicate again, right?

sphuber:
You are absolutely right @janssenhenning that is something we missed. This distinction between "remote" and "local" codes has always been confusing, especially when you have "remote" codes on localhost...

I have opened #5215 with a potential fix. It would be great if you could have a look and see if you can find any problems.

Issue #5178, opened 2021-10-18
Caching: ph process failed if previous pw.x process finished_ok and cached but remote folder cleaned. 


unkcpz:
- [ ] [AiiDA Troubleshooting Documentation](https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/troubleshooting.html)
- [ ] [AiiDA Users Forum](https://groups.google.com/forum/#!forum/aiidausers)

### Describe the bug

This problem happened when chain the `PwBaseWorkChain` and `PwBaseWorkChain` in a Workflow and using QE plugin with caching on for `aiida.calculations:quantumespresso.pw`. 
The `quantumespresso.ph` CalcJob require the remote folder of `quantumespresso.pw` run as input, but when it pw process finishes and remote folder is cleaned, the `quantumespresso.ph` will continues to fail since it can not found the saved files to do the calculation. 

### Steps to reproduce

This my workflow code to calculate the phonon frequencies, which do a SCF calculation and set the `scf_remote_folder` for next ph process step. 

```python
class PhononFrequenciesWorkChain(WorkChain):
    """WorkChain to calculate cohisive energy of input structure"""
    @classmethod
    def define(cls, spec):
    ....

    ....
    def inspect_scf(self):
        """inspect the result of scf calculation if fail do not continue."""
        workchain = self.ctx.workchain_scf

        if not workchain.is_finished_ok:
            self.report(
                f'PwBaseWorkChain failed with exit status {workchain.exit_status}'
            )
            return self.exit_codes.ERROR_SUB_PROCESS_FAILED_SCF

        try:
            self.ctx.scf_remote_folder = workchain.outputs.remote_folder
        except NotExistentAttributeError:
            return self.exit_codes.ERROR_NO_REMOTE_FOLDER

    def run_ph(self):
        """
        set the inputs and submit ph calculation to get quantities for phonon evaluation
        """
        inputs = {
            'metadata': {
                'call_link_label': 'PH'
            },
            'ph': {
                'code': self.inputs.ph_code,
                'qpoints': self.ctx.qpoints,
                'parameters': orm.Dict(dict=self.ctx.ph_parameters),
                'parent_folder': self.ctx.scf_remote_folder,
                'metadata': {
                    'options': self.ctx.options,
                },
                # CMDLINE setting for parallelization
            },
        }

        running = self.submit(PhBaseWorkflow, **inputs)
        self.report(f'Running ph calculation pk={running.pk}')
[...]

PR #5197, opened 2021-10-25
caching: order and use latest node when cache more than one

unkcpz:


greschd:
> and the impact in performance should not be very much

I'm not so sure about this: now, we can essentially make a query that can return when the first match is found. When we guarantee using the latest, worst-case (i.e., if there is no index on time) we'd have to go through all nodes and order them.

I don't have any insight query performance and what is or isn't indexed, just wanted to point out that this statement is not self-evident.

Other than that, I agree with returning the newest node being an improvement over the status quo.

ramirezfranciscof:
> I'm not so sure about this: now, we can essentially make a query that can return when the first match is found. When we guarantee using the latest, worst-case (i.e., if there is no index on time[1]) we'd have to go through all nodes and order them.

Mmm you think so? I mean, my logic was that when there is no equivalent hashed node, you are already going through all of the possible calcjobs and checking if the hash coincides, no? So although I understand why you say this would be slower than finding the first hit, I think it is still would be roughly comparable to the most common hash search query.

greschd:
I think it could make a big difference if it's a data type with lots of duplicates, like for example `Int`. If you're looking for _any_ `Int(1)` it will be found fairly quickly, whereas _the latest_ is only fast if there is a time index. Otherwise, it'll be O(N) with the number of `Int` in the DB.

This is a slightly exaggerated example, because we actually search for process nodes, not base types.

Actually, thinking about this, does anyone have experience with the caching performance on really large DBs? This was always a concern we've pushed to "when it becomes problematic". Compared to the initial design, what makes this somewhat worse is that -- if I remember correctly -- back then, different ``CalcJob``s had different node types, and we would filter by that.

sphuber:
> I think it could make a big difference if it's a data type with lots of duplicates, like for example Int

Except that `Data` nodes are not cachable anyway and so are never considered, so the problem of performance doesn't apply here.

giovannipizzi:
My 2 cents: I agree these are 2 independent problems, and (after checking for performance) I'm agreeing that returning the last one is a good change for this PR (but it would be good to do some benchmarking on a large DB?). If we use the `ctime`, and as long as it's indexed (to double check) it should be OK, but performance is always hard to predict (also, if the query for same hash returns few nodes, most probably the DB will anyway decide to do a scan search regardless of the presence or absence of an index).
Two comments:
- I would still suggest not to rely on this unless you're sure the last one is really always the right thing to do, but suggest to disable caching for nodes that shouldn't (usecase: I have 2 identical calculations run without caching, then I clean only the most recent one - this new approach would fail). But combined with the other suggestion in the PR by Sebastiaan, this would also be solved
- we're not really checking truly identical nodes, but equivalent nodes, as mentioned by Francisco. In addition, beside the fact that RemoteData are special in this sense (as they link to data that is not part of the hash), even the time itself is not part of the hash (nor the UUID) so they are not really identical, so making a choice makes sense.

Practically, I would suggest to create a few identical nodes in a otherwise big DB (millions of nodes), and then search for any equivalent one (like now) and the most recent one (like suggested) and measure speed. If there is no real problem, I'd go ahead with this (but push for the API to 'uncache' nodes and document how to use it, e.g. in the case of @unkcpz).

unkcpz:
Thanks @giovannipizzi . I am happy to do the check and performance test and then go and finish this PR. But have the following practical problems. 

> If we use the ctime, and as long as it's indexed (to double check) it should be OK

What is the meaning of `ctime` is indexed? Indexed in the PostgreSQL database, correct? You may need to elaborate on that since I am not familiar with the implementation of that part of aiida-core.

> Practically, I would suggest creating a few identical nodes in a otherwise big DB (millions of nodes), and then search for any equivalent one (like now) and the most recent one (like suggested) and measure speed.

Is there any big DB available somewhere exactly for the similar performance testing or do usually we test it on the production  DB that we are working on? If there is no such testing purpose DB, we should have one, I think.

sphuber:
> What is the meaning of `ctime` is indexed? Indexed in the PostgreSQL database, correct? You may need to elaborate on that since I am not familiar with the implementation of that part of aiida-core.

The `ctime` is a column (storing the datetime of a node's creation) on the `DbNode` table, which is where `Node`s are stored. An index can be placed on a column to make queries more efficient. You can see this in the model declaration. For example for Django, in `aiida.backends.djsite.db.models.DbNode` you can see the `ctime` is declared as:
```python
ctime = m.DateTimeField(default=timezone.now, db_index=True, editable=False)
```
The `db_index=True` means that Django will instruct Postgres to create an index. So that answers @giovannipizzi 's question for Django. Unfortunately, if you look for SqlAlchemy, there is no index on the `ctime` of a node. It has long been known that the database schemas for both implementations are not identical and @chrisjsewell has been working on a PR to make them identical, before we are dropping Django. So this does complicate the situation for this PR a bit, since you would have to really test for SqlAlchemy, but that currently doesn't have the required indices.

> Is there any big DB available somewhere exactly for the similar performance testing or do usually we test it on the production DB that we are working on? If there is no such testing purpose DB, we should have one, I think.

I think the biggest one out there is still the `sdb` one, but since that is so big, it is very unwieldy to create a clone. Just copying over the repository can take around 4 days. So until we have the new repository, which should make this a lot faster and manageable, we would probably have to find a smaller database. I don't have any medium-sized databases that I could give you unfortunately, but maybe @qiaojunfeng and @mbercx know of some?

sphuber:
As a final clarification: my point was not that we should never order the equivalent by `ctime` to determine the one used as cache source, but that the way that it was presented here was a bad solution to the problem it was trying to solve. Even if we merge this change, I think it should be made very clear that this should in no shape or form be used as a way to control caching as a user. So I am not sure if it should even be mentioned in the caching docs, at least not in the ones that are most important for the user.

unkcpz:
Back to the original caching issue where I want to work around with this PR, I now agree with @sphuber this should NOT be the standard way to control the caching. 
Although at the moment I have this PR's code in my production environment, it only serves as a workaround. 
If we have https://github.com/aiidateam/aiida-core/pull/5207 merged to control the caching of specific `CalcJob` instances, I would do in my workflow, disable the caching every time when the `clean()` function is called. 
For the already cached nodes, I will do the `clear_hash()` for all the same nodes.

So I personally prefer this be closed.

mbercx:
>  I don't have any medium-sized databases that I could give you unfortunately, but maybe @qiaojunfeng and @mbercx know of some?

I have quite a few databases of different sizes. What # nodes are you looking for? :)

unkcpz:
Close this and redirect it to the discussion https://github.com/aiidateam/aiida-core/discussions/5223.

For now, to work around the issue I mentioned in https://github.com/aiidateam/aiida-core/issues/5178, I query for the same node manually and use the latest node by invalid caching on all other old nodes, first by removing the `_hash` from extra attributes and we can do it now with set `is_valid_cache=False` for node instance after https://github.com/aiidateam/aiida-core/pull/5207 merged.

unkcpz:
> I have quite a few databases of different sizes. What # nodes are you looking for? :)

@mbercx thanks! As Giovanni mentioned, maybe 'millions' of nodes?

mbercx:
> @mbercx thanks! As Giovanni mentioned, maybe 'millions' of nodes?

I think the `3dd` database had around 3-4 million nodes when I synced it to `prnmarvelsrv`. I'll explain on Slack where you can find its repository and psql dump.

PR #5204, opened 2021-10-28
`Code`: add additional validation before storing

### LINKED ISSUE #5203: `Code` can be stored without a label whatsoever and duplicate full label

sphuber:
Fixes #5203 

Up till now it was possible to store a `Code` instance without a label;
either an empty string or even `None`. This is now checked in the
`_validate` method, which is automatically called when `store` is
called. It will raise a `ValidationError` if no valid label is defined.

In addition, a check is added for the uniqueness of the full label,
i.e., the string `label@computer.label`. Since this is usually used as
an identifier for a code, allowing duplicates will make it impossible to
load a code with it and one has to resort to using the PK or UUID.

One would imagine that these conditions should be enforced on the
database level, however, for the empty label this is not
straight-forward as the `Code` is just a special case of the node table
and there the label is optional and so `None` and the empty string are
perfectly possible. The same goes for the full label. This is a
uniqueness constraint across two tables, but it requires a `WHERE`
statement, since it should only apply if `node_type == data.code.Code.`.
This is possible in Postgres through partial indexes, but it is not
clear if both Django and SqlAlchemy support defining these on the models.


sphuber:
After discussion it was decided that the potential for breaking existing archives and databases was great and the added benefit was small, so we decide to leave this as is.

PR #5206, opened 2021-10-28
`get_entry_point`: add proper deprecation pathway for old entry points 

### LINKED ISSUE #5202: Verdi: DataParamType cannot be used with deprecated entry points

### Describe the bug

Using the ``DataParamType`` does not work with the now deprecated entry points like ``aiida.data:structure`` instead of ``aiida.data:core.structure``. It leads to errors like the following in the tests of ``aiida-fleur``.

```
________________________ ERROR collecting tests/cmdline/test_base_all_commands.py _________________________
../../aiida-core-fork/aiida/cmdline/params/types/identifier.py:62: in __init__
    entry_point = get_entry_point_from_string(entry_point_string)
../../aiida-core-fork/aiida/plugins/entry_point.py:183: in get_entry_point_from_string
    return get_entry_point(group, name)
../../aiida-core-fork/aiida/plugins/entry_point.py:265: in get_entry_point
    raise MissingEntryPointError(f"Entry point '{name}' not found in group '{group}'")
E   aiida.common.exceptions.MissingEntryPointError: Entry point 'structure' not found in group 'aiida.data'

cmdline/test_base_all_commands.py:4: in <module>
    from aiida_fleur.cmdline import cmd_root
../aiida_fleur/cmdline/__init__.py:21: in <module>
    from .launch import cmd_launch
../aiida_fleur/cmdline/launch/__init__.py:16: in <module>
    from .launch import launch_inpgen
../aiida_fleur/cmdline/launch/launch.py:18: in <module>
    from ..util import options
../aiida_fleur/cmdline/util/options.py:31: in <module>
    type=types.DataParamType(sub_classes=('aiida.data:structure',)),
../../aiida-core-fork/aiida/cmdline/params/types/identifier.py:64: in __init__
    raise ValueError(f'{entry_point_string} is not a valid entry point string: {exception}')
E   ValueError: aiida.data:structure is not a valid entry point string: Entry point 'structure' not found in group 'aiida.data'
```
### Steps to reproduce

Create a command with an option of type ``DataParamType`` with a deprecated entry point

```
import click
from aiida.cmdline.types import DataParamType

@click.command('test')
@click.option('--data', type=DataParamType(sub_classes=('aiida.data:structure',)))
def test_command(identifier):
     pass

```

### Expected behavior

The old entry point should continue to work with a deprecation warning

### Your environment

- Python version [e.g. 3.7.1]: 3.9.7
- aiida-core version [e.g. 1.2.1]: develop

sphuber:
Fixes #5202 

All entry points that ship with aiida-core were deprecated for v2.0
as they were changed to be properly prefixed with core.. To make sure
the old entry points would still be automatically loaded, the factories
were updated to automatically catch them, print a deprecation warning,
and load the new entry point instead.

However, the factory was not the correct place to put this logic, since
the `get_entry_point` method, which the factories call and is the lowest
function in the stack that actually retrieves the entry point, can also
be called directly, circumventing the deprecation mechanic added to the
factories. This would result in the deprecated entry points raising an
exception when being loaded, for example in the parameter types of the
command line that have support for specific entry points, such as the
`IdentifierParamType`.

The solution is to move the deprecation mechanic from the factories to
the lowest layer of `get_entry_point`.

### DIFF
--- a/aiida/plugins/entry_point.py
+++ b/aiida/plugins/entry_point.py
@@ -12,6 +12,7 @@ import enum
 import functools
 import traceback
 from typing import Any, List, Optional, Sequence, Set, Tuple
+from warnings import warn
 
 # importlib.metadata was introduced into the standard library in python 3.8,
 # but was then updated in python 3.10 to use an improved API.
@@ -20,6 +21,7 @@ from importlib_metadata import EntryPoint, EntryPoints
 from importlib_metadata import entry_points as _eps
 
 from aiida.common.exceptions import LoadingEntryPointError, MissingEntryPointError, MultipleEntryPointError
+from aiida.common.warnings import AiidaDeprecationWarning
 
 __all__ = ('load_entry_point', 'load_entry_point_from_string', 'parse_entry_point', 'get_entry_points')
 
@@ -70,6 +72,21 @@ ENTRY_POINT_GROUP_TO_MODULE_PATH_MAP = {
     'aiida.tools.dbimporters': 'aiida.tools.dbimporters.plugins',
     'aiida.transports': 'aiida.transports.plugins',
     'aiida.workflows': 'aiida.workflows',
+}
+
+DEPRECATED_ENTRY_POINTS_MAPPING = {
+    'aiida.calculations': ['arithmetic.add', 'templatereplacer'],
+    'aiida.data': [
+        'array', 'array.bands', 'array.kpoints', 'array.projection', 'array.trajectory', 'array.xy', 'base', 'bool',
+        'cif', 'code', 'dict', 'float', 'folder', 'int', 'list', 'numeric', 'orbital', 'remote', 'remote.stash',
+        'remote.stash.folder', 'singlefile', 'str', 'structure', 'upf'
+    ],
+    'aiida.tools.dbimporters': ['cod', 'icsd', 'materialsproject', 'mpds', 'mpod', 'nninc', 'oqmd', 'pcod', 'tcod'],
+    'aiida.tools.data.orbitals': ['orbital', 'realhydrogen'],
+    'aiida.parsers': ['arithmetic.add', 'templatereplacer.doubler'],
+    'aiida.schedulers': ['direct', 'lsf', 'pbspro', 'sge', 'slurm', 'torque'],
+    'aiida.transports': ['local', 'ssh'],
+    'aiida.workflows': ['arithmetic.multiply_add', 'arithmetic.add_multiply'],
 }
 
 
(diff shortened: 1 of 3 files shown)


aiida/plugins/entry_point.py
```diff
@@ -268,6 +270,60 @@ def get_entry_point(group: str, name: str) -> EntryPoint:
...
+    For `aiida-core==2.0` all existing entry points where properly prefixed with ``core.`` and the old entry points were
+    deprecated. To provide a smooth transition these deprecated entry points are detected in ``get_entry_point``, which
+    is the lowest function that tries to resolve an entry point string, by calling this function.
+
+    If the entry point corresponds to a deprecated one, a warning is raised and the new corresponding entry point name
+    is returned.
+
+    This method should be removed in ``aiida-core==3.0``
+    """
+    deprecated_entry_points_mapping = {
```
chrisjsewell, line 285: I would move this out of the function, so you don't have to recreate the dictionary on every function call. Probably a marginal gain, but don't want to slow down the CLI

ramirezfranciscof (commented):
Thanks! Looks good to me in general, my only question is if it makes sense to have two separate functions doing this. I understand that maybe this is somehow "inherited" of the previous mechanism in which `warn_deprecated_entry_point` was called by many different factories with different lists of possible entry-points. But now that this is only required in `get_entry_point`, having that intermediate `convert_potentially_deprecated_entry_point` that only selects and passes the correct list of entry-points feels like bit forced separation of concerns. Could these be made a single function?

aiida/plugins/entry_point.py
sphuber: Done

sphuber:
Thanks @chrisjsewell and @ramirezfranciscof , addressed both your comments.

aiida/plugins/entry_point.py
```diff
@@ -268,6 +285,37 @@ def get_entry_point(group: str, name: str) -> EntryPoint:
...
+    is the lowest function that tries to resolve an entry point string, by calling this function.
+
+    If the entry point corresponds to a deprecated one, a warning is raised and the new corresponding entry point name
+    is returned.
+
+    This method should be removed in ``aiida-core==3.0``
+    """
+    from warnings import warn
+
+    from aiida.common.warnings import AiidaDeprecationWarning
```
chrisjsewell, line 302: Is there a reason to not have these as top-level imports?

chrisjsewell (changes requested):
one more thing cheers

aiida/plugins/entry_point.py
sphuber: Because this function is the only place that uses them and when the function gets removed the obsolete imports are also automatically taken care over. Not a huge deal, but why not?

ramirezfranciscof (approved):
Thanks! Looks good to me! Feel free to merge after you finish your discussion with @chrisjsewell 

BTW this seems the best way to do multi-people review, right? I mean, each approves separately and it is understood the author would wait for approval from all before merging (this would be opposed to something like giving "approving" messages as comment but the last reviewer leaves theirs as "approve")

aiida/plugins/entry_point.py
chrisjsewell: Meh, pylint will flag unused imports, so that's not an issue.
Yeh not a big deal, but that's the python coding standard that we use (outside reasons like CLI response and circular imports), whereas "single use imports should be internal to functions" is not.
Up to you

chrisjsewell (approved):

Issue #5189, opened 2021-10-20
Add API to remove a process node from being used in caching


sphuber:
There is a method to do this already, essentially removing the `_aiida_hash` extra, but we should have a real method that is more intuitive and easy to find.

PR #5209, opened 2021-10-30
Fix `verdi calcjob cleanworkdir` command

zhubonan:
Previously the AND condition is not apply when -o and -p are both supplied, only the former is applied....
Also adds the ability to filter by `exit_status` of the calcjob.

### DIFF
--- a/aiida/cmdline/commands/cmd_calcjob.py
+++ b/aiida/cmdline/commands/cmd_calcjob.py
@@ -228,7 +228,8 @@ def calcjob_outputls(calcjob, path, color):
 @options.OLDER_THAN(default=None)
 @options.COMPUTERS(help='include only calcjobs that were ran on these computers')
 @options.FORCE()
-def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force):
+@options.EXIT_STATUS()
+def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force, exit_status):
     """
     Clean all content of all output remote folders of calcjobs.
 
@@ -237,8 +238,7 @@ def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force):
     modified AFTER [-p option] days from now, but BEFORE [-o option] days from now.
     """
     from aiida import orm
-    from aiida.orm.utils.loaders import ComputerEntityLoader, IdentifierType
-    from aiida.orm.utils.remote import clean_remote, get_calcjob_remote_paths
+    from aiida.orm.utils.remote import get_calcjob_remote_paths
 
     if calcjobs:
         if (past_days is not None and older_than is not None):
@@ -248,7 +248,14 @@ def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force):
             echo.echo_critical('if no explicit calcjobs are specified, at least one filtering option is required')
 
     calcjobs_pks = [calcjob.pk for calcjob in calcjobs]
-    path_mapping = get_calcjob_remote_paths(calcjobs_pks, past_days, older_than, computers)
+    path_mapping = get_calcjob_remote_paths(
+        calcjobs_pks,
+        past_days,
+        older_than,
+        computers,
+        exit_status=exit_status,
+        only_not_cleaned=True,
+    )
 
     if path_mapping is None:
         echo.echo_critical('no calcjobs found with the given criteria')
@@ -263,12 +270,12 @@ def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force):
     for computer_uuid, paths in path_mapping.items():
 
         counter = 0
-        computer = ComputerEntityLoader.load_entity(computer_uuid, identifier_type=IdentifierType.UUID)
+        computer = orm.load_computer(uuid=computer_uuid)
         transport = orm.AuthInfo.objects.get(dbcomputer_id=computer.id, aiidauser_id=user.id).get_transport()
 
         with transport:
-            for path in paths:
-                clean_remote(transport, path)
+            for remote_folder in paths:
+                remote_folder._clean(transport=transport)  # pylint:disable=protected-access
                 counter += 1
 
         echo.echo_success(f'{counter} remote folders cleaned on {computer.label}')
--- a/aiida/orm/nodes/data/remote/base.py
+++ b/aiida/orm/nodes/data/remote/base.py
@@ -24,6 +24,8 @@ class RemoteData(Data):
     Remember to pass a computer!
     """
 
+    KEY_EXTRA_CLEANED = 'cleaned'
+
     def __init__(self, remote_path=None, **kwargs):
         super().__init__(**kwargs)
         if remote_path is not None:
@@ -149,18 +151,32 @@ class RemoteData(Data):
                 else:
                     raise
 
-    def _clean(self):
-        """
-        Remove all content of the remote folder on the remote computer
+    def _clean(self, transport=None):
+        """Remove all content of the remote folder on the remote computer.
+
+        When the cleaning operation is successful, the extra with the key ``RemoteData.KEY_EXTRA_CLEANED`` is set.
+
+        :param transport: Provide an optional transport that is already open. If not provided, a transport will be
+            automatically opened, based on the current default user and the computer of this data node. Passing in the
+            transport can be used for efficiency if a great number of nodes need to be cleaned for the same computer.
+            Note that the user should take care that the correct transport is passed.
+        :raises ValueError: If the hostname of the provided transport does not match that of the node's computer.
         """
         from aiida.orm.utils.remote import clean_remote
 
-        authinfo = self.get_authinfo()
-        transport = authinfo.get_transport()
         remote_dir = self.get_remote_path()
 
-        with transport:
+        if transport is None:
+            with self.get_authinfo().get_transport() as transport:  # pylint: disable=redefined-argument-from-local
+                clean_remote(transport, remote_dir)
+        else:
+            if transport.hostname != self.computer.hostname:
+                raise ValueError(
+                    f'Transport hostname `{transport.hostname}` does not equal `{self.computer.hostname}` of {self}.'
+                )
             clean_remote(transport, remote_dir)
+
+        self.set_extra(self.KEY_EXTRA_CLEANED, True)
 
     def _validate(self):
         from aiida.common.exceptions import ValidationError
--- a/aiida/orm/utils/remote.py
+++ b/aiida/orm/utils/remote.py
@@ -10,6 +10,8 @@
 """Utilities for operations on files on remote computers."""
 import os
 
+from aiida.orm.nodes.data.remote.base import RemoteData
+
 
 def clean_remote(transport, path):
     """
@@ -37,7 +39,16 @@ def clean_remote(transport, path):
         pass
 
 
-def get_calcjob_remote_paths(pks=None, past_days=None, older_than=None, computers=None, user=None, backend=None):
+def get_calcjob_remote_paths(  # pylint: disable=too-many-locals
+    pks=None,
+    past_days=None,
+    older_than=None,
+    computers=None,
+    user=None,
+    backend=None,
+    exit_status=None,
+    only_not_cleaned=False,
+):
     """
     Return a mapping of computer uuids to a list of remote paths, for a given set of calcjobs. The set of
     calcjobs will be determined by a query with filters based on the pks, past_days, older_than,
@@ -48,7 +59,9 @@ def get_calcjob_remote_paths(pks=None, past_days=None, older_than=None, computer
     :param older_than: only include calcjobs older than
     :param computers: only include calcjobs that were ran on these computers
     :param user: only include calcjobs of this user
-    :return: mapping of computer uuid and list of remote paths, or None
+    :param exit_status: only select calcjob with this exit_status
+    :param only_not_cleaned: only include calcjobs whose workdir have not been cleaned
+    :return: mapping of computer uuid and list of remote folder
     """
     from datetime import timedelta
 
@@ -58,6 +71,7 @@ def get_calcjob_remote_paths(pks=None, past_days=None, older_than=None, computer
 
     filters_calc = {}
     filters_computer = {}
+    filters_remote = {}
 
     if user is None:
         user = orm.User.objects.get_default()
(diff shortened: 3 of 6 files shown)


zhubonan:
Also, since cleaning hundres of calculations can be very slow, is it worth storing a flag for those that has been cleaned in `extras`, so they can be skipped the next time the command issued?

aiida/orm/utils/remote.py
```diff
@@ -69,7 +71,16 @@ def get_calcjob_remote_paths(pks=None, past_days=None, older_than=None, computer
         filters_calc['mtime'] = {'>': timezone.now() - timedelta(days=past_days)}
 
     if older_than is not None:
-        filters_calc['mtime'] = {'<': timezone.now() - timedelta(days=older_than)}
+        older_filter = {'<': timezone.now() - timedelta(days=older_than)}
+        # Check if we need to apply the AND condition
+        if 'mtime' not in filters_calc:
+            filters_calc['mtime'] = older_filter
```
unkcpz, line 88: I think these two time filters need to be used separately, otherwise, how do you know we should use 'AND' or 'OR' when they both applied? 
I would suggest just adding a check and raising a CLI exception here said these two options can not be applied at the same time.

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -229,7 +230,10 @@ def calcjob_outputls(calcjob, path, color):
 @options.COMPUTERS(help='include only calcjobs that were ran on these computers')
 @options.FORCE()
 @options.EXIT_STATUS()
-def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force, exit_status):
+@click.option(
+    '--include-cleaned', is_flag=True, default=False, help='Included calcjobs that have been cleaned previously'
+)
+def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force, exit_status, include_cleaned):
```
unkcpz, line 236: Can we just set this internally rather than an option to set that always ignores the 'cleaned' calcjobs?  Do you see any downside of it?

unkcpz: I can see another advantage of adding an extra attribute to the calcjob that has been cleaned, is that it will be easy to turn off the caching for the process specifically.

unkcpz (changes requested):
Thanks a lot @zhubonan 
Sorry for the delay in reviewing it.

I have some minor questions about the implementation. The exit_status part looks good to me, but not pretty sure the `past` and `older` options should be applied at the same time for filtering.

unkcpz:
Hi @zhubonan, do you get a chance to have a look and finish this PR. If you need any help please let me know, I also require this feature to clear the remote folder.

aiida/orm/utils/remote.py
zhubonan: I think `and` is what makes most sense? Similar to other commandline tools like `find -mtime \+1 -mtime -10`.

aiida/cmdline/commands/cmd_calcjob.py
zhubonan: My worry was that for very large database trying to find the `cleaned` flag may be slow? Otherwise, I agree we should skip the cleaned ones automatically. 

> I can see another advantage of adding an extra attribute to the calcjob that has been cleaned, is that it will be easy to turn off the caching for the process specifically.

The flag should not affect caching at all, it is fine to cache a calcultion containing output remote folder that is cleaned. But runing a new calculation taking a cleaned remote folder as input would not make any sense.

zhubonan:
Hi @unkcpz sorry for leaving it for a while! Just replied to your comments. I think it is almost ready to merge now

aiida/orm/utils/remote.py
unkcpz: > Similar to other commandline tools like find -mtime \+1 -mtime -10

Cool! Good to know this. Then I would say I agree with using as `AND` because it is more system compatible.

aiida/cmdline/commands/cmd_calcjob.py
unkcpz: >  >I can see another advantage of adding an extra attribute to the calcjob that has been cleaned, is that it will be easy to turn off the caching for the process specifically.
>
>The flag should not affect caching at all, it is fine to cache a calcultion containing output remote folder that is cleaned. But runing a new calculation taking a cleaned remote folder as input would not make any sense.

I was to going to say since the clean workdir may have some downside influence on the caching mechenism, like in the workflow the previous Calcjob has a RemoteData output which used for the subsequent CalcJob in the workflow but the RemoteData workdir is cleaned. (For reference, check discussion https://github.com/aiidateam/aiida-core/issues/5178). Then with you implementation here you add a `workdir_cleaned` to the extra attributes, I then can check these attributes and control the caching (by the new feature implemented https://github.com/aiidateam/aiida-core/issues/5189) for this specific type of 'cleaned nodes' and turn off the caching for them.

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -228,7 +228,11 @@ def calcjob_outputls(calcjob, path, color):
 @options.OLDER_THAN(default=None)
 @options.COMPUTERS(help='include only calcjobs that were ran on these computers')
 @options.FORCE()
-def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force):
+@options.EXIT_STATUS()
+@click.option(
+    '--include-cleaned', is_flag=True, default=False, help='Included calcjobs that have been cleaned previously'
+)
```
unkcpz, line 234: I am still not sure we should complicate it here for the cases we have not encountered yet. I would prefer to add this option only when the performance issue hit. I let @ramirezfranciscof involved for some advice. 

If we agree this option is necessary then I think it better be `@options.INCLUDE_CLEANED` to compatible the code style we have.

unkcpz (changes requested):
Cheers! All tests passed after the outstanding fix on CI. I add a minor commit to pass the verdi performance test fail I hope you not mind @zhubonan. 

I am all good with the changes except the `--include-cleaned` option. pinning @ramirezfranciscof for the votes.

aiida/cmdline/commands/cmd_calcjob.py
zhubonan: To solve #5178, what you really need is have AiiDA to work out a sequence of non-cached calculations to run with a certain missing `RemoteData` end point, which would be quite complex to implement...
 
Suppose you have a workchain that runs fource calculations, *A*, *B*, *C* and *D* in a sequence, each using the `RemoteData` from the previous one.

*A* -> *RemoteA* -> *B* -> *RemoteB* -> *C* -> *RemoteC* -> *D*

If *D* crashes, restarting the workflow with caching on should fast-forward to running *D*, this is done by caching *A*, followed by *B*, then by *C* in a seqeuence. The remote folder of *A* and *B* may be cleaned and it should not affect having *C* cached. If the `RemoteFolder` of *C* is cleaned then obviously the restart of *D* should not proceed, but this is not related to caching mechanism. 

Instead, attempt to run *D* should just give an error, and one has to run the workchain without caching from scratch. However, this does not stop *A*, *B*, *C* being valid caches, as some calculations may not use `RemoteData` as inputs, and there is no ways to know how future workchain will use existing nodes. 

If we make a cleaned `RemoteFolder` invalidating the cache, then the chain would be breakon in the first step, which is a bit of overkill.

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -267,8 +279,9 @@ def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force):
         transport = orm.AuthInfo.objects.get(dbcomputer_id=computer.id, aiidauser_id=user.id).get_transport()
 
         with transport:
-            for path in paths:
+            for (calcjob_id, path) in paths:
                 clean_remote(transport, path)
                 counter += 1
+                load_node(calcjob_id).set_extra('workdir_cleaned', True)
```
sphuber, line 285: if you are going to load the node again, wouldn't it make more sense to have `get_calcjob_remote_paths` have the argument `include_calcjob` and return the node. Since you get it from the query, this will be a lot more efficient.

aiida/cmdline/commands/cmd_calcjob.py
sphuber: Defining an option as a reusable option is only really necessary if it is used more than once. The idea of having them in `options` module is to make it easy to reuse so similar options look similar across the CLI. If there is an option that is a one-off, it is fine to keep it defined here in place.

aiida/cmdline/commands/cmd_calcjob.py
sphuber, line 285: Would personally opt for just `cleaned`.
```suggestion
                load_node(calcjob_id).set_extra('cleaned', True)
```

sphuber, line 285: If we are going to make part of the way things work, we should really make this logic part of the `RemoteData` class, i.e., the `_clean` method should be responsible for setting this extra. Now the `clean_remote` function doesn't call this method because it doesn't allow specifying the transport making it very inefficient. But I would simply update `_clean` to be
```
def _clean(self, transport=None)
```
to allow passing a transport. This way you can use this method in the CLI, allowing to centralize the setting of the `cleaned` extra, and it is backwards compatible.

One problem we would have to keep in mind is really that we should maybe check in `_clean` that the passed `transport` is configured for the `Computer` that is the same as the computer of the `RemoteData`.

sphuber (commented):
Thanks @zhubonan . Realize that I am late to the party, but have some comments.

aiida/cmdline/commands/cmd_calcjob.py
zhubonan: good point. this also means we the `cleaned` should be stored in `extras` of the `RemoteData` not `CalcJobNode`. So I have updated the code to that instead.

aiida/orm/nodes/data/remote/base.py
```diff
@@ -149,18 +149,22 @@ def listdir_withattributes(self, path='.'):
...
-        transport = authinfo.get_transport()
         remote_dir = self.get_remote_path()
+        if transport is None:
+            authinfo = self.get_authinfo()
+            transport = authinfo.get_transport()
 
-        with transport:
+            with transport:
+                clean_remote(transport, remote_dir)
+        else:
```
sphuber, line 165: In this clause we should really check that the `transport` goes with the `Computer` of the `remote_dir`. There is no direct way of getting the `Computer` from `transport`, but I think we should be able to get the `hostname` and then we compare that with the `hostname` of the `Computer`. Would that be sufficient?

aiida/orm/utils/remote.py
```diff
@@ -84,8 +115,8 @@ def get_calcjob_remote_paths(pks=None, past_days=None, older_than=None, computer
 
     path_mapping = {}
 
-    for path, computer in query.all():
-        if path is not None:
-            path_mapping.setdefault(computer.uuid, []).append(path)
+    for calcjob_id, remote_data, computer in query.all():
+        path_mapping.setdefault(computer.uuid,
```
sphuber, line 119: Since you already fetch the `Computer` instance, I would return this in the `path_mapping`, especially since the CLI needs it and actually loads it again. This would be more efficient.

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -267,8 +276,8 @@ def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force):
         transport = orm.AuthInfo.objects.get(dbcomputer_id=computer.id, aiidauser_id=user.id).get_transport()
 
         with transport:
-            for path in paths:
-                clean_remote(transport, path)
+            for remote_folder in paths:
```
sphuber, line 279: If you fix `get_calcjob_remote_paths` to return the `Computer` instance, on line 275 you no longer need to load the `Computer` again, and you can simply get the transport with `with computer.get_transport(user):`

aiida/orm/utils/remote.py
```diff
@@ -69,13 +85,28 @@ def get_calcjob_remote_paths(pks=None, past_days=None, older_than=None, computer
...
+            filters_calc['mtime'] = {'and': [past_filter, older_filter]}
+
+    if exit_status is not None:
+        filters_calc['attributes.exit_status'] = exit_status
 
     if pks:
         filters_calc['id'] = {'in': pks}
 
+    if only_not_cleaned is True:
+        filters_remote['extras.cleaned'] = {'!==': True}
```
sphuber, line 103: Will this match also those that don't even have the extra? As in, should we add an explicit filter `{'!has_key': 'extras.cleaned'}` ? Not sure if this is being tested. Good to check

aiida/orm/nodes/data/remote/base.py
```diff
@@ -149,18 +149,22 @@ def listdir_withattributes(self, path='.'):
...
             clean_remote(transport, remote_dir)
+        self.set_extra('cleaned', True)
```
sphuber, line 167: Would make the `'cleaned'` constant a class attribute of `RemoteData`, e.g., `KEY_EXTRA_CLEANED = 'cleaned'`. Then you can use this attribute in `get_calcjob_remote_paths` to do
```
filters_remote[f'extras.{RemoteData.KEY_EXTRA_CLEANED}'] = {'!==': True}
```

sphuber (changes requested):
Thanks @zhubonan . I apologize for yet more requests, but there are a few more things I noticed. After this, I would be ok to merge this, promise :sweat_smile:

aiida/orm/utils/remote.py
zhubonan: I think is does match to those does not have the extra, but making it more explicity is better.

aiida/orm/nodes/data/remote/base.py
zhubonan: good point

aiida/orm/utils/remote.py
zhubonan: I think the reason for using this approach is that there are many folders, there will be each `Computer` node for each match - so lots of `Computer` to be loaded. In reality we expect only as few distinct computers among many matches, so project the ids then load the unique computer can be more efficeint?

zhubonan: or may be the query will just return a single `Computer` object but multiple times? I am not sure.

zhubonan: > It is a good point, but then we should also just project just the UUID as the query is currently projecting the entire entity.
ahh, I did not spot that. Just fixed it.

aiida/orm/nodes/data/remote/base.py
zhubonan: The hostname from the transport is stored as `Transport._machine`, but this is something defined by the plugins. I compare this to `Computer.hostname`. Hope this is OK?

aiida/cmdline/commands/cmd_calcjob.py
```diff
@@ -248,7 +248,15 @@ def calcjob_cleanworkdir(calcjobs, past_days, older_than, computers, force):
...
 
     calcjobs_pks = [calcjob.pk for calcjob in calcjobs]
-    path_mapping = get_calcjob_remote_paths(calcjobs_pks, past_days, older_than, computers)
+    path_mapping = get_calcjob_remote_paths(
+        calcjobs_pks,
+        past_days,
+        older_than,
+        computers,
+        exit_status=exit_status,
+        include_calcjob_id=False,
```
sphuber, line 257: You are adding this, but I don't think you actually use the `calcjob_id`? If it is not necessary, I would not add this argument now.

zhubonan: right yes, since now it just owrk with RemoteData. I have just removed it.

sphuber:
@zhubonan I realized that due to my requests, the scope of the PR had broadened considerably and so I thought it best to have separate commits for clarity. Since I didn't want to burden you with this additional work, I took the liberty to rebase it myself. In the process I also updated the tests for `RemoteData._clean` to check the new extra. Hope you don't mind. Let me know if you want to have a final look and are ok with the changes. If so, I will merge.

zhubonan:
@sphuber No problem at all. Please free feel to edit, rebase and merge.

sphuber (approved):

sphuber:
Thanks a lot @zhubonan !

PR #5210, opened 2021-11-01
CLI: deprecate `verdi config caching`

unkcpz:
I think it is replaced with `verdi config set/unset/list` command after 1.6.0 and should be deprecated.

### DIFF
--- a/aiida/cmdline/commands/cmd_config.py
+++ b/aiida/cmdline/commands/cmd_config.py
@@ -182,6 +182,8 @@ def verdi_config_caching(disabled):
     from aiida.manage.caching import get_use_cache
     from aiida.plugins.entry_point import ENTRY_POINT_STRING_SEPARATOR, get_entry_point_names
 
+    echo.echo_deprecated('This command is deprecated, please use set/unset to config the central `config.json` file.')
+
     for group in ['aiida.calculations', 'aiida.workflows']:
         for entry_point in get_entry_point_names(group):
             identifier = ENTRY_POINT_STRING_SEPARATOR.join([group, entry_point])


aiida/cmdline/commands/cmd_config.py
```diff
@@ -182,6 +182,8 @@ def verdi_config_caching(disabled):
     from aiida.manage.caching import get_use_cache
     from aiida.plugins.entry_point import ENTRY_POINT_STRING_SEPARATOR, get_entry_point_names
 
+    echo.echo_deprecated('This command is deprecated, please use set/unset to config the central `config.json` file.')
```
sphuber, line 185: I would use the dedicated decorator we have `aiida.cmdline.utils.decorators.deprecated_command`. And maybe for the text we could make the command more explicit and I would remove reference to the `config.json`. Also, this command was only used for listing, not for setting caching options, so I would remove that. So something like:
```
This command has been deprecated and will be removed. Please use `verdi config list caching` instead.
```

sphuber (changes requested):
Thanks @unkcpz . I think this command can indeed be deprecated in favor of `verdi config list caching`. Note that there are still references in the [documentation to `verdi config caching`](https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/run_codes.html?highlight=caching#class-level) which should be adapted.

unkcpz:
@sphuber Thanks! But I got a new feeling that this `verdi config caching` has its usage in expanding wildcard matching of the config setting into the list of plugins which show more apparently to the users.

What confuse me most is the `disable_for` and `enable_for` settings. Maybe we should only keep one list here? Do you have any good idea?

sphuber:
> But I got a new feeling that this `verdi config caching` has its usage in expanding wildcard matching of the config setting into the list of plugins which show more apparently to the users.

We should ask @chrisjsewell to be sure, but if that is the case, I would still try to merge this in the existing commands. I feel having separate ones is confusing concerning which command should be used for what purpose. It may not be possible though, to integrate this functionality in the generic `verdi config list/show` commands. This was my original concern when @chrisjsewell wanted to get rid of the dedicated `verdi config caching` command.

> What confuse me most is the `disable_for` and `enable_for` settings. Maybe we should only keep one list here? Do you have any good idea?

I am pretty sure that the idea is that it allows you to enable an entire namespace with wildcards and then just disable a few, instead of having to explicitly enable everything. For example:
```
enabled_for: aiida.calculations:quantumespresso.*
disabled_for: aiida.calculations:quantumespresso.pw
```

unkcpz:
> I am pretty sure that the idea is that it allows you to enable an entire namespace with wildcards and then just disable a few, instead of having to explicitly enable everything. For example:
>
> enabled_for: aiida.calculations:quantumespresso.*
> disabled_for: aiida.calculations:quantumespresso.pw

Thanks! This is a point that is not very clear to me. Is it means `disabled_for` will always take effect after `onabled_for` override its setting or it will only be the invert selection of the wildcard matching?

chrisjsewell:
It sounds to me like you haven't read https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/run_codes.html#how-to-run-codes-caching-configure 😉

sphuber:
> It sounds to me like you haven't read https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/run_codes.html#how-to-run-codes-caching-configure wink

This doesn't really answer whether `verdi config caching` _should_ be kept as a separate command or should be removed and its functionality integrated into one of the other existing `verdi config` commands.

chrisjsewell:
TL;DR no this command should not be deprecated; it is part of the new interface

chrisjsewell:
All that documentation and `verdi config cache` were added as part of the new interface: https://github.com/aiidateam/aiida-core/pull/4712/files

chrisjsewell:
> This was my original concern when @chrisjsewell wanted to get rid of the dedicated verdi config caching command.

I think you may be mistaken here, given that I was the one who added `verdi config caching`, in #4712 😬

unkcpz:
Okay~ All make sense. 

But it is then a subtle but necessary post-process to run `verdi config caching` every time after setting a new caching config. In order to check that there are no calculation plugins conflicting with each other.

chrisjsewell:
> But it is then a subtle but necessary post-process to run verdi config caching every time after setting a new caching config.

Certainly open to improvement suggestions 👍 But definitely I would read through #4712 first, to see the original thinking

sphuber:
> I think you may be mistaken here, given that I was the one who added `verdi config caching`, in #4712 grimacing
> 
> Before that, there was no CLI for caching

True, but there _was_ an open PR with an implementation of a dedicated command before your PR. My whole point for having a dedicated command and not integrate this with `verdi config` is exactly for the reason that I foresaw that the caching would require more complex functionality than simply setting and listing config options. I rest my case with the fact that there is now indeed an additional separate command with additional functionality just for caching and that this causes confusion. Anyway, not saying we should change this now, but it would be good if we could find a way to maybe make it clearer and more obvious what each part is for.

chrisjsewell:
> and that this causes confusion

I disagree

PR #5215, opened 2021-11-02
`verdi code setup`: validate the uniqueness of label for local codes

sphuber:
In commit d25339dd7e8f8bed40145392dd5166817f3afef5 `verdi code setup`
was improved to have a callback for the `label` option to check for its
uniqueness. However, it only implemented this for "remote" computers,
which have an associated `Computer` and so the uniqueness criterion is
on the "full label", which is the `label@computer.label`.

However, there are also "local" codes, which don't have an associated
computer and so only have the label as the identifier that should be
unique. The callback `validate_label_uniqueness` is now updated to
distinguish between these two cases.

### DIFF
--- a/aiida/cmdline/params/options/commands/code.py
+++ b/aiida/cmdline/params/options/commands/code.py
@@ -24,18 +24,34 @@ def is_not_on_computer(ctx):
 
 
 def validate_label_uniqueness(ctx, _, value):
-    """Validate the uniqueness of the full label of the code, i.e., `label@computer.label`.
+    """Validate the uniqueness of the label of the code.
 
-    .. note:: For this to work, the computer parameter already needs to have been parsed. In interactive mode, this
-        means that the computer parameter needs to be defined after the label parameter in the command definition. For
-        non-interactive mode, the parsing order will always be determined by the order the parameters are specified by
-        the caller and so this validator may get called before the computer is parsed. For that reason, this validator
-        should also be called in the command itself, to ensure it has both the label and computer parameter available.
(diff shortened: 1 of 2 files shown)


ramirezfranciscof (changes requested):
Looks good to me! The only thing I would add is some tests monkeypatching `load_code` to raise `exceptions.MultipleObjectsError` just for coverage completeness. Also, more up to you, but I think it is better, since we already know it is a `MultipleObjectsError` exception, to mention this in the message (``f'multiple copies of the code `{value}` already exist.'``).

ramirezfranciscof:
One side comment that I've just notice is that I think we are not catching "@" in the user input, no? Bah, maybe I'm missing something. Anyways this is separate and can be done in a different PR, just mentioning here for relevance.

@JPchico did you want to take a look at this before we merge? We could probably wait some days if you wanted to check this but don't have time right now, otherwise once my comments are addressed by @sphuber we would probably merge this by the end of the week. Let us know.

ramirezfranciscof:
@janssenhenning  did you want to take a look at this before we merge? We could probably wait some days if you wanted to check this but don't have time right now, otherwise once my comments are addressed by @sphuber we would probably merge this by the end of the week. Let us know.

sphuber:
@ramirezfranciscof thanks for the review. Updated the exception message and added tests.

janssenhenning:
@ramirezfranciscof Sorry I did not have time so far. The fix looks fine to me

ramirezfranciscof (approved):
Great, thanks @janssenhenning and @sphuber , this can go right in!

PR #5218, opened 2021-11-04
Add support for `Enum` types to `aiida.orm.utils.serialize`

### LINKED ISSUE #5217: Add support for `Enum` instances in the `aiida.orm.utils.serialize` YAML serializers
Enum's are often used data types that currently cannot be used in process inputs, even if the ports are `non_db`, because even the `non_db` inputs are serialized as part of the entire `Process` instance that is serialized when a checkpoint is created. Creating a generic serializer and deserializer should be straightforward as a fully qualified definition should consist of its full class identifier and the value.

sphuber:
Fix #5217 

Enum's are often used data types that currently cannot be used in
process inputs, even if the ports are `non_db`, because even the
`non_db` inputs are serialized as part of the entire `Process` instance
that is serialized when a checkpoint is created.

The custom YAML serializer and deserializer that are defined in the
`aiida.orm.utils.serialize` module that are used for serializing process
instances and their inputs, are updated with a dumper and a loader for
`Enum` instances. The full class identifier is generated from the
default object loader provided by `plumpy` and concatenated by the `|`
sign and the enum's values, it is represented as a YAML scalar.

The loader uses the same object loader to load the enum class from the
identifier string and reconstruct the enum instance from the serialized
value.

The original tests were written in `tests/common/test_serialize.py` and
are moved to `tests/orm/utils/test_serialize.py` to conform with the
standard to mirror the package hierarchy. The tests are converted to
`pytest` style and a new test is added by performing a round trip of an
enum instance (de)serialization.


sphuber:
@giovannipizzi @bosonie this should provide what we need as discussed. @chrisjsewell @ramirezfranciscof feel free to review if you'd like.

giovannipizzi (approved):
Ok with me thanks! Just to double check two things:
- this means that one needs to be able to import the Enum python declaration, right? (I think it's OK
- This does not yet allow to store this in the DB as one of the inputs when `non_db=False`? If so, would it be possible to store the string value in the DB in this case, or there are issues with this?

sphuber:
> this means that one needs to be able to import the Enum python declaration, right?

Yes, otherwise the deserialization will fail. I think there is no way around this and is an acceptable limitation.

> This does not yet allow to store this in the DB as one of the inputs when non_db=False? If so, would it be possible to store the string value in the DB in this case, or there are issues with this?

It does indeed not, but in general we do not allow to store arbitrary inputs in the database. All inputs have to be nodes so they can be properly linked up. The only exception are the `metadata.options` inputs of the `CalcJob` class which are stored as attributes.

chrisjsewell (approved):

giovannipizzi:
Wouldn't it be useful to also have a way (as an alternative) to store an input as a string (Str), but with automatic validation coming from the values of an Enum? This would be useful to store the inputs of the common workflows, for instance (e.g. the ElectronicType)

sphuber:
Might be useful indeed. Would be multiple ways to go about it. Maybe most straightforward would be to implement the `Enum` data plugin which automatically (de)serializes into the value and identifier. Opened #5224 and will look at implementing this a bit later.

